PR #23, opened 2026-04-27
fix(bulk): per-attempt idempotency + 207 Multi-Status (CAURA-602)

Eldad-Caura:
## Summary

Eliminates the **silent-create-bulk** class surfaced on loadtest-1777301515 (+1613 unacked rows on tenant A, +300 on tenant B). The bulk path is now atomic at the row level: a retry of the same logical batch can never produce a row the client doesn't see.

## What changes

**Wire contract (breaking):** `/api/v1/memories/bulk` requires the new `X-Bulk-Attempt-Id` header. Per-item results carry `client_request_id` and one of four statuses: `created`, `duplicate_attempt` (a previous call's row from the same attempt), `duplicate_content` (different attempt, same content_hash), or `error`. Status codes: `200` clean, `207` mixed, `422` all-error, `504` bulk-budget burn.

**Schema:** new `memories.client_request_id TEXT` column + partial unique index `ix_memories_attempt_unique ON (tenant_id, fleet_id, client_request_id) WHERE deleted_at IS NULL AND client_request_id IS NOT NULL`. `ALTER TABLE` is metadata-only on PostgreSQL and `CREATE INDEX CONCURRENTLY` follows the pattern from migration 005.

**Storage:** `memory_add_all` switches to `ON CONFLICT DO NOTHING RETURNING ...` plus a same-session SELECT for previously-committed rows; returns `{client_request_id, id, was_inserted}` per input item.

**core-api:** `create_memories_bulk` derives `f"{attempt_id}:{idx}"` per row, maps storage results by `client_request_id` (input order isn't guaranteed by `RETURNING ON CONFLICT`), and the legacy `httpx.ReadTimeout` reconcile branch is removed — per-row conflict resolution subsumes it.

**Budget:** bulk opts out of the 45s `RequestTimeoutMiddleware` (cancelling mid-commit was the proximate cause of the regression) and enforces its own `bulk_request_timeout_seconds = 90.0` via `asyncio.wait_for`.

**Auto-chunk + MCP:** in-process callers mint `f"auto-chunk:{uuid4()}"` / `f"mcp:{uuid4()}"` so the required-attempt-id contract holds everywhere.

## Why each part is needed

The receipt-level `Idempotency-Key` cache (PR #32) handles the fast path — server kept the response. The new per-attempt unique constraint handles the slow path — server crashed/timed-out before recording the receipt but the storage commit landed. Both layers compose.

## Test plan

- [ ] CI green on the new bulk-atomicity tests (per-attempt retry → all `duplicate_attempt`; different attempt id same content → `duplicate_content`; missing/malformed header → 400; budget-burn → 504)
- [ ] CI green on storage-writer integration test for `memory_add_all` per-attempt idempotency + post-conflict id resolution
- [ ] CI green on existing `test_idempotency.py` replay path with both headers present
- [ ] Re-run staging loadtest after merge; assert `silent-create-bulk` delta is 0 on both tenants

### DIFF
--- a/common/models/memory.py
+++ b/common/models/memory.py
@@ -2,7 +2,7 @@ import uuid
 from datetime import datetime
 
 from pgvector.sqlalchemy import Vector
-from sqlalchemy import DateTime, Float, ForeignKey, Index, Integer, Text, text
+from sqlalchemy import DateTime, Float, ForeignKey, Index, Integer, Text, func, text
 from sqlalchemy.dialects.postgresql import JSONB, TSVECTOR, UUID as PG_UUID
 from sqlalchemy.orm import Mapped, mapped_column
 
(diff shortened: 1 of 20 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@claude

Eldad-Caura:
@claude

Eldad-Caura:
@claude

Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #24, opened 2026-04-27
fix(noisy-neighbor): per-tenant storage-roundtrip bulkhead (draft, gated on #23 loadtest)

Eldad-Caura:
## Status

**Draft. Do not merge until post-#23 loadtest validates.**

The CAURA-602 silent-create fix in #23 may already absorb most of the noisy-neighbor symptom — last loadtest's +1613 silent commits inflated tenant A's effective load on the storage pool through the retry-storm path. After #23 lands and we re-run the loadtest, this PR is justified iff:

- [ ] **tenant B write p95-under-storm ≥ 1.5× baseline** (current: 2.37×; target: ≤ 1.5×)
- [ ] **tenant B writes still show storm-window failures** (current: 4 storm vs 0 baseline)

If both hold, promote out of draft. Otherwise close or downgrade.

## Summary

Adds a deeper-than-route per-tenant `asyncio.Semaphore` that wraps the storage call itself, complementing PR #33's route-entry slot. Two layers compose:

| Layer | Scope | Acquire | Held while |
|---|---|---|---|
| Route-entry (PR #33) | `"write"` / `"search"` | 50ms timeout → 429 | whole request lifecycle |
| **Storage-call (this PR)** | `"storage_write"` / `"storage_search"` | unbounded queue | only across `sc.<call>` |

Caps default to `per_tenant_storage_write_concurrency=2` and `per_tenant_storage_search_concurrency=4` — half the route-entry caps, since each request now holds the storage slot only during the fast storage call (~500ms–3s), not the slower embed/enrich. With 11 staging instances that's ~22 fleet-wide concurrent storage writes per tenant, comfortably below the ~110 storage-writer pool slots, so a single hot tenant can't park more than ~20% of pool capacity while a cold tenant's write queues.

## Wiring

Three primary storage call sites in `services/memory_service.py`:
- single-write hot path (`sc.create_memory`)
- bulk-write hot path (`sc.create_memories` — note: will conflict with #23's `create_memories_bulk` rewrite; cleanly resolvable on rebase)
- scored-search hot path (`sc.scored_search`)

Auto-chunk and other internal cascades intentionally **NOT** wrapped — they're rare and the additional slot acquire could compound internal-cascade latency. Will be revisited if loadtest data shows they're worth bulkheading too.

## Test plan

- [x] Unit tests cover unbounded queueing under saturation, tenant isolation under storm, exception-path slot release (`tests/test_per_tenant_concurrency.py`)
[...]

PR #25, opened 2026-04-27
fix(storage): align core-storage-api DB pool defaults with platform-storage (5+5)

Eldad-Caura:
## Summary

Lowers `db_pool_size` / `db_max_overflow` source defaults in `core-storage-api/config.py` from 20+20 to 5+5, matching `platform-storage-api` after PR #166. Staging is already running 5+5 via env-var overrides; the source-level mismatch was a fresh-deploy footgun — any new operator (or a clean redeploy) inherits the 20+20 source default and silently over-allocates against AlloyDB's connection ceiling.

Loadtest-1777301515's storm window surfaced this as `asyncpg.TooManyConnectionsError` on the storage-writer. The CAURA-602 silent-create work in #23 closes the symptom path (committed-but-unacked rows on retry); this PR closes the root by sizing the pool correctly out of the box.

## Test plan

- [x] No code/behavior change beyond the default values; existing tests are unaffected.
- [ ] After merge + deploy: drop the `DB_POOL_SIZE=5` / `DB_MAX_OVERFLOW=5` env-var overrides on `staging-memclaw-core-storage-writer` and `staging-memclaw-core-storage-reader` (per the prod-rollout runbook §7e). The new source default makes them no-ops.
- [ ] After staging confirmation: same env-var cleanup on the prod writer + reader during Phase B rollout.

## Notes

Operators that need a larger pool still set `DB_POOL_SIZE` / `DB_MAX_OVERFLOW` explicitly — only the source default changes.

Pairs with [#24](https://github.com/caura-ai/caura-memclaw/pull/24) (per-tenant storage bulkhead, draft) which approaches the same noisy-neighbor surface from the application layer; this PR addresses the infrastructure-default layer. Independent — both can land in either order.

### DIFF
--- a/core-storage-api/src/core_storage_api/config.py
+++ b/core-storage-api/src/core_storage_api/config.py
@@ -22,8 +22,22 @@ class Settings(BaseSettings):
     # we route.
     database_url: str = "postgresql+asyncpg://memclaw:changeme@localhost:5432/memclaw"
     read_database_url: str = ""
-    db_pool_size: int = 20
-    db_max_overflow: int = 20
+    # 5+5 matches the post-PR-#166 ``platform-storage-api`` defaults so
+    # both services share the same pool sizing without per-environment
+    # env-var overrides. Pre-this-fix the source defaults were 20+20,
+    # which staging worked around with ``--remove-env-vars`` /
+    # explicit env-var overrides on the writer + reader Cloud Run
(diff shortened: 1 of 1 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #26, opened 2026-04-27
chore(tier-2): hygiene batch — agent_add ON CONFLICT, patch-resurrects fix, README

Eldad-Caura:
## Summary

Three small Tier-2 cleanups bundled per request:

1. **`fix(storage): replace agent_add try/IntegrityError/rollback with ON CONFLICT`** — `agent_add` used a mid-session `rollback()` inside the IntegrityError handler, which aborts every other pending write in the session and forced [`test_concurrent_same_key_returns_same_id`](https://github.com/caura-ai/caura-memclaw/blob/main/tests/test_idempotency.py#L360) to pre-create the agent row via a warm-up write to dodge concurrent calls. Switched to `INSERT ... ON CONFLICT (tenant_id, agent_id) DO NOTHING RETURNING id` plus a same-session re-SELECT — same shape `memory_add_all` uses post-#23. Drops the warm-up workaround so the test drives the race directly.

2. **`fix(storage): block PATCH on soft-deleted rows`** — adds `WHERE deleted_at IS NULL` to both UPDATE branches in `memory_update` (column-set + metadata-merge). Pre-fix, a PATCH that landed after a soft DELETE silently set columns on the deleted row, leaving inconsistent state like `status='active', deleted_at=<ts>`. Loadtest-1777301515 + 1777327400 both *held* the `patch-resurrects-deleted` probe by timing luck; the gap was always there. New regression test `test_patch_after_delete_does_not_resurrect` drives both branches and asserts the row stays deleted.

3. **`docs(readme): document X-Bulk-Attempt-Id requirement on /memories/bulk`** — the breaking change shipped in #23 wasn't reflected in the README's endpoint table; SDK authors and integrators couldn't see the new contract without reading the route docstring or changelog.

## Out of scope (deferred)

- **Rate-limit test flake** ([Ticket 10](https://caura-ai.atlassian.net/browse/CAURA-513)): tried `asyncio.gather`-ing the burst, but the slowapi rate decorator binds the limit string at module-import time so per-test overrides don't take. A robust fix needs test-infra changes (env-var override before app import, or moving slowapi to a runtime-configurable limit). Left for its own PR.

## Test plan

- [x] `pytest tests/test_per_tenant_concurrency.py tests/test_request_timeout.py tests/test_bulk_write.py -m unit` — passes locally
- [ ] CI: full integration suite, including the two new tests (`test_concurrent_same_key_returns_same_id` no-warmup + `test_patch_after_delete_does_not_resurrect`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -396,7 +396,7 @@ All routes are versioned under `/api/v1/`. Interactive Swagger docs at `/api/doc
 | Endpoint | Method | Description |
 |---|---|---|
 | `/memories` | POST | Write a memory. LLM enrichment + embedding + entity extraction + contradiction detection. `"persist": false` for extract-only preview |
-| `/memories/bulk` | POST | Write up to 100 memories. Batches embeddings, parallelizes enrichment, single transaction |
+| `/memories/bulk` | POST | Write up to 100 memories. Batches embeddings, parallelizes enrichment, single transaction. Requires `X-Bulk-Attempt-Id` header (per-attempt idempotency); a retry with the same id resolves committed rows as `duplicate_attempt` instead of duplicating. Returns 200 (clean / all-error) or 207 Multi-Status (mixed) — read per-item `status` |
 | `/memories` | GET | List memories (filter by type, status, agent; paginate) |
 | `/memories/{id}` | GET | Full memory detail (embedding stats, entity links, RDF triple, temporal bounds) |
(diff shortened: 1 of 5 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #29, opened 2026-04-28
fix(search): drop UUID/hex fragments before entity FTS roundtrip

Eldad-Caura:
## Summary

- Extracts a shared `extract_entity_tokens(query)` helper that splits on whitespace **and** internal punctuation, then drops short tokens, stopwords, and hex-only fragments (UUIDs / content hashes / oid-style IDs).
- Uses the helper from **all three** entity-FTS call sites — pre-fix each rolled its own `query.split()` and only one (or none) would have been hardened by a localised patch.
- Drive-by: hoists `from sqlalchemy import update` out of `bulk_delete_by_ids` (`[Top-level imports only]`).

Closes the `empty-query-nonempty` finding in [report-loadtest-1777332393](../loadtest/report-loadtest-1777332393.md): the gibberish probe `xyzzy-{uuid_hex}-jibbet` was returning 1 hit on the larger-graph tenant because the unbroken token reached storage-side FTS, which re-tokenised on punctuation and matched UUID chunks against indexed entities.

## Refactored sites

- `core_api.pipeline.steps.search.classify_query.ClassifyQuery`
- `core_api.pipeline.steps.search.parallel_embed_entity_boost._entity_boost_via_storage`
- `core_api.services.memory_service._entity_boost_pipeline`

## Test plan

- [x] `tests/test_entity_tokens.py` — helper unit tests (7 cases)
- [x] `tests/pipeline/test_classify_query.py` — 3 new behavioural tests (loadtest probe, hyphenated phrase, pure-UUID query) + all 22 pre-existing tests still pass
- [x] `ruff check`, `ruff format --check`, pre-commit all green
- [ ] Re-run staging loadtest after merge — `empty-query-nonempty` finding should be absent for both tenants.

## Out of scope (separate work)

- Item F (move single-write enrichment off the hot path) is **already implemented in code** (`config.py:76` `enrich_on_hot_path` flag + `ScheduleBackgroundTasks` publishing `Topics.Memory.ENRICH_REQUESTED`). Staging is just running the OSS default `True`. Will be flipped via `gcloud run services update` separately, after verifying `core-worker` consumes the topic.

### DIFF
--- a/core-api/src/core_api/pipeline/steps/search/classify_query.py
+++ b/core-api/src/core_api/pipeline/steps/search/classify_query.py
@@ -12,14 +12,11 @@ from __future__ import annotations
 import asyncio
 import logging
 import re
-import string
 import types
 from uuid import UUID
 
 from core_api.clients.storage_client import get_storage_client
 from core_api.constants import (
-    ENTITY_STOPWORDS,
-    ENTITY_TOKEN_MIN_LENGTH,
     FTS_WEIGHT_BOOSTED,
     GRAPH_HOP_BOOST,
     GRAPH_MAX_BOOSTED_MEMORIES,
@@ -33,6 +30,7 @@ from core_api.pipeline.steps.search.retrieval_types import (
     RetrievalStrategy,
 )
 from core_api.schemas import EntityLinkOut
+from core_api.services.entity_tokens import extract_entity_tokens
 
 _GRAPH_HOP_BOOST_FALLBACK = GRAPH_HOP_BOOST[max(GRAPH_HOP_BOOST)]
 
@@ -70,14 +68,7 @@ class ClassifyQuery:
         graph_max_hops: int = sp["graph_max_hops"]
         top_k: int = sp["top_k"]
 
-        # Tokenize, strip punctuation, and filter to meaningful tokens.
-        tokens = [
-            stripped
-            for t in query.split()
-            if (stripped := t.strip(string.punctuation))
-            and len(stripped) >= ENTITY_TOKEN_MIN_LENGTH
-            and stripped.lower() not in ENTITY_STOPWORDS
-        ]
+        tokens = extract_entity_tokens(query)
 
         if tokens:
             try:
--- a/core-api/src/core_api/pipeline/steps/search/parallel_embed_entity_boost.py
+++ b/core-api/src/core_api/pipeline/steps/search/parallel_embed_entity_boost.py
@@ -14,14 +14,13 @@ from fastapi import HTTPException
 
 from core_api.clients.storage_client import get_storage_client
 from core_api.constants import (
-    ENTITY_STOPWORDS,
-    ENTITY_TOKEN_MIN_LENGTH,
     GRAPH_HOP_BOOST,
     GRAPH_MAX_BOOSTED_MEMORIES,
     RECALL_BOOST_CAP,
 )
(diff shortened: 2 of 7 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
Addressed in force-push (`9ee8b04` → `b64d4ca`):

- Added `_HEX_ID_MIN_LENGTH = 8` (the shortest UUID segment) so the all-hex test no longer clobbers real English words like `cafe`, `face`, `dead`, `beef`, `decade`, `facade`.
- Pulled the compound clause into a named `_is_hex_id(s)` helper for readability — the docstring now carries the rationale that was inline.
- Added two tests: `test_short_hex_english_words_kept` (the bug you flagged) + `test_hyphenated_uuid_partially_filtered` (documents the trade-off — 4-char UUID segments now survive the threshold; they reach FTS but won't match a real entity name).
- Updated the corresponding integration test in `tests/pipeline/test_classify_query.py` to reflect the new behavior.

Trade-off worth flagging: a pasted UUID like `550e8400-e29b-41d4-a716-446655440000` now sends `["e29b", "41d4", "a716"]` to entity FTS instead of skipping the call. Those 4-char hex tokens won't match real entity names in practice, so the call is a wasted roundtrip but not a correctness issue. If someone wants stricter (entity-search-by-id is unusual), we can split the threshold per call site. Happy to do that in a follow-up if preferred.

Eldad-Caura:
@claude

Eldad-Caura:
Bot review addressed in force-push (`b64d4ca` → `9d02515`):

**Medium** — case-mismatch fixed: `extract_entity_tokens` now lowercases each surviving token. The two call sites that did `.lower()` themselves drop the wrapper (`parallel_embed_entity_boost.py:51`, `memory_service.py:2603`), and `classify_query.py` now gets pre-normalised tokens too. All three FTS roundtrips agree on case.

**Low** — `_is_hex_id` docstring corrected to state the real rationale (8 sits above the longest common all-hex English word ~6 chars, not "matches the shortest UUID segment" — the 4-char UUID segments are deliberately let through).

Tests: 34 pass. New `test_tokens_are_lowercased` is the regression guard for the centralisation. Updated `test_uuid_laden_gibberish_drops_hex_fragments` (`FLIBBERTYJIBBET` → `flibbertyjibbet`) and `test_internal_punctuation_splits_tokens` (`USA` → `usa`) to match the new contract.

Eldad-Caura:
@claude

Eldad-Caura:
Bot review #2 addressed in force-push (`9d02515` → `24e2f52`):

**Medium** — `_` removed from `_TOKEN_SEP_RE`. UUIDs use hyphens (the loadtest probe pattern), and underscore-splitting silently broke snake_case entity names (`project_alpha`, `api_key`) by turning them into sub-tokens that wouldn't match an indexed entity. Comment in the source documents the omission so a future reader doesn't add it back.

**Low** — walrus refactor on the comprehension dedupes the double `stripped.lower()` call (once for the stopword check, once for emit) into a single `(lower := stripped.lower())`.

Tests: 35 pass. New `test_snake_case_identifiers_kept_intact` is the regression guard for the underscore-exclusion.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #31, opened 2026-04-28
fix(memories): wire GET /memories/:id/contradictions to detector results (CAURA-604)

Eldad-Caura:
## Summary

Loadtest reports (`report-loadtest-1777332393`, `1777364035`, `1777371451`, `1777375004`) flagged `contradictions-not-detected`: two seeded directly-conflicting facts about the same entity got neither flagged through the API. Root cause was the GET endpoint, not the detector.

`GET /memories/{id}/contradictions` was only returning the `supersedes_id` chain — what callers think of as "the detector's findings" was nowhere in the response. The detector (`services/contradiction_detector.py`) runs fire-and-forget post-commit on writes and persists its output as side-effects on the `memories` table:

- Loser row: `status` set to `outdated` (RDF path) or `conflicted` (semantic path).
- Winner row: `supersedes_id` set to the loser's id.

There is no separate contradictions table — the chain + status fields ARE the persisted detector output. So this PR re-shapes the GET response to surface that information explicitly while keeping the existing fields intact.

## What changed

`core-api/src/core_api/routes/memories.py` — additive fields on the response:

- `contradictions`: list of `{memory_id, status, reason, content_preview, direction, created_at}`.
  - `direction = "superseded_by"` for newer rows that replaced this one, `"supersedes"` for the older row this one replaced.
  - `reason` is inferred from the conflicting row's status (`outdated` -> `rdf_conflict`, `conflicted` -> `semantic_conflict`).
- `detection_status`: `"completed"` if any contradiction evidence exists for this memory (own status is outdated/conflicted, has `supersedes_id`, or another row points at it), else `"pending"`.

### Pending vs synchronous re-detect

The task left the choice between returning `pending` and running a synchronous detection. We picked `pending` — the docstring documents the trade-off. A GET should not have write side-effects, and the synchronous detector would mutate other rows' `status` and this row's `supersedes_id`. `pending` honestly reflects "no contradictions have been recorded yet" without distinguishing "ran and found nothing" from "didn't run" (the API can't tell those apart from the data alone, and that's fine for the loadtest's needs).

### Back-compat

- `memory_id`, `status`, `superseded_by`, `superseded_memories` are unchanged.
- All existing tests in `tests/test_api_memories.py` still pass.

## Test plan

- [x] `tests/test_api_contradictions.py::test_contradictions_pending_when_no_findings` — fresh memory returns `pending` + empty list.
- [x] `tests/test_api_contradictions.py::test_contradictions_completed_with_supersedes_chain` — after simulating detector side-effects, both directions of the chain are surfaced and `detection_status == "completed"`.
- [x] `uv run ruff check` + `uv run ruff format --check` clean on changed files.
- [x] `uv run mypy` clean on `routes/memories.py`.
- [x] Existing contradiction unit tests (`test_p1_contradiction`, `test_contradiction_batch`, `test_contradiction_providers`, `test_visibility_contradiction`) all pass.
- [ ] Next loadtest run should show `contradictions-not-detected` resolved (callers can read `contradictions` / `detection_status` instead of inferring from the supersedes-chain shape).

### DIFF
--- a/core-api/src/core_api/routes/memories.py
+++ b/core-api/src/core_api/routes/memories.py
@@ -588,14 +588,49 @@ async def get_contradictions(
     auth: AuthContext = Depends(get_auth_context),
     db: AsyncSession = Depends(get_db),
 ):
-    """Return memories that this memory superseded (marked outdated/conflicted)."""
+    """Return contradiction detector findings for this memory.
+
+    The contradiction detector (services/contradiction_detector.py) runs
+    fire-and-forget post-commit on writes and persists its findings via
+    two side-effects on the memories table:
+
+    1. Conflicted/outdated rows get ``status`` set to ``"conflicted"`` or
+       ``"outdated"``.
+    2. The newer (winning) row's ``supersedes_id`` is set to the older
+       (losing) row's id.
+
+    There is no separate ``contradictions`` table — the supersedes chain
+    + status fields ARE the persisted detector output (CAURA-604).
+
+    Response fields (``superseded_by``, ``superseded_memories``) are kept
+    intact for back-compat. New fields added in CAURA-604:
+
+    - ``detection_status``: ``"completed"`` if any contradiction evidence
+      exists for this memory (it has ``supersedes_id`` set, or another
+      row supersedes it, or its own status is outdated/conflicted),
+      otherwise ``"pending"``. ``"pending"`` means "no contradictions
+      have been recorded" — it covers both "detector ran and found
+      nothing" and "detector hasn't run yet"; the API has no way to
+      distinguish those without re-running detection synchronously,
+      which would side-effect from a GET. We chose the cheap, no-side-
+      effect path (return pending) over an inline re-run.
+    - ``contradictions``: detector-shaped findings derived from the
+      persisted chain (id, status, reason, content_preview, direction,
+      created_at). ``reason`` is inferred from the status of the
+      conflicting row: ``outdated`` -> ``rdf_conflict``,
+      ``conflicted`` -> ``semantic_conflict``. ``direction`` is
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
Bot review addressed in force-push:

**Medium** — `detection_status="completed"` no longer fires when `supersedes_id` is set but `older` is soft-deleted or excluded by tenant guard. The third arm now requires `older is not None and older.deleted_at is None`, so empty-`contradictions` is always paired with `detection_status="pending"`.

**Low** — added the tenant-isolation guard right after the `db.get(Memory, supersedes_id)` lookup. Cross-tenant `supersedes_id` (data corruption / manual migration) is now treated as absent instead of leaking content into `superseded_by` / `contradictions`. The pre-existing same-gap is fixed by extension.

Tests: 16/16 pass. ruff + format clean.

Eldad-Caura:
@claude

Eldad-Caura:
Round 2 bot review addressed:

**Medium** — collapsed `has_evidence` to `bool(contradictions)`. The earlier three-arm expression had the same `{completed, []}` ambiguity for the `memory.status` arm that round 1 closed for the `supersedes_id` arm. The third arm was also dead code (when its condition was True, the `older` branch above already appended to `contradictions`). One arm, no ambiguity.

**Low** — `reason` for `superseded_by` entries now falls back to `_reason_for(m.status)` when `memory.status` produces `"unknown"`, covering the async-lag window where the supersessor row exists but `memory.status` hasn't been updated yet. Documented inline.

**Suggestion** (third condition is dead code) — addressed by the same `bool(contradictions)` collapse.

Tests pass; ruff + format clean.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #32, opened 2026-04-28
fix(logging): route uvicorn / fastmcp / mcp / slowapi to root formatter (CAURA-588)

Eldad-Caura:
## Summary

Add `_route_third_party_to_root()` (called from `configure_logging`) that strips the library-installed handlers from `uvicorn`, `uvicorn.access`, `uvicorn.error`, `fastmcp`, `mcp`, and `slowapi`, and sets `propagate=True` so their lines flow through the root `ProcessorFormatter`.

## Why

Without this, those loggers ship with `propagate=False` and their own stream handler — so their access / error lines bypass the GCP-severity-aware JSON formatter that `configure_logging` installs on the root logger. They land in Cloud Logging as plain text at DEFAULT severity, breaking severity filtering and audit-log searches by severity.

The existing comment in `common/structlog_config.py` already flagged this as a known limitation ("Uvicorn installs its own handlers with propagate=False on `uvicorn` and `uvicorn.access`, so its lines bypass this handler — worth a follow-up").

## What's not changed

- The existing `_silence_noisy_dependency_loggers` block (httpx / google.auth / google.auth.transport at WARNING when min_level is INFO) — that's about level filtering, separate concern.
- No log-volume change. Same lines still emitted; they now carry GCP severity labels.

## Test plan

- [x] 2 new tests in `tests/test_logging.py`: handler-cleared + idempotency
- [x] All 12 tests pass; ruff + format clean

## Refs

- Epic: [CAURA-513](https://caura-ai.atlassian.net/browse/CAURA-513)
- Ticket: [CAURA-588](https://caura-ai.atlassian.net/browse/CAURA-588)

[CAURA-513]: https://caura-ai.atlassian.net/browse/CAURA-513?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/common/structlog_config.py
+++ b/common/structlog_config.py
@@ -214,6 +214,112 @@ def configure_logging(
         _configured = True
 
 
+# Names of third-party loggers that install their own (typically stderr,
+# plain-text) handlers OR set ``propagate=False``, bypassing the root
+# ``ProcessorFormatter`` without explicit re-routing. CAURA-588.
+#
+# Ordering caveat: ``_route_third_party_to_root`` only re-routes loggers
+# whose libraries have already initialised them at call time. uvicorn's
+# ``Config.configure_logging()`` runs at server start; if our
+# ``configure_logging()`` is called at module import (the typical app.py
+# path), the uvicorn loggers may not yet have their handlers when we
+# iterate. The function logs a debug line in that case so the caller can
+# move the call into an ASGI lifespan startup handler if production logs
+# show un-rerouted output.
+_THIRD_PARTY_LOGGERS_TO_REROUTE: tuple[str, ...] = (
+    # uvicorn ships three loggers with propagate=False + own StreamHandlers
+    # via uvicorn.config.LOGGING_CONFIG.
+    "uvicorn",
+    "uvicorn.access",
+    "uvicorn.error",
+    # FastMCP / underlying mcp library register their own stream handler.
+    "fastmcp",
+    "mcp",
+    # slowapi (rate-limit middleware) — included defensively. Some versions
+    # install their own handler when rate-limit decisions are configured to
+    # log; on versions that don't, the rerouting is a harmless no-op (no
+    # snapshotted handlers, level untouched).
+    "slowapi",
+)
+
+# Pre-config state for each rerouted logger (handlers, propagate, level).
+# Captured by ``_route_third_party_to_root`` on first invocation and
+# replayed by ``_reset_for_testing`` so test fixtures can return to a
+# state structurally identical to import-time. Handlers are *not* closed
+# in the route-to-root path so the snapshot retains usable handler objects
+# for restore — Python's GC handles FD release when the logger no longer
+# references them after the next reset.
+_third_party_logger_original_state: dict[
+    str, tuple[list[logging.Handler], bool, int]
+] = {}
+
+
+def _route_third_party_to_root() -> None:
+    """Strip third-party loggers' own handlers and force propagation to root.
+
+    Without this, uvicorn access/error and FastMCP server lines bypass the
+    root ``ProcessorFormatter`` and emit as unstructured plain text at
+    DEFAULT severity in Cloud Logging — breaking severity filtering and
+    audit log searches. After this runs, those lines flow through the
+    same JSON formatter as application logs.
+
+    Only resets the level when the library actually had its own handlers
+    — otherwise the level was an operator-/library-configured filter
+    relative to root (e.g. ``uvicorn --log-level warning``) and we mustn't
+    silently widen it.
+    """
+    for name in _THIRD_PARTY_LOGGERS_TO_REROUTE:
+        lg = logging.getLogger(name)
+        # Skip iff the snapshot says we already did the work (had handlers
+        # captured). Lets a re-call (e.g. from an ASGI lifespan startup
+        # after uvicorn initialised) idempotently no-op.
+        already_rerouted = name in _third_party_logger_original_state and bool(
+            _third_party_logger_original_state[name][0]
+        )
+        if already_rerouted:
+            continue
+        # Library may not have initialised this logger yet — there's nothing
+        # to reroute right now. Don't snapshot the empty state (would freeze
+        # us into "no work to do" forever) and warn the operator at WARNING
+        # so they notice in Cloud Logging if a uvicorn-fronted service is
+        # calling configure_logging at module import (before uvicorn's own
+        # ``Config.configure_logging`` runs at server start).
+        if not lg.handlers and lg.propagate:
+            logging.getLogger(__name__).warning(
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
Bot review addressed:

**Medium** — `_reset_for_testing()` now also clears level overrides on the rerouted loggers (`_THIRD_PARTY_LOGGERS_TO_REROUTE`), and the inline comment documents the unrecoverable limitation: the library's original handlers were already discarded, so we can't restore them. Loggers stay handler-less with `propagate=True` until the next `configure_logging()` call. Tests that need pre-config state must reimport.

**Low** — `_route_third_party_to_root()` now closes each removed handler in a try/except, mirroring the existing pattern in `_reset_for_testing()`. Any `FileHandler` / `SocketHandler` a library installed will release its file descriptor cleanly.

Tests still 12/12 pass; ruff + format clean.

Eldad-Caura:
@claude

Eldad-Caura:
Round 3 bot review addressed (`0a5dc77 → be562ee`):

**Medium #1** — added `lg.setLevel(logging.NOTSET)` after `propagate=True`. Without this, a library that set its own logger level (e.g. `uvicorn --log-level warning`) would still filter messages below that level before they reach root, defeating the rerouting. NOTSET delegates filtering entirely to root, where `configure_logging` already enforces the chosen level.

**Medium #2** — `_route_third_party_to_root()` now snapshots the original `(handlers, propagate, level)` for each rerouted logger into a module-level `_third_party_logger_original_state` dict. `_reset_for_testing()` restores from that snapshot instead of just clearing levels. Best-effort: a closed `FileHandler/SocketHandler` is re-installed but won't write again (fine for tests). Snapshot is captured ONCE per logger so re-runs don't overwrite the original.

**Low #1** — moved `_THIRD_PARTY_LOGGERS_TO_REROUTE`, the snapshot dict, and `_route_third_party_to_root` definitions to **before** `_reset_for_testing` so the forward-reference fragility the bot flagged is gone.

**Low #2** (`except Exception: pass` on `h.close()`) — **declining**. The same silent-except pattern is used in `_reset_for_testing` lines 246-254 (which the round-1 bot review explicitly asked us to mirror). Adding logging in only one of the two places creates inconsistency; if the team wants logged-warning everywhere, that's a separate change touching both. Documented in this PR's discussion.

Tests: 12/12 still pass. ruff + format clean.

Eldad-Caura:
@claude

Eldad-Caura:
Round 4 bot review addressed:

**HIGH** — `lg.setLevel(logging.NOTSET)` is now conditional: only reset when the snapshot shows the library installed its own handlers. If the library had no own handlers, the existing level was an operator-set filter (e.g. `uvicorn --log-level warning`) and silently widening it would flood Cloud Logging. Added a regression test (`test_route_third_party_to_root_preserves_operator_set_level_when_no_handlers`) that proves the WARNING setting survives the rerouting.

**Medium** — Added an explicit ordering caveat to the `_THIRD_PARTY_LOGGERS_TO_REROUTE` doc-comment + a debug-level log in the no-handlers no-op branch. Operators investigating un-rerouted output now have a trail in the logs. Did not move the `_route_third_party_to_root` call to an ASGI lifespan startup handler in this PR — that's a larger change touching `app.py` call sites and worth its own scope.

**Low** (closed handlers in snapshot) — Dropped the `h.close()` call in `_route_third_party_to_root`. The snapshot now retains usable (open) handler objects for `_reset_for_testing` to restore. FD release happens via Python GC once both the logger reference and the snapshot reference are dropped on the next reset cycle. The earlier round-1 ask to "mirror `_reset_for_testing`'s close pattern" was redundant with that function's own close call — that one stays.

**Low** (slowapi comment) — Rewritten to be honest about uncertainty: "included defensively. Some versions install their own handler when rate-limit decisions are configured to log; on versions that don't, the rerouting is a harmless no-op."

Tests: 13/13 pass (1 new). ruff + format clean.

Eldad-Caura:
@claude

Eldad-Caura:
Round 5 bot review addressed:

**HIGH (silent uvicorn rerouting failure)** — applied the lighter half of the fix: bumped the silent-skip log from DEBUG → **WARNING**, with a concrete pointer in the message ("call `_route_third_party_to_root` from an ASGI lifespan startup handler if this is a uvicorn-fronted service"). The fuller fix (expose the function publicly + remove from `_configure_logging_impl` + wire it from app.py's lifespan) remains deferred — that's a multi-file change touching every service entry point and warrants its own PR scope. WARNING in Cloud Logging surfaces in alerting, so operators won't miss a misconfigured deployment.

**LOW (retained handlers in production)** — added an explicit comment that the snapshot is intentional + bounded for the listed libraries (`uvicorn` / `fastmcp` / `mcp` / `slowapi` only install StreamHandlers pointing at stderr — ~kilobytes retained, no FD-leak risk).

**LOW (silent propagate mutation when no handlers + propagate=False)** — broadened the skip condition from `not had_own_handlers and lg.propagate` to just `not had_own_handlers`. If a library has no own handlers (whether propagate is True or False), the rerouting is now consistently a no-op + WARNING. This also preserves a library's intentional `handlers=[] + propagate=False` silence, which the previous condition would have inadvertently broken.

Tests still 13/13 pass. ruff + format clean.

Eldad-Caura:
@erni-a please approve

erni-a (dismissed):

Eldad-Caura:
Round 5 bot review addressed:

**HIGH** — silent uvicorn rerouting failure: bumped DEBUG → **WARNING** on the no-handlers branch with a concrete pointer to the lifespan-startup fix. The fuller fix (expose function publicly + invoke from ASGI lifespan) remains deferred — it's a multi-file change touching every service entry point. WARNING surfaces in Cloud Logging alerting so operators won't miss a misconfigured deployment.

**LOW (retained handlers)** — added an explicit comment that the snapshot dict is intentional + bounded for the listed libraries (uvicorn/fastmcp/mcp/slowapi only install StreamHandlers at stderr, ~kilobytes total, no FD-leak risk).

**LOW (silent propagate mutation)** — restructured the skip logic. Three cases now cleanly distinguished:
- **Already rerouted** (snapshot has handlers): no-op continue. Idempotent re-calls.
- **No work to do** (`lg.handlers=[]` AND `propagate=True`): WARNING + continue. Don't snapshot (would freeze into "no work" forever — blocks future lifespan-startup re-call from succeeding).
- **First effective call** (handlers OR propagate=False): snapshot + strip handlers + propagate=True + conditional setLevel.

This also fixed a regression introduced in round 4: the old "snapshot first, decide later" order let pytest's setup snapshot empty handlers and broke the `clears_handlers` test that adds handlers post-snapshot. Now snapshot only happens on the first call where there's actual work, which lets later test/runtime calls correctly snapshot the now-populated state.

Tests: 13/13 pass. ruff + format clean.

Eldad-Caura:
@erni-a please reapprove

erni-a (approved):

PR #30, opened 2026-04-28
fix(noisy-neighbor): widen per-tenant Semaphore caps so bench loop fits

Eldad-Caura:
## Summary

Bump per-tenant in-flight concurrency caps so routine read fan-out doesn't 5xx:

- `per_tenant_search_concurrency`: **8 → 32**
- `per_tenant_write_concurrency`: **4 → 16**

The original caps (private-PR #33, commit [`1ce87b5`](https://github.com/caura-ai/caura-memclaw/commit/1ce87b5d1e95d825178c8bd0fc98c5ae72d430b4)) were sized for a writes-only storm against a single tenant. The loadtest harness's microbench phase fans 10-30 concurrent search / list / stats / entities / agents ops out within one tenant, and `8` was too low to absorb that.

## Loadtest evidence

[`loadtest-1777371451`](../loadtest/report-loadtest-1777371451.md) (post-#29 deploy of `staging-memclaw-core-api-00188-rb6`):

| op | n | failures | tail |
|---|---:|---:|---|
| `bench_search` | 100 | **25 × 500** | — |
| `bench_agents` | 100 | **19 × 500** | — |
| `bench_entities` | 100 | **19 × 500** | — |
| `bench_search_concurrent` | 60 | several | **p99 15.5 s** (target 5 s) |
| total error rate | 1613 reqs | 392 × 5xx | **25.73%** (was 14.44% pre-#33) |

The semaphore IS firing on writes (`bulk_write:429=11`), so the mechanism is sound — only the sizes were too conservative.

## Why these specific values

- **search 32**: covers a single tenant fanning 30 concurrent searches (the harness's heaviest read fan-out) without firing 429.
- **write 16**: still bites under the storm phase (tenant A storm runs at concurrency=20 for 60 s — bounded to ~16 slots per instance instead of unbounded). Per-tenant fleet-wide cap = `16 × peak_instances`; with `staging-memclaw-core-api` at peak 38 in the latest run, that's ~600 slots vs unbounded pre-#33. Tenant isolation preserved.

## What's not changing

- `per_tenant_storage_search_concurrency=4`, `per_tenant_storage_write_concurrency=2` — the deeper bulkhead at the storage roundtrip is unchanged. Those weren't observed firing incorrectly.
- `per_tenant_acquire_timeout_seconds=0.05` — unchanged.

## Test plan

- [x] `tests/test_per_tenant_concurrency.py` — 8/8 pass; tests monkeypatch their own values so default changes don't shift assertions
[...]

PR #33, opened 2026-04-28
fix(memories): per-phase storage timeout for bulk-create (CAURA-599)

Eldad-Caura:
## Summary
- Bulk-create's hot path had per-phase deadlines on embedding (30s) and enrichment (30s) but no per-phase deadline on the storage roundtrip itself — only the 90s ``bulk_request_timeout_seconds`` umbrella.
- A hung storage call could consume embed/enrich slack before 504 fired; non-timeout errors from storage (5xx, network failures) bubbled as generic 500 with no client-actionable retry hint.
- Adds: ``storage_bulk_timeout_seconds = 40s`` setting + validator + ``asyncio.timeout`` wrap around ``sc.create_memories(...)`` + two new exception branches mapping ``httpx.HTTPStatusError`` (5xx → 504) and ``httpx.RequestError`` (network → 503 with ``Retry-After: 1``).

## Why now
Loadtest 1777385048 showed ``bulk_write`` retry rate of 24% (down from 50% pre-PR-#30/#186 but still meaningfully above target). The ``X-Bulk-Attempt-Id`` row-level idempotency contract (CAURA-602) already makes retries safe — but the *client* only retries when it sees a 5xx with a recoverable hint. Pre-this-PR, only the timeout branch surfaced that hint; storage 5xx surfaced as opaque 500.

## Architecture
```
embedding:    asyncio.wait_for(...,  30s)  ✓ per-phase
enrichment:   asyncio.timeout(...,   30s)  ✓ per-phase
storage HTTP: asyncio.timeout(...,   40s)  ✅ NEW (was unbounded under 90s umbrella)
umbrella:     asyncio.wait_for(...,  90s)  ✓ unchanged (belt-and-suspenders)
```

Total worst-case under umbrella: 30s (embed/enrich parallel) + 40s (storage) = 70s, with ~20s slack. Validator enforces ``storage_bulk_timeout < bulk_request_timeout``.

## Test plan
- [x] Unit: 5 new tests in ``tests/test_bulk_atomicity.py`` cover per-phase timeout, 5xx → 504, 4xx propagation (httpx.HTTPStatusError surfaces unchanged), network → 503 + Retry-After, retry-hint message
- [x] Existing 27 tests in test_bulk_write.py + 7 existing in test_bulk_atomicity.py: all pass
- [ ] Staging deploy — observe whether ``bulk_write:500`` count drops below 5/run and whether 503/504 with the X-Bulk-Attempt-Id hint shows up in client retries
- [ ] Re-run loadtest after deploy — expect ``bulk_write`` retry rate to keep dropping (target <20%) and any 5xx to carry the recovery hint

## Out of scope
- Phase 2 chunk-and-retry inside the handler (deferred from CAURA-599 design doc — gated on whether Phase 1 alone moves the needle)
- 4xx → handler-mapped error (4xx is intentionally a hard surface — request-shape bugs should stay loud)
- Edge-layer rate limit interaction with the new 503 (gateway already passes 503 through unchanged)

### DIFF
--- a/core-api/src/core_api/config.py
+++ b/core-api/src/core_api/config.py
@@ -128,6 +128,29 @@ class Settings(BaseSettings):
     # 90s is roughly 2x headroom while staying well under the 300s
     # Cloud Run platform ceiling.
     bulk_request_timeout_seconds: float = 90.0
+    # Per-phase cap on the storage roundtrip inside
+    # ``create_memories_bulk`` (CAURA-599). Embedding and enrichment
+    # already enforce their own 30s caps; storage was the only phase
+    # without one, so a hung storage call ate the full
+    # ``bulk_request_timeout_seconds`` umbrella before the 504 path
+    # fired. The bulk path runs embed and enrich SEQUENTIALLY (embed
+    # at memory_service.py:984, then enrich at the gather a few lines
+    # below), so worst-case time before storage starts is
+    # ``BULK_EMBEDDING_TIMEOUT + BULK_ENRICHMENT_TOTAL_TIMEOUT`` = 60s.
+    # With a 90s umbrella that leaves 30s for storage, so 25s here
+    # gives a 5s slack the validator enforces. Sized to fit the
+    # observed p99 storage roundtrip (~3-5s under load) with ~5x
+    # headroom for slow tails. The ``per_tenant_storage_slot`` acquire
+    # is unbounded — this is the only deadline on the storage phase
+    # itself.
+    storage_bulk_timeout_seconds: float = 25.0
+    # ``Retry-After`` header value (in seconds) on the 503 returned when
+    # the storage call hits a network-level error (DNS, connect refused,
+    # pool exhaustion not surfaced as TimeoutException). 5s is a balance
+    # between letting the upstream recover and not stalling the client
+    # for too long; tunable via env var so an operator can widen it
+    # during a sustained outage to avoid thundering-herd retries.
+    storage_network_error_retry_after_seconds: int = 5
     # Rate limits applied per-route via slowapi decorators
     # (middleware/rate_limit.py). Syntax: "<count>/<period>" where period
     # is second | minute | hour | day.
@@ -260,7 +283,10 @@ class Settings(BaseSettings):
     def _validate_timeout_ordering(self) -> "Settings":
         # Local import avoids a circular: constants → common.constants,
         # but this file is imported by constants.py's dependents.
-        from core_api.constants import BULK_ENRICHMENT_TOTAL_TIMEOUT_SECONDS
+        from core_api.constants import (
+            BULK_EMBEDDING_TIMEOUT_SECONDS,
+            BULK_ENRICHMENT_TOTAL_TIMEOUT_SECONDS,
+        )
 
         if self.request_timeout_seconds < BULK_ENRICHMENT_TOTAL_TIMEOUT_SECONDS:
             raise ValueError(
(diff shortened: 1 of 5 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
Bot review addressed (commit b9994a3):

**Medium — Wrong timeout value in log when per-phase deadline fires** ✅ — log now surfaces both caps so operators can tell which timer fired without having to grep the exception type out of the access log:
\`\`\`python
logger.warning(
    "bulk write timed out (storage cap %ss / request cap %ss); "
    "client should retry with same %s",
    app_settings.storage_bulk_timeout_seconds,
    app_settings.bulk_request_timeout_seconds,
    BULK_ATTEMPT_ID_HEADER,
)
\`\`\`

**Low — Hardcoded \`Retry-After: "1"\`** ✅ — added \`storage_network_error_retry_after_seconds: int = 5\` to \`Settings\` (env-tunable so an operator can widen during a sustained outage to avoid thundering-herd retries) and the \`httpx.RequestError\` handler now reads from it. Default raised from 1 → 5 to give upstreams more room to recover. Updated \`test_storage_network_error_returns_503_with_retry_after\` to assert the new default.

Both changes are config-driven so an incident response can flip values via env var without redeploying.

Eldad-Caura:
@claude

Eldad-Caura:
Round-2 review addressed (commit b3e6fe5):

**Medium — Validator missed the embed/enrich phase budget** ✅ — the previous check `storage < bulk_request` admitted configs like `storage=85, bulk=90` where `30 (BULK_ENRICHMENT_TOTAL) + 85 = 115 > 90` and the umbrella silently fired first, defeating the per-phase cap. Updated to:

```python
if (
    self.storage_bulk_timeout_seconds + BULK_ENRICHMENT_TOTAL_TIMEOUT_SECONDS
    >= self.bulk_request_timeout_seconds
):
    raise ValueError(...)
```

Now enforces the same inner-fires-first invariant the other two checks in this validator already use. Current defaults (30 + 40 = 70 < 90) continue to pass.

**Low — 4xx re-raise emitted no structured log** ✅ — added a `logger.warning` in the 4xx branch so the request-shape-bug case is observable in structured logs. Without it, only FastAPI's generic unhandled-exception log surfaced and it has no bulk context.

```python
if not (500 <= exc.response.status_code < 600):
    logger.warning(
        "bulk write got unexpected %s from storage; "
        "this is a request-shape bug",
        exc.response.status_code,
    )
    raise
```

All 39 tests still pass.

Eldad-Caura:
@claude

Eldad-Caura:
Round-3 review addressed (commit 4bd822a):

**HIGH — Storage-phase timeout did not cover the semaphore acquire wait** ✅ — confirmed the bug by re-reading Python's compound-context-manager semantics: \`async with (A, B):\` enters left-to-right, so the semaphore's \`acquire()\` (which can block indefinitely under slot contention) was running BEFORE \`asyncio.timeout\` armed its deadline. Swapped order so timeout is the outer context manager:

\`\`\`python
async with (
    asyncio.timeout(settings.storage_bulk_timeout_seconds),
    per_tenant_storage_slot("storage_write", data.tenant_id),
):
    storage_results = await sc.create_memories([d for _, d in pending])
\`\`\`

Now the 40s deadline covers both the slot wait AND the storage call, matching the original CAURA-599 intent. Cancellation during a queued acquire raises cleanly from the semaphore's \`__aenter__\` — no slot is held, no release needed on the timeout path.

Added a regression test (\`test_storage_phase_timeout_covers_slot_acquire_wait\`) that pre-saturates the storage_write semaphore for the test tenant, fires a bulk-create, and asserts the route returns 504 in <1s. With the previous (buggy) order the request would block until the 30s umbrella fired — also 504, but the elapsed-time check makes the regression unmistakable instead of just slow.

All 13 tests in \`test_bulk_atomicity.py\` pass.

Eldad-Caura:
@claude

Eldad-Caura:
Round-4 review addressed (commit 678ac52):

**Medium — Validator missed the embed phase** ✅ — confirmed embed and enrich run **sequentially** (not parallel) in the bulk path: \`memory_service.py:984\` (\`asyncio.wait_for(get_embeddings_batch, timeout=30)\`) completes before \`memory_service.py:1018\` (\`asyncio.timeout(BULK_ENRICHMENT_TOTAL_TIMEOUT_SECONDS)\` for the enrich gather). Worst-case time-to-storage is therefore \`embed + enrich\`, and my prior validator's \`storage + enrich < bulk\` check missed the embed term.

Fix shape (Option A from your prompt):
1. Promoted the magic \`30.0\` in \`memory_service.py\` to a named constant \`BULK_EMBEDDING_TIMEOUT_SECONDS\` in \`constants.py\` (alongside the existing \`BULK_ENRICHMENT_TOTAL_TIMEOUT_SECONDS\`)
2. Lowered \`storage_bulk_timeout_seconds\` default 40s → **25s** so the new check passes with the same 90s umbrella: 30 + 30 + 25 = 85 < 90 (5s slack)
3. Validator now enforces \`embed + enrich + storage < bulk\` with the embed term explicitly cited in both the error message and a code comment

25s is still ~5x the observed p99 storage roundtrip (~3-5s under load), so legitimate writes have plenty of headroom; the cap exists to fail fast on a hung storage call, not to be the actual budget.

**Low — Misleading log comment** ✅ — replaced with the accurate description: both caps are static config; the elapsed time on the access-log entry is what distinguishes which timer fired (storage cap at ~25s vs umbrella at ~90s).

**Low — Misleading \`RequestError\` handler comment** ✅ — fixed. \`httpx.PoolTimeout\` IS a \`TimeoutException\` and is caught by the earlier clause; the \`RequestError\` branch only sees DNS failure / connect refused / broken pipe. Comment says exactly that now, with a note about \`PoolTimeout\` being handled earlier so future readers don't make the same mistake.

All 40 tests still pass; ruff check + format clean.

Eldad-Caura:
@claude

Eldad-Caura:
Round-5 review addressed:

**Low — Missing \`exc_info\` on the 4xx warning** ✅ — added \`exc_info=True\` so the stack trace is preserved in structured logs. Without it, the diagnostic for a "request-shape bug" was missing the very thing you'd need to debug it.

**Low — Mixed \`asyncio.wait_for\` and \`asyncio.timeout\` in same function** ✅ — converted the embedding-phase timeout from \`asyncio.wait_for(get_embeddings_batch(...), timeout=N)\` to \`async with asyncio.timeout(N): valid_embeddings = await get_embeddings_batch(...)\` so both per-phase deadlines in \`create_memories_bulk\` use the same idiom. Functional behaviour identical; readability/consistency only.

On the suggestion (typed exception class to distinguish storage-cap vs umbrella in logs): noted but deferring — would add a new exception type for one log-distinguishability use case. Not worth the API surface for now; we can add it if/when log-bucketing on this becomes operationally important.

All 40 tests still pass; ruff check + format clean.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #34, opened 2026-04-29
fix(llm,embedding): explicit httpx.Limits on OpenAI-compatible providers (CAURA-627)

Eldad-Caura:
## Summary
- Both ``OpenAILLMProvider`` and ``OpenAIEmbeddingProvider`` were riding the SDK's default httpx pool (**100 max_connections / 20 keepalive**). Under bulk-write storm fan-out — ``per_tenant_write_concurrency=16`` × ``BULK_ENRICHMENT_CONCURRENCY=10`` = up to **160 concurrent LLM calls per worker process** from a single tenant — the pool saturates and queues subsequent requests, including other tenants'.
- This is the mechanism behind the ``noisy-neighbor-write`` HIGH finding's disproportionate cross-tenant degradation that survived the storage-pool fix in #190.
- Set explicit ``httpx.Limits(max_connections=200, max_keepalive_connections=50)`` on both providers via the ``http_client`` arg of ``AsyncOpenAI``. Env-tunable (``OPENAI_HTTPX_MAX_CONNECTIONS``, ``OPENAI_HTTPX_MAX_KEEPALIVE_CONNECTIONS``) for incident-time adjustment without a redeploy.

## Why these values
- max_connections=200: covers the worst-case 160-concurrent fan-out per process with ~25% headroom.
- max_keepalive_connections=50: 2.5× the default. Aggressive enough to keep TLS handshake costs down on bursts; not so aggressive that idle workers hold an unreasonable number of sockets.

Both are env vars so the values can be re-tuned post-loadtest if needed.

## What's NOT in scope
- Vertex / Gemini providers (use ``google-cloud-aiplatform`` SDK with its own connection management — different abstraction)
- Per-tenant httpx client pools (would require a tenant→client cache; deferred unless needed)

## Test plan
- [x] All 121 LLM/embedding-tagged tests pass locally
- [x] Lint clean (ruff check + format)
- [ ] After merge + OSS deploy + ``ENRICH_ON_HOT_PATH=false`` env flip on staging-memclaw-core-api → run loadtest and watch for:
  - ``noisy-neighbor-write`` drops below 5× (current 8.15×)
  - ``paraphrase_search`` p99 drops back into ≤4s neighborhood (since paraphrase uses an LLM call too)
  - ``freshness-search-never-visible`` improves further
  - No regression in single_write or bulk_write retry rates

## Related
- CAURA-627 deep-dive comment captures the full investigation
- Gotcha #15 in v0.99.0-prod-rollout.md is the previously-documented LLM-pool starvation; this PR is the actual fix

### DIFF
--- a/common/embedding/_registry.py
+++ b/common/embedding/_registry.py
@@ -11,6 +11,7 @@ and core-worker (platform-only) can use it.
 
 from __future__ import annotations
 
+import asyncio
 import logging
 import os
 from collections import OrderedDict
@@ -37,13 +38,29 @@ logger = logging.getLogger(__name__)
 # instead of pinning a 401-returning client forever. ``OrderedDict``
 # semantics: ``move_to_end`` on hit, ``popitem(last=False)`` on miss
 # when full. 256 is well above any realistic tenant count for a single
-# process; raise if that proves wrong. ``AsyncOpenAI`` doesn't require
-# explicit ``aclose`` — the GC drops the underlying httpx pool when
-# the evicted instance has no remaining refs.
+# process; raise if that proves wrong.
+#
+# Eviction must explicitly close the evicted provider's httpx pool
+# (CAURA-627). The OpenAI SDK now gets a user-provided ``http_client``
+# (set in ``OpenAIEmbeddingProvider.__init__`` so we control pool
+# sizing); per the SDK contract, user-provided clients are NOT closed
+# on SDK teardown — caller owns cleanup. ``_get_or_create_openai_provider``
+# below schedules ``aclose()`` on the evicted instance via
+# ``asyncio.create_task`` so the pool drains in the background instead
+# of leaking ``ResourceWarning`` and held connections under long-lived
+# processes that rotate keys past the cache cap.
 _OPENAI_CACHE_MAX = 256
 _openai_provider_cache: OrderedDict[tuple[str, str], OpenAIEmbeddingProvider] = (
     OrderedDict()
 )
+
+# Strong references to in-flight ``aclose()`` cleanup tasks so the GC
+# doesn't reap them mid-execution. ``asyncio.create_task`` returns a
+# task that's kept alive only by user references; without this set the
+# task could be collected before its coroutine finishes draining the
+# httpx pool. The ``add_done_callback`` removes the task from the set
+# once it completes so this doesn't grow unboundedly.
+_background_tasks: set[asyncio.Task[None]] = set()
 
 
 def _get_or_create_openai_provider(api_key: str, model: str) -> OpenAIEmbeddingProvider:
@@ -65,7 +82,21 @@ def _get_or_create_openai_provider(api_key: str, model: str) -> OpenAIEmbeddingP
     provider = OpenAIEmbeddingProvider(api_key=api_key, model=model)
     _openai_provider_cache[cache_key] = provider
     if len(_openai_provider_cache) > _OPENAI_CACHE_MAX:
-        _openai_provider_cache.popitem(last=False)
+        _, evicted = _openai_provider_cache.popitem(last=False)
+        # Schedule cleanup of the evicted provider's httpx pool. The
+        # SDK won't do it for us (we pass an explicit ``http_client``).
+        # Best-effort — bare ``except RuntimeError`` covers the rare
+        # case where the registry is exercised outside a running event
+        # loop (e.g. early startup); GC will reclaim the connections
+        # eventually but with the ``ResourceWarning`` we'd see in
+        # asyncio debug mode.
+        try:
+            loop = asyncio.get_running_loop()
+            task = loop.create_task(evicted.aclose())
+            _background_tasks.add(task)
+            task.add_done_callback(_background_tasks.discard)
+        except RuntimeError:
+            pass
     return provider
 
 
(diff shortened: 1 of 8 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
Round-1 review addressed:

**HIGH — bare \`int()\` in embedding/constants.py would crash at import on bad env values** ✅ — duplicated the same defensive \`_read_int_env\` helper from \`common/llm/constants.py\` into \`common/embedding/constants.py\`, with a docstring explaining why it's duplicated rather than imported (the two modules are intentionally kept independent per the embedding constants file's top docstring, and a cross-import would break that contract).

**Low — \`httpx.AsyncClient\` never closed (ResourceWarning under asyncio debug mode)** ✅ — added \`async def aclose(self)\` to both \`OpenAILLMProvider\` and \`OpenAIEmbeddingProvider\` that delegates to the SDK's \`self._client.close()\`. Documented as idempotent so callers can safely call multiple times.

All 121 LLM/embedding-tagged tests still pass; ruff check + format clean.

Eldad-Caura:
@claude

Eldad-Caura:
Round-2 review addressed (commit 5375569):

**HIGH — No positive-value guard** ✅ — \`int("0")\` and \`int("-1")\` parse cleanly but \`httpx.Limits(max_connections=0)\` is a semaphore with zero permits → every request blocks forever. The new shared helper now rejects values \`< 1\` with a stderr warning and falls back to the default.

**Medium — No cross-constraint validation between max_keepalive and max_connections** ✅ — added an explicit clamp + warning in BOTH constants files. If \`OPENAI_HTTPX_MAX_KEEPALIVE_CONNECTIONS > OPENAI_HTTPX_MAX_CONNECTIONS\`, we now print the misconfig + the clamp action to stderr at module import. httpx's silent clamp is no longer silent.

**Low — Duplicated \`_read_int_env\` in two modules** ✅ — created \`common/env_utils.py\` with a single shared \`read_int_env\` helper. Both \`common/llm/constants.py\` and \`common/embedding/constants.py\` now import from it. The new module has zero imports from peer \`common/\` subpackages so the original "modules are kept independent" invariant is preserved.

The round-2 cycle itself proved the bot's point — applying the HIGH fix to two byte-for-byte copies would have been the maintenance hazard the duplication-finding was warning about.

Added \`tests/test_env_utils.py\` with 8 new tests covering: unset, valid int, invalid string, unit suffix, zero, negative, boundary (1), and surrounding-whitespace (the last documents what \`int(...)\` tolerates).

All 8 env_utils tests + 121 LLM/embedding tests pass; ruff check + format clean on the touched files.

Eldad-Caura:
@claude

Eldad-Caura:
Round-3 review addressed (commit 6530281):

**Medium — Stale \`_registry.py\` comment + missing \`aclose()\` on LRU eviction** ✅ — this was a real resource leak that my round-1 \`aclose()\` addition didn't actually wire up. The OpenAI SDK doesn't close user-provided httpx clients on teardown (caller-owned), and the LRU eviction at \`_registry.py:67-68\` was just dropping the reference. Fix:

\`\`\`python
if len(_openai_provider_cache) > _OPENAI_CACHE_MAX:
    _, evicted = _openai_provider_cache.popitem(last=False)
    try:
        loop = asyncio.get_running_loop()
        loop.create_task(evicted.aclose())
    except RuntimeError:
        pass  # rare: registry exercised outside event loop, GC reclaims
\`\`\`

Also rewrote the comment block above the cache to reflect the new contract (caller owns httpx cleanup; eviction schedules aclose).

**Low — Duplicate clamping warning fires twice in combined LLM + embedding processes** ✅ — extracted \`clamp_keepalive(max_connections, max_keepalive)\` helper into \`common/env_utils.py\`. Both constants modules now call it instead of duplicating the if-block. Single warning text + only one \`import sys\` total.

**Suggestion — \`env_utils.py\` docstring claims deduplication that didn't exist** ✅ — rewritten. The original "duplicated byte-for-byte in [...]" was true within this PR's own history but misleading for future \`git blame\`. Now reads "without this, both modules would need to repeat the same validation logic and a bug fix in one would silently miss the other" — describes the architectural reason without claiming prior duplication.

New tests added:
- \`test_env_utils.py\`: 3 new tests for \`clamp_keepalive\` (passthrough, equal-boundary, clamps + warns)
- \`test_embedding_registry_eviction.py\` (new file): 2 tests covering the eviction \`aclose()\` is awaited; doesn't crash when called outside an event loop

134 total tests pass; ruff check + format clean on the touched files.

Eldad-Caura:
@claude

Eldad-Caura:
Round-4 fixes pushed:

**1. Background-task strong reference for the eviction \`aclose()\`** ✅ — \`asyncio.create_task\` returns a task whose only strong reference is what the caller holds; without one, GC can collect the task mid-execution and the \`aclose()\` coroutine never finishes draining the httpx pool. Added a module-level \`_background_tasks: set[asyncio.Task[None]]\` near the cache declaration in \`common/embedding/_registry.py\`. The eviction path now:

\`\`\`python
task = loop.create_task(evicted.aclose())
_background_tasks.add(task)
task.add_done_callback(_background_tasks.discard)
\`\`\`

The done-callback removes the task from the set on completion so the set doesn't grow unboundedly under sustained eviction churn.

**2. \`clamp_keepalive\` docstring inaccuracy** ✅ — corrected. The previous claim ("exactly one warning line") was about my dedupe goal in the source, not runtime behavior. Each importing module produces its own warning when misconfigured (a process loading both \`common/llm/constants\` and \`common/embedding/constants\` will see the warning twice). Docstring now says exactly that, and clarifies the helper exists to keep the warning text + clamp logic identical across modules — not to deduplicate at runtime.

**3. Generalize \`clamp_keepalive\` env var names** ✅ — added \`max_connections_var\` and \`max_keepalive_var\` keyword-only args defaulting to the current OpenAI names. A future Anthropic-only or Vertex-only client pool can reuse the helper with accurate variable names in the warning text. Added a test (\`test_clamp_keepalive_uses_custom_var_names_in_warning\`) that exercises the overrides AND checks the OpenAI defaults don't leak through when the kwargs are set.

14 of the env_utils + registry-eviction tests pass; ruff check + format clean.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #35, opened 2026-04-29
fix(memory): drop unused unscoped indexes ix_memories_status + ix_memories_visibility (CAURA-629)

Eldad-Caura:
## Summary
- Both indexes were declared **global** (no ``tenant_id`` prefix) in migration 001 but every query that filters by ``status`` or ``visibility`` already pre-narrows by ``tenant_id``. The planner uses ``ix_memories_tenant_id`` (or one of the ``tenant_id``-prefixed composites) to bound the scan to a single tenant's rows, then applies status/visibility as a cheap post-filter.
- Net effect: the unscoped indexes provide **no read benefit** on any code path; they only add per-row index maintenance on every Memory INSERT/UPDATE.
- Surfaced in the CAURA-627 deep-dive (audit_log + memory write contention investigation).

## Verification
\`\`\`bash
grep -rn "ix_memories_status\\|ix_memories_visibility" --include="*.py"
# Returns only:
#   migrations/001_initial_schema.py  (creates them — original)
#   migrations/008_drop_unscoped_memory_indexes.py  (drops them — this PR)
#   common/models/memory.py  (now just a comment reference)
# No live code, query, or test references either index name.
\`\`\`

## Migration 008 semantics
- ``DROP INDEX CONCURRENTLY`` inside alembic's ``autocommit_block`` so AlloyDB primary keeps serving writes through the operation
- ``IF EXISTS`` is belt-and-suspenders for clones that never had these indexes
- Downgrade re-creates them ``CONCURRENTLY`` for round-trip symmetry

## Expected effect
- Write throughput: marginal improvement (sub-5%) — every Memory INSERT/UPDATE skips two index updates
- Read throughput: no effect — every query that uses status/visibility is already tenant-scoped

This is a hygiene win that unblocks the larger CAURA-628 audit-batch-ingest work without introducing competing variables at deploy time.

## Test plan
- [x] 143 memory/bulk tests pass on the core-api in-memory test infra (the SQLAlchemy schema simply doesn't have those indexes anymore — tests don't care)
- [x] Storage-api integration tests: 5 passed, 2 pre-existing failures on \`main\` reproduce identically (\`test_metadata_patch_jsonb_merge\` and \`test_batch_update_status\` — unrelated to this change)
- [x] grep verified no remaining references in code/queries
- [x] Lint clean (ruff check + format)
- [ ] Staging deploy: confirm migration runs cleanly (sub-second on staging row counts; CONCURRENTLY makes it non-blocking even on bigger tables)
- [ ] Optional: \`EXPLAIN ANALYZE\` of representative tenant-scoped queries before/after to confirm planner still uses the tenant-prefixed indexes (it should — they were always preferred)

## Out of scope
- The bigger CAURA-628 (audit_log async batch ingest) — that's the actual fix for the residual noisy-neighbor write contention. This PR is the small hygiene step before it.

### DIFF
--- a/common/models/memory.py
+++ b/common/models/memory.py
@@ -103,8 +103,6 @@ class Memory(Base):
                 "deleted_at IS NULL AND client_request_id IS NOT NULL"
             ),
         ),
-        Index("ix_memories_status", "status"),
-        Index("ix_memories_visibility", "visibility"),
         Index("ix_memories_valid_range", "ts_valid_start", "ts_valid_end"),
         Index("ix_memories_subject_entity", "subject_entity_id"),
         Index("ix_memories_recall_count", "recall_count"),
--- a/core-storage-api/src/core_storage_api/database/migrations/versions/008_drop_unscoped_memory_indexes.py
+++ b/core-storage-api/src/core_storage_api/database/migrations/versions/008_drop_unscoped_memory_indexes.py
@@ -1,0 +1,54 @@
+"""Drop unused unscoped indexes ``ix_memories_status`` + ``ix_memories_visibility``.
+
+CAURA-629 (2026-04-29). Both indexes were declared global (no
+``tenant_id`` prefix) but every query in the codebase that filters by
+``status`` or ``visibility`` already pre-narrows by ``tenant_id`` —
+the planner uses ``ix_memories_tenant_id`` (or one of the
+``tenant_id``-prefixed composites in ``__table_args__``) to bound the
+scan to a single tenant's ~100-1000 rows, then applies status /
+visibility as a cheap post-filter. The unscoped indexes provide no
+read benefit on any tenant-scoped query path; they only add per-row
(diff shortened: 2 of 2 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
Round-1 review addressed:

**Removed the tombstone comment block in \`common/models/memory.py\`** ✅ — the rationale was redundant with the migration's module docstring + this PR's commit message. Per the bot's point: a migration-number reference inside \`__table_args__\` becomes a stale pointer if migrations are ever squashed or renumbered.

The \`Index("ix_memories_valid_range", ...)\` line now sits directly after the partial-unique \`ix_memories_attempt_unique\` block; no doc gap. 143 memory/bulk tests still pass; ruff check + format clean.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #36, opened 2026-04-29
fix(audit): batch + async ingest via in-memory queue (CAURA-628)

Eldad-Caura:
## Summary
- The substantive fix for the residual ``noisy-neighbor-write`` cross-tenant contention identified in CAURA-627 deep-dive (after #34 LLM-pool fix + #35 dead-index drop).
- Replaces the per-event audit POST (one HTTP roundtrip + one AlloyDB ``audit_log`` table-lock acquisition per audit) with a process-local ``asyncio.Queue`` + background batched flusher (one HTTP roundtrip + one multi-row INSERT per up-to-50-event batch).
- Tenant B's gentle audit traffic no longer queues behind tenant A's bulk-storm fan-out at the storage-api / AlloyDB layer.

## Architecture

\`\`\`
            BEFORE                                        AFTER
   ─────────────────────────                  ─────────────────────────────────────
   request handler                            request handler
       │                                          │
       └─ track_task(log_action(...))             └─ AuditEventQueue.enqueue()  (sync, O(1))
              │                                          │
              └─ HTTP POST /audit-logs                   ▼
                     │  (×N per bulk)             ┌──────────────┐
                     ▼                            │ flusher loop │  (one task per process)
              core-storage /audit_add             │  - threshold │
              SESSION + INSERT                    │  - interval  │
              (×N table-lock acquires)            └──────┬───────┘
                                                         │
                                                         ▼
                                           HTTP POST /audit-logs/bulk
                                           SESSION + add_all() multi-row INSERT
                                           (1 table-lock acquire per batch)
\`\`\`

## What ships
| Layer | Change |
|---|---|
| **core-api** | New ``services/audit_queue.py`` (queue + flusher); ``log_action`` enqueues; lifespan binds + drains; storage client gains ``create_audit_logs_bulk``; 3 new ``Settings`` env knobs |
| **core-storage-api** | New ``POST /audit-logs/bulk`` endpoint; ``audit_add_batch`` service method using ``session.add_all`` |
| **tests** | ``test_audit_queue.py`` — 7 unit tests covering threshold/interval/shutdown/overflow/failure/idempotent-start/hung-flusher |

## Operational tunables (env-tunable, no redeploy)
| Setting | Default | Purpose |
|---|---|---|
| ``AUDIT_QUEUE_MAX_SIZE`` | 10000 | Per-process queue cap (~2 MiB at default event size). **Set to 0 to disable the queue** → \`log_action\` falls through to the legacy synchronous path. Incident kill-switch. |
| ``AUDIT_QUEUE_FLUSH_THRESHOLD`` | 50 | Events accumulated → flush immediately |
| ``AUDIT_QUEUE_FLUSH_INTERVAL_SECONDS`` | 1.0 | Max staleness before any single event flushes |

## Backwards compat
- ``log_action(db, *, ...)`` signature unchanged. Existing test sites that ``patch(...)`` it keep working.
- Legacy single-event ``POST /audit-logs`` endpoint untouched — it's the synchronous-fallback target.
- Audit-list freshness lag bounded by flush interval (~1s). Acceptable for ``crystallizer_service`` + ``/audit-logs`` list (post-hoc analysis paths; not real-time).

## Expected loadtest impact
- ``noisy-neighbor-write`` (current 8.15×, target ≤2×) — should drop materially. The fan-out from N audit POSTs to 1 batch is the residual cause CAURA-627 isolated.
- ``bulk_write`` retry rate (current 26%, all 429s) — may improve marginally as the storage-api connection pool sees less audit-side pressure.
- Other metrics — should be steady or marginally better.

## Test plan
- [x] 7/7 new audit-queue unit tests pass (threshold, interval, drain, overflow, recovery, idempotent start, hung-flusher cancel)
- [x] 168 audit/memory/bulk tests pass — no regressions
- [ ] Staging deploy → loadtest → expect ``noisy-neighbor-write`` to drop below 5×
- [ ] If regression: ``AUDIT_QUEUE_MAX_SIZE=0`` env var falls back to legacy synchronous path with no redeploy

## Out of scope
- Durability: events in flight at process kill are lost. Acceptable for audit (post-hoc analysis); if durability becomes critical later, swap the in-memory queue for a Pub/Sub publisher (same shape, different transport).
- Retiring the legacy ``POST /audit-logs`` endpoint — kept for synchronous fallback during the rollout window. Can retire in a separate cycle once the queue path is proven across multiple loadtests.

### DIFF
--- a/core-api/src/core_api/app.py
+++ b/core-api/src/core_api/app.py
@@ -194,6 +194,43 @@ async def lifespan(app):
             )
         )
 
+        # CAURA-628: bind + start the audit batch flusher. ``log_action``
+        # checks for an active queue and falls back to a synchronous
+        # POST when ``audit_queue_max_size = 0`` (kill-switch) or the
+        # queue isn't bound (early startup, tests).
+        audit_queue = None
+        if app_settings.audit_queue_max_size > 0:
+            from core_api.services.audit_queue import (
+                AuditEventQueue,
+                set_audit_queue,
+            )
+
+            async def _flush_audit_batch(events: list[dict]) -> None:
+                try:
+                    await get_storage_client().create_audit_logs_bulk(events)
+                except Exception:
+                    # Log the rich detail (event count, storage-api
+                    # correlation context) here, then re-raise so
+                    # ``_drain_and_flush`` can bump ``_failed_count``
+                    # for the dashboard. The flusher loop's outer
+                    # ``except`` will emit a generic "iteration failed;
+                    # continuing" line on top of this — slight log
+                    # noise but worth it for the failure metric.
+                    logger.exception(
+                        "audit batch flush failed (events=%d); events lost from this batch",
+                        len(events),
+                    )
+                    raise
+
+            audit_queue = AuditEventQueue(
+                max_queue_size=app_settings.audit_queue_max_size,
+                flush_threshold=app_settings.audit_queue_flush_threshold,
+                flush_interval_seconds=app_settings.audit_queue_flush_interval_seconds,
+                flush_callable=_flush_audit_batch,
+            )
+            set_audit_queue(audit_queue)
+            await audit_queue.start()
+
         # Start lifecycle automation scheduler
         from core_api.services.lifecycle_service import lifecycle_scheduler
         from core_api.tasks import cancel_all_tasks, track_task
@@ -219,11 +256,23 @@ async def lifespan(app):
         # we leak the resources the later steps would have freed.
         # Wrap each in its own try/except and continue; the
         # executor.shutdown at the end always runs.
-        for coro in [
-            event_bus.stop(),
-            cancel_all_tasks(),
-            get_storage_client().close(),
-        ]:
+        #
+        # Order matters: drain the audit queue BEFORE closing the
+        # storage client — the final flush goes through that client.
+        # Bus stop also happens before storage-client close because
+        # the bus's pull-loops may still be issuing storage calls
+        # mid-cancel.
+        shutdown_steps: list = []
+        if audit_queue is not None:
+            shutdown_steps.append(audit_queue.stop(timeout=5.0))
+        shutdown_steps.extend(
+            [
+                event_bus.stop(),
+                cancel_all_tasks(),
+                get_storage_client().close(),
+            ]
+        )
+        for coro in shutdown_steps:
             try:
                 await coro
             except Exception:
--- a/core-api/src/core_api/clients/storage_client.py
+++ b/core-api/src/core_api/clients/storage_client.py
@@ -902,6 +902,12 @@ class CoreStorageClient:
     async def create_audit_log(self, data: dict) -> dict:
         return await self._post("/audit-logs", data)  # type: ignore[return-value]
 
+    async def create_audit_logs_bulk(self, events: list[dict]) -> dict:
+        """Batched audit insert (CAURA-628). One HTTP POST + one
+        multi-row INSERT regardless of batch size, vs N round-trips +
+        N table-lock acquisitions on the legacy single-event path."""
+        return await self._post("/audit-logs/bulk", {"events": events})  # type: ignore[return-value]
+
     async def list_audit_logs(
         self,
         tenant_id: str,
--- a/core-api/src/core_api/config.py
+++ b/core-api/src/core_api/config.py
@@ -151,6 +151,35 @@ class Settings(BaseSettings):
     # for too long; tunable via env var so an operator can widen it
     # during a sustained outage to avoid thundering-herd retries.
     storage_network_error_retry_after_seconds: int = 5
+    # Audit-event queue tunables (CAURA-628). The queue replaces the
+    # legacy per-event audit POST with batched flushes, removing the
+    # cross-tenant table-lock contention that surfaced as the residual
+    # noisy-neighbor-write signal after the LLM-pool fix in #34 and
+    # the dead-index drop in #35.
+    #
+    # ``audit_queue_max_size``: per-process queue cap. With ~200 bytes
+    # per event, 10000 events fits in ~2 MiB. Sized well over realistic
+    # storm rate so the queue only fills if storage-api is degraded;
+    # overflow triggers a structured warning + drop counter rather
+    # than blocking the request hot path.
+    #
+    # ``audit_queue_flush_threshold``: events accumulated → flush
+    # immediately. 50 keeps the average batch comfortable for one
+    # multi-row INSERT without stalling steady-state low-volume
+    # tenants behind a long flush cadence.
+    #
+    # ``audit_queue_flush_interval_seconds``: maximum staleness for
+    # any single event before it lands in storage. 1s matches the
+    # CAURA-627 scoping recommendation; the audit_list endpoint may
+    # see up to that delay for the most recent events, which is
+    # acceptable for the post-hoc analysis paths that consume it.
+    #
+    # Setting ``audit_queue_max_size = 0`` disables the queue entirely
+    # — ``log_action`` then falls through to the legacy synchronous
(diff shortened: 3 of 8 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
Round-1 review addressed:

**HIGH 1 — Unhandled \`ValueError\` on invalid UUID** ✅ — extracted \`_parse_resource_id(rid)\` helper that raises \`HTTPException(422)\` with the offending value. Both \`POST /audit-logs\` (single) and \`POST /audit-logs/bulk\` use it, so the same fix covers the legacy endpoint the bot flagged as having the same bug.

**HIGH 2 — KeyError on missing required fields silently loses the batch** ✅ — added a \`_REQUIRED_AUDIT_FIELDS = {"tenant_id", "action", "resource_type"}\` boundary check inside the per-event loop. Returns \`HTTPException(422)\` naming the index + sorted missing-field list on the first violation, before any DB session opens, so already-drained valid events aren't rolled back.

**Medium — Double exception logging on flush failure** ✅ — removed the \`raise\` from the \`_flush_audit_batch\` closure in \`app.py\`. The closure now logs the rich context (event count + storage-api correlation), and the outer flusher loop's \`except\` no longer fires for flush-callable failures (only for genuinely unexpected errors). Updated the comment to make the responsibility split explicit.

**Low — Unbounded batch size** ✅ — added \`_MAX_BATCH_SIZE = 500\` (10× the default \`audit_queue_flush_threshold\`) with an \`HTTPException(422)\` if exceeded. Generous enough for burst flushing if someone tunes \`flush_threshold\` higher; tight enough to keep any single INSERT predictably sized.

Validation order is documented in the bulk endpoint's docstring: batch-size cap → required-field check → UUID parse. All three return 422 so the caller can isolate the bad row and retry the rest.

155 audit/memory/bulk tests pass; ruff check + format clean on both touched files.

Eldad-Caura:
@claude

Eldad-Caura:
Round-2 fixes pushed:

**Fix 1 — split \`flushed_count\` / \`failed_count\` in \`audit_queue.py\`** ✅ — added \`_failed_count\` to \`__init__\` + \`failed_count\` property + property docstrings explaining the operational distinction (\`dropped_count\` = full-queue rejection at enqueue time; \`failed_count\` = drained but lost on flush callable failure). \`_drain_and_flush\` now wraps the callable in try/except: increments \`_failed_count\` and re-raises on failure (so the loop's outer \`except\` still fires), increments \`_flushed_count\` only on success. Updated the test \`test_flush_callable_failure_does_not_kill_loop\` to assert both counters land at 1 (1 failed, 1 flushed).

**Fix 2 — restore \`raise\` in \`_flush_audit_batch\`** ✅ — walked back round-1's "remove the raise" change. The new \`_drain_and_flush\` exception path is what bumps \`_failed_count\` for the dashboard; that's worth more than the slight log noise from the loop's outer \`except\` firing on top. Comment updated to explain the responsibility split: closure logs rich detail, loop logs generic "iteration failed; continuing", \`_drain_and_flush\` owns the metric.

**Fix 3 — list-type guard in storage router** ✅ — added \`isinstance(events, list)\` check before \`len()\`. Without it, a payload like \`{"events": {"0": {...}}}\` would raise \`TypeError\` and propagate as 500. Now returns 422 with the actual type name in the detail.

7/7 audit-queue tests pass; ruff check + format clean across all touched files.

Eldad-Caura:
@claude

Eldad-Caura:
Round-3 fixes pushed:

**HIGH — Unbounded drain can exceed \`_MAX_BATCH_SIZE\`** ✅ — real data-loss bug. With \`max_queue_size=10000\` and a backed-up flusher (storage-api outage), the drain could ship the entire queue in one call, exceeding the storage-side \`_MAX_BATCH_SIZE=500\` cap → 422 → all events lost.

Fix: added \`chunk_size: int = 500\` constructor parameter. \`_drain_and_flush\` now slices the drained list and flushes one chunk at a time, with **per-chunk failure isolation** — if chunk 2 raises, chunks 1 + 3 still flush. Counters (\`flushed_count\` / \`failed_count\`) increment per chunk independently. Default matches the storage-side cap exactly; tests pass \`chunk_size=2\` to exercise slicing without depending on the storage constant.

Two new tests:
- \`test_drain_chunks_above_chunk_size\`: 5 events at \`chunk_size=2\` → 3 callable invocations of sizes [2, 2, 1]
- \`test_chunk_failure_does_not_drop_other_chunks\`: 6 events at \`chunk_size=2\`, fail chunk 2 → \`flushed_count=4\`, \`failed_count=2\`, all 3 chunks still reach the callable

**Medium — Validation raises on first bad event, silently discarding the whole drained batch** ✅ — restructured the storage-side bulk endpoint to **two-pass validation**:
- Pass 1: walk every event, collect required-field-missing AND malformed-\`resource_id\` errors into a list, then raise \`HTTPException(422)\` once with the full list. Caller sees \`[{"index": 3, "missing_fields": ["action"]}, {"index": 7, "invalid_resource_id": "'not-a-uuid'"}]\` and fixes everything in one round-trip.
- Pass 2: normalise. \`resource_id\` already validated, so the \`UUID(rid)\` here is just the type conversion — no error path.

**Low — \`task_done()\` never called** ✅ — added \`self._queue.task_done()\` after each successful \`get_nowait()\` in the drain loop. Future \`queue.join()\` (stricter shutdown / test fixture) won't deadlock now.

Also updated the flusher loop's outer \`except\` comment — since \`_drain_and_flush\` catches per-chunk failures internally, the outer \`except\` only fires on genuinely unexpected errors (queue corruption, asyncio internal bugs). The log message is now \`"audit flush iteration failed unexpectedly; continuing"\` to match.

**Suggestion (counter exposure to /health or Prometheus)**: noted for a follow-up. Worth doing but beyond this PR's scope; the dashboards work today via log grep + the structured warning lines.

9/9 audit-queue tests pass (3 new). ruff check + format clean.

Eldad-Caura:
@claude

Eldad-Caura:
Round-4 fixes pushed:

**Fix 1 — Skipping** ❌ — \`requires-python = ">=3.12"\` in \`core-api/pyproject.toml\`, and \`memories.py\` lines 982-988 already document the convention explicitly:

> ``asyncio.wait_for`` documents raising ``asyncio.TimeoutError``, which Python 3.11 aliased to the builtin ``TimeoutError``. Project floor is 3.12 (``requires-python = ">=3.12"``), so the builtin form is correct everywhere we run; ruff UP041 also mandates this shape. Anyone porting back to Python 3.10 must change this to ``except asyncio.TimeoutError:`` first.

The defensive \`(TimeoutError, asyncio.TimeoutError)\` form would trigger ruff UP041 ("Replace aliases of TimeoutError with TimeoutError") and contradict the established convention. The bot's concern is moot on 3.12+.

**Fix 2 — Per-element type guard in storage router** ✅ — JSON arrays can hold arbitrary types, so \`{"events": [null, "oops", 123]}\` would crash on \`event.keys()\`. Added an \`isinstance(event, dict)\` guard at the start of the per-event loop that appends a structured error and \`continue\`s — collects-all-errors semantics still hold for the rest of the batch.

**Fix 3 — Remove premature \`task_done()\`** ✅ — bot is correct; my round-3 placement violated the documented \`queue.join()\` contract. Calling \`task_done()\` immediately after \`get_nowait()\` would let \`join()\` return while events were still buffered in the local list and not yet written to storage. Removed the call. Replaced the docstring section with the rationale (intentionally NOT calling it; if a future caller wires \`queue.join()\` for a stricter shutdown, the right place for \`task_done()\` is after per-chunk flush completion, not at drain time).

9/9 audit-queue tests pass; ruff check + format clean.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #42, opened 2026-04-29
fix(memories): apply visibility scoping to memory count endpoints (CAURA-000)

erni-a:
## Summary
Two count endpoints diverged from the tenant-facing list view by missing the visibility predicate that `memory_repository.list_by_filters` applies (line 137). Both surfaced the same user-visible symptom: the count widget showed more rows than the list view next to it.

| Endpoint | Bug | Fix |
|---|---|---|
| `GET /api/v1/memories/stats` | Counted `scope_agent` rows when no `agent_id` was in the URL | When `agent_id` omitted, append `Memory.visibility != "scope_agent"` (mirrors list endpoint). When `agent_id` present, author equality dominates — keep existing `Memory.agent_id == agent_id`. |
| `GET /api/v1/fleets` | Per-fleet `memory_count` / `agent_count` counted `scope_agent` rows | No `agent_id` accepted → unconditionally exclude `scope_agent` (the `else` branch of the stats fix). |

User-visible: in the Prism memory browser, the sidebar count showed e.g. **4** while the table rendered only **1** row.

## Diagnosis trace
- `routes/memories.py:256-311` (memory_stats) and `routes/memories.py:134-163` (list_fleets) — both apply `Memory.deleted_at.is_(None)` plus equality filters but no visibility predicate.
- `repositories/memory_repository.py:122-157` (list_by_filters) — applies the visibility predicate based on `caller_agent_id`.
- The two paths diverged precisely when the URL omits `agent_id` — the scenario Prism's default view triggers.

## Scope
Only the tenant-facing endpoints are patched. The admin variants (`/admin/memories/stats`, `/admin/fleets`) are intentionally left unscoped — they mirror `admin_list_memories` (line 1484-1548), which also ignores visibility because admins see everything by design.

## Out of scope
- `memory_count_all` / `memory_distinct_agent_count` soft-delete leak in the **public** `/api/stats` landing-page counter — a different bug on a different endpoint, addressed by branch `CAURA-000-public-counters-exclude-deleted` (commit `ff42e6f`).

## Test plan
- [x] `test_memory_stats_excludes_scope_agent_when_caller_unidentified` — writes 1 `scope_org` + 2 `scope_agent` (different agents), asserts `stats.total == len(list.items) == 1` for an unidentified caller.
- [x] `test_memory_stats_with_agent_id_includes_own_scope_agent` — asserts the agent-scoped path matches list semantics (alice sees her own `scope_org` + `scope_agent` = 2; bob's `scope_agent` excluded).
- [x] `test_list_fleets_excludes_scope_agent` — writes 2 `scope_agent` + 1 `scope_org` into one fleet, asserts `memory_count == 1` and `agent_count == 1` (was 3/3 pre-fix).
- [x] Existing `test_memory_stats` and `test_list_fleets` still pass (default `scope_team` writes are unaffected).
- [x] `ruff check` + `ruff format --check` clean on `core-api/src/`.

### DIFF
--- a/core-api/src/core_api/constants.py
+++ b/core-api/src/core_api/constants.py
@@ -82,8 +82,22 @@ MEMORY_STATUSES_PATTERN = (
 )
 
 # ── Memory visibility levels ──
-MEMORY_VISIBILITIES = ("scope_agent", "scope_team", "scope_org")
-MEMORY_VISIBILITIES_PATTERN = r"^(scope_agent|scope_team|scope_org)$"
+# Named constants are the SoT — ``MEMORY_VISIBILITIES`` and the regex below
+# derive from them so a rename here propagates to membership checks and
+# the Pydantic pattern automatically. SQL filters and any other call
+# site should import the named constant rather than the bare string so
+# a typo turns into a NameError at import time, not a silent miss.
+MEMORY_VISIBILITY_SCOPE_AGENT = "scope_agent"
+MEMORY_VISIBILITY_SCOPE_TEAM = "scope_team"
+MEMORY_VISIBILITY_SCOPE_ORG = "scope_org"
+MEMORY_VISIBILITIES = (
+    MEMORY_VISIBILITY_SCOPE_AGENT,
+    MEMORY_VISIBILITY_SCOPE_TEAM,
+    MEMORY_VISIBILITY_SCOPE_ORG,
+)
+MEMORY_VISIBILITIES_PATTERN = (
+    f"^({MEMORY_VISIBILITY_SCOPE_AGENT}|{MEMORY_VISIBILITY_SCOPE_TEAM}|{MEMORY_VISIBILITY_SCOPE_ORG})$"
+)
 
 MAX_CONTENT_LENGTH = 10000
 CHUNKING_THRESHOLD_CHARS = 2000  # content above this triggers auto-chunking
--- a/core-api/src/core_api/routes/memories.py
+++ b/core-api/src/core_api/routes/memories.py
@@ -8,7 +8,7 @@ from uuid import UUID
 import httpx
 from fastapi import APIRouter, Body, Depends, Header, HTTPException, Query, Request, Response
 from fastapi.responses import JSONResponse
-from sqlalchemy import func, select, tuple_, update
+from sqlalchemy import and_, func, or_, select, tuple_, update
 from sqlalchemy.exc import OperationalError, SQLAlchemyError
 from sqlalchemy.exc import TimeoutError as SQLATimeoutError
 from sqlalchemy.ext.asyncio import AsyncSession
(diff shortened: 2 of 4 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
Medium finding applied in 0cac37b (added on top of erni-a's commits — squash-merge will fold them).

The reviewer is right: today the implicit ``Memory.agent_id == agent_id`` filter happens to be equivalent to ``list_by_filters``'s explicit whitelist for the three current visibility values, but adding any new restricted scope (``scope_private`` etc.) would recreate exactly the count-vs-list mismatch this PR fixes. The new code mirrors ``list_by_filters``'s whitelist line-by-line:

```python
or_(
    Memory.visibility == "scope_org",
    Memory.visibility == "scope_team",
    and_(Memory.visibility == "scope_agent", Memory.agent_id == agent_id),
)
```

Updated the comment to reflect why the predicate is there for forward-compat, not just style.

**Tests**: existing `test_memory_stats*` (3) + broader visibility/api suite (23) all pass. A regression test for an unknown visibility value would require schema-level changes (the API enum rejects unknown values), so it's out of scope for this PR — flagging for a follow-up if/when a new scope is added.

Lint clean (`ruff check` + `ruff format --check`).

Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

Eldad-Caura:
Both follow-up findings applied in 11ee1cf.

**1. NULL-safety (Medium).** Confirmed in the model: ``Memory.visibility`` is ``Mapped[str] = mapped_column(Text, server_default=text("'scope_team'"), nullable=False)`` (common/models/memory.py:62-66) — guaranteed not NULL at the DB level, so the SQL three-valued-logic pitfall doesn't apply. Rewriting as ``or_(.. != .., .. .is_(None))`` would falsely imply NULL is possible. Added an inline note at both ``!= MEMORY_VISIBILITY_SCOPE_AGENT`` sites (list_fleets and memory_stats else-branch) explaining why ``!=`` is safe here, so a future reader doesn't reflexively defensify.

**2. Hardcoded strings (Low).** No ``VisibilityEnum`` exists; the closest equivalent was ``MEMORY_VISIBILITIES = ("scope_agent", ...)`` in ``core_api/constants.py:85``, a positional tuple. Added named constants there (``MEMORY_VISIBILITY_SCOPE_AGENT/_TEAM/_ORG``) and made them the SoT — the existing tuple and regex pattern now derive from them. Replaced literals at all three sites flagged (lines ~155, 296-307, 315). A typo on the constant now turns into a NameError at import time instead of a silent miss.

Other ``scope_*`` literals in the codebase (schemas.py:149,247, etc.) weren't touched — that's a wider sweep beyond the scope of this PR; flagging for a follow-up if/when someone wants to do it.

Tests: 26 pass. Ruff check + format clean.

Eldad-Caura:
@erni-a please approve

Eldad-Caura (approved):

PR #44, opened 2026-04-29
fix(audit): per-tenant slot in batch flusher to stop bulk_write contention (CAURA-631)

Eldad-Caura:
## Summary

Closes [CAURA-631](https://caura-ai.atlassian.net/browse/CAURA-631).

CAURA-628 (PR #36) added an `asyncio.Queue` + background flusher that batches audit events into `POST /audit-logs/bulk`. Loadtest 1777462612 surfaced a regression: `bulk_write` 429 retries jumped from **26% baseline → 72%** with the batch path live, dropping to 18% with the `AUDIT_QUEUE_MAX_SIZE=0` kill-switch.

## Root cause

`_flush_audit_batch` calls `create_audit_logs_bulk()` **without acquiring the per-tenant storage slot** that `/memories/bulk` respects (cap=2 per tenant, see `memory_service.py:369/455/641`). Both share the same httpx pool (100 max / 50 keepalive) and the same storage-api writer pool. Under tenant-A storm the flusher hoards storage connections that `/memories/bulk` requests need; the per-tenant request slot's 50ms timeout fires; clients see 429s.

The kill-switch fixed it because with the queue disabled, audit events fall back to legacy synchronous per-event POSTs that fire inline in the request context — already governed by rate-limit + request-timeout middleware.

## Fix

Group the flush chunk by `tenant_id`, then acquire `per_tenant_storage_slot("storage_write", tenant_id)` per group before the storage call. Each tenant's audit slice now queues fairly behind that tenant's own `/memories/bulk` requests instead of hoarding pool slots.

Per-tenant flush failures don't abort sibling tenants; only the last error is re-raised so `_drain_and_flush` still bumps `_failed_count`. Slightly over-reports failures when one tenant succeeds and another fails in the same chunk — acceptable vs silently swallowing errors. Per-tenant detail logged for forensics either way.

## Test plan

- [ ] DCO + CI pass
- [ ] Existing `tests/test_audit_queue.py` continues to pass — the fix is in the closure `app.py` passes as `flush_callable`; the queue's contract (fan-out chunks to the callable, count successes/failures) is unchanged
- [ ] After staging deploy: re-run the loadtest harness with `AUDIT_QUEUE_MAX_SIZE` ON (default), expect `bulk_write` retries to drop from 72% to ≤26% (matching kill-switch)
- [ ] If validated on staging: prod can drop the `AUDIT_QUEUE_MAX_SIZE=0` env var that's currently overriding

## Validation experiments

| run | AUDIT_QUEUE | observed bulk_write 429 |
|---|---|---|
| 1777411213 baseline (pre-CAURA-628) | n/a (sync POSTs) | 26% |
| 1777462612 (CAURA-628 live) | enabled | 72% |
| 1777463809 (kill-switch) | disabled | 18% |
| **post-merge re-run** | enabled | **expect ≤26%** |

[CAURA-631]: https://caura-ai.atlassian.net/browse/CAURA-631?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core-api/src/core_api/app.py
+++ b/core-api/src/core_api/app.py
@@ -57,6 +57,15 @@ from core_api.routes.settings import router as settings_router
 from core_api.routes.stats import router as stats_router
 from core_api.routes.stm import router as stm_router
 
+# CAURA-631: sentinel bucket for audit events that arrive without a
+# ``tenant_id`` field. Routed through the per-tenant flusher's
+# group-by step then explicitly skipped (events are unattributable, so
+# we'd be writing them to the wrong tenant's audit log otherwise).
+# Identity sentinel (``object()``) rather than a string so a tenant
+# whose actual ID happens to be a debug-style label can't accidentally
+# match it and get its events silently dropped.
+_UNKNOWN_TENANT_SENTINEL: object = object()
+
 _SECURITY_HEADERS = {
     "strict-transport-security": "max-age=63072000; includeSubDomains; preload",
     "x-content-type-options": "nosniff",
@@ -208,15 +217,9 @@ async def lifespan(app):
             )
 
             async def _flush_one_tenant(tid: str, tevs: list[dict]) -> None:
-                # Sentinel bucket for events without a tenant_id — log + skip
-                # rather than failing the whole batch (CAURA-631 review).
-                if tid == "_UNKNOWN_TENANT_":
-                    logger.warning(
-                        "audit batch contained %d events with no tenant_id; "
-                        "skipping write (events unattributable)",
-                        len(tevs),
-                    )
-                    return
+                # Caller separates the sentinel bucket out before scheduling
+                # ``_flush_one_tenant`` over real tenants only, so ``tid`` is
+                # always a real tenant ID here — no sentinel branch needed.
                 try:
                     async with per_tenant_storage_slot("storage_write", tid):
                         await get_storage_client().create_audit_logs_bulk(tevs)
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
Round-2 review fixes in b2912ed1b7eacb2e8dbe8383d970a2c2020f8925:

1. **HIGH `KeyError` on malformed event** — `ev["tenant_id"]` → `ev.get("tenant_id") or "_UNKNOWN_TENANT_"`. Events without tenant_id now bucket to a sentinel; the per-tenant flush function skips that bucket with a warning rather than failing the whole chunk.

2. **MEDIUM sequential tenant loop** — replaced `for` loop with `asyncio.gather(..., return_exceptions=True)`. All tenants flush concurrently; per-tenant slot still caps each tenant's own concurrency at 2. Flush latency now bounded to slowest single tenant rather than sum across tenants.

3. **LOW traceback truncation** — re-raise uses `last_error.with_traceback(last_error.__traceback__)` so Sentry / APM shows the original `create_audit_logs_bulk` failure frame.

Also: pulled the per-tenant body into a separate `_flush_one_tenant` helper for readability + testability. Added `import asyncio` at module top level (per repo convention).

Eldad-Caura:
@claude

Eldad-Caura:
Round-3 fixes in 69acb06a27512950b04f5ca042f4cb949f205884:

1. **MEDIUM `CancelledError` swallowing** — split the post-gather error filter into two buckets:
   - `cancel_errors` = `isinstance(r, asyncio.CancelledError)`
   - `errors` = `isinstance(r, Exception) and not isinstance(r, CancelledError)`
   
   Regular errors propagate first (they carry the actionable Sentry stack); cancellations only raise if they're the sole signal. No silent drops.

2. **LOW sentinel constant** — extracted `_UNKNOWN_TENANT_SENTINEL = "_UNKNOWN_TENANT_"` at module level. Both call sites now reference the constant; a future typo can't silently disconnect the guard from the assignment.

Eldad-Caura:
@claude

Eldad-Caura:
@claude

Eldad-Caura:
Round-4 fixes in 0ae25cbd0e2e014fda5b80bc23157ecc9dd88d41:

1. **MEDIUM `_failed_count` over-counts** — `_flush_audit_batch` now sets `exc.failed_event_count = total_failed_events` on the raised exception (computed by zipping `tasks` with `gather` results and summing only the failed slices). `_drain_and_flush` reads it via `getattr(exc, 'failed_event_count', len(chunk))`, credits the surviving slices to `_flushed_count`, and only the genuinely-failed events to `_failed_count`. Falls back to old whole-chunk accounting if the attribute is absent.

2. **LOW BaseException subclasses silently swallowed** — added `other_base` bucket for `BaseException` subclasses that aren't `Exception` or `CancelledError` (e.g. `SystemExit`, `KeyboardInterrupt`, future `BaseExceptionGroup`). Re-raised as-is between the regular-error branch and the cancel branch — no exception class can silently disappear now.

3. **LOW sentinel collision** — `_UNKNOWN_TENANT_SENTINEL` is now `object()` (identity sentinel). `_flush_one_tenant` uses `if tid is _UNKNOWN_TENANT_SENTINEL` (identity check). `by_tenant` is typed `dict[str | object, list[dict]]`. A tenant whose actual ID happens to be `"_UNKNOWN_TENANT_"` literal can't collide.

4. **SUGGESTION dead-code guard** — dropped `and not isinstance(r, asyncio.CancelledError)` from the `errors` list comp (CancelledError is BaseException, not Exception, since Python 3.8 — guard was always True).

Touches `audit_queue.py` for the failed_event_count read; backwards compatible with any existing flush_callable that doesn't set the attribute.

Eldad-Caura:
Round-5 fixes in :

1. **HIGH CancelledError priority inversion** — flipped raise order to `cancel → other_base → errors`. Asyncio cancellation now propagates first; co-occurring regular errors don't silence shutdown signals anymore.

2. **HIGH `failed_event_count` not set on BaseException paths** — now attached on all 3 raise branches (cancel, other_base, errors). `_drain_and_flush` reads accurate counts whichever path fires, instead of falling back to `len(chunk)` and inflating `_failed_count`.

3. **MEDIUM sentinel events credited as flushed** — pop sentinel bucket via `by_tenant.pop(_UNKNOWN_TENANT_SENTINEL, [])` BEFORE fan-out. Logged once at warning level, never enters `tasks`. Removed the now-dead sentinel branch from `_flush_one_tenant`. (Note: in the no-failure case, `_drain_and_flush` still bumps `_flushed_count` by the full `len(chunk)` since the callable returns cleanly — the sentinel bucket is now correctly excluded from `tasks`/`total_failed_events`, but the `len(chunk)`-includes-sentinel residual metric skew is the boundary of what this PR fixes without changing the callable contract. Sentinel events are documented as rare malformed-input cases.)

4. **LOW redundant `with_traceback`** — replaced `raise last_error.with_traceback(last_error.__traceback__)` with plain `raise last_error`. `__traceback__` is already attached from the `except` in `_flush_one_tenant`; the with_traceback call was a no-op.

Eldad-Caura:
@claude

Eldad-Caura:
Both follow-ups applied in 9d3c07d04581c7d23a52d80bd27f252cbbe6e044:

1. **`_drain_and_flush` accounting gap on BaseException** — `except Exception` → `except BaseException`, accounting always runs, log only on `isinstance(exc, Exception)` (skip CancelledError noise), `raise` to preserve propagation. The per-tenant flusher's CancelledError / SystemExit / BaseExceptionGroup raises (which carry `failed_event_count`) now flow through the accounting block instead of bypassing it.

2. **dead isinstance filter after sentinel pop** — replaced `[(tid, tevs) for ... if isinstance(tid, str)]` with an `assert` + plain `list(by_tenant.items())`. Sentinel was already popped above; the filter was dead code that would silently drop future non-string keys instead of surfacing the bug.

Eldad-Caura:
@claude

Eldad-Caura:
Test failure caught — `test_chunk_failure_does_not_drop_other_chunks` exercises the per-chunk failure isolation contract: chunk 1 succeeds, chunk 2 raises, **chunk 3 should still flush**. Unconditional `raise` after the accounting block exits the loop and breaks chunk 3.

Fix in : re-raise **only for non-`Exception` BaseException**. Regular `Exception` stays caught + accounted + the loop continues (preserves the existing chunk-isolation contract). `CancelledError` / `SystemExit` / `BaseExceptionGroup` re-raise to propagate shutdown. Tests pass locally.

Eldad-Caura:
Both follow-ups applied in 25cc0c35850545f9522b6dcbdc1a46907697b6cf:

1. **`assert` → `raise RuntimeError`** — runtime guard now fires under `python -O` too, instead of being optimised out.

2. **Sentinel events no longer inflate `_flushed_count`** — `_flush_audit_batch` stashes `_sentinel_count = len(sentinel_evs)` on itself unconditionally on every entry. `_drain_and_flush` subtracts `getattr(self._flush_callable, '_sentinel_count', 0)` from `_flushed_count` on **both** success and partial-failure branches. Defaults to 0 for callables that don't use this protocol — backward compatible. Net effect: sentinel-dropped events are credited as neither flushed nor failed, matching their actual disposition (intentional drop with warning log).

All 9 audit_queue tests still pass locally.

Eldad-Caura:
@claude

Eldad-Caura:
Round-7 review:

1. **MEDIUM ExceptionGroup for multi-tenant failures** — applied in 51a53117038a7167952a65df6e8cc104108e5954. `if len(errors) == 1` keeps the simple single-error raise (preserves the direct storage-frame in Sentry); `else` wraps in `ExceptionGroup(f"audit batch flush failed for {n} tenants", errors)` so every tenant's traceback is preserved. `failed_event_count` attached to the group.

2. **LOW external CancelledError over-counts** — applied. `if isinstance(exc, Exception): failed = getattr(..., len(chunk))` (regular path, conservative "all failed" fallback). `else: failed = getattr(..., 0)` (cancel/shutdown path — disposition unknown, don't spike the dashboard).

3. **MEDIUM `_sentinel_count` function-attribute side-channel** — **acknowledged but not changed**. The user explicitly chose this approach in the prior round to keep the protocol change local rather than bumping the callable signature to return `int` or `FlushResult`. Tradeoff: if anyone wraps `flush_callable` (`functools.partial`, test spy, lambda), `getattr` returns 0 and sentinel events get double-credited. In production the callable is bound directly to `AuditEventQueue` without wrappers, so the practical risk is low. Worth a follow-up ticket if/when wrapping becomes a real concern.

All 9 audit_queue tests still pass locally.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #45, opened 2026-04-29
fix(memory): restore ix_memories_status + ix_memories_visibility (CAURA-632)

Eldad-Caura:
## Summary

Closes [CAURA-632](https://caura-ai.atlassian.net/browse/CAURA-632) (test A).

Migration 009 restores the two unscoped indexes that migration 008 (CAURA-629) dropped. Bisect of the **8.15× → 14.49×** regression on `noisy-neighbor-write` between baseline `1777411213` and `1777463809` identified two contributors that compound multiplicatively:

1. **The dropped indexes (this PR)**. Sampler queries in `postgres_service.py` issue `WHERE status IN (...) AND visibility != 'scope_agent'` filters on a per-tenant scoped slice. Without the unscoped indexes, planner falls back to a `tenant_id`-only index scan + post-filter. Cheap normally, but under tenant-A storm the post-filter rows contend with INSERT row-locks. Restoring the indexes lets the planner use bitmap-AND of two narrow indexes and skip the contended rows entirely.
2. **`EMBED_ON_HOT_PATH=false` + `ENRICH_ON_HOT_PATH=false`** add ~5× write-side amplification. Independent of this migration — tracked as the second half of CAURA-632's bisect. May be addressed by separate experiment / further worker tuning.

## Why dropping was wrong

Migration 008's claim that "every query already pre-narrows by `tenant_id`" was true in steady state — `pg_stat_user_indexes` showed zero scans. It missed the noisy-neighbor write-storm shape where the per-tenant slice becomes hot and the post-filter rows materially contend with concurrent INSERT row-locks.

## Trade-off

Restoring the indexes re-adds the per-row maintenance cost on every Memory `INSERT`/`UPDATE` that 008 wanted to remove. On the noisy-neighbor metric the contention savings (~6×) dwarf the maintenance cost — clear net win. A future redesign could add tenant-scoped composites (`(tenant_id, status)` etc) to capture both wins, but that's a larger schema change tracked separately.

## Test plan

- [ ] DCO + CI pass
- [ ] After staging deploy: writer service auto-runs alembic on startup; both indexes appear in `\d+ memories` output
- [ ] Re-run loadtest harness; expected: tenant B write p95 drops from 14.49× toward 8–10× (baseline). If it stays high, the async-amplification half of the bisect is the dominant contributor and we'll tune workers / re-evaluate `EMBED_ON_HOT_PATH` separately.

## Related

- Migration 008 (CAURA-629) — original drop, now reverted by this migration
- CAURA-631 — separate fix for the audit batch ingest 429 spike (PR #44)

[CAURA-632]: https://caura-ai.atlassian.net/browse/CAURA-632?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core-storage-api/src/core_storage_api/database/migrations/versions/009_restore_unscoped_memory_indexes.py
+++ b/core-storage-api/src/core_storage_api/database/migrations/versions/009_restore_unscoped_memory_indexes.py
@@ -1,0 +1,74 @@
+"""Restore unscoped indexes ``ix_memories_status`` + ``ix_memories_visibility``.
+
+CAURA-632 (2026-04-29). Migration 008 dropped both indexes on the
+premise that "every query already pre-narrows by ``tenant_id`` so the
+unscoped indexes give no read benefit." That premise held in
+*steady-state* — ``pg_stat_user_indexes`` showed zero scans, and the
+planner does prefer the ``tenant_id``-prefixed composites for normal
+single-tenant queries.
+
+It DID NOT hold under the noisy-neighbor write storm shape that the
+loadtest harness exercises. Bisect of regression 8.15x → 14.49x on
+``noisy-neighbor-write`` between baseline ``1777411213`` and
+``1777463809`` (after kill-switching CAURA-628, ruling that out)
+identified two contributors that compound multiplicatively:
+
+  1. The dropped indexes (this migration). Sampler queries in
+     ``postgres_service.py`` (semantic-search candidate fetch,
+     contradiction detection, statistics) issue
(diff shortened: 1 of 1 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #48, opened 2026-04-30
feat(events): EVENT_BUS_PUBSUB_MAX_MESSAGES override (CAURA-636) — v2

Eldad-Caura:
## Summary
Reintroduces the env-var override reverted in #47, with both Claude review findings from #46 addressed.

- Wire `EVENT_BUS_PUBSUB_MAX_MESSAGES` into `get_event_bus()` so the Pub/Sub bus's pull batch size can be tuned without a code change. Default stays 25 (the existing `PubSubEventBus.__init__` default).
- **Review fix #1 (Medium)**: Add upper-bound check (`<= 1000`). GCP Pub/Sub's `pull()` API rejects `max_messages > 1000` with `INVALID_ARGUMENT`. Without this check, a typo would surface as a runtime error at first pull instead of failing fast at boot.
- **Review fix #2 (Low)**: Unify the two `PubSubEventBus(...)` branches into a single kwargs-splat call. Future constructor params will no longer silently use stale defaults in one branch.
- Validation raises `RuntimeError` at boot for non-int / non-positive / over-1000 values, matching the existing factory style for `GCP_PROJECT_ID` and `EVENT_BUS_SUBSCRIPTION_PREFIX`.

## Tests
- 16 tests pass (`pytest tests/test_events/test_factory.py`)
- New cases: override-applied, default-when-unset, upper-bound (1000) accepted, parametrized rejection of `\"abc\"`, `\"0\"`, `\"-3\"`, `\"1.5\"`, `\"1001\"`.

## Why
[CAURA-636](https://caura-ai.atlassian.net/browse/CAURA-636) — async-amplification regression. After CAURA-631 + 632 fixes recovered \`noisy-neighbor-write\` from 14.49x to 12.7x, the residual gap vs the 8.15x baseline is async-pipeline amplification (each write fans out to embed + enrich PATCH-backs, ~3x storage-writer ops). I want to test whether dropping the worker's \`max_messages\` from 25 to 10 narrows that gap before adding a per-tenant slot in the worker. This PR provides the knob.

## Process note
Previous PR #46 was merged via \`--admin\` while the auto-review was still pending; #47 reverted it. This v2 will go through normal review/merge — please review, no admin-merge.

## Test plan
- [x] \`pytest tests/test_events/test_factory.py\` passes locally (16 tests)
- [x] \`ruff check\` + \`ruff format --check\` clean on changed files
- [ ] Auto-review (Claude Code Review) green
- [ ] Approved + merged through normal protected-branch flow
[...]

PR #53, opened 2026-05-02
fix(worker): per-tenant storage slot to stop noisy-neighbor PATCH-back contention (CAURA-636)

Eldad-Caura:
## Summary

Closes the residual async-amplification gap that PR #44 (CAURA-631) and #45 (CAURA-632) couldn't reach: the worker's two PATCH-backs (embed result + enrich result) per memory write hammer the storage-writer pool with no per-tenant gate, so a single tenant's burst pushes other tenants into a 12.7x \`noisy-neighbor-write\` regression vs the 8.15x baseline.

This PR adds a worker-local \`per_tenant_storage_slot(tenant_id)\` mirroring \`core_api.middleware.per_tenant_concurrency.per_tenant_storage_slot\`, and wraps both PATCH calls in \`consumer.py\` (\`update_memory_embedding\` at line 160, \`update_memory_enrichment\` at line 532).

## Why not Option 2 (throughput tuning)

Loadtest \`1777548665\` ran with \`EVENT_BUS_PUBSUB_MAX_MESSAGES=10\` on the staging worker (PR #48 added the env-var knob). Result:

| metric | post-fix \`1777538050\` | max_messages=10 \`1777548665\` |
|---|---:|---:|
| \`bulk_write\` 429 retries | 18% | **30%** ⚠ regression |
| \`noisy-neighbor-write\` | 12.7x | **12.8x** flat |
| \`single_write\` p99 | 11,670 ms | **14,540 ms** ⚠ regression |
| \`freshness-search-never-visible\` | 4/10 | **0/10** ✅ improved |

\`noisy-neighbor-write\` barely moved. Slowing aggregate worker throughput just made PATCH-backs queue behind core-api's bulk publish bursts longer (worse \`bulk_write\` 429s and worse \`single_write\` p99). The issue is per-tenant burst, not aggregate concurrency.

## What changed

- **NEW** \`core_worker.per_tenant_concurrency\` — \`per_tenant_storage_slot(tenant_id)\` async context manager. Per-process \`dict[str, asyncio.Semaphore]\`. Cap from \`settings.per_tenant_storage_write_concurrency\`.
- **\`config.py\`** — adds \`per_tenant_storage_write_concurrency: int = 2\` (matches core-api default), and a module-level \`settings = Settings()\` singleton (mirrors core-api's pattern at \`core_api/config.py:452\`) so callers don't reconstruct pydantic-settings per call.
- **\`consumer.py\`** — wraps both PATCH-back call sites with the slot. Provider call upstream stays unguarded (it doesn't touch the storage-writer pool).

## Why not extracted to \`common/\`

State is per-process; worker containers don't share semaphores with core-api. core-api's variant has a second form (route-entry slot with timeout + 429) the worker doesn't need. Duplication is ~30 lines of mechanics; the per-service tuning rationale documented in each module's docstring is more useful than a shared helper that re-introduces a settings indirection.

## Tests

5 new tests in \`core-worker/tests/test_per_tenant_concurrency.py\`:
- [x] semaphore created at configured cap on first acquire
- [x] separate tenants get independent semaphores
- [x] slot released on normal exit
- [x] slot released on exception
- [x] third concurrent caller for same tenant queues; fourth caller for a different tenant proceeds (noisy-neighbor invariant)

\`pytest core-worker/tests/\` → 54 pass (49 pre-existing + 5 new). \`ruff check\` + \`ruff format --check\` clean.

## Process note

Going through normal review/merge — **no admin-merge**.

## Test plan
- [x] Unit tests pass locally
- [ ] Auto-review (Claude Code Review) green
- [ ] CI green
- [ ] Approved + merged through normal protected-branch flow
- [ ] After merge: deploy worker to staging via \`deploy-oss-services.yml\`, re-run loadtest, verify \`noisy-neighbor-write\` recovers materially toward 8.15x. If yes → close CAURA-636. If no → escalate to Option 3 (storage-api side per-tenant cap, ~half-day).

### DIFF
--- a/core-worker/src/core_worker/config.py
+++ b/core-worker/src/core_worker/config.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 from typing import Literal
 
+from pydantic import field_validator
 from pydantic_settings import BaseSettings, SettingsConfigDict
 
 
@@ -34,5 +35,38 @@ class Settings(BaseSettings):
     # than 504 + nack + redeliver.
     storage_http_timeout_s: float = 30.0
 
+    # Per-tenant cap on concurrent storage PATCH-backs (embed/enrich
+    # results). Keep aligned with core-api's
+    # ``per_tenant_storage_write_concurrency`` so a single tenant's
+    # combined occupancy of the storage-writer pool stays bounded
+    # across both services. CAURA-636 — tenant-A storm fans into ~2
+    # PATCHes per write (embed + enrich); without this cap one
+    # tenant's burst saturates the pool and pushes other tenants into
+    # 12x write-latency regression.
+    per_tenant_storage_write_concurrency: int = 2
+
+    @field_validator("per_tenant_storage_write_concurrency")
+    @classmethod
+    def _per_tenant_cap_must_be_positive(cls, v: int) -> int:
+        # ``asyncio.Semaphore(0)`` is valid Python — no ValueError at
+        # construction — but every ``acquire()`` would block forever
+        # with no error or log, silently deadlocking every storage
+        # PATCH-back. Reject at config load instead so the misconfig
+        # surfaces at startup.
+        if v < 1:
+            raise ValueError(
+                "per_tenant_storage_write_concurrency must be >= 1; 0 would deadlock every storage PATCH-back"
+            )
+        return v
+
     # Cloud Run port.
     port: int = 8080
+
+
+# Module-level singleton — matches core-api's ``core_api.config.settings``
+# pattern. Modules that need a config value should ``from core_worker.config
+# import settings`` rather than reconstruct ``Settings()``: pydantic-settings
+# isn't cached, so every reconstruction re-reads env + re-runs validators.
+# Also lets call sites drop the ``# type: ignore[call-arg]`` that mypy needs
+# at the construction site.
+settings = Settings()  # type: ignore[call-arg]
--- a/core-worker/src/core_worker/consumer.py
+++ b/core-worker/src/core_worker/consumer.py
@@ -64,6 +64,7 @@ from core_worker.clients.storage_client import (
     update_memory_embedding,
     update_memory_enrichment,
 )
+from core_worker.per_tenant_concurrency import per_tenant_storage_slot
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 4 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
Both findings + the suggestion applied in 341450d:

**Medium — Unbounded growth.** Switched `_TENANT_SEMAPHORES` to `OrderedDict` with LRU eviction, cap `_MAX_TRACKED_TENANTS = 4096`. On hit, `move_to_end` keeps the entry MRU; on miss, `popitem(last=False)` evicts the LRU entry before installing the new one. Re-acquiring an evicted tenant allocates a fresh semaphore (cap-fresh, no inherited backpressure) — covered by the new `test_evicted_tenant_reacquire_creates_fresh_semaphore` test. Eviction-during-in-flight is safe: only the dict entry drops; coroutines still hold their semaphore reference for release.

**Low — Deadlock on `0`.** Added `field_validator("per_tenant_storage_write_concurrency")` rejecting `< 1` with a clear message. Parametrized test covers `0`, `-1`, `-100`.

**Suggestion — `locked() → acquire()` race-free note.** Added an inline comment explaining why the check is race-free in single-threaded asyncio (no `await` between the check and the acquire blocks; raced state would log the same way after acquire and would be too late to matter anyway).

60 worker tests pass (49 pre-existing + 11 new — added 4: LRU eviction, evict-then-reacquire, validator-rejects-zero, validator-parametrized-non-positive). ruff clean.

Eldad-Caura:
@claude

Eldad-Caura:
Both low-severity fixes applied in 7cbca8f:

1. **Log key `cap` → `configured_cap`** (lines 100-113). Reflects that the value comes from the live `Settings` (not the semaphore's frozen internal capacity); since `Settings` isn't `frozen=True`, the two can drift. Comment explains the distinction.
2. **`_reset_for_tests()` production guard** (lines 121-131). `assert os.environ.get("PYTEST_CURRENT_TEST")` — fires if anything outside pytest tries to call it. The 60 existing tests still pass (the autouse fixture in `test_per_tenant_concurrency.py` runs under pytest, so the env var is set).

60 worker tests pass; ruff clean.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #54, opened 2026-05-03
feat(stats): real public counters + /api/v1/status + trust soft-pass (CAURA-000)

erni-a:
## Summary

Three loosely-related fixes in the public API surface, plus a one-line
local-dev fix. Stacked on top of #27 — review #27 first.

* **`fix(stats):` real `tenant_count`.** `/api/v1/stats` was returning a
  hardcoded `tenant_count: 1` regardless of how many tenants actually had
  memories. Production screenshot from net/dev/etoro confirmed all three
  show `1`. Adds `memory_distinct_tenant_count()` in core-storage,
  exposes `GET /memories/distinct-tenants`, threads through the
  storage-client, and runs the three counters in one `asyncio.gather()`.
* **`feat(status):` new `GET /api/v1/status` endpoint.** Public
  service-fingerprint: version, OSS-vs-enterprise mode, LLM/embedding
  provider names + models, dependency probes, symbolic
  `platform_init_errors`. Refactors `health()` body into a shared
  `_probe_dependencies()` helper. Tests cover shape, no-auth posture,
  no-secrets-leaked, and stub-configured providers. Gateway alias is in
  the enterprise repo's `CAURA-000-tenants-tile-and-status-gateway`
  branch.
* **`fix(trust):` API-key validity is the gate; soft-pass missing
  Agent rows.** Pre-fix, a fresh agent calling `memclaw_list` hit
  `403 "Agent not found (trust_level=0 < required 1)"` because
  `require_trust` rejected missing rows and `get_or_create_agent` only
  fires on writes. Soft-passes a missing row at `DEFAULT_TRUST_LEVEL=1`,
  which matches what every freshly-registered agent gets. Also drops the
  redundant `if not_found: 403` blocks in `routes/insights.py` and
  `routes/evolve.py` that bypassed the soft-pass. MCP tool handlers
  already use the `_, _, terr` pattern so they pick up the soft-pass for
  free.
* **`fix(core-storage-api):` ignore extra env-file keys.** Mirror
  core-api / core-worker's `extra="ignore"` on the Pydantic Settings.
  Without it every unit-test run that touches a module which
  transitively imports core-storage-api's config crashes on
  `extra_forbidden` whenever the monorepo `.env` has `OPENAI_API_KEY`
  for core-api.

## Test plan

* [x] `pytest -m unit` — 696/696 pass on this branch
* [x] Live E2E in local enterprise stack (rebuilt core-api + storage):
  - `GET /api/v1/stats` → `{"tenant_count": 11, "memory_count": 27, "agent_count": 9}` (was hardcoded 1)
  - `GET /api/v1/status` → 200 with documented shape, mode=enterprise, deps connected
  - `GET /api/v1/memories?agent_id=fresh-X&scope=fleet` → `200 {"items":[]}` (was 403)
* [ ] Auto-review (Claude Code Review) green
* [ ] Approved + merged through normal flow

## Why

Three small fixes to the unauth + auth-gated public surface. The status
endpoint replaces a planned `?status=true` flag on `/health`; clean
separation of "deploy gate" (health → 503) and "service fingerprint"
(status → always 200). The trust soft-pass closes a real bug operators
were hitting — fresh API keys 403'd until the agent happened to write.

## Coordination

Stacked on **#27** (`CAURA-000-public-counters-exclude-deleted`).
Conflict-free with **#42** and **#52** (verified via `git merge-tree`).

### DIFF
--- a/core-api/src/core_api/clients/storage_client.py
+++ b/core-api/src/core_api/clients/storage_client.py
@@ -398,6 +398,11 @@ class CoreStorageClient:
         result = await self._get("/memories/distinct-agents")
         return (result or {}).get("count", 0)
 
+    async def count_distinct_tenants(self) -> int:
+        """Global count of distinct tenants with at least one live memory."""
+        result = await self._get("/memories/distinct-tenants")
+        return (result or {}).get("count", 0)
+
     async def update_embedding(
         self,
         memory_id: str,
--- a/core-api/src/core_api/constants.py
+++ b/core-api/src/core_api/constants.py
@@ -81,6 +81,15 @@ MEMORY_STATUSES_PATTERN = (
     r"|outdated|conflicted|archived|deleted)$"
 )
 
+# ── Health / status probe timeouts ──
+# Upper bound on a single dependency probe (storage / redis / event_bus).
+# Cloud Run typically gives health checks 10-30s before marking a revision
+# unhealthy; we want to fail-fast well before that so a stalled backend
+# can't hang the whole probe. Shared between ``/health`` (binary 503 deploy
+# gate) and ``/stats`` / ``/status`` (public endpoints with the same posture
+# — return ``0`` / "unreachable" rather than block landing-page hits).
+PROBE_TIMEOUT_SECONDS = 5.0
+
 # ── Memory visibility levels ──
 # Named constants are the SoT — ``MEMORY_VISIBILITIES`` and the regex below
 # derive from them so a rename here propagates to membership checks and
--- a/core-api/src/core_api/mcp_server.py
+++ b/core-api/src/core_api/mcp_server.py
@@ -1198,7 +1198,22 @@ async def memclaw_insights(
     min_level = 1 if scope == "agent" else 2
 
     async with _mcp_session() as db:
-        _, _, terr = await _require_trust(db, tenant_id, agent_id, min_level=min_level)
+        # Mirror the REST insights gate: ``require_trust`` soft-passes a
+        # missing Agent row at ``DEFAULT_TRUST_LEVEL`` (read-only ergonomics
+        # — see ``memclaw_list`` below for the intended consumer), but
+        # this handler persists insight memories + audit-log rows keyed
+        # to ``agent_id``. Without a registered row backing the name,
+        # attribution becomes unverifiable, so re-block unregistered
+        # agents on the write path. ``terr`` is None for the soft-pass
+        # case, so without the explicit ``not_found`` check below the
+        # fabricated id would fall through and write.
+        _, not_found, terr = await _require_trust(db, tenant_id, agent_id, min_level=min_level)
+        if not_found:
+            return _with_latency(
+                f"Error (403): Agent '{agent_id}' is not registered. "
+                "Register the agent by writing one memory first.",
+                t0,
+            )
         if terr:
             return _with_latency(_error_response("FORBIDDEN", parse_trust_error(terr)), t0)
         try:
@@ -1271,7 +1286,19 @@ async def memclaw_evolve(
     min_level = 1 if scope == "agent" else 2
 
     async with _mcp_session() as db:
-        _, _, terr = await _require_trust(db, tenant_id, agent_id, min_level=min_level)
+        # Mirror the REST evolve gate (and ``memclaw_insights`` above):
+        # block unregistered agents on the write path so the
+        # outcome/rule memories + audit-log rows have a real registered
+        # ``agent_id`` backing them. Soft-pass remains in
+        # ``require_trust`` itself for the read-only ``memclaw_list``
+        # below.
+        _, not_found, terr = await _require_trust(db, tenant_id, agent_id, min_level=min_level)
+        if not_found:
+            return _with_latency(
+                f"Error (403): Agent '{agent_id}' is not registered. "
+                "Register the agent by writing one memory first.",
+                t0,
+            )
         if terr:
             return _with_latency(_error_response("FORBIDDEN", parse_trust_error(terr)), t0)
         try:
--- a/core-api/src/core_api/routes/evolve.py
+++ b/core-api/src/core_api/routes/evolve.py
@@ -141,6 +141,18 @@ async def report_outcome_endpoint(
         if not caller_agent_id:
             raise HTTPException(status_code=422, detail="agent_id is required.")
         min_level = 1 if body.scope == "agent" else 2
+        # ``require_trust`` soft-passes a missing Agent row at
+        # ``DEFAULT_TRUST_LEVEL`` so read-only callers don't 403 on a
+        # brand-new agent_id. Write paths still need identity pinning,
+        # though: this endpoint persists outcome and rule memories
+        # plus an audit-log entry keyed to ``caller_agent_id``, and
+        # without a registered row backing the name, attribution
+        # becomes unverifiable (a tenant-key holder could synthesise
+        # ``"admin-bot"`` and have that string appear in the audit
+        # trail). API-key admission proves authorization, not identity
+        # — re-block unregistered agents on writes specifically. The
+        # soft-pass remains useful for read-only consumers calling
+        # ``require_trust`` directly.
         _, not_found, terr = await require_trust(db, body.tenant_id, caller_agent_id, min_level=min_level)
         if not_found:
             raise HTTPException(
--- a/core-api/src/core_api/routes/health.py
+++ b/core-api/src/core_api/routes/health.py
@@ -9,14 +9,13 @@ from common.events.factory import get_event_bus
 from core_api.cache import redis_healthy
 from core_api.clients.storage_client import get_storage_client
 from core_api.config import settings
-from core_api.constants import VERSION
-from core_api.providers._platform import get_platform_init_errors
+from core_api.constants import PROBE_TIMEOUT_SECONDS, VERSION
+from core_api.providers._platform import (
+    get_platform_embedding,
+    get_platform_init_errors,
+    get_platform_llm,
+)
 from core_api.tools import REGISTRY  # SoT registry — populated at import time
-
-# Upper bound on a single dependency probe. Cloud Run typically gives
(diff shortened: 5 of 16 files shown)


Eldad-Caura:
Rebased + squashed onto current main in 4c12057 (force-pushed; conflicts resolved).

**Conflicts resolved**: pure whitespace conflict on ``memory_count_all`` in ``core-storage-api/services/postgres_service.py`` (both this branch and #27 added the same ``WHERE deleted_at IS NULL`` filter, just formatted differently). Kept main's single-line form since it's already canonical and matches ruff format.

**Auto-dropped during rebase** (patch contents already upstream via #55):
- ``fix(core-storage-api): ignore extra env-file keys`` (identical hunk in #55)
- ``style(core-storage): apply ruff format`` (already applied)

So the squash collapsed 8 commits → 1 (the 5 surviving + 2 auto-dropped + the soft-delete commit that's already in via #27 = original 8). Net diff against main: 13 files, +601/−66.

**Three logically related fixes folded in a single commit**:
1. Real ``tenant_count`` on ``/api/v1/stats`` (replaces hardcoded ``1``)
2. New ``/api/v1/status`` service-fingerprint endpoint
3. Trust-gate soft-pass for missing Agent rows (closes the brand-new-agent-403 bug on memclaw_list / insights / evolve)

Tests scaffold preserved: ``test_api_stats.py``, ``test_api_status.py``, ``test_trust_service_soft_pass.py``, plus inverted unknown-agent assertions in ``test_evolve_service.py`` and the ``test_integration.py`` coverage for the new ``/memories/distinct-tenants`` route.

Erni stays primary author; my Sign-off + Co-Authored-By added.

End-of-work workflow: fetched ``origin/main`` (edc79e3, current) and ``origin/CAURA-000-stats-tenant-count`` (fd412ad, the unsquashed pre-push) before the force-push so the lease compared against current state.

Eldad-Caura:
@claude

Eldad-Caura:
All three findings applied in 850d162.

**Medium — ``mode`` field removed from /status.** The reviewer was right that exposing ``mode = "oss" if settings.is_standalone else "enterprise"`` on a public unauthenticated endpoint directly leaks the tenant-auth-bypass opt-in state. Took option (a) — removed the field entirely from the response shape. The OSS/enterprise split is discoverable from a service's image tag and config in operator contexts that already have access; we don't owe it to anonymous callers. Updated ``test_api_status.py`` to drop the ``mode`` assertions and added a positive ``assert "mode" not in data`` so the field doesn't drift back in unnoticed.

**Low — ``memory_distinct_agent_count`` switched to ``get_read_session()``.** Matched the pattern the new ``memory_distinct_tenant_count`` already established. Both functions are pure ``SELECT COUNT(DISTINCT ...)`` so there's no reason to consume a write-pool connection.

**Suggestion — timeout-wrapped public_stats counts.** Refactored ``stats.py`` to use a small ``_bounded_count(label, coro)`` helper that wraps each storage call in ``asyncio.wait_for(..., _PROBE_TIMEOUT_SECONDS)`` (imported from ``health.py`` so the timeout posture is shared) and returns ``0`` on TimeoutError or any exception. A stalled storage backend can no longer hang public landing-page hits indefinitely. Logs a WARNING with the failed counter label + traceback for ops; the public response just returns ``0`` for the affected tile.

**Cleanup**: removed an unused ``import pytest`` in ``test_api_status.py`` that ruff F401 was flagging (was always unused; the project's conftest must auto-apply asyncio mode).

End-of-work workflow followed: fetched ``origin/main`` (edc79e3, current — my parent matches) and ``origin/CAURA-000-stats-tenant-count`` (4c12057, my previous push — no concurrent activity) before the force-push. Lease passed cleanly.

Single commit. Erni primary author.

Eldad-Caura:
@claude

Eldad-Caura:
4 of 5 findings applied + 1 skip-with-reasoning in 676d19a.

**High — Unregistered agent IDs persisted on write paths.** Restored ``not_found`` 403 in BOTH ``evolve.py`` and ``insights.py`` write endpoints. The soft-pass in ``trust_service.require_trust`` itself stays — it's still useful for read-only callers (memclaw_list, recall, the MCP tool wrappers that already use ``_, _, terr``) — but the two REST writes that persist memories + audit-log rows now reject brand-new agent_ids with a clear ``Agent 'X' is not registered in tenant 'Y'.`` 403. Test pair updated: ``test_evolve_rest_accepts_unknown_agent_with_default_trust`` and ``test_insights_rest_accepts_unknown_agent_with_default_trust`` renamed to ``..._rejects_unknown_agent`` and assert 403 + the expected detail substring. Operators register an agent by writing one memory first (the standard ``get_or_create_agent`` path), exactly like before.

**Medium — ``storage`` key on success — SKIP, finding doesn't apply.** Looked at ``_probe_dependencies`` line 73 in the current branch: ``result["storage"] = "connected"`` IS set on the success path (right after the ``await asyncio.wait_for`` succeeds, before the ``except``). Reviewer was reading an earlier intermediate commit — the current squash already has consistent storage-key presence. No change.

**Low — ``_PROBE_TIMEOUT_SECONDS`` cross-route private import.** Moved to ``core_api/constants.py`` as ``PROBE_TIMEOUT_SECONDS`` (public name, no underscore). Updated imports in ``health.py`` and ``stats.py``. No more cross-route private-symbol coupling.

**Low — Partial indexes for ``COUNT(DISTINCT)`` queries.** New migration **011** (``ix_memories_tenant_id_active`` and ``ix_memories_agent_id_active``) — both narrow partial indexes filtered to ``deleted_at IS NULL``, plus the ``agent_id IS NOT NULL`` predicate on the agent index to match the storage query. PostgreSQL can now satisfy both ``COUNT(DISTINCT tenant_id)`` and ``COUNT(DISTINCT agent_id)`` without a full live-rows scan on every landing-page hit. Existing ``ix_memories_tenant_created_active`` is on ``(tenant_id, created_at DESC, id DESC)`` so it doesn't help DISTINCT aggregates.

**Suggestion — ``uptime`` → ``health``.** Renamed in the ``/status`` response shape. The field carries the rolled-up health enum (``ok`` / ``degraded`` / ``unhealthy``), not a duration, so ``health`` is the accurate name. Updated test assertions; added a positive ``assert "uptime" not in data`` so the old name doesn't drift back. Inline comment in ``health.py`` explains the rename so anyone reading later understands the reason.

End-of-work workflow followed: fetched ``origin/main`` (edc79e3, current — my parent matches) and ``origin/CAURA-000-stats-tenant-count`` (850d162, my previous push — no concurrent activity) before the force-push.

Single commit. Erni primary author.

Eldad-Caura:
@claude

Eldad-Caura:
All three findings applied in b4e4c74.

**High — MCP write handlers bypass the not_found block.** Real catch — I'd added the ``not_found`` 403 to ``evolve.py`` and ``insights.py`` (REST routes) in the previous round but missed the MCP twins. Both ``memclaw_insights`` (~line 1119) and ``memclaw_evolve`` (~line 1186) in ``mcp_server.py`` now do the full three-tuple unpack and short-circuit on ``not_found`` with the ``"Error (403): Agent 'X' is not registered. Register the agent by writing one memory first."`` message. ``memclaw_list`` (~line 1006) deliberately stays on the soft-pass — it's the read-only handler the soft-pass exists to support.

**Medium — Migration 011 holds a write-blocking ShareLock.** Switched both index builds to ``CREATE INDEX CONCURRENTLY`` inside ``op.get_context().autocommit_block()``, mirroring the established pattern in ``009_restore_unscoped_memory_indexes.py`` (the repo uses raw ``op.execute`` rather than ``postgresql_concurrently=True`` — same end state, consistency wins). Added ``IF NOT EXISTS`` for idempotency against partial prior runs, and ``DROP INDEX CONCURRENTLY IF EXISTS`` on the downgrade path. Inline comment notes that without this, a fresh migration on a large table is minutes of write downtime.

**Low — Misleading deny message in ``require_trust``.** Reworded the missing-row + below-min-level error from ``Agent 'X' (trust_level=1) < required N`` to ``Agent 'X' is not registered (no row; effective trust 1 < required N). Register the agent by writing one memory first.`` Now distinguishes "no row" from a real ``trust_level`` denial so an operator looking up the 403 doesn't go hunting for a row to upgrade that doesn't exist. Updated ``test_trust_service_soft_pass.py`` assertions for both the ``min_level=2`` and ``min_level=3`` deny paths to match the new wording.

End-of-work workflow followed: fetched ``origin/main`` (edc79e3, current — my parent matches) and ``origin/CAURA-000-stats-tenant-count`` (676d19a, my previous push — no concurrent activity) before the force-push.

Single commit. Erni primary author.

Eldad-Caura:
@claude

Eldad-Caura (approved):

Eldad-Caura:
Both fixes applied in 1db62f1.

**Fix 1 — ``require_trust`` write-path warning.** Added a ``.. warning::`` block to the docstring spelling out that write paths MUST check ``not_found`` independently of ``error_str``, with the exact pattern from ``routes/evolve.py`` inlined as a code snippet. Calls out the asymmetry between ``memclaw_list`` (intentional soft-pass) and the write call sites (REST evolve/insights + MCP evolve/insights). No logic change — this is the documentation that should have shipped with the soft-pass in the first place. A future contributor adding a new write path that calls ``require_trust`` will see the warning at their cursor (most LSPs surface the docstring inline).

**Fix 2 — ``platform_init_errors`` key consistency.** Renamed the degraded-path key in ``/health`` from ``platform_provider_errors`` to ``platform_init_errors`` so both ``/health`` and ``/status`` (which already used ``platform_init_errors``) expose the result of ``get_platform_init_errors()`` under the same name. Inline comment notes the parity. No tests referenced the old key (grep was clean).

End-of-work workflow note: ``origin/main`` advanced to ``6d33caf`` (PR #58 — error contract standardization, +208 to mcp_server.py) between my last push and this one. Rebased onto current main; the new main commit didn't conflict with my ``not_found`` blocks on ``memclaw_evolve`` / ``memclaw_insights``, both still intact post-rebase. Verified before push.

Single commit. Erni primary author.

PR #55, opened 2026-05-03
feat(agents): split agent_id (opaque, stable) from display_name + plugin install_id (CAURA-000)

erni-a:
## Summary

Stop the OpenClaw plugin's default `"main"` agent from colliding on the
server's `(tenant_id, agent_id)` unique key when more than one install
shares a tenant. Splits agent identity into two fields with different
stability semantics:

| Field | Purpose | Stability | Format |
|---|---|---|---|
| `agent_id` | Internal unique key | Forever-stable | Opaque random — `main-${installId}` (8 hex chars) |
| `display_name` (new) | Human label in admin UIs | Mutable — tracks current hostname | `${shortHostname}-${baseName}` |
| `install_id` (new, indexed) | Per-plugin-install opaque suffix | Forever-stable | 8 hex chars, generated once |

Server-side migration `005_agent_display_name_install_id` adds both new
columns as nullable + indexes `install_id`. Old plugin versions (no
fields) keep working — the columns stay NULL, and the admin UI falls
back to `agent_id`.

Plugin-side, **`getInstallId()`** persists an 8-hex-char id to
`install.json` in the plugin data dir on first heartbeat.
**`getDisplayName(baseName)`** returns `${shortHostname}-${baseName}`,
where `shortHostname` strips FQDN tails (so a GCP VM's
`erni-openclaw.us-central1-c.c.alpine-theory-469016-c8.internal`
collapses to `erni-openclaw`) and respects a
`MEMCLAW_DISPLAY_NAME_OVERRIDE` env var for operators who prefer a
custom label.

## Commits

* `fix(core-storage-api): ignore extra env-file keys` — same one-line
  fix as on **#54**; harmless duplicate.
* `feat(agents): split agent_id (opaque, stable) from display_name` —
  Agent model + `005` migration + storage upsert merge keys.
* `feat(plugin): per-install id + hostname-prefixed display_name` —
  `install-id.ts`, `identity.ts`, heartbeat + resolve-agent wiring.
* `fix(plugin): ship install-id.ts + identity.ts in installer file
  lists` — caught live during E2E against a real GCP VM; without it,
  every customer upgrade hits `tsc` errors.
* `fix(plugin): trim FQDN + add MEMCLAW_DISPLAY_NAME_OVERRIDE` —
  polish from the same E2E (the GCP internal FQDN was 90 chars long).
* `fix(fleet): route heartbeat agent upserts through
  get_or_create_agent` — fixes a 500 in storage caused by the heartbeat
  upsert bypass-then-conflict path.

## Tests

* 121 plugin tests pass (`npm test` in `plugin/`):
  - `identity.test.ts` (8 tests) — sanitizer edge cases + override
  - `install-id.test.ts` (5 tests) — first-call persists, cached on
    subsequent, distinct-dirs-distinct-ids, malformed-regenerates
* 696 OSS Python unit tests pass on this branch
* **Live E2E against GCP VM `erni-openclaw`:**
  ```
  agent_id     = main-e2bce325
  display_name = erni-openclaw-main
  install_id   = e2bce325
  fleet_id     = teest
  ```
  Plugin source uploaded via reverse SSH tunnel + the install endpoint;
  heartbeat propagated all three new fields; `install.json` materialised
  on first tick.

## Behaviour change for upgrading plugin installs

Post-upgrade, the default agent's internal `agent_id` flips from
`"main"` → `"main-${installId}"`. The server sees this as a NEW agent
with no history. To preserve continuity with the pre-upgrade `"main"`
agent, set `agents.list[0].id = "main"` in the OpenClaw config and
remove the suffix.

CHANGELOG-equivalent line is in the commit body of
`feat(agents):`. Release-please will pick it up via the `feat` prefix.

## Migration

Apply migration `005_agent_display_name_install_id` (additive nullable
columns + index — purely forward-compatible). No data backfill needed.

## Coordination

Branched off `main`. Conflict-free with **#27**, **#42**, **#52**, and
**#54** (verified via `git merge-tree`). The fleet-id storage upsert
in `f4b1fd9` from earlier drafts was DROPPED during the rebase onto
`main` because `main` already has the better `INSERT … ON CONFLICT DO
NOTHING + re-SELECT` shape (good!).

### DIFF
--- a/common/models/agent.py
+++ b/common/models/agent.py
@@ -1,7 +1,7 @@
 import uuid
 from datetime import datetime
 
-from sqlalchemy import DateTime, Index, SmallInteger, Text, text
+from sqlalchemy import DateTime, Index, SmallInteger, String, Text, text
 from sqlalchemy.dialects.postgresql import JSONB
 from sqlalchemy.orm import Mapped, mapped_column
 
@@ -17,6 +17,18 @@ class Agent(Base):
     tenant_id: Mapped[str] = mapped_column(Text, nullable=False)
     fleet_id: Mapped[str | None] = mapped_column(Text)
     agent_id: Mapped[str] = mapped_column(Text, nullable=False)
+    # ``display_name``: human-readable label shown in admin UIs / dashboards
+    # / audit log columns. ``${hostname}-${baseName}`` from the plugin (e.g.
+    # ``johnsmith-laptop-main``). Mutable — heartbeats refresh it so a
+    # renamed machine propagates. Falls back to ``agent_id`` in the UI when
+    # NULL (older plugin versions don't send the field).
+    display_name: Mapped[str | None] = mapped_column(Text)
+    # ``install_id``: per-plugin-install opaque suffix that disambiguates
+    # the default ``"main"`` agent across multiple OpenClaw deployments
+    # sharing the same tenant. The plugin generates this once at first
+    # heartbeat and persists it to ``install.json`` in the plugin data dir;
+    # never rotates. Indexed because the admin UI groups agents by install.
+    install_id: Mapped[str | None] = mapped_column(String(32), index=True)
     trust_level: Mapped[int] = mapped_column(SmallInteger, nullable=False, server_default=text("1"))
     search_profile: Mapped[dict | None] = mapped_column(JSONB)
     created_at: Mapped[datetime] = mapped_column(
--- a/core-api/src/core_api/routes/fleet.py
+++ b/core-api/src/core_api/routes/fleet.py
@@ -1,11 +1,14 @@
 """Fleet heartbeat and command channel — replaces WebSocket/SSH gateway model."""
 
+import logging
 from datetime import UTC, datetime
 from uuid import UUID
 
 from fastapi import APIRouter, Depends, HTTPException, Query
-from pydantic import BaseModel
+from pydantic import BaseModel, Field
 from sqlalchemy.ext.asyncio import AsyncSession
+
+logger = logging.getLogger(__name__)
 
 from core_api.auth import AuthContext, get_auth_context
 from core_api.clients.storage_client import get_storage_client
@@ -50,6 +53,20 @@ class HeartbeatIn(BaseModel):
     tools: list | None = None
     channels: list | None = None
     metadata: dict | None = None
+    # ``install_id`` is the per-OpenClaw-install opaque suffix the plugin
+    # generates once at first heartbeat and persists locally. Used to
+    # disambiguate the default ``"main"`` agent across fleet installs
+    # so memories from different machines stop colliding on a single
+    # ``(tenant_id, agent_id="main")`` row. Optional — older plugin
+    # versions don't send it.
+    #
+    # ``max_length=32`` matches the ``agents.install_id`` column
+    # (``String(32)``); without this Pydantic constraint, an oversized
+    # value silently 422s at the storage layer in a per-agent
+    # exception handler that swallows the failure, leaving the row
+    # without an ``install_id`` and recreating the very collision
+    # this feature exists to fix. Reject at the API boundary instead.
+    install_id: str | None = Field(None, max_length=32)
 
 
 class CommandIn(BaseModel):
@@ -173,6 +190,7 @@ async def delete_fleet(
 async def heartbeat(
     body: HeartbeatIn,
     auth: AuthContext = Depends(get_auth_context),
+    db: AsyncSession = Depends(get_db),
 ):
     """Plugin pushes status; receives pending commands in response."""
     auth.enforce_tenant(body.tenant_id)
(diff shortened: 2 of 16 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@claude

Eldad-Caura:
All 4 review findings applied + branch squashed to a single commit (c0046a9, force-pushed).

**Medium — install_id entropy.** Bumped from 8 hex chars (32 bits) to 12 hex chars (48 bits) in ``plugin/src/install-id.ts:69``. ~0.003% birthday-paradox collision at 10k installs/tenant vs ~1.2% before. Test regexes updated in ``install-id.test.ts``. Migration column is ``String(32)`` so future bumps to the full UUID don't require a schema change.

**Medium — install_id overwrite guard at the storage layer.** Added an in-loop guard in ``postgres_service.agent_add``'s conflict path: when a row already has ``install_id`` set, the field is skipped (only ``display_name`` and ``fleet_id`` overwrite on change). The application-layer guard in ``agent_service.get_or_create_agent`` stays — duplicating the invariant defensively closes the gap for any future direct caller of ``agent_add`` that bypasses the service layer.

**Low — partial audit-log commit.** ``fleet.heartbeat`` now collects ``failed_agents: list[str]`` during the per-agent loop and emits a single summary warning *after* ``await db.commit()`` so the on-call can correlate partial state from the log: ``fleet.heartbeat: agent upsert failed for K/N agents in tenant=T: [agent_a, agent_c]``. Per-agent stack traces still log inline.

**Low — test-only export annotation.** Added ``@internal`` JSDoc on ``_resetInstallIdCacheForTesting`` plus a sentence in the docstring saying production code MUST NOT call it. Underscore prefix + tag are the lint-friendly signals.

**Suggestion deferred.** Server-side length cap on ``display_name`` (column is ``Text``, plugin override passes verbatim) — out of scope for the agent-identity split. Worth a small follow-up to truncate at e.g. 255 chars at the schema boundary.

**Squash.** Per repo convention, squashed the 8 incremental commits (split, plugin, file lists, FQDN trim, fleet heartbeat re-route, ruff format, migration re-chain) into one. Erni stays as the primary author; my Signed-off-by + Co-Authored-By added. Full chronology recoverable from this PR's review history.

Eldad-Caura:
All 4 review-2 findings applied in b8a251d.

**High — silent history orphaning on plugin upgrade.** Implemented option A from the reviewer's "better fix" in ``core_api.services.agent_service.get_or_create_agent``: when a fresh ``main-{install_id}`` row is created and a legacy ``"main"`` row exists for the same tenant/fleet, ``trust_level`` and ``search_profile`` carry forward into the new row. Bounded scope (only triggers for ``main-{install_id}`` ids, only when ``require_approval=False``). Legacy row stays intact so its memories remain queryable under ``agent_id="main"`` for admin recovery — operators decide later whether to delete or archive. Audit log includes ``carried_from_legacy_main: bool`` so the migration is observable per row. Also added a "Legacy main rows" section to the migration 010 docstring with operator recovery SQL (rename-in-place / archive variants).

**Medium — HeartbeatIn.install_id length.** Added ``Field(None, max_length=32)`` so an oversized value 422s at the API boundary instead of silently failing inside the per-agent exception handler downstream.

**Medium — unguarded db.commit() in heartbeat.** Wrapped the post-loop ``await db.commit()`` in try/except so a ``PendingRollbackError`` from a per-agent failure can't 500 the heartbeat and drop the pending-commands response. Best-effort observability, never break the contract.

**Low — atomic install.json write.** Changed to write-then-rename via ``.tmp`` sibling. ``rename(2)`` is atomic on POSIX same-filesystem; readers always see either the previous complete file or the new one, never a partial. Added orphan-tmp cleanup in the catch path.

---

**Process note + apology**: I momentarily overwrote your concurrent push (the mcp_server / documents / docs additions in 12df167) with a stale ``--force-with-lease``. The lease was comparing against my old ``origin/`` ref so it passed despite the remote having moved. Recovered immediately by hard-resetting to 12df167 (still in my git object DB), re-applying my 4 review fixes on top, and force-pushing the squash. The current b8a251d is the strict superset: 12df167 + my 4 review-2 fixes. ``git diff 12df167 b8a251d --stat`` shows the diff is exactly the 4 expected files, no other surprises.

Saving the lesson: always ``git fetch <branch>`` immediately before any force-push, not at the start of the work window.

Single commit, erni primary author, my Sign-off + Co-Authored-By preserved.

Eldad-Caura:
@claude

Eldad-Caura:
Both follow-up fixes applied in 4cc75eb.

**Fix 1 — Session poisoning recovery in heartbeat loop.** Added a best-effort ``await db.rollback()`` inside the per-agent except block (after the warning + ``failed_agents.append``). Clears any ``PendingRollbackError`` left on the SQLAlchemy session so subsequent agents in the same heartbeat batch can use the session normally instead of cascading rollback errors. Wrapped in its own try/except — if the rollback itself fails the session is still poisoned, but we can't do better than swallow and continue.

**Fix 2 — display_name length cap at the API boundary.** Truncates ``raw_dn`` to ``raw_dn[:255]`` before passing into ``get_or_create_agent``. Storage column is ``Text`` (unlimited) and ``MEMCLAW_DISPLAY_NAME_OVERRIDE`` passes verbatim from the plugin, so an oversized value would otherwise land in audit logs and UI rendering. 255 chars is comfortably above any realistic hostname-derived label and matches typical name-field conventions.

Single commit, clean trailers, erni primary author.

(End-of-work workflow note for me: fetched both ``origin/main`` (already current) and ``origin/CAURA-000-agent-display-name`` (matched local HEAD, no concurrent pushes) before pushing — incident from earlier this thread is now memory-tracked.)

Eldad-Caura (approved):

PR #69, opened 2026-05-04
fix(cache): include VECTOR_DIM in qemb query-embedding cache key (CAURA-644)

Eldad-Caura:
## Summary

Closes the cache-key vs schema-dim mismatch that surfaced during the v2.0.0 prod rollout (CAURA-641). After migration 012 widened the pgvector schema from `vector(768)` to `vector(1024)`, the existing `qemb2:` cache key didn't include the dim — pre-migration entries were retrieved by post-migration code and handed to SQL similarity search, which rejected them: `expected 1024 dimensions, not 768`.

Mitigated in prod by manually flushing `qemb2:*` (91 keys) via a one-shot Cloud Run Job. This PR closes the underlying bug.

## Change

`core-api/src/core_api/services/memory_service.py` — add `VECTOR_DIM` to the SHA-256 input of the `_get_or_cache_embedding` cache key:

```python
_qhash = hashlib.sha256(
    f"{_model}:{VECTOR_DIM}:{tenant_id}:{_normalized}".encode()
).hexdigest()
```

Prefix stays at `qemb2:` (namespace, not schema-version marker). Old-dim entries become unreachable by the new key shape and expire via `EMBEDDING_CACHE_TTL`.

## Why this won't regress

Two regression tests in `tests/test_qemb_cache_dim_versioning.py`:

1. **`test_qemb_cache_key_changes_with_vector_dim`** — same (query, tenant), different `VECTOR_DIM` → assert the cache keys differ. This is the locked-in property: a future schema-dim migration cannot leak stale-dim entries into the new schema.
2. **`test_qemb_cache_hit_returns_cached_value`** — sanity that the new key shape doesn't accidentally always-miss; a hit still returns the cached payload without a fresh embed.

## Audit of related cache sites

Grepped for any other site that builds an embedding-shaped cache key. Only one site: this function. Captured in CAURA-644's "done when" so it stays visible if a new cache site lands later.

## Test plan

- [x] `pytest tests/test_qemb_cache_dim_versioning.py -v` — both new tests pass
- [x] `pytest tests/pipeline/test_search_pipeline.py -k "cache or embed"` — existing pipeline tests still green
- [ ] CI green
- [ ] Cherry-picked / available on dev for the next staging+prod cut so the prod-side flush isn't required again

### DIFF
--- a/core-api/src/core_api/services/memory_service.py
+++ b/core-api/src/core_api/services/memory_service.py
@@ -34,6 +34,7 @@ except ImportError:
         pass  # type: ignore[misc]
 
 
+from common.constants import VECTOR_DIM
 from common.embedding import get_embedding, get_embeddings_batch, get_query_embedding
 from common.events import publish_memory_embed_request, publish_memory_enrich_request
 from core_api.constants import (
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
Both findings applied in `47bb586`.

**🟡 Medium — cache cold-start explicit:**
Bumped the prefix from `qemb2:` → `qemb3:` so the cache-generation boundary is visible in Redis key stats. Functionally identical to the previous version (any old `qemb2:*` entries are unreachable under both shapes since the hash input changed too), but operators can now see the cold-start at deploy time by watching `qemb3:*` key counts grow as `qemb2:*` ages out via `EMBEDDING_CACHE_TTL`. Added a comment above `_cache_key` explaining the bump.

The cold-start is still real on deploy — the working set has to be re-fetched from the embedding provider. Provider rate limits are within budget for the largest tenant we care about (text-embedding-3-small at 3000 RPM is comfortably above any realistic dashboard load), but worth a quick eye on the OpenAI dashboard during the next prod deploy.

**🟡 Low — docstring trimmed:**
Replaced the multi-paragraph incident-history block with the invariant only:

```
Get embedding from cache or generate it.

The cache key includes ``VECTOR_DIM`` so that a schema-dimension
migration doesn't surface stale cached embeddings to the new schema;
old entries with a mismatched dim become unreachable under the new
key and expire naturally via ``EMBEDDING_CACHE_TTL``.
```

CAURA-644/CAURA-641 ticket IDs, the 91-entry flush count, the v2.0.0 rollout date, and the alembic-012 migration number are still in the commit message and PR description where they belong.

Both regression tests still pass. Ruff format clean.

Eldad-Caura:
@claude

Eldad-Caura:
@claude

PR #71, opened 2026-05-04
fix(crystallizer): /latest 404→200 null on empty (CAURA-646)

Eldad-Caura:
## Summary

Fixes the API contract for `GET /api/v1/crystallize/latest` so it returns **200 with body `null`** when a tenant has no completed reports yet, instead of a 404.

## Why

The 404 conflated:

- **Resource doesn't exist** — the URL doesn't map to anything (genuine 404)
- **Empty state** — the resource exists as a concept; it just has no value yet

For a tenant-scoped "give me my latest of $X" endpoint, the URL is well-defined; only the optional value behind it is unset → 200 + `null` is correct. 404 stays for `/crystallize/reports/{report_id}` because *that* endpoint genuinely points at an opaque id that may not exist.

Real prod report today: a user hitting `/api/crystallize/latest?tenant_id=eldad` got a 404 in their network tab. Dashboard had to special-case it via `String(e).includes('404')` → `setLatest(null)`.

## Change

`core-api/src/core_api/routes/crystallizer.py`:

- Replace `raise HTTPException(404, ...)` with `return None`. FastAPI emits `null` body with 200.
- Expanded docstring spells out empty-state vs missing-resource so the next reader doesn't re-introduce the 404 thinking it's "more REST-y".

## Test

`tests/test_api_crystallizer.py::test_get_latest_report_empty_returns_200_null` — fresh tenant (per-test unique id), assert 200 + `json() is None`. Passes locally.

Existing `test_get_report_not_found` for `/reports/{id}` unchanged and still passes — that 404 is the genuine-missing-resource case.

## Test plan

- [x] `pytest tests/test_api_crystallizer.py::test_get_latest_report_empty_returns_200_null -v` — passes
- [x] `pytest tests/test_api_crystallizer.py::test_get_report_not_found -v` — still passes
- [x] `ruff check` + `ruff format --check` — clean
- [ ] CI green
- [ ] Frontend's `includes('404')` simplification (separate small PR if desired — existing branch still works since 200+null falls through `setLatest(data)` with `data=null`)

## Out of scope

Other `/latest`-shaped endpoints with the same pattern (e.g. `/security-audit/latest` if present, `/contradictions/latest`). Separate sweep — file as found.

### DIFF
--- a/core-api/src/core_api/routes/crystallizer.py
+++ b/core-api/src/core_api/routes/crystallizer.py
@@ -155,12 +155,27 @@ async def get_latest_report(
     tenant_id: str = Query(...),
     auth: AuthContext = Depends(get_auth_context),
 ):
-    """Get the most recent completed crystallization report for a tenant."""
+    """Get the most recent completed crystallization report for a tenant.
+
+    Returns ``200`` with the report body when a completed report exists;
+    returns ``200`` with body ``null`` when the tenant has none yet. The
+    URL itself is the well-defined "give me my latest report" resource —
+    the fact that no completed report exists yet is *empty state*, not a
+    missing resource. ``404`` would conflate the two and force every
+    client to special-case it as "actually empty"; see CAURA-646. The
+    sibling ``/crystallize/reports/{report_id}`` keeps its 404 because
+    *that* endpoint genuinely points at an opaque id that may not exist.
+    """
     auth.enforce_tenant(tenant_id)
     sc = get_storage_client()
     report = await sc.get_latest_report(tenant_id)
-    if not report:
-        raise HTTPException(status_code=404, detail="No completed reports found")
+    # Identity check, not truthiness — the storage client's contract is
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
Two fixes in `7e60d85`:

**CI failure** — the test was using `get_test_auth()`'s shared `"default"` tenant, which earlier tests in the same file populate with completed crystallize reports → my test got back a real report instead of `null`. Switched to a per-test unique `tenant_id` (admin API key bypasses `enforce_tenant`, so an arbitrary id routes cleanly without auth wiring). Test now passes locally and isolates correctly.

**🟡 Low — `is None` over `not report`** — applied. Tightened the guard to `if report is None:` so an empty-but-valid report dict (`{}`) from a hypothetical future storage-side change wouldn't silently null-return.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #83, opened 2026-05-05
refactor: rename tenant_settings → organization_settings (CAURA-654)

Eldad-Caura:
## Summary

- Storage rename only: `tenant_settings` / `tenant_settings_audit` → `organization_settings` / `organization_settings_audit`. Settings are conceptually per-organization (admin UI is org-scoped); the legacy per-tenant table holds 0 rows in prod, so the migration is a clean drop+recreate.
- `org_id` column intentionally typed as `text` with NO foreign-key constraint — pure-OSS deployments don't have an `organizations` table (enterprise-only concept). The `text` shape works in both: OSS-standalone passes `tenant_id` as the implicit org-key; enterprise passes the real org_id.
- ~22 lazy-import call sites updated mechanically to point at `core_api.services.organization_settings`. Public function parameter names stay `tenant_id` for back-compat; renaming those is a follow-up that touches every call site (out of scope here to keep the diff scoped).

## Changes

- **Migration 014** (`core-storage-api/src/core_storage_api/database/migrations/versions/014_rename_tenant_settings_to_organization_settings.py`): drop legacy tables + indexes, create new shape, downgrade restores legacy.
- **Model**: `common/models/tenant_settings.py` → `organization_settings.py`; classes `TenantSettings(Audit)` → `OrganizationSettings(Audit)`; column `tenant_id` → `org_id`.
- **Service**: `core-api/src/core_api/services/tenant_settings.py` → `organization_settings.py`; `ResolvedConfig.__init__(tenant_settings=…)` → `org_settings=…` (verified zero keyword call sites); cache log lines re-keyed to "organization_settings cache". Internal queries updated to `OrganizationSettings.org_id`.
- **22 consumers** in `core-api/` had their import path flipped (mechanical).
- **Test**: `test_tenant_settings_storage.py` → `test_organization_settings_storage.py` with raw-SQL `tenant_settings_audit` queries updated to `organization_settings_audit`.
- One stale docstring reference in `common/events/memory_enrich_publisher.py` updated.

## Why this shape (not a bigger refactor)

Two things deliberately stay as today's shape:

1. **Function parameters keep the name `tenant_id`**. Renaming them to `org_id` requires updating ~20 call sites that pass `auth.tenant_id`, plus ensuring `auth.org_id` is populated everywhere (including OSS-standalone middleware). Doing it here would balloon the PR without changing behavior; the empty prod table makes the semantic difference invisible today. Landing the rename now sets up the storage; the call-site migration is a follow-up.
2. **`org_id` is `text`, no FK**. The `organizations` table is enterprise-only — pure-OSS migrations don't create it. A FK would break OSS-standalone deployments at migration time.

## Test plan

- [ ] CI passes (ruff, mypy, pytest, alembic upgrade head from a fresh DB).
- [ ] `pytest tests/test_organization_settings_storage.py` runs green — covers cache lifecycle, audit row creation, no-op-skip, ResolvedConfig fallbacks.
- [ ] Manual smoke after deploy: `GET /settings` returns the merged display view; `PUT /settings` writes a row in the new `organization_settings` table.

## Out of scope

- Function parameter rename `tenant_id` → `org_id` across ~20 call sites — separate follow-up.
- FastAPI route handler names `get_tenant_settings` / `update_tenant_settings` — cosmetic; HTTP path stays /settings.
- `auth.org_id` propagation end-to-end (so call sites pass the org-key directly).

Closes CAURA-654.

### DIFF
--- a/common/events/memory_enrich_publisher.py
+++ b/common/events/memory_enrich_publisher.py
@@ -38,7 +38,7 @@ async def publish_memory_enrich_request(
     ``tenant_config`` is a ``ResolvedConfig``-shaped object (or ``None``).
     The publisher reads the enrichment-relevant attributes off it and
     packs them into the payload so the worker doesn't need to import
-    ``core_api.services.tenant_settings``.
+    ``core_api.services.organization_settings``.
 
     ``agent_provided_fields`` lists the row's columns the agent set
     explicitly at write time. The worker uses it to skip overwriting
--- a/common/models/organization_settings.py
+++ b/common/models/organization_settings.py
@@ -1,4 +1,11 @@
-"""Per-tenant settings storage: live overrides + append-only audit trail."""
+"""Per-organization settings storage: live overrides + append-only audit trail.
+
+Renamed from ``tenant_settings`` in CAURA-654. The keying column is
+``org_id`` (text); semantically the value is the organization
+identifier in enterprise deployments and the tenant identifier
+acting as an implicit single-org key in OSS-standalone deployments
+(no ``organizations`` table in pure OSS).
+"""
 
 from datetime import datetime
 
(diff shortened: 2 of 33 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #88, opened 2026-05-05
chore(embedding): remove broken Vertex embedding provider (CAURA-333)

ran-taig:
## Summary

- `VertexEmbeddingProvider` never passed `output_dimensionality` to the SDK, so every write returned the model's native dim (768/3072) and was rejected by pgvector's 1024-dim column. The provider was effectively DOA.
- Path is unreachable from the tenant UI today (`PROVIDER_OPTIONS["embedding"]` only lists `openai`) and unset in every deploy config — this is dead code, not a live regression.
- Rather than wire `output_dim` through and own a configuration we don't use, remove the in-tree provider. OSS users wanting non-OpenAI embeddings can supply their own `EmbeddingProvider` subclass at their own risk; the `VECTOR(VECTOR_DIM)` schema constraint still applies.

## Changes

- Delete `common/embedding/providers/vertex.py`
- `common/embedding/_platform.py`: drop the vertex branch + GCP plumbing; keep a warn-and-noop for legacy `PLATFORM_EMBEDDING_PROVIDER=vertex` so an upgrading operator gets a clear log line instead of silent fallback
- Drop `VERTEX_EMBEDDING_MODEL` constants and `platform_embedding_gcp_*` settings fields
- Sharpen the existing `_remap_deprecated_vertex` shim message (CAURA-333 reason)
- `.env.example` / `README.md`: remove vertex embedding env vars, point OSS users to OpenAI-compatible providers
- Tests: replace the two vertex-singleton tests with one that asserts the `vertex` key is rejected; drop the protocol-conformance test for the deleted class

## Test plan

- [x] `pytest tests/test_provider_protocols.py tests/test_platform_providers.py tests/test_api_status.py` — 105 passed
- [x] `ruff check` clean on all touched modules
- [ ] CI green
- [ ] Manual: `PLATFORM_EMBEDDING_PROVIDER=vertex` boots with a loud warning and `_platform_embedding is None` (test covers this path)


ran-taig:
@claude

arkash20 (approved):

PR #96, opened 2026-05-06
fix(core-api): read version from installed package metadata

Eldad-Caura:
## Summary

- `/api/v1/version` was returning **2.0.0** in production *after* the v2.1.0 release. Root cause: the bare-string `"VERSION"` entry in `release-please-config.json`'s `extra-files` expects an inline `x-release-please-version` marker comment, which a one-line plain-text file can't carry — release-please silently skipped the bump every release. Every other extra-file (typed entries with explicit `jsonpath`) bumped correctly.
- Make `pyproject.toml` the single source of truth: `core_api.constants.VERSION` now reads from `importlib.metadata.version("core-api")`, which `pip install` materialises from `pyproject.toml`'s `[project] version`. release-please already manages that, so the version flows from the SoT to the runtime with no coordination point.
- Cleans up the fallout from removing the `VERSION` file:
  - `scripts/bump-version.sh` retired — release-please now owns the `plugin/package.json` + `plugin/openclaw.plugin.json` bumps.
  - `scripts/gen-version.sh` rewired to read from `plugin/package.json` (still produces `plugin/src/version.ts` for `PLUGIN_VERSION`).
  - `docker-compose.dev.yml` no longer bind-mounts the deleted file.
  - `.github/ISSUE_TEMPLATE/bug_report.yml` points users at `curl /api/v1/version`.
- Regenerated `plugin/src/version.ts` from current `package.json` (`2.0.0` → `2.1.0`) so HEAD reflects what `npm run build` would produce post-merge.

## Test plan

- [x] `ruff check core-api/src/core_api/constants.py` — passes
- [x] `ruff format --check core-api/src/core_api/constants.py` — clean
- [x] `mypy core-api/src/core_api/constants.py --follow-imports=silent` — no issues
- [x] `pytest tests/test_api_status.py tests/test_api_plugin.py` — 19 passed (covers the `/api/v1/status`/`/api/v1/version` surfaces that read VERSION)
- [x] `bash scripts/gen-version.sh` produces `// Auto-generated from plugin/package.json — do not edit\nexport const PLUGIN_VERSION = "2.1.0";`
- [ ] After merge: confirm release-please opens a fresh release PR (this should be a `fix:` patch bump from 2.1.0 → 2.1.1)
- [ ] After 2.1.1 ships + AR retag + Cloud Run deploy: `curl https://memclaw.dev/api/version` and `https://memclaw.net/api/version` return `2.1.1`

## Notes

- No fallback for `importlib.metadata.PackageNotFoundError`. Correct behaviour for a service that's always `pip install`-ed (including editable installs in dev) — fail loud at import beats serving a bogus version.
- The package-name string `"core-api"` is now a hardcoded literal that must match `core-api/pyproject.toml`'s `[project] name`. A rename without updating both would surface as a `PackageNotFoundError` at import — caught by `tests/test_api_status.py` in CI.


Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

Eldad-Caura:
Round 1 fix applied: wrapped `importlib.metadata.version("core-api")` in `try/except importlib.metadata.PackageNotFoundError` with `VERSION = "dev"` fallback for source-only dev runs (PYTHONPATH set without `pip install -e`). Container builds + production deploys are unaffected — pip install puts the metadata in place. Pushed as a `--force-with-lease` amend.

@claude

erni-a (approved):

PR #104, opened 2026-05-07
fix(core-api): align write/query embedding surface (CAURA-222)

ran-taig:
## Summary

Removes the hint-prefix re-embed in `parallel_embed_enrich.py` so writes persist the **same surface** the search path embeds.

The write path was calling `compose_embedding_text(data.content, enrichment.retrieval_hint)` — which prepends `\"[Retrieval hint]: <hint>\\n\\n\"` to the content — and overwriting the content-only vector. The search path embeds the raw query through `get_query_embedding`. Same model (text-embedding-3-small / 1536d), different input — so identical content↔query produced cosine **0.693** instead of ~1.0.

## Evidence

algo-stress run \`20260507T153129Z-76c172\` on staging tenant \`ran-507cb7\`. The probe writes a fixed string then immediately searches for the byte-identical string; top hit IS the just-written memory but cosine = 0.693.

A single defect explains three independent measured regressions, all of which bottom out on a cosine threshold:

| metric | before | target |
|---|---:|---:|
| `embed.stability_top1_similarity` | 0.693 | ≥ 0.95 |
| `dedup.paraphrase.detected_rate` | 0/80 | ≥ 0.85 |
| `ranking.recall@20` | 0.512 | ≥ 0.80 |
| `ranking.ndcg@10` | 0.363 | ≥ 0.60 |
| `dispatcher.entity_lookup.recall` | 0.20 | ≥ 0.70 |

When stored vectors lived on the hint-prefixed surface and query vectors lived on the raw surface, every threshold (semantic_dedup floor, dispatcher entity gating, search `min_similarity` / blend weights) was being evaluated on a corrupted axis. Restoring surface symmetry restores the full ~[0,1] cosine dynamic range and lifts all three components together.

## Change

`core-api/src/core_api/pipeline/steps/write/parallel_embed_enrich.py`: removed lines 105–127 (the post-enrichment hint re-embed block). Writes now persist the content-only embedding produced at line 61.

No changes to:
- `min_similarity` / dedup floor / blend weights — root fix, not a threshold mask.
- Embedding model — `text-embedding-3-small` / 1536d preserved per CLAUDE.md.
- Schema, endpoints, or abstractions.
- The `retrieval_hint` enrichment field, prompt, validator, or `compose_embedding_text` helper — they stay as metadata and as a future building block (see *Future work* below).

## What I checked (other paths preserving the issue)

Audited every `get_embedding` / `provider.embed` / `compose_embedding_text` site in `core-api`, `core-worker`, `core-storage-api`, and `common`:

**Symmetric (raw content/query) — OK:**
- \`parallel_embed_enrich.py:61\` (after fix)
- \`memory_service.py:526\` (memory write)
- \`memory_service.py:1654\` (\`_reembed_after_delay\` background retry)
- \`memory_service.py:2280\` (manual update path)
- \`core-worker/consumer.py:158\` (async EMBED_REQUESTED backfill — embeds \`request.content\` raw)
- \`core-storage-api/scripts/backfill_embeddings.py:217\` (CLI backfill)
- \`mcp_server.py:846/932\`, \`routes/documents.py:280\` (document write/search both use \`get_embedding\` — symmetric pair on the document collection)
- \`memory_service.py:2594\` (\`get_query_embedding\` — symmetric provider falls through to \`embed\`)

**Still asymmetric — known follow-ups (same root pattern, intentionally out of scope for this PR per \"single-line fix at root\"):**
- \`memory_service.py:2030–2046\` — SaaS-mode background hint re-embed inside \`_enrich_memory_background\`. Fires when \`embed_on_hot_path=False\`. Will overwrite the content-only embedding with \`compose_embedding_text(content, hint)\` and re-introduce the asymmetry for SaaS deployments.
- \`memory_service.py:2095–2123\` — atomic-fact fan-out. Each child memory's stored embedding is \`compose_embedding_text(fact_content, fact_hint)\`, so atomic-fact children carry the asymmetry even after this fix.

These should be fixed in a follow-up PR (CAURA-222-followup) so SaaS-mode and atomic-fact children land on the same surface as the OSS hot path. The fix is the same shape — drop the \`compose_embedding_text\` wrap and embed raw content. They are split out to keep this PR minimal and easy to revert.

## Future work — re-introducing the hint mechanism *symmetrically*

The hint mechanism's intent was correct: bridge the **semantic gap** between content vocabulary (\"I signed a contract with my first client\") and query vocabulary (\"business milestone\"). The bug was that only the write side got the prefix.

A symmetric reintroduction has at least three viable shapes — each is a separate design decision:

1. **Two-vector store** — keep the raw \`embedding\` column for surface-form recall, add a \`hint_embedding\` column populated from \`compose_embedding_text(content, hint)\`. Search ranks against both and takes \`max\` (or a weighted blend). Pros: preserves both axes, no query-side changes needed. Cons: schema migration, doubles storage and embedding cost per memory.
2. **Query-side hint synthesis** — at query time, ask a small fast LLM to expand the query into both raw and significance-framed forms, embed both, max over candidates. Pros: no schema change, write path stays raw. Cons: per-query LLM latency on the hot search path, cache invalidation complexity.
3. **Decoder-symmetric framing** — instead of a prefix marker, use the same instruction-aware encoder protocol the project already has (\`get_query_embedding\` / \`InstructionAwareEmbedder\`). Document side encodes with a *passage* instruction, query side with a *query* instruction. Requires moving to an instruction-aware embedding model, which is gated by CLAUDE.md's \`text-embedding-3-small\` mandate.

A future PR rebuilding this should pick one explicitly, document the choice, and add an algo-stress test that asserts \`embed.stability_top1_similarity ≥ 0.95\` so this regression cannot return.

## Test plan

- [ ] CI passes on this branch.
- [ ] Existing tests \`tests/test_retrieval_hint.py\` still pass (they cover \`compose_embedding_text\`, the validator, and prompt rules — all unchanged).
- [ ] Existing tests \`tests/test_embed_off_hot_path.py\` reviewed: cases that asserted the hot-path hint-enhanced overwrite (e.g. \"hint re-embed on hot path overwrites raw embedding\") need updating to assert raw-content embedding instead. Reviewer please flag if I should include those test edits in this PR.
- [ ] Deploy to staging \`memclaw.dev\`.
- [ ] Drain tenant \`ran-507cb7\`: \`python -m scripts.drain_tenant\`.
- [ ] Re-run algo-stress: \`python -m runners.all --only embed_stability --only dedup --only search_ranking --only query_classify\`.
- [ ] Verify metric table targets above are met.
- [ ] Verify \`dedup.topical_similar\` / unrelated **false-merge stays at 0%** — removing the prefix moves to a stricter axis but I want this confirmed empirically before merge.

## Rollout

- Backwards-compatible at the storage layer: existing rows still have hint-prefixed vectors; new writes produce raw-content vectors. They will coexist and the recall floor will lift gradually as fresh writes accumulate.
- For a clean recall lift across the full corpus, run \`core-storage-api/scripts/backfill_embeddings.py\` after merge to re-embed historical rows on the raw surface. This is optional and can be done lazily.

---

### DIFF
--- a/core-api/src/core_api/config.py
+++ b/core-api/src/core_api/config.py
@@ -82,10 +82,6 @@ class Settings(BaseSettings):
     # breaching the 45s outer request budget. Must stay below
     # ``request_timeout_seconds`` so this fires first.
     enrichment_inline_timeout_seconds: float = 35.0
-    # Inner timeout for the optional hint-based re-embed roundtrip after
-    # enrichment lands a retrieval_hint. Pure quality-vs-latency knob;
-    # only fires when ``embed_on_hot_path`` is True.
-    enrichment_hint_reembed_timeout_seconds: float = 10.0
     # Per-call timeout passed to the AsyncOpenAI client (covers both LLM
     # enrichment and embedding providers). Without an explicit value the
     # SDK rides httpx's default — long enough that a single hung upstream
--- a/core-api/src/core_api/pipeline/steps/write/parallel_embed_enrich.py
+++ b/core-api/src/core_api/pipeline/steps/write/parallel_embed_enrich.py
@@ -12,10 +12,13 @@ enrichment call too. The strong-write pipeline persists the row with
 agent-provided values + schema defaults for the LLM-derived columns
 (``memory_type`` / ``weight`` / ``status``); ``ScheduleBackgroundTasks``
 publishes ``Topics.Memory.ENRICH_REQUESTED`` and ``core-worker``
-PATCHes the enrichment fields back. Hint-based re-embed is also
-deferred — when both flags are off, we accept the small recall hit on
-the initial embed and let the back-channel ENRICHED consumer trigger
-a follow-up re-embed if the hint is high-value.
+PATCHes the enrichment fields back. Hint-based re-embed is DISABLED
+(CAURA-222): writes used to embed
+``compose_embedding_text(content, retrieval_hint)`` while queries embed
+raw text, producing a write/query surface asymmetry that capped recall
+across dedup, entity-lookup, and search ranking. Until a symmetric
+reintroduction lands, both the hot-path here and the background
+re-embed in ``_enrich_memory_background`` embed raw ``content``.
 """
 
 from __future__ import annotations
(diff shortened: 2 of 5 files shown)


ran-taig:
@claude

ran-taig:
@claude

ran-taig:
@claude

ran-taig:
@claude

ran-taig:
@claude

Eldad-Caura (approved):

PR #105, opened 2026-05-08
feat(stats): add include_deleted option to memclaw_stats / /memories/stats

yanki-m:
## Summary

- Adds an opt-in `include_deleted=true` flag to both the MCP `memclaw_stats` tool and the REST `GET /api/v1/memories/stats` endpoint.
- Default behavior unchanged: `total` and the `by_type`/`by_agent`/`by_status` breakdowns continue to exclude soft-deleted rows so `/memories/stats.total` keeps matching what `/memories` returns.
- When opted in, the response also carries `deleted` (count of soft-deleted rows under the same scoping filters) and `total_including_deleted = total + deleted`.

## Why

Per a recent map of count APIs in this repo, every memory-count endpoint hard-filters `deleted_at IS NULL` with no way to surface the soft-deleted population. That's the right default (counts should match the queryable footprint), but operators occasionally want both numbers — e.g. dashboard widgets reconciling soft-delete activity, audits, or debugging cleanup pipelines.

## Shape

```jsonc
GET /api/v1/memories/stats?tenant_id=foo
// → { total, by_type, by_agent, by_status }   (unchanged)

GET /api/v1/memories/stats?tenant_id=foo&include_deleted=true
// → { total, by_type, by_agent, by_status, deleted, total_including_deleted }
```

Breakdowns deliberately stay non-deleted (aligned with `total`). Scope filters (`tenant_id`, `fleet_id`, `agent_id`+visibility, `memory_type`, `status`) apply equally to the live and deleted counts.

## Storage-api fallback

The REST route's storage-api fallback can't return a soft-deleted count, so `include_deleted=true` now joins `agent_id`/`memory_type`/`status` in the "filter combos that 503 instead of returning a degraded answer" guard — silently dropping a field the caller explicitly asked for would lie to them.

## Documentation

Second commit on this branch updates the three agent- and integrator-facing references in this repo:

- `plugin/skills/memclaw/SKILL.md` — tool card lists `include_deleted` in the signature and the additional response fields.
- `static/docs/integration-guide.md` — tools table cell extended.
- `docs/api-surfaces.md` — REST/MCP charter row notes the opt-in.

The companion enterprise frontend doc updates (for-agents page, dashboard MCP playground, docs page) are in caura-ai/caura-memclaw-enterprise (separate PR).

## Test fixtures

Updated three tool-surface baselines for the new description and `include_deleted` schema entry: `tools_list_baseline_v1.json`, `tool_descriptions_baseline_v1.json`, `tool_descriptions_enriched_baseline_v1.json`. Token budget: 4217 / 5000 (well under ceiling).

Live `mcp.list_tools()` was diffed against the updated baseline — exact match.

## Test plan

- [x] Default GET `/memories/stats` returns the original 4-key shape (no extra fields).
- [x] `?include_deleted=true` returns `deleted` and `total_including_deleted`.
- [x] Invariant `total_including_deleted == total + deleted` holds across no-filter, `memory_type=episode`, `memory_type=task`, `status=active`.
- [x] Soft-delete drives the `deleted` count from 0 → 1 as expected.
- [x] Live MCP tool registry matches `tools_list_baseline_v1.json` byte-for-byte.
- [x] Service-layer call exposes the same shape (`compute_memory_stats(..., include_deleted=True)`).


ran-taig:
@claude

PR #107, opened 2026-05-08
feat(fleet-stats): add memory-status totals to fleet_summary

yanki-m:
## Summary
- Extends `fleet_agent_stats` with `total_memories`, `conflicted_memories`, `outdated_memories`, `deleted_memories`, and `recalled_memories_24h` (distinct memories with `last_recalled_at` within 24h).
- Folds the existing `memories_24h` count into the same `COUNT(*) FILTER (...)` query so we still issue one scan over `memories` per fleet-stats request instead of adding a second.
- Powers the new General + Last 24 hours rows in the OpenClaw Fleets browser (frontend PR in `caura-memclaw-enterprise`).

## Test plan
- [ ] `GET /api/v1/storage/fleet/stats?tenant_id=...&fleet_id=...` returns the five new keys with sensible values for a tenant that has memories in mixed statuses.
- [ ] Empty-tenant case still returns zeros for every new field (verified locally against the dev container).
- [ ] No regression in `memories_24h` (same predicate as before, just inside the FILTER).

### DIFF
--- a/core-storage-api/src/core_storage_api/services/postgres_service.py
+++ b/core-storage-api/src/core_storage_api/services/postgres_service.py
@@ -2805,15 +2805,25 @@ class PostgresService:
                     }
                 )
 
-            # Memories in last 24h
-            result24 = await session.execute(
+            # Memory totals + status breakdown (single query, one table scan).
+            # ``deleted_at IS NULL`` keeps live rows separate from soft-deleted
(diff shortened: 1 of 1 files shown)


ran-taig:
@claude

ran-taig:
@claude

PR #108, opened 2026-05-10
chore(core-api): CAURA-222 follow-up — drop dead helper, add stability test, extend backfill

ran-taig:
## Summary

Three small follow-ups to PR #104 (CAURA-222 write/query embedding-surface alignment), bundled into one PR. None change production logic — they close the loop on cleanup, regression coverage, and operational tooling that was intentionally deferred to keep #104 minimal.

1. **Delete `compose_embedding_text` and unwire re-exports.** Zero production call sites since #104.
2. **Add `tests/test_embed_stability.py` regression suite.** Pins write/query surface symmetry so the asymmetry can't return silently.
3. **Extend the embedding-backfill CLI with `--rewrite-hint-prefixed`.** One-shot recall-recovery mode for tenants whose existing rows were embedded under the old hint-prefixed write path.

## 1. Drop `compose_embedding_text`

After #104 removed every production call site, the helper, its re-exports, and the `TestComposeEmbeddingText` class were dead weight. Removed:

- `common/enrichment/service.py` — function definition.
- `common/enrichment/__init__.py` — import + `__all__` entry.
- `core-api/src/core_api/services/memory_enrichment.py` — re-export.
- `tests/test_retrieval_hint.py::TestComposeEmbeddingText` — the class.
- `tests/test_embed_off_hot_path.py` — leftover mock that was a no-op since #104.

Kept on purpose:

- The `retrieval_hint` LLM field, prompt validator (\`tests/test_retrieval_hint.py::TestValidatorRetrievalHint\` and \`TestPromptContainsHintRule\` stay).
- `metadata.retrieval_hint` persistence — auditability and the selector for the backfill mode below.
- The historical-context comments in \`parallel_embed_enrich.py\` and \`memory_enriched.py\`.

If the symmetric reintroduction work (the future PR I drafted as part of #104) goes with Option A (two-vector store), it'll add a new helper rather than reviving this one — the old prefix shape is documented as the failure mode in the test suite below, not as a building block.

## 2. Regression test — `tests/test_embed_stability.py`

Three tests, deterministic spies on the embedding providers — no network, no DB, no CI infra dependencies:

| Test | Pins |
|---|---|
| `test_hot_path_embeds_raw_content_when_enrichment_yields_hint` | `ParallelEmbedEnrich` calls embedder exactly once with raw content even when enrichment yields a `retrieval_hint`. Catches re-introduced hint prefixes on the write hot path. |
| `test_atomic_fact_children_embed_raw_fact_content` | Atomic-fact fan-out children pass `fact.content` (not `compose_embedding_text(fact_content, fact_hint)`) to the embedder. |
| `test_write_and_query_embed_identical_input_for_identical_text` | **Load-bearing**: write and search paths feed the embedder byte-identical strings for byte-identical content/query. Any future divergence — re-introduced prefix, document-side instruction without a query-side counterpart, normalization on one side — trips this test. |

The third test is the one I'd want to see in any future PR that touches embedding text construction. It mocks `get_embedding` (write side) and `get_query_embedding` (search side) with recording spies, exercises both paths with the same input string, and asserts the recorded inputs are equal.

## 3. Backfill CLI — `--rewrite-hint-prefixed`

New mode in \`core-storage-api/src/core_storage_api/scripts/backfill_embeddings.py\` for one-off recall recovery on tenants that already have rows embedded under the old surface.

**Selector:**
```sql
WHERE embedding IS NOT NULL
  AND metadata_ ? 'retrieval_hint'
  AND COALESCE(metadata_->>'retrieval_hint', '') <> ''
  AND deleted_at IS NULL
```

Default IS-NULL scan path unchanged — the new flag branches inside `_iter_rows`. Entities are skipped silently in this mode (no hint metadata exists there). Per-tenant phasing via `--tenant-id` and `--dry-run` still apply.

**Usage:**
```
# Dry-run scope estimate per tenant:
docker compose run --rm core-storage-api \
  python -m core_storage_api.scripts.backfill_embeddings \
    --rewrite-hint-prefixed --tenant-id tenant-abc --dry-run

# Actual rewrite:
docker compose run --rm core-storage-api \
  python -m core_storage_api.scripts.backfill_embeddings \
    --rewrite-hint-prefixed --tenant-id tenant-abc
```

Three new tests cover this:
- `test_rewrite_hint_prefixed_targets_memories_with_hint_metadata` — asserts the SQL filter shape.
- `test_rewrite_hint_prefixed_skips_entities` — entities table is not even queried.
- `test_default_mode_uses_null_embedding_filter` — pins the default mode's WHERE shape so the new branching can't silently break the original CLI.

**Caveats called out in the existing module docstring still apply:**
- Standalone CLI runs against the process-level embedding provider; it does NOT honour per-tenant overrides. Multi-tenant prod with per-tenant overrides should still use the event-driven `core-worker backfill-embeddings` task. `--tenant-id` here scopes the SQL scan, not the provider resolution.
- A re-run of `--rewrite-hint-prefixed` will re-match the same rows because `metadata.retrieval_hint` is intentionally preserved (auditability). Operational pattern is one forward pass per tenant, then the embed-stability probe to verify, then move on.

## Test plan

- [x] `pytest tests/test_embed_stability.py tests/test_embed_off_hot_path.py tests/test_retrieval_hint.py tests/test_backfill_embeddings.py` — 43 passed locally.
- [x] `ruff format --check` clean on all touched paths.
- [x] `ruff check` clean on all touched paths.
- [ ] CI green.
- [ ] After merge: dry-run the new backfill mode against staging \`memclaw.dev\` tenant \`ran-507cb7\` to validate the SQL selector against real metadata shape; expected scope = paragraph-or-longer memories that were enriched on a deploy where the hint re-embed fired.

## Rollout

Independent of any production deploy — this PR is purely additive (regression test + cleanup + new opt-in CLI flag). Default backfill behavior, write path, and search path are all byte-identical to post-#104 main.

The recall recovery itself (running `--rewrite-hint-prefixed` against existing data) is a separate operational decision, owner-driven per-tenant, with the dry-run cost gate built into the existing CLI.

### DIFF
--- a/common/enrichment/__init__.py
+++ b/common/enrichment/__init__.py
@@ -7,9 +7,6 @@ Public surface:
 * :func:`enrich_memory` — async entrypoint with 3-tier fallback.
 * :func:`fake_enrich` — keyword heuristic fallback used as the
   ``"fake"`` provider and last-resort safety net.
-* :func:`compose_embedding_text` — content + retrieval hint composer
-  used by both the write-path embedding call and the post-enrichment
-  re-embedding step.
 * Constants: :data:`MEMORY_TYPES`, :data:`MEMORY_STATUSES`,
   :data:`DEFAULT_MEMORY_TYPE`, :data:`DEFAULT_MEMORY_WEIGHT`.
 
@@ -35,7 +32,6 @@ from common.enrichment.constants import (
 )
 from common.enrichment.schema import AtomicFact, EnrichmentResult
 from common.enrichment.service import (
-    compose_embedding_text,
     enrich_memory,
     fake_enrich,
 )
@@ -47,7 +43,6 @@ __all__ = [
     "MEMORY_TYPES",
     "AtomicFact",
     "EnrichmentResult",
-    "compose_embedding_text",
     "enrich_memory",
     "fake_enrich",
 ]
--- a/common/enrichment/service.py
+++ b/common/enrichment/service.py
@@ -4,9 +4,6 @@
 Public surface:
 
 * :func:`enrich_memory` — async entrypoint with 3-tier fallback.
-* :func:`compose_embedding_text` — helper used both pre- and post-
-  enrichment to assemble the text fed to the embedding model
-  (content + retrieval hint).
 * :func:`fake_enrich` — keyword-heuristic fallback used both as the
   ``"fake"`` provider and as the last-resort safety net.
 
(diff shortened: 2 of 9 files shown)


ran-taig:
@claude

ran-taig:
@claue

ran-taig:
@claude

Eldad-Caura (approved):

PR #159, opened 2026-05-18
CAURA-123 fix(contradiction): activate dead RDF triple path end-to-end

ran-taig:
## Summary

The RDF (deterministic, no-LLM) contradiction path was effectively dead. Three latent gaps stacked:

1. **No emission.** Ingestion never set `subject_entity_id` / `predicate` / `object_value`, so the detector's RDF branch ([`contradiction_detector.py:198`](../blob/main/core-api/src/core_api/services/contradiction_detector.py#L198)) was gated off every memory. Every contradiction — including trivial ones like preference flips or location updates — went through the LLM.
2. **Latent UUID bug.** `WriteMemoryRow` passed `link.entity_id` (a `UUID` object) into `create_entity_link`, which 500'd on `httpx.post(json=...)`. Any write with `entity_links` has failed since v1.0.0; we just had no integration coverage exercising that exact shape.
3. **Fleet scoping dropped.** The storage-api `/rdf-conflicts` router didn't forward `fleet_id`, so the SQL fell into `Memory.fleet_id IS NULL`. Even when triples did exist, RDF detection only worked for null-fleet writes — i.e., never in production.

This PR closes all three.

### New: `EmitMemoryTriple` pipeline step
- Default ON, per-tenant override `write.triple_emission_enabled`.
- Populates the three triple columns from `entity_links` (subject) + a deterministic phrase→predicate regex table (predicate, object_value).
- All emitted predicates are a subset of `SINGLE_VALUE_PREDICATES` (parity test enforces this).
- Wired into `fast` / `strong` / `persist` pipelines. **STM and extract-only deliberately excluded** (no contradiction detection runs there).
- Skip-on-error contract: any failure → `SKIPPED`, never raises. The existing LLM fallback remains the safety net for anything the heuristic misses.

### Drive-by: `WriteMemoryRow` UUID stringification
One line: `str(link.entity_id)`. Mirrors the same file's handling of `subject_entity_id` and the bulk write path. Risk analyzed as low — SQLAlchemy auto-coerces; storage tests already pass strings.

### Drive-by: forward `fleet_id` to `/rdf-conflicts`
- `core-storage-api` router accepts and forwards `fleet_id`.
- Python `storage_client.find_rdf_conflicts` accepts `fleet_id`.
- `contradiction_detector` passes `new_memory.get("fleet_id")`.

The service-side SQL was already correct; only the wire was dropped. RDF detection now respects the same isolation boundary as semantic detection.

## Test plan

- [x] 21 new unit tests covering: flag default, already-set bypass, subject resolution (none / one / ambiguous), predicate matching (none / one / ambiguous), normalization, error degradation, pipeline composition (must be in fast/strong/persist, must NOT be in STM/enrichment-only), tenant config resolution, allowlist parity.
- [x] 170 contradiction-related unit tests pass.
- [x] 914 unit tests pass repo-wide (no regressions).
- [x] Ruff clean on every touched file.
- [x] **Wet test** (`scripts/wet_test_triple_emission.py`) passes 6/6 against a live `core-api` with a real fleet — proves the deterministic RDF path fires end-to-end with zero LLM contradiction calls.

Wet-test reproduction:
```bash
docker-compose up -d
python scripts/wet_test_triple_emission.py
```

## Behavioral impact

- **Performance:** every previously-LLM-detected single-value-predicate contradiction (location updates, role assignments, status changes, preference flips, ship dates, …) now resolves in tens of milliseconds via SQL instead of a second of LLM round-trip.
- **Schema status `outdated`** (set only by the RDF path) becomes reachable for the first time. The LLM path continues to use `conflicted`.
- **Insights service** (`insights_service.py:272-450`) reads the same triple columns for multi-value mining — it will now have more signal. Not a regression; flagged for visibility.

## Rollback

- Per-tenant: set `write.triple_emission_enabled=false` via org settings. Instant.
- Global: revert this PR. The three drive-bys are independent and individually revertable.

### DIFF
--- a/core-api/src/core_api/clients/storage_client.py
+++ b/core-api/src/core_api/clients/storage_client.py
@@ -397,6 +397,8 @@ class CoreStorageClient:
         subject_entity_id: str,
         predicate: str,
         exclude_id: str | None = None,
+        fleet_id: str | None = None,
+        object_value: str | None = None,
     ) -> list[dict]:
         params: dict[str, Any] = {
             "tenant_id": tenant_id,
@@ -405,6 +407,10 @@ class CoreStorageClient:
         }
         if exclude_id is not None:
             params["exclude_id"] = exclude_id
+        if fleet_id is not None:
+            params["fleet_id"] = fleet_id
+        if object_value is not None:
+            params["object_value"] = object_value
         return await self._get_list("/memories/rdf-conflicts", **params)
 
     async def scored_search(self, data: dict) -> list[dict]:
--- a/core-api/src/core_api/pipeline/compositions/write.py
+++ b/core-api/src/core_api/pipeline/compositions/write.py
@@ -6,6 +6,7 @@ from core_api.pipeline.steps.write import (
     CheckExactDuplicate,
     CheckSemanticDuplicate,
     ComputeContentHash,
+    EmitMemoryTriple,
     LoadTenantConfig,
     MergeEnrichmentFields,
     ParallelEmbedEnrich,
@@ -35,6 +36,9 @@ def build_persist_pipeline() -> Pipeline:
     return Pipeline(
         "write_persist",
         [
+            # entity_links and content are expected to be fully enriched by the
+            # upstream enrichment pipeline before this path runs.
+            EmitMemoryTriple(),
             CheckExactDuplicate(),
             CheckSemanticDuplicate(),
             WriteMemoryRow(),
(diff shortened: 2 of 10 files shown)


ran-taig:
@claude

ran-taig:
@claude

ran-taig:
@claude

ran-taig:
@claude

Eldad-Caura (approved):

erni-a (approved):

PR #176, opened 2026-05-19
CAURA-679 fix(storage): NULL-embedding rows rank on fts_score alone

eyal-bl:
The scored-search CTE blended `(1 - w) * vec_sim + w * fts_score` unconditionally. For NULL-embedding rows (CAURA-594 deferred-embed window, or any case where the embed worker fails), the `vec_sim = 0.0` sentinel left `similarity = w * fts_score` — a hidden `w` haircut that let noise-floor-cosine embedded rows beat unembedded FTS matches and slip past `LIMIT top_k * overfetch_factor`.

Wrap the similarity expression in a CASE so NULL rows score on `fts_score` alone. PR #150's post-filter is what kept these rows discoverable via `has_embedding`; this protects them at the rank-by- score stage as well, end-to-end. Defensive against permanent embedder failure: rows stay searchable instead of silently undiscoverable.

Tradeoff: a row's similarity can step down (not up) when the embed worker patches it if the resulting vec_sim turns out to be low. Brief NULL-window rank instability vs. the current silent-drop pathology — the former is observable and self-correcting; the latter is invisible.

Adds regression test asserting `similarity` is invariant under `fts_weight` for NULL rows (under the old blend it would scale linearly with `w`).

## Type of Change

- [ ] Documentation update
- [ ] Other:

## Checklist

- [ ] `mypy` passes
- [ ] `pytest` passes locally
- [ ] I have updated relevant documentation (README, ARCHITECTURE_REVIEW, etc.)

### DIFF
--- a/core-storage-api/src/core_storage_api/services/postgres_service.py
+++ b/core-storage-api/src/core_storage_api/services/postgres_service.py
@@ -677,8 +677,7 @@ class PostgresService:
         # through the similarity blend into `score`, and PostgreSQL's
         # default `ORDER BY score DESC` sorts NULLS FIRST — putting
         # every unembedded row at the TOP of results. The CASE forces
-        # a numeric value (0.0 — the blend's identity element so NULL
-        # rows rank on fts_score alone) and also short-circuits so
+        # a numeric value (0.0) and also short-circuits so
         # cosine_distance isn't even evaluated for NULL rows.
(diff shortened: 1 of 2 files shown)


eyal-bl:
@claude

arkash20 (dismissed):

arkash20 (approved):

PR #179, opened 2026-05-20
CAURA-125 fix(contradiction): direction-invariant detection (split detection from attribution)

ran-taig:
## Summary

Audit gap **A6** — contradiction detection was measured at `dir_inv = 0.795` (41 of 200 probe pairs flipped verdict under A↔B swap). Root cause: `_candidate_is_older` was used as a hard *detection* filter at three call sites in `_detect()`, gating whether the LLM/RDF check ran at all on candidate-vs-new temporal order. Memories that were newer than the just-written row were silently dropped.

This PR splits **detection** from **attribution**:
- Detection runs symmetrically on every candidate.
- `_pick_older(a, b)` decides which row carries `outdated` / `conflicted` status *after* the conflict is confirmed.
- Supersession edge always points newer → older, regardless of which row was "new memory" at detection time.

## What changes

| File | Change |
|---|---|
| `core-api/src/core_api/services/contradiction_detector.py` | New helper `_pick_older(a, b)` with UUID-string tiebreaker; replaces `_candidate_is_older` (no external callers). Three call sites (RDF path, semantic path, entity-overlap path) rewritten to detect first, then attribute. |
| `tests/test_contradiction_direction_invariance.py` (new) | 14 unit tests against mocked storage: helper semantics + RDF/semantic paths + canonical/flipped/tied cases + symmetry property + no-conflict no-op. |
| `scripts/wet_test_direction_invariance.py` (new) | 15-fixture pair harness that calls `_llm_contradiction_check(A, B)` and `_llm_contradiction_check(B, A)` and asserts agreement. JSONL output for before/after diff. |
| `scripts/wet_test_baselines/caura-125-baseline.jsonl` | Wet-test against unchanged main. |
| `scripts/wet_test_baselines/caura-125-postfix.jsonl` | Wet-test after the fix lands. |

## Commit ordering (fixtures-first)

1. **fixtures + harness** — no production-code change; runnable against any branch.
2. **baseline JSONL** — pins current asymmetry rate per shape (87% combined).
3. **the fix** — `_pick_older` + three call-site rewrites.
4. **unit tests** — 14 cases against mocked storage, exercising the helper + end-to-end through `_detect`.
5. **post-fix JSONL** — direct diff against (2).

### Unit tests (the in-scope coverage)

| Suite | Tests | Result |
|---|---|---|
| `tests/test_contradiction_direction_invariance.py` (new) | 14 | **14/14 pass** ✓ |
| All contradiction-related unit tests (`contradiction`, `supersede`, `_detect`, `pick_older`, `direction`) | 111 | **111/111 pass** ✓ |

The unit tests prove the candidate-filter is now fully symmetric — the audit's primary 41/200 source.

### Wet-test (the LLM-side residual)

| Shape | Baseline | Post-fix | Δ |
|---|---|---|---|
| `genuine_conflict` | 14/16 (88%) | 14/16 (88%) | 0 |
| `conditional_unrealized` | 2/2 (100%) | 2/2 (100%) | 0 |
| `event_restatement` | 2/2 (100%) | 2/2 (100%) | 0 |
| `list_valued_predicate` | 2/2 (100%) | 2/2 (100%) | 0 |
| `refinement` | 2/2 (100%) | 2/2 (100%) | 0 |
| `scope_mismatch` | 1/2 (50%) | 1/2 (50%) | 0 |
| `same_name_distinct_subject` | 1/2 (50%) | **2/2 (100%)** | +1 |
| `temporal_supersession` | 2/2 (100%) | 2/2 (100%) | 0 |
| **OVERALL** | **26/30 (87%)** | **27/30 (90%)** | **+3 pts** |

Modest +3 pt wet-test delta is the *expected shape* — the wet test only exercises the LLM-side component (calls `_llm_contradiction_check` twice and checks the parser verdict agrees). The remaining 3/30 failures are LLM **prompt-framing** asymmetries (the prompt labels inputs "Statement A (NEW)" / "Statement B (EXISTING)" and the model uses that framing to decide temporal direction). Per the PR plan, prompt-framing is **deferred** — it requires either dropping the labels (loses model context) or double-calling (doubles LLM spend).

## What the audit will measure post-fix

The audit's `dir_inv = 0.795` is a *production-traffic* number — most asymmetric pairs were deferred-embedding race conditions and `created_at` ties at the candidate-filter layer, *not* LLM-side. With the candidate filter symmetrized:

- Pairs whose asymmetry was *purely* candidate-filter-driven → now invariant. ✓
- Pairs whose asymmetry was LLM-prompt-driven → still ~10% residual; covered by deferred follow-up.

Expect production `dir_inv` to move from ~0.80 to ~0.95+ on the next probe; the wet test underestimates the production gain because its fixtures are LLM-side stress tests, not deferred-embedding races.

## Out of scope (explicitly deferred per plan)

- **LLM prompt A/B framing.** Same trade-off as CAURA-124 — fixable via prompt rework or double-call, deferred for a separate ticket.
- **Multi-candidate fan-out reconciliation.** When `_detect` finds several contradicting candidates with mixed ages, per-pair direction is well-defined; chain coherence is unchanged from today.

## Validation gates

- [x] 14 new unit tests pass.
- [x] 111 contradiction-related unit tests pass repo-wide.
- [x] Ruff format + check clean on every touched file.
- [x] DCO `Signed-off-by` on every commit.
- [x] Wet test against both OpenAI gpt-4o-mini and Gemini gemini-2.5-flash with `--json-out` artifacts pinned in the tree for direct diff.

## Behavioral changes worth flagging in review

- **Previously-unreachable branch now active:** when a candidate is newer than `new_memory` (deferred-embedding scenarios — a memory written without an embedding, queued, finally embedded after newer rows landed), the candidate now carries `supersedes_id` pointing back at `new_memory`, and `new_memory` is marked outdated. Previously: silent drop.
- **`created_at` ties no longer skip detection.** Strict `<` in the old helper meant tied timestamps caused detection to bail entirely. UUID tiebreaker now produces a deterministic direction.
- **CAS race semantics preserved.** When the flipped branch writes `candidate.supersedes_id`, storage's existing `WHERE supersedes_id IS NULL` CAS handles the race where another writer already set the candidate's edge between detection runs.


ran-taig:
@claude

ran-taig:
@claude

ran-taig:
@claude

erni-a (approved):

PR #191, opened 2026-05-24
fix(plugin): conform to OpenClaw MemoryFlushPlan contract (CAURA-000)

ran-taig:
## Summary

Customer report 2026-05-21: a node running plugin **v2.6.0** threw `Error: Invalid memory flush target path` during a long-running session.

**Root cause** — our `registerMemoryFlushPlan(...)` callback returns the wrong shape. OpenClaw's `MemoryFlushPlan` type (declared in `memory-state.d.ts:182`) requires six fields:

```ts
type MemoryFlushPlan = {
  softThresholdTokens: number;
  forceFlushTranscriptBytes: number;
  reserveTokensFloor: number;
  model?: string;
  prompt: string;
  systemPrompt: string;
  relativePath: string;   // ← load-bearing
};
```

Pre-fix we returned `{instructions, softThresholdTokens}` — wrong field name (`instructions` instead of `prompt`) **and missing four other fields including `relativePath`**.

When OpenClaw triggers a memory flush (long conversation crossing `softThresholdTokens` or `forceFlushTranscriptBytes`), `agent-runner.runtime` does:

```js
const memoryFlushWritePath = activeMemoryFlushPlan.relativePath;  // undefined
await ensureMemoryFlushTargetFile({
  workspaceDir: …,
  relativePath: memoryFlushWritePath,
});
// inside ensureMemoryFlushTargetFile:
if (!workspaceDir || !relativePath || path.isAbsolute(relativePath))
  throw new Error("Invalid memory flush target path");
```

So the bug is **unconditional** but only fires once a session is long enough to require compaction — which is why it looks intermittent ("one node experienced this") across a fleet.

## Why we got it wrong

Our 2026-04 commit added the registration with the comment *"MemClaw persists server-side, no local file"* on a sibling field (`memoryFlushWritePath` on `resolveMemoryBackendConfig`, which doesn't exist on that contract either). We conflated two unrelated fields and skipped the required ones. The TypeScript types weren't enforced because `api.registerMemoryFlushPlan` is `unknown`-typed in our `api` parameter.

## Changes

| File | Change |
|---|---|
| `plugin/src/index.ts` | Rebuild the flush-plan resolver to return all six required fields. `relativePath` is `memclaw/flush-YYYY-MM-DD.md` (UTC date stamp), namespaced under `memclaw/` so we don't collide with memory-core's `memory/` layout when both plugins coexist on a host. Defaults match memory-core's reference impl: `softThresholdTokens=4000`, `forceFlushTranscriptBytes=2 MiB`, `reserveTokensFloor=20000`. `prompt`/`systemPrompt` tell the compaction sub-agent to persist via `memclaw_write` and not produce a user reply. Also drop the misleading `memoryFlushWritePath: null` from `resolveMemoryBackendConfig` (field doesn't exist on that contract). |
| `plugin/src/runtime-contract.test.ts` | + 4 tests that lock the flush-plan shape: every required field has the right primitive type; `relativePath` is non-absolute, no parent-escape, namespaced under `memclaw/`; date-stamp is deterministic on identical `nowMs`; resolver tolerates being called with no args (legacy callers). |

## Affected versions

Anything 2.4.x onwards that ships our `registerMemoryFlushPlan` callback. v2.6.0 specifically per the customer report. Symptom is **only visible on long sessions** crossing the soft / hard compaction thresholds.

## Customer mitigation while this ships

Disable memory flush on the affected node by adding to `~/.openclaw/openclaw.json`:

```jsonc
{
  "agents": {
    "defaults": {
      "compaction": { "memoryFlush": { "enabled": false } }
    }
  }
}
```

That makes OpenClaw's `buildMemoryFlushPlan` return `null` → `resolveMemoryFlushPlan` returns `null` → agent-runner skips the entire flush sequence and uses normal compaction. Plain compaction still works; just no MemClaw-persistence side effect on the way through.

## Test plan

- [x] `npx tsc --noEmit` clean
- [x] `npm test` — 230/230 (226 prior + 4 new)
- [ ] After merge, re-run wet test on `openclaw-test-ran`: trigger compaction by force-feeding a >2 MiB transcript, verify `memclaw/flush-YYYY-MM-DD.md` gets created and `memclaw_write` calls land

### DIFF
--- a/plugin/src/index.ts
+++ b/plugin/src/index.ts
@@ -445,17 +445,82 @@ const memclawPlugin = {
     }
 
     // --- Memory flush plan ---
+    //
+    // OpenClaw's MemoryFlushPlan contract (see memory-state.d.ts)
+    // requires SIX fields:
+    //
+    //   softThresholdTokens, forceFlushTranscriptBytes, reserveTokensFloor,
+    //   prompt, systemPrompt, relativePath
+    //
+    // Pre-fix we returned {instructions, softThresholdTokens} — wrong
+    // field name (instructions vs prompt) AND missing the four
+    // other fields. relativePath is the load-bearing omission: when
+    // compaction crosses softThresholdTokens, OpenClaw's
+    // agent-runner.runtime reads activeMemoryFlushPlan.relativePath
+    // and passes it to ensureMemoryFlushTargetFile, which throws
+    // "Invalid memory flush target path" on any falsy / absolute
+    // value. The error only surfaces on long sessions that actually hit
+    // the compaction threshold, so it looked intermittent — but the bug
+    // is unconditional. Customer report 2026-05-21.
+    //
+    // relativePath is the workspace-relative scratch file the
+    // compaction sub-agent has append-only write access to during the
+    // flush turn. MemClaw's server-side persistence is orthogonal — the
+    // sub-agent still calls memclaw_write to capture salient
+    // context, but it ALSO needs the file to exist because that's the
+    // only filesystem write surface OpenClaw exposes to it. Mirror
+    // memory-core's memory/YYYY-MM-DD.md layout but namespace under
+    // memclaw/ so we don't collide on hosts running both plugins.
+    // The outer try here only catches REGISTRATION-time failure. The
+    // resolver itself runs later in OpenClaw's agent-runner stack — any
+    // throw there propagates up and crashes the flush turn. Pre-fix two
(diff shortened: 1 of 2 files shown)


ran-taig:
@claude

ran-taig:
@claude

ran-taig:
@claude

erni-a (dismissed):

erni-a (approved):

PR #209, opened 2026-05-25
fix(logging): propagate stdlib logger extras into structlog JSON

Eldad-Caura:
## Summary

``logger.info(msg, extra={...})`` from stdlib was silently dropping every ``extra`` key at the GCP boundary. The JSON payload in Cloud Logging contained only ``message``, ``timestamp``, ``logger``, ``severity`` — never the structured fields the call site attached.

## Discovery context

Bringing up CAURA-682 Phase 1 write-latency instrumentation. The ``memory_write_latency`` log emits fine — verified ~1000+ entries in GCP during the staging loadtest — but none of ``tenant_id`` / ``embedding_ms`` / ``enrichment_ms`` / ``storage_ms`` / ``entity_links_ms`` are visible in ``jsonPayload``. Per-tenant phase slicing (the whole point of Phase 1) is impossible.

## Blast radius

This isn't just my new log — it's a project-wide bug. Every existing extras-using call site has been silently emitting only ``message`` + ``timestamp`` to GCP since the structlog cutover:

| Site | Lost fields |
|---|---|
| ``core-api/routes/memories.py:1322`` "search request completed" | `path`, `tenant_id`, `top_k`, `row_count`, `total_ms`, `error` |
| ``core-api/routes/memories.py:596`` "memory get" | `path`, `tenant_id`, `memory_id`, `total_ms`, `error` |
| ``core-api/middleware/per_tenant_concurrency.py:121`` "per-tenant concurrency cap reached" | `scope`, `tenant_id`, `cap` |
| ``core-api/services/memory_service.py:352`` "memory_write_latency" (CAURA-682) | every phase timing |

After this PR all four start emitting their structured fields to GCP. No call sites change.

## Root cause + fix

``_base_processors()`` in ``common/structlog_config.py`` lacked ``structlog.stdlib.ExtraAdder()`` — the processor whose specific job is reading non-standard LogRecord attributes (the ``extra`` dict) and merging them into structlog's event dict. Without it, the stdlib bridge's ``ProcessorFormatter`` drops them on the floor.

One-line config addition. Comment explains what it does and which sites it unblocks.

## Test

``tests/test_logging.py`` gains two cases:
- ``test_stdlib_logger_extras_reach_json_payload`` — exercises the configured pipeline end-to-end (stdlib ``logger.info(msg, extra={...})`` → JSON output → parse → assert each key is a top-level field). Pins the contract.
- ``test_stdlib_logger_without_extras_still_emits`` — regression guard so a future processor edit can't accidentally add empty/null extra fields when the call site didn't pass any.

JSON capture uses the handler's stream object directly (in-memory ``StringIO`` swap) rather than ``capfd`` / ``capsys``, because pytest's stdout swap happens before the conftest's ``configure_logging``, leaving the StreamHandler bound to the original FD.

## Test plan

- [x] ``uv run pytest tests/test_logging.py`` — 15/15 pass (13 pre-existing + 2 new)
- [x] ``uv run pytest tests/pipeline/`` — 79/79 pass (no regression from the new processor)
- [x] ``uv run ruff check / format --check`` clean on touched files
- [x] ``cd core-api && uv run mypy src/`` — ``Success: no issues found in 161 source files``
- [ ] After merge → auto-deploy to staging via CAURA-683 webhook → re-run noisy-neighbor loadtest → confirm ``jsonPayload.path="memory-write"`` query returns per-tenant phase timings (the goal of CAURA-682 Phase 1)

### DIFF
--- a/common/structlog_config.py
+++ b/common/structlog_config.py
@@ -70,6 +70,99 @@ class LoggingConfigurationWarning(UserWarning):
     """
 
 
+# Canonical attrs on a vanilla `logging.LogRecord`. Mirrors
+# `structlog.stdlib._LOG_RECORD_KEYS` (the set that ExtraAdder excludes)
+# — by constructing a dummy record we stay aligned with whatever the
+# stdlib reports for the current Python version, rather than hard-coding
+# the names (which change between releases — e.g. ``taskName`` was added
+# in 3.12). Frozen at import time so a downstream library's runtime
+# monkey-patching of the LogRecord class can't widen the deny set under
+# our feet.
+_LOG_RECORD_STANDARD_ATTRS: frozenset[str] = frozenset(
+    logging.LogRecord("name", 0, "pathname", 0, "msg", (), None).__dict__.keys()
+)
+
+# Keys the structlog/GCP pipeline reserves for itself. A user-supplied
+# ``extra={...}`` must not be allowed to populate any of these — each
+# would open a silent-corruption channel into a different GCP field:
+#
+#   * ``event`` — would overwrite ProcessorFormatter's
+#     ``event_dict["event"]`` (the real LogRecord message); then
+#     ``_rename_event_to_message`` would rename the extra's value as
+#     the GCP ``message`` field, silently losing the original text.
+#     (Stdlib ``logger.*()`` accepts ``extra={"event": ...}`` because
+#     ``event`` isn't a standard LogRecord attribute — structlog's
+#     convention only.)
+#
+#   * ``message`` — stdlib's own LogRecord-attr-collision guard
+#     ``raise KeyError("Attempt to overwrite 'message'")`` already
+#     blocks the ``extra={"message": ...}`` route. The deny entry
+#     here is defense-in-depth for the ``logging.Filter`` path: a
+#     filter that stamps ``record.message = "x"`` directly bypasses
+#     stdlib's ``extra=`` validation.
+#
+#   * ``severity`` — ``_add_logrecord_extras`` runs BEFORE
+#     ``_map_to_gcp_severity``, whose guard only fills the field when
+#     it is ``None`` or ``""``. A stdlib ``extra={"severity": "P1"}``
+#     (an application-domain label) would therefore propagate
+#     untouched into the GCP severity field, overriding the correct
+#     log-level-derived value, producing an invalid GCP severity, and
+#     potentially misrouting log-based alerts.
+#
+#   * ``stack`` — set by ``StackInfoRenderer`` ONLY when the caller
+#     passes ``stack_info=True``. Without that flag, the renderer is
+#     a no-op and an ``extra={"stack": "noise"}`` would propagate
+#     untouched into the JSON payload alongside any future
+#     stack-trace-based tooling.
+#
+#   * ``exception`` — set by ``format_exc_info`` ONLY when ``exc_info``
+#     is present on the call. Without exception context an
+#     ``extra={"exception": "fake"}`` persists in the JSON payload,
+#     fabricating exception data on a non-exception log line.
+_RESERVED_OUTPUT_KEYS: frozenset[str] = frozenset(
+    {"event", "message", "severity", "stack", "exception"}
+)
+
+
+def _add_logrecord_extras(
+    _logger: Any,
+    _method_name: str,
+    event_dict: dict[str, Any],
+) -> dict[str, Any]:
+    """structlog processor: copy a ``LogRecord``'s non-standard attrs
+    into the event dict, but block reserved-output-key collisions.
+
+    Equivalent to ``structlog.stdlib.ExtraAdder()`` for typical extras
+    (``path``, ``tenant_id``, ``total_ms``, etc.) but additionally
+    refuses to propagate ``event`` / ``message`` — the two keys that
+    silently corrupt the GCP ``message`` field if a caller passes them
+    via ``extra={...}``. ``ExtraAdder`` exposes only ``allow`` (since
+    structlog 21.5), not ``deny``, so we can't use it directly without
+    resorting to a brittle static allowlist of every key the codebase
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
Applied. Pushed as ``e41185e``.

Fair catch — ``ExtraAdder()`` without an ``allow`` list propagates **all** non-standard ``LogRecord`` attributes, not just ``extra={...}`` keys. My original comment claimed the narrower behavior, which would have been a footgun for the next maintainer who added a ``logging.Filter`` and was surprised by accidental JSON-field leakage.

Picked **option 2** (accurate comment, not narrowed allowlist) — grep'd ``common/`` and ``core-api/src/`` first to confirm no ``logging.Filter`` in-tree stamps custom attributes today, so the broader extraction is functionally equivalent. Comment now explicitly notes ``logging.Filter`` is a known second channel:

> ExtraAdder propagates all non-standard ``LogRecord`` attributes into the event dict — this covers ``logger.info(msg, extra={...})`` call sites as well as any ``logging.Filter`` that stamps attributes directly onto a record. (No such filter exists in-tree today; flagged so a future addition is a known channel, not a surprise.)

No code change, comment-only. Tests still 15/15; ruff/format/mypy all green.

Eldad-Caura:
@claude

Eldad-Caura:
Applied — with one substantive correction and one new finding. Pushed as ``81e917d``.

## Correction: ``ExtraAdder`` has no ``deny`` parameter

The suggested ``ExtraAdder(deny=["event", "message"])`` doesn't exist — ``ExtraAdder`` exposes only ``allow`` (verified against structlog 25.x source: ``def __init__(self, allow: Collection[str] | None = None)``). Confirmed by the test run: ``TypeError: ExtraAdder.__init__() got an unexpected keyword argument 'deny'``.

A static ``allow`` list would have to enumerate every key the codebase might log (~25 today across 4 call sites and growing), which is exactly the brittle path the reviewer flagged as undesirable.

Instead, replaced ``structlog.stdlib.ExtraAdder()`` with a custom ``_add_logrecord_extras`` processor — a ~15-line shim that mirrors ``ExtraAdder``'s implementation but adds an explicit deny set (``event``, ``message``). ``_LOG_RECORD_STANDARD_ATTRS`` is derived at import time from a dummy ``logging.LogRecord(...)`` ``.__dict__.keys()`` (same trick structlog uses) so it stays aligned with the running Python version's LogRecord shape rather than hard-coding a list that drifts (e.g., ``taskName`` was added in 3.12).

## New finding while writing the test: Path 2 isn't actually silent

While writing ``test_extra_message_key...``, Python's stdlib ``logging`` module itself rejected ``logger.info(msg, extra={"message": "x"})`` with:

```
KeyError: "Attempt to overwrite 'message' in LogRecord"
```

This is stdlib's own LogRecord-attr-collision guard, in [Python's logging/__init__.py:1636](https://github.com/python/cpython/blob/main/Lib/logging/__init__.py). So the reviewer's "Path 2 — silent corruption" is actually **impossible through the ``extra={...}`` API** — stdlib raises before the call even reaches our processor chain.

That means: the deny entry on ``message`` is still load-bearing, but as **defense-in-depth** for a different attacker shape — a ``logging.Filter`` that stamps ``record.message = "x"`` *directly* on the record (bypassing stdlib's ``extra=`` validation). No such filter exists in-tree today, but if one is added, ``_add_logrecord_extras`` blocks the collision.

Test updated to cover both paths:
- ``test_extra_message_key_is_rejected_by_stdlib_then_blocked_in_depth`` — first asserts stdlib's ``KeyError`` on the ``extra={"message": ...}`` route, then exercises a real ``logging.Filter`` that stamps ``record.message`` directly and asserts the deny entry still prevents corruption.

## Validation

- ``uv run pytest tests/test_logging.py`` → **17/17 pass** (13 pre-existing + 4 new: extras propagation × 2 + event-collision + message-collision-in-depth)
- ``uv run pytest tests/pipeline/`` → 79/79 pass (no regression)
- ``ruff check / format --check`` clean
- ``cd core-api && uv run mypy src/`` → 161 source files, clean

Ready for re-review.

Eldad-Caura:
@claude

Eldad-Caura:
Applied. Pushed as ``bc6ba21``.

Both findings valid — same corruption shape as ``event`` / ``message``, just targeting different GCP fields. Added all three (``severity`` Medium, plus ``stack`` and ``exception`` Low) to ``_RESERVED_OUTPUT_KEYS``.

## Changes

**``_RESERVED_OUTPUT_KEYS`` expanded** from ``{"event", "message"}`` to ``{"event", "message", "severity", "stack", "exception"}``. The comment block now documents each key's specific corruption vector — they're all variations on "pipeline processor only fills on a guard, so a non-empty user extra slips past":

- ``severity`` — ``_map_to_gcp_severity`` only fills None/"" → ``extra={"severity": "P1"}`` survives → invalid GCP severity / misrouted alerts.
- ``stack`` — ``StackInfoRenderer`` is a no-op without ``stack_info=True`` → ``extra={"stack": "noise"}`` propagates.
- ``exception`` — ``format_exc_info`` is a no-op without ``exc_info`` → ``extra={"exception": "fake"}`` fabricates exception data on a non-exception log line.

**Tests**: new parametrized ``test_extra_pipeline_reserved_keys_are_dropped`` covering all three. The ``severity`` case asserts the pipeline-derived value (``INFO``) survives; ``stack`` and ``exception`` cases assert the keys are absent from the JSON.

## Validation

- ``pytest tests/test_logging.py`` → **20/20 pass** (17 from previous cycle + 3 new parametrized cases)
- ``ruff check / format --check`` clean
- ``mypy src/`` clean (161 source files)

Ready for re-review.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #212, opened 2026-05-25
fix(plugin): wire keystones into WhatsApp system prompts end-to-end (CAURA-000)

ran-taig:
## Summary

Two commits, one customer-reported bug: agents on WhatsApp could fetch keystone rules via the `memclaw_keystones` tool but the `<keystone_rules>` block never appeared in the LLM's system prompt. Validated against OpenClaw 2026.5.4 source pulled via `npm pack`.

- **`95460acd` — auto-claim `slots.contextEngine`.** Install script (`core-api/.../plugin.py`), `autoFixAllowlist`, and `isMemclawFullyConfigured` now own both `slots.memory` AND `slots.contextEngine`. Without the latter, OpenClaw's `resolveContextEngine` (registry-DFFgCbcm.js:241) falls to the "legacy" engine and our `MemClawContextEngine.assemble()` is never invoked. New boot-time WARNING surfaces the gap on already-installed nodes so operators don't have to dig through gateway.log.
- **`60df6477` — fix `assemble()` to match the 2026.5.4 `AssembleResult` contract.** After the slot was set, OpenClaw started calling our assemble and the log filled with `TypeError: Cannot read properties of undefined (reading 'slice')` once per turn. Root cause: contract requires `{messages: AgentMessage[], estimatedTokens: number, systemPromptAddition?: string}`; we were returning `{system, systemPromptAddition, ...}` with `messages` undefined. Downstream consumer at selection-BfCSa_QL.js:7677 reads `assembled.messages.slice(...)` to detect mutation — `undefined.slice()` throws, the runtime's catch logs via `String(err)` (strips the stack), and `systemPromptAddition` is silently dropped. Fix: outer try/catch that logs the full stack via `logErrorCritical` + verbatim `console.warn`, defensive `params.messages` coercion, both return sites now include `messages: incomingMessages` (reference-equal echo) and `estimatedTokens`. Legacy `system` + `tokenEstimate` aliases retained for older OpenClaw.
- **Forensic anchor.** New one-shot boot line `[memclaw] BOOT: plugin vX.Y.Z, N tools registered, node A.B.C` so the next customer report pins deployed version + tool surface + node runtime via a single `grep BOOT`.

## Files changed (combined across both commits)

- `core-api/src/core_api/routes/plugin.py` — install script writes both slots.
- `plugin/src/config.ts` + `config.test.ts` — `isContextEngineSlotClaimed`, expanded `isMemclawFullyConfigured`, `autoFixAllowlist` claims both slots.
- `plugin/src/index.ts` — boot WARNING when `slots.contextEngine !== "memclaw"`; one-shot `BOOT:` diagnostic line at register time.
- `plugin/src/context-engine.ts` — `assemble` wrapper with outer try/catch + stack logging + safe fallback; `_assembleInner` is the relocated body; both return sites match `AssembleResult` v2026.5.4.
- `plugin/src/runtime-contract.test.ts` — 5 new tests in a `ContextEngine.assemble contract` block, locking the shape + the throw-safety guarantee.
- `tests/test_plugin_source_manifest.py` — `test_install_script_claims_both_slots` drift test.
- `plugin/package.json` + `openclaw.plugin.json` + `version.ts` — 2.6.1 → 2.6.3.

## Test plan

- [x] `npm test` in `plugin/` — **250/250 pass** (245 prior + 5 new assemble contract tests + the 6 new slot tests in the prior commit).
- [x] `pytest tests/test_plugin_source_manifest.py` — **11/11 pass** (10 prior + 1 new slot drift test).
- [ ] Customer reinstall + WhatsApp turn — confirm `[memclaw] BOOT:` appears once at startup, `[memclaw] ContextEngine 'memclaw' registered` appears once, and the next `memclaw_keystones`-relevant turn shows the `<keystone_rules>` block in the model's view of the system prompt (no more `assemble failed` warnings; if any inner throw occurs, the full stack is now logged via our own logger).

### DIFF
--- a/core-api/src/core_api/routes/plugin.py
+++ b/core-api/src/core_api/routes/plugin.py
@@ -493,9 +493,17 @@ if (config.plugins.entries['memory-core']) {{
   config.plugins.entries['memory-core'].enabled = false;
 }}
 
-// Claim the exclusive memory slot
+// Claim the exclusive memory slot (controls registerMemoryPromptSection +
+// registerMemoryRuntime delivery) AND the contextEngine slot (controls
+// ContextEngine.assemble() — the path that injects the <keystone_rules>
+// block into the system prompt on every turn). Without contextEngine set
+// to 'memclaw', OpenClaw falls back to the default "legacy" engine and
+// our assemble() never runs, so keystones never appear in the prompt
+// even though the tool surface is registered. Confirmed against
+// OpenClaw 2026.5.4 dist/registry-DFFgCbcm.js:241 resolveContextEngine.
 if (!config.plugins.slots) config.plugins.slots = {{}};
 config.plugins.slots.memory = 'memclaw';
+config.plugins.slots.contextEngine = 'memclaw';
 
 if (!config.plugins.load) config.plugins.load = {{}};
 if (!Array.isArray(config.plugins.load.paths)) config.plugins.load.paths = [];
--- a/plugin/openclaw.plugin.json
+++ b/plugin/openclaw.plugin.json
@@ -2,7 +2,7 @@
   "id": "memclaw",
   "name": "MemClaw",
   "kind": "memory",
-  "version": "2.6.2",
+  "version": "2.6.3",
   "description": "Central persistent memory for OpenClaw agents — write, search, and entity lookup tools.",
   "entry": "dist/index.js",
   "skills": [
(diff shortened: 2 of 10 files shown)


ran-taig:
@claude

ran-taig:
@claude

ran-taig:
@claude

Eldad-Caura (approved):

PR #219, opened 2026-05-25
feat(storage-api): add /_debug/pg_locks endpoint for live lock investigation (CAURA-686)

Eldad-Caura:
## Summary

[CAURA-686](https://caura-ai.atlassian.net/browse/CAURA-686). Adds ``GET /api/v1/storage/_debug/pg_locks`` on core-storage-api plus a small polling script (``scripts/capture_pg_locks_during_storm.py``) to capture per-query lock-holder evidence during a contention storm.

## Why

The CAURA-682 noisy-neighbor work narrowed the remaining bottleneck to PG row-level lock contention (instance-level ``alloydb.googleapis.com/instance/postgresql/wait_time`` showed **7.4s of `Lock/transactionid` + 3.0s of `Lock/tuple` cumulative** under a 30-second storm). AlloyDB Insights' per-query aggregation only captures LWLocks; ``Lock/transactionid`` / ``Lock/tuple`` waits don't surface there. So we don't yet know *which* query is the lock holder.

The canonical Postgres surface for query-level lock attribution is ``pg_locks ∩ pg_stat_activity`` + ``pg_blocking_pids()`` — but AlloyDB's private-IP-only posture rules out direct ``psql`` from a laptop. This PR adds that surface as a thin read-only endpoint on the storage-api itself.

## What's in the PR

| File | Purpose |
|---|---|
| ``core-storage-api/.../routers/debug.py`` | New ``GET /_debug/pg_locks`` route. ~80 LOC including the SQL + docstring. Filters to backends that are either waiting OR being waited-on, returns ``wait_event_type / wait_event / xact_age_sec / query / blocked_by_pids`` per row. |
| ``core-storage-api/.../routers/__init__.py`` | Export ``debug_router`` |
| ``core-storage-api/.../app.py`` | Register the new router under ``/api/v1/storage`` |
| ``core-storage-api/tests/test_integration.py`` | One shape test in ``TestHealth`` — asserts the endpoint returns the documented payload structure and runs without raising against the test DB. |
| ``scripts/capture_pg_locks_during_storm.py`` | Thin polling loop that hits the endpoint at 1Hz, writes JSONL for later ``jq`` slicing. Documents the typical ``gcloud run services proxy`` + storm workflow. |

## Posture / security

- The endpoint is on the storage-api's private VPC IP, same as every other route there. It is NOT routed through the gateway (``memclaw.dev`` / ``memclaw.net``).
- No authentication is added — core-storage-api's existing trust boundary is the private network. To reach this endpoint externally you need ``gcloud run services proxy`` with Cloud Run IAM, which my account has on staging only.
- The response carries the first 240 chars of each waiting/holding query (``LEFT(a.query, 240)``) and Postgres ``pid``s; no user-data PII surfaces.
- Read-only — runs one ``SELECT`` per request.

## What I'll do after merge

After CAURA-683 auto-deploys this to staging:

```bash
gcloud run services proxy staging-memclaw-core-storage-writer --port 8080 --region us-central1 &
python3 scripts/capture_pg_locks_during_storm.py --base http://localhost:8080 --duration 90 --out /tmp/pg-locks-storm.jsonl
# concurrently:
python3 loadtest.py dry dev
```

Then slice the JSONL by the noisy-neighbor storm window (Phase 4b — ~test-elapsed 11:00 → 11:30 from the loadtest's stdout), identify the lock holder + blocked-by chain, and propose a targeted fix on CAURA-686. Strong candidate based on the schema + storm pattern: speculative-insertion contention on ``ix_memories_attempt_unique`` partial unique index, but I want the actual evidence before guessing.

## Test plan

- [x] ``pytest core-storage-api/tests/test_integration.py::TestHealth`` — 4/4 pass (3 existing + 1 new shape test)
- [x] ``ruff check / format --check`` clean on touched files
- [x] ``cd core-storage-api && uv run mypy src/`` — Success: no issues in 48 source files
- [ ] After merge + auto-deploy: ``curl`` the endpoint via ``gcloud run services proxy``, then run the loadtest, capture the snapshots, attribute the locks

[CAURA-686]: https://caura-ai.atlassian.net/browse/CAURA-686?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core-storage-api/src/core_storage_api/app.py
+++ b/core-storage-api/src/core_storage_api/app.py
@@ -28,6 +28,7 @@ from core_storage_api.middleware import RejectWritesOnReaderMiddleware
 from core_storage_api.routers import (
     agents_router,
     audit_router,
+    debug_router,
     documents_router,
     entities_router,
     fleet_router,
@@ -119,6 +120,11 @@ def create_app() -> FastAPI:
     app.include_router(tasks_router, prefix=prefix)
     app.include_router(idempotency_router, prefix=prefix)
     app.include_router(lifecycle_audit_router, prefix=prefix)
+    # CAURA-686: ``GET /api/v1/storage/_debug/pg_locks`` for live
+    # pg_locks / pg_stat_activity snapshots during contention triage.
+    # Behind the same private-VPC posture as everything else here —
+    # not exposed via the gateway.
+    app.include_router(debug_router, prefix=prefix)
 
     return app
 
--- a/core-storage-api/src/core_storage_api/routers/__init__.py
+++ b/core-storage-api/src/core_storage_api/routers/__init__.py
@@ -1,5 +1,6 @@
 from core_storage_api.routers.agents import router as agents_router
 from core_storage_api.routers.audit import router as audit_router
+from core_storage_api.routers.debug import router as debug_router
 from core_storage_api.routers.documents import router as documents_router
 from core_storage_api.routers.entities import router as entities_router
 from core_storage_api.routers.fleet import router as fleet_router
@@ -14,6 +15,7 @@ from core_storage_api.routers.tasks import router as tasks_router
 __all__ = [
     "agents_router",
     "audit_router",
+    "debug_router",
     "documents_router",
     "entities_router",
     "fleet_router",
--- a/core-storage-api/src/core_storage_api/routers/debug.py
+++ b/core-storage-api/src/core_storage_api/routers/debug.py
@@ -1,0 +1,151 @@
+"""Diagnostic endpoints for live DB inspection.
+
+Internal-only surface — exposed on the storage-api private VPC IP and
+gated by Cloud Run IAM. Not routed through the gateway. The endpoints
+here snapshot DB-side state (``pg_locks``, ``pg_stat_activity``,
+``pg_blocking_pids``) for triaging contention storms; they're cheap
+enough to poll at 1Hz during a controlled loadtest, but expensive
+enough to NOT wire into a routine dashboard.
+
+CAURA-686 motivation: instance-level Cloud Monitoring wait-event
+metrics (``alloydb.googleapis.com/instance/postgresql/wait_time``) tell
+you THAT row-level locks are accumulating but not WHICH query holds
+them. ``pg_stat_activity`` joined with ``pg_locks`` and
+``pg_blocking_pids`` is the canonical Postgres surface for that.
+AlloyDB Insights' per-query lock_time aggregation does NOT capture
+``Lock/transactionid`` / ``Lock/tuple`` — only LWLocks — so this
+endpoint is the only way to attribute row-lock waits to specific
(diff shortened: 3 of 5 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
Applied all four. Pushed as ``5ed0deb``.

| # | Fix |
|---|---|
| 1 | ``from datetime import UTC, datetime`` moved to module level with the other imports. |
| 2 | ``_PG_LOCKS_SNAPSHOT_SQL`` rewritten as a CTE so ``pg_blocking_pids()`` is called exactly once per backend. Prior version called it ~5× per backend (once in WHERE, once in SELECT projection, twice via ``array_length`` on the same call, once in the IN-list subquery). Inline comment documents the change so a future reader doesn't undo it. |
| 3 | New ``pg_read_all_stats: true / false / null`` field as a fast visibility self-check. Three-valued so the caller can distinguish "no contention so we don't know" from "missing privilege so we're blind". Docstring spells out each value's meaning. Existing shape test gains an assertion that the field is in ``{True, False, None}``. |
| 4 | Polling script switched to ``contextlib.nullcontext(sys.stdout)`` plus ``with ctx as sink:`` — closes the file-handle-leak window where the previous shape opened the file BEFORE entering ``try``, so any exception in the loop setup would leak the handle. |

Tests still pass (4/4 in ``TestHealth``); ruff / format / mypy clean.

Eldad-Caura:
@claude

Eldad-Caura:
Applied all three. Pushed as ``bffb3a2``.

| # | Fix |
|---|---|
| 1 | Direct privilege check via ``_PG_CHECK_STATS_SQL`` — ``pg_has_role(current_user, 'pg_read_all_stats', 'USAGE') OR pg_has_role(current_user, 'pg_monitor', 'USAGE')``. Asks the catalog instead of inferring from row contents, so ``pg_read_all_stats`` in the response is now always ``True`` / ``False`` — never ``null``. The earlier heuristic couldn't tell "no contention" from "no privilege"; this can. Docstring rewritten; shape test now asserts ``isinstance(..., bool)``. |
| 2 | ``ValueError`` added to the polling script's except tuple. ``json.JSONDecodeError`` inherits from ``ValueError``, so this catches partial / malformed responses during a cold-start or proxy hiccup. Inline comment notes why. |
| 3 | ``open(...)`` moved inside the ``with`` expression so the file is acquired as part of the same context — no half-open-handle window from a ``ctx = open(...)`` assignment outside the ``with``. Comment updated. |

Tests still 4/4 in ``TestHealth``; script ``--help`` parses; ruff / format / mypy clean.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #220, opened 2026-05-25
fix(storage-api): bulk-insert memory_entity_links with ON CONFLICT DO NOTHING (CAURA-686)

Eldad-Caura:
## Summary

Phase 2 fix for [CAURA-686](https://caura-ai.atlassian.net/browse/CAURA-686) (row-lock contention during noisy-neighbor write storms).

`PATCH /memories/{id}/entities` iterated `entity_links` in a Python `for` loop, calling the singular storage method N times — N separate transactions, each doing a bare `INSERT INTO memory_entity_links` with no `ON CONFLICT` handling. Two concurrent storage-api requests targeting the same `(memory_id, entity_id)` PK serialised on `Lock/transactionid` for the full duration of the holder's transaction.

Replace the loop with a single bulk `pg_insert(...).on_conflict_do_nothing(index_elements=["memory_id", "entity_id"])` call. Identical-row conflicts now silently skip instead of blocking, idempotent re-PATCH is a no-op, and the route makes 1 DB round trip instead of N.

## Evidence

Fresh staging `/_debug/pg_locks` capture during today's loadtest storm caught the contention chain directly: pid 646055 blocked by pid 645132 on `Lock/transactionid` for ≥3s, blocked statement was `INSERT INTO memory_entity_links (memory_id, entity_id, role) VALUES ($1::UUID, $2::UUID, $3::VARCHAR)`. Tenant B write p95 during storm: 573.1ms → 3820.7ms (6.67×), well above the ≤2× [CAURA-682](https://caura-ai.atlassian.net/browse/CAURA-682) closure target. Storm-window data attached on the Jira ticket.

The prior hypothesis on memclaw memory `45414b2b…` (`ix_memories_attempt_unique` speculative-insertion contention on the `memories` table) is **refuted** by the JSONL — different table, different lock type.

## Test plan

- [ ] CI: new `test_patch_entities_is_bulk_and_idempotent` exercises bulk-insert + conflict-skip role-preservation
- [ ] After auto-deploy to staging via [CAURA-683](https://caura-ai.atlassian.net/browse/CAURA-683): re-run `loadtest.py dry dev`
- [ ] Tenant B write storm/baseline ratio drops to ≤ 2.0× (currently 6.67×)
- [ ] Prod promotion stays gated per `loadtest-autonomy` keystone + [CAURA-676](https://caura-ai.atlassian.net/browse/CAURA-676) manual gate

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[...]

PR #223, opened 2026-05-25
test(storage-api): default fake_embedding to VECTOR_DIM (1024)

Eldad-Caura:
## Summary

The integration-test fixture `fake_embedding(seed, dim=768)` in `core-storage-api/tests/test_integration.py` has been silently stale since migration `012_vector_dim_1024.py` widened the `memories.embedding` pgvector column from 768 to 1024.

Every test in `TestMemories` was failing locally with:
```
sqlalchemy.exc.StatementError: (builtins.ValueError) expected 1024 dimensions, not 768
```

CI never caught this because it provisions a fresh DB on each run and the mismatch path was masked by other factors there. But anyone running `pytest` locally on `core-storage-api` was hitting it on the very first test that exercised the write path.

Fix: switch the default to `common.constants.VECTOR_DIM`, which is the canonical constant the rest of the embedding stack already uses (`common/embedding/providers/fake.py`, etc.). One import + one default change.

## Test plan

- [x] Re-ran `tests/test_integration.py::TestMemories` locally — the three tests that previously failed on Memory creation (`test_create_and_get_memory`, `test_update_status_via_patch`, `test_patch_entities_is_bulk_and_idempotent`) now all pass.
- [x] Confirmed `tests/test_qemb_cache_dim_versioning.py` is the only other 768 hardcode in tests — it's deliberate (tests the 768→1024 cache-key versioning), out of scope.
- [ ] CI

## Background

Discovered during CAURA-686 Phase 2 work (PR #220). My new test there couldn't be exercised locally because of this drift; CI was the only path. This unblocks local TDD on `core-storage-api`.

### DIFF
--- a/core-storage-api/tests/test_integration.py
+++ b/core-storage-api/tests/test_integration.py
@@ -15,6 +15,8 @@ import uuid
 import pytest
 from httpx import AsyncClient
 
+from common.constants import VECTOR_DIM
+
 pytestmark = [pytest.mark.integration, pytest.mark.asyncio]
 
 PREFIX = "/api/v1/storage"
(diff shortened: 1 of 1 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #235, opened 2026-05-27
chore(perf+sec+tests): ran follow-ups — T5 + #22/#25/#29/#31

ran-taig:
## Summary

Five small, independent fixes from the audit follow-up queue. Each is contained to one file and none depends on the others — could be split into 5 PRs, bundled here for review efficiency since the per-change LOC is small.

| Audit ID | Surface | Type | LOC |
|---|---|---|---|
| **T5** | tests/conftest.py — `_patch_storage_client` fixture | Test hygiene (cross-test leak) | +5 |
| **#22** | routes/crystallizer.py — `get_report` | Security (cross-tenant existence leak) | +6 |
| **#25** | services/evolve_service.py — `_apply_outcome_to_db` | Perf (N+1 UPDATE → single statement) | ~80 |
| **#29** | services/insights_service.py — `_persist_findings` | Perf (serial create_memory → create_memories_bulk) | ~75 |
| **#31** | services/entity_extraction_worker.py | Cleanup (dead try/except) | de-indent only |

### T5 — try/finally on `_patch_storage_client`
The autouse fixture mutated `sc_mod._client` then `yield`ed. A setup-time exception between the mutation and yield (or any teardown error) would leak the ASGI-bridged client into subsequent tests via the module-level singleton — caused intermittent cascade failures previously. Wrap in try/finally.

### #22 — 404 (not 403) for foreign-tenant report read
`GET /crystallize/reports/{id}` used to 404 on missing rows but **403** on foreign-tenant rows (via `enforce_tenant`). Two distinguishable states let attackers enumerate cross-tenant report_ids by random-probing UUIDs (403 = exists in another tenant; 404 = doesn't exist anywhere). Collapse both to 404 with the same detail. Admin keys keep cross-tenant read via the `is_admin` bypass.

### #25 — Collapse evolve_service N+1 weight UPDATEs
One UPDATE per `related_id` inside its own savepoint → single `UPDATE FROM (VALUES …)` via `IN :mids` expanding bindparam. One outer savepoint preserves the "DB failure aborts the bulk but not the outer transaction" contract. Per-row response ordering and the "row-not-found → skip with warning" semantics preserved.

### #29 — Use `create_memories_bulk` in insights findings persist
Serial `create_memory` per finding inside per-finding savepoint → single `create_memories_bulk` call. Per-item error isolation comes from the bulk contract (`BulkItemResult(status="error")` → `None` in `insight_ids`). Empty-findings short-circuit added (`BulkMemoryCreate.items` enforces `min_length=1`).

### #31 — Delete dead try/except
Inner `try: ... except Exception: raise` re-raised into the outer except one frame above — pure no-op. Removed the wrapper and de-indented the body. Smaller stack on logged exceptions; cleaner blame.

## Test plan

- [x] **2 new regression tests** in `tests/test_api_crystallizer.py` for #22:
  - `test_get_report_foreign_tenant_returns_404_not_403`: non-admin caller, foreign report → 404 (asserts via direct route call with mocked AuthContext + storage client; no HTTP fixture required).
  - `test_get_report_foreign_tenant_admin_bypass`: admin still reads cross-tenant.
- [x] **Updated** `tests/test_insights_service.py::test_priors_restored_when_all_findings_fail` to patch `create_memories_bulk` (the new persist target) instead of `create_memory`. Still exercises the "all findings failed → restore priors" branch.
- [x] **218 tests pass** across touched surfaces (storage bulk endpoints, entity linking, insights, crystallizer, evolve, A4 status, A10 evolve rule, MCP insights, entity blocklist, A5c tenant config).
- [x] Ruff lint + format clean.
- [x] mypy on core-api clean (only pre-existing `config.py` croniter stub from main).

### DIFF
--- a/core-api/src/core_api/routes/crystallizer.py
+++ b/core-api/src/core_api/routes/crystallizer.py
@@ -131,7 +131,16 @@ async def get_report(
     report = await sc.get_report(str(report_id))
     if not report:
         raise HTTPException(status_code=404, detail="Report not found")
-    auth.enforce_tenant(report.get("tenant_id", ""))
+    # Collapse foreign-tenant (403) into not-found (404) so callers
+    # can't probe for the existence of reports in other tenants by
+    # distinguishing 403 from 404 on random UUIDs (audit finding #22).
+    # Cross-tenant read keys (``readable_tenant_ids`` widened past the
+    # home tenant) are honoured here — the set always contains
+    # ``auth.tenant_id`` by construction (``AuthContext`` line 85-90),
+    # so the 404 mask still hides reports the caller has no read
+    # access to.
+    if not auth.is_admin and report.get("tenant_id") not in auth.readable_tenant_ids:
+        raise HTTPException(status_code=404, detail="Report not found")
     return {
         "id": str(report.get("id", "")),
         "tenant_id": report.get("tenant_id"),
--- a/core-api/src/core_api/services/entity_extraction_worker.py
+++ b/core-api/src/core_api/services/entity_extraction_worker.py
@@ -83,121 +83,118 @@ async def process_entity_extraction(
 
         sc = get_storage_client()
 
-        try:
-            blocklist = tenant_cfg.entity_blocklist
+        blocklist = tenant_cfg.entity_blocklist
 
-            # Embed entity names for fuzzy resolution
-            name_embeddings: dict[str, list[float]] = {}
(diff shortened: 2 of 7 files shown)


ran-taig:
@claude

ran-taig:
@claude

ran-taig:
@claude

Eldad-Caura (approved):

PR #237, opened 2026-05-27
CAURA-687 fix(search): wire ENTITY_LOOKUP short-circuit to dedicated …

eyal-bl:
…storage endpoint

The ENTITY_LOOKUP short-circuit in ClassifyQuery._collect_memories posted a payload with ``memory_ids`` + ``entity_lookup: True`` to ``/memories/ scored-search``, but that route never read those keys — it hard-indexed ``body["embedding"]`` and 500'd. The broad ``except Exception:`` at classify_query.py:123 swallowed the failure and the pipeline silently fell through to keyword/semantic search on every entity-token query.

This was the second bug in the same code path. CAURA-684 fixed the dict- vs-tuple shape mismatch at _expand_per_fleet line 214; once that landed the next gate (_collect_memories → scored_search) became reachable and its contract bug surfaced as a storage 500 (confirmed live in staging 2026-05-27: same query, same fallthrough warning, but trace now ends at KeyError 'embedding' on the storage side instead of KeyError 0 on the client side).

Adopts Approach A from the design discussion:

- New storage method ``PostgresService.memory_load_by_ids`` that selects memories by ID with visibility/fleet/agent/memory_type/status/valid_at filters lifted verbatim from ``memory_scored_search``. No vector cosine, no FTS, no freshness scoring — sub-millisecond ``SELECT WHERE id IN (...)``. Filter clauses MUST stay in sync with ``memory_scored_search`` (any drift is a cross-tenant leak risk — comment notes this for future maintainers).

- New route ``POST /memories/load-by-ids`` (dedicated endpoint rather than a conditional branch inside ``/memories/scored-search`` — clearer contract, smaller blast radius, easier to test).

- New storage client method ``load_memories_by_ids``.

- ``_collect_memories`` switches the call from ``scored_search`` to ``load_memories_by_ids``. Also now forwards ``valid_at`` and ``readable_tenant_ids`` which were silently dropped before — the short-circuit's visibility behaviour now matches the scored-search fallthrough exactly. Drops the unused ``entity_lookup: True`` flag.

Tests:

- tests/test_entity_lookup_short_circuit.py: switched the storage mock to ``load_memories_by_ids``. Added two assertions: (1) ``valid_at`` + ``readable_tenant_ids`` are forwarded in the cross-tenant case. (2) ``readable_tenant_ids`` is OMITTED in the single-tenant case so storage uses the cheap ``tenant_id == X`` predicate. Both also assert the old stale keys (``embedding``/``query``/ ``search_params``/``entity_lookup``) are absent from the payload.

- tests/pipeline/test_classify_query.py: switched the four ENTITY_LOOKUP test paths' mocks from ``scored_search`` to ``load_memories_by_ids``, plus the ``_mock_sc`` default. The remaining scored_search default stays for non-ENTITY_LOOKUP test paths.

Follow-up worth a separate ticket: tighten the broad ``except Exception:`` at classify_query.py:123 to catch only expected exception types so the next regression doesn't take months to surface.

## Type of Change

- [ ] Documentation update
- [ ] Other:

## Checklist

- [ ] `mypy` passes
- [ ] `pytest` passes locally
- [ ] I have updated relevant documentation (README, ARCHITECTURE_REVIEW, etc.)

### DIFF
--- a/core-api/src/core_api/clients/storage_client.py
+++ b/core-api/src/core_api/clients/storage_client.py
@@ -606,6 +606,15 @@ class CoreStorageClient:
     async def scored_search(self, data: dict) -> list[dict]:
         return await self._post("/memories/scored-search", data, read=True)  # type: ignore[return-value]
 
+    async def load_memories_by_ids(self, data: dict) -> list[dict]:
+        """ENTITY_LOOKUP short-circuit endpoint (CAURA-687).
+
+        Skips vector/FTS/freshness scoring — caller supplies memory IDs,
+        server applies visibility/fleet/agent filters and returns raw
+        memory rows. Pairs with PostgresService.memory_load_by_ids.
+        """
+        return await self._post("/memories/load-by-ids", data, read=True)  # type: ignore[return-value]
+
     async def archive_expired(self, tenant_id: str, fleet_id: str | None = None) -> int:
         data: dict[str, Any] = {"tenant_id": tenant_id}
         if fleet_id is not None:
--- a/core-api/src/core_api/pipeline/steps/search/classify_query.py
+++ b/core-api/src/core_api/pipeline/steps/search/classify_query.py
@@ -13,6 +13,7 @@ import asyncio
 import logging
 import re
 import types
+from datetime import datetime
 from uuid import UUID
 
 from core_api.clients.storage_client import get_storage_client
@@ -65,6 +66,8 @@ class ClassifyQuery:
         filter_agent_id: str | None = ctx.data.get("filter_agent_id")
         memory_type_filter: str | None = ctx.data.get("memory_type_filter")
         status_filter: str | None = ctx.data.get("status_filter")
+        valid_at = ctx.data.get("valid_at")
+        readable_tenant_ids: list[str] | None = ctx.data.get("readable_tenant_ids")
         graph_max_hops: int = sp["graph_max_hops"]
         top_k: int = sp["top_k"]
 
@@ -95,6 +98,8 @@ class ClassifyQuery:
                         filter_agent_id=filter_agent_id,
                         memory_type_filter=memory_type_filter,
                         status_filter=status_filter,
+                        valid_at=valid_at,
+                        readable_tenant_ids=readable_tenant_ids,
                     )
 
                     if filtered_rows:
@@ -252,6 +257,8 @@ class ClassifyQuery:
         filter_agent_id: str | None = None,
         memory_type_filter: str | None = None,
         status_filter: str | None = None,
+        valid_at: datetime | None = None,
+        readable_tenant_ids: list[str] | None = None,
     ) -> list[types.SimpleNamespace]:
         """Load memories linked to graph-expanded entities, scored by hop distance."""
         all_entity_ids = list(entity_hops.keys())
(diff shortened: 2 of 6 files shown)


eyal-bl:
@claude

eyal-bl:
Thanks for the review — pushed fixes for all three findings as an amended single commit (`0106816`):

## Medium — `top_k` LIMIT without ORDER BY ✅ fixed
- Dropped `top_k` from `memory_load_by_ids` (removed both the `top_k` parameter and `stmt.limit(top_k)`).
- Dropped `"top_k"` from the `search_data` payload in `_collect_memories`.
- Caller still caps the result via `rows[:top_k]` after sorting by hop boost; `GRAPH_MAX_BOOSTED_MEMORIES = 50` is the natural safety cap on the input side.
- Added a comment in both places explaining why a SQL LIMIT here would silently discard high-boost rows.

## Low — parse errors before try block ✅ fixed
- Moved the `UUID(mid)` list-comp and `datetime.fromisoformat(valid_at)` parse calls inside the `try` block.
- Pre-declared `memory_ids: list[UUID] = []` outside the try so the `finally` can safely read `len(memory_ids)` for `log_request` even if the parse raised.

Note: the existing `/scored-search` route has the same parse-outside-try pattern — leaving it untouched here to keep this PR's scope tight; worth a separate follow-up.

## Low — `readable_tenant_ids` invariant ✅ fixed
- Replaced `len(readable_tenant_ids) > 1` with the explicit `readable_tenant_ids != [tenant_id]` check.
- Added `test_collect_memories_forwards_single_element_when_different_from_home` to lock in the edge case (single element naming a non-home tenant must still be forwarded).

Note: the same `len > 1` pattern lives in `core-api/src/core_api/pipeline/steps/search/execute_scored_search.py:79-81` (which is where I copied it from). Same follow-up call — leaving untouched here for scope discipline.

## Tests
30 passed locally (`tests/pipeline/test_classify_query.py` + `tests/test_entity_lookup_short_circuit.py`) — one new positive-control test for the readable_tenant_ids edge case.

eyal-bl:
@claude

eyal-bl:
Pushed both review fixes as another amended single commit (`9486077`):

## Low — `valid_at` type annotation ✅ fixed
- Added `from datetime import datetime` to the imports block.
- Annotated `valid_at: datetime | None = None` in `_collect_memories` to match every other kwarg in the signature.

## Low — `body["tenant_id"]` → 422 instead of 500 ✅ fixed
- Added an explicit guard inside the try block, just after the `memory_ids` check, that raises `HTTPException(422, "tenant_id is required")` on missing field.
- Replaced `body["tenant_id"]` with the local `tenant_id` in the `_svc.memory_load_by_ids` call.
- Verified `HTTPException` was already imported at the top of the router.
- Note: the existing `/scored-search` route has the same indexed-access pattern. Leaving untouched here for scope; same follow-up as the parse-outside-try point from the previous CR round.

## Suggestion — conditional payload fields (omit on `None`)
Declining politely. The router uses `.get()` so explicit `None` is harmless on the wire, the schema reads more obviously in one glance vs. a dozen `if x is not None: ...` lines, and the wire savings (`null` = 4 bytes) are negligible. Happy to revisit if it becomes a pattern enforced repo-wide.

## Tests
30 passed locally — same set, no new tests needed for these two changes (annotation is pure hygiene; the 422 guard is on a brand-new endpoint with no callers depending on prior 500 behaviour).

ran-taig (approved):

PR #240, opened 2026-05-27
feat(events): subscription manifest to guard infra provisioning

Eldad-Caura:
## Summary

Adds a machine-readable contract of **which Pub/Sub topics each deployed service subscribes to**, so the infra repo (caura-memclaw-enterprise) can fail CI when a consumed topic lacks a provisioned subscription.

This is the OSS half of a guard against the failure that took staging down: OSS #238 added `memclaw.lifecycle.insights-requested` with a core-api subscriber, but no Terraform subscription existed — core-api's pull loop 404'd, `/api/v1/health` 503'd, and `deploy-to-staging` failed on every run.

- `common/events/events_manifest.json` — `{service: [topics]}` for PubSub/SaaS deployment mode.
- `scripts/gen_events_manifest.py` — generator (`--check` mode for CI). Lifecycle topics are captured **dynamically** by invoking `register_pipeline_consumers` / `register_archive_consumers` against a recording bus, so a new `bus.subscribe` in a helper is picked up automatically. Memory-pipeline direct subscribes are listed explicitly (OSS CI doesn't install the worker package, so it can't be imported here).
- `tests/test_events/test_events_manifest.py` — fails if the committed manifest drifts from the generator (runs in the existing `pytest tests/` CI step; no workflow change).

## How the guard works end-to-end
1. Dev adds a consumed topic (e.g. to a lifecycle helper) → this test fails until they run `python scripts/gen_events_manifest.py` → manifest updated.
2. The enterprise repo's `check_pubsub_provisioning.py` reads this manifest and fails its CI until a matching `${var.env}-<service>--<topic>` Terraform subscription exists.

## Test plan
- [x] `python scripts/gen_events_manifest.py --check` passes
- [x] `pytest tests/test_events/test_events_manifest.py` passes (2 tests)
- [ ] Manifest reviewed: core-api = {embedded, enriched, crystallize, entity-link, insights}; core-worker = {embed-requested, enrich-requested, archive-expired, archive-stale, purge-soft-deleted}


Eldad-Caura:
@claude

Eldad-Caura:
Thanks — all three addressed (amended, force-pushed):

**🟠 High — `_MEMORY_DIRECT` blind spot.** Real gap, good catch. Added `test_memory_direct_matches_consumer_files`, which greps `core-api/src/core_api/consumer.py` and `core-worker/src/core_worker/consumer.py` for `bus.subscribe(Topics.Memory.*)` calls and asserts the resolved topic set exactly matches `_MEMORY_DIRECT` for each service. Pure file-read (no import), so it works in OSS CI even though the worker package isn't installed — a new direct subscribe added without updating `_MEMORY_DIRECT` now fails this test instead of silently shipping an incomplete manifest.

**🟡 Low — encoding.** Pinned `encoding="utf-8"` on both `read_text`/`write_text` (and the new test's reads). Note the manifest file itself is already pure-ASCII because `json.dumps` defaults to `ensure_ascii=True` (the em dash is written as `—`), but explicit is better.

**🟢 Suggestion — diff on `--check`.** `--check` now prints a `difflib.unified_diff` of committed vs regenerated on staleness, so the failure is self-diagnosing.

Tests: 3 pass (`is_in_sync`, `is_well_formed`, `memory_direct_matches_consumer_files`); ruff check + format clean.

Eldad-Caura:
@claude

Eldad-Caura:
**🟡 Medium — `build_manifest` drops `_MEMORY_DIRECT`-only services. Applied.** Good catch — a silent manifest gap is exactly what this system exists to prevent. `build_manifest` now drives the loop off `set(_LIFECYCLE_HELPERS) | set(_MEMORY_DIRECT)`, so each dict is independently authoritative and a service in one but not the other still appears. (No behavior change today — both dicts hold the same two services — but it removes the latent footgun.) Manifest output is byte-identical; all 3 tests still pass.

**🟢 Low — atomic manifest write. Declining, with reasoning.** This is a developer-run codegen script, not a production write path. A mid-write interruption produces a corrupt file that the very next `--check` / drift test flags immediately, and the fix is just re-running the generator — it's self-correcting and the file is version-controlled. The temp-file+rename adds a stray `.tmp` artifact to reason about (and a `.gitignore` entry) for a failure mode that can't reach `main` (CI regenerates and compares). I'd rather keep the one-liner. Happy to add it if you feel strongly.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #241, opened 2026-05-27
feat(storage): purge_org_data primitive for org hard-delete (CAURA-689)

Eldad-Caura:
## Summary

OSS side of the organization hard-delete (epic CAURA-513). Permanently removes all `public.*` data for a set of `tenant_ids` (every tenant of an enterprise org being deleted). Enterprise `platform-admin-api` calls this during the hard-delete sweep / force-delete; the `enterprise.*` rows are torn down separately by `platform-storage-api`.

- **core-storage-api** — `PostgresService.purge_tenant_data(tenant_id)` hard-deletes every row scoped to a tenant across the OSS schema in **one transaction, children-before-parents** (so a FK never blocks a delete), returning per-table counts. `memory_entity_links` cascades from `memories`/`entities`; `organization_settings*` are keyed by `org_id` (== tenant id in OSS). **`lifecycle_audit` is intentionally retained** — it's the audit trail. Exposed via `POST /purge/tenant-data`.
- **core-api** — admin route `POST /admin/org/purge-data` (admin-key) loops the `tenant_ids`, wrapping each storage purge in a `lifecycle_audit` row (`hard-delete-org`, `pending` → `success`/`failure`) and **isolating per-tenant failures**. The route opts out of the request-timeout middleware (terminal batch op, bounded by the storage client's own timeout).

Idempotent: re-purging an already-purged tenant deletes nothing and returns zeros, so a retried org hard-delete is safe.

## Test plan
- [x] `core-storage-api/tests/test_purge.py` — 4 integration tests (real PG): deletion + per-table counts, idempotency, tenant-isolation, validation. All pass.
- [x] `ruff check src/` + `ruff format --check src/` + `mypy src/` clean for both core-storage-api and core-api.
- [ ] Reviewer: confirm the purge table set stays in sync with the OSS schema (new tenant-scoped tables must be added to `_PURGE_TENANT_TABLES`).

### DIFF
--- a/core-api/src/core_api/app.py
+++ b/core-api/src/core_api/app.py
@@ -55,6 +55,7 @@ from core_api.routes.keystones import router as keystones_router
 from core_api.routes.lifecycle import router as lifecycle_router
 from core_api.routes.memories import admin_memories_router
 from core_api.routes.memories import router as memories_router
+from core_api.routes.org_deletion import router as org_deletion_router
 from core_api.routes.plugin import plugin_bootstrap_router
 from core_api.routes.plugin import router as plugin_router
 from core_api.routes.settings import router as settings_router
@@ -595,6 +596,7 @@ app.include_router(stm_router, prefix="/api/v1")
 app.include_router(insights_router, prefix="/api/v1")
 app.include_router(evolve_router, prefix="/api/v1")
 app.include_router(lifecycle_router, prefix="/api/v1")
+app.include_router(org_deletion_router, prefix="/api/v1")
 
 # Test-only endpoints (time-warp, etc.) — only registered when TESTING=1
 if _os.getenv("TESTING") == "1":
--- a/core-api/src/core_api/clients/storage_client.py
+++ b/core-api/src/core_api/clients/storage_client.py
@@ -646,6 +646,15 @@ class CoreStorageClient:
         result = await self._post("/memories/purge-soft-deleted", data)
         return result.get("deleted", 0)  # type: ignore[union-attr]
 
+    async def purge_tenant_data(self, tenant_id: str) -> dict[str, int]:
+        """Hard-delete ALL OSS data for ``tenant_id`` (CAURA-689 org delete).
+
+        Returns the per-table deleted counts. Idempotent at the storage
+        layer — a repeat call for an already-purged tenant returns zeros.
+        """
+        result = await self._post("/purge/tenant-data", {"tenant_id": tenant_id})
+        return result.get("deleted", {})  # type: ignore[union-attr,return-value]
+
     async def count_active(self, tenant_id: str, fleet_id: str | None = None) -> int:
         params: dict[str, Any] = {"tenant_id": tenant_id}
         if fleet_id is not None:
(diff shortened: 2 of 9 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
Thanks — all three addressed in the amended commit:

- **🟠 High — audit update not exception-safe:** the per-tenant `lifecycle_audit` writes are now isolated. `create_lifecycle_audit_row` is wrapped (falls back to `audit_id=None`), and both the success and failure updates go through a `_finalize(audit_id, **kwargs)` helper that swallows-and-logs. A transient audit-write failure can no longer abort the batch or mask a completed purge — the purge stays the source of truth and the loop continues.
- **🟡 Med — unbounded `tenant_ids`:** added `_MAX_TENANT_IDS = 100`; larger batches get a 422 (the orchestrator splits across calls).
- **🔵 Low — duplicate `tenant_ids`:** rejected with 422 (`len != len(set)`), so we can't emit two audit rows for the same tenant.

`ruff check` + `ruff format --check` + `mypy` clean on `src/` for both services; the `core-storage-api/tests/test_purge.py` integration tests still pass. (core-api has no unit-test harness here — the route is covered by e2e; the storage primitive is the part with table-level tests.)

Eldad-Caura:
@claude

Eldad-Caura:
Thanks for the re-review — all three addressed in the amended commit:

- **🟡 Med — empty `tenant_ids` silently accepted (`all([])` is True):** validation now also requires `not tenant_ids` → empty list returns 422 (matches the existing error message).
- **🔵 Low — unqualified raw DELETE:** the purge SQL is now `DELETE FROM public.{table_name} WHERE …`, so an irreversible hard-delete can't be redirected by a non-default `search_path`.
- **🔵 Low — malformed JSON body → 500:** wrapped `await request.json()` in `try/except json.JSONDecodeError` → 422 in both `core-api/.../org_deletion.py` and `core-storage-api/.../purge.py` (matches the existing storage convention, e.g. `lifecycle_audit` router). Added `test_purge_rejects_malformed_body` (5 storage tests passing).

`ruff check`, `ruff format --check`, and `mypy` all clean on `src/` for both services. (core-api has no unit-test harness for routes; the validation fixes are straightforward and the e2e path is the integration cover.)

Eldad-Caura:
@claude

Eldad-Caura:
Round 3 — both addressed in the amended commit:

- **🟡 Med — destructive storage endpoint has no service auth:** verified the established convention — *no* storage router in this service authenticates (`routers/keystones.py` even documents it: "Trust enforcement lives upstream in core-api; storage assumes its callers are trusted"), and the upstream `POST /admin/org/purge-data` already enforces `auth.enforce_admin()` with `CORE_API_ADMIN_API_KEY`. So I expanded the purge module docstring to make the trust model + VPC-only constraint explicit, matching that convention. No new auth surface that would be inconsistent with the rest of `core-storage-api`.
- **🔵 Low — HTTP 200 on batch failure:** the route now returns **207 Multi-Status** whenever `failed` is non-empty (partial or total), and **200** only when every tenant purged successfully (uniform signalling — the orchestrator can detect failure from the status alone). Updated docstring with the explicit contract, switched the return type to `JSONResponse`, added the import.

Re the **module-level `_svc = PostgresService()` singleton** suggestion: verified that *every* storage router uses it (`agents.py`, `audit.py`, `documents.py`, `entities.py`, `fleet.py`, `idempotency.py`, `keystones.py`, `lifecycle_audit.py`, `memories.py`) — already consistent, `PostgresService.__init__` does no I/O. No change needed.

5 storage purge tests still pass; ruff + format + mypy clean on `src/` for both services.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #244, opened 2026-05-28
feat(core): OSS suppression mirror + boundary guard (CAURA-694)

Eldad-Caura:
## Summary

Adds the OSS-side consumer of the `memclaw.org.suppression-changed` event that enterprise `platform-admin-api` already publishes (CAURA-691 / -695). Synchronously gates **core-api** on the mirrored state so a suppressed org cannot read or write through any auth path — even before the durable hard-delete sweep lands.

## Layers

| Layer | Change |
|---|---|
| `common/events` | `Topics.Org.SUPPRESSION_CHANGED` constant, typed `OrgSuppressionEvent` payload (`{tenant_ids, action: "suppress" \| "restore"}`), shared `register_suppression_consumer` factory. Per-tenant failures re-raise so Pub/Sub redelivers (DLQ on max-delivery, configured in CAURA-695). |
| `core-storage-api` | Migration **019** creates `public.tenant_suppression` (`tenant_id` PK + nullable `suppressed_at`). Service primitives `set_tenant_suppression` / `is_tenant_suppressed`. `ON CONFLICT` preserves the original `suppressed_at` on duplicate `suppress` so the canonical timestamp survives Pub/Sub redelivery. New `POST /tenant-suppression` + `GET /tenant-suppression/{tenant_id}` endpoints under the same VPC-internal trust model as `routers/purge.py`. |
| `core-worker` | `upsert_tenant_suppression` client method + `_SuppressionAdapter` wiring in `register_consumers`. |
| `core-api` | `suppression.is_tenant_suppressed` with a **30-s TTL** cache (mirrors the auth-api `_is_org_deleted` pattern from CAURA-690). **Fail-open** on storage transport errors so a storage flap can't take core-api down. `auth._block_if_suppressed` injected into all four tenant-resolving paths in `get_auth_context` (memclaw-key standalone / header, standalone mode, X-Tenant-ID). Admin paths bypass via `tenant_id is None`. |

## Defence-in-depth

The CAURA-690 auth-api gate is still the first-line synchronous block for JWT / cookie / API-key login. This guard catches the remaining API-key paths and OSS-standalone deployments where there is no auth-api in front.

## Test plan

- [x] `uv run pytest tests/test_suppression_handlers.py tests/test_suppression_boundary.py` → 13 passed
- [x] `uv run mypy src/` on core-api / core-storage-api / core-worker → all clean
- [x] `uv run ruff check` + `ruff format --check` on every touched file → clean
- [ ] `core-storage-api/tests/test_tenant_suppression.py` (integration suite, requires CI DB)
- [ ] End-to-end against staging once CAURA-695's pubsub resources land: publish a `suppress` event, observe row in `tenant_suppression`, observe 403 on a request keyed by the suppressed tenant.

### DIFF
--- a/common/events/__init__.py
+++ b/common/events/__init__.py
@@ -30,7 +30,12 @@ from common.events.lifecycle_publishers import (
 from common.events.memory_embed_publisher import publish_memory_embed_request
 from common.events.memory_enrich_publisher import publish_memory_enrich_request
 from common.events.memory_enriched_publisher import publish_memory_enriched
+from common.events.org_suppression_event import OrgSuppressionEvent
 from common.events.pubsub import PubSubEventBus
+from common.events.suppression_handlers import (
+    SuppressionStorageAdapter,
+    register_suppression_consumer,
+)
 from common.events.topics import Topics
 
 # Intentionally NOT re-exported:
@@ -43,7 +48,9 @@ __all__ = [
     "EventBus",
     "EventHandler",
     "InProcessEventBus",
+    "OrgSuppressionEvent",
     "PubSubEventBus",
+    "SuppressionStorageAdapter",
     "Topics",
     "get_event_bus",
     "publish_archive_expired_request",
@@ -55,4 +62,5 @@ __all__ = [
     "publish_memory_embed_request",
     "publish_memory_enrich_request",
     "publish_memory_enriched",
+    "register_suppression_consumer",
 ]
--- a/common/events/org_suppression_event.py
+++ b/common/events/org_suppression_event.py
@@ -1,0 +1,41 @@
+"""Typed payload for ``memclaw.org.suppression-changed`` (CAURA-694).
+
+Published by enterprise platform-admin-api on every soft-delete /
+restore of an organization. Carries the list of tenant_ids the action
+applies to plus an ``action`` discriminator so a single topic covers
+both directions of the lifecycle. The OSS subscriber upserts a row in
+``public.tenant_suppression`` per tenant_id: ``suppress`` sets
+``suppressed_at = now()``; ``restore`` clears it.
+
+Out of band: the **synchronous** check happens at the auth-api layer
+(CAURA-690) so an unexpired JWT cannot reach a soft-deleted org's data
+even before this event lands. This event is the **durable** mirror —
+core-api uses it to reject **API-key**-credentialed callers (which
+don't traverse auth-api) and to defend OSS standalone deployments
+where there is no auth-api in front.
+"""
+
+from __future__ import annotations
+
+from typing import Literal
+
+from pydantic import BaseModel, Field
+
+
+class OrgSuppressionEvent(BaseModel):
+    """Payload of ``memclaw.org.suppression-changed``.
+
+    ``tenant_ids`` is the explicit list of tenants the action covers —
+    one org has many tenants, and the publisher (platform-admin-api)
+    already knows the list at decision time, so we don't make the
+    consumer re-resolve it. Empty list is valid (a soft-delete of an
+    org with zero tenants is a no-op for the mirror but still
+    well-formed).
+
+    ``action`` is the discriminator. Using ``Literal`` rather than a
+    free string means Pydantic validates the value and consumers can
+    pattern-match safely without a defensive ``else: raise``.
+    """
+
+    tenant_ids: list[str] = Field(default_factory=list)
+    action: Literal["suppress", "restore"]
--- a/common/events/suppression_handlers.py
+++ b/common/events/suppression_handlers.py
@@ -1,0 +1,130 @@
+"""Consumer for ``memclaw.org.suppression-changed`` (CAURA-694).
+
+Lives in ``common/`` so the same code can run in core-worker (the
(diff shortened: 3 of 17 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
All three round-1 findings addressed in `75cb170`:

- **🟠 High — thundering herd during storage outage:**
  `core-api/src/core_api/suppression.py` now writes `(False, now + _ERROR_TTL_SECONDS)` to the cache on the exception path. New `_ERROR_TTL_SECONDS = 5.0` constant — shorter than the 30 s success-TTL so the cache absorbs the storm during the outage and still self-heals quickly once storage recovers. Comments at both the constant and the except block explain the trade-off and point back at this review.

- **🟡 Med — non-object JSON body → 500:**
  `core-storage-api/.../tenant_suppression.py` now checks `isinstance(body, dict)` immediately after `request.json()`. Returns 422 `"request body must be a JSON object"` for `[1,2,3]` / `"string"` / `42` instead of letting the bare `.get()` raise `AttributeError`.

- **🟢 Low — `UnicodeDecodeError` not caught:**
  Widened the except to `(json.JSONDecodeError, UnicodeDecodeError)`. Comment notes that `UnicodeDecodeError` is NOT a subclass of `JSONDecodeError` so a single-class catch missed it.

Tests:
- `test_storage_failure_caches_false_to_rate_limit_retries` — locks in the cache-write-on-failure shape (second call within the error-TTL is served from cache, no second storage hit).
- `test_error_cache_uses_short_ttl` — asserts `_ERROR_TTL_SECONDS < _TTL_SECONDS` AND that the cache expires on the shorter window (fast-forward `time.monotonic` past `_ERROR_TTL_SECONDS` triggers a fresh storage call).
- `test_rejects_non_object_json_body` — parametrizes over `[1,2,3]`, `"string"`, `42` → 422 with `"JSON object"` in detail.
- `test_rejects_invalid_utf8_body` — `b"\xff\xfe …"` body → 422.
- `uv run pytest tests/test_suppression_handlers.py tests/test_suppression_boundary.py` → 15 passed (13 prior + 2 new).
- mypy clean on core-api + core-storage-api; ruff check + format clean on touched files.

On the **partial-failure ordering** suggestion (handler iterates tenants sequentially; a redelivery loop on `t2` keeps `t3` live until DLQ): keeping the sequential approach for now — `asyncio.gather` would let `t3` land on the first delivery but it'd also re-run every successful tenant's idempotent upsert on every redelivery (cheap, but doubles the storage write rate on flaky messages). The current shape is the explicit "correctness > throughput" choice the docstring documents; happy to switch to gather if the throughput tail becomes the bigger concern in practice.

Ready for round 2 whenever the bot wants one.

Eldad-Caura:
@claude

Eldad-Caura:
Round-2 review addressed in `8282d00`. Both Lows + the Pydantic suggestion folded in.

- **🟢 Low — unbounded `_cache` growth:**
  Added `_prune_expired(now)` helper in `core_api/suppression.py` that drops every cache entry whose TTL has elapsed. Called immediately before each `_cache[tenant_id] = ...` write (both the success path and the error-TTL path). Per-call cost is proportional to the currently-cached tenant count, which is itself bounded by distinct tenants seen within the TTL window — so the dict stays bounded in long-running processes with high tenant turnover. Build-list-then-pop pattern dodges the "dictionary changed size during iteration" issue.

- **🟢 Low — `readable_tenant_ids` not checked for suppression:**
  New `_block_if_any_readable_suppressed(tenant_id, readable_tenants)` helper in `auth.py` iterates the readable set (skipping `tenant_id` which the caller already checked) and 403s on the first suppressed match. Wired into **Path 4** (X-Tenant-ID / enterprise ingress) AND **Path 2b** (MEMCLAW_API_KEY + X-Tenant-ID branch) for symmetry, per your suggestion. The cache keeps each extra check at one dict lookup per tenant per 30 s — cheap.

- **🟢 Suggestion — `model_validate` vs `**event.payload`:**
  Switched `common/events/suppression_handlers.py:_handle_suppression_changed` to `OrgSuppressionEvent.model_validate(event.payload)`. Comment notes the small defensiveness gain (catches `TypeError`-shaped non-dict inputs via the same `ValidationError` path that already gets logged + dropped).

Tests:
- `test_cache_pruning_evicts_expired_entries` — primes the cache with a tenant whose TTL is about to elapse, fast-forwards `time.monotonic` past `_TTL_SECONDS`, then triggers a lookup on a DIFFERENT tenant. The old key must be gone from `_cache`.
- `test_block_if_any_readable_suppressed_checks_each` — `[home, co-1, co-2]` with `co-2` suppressed → 403.
- `test_block_if_any_readable_suppressed_skips_home` — home tenant gets no storage call (caller already covered it).
- `test_block_if_any_readable_suppressed_empty_list_noop` — single-tenant credentials short-circuit cleanly.
- `uv run pytest tests/test_suppression_handlers.py tests/test_suppression_boundary.py` → **19 passed** (15 prior + 4 new).
- mypy clean on core-api; ruff clean on the touched files.

Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #246, opened 2026-05-28
feat(core): deletion-preview counts (CAURA-696 PR A — OSS half)

Eldad-Caura:
## Summary

Adds the **read-only forecast** of what a subsequent hard-delete would remove. Mirrors `purge_tenant_data` (CAURA-689) one-for-one — same table set, same scoping columns — so the preview is a faithful "what will be deleted" view for the OPS UI (CAURA-693) and the customer self-delete flow (CAURA-697).

## Layers (OSS)

| Layer | Change |
|---|---|
| `core-storage-api` | New `PostgresService.count_tenant_data(tenant_id)` — `SELECT count(*)` per table over the same `_PURGE_TENANT_TABLES` + `_PURGE_ORG_KEYED_TABLES` constants. Returns the full per-table breakdown including zeros. New `POST /api/v1/storage/preview/tenant-counts` router — same VPC-internal trust model as the existing `purge` route. Hardened JSON-decode (`JSONDecodeError` + `UnicodeDecodeError`) and non-object-body guard — same defences PR #244 round 1 added to the suppression router. |
| `core-api` | New `CoreStorageClient.count_tenant_data` HTTP wrapper. New `POST /admin/org/preview-data` admin route. Body `{tenant_ids}`. Returns `{counts: {tenant_id: {table: row_count}}, failed: {tenant_id: error}}`. Per-tenant isolation; 200 clean / 207 Multi-Status partial. **No `lifecycle_audit` row** — a preview is a query, not a lifecycle event. Same validation + 100-tenant cap as `purge_org_data`. |

## Why a separate route from purge

- Read-only — should be cheap to call from the preview UI on every confirm-dialog open.
- No state side-effects → no audit row → no on-call noise.
- Same per-tenant isolation + 207 contract → admin-api can use one code path for both.

## What's not in this PR (PR B follow-up)

- Per-org count endpoint on `platform-storage-api` (enterprise rows: org, tenants, members, api_keys, etc.).
- Aggregator at `platform-admin-api`: `GET /api/v1/org-lifecycle/{org_id}/deletion-preview` that combines platform-storage (enterprise) + core-api (OSS, all tenants in the org) into a single unified payload for the OPS UI / self-delete UI.
- OPS UI panel itself — that's CAURA-693.

## Test plan

- [x] `core-storage-api/tests/test_preview.py` — happy path (3 seeded memories → `memories=3` + zeros), unknown-tenant returns all zeros, read-only, 422 on missing/empty/malformed/non-object body.
- [x] `uv run mypy src/` on core-api + core-storage-api → clean
- [x] `uv run ruff check` + `format --check` → clean
- [ ] CI integration suite against ephemeral postgres

### DIFF
--- a/core-api/src/core_api/clients/storage_client.py
+++ b/core-api/src/core_api/clients/storage_client.py
@@ -655,6 +655,15 @@ class CoreStorageClient:
         result = await self._post("/purge/tenant-data", {"tenant_id": tenant_id})
         return result.get("deleted", {})  # type: ignore[union-attr,return-value]
 
+    async def count_tenant_data(self, tenant_id: str) -> dict[str, int]:
+        """Per-table row count for ``tenant_id`` — drives the
+        deletion-preview panel (CAURA-696). Same table set + scoping
+        as ``purge_tenant_data`` so the preview is a faithful forecast.
+        Read-only.
+        """
+        result = await self._post("/preview/tenant-counts", {"tenant_id": tenant_id})
+        return result.get("counts", {})  # type: ignore[union-attr,return-value]
+
     async def count_active(self, tenant_id: str, fleet_id: str | None = None) -> int:
         params: dict[str, Any] = {"tenant_id": tenant_id}
         if fleet_id is not None:
--- a/core-api/src/core_api/routes/org_deletion.py
+++ b/core-api/src/core_api/routes/org_deletion.py
@@ -1,4 +1,4 @@
-"""Admin org-deletion endpoint (CAURA-689).
+"""Admin org-deletion endpoints (CAURA-689 + CAURA-696).
 
 ``POST /admin/org/purge-data`` — permanently purge all OSS (``public.*``)
 data for a set of ``tenant_ids`` (every tenant of an enterprise org being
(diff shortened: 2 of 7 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
Round-1 fix shipped in `8a7b919`:

- **🟡 Med — non-dict JSON body 500s on `preview_org_data`:**
  Added `isinstance(body, dict)` guard immediately after the `JSONDecodeError` / `UnicodeDecodeError` catch in `core-api/src/core_api/routes/org_deletion.py`. A top-level JSON array / string / number now surfaces as 422 `"request body must be a JSON object"` instead of `AttributeError` → 500. Comment back-references PR #244 round 1 where the storage-suppression router gained the same defence.

No new test added in `tests/`: the OSS repo doesn't currently have route-level admin tests for `org_deletion.py` (no `tests/test_admin_org_deletion.py` exists; the parallel `purge_org_data` route shipped without one in CAURA-689 either). The hardened storage-layer router `core-storage-api/src/core_storage_api/routers/preview.py` already has `test_rejects_non_object_json_body` and `test_rejects_malformed_body` in `tests/test_preview.py`, so the per-shape behaviour is exercised end-to-end — adding a duplicate route-only test here would be belt-and-suspenders.

Happy to add one if you'd like, but treating it as a follow-up to stand up a small `tests/test_admin_routes.py` infrastructure (since `purge_org_data` would benefit from the same coverage) feels like the cleaner shape than wedging a single test into nowhere.

`uv run mypy core-api/src/` clean; `uv run ruff check` + `format --check` clean on the touched file.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #234, opened 2026-05-26
fix(plugin): delegate compaction to OpenClaw runtime SDK to unwedge over-budget groups (CAURA-000)

ran-taig:
## Summary

**Urgent hotfix for v2.6.3 regression.** Customer escalation 2026-05-26: WhatsApp group sessions stuck running, looping on `memclaw_keystones` tool calls without finalizing, emitting `Skipping auto-reply: ... final-only` warnings. Group contexts reported over budget at 312k/272k and 292k/272k tokens via `openclaw status`. Direct chats with the same agent worked normally.

**Root cause:** `MemClawContextEngine.info.ownsCompaction = true` told OpenClaw we owned compaction, but `compact()` returned `undefined` — not a valid `CompactResult` per OpenClaw 2026.5.4 `plugin-sdk/src/context-engine/types.d.ts`. Over-budget sessions delegated to us, got nothing back, never shrank. Direct chats stayed under budget so the latent bug never surfaced; group chats accumulate transcript across participants and hit the wedge first.

**Fix:** stay `ownsCompaction: true` (truthful) and have `compact()` actually do its job by delegating to `delegateCompactionToRuntime` from `openclaw/plugin-sdk` — the same helper the legacy engine uses internally. Required a runtime SDK-discovery bridge (`openclaw-sdk-bridge.ts`) because OpenClaw's plugin loader routes `.js` files through native Node import, which bypasses the alias-map that would normally resolve the bare-spec import. Plus a defensive pre-check that returns a structured `CompactResult` fallback when `sessionId`/`sessionFile` are missing, blocking degenerate calls from reaching `compactEmbeddedPiSessionDirect` (which we observed can crash the gateway with synthetic params).

Note: I considered the simpler `ownsCompaction: false` change first. Per OpenClaw docs (`docs/concepts/context-engine.md:241`), `false` does NOT auto-fall-back to legacy compaction — it just disables compaction entirely. So we have to delegate explicitly to fulfill the contract.

## Files changed

- **`plugin/src/openclaw-sdk-bridge.ts`** (new, 203 lines): runtime SDK discovery via `process.argv[1]` walk → first `package.json` with `name: "openclaw"` → absolute-path dynamic import of `dist/plugin-sdk/index.js`. Structurally-typed surface, no build-time dep on `openclaw`. Cached resolver state for the process lifetime. `_resetSdkBridgeCache()` exposed for tests.
- **`plugin/src/context-engine.ts`**:
[...]

PR #247, opened 2026-05-28
fix(plugin): tolerate undefined config in ContextEngine constructor (CAURA-000)

ran-taig:
## Summary

**Follow-up hotfix to PR #234.** Customer escalation 2026-05-28 surfaced a *second* class of regression on the same deployment — a different agent (`dbaclaw`) hit:

```
[memclaw] assemble: unexpected error (returning safe fallback)
TypeError: Cannot read properties of undefined (reading 'tenantId')
    at getTenantPrefix (...context-engine.js:29:19)
    at getSessionKey   (...context-engine.js:32:26)
    at _assembleInner  (...context-engine.js:519:28)
    at MemClawContextEngine.assemble (...context-engine.js:452:20)
    at invokeWithLegacyCompat (registry-hc1-G3yP.js:103:10)
    at Object.mutableAgent.transformContext (model-context-tokens-z5hvDVkk.js:2679:22)
```

The outer try/catch from PR #212 caught it — no gateway crash, no wedge — but every turn that hit this path silently bypassed keystones, recall, and afterTurn. **Behavioral degradation with no user-visible symptom.**

## Root cause

`getTenantPrefix(config)` accessed `config.tenantId` with no nullish guard. The throw means `this.config === undefined` inside the engine instance. Validation against OpenClaw 2026.5.4 source (`registry-DFFgCbcm.js:241-289 resolveContextEngine`) confirmed the standard factory call always passes a populated `factoryCtx` — so the bug doesn't repro on upstream 2026.5.4. The customer's bundle hash (`registry-hc1-G3yP.js`) differs from upstream (`registry-DFFgCbcm.js`); likely a custom build whose legacy-compat invocation path reaches our factory with `undefined`. Regardless of the upstream cause, the fix is uncontroversial: tolerate the input.

## Fix (3 surgical edits)

1. **Constructor coerces `undefined` / `null` → `{}`**:
   ```typescript
   constructor(config: Record<string, unknown> | undefined | null) {
     this.config = config ?? {};
   }
   ```
2. **`getTenantPrefix` optional-chains**: `config?.tenantId || MEMCLAW_TENANT_ID || "default"`.
3. **`getSessionKey` matches**: same widened signature + optional chains + adapter for the already-tolerant `resolveAgentId`.

## What's untouched

PR #234's compaction-delegation work — `compact()`, `openclaw-sdk-bridge`, `info.ownsCompaction: true` — all unchanged. This hotfix only touches the pre-compaction assemble entry helpers.

## Tests (`runtime-contract.test.ts`)

New `MemClawContextEngine.constructor — undefined-config tolerance` describe block with 2 tests:
- `new MemClawContextEngine(undefined)` → `assemble` returns populated `AssembleResult` with `systemPromptAddition.length > 0`. The length assertion specifically distinguishes "helpers worked through to the inner success path" from "helpers threw but outer try/catch caught it."
- `new MemClawContextEngine(null)` — same tolerance, defense-in-depth.

## Test plan

- [x] `npm test` in `plugin/` — **261/261 pass** (259 prior + 2 new).
- [x] `pytest tests/test_plugin_source_manifest.py` — **11/11 pass** (no new src files, just edits).
- [x] Compiled dist verified — line 35 has `config?.tenantId`, constructor has `?? {}` coercion.
- [ ] Customer post-deploy: `journalctl --user | grep "memclaw\] assemble: unexpected error" | tail -5` returns empty (was firing repeatedly pre-fix).

## Version

2.6.3 → **2.6.5**. The 2.6.4 number is taken by PR #234's published artifact (already running on the customer's machine); release-please reverted `package.json` on main back to 2.6.3 (commit `fd253f76`) as part of its release-management workflow, so bumping from main's current 2.6.3 to 2.6.4 would collide. **2.6.5 gives the customer an unambiguous version string to verify the deploy.**

### DIFF
--- a/plugin/openclaw.plugin.json
+++ b/plugin/openclaw.plugin.json
@@ -2,7 +2,7 @@
   "id": "memclaw",
   "name": "MemClaw",
   "kind": "memory",
-  "version": "2.6.3",
+  "version": "2.6.5",
   "description": "Central persistent memory for OpenClaw agents — write, search, and entity lookup tools.",
   "entry": "dist/index.js",
   "skills": [
--- a/plugin/package.json
+++ b/plugin/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@caura/memclaw",
-  "version": "2.6.3",
+  "version": "2.6.5",
   "description": "OpenClaw plugin for MemClaw central memory",
   "license": "Apache-2.0",
   "private": true,
--- a/plugin/src/context-engine.ts
+++ b/plugin/src/context-engine.ts
@@ -67,17 +67,22 @@ const MAX_INGEST_WRITES_PER_SESSION = 10;
 const sessionBuffers = new Map<string, IngestMessage[]>();
 const sessionIngestCounts = new Map<string, number>();
 
-function getTenantPrefix(config: Record<string, unknown>): string {
-  return (config.tenantId as string) || MEMCLAW_TENANT_ID || "default";
+function getTenantPrefix(
+  config: Record<string, unknown> | undefined | null,
+): string {
+  // Optional chaining — config may be missing entirely (CAURA-000).
+  return ((config?.tenantId as string) || MEMCLAW_TENANT_ID || "default");
 }
 
-function getSessionKey(config: Record<string, unknown>): string {
+function getSessionKey(
+  config: Record<string, unknown> | undefined | null,
+): string {
   const tenantPrefix = getTenantPrefix(config);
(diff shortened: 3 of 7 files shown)


ran-taig:
@claude

ran-taig:
@claude

ran-taig:
@claude

ran-taig:
@claude

Eldad-Caura (approved):

PR #249, opened 2026-05-28
feat(plugin): bump MEMCLAW_KEYSTONES_TOKEN_CAP default 500 → 1500 (CAURA-000)

ran-taig:
## Summary

Customer escalation 2026-05-28: BankingClaw agent reports the `<keystone_rules>` block in its system prompt is being truncated every turn. Of 16 keystone rules in their tenant, only 12 are injected — the prompt ends with `... (4 more rules omitted)`.

Root cause: `MEMCLAW_KEYSTONES_TOKEN_CAP` defaults to **500 tokens** (~2000 chars). With ~120 chars/rule + header/footer overhead, only 12 rules fit. The 500-token default predates the contextEngine-slot work (PR #212/234/247) — it was set when in-prompt keystone injection was experimental. Now that keystones are a stable per-turn feature, and modern context windows start at 200k+ tokens, 500 is overcautious.

## Change

Single literal: `plugin/src/env.ts:269` `500` → `1500`. Doc comment updated to explain the new headroom (~20-30 rules at ~120 chars each) and reference CAURA-000 for the precedent.

## Tests

`plugin/src/env.test.ts` — new `env.ts default-value source pins` describe block. Reads the source file and asserts the `_readIntEnv("MEMCLAW_KEYSTONES_TOKEN_CAP", N, ...)` call has `N === 1500`. **Source-file** (not runtime) pin because `keystones.test.ts:26` overrides the env var at module-load time and `node --test` shares process env across files — a runtime read could see the polluted value. Error message in the assertion points future maintainers at the customer-impact reasoning so accidental refactor reverts fail loudly.

## Test plan

- [x] `npm test` — **262/262 pass** (261 prior + 1 new source-pin).
- [x] Existing `keystones.test.ts:26` override still exercises truncation logic at cap=120 explicitly — no regression of the truncation algorithm itself, only the default ceiling.
- [ ] Customer post-upgrade: ask the agent `Search your system prompt for "<keystone_rules>" — paste the count of bullets and any "N more rules omitted" line` — should report 16 bullets, no truncation line.

## Customer mitigation pending publish

`MEMCLAW_KEYSTONES_TOKEN_CAP=1500` in `~/.openclaw/plugins/memclaw/.env` + restart gives them the same result today without waiting for a release.

### DIFF
--- a/plugin/src/env.test.ts
+++ b/plugin/src/env.test.ts
@@ -117,3 +117,44 @@ describe("resolveTenantId — network failure handling", () => {
     }
   });
 });
+
+// --- Default-value source-pin tests ---
+//
+// We don't pin defaults via runtime import because other test files in
+// this suite override env vars at their module-load time (e.g.
+// ``keystones.test.ts:26`` sets ``MEMCLAW_KEYSTONES_TOKEN_CAP=120``),
+// and ``node --test`` shares process env across files — so any
+// runtime read could see a polluted value. Reading the source TS
+// file's literal value is robust to that, and also documents the
+// chosen default in a way that fails CI loudly if someone bumps it
+// without intent.
+
+describe("env.ts default-value source pins", () => {
+  test("MEMCLAW_KEYSTONES_TOKEN_CAP default literal is 1500 (CAURA-000)", async () => {
+    const { readFile } = await import("node:fs/promises");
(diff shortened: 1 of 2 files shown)


ran-taig:
@claude

Eldad-Caura (approved):

PR #250, opened 2026-05-30
fix(security): enforce fleet/agent scope on by-id memory access

erni-a:
## Summary

By-id memory handlers authorized on `tenant_id` alone, while the list/search
paths additionally enforce `scope_agent` ownership and the cross-fleet trust
ladder. This change brings the by-id surface in line with that contract so a
single shared check governs every access path.

## What changed

- **New `agent_service.authorize_memory_access` (+ `enforce_memory_read`)** —
  the single fleet/visibility contract shared by every by-id handler.
  `caller_agent_id is None` (a tenant/user credential with no gateway
  `X-Agent-ID`) keeps full tenant access, so the dashboard is unchanged; only
  agent-scoped credentials are constrained. The identity used is always the
  authenticated one (`auth.agent_id` / `_get_agent_id()`), never a
  caller-supplied parameter.
- **Reads** (`GET /memories/{id}`, `.../contradictions`, MCP `read`/`lineage`)
  → 404 when out of scope (mirrors how list/search simply omit the row).
- **Writes** (`DELETE`, `PATCH /status`, `PATCH`, MCP
  `transition`/`update`/`delete`/`bulk_delete`) → 403, via the existing
  trust ladder.
- **`/search` and MCP `recall`** fall back to the authenticated agent for
  fleet-forcing + `enforce_fleet_read` + visibility scoping when
  `filter_agent_id` is omitted (these paths return content, so scope matters
  there too).
- **`list_memories`** uses the authenticated identity for visibility/fleet
  scoping; the `?agent_id=` query param remains only the author filter.

## Tests

`tests/test_memory_byid_authz.py` — 18 tests: unit matrix on the helper,
the MCP `read` surface, and REST integration (cross-fleet denied/allowed by
trust, scope_agent author-only, and a no-agent dashboard regression guard).
Full non-benchmark suite passes locally (pre-existing, unrelated
`google-cloud-pubsub` import failures aside).

## Notes

- Behavior for tenant/user (dashboard) credentials is unchanged.
- Error code on read denial is 404 by design (avoids confirming a row exists
  outside the caller's scope).

### DIFF
--- a/core-api/src/core_api/mcp_server.py
+++ b/core-api/src/core_api/mcp_server.py
@@ -41,7 +41,12 @@ from core_api.db.session import async_session
 from core_api.errors import code_for_status
 from core_api.repositories import memory_repo
 from core_api.schemas import BulkMemoryCreate, BulkMemoryItem, MemoryCreate, MemoryUpdate
-from core_api.services.agent_service import enforce_fleet_write
+from core_api.services.agent_service import (
+    authorize_memory_access,
+    enforce_delete,
+    enforce_fleet_read,
+    enforce_fleet_write,
+)
 from core_api.services.audit_service import log_action, log_cross_tenant_read
 from core_api.services.entity_service import get_entity
 from core_api.services.memory_service import (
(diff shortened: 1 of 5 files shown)


erni-a:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #251, opened 2026-05-30
fix(security): require admin-trust for bulk/whole-tenant memory deletes

erni-a:
## Summary

Destroy-side sibling of #250. The bulk-delete endpoints authorized on **write
capability + tenant match only** — no trust check — so a routine trust-1 agent
credential could soft-delete an entire fleet or the whole tenant in one call
(broken function-level authorization). Same root cause as #250: the tenant is
the only enforced boundary; the intra-tenant trust ladder wasn't applied here.

## What changed

- **`delete_all_memories`** (`DELETE /memories`) and **`bulk_delete_by_ids`**
  (`POST /memories/bulk-delete`): when the caller is an **agent credential**
  (gateway `X-Agent-ID` present), require **admin-trust (≥ 3)** via the existing
  `enforce_delete` — the same gate as single-delete (#250) and the trust ladder.
- **Tenant/user credentials** (no `X-Agent-ID` → the tenant owner) keep full
  reach, so dashboard reset and the load-test tagged cleanup are unchanged.
- Deny-only — can only reject more requests; no data-loss/corruption risk.

## Tests

`tests/test_memory_byid_authz.py` (+3): trust-1 agent blocked (403) on both
endpoints, trust-3 agent allowed, tenant-key control unchanged (204). Full
non-benchmark suite green (pre-existing unrelated `google-cloud-pubsub`
import failures aside).

## Note for reviewers

This closes the demonstrated path (a trust-1 *agent* key). An optional
defense-in-depth — requiring org-admin even for a *tenant* key's *unfiltered*
whole-tenant purge — is deliberately **not** included here because it needs
`org_role` plumbed on the `X-Tenant-ID` auth branch and would change dashboard
"reset tenant" behavior. Flagging as a follow-up decision rather than bundling
a UX/contract change into a security hotfix.

### DIFF
--- a/core-api/src/core_api/routes/memories.py
+++ b/core-api/src/core_api/routes/memories.py
@@ -433,6 +433,13 @@ async def delete_all_memories(
     """
     auth.enforce_read_only()
     auth.enforce_tenant(tenant_id)
+    # BFLA gate: a bulk/whole-tenant delete by an *agent* credential requires
+    # admin-trust (>= 3), matching single-delete (enforce_delete) and the trust
+    # ladder — a routine trust-1 write key must not be able to wipe a fleet or
+    # the tenant. Tenant/user credentials (no gateway X-Agent-ID → the tenant
+    # owner) keep full reach (dashboard reset, tagged cleanup) unchanged.
+    if auth.tenant_id and auth.agent_id:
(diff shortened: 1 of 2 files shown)


erni-a:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #254, opened 2026-05-30
fix(security): gateway-shared-secret gate on the X-Tenant-ID auth path (CRITICAL-1, app side)

erni-a:
## Summary

Application-layer mitigation for the **gateway-bypass** finding (CRITICAL-1):
core-api is reachable on its public `*.run.app` URL and **Path 4 trusts
`X-Tenant-ID` / `X-Agent-ID` / `X-Readable-Tenant-IDs` with no credential of its
own** — safe only behind the gateway. A direct caller could set those headers
and impersonate any tenant (confirmed live: a keyless `GET` with `X-Tenant-ID`
returned data).

## What changed

- New setting **`GATEWAY_SHARED_SECRET`**. When set, **Path 4** additionally
  requires a matching **`X-Gateway-Secret`** header (injected by the gateway on
  every core-api location), checked with `hmac.compare_digest`, before honoring
  the identity headers. A direct/gateway-bypassing request lacks it → **401**.
- **Surgical**: only the unauthenticated header-trust path is gated. Admin-key,
  `MEMCLAW_API_KEY`, and standalone paths are untouched. Unset secret
  (OSS / standalone / dev) is a **no-op** — fully backward compatible.
- Reversible (unset the env var).

## Tests

`tests/test_gateway_secret_gate.py` — missing/wrong secret → 401; correct →
resolves the tenant; unset → no-op. Full non-benchmark suite green.

## Rollout (coordinated — this is one of three pieces)

1. **This PR** (core-api): deploy with `GATEWAY_SHARED_SECRET` set (Secret
   Manager) — but only **after** the gateway is sending the header, or set it
   empty first.
2. **Gateway** (enterprise, separate PR): `proxy_set_header X-Gateway-Secret
   <secret>` on every core-api/auth/audit location, same secret value.
3. Deploy order: gateway first (sending the header is harmless if core-api
   ignores it while the secret is unset), then set the secret on core-api.
   Verify the live PoC (`curl https://<core-api>.run.app/api/v1/memories
   -H 'X-Tenant-ID: t'`) now returns **401**, and the gateway path still 200s.

Defense-in-depth follow-up (Option B): lock core-api ingress to internal +
give the gateway a VPC connector, so it isn't publicly reachable at all.

### DIFF
--- a/core-api/src/core_api/auth.py
+++ b/core-api/src/core_api/auth.py
@@ -385,6 +385,19 @@ async def get_auth_context(
     # ── Path 4: X-Tenant-ID header (set by enterprise nginx / ingress) ──
     tenant_id = request.headers.get("x-tenant-id")
     if tenant_id:
+        # Perimeter check: this path TRUSTS the X-Tenant-ID / X-Agent-ID /
+        # X-Readable-Tenant-IDs headers with no credential of its own — safe
+        # only when the request came through the gateway. When a shared secret
+        # is configured, require the gateway-injected ``X-Gateway-Secret`` so a
+        # caller hitting core-api directly (e.g. its public run.app URL) cannot
+        # impersonate a tenant by setting the identity headers itself. No-op
+        # when unset (OSS / standalone / dev).
+        gw_secret = settings.gateway_shared_secret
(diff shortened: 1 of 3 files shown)


erni-a:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #264, opened 2026-06-01
fix(search): dedup per-tenant storage_search slot on entity-lookup fall-through (C10)

arkash20:
## Summary

The search pipeline has two storage round-trips that both acquired \`per_tenant_storage_slot(\"storage_search\", tenant_id)\`:

1. \`ClassifyQuery._collect_memories\` for the entity-lookup short-circuit's \`load_memories_by_ids\` call.
2. \`ExecuteScoredSearch\` for the \`scored_search\` call.

When the entity-lookup short-circuit **succeeds** (\`filtered_rows\` non-empty), the resulting plan sets \`skip_scored_search=True\`, ExecuteScoredSearch skips, and the request consumes 1 slot.

When the short-circuit **fails** (matched entities but their linked memories all got filtered out), classify falls through to keyword / semantic and ExecuteScoredSearch runs — at which point the request consumed **2 slots** against the same per-tenant bucket. That's the \"doubles queueing latency\" C10 named, now visible since A23 fixed the classify load.

The gap list called out two fix options: a sentinel through the pipeline context, or refactoring scored_search to take an outer slot. This PR takes the **sentinel route** — minimal scope, no API change.

## Changes

- \`ClassifyQuery._collect_memories\` gains a \`slot_acquired_marker: dict | None\` parameter. After the \`per_tenant_storage_slot\` \`async with\` returns successfully, it writes \`slot_acquired_marker[\"_storage_slot_acquired\"] = True\`. The \`execute\` site passes \`ctx.data\` as the marker.
- \`ExecuteScoredSearch.execute\` reads \`data.get(\"_storage_slot_acquired\")\`; when truthy, it calls \`scored_search\` directly without re-acquiring the slot. When falsy, it acquires as before.
- The sentinel is set ONLY after a successful \`load_memories_by_ids\`. If the load raises (caught by classify's outer try/except), \`_storage_slot_acquired\` stays unset and ExecuteScoredSearch acquires the slot normally on the fall-through.

Net effect: one logical search counts once against the per-tenant \`storage_search\` bucket, regardless of which path it took.

## Tests — black-box by a subagent

I implemented the fix, then briefed a subagent on the behavioural contract only (no access to the two pipeline-step files or \`per_tenant_concurrency.py\`). It produced \`tests/test_c10_storage_slot_dedup.py\` with 7 cases:

1. Entity-lookup SUCCESS path → classify acquires 1 slot, execute skips. Sentinel set.
2. Entity-lookup FALL-THROUGH → classify acquires 1, sentinel set, execute calls scored_search without re-acquiring.
3. Non-entity-lookup path → classify never enters \`_collect_memories\`, sentinel absent, execute acquires normally.
4. Sentinel suppresses only the slot, not the scored_search call.
5. Sentinel doesn't leak across requests (two ctx objects).
6. \`load_memories_by_ids\` failure does NOT mark the sentinel — execute re-acquires.
7. Plan-says-skip wins regardless of sentinel (defensive).

## Test plan

- [x] \`tests/test_c10_storage_slot_dedup.py\`: 7 tests, 0.50s, all pass.
- [x] Pipeline-tagged tests (40 across classify_query / scored_search / search_pipeline): all pass.
- [x] Full non-benchmark suite: **2645 passed** (+7), 0 regressions.
- [x] \`ruff check\` + \`ruff format --check\` clean.
- [x] Wet-tested against a fresh core-api build on local-dev: both \`POST /api/v1/search\` and \`POST /api/v1/recall\` return 200 — no regression on the normal (sentinel-absent) path.

## Notes

- Closes C10 in the canonical task list.
- Subagent caught an ASGI-bridge subtlety in the conftest (\`get_storage_client\` resolves to a real in-process bridge by default) and patched both step modules' bindings consistently. One iteration loop on the first run, all green on the second.

### DIFF
--- a/core-api/src/core_api/pipeline/steps/search/classify_query.py
+++ b/core-api/src/core_api/pipeline/steps/search/classify_query.py
@@ -100,6 +100,7 @@ class ClassifyQuery:
                         status_filter=status_filter,
                         valid_at=valid_at,
                         readable_tenant_ids=readable_tenant_ids,
+                        slot_acquired_marker=ctx.data,
                     )
 
                     if filtered_rows:
@@ -259,6 +260,7 @@ class ClassifyQuery:
         status_filter: str | None = None,
         valid_at: datetime | None = None,
         readable_tenant_ids: list[str] | None = None,
+        slot_acquired_marker: dict | None = None,
     ) -> list[types.SimpleNamespace]:
         """Load memories linked to graph-expanded entities, scored by hop distance."""
         all_entity_ids = list(entity_hops.keys())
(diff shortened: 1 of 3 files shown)


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #265, opened 2026-06-01
fix(api): accept metadata_mode as a query param on PATCH /memories/{id} (C7)

arkash20:
## Summary

\`PATCH /api/v1/memories/{memory_id}\` read \`metadata_mode\` only from the request body. A caller sending \`?metadata_mode=replace\` as a query string got the default merge behaviour silently — FastAPI drops unknown query params on the floor, and the body-side validator never sees the value.

C7 asked for either (a) accept it as a documented query param, OR (b) return 422 on unknown query params on PATCH. This PR picks (a) — targeted to the one endpoint, no behaviour change for callers who already use the body field.

## Changes

- \`update_memory_endpoint\` gains an optional \`metadata_mode\` query param with the same \`^(merge|replace)$\` pattern as the body field.
- **Body wins on conflict.** When body did NOT set \`metadata_mode\` and the query param IS set, the query value applies (and the body field is patched in place before the service call).
- Mirror the body-side \`metadata_mode_requires_metadata\` validator: if the caller supplies the query param without a matching \`metadata\` field in the body, return 422 with the same message. Sending only the mode flag is a no-op the existing validator already rejects on the body side; this PR makes the query-side symmetric.

## Truth table

| body \`metadata_mode\` | query \`metadata_mode\` | body \`metadata\` | outcome |
|---|---|---|---|
| no | no | yes | merge (default) — unchanged |
| no | yes (\`replace\`) | yes | **REPLACE via query** (new C7) |
| yes (\`replace\`) | no | yes | replace (existing) |
| yes (\`merge\`) | yes (\`replace\`) | yes | **merge — body wins** |
| no | yes (\`replace\`) | no | **422 — query-side mirror of body validator** |
| no | yes (\`bogus\`) | yes | 422 — FastAPI regex validation |

## Tests — black-box by a subagent

I implemented the fix, then briefed a subagent on the behavioural contract only (no access to \`routes/memories.py\`). It produced \`tests/test_c7_metadata_mode_query_param.py\` with 8 cases — one per row of the truth table plus an idempotency check.

## Test plan

- [x] \`tests/test_c7_metadata_mode_query_param.py\`: 8 tests, 1.14s, all pass.
- [x] Existing body-side tests (\`tests/test_api_memories.py\`) preserved.
- [x] Full non-benchmark suite: **2653 passed** (+8), 0 regressions.
- [x] \`ruff check\` + \`ruff format --check\` clean.
- [x] Wet-tested against a fresh core-api build on local-dev:
  - PATCH \`?metadata_mode=replace\` + body \`{metadata:{c:3}}\` → resulting metadata = \`{\"c\": 3}\` (replace applied via query).
  - PATCH \`?metadata_mode=replace\` + body \`{title:\"renamed\"}\` (no metadata) → 422 with expected message.

## Notes

- Closes C7 in the canonical task list.
- Doesn't broaden to 422-on-unknown-query-params (the alternative C7 option) — that would touch every PATCH and risk breaking clients that send tracking query params. Targeted accept is the minimal change.

### DIFF
--- a/core-api/src/core_api/routes/memories.py
+++ b/core-api/src/core_api/routes/memories.py
@@ -1334,6 +1334,18 @@ async def update_memory_endpoint(
     body: MemoryUpdate,
     tenant_id: str = Query(...),
     agent_id: str | None = Query(default=None),
+    metadata_mode: str | None = Query(
+        default=None,
+        pattern="^(merge|replace)$",
+        description=(
+            "C7: alias for the body field of the same name. Body wins on "
+            "conflict. When supplied as a query param without a matching "
+            "``metadata`` field in the body, returns 422 (same contract "
+            "as the body-side validator). Provided so SDK callers that "
+            "thread the toggle through a URL query don't silently get "
+            "the default merge behaviour."
+        ),
+    ),
     auth: AuthContext = Depends(get_auth_context),
     db: AsyncSession = Depends(get_db),
 ):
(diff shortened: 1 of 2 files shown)


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #266, opened 2026-06-01
feat(api): surface near_duplicate_of in default-mode write metadata (A21)

arkash20:
## Summary
- Adds `DetectNearDuplicate`, an advisory twin of `CheckSemanticDuplicate`, wired into the fast write pipeline between `CheckExactDuplicate` and `WriteMemoryRow`. On a high-similarity hit it stashes `metadata["near_duplicate_of"]`, `metadata["near_duplicate_similarity"]`, and `metadata["near_dup_check_ms"]` on the write response — **never 409-rejects** (strong-mode keeps that contract via `CheckSemanticDuplicate`).
- Uses the JUDGE threshold (`SEMANTIC_DEDUP_JUDGE_THRESHOLD = 0.85`), not AUTO (0.97). Wider band catches real paraphrases that sit at cosine 0.85–0.94; we skip the LLM-judge call the strong path makes above this threshold (advisory ≠ rejection, so false-positives are tolerable). Same A1 identifier-bearing pre-filter + same `semantic_dedup_enabled` tenant gate as the strong-side step.
- Closes A21 from the V4 task checklist: closes the asymmetry where default-mode agents could re-state the same fact and get two rows with no signal a near-duplicate already exists.

## Test plan
- [x] 14 blackbox unit tests in `tests/test_a21_detect_near_duplicate.py` (subagent-written, no implementation access — verifies: gate behavior, metadata key contract, embedding/visibility plumbing, identifier-prefilter skip, JUDGE-threshold pin, non-blocking semantics, latency-record contract)
- [x] Full non-benchmark suite (`pytest tests/ -x -q`): 2667 passed
- [x] `ruff check` + `ruff format --check` clean on the four touched files
- [x] Local wet test against `localhost` gateway: two semantically-equivalent writes ("Dana likes her tea steeped for exactly 3 minutes" → "Dana prefers her tea steeped for 3 minutes", 15s gap, both `write_mode=fast`) — second response carries `near_duplicate_of` pointing at the first id, `similarity=0.9766`, `check_ms=3.6`. Both writes accepted (201, as designed).

### DIFF
--- a/core-api/src/core_api/pipeline/compositions/write.py
+++ b/core-api/src/core_api/pipeline/compositions/write.py
@@ -6,6 +6,7 @@ from core_api.pipeline.steps.write import (
     CheckExactDuplicate,
     CheckSemanticDuplicate,
     ComputeContentHash,
+    DetectNearDuplicate,
     EmitMemoryTriple,
     LoadTenantConfig,
     MergeEnrichmentFields,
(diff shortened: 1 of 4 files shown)


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #270, opened 2026-06-01
ci: harden the shared review script

Eldad-Caura:
Hardens the shared review script (`.github/scripts/claude_pr_review.sh`) — follow-ups surfaced by the **enterprise repo's self-test review** of the identical script (the on-open review working as intended caught these):

- **Medium — jq silent-exit:** under `set -euo pipefail`, a non-JSON response from `claude` (warning banner, rate-limit HTML) made `jq` fail and killed the script *before* the empty-result guard, posting no error. Now guarded → posts a visible parse-error instead.
- **Low — 65 536-char comment limit:** a very large review would 422 on `gh api` then silently exit. Now truncated with a `_[truncated]_` marker.
- **Low — job-summary cost table:** restored the cost + input/output-token table to `$GITHUB_STEP_SUMMARY` (alongside the comment footer).

Declined (with reasoning, documented): the "PR-branch checkout runs the PR's script with secrets" finding — pinning to base would kill the self-test that *caught these very issues*; the real mitigation is the repo's "require approval for workflow runs" setting, not workflow code.

Keeps the script byte-identical across all three repos. This PR self-tests on open (it's a `.sh` change).

### DIFF
--- a/.github/scripts/claude_pr_review.sh
+++ b/.github/scripts/claude_pr_review.sh
@@ -46,13 +46,48 @@ RESULT=$(printf '%s' "$DIFF" | claude --print --model "$MODEL" --output-format j
   exit 1
 }
 
-REVIEW=$(printf '%s' "$RESULT" | jq -r '.result // empty')
-COST=$(printf '%s' "$RESULT" | jq -r '.total_cost_usd // empty')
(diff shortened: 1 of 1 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #17, opened 2026-04-27
ci: publish multi-arch Docker images to ghcr.io on release

ran-taig:
Adds the publish-docker workflow that fires on every published GitHub Release. See diff for full description.

Re-opened after the repo recreate (delete + push from bundle); the original PR #30 was lost when the previous repo was deleted as part of the security cleanup. Same content, same branch HEAD.

PR #272, opened 2026-06-02
CAURA-130 feat(contradiction): Path C kill-switch + entity-links preflight

ran-taig:
## Summary

Two safety controls on Path C contradiction detection, bundled because they touch the same code path and share testing infrastructure.

### L3.8 — Per-tenant retraction kill-switch

CAURA-128/129 tightened the retraction judge but left no operational escape valve. This PR adds `ResolvedConfig.retraction_enabled` (default `True`) backed by `organization_settings.settings["write"]["retraction_enabled"]`. Ops can flip per-tenant in the JSONB blob without a deploy.

Mirrors the existing `triple_emission_enabled` pattern (CAURA-123):
- `DEFAULT_SETTINGS["write"]["retraction_enabled"]` — `None` placeholder; resolver returns `True` when None/absent.
- `_LEAF_TYPES["write.retraction_enabled"] = bool` — type validation on PATCH.
- `ResolvedConfig.retraction_enabled` — property reading the blob with True default.
- Early-return at the top of `_attempt_path_c_retraction` with INFO log.

### L3.4 — Entity-links subject preflight (forward Path C)

The A1 #17 subject preflight only fires when **both** memories have non-NULL `subject_entity_id`. When one side is NULL (heuristic missed but entity-extraction populated `entity_links` with a subject-role entity), same-canonical-name distinct-entity pairs — the **`priya`-collision case** from the inline TODO at `contradiction_detector.py:1100` — silently fell through to the LLM judge, which mis-classified them as contradictions.

New plumbing:
- `_extract_subject_canonical_identity(entities)` — pure helper returning `(canonical_name, entity_type, entity_id)` of the first subject-role entity, or `None`. Identity keyed on `entity_id`.
- `_fetch_entity_context`'s hydrated row now includes `entity_id` (additive; doesn't break CAURA-129 consumers).
- Forward Path C preflight (post A1 #17 filter): fetches entity context for the fall-through set (candidates where `subject_entity_id` is NULL on at least one side), compares canonical identity by `entity_id`, drops on mismatch.

### Cost guards (per D3 in the plan)

| Case | Behaviour |
|---|---|
| Both `subject_entity_id` non-NULL | L3.4 stage **skips entirely** — legacy A1 #17 already covered it. Zero new round-trips. |
| At least one side NULL | Fetch via parallel `gather` wrapped in `asyncio.wait_for(timeout=5.0)`. 1 RT for new_memory + N RTs for fall-through candidates. |
| Storage error / timeout / missing subject identity | **Fail-open** — keep candidates, let the LLM judge decide. Conservative against losing real contradictions on a transient hiccup. |

## Tests (25 collected across both files)

**`tests/test_a4_13_path_c_retraction.py`** — 2 new kill-switch tests:
- `test_kill_switch_disabled_skips_retraction`
- `test_kill_switch_enabled_default_allows_retraction`

**`tests/test_caura_130_path_c_safety.py`** *(new — 14 tests)*:
- **6** `_extract_subject_canonical_identity` helper tests (first-subject-role, empty, no-subject, name fallback, skip missing entity_id, role match case-insensitive).
- **4** `ResolvedConfig.retraction_enabled` resolver tests (defaults, explicit True/False).
- **4** forward-path preflight tests:
  - Drops collision when `subject_entity_id` NULL on either side.
  - Keeps candidate when subjects truly match (same entity_id).
  - Skipped entirely when both subject_entity_id non-NULL (zero RT cost).
  - Fails open on storage error.

All CAURA-128/129 tests preserved unchanged.

## Validation

- [x] `python -m py_compile` clean.
- [x] `uv run ruff check core-api/src/` clean.
- [x] `uv run ruff format --check` clean.
- [x] `pytest --collect-only` collects 25 tests.
- [ ] CI runs full pytest (Postgres required for autouse fixture).
- [ ] Post-merge wet test:
  - Re-run `scripts/repro_contradictions_race.py` on dev → expect ≥95% detection across all gaps.
  - Optional: build a synthetic ``priya``-collision repro to confirm L3.4 drops the false-positive.

## Rollback

Single `git revert` of the merged squash. No DB migration, no schema change. Per-tenant kill-switch state lives in JSONB; clearing the key restores default-ON behaviour.

## Follow-ups (not in this PR)

- L3.4 in retraction path — already handled by CAURA-129's entity-aware judge; no change needed.
- Per-fleet kill-switch granularity — defer until tenant-level proves insufficient.
- Tier-3 `disputed` status + tiebreaker.

### DIFF
--- a/core-api/src/core_api/services/contradiction_detector.py
+++ b/core-api/src/core_api/services/contradiction_detector.py
@@ -1008,6 +1008,17 @@ def _fake_contradiction_check(new_content: str, old_content: str) -> bool:
 _ENTITY_CONTEXT_MAX_ENTITIES = 10
 _ENTITY_CONTEXT_NAME_MAX_CHARS = 100
 
+# CAURA-130 (L3.4) — upper bound on the fall-through set size before
+# the entity-links subject preflight starts fetching. Each candidate
+# costs one parallel storage round-trip; ``find_entity_overlap_candidates``
+# can theoretically return many rows for a high-fanout subject (popular
+# entity referenced by hundreds of memories), so cap the fan-out before
+# it becomes a thundering-herd risk on the storage API. Above the cap
+# we fail open (skip the L3.4 stage, let the LLM judge decide) rather
+# than drop everything — the legacy A1 #17 gate has already done what
+# it can.
+_ENTITY_LINKS_PREFLIGHT_MAX_CANDIDATES = 20
+
 
 def _format_entity_context(entities: list[dict]) -> str:
     """Render resolved entity rows into a readable block for the prompt.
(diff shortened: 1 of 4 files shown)


ran-taig:
@claude

ran-taig:
@claude

ran-taig:
@erni-a

erni-a (approved):

PR #273, opened 2026-06-02
CAURA-698 feat(search): decline ENTITY_LOOKUP short-circuit on over-broad FTS matches

eyal-bl:
## Summary

Etoro prod data (6h sample, 2026-06-02) showed the now-working ENTITY_LOOKUP short-circuit firing on **89.8%** of all `/search` calls — median 41 matched entities, max 5,415. The path is designed for *"user named a specific entity"* queries, not for narrative queries that incidentally match many entities. Once the count is high enough the short-circuit returns 50 hop-proximate-but-low-relevance memories and skips the rest of the pipeline (vector scoring, FTS rank, freshness), so the noise can't be re-filtered downstream.

Adds a count threshold gate: when entity FTS returns more than `ENTITY_LOOKUP_MAX_MATCHES = 100`, log + decline the short-circuit and fall through to the keyword/semantic cascade.

## Threshold choice

| Threshold | Caught (% of legit queries from etoro 6h CSV) | Tradeoff |
|---|---|---|
| 50 | ~58% redirect | Aggressive, risk of false fallthroughs in 50-100 zone |
| **100** | ~37% redirect | Sits well above legit median (41), catches over-broad tail |
| 200 | ~32% redirect | Conservative; medium-broad 100-200 zone stays misclassified |

100 chosen for first-pass — easy to tune from the constant later.

## Test plan

- [x] `test_entity_lookup_fires_when_match_count_at_threshold` — boundary case; exactly `ENTITY_LOOKUP_MAX_MATCHES` allowed through (gate is `>`, not `>=`).
- [x] `test_entity_lookup_falls_through_when_match_count_exceeds_threshold` — one over; downstream storage calls (`_expand_per_fleet`, `get_memory_ids_by_entity_ids`, `load_memories_by_ids`) MUST NOT execute; decision logged at INFO for ops visibility.
- [x] Negative control: both tests fail when the gate is stashed (ImportError on the new constant).
- [x] 32 targeted tests pass with the change in place.

## What this doesn't do

- Doesn't fix the "100 matches but all weak FTS rank" middle zone — that's gate 2 from the design discussion (surface `ts_rank_cd` from storage, threshold on best rank). Separate ticket if numbers still look bad after deploy.
- Doesn't fix the upstream OR-vs-AND tsquery semantics that produce the high match counts in the first place — separate consideration.
- Doesn't tighten the broad `except Exception:` at `classify_query.py:123`. Still a tracked follow-up.

### DIFF
--- a/core-api/src/core_api/constants.py
+++ b/core-api/src/core_api/constants.py
@@ -447,6 +447,23 @@ GRAPH_HOP_BOOST = {
     2: 1.1,
 }  # boost factor per hop distance (0 = direct match)
 GRAPH_MAX_BOOSTED_MEMORIES = 50  # cap on memories receiving graph boost (prevents popular-entity fan-out)
+# CAURA-698: cap on entity FTS matches that triggers the ENTITY_LOOKUP
+# short-circuit. Above this, the query almost certainly did not name a
+# specific entity (it matched broadly against a dense entity index); the
+# precision argument for entity_lookup breaks down at high match counts,
+# so fall through to keyword/semantic search instead.
+#
+# Calibrated against etoro prod data (2026-06-02, 6h, 98 user_or_assistant
+# queries). The distribution is bimodal with a wide empty gap between
+# ~75 and ~500 matches, so any threshold in that range is equivalent on
+# this dataset. Sampling the actual queries in each band revealed they
+# are all multi-keyword topical searches (e.g., "DeFi web3 smart contract",
+# "docker caddy dns", "marketing campaign email creative") — none name a
+# specific entity. T=20 lands the entity_lookup rate in the 20-30% target
+# band on user traffic while still allowing low-count cases that might
+# represent legitimate "name a thing" queries on other tenants' data.
+# Tune via prod measurement, not bench alone.
+ENTITY_LOOKUP_MAX_MATCHES = 20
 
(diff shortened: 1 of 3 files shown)


eyal-bl:
Updated `ENTITY_LOOKUP_MAX_MATCHES` from **100 → 20** based on the actual etoro data distribution (`2b73862`).

## Why 20

The 6h etoro dataset has a bimodal entity-count distribution with a clean empty gap from ~75 to ~500 matches, so any threshold in [75, 500] produces identical outcomes — including the original 100. To hit the 20-30% entity_lookup rate on real user traffic, the threshold had to drop into the 15-25 range.

| Threshold | Entity_lookup rate (user content) | Notes |
|---|---|---|
| 100 (original PR) | 51.0% | Falls in the empty-gap zone, no effect |
| 25 | 31.6% | Just above target |
| **20** (this commit) | **25.5%** | Middle of 20-30% target |
| 15 | 21.4% | Low edge of target |

Sampling queries in each band confirmed the dataset's user content is dominated by multi-keyword topical searches (`DeFi web3 smart contract`, `docker caddy dns`, `marketing campaign email creative`) rather than named-entity queries — so a lower threshold is data-supported.

## What's the same

Same code path, same tests (32 still passing, value-agnostic), same single signed-off commit. Just the constant value + its docstring changed.

eyal-bl:
@claude

eyal-bl:
@ran-taig please approve

ran-taig (approved):

PR #274, opened 2026-06-02
fix(plugin): context-engine auto-recall, smoke cleanup & post-upgrade allowlist drift

ran-taig:
## Summary

Three independent plugin defects surfaced while validating the MemClaw plugin against **OpenClaw 2026.5.28** on a wet-test VM (`openclaw-test-ran`). All three are verified — #1 and #2 live against `memclaw.net`, #3 via unit tests.

### 1. Auto-recall injected nothing (high impact)
The REST `POST /api/v1/search` endpoint returns `{ items: [...] }` (core-api `SearchResponse(items=...)`), but `assemble()` and `searchMemories` parsed **`.results`** → always `[]`. Context-engine auto-recall has therefore been silently empty (masked because the explicit `memclaw_recall` *tool* uses the MCP path, which does return `results`). The bootstrap smoke test parsed `.results` too, so it also misfired (`SMOKE TEST FAILED: … check EMBEDDING_PROVIDER`).
- New `parseSearchItems()` helper: `items` → `results` → bare-array fallback, never throws.
- Applied at all three `/search` consumers: recall, smoke test, memory-runtime search.
- **Verified live:** recall block now injects the matching memory (`RECALL block present: true`).

### 2. Smoke-test memory leaked on every restart
Bootstrap cleanup `DELETE /memories/{id}` omitted the required `tenant_id` query param → 422 (swallowed by `.catch(()=>{})`) → a `__health_check__` memory accumulated on each start. Hoist `tid` out of the `try` and pass `{ tenant_id }`.
- **Verified live:** 0 leaked memories after bootstrap (was: one per restart).

### 3. Tools dropped from `tools.alsoAllow` after a plugin upgrade
`autoFixAllowlist` already adds all `MEMCLAW_TOOLS`, but the gate ran it **once** (`.allowlist-applied` flag). So a tool added in a later plugin version (`memclaw_keystones`) never landed in `alsoAllow` on existing installs — and OpenClaw 2026.5.28's new `tools.profile` (grants core tools + `alsoAllow` only) then **stripped it**. Extracted pure `shouldRunAutoFix()` that also re-runs on drift (a missing tool, or an unclaimed contextEngine slot); `autoFixAllowlist` is idempotent so clean installs still no-op. Also added `memclaw_keystones` to the wet-test installer's tool list.

## Tests
- `parseSearchItems` — 5 cases (items / results / both / bare array / empty·null·primitive).
- `shouldRunAutoFix` — truth table (force / opt-out / first-run / missing-tool / unclaimed-slot / clean).
- Full plugin suite green: **273 passing**.

## Scope / non-goals
- The **compaction SDK-bridge** and **`assemble` contract** were verified *unaffected* by 2026.5.28 (runtime still calls `assemble()` and injects `systemPromptAddition`) — not touched here.
- **Skills-discovery truncation** (new 300/200 caps) is intentionally **out of scope**.
- `plugin/dist` is gitignored; build artifacts not committed.

### DIFF
--- a/plugin/src/config.test.ts
+++ b/plugin/src/config.test.ts
@@ -4,6 +4,7 @@ import {
   isMemorySlotClaimed,
   isContextEngineSlotClaimed,
   isMemclawFullyConfigured,
+  shouldRunAutoFix,
 } from "./config.js";
 import { getPluginDir } from "./paths.js";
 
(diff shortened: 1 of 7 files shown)


ran-taig:
@claude

ran-taig:
@claude

ran-taig:
@erni-a

erni-a (approved):

PR #276, opened 2026-06-02
CAURA-131 feat(contradiction): entity-aware Path C detection (close priya silence)

ran-taig:
## Summary

Symmetric mirror of CAURA-129. There, Path C's **retraction** judge ran the same LLM call as Path A and got stochastic disagreement. Here, Path C's **forward detection** judge runs the same base call and inherits its surface-qualifier blindness — failing to flag genuine within-subject contradictions when the entity extractor has correctly merged the subjects.

Wet-tested case (`scripts/repro_contradictions_collision.py` on dev, 2026-06-02):

```
Memory A: "Priya from AcmeCorp lives in Tel Aviv."
Memory B: "Priya from BetaIndustries lives in Haifa."
```

Entity extractor merged both `Priya` to entity `b5aca002-…`. `find_entity_overlap_candidates` returned the pair (entities match → candidate). Then the LLM call extracted `subject_a="Priya from AcmeCorp"` / `subject_b="Priya from BetaIndustries"` → `same_subject=false` → silent. **The data was one variable away from the prompt.**

## What changed

| Change | Detail |
|---|---|
| Lifted entity-context fetch above the L3.4 preflight | `_fetch_entity_context` runs once for `new_memory + all post-A1#17 candidates`, parallel `asyncio.gather` under 5s `wait_for`. Stored in `contexts: dict[memory_id, list[dict]]`. |
| L3.4 preflight now reads from `contexts` | Cheap dict lookup; same canonical-subject-mismatch drop semantics. |
| Detection LLM call selects per-candidate | Both contexts non-empty → `_llm_entity_aware_contradiction_check` (CAURA-129's entity-aware judge). Either empty / fetch failed / cap exceeded → fallback to base `_llm_contradiction_check`. |
| Cap reused | `_ENTITY_LINKS_PREFLIGHT_MAX_CANDIDATES = 20` bounds the full candidate-set fetch (was: fall-through-set only). |

## Tests (32 collected across 4 files)

**`tests/test_caura_131_entity_aware_path_c_detection.py`** *(new, 6 tests)*:

| Test | Asserts |
|---|---|
| Entity-aware judge used when both contexts present | `entity_aware_judge.assert_called_once()` + `base.assert_not_called()` + entity_ids in call args. The priya-shape repro. |
| Fallback when new memory has empty context | Base judge runs |
| Fallback when candidate has empty context | Base judge runs |
| Fallback when context fetch fails | Base judge runs (fail-open) |
| Fallback when candidate count > cap | Fetch NOT called; base judge runs N times |
| L3.4 collision still drops (regression guard) | Neither judge runs (preflight dropped) |

Existing CAURA-128/129/130 tests preserved unchanged.

## Wet-test artifacts committed

- `scripts/repro_contradictions_race.py` — adds `--trials N` for variance measurement. Pre-128/129/130 ~50% detection on net; post-128/129/130 90% on dev (5×4 trials).
- `scripts/repro_contradictions_collision.py` — new probe for the priya-shape. Pre-merge silent; expected post-merge to fire.
- `scripts/repro_contradictions_proper_noun.py` — minor lint cleanup carried over from S1 session.

## Validation

- [x] `python -m py_compile` clean on all changed files.
- [x] `uv run ruff check core-api/src/ scripts/ tests/...` clean.
- [x] `uv run ruff format --check` clean.
- [x] `pytest --collect-only` collects 32 tests across all 4 contradiction test files.
- [ ] CI runs full pytest (Postgres autouse fixture).
- [ ] **Post-merge wet test on dev**:
  - `scripts/repro_contradictions_collision.py` — expect contradiction to fire when extractor merges subjects.
  - `scripts/repro_contradictions_race.py --trials 5` — expect ≥95% (was 90%).

## Cost analysis

Zero new round-trips on the happy path — the L3.4 preflight already fetched `new_memory`'s context. CAURA-131 just extends the same `asyncio.gather` to all surviving candidates instead of only the fall-through set. On the post-A1-#17 candidate list, fetch is `1 + N` parallel storage calls, bounded by the existing cap (20).

## Rollback

Single `git revert`. No DB migration, no schema change, no flag flip.

## What's left (deferred)

- Path A **semantic** detection — runs PRE-entity-extraction, so `entity_links` typically empty. Fixing it requires a defer-and-retry pattern (significant refactor). Path C entity-aware should catch most of what Path A misses anyway.
- L3.2 object normalization, L3.5 N>2 transitive, L3.6 MCP coverage, L3.7 cross-tenant.

### DIFF
--- a/core-api/src/core_api/services/contradiction_detector.py
+++ b/core-api/src/core_api/services/contradiction_detector.py
@@ -1019,6 +1019,18 @@ _ENTITY_CONTEXT_NAME_MAX_CHARS = 100
 # it can.
 _ENTITY_LINKS_PREFLIGHT_MAX_CANDIDATES = 20
 
+# CAURA-131 — absolute upper bound on the TOTAL candidate count we'll
+# issue parallel entity-context fetches for (fall-through PLUS A1 #17-
+# matched candidates). The fall-through cap above bounds the L3.4
+# preflight set; this constant additionally bounds the entity-aware
+# detection-judge fetch. Without it, a popular entity producing 500
+# A1-#17-matched candidates would issue 501 parallel storage calls
+# even though fall-through is tiny. Set to 2x the preflight cap on the
+# heuristic that the entity-aware lift is marginal for A1-#17-matched
+# rows (subject identity is already confirmed by the column match),
+# so we don't need to pay the full thundering-herd budget for them.
+_ENTITY_LINKS_DETECTION_FETCH_MAX_CANDIDATES = _ENTITY_LINKS_PREFLIGHT_MAX_CANDIDATES * 2
+
 
 def _format_entity_context(entities: list[dict]) -> str:
     """Render resolved entity rows into a readable block for the prompt.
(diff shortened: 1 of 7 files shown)


ran-taig:
@claude

ran-taig:
@claude

ran-taig:
@claude

ran-taig:
@arkash20

ran-taig:
@erni-a

erni-a (approved):

PR #277, opened 2026-06-03
feat(lifecycle): run all lifecycle crons at a fixed UTC hour, not boot-relative intervals

arkash20:
## What

Every `core-operations` lifecycle cron now fires **once a day at a configurable UTC hour** (default **02:00 UTC**) instead of the generic run-then-sleep interval. The interval mode fired **once immediately at startup** then every ~24h *after the previous tick finished* — anchored to the last deploy and **drifting** by each tick's duration. Now they land in a predictable off-peak window, every redeploy.

This started as just the insights cron and was expanded (per request) to align **all six** lifecycle jobs:
`lifecycle-archive-expired`, `lifecycle-archive-stale`, `lifecycle-purge-soft-deleted`, `lifecycle-crystallize`, `lifecycle-entity-link`, `lifecycle-insights`.

## How

- **scheduler**: optional `delay_provider` on `register()`. When set, the loop sleeps `delay_provider()` (recomputed every cycle → drift-free) *before* each run and **skips the boot-time tick**. Interval mode is unchanged and remains available — backward compatible.
- **`seconds_until_next_utc_hour(hour)`** helper: strictly-future, ≤24h. The strict-future guarantee also means an aligned task can't hot-loop after a fast failure (next slot is a full day out once the hour passes).
- **config**: the three interval knobs (`lifecycle_{archive,purge,pipeline}_interval_seconds`) become per-job hour knobs (`lifecycle_{archive,purge,pipeline,insights}_run_at_hour`, int `0..23`, all default `2`), validated by one shared `field_validator`. Hours are independent, so operators can **stagger** the jobs across the night.
- **app**: every job registered with a `delay_provider`; archive-expired/stale share the archive hour, crystallize/entity-link share the pipeline hour.

No change to any consumer-side gate (`auto_*_enabled`, the insights activity gate, crystallize's >1000-active-memory threshold) — only **when** each daily tick fires.

## Behavior note

Because the boot-time tick is gone, after a deploy a job won't run until the next occurrence of its hour (≤24h later) rather than immediately. That's the intended trade for predictable off-peak scheduling.

## Test

- `ruff check` ✓ · `ruff format --check` ✓ · `mypy src/` ✓ · `pytest` ✓ (34 passed)
- New `test_scheduler.py` coverage: helper math (same-day / rollover / exact-boundary / bounds / out-of-range), aligned task does **not** fire at boot, sleeps-then-runs each cycle (drift-free), failing tick recovers without killing the loop.
- New `test_app.py`: all six jobs registered + wall-clock aligned, `*_run_at_hour` overrides thread through, config validator rejects out-of-range hours.

### DIFF
--- a/core-operations/src/core_operations/app.py
+++ b/core-operations/src/core_operations/app.py
@@ -26,7 +26,7 @@ from fastapi import FastAPI, HTTPException
 
 from common.structlog_config import configure_logging
 from core_operations.config import settings
-from core_operations.scheduler import scheduler
+from core_operations.scheduler import scheduler, seconds_until_next_utc_hour
 from core_operations.tasks import (
     run_archive_expired_tick,
     run_archive_stale_tick,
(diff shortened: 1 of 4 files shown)


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #278, opened 2026-06-03
feat(lifecycle): align all lifecycle crons to a fixed UTC hour, not boot-relative intervals

arkash20:
## What

Follow-up to #277. That PR moved the **insights** cron to a fixed UTC hour; this extends the same wall-clock alignment to **all** `core-operations` lifecycle crons:

`lifecycle-archive-expired`, `lifecycle-archive-stale`, `lifecycle-purge-soft-deleted`, `lifecycle-crystallize`, `lifecycle-entity-link` (insights already landed in #277).

Each now fires **once a day at a configurable UTC hour** (default **02:00 UTC**) via the scheduler's `delay_provider`, instead of a run-then-sleep interval that fired at startup and drifted by each tick's duration from the last deploy.

## How

- **config**: replace the three interval knobs (`lifecycle_{archive,purge,pipeline}_interval_seconds`) with per-job hour knobs (`lifecycle_{archive,purge,pipeline,insights}_run_at_hour`, int `0..23`, default `2`), validated by one shared `field_validator`. Hours are independent so operators can **stagger** the jobs across the night.
- **app**: every job registered with a `delay_provider`; archive-expired/stale share the archive hour, crystallize/entity-link share the pipeline hour.

The `delay_provider` mechanism and `seconds_until_next_utc_hour` helper already landed in #277 — this PR just wires the remaining five jobs to them.

No change to any consumer-side gate (`auto_*_enabled`, the insights activity gate, crystallize's >1000-active-memory threshold) — only **when** each daily tick fires.

## Behavior note

The boot-time tick is gone, so after a deploy a job won't run until the next occurrence of its hour (≤24h later) rather than immediately — the intended trade for predictable off-peak scheduling.

## Test

- `ruff check` ✓ · `ruff format --check` ✓ · `mypy src/` ✓ · `pytest` ✓ (34 passed)
- New `test_app.py`: all six jobs registered + wall-clock aligned (no boot tick), `*_run_at_hour` overrides thread through to the scheduler, config validator rejects out-of-range hours.

### DIFF
--- a/core-operations/src/core_operations/app.py
+++ b/core-operations/src/core_operations/app.py
@@ -40,45 +40,54 @@ logger = logging.getLogger(__name__)
 
 
 def _register_scheduled_tasks() -> None:
-    # Each lifecycle op gets its own registration so an outage on one
-    # can't silently mask the others; their audit rows stay
-    # independent.
+    # Every lifecycle op is wall-clock aligned to a fixed UTC hour via
+    # delay_provider: it fires once a day at its configured hour, never at
+    # startup, and never drifts from the boot time. The ``24 * 3600`` arg
+    # is the nominal daily period (documentation only — the delay_provider
+    # drives the actual firing). Each op gets its own registration so an
+    # outage on one can't silently mask the others; their audit rows stay
+    # independent. Hours are independently configurable so operators can
+    # stagger the jobs; they all default to 02:00 UTC.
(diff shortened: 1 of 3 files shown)


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #283, opened 2026-06-05
Reword fleet-scope 403s as policy, not trust levels

erni-a:
Cross-fleet read/write and delete/update denials now read as access policy (`fleet 'legal' is not readable by principals of fleet 'rnd'`) instead of exposing the internal `trust_level >= N` ladder. The 403 body is API copy that gets quoted/screenshotted; this matches how the boundary is described externally. Enforcement unchanged — message strings only in `core-api/src/core_api/services/agent_service.py` (5 messages). No tests assert them.


erni-a:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #291, opened 2026-06-07
feat(embedding): let the platform singleton target a self-hosted endpoint

Eldad-Caura:
## Summary

The core-worker embeds documents **exclusively** through the platform embedding singleton (`core_worker/consumer.py::handle_embed_request` → `get_platform_embedding`), never the registry. `init_platform_embedding` built `OpenAIEmbeddingProvider(api_key, model)` with **no `base_url`**, so the worker write path could only ever reach `api.openai.com` — there was no env var to point it at a self-hosted OpenAI-compatible endpoint (TEI, vLLM).

This is the gap behind a subtle failure mode: pointing core-api at TEI via the registry's `OPENAI_EMBEDDING_BASE_URL` moves only the **query** path. The **write** path (and thus the stored corpus) silently stays on the old vendor — and cross-vendor cosine is ~random, so search recall collapses.

## Change

`init_platform_embedding` now also reads `PLATFORM_EMBEDDING_BASE_URL`, `PLATFORM_EMBEDDING_SEND_DIMENSIONS`, and `PLATFORM_EMBEDDING_TRUNCATE_TO_DIM` and threads them into the provider, with the same two fail-loud guards the registry uses (the `base_url` ⊕ `send_dimensions` misconfigurations). A small `_reject_openai_config` helper single-sources the `openai-embedding-config` error tag across the rejection arms.

**Backward compatible:** with `PLATFORM_EMBEDDING_BASE_URL` unset, behaviour is identical to today (hosted OpenAI, `dimensions=` enabled).

## Why

Unblocks pointing the platform/worker embedding tier at a self-hosted TEI/bge-m3 endpoint — the staging local-embedder rollout (Cloud Run GPU) needs this before a coherent query+write cutover. Same knobs the registry already exposes, so both tiers can target the same endpoint.

## Test plan

- New `TestPlatformEmbeddingSelfHosted` group in `tests/test_platform_providers.py`: TEI provider built with `base_url` + `send_dimensions=false`; hosted-OpenAI default unchanged; and the four reject paths (base_url+send_dimensions, send_dimensions-false-without-base_url, invalid bool, non-int truncate).
- `tests/test_platform_providers.py` — **32 passed**. Embedding test surface (platform + service + stampede + cache + qemb) — **54 passed**.
- `ruff check` + `ruff format --check` clean on changed files; `mypy` introduces 0 new errors in `_platform.py`.
- Note: full repo suite / CI not run locally — CI is the gate here.

### DIFF
--- a/common/embedding/_platform.py
+++ b/common/embedding/_platform.py
@@ -29,6 +29,19 @@ _platform_embedding: EmbeddingProvider | None = None
 _platform_init_errors: list[str] = []
 
 
+def _reject_openai_config(msg: str, *args: object) -> None:
+    """Record a platform OpenAI-embedding misconfiguration and warn.
+
+    Single-sources the ``"openai-embedding-config"`` error tag shared by
+    every config-rejection arm. The caller must ``return`` after calling
+    this — the singleton is left ``None`` so the worker fails loud (drops
+    embed requests with a visible log) rather than silently embedding
+    against the wrong endpoint.
+    """
+    logger.warning(msg, *args)
+    _platform_init_errors.append("openai-embedding-config")
+
+
 def init_platform_embedding() -> None:
     """Build the singleton from ``PLATFORM_EMBEDDING_*`` env vars.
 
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #295, opened 2026-06-07
fix(events): env-scoped topic prefix for PubSubEventBus (no-op default)

Eldad-Caura:
## What
Cross-env Pub/Sub **fan-out**: prod + staging share one GCP project (`alpine-theory-469016-c8`) and publish/subscribe against the **same non-env-scoped topics** (`memclaw.memory.*`, `lifecycle.*`, `audit.*`, …), so every event is delivered to **both** envs' workers — the foreign env logs `target row missing; ack-dropping`. The EventBus already env-scopes **subscriptions** (`EVENT_BUS_SUBSCRIPTION_PREFIX`) but **not topics** — that asymmetry is the root cause. (Filed as the P0 cross-env-leakage report; verified real but ~P2 — no data loss, since each env has its own subscription and processes its own copy; the foreign env just wastes work + logs noise + is a latent cross-env-write risk.)

## This PR — the safe enabling primitive only (strict no-op)
A `_topic_name()` helper that prefixes the topic id with an env-scoped `EVENT_BUS_TOPIC_PREFIX`, applied at **both** the publish (`topic_path`) and subscribe (`sub_name`) call sites so an env's publishers and subscribers always agree. **Default (unset/empty) returns the raw topic name — byte-identical to today.** `topics.py`, terraform, and the events manifest are untouched, so **this changes nothing in prod on its own** (the cross-repo `pubsub-provisioning` validator is unaffected).

- `common/events/pubsub.py` — `topic_prefix` ctor kwarg + `_topic_name()` + both call sites.
- `common/events/factory.py` — read `EVENT_BUS_TOPIC_PREFIX` (empty default ⇒ no-op).
- `tests/test_events/test_pubsub_bus.py` — prefixed + no-op coverage (`40 passed`; ruff clean).

## Deferred — the actual cutover (NOT in this PR)
Activating cross-env isolation is a **staged terraform cutover**: create per-env topics (`{prefix}--<topic>` + DLQs) in each env's terraform, dual-publish, drain, set `EVENT_BUS_TOPIC_PREFIX` per env, flip all consumers (core-api, core-worker, platform-audit-api, platform-admin-api), then delete the old shared topics. Renaming terraform topics is destroy-recreate, so it must be phased (big-bang would crash consumers + lose in-flight messages). The terraform topic name **must** equal this helper's `{prefix}--{topic}` format.

@erni-a — over to you. This PR is intentionally a no-op until the cutover wires `EVENT_BUS_TOPIC_PREFIX`; it just lands the primitive so the cutover is a config/terraform change, not a code change under time pressure.

### DIFF
--- a/common/events/factory.py
+++ b/common/events/factory.py
@@ -54,7 +54,9 @@ def get_event_bus() -> EventBus:
             # behaviour) — a strict no-op. Set EVENT_BUS_TOPIC_PREFIX per-env (e.g.
             # "prod"/"staging") to isolate Pub/Sub topics across environments that
             # share one GCP project, eliminating cross-env message fan-out.
-            topic_prefix = os.getenv("EVENT_BUS_TOPIC_PREFIX")
+            # Strip so a fat-fingered secret (e.g. "prod ") can't produce an
+            # invalid GCP topic name like "prod --memclaw.memory.embedded".
+            topic_prefix = (os.getenv("EVENT_BUS_TOPIC_PREFIX") or "").strip()
             if topic_prefix:
                 kwargs["topic_prefix"] = topic_prefix
             # Environment identity for the cross-env fan-out guard. Two
--- a/common/events/factory.py
+++ b/common/events/factory.py
@@ -49,6 +49,14 @@ def get_event_bus() -> EventBus:
                     "EVENT_BUS_BACKEND=pubsub requires GCP_PROJECT_ID (or "
                     "EVENT_BUS_PROJECT_ID) and EVENT_BUS_SUBSCRIPTION_PREFIX"
                 )
+            kwargs: dict[str, Any] = {}
+            # Env-scoped topic prefix. Unset/empty ⇒ raw topic names (today's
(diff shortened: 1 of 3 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
Cross-link: **#297** is an immediate, zero-infra app-layer mitigation for the same cross-env fan-out — it ack-drops foreign-env messages (via a `source_env` attribute) *before* the embed/enrich call, buying relief while this topic-prefix cutover is gated. `caura-ai/caura-memclaw-enterprise#408` is Phase 1 of the terraform cutover this primitive enables (per-env topics + subs, dormant until `EVENT_BUS_TOPIC_PREFIX` is set).

Note: #295 and #297 touch the same files (`pubsub.py`, `factory.py`) → trivial mechanical rebase for whichever merges second. Full relationship + the architectural fork written up on #297.

Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #297, opened 2026-06-08
fix(events): drop cross-environment Pub/Sub fan-out messages

erni-a:
## Problem

When two environments share one GCP project, they also share its Pub/Sub topic namespace — topic names are not env-scoped (`topic_path(project_id, topic)` uses bare names). Each topic has a separate subscription per environment, so Pub/Sub **fans every published message out to both environments**.

The foreign-env subscriber then does real work: the core-worker's `handle_embed_request` / `handle_enrich_request` call `provider.embed(...)` / `enrich_memory(...)` on the **payload content** *before* the DB write, then the tenant-scoped `UPDATE … WHERE tenant_id/id` no-ops against a row that doesn't exist in that env's database. The result is wasted embedding/LLM spend, doubled compute, and a flood of `target row missing; ack-dropping` logs across the memory, lifecycle, audit and suppression topics.

**No data is lost** — each environment has its own subscription and successfully processes its own copy; the foreign copy is a duplicate that gets dropped.

## Fix

Make the event bus environment-aware:

- `publish()` stamps every message with a `source_env` Pub/Sub **attribute** (an attribute, not a body field, so a later subscription `filter` can match it server-side).
- `_pull_loop` ack-drops any message whose `source_env` differs from this process's env **before decode/dispatch**, so the foreign-env handler — and its provider call — never runs.
- Env identity resolves from `EVENT_BUS_ENV`, falling back to `ENVIRONMENT` (already set per service) → **activates with no infra change**.

The guard lives in the bus, so it covers **every** topic, not just the memory ones.

### Backward-compatible / safe rollout in any order
A message with **no** `source_env` attribute is processed exactly as today. Until every publisher stamps the attribute, unstamped messages keep their current fan-out behaviour rather than being dropped — so there is no data-loss window regardless of deploy order. If neither env var is set the guard is disabled (fail-soft) with a startup warning.

## Verification
- New tests in `tests/test_events/test_pubsub_bus.py`: attribute stamping (+ omission when env unset), env normalization, the foreign-env decision matrix, and `_pull_loop` behaviour (foreign dropped+acked+not-dispatched / same-env processed / no-attribute backward-compat).
- `tests/test_events/` — 78 passed; `mypy` clean on both changed files.

## Follow-up (separate, infra-only)
Once `source_env` is confirmed flowing on both envs, recreate each subscription with a server-side `filter` on `attributes.source_env`. That moves the drop server-side (eliminates even the wasted pull) and makes this code guard defense-in-depth. **Must land after this is fully deployed** — adding the filter before publishers stamp the attribute would drop everything.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -22,6 +22,12 @@
 
 # -- Mode -------------------------------------------------------------------
 IS_STANDALONE=true
+# Also the default environment identity for the Pub/Sub event bus: when
+# EVENT_BUS_BACKEND=pubsub, the bus stamps each message with this value and
+# drops messages published by a sibling environment that shares the same GCP
+# project's topics. Override independently with EVENT_BUS_ENV if the bus
+# identity must differ from ENVIRONMENT. Irrelevant for the default
+# in-process bus (single environment per process).
 ENVIRONMENT=development
 
 # -- Image version (docker-compose only) ------------------------------------
--- a/common/events/factory.py
+++ b/common/events/factory.py
@@ -59,8 +59,12 @@ def get_event_bus() -> EventBus:
             # the broader ``ENVIRONMENT`` if needed; falling back to
             # ``ENVIRONMENT`` means existing deployments (already set to a
             # per-environment value) get the guard with no infra change.
-            env = os.getenv("EVENT_BUS_ENV") or os.getenv("ENVIRONMENT")
-            if not env or not env.strip():
+            # Strip before the fallback so a whitespace-only EVENT_BUS_ENV
+            # (truthy) doesn't short-circuit the ``or`` and block the
+            # ENVIRONMENT fallback; empty collapses to None (guard disabled).
+            event_bus_env = os.getenv("EVENT_BUS_ENV", "").strip()
+            env = event_bus_env or os.getenv("ENVIRONMENT", "").strip() or None
+            if not env:
                 # Fail-soft: the guard stays disabled (every message is
                 # treated as same-env, i.e. today's behaviour). Warn so a
                 # misconfigured multi-env deploy is visible rather than
--- a/common/events/factory.py
+++ b/common/events/factory.py
@@ -7,6 +7,7 @@ subscribers in the same process share the same bus instance.
 
 from __future__ import annotations
 
+import logging
 import os
 import threading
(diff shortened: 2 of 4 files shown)


Eldad-Caura:
Cross-linking for context: this targets the **same cross-env Pub/Sub fan-out** as the topic-prefix primitive in #295 and the enterprise terraform cutover (`caura-ai/caura-memclaw-enterprise#408`) — but at a different layer/timeline, so they're **complementary, not redundant**.

- **This PR (#297)** — immediate, zero-infra app-layer mitigation: stamps `source_env` + ack-drops foreign-env messages *before* the embed/enrich call. Stops the wasted LLM/embedding spend now, activates from `ENVIRONMENT`, backward-compatible.
- **#295 + enterprise#408** — the permanent structural fix: env-scoped topics (`{env}--{topic}`) so Pub/Sub never fans out. Only effective after the multi-day, prod-gated terraform cutover completes.

**Relationship:**
- Once the #295/#408 cutover lands, this guard becomes a **no-op** (no foreign message ever reaches a subscription) — harmless defense-in-depth.
- This PR and #295 edit the **same files** (`pubsub.py` `__init__`/`publish`/`_pull_loop`, `factory.py` env-read + `kwargs`) — #295 adds `topic_prefix`/`EVENT_BUS_TOPIC_PREFIX`, this adds `env`/`EVENT_BUS_ENV`. No logical conflict; whichever merges second needs a trivial mechanical rebase.
- This PR's **follow-up** (server-side subscription `filter` on `attributes.source_env`) is an *alternative end-state* to the #408 topic-rename cutover — both achieve "no foreign processing." Worth deciding whether hard topic isolation (#408) is still warranted once this + the filter are in place, or whether #408 downgrades to nice-to-have. (Architectural call for @erni-a / Eldad.)

Verified the code matches the description and is correct (owning env still processes its own copy; true no-op in every backward-compat case). Underlying-bug verdict unchanged across all three: real misconfig, ~P2, **no data loss** (fan-out).

Eldad-Caura:
@claude

Eldad-Caura:
@arkash20 please approve

Eldad-Caura (approved):

PR #298, opened 2026-06-08
fix: return real columns from cross-link insert instead of nonexistent id

arkash20:
## The error (prod)

```
Error bulk-inserting 1 cross-links for tenant olivier-obuchowski-3d53f3
asyncpg.exceptions.UndefinedColumnError: column "id" does not exist
```

## Root cause

`DiscoverCrossLinks` inserts into `memory_entity_links` with `RETURNING id`, but that table is a **pure junction table** — composite PK `(memory_id, entity_id)`, no surrogate `id` column (`common/models/entity.py:53-62`). So `RETURNING id` references a column that doesn't exist and Postgres rejects the statement.

**Latent since v1.0.0** (`7456e23`), not a regression. It only surfaced now because (1) the insert branch only fires when a memory has <3 entity links *and* a similar entity clears the 0.75 threshold *and* text-verify passes, and (2) the existing tests mock the DB, so the invalid SQL was never validated against a real schema. Tenant `olivier-obuchowski-3d53f3` is the first prod write to reach the branch.

**Blast radius:** non-fatal. The step catches the exception and returns `FAILED`; the parent memory write still succeeds. Effect = degraded cross-link discovery, no data loss.

## Fix

```diff
-                    RETURNING id
+                    RETURNING memory_id, entity_id
```

Behaviour-preserving for the count: with `ON CONFLICT … DO NOTHING`, `RETURNING` only emits actually-inserted rows, so `links_created` stays accurate — it just selects columns that exist.

## Test

Mock-based tests can't catch a bad column name, so this adds a SQL-shape regression test (`test_discover_insert_returns_real_columns_not_id`) asserting the INSERT returns `memory_id, entity_id` and never `RETURNING id` again — mirroring the existing CAURA-675 guard in the same file.

## Verification

- `ruff check` clean ✅
- `pytest tests/pipeline/test_entity_linking.py tests/test_entity_linking_wiring.py` → 10 passed ✅

### DIFF
--- a/core-api/src/core_api/pipeline/steps/entity_linking/discover_cross_links.py
+++ b/core-api/src/core_api/pipeline/steps/entity_linking/discover_cross_links.py
@@ -136,11 +136,15 @@ class DiscoverCrossLinks:
         links_created = 0
         if to_insert:
             try:
+                # memory_entity_links has a composite PK (memory_id, entity_id)
+                # and no surrogate `id` column, so RETURNING must reference real
+                # columns. With ON CONFLICT DO NOTHING, only actually-inserted
+                # rows are returned, so the count below stays accurate.
                 insert_link_returning = text("""
                     INSERT INTO memory_entity_links (memory_id, entity_id, role)
                     VALUES (:memory_id, :entity_id, 'mentioned')
                     ON CONFLICT (memory_id, entity_id) DO NOTHING
-                    RETURNING id
+                    RETURNING memory_id, entity_id
                 """)
                 result = await ctx.require_db.execute(insert_link_returning, to_insert)
                 links_created = len(result.all())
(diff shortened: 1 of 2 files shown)


arkash20:
### Verification update

**Unit regression test (red → green):** `test_discover_insert_returns_real_columns_not_id` confirmed to fail against the buggy `RETURNING id` and pass against the fix.

**Wet test against real Postgres** (`pgvector/pg16`, real `memory_entity_links` schema, rolled-back txns, valid FK pair):

| Step | Statement | Result |
|------|-----------|--------|
| Before (bug) | `... RETURNING id` | ❌ `ERROR: column "id" does not exist` (matches prod) |
| After (fix) | `... RETURNING memory_id, entity_id` | ✅ success, returned 1 row |

Confirmed schema: `memory_entity_links(memory_id, entity_id, role)`, PK `(memory_id, entity_id)`, no `id` column.

erni-a (approved):

Eldad-Caura (approved):

PR #301, opened 2026-06-08
fix(search): bound graph BFS frontier to prevent unbounded IN-clause (CAURA-000 F4)

ran-taig:
## Summary

Server-side fix for the F4 root cause identified during the customer-log investigation. F3 already landed via PR #298, so this PR is **F4-only**.

## Root cause (proven, not inferred)

Customer log capture (read-only ssh, `core-storage-api` container, tenant `etoro0-40f488`, 2026-06-07T16:52:05Z):

```
WHERE relations.tenant_id = $1::VARCHAR
  AND (relations.fleet_id = $4::VARCHAR OR relations.fleet_id IS NULL)
  AND relations.from_entity_id IN ($5::UUID, $6::UUID, ..., $42146::UUID)
[parameters: ('etoro0-40f488', 'yoniclaw-fleet', ..., 42046 parameters truncated ...)]
(Background on this error at: https://sqlalche.me/e/20/rvf5)
"severity": "ERROR", "message": "Exception in ASGI application\n"
```

→ **42,146 bind parameters in a single statement** — well past asyncpg's safe window. Manifests upstream as 7 `recall failed: MemClaw API 500` and 3 `SMOKE TEST ERROR` events in the gateway log over a 2-day capture.

## Mechanism

`entity_expand_graph` (`postgres_service.py:3089`) does BFS over the relations graph:

```python
frontier = set(seed_entity_ids)
for hop in range(1, max_hops + 1):
    fwd = select(...).where(Relation.from_entity_id.in_(frontier))   # ← N bind params
    rev = select(...).where(Relation.to_entity_id.in_(frontier))
    ...
    frontier = neighbor_weights.keys() - ...   # ← unbounded growth
```

A `GRAPH_MAX_EXPANDED_ENTITIES = 200` constant existed but was **applied in the wrong place**: only the OUTPUT of `entity_expand_graph` was capped (in `classify_query.py:287-291`), not the FRONTIER inside the BFS. By the time we got to the truncation, the giant SQL had already fired. And the *other* search path that consumes the BFS output (`parallel_embed_entity_boost.py:91-94`) had no cap at all.

## Fix

Two-site fix, one logical change:

1. **`entity_expand_graph`** (core-storage-api): cap the frontier at `GRAPH_MAX_EXPANDED_ENTITIES` BEFORE building the per-hop `IN (...)` query. Order by `(weight desc, id asc)` so the cap keeps the highest-weighted edges deterministically; ID tiebreak makes selection stable.

2. **`parallel_embed_entity_boost`** (core-api): apply the same cap defensively at the call boundary BEFORE `get_memory_ids_by_entity_ids`. Mirrors the existing cap in `classify_query._load_graph_memories` so both search paths behave consistently. Order by `(hop asc, weight desc)` — closest entities win.

Both sites log `INFO` when the cap fires so we get operator visibility on hot-tenant cap rates.

## Numbers

| | Bind params | Outcome |
|---|---:|---|
| Customer's failing query, before fix | 42,151 | crashed (`Exception in ASGI application`) |
| Same query, after fix | 205 | safe (≪ asyncpg ~32K ceiling) |

## Risks & trade-offs

- **Truncation can drop low-weight edges.** Since we order by weight, dropped edges are the lowest-relevance ones. The other search path already does this and we have no recall-regression evidence; this PR brings the second path in line.
- **No schema change, no migration, no constant change.** `GRAPH_MAX_EXPANDED_ENTITIES = 200` was already in `common/constants.py`. We just apply it at the right places.
- **Doesn't migrate to recursive CTE.** That's the architecturally cleaner fix (single round-trip, no parameter explosion possible) but a bigger change. Deferred to a follow-up; will re-evaluate based on post-deploy cap-fire rates.

## Tests

New `tests/test_caura_000_graph_frontier_cap.py` (8 unit tests):

- Constant sanity (cap below asyncpg window, above realistic seed counts)
- `parallel_embed_entity_boost` cap applied when entity_hops > cap
- `parallel_embed_entity_boost` no-op when within cap
- `parallel_embed_entity_boost` ordering keeps closer hops (regression guard: a sort-key inversion would silently drop seed-hop entities)
- `entity_expand_graph` cap keeps highest-weight edges
- `entity_expand_graph` cap is deterministic across calls (ID tiebreak)
- `entity_expand_graph` no-op when within cap

Regression sweep across neighboring suites: 37/37 pass (D7 + P3-3 + P3-2 + new file).

## Test plan

- [x] Unit tests: 8/8 pass
- [x] Regression sweep: 37/37 pass
- [x] Customer-scenario simulation: 42,151 → 205 bind params
- [ ] Customer rollout: ship via next on-prem release; expect `recall failed: MemClaw API 500` (search pipeline) count to drop to near-zero. Watch the new `entity_expand_graph: frontier capped at 200` INFO lines — if they fire >5/sec in steady state we should reconsider the cap or move to a CTE traversal.

### DIFF
--- a/core-api/src/core_api/pipeline/steps/search/parallel_embed_entity_boost.py
+++ b/core-api/src/core_api/pipeline/steps/search/parallel_embed_entity_boost.py
@@ -17,6 +17,7 @@ from core_api.clients.storage_client import get_storage_client
 from core_api.constants import (
     GRAPH_HOP_BOOST,
     GRAPH_MAX_BOOSTED_MEMORIES,
+    GRAPH_MAX_EXPANDED_ENTITIES,
     RECALL_BOOST_CAP,
 )
 from core_api.pipeline.context import PipelineContext
@@ -89,6 +90,28 @@ async def _entity_boost_via_storage(
 
         if matched_entity_ids:
             all_entity_ids = list(entity_hops.keys())
+
+            # Bound the IN-clause size before the downstream storage call.
+            # ``entity_expand_graph`` already applies the same cap internally
+            # (CAURA-000), but a future change there must not be able to
+            # silently feed a 40K-UUID list into ``get_memory_ids_by_entity_ids``
+            # — that's the exact 42,146-bind-parameter SQL that crashed the
+            # customer's request on 2026-06-07. Order by (hop asc, weight desc)
+            # so the closest, highest-weighted entities are kept — mirrors the
+            # cap applied in ``classify_query._load_graph_memories`` so the two
+            # search paths behave consistently.
+            if len(all_entity_ids) > GRAPH_MAX_EXPANDED_ENTITIES:
+                all_entity_ids = sorted(
+                    all_entity_ids,
+                    key=lambda eid: (entity_hops[eid][0], -entity_hops[eid][1]),
+                )[:GRAPH_MAX_EXPANDED_ENTITIES]
+                logger.info(
+                    "parallel_embed_entity_boost: entity-id list capped at %d (tenant=%s dropped=%d)",
+                    GRAPH_MAX_EXPANDED_ENTITIES,
+                    tenant_id,
+                    len(entity_hops) - GRAPH_MAX_EXPANDED_ENTITIES,
+                )
+
             raw_links = await sc.get_memory_ids_by_entity_ids(
                 [str(eid) for eid in all_entity_ids],
             )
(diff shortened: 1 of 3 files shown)


ran-taig:
@claude

ran-taig:
@erni-a

erni-a (approved):

PR #305, opened 2026-06-08
fix(fleet): bound deploy-queue runaway when a previous deploy is acked (CAURA-000)

ran-taig:
## Summary

The auto-upgrade gate in `_maybe_queue_auto_upgrade` was blind to `acked`-but-not-completed deploy commands. Every heartbeat queued another duplicate, the plugin processed it, restarted via `systemctl` (sending SIGTERM to itself), and the cycle repeated every 60 seconds — exactly matching the customer-reported "SIGTERM every 60 seconds" on the affected gateway.

This PR adds a second gate that consults the repository for any in-flight deploy (`status IN (pending, acked)`, `created_at` within a 10-minute window) and short-circuits if one is found.

## Customer ground truth (read-only ssh)

Postgres on `134.98.155.239`, `2026-06-08`:

| Metric | Value |
|---|---:|
| Deploy commands stuck at `status='acked'` | **1,381** |
| Stuck-acked on a single node (`cb457f2c-…`) | **1,223** |
| Timestamps between consecutive stuck commands | 60s |
| All target_version | `"2.8.1"` |

"Removing memclaw plugin solved it" because the plugin's `setInterval(sendHeartbeat, 60_000)` is what pulled the perpetually-fresh deploy command and triggered the restart cycle. No plugin → no heartbeat → no pull → no restart.

## Mechanism (full trace in commit message)

1. Heartbeat handler fetches `pending` commands → empty (previous deploy was `acked`)
2. Gate `_has_recent_deploy_command_from_list` sees no deploy in pending → allows queue
3. New deploy queued (`status=pending`) → re-fetched → `ack_commands` transitions to `acked` → sent to plugin
4. Plugin's `processCommand("deploy")` schedules `systemctl restart` in 2s
5. The plugin POSTs the result, but if the POST takes >2s the restart's SIGTERM kills it mid-flight
6. Command stays at `acked` forever
7. 60s later, repeat from step 1

## Fix (smallest blast radius — backend only)

| | File | Change |
|---|---|---|
| 1 | `core-api/.../repositories/fleet_repository.py` | New `has_recent_in_flight_deploy` method — direct SQLAlchemy query |
| 2 | `core-api/.../routes/fleet.py` | Import `fleet_repo`, define `DEPLOY_IN_FLIGHT_WINDOW = 10 min`, add a second gate check inside `_maybe_queue_auto_upgrade` AFTER the cheaper pending-list check, falls open on repo exception |

The cheaper pending-list check still fires first (no DB roundtrip on the happy path with a fresh pending deploy already in the list). The new DB check only runs when the auto-upgrade is otherwise eligible. The 10-minute window means a genuinely-abandoned deploy auto-recovers without operator intervention.

## What this PR explicitly does NOT do

- **No plugin change.** The plugin's `setTimeout(systemctl restart, 2000)` race with the result POST is a real bug that should be fixed in a follow-up PR. But the backend gate makes the customer's loop bounded regardless of plugin version.
- **No DB cleanup.** The 1,381 existing stuck-acked rows on the customer's prod will be naturally bypassed by the new gate (they're outside any future 10-min window). Optional follow-up: `UPDATE fleet_commands SET status='failed' WHERE status='acked' AND created_at < now() - interval '1 hour'` if the customer wants the queue trimmed.
- **No constant changes.** `MIN_RECOMMENDED_PLUGIN_VERSION`, retry semantics, ack timing — all unchanged.

## Wet test (against real customer prod data)

Read-only SQL simulation on `memclaw-postgres-1` (no writes; queries only):

```
total_queued_deploys | would_have_been_blocked_by_new_gate | pct_blocked
---------------------+-------------------------------------+-------------
                1223 |                                1218 |       99.6
```

The 5 unblocked are the **first** commands in the cycle (gate has nothing to see yet — first attempt must pass). After that the gate would have suppressed every follow-up for 10 minutes. Result: 60s → 10min cap on the SIGTERM cadence, AND zero runaway once a deploy completes successfully.

## Tests

5 new in `tests/test_fleet_heartbeat_auto_upgrade.py`:
- `test_maybe_queue_auto_upgrade_skips_when_acked_deploy_in_flight` — headline behavior
- `test_maybe_queue_auto_upgrade_queues_when_no_in_flight` — happy path unchanged
- `test_maybe_queue_auto_upgrade_fails_open_on_repo_error` — DB hiccup safety
- `test_maybe_queue_auto_upgrade_skips_pending_before_querying_repo` — gate ordering (cheaper check first)
- `test_deploy_in_flight_window_is_sensible` — pins the constant value range

| Test scope | Result |
|---|---|
| `test_fleet_heartbeat_auto_upgrade.py` (full file) | **51/51 pass** (was 46, +5) |
| Broader `pytest -k "fleet or heartbeat" -m unit` | 24/24 pass, 0 regressions |
| `ruff check` + `ruff format --check` | clean |

## Test plan

- [x] Unit tests: 51/51 pass in the auto-upgrade file
- [x] Broader fleet/heartbeat sweep: 24/24 pass
- [x] Wet test against customer prod data (read-only SQL): 1,218/1,223 (99.6%) of past stuck commands would have been blocked
- [ ] Customer redeploy: expect `[memclaw] command deploy reported` log lines to drop to ≤1 per 10 min on previously-stuck nodes; `[fleet.heartbeat] skipping auto-upgrade for node=...` INFO lines start appearing

### DIFF
--- a/core-api/src/core_api/repositories/fleet_repository.py
+++ b/core-api/src/core_api/repositories/fleet_repository.py
@@ -139,6 +139,49 @@ class FleetRepository:
         )
         return result.scalars().all()
 
+    async def has_recent_in_flight_deploy(
+        self,
+        db: AsyncSession,
+        *,
+        node_id: UUID,
+        since: datetime,
+    ) -> bool:
+        """True if a ``deploy`` command for this node is still in flight.
+
+        "In flight" means status in (``pending``, ``acked``) and
+        ``created_at >= since``. Used by the auto-upgrade gate to suppress
+        queueing duplicate deploys when a previous one has been sent to
+        the plugin but never reported back as completed/failed.
+
+        Why this matters (CAURA-000): the existing pending-only gate
+        (``_has_recent_deploy_command_from_list``) is blind to ``acked``
+        commands. After the heartbeat handler transitions a command
+        ``pending`` → ``acked`` and ships it to the plugin, the gate
+        sees an empty pending list on the next heartbeat and happily
+        queues another deploy — even if the plugin's last deploy is
+        still in flight or got killed mid-result-POST by its own
+        ``systemctl restart``. Customer prod-data observation
+        (2026-06-08): 1,381 deploy commands stuck at ``acked``, one new
+        per heartbeat, on a single node — exactly the customer-reported
+        "SIGTERM every 60s" cycle.
+
+        The ``since`` window bounds the gate: an acked command older
+        than ~10 min is assumed abandoned, and a retry is allowed. This
+        recovers nodes whose plugin crashed mid-deploy (lost result-POST)
+        without needing operator intervention to clear stale rows.
+        """
+        result = await db.execute(
+            select(FleetCommand.id)
+            .where(
+                FleetCommand.node_id == node_id,
+                FleetCommand.command == "deploy",
+                FleetCommand.status.in_(("pending", "acked")),
+                FleetCommand.created_at >= since,
+            )
+            .limit(1)
+        )
+        return result.scalar_one_or_none() is not None
+
     async def ack_commands(self, db: AsyncSession, *, command_ids: list[UUID], now: datetime) -> None:
         if not command_ids:
             return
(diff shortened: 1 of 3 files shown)


ran-taig:
@claude

ran-taig:
@erni-a

erni-a (approved):

PR #306, opened 2026-06-08
fix(plugin): schedule restart AFTER result POST resolves (CAURA-000)

ran-taig:
## Summary

`processCommand` was scheduling `setTimeout(systemctl --user restart, 2000)` INSIDE the deploy/restart success branches, BEFORE awaiting the result POST. If the POST took longer than ~2s (slow network, retried connection, queued behind other heartbeat traffic), the systemctl SIGTERM killed it mid-flight and the backend never saw `status="done"` — the command stayed at `status="acked"` forever.

Combined with the backend's pending-only auto-upgrade gate (fixed separately in #305 by adding an in-flight-deploy check), this race produced the customer's observed "SIGTERM every 60 seconds" on one gateway. This PR removes the race at the source so new deploys don't accumulate in the first place; #305 handles cleanup/safety on already-stuck-acked commands. Both together close the loop.

## Mechanism

Before:
```
deploy/restart branch:
  setTimeout(systemctl restart, 2000)    ← T=0: scheduled
  ...
  // fall through to end of processCommand
  await POST result                       ← T=X: depends on POST latency
                                          ← T=2000: setTimeout fires → SIGTERM
                                          // if X > 2000, POST is killed mid-flight
```

After:
```
deploy/restart branch:
  shouldRestart = true                    ← T=0: just a flag
  ...
  await POST result                       ← T=X: POST resolves (guaranteed done)
  if (shouldRestart) _scheduleGracefulRestart()  ← T=X: setTimeout(2000) scheduled NOW
                                          ← T=X+2000: systemctl fires
```

The 2-second setTimeout is preserved as general post-POST drain headroom, but it no longer competes with the POST.

## Customer ground truth

`134.98.155.239` postgres on 2026-06-08:
- 1,381 deploy commands stuck at `status='acked'`
- 1,223 on a single node, all `target_version: "2.8.1"`
- Latest timestamps exactly 60s apart — matching the customer's heartbeat cadence (`HEARTBEAT_INTERVAL_MS = 60_000`)
- "Removing memclaw plugin solved it" because no plugin = no heartbeat = no restart trigger

## Wet test (on openclaw-test-ran, OpenClaw 2026.6.1)

The probe spun up a fake backend that returns 200 on the result POST after a configurable delay, and substituted `_scheduleGracefulRestart` with an event recorder.

| Scenario | POST delay | Order | Gap | Verdict |
|---|---:|---|---:|---|
| A. Normal POST | 30ms | post-resolved → restart-scheduled | 13ms | ✓ |
| **B. SLOW POST** | **1500ms** | **post-resolved → restart-scheduled** | **3ms** | ✓ |
| C. ping (no restart) | 30ms | post-resolved (only) | — | ✓ |

**Scenario B is the headline**: with the 1500ms POST delay, the restart was scheduled just 3ms AFTER the POST resolved. Pre-fix, the restart's setTimeout would have been scheduled at t=0, fired at t=2000, and the POST (still in flight at t=1500) would have continued momentarily before the SIGTERM killed it. In real customer traffic with typical backend latencies, the POST never completed → stuck-acked accumulation.

## Tests (4 new in `plugin/src/heartbeat.test.ts`)

- `processCommand POSTs result BEFORE scheduling systemctl` — pins the FIXED ordering using a 30ms POST delay to amplify the race. Pre-fix this test fails; post-fix it passes deterministically.
- `a non-restart command (ping) does NOT schedule a restart` — gate check on the `shouldRestart` flag
- `unknown command does NOT schedule a restart` — fails-closed safety
- `restart still fires when the result POST itself fails` — the deploy/restart command completed; bookkeeping channel failure must not strand the customer on the old binary

## Risks

- **The 2-second setTimeout is now redundant** since the POST is already done. Considered shortening it; opted to preserve current behavior — the small extra grace period covers any tail post-POST tidy-up (e.g., log flush, file-write sync) without touching production-timing semantics. Future cleanup PR can reduce/remove it.
- **`__DEPLOY_INTERNALS__.processCommand` is now an exported test seam.** Gated to NODE_ENV=test for the restart override; the processCommand export itself is unprotected (it's already an internal helper, the export just makes it directly invokable from tests).

## Plan to ship

| | |
|---|---|
| Plugin version | 2.8.4 → 2.8.5 |
| Backend dependency | none |
| Customer deploy | refresh plugin on each gateway VM (existing flow) |
| Expected impact | New deploy commands now correctly report `status="done"` after `systemctl restart` lands the new gateway; no more accumulation of acked-stuck commands |

## Test plan

- [x] Unit tests: 298/298 pass plugin-side (+4 new)
- [x] Manifest drift: 11/11 pass
- [x] tsc clean
- [x] Wet test under OpenClaw 2026.6.1: all 3 scenarios PASS, including the 1500ms-slow-POST race trigger
- [ ] Customer redeploy: expect zero new stuck-acked commands after this plugin lands on goodclaw/yoniclaw/antclaw nodes

### DIFF
--- a/plugin/openclaw.plugin.json
+++ b/plugin/openclaw.plugin.json
@@ -2,7 +2,7 @@
   "id": "memclaw",
   "name": "MemClaw",
   "kind": "memory",
-  "version": "2.8.4",
+  "version": "2.8.5",
   "description": "Central persistent memory for OpenClaw agents — write, search, and entity lookup tools.",
   "entry": "dist/index.js",
   "skills": [
--- a/plugin/package.json
+++ b/plugin/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@caura/memclaw",
-  "version": "2.8.4",
+  "version": "2.8.5",
   "description": "OpenClaw plugin for MemClaw central memory",
   "license": "Apache-2.0",
   "private": true,
--- a/plugin/src/heartbeat.test.ts
+++ b/plugin/src/heartbeat.test.ts
@@ -160,3 +160,244 @@ describe("deploy post-restart verification", () => {
     assert.deepEqual(__DEPLOY_INTERNALS__.readCooldown(), {});
   });
 });
+
+
+// ---- CAURA-000: result POST → restart ordering (race-condition fix) ----
+//
+// Pre-fix, ``processCommand`` scheduled ``setTimeout(systemctl restart, 2000)``
+// BEFORE awaiting the result POST. If the POST took longer than 2 s, the
+// systemctl SIGTERM killed it mid-flight and the backend never saw "done"
+// — the command stayed at status=acked forever. Customer prod data
+// (2026-06-08) showed 1,381 acked-stuck deploy commands, 1,223 on a single
+// node, which the backend's pending-only auto-upgrade gate then duplicated
+// at the next 60-s heartbeat → "SIGTERM every 60 seconds" loop.
+//
+// This test pins the FIXED order: the restart MUST only be scheduled
+// AFTER the POST resolves. The pre-fix code would record the restart
+// before the POST; the post-fix code records them in the opposite order.
+
+describe("processCommand — restart-after-POST ordering (CAURA-000)", () => {
+  // Required env so transport.ts / sendHeartbeat machinery imports cleanly.
+  process.env.MEMCLAW_API_KEY = "mc_test_key_for_processcommand_tests";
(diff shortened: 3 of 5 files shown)


ran-taig:
@claude

ran-taig:
@claude

ran-taig:
@erni-a

erni-a (approved):

PR #307, opened 2026-06-08
fix(plugin): don't create plugins.allow from nothing on autoFix (CAURA-000)

ran-taig:
## Summary

Customer-reported: installing memclaw 2.8.1 on a fresh OpenClaw gateway caused the `openai/gpt-5.5` agent model to become unknown ~2 minutes after install (`FailoverError: Unknown model: openai/gpt-5.5`). The gateway log told the story directly: `http server listening (1 plugin: memclaw; 0.8s)` — OpenClaw loaded ONLY memclaw, blocking every other plugin including the bundled `openai` provider.

**The customer's suspicion was correct: our plugin was to blame.**

## Root cause

`autoFixAllowlist` step 1 in `config.ts`:
```typescript
if (!isMemclawAllowed(config)) {
  if (!Array.isArray(config.plugins.allow)) config.plugins.allow = [];
  config.plugins.allow.push("memclaw");
}
```

On a fresh `openclaw.json` with no `plugins.allow` field, this **created** `plugins.allow = ["memclaw"]`. Verified against OpenClaw 2026.6.1's plugin gate (read from `/usr/lib/node_modules/openclaw/dist/*.js`):

```js
entry?.enabled === true &&
  (plugins.allow.length === 0 || plugins.allow.includes(pluginId))
```

**A non-empty `plugins.allow` is a STRICT allowlist.** Empty or missing = no restriction. By materialising `["memclaw"]` from nothing, we flipped a permissive config into a restrictive one and locked out every other plugin.

Confirmation from the OpenClaw error strings in the same dist files:
- `"plugins.allow does not include browser. Add \"browser\" to plugins.allow or remove the restrictive plugin allowlist."`
- `"Add ${formatPluginList(ownerPluginIds)} to plugins.allow or remove plugins.allow."`

## Fix

Two minimal changes in `plugin/src/config.ts`:

1. **`isMemclawAllowed`**: missing or empty `plugins.allow` now returns `true` (matches OpenClaw's runtime semantics — memclaw can load anyway). Only returns `false` when allow is a non-empty array that excludes memclaw.

2. **`autoFixAllowlist` step 1**: only appends memclaw to an EXISTING, non-empty allowlist. Never materialises one from nothing.

The post-fix invariant:
- User had no `plugins.allow` → autoFix leaves it missing (permissive — memclaw + openai + others all load)
- User had `plugins.allow = []` → autoFix leaves it empty (permissive)
- User had `plugins.allow = ["browser", "filesystem"]` → autoFix adds memclaw → `["browser", "filesystem", "memclaw"]`
- User had `plugins.allow = ["browser", "memclaw"]` → no-op

### Unit tests
- 310/310 pass plugin-side (was 301, +9 new):
  - 5 `isMemclawAllowed` tests pinning all 4 input shapes
  - 4 `autoFixAllowlist` tests driving the real function against temp `openclaw.json` files (verifies no-creation, empty-preserved, appended-to-restrictive, no-duplicate)
- 1 existing `isMemclawFullyConfigured` test updated: pre-fix it used `plugins.allow = []` to mean "not allowlisted", which assumed empty = restrictive; now uses an explicit non-empty allowlist that excludes memclaw (the only real "not allowed" case under corrected semantics).

### Wet test on `openclaw-test-ran` (GCE, OpenClaw 2026.6.1, plugin 2.8.6)

8/8 scenarios PASS:

```
✓ A. fresh install (no plugins.allow) — autoFix must NOT create restrictive allowlist
     plugins.allow after autoFix = undefined
✓ B. plugins.allow = [] (empty) — autoFix must leave it empty
✓ C. restrictive allow (without memclaw) — autoFix adds memclaw + preserves rest
     ["browser","filesystem","memclaw"]
✓ D. existing allow already with memclaw — no duplicate added
✓ E1. isMemclawAllowed: missing allow → true (permissive)
✓ E2. isMemclawAllowed: empty allow → true (permissive)
✓ E3. isMemclawAllowed: non-empty allow with memclaw → true
✓ E4. isMemclawAllowed: non-empty allow without memclaw → false
```

Scenario A is the **direct regression test for the customer bug**: pre-fix the same input produced `plugins.allow = ["memclaw"]` (the restrictive trap); post-fix it stays `undefined`.

## Customer recovery

Once 2.8.6 ships, the customer should:
1. Pull plugin 2.8.6 on the affected gateway (via the existing deploy machinery — note that PR #305 and #306 in this same batch fix the deploy-loop pathology, so the upgrade itself is no longer fragile)
2. Manually clear the bad `plugins.allow` in their `openclaw.json` (or remove the key entirely). Future installs/upgrades won't re-introduce the problem.

## Version

Plugin 2.8.5 → 2.8.6. No backend change.

## Test plan

- [x] 310/310 unit tests pass
- [x] Wet test on OpenClaw 2026.6.1 — 8/8 scenarios pass, including the direct customer-bug regression check
- [x] tsc clean (via `npm test` prefix)
- [x] DCO sign-off

### DIFF
--- a/plugin/openclaw.plugin.json
+++ b/plugin/openclaw.plugin.json
@@ -2,7 +2,7 @@
   "id": "memclaw",
   "name": "MemClaw",
   "kind": "memory",
-  "version": "2.8.5",
+  "version": "2.8.6",
   "description": "Central persistent memory for OpenClaw agents — write, search, and entity lookup tools.",
   "entry": "dist/index.js",
   "skills": [
--- a/plugin/package.json
+++ b/plugin/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@caura/memclaw",
-  "version": "2.8.5",
+  "version": "2.8.6",
   "description": "OpenClaw plugin for MemClaw central memory",
   "license": "Apache-2.0",
   "private": true,
--- a/plugin/src/config.test.ts
+++ b/plugin/src/config.test.ts
@@ -1,9 +1,14 @@
 import { test, describe } from "node:test";
 import assert from "node:assert/strict";
+import { writeFileSync, readFileSync, mkdtempSync, mkdirSync, rmSync } from "fs";
+import { join } from "path";
+import { tmpdir } from "os";
 import {
-  isMemorySlotClaimed,
+  autoFixAllowlist,
   isContextEngineSlotClaimed,
+  isMemclawAllowed,
   isMemclawFullyConfigured,
+  isMemorySlotClaimed,
   shouldRunAutoFix,
 } from "./config.js";
 import { getPluginDir } from "./paths.js";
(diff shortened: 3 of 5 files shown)


ran-taig:
@claude

ran-taig:
@erni-a

erni-a (approved):

PR #313, opened 2026-06-10
feat(skills): auto-approve clean Forge candidates (skip HITL inbox)

ran-taig:
Adds an **opt-in** flag that promotes a Forge candidate which passes **all 6 auto-gates AND carries a clean Sentinel scan** straight to `status='active'` — skipping the human Inbox approval step. **Default off**; HITL stays the default posture.

## What's new

| File | Change |
|---|---|
| `organization_settings.py` | New `skills_factory.sentinel.auto_promote_clean` (bool, default `False`) + `_LEAF_TYPES` validator |
| `skill_promoter.py` | `promote_pending_candidates` takes `auto_promote_clean`. Per passing candidate, destination = `active` if (flag on AND scan clean: `state='clean'`, `critical=0`) else `staged`. New `_scan_is_clean` helper; `PromotionAttempt.target_status`; `PromoterRunResult.auto_approved` |
| `cron_handler.py` | Reads the flag (`_resolve_auto_promote_clean`), threads it into the promoter, surfaces `auto_approved` in audit `stats` |
| `docs/operator-forge-cron.md` | New "Auto-approve clean candidates" section — trust trade-off, what it does/doesn't bypass, audit visibility, pause path |

## Safety invariants (the important part)

- **Never bypasses the 6 auto-gates.** The flag only changes the *destination* of an already-passing candidate — never whether it passes.
- **A dirty/quarantined scan never auto-activates.** Gate G5 holds it in `candidate`; and the promoter **re-asserts** scan cleanliness at the auto-approve site, so the "only clean auto-activates" invariant is self-contained and not load-bearing on G5 alone (defends against a future gate refactor).
- **Flag-off = byte-identical prior behavior.** Clean candidates land in `staged` and surface in the Inbox.
- **Clean unwind.** Flipping back off leaves already-active skills untouched; future clean candidates route to `staged`. A human can still Reject an active skill (poisons its fingerprint).

## Tests

`TestAutoPromoteClean` (4): flag-off→staged · flag-on+clean→active · warn-but-clean→active · injected-gate-pass + dirty-scan→staged (proves the local re-assertion catches a dirty scan even if a gate is relaxed). Cron tests assert the flag is read from settings + threaded + `auto_approved` surfaces in stats.

## Validation

```
400/400 SF tests pass     (lifecycle + cron + schema + distill + sentinel
                           + cluster + outcome-inference + session-trace + eval)
mypy + ruff clean on core-api/src/
```
[...]

PR #311, opened 2026-06-09
feat(skills): wire Forge into the lifecycle cron fanout

ran-taig:
Makes the Skill Factory pipeline **self-driving**. Phase 0 (PR #293) shipped the event topic + publisher + a no-op stub consumer; this PR replaces the stub with a real cron tick handler and adds the action to the lifecycle fanout so an external scheduler drives one tick per opted-in tenant.

## What's new

| File | Change |
|---|---|
| `routes/lifecycle.py` | `"forge-distill"` added to `_ACTION_PUBLISHERS` + new route branch in `_list_tenants_for_action` that filters to opted-in tenants only |
| `services/tenants.py` | New `list_tenants_with_skills_factory_enabled()` helper. **Non-opted-in tenants pay zero per-tick cost** — no message published, no audit row written |
| `services/lifecycle_audit.py` | `resolve_publisher_kwargs` stamps a deterministic `run_label` per `(tenant, UTC minute)`. Adapter's `forge_distill` delegates to the real cron handler (was Phase-0 no-op) |
| `services/forge/cron_handler.py` (NEW) | Production entry point. Per-tick flow: resolve `ForgeConfig` from per-tenant `org_settings.skills_factory.forge.*` → wire injectables (LLM/memory_fetcher/poison/writer/status_checker) → `run_forge_distill` → `promote_pending_candidates` (same-tick promotion) → return merged stats. **Requires real LLM provider chain** — the fake-LLM fallback is intentionally CLI-only. |
| `docs/operator-forge-cron.md` (NEW) | Cloud Scheduler YAML + k8s CronJob alternative, opt-in/opt-out flow, failure-mode recovery table |

## Same-tick promotion

`run_forge_cron_tick` runs `run_forge_distill` **then** `promote_pending_candidates` on the same DB session. A candidate that passes all 6 auto-gates (volume / diversity / freshness / poison / scan / hash-binding) lands in `staged` **within the same tick** — operators don't have to wait a second cron firing.

## Tests

`tests/test_forge_cron_tick.py` — **8 tests** covering:
- Forge-distill fanout routes to the opted-in helper (regression-pinned)
- Other actions still use the broad active-tenants helper (no accidental widening)
- `resolve_publisher_kwargs('forge-distill', org)` stamps the deterministic `run_label`
- `ForgeConfig` resolution honors per-tenant overrides + falls through to defaults
- `run_forge_cron_tick` invokes both phases (Forge + promoter) and returns merged stats incl. the 5 skip buckets
[...]

PR #293, opened 2026-06-07
feat(skills): Skill Factory · Phases 0, 1, and 2 (partial — backend only)

ran-taig:
**Restores PR #284** which was accidentally closed when the source branch was deleted during a Phase-2 push. The branch + commits are intact at `98b098b1`; the GitHub PR record could not be reopened because the closed-PR head SHA (`e085bcc…`) drifted from this branch's actual tip.

## Scope

Three commits, all DCO-signed, **flag-gated** behind `org_settings.skills_factory.enabled` (default `False` — zero behavior change for non-opted-in tenants):

- `dc2f96e` **Phase 0** — schema (v1) + flag-gated validator + 5 migrations (020 forge_rejected_fingerprints, 021 session_traces, 022 skills backfill) + Sentinel stub + org_settings.skills_factory.* (8 keys) + lifecycle publisher
- `da7dd48` **Phase 1** — Forge dry-run pipeline: 6 outcome-inference signals, session-trace builder, cluster fingerprint (P1-P6 stable), distill prompt+parse, eval harness, memclawctl CLI
- `98b098b` **docs** — Live Memory pitch deck (8 SVGs) + Skill Factory implementation plan + spec

## Validation

- 339/339 unit tests pass
- ruff check + format clean on `core-api/src/` + `core-storage-api/src/`
- DCO sign-off on all 3 commits
- Migration 022 was dry-run THREE times against eToro production (read-only `BEGIN…ROLLBACK`); v2 (json/jsonb cast round-trip) is the one persisted here

## Phase 2

Phase 2 backend (HITL Inbox + Sentinel real checks) lands as a separate PR ([#285] when opened) branched off this branch's tip — keeps reviews scoped.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #314, opened 2026-06-10
CAURA-133 fix(contradiction): mechanical same_subject + worked examples in entity-aware prompt

ran-taig:
## Summary

The dominant remaining priya-silence on dev. CAURA-131 wired the entity-aware judge correctly; CAURA-132 confirmed it's invoked on the failing pairs; the wet-test on dev v2.12.1 (post Pub/Sub topology mitigation) shows the LLM **still returns `verdict=False`** on 15 of 16 shared-canonical-subject candidate-pairs because the prompt's same-subject path was prose-weak and let surface qualifiers ("from AcmeCorp") fool the model.

This PR strengthens the same-subject section of `ENTITY_AWARE_CONTRADICTION_PROMPT` with:

1. **Mechanical entity_id comparison** as the decision rule (instead of a prose instruction the LLM could interpret away).
2. **Two worked examples** anchoring both branches: same entity_id different employer → `same_subject=true`; different entity_id same name → `same_subject=false`.
3. **Explicit enumeration of surface qualifier categories** that must NOT influence the decision when entity_id matches.

## Why this works

The L3.4 preflight (CAURA-130) already drops candidates whose `entity_id`s differ before the entity-aware judge runs (the "priya" / "priya" collision case). By the time the judge sees a pair, entity_ids are trustworthy. So a mechanical `entity_id == entity_id` rule is safe to take as ground truth — no need to leave room for LLM judgment about the resolution itself.

## Evidence we're fixing the real bug

Dev v2.12.1 CAURA-132 forensic logs for the wet-test failing pairs:

```
PATH_C_DETECTION judge_selection candidates=8 entity_aware=8 base=0   ← wiring works
PATH_C_DETECTION context_fetched new_ctx_size=3 cand_ctx_sizes={…3,3,3,3,3,3,3,3}
PATH_C_DETECTION verdict … judge=entity_aware verdict=False conf=0.90  (× 15)
PATH_C_DETECTION verdict … judge=entity_aware verdict=True  conf=0.90  (× 1)
```

The entity-aware judge IS being invoked. It IS receiving the resolved entity context. It says **False** anyway. That's a prompt bug, not a wiring bug.

## Preserved (no regressions)

| Asset | Status |
|---|---|
| 4 format placeholders | unchanged |
| 6 JSON output keys consumed by `_judge_contradiction` | unchanged |
| Phrases `"RESOLVED ENTITIES for Statement A/B"`, `"authoritative"` (test-asserted) | unchanged |
[...]

PR #315, opened 2026-06-10
feat(skills): MCP-direct delivery — agents discover only active skills

ran-taig:
Gates the agent-facing `memclaw_doc` surface so a skill is **discoverable only when `status='active'`**. This is the mechanism that turns *"an operator approved it"* (or `auto_promote_clean` fired) into *"agents can find it"* — `candidate`/`staged`/`quarantined` skills are in-flight or blocked and must not surface to agents.

## The model (MCP-direct delivery)

For MCP clients, delivery is a **read, not a push** — the skill lives in the `skills` collection; the agent pulls it via `memclaw_doc op=search/read/query`. No install, no plugin runtime. This PR makes the lifecycle (`candidate → staged → active`) actually control what agents see.

## What's new

| File | Change |
|---|---|
| `document_repository.py` | `search()` gains optional `status` param (pure mechanism: `data->>'status' = :status`). REST path passes nothing → unaffected |
| `mcp_server.py` | `memclaw_doc` active-only policy for `collection='skills'` when opted in — see table below. New non-raising `_skills_factory_enabled` helper |

| op | behavior on skills (flag on) |
|---|---|
| `read` | non-active → `Not found` (no existence leak) |
| `query` | forces `where['status']='active'` (overrides caller) |
| `search` scoped | pushes `status='active'` into SQL (exact top_k) |
| `search` broad | post-filters skill rows so a candidate never leaks cross-collection |

## Invariants

- **Opt-in gated.** Applies ONLY when `skills_factory.enabled=true`. Non-opted-in tenants are **byte-identical** to pre-Skill-Factory — preserves the merge-day no-op guarantee.
- **No admin bypass on MCP.** Operators inspect non-active skills via the Inbox API, not the agent surface. Exactly one way for a non-active skill to reach an agent: it can't.
- **Skills-only.** Other collections never touched.

## Tests

`tests/test_mcp_doc.py` +12 covering all 4 ops × flag on/off + non-skills-untouched + broad-search leak prevention.

```
188 tests pass (mcp_doc + api_documents_skills + skill_schema + lifecycle + direct_mcp_skill)
mypy + ruff clean on core-api/src/
```

## Docs

`docs/mcp-skill-delivery.md` — the active-only contract, the opt-in gate, why operators use the Inbox, and the **pull-vs-install ceiling**: MCP-direct is the universal *probabilistic* floor (agent must choose to look); reliable delivery needs a per-harness install adapter (future). They're complementary.

## Context

Builds on the merged Skill Factory stack (#293 schema/inbox, #311 cron, #313 auto-approve). This is the read side — it makes `active` skills reachable by agents over any MCP-capable harness with zero integration.

### DIFF
--- a/core-api/src/core_api/mcp_server.py
+++ b/core-api/src/core_api/mcp_server.py
@@ -13,7 +13,7 @@ import json
 import logging
 import re
 import time
-from typing import Annotated, cast
+from typing import Annotated, Any, cast
 from uuid import UUID, uuid4
 
 import httpx
@@ -1165,6 +1165,107 @@ async def memclaw_tune(
 SKILLS_COLLECTION = "skills"
 _SKILL_SLUG_RE = re.compile(r"^[a-z0-9][a-z0-9._-]{0,99}$")
 
+# The agent-facing MCP surface only exposes skills in this status. The
+# Skill Factory lifecycle (candidate → staged → active) gates what an
+# agent can discover: candidate/staged/quarantined/rejected skills are
+# in-flight or blocked and must NOT surface to agents. Only this filter
+# turns "an operator approved it" into "agents can find it".
+_AGENT_VISIBLE_SKILL_STATUS = "active"
+
+
+async def _skills_factory_flag(db: Any, tenant_id: str) -> bool:
+    """Strict opt-in check — RAISES on a settings-lookup failure.
+
+    Is ``skills_factory.enabled`` true for this tenant? Callers that gate
+    a SECURITY decision (op=write / op=delete) wrap this and **fail
+    closed** (abort the mutation) on error — they must never proceed
+    unvalidated just because the flag couldn't be read. Read-path callers
+    use the lenient ``_skills_factory_enabled`` wrapper instead.
+    ``get_raw_settings`` is cache-first (5-min TTL), so this is a cheap
+    hot-path check.
+    """
+    from core_api.services.organization_settings import get_raw_settings
+
+    raw = await get_raw_settings(db, tenant_id)
+    return (
+        isinstance(raw, dict)
+        and isinstance(raw.get("skills_factory"), dict)
+        and bool(raw["skills_factory"].get("enabled"))
+    )
+
+
+async def _skills_factory_enabled(db: Any, tenant_id: str) -> bool:
+    """Non-raising opt-in check for the active-only skill-read filter
+    (read / query / search) — fails CLOSED on a settings-lookup failure.
+
+    On success returns the tenant's true ``skills_factory.enabled`` flag,
+    so non-opted-in tenants keep byte-identical (unfiltered) read
+    behavior — the merge-day no-op invariant.
+
+    On a settings-lookup FAILURE it returns ``True`` (assume enabled), so
+    the active-only filter IS applied: a transient settings outage must
+    not let a non-active skill (candidate / staged / quarantined) leak to
+    an agent. The read still succeeds — it's filtered, not 500'd — so the
+    cost is only that, during an outage, a tenant temporarily sees just
+    their ``active`` skills. For non-opted-in tenants that is a near
+    no-op (their skills are all ``active`` post-migration). This keeps
+    ALL skill gates fail-closed: write/delete abort, reads filter.
+    """
+    try:
+        return await _skills_factory_flag(db, tenant_id)
+    except Exception:
+        logger.warning(
+            "skills_factory flag lookup failed for %s; failing CLOSED — active-only "
+            "filter APPLIED (only status='active' skills surface) until cache recovers",
+            tenant_id,
+        )
+        return True
+
+
+def _safe_int(val: Any, default: int) -> int:
+    """Coerce a settings value to int, falling back to ``default`` on a
+    null, boolean, or non-numeric value. The per-tenant byte caps are
+    operator-editable JSON, so a misconfiguration (``null``, ``"auto"``,
+    ``true``/``false``) must degrade to the documented default — not
+    crash (or silently mis-cap) the skills write path.
+
+    ``bool`` is checked first because it is a subclass of ``int`` in
+    Python: ``int(True) == 1`` / ``int(False) == 0`` would otherwise pass
+    through silently and produce a 1- or 0-byte cap rather than the
+    intended default."""
+    if isinstance(val, bool):
+        return default
+    try:
+        return int(val)
+    except (TypeError, ValueError):
+        return default
+
+
+def _skill_hidden_from_agent(doc: Any, *, caller_tenant_id: str, caller_opted_in: bool) -> bool:
+    """Whether a fetched row must NOT surface on the agent MCP skill
+    surface.
+
+    A skills-collection row is hidden when it is non-active AND either:
+
+      • the caller's tenant opted in (the active-only gate applies to
+        the caller's own skills), or
+      • the row belongs to a DIFFERENT tenant than the caller — a
+        sibling tenant's in-flight / blocked skill must never leak
+        through cross-tenant credentials, regardless of the *caller's*
+        own opt-in flag. This is safe: a non-opted-in owning tenant's
+        skills are all ``active`` (backfilled by migration 022), so this
+        only ever hides a genuinely non-active row, never a legitimately
+        visible one.
+
+    Non-skills rows are never hidden. Keyed off the row's own
+    ``tenant_id`` so the decision follows the OWNING tenant, not just
+    the caller's flag (closes the cross-tenant read leak)."""
+    if getattr(doc, "collection", None) != SKILLS_COLLECTION:
+        return False
+    if (getattr(doc, "data", None) or {}).get("status") == _AGENT_VISIBLE_SKILL_STATUS:
+        return False
+    return caller_opted_in or getattr(doc, "tenant_id", caller_tenant_id) != caller_tenant_id
+
 
 async def memclaw_doc(
     op: Annotated[str, Field(description="write|read|query|delete|list_collections|search.")],
@@ -1250,6 +1351,36 @@ async def memclaw_doc(
                     fleet_id=fleet_id,
                     readable_tenant_ids=readable if op in READ_OPS else None,
                 )
+                # Active-only count correction for skills: an opted-in
+                # tenant's listing must not advertise non-active skills
+                # in the count — they're invisible to read/query/search,
+                # so a count that includes them is misleading. Recompute
+                # the skills row restricted to status='active', over the
+                # same (possibly cross-tenant) scope the listing used.
+                # One COUNT, only when a skills row is present AND the
+                # tenant opted in.
+                if any(name == SKILLS_COLLECTION for name, _ in rows) and await _skills_factory_enabled(
+                    db, tenant_id
+                ):
+                    from sqlalchemy import func as sa_func
+                    from sqlalchemy import select as sa_select
+
+                    from common.models.document import Document
+
+                    tenant_pred = (
+                        Document.tenant_id.in_(readable) if readable else Document.tenant_id == tenant_id
+                    )
+                    cnt_stmt = sa_select(sa_func.count()).where(
+                        tenant_pred,
+                        Document.collection == SKILLS_COLLECTION,
+                        Document.data["status"].astext == _AGENT_VISIBLE_SKILL_STATUS,
+                    )
+                    if fleet_id:
+                        cnt_stmt = cnt_stmt.where(Document.fleet_id == fleet_id)
+                    active_count = int((await db.execute(cnt_stmt)).scalar_one())
+                    rows = [
+                        (name, active_count if name == SKILLS_COLLECTION else count) for name, count in rows
+                    ]
                 return _with_latency(
                     json.dumps(
                         {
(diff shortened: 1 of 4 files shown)


ran-taig:
@claude please review — security-critical read-path gate. This filters the agent-facing `memclaw_doc` surface so only `status='active'` skills are discoverable (when `skills_factory.enabled`).

Focus areas:
- **Leak prevention**: can a candidate/staged/quarantined skill reach an agent through ANY op (read/query/search-scoped/search-broad)? The broad cross-collection search path is the subtle one.
- **Opt-in invariant**: flag-off tenants must be byte-identical to today (no filter).
- **No bypass**: the MCP path has no admin override by design — confirm there's no accidental way around the active-only gate.

188 tests pass; mypy + ruff clean. Looking for 🟠 high / 🔴 critical only.

ran-taig:
@claude re-review round 2 — all three findings fixed:

1. 🟠 HIGH — **op=write self-promotion bypass closed.** The MCP write path applies no lifecycle validator, so I added a guard: for `collection='skills'` + flag-on, any caller-supplied `status` is rejected (`INVALID_ARGUMENTS`) before the upsert. Status transitions go through the Inbox API only. Test asserts the DB write never happens. (Used the `_error_response` envelope for consistency with the other op=write rejections rather than a plain string.)

2. 🟡 MED — **`_skills_factory_enabled` truly non-raising.** Wrapped `get_raw_settings` in try/except → returns False (legacy/unfiltered) on failure, with an honest comment about the direction (settings outage → opted-in tenant briefly reverts to unfiltered rather than 500-ing; cache-first so rare).

3. 🟡 MED — **op=delete active-only gate.** Delete is atomic (DELETE…RETURNING), so I pre-fetch for skills+flag-on and return the same `Not found` as op=read when the skill isn't active — no existence leak, and agents can't delete in-flight skills (operator-managed via Inbox).

5 new tests (write-rejects-status / write-without-status-ok / write-status-ok-when-flag-off / delete-hides-non-active / delete-allows-active). 151 tests pass; mypy + ruff clean. PR at SHA `c962228a`.

ran-taig:
## Update — replaced the op=write band-aid with proper lifecycle wiring

The previous round's `op=write` guard blanket-rejected any caller-supplied `status` on skills writes. That was the wrong fix: per the original plan (§4 / SF-002), an agent-direct skill write is a **supported** path — it should flow through the validator → `staged` → Inbox → review → `active`, not bounce or land in limbo.

So the MCP `op=write` path now runs the **same `validate_and_normalize_skill_write`** the REST route does (mirrors `routes/documents.py` §SF-002), gated on the opt-in flag:

- **status defaults to `staged`** → the doc lands in the HITL Inbox, never agent-visible until approved to `active`. No dead end.
- **status RBAC closes self-promotion properly.** MCP callers carry no admin/forge identity (there is no `is_admin`/`org_role` accessor on this surface), so `is_admin=is_internal_forge=False`. A caller-supplied `status='active'`/`'candidate'`/system status → **403 FORBIDDEN** via the validator's `ADMIN_ONLY_STATUSES`/`INTERNAL_ONLY_STATUSES`. An admin who legitimately authors an `active` skill uses the REST/dashboard path (real auth context).
- **source RBAC** — `source='forge'` (internal-only) / `source='manual'` (admin-only) → 403; agents write `source='agent'`.
- **Sentinel pre-scan + content_hash + byte/slug caps** all run; `content_hash` is server-stamped, never trusted from the body.
- The validator raises `HTTPException`; the handler's existing `except HTTPException` maps 403/422/409/404 to the right error code — no new error plumbing.

Also retained from round 1: `_skills_factory_enabled` is wrapped in try/except (→ False on lookup failure), and `op=delete` has an active-only pre-fetch guard.

### Tests
Revised the 3 band-aid write tests to assert validator behavior, and added a source-RBAC test:
- `test_skill_write_rejects_caller_active_status_when_flag_on` → 403 FORBIDDEN, no DB write
- `test_skill_write_rejects_forge_source_when_flag_on` → 403, no DB write
- `test_skill_write_defaults_to_staged_when_flag_on` → asserts upserted `data.status=='staged'`, `source=='agent'`, server-stamped `content_hash`
- `test_skill_write_status_allowed_when_flag_off` → validator skipped, status passes through unchanged

`tests/test_mcp_doc.py`: 54 pass. mypy + ruff clean on `core-api/src/`.

Docs updated (`docs/mcp-skill-delivery.md`) with the write-goes-through-the-lifecycle contract and `read`/`write`/`delete` rows in the op table.

@claude please re-review.

ran-taig:
## Round 3 — both fixes applied

**🟡 Medium — unguarded `int()` on misconfigured byte caps**
Added a `_safe_int(val, default)` helper next to `_skills_factory_enabled` and swapped both `SkillWriteContext` cap conversions to it. A `null` or non-numeric (`"auto"`) per-tenant cap now degrades to the documented default (160 / 40 000) instead of bubbling a `TypeError`/`ValueError` into an opaque `INTERNAL_ERROR`. New regression test `test_skill_write_tolerates_misconfigured_byte_caps`.

**🟢 Low — delete TOCTOU**
Removed the `get_by_doc_id` pre-fetch and folded the status guard directly into the DELETE's WHERE clause (`Document.data['status'].astext == 'active'`, added only when the skills-factory flag is on). Check-and-delete is now a single atomic statement — no window for a concurrent promote/demote. Zero rows deleted under the skills gate returns the same `"Not found"` as `op=read`, so a non-active skill stays indistinguishable from a missing one. Updated the two delete tests to model the atomic zero-row path.

`tests/test_mcp_doc.py`: 55 pass. mypy + ruff clean on `core-api/src/` (the lone mypy note is the pre-existing `croniter` stub in `config.py`, unrelated).

@claude please re-review.

ran-taig:
## Round 4 — all five addressed

**🟠 High — case-bypass write guard** — Note this references the round-1 band-aid (`"status" in data`), which round 2 replaced with the SF-002 validator. In the validator version a `{"STATUS":"active"}` key **cannot** self-promote: the validator owns the canonical lowercase `status` (defaulting agent writes to `staged`), and every gate reads `data->>'status'` (case-sensitive in JSONB), so a variant key is inert for visibility. It would still persist as a confusing shadow field, so I added a focused guard that **rejects any case-variant `status` key** (`STATUS`/`Status`/…) with `INVALID_ARGUMENTS` before the validator runs. Test: `test_skill_write_rejects_case_variant_status_key_when_flag_on`.

**🟡 Medium — TOCTOU delete** — Already fixed in round 3: the pre-fetch is gone and the status guard lives in the DELETE's WHERE (`data->>'status' = 'active'`, added only when the flag is on), so check-and-delete is one atomic statement. The cited lines (1595-1618) are the pre-round-3 version.

**🟡 Medium — cross-tenant broad-search leak** — Real, and it applied to `read`/`query`/`search`, not just broad search. Fixed with a shared `_skill_hidden_from_agent(doc, caller_tenant_id, caller_opted_in)` helper: a skill row is hidden when non-active AND (caller opted in **OR** the row's `tenant_id != caller's`). The cross-tenant arm fires regardless of the caller's own flag, so a non-opted-in caller can never pull a sibling's in-flight skill via cross-tenant creds. Applied to `read` (single doc), `query` (post-filter when caller off; SQL force when on), and `search` (per-row post-filter over scoped + broad). Safe because a non-opted-in owner's skills are all `active` (migration 022). Tests: `test_skill_{read,query,search}_*cross_tenant*`.

**🟢 Low — list_collections count leak** — When the flag is on and a `skills` row is present, the count is recomputed active-only via one `COUNT(... WHERE data->>'status'='active')` over the same (cross-tenant-aware) scope. Tests: `test_list_collections_skill_count_{active_only_when_flag_on,unchanged_when_flag_off}`.

**🟢 Low — silent query override** — An explicit `where={"status": <non-active>}` on skills now returns `INVALID_ARGUMENTS` pointing to the Inbox API, rather than being silently rewritten to `active`. A status-less query is still transparently scoped to `active`. Tests: `test_skill_query_rejects_explicit_non_active_status_when_flag_on` + `test_skill_query_scopes_to_active_when_no_status_when_flag_on`.

`tests/test_mcp_doc.py`: 62 pass; 235 pass across the skill/mcp_doc/documents sweep. mypy + ruff clean on `core-api/src/` (lone mypy note is the pre-existing `croniter` stub in `config.py`).

@claude please re-review.

ran-taig:
## Round 5 — both addressed

**🟡 Medium — delete response format** — Fixed, and made strictly more correct than a string→JSON swap. Since the status guard now lives in the DELETE's WHERE, a non-active skill and a missing doc BOTH delete zero rows — so I dropped the special skills "Not found" branch entirely and let both fall through to the **same** generic `json.dumps({"error": "Document '<id>' not found in collection '<coll>'"})`. That's valid JSON (no bare-string `_with_latency` issue) **and** byte-for-byte indistinguishable from a missing doc — a returned message like `Not found: skills/...` would actually have been distinguishable from the generic missing-doc message and leaked the non-active/missing distinction. Test updated to parse the envelope and assert the identical generic error.

**🟢 Low — double settings read on write path** — Collapsed. The op=write skills block now makes a single `get_settings_for_display` call and derives both the opt-in flag (`skills_factory.enabled`) and the byte caps from it — no separate `_skills_factory_enabled`/`get_raw_settings` round-trip on a cold cache. read/query/search/delete keep the cheap `_skills_factory_enabled` helper (they don't need the caps). Test helper updated so the write tests drive `enabled` through that one call.

`tests/test_mcp_doc.py`: 62 pass; 235 across the skill/mcp_doc/documents sweep. mypy + ruff clean on `core-api/src/` (lone mypy note is the pre-existing `croniter` stub in `config.py`).

@claude please re-review.

ran-taig:
## Round 6 — both addressed

**🟡 Medium — fail-open on write/delete security gates** — Fixed, with one correction to the suggested patch: the snippet wraps `_skills_factory_enabled` in try/except, but that helper *already swallows exceptions → False internally*, so the `except` would be dead code and the gate would still fail **open**. The real fix is a strict, raising variant:

- Split the helper into **`_skills_factory_flag`** (strict, raises) + **`_skills_factory_enabled`** (lenient wrapper, fail-open → False). read/query/search keep the lenient one (fail-open is fine for reads — a settings blip shouldn't 500 a read).
- **op=write** no longer calls the flag helper at all (Round 5 moved it to `get_settings_for_display`); wrapped *that* call in try/except → on failure returns `INTERNAL_ERROR "skill lifecycle gate unavailable"` and **never reaches the upsert** (fail-closed).
- **op=delete** now uses `_skills_factory_flag` (strict) wrapped in try/except → fail-closed before the DELETE runs. (The `and` short-circuits for non-skills collections, so they never touch settings.)
- Tests: `test_skill_write_fails_closed_on_settings_error` (asserts no upsert) and `test_skill_delete_fails_closed_on_settings_error` (asserts `db.execute` never called).

**🟢 Low — duplicate comment block** — Removed the redundant back-to-back "Active-only gate for a SCOPED skills search" comment; kept the single accurate one.

`tests/test_mcp_doc.py`: 64 pass; 237 across the skill/mcp_doc/documents sweep. ruff clean, 0 mypy errors in `mcp_server.py`.

@claude please re-review.

ran-taig:
## Round 7 — addressed (+ rebased onto latest main)

First, FYI: the branch had been rebased onto the new `main` (#314/#316) and force-pushed; I replayed this round's delta on top of that rebased base, so the PR is now current with `main`.

**🟡 Medium — `_skills_factory_enabled` exception path (fail-open on reads)** — Switched it to **fail CLOSED**: on a settings-lookup failure the helper now returns `True` (assume enabled) so the active-only filter IS applied — a transient settings outage can no longer let a non-active skill leak to an agent. The read still succeeds (it's filtered, not 500'd); the only effect during an outage is that a tenant temporarily sees just their `active` skills, which is a near no-op for non-opted-in tenants (their skills are all `active` post-migration 022). This makes **all** skill gates fail-closed: write/delete abort, reads filter. The log line is now explicit: `failing CLOSED — active-only filter APPLIED … until cache recovers`. (Refactor from Round 6 means the strict raising check lives in `_skills_factory_flag`; this lenient wrapper is the read-path one.) Tests: `test_skills_factory_enabled_fails_closed_on_settings_error` + `test_skill_read_hides_non_active_when_flag_lookup_errors`.

**🟢 Low — duplicate comment in op=search** — Already removed in Round 6 (that review was on the pre-Round-6 push); verified there's now exactly one.

**CI — `ruff format --check`** — Ran the formatter; `core-api/src/` (182 files) is now format-clean. (I'd been running `ruff check` but not `ruff format` — fixed.)

`tests/test_mcp_doc.py`: 66 pass. ruff check + format clean, 0 mypy errors in `mcp_server.py`.

@claude please re-review.

ran-taig:
## Round 8 — addressed (option 2: separate the two concerns)

Fixed the op=query lenient-flag issue. The query gate now splits cleanly:

**Explicit non-active status (`where={"status": "staged"}`)** — the security-sensitive REJECTION path. Now gated on the **strict `_skills_factory_flag`**:
- opted in → `INVALID_ARGUMENTS` with the Inbox-API pointer (correct);
- genuinely **not** opted in (settings readable, flag false) → the caller's `where` passes through untouched (legacy behavior, **no** confusing Inbox message);
- settings **outage** (strict flag raises) → `INTERNAL_ERROR "skill lifecycle gate unavailable"`, same fail-closed pattern as op=delete — never the nonsensical Inbox pointer to a tenant we can't confirm opted in.

**No status / `status='active'`** — the common, safe path. Keeps the **lenient `_skills_factory_enabled`** for the transparent `active` injection: on an outage it fails closed by *filtering* (returns True → injects `status='active'`), consistent with op=read, and never errors a plain query.

Tests:
- `test_skill_query_explicit_status_fails_closed_on_settings_error` (strict flag raises → INTERNAL_ERROR, query not called);
- `test_skill_query_no_filter_when_flag_off` now also asserts the genuinely-not-opted-in pass-through (no Inbox rejection);
- `_patch_flag` patches both helpers so opt-in semantics stay consistent across the lenient and strict checks.

`tests/test_mcp_doc.py`: 67 pass; 240 across the skill/mcp_doc/documents sweep. ruff check + format clean (`core-api/src/`), 0 mypy errors in `mcp_server.py`.

@claude please re-review.

ran-taig:
## Round 9 — addressed

Fixed `_safe_int` to treat a JSON boolean as a misconfiguration. Added an `isinstance(val, bool)` guard *before* the `int()` attempt, since `bool` subclasses `int` (`int(True)==1`, `int(False)==0`) — without the guard a `true`/`false` cap would silently produce a 1- or 0-byte cap instead of degrading to the documented default. Docstring updated to list `true`/`false` alongside `null` / `"auto"`.

Test: `test_safe_int_degrades_bool_and_garbage_to_default` asserts `True`/`False`/`None`/`"auto"`/`[]` → default, while valid ints and numeric strings (`40000`, `"4096"`) still pass through.

`tests/test_mcp_doc.py`: 68 pass. ruff check + format clean (`core-api/src/`), 0 mypy errors in `mcp_server.py`.

@claude please re-review.

ran-taig:
@claude

ran-taig:
## Round 10 — addressed

Wrapped the `kind='update'` live-doc fetch (`get_storage_client().get_document(...)`) in its own try/except, mirroring the settings-fetch and op=delete flag guards. A transient DB/network error now logs and returns the curated `INTERNAL_ERROR "skill lifecycle gate unavailable"` and aborts before the upsert, instead of falling through to the outer handler that leaked the raw exception string. Fails closed — consistent with every other external call that gates a security decision in this path.

Test: `test_skill_update_write_fails_closed_on_live_doc_fetch_error` (storage `get_document` raises on a `kind='update'` write → INTERNAL_ERROR, upsert not called).

`tests/test_mcp_doc.py`: 69 pass. ruff check + format clean (`core-api/src/`), 0 mypy errors in `mcp_server.py`.

@claude please re-review.

ran-taig:
@claude

ran-taig:
@Eldad-Caura

Eldad-Caura (approved):

PR #317, opened 2026-06-11
feat(skills): gate the OpenClaw install path to active-only skills

ran-taig:
## Why

PR #315 closed the active-only leak on the MCP **pull** surface (`memclaw_doc`), but the **push** surface stayed open. The OpenClaw plugin reconciler (`plugin/src/reconcile-skills.ts`) mirrors the `skills` catalog to each node's disk via `POST /documents/query` with `where={}` — **all statuses**. For a Skill-Factory-opted-in tenant that pushes `candidate` / `staged` / `quarantined` skills onto every node's disk and into the agent's palette — exactly the leak #315 closed for pull, still open on push.

This wires push and pull behind **one server-side gate**.

## What's new

**`POST /api/v1/skills/installable`** (`core-api/.../routes/documents.py`) — the agent-harness install surface. Same active-only + opt-in gate as the MCP path, enforced server-side:
- opted-in tenant → only `status='active'` skills;
- non-opted-in → every visible skill (`where={}`), byte-identical to the legacy reconcile (a legacy skill predating Skill Factory may lack a `status` field; forcing `active` would wrongly drop it — so we don't filter at all);
- settings-lookup failure → **503, fail closed**, storage never queried.

The request model has **no `where`** — a caller can't widen the filter to pull non-active skills.

**Reconciler switch-over** (`plugin/src/reconcile-skills.ts`) — pulls from `/skills/installable` instead of `/documents/query{where:{}}`. The client carries zero policy: no status filter, no opt-in flag.

## Consequences

- **De-activation reactivity (free):** a skill flipping `active → rejected/quarantined` drops out of `installable`; the reconciler's existing orphan-removal path deletes it from disk next tick. Push stays in sync with the gate.
- **Fail-safe end-to-end:** server fails closed (503); reconciler fails safe on a non-2xx (preserves on-disk skills, writes nothing) — an outage can never push a non-active skill to disk.
- **No blast radius:** `/documents/query` untouched (dashboards / inbox that legitimately need non-active skills unaffected). Old plugins keep working (ungated, as before — `/documents/query` still exists); new plugins are gated. The plugin sends no version-coupled change beyond the endpoint URL.

## Scope

This is the **first** of the OpenClaw-install PRs (the "close the gate gap" slice). Deliberately out of scope, for follow-ups: install/bootstrap hardening (`/install-plugin`, gateway restart, version pinning) and operator-facing per-node reconcile observability.

## Tests

- `tests/test_skills_installable_endpoint.py` — active-only when opted-in; `where={}` when not; 503 + storage-not-queried on settings error; caller `where` ignored.
- `plugin/src/reconcile-skills.test.ts` — endpoint switch-over; de-activation removes from disk next tick; server 503 → reconciler fails safe.
- 246 server tests across the documents/installable/mcp_doc/skill sweep; 312 plugin tests. ruff + format clean; 0 mypy errors in the touched route.

## Docs

`docs/mcp-skill-delivery.md` — documents the push path and the shared gate; OpenClaw promoted from "future" to the first per-harness install.

### DIFF
--- a/core-api/src/core_api/routes/documents.py
+++ b/core-api/src/core_api/routes/documents.py
@@ -107,6 +107,19 @@ class DocQueryRequest(BaseModel):
     offset: int = Field(default=0, ge=0)
 
 
+class InstallableSkillsRequest(BaseModel):
+    """Request for the agent-harness install surface (`/skills/installable`).
+
+    Deliberately narrower than ``DocQueryRequest``: the collection is
+    fixed to ``skills`` and the ``where`` filter is server-decided (the
+    caller cannot widen it), so a harness can't ask for non-active skills.
+    """
+
+    tenant_id: str
+    fleet_id: str | None = None
+    limit: int = Field(default=1000, ge=1, le=1000)
+
+
 class DocSearchRequest(BaseModel):
     """Vector search over indexed documents.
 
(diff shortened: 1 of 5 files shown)


ran-taig:
@claude please review. This gates the OpenClaw push-to-disk install path to active-only skills via a new server-side `POST /api/v1/skills/installable` surface, mirroring #315's MCP pull gate. Key things to scrutinize: (1) the opt-in / non-opt-in branch correctly preserves the merge-day no-op invariant; (2) fail-closed (503) on settings error never reaches storage; (3) the reconciler fails *safe* on the server's fail-closed 503; (4) no regression to `/documents/query` or other skills routes.

ran-taig:
@Eldad-Caura

Eldad-Caura (approved):

PR #319, opened 2026-06-11
fix(mcp): enforce gateway perimeter on the /mcp header-trust path

erni-a:
## Summary
The MCP ASGI middleware trusted the gateway identity headers (`X-Tenant-ID`, `X-Agent-ID`, `X-Readable-Tenant-IDs`, `X-Capabilities`) verbatim, while the REST auth stack refuses that header-trust path unless a matching `X-Gateway-Secret` is presented. A caller reaching core-api's `/mcp` directly could set `X-Tenant-ID` to any tenant and fully impersonate it.

## Changes
- Validate `X-Gateway-Secret` (`hmac.compare_digest`) before honoring `X-Tenant-ID`; reject with 401 on mismatch. No-op when the secret is unset (standalone / dev).
- Honor the gateway identity headers only on the gateway-verified path, so direct callers can't self-assert an agent identity, cross-tenant read set, or capability set.
- Prepend the home tenant to the readable set (parity with REST) so recall/list/stats can't exclude the caller's own rows.
- `memclaw_entity_get`: add the canonical error tail so storage errors surface as `isError` envelopes instead of escaping unwrapped.

## Tests
`tests/test_mcp_gateway_secret.py` (7 cases) — covers secret-missing/wrong/correct, secret-unset back-compat, home-tenant prepend, and gateway-only header trust. Verified live: spoofed tenant → 401, correct secret → 200, REST parity intact.


erni-a:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #320, opened 2026-06-11
fix(api): close cross-tenant and caller-asserted-identity authz gaps

erni-a:
## Summary
Four route-level authorization gaps.

- **fleet `command_result`** — the storage UPDATE keyed only on `command_id`, so any authenticated tenant could mark another tenant's fleet command `done`/`failed` (and inject a `result`) by UUID. The UPDATE is now tenant-scoped (admin stays unscoped) and the route 404s when no row matches. The storage route also now actually persists `done`/`failed` status + result + `completed_at`, which the caller has always sent but the old handler silently dropped for every status except `acked`.
- **memories `redistribute`** — the `trust_level >= 3` gate ran against the caller-controlled `agent_id` query param instead of the authenticated identity, so a trust-1 agent credential could clear it by naming a trust-3 agent. The authenticated agent identity now takes precedence (same pattern as `delete`/`update_memory`).
- **STM `clear_notes` / `clear_bulletin` / `promote`** — added `enforce_read_only` (all three), `enforce_usage_limits` (promote), and agent-identity binding so an agent credential can't wipe a peer's notes or promote into LTM on the peer's behalf.
- **`delete_memory` audit row** — now attributes the effective identity instead of the raw query param, which logged `None` for gateway agents.

## Tests
`tests/test_route_authz_gaps.py` (11 cases) covering cross-tenant 404, same-tenant persistence, the redistribute precedence matrix, STM read-only/peer-agent rejection, and delete audit attribution.


erni-a:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #322, opened 2026-06-11
fix: cross-tenant search widening, non-default LLM structured-output kwargs, event deploy safety

erni-a:
## Summary
Four findings.

- **search** — `ExecuteScoredSearch` forwarded `readable_tenant_ids` only when `len(readable) > 1`, while the entity-lookup short-circuit uses `readable != [tenant_id]`. A cross-tenant credential whose readable set is exactly one foreign tenant was silently narrowed to home-tenant reads on the scored-search path — same request, different visibility depending on whether the query matched an entity token. Both paths now use the explicit comparison.
- **llm** — Gemini / Vertex / Fake `complete_json` rejected the structured-output kwargs (`seed`, `response_schema`) with `TypeError`. The fallback retry loop swallowed it, so entity extraction silently degraded to the regex fallback for every non-OpenAI deployment. The kwargs are now accepted-and-ignored and the provider protocol documents that contract.
- **events** — lifecycle payloads used `extra="forbid"`; an additive publisher field would fail validation and silently drop every archive/purge/forge request during a rolling deploy. Now `extra="ignore"`, matching the other consumer schemas.
- **events** — the lifecycle consumer built payloads via kwargs-splat; a malformed (non-dict) payload raised `TypeError`, escaped `except ValidationError`, and nacked into redeliver-forever. Now `model_validate`.

## Tests
`tests/test_audit_s6_c1_events.py` (8 cases) covering all four.


erni-a:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #324, opened 2026-06-11
CAURA-134 fix(contradiction): raise context-fetch timeout 5s → 30s + improve failure log

ran-taig:
## Summary

Wet test on dev v2.14.0 today (post-CAURA-133, with CAURA-131 wiring + CAURA-133 prompt both deployed) showed the priya silence still present: **0 / 5 trials surfaced a contradiction** even though every trial had both sides resolved to the same canonical `priya` entity row.

CAURA-132 forensic logs revealed the cause is **not** the prompt and **not** the wiring — it's the 5-second `asyncio.wait_for` budget on the context fetch timing out:

```
PATH_C_DETECTION entry           candidates_initial=8
PATH_C_DETECTION after_a1_17     remaining=8
PATH_C_DETECTION judge_selection candidates=8 entity_aware=0 base=8  ← !!
PATH_C_DETECTION verdict         judge=base verdict=False  (× 8)

WARNING: Path C entity-links context fetch failed for memory <id>: .
         Falling through to base LLM judge.
```

The empty bit after `failed:` is the fingerprint of `asyncio.TimeoutError` (whose default `str(e)` is empty). With dev's accumulated tenant state, the parallel `_fetch_entity_context` gather exceeds 5s — every candidate falls back to the base judge. **CAURA-131's entity-aware judge is silently bypassed; CAURA-133's prompt rewrite is never exercised.**

## What this PR changes

| Layer | Before | After |
|---|---|---|
| Context-fetch timeout | hard-coded `5.0` at both call-sites | new `_CONTEXT_FETCH_TIMEOUT_SECONDS = 30.0` module constant; both call-sites reference it |
| WARNING log shape | `… failed: %s. Falling through …` (empty for `TimeoutError`) | `… failed: %s: %s. Falling through …` with `type(e).__name__` always rendered |
| Symmetric failure observability | success log only (`PATH_C_DETECTION context_fetched`) | + new `PATH_C_DETECTION context_fetch_failed memory=… exc_type=… candidates=N` INFO log |

Both call-sites:
- Retraction path (CAURA-129, ~`contradiction_detector.py:1414`) — gets prefix `PATH_C_RETRACTION context_fetch_failed`.
- Detection path (CAURA-131, ~`contradiction_detector.py:1700`) — gets WARNING update + INFO failure log.

## Why 30s is safe (not user-facing latency)

Path C runs in a **fire-and-forget background task after the write has already returned 201**. The timeout only bounds the background task's wall-clock — it never gates user-perceived write latency. The only side effects of raising it are:
- **Time-to-flag on slow tenants improves** (was: never, because the 5s timeout silenced them; now: ~10-30s).
- Hung Pub/Sub leases extend from ≤5s to ≤30s on storage stalls. Worker concurrency is async-bound and sized for this.

## Tests

| File | New / changed | Asserts |
|---|---|---|
| `test_caura_131_entity_aware_path_c_detection.py` | extended `test_fallback_to_base_judge_when_context_fetch_fails` | WARNING includes `RuntimeError` class name; INFO `context_fetch_failed` fires with `exc_type=RuntimeError` |
|  | **NEW** `test_fallback_to_base_judge_logs_timeouterror_explicitly` | Raises `asyncio.TimeoutError` (dominant prod failure, empty `str(e)`); asserts WARNING still names the class. Locks in the diagnosability fix against the empty-str(e) trap. |
|  | **NEW** `test_context_fetch_timeout_constant_is_generous_enough` | Guards against anyone tightening `_CONTEXT_FETCH_TIMEOUT_SECONDS` back below 15s. |
| `test_a4_13_path_c_retraction.py` | **NEW** `test_retraction_context_fetch_failure_logs_exc_type` | Retraction WARNING has `PATH_C_RETRACTION context_fetch_failed` prefix + `TimeoutError` class name. |

Tests collected:
- `test_caura_131_entity_aware_path_c_detection.py`: **10** (was 8 + 2 new)
- `test_a4_13_path_c_retraction.py`: **12** (was 11 + 1 new)

Validation: `ruff check`, `ruff format --check`, `pytest --collect-only` all clean.

## Wet-test plan (post-deploy)

CAURA-134 is the **unblocker for CAURA-131/133 verification**. Today we cannot tell if those PRs work because the judge they wire to isn't being invoked.

| Probe | Verifies |
|---|---|
| W1 | GCP query for `"context fetch failed"` warnings — pre vs post — drops materially |
| W2 | `repro_contradictions_collision.py` × 5 → `context_fetched` lines reappear → `judge_selection entity_aware>0 base=0` |
| W3 | For W2 trials with `entity_aware>0`: pull `PATH_C_DETECTION verdict judge=entity_aware` → **≥4/5 verdict=True** verifies CAURA-133's prompt fix |
| W4 | 10-min sweep for any new warning classes (no regressions) |
| W5 | Retraction-path warnings also drop (same fetch, same fix) |

## Rollback

Single `git revert`. Constant bump + log format + 1 new log line. No schema, no migration, no behavior change in success paths.

### DIFF
--- a/core-api/src/core_api/services/contradiction_detector.py
+++ b/core-api/src/core_api/services/contradiction_detector.py
@@ -1069,6 +1069,27 @@ _ENTITY_CONTEXT_NAME_MAX_CHARS = 100
 # it can.
 _ENTITY_LINKS_PREFLIGHT_MAX_CANDIDATES = 20
 
+# CAURA-134 — timeout budget (seconds) for the parallel
+# ``_fetch_entity_context`` gather in both Path C entry points
+# (retraction and forward-overlap detection). Raised from 5s -> 30s
+# after CAURA-132 forensic logs showed the 5s ceiling silently firing
+# on dev v2.14.0 with accumulated tenant state, causing
+# ``contexts_fetched`` to stay False and ALL candidates to fall back
+# to the base LLM judge. The fallback bypasses CAURA-131's entity-
+# aware wiring and CAURA-133's entity-aware prompt — i.e. the priya-
+# class silence was being driven by this timeout, not by a weak
+# prompt.
+#
+# 30s is safe because Path C runs in a fire-and-forget background
+# task after the write has already returned 201; the timeout only
+# bounds the background task's wall-clock, not user-perceived
+# latency. The two storage round-trips per memory (one batch +
+# per-link gather) complete in ~50-200ms p95 in the working
+# trials we observed; 30s is comfortably above any reasonable
+# per-tenant accumulated-state cost while still cancelling truly
+# hung tasks.
+_CONTEXT_FETCH_TIMEOUT_SECONDS = 30.0
+
 # CAURA-133 — process-scoped prefix for the missing-entity_id sentinel
 # emitted by ``_format_entity_context``. A real entity_id (always a
 # UUID written by the entity-extraction worker) cannot accidentally
@@ -1407,23 +1428,37 @@ async def _attempt_path_c_retraction(
     # (degenerate inputs / extractor failure).
     # Wrap the fetch in ``asyncio.wait_for`` so a hung storage
     # round-trip cannot block Path C indefinitely (mirrors the LLM
-    # call's cancellation boundary below). 5s budget is ample for two
-    # parallel storage calls (~50-200ms p95 in practice); on failure
-    # (timeout, network, storage error), treat as "no context, leave
-    # Path A alone" rather than retrying.
+    # call's cancellation boundary below). On failure (timeout,
+    # network, storage error), treat as "no context, leave Path A
+    # alone" rather than retrying. See ``_CONTEXT_FETCH_TIMEOUT_SECONDS``
+    # for the timeout rationale (CAURA-134).
     try:
         new_entities, old_entities = await asyncio.wait_for(
             asyncio.gather(
                 _fetch_entity_context(sc, str(new_memory.get("id"))),
                 _fetch_entity_context(sc, str(candidate.get("id"))),
             ),
-            timeout=5.0,
+            timeout=_CONTEXT_FETCH_TIMEOUT_SECONDS,
         )
     except Exception as e:
+        # CAURA-134 — include exception class name in the log. The
+        # default str(e) is empty for ``asyncio.TimeoutError``, which
+        # made the old "failed: . Path A's verdict stands." message
+        # un-diagnosable; the class name disambiguates timeouts from
+        # network errors from malformed responses.
+        #
+        # No symmetric INFO line here: the retraction path has no
+        # success-side ``context_fetched`` INFO to mirror (unlike the
+        # detection path, which emits one on the happy path for GCP
+        # metric parity). A redundant failure-only INFO would skew
+        # any retraction-path success/failure counter built from
+        # ``context_fetched`` / ``context_fetch_failed`` pairs.
         logger.warning(
-            "Path C entity-context fetch failed for memory %s candidate %s: %s. Path A's verdict stands.",
+            "PATH_C_RETRACTION context_fetch_failed memory=%s candidate=%s "
+            "exc_type=%s exc=%s. Path A's verdict stands.",
             new_memory.get("id"),
             candidate.get("id"),
+            type(e).__name__,
             e,
         )
         return False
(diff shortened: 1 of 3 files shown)


ran-taig:
@claude

ran-taig:
@claude please re-review — addressed two follow-up nits in the latest force-push (commit `fa2125bd`):

**Fix 1 — Symmetric `PATH_C_RETRACTION context_fetch_failed` INFO log** added after the WARNING in the context-fetch except block, mirroring the analogous `PATH_C_DETECTION context_fetch_failed` line in the detection path. GCP queries can now grep both paths uniformly without inferring retraction-path failures from WARNING-only signals.

**Fix 2 — LLM judge WARNING rewritten** with grep-friendly `PATH_C_RETRACTION judge_failed` prefix and explicit `exc_type=%s exc=%s` fields. Same bug class as the context-fetch warning — `str(e)` is empty for `asyncio.TimeoutError`, so the previous `'judge failed for memory ... %s'` was un-diagnosable on the dominant production failure mode.

Existing CAURA-134 unit tests still cover the new shapes (the retraction-path test asserts the `PATH_C_RETRACTION context_fetch_failed` substring + `TimeoutError` class name, which both still hold).

Validation: `ruff check`, `ruff format --check`, `pytest --collect-only` (22 tests) all clean.

ran-taig:
Force-pushed `9e5be9b9` — fixes the 3 CI test failures from the previous push.

**Root cause of the test failures:** my tests mocked `sc.get_entity_links_for_memories` to raise. But `_fetch_entity_context` (the helper that wraps the storage call) has its OWN try/except that catches storage errors, logs an internal `Path C entity-context fetch failed (links) for memory ...` warning, and returns `[]`. So the outer `asyncio.wait_for` saw an empty list, NOT an exception — and the CAURA-134 outer WARNING + INFO logs never fired.

The fix: patch `_fetch_entity_context` directly instead of the inner storage method. This makes the exception surface at the OUTER `wait_for` boundary where the CAURA-134 log code lives — accurately modeling the production failure mode (which is `asyncio.TimeoutError` from `wait_for` cancelling a slow gather, not an inner storage call raising).

Three tests updated:
- `test_fallback_to_base_judge_when_context_fetch_fails`
- `test_fallback_to_base_judge_logs_timeouterror_explicitly`
- `test_retraction_context_fetch_failure_logs_exc_type` (also extended to assert the new symmetric INFO log)

@claude re-review please.

ran-taig:
Force-pushed addressing both review nits:

**Fix 1 — detection-path WARNING now uses grep-friendly key=value form** mirroring the retraction-path WARNING. Old shape was `Path C entity-links context fetch failed for memory %s: %s: %s. Falling through...`; new shape is `PATH_C_DETECTION context_fetch_failed memory=%s exc_type=%s exc=%s candidates=%d. Falling through to base LLM judge.` — same prefix string as the retraction WARNING and the INFO mirror, so a single GCP query can grep all of them.

**Fix 2 — removed the orphaned retraction-path INFO** + added a doc comment to the surviving detection-path INFO. Reasoning: the detection path has a success-side `PATH_C_DETECTION context_fetched` INFO that the failure INFO genuinely mirrors (a single GCP metric definition can match both lines and split by outcome). The retraction path has no `PATH_C_RETRACTION context_fetched` success-side counterpart, so a failure-only INFO there was a redundant log line that would skew any retraction-path success/failure counter built from the pair. The WARNING already carries the full diagnostic on the retraction path.

Tests updated:
- `test_fallback_to_base_judge_when_context_fetch_fails` — WARNING assertion strengthened to the new key=value shape (`exc_type=RuntimeError` AND `candidates=1`)
- `test_retraction_context_fetch_failure_logs_exc_type` — INFO assertion inverted: now asserts the INFO is NOT emitted on the retraction path (locks in the orphan removal)

@claude re-review please.

ran-taig:
@claude

ran-taig:
Force-pushed `0f81f669` addressing the medium-severity review (the two race-condition reviews disagreed; took the more careful analysis).

**Fix — removed the detection-path INFO mirror.** The previous round added a symmetric INFO log on the detection-path failure (intended to mirror the success-side `context_fetched` INFO for GCP metric parity), but the reviewer correctly flagged that:

- GCP log-based metrics default to `severity>=INFO`, which INCLUDES WARNING.
- The WARNING + INFO emitted on a single failure event both contain `PATH_C_DETECTION context_fetch_failed`, so any metric matching that text at default severity sees the failure 2x — distorting any success/failure ratio.
- The retraction path is WARNING-only and never had this issue; the detection path now matches that shape.

**Updated:**
- Removed the `logger.info(...)` call in the detection-path exception handler.
- Updated the inline comment block to explain why we're WARNING-only and what a paired metric would have to do (define two filters at different severities, not match both at INFO+).
- Test `test_fallback_to_base_judge_when_context_fetch_fails` updated: INFO assertion **inverted** to assert the INFO is NOT emitted (locks in the absence so a future "symmetry" refactor can't silently reintroduce the metric-distortion bug).

The retraction-path WARNING-only shape is unchanged from last round (already correct).

22 tests collected; ruff + format clean. @claude re-review please.

ran-taig:
@Eldad-Caura

Eldad-Caura (approved):

PR #323, opened 2026-06-11
fix(plugin): point agents at the memclaw skill by name, not by filesystem path (CAURA-000)

ran-taig:
## Summary

Cron-run error on bridgeclaw (plugin 2.9.0):

```
channel=cron sessionKey=agent:bridgeclaw:cron:... duration=171759ms outcome=error
error="search "~/.openclaw/agents/bridgeclaw/agent/codex-home"
       -> search "memclaw/SKILL.md$|memclaw" (agent) failed"
```

**Interpretation**: this is OpenClaw's built-in **`search` tool** failing — not a memclaw tool. The agent was trying to *locate* our skill file, driven by our own prompt/educate text: *"the operating contract lives in `skills/memclaw/SKILL.md` — read it before your first call."*

That path-style pointer was correct when the plugin wrote `SKILL.md` per-workspace. But the skill is now shipped **once at plugin-root** and **manifest-discovered** via `openclaw.plugin.json:skills` — there is no per-workspace file. A cron agent (no human in the loop) took the path literally, ran the `search` tool against its `codex-home` workspace, scanned for **2m51s**, found nothing, and failed the turn.

## Fix

Three emit sites carried the path-style pointer:
- `prompt-section.ts` — per-turn system-prompt fragment
- `educate.ts` `buildToolsMd()` — TOOLS.md (per-workspace bootstrap)
- `educate.ts` `buildAgentsMd()` — AGENTS.md (per-workspace bootstrap)

All three reworded to:
1. Reference the **memclaw** skill **by name** (not a path)
2. State the runtime loads it automatically
3. Explicitly instruct the agent **not to search the filesystem** for it

**No runtime or behavior change — prompt wording only.**

Example (AGENTS.md, after):
> Before your first MemClaw call this session, open the **memclaw** skill — your runtime loads it automatically, so do NOT search the filesystem for it — for signatures, cadences, quality, prohibitions, recall policy, and sharing.

## Tests

Three existing assertions *locked in the bug* (they required the `skills/memclaw/SKILL.md` path string). Updated all three to instead require:
- absence of the `skills/memclaw/SKILL.md` path string
- the by-name `**memclaw** skill` reference
- the `do NOT search the filesystem` guard

(`educate.test.ts` ×2, `tool-definitions.test.ts` ×1. Whitespace normalised before matching — the source template wraps phrases across lines.)

- 310/310 plugin tests pass
- manifest drift 11/11 pass
- tsc clean

## Wet test

[...]

PR #325, opened 2026-06-11
feat(skills): per-node skill-reconcile observability on the heartbeat

ran-taig:
## Why

#317 gated which skills *should* land on each OpenClaw node, but there was no way to confirm which actually *did*. This closes the loop: the plugin's reconcile summary rides the heartbeat to the backend and is readable per-node, so an operator who flips a skill to `active` can confirm it reached the fleet — and see *why* a malformed catalog row was skipped.

## What's new

**Plugin**
- `reconcile-skills.ts`: adds `installed: string[]` to `ReconcileSummary` — the catalog-active skills converged onto disk *this tick*. Unlike `added`/`removed` (per-tick deltas, empty at steady state), `installed` is the **standing truth**: what's live on this node right now.
- `heartbeat.ts`: captures the `reconcileSkills()` result (previously discarded) and sends it as a typed top-level `reconcile` field.

**Backend**
- `fleet.py` `HeartbeatIn` gains `reconcile` (8 KB anti-ballooning cap, mirroring `recall_metrics`), merged into `nodes.metadata.reconcile` (latest snapshot wins). No migration — `metadata` is already JSONB. Surfaces via `GET /api/v1/fleet/nodes`.

## Latent bug fixed (required for the read path)

`GET /fleet/nodes` read `n.get("metadata")`, but the storage layer serialises that JSONB column under its ORM attribute name **`extra`** (the column is `metadata`; the model renames it to avoid shadowing SQLAlchemy's `MetaData`). So node metadata — `recall_metrics`, `deploy_blocked_until`, and now `reconcile` — **never** surfaced through this endpoint; the field was always `null`. Fixed to read `extra` (with a `metadata` fallback). Purely additive (`null` → the actual snapshot).

## Tests

- `plugin/src/reconcile-skills.test.ts` — `installed` reflects the converged set (cold start), empty on empty catalog, **reported even on steady-state ticks with no deltas**, and shrinks on de-activation.
- `tests/test_api_fleet.py` — a heartbeat `reconcile` lands at `node.metadata.reconcile` and is readable via `/fleet/nodes` (exercises the latent-bug fix); latest snapshot wins; over-cap blob → 422.
- 312 plugin tests pass; 61 across the fleet/heartbeat sweep. ruff + format clean; 0 mypy errors in the touched route.

## Docs

`docs/mcp-skill-delivery.md` — "Verifying installs reached the fleet": the heartbeat summary, `nodes.metadata.reconcile`, and `installed` as standing truth.

## Scope

Deliberately deferred to follow-ups: install/bootstrap hardening (`/install-plugin`, gateway restart, version pinning); skipped-reason enrichment; an MCP fleet-status tool.

### DIFF
--- a/core-api/src/core_api/routes/fleet.py
+++ b/core-api/src/core_api/routes/fleet.py
@@ -100,6 +100,15 @@ class HeartbeatIn(BaseModel):
     # reading the column (e.g. "blocked since 1970?"). Fail-loud here
     # instead.
     deploy_blocked_until: int | None = Field(None, ge=1)
+    # Skill-reconcile observability: the plugin's latest
+    # ``reconcileSkills()`` summary — ``installed`` (the active skills
+    # converged onto this node's disk), per-tick ``added``/``removed``
+    # deltas, ``skipped`` (bad-shape catalog rows), ``protected``, and
+    # ``catalogCount``. Stored as the newest snapshot on the node row
+    # (``nodes.metadata.reconcile``) so an operator can confirm an
+    # approved/active skill actually landed on the fleet. Optional —
+    # older plugin versions don't send it.
+    reconcile: dict | None = None
 
     @field_validator("recall_metrics")
     @classmethod
@@ -110,6 +119,17 @@ class HeartbeatIn(BaseModel):
         # with every reason populated; 4 KB is a comfortable headroom.
         if v is not None and len(json.dumps(v)) > 4096:
             raise ValueError("recall_metrics exceeds 4 KB limit")
+        return v
+
+    @field_validator("reconcile")
+    @classmethod
+    def _cap_reconcile(cls, v: dict | None) -> dict | None:
+        # Same anti-ballooning cap as recall_metrics. The summary is a
+        # handful of slug lists; even a fleet sharing ~150 skills stays
+        # well under 8 KB. Reject beyond that at the API boundary so a
+        # misbehaving plugin can't bloat the node row.
+        if v is not None and len(json.dumps(v)) > 8192:
+            raise ValueError("reconcile summary exceeds 8 KB limit")
         return v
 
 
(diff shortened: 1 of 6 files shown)


ran-taig:
@claude please review. This threads the plugin's skill-reconcile summary onto the heartbeat into `nodes.metadata.reconcile`, surfaced via `GET /fleet/nodes`. Please scrutinize: (1) the **latent-bug fix** — `/fleet/nodes` read `n.get('metadata')` but storage serialises that column under the ORM attr `extra`, so node metadata never surfaced; confirm the `extra`-with-`metadata`-fallback fix is correct and doesn't break other consumers; (2) the 8 KB cap on `reconcile`; (3) `installed` semantics (standing truth vs per-tick delta); (4) no regression to existing heartbeat fields (recall_metrics/deploy_blocked_until now also surface — intended).

ran-taig:
## Addressed — `installed` now reflects confirmed on-disk state

Good catch. `installed` was derived from `desired.keys()` (intent), which diverges from disk when a per-skill `writeFileSync` throws — contradicting the field's "standing on-disk truth" contract.

Fixed exactly as suggested: maintain an `installedSet` seeded from `onDisk` minus cleanly-removed slugs (so a same-content skill skipped by the no-op-match branch still counts — it's genuinely present), augmented by each **confirmed** write. `summary.installed` is now `[...installedSet].filter(s => !PROTECTED_SKILLS.has(s)).sort()` — so a skill whose write failed is correctly absent, and the bundled `memclaw` stays excluded (reported via `protected`).

New regression test `installed reflects CONFIRMED disk, not desired intent: a failed write is excluded` — plants a *file* where the reconciler wants a directory so `mkdirSync` throws for one slug (real failure, no fs mocking); asserts the failed skill is excluded from `installed` while the clean one is included.

All prior `installed` assertions (cold start, empty catalog, steady-state, de-activation) still hold under the new derivation. 313 plugin tests pass.

@claude please re-review.

ran-taig:
@claude

ran-taig:
## Addressed — seed `installedSet` from `onDisk ∩ desired`

Right — `summary.removed` holds only *successful* removals, so a deactivated skill whose `rmSync` threw was absent from `removed` yet still in `onDisk`, passing the filter and getting reported as installed despite no longer being catalog-active.

Replaced the `summary.removed` filter with an intersection against the catalog-active set:

```ts
const desiredSlugs = new Set(desired.keys());
const installedSet = new Set([...onDisk].filter((s) => desiredSlugs.has(s)));
```

then `installedSet.add(slug)` on each confirmed write. This is correct in every case:
- same-content skill (no-op branch): in `onDisk ∩ desired` → kept;
- new install: not on disk → added only on successful write;
- cleanly removed orphan: not in `desired` → excluded;
- **failed removal of a deactivated skill**: not in `desired` → excluded (the bug you flagged).

All prior `installed` assertions (cold start, empty catalog, steady-state, de-activation) still hold; the failed-write regression test still passes. 313 plugin tests + 10 backend fleet tests green; ruff/format clean.

> Note: the branch had been auto-rebased onto latest `main` (#323/#326) by tooling; I've rebased this round's work onto it cleanly (no conflicts — #323's skill-by-name change is in educate/prompt text, not the reconciler) so the PR is current with `main`.

@claude please re-review.

ran-taig:
@erni-a

erni-a (approved):

PR #304, opened 2026-06-08
feat(graph-build): drop literal/attr entity nodes, block suffix-distinct merges, rerank entity_lookup by query overlap

arkash20:
# feat(graph-build): fix entity_lookup recall collapse at scale (literal/attr nodes + suffix-merge)

**Branch:** `feat/entity-graph-build-fix` (`6d9ce66`)
**Files:** `core-api/src/core_api/services/entity_extraction_worker.py`, `core-api/src/core_api/pipeline/steps/search/classify_query.py`

## Problem
`entity_lookup` recall **collapses at scale**. On the S1 benchmark (entity-named attribute lookups):

| scale | recall |
|---|---|
| K=100 | 22/25 |
| **K≈1000** | **7/25 (≈0.28)** |

This is the bench's documented "K-collapse" (condition C → 0.31 at K≥1000). Root-caused to the **graph build**, not retrieval.

## Root cause
The LLM entity-extraction over-populates the entity/relation graph two ways:

- **(A) Literal values and attribute names become entity NODES.** Dates (`2024-03-23`), percentages (`95.1%`), money, and attribute *names* (`sla_uptime`, `q3_revenue`) are extracted as entities (typed `identifier`/`concept`/`artifact`). Attribute names become **high-degree hubs** (`q3_revenue` deg 11, `sla_uptime` deg 9) that bridge every entity sharing an attribute.
- **(B) Entity resolution merges suffix-distinct names.** Names differing only by a `#NNNN` suffix embed ≈identically → cosine ≥ `ENTITY_RESOLUTION_THRESHOLD (0.85)` → merge (first-seen-wins). All `Comet #NNNN` collapse into one contaminated node carrying every Comet's edges.

**Consequence:** in `entity_lookup`, 2–3 matched entities graph-expand (`_expand_per_fleet`, 3-hop union) through these hubs/merged-nodes to **~the entire corpus** (2 seeds → 51 entities/96 mems @K=100; 151/1041 @K=1000), all at near-uniform hop-boost. The `GRAPH_MAX_BOOSTED_MEMORIES=50` cap then drops the exact entity **before** any ranking → miss. (Dump evidence: a single `Comet #0002` fact carried 5 edges incl. *other* Comets' launch_dates + a foreign `owns mark lin`.)

## Fix
**entity_extraction_worker.py**
[...]

PR #327, opened 2026-06-11
fix(search): skip entity hop-boost when classifier declined the entity match as over-broad

arkash20:
## Problem

S1 retrieval@10 collapses at corpus scale even after #304: **22/25 @ K=1000 → 11/25 @ K=10000** (10,000 entity-attribute facts, 50-query replay).

Per-failure anatomy (`Comet #0002's launch_date`, gold present in DB among 272 `Comet #NNNN launch_date` siblings / 813 distinct Comet entities):

- The CAURA-698 decline fires identically at both corpus sizes (86 and 815 entity-FTS matches > `ENTITY_LOOKUP_MAX_MATCHES=20`; the entity FTS is OR-over-tokens, so `comet` alone matches every sibling). Both route to `semantic_search` — the decline does not explain the collapse.
- On pure hybrid scoring the gold ranks **#1**: vec_sim 0.9084 vs 0.8651 for the best sibling, and it is the only row matching the FTS query (its tsvector contains `0002`).
- What buries it: `ParallelEmbedAndEntityBoost` re-runs the same tokenizer + entity FTS **with no knowledge of the classifier's decline**, re-derives the same over-broad match, and hands `GRAPH_HOP_BOOST[0]=1.3` to the first `GRAPH_MAX_BOOSTED_MEMORIES=50` memory links it encounters. With hundreds of entities tied at hop 0, link order is arbitrary — the boost is a fixed-50 lottery over the sibling pool. P(gold boosted) ≈ 50/N_siblings: ~57% at K=1000 (gold boosted → rank 1, verified live), ~6% at K=10000 (gold unboosted at ~0.66 buried under ~50 boosted siblings at 0.78–0.85).

One step concludes the entity match is too broad to mean anything and discards it; the next step recomputes the identical match and acts on it.

## Fix

Thread the verdict through the pipeline context:

- `ClassifyQuery` sets `ctx.data["entity_match_declined"] = True` when the CAURA-698 decline fires.
- `ParallelEmbedAndEntityBoost` skips its entity-boost leg on that flag (embedding — the critical path — is unaffected; without the flag, behavior is unchanged).

## Measurements (S1 replay, local stack, retrieval@10)

| configuration | K=1000 | K=10000 |
|---|---|---|
| main (#304) | 22/25 | 11/25 |
| this fix | 22/25 (no regression) | 16/25 |
| this fix, `search.recall_boost=false` pinned (deterministic replay) | 22/25 | **18/25** |

The recall_boost pin removes reused-tenant contamination: `TrackRecalls` increments `recall_count` on returned rows, so earlier replays had fed recall_boost to the wrong siblings they returned. Remaining K=10000 misses are pre-existing classes (person-hub expansion, true semantic near-duplicate dilution) tracked as A30 — the exact-first cascade.

## Tests

- `tests/test_skip_entity_boost_on_decline.py` — 4 new unit tests pinning both halves of the flag plus the no-flag default (boost still runs).
- Adjacent `tests/test_d7_parallel_embed_per_task_timeout.py` (7 tests) still green — the None-task guards preserve the D7 split-budget contract.
- `ruff check` / `ruff format` / `mypy` clean on touched files.

Regression bench used for the numbers: `caura-bench-memcomp/scripts/s1_regression_bench.py` (replay-only against seeded k1k/k10k tenants, committed baselines).

### DIFF
--- a/core-api/src/core_api/pipeline/steps/search/classify_query.py
+++ b/core-api/src/core_api/pipeline/steps/search/classify_query.py
@@ -94,6 +94,13 @@ class ClassifyQuery:
                         len(matched_ids),
                         ENTITY_LOOKUP_MAX_MATCHES,
                     )
+                    # The same over-broad match must not be re-derived by
+                    # ParallelEmbedAndEntityBoost and used for hop-boosting:
+                    # with N >> GRAPH_MAX_BOOSTED_MEMORIES sibling entities all
+                    # at hop 0, the boost degenerates into an arbitrary-50
+                    # lottery that buries rows pure scoring ranks first
+                    # (S1 @K=10000: 11/25 vs rank-1 on unboosted score).
+                    ctx.data["entity_match_declined"] = True
                     matched_ids = []
 
                 if matched_ids:
--- a/core-api/src/core_api/pipeline/steps/search/parallel_embed_entity_boost.py
+++ b/core-api/src/core_api/pipeline/steps/search/parallel_embed_entity_boost.py
@@ -177,15 +177,23 @@ class ParallelEmbedAndEntityBoost:
         emb_task = asyncio.ensure_future(
             _get_or_cache_embedding(data["query"], data["tenant_id"], data["tenant_config"])
         )
-        ent_task = asyncio.ensure_future(
-            _entity_boost_via_storage(
-                data["query"],
-                data["tenant_id"],
-                data.get("fleet_ids"),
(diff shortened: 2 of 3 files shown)


Eldad-Caura (approved):

arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #328, opened 2026-06-11
fix(logging): reroute idempotency guard mis-keys on captured handlers

Eldad-Caura:
## Why

`_route_third_party_to_root`'s already-rerouted check required the snapshot's handler list to be non-empty, so a logger rerouted from `propagate=False, no-handlers` (snapshot `([], False, lvl)`) was never considered done — a re-call (e.g. ASGI lifespan startup after uvicorn initialised) fell through to the uninitialised early-exit branch and emitted a spurious **'rerouting was a no-op' WARNING** in Cloud Logging.

## What

- Key the idempotency guard on snapshot presence alone — the snapshot is only stored when rerouting actually ran, so `name in dict` is the complete signal.
- Fix the module comment claiming the uninitialised case 'logs a debug line' (it logs a WARNING).
- Regression test for the handler-less reroute + re-call path — verified to fail pre-fix and pass post-fix.

Found by review on caura-memclaw-enterprise#423 (the enterprise vendor of this file); the enterprise copy will be re-vendored from this.

## Verification
- `pytest tests/test_logging.py` → 21 passed.
- `ruff check` + `ruff format --check` clean.
- DCO signed.

### DIFF
--- a/common/structlog_config.py
+++ b/common/structlog_config.py
@@ -329,7 +329,7 @@ def configure_logging(
 # ``Config.configure_logging()`` runs at server start; if our
 # ``configure_logging()`` is called at module import (the typical app.py
 # path), the uvicorn loggers may not yet have their handlers when we
-# iterate. The function logs a debug line in that case so the caller can
+# iterate. The function logs a WARNING in that case so the caller can
 # move the call into an ASGI lifespan startup handler if production logs
 # show un-rerouted output.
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #330, opened 2026-06-11
fix(events): pubsub shutdown never flushed the publisher — final batches silently lost

Eldad-Caura:
## Why

`PubSubEventBus.stop()` calls `publisher.close()` — **which doesn't exist on `PublisherClient`** (that's a SubscriberClient method). The AttributeError is swallowed by the per-step guard and logged as `publisher.close() failed; continuing teardown` on every service shutdown (visible in prod logs today). The real API is `PublisherClient.stop()`: commits outstanding batches + joins the commit thread — and it's the **only** flush in the pipeline, since `publish()` is fire-and-forget into the client's batch queue.

**Impact**: long-running services rarely notice (the commit thread transmits within ~10ms), but short-lived processes lose their entire final batch. Verified in prod 2026-06-11: a tenant re-embed backfill (`core_worker.cli backfill-embeddings`) reported `published=16` while **zero events reached the topic** — confirmed via topic `send_message_operation_count` and a probe job that awaited the publish future.

## What

- `stop()` → `publisher.stop()` (drains); log message updated.
- `core_worker.cli` drains the bus in its `finally` (the prod incident's direct trigger).
- Corrected `publish()`'s at-least-once comment (only holds when `stop()` runs before exit).
- Test publisher mocks **spec-limited** to the real client surface — a permissive `MagicMock` accepting `.close()` is exactly how this survived the existing `stop()` tests. Assertions moved to `.stop()`.

## Verification

- `pytest tests/test_events/` → 80 passed; `tests/test_logging.py` unaffected (21 passed).
- `ruff check` + `ruff format --check` clean. DCO signed.
- Operationally proven in prod: re-running the same backfill with an explicit `PublisherClient.stop()` flush delivered 16/16 events (tenant fully re-embedded).

### DIFF
--- a/common/events/pubsub.py
+++ b/common/events/pubsub.py
@@ -396,11 +396,16 @@ class PubSubEventBus(EventBus):
         # (admin-API request handlers await this, so queued publish
         # threads translate directly to queued requests).
         #
-        # At-least-once is preserved: stop()'s shutdown(wait=True) on the
-        # publish executor drains any in-flight batches before the
-        # process exits. The tradeoff is that we no longer surface
-        # per-message publish errors to the caller — publisher-side
-        # failures land in the SDK's background-thread log instead.
+        # At-least-once is preserved ONLY if stop() runs before process
+        # exit: the executor shutdown drains the enqueue calls, and
+        # stop()'s publisher.stop() commits the client's outstanding
+        # batches (the actual transmission happens on the SDK's
+        # background commit thread, NOT in the executor). Short-lived
+        # processes that publish and exit without awaiting stop() lose
+        # whatever is still batched. The tradeoff is that we no longer
+        # surface per-message publish errors to the caller —
+        # publisher-side failures (e.g. a 403 on the topic) land in the
+        # SDK's background-thread log instead.
         # For a fire-and-forget audit path that is the right shape.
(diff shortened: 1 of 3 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
Applied (amended): the `finally`-block drain is now guarded — `get_event_bus()` factory errors (RuntimeError on missing pubsub env, ValueError on unknown backend) are caught and logged as 'event bus stop failed; continuing teardown' instead of masking the backfill's original exception. Tests + ruff clean.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

Eldad-Caura:
Extended (amended): the same nonexistent-`close()` bug existed on the three publisher **candidate** cleanup paths (`_close_pending` / `_close_post_stop` / `_close_loser`) — no data loss there (an uninstalled candidate has no batches) but the swallowed AttributeError meant the SDK commit thread was never joined (thread leak, debug-log noise). All three now call `candidate.stop()`; the TOCTOU test fakes now model the real PublisherClient surface (`stop()`, no `close()`) so a regression can't pass. 80 events tests green; ruff clean. The subscriber-side `close()` calls are untouched (valid API there).

PR #331, opened 2026-06-11
fix(events): publisher candidate cleanups also called the nonexistent close()

Eldad-Caura:
Follow-up to #330 (raced its merge — this was meant as an amendment). The three fire-and-forget publisher-**candidate** cleanup paths (`_close_pending`, `_close_post_stop`, `_close_loser`) had the same nonexistent-`PublisherClient.close()` bug: no data loss (an uninstalled candidate has no batches), but the swallowed AttributeError meant the SDK commit thread was never joined — a thread leak + debug-log noise on every TOCTOU/cancelled construction.

- All three now call `candidate.stop()`; log messages updated.
- TOCTOU test fakes now model the real `PublisherClient` surface (`stop()`, no `close()`) so a regression can't pass.
- Subscriber-side `close()` calls untouched (valid API there).

80 events tests green; ruff check + format clean; DCO signed.

### DIFF
--- a/common/events/pubsub.py
+++ b/common/events/pubsub.py
@@ -240,7 +240,11 @@ class PubSubEventBus(EventBus):
             async def _close_pending() -> None:
                 try:
                     client = await ctor_fut
-                    await loop.run_in_executor(None, client.close)
+                    # stop(), not close(): PublisherClient has no close()
+                    # (same nonexistent-method bug as the stop() teardown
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #332, opened 2026-06-11
fix(llm): explicit httpx per-phase timeouts — 5s connect too tight behind VPC connector

Eldad-Caura:
## Why

Prod core-api shows a steady trickle of `httpcore.ConnectTimeout` from the enrichment/entity-extraction handlers (a few per hour, predates recent deploys). Root cause: passing a bare float to `AsyncOpenAI(timeout=...)` keeps httpx's **default 5s connect/pool phases**, and with `all-traffic` VPC egress every outbound call rides the connector + Cloud NAT — cold connections intermittently exceed 5s. Two effects: nack/redelivery noise (self-healing — 12h DLQ deliveries: zero), and **`Entity extraction failed` is caught as non-fatal, so the memory is acked without entity links — a permanent silent gap**.

## What

- `httpx.Timeout(connect=15, read=<request budget>, write=15, pool=10)` on the OpenAI-compatible client; connect/pool env-tunable (`OPENAI_HTTPX_CONNECT_TIMEOUT_SECONDS`, `OPENAI_HTTPX_POOL_TIMEOUT_SECONDS`).
- The bespoke `OPENAI_REQUEST_TIMEOUT_SECONDS` reader generalised to `_read_float_env` and reused.

Deliberately scoped: no keepalive-pool resizing and no handler-level retry in this PR — the runbook'd next steps if the trickle persists after this lands (expected reduction: most of it).

## Verification
- `pytest tests/test_p4_1_llm_fallback.py tests/test_events` → 118 passed (full-suite collection errors are a pre-existing local `tiktoken` gap, also on main).
- `ruff check` + `format --check` clean. DCO signed.

### DIFF
--- a/common/llm/constants.py
+++ b/common/llm/constants.py
@@ -53,8 +53,8 @@ LLM_FALLBACK_MODEL_OPENAI = os.environ.get("LLM_FALLBACK_MODEL_OPENAI", "gpt-5.4
 # enough that a single hung upstream call eats the whole enrichment
 # budget silently. 25s gives the provider room to respond while still
 # leaving budget for one retry under the inline ceiling.
-def _read_openai_request_timeout_seconds() -> float:
-    """Parse ``OPENAI_REQUEST_TIMEOUT_SECONDS`` defensively.
+def _read_float_env(name: str, default: float | None) -> float | None:
+    """Parse a float env var defensively.
 
     Bare ``float(os.environ.get(...))`` would raise ``ValueError`` at
     module import on a misconfigured value (e.g. ``"25s"`` instead of
@@ -62,7 +62,9 @@ def _read_openai_request_timeout_seconds() -> float:
     structured-logging is wired. Catch it here, write a warning to
     stderr, and fall back to the documented default.
     """
-    raw = os.environ.get("OPENAI_REQUEST_TIMEOUT_SECONDS", "25.0")
+    raw = os.environ.get(name)
+    if raw is None:
+        return default
     try:
         return float(raw)
     except (TypeError, ValueError):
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
Both applied (amended):
- **write timeout** restored to the full request budget (`write=request_timeout_seconds`) — the 15s value was an unintended regression vs the bare-float behaviour, exactly as flagged; large prompt uploads keep their 25s.
- **pool timeout** default raised to 25.0 for parity with the previous implicit behaviour (the fail-fast rationale conflated connection-slot waits with NAT cold paths); stays env-tunable downward if fail-fast under pool pressure is ever wanted. Comment added at the constant.

Net: only the **connect** phase gets new headroom (5→15s), which is the actual fix. Tests + ruff clean.

Eldad-Caura:
@claude

Eldad-Caura:
Applied (amended): `OPENAI_HTTPX_POOL_TIMEOUT_SECONDS` now defaults to `None` ⇒ the pool phase **tracks `request_timeout_seconds`** — exact parity with the bare-float behaviour for every configuration, including deployments running `OPENAI_REQUEST_TIMEOUT_SECONDS` above the default. The env var remains as an explicit decoupling override (`is not None` resolution, so an explicit `0.0` means don't-wait rather than unset). `_read_float_env` widened to `float | None`. Verified: default resolves to None→request budget; env override parses. 118 tests + ruff clean.

Net behavior change of this PR is now solely: **connect 5s → 15s** (env-tunable). Every other phase is at exact parity with what the bare float produced.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #333, opened 2026-06-11
fix(storage-client): retry POSTs on connection-phase failures — never-sent requests cannot double-insert

Eldad-Caura:
## Why

Prod 2026-06-11 (11:30–17:30Z): **42×** `Async contradiction detection failed` (`find_similar_candidates`) and **2×** audit flush failures (`create_audit_logs_bulk` — events permanently lost for the affected tenant). Every single one was a **first-attempt `httpx.ConnectTimeout`** reaching core-storage-api through the VPC connector during the Phase J re-embed backlog drain. Idempotent siblings (GET/PATCH/DELETE) recovered via the F5 retry wrapper; POST was excluded from retry because a retry after partial success could double-insert.

Note this is a **different leg** than #332: that fixed the AsyncOpenAI/LLM client's connect timeout; these failures are on the core-storage-api client's unretried POSTs and would have continued even with #332 deployed.

## What

The blanket POST exclusion is broader than the safety argument requires. `ConnectTimeout` / `ConnectError` / `PoolTimeout` are all raised **before a single request byte is written** — retrying them cannot duplicate anything. `ReadTimeout` and 5xx stay non-retried for POST (request reached storage, may have committed; safe retry there still needs storage-side idempotency keys).

- `_with_retry` gains `retryable_exceptions` / `retryable_statuses` params — defaults unchanged for idempotent callers.
- `_post_with_retry` encodes the POST policy once (connection-phase exceptions, no status retries); wired into `_post`, `_post_optional`, and the bespoke `claim_idempotency` POST.

## Verification

- `pytest tests/test_storage_client_retry.py tests/test_audit_queue.py tests/test_storage_client_routing.py` → 46 passed. New tests: POST retries ConnectTimeout/ConnectError and recovers; gives up after max attempts; does **not** retry ReadTimeout or 5xx; `_post_optional` covered.
- `ruff check` + `format --check` clean; mypy delta vs main: none (19 pre-existing transitive errors unchanged).

## Follow-up (separate PR)

core-worker's storage client has four unretried POSTs (`archive_expired`, `archive_stale`, `purge_soft_deleted`, `upsert_tenant_suppression`) with the same exposure.

### DIFF
--- a/core-api/src/core_api/clients/storage_client.py
+++ b/core-api/src/core_api/clients/storage_client.py
@@ -34,11 +34,18 @@ logger = logging.getLogger(__name__)
 #  - Max 3 attempts (1 initial + 2 retries)
 #  - Exponential backoff with jitter: ~0.2s, ~0.4s
 #  - Worst-case added latency: < 1s
-#  - Retryable on idempotent HTTP methods only (GET, PATCH, DELETE).
-#    POST is intentionally NOT retried: storage creates don't have
-#    server-side idempotency keys, so a POST retry after a partial
-#    success would double-insert. POST retries are a follow-up
-#    once idempotency-key support lands storage-side.
+#  - Idempotent HTTP methods (GET, PATCH, DELETE) retry the full
+#    transient set below plus retryable 5xx statuses.
+#  - POST retries connection-phase failures ONLY (ConnectTimeout /
+#    ConnectError / PoolTimeout — all raised before a single request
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@erni-a pleasr approvr

erni-a (approved):

PR #334, opened 2026-06-11
fix(core-worker): retry connection-phase failures on storage calls — shared policy in common/http_retry

Eldad-Caura:
## Why

Follow-up to #333, which gave core-api's storage POSTs connection-phase retry after the 2026-06-11 prod incident (42× contradiction-detection failures + dropped audit events — all first-attempt `ConnectTimeout` behind the VPC connector). core-worker's storage client had the **same exposure with no retry at all** on its four POSTed SQL primitives (`archive_expired`, `archive_stale`, `purge_soft_deleted`, `upsert_tenant_suppression`) — and in the worker each connect failure is worse: it burns a Pub/Sub delivery attempt against the DLQ budget for a request that never reached storage.

## What

- **`common/http_retry.py` (new)** — the retry policy extracted from core-api's client (`with_retry`, `with_connect_phase_retry`, exception/status constants, incident history) so the two storage clients cannot silently diverge. core-api behavior is unchanged — its helpers now import the shared module (its retry tests pass unmodified).
- **core-worker** — `with_connect_phase_retry` wired into `_signed_call`, the single choke point every storage call already goes through (same drift-proofing argument as its auth signing). **Connection-phase only, for all methods**: the worker's designed retry path for sent-but-failed requests is raise → nack → Pub/Sub redelivery; an in-process ReadTimeout/5xx retry would double-layer that. Connection-phase failures are the exception because the request provably never reached storage.

## Verification

- `core-worker/tests/test_storage_client_retry.py` (new, mirrors the core-api retry tests through the public endpoint helpers): POST primitives retry ConnectTimeout/ConnectError and recover; give up after 3 attempts; ReadTimeout and 5xx propagate un-retried for the nack path; GET/PATCH ride the same choke point. → 18 passed (incl. existing `test_storage_client_auth.py` — the 401-rotation interleaving is unchanged).
- `pytest tests/test_storage_client_retry.py tests/test_audit_queue.py tests/test_storage_client_routing.py tests/test_embedding_backfill.py` (core-api side) → 53 passed.
- `ruff check` + `format --check` clean; `mypy core-worker/src` clean; core-api mypy delta vs main: none. (`test_consumer_enrich.py::test_enrichment_pending_cleared_even_with_heuristic_fallback` fails identically on clean main — pre-existing.)

## Enterprise vendoring note

No manifest obligation: `scripts/check_vendored_drift.py` in the enterprise repo only requires classification for `common/` files present in **both** trees. `common/http_retry.py` is OSS-only until enterprise chooses to vendor it (at which point it should be added as `identical`).

### DIFF
--- a/common/http_retry.py
+++ b/common/http_retry.py
@@ -1,0 +1,161 @@
+"""Transient-error retry policy for core-storage-api HTTP clients.
+
+Shared by core-api's ``CoreStorageClient`` and core-worker's storage
+client so the two cannot silently diverge on what is safe to retry.
+
+History (load-bearing for the policy split):
+
+* F5 — Cloud Run logs over 7 days showed a 31% silent-failure rate on
+  ``process_entity_extraction`` on ``staging-memclaw-core-api``. Every
+  failure traced to ``httpx.ConnectTimeout`` reaching core-storage-api
+  (cold starts / autoscaling). Retry landed for idempotent methods
+  (GET, PATCH, DELETE) only.
+* 2026-06-11 prod incident — ``find_similar_candidates`` (contradiction
+  detection, 42 failures) and ``create_audit_logs_bulk`` (audit events
+  dropped) both died on first-attempt ``ConnectTimeout`` behind the VPC
+  connector because POSTs had no retry at all. Connection-phase retry
+  for non-idempotent methods landed in response (caura-memclaw#333).
+
+Retry policy:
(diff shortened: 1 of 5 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #337, opened 2026-06-12
fix(cross-links): single multi-VALUES insert — executemany cannot RETURN rows

Eldad-Caura:
## Why

Prod 2026-06-12 06:31Z + 06:46Z (`prod-memclaw-core-api`, tenant `ct-a-synthetic-prod-77abc3`):

```
Error bulk-inserting 3 cross-links for tenant ct-a-synthetic-prod-77abc3
sqlalchemy.exc.ResourceClosedError: This result object does not return rows. It has been closed automatically.
```

#298 fixed the `RETURNING id` → real-columns issue but kept `execute(text_stmt, to_insert)` — passing a **params list** flips SQLAlchemy into the executemany path, where RETURNING rows are unavailable and `result.all()` raises. Every **multi-row** batch fails (step → FAILED, links lost); the 1-row batches seen before now happened to be the only shape exercised, which is why this stayed latent since 06-08.

## What

One `pg_insert(MemoryEntityLink).values(rows)` multi-VALUES statement (the CAURA-686 pattern from core-storage's memories router) + `.on_conflict_do_nothing(index_elements=["memory_id","entity_id"]).returning(memory_id, entity_id)` — RETURNING works for any row count, conflicts still excluded from the count. Bounds: worst case `CROSS_LINK_MEMORY_BATCH_SIZE (200) × LIMIT 10 = 2,000` inlined rows — routine for Postgres. A repo sweep found no other executemany+RETURNING sites.

## Verification

- `tests/pipeline/test_entity_linking.py` — the shape-pinning test now also asserts the executemany trigger is absent (exactly one positional `execute` arg) and that a 3-row batch compiles to a **single** statement carrying all rows + `ON CONFLICT ... DO NOTHING` + real-columns RETURNING. 10 passed (incl. wiring tests).
- `ruff check` + `format --check` clean; mypy clean on the touched file. DCO signed.

### DIFF
--- a/core-api/src/core_api/pipeline/steps/entity_linking/discover_cross_links.py
+++ b/core-api/src/core_api/pipeline/steps/entity_linking/discover_cross_links.py
@@ -5,7 +5,9 @@ from __future__ import annotations
 import logging
 
 from sqlalchemy import text
+from sqlalchemy.dialects.postgresql import insert as pg_insert
 
+from common.models.entity import MemoryEntityLink
 from core_api.constants import (
     CROSS_LINK_MEMORY_BATCH_SIZE,
     CROSS_LINK_SIMILARITY_THRESHOLD,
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #338, opened 2026-06-12
fix(mcp): memclaw_tune crashed with 'dict' object has no attribute 'id'

ran-taig:
## Summary

`memclaw_tune` raises `'dict' object has no attribute 'id'` on **any** call that actually sets a tuning parameter (e.g. `top_k=5`, `min_similarity=0.5`). A no-argument call succeeds, which is why the failure wasn't obvious.

## Root cause

`get_or_create_agent()` returns a **dict**. The handler reads it dict-safely one line before the bug, then reverts to attribute access:

```python
agent = await get_or_create_agent(db, tenant_id, agent_id)
current = (agent.get("search_profile") if isinstance(agent, dict) else agent.search_profile) or {}  # dict-safe
if updates:                                   # only true when a param was supplied
    ...
    await agent_repo.update_search_profile(db, agent.id, current)  # agent.id on a dict -> AttributeError
```

`agent.id` is attribute access on a dict → `AttributeError`. The crash lives **inside** `if updates:`, so a no-arg `memclaw_tune` skips it entirely and returns fine — which is also why the existing tests (registration, scope/auth guards) never caught it: none exercise the `update_search_profile` write branch.

## Fix

Resolve the primary key dict-safely, mirroring `routes/agents.py:141` (which already does `agent["id"]`) and the `isinstance` guard used two lines up:

```python
agent_pk = agent["id"] if isinstance(agent, dict) else agent.id
await agent_repo.update_search_profile(db, agent_pk, current)
```

`update_search_profile` expects the `Agent.id` primary key; `agent["id"]` is confirmed present on the returned dict (same field the REST tune route uses).

## Verification

- `python -m py_compile` on the module passes.
- Reproduced originally via direct MCP call on a live instance: `memclaw_tune {"agent_id":"code-reviewer","top_k":5}` returned the `'dict' object has no attribute 'id'` error, while `{"agent_id":"code-reviewer"}` (no params) succeeded — exactly matching the `if updates:` analysis above.

## Scope

Minimal one-line-of-logic fix. Worth a follow-up: a regression test covering the `memclaw_tune` write branch (mock `get_or_create_agent` → dict, assert `update_search_profile` is called with the PK), since that path is currently untested.

### DIFF
--- a/core-api/src/core_api/mcp_server.py
+++ b/core-api/src/core_api/mcp_server.py
@@ -1197,13 +1197,13 @@ async def memclaw_tune(
             from core_api.services.agent_service import get_or_create_agent
 
             agent = await get_or_create_agent(db, tenant_id, agent_id)
-            current = (agent.get("search_profile") if isinstance(agent, dict) else agent.search_profile) or {}
+            current = agent.get("search_profile") or {}
             if updates:
                 current.update(updates)
                 from core_api.services.organization_settings import validate_search_profile
 
                 current = validate_search_profile(current)
-                await agent_repo.update_search_profile(db, agent.id, current)
+                await agent_repo.update_search_profile(db, agent["id"], current)
                 await db.commit()
             return _with_latency(json.dumps({"agent_id": agent_id, "search_profile": current}, indent=2), t0)
         except HTTPException as e:


ran-taig:
@claude

ran-taig:
@claude

ran-taig:
@erni-a

erni-a (approved):

PR #339, opened 2026-06-12
fix(api): default agent_id to "mcp-agent" on standalone POST /memories

ran-taig:
## Summary

The documented quickstart — `POST /api/v1/memories` with `{tenant_id, content}` — returns `422 {"loc":["body","agent_id"],"msg":"Field required"}`. It's the reader's first hands-on API call **and** the Step-1 pipeline smoke test ("send one raw sentence, watch it get enriched"), so the failure cascades: nothing is written, and the follow-up search demo returns empty.

Root cause: `MemoryCreate.agent_id` was required with no default, while the curl (and `README.md:175`) omits it. The MCP `memclaw_write` tool doesn't hit this because its tool param defaults `agent_id="mcp-agent"`.

## Change

Make `agent_id` optional (`str | None = None`, matching the existing `BulkMemoryCreate.agent_id`) and resolve a missing value in `write_memory`:

- **Standalone** (`settings.is_standalone`) → default to the reserved `"mcp-agent"` identity. Mirrors `mcp_server._DEFAULT_AGENT_ID` and the `evolve`/`insights` REST routes. This is what makes the quickstart work.
- **Tenant-scoped key / enterprise gateway** → still require an explicit `agent_id` (now a clearer 422 instead of the raw Pydantic one).

## Why gate on `is_standalone` (the safety-critical part)

A naive "just default the schema to `mcp-agent`" would regress a deliberate protection. The MCP layer **refuses** `"mcp-agent"` when a tenant-scoped key comes through the gateway (`_refuse_default_agent_on_gateway`), because otherwise every anonymous write collapses onto one shared identity (a documented friction-report footgun: "agent row missing from list_agents"). The REST write route has **no** such guard, and its only backstop — `trust_level == 0 → 403` — doesn't fire by default (`require_agent_approval` defaults to `False`, `DEFAULT_TRUST_LEVEL = 1`), so a defaulted `mcp-agent` would auto-register and write successfully.

Gating the default on `is_standalone` keeps the gateway/managed path requiring a real agent_id, preserving that boundary.

## Impact analysis (checked, no regressions)

- **Internal `MemoryCreate(...)` constructors** — crystallizer (`"crystallizer"`), evolve (`agent_id=`), stm (`agent_id=`) all pass it explicitly. Default never applies.
- **OpenClaw plugin** — its REST `POST /memories` (`plugin/src/context-engine.ts:518`) passes `agent_id: "__health_check__"`. Unaffected.
- **Tenant-scoped / gateway REST callers** — behavior unchanged: still rejected when omitting agent_id (clearer message).
- **Tests** — no existing test asserts a 422 specifically on missing `agent_id`; no contract broken.

## Verification

- `py_compile` clean on both files.
- Pydantic check: missing `agent_id` → `None` (route then resolves); explicit `agent_id` → preserved.
- `HTTPException` and `app_settings` already imported in the route.

## Scope

Single-write path only. `BulkMemoryCreate` already allowed `agent_id=None` (and passes it straight through without this standalone gate); left as-is to keep the change focused — worth a follow-up to give bulk the same guard.

### DIFF
--- a/core-api/src/core_api/agent_ids.py
+++ b/core-api/src/core_api/agent_ids.py
@@ -1,0 +1,13 @@
+"""Shared agent-id constants.
+
+A tiny, dependency-free module so both the MCP server (``mcp_server``) and the
+REST routes (``routes.memories``) can reference the reserved default identity
+without importing each other — avoids the heavy/cross-private import that would
+otherwise couple the route module to the whole MCP tool surface.
+"""
+
+# Reserved fallback identity used when a caller omits ``agent_id`` on the
+# single-tenant / standalone path. On the enterprise gateway this value is
+# explicitly refused (see ``mcp_server._refuse_default_agent_on_gateway``) so
+# anonymous writes are never silently attributed to one shared identity.
+DEFAULT_AGENT_ID = "mcp-agent"
--- a/core-api/src/core_api/mcp_server.py
+++ b/core-api/src/core_api/mcp_server.py
@@ -26,6 +26,7 @@ from sqlalchemy import text as sa_text
 from starlette.types import ASGIApp, Receive, Scope, Send
 
 from common.enrichment.constants import SERVER_RESERVED_MEMORY_TYPES
+from core_api.agent_ids import DEFAULT_AGENT_ID
 from core_api.auth import get_admin_key
 from core_api.clients.storage_client import KeystoneUpsertPayload, get_storage_client
 from core_api.constants import (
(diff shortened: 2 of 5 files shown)


ran-taig:
@claude

ran-taig:
@claude

ran-taig:
@claude

ran-taig:
@claude

ran-taig:
@claude

ran-taig:
@claude

ran-taig:
@claude

ran-taig:
@claude

ran-taig:
@Eldad-Caura

Eldad-Caura (approved):

PR #341, opened 2026-06-13
fix(entity-linking): bind reinforce weight directly — LEAST over untyped params coerced to text

Eldad-Caura:
## Why

Prod 2026-06-13: **every** relation-reinforce batch in `infer_relations` fails with

```
asyncpg.exceptions.DatatypeMismatchError: column "weight" is of type double precision but expression is of type text
  at core-api/.../entity_linking/infer_relations.py:137
```

**45 occurrences in ~9h across prod (26) + staging (19)**, multiple tenants. Postgres resolves `LEAST(:new_weight, :max_weight)` over two **untyped** bind params as `text`, then rejects assigning that to the `double precision` `weight` column. The step returns FAILED, so co-occurrence relations are **silently never reinforced**. (Surfaced from my staging cutover-soak scan; it's the same class in the prod alert list — distinct from the connect-timeout residual.)

## What

`new_weight` is **already** clamped to `MAX_RELATION_WEIGHT` in Python (`min(current_weight + cooccur*DELTA, MAX_RELATION_WEIGHT)`), so the SQL-side `LEAST` was redundant double-clamping — and the source of the type ambiguity. Bind `:new_weight` directly (`SET weight = :new_weight`), exactly like the sibling INSERT path binds `:weight` (which is why INSERT was unaffected — column context infers the type). Drop the now-unused `max_weight` key.

Likely surfaced post-Phase-J: the re-embed created many more entity links/relations, so the reinforce path (UPDATE existing) started firing where before it rarely did.

## Verification

- New `tests/pipeline/test_infer_relations.py` (mock-based → asserts statement SHAPE + the Python clamp the fix relies on): reinforce UPDATE binds `:new_weight` with **no `LEAST`**; weight stays clamped to MAX in Python; INSERT path unchanged; empty co-occurrence skips. **Verified red against the unfixed SQL.**
- `ruff check` + `format --check` clean; `mypy` clean. DCO signed.

### DIFF
--- a/core-api/src/core_api/pipeline/steps/entity_linking/infer_relations.py
+++ b/core-api/src/core_api/pipeline/steps/entity_linking/infer_relations.py
@@ -114,8 +114,10 @@ class InferRelations:
                 reinforce_batch.append(
                     {
                         "rel_id": rel_id,
+                        # Already clamped to MAX_RELATION_WEIGHT above — bound
+                        # directly below (no SQL-side LEAST), matching the
+                        # INSERT path's ``:weight``.
                         "new_weight": new_weight,
-                        "max_weight": MAX_RELATION_WEIGHT,
                         "tenant_id": tenant_id,
                     }
                 )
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #342, opened 2026-06-13
fix(entity-linking): synchronize_session=False on the entity-embedding backfill bulk update

Eldad-Caura:
## Why

Prod 2026-06-13: **every** `backfill_entity_embeddings` batch fails with

```
sqlalchemy.exc.InvalidRequestError: bulk synchronize of persistent objects not supported
when using bulk update with additional WHERE criteria right now.
add synchronize_session=None execution option to bypass synchronize of persistent objects.
  at core-api/.../entity_linking/backfill_entity_embeddings.py:68
```

6 occurrences in ~10h. The step runs an ORM `update(Entity).where(id==:eid, tenant_id==…).values(name_embedding=:emb)` over an **executemany param list**; SQLAlchemy refuses to session-synchronize that shape. Result: entities with a NULL `name_embedding` are **silently never backfilled**, degrading entity-name vector matching. (Surfaced from the same full prod-error sweep that found infer_relations / #341.)

## What

Add `.execution_options(synchronize_session=False)` — there's no live ORM session state to keep in sync (the backfill writes the embedding and moves on), so skipping synchronization is correct, exactly as the error message instructs.

**Third `entity_linking_full` sibling to surface post-Phase-J** (the re-embed created the entities/links that make these steps fire): `discover_cross_links` (#337, executemany+RETURNING), `infer_relations` (#341, LEAST over untyped binds), and now this. A repo-wide sweep found **no other** `update().where().values()` over a param list with this latent bug.

## Verification

- New `tests/pipeline/test_backfill_entity_embeddings.py`: asserts the UPDATE statement carries `synchronize_session=False` (mock-based shape pin, **verified red** against the unfixed step) + the empty-rows skip path.
- `ruff check` + `format --check` clean; `mypy` clean. DCO signed.

### DIFF
--- a/core-api/src/core_api/pipeline/steps/entity_linking/backfill_entity_embeddings.py
+++ b/core-api/src/core_api/pipeline/steps/entity_linking/backfill_entity_embeddings.py
@@ -66,9 +66,16 @@ class BackfillEntityEmbeddings:
         backfill_count = 0
         if updates:
             await ctx.require_db.execute(
+                # ``synchronize_session=False``: this is a bulk ORM UPDATE with
+                # extra WHERE criteria over an executemany param list, which
+                # SQLAlchemy refuses to session-synchronize (InvalidRequestError,
+                # prod 2026-06-13). There is no live ORM session state to keep in
+                # sync here — the backfill writes name_embedding and moves on — so
+                # skipping synchronization is correct, not just a silencer.
                 update(Entity)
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #344, opened 2026-06-13
docs: fix README/AGENT-INSTALL inaccuracies (PII, health shape, write response)

ran-taig:
## Summary

Three doc claims that don't match the running OSS stack. All verified against a live instance.

### 1. PII is detected & tagged, not "quarantined"

`README.md` (Governance) claimed:
> PII auto-detected **and quarantined before it can cross fleet boundaries**

The code only sets `metadata.contains_pii` / `pii_types` on write (see `merge_enrichment_fields.py`). Nothing changes a memory's `status`, `visibility`, or content, and it stays `scope_team` and fully recallable. Reworded to "flagged on every write (surfaced in memory metadata as `contains_pii`/`pii_types`)".

### 2. `/api/v1/health` key is `storage`, not `database`

Docs showed `{"status":"ok","database":"connected",...}`. Actual response:
```json
{"status":"ok","storage":"connected","redis":"connected","event_bus":"ok"}
```
Fixed in `README.md` and both `AGENT-INSTALL.md` verify steps (3 spots).

### 3. Write response: `weight`, and async on fast-write

`README.md` said the write response includes `importance_score`. The field is **`weight`** (the importance score); there is no `importance_score`. Also, on the **default fast-write path** the immediate response is `enrichment_pending` and the inferred `title`/`summary`/`tags` populate asynchronously — the previous wording implied a synchronous, fully-enriched response.

## Scope

Docs-only; no code changes. Kept separate from the MCP-setup docs PR (#343) since these are unrelated accuracy fixes.

(Three further tutorial-only items — the `memclaw_evolve` `scope="fleet"` nuance for cross-agent reinforcement, the "9 tools" undercount vs the actual 12, and the simplified architecture diagram — live only in the external tutorial draft and aren't repo-doc bugs.)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #345, opened 2026-06-14
docs: comparison table — "PII detection & flagging", not "quarantine"

ran-taig:
## Summary

Follow-up to #344. That PR fixed the Governance bullet (PII *"auto-detected and flagged"*) but missed the **capability comparison table**, which still advertised **`PII detection & quarantine ✅`** for MemClaw.

The OSS engine only **detects and flags** PII — `merge_enrichment_fields.py` sets `metadata.contains_pii` / `pii_types` and nothing else (no status change, no visibility downgrade, no redaction, no blocking). Quarantine / boundary enforcement is an enterprise-gateway capability, not OSS.

Relabel the row to `PII detection & flagging` so the table matches both the OSS behavior and the corrected bullet.

Docs-only.


ran-taig:
@erni-a

erni-a (approved):

PR #346, opened 2026-06-14
fix(compose): stop shadowing .env provider keys with empty "${VAR:-}" passthroughs

ran-taig:
## Summary

`core-api` already loads `.env` through `env_file`, but the `environment:` block re-declared the provider/embedding vars as `"${VAR:-}"`. Docker Compose resolves `${...}` from the operator's **shell** with precedence over `.env`, which causes two real failures:

1. **Empty-string injection.** When a var is set only in `.env`, the `${VAR:-}` passthrough injects an empty string that *overrides* it. This is the class behind the `OPENAI_EMBEDDING_MODEL` → `400 "you must provide a model parameter"` break (the code's `os.environ.get(...) or default` is defeated because the key is present-but-empty). #336 hardened the model read, but the same trap applied to every empty-default passthrough.
2. **Stale-shell shadow.** A value exported in the operator's shell silently shadows the value they just set in `.env`. Observed in the field: a dead shell `OPENAI_API_KEY` beat a funded `.env` key → `429 insufficient_quota` on every embedding/enrichment call, with a `curl` of the `.env` key succeeding — baffling until you inspect the container env.

## Change

Drop the empty-default passthroughs and let `env_file` (`.env`) supply them directly — `env_file` values are **not** shell-shadowed:

- removed: `OPENAI_API_KEY`, `ANTHROPIC_API_KEY`, `OPENROUTER_API_KEY`, `OPENAI_EMBEDDING_BASE_URL`, `OPENAI_EMBEDDING_MODEL`, `OPENAI_EMBEDDING_TRUNCATE_TO_DIM`, `EMBEDDING_QUERY_INSTRUCTION`
- **kept**: `POSTGRES_HOST` / `CORE_STORAGE_API_URL` / `REDIS_URL` (literal Docker-network overrides), and `OPENAI_EMBEDDING_SEND_DIMENSIONS` — it needs a non-empty default (`true`) that the code does **not** supply when the var is absent (`common/embedding/_platform.py:89` → `false` if unset). Removing it would flip the hosted-OpenAI default.

## Verification

- `env.dev` sets none of the removed vars → `.env` is the sole source, so nothing is lost.
- `docker compose config` resolves cleanly.

## Known follow-up
`SEND_DIMENSIONS` still has the same shell-shadow for local-embedder users who set it to `false` in `.env`. Fixing that cleanly means giving the code a `true` default when unset and then dropping this last passthrough too — left out here to keep the change scoped and low-risk.

This is the top remaining item (F-2) from the self-host dry-run audit.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -111,21 +111,21 @@ services:
       CORE_STORAGE_API_URL: http://core-storage-api:8002
       REDIS_URL: redis://redis:6379/0
       # Provider keys and embedding config (OPENAI_API_KEY, ANTHROPIC_API_KEY,
-      # OPENROUTER_API_KEY, OPENAI_EMBEDDING_BASE_URL / _MODEL / _TRUNCATE_TO_DIM,
-      # EMBEDDING_QUERY_INSTRUCTION) come from ``env_file`` (.env) directly — they
-      # are intentionally NOT re-declared here. A "${VAR:-}" passthrough would
-      # (a) inject an empty string when the var is set only in .env, defeating
-      # the code's own defaults (the OPENAI_EMBEDDING_MODEL → 400 break), and
-      # (b) let a stale value exported in the operator's shell SHADOW the value
-      # they just set in .env, since Compose gives the shell precedence over
+      # OPENROUTER_API_KEY, OPENAI_EMBEDDING_BASE_URL / _MODEL / _SEND_DIMENSIONS /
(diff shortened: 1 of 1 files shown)


ran-taig:
@claude

ran-taig:
@claude

ran-taig:
@arkash20

ran-taig:
@Eldad-Caura

Eldad-Caura (approved):

PR #349, opened 2026-06-14
docs: add top-level BENCHMARKS.md as a citable benchmark asset

arkash20:
## What

Adds a root-level **`BENCHMARKS.md`** — the LoCoMo / LongMemEval results table, what we measure and how, the fleet-shape dimensions single-agent benchmarks can't capture, and a step-by-step reproduction recipe against the **public** datasets + the OSS local stack.

- Cross-links it from the README **Performance** section.
- Repoints `docs/performance.md` at `BENCHMARKS.md` for methodology/reproduction (it previously pointed readers to the **private**, different-purpose `keystone-benchmark` repo — that's a policy-enforcement benchmark, not the memory benchmark these numbers come from, so the old link would 404 for the public).

## Why

From the OSS SEO/GEO review: a reproducible, in-repo benchmark is our most citation-worthy asset, but it was only reachable via a blog link and a buried `docs/performance.md`. `BENCHMARKS.md` at the repo root is the conventional, citable location evaluators and AI engines look for.

## Notes

- Numbers are unchanged — reused verbatim from `docs/performance.md` / the blog (point-in-time, last run 2026-04-19; blog remains canonical).
- No turnkey runner is bundled (datasets are large + public; the internal runner isn't open yet). The doc is explicit about that and describes the exact methodology instead — chosen over shipping an unverifiable script.

### DIFF
--- a/BENCHMARKS.md
+++ b/BENCHMARKS.md
@@ -1,0 +1,98 @@
+# MemClaw Benchmarks
+
+How MemClaw performs on the two most-cited public agent-memory benchmarks —
+**LoCoMo** and **LongMemEval** — plus the fleet-shaped dimensions those
+single-agent benchmarks can't measure.
+
+> **TL;DR** — On accuracy, MemClaw sits inside the leading cluster (MemClaw,
+> Mem0, Zep land in a narrow band). Where we push hardest, and where it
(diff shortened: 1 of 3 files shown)


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #351, opened 2026-06-14
fix(fleet): cap auto-upgrade re-queue with a per-(node,target) attempt budget (CAURA-000)

ran-taig:
## Summary

Implements **L1** from the auto-upgrade re-queue bug investigation: a backend-side attempt budget that stops the gate from re-queuing deploys indefinitely against a node that never converges to the target version.

Today `_maybe_queue_auto_upgrade` has no attempt budget. Its only brakes are the 10-min in-flight window (stops *concurrent* storms only), the plugin-signalled `deploy_blocked_until` cooldown, and the hardcoded `KNOWN_BROKEN` set. A node that never converges is re-queued **every heartbeat (~60s ≈ 1,440/day)** with no brake and no operator-visible signal — for any failure mode the plugin can't self-detect.

## Validated failure modes (all un-braked today)

| Mode | Plugin writes cooldown? |
|---|---|
| Persistent `/plugin-source` fetch error (`fetchOk=false`) | **No** — `status=failed`, no cooldown |
| Unsafe-filename manifest abort | **No** |
| **"Succeeds but never advances"** — served manifest version < gate's `MIN_RECOMMENDED` | **No** — every deploy builds, restarts, post-restart-verifies OK (clears cooldown), yet backend still sees `plugin_version < target` → re-queues forever, each attempt reporting *success* |

The plugin's 24h cooldown only covers its *own* self-detected failures (`build-failed`, `post-restart-version-mismatch`). `command_result` records failed deploys but feeds nothing back into the gate. Confirmed: no per-(node,target) counter anywhere.

## Fix (backend-only)

After `AUTO_UPGRADE_MAX_ATTEMPTS` (5) deploys for the same `target_version` within `AUTO_UPGRADE_ATTEMPT_WINDOW` (24h), the gate stops queuing and logs a `warning` for operator follow-up. Brakes **all three** modes at the source — independent of plugin self-detection.

- **New repo method** `count_recent_deploys_for_target(node_id, target_version, since)` — counts ALL statuses (the nastiest mode is `status=done` with no version progress; a status filter would miss it). JSONB `payload->>'target_version'` filter excludes manual/non-auto deploys.
- **New gate branch** placed *after* the in-flight check (rarest path) so the count query only runs on outdated-eligible-and-not-in-flight heartbeats. **Fail-open on DB error**, matching the in-flight check.

### Why it's invisible to healthy upgrades
A legitimate upgrade converges in **one** attempt — the next heartbeat reports the new version and the gate exits at the `_semver_lt` check *before* the budget is ever consulted. Keying on `target_version` means a **new release (bumped target) gets a fresh budget** automatically; only a non-advancing node accumulates >1 row for a given target.

## Performance / load (asked during review)

- **Agent turn path:** untouched — zero latency impact.
- **Steady-state (converged fleet):** zero — the gate short-circuits at `_semver_lt` before any DB work.
- **Wedged node:** large load **reduction** — ~1,440 command-inserts + restart cycles/day collapse to 5.

The new query is one indexed SELECT, only on outdated-eligible heartbeats; `has_recent_in_flight_deploy` is left untouched (lower blast radius than folding the two into one query).

## Production context

Validated read-only on the customer prod box: the gap is **currently latent** there — `MIN_RECOMMENDED` == served manifest == `2.9.0`, **0 deploy churn in 48h**, fleet converged. This lands as **preventive hardening** before the next release bump, where a manifest/target drift on an on-prem image would re-activate the loop. (The earlier 1,381-stuck-deploy incident was the pre-#305/#306 ack-storm — already resolved, not recurring.)

## Tests

5 new unit tests mirroring the in-flight gate tests:
- skip-when-exhausted (+ asserts budget is keyed on target_version)
- allow-under-cap
- fresh-target-gets-fresh-budget
- fail-open-on-repo-error
- constants-are-sane

56/56 in `test_fleet_heartbeat_auto_upgrade.py`; **126/126** across the fleet/heartbeat/repository sweep. `ruff` clean. JSONB query compile-verified to `payload ->> 'target_version'`.

## Scope

- **L1 only**, per the agreed gradual plan. L2 (full `nodes.metadata` stuck-stamp + UI), L3 (version-consistency guard), L4 (plugin-side cooldown on un-braked exits) are deferred. The exhaustion `logger.warning` here delivers the cheap ~80% of L2's observability.
- No plugin change, no version bump, no migration.

## Test plan
- [x] 56/56 auto-upgrade unit tests
- [x] 126/126 fleet/heartbeat/repository sweep
- [x] JSONB query compiles to correct SQL
- [x] DCO sign-off

### DIFF
--- a/common/models/fleet.py
+++ b/common/models/fleet.py
@@ -1,7 +1,7 @@
 import uuid
 from datetime import datetime
 
-from sqlalchemy import DateTime, ForeignKey, Text, UniqueConstraint, text
+from sqlalchemy import DateTime, ForeignKey, Index, Text, UniqueConstraint, text
 from sqlalchemy.dialects.postgresql import JSONB, UUID
 from sqlalchemy.orm import Mapped, mapped_column
 
@@ -40,13 +40,34 @@ class FleetNode(Base):
 
 class FleetCommand(Base):
     __tablename__ = "fleet_commands"
+    __table_args__ = (
+        # Backs the two auto-upgrade gate queries in
+        # ``routes/fleet.py`` — ``has_recent_in_flight_deploy`` and
+        # ``count_recent_deploys_for_target`` — both of which filter
+        # ``node_id = ? AND command = 'deploy' AND created_at >= ?``.
+        # Partial on ``command = 'deploy'`` keeps the index small (deploy
+        # is a tiny fraction of all fleet commands) and lets Postgres
+        # seek straight to a node's deploy rows; the JSONB
+        # ``payload->>'target_version'`` predicate is then applied on the
+        # O(1)–O(10) rows that survive, so it doesn't need to be in the
+        # index. Bare column names (not ``.desc()``) so Alembic autogen
+        # can reflect/compare — matches the style in ``memory.py``.
+        Index(
+            "ix_fleet_commands_node_created_deploy",
+            "node_id",
+            "created_at",
+            postgresql_where=text("command = 'deploy'"),
+        ),
+    )
 
     id: Mapped[uuid.UUID] = mapped_column(
         primary_key=True, server_default=text("gen_random_uuid()")
     )
     tenant_id: Mapped[str] = mapped_column(Text, nullable=False)
     node_id: Mapped[uuid.UUID] = mapped_column(
-        UUID(as_uuid=True), ForeignKey("fleet_nodes.id", ondelete="CASCADE"), nullable=False
+        UUID(as_uuid=True),
+        ForeignKey("fleet_nodes.id", ondelete="CASCADE"),
+        nullable=False,
     )
     command: Mapped[str] = mapped_column(Text, nullable=False)
     payload: Mapped[dict | None] = mapped_column(JSONB)
--- a/core-api/src/core_api/repositories/fleet_repository.py
+++ b/core-api/src/core_api/repositories/fleet_repository.py
@@ -182,6 +182,54 @@ class FleetRepository:
         )
         return result.scalar_one_or_none() is not None
 
+    async def count_recent_deploys_for_target(
+        self,
+        db: AsyncSession,
+        *,
+        node_id: UUID,
+        target_version: str,
+        since: datetime,
+    ) -> int:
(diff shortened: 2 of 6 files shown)


ran-taig:
@claude

ran-taig:
Addressed the index review — valid, fixed.

**Added** a partial btree index `ix_fleet_commands_node_created_deploy ON fleet_commands (node_id, created_at) WHERE command = 'deploy'`, in **both** places this codebase manages schema:
- **Model** — `FleetCommand.__table_args__` (for `metadata.create_all` / fresh installs + alembic autogen reflection), mirroring `FleetNode`'s existing `__table_args__` and the partial-index style in `common/models/memory.py`.
- **Migration** — alembic `023_fleet_commands_autoupgrade_index.py`, `CREATE INDEX CONCURRENTLY` inside an `autocommit_block()`, mirroring `017`'s pattern (the established convention for concurrent index builds: 005/007/011/016/017).

Two refinements vs. the suggestion:
1. **Index name** is `ix_fleet_commands_node_created_deploy` (column-descriptive, matching the repo's `ix_memories_tenant_created_active` convention) rather than `..._autoupgrade_count`, because it backs **both** auto-upgrade gate queries — the new `count_recent_deploys_for_target` *and* the pre-existing `has_recent_in_flight_deploy`, which also seq-scanned (same `node_id + command='deploy' + created_at` filter). One index, both consumers.
2. As you noted, the JSONB `payload->>'target_version'` predicate is intentionally **not** in the index — the (node_id, created_at, partial) seek reduces the candidate set to O(1)-O(10) rows and the JSONB extract filters those cheaply.

Verified: model-generated DDL matches the migration byte-for-byte; alembic chain is single-head (023→022); 126/126 fleet/heartbeat/repository + 30/30 migration/preflight tests pass; ruff clean.

ran-taig:
@claude

ran-taig:
@claude

ran-taig:
@erni-a

erni-a (approved):

PR #353, opened 2026-06-14
feat(core-api): capability-usage adoption counters (MCP + REST)

erni-a:
## What

Adds a lightweight **capability-usage** signal: which product capabilities (recall, write, search, doc, keystones, …) get used, broken down by access path (MCP vs REST) and attributable per org, persisted to a queryable table.

### How

- **`capability_usage` table** (migration `023` + `common.models.CapabilityUsage`) — append-only aggregate counters. No unique constraint and no upsert: each instance appends its own rows and consumers `SUM` at query time (contention-free). Deliberately cross-tenant analytics, so no RLS.
- **In-process aggregator** (`services/capability_usage.py`) folds each operation into per-`(tenant, capability, op, transport, minute-bucket)` counters and flushes on a short interval (default 15s; final flush on shutdown). The request hot path pays only a dict update — never a DB write. `record_usage()` is a no-op when unwired and never raises into a request.
- **MCP emitter**: `call_tool` is overridden on the FastMCP instance — the single dispatch point every `tools/call` routes through — so the tool name (capability) and `op` argument are captured without touching any handler's signature (generated input schemas are unchanged).
- **REST emitter**: `RequestObservationMiddleware` maps the matched route template → `(capability, op)` and records `transport=rest`. It also emits one structured `http.request` log line per request (route template, method, status, duration). Registered innermost so the route template is reliably readable.
- `get_auth_context` stashes the resolved `tenant_id` on `request.state` so emitters can attribute usage to an org.

Gated by `capability_usage_enabled` (default on). No new dependencies.

### Notes

- The middleware is registered **innermost** on purpose: `SlowAPIMiddleware` is a `BaseHTTPMiddleware`, which runs the downstream app in a separate task, so `scope["route"]` set by the router does not reliably propagate to a middleware outside it. Sitting innermost guarantees the template resolves. Trade-off: it does not observe 504/429 from the outer middlewares (those requests didn't execute a capability anyway).
- On Cloud Logging deployments the `http.request` events can also back a log-based metric; on file-logging deployments they land in the JSON log file. The capability counters are independent of log level.

### Testing

- Unit tests for the aggregator (folding, non-tenant filtering, drain, interval flush) and both emitters (`tests/test_capability_usage.py`, `tests/test_request_observation.py`).
- Verified end-to-end against a live stack: migration applies in the full chain; REST and MCP both record correct capability/op/transport/duration; the transport `CHECK` constraint and the adoption `GROUP BY capability, transport` query work against Postgres.


erni-a:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #354, opened 2026-06-14
fix(keystones): let the standalone operator author keystones out-of-the-box (F-7)

ran-taig:
## Summary (audit finding F-7)

On a fresh **standalone** install, authoring a keystone over REST 403s:
```
POST /api/v1/memclaw/keystones   (X-API-Key: standalone, no X-Agent-ID)
→ 403 "Agent 'rest-admin' is not registered. Register the agent by writing one memory first."
```
A bare standalone request has no caller agent identity, so `_resolve_caller_identity` falls back to the synthetic **`rest-admin`**, which has no agent row → `_require_trust` not_found → 403 with a misleading message. The flagship governance feature looks broken on first contact.

The trust gate is **correct and necessary for multi-tenant / gateway** deployments — it's the anti-spoof defense (`_effective_min_for_caller` floor-bump + X-Agent-ID mismatch rejection). But under `IS_STANDALONE` the API-key holder *is* the admin and there's nobody to impersonate, so the gate is pure friction.

## Change (REST keystone route only)
- **`_is_standalone_admin(auth, x_agent_id)`** — true **only** when `IS_STANDALONE` *and* the caller asserts no identity at all (no agent-scoped credential, no `X-Agent-ID`). The `upsert`/`delete` handlers skip the caller-trust gate for that caller; storage-side shape validation (scope/fleet_id/agent_id) still applies.
- **Reworded the not_found 403** to name the real cause: register the agent, or pass `X-Agent-ID` / an agent-scoped credential for an agent at trust ≥ 2.

## Why it's safe (validated against code + tests)
- **No-op off standalone** — the bypass requires `IS_STANDALONE` *and* the no-identity fallback, so the multi-tenant/gateway anti-spoof defenses are fully intact.
- **Doesn't disturb existing tests** — audited every keystone endpoint call in `test_api_keystones.py`: all authenticate via the `_author_headers` helper (`X-Agent-ID` set), so they hit the `x_agent_id` branch, never the `rest-admin` fallback the bypass keys on. `test_set_rejected_when_agent_unknown` (X-Agent-ID=`ghost`) still 403s.
- **No test pins the reworded string** — keystone tests assert only `status_code`; the `"is not registered"` assertions live on the MCP path + `trust_service`, neither touched.
- **MCP keystone path unaffected** — it resolves the caller to `mcp-agent`, not `rest-admin`.

## Tests
Adds `test_set_allowed_for_standalone_admin_without_agent_id` — bare standalone admin (no X-Agent-ID) can now author a keystone (200), locking the fix.

### DIFF
--- a/core-api/src/core_api/routes/keystones.py
+++ b/core-api/src/core_api/routes/keystones.py
@@ -32,6 +32,7 @@ from sqlalchemy.ext.asyncio import AsyncSession
 
 from core_api.auth import AuthContext, get_auth_context
 from core_api.clients.storage_client import KeystoneUpsertPayload, get_storage_client
+from core_api.config import settings as app_settings
 from core_api.db.session import get_db
 from core_api.services.audit_service import log_action
 from core_api.services.trust_service import parse_trust_error
@@ -76,6 +77,7 @@ async def _enforce_author_trust(
     agent_id: str,
     *,
     min_level: int,
+    standalone_admin: bool = False,
 ) -> None:
     """Block keystone writes / deletes from principals below ``min_level``.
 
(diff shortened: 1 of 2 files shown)


ran-taig:
@claude

ran-taig:
@claude

ran-taig:
@erni-a

erni-a (approved):

PR #357, opened 2026-06-14
fix(search): surface conflicted memories that exactly match the query

arkash20:
## Problem

S1 retrieval@10 collapses at corpus scale even after #304/#327: **22/25 @ K=1000 → 18/25 @ K=10000**, and per-failure anatomy showed **every residual miss was a gold memory with `status='conflicted'`** — flipping just those statuses to `confirmed` recovered **25/25**.

Two retrieval paths hard-excluded `conflicted`/`outdated` rows from default results:
- `memory_scored_search` (semantic/keyword path): `WHERE status NOT IN ('outdated','conflicted')`
- `memory_load_by_ids` (entity_lookup short-circuit): same exclusion

The semantic contradiction detector marks the older of two embedding-near rows `conflicted`, and routinely mismarks **distinct entities that merely share a name prefix** (e.g. `Wayne Enterprises #0000` was marked conflicted against `Wayne Enterprises #0704`). The blanket exclusion then silently dropped the exact row the caller asked for, leaving only confirmed near-duplicate siblings about *other* entities. ~10% of a 10k-fact corpus was marked conflicted this way.

## Fix (retrieval-only — write/contradiction path untouched)

- **`memory_scored_search`**: keep a `conflicted` row when it is an exact lexical (FTS) match for the query, and exempt that row from the `0.5` `status_penalty` so it isn't re-buried beneath confirmed siblings. `outdated` (a definitive retraction) stays fully excluded.
- **`memory_load_by_ids`** (entity_lookup): rows here are already scoped to the resolved entity — the entity-graph analogue of an exact match — so keep `conflicted` and exclude only `outdated`.

`load_and_serialize` still injects the supersedes successor for any returned superseded row, so genuine knowledge-updates continue to show both sides.

## Results (S1 replay, local stack, retrieval@10)

| | before | after |
|---|---|---|
| **k10k** | 18/25 | **25/25** |
| k1k | 22/25 | 23/25 |

The 2 residual k1k misses are person-hub graph dilution in entity_lookup (gold's link dropped at the `GRAPH_MAX_BOOSTED_MEMORIES` cap) — a separate, deeper issue (tracked toward the exact-first cascade, A30).

## Regression safety

- New integration tests (`tests/test_integration_search.py::TestConflictedExactMatchSurfaces`) pin all three semantics: conflicted exact-match surfaces; conflicted non-match stays excluded; outdated exact-match stays excluded. They use `fake_embedding` (no external dependency).
- `knowledge-update` regression category (the relevant risk — surfacing conflicted rows): **acc 91.7%, recall 1.0** vs baseline 92.9% / 1.0 — within noise, no regression. (The full 5-category local run was blocked by unrelated write-path LLM timeouts in the local env; the categories that completed writes matched baseline exactly.)

### DIFF
--- a/core-storage-api/src/core_storage_api/services/postgres_service.py
+++ b/core-storage-api/src/core_storage_api/services/postgres_service.py
@@ -1015,8 +1015,27 @@ class PostgresService:
         else:
             date_range_boost = literal_column("1.0").label("date_range_boost")
 
+        # Status demotion. ``outdated`` is always demoted (a definitively
+        # superseded fact). ``conflicted`` is demoted EXCEPT when the row is an
+        # exact lexical match for the query: a conflicted memory carries a
+        # competing claim, not a retraction, and when it is the exact thing the
+        # caller asked for it must not be buried beneath unrelated near-duplicate
+        # siblings (different entities that merely share a name prefix). The
+        # competing successor is still surfaced via load_and_serialize's
+        # supersedes-chain injection, so the caller sees both sides.
+        #
+        # Why an FTS match is the right gate: at corpus scale a conflicted row's
+        # blended relevance (vec+fts) sits only marginally above its near-dup
+        # siblings, so the 0.5 multiplier reliably sinks it below them even when
+        # it is the single best match — the exact-match signal is what
+        # distinguishes "the row the caller wants" from "a sibling about a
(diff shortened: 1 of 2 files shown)


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #362, opened 2026-06-14
build(core-api): commit uv.lock + install frozen in Docker (pin against dep drift)

Eldad-Caura:
## Why
The 2026-06-14 prod outage was dependency drift: `core-api`'s Docker build ran a **lock-less** `pip install './core-api[pubsub]'`, re-resolving on every build. The moment **FastAPI 0.137.0** released, it floated in — its `include_router` change drops prefixed paths from `app.routes`, tripping the CAURA-602 startup guard → core-api `exit(1)` → every staging deploy failed. `#359`'s `fastapi<0.137` cap is a reactive stopgap; this fixes the **root cause**.

## What (pilot — core-api only)
- **Commit `core-api/uv.lock`** (universal, multi-platform) and un-ignore just this one in `.gitignore`.
- **Dockerfile installs frozen** from the lock — `uv export --frozen … | uv pip install` instead of re-resolving the pyproject ranges. The deployed image is now reproducible and **cannot float** between builds; dependency bumps become deliberate, reviewable `uv lock` changes (Renovate/Dependabot-friendly).

## Verification (the historical objection, retired with evidence)
`uv.lock` was gitignored originally because mac-pinned locks broke Linux CI (the `greenlet<3.5.0` saga). Modern `uv` does **universal** locking, so that no longer holds — proven by building this exact Dockerfile on **`linux/amd64`**:
- frozen install resolved **101 packages, all wheels present** (no missing-Linux-wheel failure), `fastapi==0.136.3`
- `import core_api.app` inside the image → **green, 106 routes** (the guard passes; under the broken float this import is what crashed)

Note: OSS PR CI doesn't build the Docker image (only `publish-docker` on release does, and the deployed image is built by the enterprise pipeline), so this was verified by a local `linux/amd64` build of the **real** Dockerfile rather than in PR CI.

## Scope / not in this PR
- CI test-install and the `fastapi<0.137` cap are **unchanged** — the cap still guards CI's lock-less install; both can be revisited once the durable CAURA-602 guard fix lands and locks roll out.
- `core-worker` / `core-operations` / `core-storage-api` stay lock-less until this pilot is expanded to them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,7 +42,10 @@ Screenshot*.png
 .idea/
 reports/
 
-# Python package manager lockfiles — produced by uv, consumed only by
-# local dev (uv sync). Prod/CI Dockerfiles pip install from pyproject
-# directly, so committing these adds diff noise without a consumer.
+# Python package manager lockfiles. Ignored by default — but core-api commits
+# its lock (CAURA-000 pilot): its Dockerfile installs FROZEN from uv.lock so the
+# deployed image can't silently float to a new release (a lock-less pip install
+# floated to FastAPI 0.137 and crashed prod 2026-06-14). The other services stay
+# lock-less until the pilot is rolled out to them.
 **/uv.lock
+!core-api/uv.lock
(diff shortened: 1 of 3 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #397, opened 2026-06-15
feat(audit): tamper-evident per-tenant hash chain for audit_log (eToro governance)

Eldad-Caura:
## What & why

First of three coordinated PRs delivering the **eToro governance increment** (PII masking/drop/flag + business-vs-personal gate + admin UI, due 18 Jun). This PR makes the audit log **tamper-evident**: every audit event (and specifically every future PII mask/drop/flag and non-business drop) is bound into a per-tenant linked-hash chain, so silent mutation, deletion, or reordering is detectable.

This PR is **independent** and lands first — the governance gate (PR-B) and admin UI (PR-C) build on top of it.

## How it works

- **migration 025** adds nullable `seq` / `prev_hash` / `event_hash` to `audit_log` + an `audit_chain_head` serialization-point table. Columns are nullable on purpose: the chain **genesis is the migration boundary** — pre-existing append-only rows stay unchained (excluded by a partial unique index). We deliberately do **not** backfill, since hashing historical rows would launder unprovable history into a valid-looking chain.
- **`audit_chain.py`** holds the pure primitives — deterministic canonicalization (`sort_keys` JSON, UTC µs timestamp) and `event_hash = SHA256(canonical_event ‖ prev_hash)` — defined once so the writer and verifier agree byte-for-byte. Plus a scrub-before-hash `assert_pii_safe` guard (the audit log must never store the raw secret it audits).
- **chained insert** (`_audit_chain_one_tenant`) folds a batch under a per-tenant `audit_chain_head` `SELECT … FOR UPDATE` lock — same-tenant writers serialize, different tenants never block, and the concurrent-genesis race is handled with `ON CONFLICT DO NOTHING`. `audit_add` / `audit_add_batch` route through it, so single-event, sync-fallback and keystone audit writes all join the chain.
- **`GET /audit-logs/verify`** walks the chain in `seq` order, recomputes hashes, checks linkage + genesis + tail-vs-head, and returns the first broken link (`seq_gap` / `prev_hash_mismatch` / `event_hash_mismatch` / `tail_truncated`). Exposed to the enterprise governance UI via a storage-client method.

## Tests (`tests/test_audit_chain.py`, 15 cases)

Chain continuity across batches, **concurrent same-tenant serialization** (5×4 events → gap-free chain of 20), cross-tenant independence, and detection of **tampered rows / seq gaps / tail truncation**; PII-safe rejection rolls back without persisting. Unit tests cover canonicalization determinism + genesis. Migration verified to apply cleanly 024→025 against pgvector pg16; existing audit + keystone + lifecycle suites pass unchanged.

## Notes
- All DB logic stays in core-storage-api (Rule 14); schema change is via the internal Alembic flow only (Rule 11). ruff check + format + mypy clean on both projects.

### DIFF
--- a/common/models/__init__.py
+++ b/common/models/__init__.py
@@ -1,7 +1,7 @@
 # Shared SQLAlchemy models — used by core-api and core-storage-api.
 from common.models.agent import Agent
 from common.models.analysis_report import CrystallizationReport
-from common.models.audit import AuditLog
+from common.models.audit import AuditChainHead, AuditLog
 from common.models.background_task import BackgroundTaskLog
 from common.models.base import Base
 from common.models.capability_usage import CapabilityUsage
@@ -15,6 +15,7 @@ from common.models.memory import Memory
 
 __all__ = [
     "Agent",
+    "AuditChainHead",
     "AuditLog",
     "BackgroundTaskLog",
     "Base",
--- a/common/models/audit.py
+++ b/common/models/audit.py
@@ -1,7 +1,7 @@
 import uuid
 from datetime import datetime
 
-from sqlalchemy import DateTime, Text, text
+from sqlalchemy import BigInteger, DateTime, LargeBinary, Text, text
 from sqlalchemy.dialects.postgresql import JSONB
 from sqlalchemy.orm import Mapped, mapped_column
 
@@ -22,4 +22,36 @@ class AuditLog(Base):
     detail: Mapped[dict | None] = mapped_column(JSONB)
     created_at: Mapped[datetime] = mapped_column(
         DateTime(timezone=True), server_default=text("now()")
+    )
+    # ── Tamper-evident hash chain (migration 025) ────────────────────
+    # Per-tenant linked-hash chain. These are NULL on pre-migration rows
+    # (legacy append-only): the chain genesis is the migration boundary,
+    # so existing rows are intentionally left unchained and excluded from
+    # verification (a partial unique index covers ``seq IS NOT NULL`` only).
+    # Chained rows carry a monotonic per-tenant ``seq`` (starts at 1), the
+    # prior event's ``event_hash`` as ``prev_hash`` (a genesis sentinel for
+    # seq=1), and ``event_hash = SHA256(canonical_event || prev_hash)``.
+    seq: Mapped[int | None] = mapped_column(BigInteger)
+    prev_hash: Mapped[bytes | None] = mapped_column(LargeBinary)
+    event_hash: Mapped[bytes | None] = mapped_column(LargeBinary)
+
+
+class AuditChainHead(Base):
+    """Per-tenant serialization point for the audit hash chain.
+
+    One row per tenant — locked ``FOR UPDATE`` during a chained insert so
+    concurrent same-tenant batches serialize against each other without
+    taking a lock on the (large, append-only) ``audit_log`` table, and
+    different tenants never block one another. ``last_hash`` is the
+    ``event_hash`` of the most recent chained event (a genesis sentinel
+    when ``last_seq=0``); the next insert chains onto it.
+    """
+
+    __tablename__ = "audit_chain_head"
+
+    tenant_id: Mapped[str] = mapped_column(Text, primary_key=True)
+    last_seq: Mapped[int] = mapped_column(BigInteger, nullable=False, server_default=text("0"))
+    last_hash: Mapped[bytes] = mapped_column(LargeBinary, nullable=False)
+    updated_at: Mapped[datetime] = mapped_column(
+        DateTime(timezone=True), nullable=False, server_default=text("now()")
     )
--- a/core-api/src/core_api/clients/storage_client.py
+++ b/core-api/src/core_api/clients/storage_client.py
@@ -1246,6 +1246,21 @@ class CoreStorageClient:
             params["resource_type"] = resource_type
         return await self._get_list("/audit-logs", **params)
 
+    async def verify_audit_chain(self, tenant_id: str, limit: int = 100_000) -> dict:
+        """Verify a tenant's tamper-evident audit hash chain.
+
+        Returns ``{valid, verified_count, head_seq}`` (or ``first_broken``
+        on a detected break). Used by the enterprise governance UI's
+        "chain intact" check.
+        """
+        result = await self._get("/audit-logs/verify", tenant_id=tenant_id, limit=limit)
+        # Propagate failures: a None here means a network/5xx error. Returning
+        # {} would hand callers a dict with no "valid" key, turning the real
+        # error into a confusing KeyError downstream.
+        if result is None:
+            raise RuntimeError("verify_audit_chain: empty response from storage service")
+        return result
+
     # =====================================================================
     # Lifecycle audit (CAURA-655)
     # =====================================================================
--- a/core-storage-api/src/core_storage_api/database/migrations/versions/025_audit_hash_chain.py
+++ b/core-storage-api/src/core_storage_api/database/migrations/versions/025_audit_hash_chain.py
@@ -1,0 +1,68 @@
+"""Tamper-evident linked-hash chain for audit_log (eToro governance).
+
+Adds the per-tenant chain columns (``seq``, ``prev_hash``, ``event_hash``) to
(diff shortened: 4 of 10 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
Addressed all three. Thanks — the High and Low were spot-on; the Medium needed a slightly different fix than proposed.

**🔴 High — unbounded `/verify` limit:** fixed. `limit: int = Query(default=100_000, ge=1, le=500_000)` — bounded so a caller can't force an OOM.

**🟡 Medium — JSONB float normalisation:** real risk, fixed — but the proposed `json.loads(json.dumps(...))` is a **no-op for floats** (Python preserves `1.0` through a JSON round-trip), so it wouldn't have closed the gap. I empirically checked the actual round-trip on **pg16** (the image CI + staging use) and Postgres JSONB *preserves* `1.0` (returns it as `1.0`, not `1`) — so there's no mismatch on our deployed PG today. To make the tamper-evident hash robust across **any** PG version (AlloyDB in prod) and any future float-bearing audit detail, I added a `_normalise_numbers` helper that collapses integral floats → int (recursively, leaving bools and non-integral floats alone) inside `canonical_event`, which both the writer and verifier call — so they always agree. Locked with a unit test (`test_canonical_event_normalises_integral_floats`).

**🟢 Low — storage-client error masking:** fixed. `verify_audit_chain` now raises `RuntimeError` on a `None` response instead of `return result or {}`, so a network/5xx surfaces instead of a confusing `KeyError` on a missing `"valid"` key.

All audit-chain tests (16) + lint + mypy green. The hash-chain design points you validated (head-row lock serialization, genesis, tail-check) are unchanged.

Eldad-Caura:
@claude

Eldad-Caura:
Addressed both Mediums; the Low is a no-op here — details below.

**🟡 Medium — false `tail_truncated` under concurrent writes:** real TOCTOU, fixed. `audit_verify_chain` now runs `SET TRANSACTION ISOLATION LEVEL REPEATABLE READ` before its two reads (rows + head) so they share one snapshot. I verified the window empirically against pg16: under READ COMMITTED the second read sees a concurrent same-tenant insert (`c1=1 → c2=2`); under REPEATABLE READ it doesn't (`c1=1 = c2`). New regression test covers it.

**🟡 Medium — `_CARD_SHAPE` over-broad:** fixed. Tightened from "any 13–19 digit run" to separated 4‑4‑4‑4 (`\b\d{4}[ -]\d{4}[ -]\d{4}\d{4}\b`). This is a defense-in-depth backstop where a hit rolls back the whole batch (→ chain gap), so precision matters more than recall — benign 16-digit order IDs / txn refs / phone numbers in audit detail no longer trip it, while a clearly card-shaped value still does. The authoritative detector is the deterministic `common.governance` library on the write path. Test added (`test_assert_pii_safe_allows_benign_numeric_ids`).

**🟢 Low — full ORM load in verify:** respectfully skipping — it's a no-op here. The premise ("`detail` not needed") is inverted: `detail` **is** a hash input (`canonical_event` hashes it), so it must be loaded; and `AuditLog` has *exactly* the 11 columns the verifier uses (id, tenant_id, agent_id, action, resource_type, resource_id, detail, created_at, seq, prev_hash, event_hash) — the suggested column-select lists the same 11, so it loads identical data and saves no memory. The earlier `limit` bound (≤500k) already caps the row count.

ruff + mypy + 17 audit-chain tests green.

Eldad-Caura:
@claude

Eldad-Caura:
Addressed in `f0ba82a` (amend + force-with-lease — branch stays at one commit). 20 chain tests pass (17 + 3 new); ruff + mypy clean on `core-storage-api`.

**🔴 High — CPU verify loop blocks the async event loop** (`postgres_service.py` `audit_verify_chain`): Fixed. Extracted the walk into a pure module-level `_verify_audit_chain_rows(tenant_id, rows, head, limit)` and call it via `asyncio.to_thread(...)`. The DB reads stay awaited; only the SHA-256 + JSON-canonicalization loop (up to the 500k-row cap) moves off the loop. The rows/head are fully materialized before the thread runs and the session is already closed, so the thread only touches in-memory attributes (no lazy-load off-loop). `asyncio.to_thread` matches the existing CPU-offload pattern in `common/llm/providers/*`.

**🟠 Medium — `resource_id` hashed un-normalized at write but normalized at verify**: Fixed, and at a slightly deeper altitude than the suggestion. Rather than normalizing only at the write call site, I moved the canonicalization **into `canonical_event` itself** (`audit_chain.py`), right beside the existing `_normalise_numbers` for `detail`. `canonical_event` is the documented single home for "writer and verifier agree byte-for-byte," and `detail` was already normalized there while `resource_id` was left to callers — that asymmetry is exactly what lets the two paths drift. Now both the write path (`_audit_chain_one_tenant`) and the verify path pass the raw value and `canonical_event` canonicalizes via `_canonical_resource_id` (uppercase/non-canonical UUID string, `UUID` object, and `None`/falsy all map to the same lowercase-hyphenated form a UUID column reads back). Scoped to `resource_id` only — the other hashed fields (`tenant_id`/`agent_id`/`action`/`resource_type`) are `Text` columns with no string-vs-typed hazard.

Tests added: `test_canonical_event_normalises_resource_id_case_and_type` (uppercase string vs `UUID` object → identical canonical bytes), `test_canonical_resource_id_normalises_case_and_type` (the helper directly), and `test_chain_verifies_batch_event_with_uppercase_uuid_resource_id` (end-to-end: an uppercase-UUID batch event now verifies `valid: true` — it would have failed `event_hash_mismatch` before).

CI re-running on `f0ba82a`.

Eldad-Caura:
@claude

Eldad-Caura:
Evaluated both — I'm **not applying either as suggested**; neither is a current correctness bug, and both suggested fixes have downsides that matter here. Rationale + follow-up offers below. #397 stays green @ `fcb115c`.

**🟡 Medium — sequential per-tenant writes → `asyncio.gather`**: Declining the *unbounded* gather. `_audit_chain_one_tenant` opens a **writer-pool** session (`get_session()`), and the audit flush runs inside core-storage-api sharing that pool with the write-request handlers. That pool is deliberately conservative — `DB_POOL_SIZE=20` / `DB_MAX_OVERFLOW=5`, sized to `containerConcurrency=20` so handlers never queue on checkout (**CAURA-685**). An unbounded gather over a many-tenant batch ("dozens of tenants," as you note) lets one flush grab dozens of writer connections at once and starve concurrent memory writes — the exact pool-checkout tail-latency failure CAURA-685 was filed to fix.

Two things make sequential a reasonable default rather than a regression to undo:
- The per-tenant **transaction boundary is intrinsic** to the design — each tenant commits independently under its own `audit_chain_head` `FOR UPDATE` lock so one tenant's failure can't roll back another's; it can't collapse back to the prior single `add_all`.
- The flusher is a **fixed-interval background process**, off the request hot path, so its wall-clock isn't user-facing.

The throughput point is fair for large batches, but the safe form is **bounded** concurrency (semaphore), and the bound should be sized from writer-pool headroom under real concurrent-flush load — not a blind fan-out. Happy to add that as a follow-up once it can be sized from staging pool metrics; I'd rather not add concurrency to the tamper-evident write path on a guessed cap right before merge.

**🔵 Low — ORM instances in the worker thread**: Declining the suggested dict-materialization because it would undo this PR's own High fix. Building a 500k-element list of 11-key dicts in a comprehension runs **on the event loop** (before `to_thread`), reintroducing the O(n) blocking that the `asyncio.to_thread` offload was added to eliminate — a concrete regression for a speculative risk.

You're right it's safe today (`get_read_session` doesn't commit → `expire_on_commit` never fires → loaded columns stay in `__dict__`; `AuditLog` has no relationships), and the helper docstring already states the contract ("Operates only on already-loaded ORM attributes, so it's safe off the event loop"). If we want to harden against a future relationship, the zero-event-loop-cost way is a **column-select** (`select(AuditLog.seq, AuditLog.prev_hash, …)` → lightweight `Row`s straight from the driver, no ORM identity map / lazy-load), not an extra dict pass — happy to do that as a follow-up.

Net: no code change this round; #397 is green and ready. Both captured as follow-ups if you want them.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #398, opened 2026-06-15
feat(governance): ingestion-boundary PII + business/personal gate (eToro)

Eldad-Caura:
## What & why

Second of the **eToro governance increment** PRs (companion to the tamper-evident audit chain in #397). Screens every memory for **PII / PCI / secrets / credentials** and classifies it **business vs personal** *before* it persists to the knowledge graph, enforcing a per-tenant policy. Opt-in (disabled by default); every enforcement action is recorded to the audit log.

Depends conceptually on #397 for tamper-evident audit, but is independent at the code level (emits via the existing `log_action`).

## Controls (all per-tenant, configurable, no redeploy)

- **PII** — `enabled` · `action` ∈ {mask, drop, flag} · per-category toggles (email, phone, credit_card, iban, national_id, api_key, secret).
- **Non-business** — `enabled` · `disposition` ∈ {drop, keep_private, store}.

## How it works

- **`common/governance/`** — a deterministic **60+ pattern library** (emails, phones, cards w/ **Luhn**, IBAN w/ **mod-97**, national IDs, provider API keys, secrets w/ **Shannon-entropy** gating), seeded from the Skill-Factory Sentinel regexes. `scan()` returns `Finding`s carrying category + offsets only (**never raw text**); `mask()` redacts spans.
- **`business_relevance`** rides the *existing* enrichment LLM call (one new field — no extra round-trip), wired through the worker routing tuples + merge step.
- **Config** lives in OSS `organization_settings` `DEFAULT_SETTINGS` (Rule: one settings store) + enum validation + `ResolvedConfig` accessors. Booleans default off; action/disposition default to the safe `flag`/`store`.
- **Two-track enforcement** — fast mode (the default) defers enrichment to the worker, so the LLM signal isn't available pre-write:
  | Signal | strong | fast |
  |---|---|---|
  | Deterministic scan (regex+Luhn/IBAN/entropy) | sync pre-write | **sync pre-write** |
  | LLM `contains_pii` / `business_relevance` | sync pre-write | **async post-write remediation** |
  - `GovernanceScanContent` runs in **both** modes after `LoadTenantConfig`, before `ComputeContentHash` — masking mutates `data.content` **in place** so the hash/dedup/embedding/stored row all see the redacted form (solves the mask-vs-hash ordering structurally); `drop` raises 422 pre-persist (same mechanism as `CheckSemanticDuplicate`'s 409).
  - `GovernanceDecision` (strong) acts on the LLM signal; fast mode applies the same via `remediate_after_enrichment` in the ENRICHED consumer (mask→update, drop→soft-delete, keep-private→`visibility=scope_agent`). Bulk + STM paths gated too.
- **Fail-closed** — the deterministic scan is the always-on high-risk backstop (validators are deterministic, never "uncertain"); `GovernanceDecision` takes no destructive action on an absent/uncertain LLM signal (records `governance_llm_uncertain`).
- **PII-safe audit** — details carry categories + spans + content hash, **never raw values**.

## Tests

`test_governance_patterns.py` (validators incl. Luhn/IBAN/entropy/overlap + false-positive corpus), `test_governance_config.py` (resolution + enum validation), `test_governance_gate.py` (step-level mask/drop/flag/keep-private/business/fail-closed with audit-emit capture + E2E mask/drop/clean writes through the real app). **41 governance + 194 write-path/bulk/consumer/pipeline regression tests pass.** ruff + mypy clean on core-api / core-storage-api / core-worker / common.

> Note: provider-key test fixtures are assembled from fragments so no contiguous secret literal sits in source (secret-scanning push protection).

## Demo recommendation
eToro's default write mode is **fast**; the deterministic scan blocks regex-detectable PII pre-persist in both modes, but for the LLM-classified cases to also be synchronous-pre-write, set the demo tenant to `write.default_write_mode="strong"` (config flip, no code).

### DIFF
--- a/common/enrichment/_prompts.py
+++ b/common/enrichment/_prompts.py
@@ -144,8 +144,15 @@ Analyze the following memory content and return a JSON object with these fields:
         "retrieval_hint" : same guidance as field 11 — short, query-aligned
     - When in doubt, leave atomic_facts as null.
 
+13. "business_relevance": one of "business" | "personal" (default "business")
+    - "personal": private life unrelated to work — health, family, personal
+      finance, relationships, errands, vacation planning, casual chat, idle ideas.
+    - "business": work / professional / operational content (the default).
+    - When unsure, choose "business" — only mark "personal" when you are
+      confident the content is non-work.
+
 Return ONLY valid JSON (no markdown fences):
-{{"memory_type": "...", "weight": 0.0, "title": "...", "summary": "...", "tags": ["..."], "status": "active", "ts_valid_start": null, "ts_valid_end": null, "contains_pii": false, "pii_types": [], "retrieval_hint": "", "atomic_facts": null}}
+{{"memory_type": "...", "weight": 0.0, "title": "...", "summary": "...", "tags": ["..."], "status": "active", "ts_valid_start": null, "ts_valid_end": null, "contains_pii": false, "pii_types": [], "retrieval_hint": "", "atomic_facts": null, "business_relevance": "business"}}
 
 Content:
 {content}
--- a/common/enrichment/schema.py
+++ b/common/enrichment/schema.py
@@ -105,6 +105,11 @@ class EnrichmentResult(BaseModel):
     ts_valid_end: str | None = None
     contains_pii: bool = False
     pii_types: list[str] = []
+    # Governance business-vs-personal gate (eToro): "business" (work-relevant,
+    # default) vs "personal" (vacation planning, casual chat, idle ideas). The
+    # admin policy decides what to do with "personal" content; the default is
+    # the fail-closed-safe value (only a confident "personal" triggers the gate).
+    business_relevance: str = "business"
     retrieval_hint: str = ""
     atomic_facts: list[AtomicFact] | None = None
     llm_ms: int = 0
--- a/common/enrichment/service.py
+++ b/common/enrichment/service.py
@@ -166,6 +166,10 @@ def _validate_enrichment(raw: dict, llm_ms: int) -> EnrichmentResult:
     raw["atomic_facts"] = cleaned_facts if cleaned_facts else None
     raw["contains_pii"] = bool(raw.get("contains_pii", False))
     raw["pii_types"] = raw.get("pii_types") or []
+    # Governance gate: clamp to the allowed set; anything off-spec (including a
+    # missing value) falls back to "business" — the fail-closed-safe default, so
+    # only a confident "personal" from the LLM ever triggers the disposition gate.
+    raw["business_relevance"] = raw.get("business_relevance") if raw.get("business_relevance") in ("business", "personal") else "business"
     raw["llm_ms"] = llm_ms
     return EnrichmentResult(**raw)
 
--- a/common/governance/__init__.py
+++ b/common/governance/__init__.py
@@ -1,0 +1,32 @@
+"""Deterministic PII / PCI / secret detection for the governance gate.
+
+The ingestion-boundary governance gate (eToro) screens every memory for
+sensitive data before persistence using these deterministic patterns combined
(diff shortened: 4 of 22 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
Addressed in `c542440` (amend + force-with-lease — branch stays at one commit). All 52 governance tests pass; ruff + mypy clean on `core-api`.

**🔴 High — `memory_id = str(memory.get("id"))` → `"None"`** (`governance_remediation.py`): Fixed as suggested. Guard `memory.get("id") is None` → log a warning and return `False` (no-op), so a malformed enriched-event payload can't soft-delete the literal `"None"` or stamp `resource_id="None"` on audit rows. The fast-mode remediation path had no unit coverage before — added `tests/test_governance_remediation.py` (incl. `test_missing_id_skips_without_side_effects`).

**🟡 Medium — metadata fallback not written back** (`governance_decision.py`): Fixed as suggested. Added `data.metadata = metadata` in the `metadata is None` fallback, mirroring `GovernanceScanContent`. Added `test_decision_writes_metadata_back_when_fields_absent`.

**🟡 Medium — fast-mode logs `pii_flag` when configured action is `mask`**: Genuine gap — resolved **differently** than the suggested `ACTION_PII_MASK`, and the difference matters for the compliance story:

- Emitting `pii_mask` here would make the audit *verb* claim a redaction that didn't happen. The LLM free-form signal has no span offsets, so nothing is actually masked (and in fast mode the row is already persisted). `pii_mask` currently has a precise, truthful meaning — it's emitted **only** by the deterministic, span-aware path in `GovernanceScanContent` when content is genuinely redacted in place. Overloading it would (a) mislead a compliance reviewer counting "masked" rows and (b) diverge from the deliberate strong-mode convention (`governance_decision.py`: *"mask or flag … record the PII judgement on the row (flag) either way"*).
- Instead I kept `action=pii_flag` (truthful) and recorded the tenant's intent in the audit **detail** as `configured_action: "pii_mask"` — closing the distinguishability gap (mask-configured fallback vs. genuine flag policy) without claiming a redaction. Applied consistently in **both** fast (`remediate_after_enrichment`) and strong (`GovernanceDecision`) modes via the shared `llm_pii_audit_detail` helper (which suppresses `configured_action` when it equals the action taken).
- Covered by `test_decision_pii_mask_config_flags_and_records_intent` / `..._flag_config_records_no_configured_action` (gate) and the equivalents in the new remediation test file.

CI re-running on `c542440`.

Eldad-Caura:
@claude

Eldad-Caura:
Addressed in `1e9ba5e` (amend + force-with-lease — branch stays at one commit). 10 remediation tests pass; ruff + mypy clean on `core-api`.

**🟡 Medium — soft-delete fires before audit in fast-mode drop paths**: Fixed in both drop branches. `emit_governance_audit(...)` now runs **before** `sc.soft_delete_memory(memory_id)` in the PII-drop and non-business-drop branches, matching the audit-before-mutate invariant `GovernanceScanContent` documents. As you note: a delete that succeeds before a failing audit would leave an untracked deletion in the tamper-evident log, whereas audit-then-failed-delete leaves a remediable "intended to drop" trace.

Added `test_drop_audits_before_soft_delete` — a parametrized test (PII + non-business) that captures audit and soft-delete into one ordered log and asserts `["audit:<drop>", "soft_delete"]`. It fails on the old ordering and passes now, locking the invariant against regression.

Note: the keep-private branch (`update_memory` → audit) keeps update-then-audit, since `keep_private` is non-destructive (visibility change, fully reversible) — the audit-first ordering specifically guards the irreversible **delete** paths. Happy to make that consistent too if you'd prefer.

CI re-running on `1e9ba5e`.

Eldad-Caura:
@erni-a plesae approve

erni-a (approved):

PR #403, opened 2026-06-16
feat(core-api): fast business/personal pre-gate before enrichment (opt-in)

Eldad-Caura:
Implements the approved design (`DESIGN-business-personal-pregate.md`) — **Design A (additive), disabled by default.**

## What it does
Adds an optional **fast business-vs-personal go/no-go** at the front of the write pipeline — after the deterministic PII scan, **before** enrichment, embedding, and the (already-deferred) entity extraction. On a confident **`personal`** verdict when the tenant's disposition is **`drop`**, the write is rejected with `422` *before* any of that work runs and before a row is written.

## Highlights
- **New step `BusinessPersonalPregate`**, wired into the **strong + fast** write pipelines (after `GovernanceScanContent`, before `ParallelEmbedEnrich`); correctly absent from STM / enrichment-only.
- **Its own provider/model** (via `common.llm` `call_with_fallback`), so the signal is **independent of the enrichment provider** — it survives `ENTITY_EXTRACTION_PROVIDER=none` (the green-local/red-CI blind spot).
- **Disabled by default.** New `governance.non_business.pregate` settings (`enabled` / `provider` / `model` / `min_confidence`), all opt-in, validated + resolved alongside the existing governance settings. **Prod behavior is unchanged until a tenant enables it.**
- **Truthful audit verb** `nonbusiness_pregate_drop` (nothing persisted — distinct from `nonbusiness_drop` which soft-deletes a stored row), emitted **before** the raise.
- **Additive (Design A):** the post-enrichment `GovernanceDecision` remains the accurate backstop; `keep_private`/`store` stay there (they persist the row, so an early exit saves nothing).
- **Fail-open:** a `none` provider or any classifier failure never blocks a write.

## Open design decisions (per the doc, as approved)
D1 additive-first · D2 model configurable (defaults to the cheap tier) · D3 drop-only · D4 fail-open default with a `min_confidence` knob · D5 default-off. Graduating to full separation (B) remains a documented follow-up.

## Verification
- 43 tests green (new `tests/test_business_personal_pregate.py` + existing governance suites — no regression): enablement gating, drop/flow-through, `min_confidence`, **provider-independence + fail-open reproduced under `provider=none`**, settings default-off + validation, and pipeline wiring.
- `ruff check` + `ruff format --check` clean; `mypy` clean on touched files.
- `/simplify` run before commit (one cleanup applied: removed dead `isinstance` guards + a redundant cast in the settings validator).

### DIFF
--- a/core-api/src/core_api/pipeline/compositions/write.py
+++ b/core-api/src/core_api/pipeline/compositions/write.py
@@ -2,6 +2,7 @@
 
 from core_api.pipeline.runner import Pipeline
 from core_api.pipeline.steps.write import (
+    BusinessPersonalPregate,
     CheckContentLength,
     CheckExactDuplicate,
     CheckSemanticDuplicate,
@@ -69,6 +70,9 @@ def build_fast_write_pipeline() -> Pipeline:
             CheckContentLength(),
             LoadTenantConfig(),
             GovernanceScanContent(),
+            # Opt-in fast business/personal go/no-go: reject personal content
+            # (disposition=drop) before enrichment/embedding/extraction run.
+            BusinessPersonalPregate(),
             ComputeContentHash(),
             ParallelEmbedEnrich(),
             MergeEnrichmentFields(),
@@ -105,6 +109,9 @@ def build_strong_write_pipeline() -> Pipeline:
             CheckContentLength(),
             LoadTenantConfig(),
             GovernanceScanContent(),
+            # Opt-in fast business/personal go/no-go: reject personal content
+            # (disposition=drop) before enrichment/embedding/extraction run.
+            BusinessPersonalPregate(),
             ComputeContentHash(),
             ParallelEmbedEnrich(),
             MergeEnrichmentFields(),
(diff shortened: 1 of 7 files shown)


Eldad-Caura:
@claude please review — opt-in fast business/personal pre-gate before enrichment (Design A, default-off). Key things to sanity-check: the step is a true no-op when disabled, the audit verb is truthful (nothing persisted), the new settings are validated/resolved like the sibling governance block, and the classifier is fail-open + independent of the enrichment provider.

Eldad-Caura:
@claude fixed — removed the brace-escaping in the classifier; `content` is now passed directly to `str.format` (which inserts substituted values literally, so escaping only corrupted JSON/code/braces). Extracted `_build_prompt` and added a regression test asserting brace-containing content survives into the prompt. Verified empirically + ruff/mypy/tests green. (Note: the same latent bug exists in `common/enrichment/service.py` `enrich_memory` — I modeled the classifier on it — flagged separately as out-of-scope for this PR.)

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #404, opened 2026-06-16
fix(prompts): stop brace-escaping content before str.format

Eldad-Caura:
## The bug
`enrich_memory` (`common/enrichment/service.py`), `_generate_rule` (`evolve_service.py`), and `synthesize_insights` (`insights_service.py`) all pre-escaped content (`{` → `{{`) before passing it as a **substitution value** to `str.format()`, with comments claiming an un-escaped value would raise `KeyError`.

That's incorrect. `str.format()` parses only the **template** for replacement fields — a substituted value is inserted **literally** and never re-scanned, so it never raises `KeyError` and never needs escaping. The escaping instead **corrupts** any brace-containing content before the LLM sees it:

```python
>>> "{c}".format(c='{"a": 1}')      # substituted value, literal — no error
'{"a": 1}'
>>> "{c}".format(c='{{"a": 1}}')    # escaped form is inserted as-is → corrupted
'{{"a": 1}}'
```

So JSON, code, regex, templates, and dict reprs (insights cluster mode) reached the enrichment/rule/insight LLM with doubled braces — degrading exactly the technical/business content these paths handle.

## The fix
Pass `content` / `outcome` / `memories` directly to `.format()` at all three sites and correct the misleading comments. The evolve `_sanitize_content(outcome, max_len=2000)` size cap is preserved — only the escaping is removed.

## Tests
- `test_enrich_memory_prompt_preserves_braces` — exercises `enrich_memory` with a mocked provider and asserts brace-containing content reaches the prompt **verbatim** (fails on the old escaping).
- Brace-preservation invariants for the evolve (`_RULE_GENERATION_PROMPT`) and insights (`_PROMPT_DISPATCH`) templates.
- Verified empirically; `ruff check` + `format --check` + `mypy` clean; enrichment suites green (no regression).

## Context
Same class of bug was just fixed in `core_api/services/business_classifier.py` in #403 (the new pre-gate classifier was modeled on `enrich_memory`); this PR fixes the pre-existing occurrences in the enrichment / evolve / insights paths.

### DIFF
--- a/common/enrichment/service.py
+++ b/common/enrichment/service.py
@@ -169,7 +169,11 @@ def _validate_enrichment(raw: dict, llm_ms: int) -> EnrichmentResult:
     # Governance gate: clamp to the allowed set; anything off-spec (including a
     # missing value) falls back to "business" — the fail-closed-safe default, so
     # only a confident "personal" from the LLM ever triggers the disposition gate.
-    raw["business_relevance"] = raw.get("business_relevance") if raw.get("business_relevance") in ("business", "personal") else "business"
+    raw["business_relevance"] = (
+        raw.get("business_relevance")
+        if raw.get("business_relevance") in ("business", "personal")
+        else "business"
+    )
     raw["llm_ms"] = llm_ms
     return EnrichmentResult(**raw)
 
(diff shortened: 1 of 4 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #405, opened 2026-06-16
fix(entity-linking): backfill embeddings via Core table, not ORM bulk-by-PK

Eldad-Caura:
## Prod incident

The nightly `entity_linking_full` pipeline's **`backfill_entity_embeddings`** step has been failing in prod (observed 2026-06-16, ~6 occurrences/night, all in the 02:00–02:11 UTC cron window) with:

```
entity_linking_full.backfill_entity_embeddings: FAILED — No primary key value supplied
for column(s) entities.id; per-row ORM Bulk UPDATE by Primary Key requires that records
contain primary key values
```

Effect: entity `name_embedding`s are silently **not** backfilled for affected tenants, and the runner `break`s the rest of that org's linking run for the night.

## Root cause

`session.execute(update(Entity), [param dicts])` (an **ORM-mapped** class + an executemany param list) routes to SQLAlchemy's *"ORM Bulk UPDATE by Primary Key"*, which requires every param dict to carry the PK column `id`. Our dicts key the PK off a custom `eid` bindparam in the `WHERE` clause, so the bulk-by-PK machinery finds no `id` and raises.

The earlier `synchronize_session=False` change (prod 2026-06-13) only silenced a *different* error on the same ORM path — it masked this second failure mode rather than removing the ORM bulk-update path.

This is the same `entity_linking_full` family as the prior `discover_cross_links` (#337) and `infer_relations` (#341) fixes.

## Fix

Target the **Core table** `update(Entity.__table__)` instead of the ORM-mapped `Entity`. That is a plain Core executemany `UPDATE` which honours the custom `eid`/`emb` bindparams and has **no** ORM bulk-by-PK or session-synchronisation behaviour at all. The `.execution_options(synchronize_session=False)` is removed (irrelevant on a Core statement).

## Verification

- Reproduced against **SQLAlchemy 2.0.49**: the ORM form (`update(Entity)` + executemany) raises the exact prod `InvalidRequestError`; the Core form (`update(Entity.__table__)`) succeeds.
- The regression test now asserts the statement is a Core `UPDATE` (no ORM compile plugin) — the precise property that prevents the bulk-by-PK path — instead of the now-removed `synchronize_session` option.
- `pytest tests/pipeline/` → 88 passed. `ruff check` / `ruff format --check` / `mypy` clean.

### DIFF
--- a/core-api/src/core_api/pipeline/steps/entity_linking/backfill_entity_embeddings.py
+++ b/core-api/src/core_api/pipeline/steps/entity_linking/backfill_entity_embeddings.py
@@ -66,16 +66,23 @@ class BackfillEntityEmbeddings:
         backfill_count = 0
         if updates:
             await ctx.require_db.execute(
-                # ``synchronize_session=False``: this is a bulk ORM UPDATE with
-                # extra WHERE criteria over an executemany param list, which
-                # SQLAlchemy refuses to session-synchronize (InvalidRequestError,
-                # prod 2026-06-13). There is no live ORM session state to keep in
-                # sync here — the backfill writes name_embedding and moves on — so
-                # skipping synchronization is correct, not just a silencer.
-                update(Entity)
-                .where(Entity.id == sa.bindparam("eid"), Entity.tenant_id == tenant_id)
-                .values(name_embedding=sa.bindparam("emb"))
-                .execution_options(synchronize_session=False),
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #406, opened 2026-06-16
fix(http-retry): widen connection-phase retry to ride out storage cold starts

Eldad-Caura:
## Prod signal (2026-06-16)

A steady trickle of `httpx.ConnectTimeout` reached `prod-memclaw-core-api` across **every** core-api → core-storage-api path — Pub/Sub enrichment (`pubsub handler raised; nacking for redelivery`), entity extraction, contradiction detection, and audit flush. All are storage-api connect timeouts during cold starts / instance recycles.

The current policy is 3 attempts (~0.6s of backoff). That is too short to ride out a Cloud Run cold start, so a transient connect blip surfaces as a handler failure — and on the Pub/Sub path it triggers a nack → immediate-redelivery storm (`common/events/pubsub.py` nacks with `ack_deadline=0`).

## Change

Widen **only** the connection-phase path (`with_connect_phase_retry`) to `CONNECT_PHASE_MAX_ATTEMPTS = 5`.

Why only that path: `ConnectTimeout` / `ConnectError` / `PoolTimeout` are raised *before* the request is sent, and against an instance that isn't accepting connections yet — so retrying them:
- cannot double-insert (request provably never sent), and
- adds **zero load to a healthy server** (the connection never lands), so extra attempts just wait out the cold start.

The idempotent path (`with_retry`, which also retries `ReadTimeout`/5xx) **stays at 3 attempts** so a genuine 5xx storm isn't amplified by extra load against a struggling server.

Also:
- `RETRY_BACKOFF_MAX_S = 2.0` caps a single backoff sleep so the higher attempt count stays bounded.
- Factored the capped + jittered delay into a small `_backoff_delay()` helper (dedups the two call sites).

This **complements, not replaces**, keeping storage-api warm (Cloud Run min-instances) — the durable fix for the cold starts themselves.

## Tests

- `test_post_gives_up_after_max_attempts_on_connect_timeout` now asserts 5 attempts for the connect-phase path; `test_get_retries_then_gives_up_after_max_attempts` still asserts 3 for the idempotent path — the contrast is the regression guard.
- `pytest tests/test_storage_client_retry.py` → 16 passed. `ruff check` / `ruff format --check` / `mypy common/http_retry.py` clean.

### DIFF
--- a/common/http_retry.py
+++ b/common/http_retry.py
@@ -15,19 +15,29 @@ History (load-bearing for the policy split):
   dropped) both died on first-attempt ``ConnectTimeout`` behind the VPC
   connector because POSTs had no retry at all. Connection-phase retry
   for non-idempotent methods landed in response (caura-memclaw#333).
+* 2026-06-16 prod — a steady trickle of ``httpx.ConnectTimeout`` still
+  reached core-api on the Pub/Sub enrichment, entity-extraction,
+  contradiction and audit-flush paths (storage-api cold starts /
+  instance recycles). 3 attempts (~0.6s of backoff) is too short to
+  ride out a cold start, so the connect-phase path now retries more
+  times — see ``CONNECT_PHASE_MAX_ATTEMPTS``. This complements, not
+  replaces, keeping storage-api warm (Cloud Run min-instances).
 
 Retry policy:
- - Max 3 attempts (1 initial + 2 retries)
- - Exponential backoff with jitter: ~0.2s, ~0.4s
- - Worst-case added latency: < 1s
- - Idempotent HTTP methods retry ``RETRYABLE_EXCEPTIONS`` plus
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
Addressed in `7ad53d6` (force-pushed) — took **option A**: apply the `RETRY_BACKOFF_MAX_S` cap *after* jitter in `_backoff_delay`, so it's a true hard ceiling (≤ 2.0s) and the docstring + module policy comment are now accurate. Verified the finding was correct (jitter-after-cap reached MAX × 1.1 = 2.2s).

Added a regression test `test_backoff_delay_cap_is_a_hard_ceiling` asserting the bound holds across attempts 1–8. `ruff`/`mypy` clean; 17 retry tests pass.

@claude please re-review.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #407, opened 2026-06-16
fix(audit): idempotent bulk flush via client_event_id — stop dropping audit rows

Eldad-Caura:
## Problem

The async audit flush (`POST /audit-logs/bulk`) used **connect-phase-only retry**, so a `ReadTimeout`/5xx dropped a whole tenant's slice (logged `audit batch flush failed … events lost`, never requeued) — **silent audit-log data loss**. This was the known gap behind the `audit chunk flush failed` / `audit batch flush failed` signatures (the connect-timeout occurrences are separately addressed by widening connect-phase retry).

Naively switching to full retry was **unsafe**: `audit_log` is a tamper-evident **per-tenant hash chain** with **no per-event identity**, so re-sending an already-committed batch would re-append every event with fresh `seq`s and silently **double the chain** — and it would still pass `verify_chain`, so the corruption is invisible.

## Fix — make the flush idempotent, then retry is safe

1. **core-api `log_action`** mints a UUID `client_event_id` per event (stable across a retry of the same enqueued event).
2. **Migration 026**: `audit_log` gains a nullable `client_event_id` column + a partial unique index `(tenant_id, client_event_id) WHERE client_event_id IS NOT NULL`, built with `CREATE UNIQUE INDEX CONCURRENTLY` inside an `autocommit_block`. This follows the **007** online pattern and deliberately **not** the **025** plain in-transaction `CREATE INDEX` that stalled storage-writer startup past the 300s migration-lock timeout.
3. **`_audit_chain_one_tenant`** dedups already-chained `client_event_id`s **under the per-tenant head `FOR UPDATE` lock, BEFORE assigning `seq`** (plus intra-batch duplicates). Survivors get **contiguous** seqs, so the chain stays gap-free and the head consistent — avoiding the seq-gap trap a bare `ON CONFLICT` would create. Events with no `client_event_id` (legacy single-event path) are never deduped.
4. **`create_audit_logs_bulk`** switches to full `with_retry` (`ReadTimeout` + 5xx), now safe: a replayed batch dedups instead of double-appending. Recovers the lost-ack case that previously dropped the slice.

## Why it's safe

`_audit_chain_one_tenant` already commits one atomic transaction per tenant. With the dedup, the three retry cases are all correct:
- **pre-commit failure** → rolled back, nothing chained → retry re-inserts.
- **committed but lost ack** (`ReadTimeout`) → retry's dedup SELECT (under the head lock) sees the committed rows → skips them, no double-append.
- **partial** → impossible (single transaction).

## Deploy ordering

Storage-first (storage owns the schema; the migration runs in the lifespan startup before serving). New core-api + old storage is safe (extra field ignored); new storage + old core-api is safe (column nullable).

## Tests

- `test_audit_chain.py` *(integration, runs against Postgres in CI)*: a replayed batch dedups — no duplicate rows, contiguous seqs, `verify_chain` stays valid; intra-batch duplicate chained once.
- `test_storage_client_retry.py`: `create_audit_logs_bulk` retries `ReadTimeout` and 5xx (unlike a bare POST).
- `test_audit_queue.py`: `log_action` mints a unique `client_event_id` per event.
- `ruff check` / `ruff format --check` / `mypy` clean on the `core-api/src` + `core-storage-api/src` files (28 mock tests pass locally; the chain-integration tests run in CI).

### DIFF
--- a/common/models/audit.py
+++ b/common/models/audit.py
@@ -1,7 +1,7 @@
 import uuid
 from datetime import datetime
 
-from sqlalchemy import BigInteger, DateTime, LargeBinary, Text, text
+from sqlalchemy import BigInteger, DateTime, Index, LargeBinary, Text, text
 from sqlalchemy.dialects.postgresql import JSONB
 from sqlalchemy.orm import Mapped, mapped_column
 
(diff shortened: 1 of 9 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
Both findings addressed (verified correct) in the amended commit:

- **Medium — migration retry-safety:** `upgrade()` now uses `ALTER TABLE audit_log ADD COLUMN IF NOT EXISTS client_event_id text` instead of `op.add_column`. Confirmed the risk: entering `autocommit_block()` commits the column add, so a failed `CONCURRENTLY` build would leave the column present but `alembic_version` unstamped → a plain `ADD COLUMN` raises "column already exists" on retry. `IF NOT EXISTS` makes it idempotent (metadata-only on a clean run).
- **Low — guard consistency:** `incoming_ids` now uses `is not None`, matching the per-event loop guard. (UUIDs are never empty in practice, but the guards are now aligned.)

`ruff`/`mypy` clean; migration-chain single-head guard bumped to 026 (the original CI failure); rebased onto current main (incl. #406). @claude please re-review.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #408, opened 2026-06-17
fix(storage): harden migration advisory-lock so a slow migration can't crash booting writers

Eldad-Caura:
## Background — the 2026-06-16 incident

On a `core-storage-writer` deploy, all N replicas race for a Postgres advisory lock (`8675309`); one wins and runs `alembic upgrade head` **while holding the lock**, the rest poll for it and only serve once it's released. The wait was a **hardcoded 300s**. But the lock is held for the *entire* migration — including minutes-long `CREATE INDEX CONCURRENTLY` builds — so a slow-but-healthy migration crashes the N-1 waiting replicas.

On 2026-06-16, migration **025** (a plain, non-`CONCURRENTLY` `CREATE INDEX` on the large `audit_log` table) ran past 300s and failed **6** `prod-memclaw-core-storage-writer` boots:
```
TimeoutError: Migration advisory lock 8675309 not acquired within 300s → Application startup failed. Exiting.
```
It self-healed (Cloud Run restarted them once the winner finished), but the next slow migration repeats it.

## Approach (A + B + D, chosen with the team)

Keeps migrations in the lifespan startup — the deliberate **CAURA-621** atomic-deploy property stays intact (no pre-deploy migration job / deploy-ordering orchestration).

1. **Configurable + generous wait (A):** `MIGRATION_LOCK_WAIT_SECONDS` (`settings.migration_lock_wait_seconds`), default raised **300s → 1800s**. The wait is now a *stuck-migration backstop*, not a routine cap — it must exceed the slowest real migration so a progressing one never crashes a booting replica.
2. **Wait-don't-crash, observably (B):** keeps the **safe** `pg_try_advisory_lock`-in-`AUTOCOMMIT` poll (a blocking `pg_advisory_lock` would hold an idle-in-transaction that stalls the winner's `CONCURRENTLY` build — the existing comment documents this), and **logs progress every 30s** while waiting, so a long-but-healthy migration reads as progress in the logs instead of a silently hung boot. The deadline still raises a clear *stuck-migration* error (better than letting the Cloud Run probe kill the instance silently).
3. **CI guard (D):** `test_no_plain_create_index_on_large_tables` fails if a migration from **005 onward** does a plain `CREATE INDEX` on a large pre-existing table (`audit_log` / `memories` / `entities` / `documents` / `memory_entity_links` / `relations`) — it must be `CREATE INDEX CONCURRENTLY` in an `autocommit_block` (see 007 / 026). `025` is allow-listed as documented, already-applied debt; `001–004` build the initial schema (indexes on empty tables) and are exempt.

**Caveat:** the effective wait is `min(MIGRATION_LOCK_WAIT_SECONDS, Cloud Run startup-probe deadline)`. The 300s error fired first in the incident, so the probe grace was ≥300s — but the storage-writer's Cloud Run startup timeout should be confirmed ≥ 1800s (Terraform, out of this repo) or raised alongside, else the probe kills the instance before the longer wait helps.

This is the root-cause follow-up to #407's migration 026 (which already follows the `CONCURRENTLY` pattern). It does **not** retroactively change 025 (already applied in prod) — it stops the *next* slow migration from repeating the crash.

## Tests
- `test_no_plain_create_index_on_large_tables` (the new guard) + the existing migration-chain tests pass; full `test_skill_schema_v1.py` → 79 passed.
- `ruff check` / `ruff format --check` / `mypy` clean on `config.py` + `init.py`. The acquire→upgrade happy path is exercised by CI's "Run migrations" step.

### DIFF
--- a/core-storage-api/src/core_storage_api/config.py
+++ b/core-storage-api/src/core_storage_api/config.py
@@ -47,6 +47,19 @@ class Settings(BaseSettings):
     db_pool_timeout: int = 60
     db_pool_recycle: int = 1800
 
+    # How long a booting writer waits to acquire the migration advisory lock
+    # before giving up. One replica runs ``alembic upgrade head`` while holding
+    # the lock; the others poll for it and only serve once it's released. This
+    # must comfortably exceed the slowest real migration — a large
+    # ``CREATE INDEX CONCURRENTLY`` / backfill can run for minutes, and the lock
+    # is held for the whole run — so a legitimately slow migration doesn't crash
+    # the N-1 booting replicas waiting on it. It was a hardcoded 300s, which
+    # migration 025 blew past on 2026-06-16, failing 6 writer boots. The wait is
+    # now a STUCK-migration backstop, not a routine cap. Keep it <= the Cloud Run
+    # startup-probe deadline, or the probe kills the instance before this fires.
+    # Env: MIGRATION_LOCK_WAIT_SECONDS.
+    migration_lock_wait_seconds: int = 1800
+
     # Service role (CAURA-591 Part B). "hybrid" keeps the original
     # single-service behaviour and is the safe default for OSS + any
     # deploy that hasn't opted into the split. Enterprise SaaS runs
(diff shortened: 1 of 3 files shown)


Eldad-Caura:
@claude please review — CI is green (CI / dco / pre-checks all pass). This is the migration-lock hardening (A+B+D) from the #407 root-cause follow-up.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #409, opened 2026-06-17
fix(search): rank entity-lookup fan-out cap by query-match count (A30)

arkash20:
## What

`ClassifyQuery._collect_memories` caps the entity-lookup candidate pool to `GRAPH_MAX_BOOSTED_MEMORIES` (50) **before** loading content, sorting only by **near-uniform hop-boost**. When a query hits a hub entity (e.g. a bare `john smith` linking 100+ memories), the gold is one of N near-equal-boost candidates and gets cut by the arbitrary cap — *before* the discriminator-aware `_query_overlap` rerank ever runs. This is the person-hub dilution behind the S1 K=10000 residual misses (**A30**).

## Fix

Rank the cap by **how many of the query's directly-matched (hop-0) entities each memory links to** first, with hop-boost as the tiebreak. A `"What is X's manager?"` gold links to **both** the person hub *and* the matched `manager` role (match-count 2), whereas sibling facts about other attributes link to only the hub (count 1) — so the gold survives the cut regardless of hub size (robust even for the 1000+-member hubs seen at K=10000). The `#NNNN` discriminator is then resolved by the existing `_query_overlap` rerank after load.

- No new constants, no routing change (strategy split unchanged).
- 21-line change, isolated to `_collect_memories`.

## Validation

S1 retrieval bench (replay-only against already-seeded tenants, `text-embedding-3-small@1024`):

| Tenant | Before | After |
|---|---|---|
| k1k | 23/25 | **25/25** |
| k10k | 25/25 | **25/25** (no regression) |

Distractors 25/25 on both; strategy split unchanged (k1k 34/16, k10k 42/8).

New regression test `test_collect_memories_cap_prefers_multi_matched_entity_gold_over_hub_siblings` — constructed so the gold is enqueued **last** (a boost-only stable-sort cap drops it); **verified to fail on pre-A30 code** and pass with the fix.

`ruff check` / `ruff format` / `mypy` clean; related search-pipeline tests (45 + 8) pass.

### DIFF
--- a/core-api/src/core_api/pipeline/steps/search/classify_query.py
+++ b/core-api/src/core_api/pipeline/steps/search/classify_query.py
@@ -312,7 +312,11 @@ class ClassifyQuery:
         )
 
         # Best (lowest hop → highest boost) per memory + collect entity links.
+        # ``memory_match_count`` tracks how many DIRECTLY-MATCHED (hop-0) query
+        # entities each memory links to — the pre-load relevance signal that
+        # survives the fan-out cap below.
         memory_boost: dict[str, float] = {}
+        memory_match_count: dict[str, int] = {}
         memory_entity_links: dict[str, list[EntityLinkOut]] = {}
         for link in all_links:
             mem_id, ent_id_str, role = link["memory_id"], link["entity_id"], link.get("role")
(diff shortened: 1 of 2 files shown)


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #410, opened 2026-06-17
fix(rate-limit): fail open on Redis outage instead of 500 (view_rate_limit AttributeError)

Eldad-Caura:
## Prod incident (2026-06-17)

A Redis connection reset turned into **500s on `POST /api/v1/search`** (found while reviewing today's prod alerts):
```
slowapi: Failed to rate limit. Swallowing error → redis.ConnectionError: Connection reset by peer
Unhandled exception on POST /api/v1/search → AttributeError: 'State' object has no attribute 'view_rate_limit'
```
The limiter is configured `swallow_errors=True` — the *intent* is to fail open when the rate-limit backend is down. But it doesn't fully work.

## Root cause (slowapi 0.1.10)

slowapi sets `request.state.view_rate_limit` only **after** it hits the storage backend:
```
extension.py:514   if not self.limiter.hit(lim.limit, *args, cost=cost): ...
extension.py:530   request.state.view_rate_limit = limit_for_header
```
When `hit()` raises (Redis down) and `swallow_errors=True` catches it (extension.py:647, the `"Failed to rate limit. Swallowing error"` log), line 530 is **skipped**. But the limit decorator then reads `request.state.view_rate_limit` **unconditionally** on the way out to inject headers (extension.py `async_wrapper`), so the swallow path `AttributeError`s → 500 instead of failing open. (`_inject_headers` itself tolerates `None`, and headers are off — the crash is purely the eager attribute read.)

## Fix

Seed `request.state.view_rate_limit = None` in **`_key_func`**, which runs first in every limit check (to compute the bucket key) and never touches the backend:
- a successful check overwrites it with the real limit;
- a swallowed one leaves it `None`, which `_inject_headers` tolerates — so `swallow_errors` now genuinely fails open.

Also harden the 429 handler (`app.py`) to read it via `getattr(request.state, "view_rate_limit", None)` for defense in depth.

**Why `_key_func` and not a new middleware:** it runs in the route's own execution context (no BaseHTTPMiddleware scope-propagation caveat — see the existing `SlowAPIMiddleware` note in `app.py`) and adds **zero** request-path overhead.

## Test

`test_search_fails_open_when_rate_limit_storage_errors` monkeypatches the limiter backend (`limiter.limiter.hit`) to raise and asserts `/api/v1/search` returns **200, not 500** — reproduces the prod crash without the seed. Full `tests/test_rate_limit.py` → 8 passed; `ruff`/`mypy` clean.

Found during the 2026-06-17 alert review; the only new category not already covered by #405–#408.

### DIFF
--- a/core-api/src/core_api/app.py
+++ b/core-api/src/core_api/app.py
@@ -488,7 +488,14 @@ async def _json_rate_limit_handler(request: Request, exc: RateLimitExceeded) -> 
     detail_str = f"Rate limit exceeded: {exc.detail}. Try again later."
     body = {"detail": detail_str, **make_error_payload("RATE_LIMITED", detail_str)}
     response = JSONResponse(body, status_code=429)
-    response = request.app.state.limiter._inject_headers(response, request.state.view_rate_limit)
+    # Inject X-RateLimit headers only when the limit was actually evaluated. A
+    # swallowed storage error (see _key_func) leaves view_rate_limit None; skip
+    # the private _inject_headers call entirely rather than depending on slowapi's
+    # None-handling for OUR call site. (On the 429 path it's normally set — a 429
+    # means a limit was hit — so this is defence in depth.)
+    view_rate_limit = getattr(request.state, "view_rate_limit", None)
+    if view_rate_limit is not None:
+        response = request.app.state.limiter._inject_headers(response, view_rate_limit)
     return response
 
 
--- a/core-api/src/core_api/middleware/rate_limit.py
+++ b/core-api/src/core_api/middleware/rate_limit.py
@@ -41,6 +41,29 @@ def _key_func(request: Request) -> str:
     storage backend or access logs, while keeping buckets unique for
     keys that happen to share a prefix.
     """
+    # Fail-open seed for slowapi 0.1.10's swallow_errors gap. slowapi sets
+    # ``request.state.view_rate_limit`` only AFTER it hits the storage backend
+    # (extension.py: ``self.limiter.hit(...)`` then ``view_rate_limit = ...``).
+    # With ``swallow_errors=True`` a Redis outage is swallowed BEFORE that
+    # assignment ("Failed to rate limit. Swallowing error"), yet the limit
+    # decorator then reads ``request.state.view_rate_limit`` unconditionally to
+    # inject headers — so the swallow path raised AttributeError → 500 instead of
+    # failing open (prod 2026-06-17: Redis connection reset → 500s on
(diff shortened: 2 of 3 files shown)


Eldad-Caura:
@claude please review — CI green (CI / dco / pre-checks pass). Small fail-open fix for the slowapi/Redis rate-limit 500 on /api/v1/search found in today's prod alert review.

Eldad-Caura:
Addressed in the amended commit — the finding is correct and verified: slowapi calls `key_func` once **per applied limit** (`extension.py:502` `lim.key_func(request)` inside the per-limit loop, before each `hit()`), so an unconditional seed would reset a value an earlier limit's `hit()` already wrote.

Fix: guarded the seed with `if not hasattr(request.state, "view_rate_limit")` — the attribute still always exists before the first `hit()` (fail-open guarantee intact), but a later limit's `key_func` call no longer clobbers a prior limit's result. Added `test_key_func_seeds_fail_open_default_without_clobbering` covering exactly that sequence.

`ruff`/`mypy` clean; 9 rate-limit tests pass. @claude please re-review.

Eldad-Caura:
Both addressed in the amended commit:

**Medium — reliance on slowapi's private `_inject_headers(None)`:**
- Our 429 handler now **skips** `_inject_headers` entirely when `view_rate_limit` is `None`, so we no longer pass `None` into the private method from our own call site.
- Added a **contract test** `test_inject_headers_is_noop_for_none_limit` (with `headers_enabled=True` so the `current_limit is not None` guard is the operative one) that pins slowapi's None-no-op. I went with a CI contract test rather than the suggested runtime version-warning: a slowapi upgrade that drops the guard then **fails CI**, instead of emitting a warning that's easy to miss at runtime. Comments now cite the verified guard (0.1.10 `extension.py`).

**Low — `_key_func` side effect:** kept deliberately, and heavily commented. Rationale: it runs in the route's **own execution context** (guaranteed-correct, no scope mechanics), adds **zero** request-path overhead, and the suggested `BaseHTTPMiddleware` is explicitly cautioned against a few lines away in `app.py` (the `SlowAPIMiddleware` BaseHTTPMiddleware-separate-task note) — a pure-ASGI alternative would rely on `scope['state']` propagating through that intervening `BaseHTTPMiddleware`, which is subtler than the `key_func` placement. Flagging for @Eldad-Caura to decide at merge if he'd prefer the middleware regardless.

`ruff`/`mypy` clean; 10 rate-limit tests pass. @claude please re-review.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #411, opened 2026-06-17
fix(search): dampen recall_boost so it can't hijack rankings (A26)

arkash20:
## What

`recall_count` is bumped for **every returned row, used or not** (`TrackRecalls` → `memory_increment_recall`) and feeds `recall_boost` back into the rank score (`memory_scored_search` / `memory_repository`) — a self-reinforcing **returned → boosted → returned** loop with no usefulness signal. Popular-but-unused memories asymptote toward the boost cap and entrench at the top; the S1 bench has to pin `recall_boost=false` to avoid run-over-run drift, which is the same bug in miniature.

## Why this PR is the interim, not the full fix

Truly breaking the loop requires a **usefulness signal** (confirmation-gated bump — only count a recall when the memory was actually used/cited). That needs new signal plumbing that overlaps the evals + write-time-signal work (D5/D1), so it's filed as a separate follow-up. This PR is the low-risk damage-reducer: dial the boost down so it can no longer overcome a base-score gap.

## Change

- `RECALL_BOOST_CAP` **1.5 → 1.1** — shrinks the "hijack zone" (how much more relevant a row must be to be safe from a saturated popular one) from **~50% → ~10%**.
- `RECALL_DECAY_WINDOW_DAYS` **90 → 14** — stale popularity fades in ~2 weeks instead of a quarter.

Single source of truth: `resolve_search_profile` propagates both to storage via `search_params` (verified). The boost stays enabled — bounded, not removed.

## Tests

- New guard `test_a26_boost_cannot_hijack_a_more_relevant_memory` — encodes the guarantee (a saturated popular row can't out-rank a >12%-more-relevant fresh one). **Verified to fail at cap=1.5 and pass at 1.1.**
- Updated two existing assertions that hard-coded the old 1.5-cap magnitude to assert *shape relative to the constant* (survives future tuning).
- Full recall/scoring unit suite: 29 passing; ruff/format/mypy clean.

Live S1 replay intentionally skipped: the bench runs with `recall_boost=false`, so a cap/window constant cannot change its result.

### DIFF
--- a/core-api/src/core_api/constants.py
+++ b/core-api/src/core_api/constants.py
@@ -471,8 +471,16 @@ FTS_BOOST_MAX_TOKENS = 3  # queries with more meaningful tokens than this stay a
 FTS_BOOST_SPECIFICITY_RATIO = 0.4  # strict >; at N=2 this means >=1 specific token triggers boost
 SIMILARITY_BLEND = 0.85  # base_score = SIMILARITY_BLEND * similarity + (1 - SIMILARITY_BLEND) * weight (raised from 0.75 — LoCoMo sweep showed +13pp recall)
 SEARCH_OVERFETCH_FACTOR = 2  # fetch top_k * N candidates from storage, trim to top_k after min_similarity filter — gives post-filter headroom
-RECALL_BOOST_CAP = 1.5  # max multiplier from frequent recall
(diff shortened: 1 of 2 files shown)


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #28, opened 2026-04-28
feat(infra): docker-compose pulls published images by default

ran-taig:
## Why

The README's Quick Start says `docker compose up -d`, which currently does a full local build (3–5 min on first run). Now that we publish multi-arch images to ghcr.io on every release ([PR #17](https://github.com/caura-ai/caura-memclaw/pull/17)), users can pull instead — same Dockerfile, same source at the same tag, ~30s instead of minutes.

## What

Adds the standard `image: + build:` Compose pattern to `core-api` and `core-storage-api`:

| Command | Behaviour |
|---|---|
| `docker compose up -d` (default) | Pulls `ghcr.io/caura-ai/caura-memclaw-*:latest` |
| `docker compose pull` | Refresh to the current published images |
| `docker compose build` | Build from local source (forks, dev) |

Tag is parameterised: `MEMCLAW_VERSION=v1.2.3` in `.env` pins a specific release. Defaults to `latest` for the most recent stable release.

## Compose verification

- `docker compose -f docker-compose.yml config --quiet` passes (compose recognises the dual image+build block).
- YAML still parses cleanly via Python yaml.
- `docker-compose.dev.yml` deliberately untouched — dev profile keeps build-only since iterating-on-source is the point.

## What stays the same

- Dockerfiles and source are identical between local-built and pulled images
- All env vars and services unchanged
- New users hitting Quick Start get the fast path automatically

## Sequencing note

This PR depends on [PR #17](https://github.com/caura-ai/caura-memclaw/pull/17) being merged + the v1.0.0 images actually being published to ghcr.io (manual workflow_dispatch with `tag=v1.0.0` and `push_latest=true`). Don't merge this PR until those images exist, or new `docker compose up` runs hit a missing image and fall back to building locally — which works but defeats the point.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #412, opened 2026-06-17
fix(extraction): re-fold split-off subject discriminators (A33)

arkash20:
## What

Entity extraction inconsistently **splits a trailing disambiguator off its subject**: `"Acme Corp #0033's …"` becomes a bare subject `acme corp` **+** a separate identifier `#0033` (`role=mentioned`). So every `Acme Corp #NNNN` collapses onto one shared `acme corp` subject entity. With `same_subject = (entity_id == entity_id)` (CAURA-133), that shared id is a **guaranteed false contradiction**, and the bare hub also **dilutes `entity_lookup`** (the A30 person-hub class). This is the dominant mechanism behind the V2 ~6.3% false-`conflicted` rate.

The #304 B-guard can't help here — the name being resolved is literally `acme corp` (no digits), which correctly matches the existing bare entity. The discriminator already left the subject upstream, in extraction.

## Fix

`_reattach_subject_discriminators` — a deterministic normalization on the `ExtractedGraph` **before any resolution**: re-fold a trailing disambiguator (a `#tag` or a `(qualifier)`) back into its subject when the content shows the two adjacent, remapping relation endpoints + mention links.

Guards: only `#`/parenthetical shapes qualify, and only when **content-adjacent** to the subject — so real named identifiers (`pr-2025-a`, `build-734`) and non-adjacent tags are left untouched.

**Scope**: forward-only (corrects new extractions, not already-baked corpora). Covers mechanism ① (the split). The cosine-merge of *already-distinct* qualified names (two `John Smith`s, `Acme (Delaware)` vs `(Ohio)`) is a separate mechanism, tracked as a follow-up.

## Tests

- **7 unit** (`test_a33_subject_discriminator.py`): fold, sibling-distinctness, relation/mention remap, parenthetical, + 3 negatives (non-adjacent, real named identifier, no-subject).
- **Wet-tested on the local stack** (real OpenAI extraction): two sibling writes resolve to **distinct** subject entities — `acme corp #0028` (active) vs `acme corp #0033` (confirmed), **neither marked `conflicted`**. Pre-fix these collapsed to one bare `acme corp` + a false conflict (V2 trace).
- Entity-extraction/resolution suite: 49 passed, no regression. ruff/format/mypy clean.

### DIFF
--- a/core-api/src/core_api/services/entity_extraction.py
+++ b/core-api/src/core_api/services/entity_extraction.py
@@ -77,6 +77,53 @@ def _fake_extract(content: str) -> ExtractedGraph:
     return ExtractedGraph(entities=entities, relations=[])
 
 
+# A33 (mechanism ①): the extractor inconsistently splits a trailing
+# disambiguator off its subject — "Acme Corp #0033's …" becomes subject
+# "acme corp" + a SEPARATE identifier "#0033" (role=mentioned) — so every
+# "<Name> #NNNN" collapses onto one bare "<Name>" subject entity. With
+# same_subject = (entity_id == entity_id) (CAURA-133) that shared id is a
+# guaranteed false contradiction, and the bare hub also dilutes entity_lookup
+# (A30). Re-fold a trailing disambiguator (a "#tag" or a "(qualifier)") back
+# into its subject when the content shows the two adjacent. Only "#"/paren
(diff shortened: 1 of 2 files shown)


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #413, opened 2026-06-17
refactor(plugin): configurable skill-reconcile targets (PR1 — refactor + config plumbing)

ran-taig:
First slice of "reconcile MemClaw skills into additional folders." **No-behavior-change refactor + config scaffolding only** — `owned` mode is implemented; `additive` mode is a follow-up (PR2).

## What changes
- Extract per-dir converge logic into **`reconcileOwnedDir(dir, desired)`** (read on-disk → prune orphans except `PROTECTED_SKILLS` → write/update → report installed). Owned-mode behaviour moved **verbatim**.
- `reconcileSkills()` fetches the catalog + builds `desired` **once**, then reconciles each configured target and aggregates into the unchanged flat `ReconcileSummary` (`installed` deduped+sorted across targets).
- New **`SkillTarget { dir, mode }` + `resolveSkillTargets()`**:
  - Always includes the plugin's own dir (`getPluginDir()/skills`, `owned`).
  - Extra targets via **`MEMCLAW_SKILL_TARGETS`** (JSON array, read at call time).
  - Parse **fails safe**: invalid JSON / non-array / malformed entry / bad mode / owned-dir duplicate → logged + ignored. A bad config can't crash the heartbeat or alter the owned dir.
- **`additive` targets are parsed but skipped with a warning** (not implemented yet) → a misconfigured additive dir is a no-op, never a destructive prune.

## Why a refactor-only PR
The reconciler's prune is destructive (`onDisk ∉ catalog → rmSync`, except `memclaw`). Before we point it at *shared* folders, the plumbing + the owned/additive split need to land cleanly so the additive (ownership-tracked, non-destructive) mode can slot in without touching the proven owned path.

## Invariants preserved (verified)
- **Default (env unset) = byte-identical** to before: one `owned` target = the plugin skills dir. All pre-existing reconcile tests pass unchanged.
- Empty-catalog / `memclaw`-protected / convergence / idempotency / bad-slug-skip / fail-open are unchanged (now inside `reconcileOwnedDir`).

## Tests (+10)
- `resolveSkillTargets`: default → single owned; valid JSON appends owned+additive; invalid/non-array JSON → owned-only; malformed entries skipped; owned-dir dup dropped.
- configured targets: an `additive` target is parsed but skipped and its foreign dir is left **completely untouched**; an extra `owned` target IS reconciled alongside the default.
- 323 plugin tests pass; tsc clean.

## Follow-ups (per plan)
- **PR2:** `additive` mode + per-skill ownership marker + same-slug collision policy.
- **PR3:** per-target heartbeat observability; optional auto-register targets into OpenClaw `settings.skills.customDirectories`.

### DIFF
--- a/plugin/src/reconcile-skills.test.ts
+++ b/plugin/src/reconcile-skills.test.ts
@@ -38,7 +38,7 @@ const originalHome = process.env.HOME;
 const tmpHome = mkdtempSync(join(tmpdir(), "reconcile-skills-test-home-"));
 process.env.HOME = tmpHome;
 
-const { reconcileSkills, PROTECTED_SKILLS } = await import("./reconcile-skills.js");
+const { reconcileSkills, PROTECTED_SKILLS, resolveSkillTargets } = await import("./reconcile-skills.js");
 
 const SKILLS_ROOT = join(tmpHome, ".openclaw", "plugins", "memclaw", "skills");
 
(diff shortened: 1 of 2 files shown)


ran-taig:
@claude please review. This is a **no-behavior-change refactor + config plumbing** (PR1 of the configurable-skill-targets plan). Key things to scrutinize: (1) the default path (env unset → single `owned` target) is byte-identical to the prior reconciler — confirm `reconcileOwnedDir` preserves orphan-prune + `PROTECTED_SKILLS` + installed-tracking + fail-open exactly; (2) `resolveSkillTargets` fails safe on every bad-config shape (invalid JSON / non-array / missing dir / bad mode / owned-dir dup) and can never crash the heartbeat or re-mode the owned dir; (3) `additive` targets are parsed but **skipped** — never destructively pruned (the safety property for PR2); (4) multi-target summary aggregation (dedupe+sort of `installed`) is correct.

ran-taig:
@claude

ran-taig:
Both findings were valid — applied:

**FIX 1 — dedupe `added`/`removed`/`protected` across targets.** They now match `installed`'s slug-level dedupe+sort (`[...new Set(...)].sort()`) — a slug touched in multiple targets counts once. Default single-target output is unchanged (one multi-element existing assertion already calls `.sort()`).

**FIX 2 — `resolve(dir)` instead of `join(dir)`** in `resolveSkillTargets`, and `resolve` added to the `path` import. Target dirs are now absolutized (relative to cwd), which is correct for both the dedupe-vs-`ownedDir` (absolute) check and predictable filesystem writes.

323 plugin tests pass; tsc clean.

@claude please re-review.

ran-taig:
Addressed:

**Finding A (valid, applied):** `resolveSkillTargets` now seeds the dedup set + `targets[0]` with `resolve(ownedSkillsDir())` instead of the raw `ownedSkillsDir()`, so the owned-dir seed and the `resolve()`d entries share one canonical representation — a non-canonical owned path (e.g. with `..`) can no longer let a same-dir entry slip past the seen-set and reconcile the owned dir twice.

**Findings B & C — already in the branch (from the prior review round):**
- B: line 159 is already `const normalized = resolve(dir);`.
- C: the loop already aggregates via `addedAll`/`removedAll`/`protectedAll` collectors and assigns `summary.{added,removed,protected,installed} = [...new Set(...)].sort()`. All four arrays are deduped+sorted consistently.

323 plugin tests pass; tsc clean.

@claude please re-review.

ran-taig:
Both applied:

**Stale log prefixes** — the three log lines inside `reconcileOwnedDir` (read-dir / rm / write failures) now use the `reconcileOwnedDir:` prefix. The catalog-query log stays `reconcileSkills:` since it's still in the parent.

**Shallow-path guard** — `resolveSkillTargets` now rejects too-shallow target dirs (`< 2` path segments, e.g. `/`, `/tmp`) with a warning + skip, before the dedupe/push — preventing a misconfigured shallow target from driving owned-mode `rmSync`. The default owned dir is seeded before this loop, so it's unaffected. Added a test (`too-shallow target dirs … skipped`).

324 plugin tests pass; tsc clean.

@claude please re-review.

ran-taig:
@claude

ran-taig:
@erni-a

erni-a (approved):

PR #414, opened 2026-06-17
fix(core-api): stamp version into image so /api/v1/version stops serving "dev" (CAURA-000)

erni-a:
## Summary

`/api/v1/version` (and the `version` field of `/api/v1/status`) returns `"dev"` instead of the real release version.

**Root cause:** the Dockerfile installs dependencies with `uv export --frozen --no-emit-project`, which deliberately excludes the project itself — so **core-api is never installed as a package** in the image. `core_api.constants.VERSION` resolves via `importlib.metadata.version("core-api")`, which raises `PackageNotFoundError`, and the silent `except` fallback returns `"dev"`. Confirmed by inspecting the built image: `pip show core-api` → not found, and the module loads from source via `PYTHONPATH`.

This is invisible in CI (which installs core-api editable, so metadata resolves) — the bug only exists inside the built image.

## Fix

Bake the version into the image at build time and read it deterministically:

- **Dockerfile** — stamp `/app/VERSION` from `pyproject.toml` (the source of truth) in the builder, with an optional `MEMCLAW_VERSION` build-arg override (e.g. a release tag).
- **constants.py** — `_resolve_version()` precedence: `MEMCLAW_VERSION` env → `/app/VERSION` file → installed-package metadata → `"dev"`. Independent of install method / `PYTHONPATH` ordering.
- **.dockerignore** — add recursive `**/*.egg-info` (a bare `*.egg-info` only matches the context root) so a stray nested `core_api.egg-info` from a local editable install can't bake into the image and shadow the stamp.
- **ci.yml** — new `image-version` job that builds the image and asserts the baked version is a real semver matching `pyproject.toml` (not `"dev"`). A pytest assertion can't catch this; the job builds the actual image.
- **tests** — version-resolution precedence guards in `test_api_status.py`.

## Test plan

- [x] `docker build` of the image → `VERSION = <pyproject version>` (was `"dev"`)
- [x] Verified `pip show core-api` is absent in the image — the stamp is the sole source
- [x] `ruff check` / `ruff format --check` clean; `mypy` clean
- [x] `pytest tests/test_api_status.py -k version` — 3 passed
- [x] New `image-version` CI job fails loudly if the baked version is `"dev"` or drifts from `pyproject.toml`

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -7,6 +7,11 @@ node_modules
 memclaw/node_modules
 memclaw/.next
 *.egg-info
+# Bare globs only match the context root — add a recursive form so a stray
+# nested build artifact (e.g. core-api/src/core_api.egg-info from a local
+# editable install) can't bake into the image and shadow the version stamp.
+**/*.egg-info
+**/__pycache__
 .pytest_cache
 .mypy_cache
 .venv
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -144,3 +144,39 @@ jobs:
         run: |
           cd core-api && uv build && cd ..
           cd core-storage-api && uv build
+
+  image-version:
+    # Guards the /api/v1/version surface end-to-end: builds the actual
+    # core-api image and asserts it bakes a real semver (== pyproject),
+    # not "dev". This is the regression that silently shipped for weeks —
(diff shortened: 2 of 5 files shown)


erni-a:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #415, opened 2026-06-17
fix(core-api): self-heal storage pool + cancellation-safe requests (CAURA-000)

ran-taig:
## Why

Root cause of the **2026-06-16 core-api pool-starvation SEV-2** on eToro on-prem (~2h recall/write stall). `CoreStorageClient` holds a process-wide httpx pool to core-storage (`max_connections=200`, `pool=5s`). When an in-flight call was cancelled mid-request — the 35s enrichment `wait_for` or the 45s request timeout — httpcore (1.0.9, current) did **not** return the connection to the pool. Leaked slots accumulated over ~4.5 days of uninterrupted uptime until every acquire hit `PoolTimeout`; the singleton never self-healed → **restart-only recovery**.

Forensics (incident addendum): there was **no acute trigger** that morning — the prior day had *more* slow upstream calls and *more* timeouts yet survived; the leak simply crossed the 200-slot ceiling at 07:37. So this recurs on a cadence set by the cancellation rate until the pool defect is fixed — which is what this PR does.

## What

Two layered fixes in `CoreStorageClient`, both routed through a single `_execute()` funnel that all six request helpers (`_get`/`_get_list`/`_post`/`_post_optional`/`_patch`/`_delete`) **and** the two direct call sites (`list_keystones`, `claim_idempotency`) now use:

- **Fix B — prevention (`_cancel_safe`)**: shields the in-flight request via `asyncio.shield` so a cancelled caller still lets the request finish and release its slot. The caller observes `CancelledError` immediately; the orphaned result is consumed so it isn't logged as never-retrieved. Bounded by the existing 120s read timeout, so worst case it self-releases instead of leaking forever.
- **Fix A — recovery (`_recycle_pools`)**: on a `PoolTimeout` that survives the retry policy (i.e. genuine exhaustion), rebuild the pool **once** under a lock with a generation guard (concurrent `PoolTimeout`s collapse to one rebuild), background-close the old pool, and retry on the fresh one. Converts a restart-only outage into a **seconds-long self-heal** and emits an `ERROR` log (also closes the observability gap from the incident).

Helpers read `self._http`/`self._read_http` **lazily** so a recycle is picked up on retry. No behavioural change on the happy path or to existing retry semantics.

## Tests

New `tests/test_storage_client_pool_recovery.py`:
- cancellation mid-request lets the shielded call complete (no leak)
- POST + GET recycle the pool on exhaustion, then succeed on the fresh pool
- a transient `PoolTimeout` ridden out by retries does **not** recycle (no churn)

The existing F5 / 2026-06-11 retry suite (`test_storage_client_retry.py`, 19 tests) passes unchanged — the `_execute` refactor preserves all retry behaviour. `ruff check`, `ruff format --check`, and `mypy` are clean.

## Tradeoff

A write cancelled by the 45s budget now completes in the background rather than being abandoned — benign, because writes are idempotency-keyed (`client_request_id`) and reads are side-effect-free.

## Rollout

- Targets `caura-memclaw`; rides the next on-prem release. **Do not merge without staging on erni-onprem first** per the 5-phase upgrade protocol, then eToro.
- Rename `CAURA-000` to the real ticket before merge.
- Out of scope (separate, fill-rate-only PRs): the `cluster_id` ValidationError fix and 429-aware backoff — they reduce how fast the pool fills, not the leak itself.

### DIFF
--- a/core-api/src/core_api/clients/storage_client.py
+++ b/core-api/src/core_api/clients/storage_client.py
@@ -2,7 +2,9 @@
 
 from __future__ import annotations
 
+import asyncio
 import logging
+from collections.abc import Awaitable, Callable
 from typing import Any, Literal, NotRequired, TypedDict
 
 import httpx
@@ -96,6 +98,16 @@ class CoreStorageClient:
             # against a single service for no benefit.
             self._read_http = self._http
 
+        # Pool self-healing (incident 2026-06-16): the singleton pool could
+        # leak connection slots when in-flight calls were cancelled and never
+        # recovered without a process restart. ``_pool_generation`` lets
+        # concurrent ``PoolTimeout``s collapse into exactly one rebuild.
+        self._pool_lock = asyncio.Lock()
+        self._pool_generation = 0
+        # Anchor fire-and-forget pool-close tasks so they aren't GC'd while
+        # pending (Python 3.12+ warns otherwise); discard on completion.
+        self._background_tasks: set[asyncio.Task] = set()
+
     @staticmethod
     def _make_pool() -> httpx.AsyncClient:
         """Construct an httpx pool with the tuned timeouts + limits.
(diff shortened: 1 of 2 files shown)


ran-taig:
@claude

ran-taig:
@claude please re-review — applied the review feedback and force-pushed (amended onto the rebased tip).

Changes since last review:
- `_cancel_safe`: replaced the silent done-callback lambda with a named `_consume_result` that still consumes the orphaned task's exception (so it's never logged as "never retrieved") but now also logs it at `debug` for visibility.
- `asyncio.ensure_future(...)` → `asyncio.create_task(...)` in both `_cancel_safe` (the shielded request) and `_recycle_pools` (the background pool close) — both are always coroutines invoked within a running loop, so `create_task` is the correct, more idiomatic call.

No behavioural change to the fix itself. Re-verified locally: full pool-recovery suite + the existing retry suite (23 passed), `ruff check`, `ruff format --check`, and `mypy` all clean.

ran-taig:
@claude

ran-taig:
@claude one more revision pushed (amended).

- Added a `_background_tasks: set[asyncio.Task]` on the client and anchored the fire-and-forget pool-close tasks in `_recycle_pools` to it (`add` + `add_done_callback(...discard)`). This keeps them strongly referenced while pending so they aren't GC'd mid-close and don't trigger the Python 3.12+ "Task was destroyed but it is pending" / create_task-reference DeprecationWarnings.

No behavioural change. Re-verified: 23 tests pass (pool-recovery + retry suites), `ruff check`, `ruff format --check`, `mypy` all clean.

ran-taig:
@erni-a

erni-a (approved):

PR #416, opened 2026-06-17
fix(core-api): harden business/personal pre-gate (timeout, audit truth, fail-closed, audit durability)

Eldad-Caura:
Hardens the opt-in business/personal pre-gate ([#403](https://github.com/caura-ai/caura-memclaw/pull/403)) against five findings. **Opt-in throughout — no behavior change for tenants that haven't enabled it; `fail_closed` defaults to false (current fail-open preserved).**

## Findings addressed

| # | Finding | Fix |
|---|---------|-----|
| **1** | Classifier ran inline on the write path with **no whole-chain timeout** — a slow/unreachable provider could stall a write ~100s (retries × fallback). | Hard ceiling `PREGATE_CLASSIFIER_TIMEOUT_SECONDS` (env-tunable, default **5s**) via `asyncio.wait_for`; on timeout fail-open + alertable warning. `call_with_fallback`'s `timeout=` is per-attempt, so it can't bound the chain — hence the outer wrap. |
| **2** | `min_confidence` does **not** reduce false rejects as documented — the post-enrichment backstop drops any `"personal"` verdict unconditionally (no enrichment confidence exists to gate on). | Doc-only: corrected the settings comment + `_GovNBPregate` docstring to describe the real behavior (gates the *early* drop only; defers borderline content to the accurate backstop). |
| **3** | Under provider fallback the audit recorded the **intended** provider/model, not the one that actually classified. | Capture the live `provider`/`model` on `BusinessClassification` (from the `LLMProvider` instance) and audit those. |
| **4** | A `drop` tenant whose classifier is unavailable **silently stored** unclassified content. | Detect it (`result.provider is None`) + alertable warning; add opt-in `pregate.fail_closed` (default false) that rejects the write (**503** + `nonbusiness_pregate_unavailable` audit) instead of storing unclassified content. |
| **5** | Governance enforcement audits could be **dropped under audit-queue overflow** while the write was still refused — contradicting `emit_governance_audit`'s "can't be silently disabled" promise. | `log_action(critical=True)`: on overflow, critical events fall back to a synchronous storage write instead of dropping; `enqueue()` now reports whether it queued. Applied to **every** reject/destructive-enforcement audit (pregate drop + unavailable, deterministic + LLM PII drop, non-business drop, fast-mode remediation soft-deletes, bulk drop). Flag/mask/keep-private stay best-effort. |

## Decisions
- **Finding 2** is doc-only because the enrichment signal carries no confidence (`business_relevance: str` in `common/enrichment/schema.py`), so wiring a confidence floor into the backstop would need a broad enrichment-schema change affecting all tenants.
- **Finding 4** `fail_closed` setting lives in OSS `organization_settings` `DEFAULT_SETTINGS` (no split). Enterprise "Content Policy" UI toggle is a follow-up in the enterprise repo.
- **Finding 5** `critical` is scoped to reject/destructive paths (audit is the only record); high-volume flag/keep-private paths keep the fire-and-forget queue, so the hot path is unaffected.

## Verification
- New/updated tests: `tests/test_business_personal_pregate.py` (timeout→fail-open, fallback audit attribution, fail-open observability default, `fail_closed` 503) and `tests/test_audit_queue.py` (`enqueue` bool contract, critical sync-fallback on overflow). All green.
- `ruff check` + `ruff format --check` clean (core-api/src, core-storage-api/src); `mypy` clean on **both** core-api and core-storage-api.
- The DB-backed `test_governance_e2e.py` / `test_audit_chain.py` integration suites fail identically on clean `main` in this sandbox (no database) — they'll exercise these changes under CI's DB.

### DIFF
--- a/common/llm/constants.py
+++ b/common/llm/constants.py
@@ -83,6 +83,20 @@ def _read_float_env(name: str, default: float | None) -> float | None:
 OPENAI_REQUEST_TIMEOUT_SECONDS = _read_float_env("OPENAI_REQUEST_TIMEOUT_SECONDS", 25.0)
 
 
+# Hard ceiling on the *whole* business/personal pre-gate classification —
+# across retries and any provider fallback — enforced by the classifier itself
+# (``classify_business_personal`` wraps the call in ``asyncio.wait_for``). The
+# pre-gate is a fast, fail-open go/no-go that runs inline on the write path
+# BEFORE the row is written, so a slow or unreachable provider must never stall
+# a write: exceeding this ceiling fails open (the post-enrichment gate remains
+# the backstop). Deliberately aggressive — much tighter than the per-call SDK
+# read timeout above — because the cost of a too-low value is only a missed
+# early-reject, never a blocked write. Env-tunable per deployment.
+PREGATE_CLASSIFIER_TIMEOUT_SECONDS = _read_float_env(
+    "PREGATE_CLASSIFIER_TIMEOUT_SECONDS", 5.0
+)
+
+
 # ── httpx pool sizing for OpenAI-compatible providers ────────────────
 #
 # CAURA-627: the SDK rides httpx's default (100 max_connections / 20
--- a/core-api/src/core_api/pipeline/steps/write/business_personal_pregate.py
+++ b/core-api/src/core_api/pipeline/steps/write/business_personal_pregate.py
@@ -15,6 +15,8 @@ a classifier failure or a ``none`` provider never blocks a write.
 
 from __future__ import annotations
 
+import logging
+
 from fastapi import HTTPException
 
 from core_api.pipeline.context import PipelineContext
@@ -22,9 +24,12 @@ from core_api.pipeline.step import StepOutcome, StepResult
 from core_api.services.business_classifier import classify_business_personal
 from core_api.services.governance_gate import (
     ACTION_NB_PREGATE_DROP,
+    ACTION_NB_PREGATE_UNAVAILABLE,
     emit_governance_audit,
     nonbusiness_pregate_audit_detail,
 )
+
+logger = logging.getLogger(__name__)
 
 
 class BusinessPersonalPregate:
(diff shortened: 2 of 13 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
Addressed in \`3eae486\` (force-pushed, single commit).

**Dropped counter inflates for critical events recovered via sync fallback** — fixed. \`AuditEventQueue.enqueue\` now takes a \`silent\` flag and \`log_action\` passes \`silent=critical\`: when a critical event hits a full queue it returns \`False\` **without** incrementing \`_dropped_count\` or logging a "dropped" warning — the synchronous-write fallback recovers it, so it's no longer miscounted as a loss or logged as lost. Non-critical drops still count + warn exactly as before.

Tests: added \`test_enqueue_silent_drop_does_not_count_or_warn\`; \`test_log_action_critical_falls_back_to_sync_on_full_queue\` now also asserts \`silent=critical\` is threaded through. ruff check + format and mypy (core-api + core-storage-api) green; audit-queue + pre-gate suites pass.

Eldad-Caura:
@claude

Eldad-Caura:
@claude

Eldad-Caura:
Addressed in \`d08a31c\` (force-pushed, single commit, rebased onto the current base).

- **\`critical=True\` sync fallback could skip \`soft_delete_memory\` (Medium)** — fixed at the chokepoint: \`log_action\`'s critical sync-fallback now **logs-and-swallows** on failure instead of propagating, so a failing audit can't block the enforcement action it precedes (a 4xx reject, or a soft-delete that would otherwise leave the row the policy means to drop). Done in \`log_action\` rather than per-call-site so it covers all critical paths (remediation, decision, scan, bulk) once. The queue-inactive sync POST is **deliberately left raising** (it's the primary write for *all* events in kill-switch/startup mode — swallowing there would be a silent audit blackout); added a guard comment. New test \`test_log_action_critical_sync_failure_does_not_propagate\`.
- **Warning factually wrong when \`fail_closed=True\` (Low)** — split into two branches: fail-closed logs \"REJECTING write\", fail-open logs \"content stored WITHOUT enforcement\". Test asserts the correct message per branch.
- **\`except TimeoutError\` vs \`asyncio.TimeoutError\` (Medium)** — kept the builtin \`except TimeoutError\`. This project is Python 3.12 where \`asyncio.TimeoutError IS TimeoutError\`, and **\`ruff UP041\` enforces the builtin spelling** (switching to \`asyncio.TimeoutError\` fails lint). Added a comment so it isn't re-flagged.

ruff check + format and mypy green on core-api + core-storage-api; pre-gate + audit-queue suites pass.

Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #417, opened 2026-06-17
feat(plugin): additive-mode skill reconcile with ownership marker (PR2)

ran-taig:
## What

PR1 (#413) made reconcile targets configurable but left **`additive`** mode a warn-skip no-op. This PR implements it: MemClaw can now reconcile its active skills into **shared/foreign** target dirs **without ever deleting or overwriting an entry it does not own**.

## Mechanism — per-skill ownership marker

A sentinel file `<dir>/<slug>/.memclaw-owned` stamps every skill MemClaw writes into an `additive` dir. `reconcileAdditiveDir`:

- **write** — a desired slug is written only when its dir is **absent** (new → stamp marker) or **already MemClaw-owned** (update in place). A slug occupied by an **unowned** dir is a **collision** → skipped, never overwritten (reported in `summary.skipped`).
- **remove** — an on-disk slug not in the catalog is removed **only if it carries the marker**. Unowned (foreign) entries are left untouched.

This is **fail-safe**: a missing marker means *"leave it alone"*, never *"delete it"*. So an empty catalog — or a misconfigured tenant returning an empty installable set — prunes only MemClaw's own entries in additive dirs. The destructive "wipe the dir" hazard stays confined to **`owned`** dirs.

The marker lives **inside** the skill dir; OpenClaw's loader reads only `SKILL.md` and halts recursion at a skill root, so the marker is invisible to the agent (verified against agent-core `skills.ts` / `session.ts`).

## Changes

- `OWNED_MARKER` + `isMemclawOwned` helpers.
- `DirReconcileResult.skipped` (additive collisions); `owned` dirs always return it empty.
- `reconcileAdditiveDir` — mirrors `reconcileOwnedDir`'s shape with marker-gated removal + collision-skip writes.
- `reconcileSkills` loop dispatches `additive` → the new path; `summary.skipped` now merges bad-shape catalog rows + collisions (deduped + sorted, consistent with the other arrays).
- `reconcileOwnedDir` and the default single-target behavior are **unchanged**.
- Docs (`mcp-skill-delivery.md`) section describing `owned` vs `additive` (collision/never-remove/marker semantics).

## Tests

5 new `additive` tests: write + marker stamp; foreign skill never removed (incl. empty catalog); collision → skipped + foreign bytes untouched; owned orphan removed (foreign neighbour survives); owned skill updated in place. Replaces the PR1 "additive not implemented" placeholder. **328 plugin tests pass; tsc clean.**

### DIFF
--- a/docs/mcp-skill-delivery.md
+++ b/docs/mcp-skill-delivery.md
@@ -173,6 +173,29 @@ malformed catalog row was skipped — closing the loop from "approved" to
 tick (even steady-state ticks with empty `added`/`removed`), so a node's
 current live-skill set is always legible.
 
+### Reconcile targets: `owned` vs `additive`
+
+By default the reconciler manages one **`owned`** dir — the plugin's own
+`skills/` — where it has full authority: anything on disk not in the
+catalog is pruned. Operators can add extra target dirs via the
+`MEMCLAW_SKILL_TARGETS` env var (JSON array of `{ dir, mode }`):
+
+- **`owned`** — fully MemClaw-managed (destructive prune). Use only for
+  dirs MemClaw exclusively controls.
+- **`additive`** — a **shared/foreign** dir. MemClaw writes its active
+  skills there but **only ever touches entries it wrote**, tracked by a
+  per-skill `.memclaw-owned` marker file:
+  - a slug already occupied by an *unowned* skill is a **collision** —
+    skipped, never overwritten (reported in `skipped`);
+  - an unowned skill is **never removed**, even on an empty catalog;
+  - only marker-bearing entries are updated/pruned.
+
+This makes the "empty catalog / wrong tenant wipes the dir" hazard apply
+only to `owned` dirs — `additive` dirs lose only MemClaw's own entries,
+never the client's. For an additive target to actually reach agents it
+must be on OpenClaw's skill load path (e.g. registered in
+`settings.skills.customDirectories`).
+
 ## What makes a skill findable: the summary
 
 Scoped `op=search` ranks on the embedded `data.summary`. So the
--- a/docs/mcp-skill-delivery.md
+++ b/docs/mcp-skill-delivery.md
@@ -186,7 +186,8 @@ catalog is pruned. Operators can add extra target dirs via the
   skills there but **only ever touches entries it wrote**, tracked by a
   per-skill `.memclaw-owned` marker file:
   - a slug already occupied by an *unowned* skill is a **collision** —
-    skipped, never overwritten (reported in `skipped`);
+    skipped, never overwritten (reported in the summary's `collisions`
+    list, kept distinct from the catalog-shape `skipped` list);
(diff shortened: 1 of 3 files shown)


ran-taig:
@claude please re-review

ran-taig:
@claude

ran-taig:
Both review comments addressed in dad57903:

1. **TOCTOU clobber** — `reconcileAdditiveDir` now re-checks liveness at write time (`dirExistsNow = onDisk.has(slug) || existsSync(skillDir)`) for the collision guard, marker stamp, unchanged-content skip, and log distinction. A foreign dir created after the initial `readdirSync` snapshot can no longer slip past the guard and get clobbered/stamped.
2. **`skipped` conflation** — split into a distinct `collisions` field: `DirReconcileResult.collisions` + `ReconcileSummary.collisions` (aggregated from additive dirs only); `skipped` stays catalog-shape errors only. The whole summary already rides the heartbeat, so `collisions` surfaces on `/fleet/nodes` with no backend change.

Tests + docs updated (collision test asserts the slug is in `collisions` and **not** `skipped`). 328 plugin tests pass; tsc clean.

@claude please re-review

ran-taig:
Round-3 review addressed in 8ca583a1 — all four correctness items + the DRY suggestion:

1. **TOCTOU actually closed** — replaced the idempotent `mkdirSync({recursive:true})` with a non-recursive `mkdirSync` that throws `EEXIST` (atomic create-or-detect). On EEXIST we re-verify ownership and report a collision if the racing dir is foreign. The `existsSync` re-stat now only *narrows*; mkdir is the real guard (comment corrected).
2. **Marker self-heal** — `isNew || !existsSync(marker)` re-stamps a marker that vanishes mid-pass (defensive; a marker absent at guard time is already a foreign collision).
3. **`dirExistsNow` uses live `existsSync(skillDir)` only** — no longer trusts the stale start-of-function `onDisk` snapshot, so a since-deleted owned dir isn't misread as a foreign collision.
4. **`installedSet` pre-seeded** with on-disk + owned + catalog-active slugs (mirrors `reconcileOwnedDir`), so a transient I/O failure can't drop a present skill from `installed`.

Plus the suggested **`readDirSlugs(root, caller)` helper** — extracted the duplicated readdir/stat block from both reconcile functions (returns null → caller bails). No behavioral change.

+1 test (marker-deleted dir → treated as foreign collision, never clobbered). **329 plugin tests pass; tsc clean.**

@claude please re-review

ran-taig:
Round-4 addressed in 4fab910e — ownership now gates the protected check in the additive removal loop. A foreign dir whose name collides with a protected slug (e.g. a client's own `memclaw` dir) is left alone and no longer misreported in `protected`; an OWNED protected dir still survives.

+2 tests (foreign protected-name dir ignored; owned protected-name dir survives + reported). **331 plugin tests pass; tsc clean.**

@claude please re-review

ran-taig:
@erni-a

erni-a (approved):

PR #418, opened 2026-06-17
feat(core-api): cross-worker settings-cache invalidation (CAURA-571)

Eldad-Caura:
Closes the cross-worker settings-cache staleness gap (the PII findings' "5-min cross-worker settings staleness" — confirmed live).

## Problem
The per-process org-settings cache (`TTLCache`, 5-min TTL in `organization_settings.py`) only invalidates **locally** on a write — `invalidate_cache` evicts the serving worker only. Sibling uvicorn workers and other Cloud Run instances keep serving the **stale** value for up to the TTL. For a governance control that's a correctness gap: flipping a policy to a **stricter** action (e.g. PII `flag → drop/mask`) has an up-to-5-minute window where some workers still apply the looser one.

## Fix — fan out invalidation over the event bus
- New **broadcast** topic `Topics.Org.SETTINGS_CHANGED` + `OrgSettingsChangedEvent` payload (carries only `org_id` — nothing to keep in sync, no secret on the wire).
- `update_settings` publishes it after `commit()` (**best-effort**: a publish failure can't fail an already-committed write — siblings then fall back to the TTL, exactly as today).
- `core-api` consumer `handle_org_settings_changed` → `invalidate_cache(org_id)`.

### Broadcast vs work-queue (the crux)
The existing bus names subscriptions `{prefix}--{topic}` per **service**, so every core-api process shares one subscription → **competing consumers**: each message reaches exactly **one** worker. That's correct for the existing enrich/embed/suppression consumers, but wrong for cache invalidation, which needs **every** process to act.

So this adds `EventBus.subscribe(broadcast=True)`:
- **InProcessEventBus**: no-op (a single process already dispatches to all local handlers).
- **PubSubEventBus**: gives a broadcast topic a **per-process ephemeral subscription** (`{prefix}--{topic}--{instance_id}`), created at `start()` with an `expiration_policy` backstop (a crashed process's sub self-reaps after 1 day) and deleted at `stop()`. Create-failure (e.g. missing IAM) **degrades to TTL-only invalidation** rather than crashing startup; the consumer is idempotent + TTL-backstopped.

Honors the standing rules: no new direct DB pool (reuses the event bus), no ad-hoc migration.

## ⚠️ Deploy dependency — IAM grant required (infra)
Durable subscriptions are Terraform-provisioned, but per-process ephemeral subscriptions embed a runtime `instance_id`, so they're created at runtime. The **core-api service account** needs subscription create/delete. Until this lands, core-api boots fine and invalidation falls back to the cache TTL (no regression).

```hcl
# Grant the core-api runtime SA permission to create/delete its own
# ephemeral broadcast subscriptions (CAURA-571).
resource "google_pubsub_topic_iam_member" "core_api_settings_changed_subscriber" {
  topic  = google_pubsub_topic.memclaw_org_settings_changed.name  # new topic
  role   = "roles/pubsub.editor"   # or a custom role with
                                   # pubsub.subscriptions.{create,delete,consume}
  member = "serviceAccount:${var.core_api_service_account_email}"
}
```
Also provision the topic `memclaw.org.settings-changed` (mirroring the other `memclaw.*` topics; the per-env `EVENT_BUS_TOPIC_PREFIX` applies as usual). I can adjust to a tighter custom role if you prefer least-privilege over `pubsub.editor`.

## Verification
- New tests: `test_pubsub_bus.py` (per-process sub create / already-exists / degrade-on-failure / stop-delete), `test_inprocess_bus.py` (broadcast no-op dispatch), `test_cross_worker_settings_invalidation.py` (handler + payload + broadcast wiring); `test_consumer_enriched.py` register assertion updated. All green.
- `ruff check` + `ruff format --check` and `mypy` clean on core-api + core-storage-api.
- (Pre-existing DB-backed settings/e2e suites need a database — not runnable in this sandbox; they exercise the publish path under CI's DB.)

### DIFF
--- a/common/events/base.py
+++ b/common/events/base.py
@@ -66,7 +66,9 @@ class EventBus(ABC):
         """
 
     @abstractmethod
-    def subscribe(self, topic: str, handler: EventHandler) -> None:
+    def subscribe(
+        self, topic: str, handler: EventHandler, *, broadcast: bool = False
+    ) -> None:
         """Register *handler* as a subscriber to *topic*. May be called
         at startup only — not thread-safe during `publish`.
 
@@ -75,6 +77,16 @@ class EventBus(ABC):
         handler raises, gets redelivered. Use `event.event_id` as a
         natural dedup key when the operation isn't inherently
         idempotent.
+
+        ``broadcast``: when True, *every* subscribing process must receive
+        each event (fan-out), not just one. The default (False) is the
+        work-queue semantics every existing consumer relies on — a shared
+        per-service subscription delivers each message to a single
+        process. Broadcast is for cross-process cache invalidation, where
+        one worker handling the event would leave the others stale. The
+        in-process bus dispatches to all local handlers regardless, so it
+        ignores the flag; the Pub/Sub bus gives broadcast topics a
+        per-process subscription.
         """
 
     async def start(self) -> None:
--- a/common/events/inprocess.py
+++ b/common/events/inprocess.py
@@ -33,7 +33,12 @@ class InProcessEventBus(EventBus):
         self._handlers: dict[str, list[EventHandler]] = defaultdict(list)
         self._tasks: set[asyncio.Task[None]] = set()
 
-    def subscribe(self, topic: str, handler: EventHandler) -> None:
+    def subscribe(
+        self, topic: str, handler: EventHandler, *, broadcast: bool = False
+    ) -> None:
+        # ``broadcast`` is a no-op here: a single-process bus already
+        # dispatches every event to all local handlers (fan-out within the
+        # process). Cross-worker fan-out is a Pub/Sub-backend concern.
         self._handlers[topic].append(handler)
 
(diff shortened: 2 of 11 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@claude

Eldad-Caura:
Addressed in `b36f420` (force-pushed, rebased onto current base). Corrected the `AlreadyExists` comment in `_ensure_broadcast_subscription`: it no longer claims "expiration_policy is reset" (the create failed, so it isn't) — now states the TTL continues from original creation and the pull loop's ongoing activity is what keeps Pub/Sub from expiring the subscription. Comment-only; ruff clean, pubsub tests pass.

Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #420, opened 2026-06-17
ci: add static security scanning (bandit SAST + pip-audit)

Eldad-Caura:
## What
Adds a **Security Scan** workflow — **bandit** (Python SAST) + **pip-audit** (dependency CVEs) over `core-api` / `core-storage-api` source — running per-PR + weekly.

## Why
`ci.yml` already gates **ruff (lint) + mypy (types) + pytest** every PR, and `claude_code_review.yml` adds an LLM correctness review. The one layer nothing scanned is **security/SAST + known-vulnerable dependencies** — this closes that gap.

## Notes
- **Report-only for now** (`continue-on-error`) so the existing backlog doesn't block merges — findings surface in the job log. Drop `continue-on-error` to make it a hard gate once triaged. `bandit` is filtered to medium+ severity/confidence to keep noise down.
- Reuses the pinned `actions/checkout` + `astral-sh/setup-uv` SHAs and the `uv` setup from `ci.yml`. No new GHAS/Advanced-Security requirement.
- **Optional follow-up:** this repo is public, so **GitHub CodeQL** (deeper SAST, covers TypeScript too) is free — I can add `.github/workflows/codeql.yml` next if you want it.

_Draft — proposal for review; tune the scope / severity thresholds / blocking-vs-reporting before merge._

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #421, opened 2026-06-17
test: de-flake scope_agent search test (synchronous write_mode)

Eldad-Caura:
## What
De-flakes `test_rest_search_scope_agent_uses_authenticated_identity`.

## Root cause
The test writes a `scope_agent` memory then immediately `POST /api/v1/search` for it. The write→search index (embedding) is **eventually consistent**, so the search could transiently return `[]` and the "author sees own row" assertion failed intermittently — this is the flake that reddened CI on the security-scan PR #420 (which touched no source).

## Fix
Write with **`write_mode="strong"`** so embedding/indexing is synchronous and the row is searchable immediately — the same deterministic pattern `test_a13_fleet_less_visible_to_fleet_scoped_search` documents and uses. `_write` gains an optional `write_mode` (default unset), so the by-id/list tests (which read from the DB directly, not search) are unaffected.

As a bonus, the negative check (bob can't see it) is now meaningful — the row is guaranteed indexed when bob searches, rather than passing vacuously on an empty index.

### DIFF
--- a/tests/test_memory_byid_authz.py
+++ b/tests/test_memory_byid_authz.py
@@ -225,19 +225,22 @@ def as_agent(monkeypatch):
     _app.dependency_overrides.pop(_gac, None)
 
 
-async def _write(client, headers, tenant_id, *, agent_id, fleet_id, visibility, content=None):
-    resp = await client.post(
-        "/api/v1/memories",
-        json={
-            "tenant_id": tenant_id,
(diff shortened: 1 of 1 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #424, opened 2026-06-18
feat(plugin): per-target reconcile observability (PR3a)

ran-taig:
## What

Third phase of the configurable-skill-targets track (after [#413](https://github.com/caura-ai/caura-memclaw/pull/413) PR1 and [#417](https://github.com/caura-ai/caura-memclaw/pull/417) PR2). This is **PR3a** — the observability half. The auto-register-into-OpenClaw half (**PR3b**) is deferred until the OpenClaw `customDirectories` key path + reload behavior are confirmed against OpenClaw source.

The reconcile summary aggregates `installed` / `added` / `removed` / `collisions` / `protected` across **all** configured targets (deduped + sorted). With more than one target — an extra `owned` dir, or an `additive` shared dir — an operator can't tell **which** dir a skill landed in, or collided in.

## Change

`ReconcileSummary` gains a `targets[]` field — one `TargetReconcileResult` per dir reconciled this tick:

```ts
{ dir, mode, installed, added, removed, collisions, protected }
```

in `resolveSkillTargets()` order (owned dir first), each array sorted. The existing top-level arrays are **unchanged** (still the cross-target dedup), so this is purely additive.

The whole summary already rides the heartbeat, and the backend stores `nodes.metadata.reconcile` as an opaque blob — so `targets[]` surfaces on `GET /api/v1/fleet/nodes` with **no backend change**.

Also fixes a stale doc comment on `SkillTargetMode` that still claimed `additive` was "parsed but NOT yet acted on … skipped with a warning" — PR2 implemented it.

## Tests

+2: default single target → one entry mirroring the aggregate; owned + additive each report their own slice while the aggregate dedups (install from owned, collision from additive). Docs updated. **333 plugin tests pass; tsc clean.**

### DIFF
--- a/docs/mcp-skill-delivery.md
+++ b/docs/mcp-skill-delivery.md
@@ -162,16 +162,24 @@ an agent may see.
 Gating decides what *should* land; observability confirms what *did*.
 Each heartbeat carries the reconciler's summary — `installed` (the
 active skills converged onto that node's disk right now), plus per-tick
-`added` / `removed` / `skipped` / `protected` deltas — which the backend
-stores as the latest snapshot at `nodes.metadata.reconcile` and surfaces
-via `GET /api/v1/fleet/nodes`. So an operator who flips a skill to
-`active` can confirm it actually reached each node, and see *why* a
-malformed catalog row was skipped — closing the loop from "approved" to
-"installed on the fleet."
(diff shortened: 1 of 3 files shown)


ran-taig:
@claude please review

ran-taig:
@erni-a

erni-a (approved):

PR #425, opened 2026-06-18
feat(plugin): auto-register additive skill dirs on OpenClaw's load path (PR3b)

ran-taig:
## What

Final phase of the configurable-skill-targets track ([#413](https://github.com/caura-ai/caura-memclaw/pull/413) PR1 · [#417](https://github.com/caura-ai/caura-memclaw/pull/417) PR2 · [#424](https://github.com/caura-ai/caura-memclaw/pull/424) PR3a).

An `additive` target dir is reconciled to disk, but — unlike the plugin's own `owned` dir, which is published as a plugin skill — it is **not** on OpenClaw's skill load path, so its skills land yet stay invisible to agents. This closes that gap behind an explicit **opt-in**.

## Grounding (against `openclaw/openclaw`)

The documented, watched mechanism for extra skill dirs is **`skills.load.extraDirs: string[]` in `~/.openclaw/openclaw.json`** — `docs/tools/skills-config.md`, consumed by `src/skills/runtime/refresh.ts`.

- **NOT** `settings.skills.customDirectories` (the value the PR2 docs guessed) — that nested form is **deprecated**, auto-migrated by `settings-manager.ts` to a flat *per-agent* `settings.skills` array. This PR also corrects that stale doc line.
- `skills.load.watch` defaults `true` → a newly added dir is picked up **without a restart**. Caveat: a long-lived agent *session* keeps its cached `<available_skills>` snapshot until a fresh session (`--session-key`).
- OpenClaw's own `plugins-install-command` uses the same read→merge→write pattern for `hooks.internal.load.extraDirs` — direct precedent.

## Change

- `SkillTarget` gains **`register: boolean`** (default false; owned default dir always false). Parsed from `MEMCLAW_SKILL_TARGETS` entries (`{ dir, mode, register? }`).
- `config.ts: ensureExtraSkillDirs(dirs)` — append-only, **idempotent** (canonical-path dedup, so a `~` entry matches an absolute dir), writes `openclaw.json` only when a dir is newly added (mirrors the `autoFixAllowlist` write idiom). **Fail-safe**: missing/unreadable config or a write error returns an `error`, never throws.
- `reconcileSkills` registers `register: true` dirs after the reconcile loop and reports them in a new `summary.registeredDirs` (standing truth; rides the heartbeat → `/fleet/nodes`, **no backend change**). A registration failure is logged, never aborts the heartbeat.
- Docs corrected + `register`/watch/stale-session documented.

## Tests

+11: `ensureExtraSkillDirs` (create/append/preserve/idempotent/`~`-canonical-match/missing-config-fail-safe/empty-noop), `register` parsing (default false, per-entry, owned never registered), `reconcileSkills` integration (registers an additive dir, untouched without the flag, idempotent re-run). **344 plugin tests pass; tsc clean.**

### DIFF
--- a/docs/mcp-skill-delivery.md
+++ b/docs/mcp-skill-delivery.md
@@ -179,14 +179,16 @@ arrays are deduped *across* targets. The summary also carries a
 (`{ dir, mode, installed, added, removed, collisions, protected }`) — so
 an operator can see exactly *which* dir a skill landed in or collided in.
 For the default single-target case it's one entry mirroring the top-level
-arrays.
+arrays. A `registeredDirs[]` array lists the target dirs MemClaw has
+ensured are on OpenClaw's skill load path this tick (the `register: true`
+opt-in below); it's empty unless a target opts in.
 
 ### Reconcile targets: `owned` vs `additive`
 
 By default the reconciler manages one **`owned`** dir — the plugin's own
 `skills/` — where it has full authority: anything on disk not in the
 catalog is pruned. Operators can add extra target dirs via the
-`MEMCLAW_SKILL_TARGETS` env var (JSON array of `{ dir, mode }`):
+`MEMCLAW_SKILL_TARGETS` env var (JSON array of `{ dir, mode, register? }`):
 
 - **`owned`** — fully MemClaw-managed (destructive prune). Use only for
   dirs MemClaw exclusively controls.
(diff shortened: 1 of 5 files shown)


ran-taig:
@claude please review

ran-taig:
Round-2 addressed in a13e8b07 — `summary.registeredDirs` is now populated **only after** `ensureExtraSkillDirs` returns no error (success or the idempotent already-present no-op). On a write failure we log a warning and leave the field empty, so it always reflects what's actually on the load path.

+1 test: a `register:true` target whose `openclaw.json` write fails is not reported in `registeredDirs` (the skill is still reconciled to disk). **345 plugin tests pass; tsc clean.**

@claude please re-review

ran-taig:
Round-3 addressed in the latest push — `ensureExtraSkillDirs` now builds the written-back array from the raw original `extraDirs` (string-only view used only for canonical-path dedup), so non-string entries are preserved verbatim. Makes the "existing entries preserved" docstring true.

+1 test (non-string `number`/`object` entries survive while the new dir is appended). **346 plugin tests pass; tsc clean.**

@claude please re-review

ran-taig:
Round-4 addressed in the latest push:

1. **Non-object config guard** — `ensureExtraSkillDirs` now rejects a top-level JSON array (or any non-object) with an error instead of letting it slip past the null check and get rewritten as `[]`. The file is left untouched.
2. **Literal relative-dir dedup** — `canonicalDir` expands `~`, resolves absolute paths, and compares relative entries **literally** (no `process.cwd()` guess), avoiding false dedup misses against OpenClaw's own relative-path interpretation.

+2 tests (array rejected + not clobbered; relative entry not falsely deduped). **348 plugin tests pass; tsc clean.**

@claude please re-review

ran-taig:
Round-5 addressed in the latest push:

1. **MEDIUM — registeredDirs on write failure:** `ensureExtraSkillDirs` now returns `alreadyPresent: string[]` in every path; on the error branch the caller populates `summary.registeredDirs` from it. Dirs already on the load path are reported as registered even when the write for new additions fails.
2. **LOW — split error message:** the non-object guard is now two checks — `not found or unreadable` vs `is not a JSON object`. Existing `/not found/` and `/not a JSON object/` tests still match.

+1 test (alreadyPresent reported alongside a newly added dir). **349 plugin tests pass; tsc clean.**

@claude please re-review

ran-taig:
Round-6 addressed in e5c4633c — the nested `config.skills` / `config.skills.load` normalization no longer clobbers a truthy non-object. Null/undefined still initialize to `{}`; a truthy non-object or array now returns a field-specific error (`'skills' is not an object` / `'skills.load' is not an object`) and leaves the file untouched, consistent with the top-level guard.

+2 tests (malformed `skills` and `skills.load` rejected + untouched). **351 plugin tests pass; tsc clean.**

@claude please re-review

ran-taig:
@erni-a

erni-a (approved):

PR #426, opened 2026-06-19
fix(core-api): cut prod connect-timeout + rate-limit-Redis log noise

Eldad-Caura:
## Why
Follow-up from the 2026-06-19 prod error review. Two transient signatures dominate the (low) prod error volume — both already retry-handled / fail-open, so this **reduces noise**, not a correctness fix:
- `pubsub handler raised; nacking for redelivery` — `httpx.ConnectTimeout` from the `memory.embedded`/`enriched` consumers' `get_memory` reads to core-storage during cold starts/recycles.
- `Failed to rate limit. Swallowing error` — Redis `ConnectionReset` (single-zone Memorystore maintenance / idle eviction) → slowapi fails open.

## What
**1. Idempotent storage reads ride out a storage cold start** (`common/http_retry.py` + `storage_client.py`)
`with_retry` gains `connect_phase_max_attempts`: connection-phase exceptions (`ConnectTimeout`/`ConnectError`/`PoolTimeout` — provably never sent) get a higher retry budget, while `ReadTimeout`/5xx stay capped at `max_attempts` (no 5xx-storm amplification; backoff still capped at `RETRY_BACKOFF_MAX_S`). `_get`/`_get_list` opt in via a small `_read_retry` wrapper using `CONNECT_PHASE_MAX_ATTEMPTS` — **the same budget POSTs already use** (`with_connect_phase_retry`). So a multi-second storage blip no longer exhausts the 3 GET attempts and surfaces as a handler failure → nack → redelivery on the Pub/Sub path.

**2. Rate-limiter Redis resilience** (`rate_limit.py`)
The slowapi `Limiter`'s Redis client now gets `socket_keepalive` + `health_check_interval=30s` + bounded connect/op timeouts via `storage_options` (Redis-only — `memory://` takes none). A connection the server already closed is PING-detected and reconnected **before** use rather than reset mid-command. `swallow_errors` fail-open stays the safety net.

## Scope / safety
- GET-only (the connect-timeout noise is on the read path); `_patch`/`_delete` unchanged (no documented noise there).
- No behavior change when healthy; backoff bounded; 5xx not over-retried.
- `ruff check` + `ruff format --check` + `mypy` clean on core-api + core-storage-api; `/simplify` applied.

### DIFF
--- a/common/http_retry.py
+++ b/common/http_retry.py
@@ -102,6 +102,7 @@ async def with_retry(
     retryable_exceptions: tuple[type[Exception], ...] = RETRYABLE_EXCEPTIONS,
     retryable_statuses: frozenset[int] = RETRYABLE_STATUS_CODES,
     max_attempts: int = RETRY_MAX_ATTEMPTS,
+    connect_phase_max_attempts: int | None = None,
 ) -> httpx.Response:
     """Wrap a single HTTP call with retry on transient errors.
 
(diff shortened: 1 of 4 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@erni-a 
please apreove

erni-a (approved):

PR #427, opened 2026-06-19
feat(core-api): route organization_settings through core-storage-api (Fix 2 Phase 0)

Eldad-Caura:
**Fix 2 Phase 0** — first slice of the Tier-2 *"no DB outside core-storage-api"* migration (ticket task_eda5ab48). Moves the per-org settings read/write off core-api's direct SQLAlchemy pool and behind new core-storage-api endpoints.

Settings are the highest-traffic pool acquisition and were the root of the 06-19 prod `ConnectTimeout` — the fire-and-forget `get_raw_settings(db=None)` cold-cache path opened its own DB session. This removes that DB dependency entirely. **No schema change** (`organization_settings`/`_audit` already live in core-storage-api's migrations).

### core-storage-api (now owns the DB access)
- `postgres_service.organization_settings_get` — read overrides (reader replica).
- `postgres_service.organization_settings_update` — the transactional upsert (`FOR UPDATE` read → flat diff → JSONB `||` merge → audit row) kept in **one** server-side transaction so the lost-update guard still holds (it can't span an HTTP read+write). Returns `{settings, changed}`; a no-op payload writes neither row.
- `routers/organization_settings.py` — `GET` + `POST /organization-settings/{org_id}`. POST (not PUT) so the storage_client's non-idempotent `_post` path applies (connection-phase retry only); re-applying an identical payload is a no-op anyway.

### core-api (now a client)
- `storage_client.get_org_settings` / `update_org_settings`.
- `services/organization_settings.py` now calls the storage client. The **TTL cache** (cache hits still short-circuit before any HTTP call), the `SETTINGS_CHANGED` publish, the schema validators, and `DEFAULT_SETTINGS` resolution all stay client-side. The old `db=None` self-session fallback is gone — every caller takes the same storage-client path. The `db` params are retained (ignored) for call-site back-compat; removing them across ~49 sites is a later phase.

### common
- `common/organization_settings_merge.py` — `deep_merge` / `diff_settings` lifted here (dependency-free) so both services share one implementation.

### Validation
- ruff + mypy clean on **both** core-api and core-storage-api.
- Full core-api suite green: **3464 passed**.
- `/simplify` applied.

### DIFF
--- a/common/organization_settings_merge.py
+++ b/common/organization_settings_merge.py
@@ -1,0 +1,50 @@
+"""Pure merge/diff helpers for organization-settings overrides.
+
+Shared by core-api (the settings service + display) and core-storage-api
+(the transactional ``POST /organization-settings`` upsert, which must compute
+the diff against the FOR-UPDATE'd row server-side so the read and write stay
+in one transaction). Kept dependency-free so both services can import it
+without pulling in a service's config.
+"""
+
+from __future__ import annotations
+
+from typing import Any
+
+
+def deep_merge(old: Any, new: Any) -> Any:
+    """Return ``old`` with ``new`` merged recursively for nested dicts.
+
+    Non-dict values in ``new`` overwrite ``old`` wholesale (including lists).
+    """
+    if not isinstance(old, dict) or not isinstance(new, dict):
+        return new
+    out = dict(old)
+    for k, v in new.items():
+        if isinstance(v, dict) and isinstance(out.get(k), dict):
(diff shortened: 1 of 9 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #428, opened 2026-06-20
feat(core-api): route tenant-discovery through core-storage-api (Fix 2 Phase 1)

Eldad-Caura:
**Fix 2 Phase 1** — second slice of the Tier-2 *"no DB outside core-storage-api"* migration (after Phase 0 settings, #427). Moves the lifecycle-fanout's tenant-discovery queries off core-api's direct SQLAlchemy pool and behind new core-storage-api endpoints. **No schema change.**

### core-storage-api (now owns the DB access)
- `postgres_service.tenants_list_active / _purgeable / _skills_factory_enabled` — the distinct-tenant and JSONB-flag SELECTs, on the **reader replica** (`get_read_session`). Same SQL that lived in `core-api/services/tenants.py`.
- `routers/tenants.py` — `GET /tenants/active`, `/tenants/purgeable`, `/tenants/skills-factory-enabled`. Envelope responses: `{tenant_ids:[...]}` for the memory-backed reads, `{org_ids:[...]}` for the settings-backed one.

### core-api (now a client)
- `storage_client.list_active_tenants / list_purgeable_tenants / list_skills_factory_enabled_orgs`.
- `services/tenants.py` — the 3 helpers are now thin storage-client wrappers (domain "why" stays in docstrings; SQL moved to storage). The `db` param is retained (ignored) for call-site back-compat (e.g. `routes/memories.py` still passes it), matching the Phase 0 pattern.
- `routes/lifecycle.py` — the cron fanout no longer opens its own `async_session` just to fetch the org list; it calls the now-sessionless discovery helpers directly, so the fan-out holds **no DB connection at all**. Drops the `core_api.db.session` import.

### Validation
- ruff + mypy clean on **both** core-api and core-storage-api.
- Full core-api suite green: **3469 passed**.
- `/simplify` ran (clean). New `test_tenant_discovery_storage` exercises the full core-api → storage-api → DB path; forge-cron routing tests updated for the dropped `db` arg.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #429, opened 2026-06-20
feat(core-api): route memory REST surface through core-storage-api (Fix 2 Phase 2)

Eldad-Caura:
**Fix 2 Phase 2** — third slice of the Tier-2 *"no DB outside core-storage-api"* migration (after Phase 0 settings #427, Phase 1 tenant-discovery #428). Moves the remaining direct-DB handlers in `routes/memories.py` off core-api's pool and behind new core-storage-api endpoints. **No schema change.**

**Key framing:** the feared "mutation + audit in one transaction" requirement doesn't exist today — `log_action` is already a decoupled async `POST /audit-logs`, `usage_service` is a no-op stub in OSS, and `soft_delete`/`update`/`agent_service` already route through the storage client. So this migrates the inline `db.execute`/`db.get`/`db.commit` that remained; the `db.commit()`s become no-ops and are removed; **audit stays the existing async POST** (not folded into storage transactions).

### core-storage-api (now owns the DB access) — 9 `postgres_service` methods + routes on the memories router
- `memory_fleet_distribution` (reader) — serves `/fleets` + `/admin/fleets` via an `exclude_scope_agent` flag.
- `memory_get_detail` (reader) — row + entity-link outerjoin in one round-trip; **embedding stats computed server-side**, raw vector + `search_vector` stripped.
- `memory_contradiction_rows` (reader) — 3 reads bundled; cross-tenant `older` guard server-side.
- `memory_soft_delete_by_filter` / `_by_ids` / `_by_run` (writer) — by-filter JSONB metadata predicate via **bound params** (no string interpolation).
- `memory_redistribute` (writer) — `SELECT … FOR UPDATE` + reassign + auto-promote `scope_agent`→`scope_team` + `not_found`, all in **one transaction**.
- `memory_admin_list` (reader) — cross-tenant, no visibility scoping, cursor.
- `memory_admin_stats` (reader) — single **GROUPING SETS** scan → `{total,by_type,by_agent,by_status}` (bound-param `text()` with `CAST(:p AS text)` so asyncpg can type the optional NULL filters).

### core-api
- `storage_client` gains 9 methods; list/dict GETs raise on 404, by-id reads return `None` for the caller to 404-translate.
- 12 handlers switched to the client. Trust gates / validation / cursor encode-decode / response shaping / audit stay in core-api. **`get_db` is retained** on the router (write/search/recall/ingest/list handlers still use it — deferred to later phases); the `db` Depends is dropped only from handlers that no longer touch db at all.

### Validation
- ruff@0.15.4 + mypy clean on **both** projects.
- Full core-api suite green: **3490 passed**.
- `/simplify` ran (clean); I reviewed the correctness-critical methods (redistribute txn, JSONB binding, server-side embedding stats, GROUPING SETS) by hand.
- New `tests/test_memory_storage_phase2.py` (21 tests) exercises the full path.

### DIFF
--- a/core-api/src/core_api/clients/storage_client.py
+++ b/core-api/src/core_api/clients/storage_client.py
@@ -835,6 +835,118 @@ class CoreStorageClient:
         )
 
     # =====================================================================
+    # Fix 2 Phase 2 — fleet/admin discovery, detail, bulk mutations
+    # =====================================================================
+    #
+    # The list/dict GET helpers below always return 200 with an envelope
+    # (an empty list / zeroed stats when there's no data), so a None / 404
+    # from ``_get`` means the endpoint is MISSING (version skew / routing),
+    # not "no data" — raise rather than silently degrade (mirrors the
+    # Phase 1 tenant-discovery methods). The two by-id reads
+    # (``get_memory_detail`` / ``get_memory_contradictions``) DO treat 404
+    # as a legitimate "memory not found" and return None for the caller to
+    # translate into its own 404.
+
+    async def memory_fleet_distribution(
+        self,
+        tenant_id: str | None = None,
+        *,
+        exclude_scope_agent: bool = False,
+    ) -> list[dict]:
+        """Distinct ``fleet_id`` with memory + agent counts, desc."""
+        params: dict[str, Any] = {"exclude_scope_agent": exclude_scope_agent}
+        if tenant_id is not None:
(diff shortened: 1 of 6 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #431, opened 2026-06-21
feat(core-api): route documents + fleet residuals through core-storage-api (Fix 2 Phase 3)

Eldad-Caura:
## Fix 2 Phase 3 — documents + fleet residuals via core-storage-api

Continues the "no DB outside core-storage-api" migration. Moves the last direct-DB **document** and **fleet** usages out of core-api route handlers and behind core-storage-api HTTP. Builds on Phase 2 (#429).

### New core-storage-api endpoints (+ `PostgresService` methods)
| Method / path | Notes |
|---|---|
| `POST /documents/upsert-xmax` | `INSERT … ON CONFLICT … RETURNING xmax` for insert-vs-update detection. Extends the existing service method to write the optional `embedding` column — passing `None` on a re-write clears a previously-indexed vector (intentional). |
| `GET /documents/collections` | Per-collection document counts; cross-tenant (`readable_tenant_ids`) widening preserved. |
| `POST /documents/search` | Vector (cosine) search, optional `status` filter. core-api still owns the embedding step and passes the vector across (storage stays a pure DB choke-point). |
| `GET /fleet/commands/in-flight-deploy` | Auto-upgrade in-flight deploy gate. |
| `GET /fleet/commands/deploy-attempt-count` | Auto-upgrade attempt budget. |

All service methods are verbatim ports of the existing `DocumentRepository` / `FleetRepository` logic — same SQL, predicates, and semantics; only the execution moves to storage-api (reads → `get_read_session`, writes → `get_session`).

### core-api migration
- **`routes/documents.py`** — `upsert` / `list_collections` / `search` now call the storage client. Dropped the intermediate `db.commit()` (storage commits its own session); audit `log_action` + final commit preserved. `get_embedding` hoisted to a top-level import; dead `db` dependency dropped from `list_collections`.
- **`routes/fleet.py`** — the two auto-upgrade gates call the storage client, preserving the fail-open `try/except`; removed the `fleet_repo` import.
- **`clients/storage_client.py`** — 5 new client methods (read/write-routed).

### Deferred (by design)
- `document_repo` / `fleet_repo` modules **stay** — `mcp_server.py` still uses them (Phase 4).
- `POST /documents/update-status` is **not** added here — its only consumers are the skill-lifecycle paths (`skills_inbox` / forge `cron_handler`), which migrate in Phase 5.

### No schema change
`documents.embedding` is already mapped (migration 003); the fleet gates use the partial index `ix_fleet_commands_node_created_deploy` from migration 024.

### Gates
- `ruff check` ✓ · `ruff format --check` ✓
- `mypy` core-api ✓ · `mypy` core-storage-api ✓
- **3506 tests pass** (+14 new Phase 3 tests covering xmax INSERT/UPDATE + embedding store/clear, collections counts/scoping/widening, search ordering/NULL-exclusion/status/widening, and both fleet gates). conftest ASGI-bridges the storage client to an in-process core-storage-api, so the route swaps are test-transparent.

### DIFF
--- a/core-api/src/core_api/clients/storage_client.py
+++ b/core-api/src/core_api/clients/storage_client.py
@@ -5,7 +5,9 @@ from __future__ import annotations
 import asyncio
 import logging
 from collections.abc import Awaitable, Callable
+from datetime import datetime
 from typing import Any, Literal, NotRequired, TypedDict
+from uuid import UUID
 
 import httpx
 
@@ -1190,14 +1192,45 @@ class CoreStorageClient:
     async def upsert_document(self, data: dict) -> dict:
         return await self._post("/documents", data)  # type: ignore[return-value]
 
+    async def upsert_document_xmax(self, data: dict) -> dict:
+        return await self._post("/documents/upsert-xmax", data, read=False)  # type: ignore[return-value]
+
+    async def list_document_collections(
+        self,
+        *,
+        tenant_id: str,
+        fleet_id: str | None = None,
+        readable_tenant_ids: list[str] | None = None,
+    ) -> dict:
+        params: dict[str, Any] = {"tenant_id": tenant_id}
+        if fleet_id is not None:
+            params["fleet_id"] = fleet_id
+        if readable_tenant_ids is not None:
+            params["readable_tenant_ids"] = readable_tenant_ids
+        result = await self._get("/documents/collections", read=True, **params)
+        if result is None:
+            # _get returns None only on 404 — i.e. the endpoint is missing
+            # (version skew / undeployed). Surface it instead of masking it as
+            # an empty collection list with `or {}`.
+            raise RuntimeError("GET /documents/collections returned 404 — core-storage-api version skew?")
+        return result
+
+    async def search_documents_vector(self, data: dict) -> list[dict]:
+        return await self._post("/documents/search", data, read=True)  # type: ignore[return-value]
+
     async def get_document(
         self,
         tenant_id: str,
         collection: str,
         doc_id: str,
+        *,
+        read: bool = True,
     ) -> dict | None:
+        # read=False forces the primary — use it for read-after-write re-fetches
+        # (e.g. immediately after an upsert) so replication lag can't yield None.
         return await self._get(
             f"/documents/{collection}/{doc_id}",
+            read=read,
             tenant_id=tenant_id,
         )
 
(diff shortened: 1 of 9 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
Addressed in `d91ba75` (force-pushed).

**Verified the finding is real** before applying: `get_read_session()` routes to the replica engine when `read_database_url` is set, and its own docstring warns *"Do NOT use for read-your-writes flows… use `get_session`"*. These two gates are exactly read-after-write — heartbeat N queues a deploy command (primary), heartbeat N+1's gate must see it or it queues a duplicate (the acked-stuck-queue storm the gate exists to prevent). The original core-api path read from its own AlloyDB-primary pool, so this was a silent downgrade.

**Fix:**
- `storage_client.py`: `fleet_in_flight_deploy` and `fleet_deploy_attempt_count` now call `_get(..., read=False)` (primary pool).
- `postgres_service.py`: `fleet_has_recent_in_flight_deploy` and `fleet_count_recent_deploys_for_target` now use `get_session()` (primary), with a comment so they don't get "optimized" back to the replica.

Scoped to the gates only — `/documents/search` + `/documents/collections` stay on `get_read_session` (staleness-tolerant; that's the point of the read/write split).

Gates green: `ruff check` + `format --check`, `mypy` (core-api + core-storage-api), **3506 tests pass**.

Eldad-Caura:
@claude

Eldad-Caura:
Addressed all four in `1807738` (force-pushed). Verified each before applying:

**#1 (High) — post-upsert re-fetch hit the replica.** Confirmed: `get_document` called `_get` with no `read=`, defaulting to the replica, so a successful indexed upsert could 500 under lag. Same read-after-write class as the fleet gates. Fix: `get_document` gains a `read` kwarg (default replica, all existing callers unchanged); the post-upsert re-fetch in `routes/documents.py` passes `read=False` (primary).

**#2 (High) — upsert-xmax bypassed the system-collection guard.** Confirmed: `document_upsert` guards `_`-prefixed collections but `document_upsert_returning_xmax` didn't, so the embedding path could write `_keystones` while the non-embedding path (same core-api route) 400s — inconsistent. Fix: added the `system: bool = False` param + guard to the service method (mirrors `document_upsert`), and the router wraps it in `try/except ValueError → HTTPException(400)`. New test `test_upsert_xmax_rejects_system_collection`.

**#3 (Medium) — dead `result is None` guard.** Confirmed `_post` always `raise_for_status()`s and never returns None. Removed the dead guard (and the unused assignment); the existing `if doc is None` check on the re-fetch — now a true read-after-write read — remains the real guard. Didn't add a `try/except httpx` mapping to "returned no rows" since that message is wrong for a storage 5xx and no sibling write path (e.g. the non-embedding `sc.upsert_document`) wraps storage errors — kept it consistent.

**#4 (Low) — `or {}` masked a 404.** Confirmed `_get` returns None only on 404. `list_document_collections` now raises (version-skew signal) instead of reporting an empty list with 200.

Gates green: `ruff check` + `format --check` (CI scope), `mypy` (both), **3507 tests pass** (+1 guard test).

Eldad-Caura:
@claude

Eldad-Caura:
@erni-a pleaee review

erni-a (approved):

PR #432, opened 2026-06-21
feat(core-api): route mcp_server's 9 ready tools through core-storage-api (Fix 2 Phase 4)

Eldad-Caura:
## Fix 2 Phase 4 — route mcp_server's 9 ready tools through core-storage-api

Continues the "no DB outside core-storage-api" migration. Routes the DB access of **9 MCP tools** off the local `_mcp_session()` (which set the RLS GUCs `app.tenant_id` / `app.readable_tenant_ids`) onto the storage HTTP client, carrying tenant scope **explicitly** as params. Builds on Ph0–3 (#427/#428/#429/#431).

**Tools migrated:** `memclaw_recall`, `memclaw_write`, `memclaw_manage`, `memclaw_entity_get`, `memclaw_tune`, `memclaw_doc`, `memclaw_list`, `memclaw_stats`, `memclaw_keystones_set`.

### New core-storage-api endpoints (+ `PostgresService` methods, + storage_client methods)
| Endpoint | Notes |
|---|---|
| `POST /memories/list` → `memory_list_by_filters` | **Visibility-scoped** list + cursor pagination + `readable_tenant_ids` widening. Verbatim port of `memory_repository.list_by_filters`. Distinct from the existing *unscoped* admin-list. |
| `POST /memories/stats-breakdown` → `memory_stats_breakdown` | GROUPING SETS (`total`/`by_type`/`by_agent`/`by_status`/`by_tenant`) + `include_deleted` CTE. Verbatim port of `compute_memory_stats`. |
| `GET /documents/collection-count` | Status-filtered active-only skills count (was a raw `db.execute` in mcp_server). |
| Widened `GET /documents/{collection}/{doc_id}` (+`readable_tenant_ids`), `DELETE` (+`require_status` atomic guard) | Cross-tenant read widening + atomic status-guarded skills delete. |

Everything else **reuses** existing Ph2/Ph3 client methods (memory CRUD/search/`soft_delete_by_ids`/`update_search_profile`, the `/documents/*` + `/fleet/*` surface, audit).

### Tenant isolation (the correctness-critical part)
Removing `_mcp_session()` for these tools means isolation is carried explicitly. Preserved exactly: **writes → home tenant only**; **reads → home + `readable_tenant_ids` only on cross-tenant reads**; **`scope='agent'` stays home-only** (forces `caller_agent_id`/`written_by`); `_UNAUTH`/`_ADMIN`/`_NO_AUTH` → empty readable ⇒ no widening. The new SQL ports keep the `tenant_id == X` vs `IN (readable)` split and the `scope_agent` visibility predicate verbatim.

**Verification beyond the gates:** ran an adversarial tenant-isolation review across the diff (writes, read-widening, the new SQL ports, sentinel handling, and whether any test assertion was weakened) — clean on all axes (one list-scope test was *strengthened*). It also caught + fixed one real regression in the in-progress work: a soft-deleted `older` memory could have leaked into lineage's `superseded_by`; restored the `older.deleted_at is None` filter client-side (matching REST `/contradictions`).

### Deferred to Ph5 (unchanged roadmap)
`memclaw_insights` + `memclaw_evolve` stay on `_mcp_session()`; capability_usage; **`_mcp_session()` deletion is the last step of Ph5**. After this PR, `fleet_repo` is fully unused and `document_repo`'s last callers are gone (both dropped at cleanup).

### Quality (`/simplify`)
Extracted `_storage_error_envelope()` (folded 6 duplicate `httpx`-error tails → 1 helper); dropped now-dead ORM-shape fallbacks in `memclaw_recall` (`sc.get_agent` returns `dict | None`).

### Gates
`ruff check` + `format --check` ✓ · `mypy` (core-api + core-storage-api) ✓ · **3507 tests pass** (incl. all `tests/test_mcp_*.py` + the cross-tenant-audit / by-id-authz isolation guards). conftest ASGI-bridges the client to in-process storage on the test engine, so the migration is test-transparent.

### DIFF
--- a/core-api/src/core_api/clients/storage_client.py
+++ b/core-api/src/core_api/clients/storage_client.py
@@ -448,10 +448,16 @@ class CoreStorageClient:
         status: str,
         supersedes_id: str | None = None,
         *,
+        tenant_id: str,
         unset_supersedes: bool = False,
         expected_supersedes_id: str | None = None,
     ) -> dict | None:
         """Update status and optionally set or clear ``supersedes_id``.
+
+        ``tenant_id`` is the home tenant of the target memory and is REQUIRED:
+        the storage route scopes every write to ``tenant_id == memory.tenant_id``
+        so a caller in tenant B cannot flip the status of tenant A's memory by
+        id (cross-tenant write guard).
 
         Set path (existing behaviour):
             ``supersedes_id=<uuid>`` sets the row's pointer, guarded by
@@ -484,7 +490,7 @@ class CoreStorageClient:
                     "for the CAS anchor; clearing without one would race "
                     "concurrent setters."
                 )
-        payload: dict[str, Any] = {"status": status}
+        payload: dict[str, Any] = {"status": status, "tenant_id": tenant_id}
         if supersedes_id is not None:
             payload["supersedes_id"] = supersedes_id
         if unset_supersedes:
@@ -820,8 +826,16 @@ class CoreStorageClient:
     async def mark_dedup_checked(self, memory_ids: list[str]) -> dict:
         return await self._post("/memories/mark-dedup-checked", {"memory_ids": memory_ids})  # type: ignore[return-value]
 
-    async def batch_update_status(self, data: dict) -> dict:
-        return await self._post("/memories/batch-update-status", data)  # type: ignore[return-value]
+    async def batch_update_status(self, data: dict, *, tenant_id: str) -> dict:
+        """Apply status updates to many memories within one tenant.
+
+        ``tenant_id`` is the trigger tenant shared by every memory in the
+        batch and is REQUIRED: the storage route scopes each row's write to
+        ``tenant_id == memory.tenant_id`` (cross-tenant write guard). It is
+        injected at the batch level rather than per row.
+        """
+        payload = {**data, "tenant_id": tenant_id}
+        return await self._post("/memories/batch-update-status", payload)  # type: ignore[return-value]
 
     async def bulk_get_memories(
         self,
@@ -921,6 +935,35 @@ class CoreStorageClient:
         """Soft-delete live memories by id (tenant-scoped); returns count."""
         result = await self._post("/memories/soft-delete-by-ids", {"tenant_id": tenant_id, "ids": ids})
         return (result or {}).get("deleted", 0)  # type: ignore[union-attr]
+
+    async def list_memories_by_filters(self, data: dict) -> list[dict]:
+        """Non-admin memory list WITH visibility scoping (MCP ``memclaw_list``).
+
+        ``data`` carries the filter/sort/cursor params (incl. ``caller_agent_id``
+        for the scope_agent visibility gate and an optional ``readable_tenant_ids``
+        widening) plus ``limit`` (the desired page size). The storage service
+        over-fetches ``limit+1`` internally for has_more detection; the caller
+        slices to ``limit`` + builds the next cursor. Distinct from
+        ``admin_list_memories`` (NO visibility scoping).
+        """
+        result = await self._post("/memories/list", data, read=True)
+        if result is None:
+            raise RuntimeError("core-storage-api /memories/list returned 404")
+        return result  # type: ignore[return-value]
+
+    async def memory_stats_breakdown(self, data: dict) -> dict:
+        """Visibility-scoped stats breakdown (MCP ``memclaw_stats``).
+
+        ``data`` carries ``{tenant_id?, fleet_id?, agent_id?, memory_type?,
+        status?, include_deleted?, readable_tenant_ids?}``. Returns
+        ``{total, by_type, by_agent, by_status}`` plus optional ``by_tenant`` /
+        ``deleted`` / ``total_including_deleted``. Distinct from
+        ``admin_memory_stats`` (no scoping).
+        """
+        result = await self._post("/memories/stats-breakdown", data, read=True)
+        if result is None:
+            raise RuntimeError("core-storage-api /memories/stats-breakdown returned 404")
+        return result  # type: ignore[return-value]
 
     async def soft_delete_by_run(
         self,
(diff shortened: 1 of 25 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
Addressed all three in `b7c4572` (force-pushed). Verified each first:

**HIGH — cross-tenant leak on `/memories/stats-breakdown`.** Confirmed real: `/list` requires `tenant_id` (422), `/stats-breakdown` didn't, and `memory_stats_breakdown`'s `elif tenant_id:` left `scope_filters` empty when both `tenant_id` and `readable_tenant_ids` were absent → unscoped aggregation across all tenants. Good catch — my adversarial pass missed it because it only checked the MCP caller (which always sends `tenant_id`); the gap is at the endpoint. Fix (defense-in-depth, both layers):
- Router: `/stats-breakdown` now requires `tenant_id` (422), mirroring `/list`.
- Service: changed `elif tenant_id:` → unconditional `else: Memory.tenant_id == tenant_id` so it's **safe-by-default** like `memory_list_by_filters` — with `tenant_id=None` it renders `tenant_id IS NULL` (matches nothing), never cross-tenant.
- Added regression test `test_ph4_stats_tenant_guard.py` (422 without tenant_id).

**MED — `literal_binds=True` over user-supplied filters.** Confirmed: `_predicate_sql` inlined `fleet_id`/`agent_id`/`memory_type`/`status` via `literal_binds`. Converted to **bound parameters** (`compile(paramstyle='named')` → `text(sql, params)`), keeping the dynamic GROUPING SETS structure. Updated the (now-incorrect) "literal_binds is safe" comment. Validated against the real test DB via the `memclaw_stats` path (`test_cross_tenant_audit_surfaces`).

**LOW — misleading `limit+1` docstring.** Corrected on both `storage_client.list_memories_by_filters` and the `/list` router (the *service* over-fetches +1, not the caller).

Gates: `ruff check` + `format --check` ✓, `mypy` (both) ✓, **3508 tests pass** (+1 guard). (One pre-existing flaky test, `test_security_tenant_isolation::test_own_tenant_memory_access_ok`, intermittently fails under full-suite load — unrelated to this PR's code paths; flagging it separately.)

Eldad-Caura:
@claude

Eldad-Caura:
Addressed both in `7470b2f` (force-pushed). Verified each first:

**HIGH — `memclaw_recall` missing `httpx.HTTPStatusError` handler.** Confirmed: its outer `try` caught only `HTTPException`, but `sc.get_agent()` / `enforce_fleet_read()` became storage HTTP calls in this PR (raise on non-2xx). Added `except httpx.HTTPStatusError: return _storage_error_envelope(e, t0)` — matching the handler the other migrated tools got.

**HIGH — `memclaw_list` had no try/except.** Confirmed: `_require_trust` (2161) and `list_memories_by_filters` (2223) are both storage calls with no surrounding handler → a storage 5xx escaped raw. Wrapped the `async with _no_db()` body in `try/except HTTPException / httpx.HTTPStatusError / Exception`, mirroring the manage/doc pattern. (The mcp_server.py diff looks large because the wrap re-indents the block body — it's mechanically just the wrap.)

Both gaps were the two migrated tools the first round's `_storage_error_envelope` helper hadn't been wired into.

**Regression tests** (`test_ph4_mcp_storage_error_envelope.py`): stub the storage method to raise `httpx.HTTPStatusError` and assert each tool returns the canonical error envelope rather than raising. Both fail without the fix.

Gates: `ruff check` + `format --check` ✓, `mypy` (both) ✓, **3510 tests pass** (+2 storage-error + the prior +1 stats-guard).

Eldad-Caura:
@claude

Eldad-Caura:
@claude

Eldad-Caura:
Addressed all three in `f9d1c5a` (rebased onto the new main now that #433/#434 merged; force-pushed). Verified each + hand-reviewed the isolation surface:

**HIGH — `memory_update_status` cross-tenant write gap.** Confirmed: the service UPDATE was `where(Memory.id == memory_id)` with no tenant predicate. Fixed at the storage layer + threaded `tenant_id` through every caller:
- Service: added required `tenant_id` → `where(Memory.id == ..., Memory.tenant_id == tenant_id)`.
- PATCH `/{id}/status` route: 422 on missing `tenant_id`, and **all three** write paths scoped — the `memory_update_status` call **and the two inline `sql_update(Memory).where(Memory.id == ...)` statements** (the CAS-retraction + set-supersedes UPDATEs were unscoped too).
- Batch route: batch-level `tenant_id` (422 if missing) threaded into each update.
- Client `update_memory_status` + `batch_update_status`: required `tenant_id`.
- Callers all pass the row's HOME tenant: mcp transition (`_get_tenant()`), REST route (auth-enforced), retraction (each row's own `tenant_id` — candidate + new_memory), crystallizer + contradiction batch (trigger tenant), legacy write (memory's tenant).

**MED — cursor direction.** `memory_list_by_filters` cursor predicate now branches on `order` (desc → `<`, asc → `>`), so an asc page walks forward. (Left the pre-existing `memory_admin_list` cursor alone — out of scope.)

**MED — `_predicate_sql` fragility.** Replaced the `else "TRUE"` fallback with an explicit `raise RuntimeError` (not assert — survives `-O`): a missing compiled WHERE now fails loudly instead of silently running an unscoped aggregation.

**Regression tests** (`test_ph4_status_tenant_guard.py`): wrong-tenant `memory_update_status` is a no-op (row unchanged) + cross-tenant PATCH 404s; correct-tenant updates; asc-cursor forward page (desc regression guard).

Gates: `ruff check` + `format --check` ✓, `mypy` (both) ✓, **3516 tests pass**.

Related (NOT in this PR, flagging): `memory_update_embedding` is a sibling internal update still scoped by id only — worth the same guard as a follow-up; it's a worker/backfill path (not user-reachable with arbitrary ids), so lower risk.

Eldad-Caura:
@claude

Eldad-Caura:
Addressed both in `9a90fb5` (force-pushed). Verified each against the code first:

1. **Medium — stale cursor comment** (`postgres_service.py:3113`): confirmed the `desc → tuple < cursor` / `asc → tuple > cursor` branches are already correct; only the comment's trailing sentence still described the pre-fix bug. Rewrote it to describe the current behavior and cite the regression test (`test_list_by_filters_asc_cursor_returns_forward_page`).
2. **Low — `memclaw_entity_get` inline error handler** (`mcp_server.py:1232`): confirmed the inline `except httpx.HTTPStatusError` block was equivalent to `_storage_error_envelope` (and `memclaw_tune`, also `-> str`, already uses the helper). Replaced it with `return _storage_error_envelope(e, t0)`.

Gates green: ruff check + `format --check` (0.15.4), mypy both projects, targeted tests (40 passed — ph4 status/stats guard, entity cross-tenant, direct-mcp, tools-registry).

Eldad-Caura:
@claude

Eldad-Caura:
Addressed both in `5f3704c` (force-pushed). Verified each against the code first:

1. **Medium — `str(None)` tenant in `_attempt_path_c_retraction`** (`contradiction_detector.py:1553`): `get_memory` always serialises `tenant_id` (NOT NULL column, present in `MEMORY_FIELDS`), so the `str(None)` → `"None"` path is unreachable today. Still hardened it: replaced the two `str(x.get("tenant_id"))` calls with extracted `cand_tenant`/`new_tenant` and a single upfront guard that raises rather than silently no-op'ing if that invariant ever breaks.
2. **Low — unbounded `limit` on `/list`** (`memories.py` `list_by_filters`): added `limit = max(1, min(int(raw_limit), 5000))` with a 422 on non-int, before the `limit + 1` over-fetch. Capped at **5000** to match the `/null-embedding-ids` bound in this same file, rather than the suggested 500 — consistency with the existing sibling endpoint.

Gates green: ruff check + `format --check` (0.15.4), mypy both projects, targeted tests (182 passed — ph4 status/envelope + all Path C retraction suites).

Eldad-Caura:
@erni-a

erni-a (approved):

PR #434, opened 2026-06-21
fix(core-api): keep-alive pool + bounded entity-context fan-out (VPC-connector ConnectTimeout residual)

Eldad-Caura:
## Keep-alive pool + bounded entity-context fan-out — stop the VPC-connector ConnectTimeout storms

Addresses the **residual VPC-connector cold-handshake class** (Ran's `BUG-staging-vpc-connector-saturation`, re-confirmed in the 2026-06-21 post-promote triage). The `e2-standard-4` connector upgrade raised the ceiling, but **bursts of simultaneous cold TCP handshakes** still trip it. Triage data (staging core-api, 6h): **174 `ConnectTimeout`s**, dominated by the Path-C entity-context fan-out — `GET /memories/<id>` ×124, `/entities/*` ×26 — plus `/audit-logs/bulk` ×16 (the audit "events lost"). Almost all `attempt 1/5` = brand-new connections.

### Two app-side levers (no infra change)
1. **`storage_client._make_pool`: `keepalive_expiry=90s`** (was httpx's 5s default). The fan-out's bursts arrive seconds-to-tens-of-seconds apart; at 5s the warm keep-alive pool drained *between* bursts, so the next burst re-handshook. 90s keeps connections warm across the gap → the fan-out **reuses** them. (`max_keepalive_connections=150` was already generous — expiry was the missing piece.)
2. **`contradiction_detector`: bound the Path-C entity-context fan-out** via a `_bounded_gather` helper (limit 8), applied to both the per-candidate `_fetch_entity_context` fan-out (candidates up to **40**) and the per-link `_hydrate` fan-out — so one burst can't open dozens of simultaneous handshakes. Order preserved (gather semantics); fresh per-call semaphore (loop-safe).

Together: warm reuse eliminates the cold handshakes; the cap keeps any cold burst within the connector's handshake headroom. **Benefits all storage callers** (entity-context, audit-flush, etc.), so it also reduces the `/audit-logs/bulk` connect-timeout "events lost".

### Verification
- New unit test: `_bounded_gather` caps concurrency (peak ≤ limit) + preserves order.
- The **41 existing Path-C tests** (`caura_131` entity-aware detection, `caura_130` safety, `a4_13` retraction, storage-client pool recovery) still pass — behavior unchanged.
- `ruff check` + `format --check` ✓, `mypy` (core-api + core-storage-api) ✓, full suite green.

Independent of the Fix 2 migration.

### Related / not in this PR
- **Embedding (TEI) timeouts** ride a *separate* httpx client (already timeout-tuned per #332). Same `keepalive_expiry` idea may apply there — flagging as a follow-up rather than touching a second client here.
- **Audit-flush durability** (don't silently drop events on connect failure — retry/dead-letter) is the other lever from triage; this PR removes most of the *triggering* timeouts, but durable audit is worth its own change.

### DIFF
--- a/core-api/src/core_api/clients/storage_client.py
+++ b/core-api/src/core_api/clients/storage_client.py
@@ -151,7 +151,14 @@ class CoreStorageClient:
             # Sized for 20 concurrent storm writes x ~5 storage calls
             # each = 100 in-flight at peak, plus tenant-B probe headroom
             # and a 33% burst margin.
-            limits=httpx.Limits(max_connections=200, max_keepalive_connections=150),
+            # ``keepalive_expiry`` defaults to 5s in httpx — too short for the
+            # bursty entity-context / enrichment fan-out, whose bursts arrive
+            # seconds-to-tens-of-seconds apart. At 5s the warm pool drains
+            # between bursts, so the next burst opens cold TCP handshakes that
+            # pile up at the Cloud Run VPC connector (the residual ConnectTimeout
+            # class). 90s keeps connections warm across the inter-burst gap so the
+            # fan-out reuses them instead of re-handshaking.
+            limits=httpx.Limits(max_connections=200, max_keepalive_connections=150, keepalive_expiry=90.0),
             follow_redirects=True,
         )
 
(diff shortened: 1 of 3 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #464, opened 2026-06-22
fix(core-api): per-tenant embed concurrency cap (noisy-neighbor-search)

Eldad-Caura:
## Problem
`noisy-neighbor-search` from the latest loadtest: a single tenant's 20-concurrency search storm degraded **another** tenant's search p95 by **6.43×** while Cloud Run CPU was idle. Root cause confirmed via live infra pull — the embedding backend `staging-memclaw-tei` is pinned at `minScale=2, maxScale=2, containerConcurrency=10` (1 GPU each) → only **~20 concurrent embedding slots platform-wide, no autoscale**. A hot tenant's query embeddings fill all 20 slots; other tenants' query embeds queue behind them. Route-level `per_tenant_slot("search")=32` is per-tenant and too coarse — it doesn't protect *across* tenants at the shared embedding backend.

## Fix
Cap concurrent embedding-backend calls **per tenant**, reusing the existing `per_tenant_slot` bulkhead:
- New `"embed"` scope in `middleware/per_tenant_concurrency.py` + `per_tenant_embed_concurrency` config knob (**default 6**).
- The slot wraps **only the single-flight cold-miss leader** in `memory_service._get_or_cache_embedding` (the one call that actually hits TEI) — cache hits and in-flight joiners return earlier and take no slot.
- A hot tenant fast-fails **429** on excess concurrent cold embeds rather than starving the shared pool. With cap `N` on `M` core-api instances, one tenant holds at most `N*M` of the ~20 TEI slots, leaving headroom for others.

Cap rationale: on 2 staging core-api instances a single tenant holds ≤`6×2=12`/20 (60%), ≥8 free — enough that a cold tenant-B query never queues (failure mode was 100% occupancy). Sits under the route-level search cap of 32 (many searches in flight, few concurrent cold embeds).

## Tests
`tests/test_per_tenant_concurrency.py`: `embed`-scope 429-on-exhaustion, scope isolation vs `search`, tenant isolation, release-on-exception. `tests/test_query_embedding_stampede_guard.py`: `test_embed_slot_only_held_by_leader` — N concurrent same-key cold callers all succeed at cap=1, proving joiners/cache-hits take no slot.

Local gate (CI-exact): `ruff check core-api/src/` ✓, `ruff format --check core-api/src/` ✓, `mypy core-api/src/` ✓ (192 files), `pytest` ✓ (20 passed, no DB needed for these).

## Follow-up (not in this PR)
- Ingest/worker batch embed (`ingest_service.py`) left ungated — storm is search-side; worker already bulkheaded by `per_tenant_storage_write_concurrency`.
- Pairs with a staging infra bump to TEI capacity (raise maxScale) — tracked separately.

### DIFF
--- a/core-api/src/core_api/config.py
+++ b/core-api/src/core_api/config.py
@@ -208,6 +208,20 @@ class Settings(BaseSettings):
     # search/list ops) while still tripping under a genuine storm.
     per_tenant_search_concurrency: int = 32
     per_tenant_write_concurrency: int = 16
+    # Per-tenant cap on concurrent embedding-backend (TEI) calls. Gates
+    # only the single-flight cold-miss leader in
+    # ``memory_service._get_or_cache_embedding`` (cache hits / in-flight
+    # joiners take no slot), so one hot tenant's search storm can't
+    # occupy the whole embedding service and starve other tenants
+    # (noisy-neighbor-search). The TEI backend is a fixed pool
+    # (``staging-memclaw-tei``: 2 instances x containerConcurrency 10 =
+    # ~20 slots, no autoscale); with cap N on M core-api instances a
+    # single tenant holds at most ``N * M`` of those, leaving headroom
+    # for everyone else. Tighter than ``per_tenant_search_concurrency``
+    # on purpose: a tenant may have many searches in flight but only a
+    # few concurrent cold embeds. Fast-fails 429 like the other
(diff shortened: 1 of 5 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #465, opened 2026-06-22
fix(core-api): resolve home fleet on omitted fleet_id for MCP writes

ran-taig:
## Why

Validated follow-up **B1** from the "One Skill Company Brain" review. An MCP write that **omits `fleet_id`** persists `fleet_id=NULL`, and a NULL-fleet row silently drops out of teammates' `scope_team` / fleet-filtered recall. REST already resolves the agent's **home fleet** on omit (`routes/memories.py:937` single, `:1125` bulk) — MCP did not, so the *same* write scoped differently by transport. This also hit `memclaw_doc op=write collection=skills`, i.e. the "publish a skill, the fleet inherits it" flow could strand a skill at `fleet_id=NULL`.

Grounded: `enforce_fleet_write` returns the home-fleet-bearing agent but the MCP paths discarded it; the shared `create_memory` service uses `data.fleet_id` as-is (no backfill). No fleet ContextVar / `X-Fleet-ID` header / `?fleet_id=` parse exists at the MCP layer, so the kwarg was the only source.

## Change

Two sites in `core_api/mcp_server.py` (`memclaw_write` and `memclaw_doc op=write`): capture the agent `enforce_fleet_write` already returns and backfill before the write:

```python
agent = await enforce_fleet_write(db, tenant_id, agent_id, fleet_id)
if not fleet_id and agent.get("fleet_id"):
    fleet_id = agent["fleet_id"]
```

### Safety (no existing flow touched)
- **Explicit writes untouched** — backfill fires only when `fleet_id` is falsy; an explicit (incl. cross-fleet) value is unchanged.
- **No new cross-fleet surface** — `enforce_fleet_write` runs *first* with the original value; the backfill target is always the agent's **own** home fleet (always self-writable), so no re-enforcement needed (mirrors REST's order-independence here).
- **Worst case unchanged** — agent with no home fleet → no-op → row stays NULL. Closing that residual (seed home fleet from the connection `?fleet_id=`) is a separate decision, not this PR.
- **Batch** — `BulkMemoryItem` inherits fleet from the parent, so one top-level backfill covers the whole batch.

## Tests
`memclaw_write`: omit→home · explicit-untouched · batch omit→home · **no-home-fleet→NULL** (guards against over-claiming). `memclaw_doc op=write`: omit→home · explicit-untouched. The pre-existing `persists_null` test is clarified to the no-home-fleet case it actually covers. All 6 pass; `ruff check core-api/src/` + mypy clean on the touched file.

## Out of scope (flagged)
`memclaw_evolve` / `memclaw_insights` / `memclaw_keystones_set` also take `fleet_id` (different default scope) — audit separately. Layer 2 (connection-fleet seeding) is its own ticket. Static skill's `fleet_id` warning text becomes partially stale post-merge — to be updated in the separate skill-reconciliation work.

### DIFF
--- a/core-api/src/core_api/mcp_server.py
+++ b/core-api/src/core_api/mcp_server.py
@@ -782,7 +782,19 @@ async def memclaw_write(
             # REST write path uses. Without this, MCP writes succeed without
             # ever creating an Agent row, so a follow-up
             # PATCH /agents/{id}/trust 404s.
-            await enforce_fleet_write(db, tenant_id, agent_id, fleet_id)
+            agent = await enforce_fleet_write(db, tenant_id, agent_id, fleet_id)
+            # Mirror the REST write paths (routes/memories.py:937, :1125): an
+            # omitted fleet_id resolves to the caller's home fleet, so an MCP
+            # write scopes like a REST write instead of persisting
+            # fleet_id=NULL — a NULL-fleet row silently drops out of
+            # teammates' scope_team / fleet-filtered recall. Fires only when
+            # fleet_id is falsy, so an explicit (incl. cross-fleet) fleet_id
+            # is left untouched and still passes through the trust gate
+            # enforce_fleet_write applied above. If the agent has no home
+            # fleet (never been scoped), this is a no-op and the row stays
+            # NULL — unchanged from prior behavior.
+            if not fleet_id and agent.get("fleet_id"):
+                fleet_id = agent["fleet_id"]
             if content is not None:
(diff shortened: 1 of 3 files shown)


ran-taig:
@claude please review

ran-taig:
@claude

ran-taig:
@claude

ran-taig:
@erni-a

erni-a (approved):

PR #466, opened 2026-06-22
chore(core-api): attribute single_write latency (dedup lookup + storage_total_ms)

Eldad-Caura:
## Why
`slo-p99-single_write`: ~**18.6s p99** with ~**575ms p50** on staging. Investigation showed the report's "move enrichment to a worker" fix is **already live** (staging defers both enrichment and embedding off the hot path), so the tail is downstream — storage-DB contention (likely **CAURA-686** PG row-lock on the insert) or core-api-side overhead. To localise it before changing code, the write path needs to attribute time across its storage roundtrips.

## What
Today only `WriteMemoryRow`'s insert is timed (`storage_ms`). The dedup lookup (`GET /memories/by-content-hash` in `CheckExactDuplicate`) was an **untimed** slice of `total_ms`, leaving `total_ms - storage_ms` an undifferentiated bucket. This PR:
- Times the dedup lookup → `phase_timings["dedup_lookup_ms"]` (mirrors `storage_ms`/`entity_links_ms`).
- Surfaces `dedup_lookup_ms` + a derived `storage_total_ms` (dedup + insert + entity-links) in the existing `memory_write_latency` log.

`total_ms - storage_total_ms` is then **pure core-api-side time** — the split that tells us whether the 18.6s tail is storage-DB or core-api. Additive only, no behaviour change.

## How we'll use it
Query `memory_write_latency` p99 of `storage_total_ms` vs the residual, and pair with the existing `GET /api/v1/storage/_debug/pg_locks` toolkit (+ `scripts/capture_pg_locks_during_storm.py`) under a same-(tenant,fleet) write burst to confirm/refute `Lock/transactionid` on `INSERT INTO memories` (CAURA-686).

## Tests
`tests/pipeline/test_write_pipeline.py::test_check_exact_duplicate_records_dedup_lookup_ms` (pure unit). Local gate: `ruff check core-api/src/` ✓, `ruff format --check` ✓, `mypy core-api/src/` ✓ (192 files), unit tests ✓.

Companion to the noisy-neighbor fix in #464.

### DIFF
--- a/core-api/src/core_api/pipeline/steps/write/check_exact_duplicate.py
+++ b/core-api/src/core_api/pipeline/steps/write/check_exact_duplicate.py
@@ -4,6 +4,8 @@ writes of identical content no longer collide (friction §2.8)."""
 
 from __future__ import annotations
 
+import time
+
 from fastapi import HTTPException
 
 from core_api.clients.storage_client import get_storage_client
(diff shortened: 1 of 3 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #467, opened 2026-06-22
fix(core-storage-api): tenant-guard by-id memory UPDATEs

Eldad-Caura:
## Tenant-guard the by-id memory UPDATEs

Continues the "no DB outside core-storage-api" migration's hardening: every by-id memory **write** that core-api routes through core-storage-api is now scoped to the row's **home tenant**, so a wrong-tenant `memory_id` can never mutate a foreign row.

### Ops hardened (`PostgresService`)
| Op | Route | Guard |
|---|---|---|
| `memory_update` | `PATCH /memories/{id}` (content PATCH — user-reachable) | scope the `SELECT … FOR UPDATE` existence check **and both** UPDATEs (column `values` + the raw JSONB metadata-merge) to `tenant_id` |
| `memory_update_embedding` | `PATCH /memories/{id}/embedding` (worker/backfill) | scope the UPDATE |
| `memory_mark_dedup_checked` | `POST /memories/mark-dedup-checked` | scope the bulk `last_dedup_checked_at` stamp to `Memory.id.in_(ids) AND tenant_id` |

### Threading
`tenant_id` is a **required** argument end-to-end: storage routes (422 if missing) → typed storage client → every caller. Each caller passes the row's **HOME** tenant — `routes/testing.py` uses the already-verified `row["tenant_id"]`; the governance / crystallizer / memory_service callers pass their in-scope `tenant_id`. The two `memory_service` callers that previously reached into the private `sc._patch` now go through the guarded `sc.update_memory`.

### Out of scope (by design)
- `memory_update_status` — its cross-tenant write guard lands on the Phase 4 PR (#432); left untouched here to avoid a merge conflict.
- `memory_increment_recall` — deferred (internal counter, repo-direct, already tenant-scoped ids).
- Entity by-id updates are already tenant-guarded.

### Tests
`tests/test_tenant_guard_byid_updates.py` (mirrors `tests/test_ph3_storage_documents_fleet.py`): wrong-tenant write = no-op (row untouched / 404 / silent skip), same-tenant write applies, and each route 422s without `tenant_id`. Existing caller tests updated for the new signature.

Gates: ruff check + `format --check` (0.15.4), mypy both projects, full suite **3516 passed**.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #468, opened 2026-06-22
fix(core-storage-api): 404 (not silent 200) on wrong-tenant embedding write

Eldad-Caura:
## 404 (not silent 200) on wrong-tenant embedding write

Follow-up to the post-merge review on [#467](https://github.com/caura-ai/caura-memclaw/pull/467#issuecomment-4767916672).

#467 scoped `memory_update_embedding`'s UPDATE to `tenant_id`, but the method stayed `-> None` and the route always returned `{"ok": True}` — so a wrong-tenant call was a silent 200 no-op, and `crystallizer_service._remediate_missing_embeddings` counted it as a successful patch (inflating the remediated tally in its log).

### Change
- `memory_update_embedding` → `-> bool`, returns `(result.rowcount or 0) > 0` (mirrors `memory_update_status`).
- `PATCH /memories/{id}/embedding` raises **404** on a no-op instead of 200 (consistent with `PATCH /memories/{id}`); the client then returns `None`.
- crystallizer increments `patched` only when the write actually landed.

### Reachability
The mismatch is **unreachable in practice today** — `_remediate_missing_embeddings` fetches candidates scoped to `tenant_id` and writes with that same tenant, so the counter was already accurate. This is a consistency + defense-in-depth fix so the three by-id ops behave uniformly. (`mark_dedup_checked` is intentionally left as a bulk 200 — per-row 404 doesn't fit a bulk stamp, and it has no success-counter caller.)

### Tests
Extended `test_update_embedding_wrong_tenant_is_noop` to assert the wrong-tenant call now returns `None` (404) and the same-tenant call returns `{"ok": True}`. Gates: ruff@0.15.4 check + `format --check`, mypy both projects, full suite **3530 passed**.

### DIFF
--- a/core-api/src/core_api/services/crystallizer_service.py
+++ b/core-api/src/core_api/services/crystallizer_service.py
@@ -242,8 +242,12 @@ async def _remediate_missing_embeddings(
         embedding = await get_embedding(content)
         if embedding is None:
             continue
-        await sc.update_embedding(str(mem_id), tenant_id, embedding)
-        patched += 1
+        # Count only writes that actually landed: update_embedding returns
+        # None when the row didn't match (404), so a no-op never inflates
+        # the remediated tally. (Candidates are tenant-scoped, so a mismatch
+        # shouldn't occur here — this just keeps the counter honest.)
+        if await sc.update_embedding(str(mem_id), tenant_id, embedding) is not None:
+            patched += 1
 
     if patched:
         logger.info(
--- a/core-storage-api/src/core_storage_api/routers/memories.py
+++ b/core-storage-api/src/core_storage_api/routers/memories.py
@@ -1371,12 +1371,17 @@ async def update_embedding(memory_id: UUID, request: Request) -> dict:
     tenant_id = body.get("tenant_id")
     if not tenant_id:
         raise HTTPException(status_code=422, detail="tenant_id is required")
-    await _svc.memory_update_embedding(
+    updated = await _svc.memory_update_embedding(
         memory_id,
         tenant_id=tenant_id,
         embedding=body["embedding"],
         metadata=body.get("metadata"),
     )
+    if not updated:
+        # No row for this (id, tenant) — surface 404 rather than a silent
+        # 200 (consistent with PATCH /memories/{id}); the client returns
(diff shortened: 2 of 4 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
Addressed in `5f0e718` (force-pushed). Verified first:

**Low — `(rowcount or 0) > 0` doesn't guard `-1`**: confirmed the bug — `-1 or 0` evaluates to `-1` (truthy), so `-1 > 0` is `False`, which would misreport a real update as a no-op (false 404) if a dialect ever returns the indeterminate `-1` sentinel. Replaced with `rc = result.rowcount; return rc is not None and rc > 0`.

Applied to **both** the new `memory_update_embedding` check **and** the existing sibling `memory_update_status` (same idiom from #432) so they stay consistent and both are correct-by-construction. asyncpg returns 0/positive today, so this is by-construction hardening rather than a live bug.

Gates green: ruff@0.15.4 check + `format --check`, mypy (core-storage-api), targeted guard tests (35 passed — tenant-guard by-id + ph4 status guard + crystallizer).

Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

Eldad-Caura:
Round 2 is correct — and it means round 1's premise was wrong. `(-1 or 0)` is `-1` (truthy), so `(rc or 0) > 0` and `rc is not None and rc > 0` are **identical for every value** (None/0/-1 → no-match, positive → match). There was never a "false 404 on -1" — the change was purely cosmetic and the comment I added (parroting round 1) was inaccurate.

Rather than keep a change justified by a false premise, I **reverted** the rowcount-expression edit at both sites (`b3a0a68`), including the gratuitous touch to `memory_update_status` — that method is now back to its #432 form. #468 is once again just the substantive fix: `memory_update_embedding -> bool`, route `404` on no-match (consistent with `PATCH /memories/{id}`), and the crystallizer counting only writes that landed — which both review passes validated as sound end-to-end.

Gates green: ruff@0.15.4 check + `format --check`, mypy (core-storage-api), guard tests (13 passed).

PR #469, opened 2026-06-22
perf(core-api): make document-ingest (kreuzberg) an opt-in extra to cut cold start

Eldad-Caura:
## Why
core-api Cloud Run `startup_latencies` p99 is **36–58s** — the real source of the load-test `single_write`/search latency tails (the handler is fast: p99 ~1.6s; pgvector 241ms; TEI 17ms). Profiling showed the **logged** app startup is only ~9s (and the `agent-backfill` lifespan step is already a no-op in this mode), so the remaining 30s+ is **pre-process**: container image pull (**183MB** compressed) + cold-instance interpreter/imports.

## What
Shrink the default image by making the **64MB `kreuzberg`** document-extraction lib an **opt-in `ingest` extra**:
- `kreuzberg` → `[project.optional-dependencies] ingest`; the `dev` extra keeps it so CI/local still run the ingest tests.
- `ingest_service` guards the import at **top level** (`try/except ImportError → kreuzberg = None`, rule-compliant — no inline imports); `_extract_with_kreuzberg` raises a clear **501** when the extra is absent. Every non-ingest path is unchanged.
- Dockerfile builds the slim image by default; **`--build-arg WITH_INGEST=1`** re-adds the extra for deployments serving `/ingest/*`.

## Honest scope
This is a **moderate** cold-start reduction (drops the kreuzberg tree + its import cost), **not** a full fix — a 150MB+ Python image has an inherent ~20s+ cold-start floor. The next-largest lever is **google-cloud-aiplatform (~99MB)**; making it conditional on the Vertex provider is a follow-up.

## Verification
- `ruff check` + `ruff format --check` (core-api/src) ✓; `mypy core-api/src` ✓ (192 files).
- `uv.lock` regenerated: default export (`--extra pubsub`) **excludes** kreuzberg; `--extra ingest` **includes** it.
- `pytest` ✓ — **54 ingest tests pass**, including a new test that `_extract_with_kreuzberg` returns 501 when kreuzberg is absent.

Companion to the harness cold-start-attribution finding ([caura-test-automation #172](https://github.com/caura-ai/caura-test-automation/pull/172)).

### DIFF
--- a/core-api/Dockerfile
+++ b/core-api/Dockerfile
@@ -22,7 +22,13 @@ COPY --from=ghcr.io/astral-sh/uv:0.9.18 /uv /usr/local/bin/uv
 COPY core-api/pyproject.toml core-api/uv.lock core-api/README.md* ./core-api/
 # Dummy src so the project metadata resolves during export.
 RUN mkdir -p core-api/src/core_api && touch core-api/src/core_api/__init__.py
-RUN uv export --frozen --no-dev --extra pubsub --no-emit-project \
+# Document-ingest (kreuzberg, ~64MB) is an opt-in extra — off by default to
+# keep the image small and cold-start fast. Build with --build-arg WITH_INGEST=1
+# for a deployment that serves the /ingest/* endpoints.
+ARG WITH_INGEST=0
+RUN EXTRAS="--extra pubsub"; \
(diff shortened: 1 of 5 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #470, opened 2026-06-22
feat(core-storage-api): route the skill-factory pipeline through storage (Fix 2 Ph5a)

Eldad-Caura:
## Fix 2 Ph5a — route the skill-factory pipeline through core-storage-api

First of three Ph5 PRs (**5a** skill-factory / 5b insights+evolve+`_mcp_session` deletion / 5c capability_usage). Moves the skill-factory background pipeline's direct DB access onto the storage HTTP client, removing `get_db`/`async_session`/`db.execute` from `skill_promoter`, `forge/poison`, `forge/cron_handler`, `session_trace`, and `outcome_inference/*`.

### New core-storage-api endpoints (+ PostgresService + storage_client methods)
All explicit-tenant-scoped; PG SQL ported verbatim (DISTINCT ON, supersedes self-join, `ON CONFLICT`, `jsonb_set`, `INTERVAL '1 day' * cooloff_days`, `ANY(CAST(:ids AS uuid[]))`):
- **`POST /documents/update-status`** — conditional CAS status flip (the deferred Ph3 endpoint). 404 on a no-op → client `None` → promoter raises `AlreadyTransitionedError`.
- **forge rejected-fingerprints** — write + 3-arm-fleet cooloff-window poison check.
- **session-traces** — batch upsert `ON CONFLICT (tenant_id, run_id, agent_id)`.
- **bespoke memory reads** — session-window, entity-links-batch, memory-content, + the 4 outcome signals (contradiction / supersession / cross-agent-reuse / terminal).
- **`GET`/`DELETE /documents/{collection}/{doc_id:path}`** — now accept slashed doc_ids; forge-distilled skills use `forge/<slug>` (the doc_id is only ever a bound SQL param, never a filesystem path).

### core-api migration
Callers moved to the typed client; `lifecycle_audit.forge_distill` stops opening its own session and threads `tenant_id`. Adds ORM models `ForgeRejectedFingerprint`/`SessionTrace` mirroring migrations 020/021 for the test `create_all` path (prod schema stays alembic-owned).

### Notable
The `contradictions` signal windows on `created_at`, not the source's `COALESCE(updated_at, created_at)` — OSS `memories` has **no** `updated_at` column, so the original reference was dead against the real schema.

### Out of scope (later PRs)
insights/evolve + `_mcp_session` deletion (Ph5b); capability_usage (Ph5c).

### Gates
ruff@0.15.4 check + `format --check`, mypy both projects, full suite **3560 passed**. New `tests/test_ph5a_skill_factory_storage.py` (25 tests) covers each endpoint incl. tenant isolation, CAS hit/miss, poison cooloff window, ON-CONFLICT refresh, and the slashed-doc_id round-trip.

### DIFF
--- a/common/models/__init__.py
+++ b/common/models/__init__.py
@@ -12,6 +12,7 @@ from common.models.fleet import FleetCommand, FleetNode
 from common.models.idempotency import IdempotencyResponse
 from common.models.lifecycle_audit import LifecycleAudit
 from common.models.memory import Memory
+from common.models.skill_factory import ForgeRejectedFingerprint, SessionTrace
 
 __all__ = [
     "Agent",
@@ -26,9 +27,11 @@ __all__ = [
     "Entity",
     "FleetCommand",
     "FleetNode",
+    "ForgeRejectedFingerprint",
     "IdempotencyResponse",
     "LifecycleAudit",
     "Memory",
     "MemoryEntityLink",
     "Relation",
+    "SessionTrace",
 ]
--- a/common/models/skill_factory.py
+++ b/common/models/skill_factory.py
@@ -1,0 +1,110 @@
+"""Skill Factory ORM models — ``forge_rejected_fingerprints`` (SF-003) +
+``session_traces`` (SF-004).
+
+These tables were introduced migration-first (``020_forge_rejected_fingerprints``
++ ``021_session_traces``) and accessed via raw ``text()`` SQL. Fix 2 Ph5a
+moves that access into core-storage-api; the models here mirror the migration
+DDL so the table definitions live in SQLAlchemy metadata too — which is what
+``Base.metadata.create_all`` (the test-schema path) builds from. Production
+schema is still owned by the alembic migrations; keep both in lockstep.
+"""
+
+from __future__ import annotations
+
+import uuid
+from datetime import datetime
+
+from sqlalchemy import CheckConstraint, DateTime, Index, Integer, Text, UniqueConstraint, text
+from sqlalchemy.dialects.postgresql import JSONB
+from sqlalchemy.dialects.postgresql import UUID as PG_UUID
+from sqlalchemy.orm import Mapped, mapped_column
+
+from common.models.base import Base
+
+
+class ForgeRejectedFingerprint(Base):
+    """Anti-poison memory for the Forge resident (migration 020).
+
+    One row per rejection event (the same fingerprint can be rejected
+    multiple times); the cooloff-window check is satisfied by any live row.
+    """
+
+    __tablename__ = "forge_rejected_fingerprints"
+    __table_args__ = (
+        Index(
+            "idx_forge_rejected_fp_lookup",
+            "tenant_id",
+            "cluster_fingerprint",
+            text("fleet_id NULLS FIRST"),
+            text("rejected_at DESC"),
+        ),
+    )
+
+    id: Mapped[uuid.UUID] = mapped_column(
+        PG_UUID(as_uuid=True),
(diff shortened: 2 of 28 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@claude

Eldad-Caura:
Addressed in `3d5939c` (force-pushed). Verified first:

**Medium — poison check read from replica**: confirmed. `forge_is_fingerprint_poisoned` is a write-path **guard** (the forge tick gates candidate creation on it), so a replica-lag miss could re-propose a just-rejected cluster — a regression vs the original, which ran the check in the same session as the write. Switched it to `get_session()` (primary), matching `memory_find_by_content_hash` (the existing write-path dedup guard). The pure analytics reads in the same block correctly stay on the replica (`get_read_session`).

Gates green: ruff@0.15.4 + `format --check`, mypy (core-storage-api), Ph5a + forge-cron tests (34 passed).

Eldad-Caura:
@claude

Eldad-Caura:
Addressed in `0bfc875` (force-pushed). Verified each:

**Medium — tenant isolation on the two by-id reads** (`/session-traces/entity-links`, `/forge/memories-content`): real gap, fixed. Both endpoints now `_require(body, "tenant_id")`; `memory_entity_links_batch` joins `memories` on `tenant_id` (the link table has no tenant column) and `memory_content_by_ids` adds `AND tenant_id = :tenant_id`. Threaded through the client methods and the callers (`session_trace._query_entity_ids_for_memories`, `cron_handler._make_memory_fetcher` — both already had `tenant_id` in scope). Added cross-tenant no-op assertions to both tests.

**Low — `query_documents` order_by** (`skill_promoter.py`): **declining — verified false positive.** `document_query` orders by `Document.data[order_by].astext` (i.e. `data->>'created_at'`), exactly the original `ORDER BY (data->>'created_at') ASC`, not the top-level column. FIFO promotion order is preserved.

**Suggestion — `cooloff_days >= 1` at the storage boundary**: applied (`isinstance(int) and >= 1`) so the storage layer is self-consistent (core-api already guarded it upstream).

Gates green: ruff@0.15.4 check + `format --check`, mypy both, full suite **3560 passed** (incl. the new cross-tenant isolation assertions).

Eldad-Caura:
@claude

Eldad-Caura:
Addressed both in `58d66c2` (force-pushed):

**Medium — `/reject` poison-write / status-flip inconsistency**: real operability gap. The poison write commits storage-side before `_persist_status_transition`, so a status-flip failure *after* that commit left the cluster blocked for `cooloff_days` with the doc still in a rejectable status and only a generic 500. Wrapped the status flip in a `try/except` that logs a loud operator alert ("poison row committed but doc NOT flipped … manual intervention required") and re-raises. Restoring true cross-HTTP atomicity isn't possible without a 2-phase protocol; making the rare inconsistent state *visible* is the actionable fix.

**Low — `_upsert_session_traces` `bind.pop("tenant_id")`**: added `assert bind.get("tenant_id") in (None, tenant_id)` before the pop — belt-and-braces against a smuggled cross-tenant row (storage already enforces the batch tenant).

The two "actually correct, just noting" items (the entity-links JOIN scoping; `_require` 0-falsy on the current string-only fields) need no change, per your own verification.

Gates green: ruff@0.15.4 check + `format --check`, mypy both projects, full suite **3560 passed**.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #471, opened 2026-06-22
feat(core-storage-api): route insights_service through storage (Fix 2 Ph5b, PR1 — insights)

Eldad-Caura:
## Fix 2 Ph5b — PR1: route insights_service through core-storage-api

First of two Ph5b PRs (**insights** here; **evolve + the `_mcp_session` deletion** in PR2). Moves the insights pipeline off core-api's direct DB pool. **`_mcp_session` intentionally remains** — `memclaw_evolve` still uses it; PR2 deletes it as its final step.

### New core-storage-api endpoints (router `insights.py` + PostgresService methods)
All explicit-tenant-scoped; analytic reads use SQLAlchemy ORM (sidesteps asyncpg array-cast pitfalls):
- `POST /insights/{contradictions,failures,stale,divergence,patterns}` — the 5 focus reads. `contradictions`/`divergence` keep the `GROUP BY … HAVING COUNT(DISTINCT …)` entity pre-query. Return the MEMORY_LIST shape (no embedding).
- `POST /insights/discover-sample` — embedding-bearing sample; numpy k-means + cluster summarisation stay client-side.
- `POST /insights/supersede-priors` + `POST /insights/restore-priors` — `_persist_findings`'s prior-outdate (JSONB `metadata->>'insight_focus'`/`scope` select + `UPDATE status='outdated'`, one atomic txn) and the all-failed restore.
- `POST /insights/activity-gate` — `MAX(created_at)` by insight/non-insight for the lifecycle discover gate.

### core-api migration
`insights_service` (`_query_*` + `_persist_findings`), `memclaw_insights` (both phases now `_no_db`), `/insights/generate` (drops `Depends(get_db)`), and `lifecycle_audit`'s discover pass call the storage client. `resolve_caller_and_gate` + the db-ignoring storage helpers widened to `db: AsyncSession | None` (pure type widening) so the route's `None` threads cleanly.

### Atomicity note
`_persist_findings`'s `db.begin_nested()` savepoint becomes client-orchestrated: each storage call is its own atomic txn, and the existing "all findings failed → restore priors" branch now calls `/insights/restore-priors`. The only behavioural delta — priors are committed-outdated *before* the bulk-create (vs. held in a savepoint) — is documented in the docstring.

### Out of scope (PR2)
`evolve_service` + `memclaw_evolve` + `/evolve/report`, and the `_mcp_session` deletion.

### Gates
ruff@0.15.4 check + `format --check`, mypy both projects, full suite **3584 passed**. New `tests/test_ph5b_insights_storage.py` covers every endpoint incl. tenant isolation, scope filters, the COUNT(DISTINCT) reads, the JSONB supersede + fleet arm, restore, activity-gate, and 422 guards.

### DIFF
--- a/core-api/src/core_api/clients/storage_client.py
+++ b/core-api/src/core_api/clients/storage_client.py
@@ -1581,6 +1581,152 @@ class CoreStorageClient:
         )
 
     # =====================================================================
+    # Insights (Fix 2 Ph5b)
+    # =====================================================================
+    #
+    # Analytic memory reads (one per focus), the discover-sample row fetch
+    # (embedding included; numpy k-means stays client-side), the
+    # supersede/restore priors writes, and the lifecycle activity gate. The 6
+    # reads + the gate route to the replica (post-commit analytics — replica
+    # lag is harmless); the two priors writes hit the primary. Each takes an
+    # explicit tenant_id (+ scope params) — there are no RLS GUCs server-side.
+    # ``max_memories`` / ``sample_size`` forward the core-api tuning constants
+    # so they stay the single source of truth here.
+
+    async def insights_query_contradictions(
+        self, *, tenant_id: str, fleet_id: str | None, agent_id: str, scope: str, max_memories: int
+    ) -> list[dict]:
+        return await self._post(  # type: ignore[return-value]
+            "/insights/contradictions",
+            {
+                "tenant_id": tenant_id,
+                "fleet_id": fleet_id,
+                "agent_id": agent_id,
+                "scope": scope,
+                "max_memories": max_memories,
+            },
+            read=True,
+        )
+
+    async def insights_query_failures(
+        self, *, tenant_id: str, fleet_id: str | None, agent_id: str, scope: str, max_memories: int
+    ) -> list[dict]:
+        return await self._post(  # type: ignore[return-value]
+            "/insights/failures",
+            {
+                "tenant_id": tenant_id,
+                "fleet_id": fleet_id,
+                "agent_id": agent_id,
+                "scope": scope,
+                "max_memories": max_memories,
+            },
+            read=True,
+        )
+
+    async def insights_query_stale(
+        self,
+        *,
+        tenant_id: str,
+        fleet_id: str | None,
+        agent_id: str,
+        scope: str,
+        thirty_days_ago: datetime,
+        fourteen_days_ago: datetime,
+        max_memories: int,
+    ) -> list[dict]:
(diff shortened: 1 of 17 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@claude

Eldad-Caura:
Addressed in `974b01d` (force-pushed). Verified the race:

**Medium — `insights_supersede_priors` TOCTOU**: confirmed. The SELECT-then-UPDATE captured `prior_ids` at SELECT time, not what the UPDATE actually transitioned — so under concurrent insights generation for the same focus/scope/agent, caller B could return ids caller A already outdated, and B's total-failure restore would re-activate A's legitimately-outdated priors (two generations active at once). Collapsed both into a single atomic `UPDATE … RETURNING id`: `prior_ids` is now EXACTLY the rows this call transitioned, so the restore can only touch this call's own work (a concurrent caller sees zero active priors or blocks until commit). Simpler, too — one statement.

Gates green: ruff@0.15.4 check + `format --check`, mypy both projects, full suite **3584 passed**.

Eldad-Caura:
@claude

Eldad-Caura:
Two findings from the review pass that raced the pre-fix commit (`376c9bc`) — both resolved in `b82ecc9`:

**Medium — `_persist_findings` swallowed 4xx from the supersede/restore calls**: real. `except Exception` caught `httpx.HTTPStatusError`, so a 422 (a code/contract bug, e.g. a bad `tenant_id`/`focus`) silently returned empty insights with a 200. Added `except httpx.HTTPStatusError: raise` before the broad catch in BOTH the supersede and restore blocks — 4xx now surfaces; 5xx / network still degrade gracefully.

**Low — `tenant_id` guard on the supersede UPDATE**: already addressed. The round-1 TOCTOU fix replaced the SELECT-then-UPDATE with a single `UPDATE … WHERE tenant_id = :tenant_id … RETURNING`, so the guard the symmetric `restore` has is already present. (This pass reviewed the pre-fix commit, hence the flag.)

For sequencing clarity: the fresh review of `974b01d` came back ✅ clean; `b82ecc9` = `974b01d` + the 4xx re-raise above.

Gates green: ruff@0.15.4 check + `format --check`, mypy both projects, full suite **3584 passed**.

Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please apporve

erni-a (approved):

PR #472, opened 2026-06-23
feat(core-storage-api): route evolve through storage + delete _mcp_session (Fix 2 Ph5b, PR2)

Eldad-Caura:
## Fix 2 Ph5b — PR2: route evolve through storage + **delete `_mcp_session`**

Second of the two Ph5b PRs (PR1 insights = #471, merged). Routes `evolve_service` + `memclaw_evolve` + `/evolve/report` off core-api's direct DB, and as the final step **deletes `_mcp_session`** — `memclaw_evolve` was its last consumer once PR1 moved insights to `_no_db`. This completes the mcp_server DB migration.

### New core-storage-api endpoints
- `POST /evolve/filter-by-scope` — SELECT ids by tenant+scope (ORM `.in_()`).
- `POST /evolve/apply-weights` — the weight-clamp CTE (`GREATEST`/`LEAST`, `RETURNING` old/new) **and** the rule→outcome `jsonb_set` backfill, folded into **one atomic transaction** so evolve's documented split-commit doesn't widen. `_apply_outcome_to_db` is re-sequenced so rule/outcome ids exist before this single call.

### `_mcp_session` deleted
Removed the `_mcp_session` def + the `async_session` import + an orphaned `sa_text` import. `grep -rn "_mcp_session" core-api/src/` → zero consumers.

### ⚠️ Write-pipeline made db-free (broader than evolve — flagging for review)
Evolve's outcome/rule persist now calls `create_memory(db=None)` (the STM path). Four `write_fast` steps — `load_tenant_config`, `detect_near_duplicate`, `check_semantic_duplicate`, and `write_memory_row`'s audit hook — used `ctx.require_db` (which **raises** on `db=None`) to call functions that are **already storage-routed and ignore `db`** (`resolve_config`, `_find_semantic_duplicate`, `log_action`). Switched all four to the nullable `ctx.db`: transparent for db-present callers, and it unblocks the db-free STM write path. The warm-cache full suite masked this (near-dup got skipped under the cached org config); a **clean-DB** run surfaced it.

Also cast `metadata::jsonb` in the apply-weights backfill — CAURA-595 `json`/`jsonb` column drift (mirrors the merged `memory_update` fix).

### Deviation (documented in-code)
The persisted outcome memory's `metadata.weight_adjustments` is `[]` on the apply path — the exact post-clamp deltas are in the `report_outcome` response and the audit log, but the outcome must be persisted *before* apply-weights to fold the backfill atomically. Nothing reads the persisted field.

### Gates
ruff@0.15.4 check + `format --check`, mypy both projects, **full suite 3604 passed on a clean DB** (the rigorous run — no warm-cache masking). New `tests/test_ph5b_evolve_storage.py` (filter-by-scope agent/fleet/all + tenant isolation; apply-weights clamp at floor/cap, RETURNING, backfill / no-backfill, isolation; 422 guards); reworked `test_evolve_service.py` to seed/assert via the storage client.

### DIFF
--- a/core-api/src/core_api/clients/storage_client.py
+++ b/core-api/src/core_api/clients/storage_client.py
@@ -1727,6 +1727,68 @@ class CoreStorageClient:
         )
 
     # =====================================================================
+    # Evolve (Fix 2 Ph5b, PR2)
+    # =====================================================================
+    #
+    # The scope-filter SELECT (read → replica; the filter is a pre-write
+    # visibility check and replica lag is harmless) and the atomic
+    # weight-adjust + rule→outcome backfill (write → primary, ONE txn). Each
+    # takes an explicit tenant_id (+ scope params) — there are no RLS GUCs
+    # server-side. The dedup / UUID-parse / cap / rounding / skip-reason logic
+    # stays client-side in ``evolve_service``; only the DB passes route here.
+
+    async def evolve_filter_by_scope(
+        self,
+        *,
+        tenant_id: str,
+        caller_agent_id: str,
+        fleet_id: str | None,
+        scope: str,
+        ids: list[str],
+    ) -> list[str]:
+        """Return the subset of ``ids`` the caller can touch under ``scope``."""
+        resp = await self._post(
+            "/evolve/filter-by-scope",
+            {
+                "tenant_id": tenant_id,
+                "caller_agent_id": caller_agent_id,
+                "fleet_id": fleet_id,
+                "scope": scope,
+                "ids": ids,
+            },
+            read=True,
+        )
+        return resp["allowed_ids"]  # type: ignore[index,return-value]
+
+    async def evolve_apply_weights(
+        self,
+        *,
+        tenant_id: str,
+        ids: list[str],
+        delta: float,
+        floor: float,
+        cap: float,
+        rule_id: str | None = None,
+        outcome_id: str | None = None,
+    ) -> dict:
+        """Clamp-adjust weights + (atomically) backfill the rule→outcome link;
+        primary write, ONE transaction. Returns
+        ``{adjustments:[{id, old_weight, new_weight}], backfilled}``."""
+        return await self._post(  # type: ignore[return-value]
+            "/evolve/apply-weights",
+            {
+                "tenant_id": tenant_id,
+                "ids": ids,
+                "delta": delta,
+                "floor": floor,
+                "cap": cap,
+                "rule_id": rule_id,
+                "outcome_id": outcome_id,
+            },
+            read=False,
+        )
+
+    # =====================================================================
     # Keystones (CAURA-000)
     # =====================================================================
     #
--- a/core-api/src/core_api/mcp_server.py
+++ b/core-api/src/core_api/mcp_server.py
@@ -23,7 +23,6 @@ from mcp.server.fastmcp import FastMCP
 from mcp.server.transport_security import TransportSecuritySettings
 from mcp.types import CallToolResult, TextContent
 from pydantic import Field, ValidationError
-from sqlalchemy import text as sa_text
 from starlette.types import ASGIApp, Receive, Scope, Send
 
 from common.enrichment.constants import SERVER_RESERVED_MEMORY_TYPES
@@ -40,7 +39,6 @@ from core_api.constants import (
     VALID_SCOPES,
     VERSION,
 )
-from core_api.db.session import async_session
 from core_api.errors import code_for_status
 from core_api.pagination import decode_cursor, encode_cursor
 from core_api.schemas import (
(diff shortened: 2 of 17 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
## Review fixes — pushed `6d9ab42` (amend of `d48d984`)

### ① HIGH — `contextlib.nullcontext()` crashes as `async with` on Python 3.9 → **declined (premise doesn't hold)**

The finding is correct that `contextlib.nullcontext()` only gained `__aenter__`/`__aexit__` in Python 3.10 — but the project does not target 3.9:

- `core-api` and `core-storage-api` both declare `requires-python = ">=3.12"`
- CI runs `python-version: "3.12"`; **all** Dockerfiles are `FROM python:3.12-slim` (core-api, core-storage-api, core-worker, core-operations)
- `from __future__ import annotations` here is a forward-reference convenience, not a 3.9-compat signal

On 3.12, `async with contextlib.nullcontext()` is valid, so the `db=None` path in `_persist_outcome` / `_persist_rule` does not crash. No change made.

### ② Medium — `backfilled = True` set unconditionally → **fixed**

[`postgres_service.py`](core-storage-api/src/core_storage_api/services/postgres_service.py) `evolve_apply_weights` now captures the UPDATE result and reports `backfilled = (br.rowcount or 0) > 0`, so a soft-deleted / never-committed / cross-tenant `rule_id` (0 rows touched) reports `false`. Matches the existing `(result.rowcount or 0) > 0  # type: ignore[attr-defined]` idiom used elsewhere in this file (`rowcount` lives on `CursorResult`, not the base `Result`). *(Note: no caller currently reads `backfilled` — `_adjust_weights` reads only `adjustments` — so this hardens the contract for future consumers rather than fixing an active bug.)*

### Gates (run locally, clean-DB)
- full suite — **3604 passed**, 3 skipped, 1 xfailed, 3 xpassed (TRUNCATE'd DB)

Eldad-Caura:
@claude

Eldad-Caura:
## Round-2 review fixes — pushed `0cefc2d` (amend of `6d9ab42`)

### ① Medium — `scope='fleet'` without `fleet_id` → 500 instead of 422 → **fixed**

Confirmed: the service raises `ValueError` ([`postgres_service.py:5968`](core-storage-api/src/core_storage_api/services/postgres_service.py)) which FastAPI surfaces as a 500 for a *direct* storage call. [`routers/evolve.py`](core-storage-api/src/core_storage_api/routers/evolve.py) `evolve_filter_by_scope` now returns a 422 (`"fleet_id is required when scope is 'fleet'"`) before the service call; the service `ValueError` stays as the deeper safety net. Misleading docstring clause replaced. Added `test_filter_fleet_scope_missing_fleet_id_422`.

### ② Medium — outcome memory persists `weight_adjustments: []` → **declined (deliberate, documented trade-off)**

This is a considered design decision, not a regression:

- **No consumer reads the persisted field.** I grepped every reference: the real post-clamp deltas are returned in the `report_outcome` response (`evolve_service.py:948` — what MCP/REST callers and the tests actually read) and recorded in the audit log. Nothing reads `outcome.metadata.weight_adjustments` back from the stored memory, so there is no audit consumer to break.
- **Capturing it would cost the atomicity this PR is built around.** The deltas aren't known until after `apply-weights`; writing them into the outcome would require a second outcome-memory write, widening the deliberately-single atomic clamp+backfill transaction. (The suggested reorder is also internally inconsistent — moving persist-outcome *after* `apply-weights` while keeping the rule→outcome backfill in that same call can't both hold.)
- I've tightened the in-code comment at `_apply_outcome_to_db` to state explicitly that the empty value is by-design and where the real deltas live.

If a future consumer needs the deltas on the stored outcome, the clean follow-up is one `sc.update_memory` write after `apply-weights` — out of scope for this PR.

### Gates (run locally, clean-DB)
- `ruff@0.15.4 check` + `format --check` (core-api/src, core-storage-api/src) — clean
- full suite — **3605 passed**, 3 skipped, 1 xfailed, 3 xpassed (TRUNCATE'd DB; +1 = the new 422 test)

Eldad-Caura:
@claude

Eldad-Caura:
## Round-3 review fix — pushed `dc4a1f9` (amend of `0cefc2d`)

### Low — non-UUID `ids` raise unhandled `ValueError` → 500 → **fixed**

Same boundary-hardening class as round-2's `fleet_id` guard. `evolve_filter_by_scope`'s `UUID(s)` parse raises `ValueError` on a non-UUID string, which FastAPI surfaces as a 500 for a *direct* storage caller. [`routers/evolve.py`](core-storage-api/src/core_storage_api/routers/evolve.py) now wraps the service call and re-raises as a 422 (`"invalid ids: …"`). Kept the explicit `fleet_id` guard above it (the fleet `ValueError` is pre-guarded, so only the UUID one reaches the catch — the message stays accurate). Docstring updated; added `test_filter_invalid_uuid_in_ids_422`. Happy path unaffected — the typed client always sends pre-validated UUIDs.

*(Note: `apply-weights` passes `ids` to `ANY(CAST(:ids AS uuid[]))`, so a bad UUID there surfaces as an asyncpg `DataError` rather than a Python `ValueError` — a different mechanism, and its `ids` always arrive pre-validated from `_adjust_weights`. Left as-is to keep this fix scoped; happy to file a follow-up if you want symmetric boundary validation across the storage routers.)*

### Gates (run locally, clean-DB)
- `ruff@0.15.4 check` + `format --check` (core-api/src, core-storage-api/src) — clean
- full suite — **3606 passed**, 3 skipped, 1 xfailed, 3 xpassed (TRUNCATE'd DB; +1 = new UUID test)

Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please arppveo

erni-a (dismissed):

Eldad-Caura:
## Round-4 review fix — pushed `206ff45` (amend of `dc4a1f9`)

### Low — missing `scope` validation (fail-open) → **fixed**

Verified real: the service's `if scope=='agent' … elif scope=='fleet' …` has no `else`, so an unrecognized `scope` (typo like `"AGENT"`) fell through to the no-extra-filter (`scope='all'`) arm — fail-open for a direct storage caller. [`routers/evolve.py`](core-storage-api/src/core_storage_api/routers/evolve.py) now validates against an inlined `_VALID_SCOPES = {"agent","fleet","all"}` and returns 422 on anything else (inlined rather than importing core-api's `constants.py` — core-storage-api is an independent service). Added `test_filter_invalid_scope_422`. core-api still validates `scope` via Pydantic upstream, so the normal path is unchanged.

### 🟢 Suggestion — `_processed_ids` discard → **addressed**

Added a comment in `_apply_outcome_to_db` clarifying that `_processed_ids` is *intentionally* discarded (the outcome was persisted in Phase 3 with the scope-filtered `related_ids` and isn't rewritten; per-id deltas live in the response) so the `_` prefix doesn't read as "simply unused."

### Gates (run locally, clean-DB)
- `ruff@0.15.4 check` + `format --check` (core-api/src, core-storage-api/src) — clean
- full suite — **3607 passed**, 3 skipped, 1 xfailed, 3 xpassed (TRUNCATE'd DB; +1 = new scope test)

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #473, opened 2026-06-23
fix(core-api): re-route third-party loggers from the ASGI lifespan startup

Eldad-Caura:
## Why

`configure_logging()` runs at import time (`app.py`) and calls `_route_third_party_to_root()` then — but `uvicorn` / `fastmcp` / `mcp` / `slowapi` are imported **after** that call (slowapi + `mcp_server` further down in `app.py`, uvicorn by the server). So the import-time pass **no-ops for those loggers** — it even logs the warning it was built to emit:

> `logger 'fastmcp' has no own handlers at rerouting time; rerouting was a no-op (library may not be imported yet — call _route_third_party_to_root from an ASGI lifespan startup handler if this is a uvicorn-fronted service)`

Result: those libraries' records never reach the JSON/GCP handler. **Most consequentially, FastMCP's `Error executing tool …` tool-error lines are invisible in prod** — an MCP tool that raises surfaces the error to the client but leaves **no server-side log to debug from**. (Concretely: this is why a recently-surfaced MCP `memclaw_write` failure produced a client-side `'memory'` error with nothing in Cloud Logging from FastMCP.)

## What

Add an **idempotent** `_route_third_party_to_root()` call at the start of the ASGI **lifespan startup** — by then `uvicorn`/`fastmcp`/`mcp`/`slowapi` are imported, so the re-route actually takes. This is exactly what the no-op warning instructs. The import-time call in `configure_logging()` is unchanged (still routes everything imported by then).

## Verification

- `ruff check` ✓, `ruff format --check` ✓ (repo-pinned ruff 0.15.10), `mypy` ✓ on `core-api/src/core_api/app.py`.
- The routing function is already covered by `tests/test_logging.py` (incl. an idempotency test and a "re-call after a handlerless no-op reroute" test), so the second call is safe.

### DIFF
--- a/common/structlog_config.py
+++ b/common/structlog_config.py
@@ -395,7 +395,7 @@ def _route_third_party_to_root() -> None:
         # ``Config.configure_logging`` runs at server start).
         if not lg.handlers and lg.propagate:
             logging.getLogger(__name__).warning(
-                "logger %r has no own handlers at rerouting time; rerouting was a no-op (library may not be imported yet — call _route_third_party_to_root from an ASGI lifespan startup handler if this is a uvicorn-fronted service)",
+                "logger %r has no own handlers at rerouting time; rerouting was a no-op (library may not be imported yet — call reroute_third_party_loggers from an ASGI lifespan startup handler if this is a uvicorn-fronted service)",
                 name,
             )
             continue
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@claude please re-review — addressed in `83fb901` (force-pushed after rebasing onto current main).

**Low — importing a private symbol across module boundary** ✅ Fixed exactly as suggested. Added a public wrapper in `common/structlog_config.py`:
```python
def reroute_third_party_loggers() -> None:
    """Public entry point to (re-)route third-party loggers to the root handler."""
    _route_third_party_to_root()
```
`app.py` now imports and calls `reroute_third_party_loggers()` instead of the private `_route_third_party_to_root`. Also updated the no-op WARNING text in `structlog_config.py` to name the **public** wrapper (so it no longer tells operators to call a private function).

`ruff check` + `ruff format --check` + `mypy` all clean on both files.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #474, opened 2026-06-23
fix(core-api): surface write-pipeline step failures instead of masking as KeyError

Eldad-Caura:
## Why

When a write-pipeline step raises a non-`HTTPException`, the runner (`pipeline/runner.py`) records `StepResult(FAILED)`, sets `result.failed=True`, and **`break`s without re-raising**. `create_memory` ignored `result.failed` and read `ctx.data["memory"]` directly — but `WriteMemoryRow` never ran, so that read raised **`KeyError: 'memory'`**, masking the real cause.

Concretely (the prod MCP-write regression): an MCP `memclaw_write` runs under `_no_db()` (`db=None`); the `load_tenant_config` step raises *"This pipeline step requires a DB session"*, the runner breaks, and the client saw the cryptic `Error executing tool memclaw_write: 'memory'` with **no actionable error**. Prod log confirms:
```
write_fast.load_tenant_config: FAILED — This pipeline step requires a DB session.
PipelineContext was built without one (STM path).
```

## What

Mirror the existing `_search_memories` pattern: after `pipeline.run(ctx)`, if `result.failed`, log the failing step's real error and raise `HTTPException(500)` — so the real cause is logged and the caller gets a clean error envelope instead of a `KeyError`.

- **Success path unchanged** — `result.failed` is `False` on success; the new block is skipped.
- `_exc` is set so the `memory_write_latency` `finally` still records `success=False`.
- Uses the same local `from core_api.pipeline.step import StepOutcome` import the search path uses (this module imports pipeline symbols locally throughout, to avoid an import cycle).

## Scope

**Masking fix only.** The underlying **trigger** — `load_tenant_config` requiring a DB session on the MCP `_no_db()` write path (an incomplete storage-routing migration; REST passes a real session, so it's unaffected) — is being fixed separately (ph5b storage-routing). With this change, that failure (and any future write-step failure) surfaces as a clear 500 + logged cause instead of `'memory'`.

**Follow-ups:** (1) the same `result.failed` guard should be applied to the auto-chunk persist path (the second `ctx.data["memory"]` read in this file); (2) a unit test (`db=None` → `HTTPException(500)`, not `KeyError`) is recommended alongside the trigger fix — note the analogous search-pipeline-failure handling is currently untested too.

## Verification

`ruff check` ✓, `ruff format --check` ✓ (repo-pinned ruff 0.15.10), `mypy` ✓ on `core-api/src/core_api/services/memory_service.py`.

### DIFF
--- a/core-api/src/core_api/services/memory_service.py
+++ b/core-api/src/core_api/services/memory_service.py
@@ -277,7 +277,9 @@ async def _create_memory_pipeline(db: AsyncSession, data: MemoryCreate) -> Memor
             tenant_config=stm_config,
         )
         pipeline = build_stm_write_pipeline()
-        await pipeline.run(ctx)
+        result = await pipeline.run(ctx)
+        if result.failed:
+            raise HTTPException(status_code=500, detail="STM write pipeline failed unexpectedly")
         return ctx.data["stm_response"]
 
     # Resolve tenant config BEFORE building the pipeline
@@ -291,7 +293,9 @@ async def _create_memory_pipeline(db: AsyncSession, data: MemoryCreate) -> Memor
 
         # Phase 1: Enrichment (always runs)
         enrichment_pipeline = build_enrichment_pipeline()
-        await enrichment_pipeline.run(ctx)
+        enrichment_result = await enrichment_pipeline.run(ctx)
+        if enrichment_result.failed:
+            raise HTTPException(status_code=500, detail="Memory enrichment pipeline failed unexpectedly")
 
         fields = ctx.data["memory_fields"]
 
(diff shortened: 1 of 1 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@claude please re-review — addressed in `e092bdb` (force-pushed after rebasing onto current main, post-#473-merge).

**HIGH — STM (`:280`) + enrichment (`:294`) branches still masked** ✅ Fixed. Guarded **every** `create_memory` pipeline.run site, not just the main path:
- main fast/strong write (existing)
- STM branch → `HTTPException(500, "STM write pipeline failed unexpectedly")`
- enrichment / auto-chunk branch → `HTTPException(500, "Memory enrichment pipeline failed unexpectedly")`
- the persist fall-through (the 2nd `ctx.data["memory"]` read) → 500

Each: `result = await <pipeline>.run(ctx); if result.failed: raise HTTPException(500, …)` before the `ctx.data[...]` read. (Inline at each site, matching the existing `_search_memories` form in this file.)

**LOW — double logging** ✅ Fixed. Dropped the redundant `logger.error(...)` from the guard — the runner already logs the failing step with full traceback + step/timing (`exc_info=True`), so the guards now only convert `result.failed` into the 500, no second ERROR line.

`ruff check` + `ruff format --check` + `mypy` clean on `memory_service.py`.

Eldad-Caura:
Verified this one and I'm **holding** — not applying as suggested:

1. **The suggested fix doesn't compile.** `StepResult` (`pipeline/step.py`) is `@dataclass` of `{outcome, error, detail}` — it has **no `.name`**. The step name lives on the `Step` protocol (`step.name`), which the runner uses for logging, but it is **not** stored on the `StepResult` in `result.steps`. So `result.steps[i].name` would `AttributeError` at runtime / fail mypy — the failing step's name isn't recoverable from `PipelineResult`.

2. **The signal isn't lost.** The runner already logs the failing step — with its **name, timing, and full traceback** — to GCP before it sets `result.failed`:
   ```python
   logger.error("%s.%s: FAILED in %.1fms — %s", self._name, step.name, step_ms, exc,
                exc_info=True, extra={"pipeline": ..., "step": step.name, ...})
   ```
   That's the surface on-call actually debugs from. The HTTP 500 `detail` is client-facing; keeping it generic is intentional and avoids leaking internal step/exception details (which the review also cautioned against for the message).

The only safely-extractable bit from `result` is the exception **type** (`type(result.steps[-1].error).__name__`) — I can add that to the detail if you'd prefer a little more, but since the runner already logs type + traceback + step name, the generic detail is the right call. Leaving the code as-is at `e092bdb`.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #477, opened 2026-06-23
feat(core-storage-api): route entity-linking pipeline through storage (Fix 2 Ph6)

Eldad-Caura:
## Fix 2 Ph6 — entity-linking pipeline → core-storage-api

Moves all direct DB access in the entity-linking pipeline out of core-api and behind `core-storage-api` HTTP (rule: no DB outside core-storage-api), following the proven coarse "run-op" pattern (one atomic storage call folds a multi-statement transaction). Removes 2 of the 6 remaining `async_session()` sites in core-api; the other 4 are the final-cleanup phase.

### New endpoints (on the existing `entities` router)
- **`POST /entities/resolve`** — the whole `resolve_entities` step in ONE `get_session()` txn: pgvector pair-find + union-find clustering + the per-duplicate **SAVEPOINT** merge (`begin_nested()` per dupe), with the pair-find/cluster-load reads on the *same write session* (read-your-writes). A SAVEPOINT can't cross an HTTP boundary, so the clustering+merge logic moves storage-side.
- **`POST /entities/discover-cross-links`** — candidate-find + LATERAL pgvector match + text-verify filter + bulk `ON CONFLICT` insert (single multi-VALUES `RETURNING`, CAURA-686) in one txn. Targeted + batch modes.
- **`POST /entities/infer-relations`** — co-occurrence scan + existing-relation lookup + reinforce-vs-create split + batched `UPDATE`/`INSERT ... ON CONFLICT` in one txn.
- **`POST /entities/list-null-embeddings`** + **`POST /entities/set-embeddings`** — the backfill read/write halves. The LLM `get_embedding` loop **stays in core-api** (storage has no embedding/provider chain and embedding needs per-tenant config), so backfill is read → core-api embed loop → write (2 calls).

### core-api side
The 4 entity-linking steps become DB-free (call the storage client); `lifecycle_audit.entity_link` and `entity_extraction_worker._discover_cross_links_for_memory` drop their `async_session` blocks. `db/session.py` and `PipelineContext.db` are untouched (still used by the search pipeline — later cleanup).

### Correctness notes
- SQL ported **verbatim** — prod-fixed asyncpg quirks preserved byte-for-byte: `ANY(CAST(:x AS uuid[]))`, `SET weight=:w` (not `LEAST(...)`, prod 2026-06-13), `update(Entity.__table__)` executemany (not `update(Entity)`, prod 2026-06-16), single multi-VALUES `RETURNING` (CAURA-686).
- Every statement scopes `tenant_id` (resolve's link ops via `JOIN memories`). New endpoints fail closed (422 on missing `tenant_id` / non-numeric params / invalid UUID) — they validate their own contract rather than trusting core-api.
- Behavior note: each step now commits in its own storage txn (vs the old single end-of-pipeline commit). This matches today's observable behavior — `Pipeline.run` already swallows step errors and commits the prefix, and every step is idempotent (resolve re-merges, discover/infer are `ON CONFLICT`, backfill only fills NULLs).

### Verification
- `ruff@0.15.4 check` + `format --check` (core-api/src, core-storage-api/src) — clean
- full suite — **3628 passed** on a clean (TRUNCATE'd) DB; +21 from new `tests/test_ph6_entity_linking_storage.py` (resolve merge/link-repoint/dupe-delete/tenant-isolation, discover targeted+batch+text-verify, infer create+reinforce, backfill read/write, and fail-closed 422 cases). The 3 mock-`db` pipeline tests were rewritten to mock the storage client; the SQL-shape regression anchors moved to the real-DB storage test.

### DIFF
--- a/core-api/src/core_api/clients/storage_client.py
+++ b/core-api/src/core_api/clients/storage_client.py
@@ -1190,6 +1190,125 @@ class CoreStorageClient:
     async def find_broken_entity_links(self, tenant_id: str) -> list[dict]:
         return await self._get_list("/entities/broken-links", tenant_id=tenant_id)
 
+    # ---------------------------------------------------------------------
+    # Entity-linking pipeline (Fix 2 Ph6) — thin wrappers over the coarse
+    # run-op endpoints that fold each entity-linking step into one atomic
+    # storage txn. The 4 pipeline steps call these instead of holding a
+    # direct DB session; tuning constants are passed from ``core_api.constants``.
+    # ---------------------------------------------------------------------
+
+    async def resolve_entities(
+        self,
+        *,
+        tenant_id: str,
+        fleet_id: str | None,
+        batch_size: int,
+        threshold: float,
+        candidate_limit: int,
+    ) -> dict:
+        """Merge duplicate entities (full resolve step, ONE atomic txn with a
+        SAVEPOINT per dupe). Returns ``{merge_count, clusters, cluster_errors,
+        merged_entity_ids}`` (or ``{error, cluster_errors}`` when every cluster
+        failed)."""
+        return await self._post(  # type: ignore[return-value]
+            "/entities/resolve",
+            {
+                "tenant_id": tenant_id,
+                "fleet_id": fleet_id,
+                "batch_size": batch_size,
+                "threshold": threshold,
+                "candidate_limit": candidate_limit,
+            },
+            read=False,
+        )
+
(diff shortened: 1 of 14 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #479, opened 2026-06-23
refactor(core-storage-api): consolidate router validation guards into _validation

Eldad-Caura:
## Consolidate the duplicated router validation guards

The fail-closed `_require(body, key)` guard was **byte-identical across four routers** (`evolve`, `skill_factory`, `insights`, `entities`), and `entities` had added `_require_number` while `evolve` still hand-rolled the same numeric check inline. This extracts both into `core_storage_api/routers/_validation.py` and imports them across the four routers, replacing evolve's inline `isinstance(..., (int, float))` loop with `_require_number`.

Surfaced during the Fix 2 Ph6 (#477) `/simplify` review (the duplication was flagged as out-of-scope for that PR and spun off here).

**Pure refactor** — no change to the happy path or the 422 detail messages. One intentional tightening: evolve's `delta`/`floor`/`cap` now go through `_require_number`, which rejects `bool` (a subclass of `int` that was never a valid tuning value); the 422 message is unchanged.

### Verification
- full suite — **3633 passed** (clean DB), unchanged (behavior-preserving; existing evolve/entities numeric-422 tests exercise the shared guards)
- Net −6 lines across the routers.

### DIFF
--- a/core-storage-api/src/core_storage_api/routers/_validation.py
+++ b/core-storage-api/src/core_storage_api/routers/_validation.py
@@ -1,0 +1,30 @@
+"""Shared fail-closed request-body validation guards for the storage routers.
+
+Each storage router validates its own request contract (it never trusts the
+calling service). ``_require`` / ``_require_number`` are the common primitives,
+kept in one place so the four routers that use them don't drift.
+"""
+
+from __future__ import annotations
+
+from fastapi import HTTPException
+
+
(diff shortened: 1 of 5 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #480, opened 2026-06-23
feat(recall): opt-in logging of agent-chosen recalls

arkash20:
## What

Opt-in, per-tenant logging of **agent-chosen recalls** (`memclaw_recall`) to two new tables, so we can finally answer **"why aren't good memories recalled?"** — separating *no-demand* from *below-floor near-miss* from *outranked*. Today this is unanswerable: there is no per-recall event log (`recall_count` is a bare counter).

## How

- **Migration 027** — `recall_event` (query + scope + strategy + result_count) and `recall_candidate` (one row per memory considered: rank, `memory_id` **pointer**, raw `vec_sim` cosine, `final_score`, `recall_boost`, `returned`). Candidates store only `memory_id`; content is JOINed from `memories` on demand.
- **`LogRecallEvent`** pipeline step logs the returned top-k **plus a few near-misses below the similarity floor** — the signal that distinguishes "just missed the cutoff" from "never close."
- **Gated two ways**, both cheap and evaluated before any DB work:
  1. `source == "mcp_recall"` — the plugin's automatic `/search` firehose is **never** logged.
  2. `observability.recall_logging_enabled` org setting (**default off**) — opt-in per tenant.
- Writes run **fire-and-forget** in a background task like `TrackRecalls` — zero added request latency; failures are swallowed so logging can never break a recall.

## Validation

- 3 unit tests (gating + candidate construction).
- **Wet-tested end-to-end** on a full local stack: 1 / 100 / 1000 real `memclaw_recall` calls → 1104 events logged exactly, 0 errors; near-misses + cosine scores + empty-result events all captured; a REST `/search` produced **zero** events (tool-only gating confirmed); ~1000 calls in 25s with no latency impact.

## Notes

- No behavior change to recall ranking/scoring — logging only.
- Tenant isolation is app-level (consistent with the rest of the schema); `recall_event` carries `tenant_id`.

### DIFF
--- a/common/models/recall_log.py
+++ b/common/models/recall_log.py
@@ -1,0 +1,73 @@
+"""Recall logging models — ``recall_event`` + ``recall_candidate``.
+
+A per-tenant, opt-in diagnostic log of agent-chosen ``memclaw_recall`` calls:
+what was queried, with what scope, and which memories were considered (raw
+cosine + final score), including a few near-misses below the similarity floor.
+Used to answer "why aren't good memories recalled?" — see migration 027.
+
+Candidates store only ``memory_id`` (a pointer); content is JOINed from
+``memories`` on demand, never duplicated here.
+"""
+
+import uuid
+from datetime import datetime
+
+from sqlalchemy import (
+    BigInteger,
+    Boolean,
+    DateTime,
+    Float,
+    ForeignKey,
+    Integer,
+    Text,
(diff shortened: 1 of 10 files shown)


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #482, opened 2026-06-23
feat(skills): refresh canonical + plugin memclaw skills, add company-brain

ran-taig:
Lands the reviewed, optimized skill texts and re-pins the structural invariant tests to the approved design. Skill-content only + their tests — no runtime/endpoint changes.

## Skills

- **`static/skills/memclaw/SKILL.md`** (direct-MCP canonical) — rewritten: stronger "THE long-term memory" framing; **`fleet_id`** corrected to home-fleet-resolve-on-omit (post-#465) + the no-home-fleet/cross-fleet residual; precise enrichment timing (type+PII **inline**, entities+contradiction **async**); `evolve` failure → **private rule by default**; loop spine (orient most-binding-first → write → evolve), dedicated **"How and when to write a memory"**, **trust self-awareness**, **cross-store pointer-memory**. Per-tool signature cards dropped → lean **"Behaviors the schema won't tell you"** (signatures already live in the always-in-context MCP schemas). Host normalised to **`memclaw.net`** (retires stale `.dev`).
- **`plugin/skills/memclaw/SKILL.md`** (OpenClaw variant) — same spine + fixes, OpenClaw layers kept (`<keystone_rules>` auto-injection, MEMORY.md vs workspace files, L1/L2/L3, orchestrator/subagent, recall auto-gating, TOOLS.md). Preface now makes the **automatic baseline explicit** (auto-recall, keystone inject, `MEMCLAW_AUTO_WRITE_TURNS` turn-summary) **and** that the agent must still use the tools deliberately. `keystones_set` stays withheld (`plugin_exposed=false`); `memclaw_stats` documented.
- **`static/skills/company-brain/SKILL.md`** (NEW) — concept-first posture skill layering on `memclaw`, deferring all mechanics to it.

## Tests re-pinned (approved structural change)

The redesign deliberately changed section names, **dropped signature cards** (schema-deferral), softened MUST-walls, and renamed Prohibitions → Anti-patterns. The invariant tests are updated to assert the **new** structure while keeping every meaningful guarantee (frontmatter, no-openclaw/no-TOOLS.md in static, install paths, error codes, all tool *names*, cross-store, etc.), plus a regression guard that signature cards are **not** reintroduced.

- `tests/test_direct_mcp_skill.py`
- `plugin/src/educate.test.ts`
- `plugin/src/tool-definitions.test.ts`

**Validation:** 351 plugin tests pass; 10 direct-MCP-adapter invariants pass.

## Follow-up (PR-2, intentionally split)
[...]

PR #483, opened 2026-06-23
feat(core-api): install-skill ?skill= selector + company-brain import path

ran-taig:
**PR-2** of the skills refresh (stacked on #482). Makes the new `company-brain` posture skill installable via the same one-liner as `memclaw`, plus docs.

> Base is `claude/skills-canonical-refresh` (#482) — it adds `static/skills/company-brain/SKILL.md`, which this PR serves. Retarget to `main` once #482 merges.

## Endpoint (`routes/plugin.py`)
- `GET /install-skill` gains a **`skill`** query param (default `memclaw`), **allowlisted** to `{memclaw, company-brain}` → invalid value returns 400, so the value never reaches a filesystem path or the generated script.
- The script generator threads the skill name through install dirs (`~/.claude/skills/<skill>`, `~/.agents/skills/<skill>`), the fetch URL (`/api/v1/skill/<skill>`), and the title. **For `skill=memclaw` the output is byte-for-byte the original.**
- New `GET /skill/{skill}` serving route (allowlisted) serves `static/skills/<skill>/SKILL.md`; the explicit `/skill/memclaw` route is kept and matches first, so the default resolves through the unchanged handler.

## Default-unaffected guarantee (your hard requirement)
No `?skill=` ⇒ identical to today. Pinned by the pre-existing memclaw tests **plus** a new `test_default_skill_is_memclaw_and_unchanged` (title + memclaw paths + `/skill/memclaw` URL + **absence** of any `company-brain` string).

## Docs (`README.md`)
"Install the skill" → new **Optional: the Company Brain skill** subsection + one-liner (`curl -s ".../install-skill?skill=company-brain" | bash`), a `?skill=` Options row, and the endpoints reference updated.

## Tests
`tests/test_install_skill_endpoint.py` +6 (default-unchanged guard; company-brain paths/url/title; invalid-skill 400; allowlist rejects traversal-looking value; `/skill/company-brain` serves; unknown → 404). **16 pass; ruff check + format clean.**

### DIFF
--- a/README.md
+++ b/README.md
@@ -519,9 +519,10 @@ bash /tmp/install-memclaw-skill.sh
 
 | Query param | Effect |
 |---|---|
-| (none) | Install for both Claude Code and Codex (default) |
-| `?agent=claude-code` | Only Claude Code → `~/.claude/skills/memclaw/SKILL.md` |
-| `?agent=codex` | Only Codex → `~/.agents/skills/memclaw/SKILL.md` |
+| (none) | Install the **memclaw** skill for both Claude Code and Codex (default) |
+| `?agent=claude-code` | Only Claude Code → `~/.claude/skills/<skill>/SKILL.md` |
+| `?agent=codex` | Only Codex → `~/.agents/skills/<skill>/SKILL.md` |
+| `?skill=company-brain` | Install the optional **Company Brain** posture skill instead of memclaw (see below; combine with `?agent=`) |
 
 #### Verify
 
@@ -535,6 +536,22 @@ Re-run the installer any time to pull the latest version.
 
 OpenClaw-plugin users get the skill automatically when the plugin
 installs; skip this step.
+
+#### Optional: the Company Brain skill
+
+`memclaw` teaches the agent the tools. **`company-brain`** is a thin,
+concept-first *posture* skill that layers on top: it frames the agent as one
+mind in a shared **Company Brain** and defers all tool mechanics back to the
+`memclaw` skill. Install it alongside `memclaw` when you want that framing:
+
+```bash
+curl -s "https://memclaw.net/api/v1/install-skill?skill=company-brain" | bash
+```
+
+It installs to `~/.claude/skills/company-brain/SKILL.md` (Claude Code) and/or
+`~/.agents/skills/company-brain/SKILL.md` (Codex), and obeys the same
+`?agent=` filter. The default install (no `?skill=`) is unchanged — it
+installs `memclaw` only.
 
 ---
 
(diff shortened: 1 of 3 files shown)


ran-taig:
@claude please review

ran-taig:
Both review fixes applied in f0b6dde4 (amended + force-pushed):

1. **Single source of truth** — `_VALID_SKILLS = frozenset(_SKILL_LABELS)` (derived from the labels map; the two can no longer diverge).
2. **Stale docstring** — `install_skill_script` no longer hard-codes the memclaw paths; now: *"fetches the SKILL.md adapter for the requested skill (default: memclaw) and writes it to the user-scope skills directory for the selected agent runtime(s)."*

16 endpoint tests pass; ruff check + format clean.

@claude please review

ran-taig:
Applied in b642669c (amended + force-pushed):

1. **Next-steps echo** parameterized → `echo "  2. Your agent now has {label} usage guidance available on-demand."`
2. **Sensitive-script header** parameterized → `f"# Installer for the {label} direct-MCP skill. ..."`
3. **CodeQL path-injection (High)** — `/skill/{skill}` now resolves the path and containment-checks it (`path.relative_to(base_dir)`) after the allowlist, clearing the taint flow and adding defense-in-depth.

So every user-facing string in the installer now follows the skill (`{label}`), and the serving route is doubly guarded (allowlist + path containment). 16 endpoint tests pass; ruff check + format clean.

@claude please review

ran-taig:
CI fix (CodeQL `py/path-injection`, 3 high) in 68f86ece:

The 3 alerts were all in `skill_by_name` — CodeQL doesn't accept `resolve()`+`relative_to()` as a barrier, so it still tracked `skill` → path → `is_file`/`read_text`. Fixed by removing user input from the path expression entirely: a module-level **`_SKILL_FILES = {name: _static_skills_dir / name / "SKILL.md" for name in _SKILL_LABELS}`** (keys are the constant allowlist), and the handler now does `path = _SKILL_FILES.get(skill)` — `skill` is only a dict key, the served path is always a fixed constant. That breaks the taint flow (canonical CodeQL-accepted pattern) and is simpler than the resolve/relative_to dance.

(Note: the two pre-existing `plugin_source` path-injection alerts on `main` (lines ~138-139) are the same allowlist-then-build pattern; not introduced by this PR, so out of scope here — happy to clean them up separately with the same map approach.)

16 endpoint tests pass; ruff check + format clean.

@claude please review

ran-taig:
@erni-a

erni-a (approved):

PR #484, opened 2026-06-23
feat(core-storage-api): recall/capability/ingest endpoints for Fix-2 final cleanup (PR1)

Eldad-Caura:
## Fix-2 final cleanup — PR1 of 3 (storage endpoints, additive)

Storage-side groundwork for the last Fix-2 phase: deleting core-api's direct DB pool (rule `6440b9a6`). Adds 4 endpoints + `PostgresService` methods so the remaining core-api direct-DB sites can route through HTTP in PR2/PR3. **Storage-only — no core-api files touched, nothing deleted.**

### Endpoints (all under `/api/v1/storage`)
- `POST /memories/increment-recall` — exposes the existing `memory_increment_recall` (by-id `recall_count` bump; the `track_recalls` hook target).
- `POST /memories/recall-log` — `recall_event` + N `recall_candidate` rows in one txn (the recall-logging writer that currently bypasses storage — that **rule regression is fixed in PR2**).
- `POST /capability-usage` — cross-tenant / RLS-free adoption-counter bulk insert (`capability_usage`, migration 023).
- `POST /memories/prior-ingest-by-doc-hash` — doc-hash idempotency lookup, SQL ported verbatim from `ingest_service` (writer session — replica lag would re-ingest).

Each endpoint validates its own contract and fails closed (422). The doc-hash and recall SQL is ported faithfully from the current core-api implementations.

### Also
Registers `recall_log` + `capability_usage` models in `conftest._import_all_models` — the recall-logging feature added the models but not their registration, so `create_all` wasn't building `recall_event`/`recall_candidate` in the test schema.

### Verification
- `mypy` core-storage-api — clean
- full suite — **3664 passed** on a clean (TRUNCATE'd) DB; +22 new tests in `test_fix2_cleanup_storage.py` (happy-path DB-effect assertions + fail-closed 422 cases for each endpoint)

### Plan
PR1 (this) deploys first. **PR2** rewires the 3 recall/capability core-api sites through these endpoints (fixes the live `6440b9a6` regression). **PR3** migrates the REST/ingest sites, drops the crystallizer enrichments, strips dead `get_db` params, and deletes `repositories/*` + `db/session.py`'s engine/pool + the `postgres_*` config.

### DIFF
--- a/core-storage-api/src/core_storage_api/app.py
+++ b/core-storage-api/src/core_storage_api/app.py
@@ -3,11 +3,13 @@
 from __future__ import annotations
 
 import contextlib
+import json
 import logging
 from collections.abc import AsyncIterator
 
-from fastapi import FastAPI
+from fastapi import FastAPI, Request
 from fastapi.middleware.cors import CORSMiddleware
+from fastapi.responses import JSONResponse
 
 from common.structlog_config import configure_logging
 from core_storage_api.config import settings
@@ -28,6 +30,7 @@ from core_storage_api.middleware import RejectWritesOnReaderMiddleware
 from core_storage_api.routers import (
     agents_router,
     audit_router,
+    capability_usage_router,
     debug_router,
     documents_router,
     entities_router,
@@ -83,6 +86,14 @@ def create_app() -> FastAPI:
         lifespan=lifespan,
         redirect_slashes=False,
     )
+
+    @app.exception_handler(json.JSONDecodeError)
+    async def _malformed_json_handler(request: Request, exc: json.JSONDecodeError) -> JSONResponse:
+        # Every router parses the body with a bare ``await request.json()``;
+        # malformed JSON would otherwise surface as an unhandled 500. Convert it
+        # to a fail-closed 422 app-wide so the body-validation contract holds at
+        # the transport layer too (rather than guarding 100+ call sites).
+        return JSONResponse(status_code=422, content={"detail": "request body must be valid JSON"})
 
     # Order matters: Starlette executes middlewares in reverse registration
     # order (last added = outermost). Register the reader filter FIRST so
@@ -164,6 +175,11 @@ def create_app() -> FastAPI:
     # Behind the same private-VPC posture as everything else here —
     # not exposed via the gateway.
     app.include_router(debug_router, prefix=prefix)
+    # Fix 2 final-cleanup (PR1): adoption-counter flush, moved off core-api's
+    # direct DB pool. Intentionally cross-tenant / RLS-free (migration 023) —
+    # one flush batch carries many tenants' counters. core-api's in-process
+    # aggregator POSTs here on its flush interval via the storage_client.
+    app.include_router(capability_usage_router, prefix=prefix)
 
     return app
 
--- a/core-storage-api/src/core_storage_api/routers/__init__.py
+++ b/core-storage-api/src/core_storage_api/routers/__init__.py
@@ -1,5 +1,6 @@
 from core_storage_api.routers.agents import router as agents_router
 from core_storage_api.routers.audit import router as audit_router
+from core_storage_api.routers.capability_usage import router as capability_usage_router
 from core_storage_api.routers.debug import router as debug_router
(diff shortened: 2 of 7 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
## Review fixes — pushed `674c12c`

### Medium — `increment_recall` reported input count not rows updated → **fixed**
`memory_increment_recall` now returns `result.rowcount` (typed `int`), and the endpoint returns that — so stale/deleted/cross-tenant ids no longer inflate `"updated"`. Its only caller is this new endpoint, so the return-type change is contained. Added `test_increment_recall_stale_id_not_counted` (1 real + 1 random id → `{"updated": 1}`).

### Low — `find_prior_ingest_by_doc_hash` no LIMIT → **declined (verbatim port + bounded in practice)**
This is a faithful port of core-api's `ingest_service._find_prior_ingest_by_doc_hash` (which also `SELECT`s all matching rows and Python-filters to the newest `run_id` — confirmed against `origin/main`). Two reasons to keep it as-is:
1. It's an **idempotency cache** — the lookup runs *before* ingest precisely to prevent re-ingesting the same `doc_hash`, so the "ingested many times" fan-out it worries about is the case this gate suppresses; matches are typically 0–1.
2. The suggested two-step `run_id` subquery would **change NULL semantics** — SQL `run_id = NULL` matches nothing, whereas the current/original Python `r.run_id == newest_run_id` returns the rows when `run_id` is `None`. Preserving the verbatim behavior avoids a subtle correctness drift for a Low edge case.

### Gates (clean-DB)
- `mypy` core-storage-api — clean
- full suite — **3666 passed** (+ the new stale-id test)

Eldad-Caura:
@claude

Eldad-Caura:
## Round-2 fixes — pushed `4e4ef7e`

### Medium — `run_id=None` collapsed multiple anonymous ingests → **fixed**
Good catch. `find_prior_ingest_by_doc_hash` now guards the NULL case: when the newest matching row has `run_id IS NULL`, it returns only that single newest row (instead of `None == None` matching every null-run_id ingest across runs). Behavior is unchanged when `run_id` is set (the real-ingest path). Added `test_prior_ingest_null_run_id_returns_single_newest`. *(This also retires the round-1 NULL concern I raised — the guard makes it correct rather than just preserving the original's latent bug.)*

### Low — bare `request.json()` → 500 on malformed JSON → **fixed app-wide**
Rather than patch my 4 handlers (the codebase has **101** bare `await request.json()` calls across all storage routers — patching only 4 would be inconsistent), I added a single app-level `@app.exception_handler(json.JSONDecodeError)` → 422 in `core-storage-api/app.py`. This makes the fail-closed-422 contract hold for malformed JSON across *every* storage endpoint, at the right altitude. Added `test_malformed_json_body_returns_422`.

### Gates (clean-DB)
- `mypy` core-storage-api — clean
- full suite — **3668 passed** (+ the 2 new tests; the app-level handler caused no regressions)

Eldad-Caura:
@claude

Eldad-Caura:
## Round-3 fix — pushed `33db566`

### Low — falsy non-None `candidates` coerced to `[]` → **fixed**
`recall_log` now uses an explicit `None` sentinel instead of `... or []`, so a falsy non-None `candidates` (`0`/`false`/`""`) 422s instead of slipping through as an empty-candidate call. Added `test_recall_log_falsy_non_list_candidates_422`.

### Gates (clean-DB)
- `ruff@0.15.4 check` + `format --check` — clean · `mypy` — clean · full suite — **3669 passed** (+1 test)

Per the trivial-fix workflow rule I'm **not re-tagging** for a Low validation tweak — surfacing **ready to merge**. All findings across 3 rounds are resolved (recall rowcount, run_id-None guard, app-wide malformed-JSON 422, and this); the only declined item was round-1's LIMIT (verbatim port; the run_id-None concern behind it is now fixed). PR1 is additive/storage-only and deploys first.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

erni-a (approved):

PR #485, opened 2026-06-23
fix(core-api): route recall/capability writes through storage (Fix-2 cleanup PR2)

Eldad-Caura:
## Fix-2 final cleanup — PR2 of 3 (regression fix)

Removes core-api's last recall/capability **direct-DB writes**, restoring rule `6440b9a6`. One was a **fresh regression**: the recall-logging feature shipped `log_recall_event.py` writing directly via `async_session` (bypassing storage). Builds on PR1 ([#484](https://github.com/caura-ai/caura-memclaw/pull/484), deployed to staging).

### Changes
- **`storage_client`** — adds `increment_recall` / `log_recall` / `flush_capability_usage` (thin `_post` wrappers over PR1's endpoints).
- **`log_recall_event._persist`** → `log_recall(event, candidates)`; candidate `memory_id` UUIDs stringified for the JSON payload; inline `async_session`/model imports replaced with a top-level client import.
- **`track_recalls`** → `increment_recall([str(m) for m in memory_ids])`.
- **`capability_usage._default_flush`** → `flush_capability_usage(rows)` with `ts_bucket` ISO-formatted (httpx can't encode a bare datetime; the endpoint coerces it back).
- **`app.py`** lifespan backfill drops the `async_session` wrapper (`backfill_agents` is already storage-routed; its `db` param is now optional).

### Scope note
The `on_recall` hook (`memory_repo.increment_recall`) is **kept** — it still has live `db`-bound consumers in `memory_service`/`entity_service`. Those + `memory_repo`/`db/session.py`/`repositories/*` deletion are **PR3**. (The build-spec's "no db-bound consumer" premise was inaccurate; this is the minimal consistent split.)

### Verification
- full suite — **3672 passed** clean-DB; +3 tests (each rewired site asserts it routes to the client with the right serialized payload — stringified ids, ISO `ts_bucket`); the end-to-end recall test passes via the conftest ASGI bridge to the new endpoint.

### DIFF
--- a/core-api/src/core_api/app.py
+++ b/core-api/src/core_api/app.py
@@ -198,16 +198,15 @@ async def lifespan(app):
 
             init_standalone()
 
-        # Backfill agent rows for any memories written before agent tracking
+        # Backfill agent rows for any memories written before agent tracking.
+        # Fully storage-routed (backfill_agents → sc.backfill_from_memories), so
+        # no core-api DB session is opened here.
         try:
-            from core_api.db.session import async_session
             from core_api.services.agent_service import backfill_agents
 
-            async with async_session() as db:
-                count = await backfill_agents(db)
-                if count:
-                    await db.commit()
-                    print(f"[startup] Backfilled {count} agent(s) from memories")
+            count = await backfill_agents()
+            if count:
+                print(f"[startup] Backfilled {count} agent(s) from memories")
         except Exception as e:
             print(f"[startup] Agent backfill skipped: {e}")
 
--- a/core-api/src/core_api/clients/storage_client.py
+++ b/core-api/src/core_api/clients/storage_client.py
@@ -1012,6 +1012,43 @@ class CoreStorageClient:
         return result  # type: ignore[return-value]
 
     # =====================================================================
+    # Recall logging + capability usage (Fix 2 final cleanup — the last
+    # core-api direct-DB writes, now routed through storage). All three are
+    # fire-and-forget background writes, so ``read=False`` (writer fleet).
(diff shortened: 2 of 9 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
## Review fixes — pushed `c43646d`

### Medium — `on_recall` hook bypassed in `track_recalls` → **addressed (verified behavior-preserving + documented)**
Verified the hook's premise: `ServiceHooks.on_recall` has exactly **one** registration in the codebase — `on_recall=memory_repo.increment_recall` (app.py:221) — i.e. precisely this recall-count bump, which now lives behind the storage `increment_recall` endpoint. There is no custom/extension `on_recall` with other side effects, so calling storage directly is behavior-preserving (not a silent regression). Added a comment in `_track_recalls_background` explaining the supersession (matches the prompt's "if the hook only does the DB increment" path). The hook itself stays for its other consumers (`memory_service`/`entity_service`) → PR3.

### Low — `backfill_agents` accepted an ignored `db` → **fixed**
Its only caller (`app.py` lifespan) now invokes it without a session, so I dropped the `db` param entirely rather than leaving a confusing accepted-but-ignored shim.

### Gates (clean-DB)
- `ruff@0.15.4 check` + `format --check` — clean · `mypy` core-api + core-storage-api — clean · full suite — **3672 passed**

Surfacing **ready to merge** (not re-tagging — a verified-no-op + clarifying comment plus a Low param removal, per the trivial-fix rule). Both findings resolved.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #486, opened 2026-06-23
feat(core-storage-api): crystallizer entity-coverage + audit-usage reads (Fix 2 final-cleanup PR3a)

Eldad-Caura:
## What

Additive **storage-only** step toward deleting core-api's direct DB pool (rule `6440b9a6`). Adds two read endpoints on the memories router that will replace `crystallizer_service`'s last two service-level direct-DB sites — the cutover + pool deletion land in **PR3b** after this deploys.

- `GET /memories/entity-coverage` → `{memories_with_entities}` — distinct memories with ≥1 entity link; ports `_compute_health`'s cross-table `COUNT(DISTINCT mel.memory_id)` verbatim, scoped by `tenant_id` (+ optional `fleet_id`). The caller divides by total memories for the pct.
- `GET /memories/audit-usage` → `{agent_activity, peak_hours}` — ports `_compute_usage`'s two `audit_log` queries verbatim. **`search_write_ratio` is intentionally dropped** — its `usage_counters` table does not exist in the OSS schema (verified: only referenced in `crystallizer_service.py`, no model/migration).

Both reads use `get_read_session` (replica). No core-api changes in this PR.

## Why split

Mirrors the PR1→PR2 rhythm: this additive storage PR deploys first so PR3b's core-api cutover never calls an endpoint that isn't live yet.

## Gates (local, clean-DB)
- `mypy` core-storage-api — clean
- full suite — **3688 passed**, 3 skipped, 1 xfailed, 3 xpassed (+9 new crystallizer-storage tests)

### DIFF
--- a/core-storage-api/src/core_storage_api/routers/memories.py
+++ b/core-storage-api/src/core_storage_api/routers/memories.py
@@ -720,6 +720,24 @@ async def get_type_distribution(
     return {"type_distribution": stats.get("type_distribution", {})}
 
 
+@router.get("/entity-coverage")
+async def get_entity_coverage(
+    tenant_id: str,
+    fleet_id: str | None = None,
+) -> dict:
+    """Crystallizer entity-extraction coverage: distinct memories with >=1 entity
+    link. The caller divides by total memories for the pct."""
+    count = await _svc.memory_entity_coverage_count(tenant_id, fleet_id)
+    return {"memories_with_entities": count}
+
+
+@router.get("/audit-usage")
+async def get_audit_usage(tenant_id: str) -> dict:
+    """Crystallizer usage metrics from audit_log: top agent activity + peak
+    hours. (search_write_ratio is omitted — no usage_counters in OSS.)"""
+    return await _svc.memory_audit_usage_stats(tenant_id)
+
+
 @router.get("/recent")
 async def get_recent_memories(
     tenant_id: str,
--- a/core-storage-api/src/core_storage_api/services/postgres_service.py
+++ b/core-storage-api/src/core_storage_api/services/postgres_service.py
@@ -2155,6 +2155,93 @@ class PostgresService:
             result = await session.execute(stmt)
             return result.scalar() or 0
 
+    async def memory_entity_coverage_count(
+        self,
+        tenant_id: str,
+        fleet_id: str | None = None,
+    ) -> int:
+        """Count distinct memories that have at least one entity link.
+
+        Ports the crystallizer ``_compute_health`` cross-table COUNT; the
+        coverage pct is computed caller-side against total memories. The query
+        is a fully static string (no f-string interpolation) — fleet stays
+        optional via ``CAST(:fleet_id AS text) IS NULL``. The CAST is required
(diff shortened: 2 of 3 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
## Review fixes — pushed `0daef9e` (amend of `5969ff2`)

These queries are verbatim ports of `crystallizer_service._compute_usage`, so I weighed each finding against preserving the original behavior:

### ① NULL `agent_id` rows in the response (Medium) → **fixed**
Added `AND a.agent_id IS NOT NULL` to the `agent_activity` query. A `{"agent_id": null}` group is meaningless in a per-agent breakdown; filtering it removes the noise row without changing the real per-agent totals.

### ② `searches` not scoped to `resource_type = 'memory'` (Low) → **fixed**
Scoped the `searches` FILTER to `resource_type = 'memory'`, symmetric with `writes`. (In OSS there's no `action='search'` audit-logging path today, so this is effectively a no-op that just makes the two FILTERs consistent.)

### ③ Unbounded audit scan / no time window (Low) → **declined (preserve behavior)**
The source query had no time bound; adding a rolling window changes the report's coverage (all-time vs windowed), which is a product/semantic decision rather than a migration concern — and the `tenant_id` index already prevents a full-table scan. Left verbatim; flagged as a candidate for a separate perf ticket if a high-volume tenant ever makes it matter.

Added `test_audit_usage_excludes_null_agent_and_non_memory_searches` to lock ①+②.

### Gates (local, clean-DB)
- `mypy` core-storage-api — clean
- full suite — **3689 passed**, 3 skipped, 1 xfailed, 3 xpassed (+1 new test)

Eldad-Caura:
@claude

Eldad-Caura:
## Round-2 review fixes — pushed `7ddb8ff` (amend of `0daef9e`)

### ① Medium — f-string SQL fragment in `memory_entity_coverage_count` → **fixed (with a correction)**
Replaced the f-string clause with a fully static query. **One deviation from the suggested prompt:** the proposed `(:fleet_id IS NULL OR m.fleet_id = :fleet_id)` form fails under asyncpg — `ProgrammingError: could not determine data type of parameter` (verified: all four entity-coverage tests errored with it). Used the asyncpg-safe `CAST(:fleet_id AS text) IS NULL OR m.fleet_id = CAST(:fleet_id AS text)` instead (the CAURA-595 NULL-bind cast rule). Same goal — no f-string interpolation — and it actually runs.

### ② Low — `peak_hours` not scoped to `resource_type='memory'` → **fixed**
Added `AND a.resource_type = 'memory'` so peak hours can't be distorted by non-memory events, symmetric with `agent_activity`. New test `test_audit_usage_peak_hours_excludes_non_memory` locks it in.

### Gates (local, clean-DB)
- `mypy` core-storage-api — clean
- full suite — **3690 passed**, 3 skipped, 1 xfailed, 3 xpassed

Both findings addressed; `7ddb8ff` is ready to merge.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #487, opened 2026-06-24
fix(core-storage-api): followers skip the migration lock once schema is at head

Eldad-Caura:
## Problem

core-storage-api runs alembic migrations on startup, serialized by a Postgres advisory lock: one replica (leader) migrates; the others (followers) **poll to acquire the lock, then run a redundant no-op `upgrade`**. During a multi-replica scale-up, that serialized `acquire → no-op → release` tail means tail replicas wait far longer than the leader's actual migration — long enough to blow the Cloud Run **240s startup TCP probe** (`failureThreshold: 1`), so Cloud Run kills them.

**Observed (prod 2026-06-24 02:04Z):** `prod-memclaw-core-storage-writer` — `Default STARTUP TCP probe failed … DEADLINE_EXCEEDED. The instance was not started.` Killed instances → brief storage capacity dips → the `core-api → core-storage-api` `httpx.ConnectTimeout`s that raise + nack the `memclaw.memory.embedded` pubsub handler (`consumer.py:184 → storage_client.get_memory`). This PR addresses the storage-startup root cause of that connect-timeout class.

## Change

`_schema_is_at_head(connection, head)` lets a non-leader replica start serving **as soon as the schema reaches head, without acquiring the lock**:
- **Fast path** before the lock loop: if nothing is pending, return immediately — no replica even contends on the lock (steady-state boot).
- **Follower poll loop**: bail the moment the leader's migration lands, so all waiting followers unblock *together* instead of draining one-at-a-time through the lock.

Unchanged / preserved:
- **Leader path** still acquires the advisory lock and runs the real `upgrade`/`stamp`.
- **Fresh DB** (`current is None`) still falls through to the stamp/upgrade path under the lock.
- `head` is computed **once** (constant per process), so the per-poll follower check is a single `alembic_version` SELECT — not a migrations-dir re-walk each poll.

## Test

`tests/test_init_schema_at_head.py` — unit-tests the predicate (at-head → True; pending → False; fresh/`None` → False) by stubbing `MigrationContext` (no DB). `ruff`/`mypy` clean.

## Residual (by design)

This removes the **serialization** overhead, not the **per-migration runtime floor**: a single migration that runs longer than the startup-probe deadline still kills followers (they legitimately can't serve until the schema is ready). That case requires raising the startup-probe deadline or moving migrations off the boot path (the known longer-term direction) — out of scope here.

### DIFF
--- a/core-storage-api/src/core_storage_api/database/init.py
+++ b/core-storage-api/src/core_storage_api/database/init.py
@@ -17,6 +17,31 @@ _engine: AsyncEngine | None = None
 _read_engine: AsyncEngine | None = None
 
 
+def _schema_is_at_head(connection: Connection, head: str | None) -> bool:
+    """True when the DB's Alembic revision already equals ``head``.
+
+    Lets a replica that isn't the one running the migration start serving as
+    soon as the schema is current — WITHOUT acquiring the migration advisory
+    lock — instead of queueing behind every other booting replica to acquire
+    the lock and run a redundant no-op ``upgrade``. That serialized
+    acquire-then-no-op is what pushed tail replicas past the Cloud Run startup
+    probe during multi-replica scale-ups (prod 2026-06-24 02:04: storage-writer
+    instances killed on a 240s STARTUP TCP probe DEADLINE_EXCEEDED).
+
+    ``head`` (the script head) is computed once by the caller and passed in —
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #488, opened 2026-06-24
fix(logging): demote handler-less reroute 'no-op' to DEBUG (records aren't lost)

Eldad-Caura:
## What

`common/structlog_config.py`'s `_route_third_party_to_root` emitted a **WARNING** ("logger X has no own handlers at rerouting time; rerouting was a no-op …") whenever a listed third-party logger (`mcp`/`fastmcp`/`uvicorn.error`/`slowapi`) was handler-less + `propagate=True`. Demote it to **DEBUG** + reword.

## Why

That state is **benign, not a routing failure**: a handler-less, `propagate=True` logger already flows to the root handler. **Verified in prod** — `mcp.server.streamable_http` / `uvicorn.error` records land in Cloud Logging as structured JSON via propagation. So the WARNING was a **misleading false alarm**, firing ~once per instance start (~57/day for `fastmcp` in prod core-api), implying records wouldn't reach GCP when they do. It nearly mis-led triage into thinking FastMCP tool-errors were invisible (the very thing #473 was about — which is actually working).

## Safety

Routing behavior is unchanged — only the log level + message. The snapshot/idempotency guard is untouched, and the empty state is still **not** snapshotted, so a later reroute can still act if a library ever installs its own handlers / sets `propagate=False`.

## Test

Existing 21 logging tests pass; adds `test_route_third_party_to_root_pristine_logger_no_warning` asserting a pristine handler-less logger emits no WARNING. ruff + mypy clean.

### DIFF
--- a/common/structlog_config.py
+++ b/common/structlog_config.py
@@ -387,15 +387,20 @@ def _route_third_party_to_root() -> None:
         # emit a spurious WARNING about rerouting being a no-op.
         if name in _third_party_logger_original_state:
             continue
-        # Library may not have initialised this logger yet — there's nothing
-        # to reroute right now. Don't snapshot the empty state (would freeze
-        # us into "no work to do" forever) and warn the operator at WARNING
-        # so they notice in Cloud Logging if a uvicorn-fronted service is
-        # calling configure_logging at module import (before uvicorn's own
-        # ``Config.configure_logging`` runs at server start).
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #489, opened 2026-06-24
fix(recall): tighten recall-brief grounding to stop invented details

Eldad-Caura:
## What
Strengthen `RECALL_PROMPT` (recall-brief summarization) grounding.

## Why
Nightly `recall-brief-ungrounded` (HIGH): the summary judge flagged an unsupported claim — *"Project Zephyr is scheduled to be completed by September 14, 2026"* — a date present in **no** memory (tenant A: 4/5 grounded, mean 4.8). The prompt already had grounding rules, but the model still *completed* a missing date.

## Change
- Lead with: answer using **ONLY** the memories as source of truth — no outside/world knowledge.
- Every name/date/number/title/identifier in the answer **must appear verbatim** in the memories; never invent/estimate/approximate/complete a missing value (explicitly: don't supply a completion date if none is stated).
- A detail the question asks for but the memories lack → say it's **not recorded** rather than supplying one.

Prompt-only; no logic/API-shape change. No unit test pins the prompt text; grounding is judged by the nightly LLM judge (which caught this), so the next run re-measures.

## Follow-up (not this PR)
Gate releases on a groundedness threshold — the nightly already computes the score; making it a hard gate is @ran / test-automation side.

### DIFF
--- a/core-api/src/core_api/services/recall_service.py
+++ b/core-api/src/core_api/services/recall_service.py
@@ -19,18 +19,23 @@ logger = logging.getLogger(__name__)
 
 RECALL_PROMPT = """\
 I will give you several facts and observations from past interactions and ingested content. \
-Please answer the question based on the relevant facts. Answer the question step by step: \
-first extract all the relevant information, then reason over the information to get the answer.
+Answer the question using ONLY those memories as your source of truth — do not use any outside \
+or world knowledge. Answer step by step: first extract the relevant facts from the memories, \
+then reason over only those facts to reach the answer.
 
 When the question requires combining facts from different memories, trace the connection \
 explicitly. Pay attention to dates within the facts — events described in past tense \
 occurred before the date the memory was recorded.
 
 Grounding rules — follow strictly:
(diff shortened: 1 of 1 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
Both applied in `2a5b2dd` — good catches, I'd narrowed the rules when rewriting:

- **[Medium] verbatim-quotation guard** — re-added as its own bullet: *"When using quotation marks, quote only text that appears word-for-word in the memories; do not paraphrase inside quotes."* Closes the `"completed successfully"`-in-quotes surface the old rule covered.
- **[Low] "field names"** — added back to the verbatim list (`name, date, number, title, field name, and identifier`), since memories are serialized as JSON so schema keys are a real hallucination surface.

Kept the stronger new wording (memories-only / no world knowledge / no estimated-or-completed values). `py_compile` + ruff clean.

Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #491, opened 2026-06-24
refactor(core-api): delete direct DB pool, route all DB access via core-storage-api (Fix 2 final cleanup)

Eldad-Caura:
## What

Final Fix-2 cleanup: **delete core-api's direct DB pool** and route every remaining DB access through core-storage-api over HTTP. This fulfils the "no direct DB outside core-storage-api" invariant for core-api. Last phase of the series — follows OSS #484 (PR1), OSS #485 (PR2), OSS #486 (PR3a), all merged.

## Changes

- Delete `db/session.py` (engine / `async_session` / `get_db`) and all of `repositories/*`; delete `services/memory_stats.py`.
- Relocate the tenant-context setters to `core_api/tenant_context.py` (`auth.py` updated) and `_relation_weight` to `constants.py`.
- Remove the `on_recall` hook; recall paths call the storage client's `increment_recall` directly.
- Reroute the remaining class-A consumers to storage-client methods: memories list/stats/tenants, ingest doc-hash idempotency, crystallizer entity-coverage + audit-usage, insights queries, and evolve persistence. These consumers now read storage **dict rows by key** (not ORM attributes).
- Add storage-client methods `get_entity_coverage` / `get_audit_usage` / `find_prior_ingest_by_doc_hash` (the endpoints already landed in #484/#486).
- Strip ~51 vestigial `db` params, 41 `Depends(get_db)`, dead `commit`/`rollback`, and `postgres_*` / `db_pool_size` / `db_max_overflow` / `database_url` from config; drop the now-deleted `db.session` / `repositories` entries from the mypy overrides.
- Add `tests/test_no_direct_db_outside_storage.py` (AST guard) so a direct DB pool can't be reintroduced into core-api.

## Compatibility

No public contract change — routes, response models, request schemas, and MCP tool schemas are unchanged. This is internal DB plumbing only; the ~550 internal test updates are because they call internal functions directly.

## Verification

- `ruff check` + `ruff format --check` (core-api/src, core-storage-api/src, common) — clean
- `mypy` core-api (182 files) + core-storage-api (68 files) — clean
- Full pytest (no `--timeout`, mirrors CI): **3669 passed, 0 failed**, no hangs

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #493, opened 2026-06-25
fix(core-api): correct get_memory_ids_by_entity_ids return type to list[dict]

Eldad-Caura:
## What

Correct a misleading return-type annotation on `CoreStorageClient.get_memory_ids_by_entity_ids`.

The method was annotated `-> list[tuple]`, but the storage endpoint it calls (`POST /entities/memory-ids-by-entity-ids`, see `core-storage-api/.../routers/entities.py`) returns `-> list[dict]` — rows of `{"memory_id", "entity_id", "role"}`. The sole consumer, `memory_service._entity_boost_pipeline`, already reads the result by key (`item["memory_id"]` / `item["entity_id"]`) and constructs tuples *from* those dicts — which is likely how the `list[tuple]` annotation crept in.

## Why it's safe

Annotation-only change — no runtime behaviour changes. The method already returned dicts; this just makes the type hint match reality (and the storage endpoint's own signature). `storage_client` is in core-api's mypy `ignore_errors` set, so this was never type-checked; the fix is purely for human readers. Spun out of the #491 review as a noted non-blocking follow-up.

## Verification

- `ruff check` + `ruff format --check` (core-api/src, core-storage-api/src, common) — clean
- `mypy` core-api (182 files) — clean

---

### DIFF
--- a/core-api/src/core_api/clients/storage_client.py
+++ b/core-api/src/core_api/clients/storage_client.py
@@ -1239,7 +1239,7 @@ class CoreStorageClient:
     async def get_memory_ids_by_entity_ids(
         self,
         entity_ids: list[str],
-    ) -> list[tuple]:
+    ) -> list[dict]:
         result = await self._post(
             "/entities/memory-ids-by-entity-ids",
             {"entity_ids": entity_ids},


Eldad-Caura:
@claude please review

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #495, opened 2026-06-28
fix(core-api): stop passing removed positional db to log_action on write path

ran-taig:
## Why (production regression — found on eToro v2.11.0 cutover)

`#491` (delete core-api's direct DB pool; route all DB via core-storage) made `audit_service.log_action` **fully keyword-only** by dropping its `db` parameter, and updated the `mcp_server` / `agents` / `documents` call sites — **but missed the pipeline write path.**

`write_memory_row.execute()` still passed `ctx.db` **positionally** to the `audit_log` hook (wired directly to `log_action`), so on **every memory create**:

```
TypeError: log_action() takes 0 positional arguments but 1 positional argument (and 6 keyword-only arguments) were given
```

This is caught by the step's `try/except` and logged as `Audit hook failed (non-critical)` — so **writes still persist, but no audit-log entry is recorded for memory creates.** Other audited actions (agents, documents, MCP) are unaffected (their call sites were updated by #491).

**Still present on `main`** → affects all v2.18.0 deployments, including SaaS.

## Fix
Drop the positional `ctx.db` argument (the call is otherwise correct — keyword-only). One line.

## Tests — `tests/test_write_path_audit_signature_regression.py`
- **unit:** asserts `log_action` stays fully keyword-only (no positional `db`) — guards the contract callers rely on. (Runs without a DB.)
- **integration:** drives the real write pipeline and asserts the audit hook does **not** log `Audit hook failed`. The existing `test_mcp_write_db_none_regression` can't catch this — it only asserts the row persists, which it does *despite* the swallowed audit failure.

Verified locally: unit test passes, `ruff check`, `ruff format --check`, `mypy` clean. (Integration test runs in CI — needs the pgvector DB.)

## Rollout
Targets `caura-memclaw`. eToro is live on v2.11.0 with this regression (memory-write audit gap since cutover); once merged we cut `backend-v2.18.1` → rebuild on-prem `v2.11.1` → redeploy eToro via the gated `upgrade.sh`. Service is otherwise healthy (pool fix working); rollback was rejected (would reintroduce the pool leak).

### DIFF
--- a/core-api/src/core_api/pipeline/steps/write/write_memory_row.py
+++ b/core-api/src/core_api/pipeline/steps/write/write_memory_row.py
@@ -106,7 +106,8 @@ class WriteMemoryRow:
         if _hooks.audit_log:
             try:
                 await _hooks.audit_log(
-                    ctx.db,  # log_action ignores db (storage-routed) — allow STM path
+                    # log_action is keyword-only since #491 dropped the direct DB
+                    # pool (storage-routed); do NOT pass ctx.db positionally.
                     tenant_id=data.tenant_id,
                     agent_id=data.agent_id,
                     action="create",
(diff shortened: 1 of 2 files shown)


ran-taig:
@claude

ran-taig:
@erni-a

erni-a (approved):

PR #497, opened 2026-06-28
fix(core-api): don't 422 the whole heartbeat on an oversized observability blob

ran-taig:
## Why

`HeartbeatIn`'s `recall_metrics` (4 KB) and `reconcile` (8 KB) field-validators **raised** when oversized. A raised `field_validator` fails the *entire* request model → FastAPI **422** → the heartbeat is rejected wholesale, dropping the node's **registration + command channel** (auto-upgrade, remote ops) over one bloated **optional** observability field.

**eToro 2026-06-28:** a node whose (legacy, unfiltered) skill catalog pushed the `reconcile` summary past 8 KB 422'd on *every* heartbeat → went stale in the fleet view and uncommandable. Memory tooling was unaffected (different path) and the on-disk skill reconcile still ran (it happens client-side *before* the POST), so it was non-blocking for the agent — but the fleet control channel was broken for that node.

## Fix
Degrade gracefully. A shared `_cap_or_drop()` helper replaces an oversized blob with a small `{"_truncated": True, "_original_bytes": N}` marker (and logs a warning) instead of raising — so the load-bearing heartbeat still lands; only the detailed snapshot for that tick is lost. Still bounds `nodes.metadata` growth.

Applied to **both** caps (same bug class). It's unfixed on `main`, so this affects all deployments.

## Tests — `tests/test_fleet_heartbeat_oversized_caps.py`
- oversized `reconcile` / `recall_metrics` → **dropped, not rejected** (no `ValidationError`, replaced with the marker)
- normal blobs pass through unchanged
- absent fields stay `None`

Verified locally: 4/4 pass, `ruff check`, `ruff format --check`, `mypy` all clean.

## Context / rollout
Not urgent for eToro right now — the server skill catalog is empty there, so summaries are small and heartbeats are landing fleet-wide; the observed 422 was transient. This lands as **defense-in-depth** so a future catalog growth (or any large node) can't silently knock nodes off the fleet command channel. Can ride a normal release; no redeploy required immediately.

### DIFF
--- a/core-api/src/core_api/routes/fleet.py
+++ b/core-api/src/core_api/routes/fleet.py
@@ -74,6 +74,31 @@ class FleetCreateIn(BaseModel):
         return v
 
 
+def _cap_or_drop(v: dict | None, limit: int, field: str) -> dict | None:
+    """Cap an OPTIONAL observability blob without failing the whole heartbeat.
+
+    ``recall_metrics`` / ``reconcile`` are best-effort observability. A
+    ``field_validator`` that *raises* on an oversized value fails the entire
+    request model → 422 → the node's registration AND command channel are
+    dropped over one bloated optional field (eToro 2026-06-28: a node with a
+    large skill catalog 422'd every heartbeat, going stale + uncommandable).
+    Degrade gracefully instead: replace the oversized blob with a small marker
+    so the load-bearing heartbeat still lands; only the detailed snapshot is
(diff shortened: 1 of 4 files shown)


ran-taig:
@claude

ran-taig:
@claude

ran-taig:
@erni-a

erni-a (approved):

PR #503, opened 2026-06-29
feat(observability): log auto /search recalls behind a separate flag

arkash20:
## What
Adds a second, independent per-tenant flag **`observability.search_recall_logging_enabled`** (default off) that turns on recall logging for the high-volume automatic plugin **`/search`** path — separately from the existing `recall_logging_enabled` (which gates the agent-chosen `memclaw_recall` tool).

`LogRecallEvent` now logs two modes through the one shared search pipeline:
| source | flag | what's logged |
|---|---|---|
| `mcp_recall` | `recall_logging_enabled` | full: returned candidates **+** below-floor near-misses |
| `search` | `search_recall_logging_enabled` | **light**: returned candidates only (id + `vec_sim` + `final_score`), no near-misses |

`event.source` now reflects the real source instead of being hardcoded to `mcp_recall`.

## Why
We want to study *why good memories aren't recalled* on the path that actually dominates real traffic — the auto `/search` (the MCP tool is rarely used). The light mode keeps write volume down on a busy path while still capturing which memories surfaced and their scores.

## Operational note (volume)
`/search` is high-volume (≈12.7k/day on a large tenant). This is intended to be enabled on **a couple of tenants for a short diagnostic window, then turned off** — no deploy needed to toggle (settings API). Near-misses are dropped for this path specifically to limit row growth. A retention/prune policy is advisable before any broad/long enablement.

## Tests
`tests/pipeline/test_log_recall_event.py` extended: search-flag-off skip, mcp-flag-off skip, unknown-source skip, mcp full (returned+near-misses), search light (returned-only). ruff + format + mypy clean; 28 passing.

### DIFF
--- a/core-api/src/core_api/pipeline/steps/search/log_recall_event.py
+++ b/core-api/src/core_api/pipeline/steps/search/log_recall_event.py
@@ -5,10 +5,13 @@ returned top-k *and* a few near-misses below the similarity floor) so we can
 answer "why aren't good memories recalled?" — distinguishing *nobody asked*
 from *just missed the cutoff* from *outranked*.
 
-Gating (both cheap, evaluated before any DB work):
-  1. ``source == "mcp_recall"`` — ONLY the agent-chosen tool. The plugin's
-     automatic ``/search`` is never logged.
-  2. ``tenant_config.recall_logging_enabled`` — per-tenant opt-in (default off).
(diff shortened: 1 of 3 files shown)


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #505, opened 2026-06-29
feat(core-api): reserve agent_id "main" on the write path (Phase 1)

ran-taig:
## Why

Bare `agent_id="main"` is the OpenClaw plugin's **unset default** — without a per-install `MEMCLAW_AGENT_ID`, every install collapses onto one shared identity (the eToro "firehose": ~52K memories, ≥15 installs under one id). This PR reserves `main` on the write path so new writes stop refilling that bucket. It's the **delete-gate** for the bare-`main` cleanup and the first slice of the identity fix.

## What (Phase 1 — behind a flag, default `warn`)

- **`enforce_reserved_write_id`** (new `services/agent_identity.py`) enforced at the **`create_memory` / `create_memories_bulk` service boundary** — the single funnel for **REST, MCP and STM**. `data.agent_id` is the resolved *effective* id there, so:
  - a write supplying a **unique** `agent_id` (incl. the plugin's `main-<install_id>`) **passes** — the escape hatch;
  - only bare `main` is refused, with an **actionable 409** telling the agent how to set its identity (`MEMCLAW_AGENT_ID` / per-call arg / `main-<install_id>` from `install.json`). The MCP write tool's existing `except HTTPException` surfaces the detail into its error envelope.
- Reserved-id set **unified** in `core_api.agent_ids` (shared with the existing `mcp-agent` gateway guard). `mcp-agent` stays **standalone-safe** (guarded conditionally there, not unconditionally here).
- `warn` mode logs `reserved_agent_write` (core-api has no metrics backend → the log is the counter) so we confirm bare-`main` writes → ~0 before flipping to `reject`.

## Phase 2 (ships dark, `bind_write_identity_to_auth=false`)

Binds a write's `agent_id` to the verified credential identity, ignoring a body override (spoof hardening). Wired into both REST inner handlers but **off**; enable only **after** the reserved-`main` credentials are re-identified (else it pins them back onto `main`).

## Rollout

`warn` → measure → flip `reserved_agent_id_policy=reject` (config, no deploy) → run the bare-`main` delete → re-identify creds → flip `bind_write_identity_to_auth=true`. Instant rollback via the flags.

## Tests

`tests/test_agent_identity_guard.py` — 19 cases (policy ladder, escape hatch, `mcp-agent` standalone-safety, 409 guidance content, warn log). All green locally.

## Reviewer notes

1. **`mcp-agent` intentionally *not* unconditionally reserved** — standalone uses it; only `main` is unconditional. (Softens the "absorb both" suggestion to avoid a standalone regression.)
2. Enforcement is at the **service boundary**, not the REST handlers, so the MCP path (`memclaw_write → create_memory`) is covered — the gap the design review caught.

Refs: design `pr-plan-reserved-main-identity.md`, sequencing `roadmap-main-identity-fix.md`. Companion: PR-2 (plugin default-id), enterprise provisioning guard, re-identify script.

### DIFF
--- a/core-api/src/core_api/agent_ids.py
+++ b/core-api/src/core_api/agent_ids.py
@@ -11,3 +11,17 @@ otherwise couple the route module to the whole MCP tool surface.
 # explicitly refused (see ``mcp_server._refuse_default_agent_on_gateway``) so
 # anonymous writes are never silently attributed to one shared identity.
 DEFAULT_AGENT_ID = "mcp-agent"
+
+# Bare ``"main"`` is the OpenClaw plugin's *unset* default agent_id: when an
+# operator never sets ``MEMCLAW_AGENT_ID`` every install collapses onto this one
+# shared identity (the eToro "firehose"). Unlike ``DEFAULT_AGENT_ID``
+# ("mcp-agent", legitimate on the standalone path), bare ``"main"`` is never a
+# valid *persisted* write identity and is refused on every write path.
+RESERVED_MAIN_AGENT_ID = "main"
+
+# Ids rejected on EVERY write path regardless of context (never legitimate,
+# including standalone), enforced by the write-path guard
+# (``services.agent_identity``). ``DEFAULT_AGENT_ID`` ("mcp-agent") is
+# intentionally excluded — standalone uses it and it is guarded *conditionally*
+# by ``mcp_server._refuse_default_agent_on_gateway`` instead.
+ALWAYS_RESERVED_AGENT_IDS = frozenset({RESERVED_MAIN_AGENT_ID})
--- a/core-api/src/core_api/config.py
+++ b/core-api/src/core_api/config.py
@@ -49,6 +49,18 @@ class Settings(BaseSettings):
     # ``inline_embedding`` / ``inline_enrichment`` helpers; Phase 3
     # (this revision) deleted the legacy flags + derivation validator.
     deployment_mode: Literal["inline", "deferred"] = "inline"
+    # Reserved-agent-id write guard (`main` identity fix). Bare `agent_id="main"`
+    # is the plugin's unset default; many installs collapse onto it. Phase 1:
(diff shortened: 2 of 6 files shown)


ran-taig:
@claude

ran-taig:
@claude — addressed the review (rebased on `main`, DCO signed-off, force-pushed `9d7edb85`):

**1. Domain exception instead of `HTTPException` in the service layer** ✅
- `agent_identity.py`: added `ReservedAgentIdError`; `enforce_reserved_write_id` now raises it (the module no longer imports `fastapi`).
- `memory_service.py`: imports `ReservedAgentIdError`; both call sites (`create_memory`, `create_memories_bulk`) wrap `try/except ReservedAgentIdError → raise HTTPException(status_code=409, detail=str(exc)) from exc`. (`HTTPException` was already imported.)
- **Item (e) intentionally skipped** — it contradicts (d). Once (d) re-raises `HTTPException(409)` inside `create_memory`, the MCP tool only ever receives `HTTPException`, and its existing `except HTTPException` already surfaces `.detail`. Switching MCP to catch `ReservedAgentIdError` "instead" would never fire *and* would break the duplicate-id / other-status handling in that block. Happy to revisit if you intended a no-wrap design.

**2. Null-guard on `create_memories_bulk`** ✅ — added `if not data.agent_id: raise ValueError("agent_id must be resolved before calling create_memories_bulk")`, mirroring `create_memory`.

**3. Removed unused `RESERVED_AGENT_IDS`** ✅ — dropped the constant + its block comment; `ALWAYS_RESERVED_AGENT_IDS` remains the live set (folded the mcp-agent rationale into its comment). Confirmed no production/test references remain.

Tests updated (`enforce_reserved_write_id` now asserts `ReservedAgentIdError`); **19/19 green** locally. PTAL.

ran-taig:
@claude

ran-taig:
@erni-a

erni-a (approved):

PR #506, opened 2026-06-30
feat(observability): sample below-floor near-misses on the /search recall log

arkash20:
## What
Follow-up to #503. The `/search` recall log is **returned-only** (to bound volume on the high-traffic auto path), which drops below-floor / outside-top-k **near-misses** — the exact signal needed to distinguish *just-missed* (the good memory was rank 7 at cosine 0.27) from *never close* when validating claims like eToro's "95% never recalled".

Adds a per-tenant knob **`observability.search_recall_near_miss_sample_rate`** (0.0–1.0, default **0.0**):
- `0.0` (default) → unchanged: returned-only.
- `>0` → keep near-misses on that random fraction of `/search` events, capped at the same `_NEAR_MISS_LIMIT` (5) as `mcp_recall`.

So e.g. `0.01` keeps the just-missed distribution estimable on ~1% of search events without the full candidate-row volume. `mcp_recall` is unchanged (always full incl. near-misses).

## Why
On the dominant path you could see *what was returned* but not *what just missed*. This restores that lens cheaply, behind a sampling rate, so it's safe to leave on for a busy tenant during a diagnostic window.

## Per-memory certainty (note)
Sampled near-misses characterise the **marginal band** (cap = 5). For the per-memory *"is THIS memory ever close to any query?"* question, offline replay against the logged queries remains the rigorous tool — no bulk near-miss firehose required.

## Tests
`tests/pipeline/test_log_recall_event.py`: added `test_search_keeps_near_misses_when_sampled` (rate=1.0 → near-misses kept); existing returned-only test now exercises the default rate=0.0. ruff + format + mypy clean; 30 passing.

### DIFF
--- a/core-api/src/core_api/pipeline/steps/search/log_recall_event.py
+++ b/core-api/src/core_api/pipeline/steps/search/log_recall_event.py
@@ -10,7 +10,11 @@ Gating (cheap, evaluated before any DB work) — two independent modes:
     → FULL log: returned candidates + a few below-floor near-misses.
   * ``source == "search"`` (the plugin's automatic path) +
     ``tenant_config.search_recall_logging_enabled`` → LIGHT log: returned
-    candidates only (no near-misses), since ``/search`` is high-volume.
+    candidates only, since ``/search`` is high-volume. Below-floor
+    near-misses are kept on a sampled fraction of events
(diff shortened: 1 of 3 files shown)


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #507, opened 2026-06-30
feat(core-api,plugin): self-identify MCP writes when the verified id is reserved "main"

ran-taig:
Follow-up to #505 (reserved-`main` write guard). #505 reserved `main`, but for the firehose creds the **escape hatch / plugin default couldn't take effect** — this PR fixes that, de-collapsing them **without re-provisioning credentials**.

## The gap this closes

The MCP write path resolved identity as `agent_id = _get_agent_id() or agent_id` — the gateway-injected `X-Agent-ID` (= credential `home_agent_id`) **always won over the body**. For the ~28 firehose creds (`home_agent_id="main"`), the body was ignored, so `MEMCLAW_AGENT_ID`, an explicit `agent_id` arg, and any plugin default were all no-ops. Only re-provisioning the credential could de-collapse them.

## Change

**Server** — new `agent_ids.effective_write_agent_id(verified, body)`: the verified id wins **unless** it's a reserved placeholder (`main`/`mcp-agent`), in which case a non-placeholder **body** id is honored (the install self-identifies). A *real* verified id is never overridden, so a properly-provisioned cred can't be spoofed via the body. Applied to **`memclaw_write` only** (line 746); the ~8 read/manage resolution sites keep `_get_agent_id() or agent_id`, so **visibility scoping is unchanged**.

**Plugin** — `memclaw_write` defaults `agent_id` to `main-${getInstallId()}` when `MEMCLAW_AGENT_ID` is unset. **Write-scoped at the dispatch, NOT in shared `enrichBody`**, so read scoping is unaffected. Plus `MEMCLAW_AGENT_ID` declared in `configSchema` (install prompt) + documented in `AGENT-INSTALL.md`.

**Net:** `home_agent_id="main"` creds (existing **and** future) de-collapse with no ops re-provisioning, and #505's reject/escape-hatch becomes actionable for them.

## Security note (please review this trade-off)

This loosens MCP write-identity resolution, so it deserves a deliberate look — but it is bounded:
- It applies **only when the verified id is a reserved placeholder** (a cred with no real assigned identity). A properly-provisioned cred (`home_agent_id != main`) still has the header win — **no new spoofing of real identities**.
- It is **intra-tenant only** (`agent_id` is within-tenant scoping; tenant isolation is enforced separately).
- It is **strictly tighter than the REST write path**, which already trusts `body.agent_id` unconditionally today (`_write_memory_inner`, bind flag off by default). So this introduces no risk class that doesn't already exist on REST.
- It's **transitional** — once a cred is re-identified (`home_agent_id != main`), the fallback never fires for it.

## Tests
- `tests/test_effective_write_agent_id.py` — 11 cases (real-id-wins/anti-spoof, `main`+body→body, both-placeholder→`main` so no `mcp-agent` collapse, tenant-key passthrough).
- #505 guard tests (19) still green; `tsc` clean; plugin contract tests (34) green; ruff lint+format clean.

## Roadmap impact
Makes credential **re-identification (PR-3 / ops) optional cleanup** rather than the firehose fix; the enterprise provisioning guard becomes defense-in-depth.

Refs: `pr-plan-reserved-main-identity.md`, `roadmap-main-identity-fix.md`.

### DIFF
--- a/AGENT-INSTALL.md
+++ b/AGENT-INSTALL.md
@@ -138,6 +138,10 @@ If you're an OpenClaw agent running on a gateway, install the plugin instead:
 MEMCLAW_URL=http://localhost:8000
 MEMCLAW_KEY=YOUR_KEY_HERE      # admin key (Path 2) or shared gate key (Path 3)
 MEMCLAW_FLEET=my-fleet
+MEMCLAW_AGENT_ID=my-agent      # optional but recommended: a stable, human-readable
+                              # identity for THIS install (e.g. webclaw, vm-01). If
+                              # unset, the plugin uses a stable per-install id
+                              # (main-<install_id>) so installs don't collide.
 
 # Run the install script (API key in header, not query param)
 curl -sf -H "X-API-Key: $MEMCLAW_KEY" "$MEMCLAW_URL/api/v1/install-plugin?fleet_id=$MEMCLAW_FLEET&api_url=$MEMCLAW_URL" | bash
--- a/core-api/src/core_api/agent_ids.py
+++ b/core-api/src/core_api/agent_ids.py
@@ -25,3 +25,39 @@ RESERVED_MAIN_AGENT_ID = "main"
 # intentionally excluded — standalone uses it and it is guarded *conditionally*
 # by ``mcp_server._refuse_default_agent_on_gateway`` instead.
 ALWAYS_RESERVED_AGENT_IDS = frozenset({RESERVED_MAIN_AGENT_ID})
+
+# Ids that must NOT be accepted as a *body-supplied* (client self-declared)
+# write identity: the unset plugin default ("main") and the MCP tool-param
+# default ("mcp-agent", which means "caller specified nothing"). Gates the BODY
+# only — see ``effective_write_agent_id``. The VERIFIED id is checked against
+# ``ALWAYS_RESERVED_AGENT_IDS`` ({main}) so a real (if poorly-named)
+# ``home_agent_id="mcp-agent"`` credential still wins and cannot be overridden
+# by the body (closing the spoof vector that treating verified "mcp-agent" as a
+# placeholder would open).
+_PLACEHOLDER_BODY_AGENT_IDS = frozenset({RESERVED_MAIN_AGENT_ID, DEFAULT_AGENT_ID})
+
+
+def effective_write_agent_id(verified_id: str | None, body_id: str | None) -> str | None:
+    """Resolve the agent_id a WRITE is attributed to from the verified
+    credential identity (gateway ``X-Agent-ID`` = ``home_agent_id``) and the
+    client-supplied body id.
+
+    Rule: the verified id wins — **unless** it is the reserved ``main``
+    placeholder (a misconfigured ``home_agent_id="main"`` cred), in which case a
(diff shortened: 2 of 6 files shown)


ran-taig:
@claude — review please. The load-bearing part is the **security trade-off**: this lets a cred whose verified identity is the reserved placeholder (`main`/`mcp-agent`) self-identify from the request body, instead of the gateway header always winning.

Key points for the review:
- A **real** (non-placeholder) verified id still always wins → no new spoofing of properly-provisioned creds.
- Intra-tenant only; tenant isolation unchanged.
- **Strictly tighter than REST**, which already trusts `body.agent_id` unconditionally today — so no new risk class vs. the existing REST write path.
- Applied to `memclaw_write` only; the read/manage resolution sites are untouched (visibility scoping unchanged).
- Plugin default is **write-scoped** (at the `memclaw_write` dispatch, not shared `enrichBody`), so read scoping is unaffected.

Would value a second opinion on whether the placeholder-only body fallback is the right call vs. requiring credential re-identification. Tests: `test_effective_write_agent_id.py` (11), #505 guard (19), plugin contract (34), tsc + ruff green.

ran-taig:
@claude — both review points addressed (amended `b6261856`, force-pushed):

**1. Docs contradiction — fixed as requested.** `AGENT-INSTALL.md` + `openclaw.plugin.json` now read "optional but recommended human-readable identity; if unset, the plugin uses a stable per-install id (`main-<install_id>`) so installs don't collide." The "collapses onto a shared main" clause is gone from both.

**2. Security (verified `mcp-agent` spoof) — concern closed, via an asymmetric check rather than the literal single-set.** You're right that the bug is treating a *verified* `mcp-agent` as a placeholder (a `home_agent_id="mcp-agent"` gateway cred could then body-override → spoof). That's now closed: a verified `mcp-agent` **wins and can't be overridden** (test: `("mcp-agent","evilclaw") → "mcp-agent"`).

I did **not** drop `mcp-agent` from *both* checks, though, because `_PLACEHOLDER = {main}`-for-both reopens a different hole: `verified="main"` + `body="mcp-agent"` → `"mcp-agent"`, and **nothing rejects `mcp-agent`** here (the reject guard is `{main}`-only; `_refuse_default_agent_on_gateway` short-circuits since `_get_agent_id()` is `"main"`, not `None`). Since `mcp-agent` is the tool-param default ("caller specified nothing"), that would persist memories under the sentinel — a re-collapse the guard can't catch.

So the check is asymmetric:
- **verified** id → checked against `ALWAYS_RESERVED_AGENT_IDS` = `{main}` ⇒ a real verified id (incl. `mcp-agent`) always wins; no body spoof. ← your fix
- **body** id → checked against `_PLACEHOLDER_BODY_AGENT_IDS` = `{main, mcp-agent}` ⇒ a defaulted body can't become the identity. ← prevents the re-collapse

Net: spoof vector closed **and** no `mcp-agent` re-collapse. Tests: 11 resolver cases (incl. the two new `mcp-agent`-wins), #505 guard (19) green, ruff clean. Shout if you'd prefer the literal single-set with the caveat documented instead.

ran-taig:
@erni-a

erni-a (approved):

PR #515, opened 2026-07-01
perf(recall): cap summary max_tokens and fail fast on slow provider

Eldad-Caura:
## What

Two latency levers on the `POST /api/v1/recall` hot path, from the investigation into the prod recall tail (and its link to the Cloud Run 503 alert).

1. **Cap `MEMORY_RECALL_SUMMARY_MAX_TOKENS` 1000 → 500** (`core-api/constants.py`). The recall summary is generated non-streaming, so the full generation time is exposed to the caller and long completions dominate the tail. 500 halves the worst-case output-bound generation.
2. **Recall calls `call_with_fallback(timeout=15.0, max_attempts=1)`** (`recall_service.py`). Adds a backward-compatible `max_attempts` kwarg to `call_with_fallback` (default `LLM_RETRY_ATTEMPTS`, threaded to both the primary and fallback `call_with_retry` — no change for any other caller).

## Why

Prod `POST /api/v1/recall` p95+ was routinely **5–15s** — all HTTP 200, just slow. Traced to the LLM summarization step (`summarize_memories` → `llm.complete_text`, non-streaming); search/embed is not the bottleneck (`DEFAULT_SEARCH_TOP_K=5`, small input).

The retry path is also a **503 amplifier**: `call_with_fallback` ran `LLM_RETRY_ATTEMPTS=2` per-attempt-timeout retries across **both** primary and fallback at a **30s** per-attempt timeout — worst case ~122s, capped only by the 45s request-timeout middleware / 60s Cloud Run, at which point it becomes one of the *"malformed response / connection to the instance had an error"* 503s. Capping the timeout **and** dropping to a single attempt per provider bounds recall at ~30s worst case (15s primary → 15s fallback → `_fake_recall`) and fails fast instead of hanging.

## Trade-offs

- **max_tokens=500, not lower:** `RECALL_PROMPT` still emits step-by-step chain-of-thought *before* the answer, so a tighter cap risks truncating the answer. If we later trim the CoT from the prompt (the separate prompt lever), this can drop toward ~400.
- **max_attempts=1 for recall only:** removes *same-provider* retries on this read path. Cross-provider resilience (primary → fallback → fake) is unchanged; a same-provider retry after a timeout rarely succeeds and doubles user-facing latency. Write-path enrichment/extraction keep the default 2 attempts.

## Tests

`tests/test_p4_1_llm_fallback.py`: added `TestCallWithFallbackMaxAttempts` — `max_attempts=1` invokes the primary once (no retry) then falls through; default still retries `LLM_RETRY_ATTEMPTS` times. Full fallback suite (40) + recall suites (43) green; ruff clean.

## Follow-up / verification

Now that DD APM + LLMObs are live in prod, the `POST /api/v1/recall` APM span waterfall + `ml_obs.span.llm.duration` for the recall model will show the before/after split directly (these are the traces the `--no-cpu-throttling` fix stopped dropping).

### DIFF
--- a/common/llm/retry.py
+++ b/common/llm/retry.py
@@ -88,6 +88,7 @@ async def call_with_fallback(
     model_override: str | None = None,
     model_attr: str = "enrichment_model",
     timeout: float | None = None,
+    max_attempts: int = LLM_RETRY_ATTEMPTS,
     provider_factory: Callable[..., Any] | None = None,
 ) -> T:
     """3-tier fallback chain for LLM calls.
@@ -117,6 +118,11 @@ async def call_with_fallback(
         Attribute name forwarded to the provider factory for resolving the
         default Vertex model. Defaults to ``"enrichment_model"``; entity
         extraction should pass ``"entity_extraction_model"``.
+    max_attempts:
+        Per-provider attempt count, forwarded to ``call_with_retry`` for both the
+        primary and fallback provider. Defaults to ``LLM_RETRY_ATTEMPTS``. Pass
+        ``1`` on latency-sensitive read paths (e.g. recall) to fail fast to the
+        fallback provider instead of retrying a slow/hung primary.
     provider_factory:
         Callable ``(name, tenant_config) -> LLMProvider``. Defaults to
         ``common.llm.registry.get_llm_provider`` (imported lazily to avoid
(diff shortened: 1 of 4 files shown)


Eldad-Caura:
@claude please review this PR.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #518, opened 2026-07-02
fix(lifecycle): correct run_crystallization args in crystallize adapter

Eldad-Caura:
## What
One-line fix at `core-api/src/core_api/services/lifecycle_audit.py` — the lifecycle `crystallize` adapter called `run_crystallization` with a stray leading `None` and a colliding `trigger`:

```python
# before
run_crystallization(None, org_id, fleet_id, trigger="lifecycle")
# after
run_crystallization(tenant_id=org_id, fleet_id=fleet_id, trigger="lifecycle")
```

## Why
`run_crystallization(tenant_id, fleet_id=None, trigger="manual", auto_crystallize=True)`. The stray `None` shifted every positional — `tenant_id=None`, `fleet_id=org_id`, `trigger=fleet_id` — and then `trigger="lifecycle"` was *also* passed by keyword, so **every** lifecycle-triggered crystallization raised:

```
TypeError: run_crystallization() got multiple values for argument 'trigger'
```

The pubsub lifecycle handler (`common/events/lifecycle_handlers.py` → `crystallize_op` → `adapter.crystallize`) then nacked + redelivered — **~20 occurrences/day on prod core-api**, and lifecycle-driven crystallization never actually ran.

`org_id` is the tenant scope at this boundary — the adapter's own docstring says *"the storage client's primitives still call the column `tenant_id`; translate at the boundary"*, and `resolve_config(org_id)` here is the same call the HTTP routes make as `resolve_config(tenant_id)` before `run_crystallization(tenant_id, …)`. So `tenant_id=org_id`. Keyword args are used so this positional class of bug can't recur.

## Tests
Adds `tests/test_lifecycle_audit.py` (3 tests): the happy path asserts the exact call `run_crystallization(tenant_id="t1", fleet_id="f1", trigger="lifecycle")` via an **`autospec`'d** mock (which enforces the real signature — the old buggy call would raise `TypeError` in the test), plus the two skip gates (auto-crystallize disabled, below active-memory threshold). The existing `test_lifecycle_handlers.py` fully fakes the adapter, which is why this call site was uncovered. Full run green; ruff clean.

## Note
Auto-review will skip on open (public OSS repo, non-public-member author) — `@claude` if you want a review pass.

### DIFF
--- a/core-api/src/core_api/services/lifecycle_audit.py
+++ b/core-api/src/core_api/services/lifecycle_audit.py
@@ -92,7 +92,7 @@ class _CoreApiLifecycleAdapter:
         # core-api startup just for the lifecycle adapter wiring.
         from core_api.services.crystallizer_service import run_crystallization
 
-        report_id = await run_crystallization(None, org_id, fleet_id, trigger="lifecycle")
+        report_id = await run_crystallization(tenant_id=org_id, fleet_id=fleet_id, trigger="lifecycle")
         return 1 if report_id is not None else 0
 
     async def insights(self, *, org_id: str, fleet_id: str | None) -> int:
--- a/tests/test_lifecycle_audit.py
+++ b/tests/test_lifecycle_audit.py
@@ -1,0 +1,90 @@
+"""Regression tests for _CoreApiLifecycleAdapter.crystallize (CAURA-657).
+
+The lifecycle crystallize adapter previously called run_crystallization as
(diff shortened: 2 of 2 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

ran-taig (approved):

PR #522, opened 2026-07-04
fix(logging): floor ddtrace trace/LLMObs writer loggers at CRITICAL

Eldad-Caura:
## What
Floor the `ddtrace.internal.writer.writer` and `ddtrace.llmobs._writer` loggers at CRITICAL in the shared `configure_logging` (when the root level is INFO/WARNING; skipped at DEBUG). `_reset_for_testing` clears the floor back to NOTSET.

## Why
Over the last 36h these two loggers produced **~95% of all `status:error` log volume** across prod+staging — every failed flush to the in-container agent at `localhost:8126` logs at ERROR:

```
failed to send, dropping N traces to intake at http://localhost:8126/v0.5/traces: timed out
failed to send N LLMObs span events to http://localhost:8126 … TimeoutError / 502 / 429 / ConnectionReset
```

These are **dropped-telemetry events** (`send_to_telemetry:false`), not application errors — the app is unaffected. But because they emit at ERROR they surface as errors in Datadog Logs, GCP Logging, and the **caura-ops error alerter**, where they drive alert flapping in both envs. CPU-throttling fixes (ent#701/#717) reduce the CPU-starvation timeouts but the agent-side variants (502/429/RemoteDisconnected) persist, so silencing at the source is the durable fix.

Flooring at CRITICAL drops the ERROR/WARNING noise while still surfacing a genuine writer CRITICAL. Lives in the shared structlog config, so it applies to every service (core-api, core-worker, core-operations, platform-*) at once.

## Test
Added `tests/test_logging.py` coverage: floored at CRITICAL at INFO **and** WARNING, not floored at DEBUG, and reset restores NOTSET. `uv run pytest tests/test_logging.py` green; ruff clean.

### DIFF
--- a/common/structlog_config.py
+++ b/common/structlog_config.py
@@ -494,6 +494,11 @@ def _reset_for_testing() -> None:
         # silent divergence from production behavior.
         for dep in ("httpx", "httpcore", "google.auth", "google.auth.transport"):
             logging.getLogger(dep).setLevel(logging.NOTSET)
+        # Clear the ddtrace-writer CRITICAL floor installed by
+        # _configure_logging_impl so a test reconfiguring at DEBUG doesn't see
+        # these stuck at CRITICAL — same rationale as the overrides above.
+        for dep in ("ddtrace.internal.writer.writer", "ddtrace.llmobs._writer"):
+            logging.getLogger(dep).setLevel(logging.NOTSET)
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@claude please review

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #523, opened 2026-07-04
fix(core-operations): configure logging at import so uvicorn startup lines aren't tagged ERROR

Eldad-Caura:
## What
Move `configure_logging()` from the lifespan startup hook to **module import** in `core-operations/app.py`, and call `reroute_third_party_loggers()` in the lifespan — mirroring core-api.

## Why
The 36h error sweep flagged `core-operations` (and `platform-operations`) emitting `status:error` logs that are actually uvicorn **INFO** startup lines:

```
INFO:     Started server process [8]
INFO:     Waiting for application startup.
```

uvicorn logs these via `uvicorn.error` **before** the ASGI lifespan runs. core-operations configured logging *inside* the lifespan, so at that moment the logger reroute wasn't active yet — the two lines fell through uvicorn's default plain-text stderr handler, and Cloud Logging tagged them `ERROR`. They're benign INFO (and pre-caura-ops#192 could even page).

core-api doesn't have this because it calls `configure_logging()` at import (uvicorn imports the app module during config load, so the reroute is in effect before the startup lines) and only `reroute_third_party_loggers()` in the lifespan. This change brings core-operations in line.

## Test
Full `core-operations` suite green (**34 passed**), including `test_app.py` (imports the module) and `test_scheduler.py` (uses caplog) — the deferral was originally to protect pytest caplog, and this proves import-time config doesn't break it (the repo's per-test reset handles it, same as core-api). ruff clean.

## Note
`platform-operations` (enterprise repo) has the same lifespan-deferred pattern and needs the identical one-line move — happy to follow up with that PR.

### DIFF
--- a/core-operations/src/core_operations/app.py
+++ b/core-operations/src/core_operations/app.py
@@ -4,8 +4,12 @@ Hosts cron/scheduled background jobs that operate on OSS data
 (memories, organizations, tenants). No business HTTP surface — only
 ``/healthz`` for Cloud Run probes.
 
+Logging is configured at MODULE IMPORT (see below) so uvicorn's own startup
+lines route through the JSON/GCP handler; the lifespan only re-routes
+third-party loggers once they're all imported.
+
 Lifespan ordering:
-1. ``configure_logging`` reads env BEFORE any module that emits log records.
+1. Re-route third-party loggers (uvicorn / scheduler) onto the JSON handler.
 2. If ``settings.standalone``: skip scheduler entirely. The service runs
    as a no-op; OSS standalone deployments should not deploy this image
    at all, but the flag is a defensive short-circuit.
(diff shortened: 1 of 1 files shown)


Eldad-Caura:
@claude please review

Eldad-Caura:
@cladue

Eldad-Caura:
Verified the caplog concern before acting — it's a **latent** fragility, not a live bug, so I addressed it with the clarifying comment (your stated fallback) rather than adding an autouse reset fixture:

- The scheduler/task tests make **real caplog assertions** (`test_scheduler.py:110` and `:305` assert on `caplog.records`) and **pass** (full suite 34 green). They log via stdlib loggers that reach caplog through root propagation, independent of structlog's import-time config.
- **No conftest anywhere in the repo calls `_reset_for_testing`** — including core-api, which runs this exact import-time `configure_logging()` pattern and passes. So the fixture the finding suggests isn't the mechanism that makes import-time config safe; it isn't needed here, and adding it only to core-operations would be a one-off the mirrored reference (core-api) doesn't have.

Added a comment on the `configure_logging()` call documenting the test-safety + the core-api parallel (commit `7a8408d`). No behavior change; ruff clean, suite still green.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #529, opened 2026-07-06
fix(scoring): cap freshness at 1.0 for future ts_valid_start (A43)

arkash20:
## What
Cap the freshness factor at 1.0 in `memory_scored_search` so a **future `ts_valid_start`** (with `ts_valid_end` NULL) can't inflate freshness above 1.0.

## Why
Freshness anchors on `greatest(created_at, coalesce(ts_valid_start, created_at))`; the age branch `1 - (age/type_decay)*(1-floor)` had no lower bound on age. A future anchor drives `age_days` negative → freshness climbs to several× → final `score` exceeds 1.0 → a low-similarity memory dominates ranking for any query.

**Evidence (eToro `etoro0`):** 28 memories had a future `ts_valid_start` + NULL `ts_valid_end`. The AI-Act memory (`ts_valid_start = 2027-08-02`, extracted from its own text) reached `recall_count` 3,773 and ranked **#1 for 159 unrelated queries** at cosine ~0.33 — this was the true cause of that memory's day-0 explosion.

## Fix
Clamp age to ≥ 0 via `greatest(0.0, age_days)` — a memory can never be "fresher than now"; freshness stays in `[floor, 1.0]`. One-line change in the scoring CASE.

## Test
Adds `test_scored_search_future_valid_start_does_not_inflate_freshness`: a future-dated memory must not outscore an otherwise-identical present-dated one (same embedding/weight/type).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #530, opened 2026-07-06
feat(plugin): opt-in recall-gating + cross-agent recall (F2/A44)

arkash20:
## What
Two **plugin-side**, **flag-gated (default off)** recall changes. Merging is behavior-neutral until opted in.

### F2 — recall-gating (`shouldRecall`)
Stops auto `/search` from firing on non-question turns. 5 new noise-skip reasons: machine/automation patterns, agent-name self-match, `@mention`/number-only, `[Subagent Context]` blobs, instruction-to-3rd-party.
- **`MEMCLAW_RECALL_GATE`** = `off` (default) · `shadow` (log would-skip, don't enforce) · `on`.
- **`MEMCLAW_RECALL_MACHINE_PATTERNS`** — machine regex list is per-deployment configurable (a term that's noise here may be real content elsewhere, e.g. "health check"), seeded from observed eToro automation traffic.
- Explicit recall keywords still win.

### A44 — cross-agent recall (silo removal)
- **`MEMCLAW_RECALL_CROSS_AGENT`** (default `false`): auto-recall omits `filter_agent_id` so an agent can retrieve **sibling agents' knowledge**, still bounded by the server-side fleet/trust scope. The silo was the plugin sending `filter_agent_id: agentId` on every recall.
- **Enable alongside the A43 freshness cap (#529)** — otherwise a future-dated hub memory dominates cross-agent results.

## Evidence
Held-out on `etoro-core` (rules built on `etoro0`): the noise gate skips **~76% of recall calls with 0/29 genuine questions lost**. Only 5–35% of `/search` calls are genuine questions; the rest are machine/probe/chatter.

## Tests
8 new unit tests for the noise-skip rules — **38/38 context-engine tests pass**, typecheck clean.

## Rollout
Default off → recommended path: `shadow` first (measure), then `on`; enable cross-agent only after #529 merges.

### DIFF
--- a/plugin/src/context-engine.test.ts
+++ b/plugin/src/context-engine.test.ts
@@ -45,6 +45,7 @@ function input(overrides: Partial<ShouldRecallInput> = {}): ShouldRecallInput {
     triggerKeywords: DEFAULT_KEYWORDS,
     sessionKey: "tenant:agent:default",
     denySessions: [],
+    applyNoiseRules: true,
     ...overrides,
   };
 }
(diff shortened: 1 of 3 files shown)


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #535, opened 2026-07-06
fix(reports): count agent-private durable writes in report aggregates

ran-taig:
## Problem

The report's `durable_memories_written` (and `by_type` / `per_agent` / `trend` / `quality`) excluded `scope_agent` (agent-private) rows. So the top-line metric silently measured **team-visible** durable writes, not what was actually **written**.

Visibility is an *audience* attribute — an agent keeping a decision/fact private doesn't make it less durable or less decision-bearing. The exclusion also disproportionately undercounts privacy-heavy tenants. Verified against prod:

| tenant | current report | with private counted | delta |
|---|---|---|---|
| etoro0 (noisy/private-heavy) | 98 | **161** | +63 (+64%) |
| etoro-core (lean) | 254 | **275** | +21 (+8%) |

The recovered rows are real knowledge (rule/outcome/fact), not noise — the cohesive/episode/`main` filters already ran. The correction is self-targeting: it fixes the undercount where it was wrong and barely moves the already-clean tenant.

## Change

- Add an opt-in **`include_scope_agent`** flag (default `False`, so MCP `memclaw_stats` behavior is unchanged) to `memory_stats_breakdown`, `memory_daily_durable_counts`, and `memory_quality_metrics`.
- The report sets it `True` on its count/trend/quality queries; it rides `scope_keys` too, so the funnel's `written` and the insight-freshness corpus stay consistent (`written >= durable` holds).

This is a **pure count** change — no memory content is exposed. The content-bearing sections (`value_highlights` / `learning` / `spotlight` / `working_on`) still exclude `scope_agent`, so no agent's private memory is surfaced to the group.

## Test

`test_report_group_counts_private_writes_but_hides_content`: a private-scope durable write is counted in `durable_memories_written` / `by_type` / `per_agent` / `trend`, but never appears in `value_highlights`. Full `tests/test_reports.py` green locally (12 passed); the 11 pre-existing tests are unchanged (they seed only `scope_team`).

### DIFF
--- a/core-api/src/core_api/routes/reports.py
+++ b/core-api/src/core_api/routes/reports.py
@@ -268,6 +268,16 @@ async def get_report(
         "exclude_memory_types": list(NON_DURABLE_TYPES),
         "exclude_agent_ids": list(RESERVED_FIREHOSE_AGENTS),
         "exclude_title_regex": NON_COHESIVE_TITLE_REGEX,
+        # Count agent-private (scope_agent) durable writes in the AGGREGATES
+        # (totals/by_type/by_agent/quality/trend). These are real, decision-
+        # bearing memories an agent kept private — excluding them made
+        # durable_memories_written measure "team-visible" rather than "written",
+        # and disproportionately undercounted privacy-heavy tenants. This is a
+        # pure count; the content sections (value_highlights/learning/spotlight/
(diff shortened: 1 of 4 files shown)


ran-taig:
@claude

ran-taig:
@erni-a

erni-a (approved):

Issue #510, opened 2026-07-01
search() returns empty list silently on unexpected API response shape


jy7lsna:
### MemClaw version

2.1.0

### Deployment mode

Self-hosted (Docker Compose)

### Operating system

Windows 11

### Python version

3.12.7

### PostgreSQL version

16.4

### LLM provider in use

openai

### Describe the bug

Both the Python and TypeScript clients silently return an empty list from search() when the API response does not match the expected shape no exception, no warning, no indication that anything went wrong.

Python: client.py line 97: `data.get("items", []) if isinstance(data, dict) else []`
TypeScript: index.ts line 132: `Array.isArray(items) ? items.map(...) : []`

If the API returns a non-dict body, or returns a dict without an "items" key (e.g. a 200 with an error envelope, a shape change, or a proxy response), the caller gets back [] and has no way to distinguish "no results" from "something went wrong."

### Steps to reproduce

1. Instantiate the Python client pointed at a mock server.
2. Have the mock return HTTP 200 with body `{"error": "quota exceeded"}` (no "items" key).
3. Call `mc.search("anything")`.
4. Observe: returns `[]` with no exception raised.

### Expected behavior

If the response is 200 but the body does not contain an "items" list, search() should raise MemClawAPIError (Python) / MemClawApiError (TypeScript) so the caller knows the result is not trustworthy not silently return an empty list.

### Actual behavior

Returns [] with no error. The caller cannot distinguish zero results from a malformed or unexpected response.

### Relevant logs / config

```shell
Same silent-fallback pattern exists in both clients at:
- clients/python/src/memclaw_client/client.py:97-98
- clients/typescript/src/index.ts:131-132

A fix should raise on unexpected shape, with a test added to clients/python/tests/test_client.py and clients/typescript/src/index.test.ts using the existing mock transport patterns already in those files.
```

### Pre-flight

- [x] I searched existing issues and Discussions for duplicates.
- [x] This is not a security vulnerability (those go through Security Advisories).

kimberly-WY:
Hit this same issue — the root cause was the missing null check before deserialization. Wrapping in a try/catch fixed it on our end.

wulansari999:
Noted — analyzing this issue for a potential fix. Will submit a PR if a solution is found.

wulansari999:
Noted — analyzing this issue for a potential fix. Will submit a PR if a solution is found.

wulansari999:
[...]

PR #539, opened 2026-07-08
fix(clients): raise on unexpected search() response shape (#510)

gitcommit90:
## Problem

On HTTP 200,  silently returned  when the JSON body was not a dict with an `items` list (e.g. `{ "error": "quota exceeded" }`), so callers could not distinguish zero hits from a malformed response.

Refs: caura-ai/caura-memclaw#510

## Proposed change

Validate the search response shape in both Python and TypeScript clients and raise `MemClawAPIError` / `MemClawApiError` with status 200 when the body is unexpected.

## Test results

- clients/python: 12 passed
- clients/typescript: 10 passed

### DIFF
--- a/clients/python/tests/test_client.py
+++ b/clients/python/tests/test_client.py
@@ -63,6 +63,36 @@ def test_search_returns_list():
     assert [m.id for m in results] == ["m1", "m2"]
 
 
+def test_search_raises_on_missing_items_key():
+    def handler(request):
+        return httpx.Response(200, json={"error": "quota exceeded"})
+
+    with pytest.raises(MemClawAPIError) as exc:
+        make_client(handler).search("q")
+    assert exc.value.status_code == 200
+    assert str(exc.value) == '[200] search response missing "items" list'
+
+
+def test_search_raises_on_items_not_a_list():
+    def handler(request):
+        return httpx.Response(200, json={"items": "not-a-list"})
+
+    with pytest.raises(MemClawAPIError) as exc:
+        make_client(handler).search("q")
+    assert exc.value.status_code == 200
+    assert str(exc.value) == '[200] search response "items" must be a list'
+
+
+def test_search_raises_on_non_dict_body():
+    def handler(request):
+        return httpx.Response(200, json=["not", "a", "dict"])
+
+    with pytest.raises(MemClawAPIError) as exc:
+        make_client(handler).search("q")
+    assert exc.value.status_code == 200
+    assert str(exc.value) == "[200] search response must be a JSON object"
+
+
 def test_recall_returns_summary():
     def handler(request):
         assert request.url.path == "/api/v1/recall"
--- a/clients/typescript/src/index.ts
+++ b/clients/typescript/src/index.ts
@@ -128,8 +128,17 @@ export class MemClaw {
     if (filterAgentId) body.filter_agent_id = filterAgentId;
     Object.assign(body, extra);
     const data = await this.request("POST", "/api/v1/search", body);
-    const items: unknown = data?.items;
-    return Array.isArray(items) ? items.map((m) => toMemory(m as Record<string, any>)) : [];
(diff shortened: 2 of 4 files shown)


clients/typescript/src/index.ts
```diff
@@ -128,8 +128,17 @@ export class MemClaw {
...
-    const items: unknown = data?.items;
-    return Array.isArray(items) ? items.map((m) => toMemory(m as Record<string, any>)) : [];
+    if (!data || typeof data !== "object" || Array.isArray(data)) {
+      throw new MemClawApiError(200, "search response must be a JSON object");
+    }
+    if (!("items" in data)) {
+      throw new MemClawApiError(200, 'search response missing "items" list');
+    }
+    const items = (data as Record<string, unknown>).items;
+    if (!Array.isArray(items)) {
```
aaron-seq, line 138: The `!Array.isArray(items)` guard is not covered by any test in `index.test.ts`. The single added test (`"search throws when 200 body lacks items"`) only exercises the non-dict/missing-items path. A second test case with a response such as `{ items: "not-a-list" }` is needed to verify this branch raises `MemClawApiError` with status 200.

clients/python/tests/test_client.py
```diff
@@ -63,6 +63,25 @@ def handler(request):
...
 
 
+def test_search_raises_on_missing_items_key():
+    def handler(request):
+        return httpx.Response(200, json={"error": "quota exceeded"})
+
+    with pytest.raises(MemClawAPIError) as exc:
+        make_client(handler).search("q")
+    assert exc.value.status_code == 200
+    assert "items" in str(exc.value)
```
aaron-seq: `assert "items" in str(exc.value)` is a fragile assertion: it checks whether the substring `"items"` appears anywhere in the string representation of the exception, which would pass even if the message changed to something unrelated but happened to contain that substring. Consider asserting the exact message instead, e.g. `assert exc.value.args[0] == 'search response missing "items" list'`, to make the test fail loudly if the error message is ever refactored.

clients/python/tests/test_client.py
```diff
@@ -63,6 +63,25 @@ def handler(request):
...
+
+
+def test_search_raises_on_non_dict_body():
+    def handler(request):
+        return httpx.Response(200, json=["not", "a", "dict"])
+
+    with pytest.raises(MemClawAPIError) as exc:
+        make_client(handler).search("q")
+    assert exc.value.status_code == 200
```
aaron-seq, line 92: Neither Python test covers the case where `items` is present but is not a list (the `if not isinstance(items, list)` branch in `client.py` line 102). A third test is needed, e.g. a handler returning `httpx.Response(200, json={"items": "not-a-list"})`, asserting that `MemClawAPIError` is raised with status 200. Without it, a future regression on that guard would not be caught by the test suite.

aaron-seq (changes requested):
The approach here is correct: replacing the silent fallback `.get("items", [])` with explicit guards that raise `MemClawAPIError`/`MemClawApiError` is the right fix for issue #510, and the error messages are clear and consistent across both clients. Tests are included, which is appreciated.

Blocking:
1. No `Signed-off-by` trailer in the commit (`ddf6fe0`). The DCO check is required for this repo. Please amend the commit to add `Signed-off-by: <name> <email>` before merging.
2. Missing test coverage for the `items-not-a-list` guard in both Python and TypeScript clients. The guard exists in the code but neither test suite exercises it. Details in inline comments.

Non-blocking suggestions are noted inline.

I reviewed suggestions generated with AI assistance and manually verified each comment against the diff and the repository's contributing guidelines before posting.

ran-taig (approved):

Eldad-Caura:
@claude

Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #540, opened 2026-07-08
feat(insights): give the nightly insights run a dedicated registered identity

ran-taig:
## Problem

The scheduled nightly insights pass had **no caller identity**, so it fell through to the generic `DEFAULT_AGENT_ID = "mcp-agent"` — an *unregistered*, anonymous id that also catches any caller who specifies nothing. Consequences:

- Its writes never surfaced in **Prism** or the **per-agent report** — there's no matching `agents` row to join against.
- It was conflated with unrelated anonymous traffic under one shared `mcp-agent` bucket.

(On prod, `mcp-agent` was the top "agent" on etoro0/etoro-core purely because of this — 100% of those writes were `client_request_id='insights:%'`.)

## Change

Give the automated run its own registerable identity, `memclaw-insighter`:

| File | Change |
|---|---|
| `agent_ids.py` | Add `INSIGHTER_AGENT_ID = "memclaw-insighter"` + `INSIGHTER_TRUST_LEVEL = 3`, distinct from the generic `mcp-agent` fallback |
| `services/lifecycle_audit.py` | `insights()` attributes `generate_insights(agent_id=INSIGHTER_AGENT_ID)` and **self-registers** the agent (idempotent `create_or_update_agent`, `ON CONFLICT DO NOTHING`) as `belonging_type='service'`, trust 3, before persisting |
| `migrations/030` | Backfill the same registration for tenants that already hold `memclaw-insighter` memories, so Prism/reports show it immediately |

**Registration timing** = lazy, on first run: the first nightly insights run for any tenant (new or existing) registers the agent; a tenant with `auto_insights_enabled` off registers nothing (no phantom agent). No cross-repo tenant-provisioning hook needed.

**Privileges:** `service`/trust 3 grants cross-fleet read (≥2, needed for `scope='all'`) and write (≥3, it persists `scope_org`). The automated path calls the service directly and bypasses the route trust gate, so this is future-proofing for any gated re-route.

**Scope guard:** interactive MCP/REST callers are unaffected — they pass their real `agent_id` via `resolve_caller_and_gate`. Only the identity-less automated path changes. The generic `mcp-agent` fallback is intentionally left in place.

## Testing

- `test_lifecycle_audit.py`: automated run attributes to `memclaw-insighter` + registers the service agent (trust 3); opt-out short-circuits with no registration. 5/5 + 14 handler tests pass locally.
- Migration 030 validated against the live ORM schema: upgrade registers only tenants with insighter memories, re-run is a no-op, downgrade removes them.

> Note: this is the persistence follow-up to the manual prod DB rename of `mcp-agent → memclaw-insighter` on etoro0/etoro-core.

### DIFF
--- a/core-api/src/core_api/agent_ids.py
+++ b/core-api/src/core_api/agent_ids.py
@@ -12,6 +12,22 @@ otherwise couple the route module to the whole MCP tool surface.
 # anonymous writes are never silently attributed to one shared identity.
 DEFAULT_AGENT_ID = "mcp-agent"
 
+# Dedicated system identity for the automated nightly insights run
+# (``lifecycle_audit._CoreApiLifecycleAdapter.insights`` → ``generate_insights``).
+# Distinct from ``DEFAULT_AGENT_ID`` above: that is the generic "caller specified
+# nothing" fallback, whereas the scheduled insights pass has no human/agent
+# caller and must write under a stable, registerable identity rather than
+# collapsing onto the anonymous default (which is unregistered, so its writes
+# never surface in Prism or the per-agent report). Registered per-tenant with
+# ``belonging_type='service'`` the first time the job runs for that tenant.
+INSIGHTER_AGENT_ID = "memclaw-insighter"
+
+# Trust tier the insighter self-registers at: service level, granting cross-fleet
+# read (>=2, needed for scope='all') and write (>=3, it persists scope_org
+# insights). The automated path calls the service directly and bypasses the
+# route trust gate, so this is future-proofing for any gated re-route.
+INSIGHTER_TRUST_LEVEL = 3
+
 # Bare ``"main"`` is the OpenClaw plugin's *unset* default agent_id: when an
 # operator never sets ``MEMCLAW_AGENT_ID`` every install collapses onto this one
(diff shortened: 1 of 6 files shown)


ran-taig:
@claude

ran-taig:
@claude re-review please.

Addressed review finding: the nightly registration was a bare `create_or_update_agent` upsert, but `agent_add`'s conflict path *updates* `trust_level`/`display_name` when they differ — so every run would silently overwrite any operator customisation of the insighter agent.

Fix (folded into the feature commit): guard the registration with an existence check so it's true one-time setup —

```python
existing = await self._storage.get_agent(INSIGHTER_AGENT_ID, org_id)
if existing is None:
    await self._storage.create_or_update_agent({...})
```

Also corrected the inline comment (dropped the inaccurate "idempotent — ON CONFLICT DO NOTHING" claim) and added a regression test asserting an existing agent is **not** re-registered. ruff/format/mypy clean; lifecycle tests 6/6.

ran-taig:
@claude

ran-taig:
@claude

ran-taig:
@erni-a

erni-a (approved):

PR #543, opened 2026-07-09
feat(reports): weekly agent-digest cron tick [CAURA-222]

ran-taig:
## Weekly agent-digest generation

Phase 2c registered a nightly tick that only generates `period=day`, so the reports UI's **"week" toggle shows nothing**. This adds a weekly wall-clock-aligned tick that fires `period=week`.

- **scheduler**: `seconds_until_next_utc_weekday_hour(weekday, hour)` — Mon..Sun alignment with the same strict-future / no-hot-loop guarantee as the existing daily helper.
- **config**: `AGENT_DIGEST_WEEKLY_RUN_AT_WEEKDAY` (0=Mon, default) + `AGENT_DIGEST_WEEKLY_RUN_AT_HOUR` (default 03:00 UTC, just after Monday's daily tick so the just-closed Mon–Mon window is covered), both validated.
- **tasks**: `run_agent_digest_weekly_tick` → `POST /admin/reports/agent-digest/run?period=week`.
- **app**: registers `agent-digest-weekly` (7-day interval, weekly `delay_provider`).

### Also fixes a stale test
`test_app._EXPECTED_JOBS` never included the daily `agent-digest` job added in Phase 2c (core-operations tests aren't in the CI gate, so it slipped through). This adds both jobs and gives the weekly job a 7-day delay bound in the alignment assertion.

### Testing
Full core-operations suite green (41 tests), incl. 5 new scheduler cases for the weekday helper + the weekly-tick test. ruff + format + mypy clean.

### DIFF
--- a/core-operations/src/core_operations/app.py
+++ b/core-operations/src/core_operations/app.py
@@ -49,9 +49,14 @@ configure_logging(
     log_file=settings.log_file or None,
 )
 
-from core_operations.scheduler import scheduler, seconds_until_next_utc_hour
+from core_operations.scheduler import (
+    scheduler,
+    seconds_until_next_utc_hour,
+    seconds_until_next_utc_weekday_hour,
+)
 from core_operations.tasks import (
     run_agent_digest_tick,
+    run_agent_digest_weekly_tick,
     run_archive_expired_tick,
     run_archive_stale_tick,
     run_crystallize_tick,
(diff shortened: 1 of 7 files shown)


ran-taig:
@claude

ran-taig:
@erni-a

erni-a (approved):

PR #545, opened 2026-07-12
fix(reports): salvage empty-narrative digests + period-scaled activity floor

ran-taig:
## Why
Prod review of the merged daily agent-activity digest (both eToro fleets) showed near-empty output. The latest daily run produced **3 rows across two 300+-agent tenants**, and **2 of 3 were dead `error` rows** (`memclaw-insighter`, `"LLM returned no narrative"`) that render blank in the UI.

Grounded findings (read-only prod queries):
- **Sparse daily activity is the real bottleneck**, not the corpus: only ~3 distinct agents/tenant write durable memories per *day* (vs 8–13 per *week*). `episode` (32k/7d — the raw firehose) is the only excluded type and correctly so; the shared noise-title filter drops genuine heartbeat/health noise (34% of `action`, 61% of `outcome`). **Corpus left untouched** — loosening it mostly re-admits noise and is shared with the live report.
- `min_activity_threshold=3` over a 1-day window then trims the ~3 daily agents down to ~1.
- Empty-narrative responses became blank `error` rows (worse than a placeholder).

## What
1. **Salvage empty narratives** → `fallback` (not `error`). If a real response omits the narrative, synthesize a terse one from the structured `sections` (`"Agent: 2 decisions, 1 shipped item across N durable memories."`); `error_detail` records why. Actual exceptions still → `error`.
2. **Period-scaled activity floor**: daily floors to `1` (surface every active agent); weekly keeps the configurable `min_activity_threshold`.

Not included (separate, by design): making **weekly the primary UI surface** (data is 4–5× richer weekly) and verifying the weekly cron is live on-prem for Monday — tracked separately.

## Tests
`test_daily_threshold_is_one` (day floor 1 vs week floor 3), rewrote the empty-narrative test to assert `fallback` + synthesized narrative, updated the threshold test to `week`. ruff + format + mypy clean (186 files); 16 generator unit tests green.

### DIFF
--- a/core-api/src/core_api/services/agent_digest.py
+++ b/core-api/src/core_api/services/agent_digest.py
@@ -57,6 +57,26 @@ DIGEST_SCHEMA: dict = {
 }
 
 _SECTION_KEYS = ("decisions", "shipped", "learned", "open_threads")
+_SECTION_LABELS = (
+    ("decisions", "decision"),
+    ("shipped", "shipped item"),
+    ("learned", "learning"),
+    ("open_threads", "open thread"),
+)
+
+
+def _narrative_from_sections(agent: dict, mems: list[dict], sections: dict) -> str:
+    """Terse prose salvaged from the structured sections when the model returned
+    them but omitted the narrative — a useful placeholder beats a blank row."""
+    name = agent.get("display_name") or agent.get("agent_id")
+    parts = []
+    for key, label in _SECTION_LABELS:
+        n = len(sections.get(key) or [])
+        if n:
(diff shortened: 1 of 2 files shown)


ran-taig:
@claude please review — two grounded fixes from a prod review of the daily digest (details in the PR body): empty-narrative responses are now salvaged from sections as `fallback` instead of blank `error` rows, and the activity floor is period-scaled (daily=1, weekly=config) since only ~3 agents/day clear a threshold of 3. Corpus deliberately untouched. 16 unit tests green.

ran-taig:
@erni-a

erni-a (approved):

PR #549, opened 2026-07-13
fix(recall): don't inflate recall_count on agentless recalls (A26)

arkash20:
## Problem

`recall_count` is a per-memory *"an agent found this useful"* signal, and it feeds `recall_boost` in scoring. `TrackRecalls` bumps it for **every** `/search` return — including **agentless** traffic: liveness/health probes, monitoring pollers, and dashboard/admin queries that hit `/search` with no agent identity.

None of those are an agent using memory, yet they:
- inflate `recall_count` (and the `recall_boost` that rides on it) with non-agent noise, and
- let a repeating automated caller **pin a single memory's counter** — e.g. a health check calling `/search` thousands of times a day at `top_k=1` hammers one memory's `recall_count`, which then gets boosted for everyone.

## Fix

Skip the `recall_count` bump when there's no caller agent identity:

```python
if not ctx.data.get("caller_agent_id"):
    return None
```

`caller_agent_id` is the authenticated identity (`filter_agent_id` or `auth.agent_id`), so **genuine agent recalls still count** — including cross-agent recalls that omit `filter_agent_id` and fall back to `auth.agent_id`. Search **results are returned unchanged**; only the counter increment is skipped.

## Why it generalizes

This isn't specific to one deployment — any environment with automated `/search` callers (health checks, monitoring, pipelines, dashboards) currently skews the usefulness signal and the `recall_boost` derived from it. Restricting the counter to identified-agent recalls makes `recall_count` mean what it's supposed to mean everywhere. (Addresses gap **A26** — self-reinforcing `recall_boost` with no usefulness signal.)

## Tests

`tests/test_track_recalls_agentless.py` — bump fires for a real agent, is skipped for `None`/empty caller identity, and is a no-op when there are no rows. Guard logic also verified standalone against the real `TrackRecalls.execute`.

### DIFF
--- a/core-api/src/core_api/pipeline/steps/search/track_recalls.py
+++ b/core-api/src/core_api/pipeline/steps/search/track_recalls.py
@@ -42,6 +42,19 @@ class TrackRecalls:
         return "track_recalls"
 
     async def execute(self, ctx: PipelineContext) -> StepResult | None:
+        # ``recall_count`` is a per-memory "an agent found this useful" signal,
+        # and it feeds ``recall_boost`` in scoring. Only bump it for recalls
+        # that carry a caller agent identity. Agentless ``/search`` traffic —
+        # liveness/health probes, monitoring pollers, dashboard/admin queries
+        # that hit the endpoint with no agent — is not an agent using memory;
+        # counting it inflates ``recall_count`` (and thus ``recall_boost``) with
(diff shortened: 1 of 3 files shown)


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #556, opened 2026-07-15
feat(memories): attribute broker bulk writes to the originating agent

Eldad-Caura:
## What

A broker (install-credential) bulk write carries no request-level `agent_id`, so the cloud attributed every such write to a synthetic `broker:<install_uuid>` identity — the memory view (Prism) shows `broker-<id>`, with no way to tell **which agent** produced a memory (feedback from the broker walkthrough).

## How

memclawd **Layer 1** (already merged, `caura-ai/memclawd#82`) now stamps each capture's `metadata.agent_id` with the capturing agent. **Layer 2** (this PR): when a broker batch's per-item metadata **unanimously** names one agent, attribute the write to that agent — so the memory view names it. A **mixed batch**, or a **pre-Layer-1 broker** (v0.8.3, no per-item `agent_id`), falls back to `broker:<install_uuid>` — the exact prior behavior, so a write is **never mis-attributed**.

New pure helper `_broker_write_agent_id(items, install_uuid)` in `routes/memories.py`, wired into the install-credential branch of the bulk-write handler.

## Scope

Attribution stays **request-level** (one `agent_id` per batch — `get_or_create_agent` / fleet-enforcement / quota / idempotency all assume this). Per-item attribution (each memory to its own agent) would require re-scoping all of those and is a larger multi-layer change, out of scope for this cut. In practice memclawd flushes one agent's captures per batch, so the unanimous case is the common one; mixed batches simply degrade to today's behavior.

## Tests

`core-api/tests/test_broker_write_attribution.py` — unit tests for the helper: unanimous → agent, mixed → install, no-metadata → install, empty-id ignored, single-named-among-unnamed → agent, missing install → `broker:unknown`. ruff + mypy clean.

### DIFF
--- a/core-api/src/core_api/routes/memories.py
+++ b/core-api/src/core_api/routes/memories.py
@@ -42,6 +42,7 @@ from core_api.middleware.rate_limit import search_limit, write_bulk_limit, write
 from core_api.pagination import decode_cursor, encode_cursor
 from core_api.schemas import (
     BulkMemoryCreate,
+    BulkMemoryItem,
     BulkMemoryResponse,
     IngestCommitRequest,
     IngestRequest,
@@ -931,7 +932,9 @@ async def write_memories_bulk(
     # — the broker's own per-
     # session ``client_hash`` de-dup is the design contract per
     # cloud-data-plane.md §2.4 (gap G3). Server-derive a per-request
-    # attempt id and attribute writes to the install. Non-broker
+    # attempt id and attribute the write to the agent the batch's per-item
+    # metadata names (memclawd Layer 1 stamps ``metadata.agent_id``); a mixed
+    # or pre-Layer-1 batch falls back to the install. Non-broker
     # callers (dashboard, SDK) keep the CAURA-602 invariants in full.
     if auth.is_install_credential:
         if not bulk_attempt_id:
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@claude

Eldad-Caura:
**Fixed (Low — docstring):** `_broker_write_agent_id`'s docstring now states the actual contract — items without an `agent_id` **abstain** rather than veto, so `[None, X, X]` attributes to `X`; only *disagreeing named items*, or *no named item at all*, fall back to the install identity. (amended + force-pushed)

**Re: the first pass's Medium (metadata attribution ownership):** verified it's real — a broker can attribute a write to another in-tenant agent by naming it in item metadata, because the write path defaults `body.fleet_id` to the *target* agent's own fleet, so `enforce_fleet_write` compares that fleet to itself and passes. Accepting it as a **within-tenant, attribution-only** trade-off for now: brokers are trusted customer components, it isn't cross-tenant, and a caller-supplied top-level `agent_id` could already do the same before this change.

**Follow-up (noted, not blocking):** install-ownership verification — trust a metadata `agent_id` only if the agent is owned / first-touched by the requesting install (via `agents.install_id`), else fall back to `broker:<install>`.

Gates green: ruff, ruff-format, mypy, 6 unit tests. @erni-a please review.

erni-a (approved):

PR #559, opened 2026-07-16
feat(memories): surface agent display_name on memory responses

Eldad-Caura:
## What

Every memory surface — REST `GET /memories` (list + get-by-id), MCP `memclaw_recall` / `memclaw_list` — returned the raw `agent_id`. The `agents` table already carries a human-readable `display_name` (migration 010); it just was never joined into memory responses. This adds a **NULL-safe `agent_display_name`** so clients can show a friendly agent name and fall back to `agent_id` when there isn't one.

## Changes

**core-storage-api**
- LEFT JOIN `agents` on `(tenant_id, agent_id)` in the five memory read paths (`memory_scored_search`, `memory_list_by_filters`, `memory_get_detail`, `memory_load_by_ids`, `memory_find_successors`), attaching `display_name` onto the returned row. A shared `_AGENT_DISPLAY_JOIN` predicate + `_attach_agent_display_names` helper keep them from drifting.
- `MEMORY_FIELDS` gains `"agent_display_name"` (a join output, not a column; `orm_to_dict` reads it via getattr).

**core-api**
- `MemoryOut.agent_display_name: str | None`, populated in both serializers + the get-by-id response.

## Behavior / safety

- The join is **1:0..1** (`agents` has `UNIQUE(tenant_id, agent_id)`), so it never multiplies rows.
- An agent with no `agents` row (e.g. `broker:<install>` ids) or a NULL `display_name` yields `None` — clients fall back to `agent_id`. Broker-captured agents currently have NULL `display_name`; populating those is separate broker-side work.
- **No migration** — the `display_name` column already exists.

## Tests

`test_memory_carries_agent_display_name`: a seeded agent `display_name` surfaces on the list, get, and search paths; a memory whose agent has no `display_name` reads back `None`. Ran locally against Postgres — new test + memory/reports/recall suites (55 tests) pass; ruff + mypy clean on both packages.


Eldad-Caura:
@claude

Eldad-Caura:
@claude — updated: fixed the MemoryOut validation in the search-pipeline test, and converted `memory_get_detail` to the shared `_AGENT_DISPLAY_JOIN` (no more third query), addressing the earlier review note.

Eldad-Caura:
@erni-a clean review (✅ no issues) and all CI green — ready for your approval when you have a moment.

Surfaces the agent's human-readable name (`agents.display_name`) on memory responses (REST `/memories` list + get, MCP `recall`/`list`) via a NULL-safe `agent_display_name`; clients fall back to `agent_id` when absent. No migration — the column already existed. The earlier review note (extra query in `memory_get_detail`) is resolved: it now uses the shared `_AGENT_DISPLAY_JOIN` like the other read paths. DCO-signed.

erni-a (approved):

PR #560, opened 2026-07-16
feat(memories): verify broker-write agent ownership (lenient)

Eldad-Caura:
## What

Hardens broker-write attribution (the **lenient** variant of backlog #2). Broker (install-credential) bulk writes trusted the client-supplied `metadata.agent_id` verbatim (#556), so an install could attribute a memory to **any** agent in its tenant — within-tenant and label-only, but the "which agent" attribution wasn't trustworthy.

## Approach: lenient ownership check

- **New `agents.owner_install_uuid`** (migration 031) — the credential-install (`auth.install_uuid`) that *first* wrote memories as the agent. First-touch, never overwritten (mirrors `install_id`'s stability), but a **separate** field from `install_id` (which stays the plugin's admin-grouping suffix — a different id namespace, so a dedicated column rather than overloading it).
- `get_or_create_agent` threads `owner_install_uuid`; the broker bulk-write path stamps it first-touch, **only for install credentials** (symmetric with the attribution gate).
- **Lenient gate** `_broker_owned_agent_id`: degrade the chosen `agent_id` to `broker:<install>` **only** when the named agent's `owner_install_uuid` is non-NULL and belongs to a *different* install. Never rejects a write; unclaimed / nonexistent / same-install agents keep their attribution.

## Behavior / safety

- No change for legitimate single-install use; a degrade happens only on genuine cross-install attribution.
- `install_id` (plugin admin grouping) is untouched; `owner_install_uuid` is nullable (no default, no index — read per-agent, never a query predicate).
- Migration 031 is a non-destructive `add_column`, applied via the storage-service alembic lifespan flow.

## Tests

`tests/test_broker_owned_agent_id.py` covers all gate branches (owned-by-this / owned-by-different → degraded / null-owner / nonexistent / already-fallback short-circuit). Locally: gate + memory/reports suites pass against Postgres; ruff + mypy clean on both packages.

### DIFF
--- a/common/models/agent.py
+++ b/common/models/agent.py
@@ -29,6 +29,13 @@ class Agent(Base):
     # heartbeat and persists it to ``install.json`` in the plugin data dir;
     # never rotates. Indexed because the admin UI groups agents by install.
     install_id: Mapped[str | None] = mapped_column(String(32), index=True)
+    # ``owner_install_uuid``: the credential-install (``x-install-uuid``, the
+    # 36-char value) that FIRST wrote memories as this agent. First-touch,
+    # never overwritten (mirrors ``install_id``'s stability). SEPARATE from
+    # ``install_id`` — this is the write-credential's install, used only for
+    # the broker lenient ownership gate; ``install_id`` stays the plugin's
+    # admin-grouping suffix. No index: read per-agent, never a query predicate.
+    owner_install_uuid: Mapped[str | None] = mapped_column(String(36))
     trust_level: Mapped[int] = mapped_column(SmallInteger, nullable=False, server_default=text("1"))
     search_profile: Mapped[dict | None] = mapped_column(JSONB)
     # ``belonging_type`` / ``owner_ref``: the typed agent→owner relationship the
--- a/core-api/src/core_api/routes/memories.py
+++ b/core-api/src/core_api/routes/memories.py
@@ -944,6 +944,13 @@ async def write_memories_bulk(
             bulk_attempt_id = f"broker-{auth.install_uuid or 'unknown'}-{_uuid.uuid4()}"
         if not body.agent_id:
             body.agent_id = _broker_write_agent_id(body.items, auth.install_uuid)
+        # Ownership gate runs on EVERY broker write, not just the
+        # auto-derived case: a broker caller that pre-populates ``agent_id``
+        # in the request body must not bypass the check and write under an
+        # agent owned by a different install. The gate short-circuits
+        # (chosen == fallback, or owner NULL / this install) so the common
+        # paths stay lookup-free.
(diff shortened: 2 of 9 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@claude

Addressed both review findings on the latest push (`970915f`):
- **High (explicit `agent_id` bypass):** the ownership gate now runs on *every* broker write, not just the auto-derived path — moved `_broker_owned_agent_id` out of the `if not body.agent_id:` block.
- **Medium (first-touch TOCTOU):** added a post-create ownership re-check in `_write_memories_bulk_inner` that reads the committed `owner_install_uuid` (storage re-selects FOR UPDATE, never overwrites a set owner) and degrades the race loser to `broker:<install>`.

Also fixed the CI failure (migration-head guard test bumped to head `031`) and added route-level tests for both gaps (explicit-agent_id gating; post-create degrade / no-degrade / non-broker skip). Please re-review.

Eldad-Caura:
@claude

Addressed the High finding from the last review on `8d3192c`.

**Fallback identity pre-claim** — fixed at the root by *reserving* the `broker:` namespace rather than detecting the bad state after the fact. In `_broker_owned_agent_id`, any chosen id in the `broker:` namespace that isn't *this* install's own fallback is degraded to this install's own fallback, with **no lookup**:

```python
if chosen.startswith(_BROKER_LABEL_PREFIX):
    return fallback
```

So an attacker naming `broker:<victim-uuid>` (in `body.agent_id` or item metadata) is degraded to `broker:<attacker>` and can never first-touch/own the victim's fallback — the capture path is closed before any ownership stamp. This fails **safe** (degrade to self), avoiding a 500-on-write that the raise-based approach would introduce. Tests: `test_foreign_broker_label_degraded_without_lookup` (unit) and `test_cannot_preclaim_another_installs_fallback` (route).

For the Low finding (wasted lookup under the dark `bind_write_identity_to_auth` flag): documented that the post-create re-check is the authoritative guard and the outer gate is a lookup-free fast path (the flag is off by default and `auth.agent_id` is normally None for broker credentials, so the override rarely applies).

Note: the single-write `POST /memories` path lacks this broker boundary entirely — tracked as a separate follow-up rather than expanding this PR's scope. Please re-review.

Eldad-Caura:
@erni-a — CI is green and the Claude review is clean (✅ no issues found) after addressing all findings (explicit-agent_id gate bypass, first-touch TOCTOU, and the reserved-namespace fallback pre-claim). Ready for your approval.

erni-a (approved):

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #561, opened 2026-07-17
fix(logging): emit Datadog `status` field so log levels survive intake

Eldad-Caura:
## Problem

Every log the fleet ships to Datadog arrives as `status:info` — verified 2026-07-17: ~1.6M logs over 3 days across prod+staging (core-api, platform-storage-api, platform-auth-api, core-storage-reader/writer, core-worker, platform-admin-api, platform-audit-api, platform/core-operations), **zero** logs of any other status. `status:error` queries and any log-based alerting are blind — the July 5–12 staging Pub/Sub dead-letter storm (thousands of consumer failures) produced no error-status logs.

## Root cause

`common/structlog_config.py` deliberately emits GCP-style `severity` (uppercase; Cloud Logging promotes it) and drops structlog's lowercase `level`. But Datadog's JSON preprocessing derives the official log status from the **`status`** attribute — and the only `status` on the wire comes from the serverless-init envelope, which stamps the stream default (`info` for stdout) onto every line. Observed on live logs: `@severity:WARNING` sitting next to `@status:info` and official status `info`.

## Fix

New `_add_datadog_status` processor in the JSON chain (both the native structlog path and the stdlib bridge, via `_json_processors()`):

- Any recognized GCP severity — method-derived or an explicit `severity=` override, NOTICE/ALERT/EMERGENCY included — lowercases to a syslog-style status Datadog's vocabulary understands, so both backends see the same level.
- GCP's `DEFAULT` and garbage/non-string severities fall back to the method-derived level rather than leak a value intake would degrade to `info`.
- The membership set is derived from `_LEVEL_TO_SEVERITY` so the two can't drift.
- `status` joins `_RESERVED_OUTPUT_KEYS`: an `extra={"status": 200}` (app-domain value) can no longer hijack the official status at intake.

`common/` is vendored OSS→enterprise, so the re-vendor bot propagates this to the enterprise fleet after merge.

## Verification

- 32/32 tests in `tests/test_logging.py` (5 new unit tests + reserved-key param case + an end-to-end stdlib-bridge test asserting `status`+`severity` pairs).
- `ruff check` clean; `ruff format --diff` produces no hunks touching the new code (the files carry pre-existing formatter drift that CI doesn't police on these paths — left untouched to keep the diff focused).
- `mypy` clean for both core-api (188 files) and core-storage-api (71 files).
- Post-merge check once staging picks it up: `env:staging status:(warn OR error)` in Datadog Logs should start returning rows (e.g. the recurring "Contradiction check failed" WARNINGs).

## Fallback if intake still pins info

If the serverless-init envelope status somehow still wins over the in-payload field, the 2-minute Datadog-side alternative is a pipeline **status remapper** on `@severity` (Logs → Pipelines). Not included here since the emit-side field is the canonical mechanism and works for any future log consumer.

### DIFF
--- a/common/structlog_config.py
+++ b/common/structlog_config.py
@@ -6,7 +6,17 @@ this, log lines written through `structlog` end up as severity `DEFAULT`
 (unlabeled) in Cloud Logging, which breaks severity filtering and distorts
 severity-based histograms.
 
-This module wires those two fields in via processors, and is safe to call
+Datadog derives a log's official status from the JSON `status` field (the
+first attribute its JSON preprocessing checks). Observed intake behavior for
+our org: `severity` is NOT consulted, and the serverless-init sidecar that
+ships Cloud Run stdout stamps every line with the stream-derived status
+`info` — so without an explicit `status` field every log, ERRORs included,
+lands in Datadog as status:info, leaving `status:error` queries and log
+monitors blind (verified 2026-07-17 against prod/staging logs, all 1.6M/3d
+status:info). `_add_datadog_status` emits the field so both backends see
+the real level.
+
+This module wires those fields in via processors, and is safe to call
 from multiple service entry points because it's idempotent. Callers pass
 their own `environment` and `log_level` so this module stays agnostic of
 per-service settings schemas.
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@erni-a — clean Claude review (no findings) and all checks green (CI, DCO, CodeQL, security scan). Ready for approval.

erni-a (approved):

PR #562, opened 2026-07-17
feat(memories): extend broker-write ownership boundary to single writes

Eldad-Caura:
## What

Follow-up to #560. The broker (install-credential) agent-ownership boundary added there only covered **`POST /memories/bulk`**. The single-write **`POST /memories`** path (`_write_memory_inner`) had none of it — no ownership gate, no `owner_install_uuid` stamp, no post-create re-check — so a broker credential could bypass the boundary just by writing one memory at a time and naming an agent owned by a different install. `is_install_credential` is set from the gateway header on **every** route, so nothing confined the gap to the bulk endpoint.

In normal operation memclawd uses bulk writes (cloud-data-plane.md §2.4), so this is defense-in-depth rather than an actively-exploited path — but the boundary shouldn't depend on the broker's choice of endpoint.

## How

Extract a shared helper and route **both** write paths through it, so the boundary can't be bypassed by an install's choice of endpoint:

```python
async def resolve_write_agent(chosen_agent_id, auth, tenant_id, fleet_id, *, require_approval=False) -> tuple[dict, str]
```

It performs the whole boundary in one place: **gate** (degrade a foreign / reserved `broker:` id via `_broker_owned_agent_id`) → **first-touch owner stamp** → **post-create TOCTOU re-check**. `_write_memory_inner` and `_write_memories_bulk_inner` both call it.

- The gate moves out of the bulk outer `write_memories_bulk` (which keeps only its bulk-specific `_broker_write_agent_id` fan-in) into the shared helper. It now runs **after** the idempotency-replay short-circuit (so it's skipped on replays — a small win) and **after** the Phase-2 `bind_write_identity_to_auth` override (so it applies to the bound identity, with no discarded work).
- **Non-broker** (dashboard / SDK) callers pass straight through `get_or_create_agent` unchanged — the stamp and gate are broker-only, keyed on `is_install_credential` (a stray `x-install-uuid` on a non-broker call is ignored).
- The single-write `require_approval` + `trust_level == 0` approval check is preserved (`require_approval` threads through the helper).
- Kept out of `get_or_create_agent` / storage `agent_add` deliberately: those are called by non-broker paths (fleet routes, MCP, `enforce_fleet_write`) and shouldn't rewrite which agent a memory is attributed to. The storage layer already enforces the *field* (FOR UPDATE re-select, never-overwrite of `owner_install_uuid`); the route-service layer is the right altitude for the *attribution* decision.

No storage/schema change — `owner_install_uuid` + migration 031 landed with #560.

## Testing

- `resolve_write_agent` unit coverage: gate degrade (foreign agent), reserved-namespace degrade, self-owned keep-through, post-create race-lost degrade, non-broker passthrough, `require_approval` passthrough.
- End-to-end that **both** routes attribute through the helper — the single-write foreign-agent degrade is the exact gap this closes; the bulk path still degrades via the shared helper.
- Pure gate branches remain unit-tested in `test_broker_owned_agent_id.py`.
- Local: `ruff check` + `ruff format --check`, `mypy` (core-api + core-storage-api), and the broker/write-path unit suites all green.

## Note / follow-up (out of scope)

The MCP write path (`mcp_server.py` `memclaw_write`) also attributes under a caller-supplied `agent_id` and bypasses this boundary — but it is **not broker-reachable today** (the MCP middleware never plumbs `x-caura-credential-kind` / `x-install-uuid`, so `is_install_credential` is never set there). It's a latent defense-in-depth gap only if install credentials are ever routed to `/mcp`; noting it here rather than expanding this PR's scope.


Eldad-Caura:
@claude

Follow-up to #560. Extends the broker agent-ownership boundary to the single-write `POST /memories` path, which #560 left uncovered. Extracts a shared `resolve_write_agent` helper (gate + owner stamp + post-create re-check) and routes both `_write_memory_inner` and `_write_memories_bulk_inner` through it, so an install can't bypass the boundary by choosing the single- vs bulk-write endpoint. Non-broker callers pass through unchanged; the single-write `require_approval`/trust check is preserved. Please review.

Eldad-Caura:
@claude

Eldad-Caura:
@erni-a pleae approve

erni-a (approved):

PR #565, opened 2026-07-17
feat(memories): enforce broker-write ownership boundary on the MCP write path

Eldad-Caura:
## What

Follow-up to #560 / #562. The broker (install-credential) agent-ownership boundary — *an install may only attribute a memory to an agent it owns* — was enforced on the REST write paths (`/memories`, `/memories/bulk`) but **not on the MCP write tool** (`memclaw_write`): it resolved its agent id via `effective_write_agent_id` and went straight to `enforce_fleet_write` → `create_memory`, with no gate, no `owner_install_uuid` stamp, and no post-create re-check. The MCP ASGI middleware also never plumbed the credential kind, so the MCP surface couldn't even *identify* a broker.

Not broker-reachable in today's deployment (memclawd writes via REST), so this is defense-in-depth — but the boundary shouldn't depend on the caller's choice of transport.

## How

Chosen approach (per discussion): **enforce uniformly** — make the MCP path apply the same boundary as REST rather than refusing broker credentials.

- **Lift `resolve_write_agent` + the broker helpers** (`_broker_owned_agent_id`, `_broker_label`, `_owned_by_other_install`, `_BROKER_LABEL_PREFIX`) from `routes/memories.py` into `services/agent_service.py` — the shared home next to `get_or_create_agent` / `lookup_agent` — so all three write entry points use one definition. The signature now takes `is_install_credential` + `install_uuid` instead of an `AuthContext`, keeping `agent_service` free of the web-auth type and letting the MCP path (which has no `AuthContext`) call it from its context vars.
- **Plumb `X-Caura-Credential-Kind` + `X-Install-UUID`** through the MCP ASGI middleware into context vars (gateway-verified path only, mirroring the existing `x-agent-id` / `x-capabilities` plumbing), with `_is_install_credential` / `_get_install_uuid` helpers.
- **`memclaw_write` routes attribution through `resolve_write_agent`** before `enforce_fleet_write`, so a broker MCP write degrades a foreign agent to `broker:<install>` exactly like REST. Non-broker (interactive) writes pass straight through, keyed on `is_install_credential`.

No storage/schema change (`owner_install_uuid` + migration 031 landed with #560).

## Testing

- `resolve_write_agent` unit coverage (gate degrade, reserved-namespace degrade, self-owned keep, post-create race, non-broker passthrough, `require_approval` passthrough) + the gate unit tests, now against `agent_service`.
- End-to-end that **all three write surfaces** — REST single (`_write_memory_inner`), REST bulk (`_write_memories_bulk_inner`), and MCP (`memclaw_write`) — attribute through the shared helper, including the MCP context-var → credential-identity plumbing.
- The `mcp_env` fixture gains a `resolve_write_agent` passthrough stub (write tools now hit it); the boundary tests restore the real function and mock `agent_service` storage.
- Local: `ruff check` + `ruff format --check`, `mypy` (core-api + core-storage-api), and the broker / MCP-write / REST-write unit suites all green.

## Scope note / follow-up (out of scope here)

The altitude review surfaced that the broker-reachable **`/evolve/report` and `/insights/generate`** paths resolve identity via a *separate* helper (`resolve_caller_and_gate`) and still attribute memories under a caller-named `agent_id` with only a trust gate — no ownership check. That's a distinct mechanism from `resolve_write_agent`; **tracked as a follow-up**, not folded in here to keep this PR to the MCP write path. (`/ingest/commit` + `/stm/promote` are the same latent class but not exercised by the current plugin client.)

There's also a pre-existing minor redundancy — `resolve_write_agent` and `enforce_fleet_write` each call `get_or_create_agent` (one extra read on every write, REST and MCP) — left as a separate optimization.

### DIFF
--- a/core-api/src/core_api/mcp_server.py
+++ b/core-api/src/core_api/mcp_server.py
@@ -54,6 +54,7 @@ from core_api.services.agent_service import (
     enforce_fleet_read,
     enforce_fleet_write,
     get_or_create_agent,
+    resolve_write_agent,
 )
 from core_api.services.audit_service import log_action, log_cross_tenant_read
 from core_api.services.capability_usage import record_usage
@@ -105,6 +106,16 @@ _readable_tenant_ids_var: contextvars.ContextVar[list[str] | None] = contextvars
     "mcp_readable_tenant_ids", default=None
 )
 _scopes_var: contextvars.ContextVar[set[str] | None] = contextvars.ContextVar("mcp_scopes", default=None)
+# memclawd (broker) identity, plumbed from the gateway on the verified path.
+# X-Caura-Credential-Kind distinguishes ``install_credential`` (broker) from
+# ``user_api_key`` (dashboard/SDK); X-Install-UUID is the broker's install id.
+# Drives the agent-ownership boundary on MCP writes (parity with REST auth).
+_credential_kind_var: contextvars.ContextVar[str | None] = contextvars.ContextVar(
+    "mcp_credential_kind", default=None
+)
+_install_uuid_var: contextvars.ContextVar[str | None] = contextvars.ContextVar(
+    "mcp_install_uuid", default=None
+)
 
 _UNAUTH = "__unauthenticated__"
 _ADMIN = "__admin__"
@@ -276,6 +287,18 @@ class MCPAuthMiddleware:
                 else ""
             )
             _scopes_var.set({s.strip() for s in caps_header.split(",") if s.strip()} if caps_header else None)
+            # Credential provenance for the write-ownership boundary. Only
+            # honored on the gateway-verified path — a direct client must not
+            # be able to self-assert an install identity and write under a
+            # broker's ownership namespace.
+            _credential_kind_var.set(
+                (headers.get(b"x-caura-credential-kind", b"").decode().lower() or None)
+                if via_gateway
+                else None
+            )
+            _install_uuid_var.set(
+                (headers.get(b"x-install-uuid", b"").decode() or None) if via_gateway else None
+            )
 
         await self.app(scope, receive, send)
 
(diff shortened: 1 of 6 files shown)


Eldad-Caura:
@claude

Follow-up to #560/#562 — closes the MCP write-path gap. Lifts `resolve_write_agent` + the broker helpers into `agent_service` (shared home; signature switched from `AuthContext` to `is_install_credential`+`install_uuid` fields so the MCP path can call it from its context vars), plumbs `X-Caura-Credential-Kind`/`X-Install-UUID` through the MCP ASGI middleware, and routes `memclaw_write` through the boundary before `enforce_fleet_write` — so a broker MCP write degrades a foreign agent to `broker:<install>` exactly like REST. Non-broker (interactive) writes pass through unchanged.

Please review. Note the scope note in the description: `/evolve/report` + `/insights/generate` use a separate identity helper (`resolve_caller_and_gate`) and are tracked as a follow-up, not in this PR.

Eldad-Caura:
@claude

Addressed on the latest push (`fa09817`):

- **Low (private cross-module import):** renamed `_broker_label` → `broker_label` (public) in `agent_service`, updated the import + call site in `memories.py`.
- **Medium (trust_level==0 approval gate on MCP):** intentionally NOT folded into this PR, filed as a separate follow-up. Rationale: that gate is a distinct feature (per-agent **approval** via `require_agent_approval`), orthogonal to the **ownership** boundary this PR establishes. It's pre-existing (MCP never had it), and I verified REST **bulk** also lacks it — only REST single-write enforces it (`routes/memories.py:833`; bulk is the broker fan-in path that auto-creates many agents). So MCP currently matches REST bulk; whether MCP (and bulk) should enforce per-agent approval is a separate design decision, not part of establishing the ownership boundary.

Please re-review.

Eldad-Caura:
@erni-a — CI green and the Claude review is clean (✅ no issues) after addressing both findings (renamed `_broker_label` → `broker_label`; the trust-level approval-gate finding is a distinct pre-existing concern — REST bulk also lacks it — filed as a separate follow-up). Ready for your approval.

erni-a (approved):

PR #563, opened 2026-07-17
fix(logging): format-stabilize the severity set for the vendored copy

Eldad-Caura:
Follow-up to #561. The 95-char `_DATADOG_STATUS_SEVERITIES` union passes OSS lint but fails caura-memclaw-enterprise's `ruff format --check common/` gate (which runs at ruff-default 88 — no config resolves for `common/` there), blocking the byte-identical re-vendor and turning enterprise Backend CI red on every PR (first hit: enterprise #811).

- Inline the single-use `_OVERRIDE_ONLY_SEVERITIES` into a magic-trailing-comma set literal — the one form `ruff format` preserves at any line width (verified stable at 88/100/110, and against the enterprise pre-commit hook directly).
- Document the vendoring constraint in the module docstring so future edits aren't authored blind to the enterprise 88-col gate (the docstring propagates to the enterprise copy on re-vendor).

No behavior change. 32/32 logging tests pass; `ruff check` clean; validated against the exact enterprise `ruff-format` hook.

Follow-up (tracked separately): add an OSS-side `ruff format --check` at 88 over the identical-policy vendored files so this class of drift is caught pre-merge here instead of post-merge there.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #567, opened 2026-07-17
ci: gate vendored common/ files with ruff format at enterprise defaults

Eldad-Caura:
Closes the loop on the 2026-07-17 incident chain (#561 merged clean here → every enterprise PR's Backend CI turned red on the vendored-drift/format gates → #563 reflow): OSS CI now format-checks exactly the two identical-policy vendored files (`common/__init__.py`, `common/structlog_config.py`) at ruff defaults (88 cols) — the same contract enterprise CI enforces on the vendored copies — so the breakage is caught pre-merge here instead of post-merge there.

- List mirrors the `policy=identical` entries of enterprise's `scripts/vendored_files_manifest.json` (source of truth, noted in the step comment). Cross-repo fetch was rejected: the manifest lives in a private repo and this is a public-repo `pull_request` workflow (fork PRs get no secrets).
- `--isolated` pins ruff-default behavior even if a root-level ruff config is ever added.
- Deliberately not gating all of `common/` — 7 non-vendored files exceed 88 today and aren't held to that contract.
- Verified: the exact gated command passes on both files locally (`2 files already formatted`); YAML/actionlint clean (one pre-existing info finding untouched).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,6 +76,17 @@ jobs:
       - name: Run Ruff formatter check (core-storage-api)
         run: uv run ruff format --check core-storage-api/src/
 
+      - name: Run Ruff formatter check (vendored common/ files at ruff defaults)
+        # This list mirrors the policy=identical entries of caura-memclaw-enterprise's
+        # scripts/vendored_files_manifest.json — the source of truth; keep in sync.
+        # Enterprise CI format-checks the vendored copies at ruff DEFAULTS (88 cols),
+        # so gate them here too or a merge turns enterprise CI red until a reflow
+        # lands (full story: the vendoring note in common/structlog_config.py's
+        # docstring). ``--isolated`` pins default behavior even if a root ruff
+        # config is ever added; deliberately NOT all of common/ — non-vendored
+        # files there are not held to the 88-col contract.
+        run: uv run ruff format --check --isolated common/__init__.py common/structlog_config.py
+
       - name: Guard against raw-string MCP error returns (audit B4 / CAURA-000 #147)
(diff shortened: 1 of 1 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
Review verdict handled — the one Low finding (hardcoded list can drift from the enterprise manifest) is a real risk but I'm declining the suggested remedies, because both require OSS CI to read the enterprise-private manifest, which the public/fork-PR constraint disallows:

- **Cross-repo / jq-the-manifest**: the manifest lives in the private caura-memclaw-enterprise repo. This is a public-repo `pull_request` workflow — fork PRs get no secrets, so a checkout/fetch of the private manifest breaks CI for outside contributors (or forces a `pull_request_target` hazard). Disqualified.
- **Mirror the manifest into OSS**: creates a second source of truth that itself drifts (the same class of problem), and publishes enterprise-internal vendoring policy into a public repo — for a list of two.

The residual risk is bounded and self-announcing: a newly vendored identical-policy file not added here regresses only that one file to the *pre-#567* status quo, and enterprise CI's own drift/format gate still goes red on it (exactly the 2026-07-17 detection path). So this PR is a strict improvement, not a new gap. The coherent place to make coverage self-maintaining is **enterprise-side** — the repo that owns the manifest asserting every `identical` entry is covered by this OSS gate — which I'll note as a follow-up there rather than bloat this PR.

Verified: manifest has exactly two `identical` entries (`common/__init__.py`, `common/structlog_config.py`) = this step's list; all checks green.

@erni-a — ready for approval.

erni-a (approved):

PR #568, opened 2026-07-17
feat(memories): enforce per-agent approval gate on MCP writes

Eldad-Caura:
## What

Follow-up flagged in the #565 review. The per-agent **approval** gate (`require_agent_approval` → new agents start at `trust_level 0`, blocked until an admin sets trust ≥ 1) was enforced only on the REST **single-write** path. The MCP write tool (`memclaw_write`) bypassed it entirely — it neither threaded `require_approval` into agent creation nor checked `trust_level == 0`, so a brand-new agent could write via MCP without approval.

This is distinct from the broker **ownership** boundary (#560/#562/#565/#566) — it's the agent-**approval** feature.

**Decision (confirmed):** enforce on **MCP** (interactive / single-agent, analogous to REST single-write); leave REST **bulk** unchanged — it's the broker (memclawd) fan-in path where per-agent approval would create trust-0 rows and 403 whole capture batches.

## How

- `memclaw_write` resolves the tenant config and threads `require_approval=config.require_agent_approval` into `resolve_write_agent` (so a new agent is created at `trust_level 0` when required), then refuses a trust-0 agent with an `AGENT_NOT_APPROVED` error envelope **before** the write — mirroring `_write_memory_inner`'s 403.
- Documented in `_write_memories_bulk_inner` **why** bulk deliberately does NOT gate approval (broker fan-in path; gating would create trust-0 rows and 403 whole batches, breaking capture).

## Testing

- `memclaw_write` refuses an unapproved (trust-0) agent (`AGENT_NOT_APPROVED`, no persist); `require_approval` threads through to the resolver.
- The `mcp_env` fixture gains a default `resolve_config` stub (approval off) so write tools that now resolve config don't hit storage.
- Local: `ruff check` + `ruff format --check`, `mypy` (both packages), and the MCP write + recall suites green.

## Note

Independent of #566 (evolve/insights). Rebased onto `main` after #565 merged.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #569, opened 2026-07-17
feat(memories): extend broker-write ownership boundary to ingest/stm

Eldad-Caura:
## What

Completes the broker (install-credential) agent-ownership boundary. #560/#562/#565/#566 covered the memory-write paths and #568 added the approval gate; this closes the **last two broker-reachable memory-write surfaces**: `POST /ingest/commit` and `POST /stm/promote`.

Both attribute a memory under a caller-supplied `agent_id` (via `create_memory` / `create_memories_bulk`) with **no trust/registration gate** — so, like evolve/insights, a broker (whose `auth.agent_id` is None) could name another install's agent. Neither is exercised by the current plugin client, so this is **defense-in-depth**.

## How

- **`/stm/promote`** (`routes/stm.py`): the existing `auth.agent_id != body.agent_id` guard is a no-op for brokers (`auth.agent_id` is None). Degrade `body.agent_id` via `broker_owned_agent_id` for install-credential callers before `promote()`.
- **`/ingest/commit`** (`routes/memories.py`): `ingest_commit` takes no `AuthContext`, so degrade `body.agent_id` in the endpoint before calling it.

Gate-only (no `owner_install_uuid` stamp) — neither path first-touches an agent, so this only redirects a foreign attribution to the caller's own `broker:<install>` fallback (matching the evolve/insights treatment). `broker_owned_agent_id` is the shared gate (unit-tested in `test_broker_owned_agent_id.py`).

## Testing

- Route-level broker-degrade + non-broker-passthrough for both endpoints.
- Local: `ruff check` + `ruff format --check`, `mypy` (both packages), and the ingest / stm / gate suites green.

## Result

Every broker-reachable memory-write surface — REST single/bulk, MCP, evolve, insights, ingest, stm — now enforces the ownership boundary through the one shared `broker_owned_agent_id` / `resolve_write_agent` in `agent_service`.

### DIFF
--- a/core-api/src/core_api/routes/memories.py
+++ b/core-api/src/core_api/routes/memories.py
@@ -60,6 +60,7 @@ from core_api.schemas import (
 from core_api.services.agent_service import (
     authorize_memory_access,
     broker_label,
+    broker_owned_agent_id,
     enforce_delete,
     enforce_fleet_read,
     enforce_fleet_write,
@@ -1492,6 +1493,12 @@ async def ingest_commit_endpoint(
     auth.enforce_read_only()
     auth.enforce_usage_limits()
     auth.enforce_tenant(body.tenant_id)
+    # Broker ownership boundary: degrade a foreign / reserved agent id to the
+    # install's own broker:<install> fallback so a broker can't attribute an
+    # ingested memory to an agent owned by another install (parity with the
+    # data-plane write paths; ingest_commit itself takes no AuthContext).
+    if auth.is_install_credential and body.agent_id:
+        body.agent_id = await broker_owned_agent_id(body.agent_id, auth.install_uuid, body.tenant_id)
     if auth.tenant_id:  # skip for admin
         await check_and_increment(body.tenant_id, "write")
     return await ingest_commit(body)
--- a/core-api/src/core_api/routes/stm.py
+++ b/core-api/src/core_api/routes/stm.py
@@ -9,6 +9,7 @@ from pydantic import BaseModel
 
 from core_api.auth import AuthContext, get_auth_context
 from core_api.config import settings
+from core_api.services.agent_service import broker_owned_agent_id
 
 logger = logging.getLogger(__name__)
 
@@ -142,6 +143,14 @@ async def promote_stm(
             status_code=403,
             detail=f"agent_id '{body.agent_id}' does not match the authenticated agent identity.",
         )
+    # Broker ownership boundary: an install-credential caller may only promote
+    # under an agent it owns. Degrade a foreign / reserved-namespace agent id to
+    # its own broker:<install> fallback (parity with the data-plane write paths).
+    # The auth.agent_id guard above is a no-op for brokers (auth.agent_id is None),
+    # so this is the check that actually constrains a broker's promote target.
+    if auth.is_install_credential and body.agent_id:
+        body.agent_id = await broker_owned_agent_id(body.agent_id, auth.install_uuid, tenant_id)
+
     from core_api.services.stm_service import promote
 
     result = await promote(
--- a/core-api/src/core_api/services/agent_service.py
+++ b/core-api/src/core_api/services/agent_service.py
@@ -153,6 +153,7 @@ async def broker_owned_agent_id(chosen: str, install_uuid: str | None, tenant_id
     install can't write under another install's agent id.
 
     Lenient — never blocks a legitimate first write:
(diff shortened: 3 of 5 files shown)


Eldad-Caura:
@claude

Closes the last two broker-reachable memory-write surfaces (`/ingest/commit` + `/stm/promote`) — the latent gaps noted in the earlier reviews. Both attribute under a caller-supplied `agent_id` with no trust gate, so (like evolve/insights) the route now degrades a foreign id via `broker_owned_agent_id` for install-credential callers before the write. Gate-only (neither path first-touches an agent). Defense-in-depth — not broker-reachable by the current plugin client. Please review.

Eldad-Caura:
@claude

Addressed the Medium finding from the last review (missing `install_uuid` guard → shared `broker:unknown` fallback on headerless install credentials).

Fix, applied in the shared gate `broker_owned_agent_id` (agent_service.py) so it covers ingest/stm plus the already-merged `resolve_write_agent` / `resolve_caller_and_gate` paths consistently:

```python
# No install identity means there is nothing to enforce ownership against ...
if install_uuid is None:
    return chosen
```

When an install credential arrives without an `install_uuid`, the gate now falls through UNGATED (write as named, pre-boundary behavior) instead of degrading every foreign-owned agent onto the shared `broker:unknown` identity. Rationale: the gateway couples `x-caura-credential-kind` with `x-install-uuid`, so a set credential kind without a uuid is a contract violation — never the real adversary (a *different* install always carries its own uuid and stays gated).

Added `test_null_install_uuid_kept_without_lookup` (keeps the chosen id, asserts no `lookup_agent` call). Folded into the single PR commit (amended). Please re-review.

Eldad-Caura:
@claude

Addressed the Low finding (ungated fallthrough when `install_uuid is None`).

**Accepted — made the no-op observable, not silent.** The `install_uuid is None` branch in `broker_owned_agent_id` now emits a `logger.warning` (tenant + chosen agent_id) before returning ungated, so if the gateway ever ships an install credential without a uuid it surfaces in telemetry. It still does not block (this PR keeps the no-new-403 contract). Unit test asserts the warning fires.

**Declined — the route-guard change (`and auth.install_uuid`), for consistency.** `broker_owned_agent_id` has four call sites (`resolve_write_agent`, `resolve_caller_and_gate`, and the two routes here). Adding `and auth.install_uuid` at only the two ingest/stm sites would leave the merged `resolve_write_agent`/`resolve_caller_and_gate` paths relying on the in-gate guard — the exact per-site inconsistency the uniform in-gate guard exists to prevent. Keeping the decision in the single shared gate is the correct altitude.

On the "attacker would produce this" concern: producing `is_install_credential=True` with `install_uuid=None` requires the caller to be inside the gateway's shared-secret perimeter (`auth.py` Path 4 trusts `x-caura-credential-kind`/`x-install-uuid`/`x-agent-id` only behind the `X-Gateway-Secret` HMAC check). Anyone holding that secret can already set `x-agent-id` to any victim directly, bypassing the broker gate entirely — so this fallthrough doesn't widen the attack surface. The gate's real adversary is a legitimate *different* install, which always carries its own uuid and stays gated. Documented in-code. Please re-review.

Eldad-Caura:
@erni-a this is ready for your review/approval — CI is green on `ff1b11e`.

The automated review's Medium finding (the install_uuid=None guard → broker:unknown pooling) is fixed: `broker_owned_agent_id` now returns the chosen id ungated, in the single shared gate, so ingest/stm and the already-merged data-plane paths stay consistent, and it emits a `logger.warning` so the contract violation is observable in telemetry.

Two follow-up findings from the last re-review round were assessed and intentionally **not** applied:

1. **Raise→500 on install_uuid=None (Medium):** declining in favor of the current log-and-write-as-named. The state is only reachable behind the gateway's shared-secret perimeter (`auth.py` Path 4), and a secret-holder can already spoof `x-agent-id` directly — so a hard-fail adds no real security while introducing a new 500 path on ingest/promote for a mere gateway misconfig. Fail-open + telemetry matches this thread's deliberate graceful-degrade stance (same rationale as bulk deliberately skipping the approval gate). Confirmed with the maintainer.

2. **`and body.agent_id` "always truthy" (Low):** not dead code — a caller can send `agent_id: ""` explicitly (falsy) on both endpoints, and the guard keeps parity with the `/stm/promote` site. Left as-is.

Single commit, DCO-signed. Not merging — leaving that to the maintainer.

erni-a (approved):

PR #574, opened 2026-07-19
fix(fleet): dedup the outdated-plugin heartbeat warning

Eldad-Caura:
## Why (finding #2 from the 2026-07-19 prod error review)
`POST /fleet/heartbeat` logs a WARNING every tick a plugin is below `MIN_RECOMMENDED_PLUGIN_VERSION`. A node stuck on an old plugin heartbeats ~every 20–60s forever, so it floods: prod produced **6,301 lines/day** from just **3 internal nodes** — `eyal-open-claw-1` on plugin `1.0.1` (~every 20s), a container on `2.8.0`, and `shushants-MacBook-Air` on `2.5.0`. That volume drowns real warnings now that log severity classifies correctly (#561).

## What
Gate the warning through a per-`(node, version)` dedup with a 1h window — a module-level TTL cache, bounded (5000) + oldest-evicted, mirroring the auth-api cache idiom. The fleet-health signal still surfaces (once/hour/stuck node) without the per-tick spam. Keyed on `(node, version)` so a version change re-logs promptly. Per-instance (not shared across Cloud Run instances), so worst case is one line per instance per node per hour — still orders of magnitude below one per tick.

**Behaviour otherwise unchanged** — the auto-upgrade queueing path is untouched; only the log is throttled.

## Tests
5 new unit tests (appended to `test_fleet_heartbeat_auto_upgrade.py`): dedup, version-change re-log, per-node isolation, window expiry, size cap. Full file 61 pass; `ruff`/`mypy` clean.

## Ops follow-up (not code)
The 3 stuck nodes are all internal (a teammate's box, a container, a laptop) — worth nudging them to upgrade or decommissioning, but not a customer-fleet problem.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #575, opened 2026-07-19
fix(core-api): map unhandled upstream 5xx/429 to a retryable 503

Eldad-Caura:
## Why (finding #3 from the 2026-07-19 prod error review — option a)
The one prod 500 in 3 days was `POST /fleet/heartbeat` → `upsert_node()` → core-storage-writer returned a transient **503**, and the storage client's `raise_for_status()` let it bubble as an **unhandled** exception → the catch-all returned 500 (`INTERNAL_ERROR`) and opened an Error-Tracking issue. A transient dependency blip on an idempotent, frequently-retried endpoint shouldn't read as a hard server bug — the plugin should just retry next tick.

## What
App-level `@app.exception_handler(httpx.HTTPStatusError)` that maps an **upstream 5xx/429** to a retryable **503** (canonical `UNAVAILABLE` envelope) instead of the 500 catch-all. Fixing it at the exception mechanism covers every endpoint uniformly rather than bolting a try/except onto the heartbeat handler (the "fix the mechanism, not the symptom" altitude).

A **4xx** from an upstream means *our* request was malformed — a genuine bug, not a blip — so it still flows to the 500 catch-all unchanged.

## Tests (`tests/test_upstream_error_handler.py`)
- 500/502/503/504/429 → 503 `UNAVAILABLE`;
- 400/404/409/422 → 500 `INTERNAL_ERROR`;
- handler is registered for `httpx.HTTPStatusError`.
`ruff`/`mypy` clean.

Companion to #574 (which dedups the heartbeat's outdated-plugin log). Together they close the fleet-heartbeat findings from the error review.

### DIFF
--- a/core-api/src/core_api/app.py
+++ b/core-api/src/core_api/app.py
@@ -4,6 +4,7 @@ import os as _os
 from contextlib import asynccontextmanager
 from pathlib import Path
 
+import httpx
 from fastapi import FastAPI, HTTPException, Request
 from fastapi.exceptions import RequestValidationError
 from fastapi.middleware.cors import CORSMiddleware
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
Fixed — and the review round caught a real regression, not just the nit:

1. **CI failure (the important one):** my app-wide handler was intercepting an upstream **4xx** and turning it into a handled response, which broke `test_bulk_atomicity.py::test_storage_4xx_does_not_get_5xx_recovery_hint` — the bulk-write path deliberately lets a storage 4xx propagate as a raw `HTTPStatusError` (its own try/except only converts 5xx→504). The handler now **re-raises** any non-5xx/429 unchanged, so 4xx surfaces exactly as before (catch-all 500) and callers that rely on propagation keep working. Only genuinely-unhandled upstream **5xx/429** become a retryable 503. Verified: `test_storage_4xx_does_not_get_5xx_recovery_hint` and `test_storage_5xx_returns_504_not_500` both pass, plus my handler tests.
2. **Low finding applied:** dropped the dead `exc.response is not None else 500` guard — `HTTPStatusError.response` is always present, so `status = exc.response.status_code`.

Updated my test to assert the 4xx path re-raises (was asserting a 500 response). ruff/mypy clean; amended into the single commit.

@claude please re-review the amended commit.

Eldad-Caura:
@erni-a — clean Claude re-review (no issues) and all checks green on the rebased commit. Ready for approval. (Upstream 5xx/429 → retryable 503; the review round caught + I fixed a 4xx-propagation regression, now re-raised so only unhandled 5xx/429 change.)

erni-a (approved):

PR #579, opened 2026-07-19
feat(search): cosine-dominant candidate pool selection (A49), off by default

arkash20:
## What

Adds an opt-in **per-tenant** knob `candidate_pool_size` (default **0 = off**). When `>0`, the first-stage candidate pool is selected by semantic relevance (`similarity`) instead of the boost-distorted `score`, so **boost-demoted-but-strong matches survive the LIMIT** into the ranking/rerank stage (A50) rather than being evicted by the boost stack.

Resolution: agent profile → tenant `default_search_profile` → global constant (0). Validator caps it to 0–200.

## Why

LongMemEval 500-case failure analysis: **64%** of the gold memories production dropped from its top-20 rank **≤20 by pure cosine** — i.e. the ranking demoted strong semantic matches, it didn't fail to find them. A cross-encoder reranker (A50) only re-scores the candidate pool it's handed, so those demoted matches must first be *in* the pool. Offline ceiling: a cosine-selected pool of **50 recovers 95%** of the dropped gold; >70 adds nothing.

## Scope / behavior

- **Off by default → zero behavior change.** A49 alone widens/relevance-selects the candidate pool; the outer query still orders by `score` and `PostFilterResults` still trims to the caller's `top_k`, so the **final reorder is A50's job**. This is dark-launch infrastructure for A50.
- Enable per tenant via `default_search_profile.candidate_pool_size` for the A49+A50 A/B.

## Changes

- **core-api**: `CANDIDATE_POOL_SIZE` constant, `_SEARCH_PROFILE_RULES` entry (int 0–200), thread `candidate_pool_size` into `search_params`.
- **core-storage-api**: order the LIMIT pool by `similarity` when `candidate_pool_size > 0` (else unchanged `score` order).

## Tests

- core-api plumbing unit tests (6, passing): default-off, validate/clamp, resolve threading.
- storage integration test (real pgvector, passing): a boost-demoted high-similarity row is dropped by score-order but retained by the cosine pool; `candidate_pool_size=0` is byte-identical to default.
- Existing `/scored-search` tests: no regression (4/4).

## Not yet validated

End-to-end accuracy — that requires the **A49+A50 A/B** (re-seed + full-500), since A49 alone is pool infrastructure. Merging is safe because the flag is off.

### DIFF
--- a/core-api/src/core_api/constants.py
+++ b/core-api/src/core_api/constants.py
@@ -153,6 +153,15 @@ MAX_QUERY_LENGTH = 5000
 DEFAULT_SEARCH_TOP_K = 5
 MAX_SEARCH_TOP_K = 20
 MIN_SEARCH_SIMILARITY = 0.3
+# A49: cosine-dominant candidate selection. 0 = OFF (the first-stage candidate pool
+# is selected by the boost-distorted ``score`` — current behaviour). >0 = select the
+# candidate pool by semantic relevance (``similarity``, boost-free) at this size, so
+# boost-demoted-but-strong matches survive the LIMIT into the ranking/rerank (A50)
+# stage instead of being evicted by the boost stack. Enable per-tenant via
+# ``default_search_profile.candidate_pool_size``; keep 0 globally until the A49/A50
+# A/B validates it. Offline ceiling (LongMemEval): pool=50 recovers 95% of the gold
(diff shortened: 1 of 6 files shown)


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #580, opened 2026-07-19
fix(ci): publish semver image tags for component-prefixed release tags

ran-taig:
## Summary

Found during the Interviewer real-LLM pilot: pulling `ghcr.io/caura-ai/caura-memclaw-core-api:2.22.0` (and `:v2.22.0`, and `:2.21.0`) fails — **no version tag has been published since release-please component prefixes began; every release ships only `:latest`.**

Verified in the `backend: v2.22.0` publish run logs — the metadata step computed exactly one tag per image (`:latest`, `image.version=latest`): `docker/metadata-action`'s `type=semver` patterns can't parse `backend-v2.22.0` and drop silently.

Consequences until now: no version pinning (`MEMCLAW_VERSION`), no rollback-by-tag, README's documented pins broken. Also found: the workflow has no component gate, so **`plugin-v*` releases rebuilt the backend images** from their checkout and re-pushed `:latest` outside any backend release.

## Changes

1. **Resolve step strips the `backend-` prefix** → parser sees `vX.Y.Z` → publishes `X.Y.Z`, `X.Y`, `X` (+`latest` on release).
2. **Job gated to backend releases** (`workflow_dispatch` unaffected) — plugin releases no longer touch backend images.
3. **Tag-shape validation now runs on both event types** — a future tag-scheme change fails the run loudly instead of silently publishing latest-only again (exactly this incident's failure mode).
4. **Adds `v{{version}}` alias tags** (`vX.Y.Z`) so the README's v-prefixed pin examples work as documented, alongside bare semver.

## Test plan

- YAML parses; strip + validation regex exercised against all three real tag shapes (`backend-v2.22.0` ✓, `backend-v2.23.0-rc.1` ✓, `plugin-v2.15.0` → gated/rejected).
- End-to-end validation: the next backend release (release-please #576 is open) should publish `2.X.Y` / `v2.X.Y` / `2.X` / `2` / `latest`. Suggest merging this before #576 so that release exercises it; if needed, `workflow_dispatch` re-publish of `v2.22.0` also works as recovery.

### DIFF
--- a/.github/workflows/publish-docker.yml
+++ b/.github/workflows/publish-docker.yml
@@ -18,7 +18,7 @@ on:
   workflow_dispatch:
     inputs:
       tag:
-        description: "Release tag to (re-)publish (e.g. v1.0.0)"
+        description: "Release tag to (re-)publish (e.g. backend-v1.0.0 — the full git tag created by release-please)"
         required: true
         type: string
       push_latest:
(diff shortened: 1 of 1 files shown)


ran-taig:
@claude

ran-taig:
@claude

ran-taig:
@claude

ran-taig:
@arkash20

ran-taig:
@erni-a

erni-a (approved):

PR #582, opened 2026-07-20
fix(skills-inbox): converge the inbox card contract with the dashboard UI (A1)

ran-taig:
Phase A1 of the Skills Inbox GA plan: make the OSS backend serve the card contract the enterprise dashboard's Skill Factory page (`app/(dashboard)/skill-factory/`) already consumes, and fix the defects found while grounding it.

## What changed

**Bug fixes**
- **`evidence` 500**: Forge writes `evidence` as a free-text rationale string (`forge_service.py`), but `InboxCard.evidence` was typed `dict` — the first Forge-minted staged card failed response validation and 500'd the entire `GET /api/v1/skills-inbox`. Now `str | dict | None`, with a regression test covering string / dict / absent.
- **Trailing-slash 307**: the list route was only registered at `/skills-inbox/`; the dashboard calls the bare path, and Starlette's redirect answers with a Location built from the internal upstream host behind the gateway. The route now answers both spellings directly (bare path is the canonical schema entry).
- **Bodyless `defer` 422**: every `DeferRequest` field is optional, but the body envelope itself was required — a plain `POST …/defer` (curl operators, the documented contract) 422'd. Body is now optional.

**Card contract convergence** (all additive; flat `scan_state`/`scan_critical`/`scan_warn` and `origin` are kept for existing consumers)
- `content` — the SKILL.md body. The list is the only inbox read surface (no per-slug GET) and the UI's Edit form pre-fills from it; without it a UI edit would silently blank the content.
- `updated_at` — bumped by every inbox transition; the UI sorts/labels with it.
- `sentinel_scan: { status, critical_count, warning_count }` — nested scan verdict in the shape `normalizeInboxCard` reads.
- `forge_evidence: { cluster_size, distinct_agents }` — only on `source=forge` cards (for hand-authored docs `origin` describes the writer, not a cluster).
- `cites` — memory-ID provenance, so the drawer can link a skill back to its source memories (Phase B needs no second backend PR).

**Tests** — first dedicated coverage for the inbox routes (54 tests, unit-marked, no DB): golden Forge-card contract test (doubles as the shared fixture the frontend can assert against), the evidence-string regression, no-redirect on the bare path, admin-only actions vs member-readable list, `SKILLS_FACTORY_DISABLED` gate, per-action status matrices, 422 body validation, and the TOCTOU 409 guards (approve content-hash drift, reject-vs-approve race refusing the poison write, edit-vs-approve race).

**Docs** — `docs/skills-inbox-api.md` response example + field notes updated for the enriched shape; defer documented as body-optional.

## Not in this PR (per plan)
- A2: plumbing `X-Org-Role` through auth-api → gateway → core-api (enterprise repo + a small core-api reader) — without it, SaaS dashboard actions still 403.
- A3: enterprise frontend repairs (flag-name copy bug, deferred chip, limit heuristic).

## Testing
- `pytest tests/test_skills_inbox_routes.py` — 54 passed.
- `pytest tests/test_skill_lifecycle_transitions.py tests/test_sentinel_scanner.py tests/test_route_authz_gaps.py tests/test_mcp_doc.py` — 181 passed.
- `ruff check` / `ruff format --check` clean on both files; `mypy` reports no issues in `skills_inbox.py`.

### DIFF
--- a/core-api/src/core_api/routes/skills_inbox.py
+++ b/core-api/src/core_api/routes/skills_inbox.py
@@ -121,6 +121,31 @@ def _require_inbox_admin(auth: AuthContext) -> None:
 # ── Pydantic shapes ────────────────────────────────────────────────
 
 
+class SentinelScanSummary(BaseModel):
+    """Nested scan verdict, shaped for the dashboard card UI.
+
+    ``status`` mirrors ``data.scan.state`` (``clean`` / ``quarantined``
+    / ``failed``); the counts mirror ``critical`` / ``warn``. The flat
+    ``scan_state`` / ``scan_critical`` / ``scan_warn`` card fields carry
+    the same values for pre-existing consumers.
+    """
+
+    status: str | None = None
+    critical_count: int = 0
+    warning_count: int = 0
+
+
+class ForgeEvidence(BaseModel):
+    """Forge cluster evidence, surfaced only for ``source='forge'``
+    cards: how many behavior traces the candidate was distilled from
+    and how many distinct agents produced them. Values live in
+    ``data.origin`` on disk (the auto-gate evaluator reads them there).
+    """
+
+    cluster_size: int = 0
+    distinct_agents: int = 0
+
+
 class InboxCard(BaseModel):
     """One row in the Inbox list response. Shape matches what the
     card-UI surfaces — keep the field list in sync with plan §10.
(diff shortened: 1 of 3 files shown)


ran-taig:
@claude

ran-taig:
@claude please re-review — both review findings addressed in the amended commit (`3da1b47f`):

**Fix 1 — `content` is now opt-in.** `GET /api/v1/skills-inbox` takes `include_content: bool = false`; the default page is lean (`content: null` on every card) and the edit UI opts in with `?include_content=true`. Covered by a new test (`test_list_content_is_opt_in`) asserting the lean default, the opt-in body, and that everything else is untouched by the lean pass.

**Fix 2 — `evidence` is never `null` on the wire.** Audited the downstream consumer: the dashboard's `normalizeInboxCard`/`extractForgeEvidence` read `forge_evidence` (null-guarded), not `evidence` — so nothing breaks today — but restored `{}` as the absent-value default anyway per the review, keeping the union for Forge strings. The model now enforces it (`evidence: str | dict = Field(default_factory=dict)`, `None` deliberately excluded) and the regression test asserts string / dict / `{}`-when-absent.

Docs updated to match (query-param table row for `include_content`, merged the evidence field notes, example response labeled as the `include_content=true` shape). 55 tests pass; ruff/format clean.

ran-taig:
@erni-a

erni-a (dismissed):

ran-taig:
@erni-a

erni-a (approved):

PR #583, opened 2026-07-20
feat(auth): read gateway-plumbed X-Org-Role into AuthContext (Skills Inbox A2, reader side)

ran-taig:
## Summary

Part 1 of 2 of the Skills Inbox admin-identity fix (A2). Today every inbox action (`approve`/`edit`/`defer`/`quarantine`/`reject`) returns **403 for everyone on the SaaS gateway path** — `_require_inbox_admin` checks `org_role == 'admin'`, but Path 4 of `get_auth_context` builds the `AuthContext` from gateway headers that never carried the org role. Only standalone mode (hardcoded `org_role='admin'`) works.

This PR is the **reader side**: Path 4 now reads `X-Org-Role` into `AuthContext.org_role`.

- **Allowlisted** to the org-membership model (`admin` | `member`, normalized); unknown values are dropped so `== 'admin'` gates never see a smuggled third role.
- **Trust boundary unchanged**: read only on the branch already gated by `X-Gateway-Secret` (when configured); the gateway overwrites any client-supplied `X-Org-Role` via `proxy_set_header`.
- **Deliberate non-goal**: the `MEMCLAW_API_KEY` path (Path 2) does NOT read the header — API-key/agent callers must not gain inbox-action rights from a self-asserted header.
- **Safe to deploy first**: absent header → `org_role=None` → behavior identical to today.

Part 2 (enterprise repo, follows): platform-auth-api `/validate` emits `X-Org-Role` from the session row / JWT claim; nginx captures + forwards it on the `/api/v1/` and legacy `/api/` blocks.

## Testing

- `tests/test_org_role_plumb.py`: admin/member parse + normalization, unknown-role allowlist drop, absent-header no-op, Path 2 ignores the header, and an end-to-end check that a gateway-shaped request opens the Skills Inbox admin gate (and 403s without the role / with `member`).
- Neighboring suites green: `test_gateway_secret_gate`, `test_skills_inbox_routes`, `test_install_credential_auth`, `test_cross_tenant_auth` (89 tests).
- `cd core-api && uv run mypy src/` (CI's invocation) clean for `auth.py`; ruff check + format clean.

### DIFF
--- a/core-api/src/core_api/auth.py
+++ b/core-api/src/core_api/auth.py
@@ -451,6 +451,24 @@ async def get_auth_context(
         credential_kind = (request.headers.get("x-caura-credential-kind") or "").lower()
         is_install_credential = credential_kind == "install_credential"
         install_uuid = request.headers.get("x-install-uuid") or None
+        # Org-level role of the human principal behind the request
+        # (``org_members``: "admin" | "member"), plumbed by the
+        # gateway's /_auth subrequest from the session row / JWT
+        # claim. Drives the admin gates on operator surfaces (e.g.
+        # the Skills Inbox actions). Read on THIS branch only: the
+        # header is trustworthy exactly when the request came through
+        # the gateway — enforced by the X-Gateway-Secret check above
+        # when configured, and the gateway overwrites any
+        # client-supplied X-Org-Role via proxy_set_header. (Without a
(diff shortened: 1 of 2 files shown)


ran-taig:
@claude

ran-taig:
@erni-a

erni-a (approved):

PR #585, opened 2026-07-20
ci: fix Claude review retrigger checkout for fork PRs

Eldad-Caura:
## Problem

The `claude-retrigger` job (fires when someone comments `@claude` on a PR) checked out the PR code with `ref: <PR head branch>` against the **base** repo. A fork PR's head branch only exists on the contributor's fork, not on `caura-ai/caura-memclaw`, so `actions/checkout` couldn't resolve it and the job exited 1 — **no review verdict was ever posted** for external fork PRs. It only worked when a fork's head branch happened to be named `main` (which exists in the base repo). Tracked as `.github#111`.

## Fix

The shared review script (`.github/scripts/claude_pr_review.sh`) fetches the PR diff via the **GitHub API**, not from the working tree — so this job never actually needs the PR's code checked out.

- Check out the **base repo's default branch** (trusted) instead of the fork's branch, and drop the now-unused `head_ref` lookup.
- **Security:** this job runs on `issue_comment`, which — unlike fork `pull_request` events — **does** receive repo secrets (`ANTHROPIC_API_KEY`, a write token). Checking out base code (not fork code) ensures a malicious fork PR can't have its code executed with our secrets when an org member triggers a review (a "pwn request" avoided).

## Hardening

- Gate the job on `author_association` (`MEMBER`/`OWNER`) at the **job level**, so the runner never even starts for a non-member's `@claude` comment. The existing step-level org-membership check stays as defense-in-depth (both read the same `author_association` field, so they can't drift).

## Testing

- YAML validated locally.
- Full end-to-end can only be confirmed after merge, since `issue_comment` workflows run from the **default branch** version of the workflow file.

### DIFF
--- a/.github/workflows/claude_code_review.yml
+++ b/.github/workflows/claude_code_review.yml
@@ -215,9 +215,15 @@ jobs:
 
   # Runs on @claude comment — only for public org members (MEMBER or OWNER).
   claude-retrigger:
+    # Hardening: gate at the job level on author_association so the runner never
+    # even starts for a non-member's @claude comment. The step-level "Check
+    # commenter is org member" below still re-validates and sets the
+    # ``authorized`` output for defense-in-depth — both read the same
+    # ``author_association`` field, so the two conditions cannot disagree.
     if: >-
       github.event_name == 'issue_comment'
       && contains(github.event.comment.body, '@claude')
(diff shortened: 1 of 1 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #586, opened 2026-07-20
CAURA-703 feat(write): record memory_type provenance in metadata

eyal-bl:
## What

Adds a `metadata.memory_type_agent_set` boolean recording **who chose a memory's type**:
- `true` — the calling agent supplied `memory_type` explicitly.
- `false` — the enrichment classifier, the default fallback, or the ingest extractor chose it.

This was previously unrecoverable: nothing persisted the distinction (the `agent_provided_fields` snapshot lives only in the transient enrich event and is discarded). It's set **synchronously at write time** — **no worker change, no DB migration** (plain JSONB key).

## Why

We needed to answer, per memory, "did the agent hand-pick this type or did the system?" — e.g. to audit classification quality and to tell agent-authored types apart from auto-classified ones. Binary was chosen over a multi-value source because it's fully known at write time (never changes), so it avoids any async/worker coupling; the finer classifier-vs-heuristic-vs-default split remains derivable from `metadata.llm_ms`.

## Changes

- **`merge_enrichment_fields`** (single write, fast + strong): capture the flag from the raw agent value **before** the CAURA-701 deprecation demotion, so an agent who supplied a now-deprecated type (`semantic` → coerced to `fact`) still counts as agent-set.
- **`create_memories_bulk`**: `metadata.setdefault("memory_type_agent_set", item.memory_type is not None)` — honours an ingest pre-stamp, else infers from the item.
- **`ingest_service`**: pre-stamp `False` — ingested types come from the extraction LLM's `suggested_type`, not the calling agent.

## Not in scope
- Historical backfill (unrecoverable; absent key = legacy/unknown).
- The bulk/ingest `semantic → fact` demotion gap (separate PR, CAURA-702).

## Tests
- Merge-step unit tests: agent-set wins & flagged `true`; classifier-fill → `false`; deprecated-demote still `true`; default fallback → `false`.
- Ingest: each committed fact is stamped `memory_type_agent_set = false`.

### DIFF
--- a/core-api/src/core_api/pipeline/steps/write/merge_enrichment_fields.py
+++ b/core-api/src/core_api/pipeline/steps/write/merge_enrichment_fields.py
@@ -26,6 +26,15 @@ class MergeEnrichmentFields:
         ts_valid_start = data.ts_valid_start
         ts_valid_end = data.ts_valid_end
 
+        # CAURA-703: record whether the CALLER chose the type, vs the enrichment
+        # classifier / default fallback filling it. Captured from the raw agent
+        # value BEFORE the deprecation demotion below, so an agent who supplied a
+        # now-deprecated type (e.g. "semantic", coerced to fact) still counts as
+        # agent-set. Fully determined at write time — the async enrichment worker
+        # never overrides memory_type when the agent set it (``agent_provided_fields``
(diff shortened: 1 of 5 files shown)


eyal-bl:
@claude

ran-taig (approved):

PR #587, opened 2026-07-20
feat(skills-inbox): scan findings on the card + truncated page flag (Phase B backend)

ran-taig:
## Summary

Backend half of the Skills Inbox Phase B polish (pairs with an enterprise dashboard PR). Two **additive** list-contract changes:

- **`sentinel_scan.findings`** — up to 20 well-formed `{code, severity, message, fatal, locator?}` entries mapped from `data.scan.findings`, so the UI can show *why* a scan tripped (and disable Approve with a reason when `critical_count > 0`). Malformed entries are skipped without consuming a capped slot; the full list stays on the doc.
- **`truncated`** top-level response flag — `true` when more staged cards exist than the page returned (effective limit `min(limit, inbox_max_pending, 200)` cut the list, or the oversample window saturated). Without it, a page of exactly N items is indistinguishable from a capped one, so no client-side "showing first N" heuristic can be accurate.

Both changes are backward-compatible (new fields only); the dashboard feature-detects them, so deploy order doesn't matter.

## Testing

- Golden Forge-card contract updated (finding entry now rides the golden doc's `warn: 1`).
- New: findings cap at 20 with malformed-entry skip (the test caught a slice-before-filter bug where junk consumed a slot — fixed), legacy scan blocks without a `findings` key, and `truncated` across cap-cut / oversample-saturated / partial-page cases.
- 60 route tests green; ruff + format clean; CI-invocation mypy clean for the touched file.
- `docs/skills-inbox-api.md` updated (response example + field notes).

### DIFF
--- a/core-api/src/core_api/routes/skills_inbox.py
+++ b/core-api/src/core_api/routes/skills_inbox.py
@@ -121,18 +121,40 @@ def _require_inbox_admin(auth: AuthContext) -> None:
 # ── Pydantic shapes ────────────────────────────────────────────────
 
 
+# Cap on findings surfaced per card. A pathological doc can carry an
+# unbounded findings list; the card only needs enough for an operator
+# to see WHY a scan tripped — the full list stays on ``data.scan``.
+_MAX_CARD_FINDINGS = 20
+
+
+class ScanFindingOut(BaseModel):
+    """One Sentinel finding, shaped for the card UI: enough to render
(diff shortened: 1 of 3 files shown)


ran-taig:
@claude

ran-taig:
@claude

ran-taig:
@erni-a

erni-a (approved):

PR #610, opened 2026-07-21
ci: type-check + build the plugin in CI

Eldad-Caura:
## What
Adds a `cd plugin && npm ci && npm run build` (tsc) step to the CI job.

## Why
The plugin currently has **no type-check gate in CI**:
- The only plugin step is `npm run check:version` (version-sync — runs no compiler).
- The real TS `Build + test` job lives in `client-typescript-ci.yml`, which is **path-scoped to the client SDK, not `/plugin`**.

So a `/plugin`-only PR can be fully green while `tsc` would fail. This surfaced when Dependabot's **typescript 5→7** bump (#547) would have broken `main` with ~47 `TS2591` errors despite a green checkmark. This step closes that blind spot (and runs `npm ci` on Linux, giving us lockfile validation too).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,7 +35,7 @@ jobs:
         with:
           persist-credentials: false
 
-      - name: Set up Node.js (for plugin version-sync check)
+      - name: Set up Node.js (for plugin version-sync + type-check)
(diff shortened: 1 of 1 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #614, opened 2026-07-21
fix(logging): correlate structlog output with Datadog APM traces

Eldad-Caura:
## Problem

`DD_LOGS_INJECTION=true` is set on the deployed services, but ddtrace's log injection patches the **stdlib** `logging.LogRecord` only. Our structlog pipeline uses `PrintLoggerFactory` (writes straight to stdout, bypassing stdlib), so injection never fires — `dd.trace_id` / `dd.span_id` never reach the emitted JSON, and Datadog can't correlate a log line to its APM trace. Surfaced during the caura-test-automation APM review (caura-test-automation#208) and deferred there as a pre-existing, fleet-wide issue.

## Fix

Add a `_add_dd_trace_context` structlog processor that reads the active span directly via ddtrace's public `tracer.get_log_correlation_context()` helper (verified against ddtrace 4.11.1 — its keys are already `dd.trace_id`/`dd.span_id`) and injects the IDs into the event dict. Placed in `_base_processors()` right after `merge_contextvars`, so it runs on **both** the native structlog chain and the stdlib-bridge `foreign_pre_chain`.

- **Guarded import:** `ddtrace` is the optional `datadog` extra, so it's imported `try/except ImportError` at module top; the processor no-ops when absent (OSS / on-prem / local) — those imports and log calls are unaffected.
- **No-op without an active span** (`trace_id == "0"`), so ambient (non-request) logs aren't stamped with a null trace.
- `DD_LOGS_INJECTION` stays `true` — it still covers third-party libs that log via stdlib directly.

## Tests
Three new cases in `tests/test_logging.py` (no ddtrace → no-op; active span → IDs injected; no active span → skipped). Full file: **35 passed** locally. `ruff format --check --isolated` (the vendored-file gate) passes.

## Rollout
- This file is **vendored `policy=identical`** into `caura-memclaw-enterprise` — the re-vendor bot syncs it after merge (no manual enterprise PR).
- `caura-ops` and `caura-test-automation` carry their own copies of this config; they get the same fix in their own repos (separate PRs).
- **Verify after deploy:** a log line emitted inside a traced request should show `dd.trace_id` matching the trace in APM. If Datadog doesn't auto-link, the remaining step is a Trace ID Remapper on the log pipeline (DD-side config, not code).

### DIFF
--- a/common/structlog_config.py
+++ b/common/structlog_config.py
@@ -38,6 +38,16 @@ from typing import Any
 
 import structlog
 
+# ddtrace is an optional dependency — only the `datadog` extra (SaaS + APM
+# builds) installs it. Import it guarded so OSS / on-prem / local, which never
+# install it, keep importing this module. `_add_dd_trace_context` no-ops when
+# this is None. Top-level (not a deferred import) so the name is patchable in
+# tests and the dependency is visible at module scope.
+try:
+    from ddtrace import tracer as _dd_tracer
+except ImportError:  # pragma: no cover - exercised only in non-datadog installs
+    _dd_tracer = None
+
 # structlog's log-level method names → GCP Cloud Logging's severity enum.
 # Severities beyond ERROR (CRITICAL/ALERT/EMERGENCY) aren't emitted by the
 # standard log-level methods; use `logger.critical(...)` for CRITICAL, or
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@claude please review

Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

Eldad-Caura:
Applied the fix exactly as suggested: both `dd.trace_id` and `dd.span_id` now use `.get(..., "0")` and I only inject when **both** are present and non-zero — so a partial context can never KeyError out of the processor or stamp an uncorrelatable trace-without-span. Added a `test_add_dd_trace_context_skips_partial_context` case for that path (36 tests pass; `ruff format --check --isolated` green).

@claude please re-review.

Eldad-Caura:
@erni-a ready for approval — log↔trace correlation processor for the OSS structlog config. Re-review came back clean after the defensive-`.get` fix; CI + dco green. On merge, the re-vendor bot syncs the identical file into caura-memclaw-enterprise.

erni-a (approved):

PR #615, opened 2026-07-21
fix(common): guard ddtrace call in the structlog processor

Eldad-Caura:
## What
Wraps the `_dd_tracer.get_log_correlation_context()` call in `common/structlog_config.py` (`_add_dd_trace_context`) in a `try/except`, returning the unmodified `event_dict` on failure.

## Why
This processor runs on **every** log call. If ddtrace raises (tracer shutdown, forked-process re-init, internal bug), the exception currently propagates through the structlog chain — dropping the very log line that would explain the failure, and in FastAPI potentially surfacing as an unhandled 500. Logging infrastructure must be resilient to a tracer-side fault.

Surfaced by the `caura-memclaw-enterprise` `claude-review` when the `common/` re-vendor brought this file over (identical-policy vendored file — fixing it here at the OSS source so the fix re-vendors downstream).

Verified locally: a tracer stub that raises no longer breaks the chain (returns `event_dict` unmodified); `ruff check`/`format` clean.

### DIFF
--- a/common/structlog_config.py
+++ b/common/structlog_config.py
@@ -48,6 +48,11 @@ try:
 except ImportError:  # pragma: no cover - exercised only in non-datadog installs
     _dd_tracer = None
 
+# Flips true after the first ddtrace correlation failure so we warn once (to
+# stderr, never through structlog — that would recurse), keeping a persistent
+# tracer fault visible without spamming every log call.
+_dd_tracer_warned = False
+
 # structlog's log-level method names → GCP Cloud Logging's severity enum.
 # Severities beyond ERROR (CRITICAL/ALERT/EMERGENCY) aren't emitted by the
 # standard log-level methods; use `logger.critical(...)` for CRITICAL, or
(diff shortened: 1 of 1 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@claude

Eldad-Caura:
@claude

Eldad-Caura:
Shipping as-is. The remaining review findings are Low/cosmetic and consciously accepted: the warn-once flag's thread-safety race is benign — worst case a few duplicate **stderr** diagnostics if multiple threads hit the *first* tracer failure at the same instant, after which `_dd_tracer = None` self-disables the path (global assignment is atomic under the GIL, so no torn state). Adding a `threading.Lock` to a per-log-call processor to dedupe a once-ever diagnostic would be over-engineering. The guard itself (never crash the chain, surface the fault, self-disable) is correct and complete. CI + dco green.

Eldad-Caura:
@erni-a

erni-a (approved):

PR #616, opened 2026-07-21
fix(logging): make dd trace-context processor resilient to tracer errors

Eldad-Caura:
Follow-up hardening to #614 (flagged on the enterprise re-vendor #843 review).

**What:** wrap `_dd_tracer.get_log_correlation_context()` in `try/except` inside `_add_dd_trace_context`, failing open to an un-annotated log.

**Why:** this processor runs on **every** log call via `_base_processors()`. ddtrace's helper is safe in normal operation, but a tracer-side raise (shutdown, forked-process re-init, internal ddtrace bug) would otherwise propagate out of the structlog chain — dropping the log line (or surfacing as a 500 in FastAPI), masking the very error worth seeing. Logging infra must be resilient.

Adds a test with a raising-tracer double. `ruff format --check --isolated` + 37 tests green. Re-vendors to enterprise on merge.

### DIFF
--- a/common/structlog_config.py
+++ b/common/structlog_config.py
@@ -304,7 +304,15 @@ def _add_dd_trace_context(
     """
     if _dd_tracer is None:
         return event_dict
-    ctx = _dd_tracer.get_log_correlation_context()
+    # This processor runs on EVERY log call. ddtrace's helper is safe in normal
+    # operation, but a tracer-side raise (shutdown, forked-process re-init, an
+    # internal ddtrace bug) must never propagate out of the logging chain — it
+    # would drop the log line (or surface as a 500 in FastAPI), masking the very
+    # error worth seeing. Fail open to an un-annotated log.
+    try:
+        ctx = _dd_tracer.get_log_correlation_context()
+    except Exception:
+        return event_dict
     # "0" is ddtrace's sentinel for "no active span". Read both keys defensively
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
On the Low (silent exception suppression): declining, with rationale. The fail-open behavior is the important, correct part and it's in place. The one-shot stderr warning is a reasonable nice-to-have, but two of the three fleet reviewers didn't flag it — caura-ops's review called the silent discard `a defensible choice for a per-log-call processor` — and adding a mutable module-level flag to a hot-path logging processor for a rare-within-rare *persistent* tracer failure is disproportionate. A persistent failure is also detectable through other channels (ddtrace's own error logging; absent trace correlation in DD). Keeping the four structlog copies byte-consistent is worth more here than the marginal diagnostic.

Easy to add across all four if we'd rather have it — happy to, just say so.

@erni-a ready for approval — try/except fail-open hardening; CI + dco green. Re-vendors to enterprise on merge.

erni-a (approved):

Eldad-Caura:
Closing as superseded by #615, which landed the same guard in a more complete form — try/except fail-open **plus** a one-shot stderr warning and disabling the tracer after a persistent failure (so the `_dd_tracer is None` guard short-circuits future calls). #615 is already on main; this PR only added the bare try/except. No action needed here.

PR #617, opened 2026-07-22
fix(mcp): allow own-fleet list/stats reads at trust 1

ran-taig:
## Problem

A client reported that `elisheva-claude` (trust 1, RND fleet) was rejected with `FORBIDDEN: trust_level=1 < required 2` when trying to see other RND agents' memories.

Per the product spec, **trust level 1 = read/write within its own fleet**; level 2 *adds* cross-fleet reads. But `memclaw_list` and `memclaw_stats` gated **every** non-agent scope at trust ≥ 2:

```python
min_level = 1 if scope == "agent" else 2
```

This charged a **same-fleet** read as if it were **cross-fleet**, contradicting the spec. It also diverged from the platform's other read surfaces, which already implement the spec:
- `memclaw_recall` — auto-scopes a constrained agent to its own fleet, gates only cross-fleet via `enforce_fleet_read` ([mcp_server.py:678](../blob/main/core-api/src/core_api/mcp_server.py))
- REST `GET /memories` + `POST /search` — allow own-fleet at any trust, require L2 only for a different fleet

## Fix

Resolve the trust floor by the **target fleet**, not the scope string (`_resolve_read_fleet_gate`, shared by both tools):

| scope | target | min trust |
|---|---|---|
| `agent` | own memories | 1 |
| `fleet` | own fleet | 1 |
| `fleet` | different fleet | 2 |
| `all` | tenant-wide | 2 |

**Security detail:** a constrained caller (trust < 2) that omits `fleet_id` on `scope='fleet'` is **pinned to its home fleet**. `memory_list_by_filters` / `memory_stats_breakdown` only apply the fleet filter when `fleet_id` is set, and their visibility predicate returns `scope_team` rows across *all* fleets — so without the pin, `scope='fleet'` with no `fleet_id` would fan out to other fleets' shared rows at L1. This mirrors the existing `memclaw_recall` guard. Unknown agents soft-pass at L1 (read ergonomics), consistent with `require_trust` / `recall`. The `scope_agent` visibility predicate is unchanged, so no private rows are exposed.

## Deliberately out of scope

`insights` and `evolve` share the same flat gate but are **not** touched — investigation found each needs a separate change first:
- **`insights`** `scope='fleet'` read has **no `scope_agent` exclusion** — lowering it would let a trust-1 agent read peers' *private* rows via synthesis.
- **`evolve`** `scope='fleet'` **mutates peers' memories** (weights + generated rules), not a read.

## Verification

No local Postgres/Docker, so the DB-backed suite wasn't run here. Verified:
- `_resolve_read_fleet_gate` logic across all 7 scope/fleet/trust cases
- End-to-end handler behavior with stubs: own-fleet allowed + `fleet_id` pinned; cross-fleet blocked; `scope=agent`/`all` skip the lookup
- Baselines regenerated with the canonical producers (list/stats descriptions only); `tools/list` = 4885 cl100k tokens (ceiling 5200)
- New/updated unit tests: split `list` fleet test into own-allowed vs cross-blocked, 7 helper tests, `get_agent` stubs added to the two cross-tenant audit tests


ran-taig:
@claude

ran-taig:
@claude re-review please — addressed the DCO sign-off, fixed the Ruff formatter failure, and applied the requested `_resolve_read_fleet_gate` hardening (unregistered/fleet-less caller + explicit fleet_id now requires L2). Note on the trust-sync request: there is a single source of truth (`agents.trust_level`), read live/uncached by `require_trust` and written directly by `update_trust_level` — no separate trust store to write through to, so I documented that invariant at the mutation site rather than adding a redundant sync.

ran-taig:
@claude

ran-taig:
@claude follow-up: closed a second fleet-less gap in `_resolve_read_fleet_gate`. A **registered** trust-1 agent with a NULL `fleet_id` passing no `fleet_id` param previously returned `(1, None)` → unfiltered `scope_team` scan across all fleets. Now resolved in three cases: (a) has home fleet → pin; (b) registered but fleet-less → force L2 (`require_trust` rejects); (c) unregistered → soft-pass at L1. Added a regression test for case (b). Full suite + ruff green locally; CI re-running.

ran-taig:
@claude

ran-taig:
@claude

ran-taig:
@claude

ran-taig:
@claude

ran-taig:
@erni-a

erni-a (approved):

PR #619, opened 2026-07-24
fix(core-api): bulk write reports oversized items per-item (207), not whole-batch 422

Eldad-Caura:
## What
Make the core-api bulk memory-write endpoint (`POST /api/v1/memories/bulk`) report an oversized item as a **per-item `status="error"` (207 multi-status)** instead of rejecting the **entire batch with 422**.

## Why
`BulkMemoryItem.content` carried a schema-level `max_length=MAX_CONTENT_LENGTH` (10 000), so FastAPI validated the whole body before the handler ran and **one** oversized item 422'd **all** items. The broker↔cloud API-versioning RFC (`docs/design/broker-cloud-api-versioning.md`, enterprise repo) requires this to be additive-tolerant: valid items must still be written, the oversized one surfaced as a typed per-item error.

## Change
- Remove `max_length` from `BulkMemoryItem.content` **only** (`MemoryCreate.content` and `MemoryUpdate.content` keep it — single-write still hard-rejects). `min_length=1` stays (an empty item is a malformed request, not a per-item business error).
- Enforce the cap per-item in `create_memories_bulk`: an oversized item is excluded from embed/enrich/dedup/write and recorded as `BulkItemResult(status="error", error="content exceeds maximum length of 10000 characters")` — the **same shape** as the existing short-content and governance-drop paths. `_bulk_response` then yields 207 for a mix, 200 for all-valid (unchanged). Length is measured on the raw string, matching the old `max_length` semantics.

## Tests
- Mixed batch (1 oversized + 2 valid) → `created=2, errors=1`, HTTP **207**, oversized row `status="error"` with `id=null`.
- All-valid batch (incl. one at exactly 10 000 chars) → HTTP **200**.
- Schema now accepts oversized bulk content; `MemoryCreate` still rejects it (scope guard).

`ruff check`, `ruff format --check`, `mypy` clean; bulk test modules + a regression sweep of every `create_memories_bulk` caller pass locally.

Gate C / Phase 0; **T0.2 core-api side** — complements the broker content cap ([memclawd #92](https://github.com/caura-ai/memclawd/pull/92)) and the gateway body limit (enterprise #888).

### DIFF
--- a/core-api/src/core_api/schemas.py
+++ b/core-api/src/core_api/schemas.py
@@ -69,8 +69,22 @@ class BulkMemoryItem(BaseModel):
     """Single item in a bulk write request. tenant_id/fleet_id/agent_id inherited from parent."""
 
     memory_type: MemoryType | None = Field(default=None, description=MEMORY_TYPES_WRITE_DESCRIPTION)
-    content: str = Field(min_length=1, max_length=MAX_CONTENT_LENGTH)
-    weight: float | None = Field(default=None, ge=0.0, le=1.0)
+    # Additive-tolerance policy (broker↔cloud API-versioning RFC): a bulk write
+    # must NOT 422 the whole batch because one item has a bad field — the valid
+    # items must still be written. So schema-level constraints that would reject
+    # the ENTIRE request are deliberately dropped here (unlike single-write
+    # ``MemoryCreate`` / ``MemoryUpdate``, which keep them) and re-enforced PER
+    # ITEM in ``create_memories_bulk`` as a ``status="error"`` 207 row:
+    #   - content length → short_content_errors
+    #     (``< CRYSTALLIZER_SHORT_CONTENT_CHARS`` = 10, subsumes the old
+    #     ``min_length=1``) + oversized_content_errors (``> MAX_CONTENT_LENGTH``).
+    #   - weight range [0.0, 1.0] → weight_errors.
+    #   - status enum → status_errors.
+    # ``memory_type`` still uses its typed enum below (an invalid value 422s the
+    # whole batch); making it per-item needs an enum→str change with downstream
+    # impact, tracked as a follow-up.
+    content: str
+    weight: float | None = Field(default=None)
     source_uri: str | None = None
     run_id: str | None = None
     metadata: dict | None = None
@@ -82,7 +96,7 @@ class BulkMemoryItem(BaseModel):
     ts_valid_start: datetime | None = None
     ts_valid_end: datetime | None = None
     reference_datetime: datetime | None = None
-    status: str | None = Field(default=None, pattern=MEMORY_STATUSES_PATTERN)
+    status: str | None = Field(default=None)  # validated per-item (status_errors); see note above
 
 
 class BulkMemoryCreate(BaseModel):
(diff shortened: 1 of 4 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@claude addressed the empty-content asymmetry. Rather than add a separate `empty_content_errors` dict, I removed `min_length=1` from `BulkMemoryItem.content` entirely — an empty/whitespace item then falls into the **existing** `short_content_errors` path (`len(content.strip()) < CRYSTALLIZER_SHORT_CONTENT_CHARS`, which is 10 and thus subsumes empty), so it surfaces as a per-item 207 error just like oversized. A dedicated empty-dict would have been dead code alongside that check. Empty and oversized now behave identically at the batch level.

Added `test_empty_content_in_mixed_batch_returns_207`; `ruff check`/`mypy` clean on `core-api/src`; the oversized + empty + all-valid bulk tests pass locally. Single commit amended + force-pushed.

Eldad-Caura:
@erni-a ready for your approval. The auto-review's one Medium finding (empty content still 422'd the whole batch via `min_length=1`, inconsistent with the additive-tolerant contract) is fixed: I removed `min_length=1` so empty/whitespace items fall into the existing per-item short-content error path (`< 10` chars, which subsumes empty) — empty and oversized now both surface as 207 per-item errors. Added a regression test; `ruff`/`mypy` clean, bulk tests pass, CI green. Single commit.

(OSS auto-review skips on-open for my authorship per usual; the manual `@claude` review ran and its finding is addressed.)

erni-a (dismissed):

Eldad-Caura:
@erni-a please approve

erni-a (approved):

Eldad-Caura:
CI is green again — the previous red was a pre-existing schema test (`test_item_content_validation`) that asserted empty bulk content raises at the schema layer; with `min_length=1` removed that assertion no longer holds, so I updated it to assert empty is accepted at the schema and handled per-item (mirroring the oversized schema test). Full `test_bulk_write.py` + `test_bulk_atomicity.py` pass; CI run 30127499119 green. Still ready for your review, @erni-a.

erni-a (dismissed):

Eldad-Caura:
@claude extended per-item additive-tolerance to `weight` (out of [0.0,1.0]) and `status` (not a MEMORY_STATUSES lifecycle value) — both now 207 per-item errors instead of a whole-batch 422, same pattern as content. `memory_type` is deliberately kept as a typed enum (still 422s) and deferred to a follow-up (the enum→str change ripples downstream). Removed the two now-stale schema-rejection tests, added schema + integration tests for weight/status. core-api/src ruff+mypy clean; full test_bulk_write.py + test_bulk_atomicity.py pass (45).

Eldad-Caura:
@claude addressed: the per-item validation errors (short/oversized content, weight, status) are now **aggregated into one error row** per item — an item that violates several constraints reports them all (`"; ".join(...)`) instead of only the first, so no multi-round-trip. Governance drops stay a separate row (disjoint — governance only runs on items that passed input validation). Added `test_item_failing_multiple_validations_aggregates_errors`; core-api/src ruff+mypy clean, full bulk modules pass (46).

Eldad-Caura:
@erni-a pleae approve

erni-a (approved):

PR #620, opened 2026-07-24
ci(core-api): freeze v1 broker OpenAPI baseline + oasdiff gate (Gate C Phase 1)

Eldad-Caura:
## What
**Freeze the v1 broker↔cloud OpenAPI contract for core-api** and enforce it on every PR with an `oasdiff` breaking-change gate. Phase 1 (Deliverables A+B) of the broker API-versioning RFC (`docs/design/broker-cloud-api-versioning.md`, enterprise repo).

## Why
v1 was a *label*, not an enforced contract — breaking changes shipped inside `/v1` and silently broke installed brokers (the Phase-0 incident). This commits the broker-facing contract as an artifact and fails CI on any breaking change to it, so a break must ship a new `/vN`.

## Contents
- **`core-api/scripts/gen_broker_openapi.py`** — imports `core_api.app` offline, filters `paths` to the 4 broker endpoints (`POST /api/v1/memories/bulk`, `POST /api/v1/search`, `GET /api/v1/health`, `GET /api/v1/version`; a missing path is a hard error), keeps only the transitively-reachable schemas + referenced security schemes, **normalizes `info` to a frozen contract version (`v1`)** decoupled from the rolling package semver (so releases don't churn the baseline), and emits deterministic JSON. `--check` fails if the committed baseline is stale.
- **`core-api/openapi.broker.json`** — the frozen v1 baseline (4 paths, 14 schemas, `APIKeyHeader`).
- **`ci.yml`** — stale-baseline guard + `oasdiff breaking --fail-on ERR` vs `main`'s baseline (pinned `tufin/oasdiff:v1.26.0`; `--fail-on ERR` required — plain `oasdiff breaking` exits 0 even on breaks). First-merge bootstrap skips the diff when main has no baseline yet.
- **`core-api/scripts/README.md`** — regeneration + contract-bump rules.

## Validation
`--check` clean and deterministic (byte-identical across runs); **proof the gate bites**: a synthetic required field on `BulkMemoryCreate` → `oasdiff` reports `new-required-request-property`, exit 1 — then reverted. `ruff` clean on the script.

Regenerate after an intentional broker contract change: `python core-api/scripts/gen_broker_openapi.py`.

Gate C / Phase 1. Companion to the enterprise auth/audit gate (#889) and the Phase-0 fixes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,6 +73,51 @@ jobs:
           uv pip install -e core-storage-api/[dev] -e core-api/[dev]
           echo "$VIRTUAL_ENV/bin" >> $GITHUB_PATH
 
+      # --- Broker↔cloud API-versioning gate (Gate C / Phase 1, deliverables A+B) ---
+      # core-api/openapi.broker.json is the FROZEN v1 contract for the four
+      # broker-facing gateway endpoints (POST /api/v1/memories/bulk,
+      # POST /api/v1/search, GET /api/v1/health, GET /api/v1/version) that
+      # memclawd calls against Caura cloud. It is a subset of the full ~91-path
+      # OpenAPI surface, with info normalized to the frozen contract version so
+      # it changes ONLY when a broker endpoint's shape changes. Regenerate with
+      # `python core-api/scripts/gen_broker_openapi.py` and commit the result;
+      # breaking changes require a deliberate contract bump per the RFC
+      # (see core-api/scripts/README.md).
+      - name: Broker OpenAPI baseline is current (freeze v1)
+        # Regenerates the broker-contract subset from the live app and fails if
+        # the committed baseline is stale vs the code on this branch.
+        run: uv run python core-api/scripts/gen_broker_openapi.py --check
+        env:
+          # core_api is installed editable above; `.` puts the repo root on the
+          # path for the `common` package (mirrors the test/migration steps).
+          PYTHONPATH: core-api/src:core-storage-api/src:.
+
+      - name: Broker OpenAPI has no breaking change vs main
+        # oasdiff compares main's committed baseline (base) against this branch's
+        # baseline (revision) and fails on any BREAKING change (removed path or
+        # param, new required request field, narrowed type, ...). oasdiff is
+        # pinned to the tufin/oasdiff Docker image by tag; `--fail-on ERR` is
+        # required for a non-zero exit on breaking changes. On the very first
+        # merge main has no baseline yet — the gate detects that and skips.
+        run: |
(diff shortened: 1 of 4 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@claude

Eldad-Caura:
Fixed the Low finding (`--out` + `--check` now errors explicitly instead of silently ignoring `--out`).

On the Medium (baseline advertises server-reserved `memory_type`s `outcome`/`rule`/`insight`): the right fix is **#621**, which changes `BulkMemoryItem.memory_type` from the full `MemoryType` enum to `str` (additive-tolerance — unknown types become per-item 207s, matching content/weight/status). With that, this baseline regenerates to `memory_type: {"type": "string"}` — no reserved-type enum to mis-advertise. I deliberately did **not** put a conflicting `Literal[MEMORY_TYPES_WRITE]` change here.

Suggested order: merge **#621 first** (it's rebased onto main, clean, mergeable), then I'll rebase this PR and regenerate `openapi.broker.json` so the baseline reflects the `str` field — which resolves this Medium. @erni-a — flagging the sequencing; happy to do it the other way (Literal here, #621 supersedes) if you prefer.

erni-a (dismissed):

Eldad-Caura:
@claude rebased onto main after #621 merged and **regenerated the baseline** — `memory_type` in `openapi.broker.json` is now a plain `string` (`$ref: MemoryType` enum removed), so the contract no longer advertises the server-reserved types; that resolves the Medium. The Low finding (`--out`+`--check`) is fixed. `gen --check` passes locally (baseline current). Please re-verify.

Eldad-Caura:
@claude fixed: `_security_scheme_names` now collects each broker operation's **effective** security (its own `security`, else the spec default) instead of unconditionally unioning the global `security` block — so a scheme in the global block that every broker op overrides no longer leaks into the contract, and there's no risk of a dangling top-level security ref. Current spec is unaffected (global security is `None`; `bulk`/`search` declare `APIKeyHeader`, `health`/`version` are unauthenticated), so the regenerated baseline is byte-identical. `gen --check` + ruff clean.

Eldad-Caura:
@claude fixed the Low: `build_broker_spec` now uses an explicit top-level allowlist (`{"openapi": ...}` + info/paths/components) instead of the denylist, so a future top-level key from app.openapi() can't leak into the frozen baseline. Regenerated — baseline byte-identical; `gen --check` + ruff clean.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #621, opened 2026-07-24
fix(core-api): bulk unknown memory_type → per-item 207 (stacked on #619)

Eldad-Caura:
## What
Completes bulk **additive-tolerance** for `BulkMemoryItem`: an unknown `memory_type` on one item now surfaces as a per-item `status="error"` (207) row instead of 422-ing the whole batch. Follow-up to #619 (content/weight/status), which deferred `memory_type` because it required an enum→str change.

> **Stacked on #619** (`CAURA-000-coreapi-bulk-oversized-207`) — base is that branch so the diff shows only the memory_type delta. **Merge after #619.**

## Change
- `BulkMemoryItem.memory_type`: `MemoryType | None` → `str | None` (single-write `MemoryCreate`/`MemoryUpdate` keep the typed enum — they still hard-reject).
- `create_memories_bulk`: new `memory_type_errors` validating against the **full `MEMORY_TYPES`** vocabulary the enum accepted; it joins the aggregated per-item error row and is excluded from `valid_indices`.
- Valid types — including reserved/deprecated-but-known ones — flow downstream **unchanged** (deprecated folding, defaulting, storage): `MemoryType` is a `str`-Enum, so valid values compare and persist identically to before.

## Tests
- Integration: mixed batch with a bad `memory_type` **and a valid explicit one** → 207, bad row errored, valid items (explicit + default type) created.
- Schema: `BulkMemoryItem` now accepts an unknown `memory_type`; `MemoryCreate` still rejects it (scope guard).
- Full `test_bulk_write.py` + `test_bulk_atomicity.py` pass (47), plus the deprecated-type (`test_caura702_*`) and bulk-caller (ingest/mcp/broker-write) regression modules (65). `ruff`/`mypy` clean on `core-api/src`.

Gate C / Phase 0; T0.2 core-api. Closes the `memory_type` follow-up from #619.

### DIFF
--- a/core-api/src/core_api/schemas.py
+++ b/core-api/src/core_api/schemas.py
@@ -68,21 +68,22 @@ class MemoryCreate(BaseModel):
 class BulkMemoryItem(BaseModel):
     """Single item in a bulk write request. tenant_id/fleet_id/agent_id inherited from parent."""
 
-    memory_type: MemoryType | None = Field(default=None, description=MEMORY_TYPES_WRITE_DESCRIPTION)
     # Additive-tolerance policy (broker↔cloud API-versioning RFC): a bulk write
     # must NOT 422 the whole batch because one item has a bad field — the valid
     # items must still be written. So schema-level constraints that would reject
     # the ENTIRE request are deliberately dropped here (unlike single-write
     # ``MemoryCreate`` / ``MemoryUpdate``, which keep them) and re-enforced PER
-    # ITEM in ``create_memories_bulk`` as a ``status="error"`` 207 row:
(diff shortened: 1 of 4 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@claude rebased onto main after #619 merged — this branch now carries only the memory_type delta (schema `str` + per-item `memory_type_errors` + tests). Please re-verify.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #623, opened 2026-07-25
fix(interviewer): preserve exact figures in synthesis + document list-limit cap

ran-taig:
Two fixes surfaced by the Interviewer staging rollout rehearsal (OpenClaw push path, validated end-to-end on `memclaw.dev`).

## 1. Synthesis preserves exact figures + whole decisions (substantive)
The staging quality eval ran realistic work windows through the live **Gemini flash-lite** synthesis and scored the output against known ground truth. It was strong on type-mapping, faithfulness (0 hallucinations), and negative-outcome discipline — but had one consistent weakness:
- **dropped exact numbers** (`recall@10 0.87 vs 0.83, index -33%` → "better recall"), and
- **flattened compound/conditional decisions** (`migrate to X after the Y fix; keep Z as fallback` → "migrate to X").

Fix: strengthen `_REPORT_SCHEMA_INSTRUCTION` in `interview_service.py` to require preserving figures **verbatim** (metrics, thresholds, counts, durations, versions, names) and keeping each decision **whole** (conditions, sequencing, alternatives). **Prompt-only — no model change** (flash-lite stays; it's cheap, fast, and faithful, just needed steering). Adds a unit test pinning the new instructions.

## 2. Document the list-limit cap (minor)
`GET /api/v1/memories?limit=1000` returns **422** (cap is `MAX_LIST_LIMIT=500`), which is easy to mistake for an empty result if the error body isn't parsed. Document the cap + 422 behaviour in the OpenAPI `limit` param description on both list endpoints. **Behaviour unchanged** — the memories endpoint is deliberately strict (`le=`), unlike the clamp-style `skills_inbox` endpoint; I kept that choice and made it discoverable rather than silently changing it.

## Verification
`tests/test_api_interview.py` 22/22 pass (incl. the new `test_build_prompt_demands_verbatim_figures_and_whole_decisions`); ruff clean.

### DIFF
--- a/core-api/src/core_api/routes/memories.py
+++ b/core-api/src/core_api/routes/memories.py
@@ -200,7 +200,15 @@ async def list_memories(
     ),
     order: str = Query(default="desc", pattern=r"^(asc|desc)$"),
     offset: int = Query(default=0, ge=0),
-    limit: int = Query(default=DEFAULT_LIST_LIMIT, ge=1, le=MAX_LIST_LIMIT),
+    limit: int = Query(
+        default=DEFAULT_LIST_LIMIT,
+        ge=1,
+        le=MAX_LIST_LIMIT,
+        description=(
+            f"Items per page, 1-{MAX_LIST_LIMIT}. A value above {MAX_LIST_LIMIT} is rejected "
+            "with 422 (not silently clamped); page via `cursor`/`offset` for more."
+        ),
+    ),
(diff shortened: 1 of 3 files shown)


ran-taig:
@claude

ran-taig:
@erni-a

erni-a (approved):

PR #626, opened 2026-07-26
fix(memories): /memories/count returns 0 for enriched tenants — count the live status set

Eldad-Caura:
## The bug

`GET /memories/count` returns **0** for tenants that have live memories, while `GET /memories` lists them normally. Reported against tenant `beacon-4c8fca` (7 rows listed, count 0).

`memory_count_active` filtered on the literal status value:

```python
Memory.status == "active"
```

…but "active" is being used in two different senses. The public endpoint's docstring promises *liveness* ("Count **active (non-deleted)** memories"), while the SQL tests one enum value. `confirmed` is a **stronger** liveness state than `active`, yet it was excluded.

## Why this is the count query's bug, not the caller's

Every sibling query in the same service already defines live as a **set**:

```python
Memory.status.in_(("active", "confirmed", "pending"))
```

— `memory_find_semantic_duplicate`, `memory_find_entity_overlap_candidates`, `memory_find_rdf_conflicts`, `memory_find_similar_candidates`. `memory_count_active` was the lone outlier. `/memories/stats` is also already liveness-based (`deleted_at IS NULL` only), so `/count` and `/stats` contradicted each other on the same tenant.

And `confirmed`/`pending` are reachable on the **normal** write path, not just by explicit transition:

- the enrichment classifier assigns `confirmed` to outcome-shaped content and `pending` to task/plan/commitment content (`common/enrichment/service.py`)
- the crystallizer writes `status="confirmed"` directly (`crystallizer_service.py`)
- the enrichment prompt explicitly offers all three (`_prompts.py`)

So any tenant whose writes get enriched drifts toward `count=0` while its list fills up.

## The fix

Adds `LIVE_MEMORY_STATUSES = ("active", "confirmed", "pending")` to `common/constants.py` — the module that already exists for "DB-query constants shared between core-api and core-storage-api" — as the single definition of liveness.

`memory_count_active(status=None)` counts that set; an explicit `status` counts exactly one. Threaded through:

| Layer | Change |
|---|---|
| `memory_count_active` | `status` param; default = live set |
| storage `/count`, `/count-active` | `status` query param, passed through |
| `storage_client.count_active` | forwards `status` |
| public `/memories/count` | `status` query param, **pattern-validated** so a typo 422s instead of silently counting 0 |

### Also: `/embedding-coverage` could report a negative percentage

Same mechanism, second-order. `memory_find_missing_embeddings` (its numerator) filtered `deleted_at IS NULL` with **no** status filter, while `memory_count_active` (its denominator) counted only `active`:

```
coverage_pct = (total - len(missing)) / total * 100
```

1 `active` + 6 `confirmed` rows all missing embeddings → `(1 - 7) / 1 * 100` = **−600%**. The reporting tenant only dodged it because `total=0` tripped the zero-guard. The numerator is now scoped to the same live set, so `missing ≤ total` holds by construction.

## Behaviour change

`/memories/count` and `/memories/count-active` now report the **live set** by default. Callers who want the old narrow number pass `status=active`. `/count-active` keeps its name but its default widens — flagging explicitly since the name now reads narrower than it behaves.

## Testing

Five new tests. The key one **fails on pre-fix code with exactly the reported symptom** — verified by reverting the filter and re-running (`assert {'count': 1} == {'count': 3}`):

- `test_memory_count_includes_enriched_statuses` — 1 active + 1 confirmed + 1 pending → count 3, and the count now **agrees with the list length**, the exact discrepancy that surfaced this
- `test_memory_count_status_param_narrows` — `?status=active` → 1, `?status=confirmed` → 1
- `test_memory_count_excludes_shelved_statuses` — `archived` drops out (liveness, not just non-deleted)
- `test_memory_count_rejects_invalid_status` — 422 on a bad status
- `test_embedding_coverage_spans_the_live_status_set` — denominator counts the confirmed row; asserts `missing ≤ total` and `coverage_pct ≥ 0`. This endpoint previously had **no** tests.

Gates:

- CI suite (`tests/`): **3954 passed**, 3 skipped, 1 xfailed, 3 xpassed — zero failures
- `ruff check` + `ruff format --check` (core-api, core-storage-api) clean
- `mypy` clean for both configs (188 + 73 files)

> `core-storage-api/tests/` is not run by CI and has 31 failures on `origin/main` in a local environment. Verified my branch is identical: **31 failed / 69 passed** clean vs **31 failed / 70 passed** here — the +1 is the new test. No regressions introduced; the pre-existing rot is untouched and out of scope.

### DIFF
--- a/common/constants.py
+++ b/common/constants.py
@@ -1,5 +1,17 @@
 """DB-query constants shared between core-api and core-storage-api."""
 
+# ── Memory liveness ──
+# The statuses that mean "this memory is live". Broader than the literal
+# ``active`` value: enrichment promotes writes past ``active`` on the normal
+# path (the classifier assigns ``confirmed`` to outcome-shaped content and
+# ``pending`` to task/plan/commitment content, and the crystallizer writes
+# ``confirmed`` directly), so a liveness check that tests ``status ==
+# "active"`` silently drops enriched rows. Excludes the terminal/shelved
+# states (``cancelled``, ``outdated``, ``conflicted``, ``archived``,
+# ``deleted``). Callers that genuinely want one exact status filter on it
+# explicitly instead.
+LIVE_MEMORY_STATUSES = ("active", "confirmed", "pending")
+
 # ── Embeddings ──
 # Native dim of the default embedder (BAAI/bge-m3, see local-embedder docs).
 # Schema upgrade lives in alembic migration 012_vector_dim_1024.py — keep
--- a/core-api/src/core_api/clients/storage_client.py
+++ b/core-api/src/core_api/clients/storage_client.py
@@ -727,10 +727,18 @@ class CoreStorageClient:
         result = await self._post("/preview/tenant-counts", {"tenant_id": tenant_id})
         return result.get("counts", {})  # type: ignore[union-attr,return-value]
 
-    async def count_active(self, tenant_id: str, fleet_id: str | None = None) -> int:
+    async def count_active(
+        self,
+        tenant_id: str,
+        fleet_id: str | None = None,
+        status: str | None = None,
+    ) -> int:
+        """Live-memory count; ``status`` narrows to one exact status."""
         params: dict[str, Any] = {"tenant_id": tenant_id}
         if fleet_id is not None:
             params["fleet_id"] = fleet_id
+        if status is not None:
+            params["status"] = status
         result = await self._get("/memories/count-active", **params)
         return (result or {}).get("count", 0)
 
--- a/core-api/src/core_api/routes/memories.py
+++ b/core-api/src/core_api/routes/memories.py
@@ -30,6 +30,7 @@ from core_api.config import settings as app_settings
 from core_api.constants import (
     DEFAULT_LIST_LIMIT,
     MAX_LIST_LIMIT,
+    MEMORY_STATUSES_PATTERN,
 )
 from core_api.middleware.idempotency import (
     IDEMPOTENCY_HEADER,
@@ -361,9 +362,16 @@ async def memory_stats(
 async def memory_count(
     tenant_id: str | None = Query(default=None),
     fleet_id: str | None = Query(default=None),
+    status: str | None = Query(default=None, pattern=MEMORY_STATUSES_PATTERN),
     auth: AuthContext = Depends(get_auth_context),
 ):
-    """Count active (non-deleted) memories for a tenant, optionally a fleet.
+    """Count live (non-deleted) memories for a tenant, optionally a fleet.
+
+    "Live" is the whole ``LIVE_MEMORY_STATUSES`` set (``active``,
+    ``confirmed``, ``pending``) — NOT just the literal ``active`` value, which
+    is what this counted before and why tenants whose rows enrichment had
+    promoted to ``confirmed``/``pending`` saw 0 while ``GET /memories`` listed
+    them. Pass ``status`` to count exactly one status instead.
 
     A lightweight operational primitive — for type/agent/status breakdowns use
     ``GET /memories/stats``. Declared BEFORE ``/memories/{memory_id}`` so the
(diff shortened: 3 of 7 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
Verified and applied the Low finding.

**Confirmed the premise** before touching it: `memory_count_all(self)` genuinely takes no arguments, and `tenant_id: str` carries no `min_length`, so `?tenant_id=` passes validation into that branch (an *omitted* param is a 422). My diff added a docstring that promised `status` narrowing unconditionally, which is false on that one path — so the inconsistency is real and mine.

**Fix:** documented the actual contract rather than changing behaviour. The empty-tenant branch predates this endpoint's filters and routes to the whole-corpus counter; making it 422 or growing a second filtered path would be a behavioural change to an unrelated code path, which doesn't belong in a bug-fix PR. The docstring now states plainly that `fleet_id` and `status` do not apply there, so it's no longer silent.

Gates re-run after the amend: `ruff check` + `ruff format --check` clean, `mypy` clean (73 files), count tests 5/5.

Out of scope, worth its own ticket: that branch reveals a pre-existing wrinkle — because `tenant_id: str` is required, calling `/count` with the param fully omitted 422s, so the whole-corpus path is only reachable via the empty-string form. `core-storage-api/tests/test_integration.py::test_public_counters_exclude_soft_deleted` expects the omitted-param form and fails on `origin/main` today (that suite isn't CI-gated — 31 failures on a clean main, unchanged by this PR).

Eldad-Caura:

Claude review clean — one Low finding (docstring contract on the empty-tenant branch), verified and handled. All checks green on the amended commit: CI ✅ · CodeQL python + js/ts ✅ · Security Scan ✅ · image-version ✅ · DCO ✅ · pre-checks ✅. Single signed commit.

Heads-up for the reviewer: this is a **behaviour change** — `/memories/count` and `/memories/count-active` now report the live status set (`active`, `confirmed`, `pending`) rather than literal `status='active'`. Callers wanting the old narrow number pass `status=active`.

erni-a (approved):

Eldad-Caura:
**Blast-radius audit of the behaviour change** (asked before merge — posting the findings).

Searched every consumer repo (`memclaw`, `memclawd`, `caura-ops`, `memclaw-enterprise`) for `/memories/count`, `/count-active`, `count_active`, and `/embedding-coverage`.

**Not affected:**
- **Broker (memclawd)** — its complete cloud surface is `/api/v1/memories`, `/memories/bulk`, `/memories/{id}`, `/memories/{id}/links`, `/search`, plus auth/installs/health/audit-log. It never calls a count endpoint.
- **MCP (`memclaw_stats`)** — routes to `compute_memory_stats` → storage `memory_stats_breakdown` (GROUPING SETS), a separate query path that already filtered on `deleted_at IS NULL` only. MCP stats were already liveness-correct; this change actually brings `/memories/count` *into agreement* with them rather than perturbing them.
- **caura-ops** — reads the MemClaw replica directly via SQLAlchemy models for its analytics; no calls to these endpoints.
- **memclaw-enterprise** — its `count_active_*` hits are `count_active_api_credentials`, unrelated.

**One real internal consumer — worth a reviewer's eye:**

`core-api/src/core_api/services/lifecycle_audit.py:88`
```python
active = await self._storage.count_active(org_id, fleet_id)
if active <= _CRYSTALLIZE_MIN_ACTIVE_MEMORIES:  # 1000
    return 0
```

This gates **auto-crystallization** on corpus size. Because the count was under-reporting, orgs whose live corpus exceeded 1000 but whose literal-`active` count did not were being skipped. After this fix their count rises and the gate opens, so **auto-crystallization (LLM work) can start running for orgs that were previously silently skipped**.

I believe that is the gate behaving as intended — it was meant to measure corpus size and was measuring the wrong thing — but it is a real cost-side consequence of the fix rather than a pure read-path change, so it should be a conscious call, not a surprise. Happy to add an explicit `status="active"` at that call site to freeze the old gating behaviour if you'd prefer to decouple the two changes.

PR #624, opened 2026-07-26
feat(agent-digest): stop over-filtering agents — floor, episode-fallback, subagent rollup

ran-taig:
Grounded in the eToro over-filtering finding: active agents were invisible in the digest because their work is **enrichment-classified as `episode`** (which we excluded wholesale) and/or **split across many subagents** (e.g. `cmoclaw` + ~40 `cmoclaw-*`). Core-api side (PR 1 of 2; UI PR follows).

## Fixes
1. **Durable floor 3→1** (default; removed the day/week special-case) — a single decision/fact is worth surfacing. Configurable via `min_activity_threshold`.
2. **Durable-first, episode-fallback** (`report_corpus` split into `passes_noise_filter` + `is_cohesive`). `generate_for_org` **reuses the already-fetched rows** (no extra query): an agent with `< floor` durable rows is summarized from its non-noise **episodes** (ranked by recall) with status `activity`. Same LLM budget; in-memory partition.
3. **Count-only "listed" tail** — active families past `top_n` (or below the summarize floors) get a no-LLM row, so no active agent is invisible.
4. **Subagent rollup** — `resolve_parent` derives the parent from id conventions (`agent:<p>:subagent:<uuid>` and `<parent>-<task>`, longest-prefix, transitive; `owner_ref` is unpopulated in prod). Subagents roll into the parent family; the parent row carries `source_count` = family total + a `subagents[]` list (`{agent_id, fleet_id, source_count}`) for the collapsed UI + child deeplinks. **Migration 032** adds the `subagents` JSONB column.

`source_count` now means the family's **meaningful-writes total** (the bar value). New statuses: `activity`, `listed`.

## Config (`agent_digest`)
`min_activity_threshold=1`, `event_floor=1`, `listed_max=25`.

## Tests
- `report_corpus`: `resolve_parent` (both conventions, boundary guard, transitive) + filter split.
- generator: floor→summarize/list, episode-only→`activity`, cmoclaw rollup (both id forms), active-beyond-top_n→`listed`.
- storage: `subagents` round-trip + empty default; migration head→032.
- 23 core-api + 13 storage tests green; ruff + mypy clean (both packages).

Cost: no new queries; LLM spend unchanged/lower (one call per family, `top_n`-bounded). Depends on nothing; the enterprise UI PR (collapsed subagent group + `activity`/`listed` rendering + child deeplinks) follows.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #630, opened 2026-07-26
fix(storage): provision test DB via migrations; accept omitted tenant_id on /count

Eldad-Caura:
Follow-up to #626 / #628. Fixes the two issues found there, plus the root cause behind most of the `core-storage-api/tests/` failures.

## 1. Test DB provisioning — the big one

`core-storage-api/tests/conftest.py` built the schema with `Base.metadata.create_all` over a **hand-maintained list of model imports**. Production provisions by running the Alembic chain (`init_database()`, called from the FastAPI lifespan hook). The two had drifted in both directions:

- **Tables with no SQLAlchemy model were never created.** Migration `019_tenant_suppression` owns `tenant_suppression`, which has no model — so every suppression test failed on a pristine database, on any machine.
- **`create_all` only CREATEs, never ALTERs.** A database carried over from an earlier revision keeps its old columns, so a migration that adds one (e.g. `agent_activity_digests.subagents`, added recently by #624) leaves the suite red until someone rebuilds the DB by hand.

Those two pull in opposite directions, which is why the failure count depended on *which* stale database you happened to have — and why "is this real rot or just my local DB?" had no stable answer.

Calling `init_database()` makes fresh and reused databases provision identically. Measured on this tree:

| Provisioning | Result |
|---|---|
| Pristine DB, `create_all` (before) | 35 failed / 65 passed |
| Carried-over dev DB, `create_all` (before) | 31 failed / 69 passed |
| **Any DB, migrations (after)** | **19 failed / 81 passed** |

All 12–16 schema-related failures are gone, and the number is now deterministic.

## 2. `/count` rejected an omitted `tenant_id`

`tenant_id` was annotated as a bare `str` — required — so `?tenant_id=` (empty) reached the whole-corpus branch but omitting the param entirely returned 422.

**The branch is not dead code**, which is worth stating clearly: `core-api`'s public-stats caller reaches it via `count_all(tenant_id="")` (`core-api/src/core_api/routes/stats.py`). Deleting it would have broken `/stats`. What didn't work was the *omitted* spelling the branch and its test were both written for.

Widening the annotation to `str | None = None` accepts both. The empty-string form is byte-for-byte unchanged, so the stats caller is unaffected.

This also un-breaks `test_public_counters_exclude_soft_deleted`, which had been failing with `KeyError: 'count'` (reading `["count"]` off a 422 body).

## Testing

New `test_count_whole_corpus_accepts_omitted_and_empty_tenant` pins **both** spellings — the empty-string one because it is load-bearing in production, the omitted one because it is what just got fixed — and asserts the tenant-scoped count is `<=` the global one, so the global path can't silently degrade into a tenant count.

- `tests/` (the CI suite): **3957 passed**, 3 skipped, 1 xfailed, 3 xpassed — zero failures
- `core-storage-api/tests/` on a pristine DB: **18 failed / 84 passed** (from 35/65)
- `ruff check` + `ruff format --check` clean on both `src/` trees; `mypy` clean for both configs (188 + 73)

## What's left, and why CI adoption isn't in this PR

18 failures remain. I triaged all of them — every one is **stale-test drift against tightened API contracts**, and I found no product bugs:

| Class | Count |
|---|---|
| Endpoint now requires `tenant_id` (explicit 400/422) | ~5 |
| Route moved / method changed (404, 405) | ~6 |
| Response shape changed (`KeyError`, `TypeError: list indices`) | ~3 |
| Bulk write now requires `client_request_id` | 1 |
| Other assorted contract drift | ~3 |

They need a separate sweep, since fixing them means deciding per-test whether the test or the contract is the intended behaviour — not something to bundle behind a provisioning fix.

**Recommendation:** land this first so the suite is deterministic, do the 18-test sweep next, then add `core-storage-api/tests/` to `ci.yml` in the same PR that gets it to green. Adding it to CI before then would just pin main red. Right now `ci.yml` only runs `uv run pytest tests/`, which is how this drifted for so long.

### DIFF
--- a/core-storage-api/src/core_storage_api/routers/memories.py
+++ b/core-storage-api/src/core_storage_api/routers/memories.py
@@ -784,17 +784,22 @@ async def get_lifecycle_candidates(tenant_id: str) -> dict:
 
 @router.get("/count")
 async def count_memories(
-    tenant_id: str,
+    tenant_id: str | None = None,
     fleet_id: str | None = None,
     status: str | None = None,
 ) -> dict:
     """Count live memories for a tenant. ``status`` narrows to one exact status.
 
-    An empty ``tenant_id`` (``?tenant_id=`` — an omitted param is a 422, since
-    the annotation is non-optional) falls through to the whole-corpus counter,
-    which takes no filters: ``fleet_id`` and ``status`` do NOT apply there. That
-    branch predates this endpoint's filters; it stays as-is rather than growing
-    a second filtered code path.
+    Omitting ``tenant_id`` (or passing it empty) returns the whole-corpus
+    counter, which takes no filters: ``fleet_id`` and ``status`` do NOT apply
+    on that path. It stays filterless rather than growing a second filtered
+    code path.
+
+    The annotation is optional so both forms reach that branch. It used to be
+    a bare ``str``, i.e. required, so only the empty-string form worked and an
+    omitted param 422'd — even though ``core-api``'s public-stats caller
+    (``routes/stats.py``, ``count_all(tenant_id="")``) and the whole-corpus
+    branch were both written for "no tenant".
     """
     if not tenant_id:
         count = await _svc.memory_count_all()
--- a/core-storage-api/tests/conftest.py
+++ b/core-storage-api/tests/conftest.py
@@ -34,30 +34,40 @@ _schema_ready = False
 
 @pytest.fixture(scope="session")
 async def _ensure_schema():
-    """Create extensions + tables for the test database once per session."""
+    """Provision the test database once per session, the way the app does.
+
+    Runs the Alembic migrations via ``init_database()`` — the same call the
+    FastAPI lifespan hook makes — rather than ``Base.metadata.create_all``
+    over a hand-maintained model import list.
+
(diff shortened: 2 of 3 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

Eldad-Caura:

Two notes for the reviewer:

1. **`/count` signature widening is additive.** `tenant_id` goes from bare `str` to `str | None`. The empty-string form is unchanged, which matters because `core-api/routes/stats.py` reaches the whole-corpus branch via `count_all(tenant_id="")` — that branch is live in production, not dead code. Only the previously-422'ing *omitted* form changes.

2. **The conftest change swaps `create_all` for the real migration chain**, so `core-storage-api/tests/` now provisions the way production does. It takes that suite from 35 failures on a pristine DB (or 31 on a carried-over one) to a deterministic 18. The remaining 18 are stale-test drift against tightened contracts, triaged in the PR body — they need their own sweep before this suite can join CI.

erni-a (approved):

PR #631, opened 2026-07-26
fix(storage): count missing embeddings instead of len()-ing a LIMITed fetch

Eldad-Caura:
Closes the last wrong number from the #626 → #628 → #630 chain.

## The bug

`/embedding-coverage` built its numerator by fetching rows and calling `len()` on them:

```python
missing = await memory_find_missing_embeddings(tenant_id, fleet_id)   # LIMIT :batch_size
...
"missing_embeddings": len(missing),
"coverage_pct": round((total - len(missing)) / total * 100, 1)
```

That query carries `LIMIT :batch_size` (default **100**). So any tenant with more than 100 un-embedded memories under-reports how many are missing, and therefore **over-reports coverage**:

| Actual | Reported before | Reported now |
|---|---|---|
| 105 missing of 108 → 2.8% covered | `missing=100` → **5.4%** | `missing=105` → **2.8%** |

The failure mode is the bad direction: the number looks *better* the worse the real backlog is, and it silently saturates. A tenant with 10,000 un-embedded rows and a tenant with 100 report identically.

#630 fixed this endpoint's *population* mismatch (numerator and denominator disagreeing on status), which is what made `coverage_pct` able to go negative. This is the remaining half — the numerator was also **capped**.

## The fix

New `memory_count_missing_embeddings` — an exact `COUNT` over the same population the denominator uses (`LIVE_MEMORY_STATUSES`, not soft-deleted, same tenant/fleet scope).

It sits **directly next to `memory_count_active`** on purpose: those two are the numerator and denominator of one ratio, and they only produce a meaningful percentage while their predicates stay identical. Splitting them across the file is what allowed them to drift apart in the first place.

## Removing `memory_find_missing_embeddings`

The coverage endpoint was its **only** caller, so it's now dead. Removed rather than left behind:

- the backfill path that genuinely needs rows already has `memory_list_null_embedding_rows`, which pages on a cursor and returns its own exact `total_remaining` — the same COUNT-not-`len()` pattern this PR applies;
- it selected `m.content` for every row and never read it, so the replacement is cheaper as well as correct.

## Testing

`test_embedding_coverage_is_not_capped_by_a_batch_size` seeds **105** un-embedded rows plus 3 embedded ones. The 105 is the point — at 100 or fewer the test passes with or without the fix, which is exactly why the existing coverage test never caught this.

Verified the test has teeth by simulating the old cap: it fails with `assert 100 == 105`. Worth noting the pre-existing `test_embedding_coverage_spans_the_live_status_set` **passed** under that simulated cap — so the new test isn't redundant with it.

- `tests/` (CI suite): **3958 passed**, 3 skipped, 1 xfailed, 3 xpassed — zero failures
- `core-storage-api/tests/`: **18 failed / 85 passed** — same 18 known-drift failures as `main`'s post-#630 baseline, +1 new pass
- No dangling references to the removed method (`grep` → 0)
- `ruff check` + `ruff format --check` clean on both `src/` trees; `mypy` clean for both configs (188 + 73)

### DIFF
--- a/core-storage-api/src/core_storage_api/routers/memories.py
+++ b/core-storage-api/src/core_storage_api/routers/memories.py
@@ -721,17 +721,20 @@ async def get_embedding_coverage(
 ) -> dict:
     """Share of live memories that have an embedding.
 
-    Numerator and denominator both scope to ``LIVE_MEMORY_STATUSES`` — they
-    used to disagree (missing spanned every status, total only ``active``),
-    which let coverage_pct exceed 100% or go negative. ``missing_embeddings``
-    is still capped by the finder's batch_size, so treat it as a floor.
+    Numerator and denominator scope to the same population
+    (``LIVE_MEMORY_STATUSES``, not soft-deleted, same tenant/fleet), so
+    ``missing_embeddings <= total_active`` holds and ``coverage_pct`` stays
+    within 0-100. Both are exact COUNTs; the numerator used to be
+    ``len(rows)`` off a ``LIMIT``-ed query, which capped it at the batch size
+    and made coverage read high on any tenant with more missing rows than
+    that.
     """
-    missing = await _svc.memory_find_missing_embeddings(tenant_id, fleet_id)
+    missing = await _svc.memory_count_missing_embeddings(tenant_id, fleet_id)
     total = await _svc.memory_count_active(tenant_id, fleet_id)
     return {
         "total_active": total,
-        "missing_embeddings": len(missing),
-        "coverage_pct": round((total - len(missing)) / total * 100, 1) if total > 0 else 0.0,
+        "missing_embeddings": missing,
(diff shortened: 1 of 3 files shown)


Eldad-Caura:
@claude

Eldad-Caura:

For the reviewer: this replaces a `len()` over a `LIMIT 100` fetch with an exact COUNT, so `/embedding-coverage` stops over-reporting on any tenant with >100 un-embedded rows. It also deletes `memory_find_missing_embeddings`, which the coverage endpoint was the sole caller of — the backfill path already uses `memory_list_null_embedding_rows` (cursor-paged, own exact total). The new test seeds 105 rows specifically to cross the old boundary; below 100 it would pass either way, which is why the existing coverage test never caught this.

erni-a (approved):

PR #647, opened 2026-07-27
fix(storage): green the core-storage-api integration suite and add it to CI

Eldad-Caura:
The sweep flagged in #630. `core-storage-api/tests/` goes from **18 failures → 103 passing**, and into CI so it can't drift again.

## Why it drifted

`ci.yml` only ran `tests/`. This suite was neither run nor linted, so nothing caught it.

## 17 of 18: stale tests, one shared root cause

Nearly all of them addressed the **internal storage API using the public core-api URL grammar**. The two layers deliberately differ — and `core-api`'s `storage_client`, the only production consumer, matches the real routes in every single case. That's what settles which side was wrong.

| Test expected | Actual route | Symptom |
|---|---|---|
| `PATCH /memories/{id}` without `tenant_id` | body `tenant_id` required (cross-tenant write guard) | 422 ×6 |
| `PATCH /memories/batch-status` | `POST /memories/batch-update-status`, batch-level `tenant_id` | fell through to `/{memory_id}`, 422 on UUID coercion |
| `/entities/by-name` | `/entities/exact` | 422 |
| `/entities/memory-links/create` | `/entities/links` | 404 |
| `/entities/linked-memories/{id}` | `/entities/{id}/with-memories` | 404 |
| `/documents/{doc_id}?collection=` | `/documents/{collection}/{doc_id:path}` | matched list-documents → `200 []` → `TypeError` |
| `/agents/{id}/trust` | `/agents/{id}/trust-level` (`/trust` is the *public* route) | 404 |
| `/fleet/commands/pending/{node}` | `node_name` is a query param | 404 |
| `{"memory": {...}}` / `{"entity": {...}}` envelopes | flat ORM dict | `KeyError` |
| `200` + null on miss | `404` | assertion mismatch |
| `metadata` | `metadata_` | `KeyError` |

The 404-on-miss is worth calling out as deliberate rather than sloppy: `storage_client._get` maps 404 → `None`, which is precisely how those callers' optional returns work. A `200` with a null envelope would make `check_exact_duplicate` treat **every** write as a duplicate.

## The 18th: a real defect

`POST /memories/bulk` let `memory_add_all`'s `ValueError` escape unhandled, so a batch missing `client_request_id` — or mixing `tenant_id`/`fleet_id` — returned **500**. A server-fault code for something only the caller can fix: it pages, and it lands in the DLQ.

Now 422, matching the convention this same router already states three times (`_parse_datetimes`, `_validate_pg_regex`, `dedup_review_enqueue`).

**Not a retry-behaviour change** — I checked before touching it: core-api retries POSTs on connection-phase failures only ("request provably never sent, so a retry cannot double-insert"), never on status code. And the guard runs before any session opens, so nothing was written either way.

## Two assertions were vacuous

Both could never have failed, and now can:

- the resurrect check read a `metadata` key that **never exists**, so `(after.get("metadata") or {})` was always `{}`;
- the bulk test accepted any `>= 400`, including the 500 it was actually getting. Now pinned to `== 422`.

## CI wiring — separate database, deliberately

The suite gets its **own** database rather than sharing `memclaw`. The two suites provision differently: `tests/` uses `create_all`, this one runs the real Alembic chain. On a shared database, whichever ran first would win — and `create_all` first leaves tables with no `alembic_version`, so `init_database()` stamps head, skips the migrations, and every migration-only table (`tenant_suppression`, which has no ORM model) goes missing again. That's the exact bug #630 fixed, reintroduced through step ordering.

Separate databases remove the coupling instead of depending on step order holding forever.

The test tree is now ruff-linted too (2 pre-existing errors fixed) — it was unlinted for exactly as long as it was un-run, and leaving that half open just invites the same drift.

## Verification

- `core-storage-api/tests/`: **103 passed, 0 failed** — on a pristine DB, again on the reused one (no state leakage), and via the exact CI env (`DATABASE_URL` only, DB named `memclaw_storage`)
- `tests/` (existing CI suite): **3958 passed**, 3 skipped, 1 xfailed, 3 xpassed — the 422 change doesn't disturb core-api
- `ruff check` + `ruff format --check` clean across both `src/` trees **and** the storage test tree; `mypy` clean for both configs (188 + 73)
- `ci.yml` parses

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -124,11 +124,19 @@ jobs:
       - name: Run Ruff linter (core-storage-api)
         run: uv run ruff check core-storage-api/src/
 
+      - name: Run Ruff linter (core-storage-api tests)
+        # The test tree was unlinted for as long as it was un-run; both halves
+        # of that blind spot close together or it just drifts again.
+        run: uv run ruff check core-storage-api/tests/
+
       - name: Run Ruff formatter check (core-api)
         run: uv run ruff format --check core-api/src/
 
       - name: Run Ruff formatter check (core-storage-api)
         run: uv run ruff format --check core-storage-api/src/
+
+      - name: Run Ruff formatter check (core-storage-api tests)
+        run: uv run ruff format --check core-storage-api/tests/
 
       - name: Run Ruff formatter check (vendored common/ files at ruff defaults)
         # This list mirrors the policy=identical entries of caura-memclaw-enterprise's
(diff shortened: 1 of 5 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
Verified and applied the Low finding — it was right, and it mattered more than its severity suggests.

**Confirmed the premise:** all three validation guards in `memory_add_all` (postgres_service.py) do run before `async with get_session()`, but my `try` block spanned the whole call. So a plain `except ValueError` would also have swallowed anything raised from *inside* the session and relabelled a genuine server fault as a 422 — the exact inverse of the 500-masking-a-4xx bug this PR set out to fix.

**Fix:** added `BulkValidationError(ValueError)` and raised it from the three guards; the router now catches that instead of bare `ValueError`. It subclasses `ValueError` so any caller catching broadly is unaffected. This follows the existing house pattern — `PIIInAuditError(ValueError)` in `services/audit_chain.py`.

Re-verified after the amend: storage suite **103 passed** via the exact CI env, `tests/` **3958 passed**, ruff check + format clean across both src trees and the storage test tree, mypy clean for both configs.

Re-tagging for another pass since this changed production error-handling semantics after the first review.

Eldad-Caura:
@erni-a please approve

Eldad-Caura:
@claude

erni-a (dismissed):

Eldad-Caura:
Applied — but the suggested command was broken, so I used a different form and verified it.

**On the suggestion itself:** `\gexec` is a psql *meta-command* and does not work with `-c`, which sends its argument as a single SQL string. I ran the proposed line against a real Postgres and got a syntax error at `\gexec` — and, worse, it exits 0, so it would have silently failed to create the database and left the pytest step to fail confusingly later.

**What I used instead** — same guard, fed via stdin so meta-commands are processed:

```
PGPASSWORD=changeme psql -h localhost -U memclaw -d memclaw -v ON_ERROR_STOP=1 <<'SQL'
SELECT 'CREATE DATABASE memclaw_storage'
WHERE NOT EXISTS (SELECT FROM pg_database WHERE datname = 'memclaw_storage')\gexec
SQL
```

I kept `ON_ERROR_STOP=1` rather than the simpler `|| true`: that idiom would also swallow a genuine connection or auth failure and defer the confusion to the pytest step, which is the same fail-quietly shape this PR is otherwise removing.

**Verified** by parsing the workflow, extracting the step's actual `run` script, and executing it twice against a real Postgres: `CREATE DATABASE` then a clean no-op, exit 0 both times, database present.

**One caveat on the premise, for the record:** GitHub-hosted runners provision fresh service containers per job execution, so "Re-run failed jobs" would not actually meet a still-live Postgres — the literal failure mode described is unlikely there. I made the change anyway because it costs nothing, and it does matter for self-hosted runners, reused services, and anyone reproducing the CI steps locally.

Re-verified after the amend: storage suite **103 passed**, ruff check + format clean.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

Eldad-Caura:

Two review rounds, both findings verified and applied. All checks green — CI ✅ (now including the new core-storage-api suite step) · CodeQL python + js/ts ✅ · Security Scan ✅ · image-version ✅ · DCO ✅ · pre-checks ✅. Single signed commit.

Summary for the reviewer:
- `core-storage-api/tests/` goes 18 failures → **103 passing**, and is now run and ruff-linted in CI so it can't drift again.
- 17 of the 18 were stale tests addressing the internal storage API with the public core-api URL grammar; `storage_client` (the only production consumer) matched the real routes in every case, which is what settled each one.
- The 18th was a real defect: `POST /memories/bulk` returned **500** for a malformed batch. Now 422, scoped to a dedicated `BulkValidationError` so unexpected in-session failures still surface as 500s.
- The suite runs against its own database — sharing one would let `create_all` win the race and silently skip the migrations again.

erni-a (approved):

PR #652, opened 2026-07-27
fix(embedding): cap in-flight embed calls to stop pool-exhaustion cascade

Eldad-Caura:
## Incident

**2026-07-27 07:26–09:10Z.** ~77% of embeds failed for ~1h45m; **~430 memories were left permanently unembedded** — stored fine, all HTTP 200s, but not semantically recallable. Silent data-quality loss.

## Root cause: oversubscription of our OWN backend, not a provider outage

Prod points `OPENAI_EMBEDDING_BASE_URL` at a **self-hosted TEI service** (`BAAI/bge-m3`, L4 GPU), not OpenAI:

| side | capacity |
|---|---|
| TEI backend | `maxScale 2` × `containerConcurrency 10` = **20 concurrent** |
| core-api demand | `minScale 10` (max 200) instances × **200-connection** pool = **≥2,000** |

When bulk volume spiked ~18× (181/hr → 3,274/hr), TEI's 20 slots saturated. Callers queued at the HTTP pool and died on `httpx.PoolTimeout` → `openai.APITimeoutError` — **while TEI itself stayed healthy, reporting 3.5 ms inference throughout.**

The self-sustaining part: a failed 50-item batch falls back to 50 per-item calls, re-jamming the same 20 slots → more failures → more fallback.

## Changes

- **`EMBEDDING_MAX_CONCURRENCY`** (default 16) caps in-flight provider calls per process — surplus work waits on a cheap semaphore instead of holding a connection until the pool times out. Applied at both provider chokepoints, acquired **per attempt** so a backing-off retry never squats on a slot.
- **`EMBEDDING_GATE_TIMEOUT_SECONDS`** (default 5.0) bounds the wait. Deliberately **below** the tightest caller cap (the search path's 10 s) — at or above it the outer timer always fires first and the gate could never attribute the failure to backpressure. On timeout the caller degrades down the *pre-existing* provider-failure path.
- **Gate logs saturation and timeout distinctly.** During the incident the backend reported 3.5 ms while callers timed out and nothing said where the time went.
- **`EMBEDDING_HTTPX_MAX_CONNECTIONS` / `_KEEPALIVE`** split the embedding pool from the LLM pool. Both previously read `OPENAI_HTTPX_MAX_CONNECTIONS`, so shrinking the embedding pool to match a small self-hosted backend would also throttle every LLM call — making the one knob you need during an embedding incident unusable.
- **`read_float_env` promoted to `common/env_utils`** (beside `read_int_env`, which that module exists to share); `common/llm/constants` delegates to it. The new gate timeout would otherwise have used a bare `float(os.environ.get(...))` that raises at module import on a bad value.

## Honest limits of this PR

**This reduces thrash amplitude; it does not restore a capacity invariant.** The cap is *per process*, so 16 × instance count is still well above TEI's capacity. Called out in the constant's comment rather than papered over.

**Two structural follow-ups I deliberately did not fold in** (both surfaced by review, both deserve their own focused change):

1. **The request-count amplifier survives.** The gate removes *connection* amplification, but one failed 50-item batch still spawns 50 × 3 re-embed retries × 2 attempts ≈ 300 gated acquisitions against 16 slots. Most time out → row stays NULL → same terminal outcome as the incident, with no automatic recovery (backfill is a manual CLI). The right fix is one function away: swap the in-process `track_task(_reembed_memory(..., is_failure_fallback=True))` fallbacks for a durable `EMBED_REQUESTED` publish, whose consumer already has Pub/Sub redelivery + DLQ + per-tenant slots. That would make the permanently-unembedded class structurally impossible — but it changes retry semantics and contradicts a deliberate existing comment, so it needs its own review.
2. **The gate is a single unkeyed global**, while the constraint is per-backend. The registry supports simultaneous distinct backends, so a saturated TEI tenant currently throttles an OpenAI tenant. Keying the gate on the same backend identity the provider cache already uses is the deeper form.

## Related (already live, not in this PR)

TEI `maxScale` raised **2 → 3** (20 → 30 concurrent). Capped there by GPU quota — `NvidiaL4GpuAllocNoZonalRedundancy` allows 3; a quota increase is needed to go further.

## Validation

- 4 new tests in `tests/test_embedding_concurrency_gate.py`: cap enforced (30 callers → 3 in-flight), slot released on exception, blocked waiter degrades instead of hanging, gate rebinds per event loop.
- `ruff check` / `ruff format` clean on changed files (the 3 remaining `G201`/`RUF100` hits are pre-existing on `main`, verified by linting the pristine tree).
- Defensive parsing verified: a misconfigured `EMBEDDING_GATE_TIMEOUT_SECONDS=30s` warns and falls back instead of crashing import.

### DIFF
--- a/common/embedding/_service.py
+++ b/common/embedding/_service.py
@@ -14,7 +14,12 @@ import time
 from collections.abc import Awaitable, Callable
 
 from common.embedding._registry import get_embedding_provider
-from common.embedding.constants import EMBEDDING_RETRY_ATTEMPTS, EMBEDDING_RETRY_DELAY_S
+from common.embedding.constants import (
+    EMBEDDING_GATE_TIMEOUT_SECONDS,
+    EMBEDDING_MAX_CONCURRENCY,
+    EMBEDDING_RETRY_ATTEMPTS,
+    EMBEDDING_RETRY_DELAY_S,
+)
 from common.embedding.protocols import InstructionAwareEmbedder
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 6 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #653, opened 2026-07-27
fix(embedding): make batch-failure re-embed fallback durable

Eldad-Caura:
## Why

Closes the tail of the **2026-07-27 incident**. #652 capped concurrency so the backend stops being thrashed — but it did **not** remove the request-count amplification, nor the permanence.

A failed 50-item bulk re-embed batch fanned out to **50 in-process tasks × 3 re-embed retries × 2 embedding attempts ≈ 300 provider calls**, all contending for the same already-saturated backend. When they exhausted, the rows stayed `embedding=NULL` **with no further recovery** — a manual CLI backfill was the only way out. That is exactly how **~430 memories were stranded**: stored, HTTP 200, but not semantically recallable.

I originally claimed #652's gate subsumed this. It doesn't — the gate removes *connection* amplification only. This PR is the actual fix.

## What changed

The five batch-failure fallbacks now go through the **existing** `_schedule_embed_or_reembed` router instead of calling `_reembed_memory` directly:

| mode | behaviour |
|---|---|
| **deferred** (prod) | one cheap `EMBED_REQUESTED` publish per item → retry lives on **Pub/Sub with redelivery + DLQ**, and survives instance restart |
| **inline** (OSS, no worker fleet) | unchanged: in-process retry |

This makes the permanently-unembedded class **structurally impossible** rather than merely less likely.

`is_failure_fallback` is threaded through the router so the **inline** branch keeps its thundering-herd backoff (the provider just failed for the whole batch); the deferred branch ignores it because Pub/Sub owns backoff.

**No new mechanism** — this reuses the same router the hot-path offload already uses, which is why the diff is small. The deliberate comment that previously documented the in-process behaviour is updated to explain the durability rationale.

## Validation

- `ruff check` / `ruff format` clean; module compiles; all 5 sites routed, 0 direct calls remain.
- 3 new tests in `tests/test_durable_reembed_fallback.py`: deferred publishes and does **not** retry in-process; inline retries in-process **and** receives the backoff flag; the hot-path default does **not** inherit the backoff.
- ⚠️ **The new tests were not run locally** — `memory_service` pulls in the full app dependency graph that isn't installed in my sandbox (`cachetools`, `croniter`, …). CI runs them in the real environment.

### DIFF
--- a/core-api/src/core_api/services/memory_service.py
+++ b/core-api/src/core_api/services/memory_service.py
@@ -1736,6 +1736,7 @@ async def _schedule_embed_or_reembed(
     tenant_id: str,
     *,
     content_hash: str | None = None,
+    is_failure_fallback: bool = False,
 ) -> None:
     """Backfill the embedding for a memory persisted with ``embedding=NULL``.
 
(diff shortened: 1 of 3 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@erni-a pleae approve

erni-a (approved):

PR #655, opened 2026-07-27
fix(client-ci): pin ruff to the 0.15 line so client CI stops floating into 0.16

Eldad-Caura:
## What
Pins the `memclaw_client` dev extra to `ruff>=0.15.4,<0.16` (from `ruff>=0.6`).

## Why
`client-python-ci.yml` installs ruff via `pip install -e ".[dev]"`. With `ruff>=0.6` it floated to **0.16**, whose new default rules fire on existing code — **18 findings** (RUF022 `__all__`-not-sorted, C408, …). That silently red-lined the *Lint (ruff)* step on every `clients/python` PR, including the Dependabot `setup-python` bump #640 (which has nothing to do with ruff). Note the pre-commit ruff hook is scoped to `core-api`/`core-storage-api`, so it never covered the client package.

## Fix
Pin to the **0.15 line**, matching `.pre-commit-config.yaml`'s `rev: v0.15.4`. Verified locally:
- ruff **0.15.4** → `All checks passed!`
- ruff **0.16.0** → `Found 18 errors`

Adopting ruff 0.16 for the client (fixing the 18 findings) is a deliberate follow-up — same reasoning as the enterprise frozen-lint work — not a silent CI-breaking float. This unblocks #640 and other client PRs.

PR #245, opened 2026-05-28
perf(entity-extraction): collapse worker N+1 storage HTTPs (P1)

ran-taig:
## Summary

Audit P1 — `process_entity_extraction` previously issued up to **6 HTTPs per entity**: 1 embed, 2-3 inside `upsert_entity` (find_exact / find_similarity / update_or_create), 2 for the entity link (find / create). For a 10-entity memory that's ~60 HTTPs; the audit-validation flagged ceilings around 240.

After P1: **one ``asyncio.gather`` round + 3 bulk HTTPs per memory**, regardless of entity count.

## Sites collapsed

| Stage | Before | After |
|---|---|---|
| Embeddings | Serial `await get_embedding(name)` per entity | One `asyncio.gather` with `return_exceptions=True` |
| Entity upsert | Per-entity `upsert_entity` (2-3 HTTPs internally) | One `sc.bulk_resolve_entities` + client-side merge + one `sc.bulk_upsert_entities` |
| Entity links | Per-entity `find_entity_link` + `create_entity_link` (2 HTTPs) | One `sc.bulk_upsert_entity_links` |

## Correctness gate: first-seen-wins canonical_name

The audit's most important correctness rule — see [`entity_service.py:91-100`](../blob/main/core-api/src/core_api/services/entity_service.py#L91-L100). When an existing entity matches (exact or similarity), the **existing** `canonical_name` wins and the new surface form accumulates in `attributes["_aliases"]`. This guards a real prior regression where "longest-wins" promoted LLM hallucinations to canonical rows.

The worker ports ``entity_service.upsert_entity``'s merge logic (lines 74-100) verbatim into a client-side step between resolve and upsert.

## Behavior preserved

- Blocklist + min-length filter applied **before** the bulk batch (was previously interleaved with per-entity HTTPs)
- Duplicate `canonical_name` in `graph.entities` collapses to first occurrence (matches old `name_to_id`-dict semantics)
- Per-entity embed failure → `None` in resolve payload → Phase 2 skipped server-side; exact-match still runs (same skip semantic as old per-entity try/except)
- Audit log, contradiction trigger, cross-link discovery wiring unchanged

## Test plan

- [x] **New** `tests/test_p1_entity_extraction_bulk.py` (11 tests):
  - **Shape** (5): one of each bulk call per memory; zero per-row HTTPs; empty graph skips bulk; blocklist filter; duplicate-name collapse with role binding
[...]

PR #656, opened 2026-07-27
ci: add Dependabot minor/patch auto-merge workflow (OSS-tailored)

Eldad-Caura:
## What
Ports enterprise's `dependabot-automerge.yml` (incl. the enterprise-#929 `-R "$REPO"` fix) so Dependabot's **`pip-`/`uv-`/`npm-minor-patch`** GROUP PRs self-merge once required CI is green — draining the ~12-PR manual backlog.

Tailored for this repo's model:
- **Groups:** pip/uv/npm-minor-patch (this repo's group names).
- **Required checks:** `CI` + `dco`, read from *classic branch protection* (this repo uses branch protection, not a ruleset).
- **Defensive merge:** if the merge is blocked it logs and exits 0 (no-op) instead of failing — so it's **safe to merge before the bypass below is set up**.

Scope mirrors the enterprise carve-out: only those minor/patch groups auto-merge; majors, constraint-updates, and the github-actions group stay manual. Merge runs under the DEPLOY_APP token, gated by the script (not `--auto`) and pinned via `--match-head-commit`.

## ⚠️ Prerequisite before it actually merges
`main` requires **1 approving review**, and there is currently **no App bypass** configured for it (no ruleset bypass actor / no `bypass_pull_request_allowances`). So an App-token merge is rejected on "review required" until you add the deploy App to **Settings → Branches → main → Allow specified actors to bypass required pull requests**. (The `DEPLOY_APP` org secrets are already shared with this repo via release-please.)

This is a **public-repo** auto-merge, so it also means Dependabot minor/patch groups would merge without the 1 human approval — flagging that explicitly since it's your call.

Verified: `CI` + `dco` both pass on this repo's Dependabot PRs (so the gate is satisfiable), and this pairs with #655 which unblocks the client CI lint.

### DIFF
--- a/.github/workflows/dependabot-automerge.yml
+++ b/.github/workflows/dependabot-automerge.yml
@@ -1,0 +1,208 @@
+name: Dependabot auto-merge (minor/patch groups)
+
+# Auto-merges Dependabot's minor/patch GROUP PRs once the required CI is green,
+# so the high-volume dependency backlog stops needing a manual merge each time.
+# Ported from caura-memclaw-enterprise's workflow of the same name and tailored
+# for this (OSS, public) repo — see the OSS-SPECIFIC notes below.
+#
+# Scope is deliberately narrow — a carve-out from the never-auto-merge policy:
+#   * ONLY the `pip-minor-patch`, `uv-minor-patch`, and `npm-minor-patch` groups
+#     qualify. Those groups are defined in .github/dependabot.yml with
+#     update-types [minor, patch], so a group PR can only contain minor/patch
+#     bumps. Majors arrive as separate, ungrouped PRs (different branch names)
+#     and are intentionally left for manual review.
+#   * Also intentionally MANUAL: the github-actions group and any ungrouped
+#     constraint-update / major PRs.
+#
+# pull_request_target (not pull_request): Dependabot-triggered `pull_request`
+# runs get a read-only token and NO secrets, so they cannot mint the App token
+# needed to merge under branch protection. pull_request_target runs in the base
+# repo context with secrets. Safe here because this workflow NEVER checks out or
+# executes PR-provided code — it only reads the trusted event payload (PR
+# number/author/branch) and calls gh. The `if` also gates on the dependabot[bot]
+# actor, which cannot be spoofed (it is the authenticated actor of the event).
+#
+# OSS-SPECIFIC vs the enterprise version:
+#   * This repo's `main` uses CLASSIC branch protection (not a ruleset): required
+#     checks are `CI` + `dco`, and it requires 1 approving review.
+#   * PREREQUISITE for the merge to succeed: the deploy App must be allowed to
+#     bypass that required review — add it to
+#     Settings → Branches → main → "Allow specified actors to bypass required
+#     pull requests" (and it must be able to merge with the strict/up-to-date
+#     rule). Until then the merge is blocked; this workflow DETECTS that and
+#     no-ops (logs + exits 0) rather than failing the job, so it's safe to land
+#     before the bypass is configured.
+#   * DEPLOY_APP_ID / DEPLOY_APP_PRIVATE_KEY org secrets are already shared with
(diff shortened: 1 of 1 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@claude

Eldad-Caura:
Addressed the review:
- **High (StatusContext read as MISSING):** the check-state extraction now coalesces `.state` — `(.conclusion // .state // .status)` — so `dco` (a StatusContext, not a CheckRun) is read correctly instead of always MISSING. This was the bug that would have kept the merge from ever firing on this repo.
- **Low (STALE):** added `STALE` to the fast-exit case so a superseded check bails immediately instead of polling to the 20-min deadline.
- Also folded in the merge-error handling from the #245 review: head-moved → graceful `exit 0`; genuine errors → `exit 1`; the not-yet-bypassed (review-required) case stays a logged no-op so it's safe to land before the bypass is configured.

Verified `bash -n` clean.

Eldad-Caura:
@claude

Eldad-Caura:
Round 2 addressed:
- **High (StatusContext `.context`):** verified empirically that `dco` on this repo is a **CheckRun** (`name=dco, conclusion=SUCCESS`; `context`/`state` null), not a legacy StatusContext — so the workflow was **not actually blocked**. Applied the robust handling anyway: selector now matches `(.name // .context)`, and the extracted value is upper-cased (`| tr`, since legacy StatusContext `.state` is lowercase). Correct for both node types now.
- **Medium (actor vs opener):** added `github.actor == 'dependabot[bot]'` so a maintainer pushing to a Dependabot branch can't get their hand-written commit auto-merged on `synchronize`; legitimate Dependabot rebases/recreates still run as `dependabot[bot]`.
- **Low (fragile error grep):** the raw merge output is now logged unconditionally, and the review-blocked pattern includes `422|approval`.

`bash -n` + YAML verified.

Eldad-Caura:
@claude

Eldad-Caura:
Round 3 addressed (both Medium/Low, both real):
- **Empty `$head` guard (Medium):** added `[ -n "$head" ] || { …; exit 0; }` before the merge — `gh` treats `--match-head-commit ""` as valid-and-matches-any, so an absent `.headRefOid` would've silently neutered the race guard.
- **head-moved regex (Low):** broadened to match GitHub's real 409 (`Head/Base branch was modified…`) via `branch was modified|409`, so the expected synchronize-race is the intended graceful `exit 0`, not a red run.

`bash -n` + YAML verified.

Eldad-Caura:
@claude

Eldad-Caura:
Round 4 handled:
- Applied the branch-protection read fix — the required-check query now unions `checks[].context` + `contexts[]` (both the modern and legacy API formats) so the list can't silently fall back. (Verified `contexts[]` does return `[CI, dco]` for this repo today and the fallback is exactly that pair, so there was no actual gap — but this is more correct for an auto-merge gate.)
- The merge-classifier `not mergeable` log nitpick is cosmetic (behavior correct) — leaving it.

The earlier CI failure was a transient Docker Hub timeout (`i/o timeout` pulling `python:3.12-slim`), unrelated to this change; the re-run passed. Stopping the review loop here — the logic is verified correct across 4 rounds. Will tag @erni-a once the new head's CI is green.

erni-a (approved):

Eldad-Caura:
@erni-a — ready for your approval. Required checks green on the current head (`CI` + `dco`), review handled across 4 rounds (security model, actor gate, StatusContext handling, empty-head guard, real-409 head-moved detection, and the dual-format required-check read are all in). Reminder from the PR description: this won't actually merge Dependabot PRs until the deploy App is added to `main`'s required-review bypass allowances — and since it's a public repo, that means Dependabot minor/patch would merge without the 1 human approval (your call). The workflow no-ops safely until then. Merge stays with Eldad.

erni-a (approved):

PR #660, opened 2026-07-28
perf(entities): add HNSW index on entities.name_embedding

Eldad-Caura:
## What

Adds an HNSW index on `entities.name_embedding` (migration `033`), mirroring `ix_memories_embedding_hnsw` from `001`.

## Why

The entity cross-link discovery path — `PostgresService.entity_discover_cross_links` → `POST /api/v1/storage/entities/discover-cross-links` — runs a per-candidate ANN search over a tenant's entities:

```sql
JOIN LATERAL (
    SELECT ... FROM entities e
    WHERE e.tenant_id = :tenant_id AND e.name_embedding IS NOT NULL ...
    ORDER BY e.name_embedding <=> m.embedding
    LIMIT 10
) e ON true
```

`entities.name_embedding` has had **no ANN index since `001`** (migration `012` notes this explicitly). Without one, every candidate memory triggers a full sequential scan of the tenant's entities — `O(candidates × entities)` cosine computations in a single request. On large tenants this exceeds the caller's 120 s timeout, observed as a **`504 Gateway Timeout`** on the storage endpoint in prod + staging (2026-07). The index makes each lookup `O(log n)` and speeds up entity linking generally.

## How

- `CREATE INDEX CONCURRENTLY IF NOT EXISTS ix_entities_name_embedding_hnsw ON entities USING hnsw (name_embedding vector_cosine_ops) WITH (m = 16, ef_construction = 64)`.
- Built `CONCURRENTLY` inside `op.get_context().autocommit_block()` so it doesn't take an AccessExclusiveLock on `entities` during the build — same idiom as `012`/`024`. `IF NOT EXISTS` keeps it idempotent on the advisory-lock-serialised startup migration path.
- Params (`m`, `ef_construction`) and opclass (`vector_cosine_ops`) match the memories index.

## Notes / risk

- **No behaviour change** — results are identical; only the query plan changes. No test changes needed; CI validates the migration applies.
- The filtered-ANN-by-`tenant_id` characteristic is the same as the existing `memories.embedding` recall path (single-column HNSW + a `tenant_id` filter), so no new failure mode is introduced.
- One-time cost: the `CONCURRENTLY` build runs once in the startup migration on first deploy; on a large `entities` table it adds a one-time build to that replica's startup (other replicas wait on the advisory lock). Acceptable and non-locking.

### DIFF
--- a/core-storage-api/src/core_storage_api/database/migrations/versions/033_entities_name_embedding_hnsw.py
+++ b/core-storage-api/src/core_storage_api/database/migrations/versions/033_entities_name_embedding_hnsw.py
@@ -1,0 +1,60 @@
+"""Add HNSW index on ``entities.name_embedding`` for cross-link discovery.
+
+The entity cross-link discovery path
+(``PostgresService.entity_discover_cross_links`` →
+``POST /api/v1/storage/entities/discover-cross-links``) runs a per-candidate
+approximate-nearest-neighbour search over a tenant's entities:
+
+    JOIN LATERAL (
+        SELECT ... FROM entities e
+        WHERE e.tenant_id = :tenant_id AND e.name_embedding IS NOT NULL ...
+        ORDER BY e.name_embedding <=> m.embedding
+        LIMIT 10
+    ) e ON true
+
+``entities.name_embedding`` has had NO ANN index since 001 (migration 012
+notes this explicitly). Without one, every candidate memory triggers a full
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #661, opened 2026-07-28
fix(logging): route uvicorn parent-supervisor lines through structlog

Eldad-Caura:
## What

Adds a `common/serve.py` launcher and points the **core-api** and **core-storage-api** Docker `CMD`s at it (`python -m common.serve <app> --settings <module:attr> ...`), so the uvicorn multiprocess **parent supervisor** configures structlog before starting.

## Why

With `uvicorn --workers 2`, the parent supervisor never imports the ASGI app module, so the import-time `configure_logging()` in `<service>/app.py` runs **only in the workers**. The parent's own lifecycle lines — emitted via the `uvicorn.error` logger — bypass the structlog JSON pipeline and reach Datadog with no `status` field, where they're mis-classified as `status:error`:

```
INFO:     Started parent process [8]
INFO:     Waiting for child process [22]
INFO:     Received SIGTERM, exiting.
INFO:     Uvicorn running on http://0.0.0.0:8000 ...
```

This was the **single biggest source of false `status:error`** across the Cloud Run services in the 2026-07 error sweeps. It's functionally harmless, but it pollutes `status:error` queries and log monitors (the alert added after the embedding incident keys off message content, not `status`, so nothing is broken — this is log-hygiene).

## How

- `common/serve.py`: calls `configure_logging(...)` from the **service's own settings object** (identical args to `app.py`), then `uvicorn.run(..., log_config=None)`.
- `log_config=None` is the key: uvicorn then installs **no** handlers of its own, so `uvicorn.{error,access}` keep `propagate=True` and reach the structlog root handler in **both** the parent and the workers. The parent's lifecycle lines emit as JSON with `status:info`. (Letting uvicorn apply its default `LOGGING_CONFIG` would re-plain-text them.)
- Reuses the existing `configure_logging()` / `_route_third_party_to_root()` machinery — no new formatter.

## Tests

- `tests/test_serve.py`: the launcher reads settings and calls `configure_logging` from those fields + `uvicorn.run(log_config=None)` (no server started).
- `tests/test_logging.py`: an INFO record on the `uvicorn.error` logger renders as `severity:INFO` / `status:info` through the pipeline.

## Verification gate ⚠️

The **end-to-end** parent-supervisor output can only be confirmed with a running multi-worker server. **Please verify in staging before merge** — `env:staging` is off in Datadog, so check via `gcloud logging read` on the staging revisions that the supervisor lines (`Started parent process`, `Received SIGTERM`) now emit as JSON with `status:info` and that nothing else regressed. This touches the shared logging path for both services, so the staging check is the real gate.

## Follow-up

The enterprise **platform-\*** services share this same pattern; they'll be wired in a follow-up once this `common/` change vendors over.

### DIFF
--- a/common/serve.py
+++ b/common/serve.py
@@ -1,0 +1,97 @@
+"""Uvicorn launcher that configures structlog logging in the *parent* process.
+
+With ``uvicorn --workers N`` the multiprocess supervisor (the parent process)
+never imports the ASGI app module, so the import-time ``configure_logging()``
+call in ``<service>/app.py`` runs only inside the workers. The parent's own
+lifecycle lines — ``Started parent process [N]``, ``Waiting for child process
+[N]``, ``Received SIGTERM, exiting.``, ``Uvicorn running on ...`` — are emitted
+via the ``uvicorn.error`` logger with uvicorn's *default* configuration, which
+writes plain ``INFO:`` text to stdout. Those lines bypass the structlog JSON
+pipeline and reach Datadog without a ``status`` field, where they are
+mis-classified as ``status:error`` (the dominant false-error source across the
+Cloud Run services in the 2026-07 error sweeps).
+
+This launcher fixes that by calling ``configure_logging()`` — from the *same*
+settings object the app uses — before starting the server, then running uvicorn
+with ``log_config=None``. With ``log_config=None`` uvicorn does not install its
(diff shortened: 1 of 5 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #662, opened 2026-07-28
style(serve): keep common/serve.py <=88 cols for the enterprise vendoring gate

Eldad-Caura:
## Why

`common/serve.py` (added in #661) has a 91-col line. OSS CI doesn't format-check `common/`, so it merged clean — but that file is a prerequisite for launching the **enterprise** `platform-*` services the same way, which means vendoring it byte-for-byte into `caura-memclaw-enterprise` as an `identical`-policy `common/` file.

Enterprise CI runs `ruff format --check common/` with **no resolved config** → ruff's **88-col default**. A >88-col vendored line fails that gate, while the byte-identical drift check simultaneously forbids reformatting *only* the enterprise copy. That's the exact wedge from the 2026-06 `structlog_config` drift incident (fixed by #563). Fixing it here on the OSS source keeps both copies byte-identical **and** 88-col-clean.

## What

- Reflow the one offending line (`--timeout-keep-alive` arg) to the 88-col form via `ruff format --isolated` — the width both repos' `common/` gates use.
- Document the ≤88-col vendoring constraint in the module docstring so it isn't reintroduced.

No behaviour change. `tests/test_serve.py` unchanged and green; `ruff check` clean; still valid at the repo's 110-col config too.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -147,7 +147,7 @@ jobs:
         # docstring). ``--isolated`` pins default behavior even if a root ruff
         # config is ever added; deliberately NOT all of common/ — non-vendored
         # files there are not held to the 88-col contract.
-        run: uv run ruff format --check --isolated common/__init__.py common/structlog_config.py
+        run: uv run ruff format --check --isolated common/__init__.py common/structlog_config.py common/serve.py
 
       - name: Guard against raw-string MCP error returns (audit B4 / CAURA-000 #147)
         # MCP tool handlers must build error envelopes via ``_error_response(...)``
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
Addressed the High: added `common/serve.py` to the "vendored common/ files at ruff defaults" `--isolated` format-check step in `ci.yml`. This is required for enterprise's `check_vendored_drift` coverage invariant (every identical-policy file must be gated by this OSS step) — good catch, it would have blocked the enterprise re-vendor. @claude re-review please.

Eldad-Caura:
@erni-a please apporve

erni-a (approved):

PR #667, opened 2026-07-29
feat(interviewer): async submit — persist window, accept fast, synthesize in the scheduler sweep

### LINKED ISSUE #665: Interviewer submit 504s (gateway timeout) on highest-volume agents — inline synthesis exceeds proxy timeout
## Summary

On the highest-volume agents, the Interviewer's `interview_request` submit **times out at the gateway (504)** and produces no report, while normal-volume agents succeed. The interview submit → server-side synthesis round-trip for a large task-trail window exceeds the gateway (caddy) proxy timeout.

Surfaced immediately after plugin **2.16.0** / onprem-**v2.11.12** (backend-v2.25.0) armed the #658 task-trail capture on eToro prod (2026-07-29).

## Evidence (eToro prod, 2026-07-29)

`oc-webclaw-prod` — the single most active agent on the fleet (~360+ memories / 3h), on plugin 2.16.0, armed (`MEMCLAW_INTERVIEWER` + `MEMCLAW_INTERVIEWER_TASKS`):

| tick | result |
|---|---|
| 07:00 | `done` — `{"ok":true,"reason":"no new events since cursor","submitted":false}` (cursor init) |
| 08:00 | **`failed`** — `{"error":"MemClaw API 504: <html>…504 Gateway Time-out…"}` |

Contrast — same 08:00 tick, other armed agents on 2.16.0 succeeded:
- `devopsclaw` — `{"ok":true,"events":200,"submitted":true,"watermark":199}` → 17 interviewer memories
- `cmoclaw` → 49 interviewer memories; `main-<install>` → 63

So the loop works; only the **highest-volume** agent 504s. webclaw's task-trail window is large enough (up to the ≤500-event / ≤150k-char cap) that the submit + inline synthesis exceeds the gateway timeout.

## Likely root cause

The `interview_request` handler submits the window to the MemClaw API and the server synthesizes the report **inline** on that request. For a max-size window the LLM synthesis runs long, and the **caddy/gateway proxy timeout** cuts the connection → the plugin receives a 504 and marks the tick failed (watermark not advanced, so it will retry the same oversized window every hour and keep 504-ing).

## Proposed fixes (owner decision)

1. **Async submit** (preferred): submit returns quickly after persisting events; synthesis runs in the worker; the plugin advances its watermark on the 2xx accept, not on synthesis completion. Removes the long request from the gateway path entirely.
2. **Raise the gateway timeout** for the interview submit/synthesis route specifically (caddy `transport`/`timeouts` or the enterprise gateway route config) to comfortably exceed the interview LLM budget.
3. **Shrink the effective window for high-volume agents** — lower the per-tick event/char cap (or chunk the submit) so any single synthesis stays under the timeout.

Fix (1) or (2) also stops the **retry-storm**: a 504 leaves the watermark unadvanced, so the same oversized window re-submits every hour and 504s again indefinitely.

## Repro

1. Arm a very-high-activity agent (plugin ≥2.16.0, `MEMCLAW_INTERVIEWER`/`_TASKS` on, tenant `interviewer.enabled`).
2. Let it accumulate a near-cap task-trail window (≤500 events / ≤150k chars).
3. Observe the hourly `interview_request` return `MemClaw API 504: Gateway Time-out`; watermark does not advance; it repeats next tick.

## Not this issue

This is **not** the #654 capture gap (fixed by #658 — task-trail capture is working; most agents now produce reports). This is a downstream latency/volume limit on the submit→synthesis path for the busiest agents.

ran-taig:
Fixes #665 (durable track; the immediate on-prem gateway timeout bump is already live on eToro).

## Problem
`POST /interview/submit` ran 60–90s of inline LLM synthesis on the request. Intermediate proxies (on-prem nginx default `proxy_read_timeout` 60s) 504'd the plugin mid-flight **while the server completed and committed anyway** — ticks misreported `failed`, buffers never pruned, and the busiest agents' reports appeared "missing" (see the corrected diagnosis on #665: webclaw's 63 memories had committed at 08:01:45 under the install-default agent).

## Design
Default path (`interview_async_submit: bool = True`, escape hatch back to inline):
1. Validate exactly as before → **mask** → persist the window as a durable `interview_jobs` doc (doc id = `job_<sha1(node:from:to)>` — duplicate submits upsert one job).
2. **Advance the forward-only watermark and return 200 `accepted` immediately** — the events are durable server-side, so the node may safely prune on this response.
3. Synthesis runs off-request: a fire-and-forget task per submit (strong-ref'd) plus the hourly scheduler sweep (`process_pending_interview_jobs`, merged into `run_interview_schedule`) as the durable retry path — attempts capped (`interview_job_max_attempts=3` → `failed_permanent`), and the deterministic bulk attempt id keeps every retry idempotent.

**Wire-compatible with deployed plugins**: they advance/prune on any 2xx with a numeric watermark and don't gate on `status`. Legacy inline path preserved verbatim behind the flag (`run_interview` refactored into mask + `_synthesize_and_write(advance_watermark_after=…)`, byte-identical behavior).

## Tests
- New `tests/test_interview_async_submit.py` (10 tests): accepted+watermark+job-doc, processor→memories→done, idempotent re-submit (duplicate_attempt dedup), failure→pending→retry→`failed_permanent` at cap, flag-off legacy behavior, sweep summary.
- `tests/test_api_interview.py`: 22/22 unchanged (conftest pins `INTERVIEW_ASYNC_SUBMIT=false` for legacy tests; the new file opts in per-test).
- Full root suite: 3985 passed; 12 failures verified pre-existing (identical with this diff stashed).
- Ruff clean.

## Review pointers
- `query_documents` uses `where: {"status": "pending"}` (same JSONB pattern as `skills_inbox.py`).
- Done job docs retain their **masked** events (no cleanup yet — a future prune could shrink the collection).

### DIFF
--- a/core-api/src/core_api/config.py
+++ b/core-api/src/core_api/config.py
@@ -139,6 +139,18 @@ class Settings(BaseSettings):
     # raise the platform timeout BEFORE raising this budget (the
     # startup validator enforces the ceiling).
     interview_request_timeout_seconds: float = 90.0
+    # Async interview submit (#665). When True (default), the submit route
+    # persists the masked window as a durable ``interview_jobs`` doc,
+    # advances the watermark, and returns 200 ``accepted`` immediately;
+    # synthesis runs off the request path (fire-and-forget task + the
+    # hourly scheduler sweep). False is the escape hatch back to the
+    # legacy inline path, whose 60-90s synthesis intermediate proxies
+    # (on-prem nginx default 60s) 504'd mid-flight while the server
+    # committed anyway.
+    interview_async_submit: bool = True
+    # Max synthesis attempts per persisted interview job before it is
+    # parked as ``failed_permanent`` instead of retried by the sweep (#665).
+    interview_job_max_attempts: int = 3
     # Per-phase cap on the storage roundtrip inside
     # ``create_memories_bulk`` (CAURA-599). Embedding and enrichment
     # already enforce their own 30s caps; storage was the only phase
--- a/core-api/src/core_api/routes/interview.py
+++ b/core-api/src/core_api/routes/interview.py
@@ -2,9 +2,16 @@
 
 ``POST /api/v1/interview/submit`` receives one node's buffered event
 window from the OpenClaw plugin (delivered in response to an
-``interview_request`` fleet command) and hands it to the interview worker
-(``services/interview_service.py``): mask → chunked interview → typed
-memories via the idempotent bulk path → forward-only watermark.
+``interview_request`` fleet command).
+
+Default path (``interview_async_submit``, #665): persist-and-accept — the
+window is masked and stored as a durable ``interview_jobs`` doc, the
+forward-only watermark advances, and the route replies 200 ``accepted``
+immediately; synthesis (chunked interview → typed memories via the
+idempotent bulk path) runs off the request in a fire-and-forget task plus
+the hourly scheduler sweep. The legacy inline path (flag off) runs the
+full worker on the request: mask → chunked interview → typed memories →
+forward-only watermark.
 
 Dark by default: the per-tenant ``interviewer.enabled`` flag gates the
 endpoint (the scheduler also never queues commands for disabled tenants —
@@ -14,6 +21,7 @@ this check is defense in depth, mirroring the skills_factory pattern).
 from __future__ import annotations
 
 import asyncio
+import logging
 from datetime import datetime
 
 from fastapi import APIRouter, Depends, HTTPException
@@ -23,10 +31,53 @@ from pydantic import BaseModel, Field
 from core_api.auth import AuthContext, get_auth_context
 from core_api.config import settings as app_settings
 from core_api.constants import INTERVIEW_EVENT_MAX_CHARS, INTERVIEW_MAX_EVENTS_PER_SUBMIT
-from core_api.services.interview_service import run_interview, run_interview_schedule
+from core_api.services.interview_service import (
+    InterviewJobPermanentlyFailedError,
+    advance_watermark,
+    enqueue_interview_job,
+    process_interview_job,
+    run_interview,
+    run_interview_schedule,
+    synthesis_sem,
+)
 from core_api.services.organization_settings import get_settings_for_display
 
+logger = logging.getLogger(__name__)
+
 router = APIRouter(tags=["Interview"])
+
+# Strong references to the fire-and-forget per-submit synthesis tasks —
+# asyncio only holds weak refs to tasks, so without this set the GC could
+# cancel one mid-synthesis (#665). The scheduler sweep remains the durable
+# retry path if the process dies anyway.
+_inflight_jobs: set[asyncio.Task] = set()
+
+
+# Bound the route's fire-and-forget synthesis like the scheduler sweep bounds
+# its own: without this, a burst of simultaneous submits spawns an unbounded
+# number of concurrent LLM map-reduce tasks (rate-quota + storage-pool
+# exhaustion). Queued tasks just wait — jobs are durable and the sweep is the
+# backstop either way.
+async def _bounded_process(tenant_id: str, doc_id: str) -> None:
+    # Shares interview_service.synthesis_sem with the scheduler sweep so the
+    # global synthesis cap holds even when both paths run concurrently.
+    async with synthesis_sem:
+        await process_interview_job(tenant_id, doc_id)
+
+
+def _log_task_exc(task: asyncio.Task) -> None:
+    """Surface fire-and-forget synthesis failures in the logs (#667):
+    without a done-callback retrieving the exception, asyncio defers the
+    'Task exception was never retrieved' report to GC time (and drops it
+    entirely on shutdown). ``process_interview_job`` never raises ordinary
+    exceptions, so anything landing here is a genuine bug — log loudly.
+    ``cancelled()`` is guarded first: calling ``exception()`` on a
+    cancelled task raises ``CancelledError``."""
+    if task.cancelled():
+        return
+    exc = task.exception()
+    if exc is not None:
+        logger.error("interview submit: fire-and-forget synthesis task crashed", exc_info=exc)
 
 
 class InterviewEventIn(BaseModel):
@@ -58,7 +109,7 @@ class InterviewSubmitIn(BaseModel):
 
 
 class InterviewSubmitOut(BaseModel):
-    status: str  # committed | partial | failed
+    status: str  # accepted | committed | partial | failed
     watermark: int | None
     memories_written: int
     errors: int
@@ -71,10 +122,17 @@ async def submit_interview(
 ):
     """Interview one node's event window and persist the report as memories.
 
+    Default (``interview_async_submit``, #665): persist the masked window
+    durably, advance the watermark, and return 200 ``accepted`` fast —
+    synthesis runs off the request path. Flag off: legacy inline synthesis
+    returning ``committed``/``partial``/``failed``.
+
     Idempotent per (node, window): the worker derives the bulk attempt id
     from ``sha1(node_id:cursor_from:cursor_to)`` server-side, so any retry
     of the same window resolves to ``duplicate_attempt`` rows and a
-    forward-only watermark — never duplicates, never a gap.
+    forward-only watermark — never duplicates, never a gap (the async job
+    doc id is derived from the same identity, so duplicate submits upsert
+    one job).
     """
     auth.enforce_read_only()
     auth.enforce_usage_limits()
@@ -101,6 +159,50 @@ async def submit_interview(
         # Defense in depth: the scheduler shouldn't have queued a command
         # for a disabled tenant; refuse rather than silently ingest.
         raise HTTPException(status_code=403, detail="interviewer is not enabled for this tenant")
+
+    if app_settings.interview_async_submit:
+        # Persist-and-accept (#665). The 60-90s inline synthesis outlived
+        # intermediate proxy budgets (on-prem nginx defaults to 60s), so
+        # plugins saw a 504 while the server committed anyway — misreported
+        # failure, never pruned. Order matters: the masked window is durable
+        # server-side BEFORE the watermark advances and the 2xx goes out, so
+        # the node may safely prune its buffer on this response.
+        try:
+            doc_id = await enqueue_interview_job(
+                tenant_id=tenant_id,
+                fleet_id=body.fleet_id,
+                agent_id=body.agent_id,
+                node_id=body.node_id,
+                command_id=body.command_id,
+                cursor_from=body.cursor_from,
+                cursor_to=body.cursor_to,
+                events=[ev.model_dump(mode="json") for ev in body.events],
+            )
+        except InterviewJobPermanentlyFailedError:
+            # 409 (not 500): the window is parked after exhausting its retry
+            # budget — permanence deserves a distinct status in access logs
+            # and lets a future-smarter plugin stop resubmitting.
+            raise HTTPException(
+                status_code=409,
+                detail="interview job is permanently failed — operator intervention required",
+            )
+        watermark = await advance_watermark(
+            tenant_id,
+            node_id=body.node_id,
+            agent_id=body.agent_id,
+            cursor_to=body.cursor_to,
+            command_id=body.command_id,
+        )
+        # Best-effort immediate synthesis WITHOUT holding the request; the
+        # scheduler sweep (process_pending_interview_jobs) is the durable
+        # retry path if this task dies with the process.
+        task = asyncio.create_task(_bounded_process(tenant_id, doc_id))
+        _inflight_jobs.add(task)
+        task.add_done_callback(_inflight_jobs.discard)
+        task.add_done_callback(_log_task_exc)
+        # Plain 200: deployed plugins advance/prune on any 2xx with a
+        # numeric watermark and do not gate on ``status`` — wire-compatible.
+        return InterviewSubmitOut(status="accepted", watermark=watermark, memories_written=0, errors=0)
 
     # Route-enforced deadline (the path opts out of the blanket 45s
     # middleware, which 504'd every realistic window — the synchronous
--- a/core-api/src/core_api/services/interview_service.py
+++ b/core-api/src/core_api/services/interview_service.py
@@ -23,6 +23,7 @@ Design notes (see ``docs/plans/interviewer-phase1-decisions.md``):
 
 from __future__ import annotations
 
+import asyncio
 import hashlib
 import logging
 from datetime import UTC, datetime
@@ -30,6 +31,7 @@ from typing import Any
 
 from common.governance import mask, scan
 from core_api.clients.storage_client import get_storage_client
+from core_api.config import settings as app_settings
 from core_api.constants import (
     INTERVIEW_CHUNK_MAX_CHARS,
     INTERVIEW_EVENT_MAX_CHARS,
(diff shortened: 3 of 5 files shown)


ran-taig:
@claude

ran-taig:
@claude

ran-taig:
@claude Please re-review — two fixes amended in (32fcb9f1): (1) process_interview_job now wraps synthesis+finalize in a BaseException guard — a cancelled task best-effort reverts the job to "pending" before re-raising, so cancellation can no longer strand a "processing" job the sweep (pending-only) would never retry; (2) enqueue_interview_job preserves the existing doc's attempts counter on duplicate upserts instead of resetting to 0, so plugin resubmits can't cycle a failing job past interview_job_max_attempts. Also: the CodeQL high alert was dismissed as false-positive with rationale (sha1 as deterministic identifier in a frozen wire contract, not security hashing). Gate: 32/32 interview tests, ruff clean.

ran-taig:
@claude Please re-review — round 2 amended in (4a341228). All four findings applied, one with a load-bearing deviation: (1) stale-processing recovery — processing_started_at stamped on the processing transition; the sweep also fetches processing-status jobs and recovers those stale >600s (client-side filter — the documents where-query is scalar-equality only); missing/unparseable timestamps count as stale so pre-fix stranded docs recover too. (2) no-downgrade enqueue — done/processing jobs skip the upsert entirely; AND _set_state now re-fetches the current doc and merges transition keys onto the fresh snapshot: the prescribed write-only-transitioned-keys shape would have DESTROYED the events payload, because document_upsert is REPLACE (ON CONFLICT DO UPDATE SET data = :data), not merge. (3) convergence-to-pending via a wrote_done flag + finally with two best-effort reset attempts — every non-done exit (synthesis failure, done-write failure, cancellation) ends pending; CancelledError still propagates. (4) prior-attempts fetch failure now logger.warning, not silent. Tests 36/36 (4 new: stale-vs-fresh recovery, no-downgrade x2, reset-retry-survives-write-failure, attempts preserved across re-enqueue); one pre-existing test rewritten because it asserted the old done→pending downgrade. Ruff clean.

ran-taig:
@claude Please re-review — round 3 amended in (d971a3aa), all three findings applied as prescribed: (1) enqueue's prior-read Exception path now returns early with an explicit skip-upsert warning — unknown state is never overwritten (the narrower TypeError/ValueError parse branch still falls through by design, prior_status is already captured there); (2) _set_state gained increment_attempts=True and computes the increment from the FRESH base, so concurrent processors land 1→2 instead of 1→1; the pending/done/finally writes intentionally keep passing the local attempts — they conclude an already-counted attempt, and dropping the key would silently regress the counter in _set_state's fetch-fallback path; (3) processing-status query over-fetches at limit*4 (stale filter is client-side, sorted by created_at not processing_started_at). Tests 38/38 (2 new: read-failure-over-done-job leaves the doc byte-identical; interleaved-processors attempts spy asserts [1,2]); ruff clean.

ran-taig:
@claude Please re-review — round 4 amended in (9565e4b2). Both findings applied, with one evidence-driven deviation: (A1) your prescribed plain guard ("processing" → return None) empirically BREAKS the round-2 stale-recovery feature — the sweep hands stale rows to process_interview_job while their stored status is still "processing", so the in-function guard can't tell the sweep's reclaim from a duplicate racer (test_sweep_recovers_stale_processing_but_not_fresh failed with the doc stranded). Implemented instead: allow_stale_processing opt-in that only the sweep sets for its stale rows, AND the function re-verifies staleness itself (flag alone cannot reclaim a fresh run). (A2) finally-reset now re-fetches and breaks on fresh done/failed_permanent, resets without the attempts key; (A3) attempts key removed from the done write; _set_state falls back to the last-successfully-written snapshot (not the pre-increment closure), so attempts-less writes can't regress the counter on the fetch-fallback path. (B) no conditional upsert exists in core-storage (both document_upsert variants are unconditional ON CONFLICT DO UPDATE) — applied the debug-log + honest race documentation fallback (worst case: attempts reset / brief re-open of a consumed window; re-synthesis dedups, cost only). Tests 40/40 (2 new: concurrent-done survives the reset; direct-processor skip/reclaim matrix). Ruff clean.

ran-taig:
@claude Please re-review — round 5 amended in (a74e3847). (CI) the ruff failure was ruff-format line-length: one call in process_interview_job was wrapped at 88 cols against core-api's 110-col config — joined; all six CI ruff steps replicated locally and pass. (High — applied) enqueue_interview_job's prior-read Exception handler now RE-RAISES after the warning: the round-4 early return silently accepted first-time submits whose job doc never persisted, letting the route advance the watermark and the plugin prune — permanent loss. Raising → route 500 → no prune, plugin retries. Route-level regression test asserts 500 + watermark doc never created + no job doc (via the repo's raise_app_exceptions=False client pattern); the round-4 test was rewritten to the new semantics (raises over an existing done doc, done doc byte-identical). (Medium — applied) the (TypeError, ValueError) guard now wraps only the attempts int-cast, so get_document errors reach the re-raising outer handler. Tests 41/41; CI-style ruff all pass. Note for a future chore: root tests/ has no ruff config so local format-check disagrees with core-api's 110-col style — pre-existing, out of scope.

ran-taig:
@claude Please re-review — round 6 amended in (375f1f96). All three applied: (1) _set_state gained skip_if_terminal — the terminal check runs on the SAME snapshot the merge writes, collapsing the reset path to a single fetch-write gap; on the fetch-fallback path the guard is best-effort by construction (the fallback base is our own last-written snapshot, which cannot show a concurrent terminal write) — the residual race under plain-replace upsert is documented in-code and remains idempotency-safe. (2) sweep restructured to collect-then-gather bounded by INTERVIEW_SWEEP_CONCURRENCY=5; verified process_interview_job cannot raise ordinary exceptions on any path, so gather is bare and only cancellation propagates (intentionally aborting the sweep); summary counting is order-insensitive. (3) _log_task_exc appended with the cancelled() guard — note your review said 'two existing callbacks': the actual count was ONE (_inflight_jobs.discard); also added the missing module logger to routes/interview.py. Tests 43/43 (2 new: single-fetch-guard pin asserting done-survives + exactly-one-post-failure-fetch; 7-job/2-tenant sweep with an in-flight gauge asserting max ≤5 and identical summary counts) — all 41 prior tests green unmodified. CI-style ruff check+format green. One honest note: a fails-on-old/passes-on-new differential for the reset race is impossible at this seam (a done injected after the single fetch is the irreducible plain-replace residual), so the new test pins the two properties the fix actually guarantees.

ran-taig:
@claude Please re-review — round 7 amended in (ec5b61a5), applied directly (no delegation): (1) the failed_permanent park write now passes skip_if_terminal=True — a stale-reclaim racing a slow-but-alive run that just wrote done can no longer falsely discard the completed synthesis (same guard as the finally reset); (2) the park branch returns the {status: failed_permanent} sentinel instead of None, and the sweep summary distinguishes jobs_parked from jobs_retried; (3) run_interview_schedule's merge now carries jobs_parked. One footnote on the (1)+(2) interaction: when skip_if_terminal actually skips the park because the concurrent run won with done, the sentinel still reports failed_permanent for THIS runner and the sweep counts it parked — per-runner semantics, as prescribed; the winning done was counted by the other path. Two tests updated to the new shapes (park sentinel; summary key set), all else untouched: 43/43, CI-style ruff check+format green.

ran-taig:
@claude Please re-review — round 8 amended in (6a495e89), applied directly: (1) the enqueue skip guard now also covers failed_permanent (warning-level log — operators should see a plugin re-hammering a window the server gave up on; re-opening it via this racy path would hand out a fresh attempts budget); (2) the non-atomic comment now documents the ACTUAL bound: a multi-worker resubmit race can exceed interview_job_max_attempts by at most (workers − 1) before every snapshot sees the high value, with the conditional-upsert (WHERE attempts <= :prior_attempts) named as the core-storage change that would close it; (3) after the processing transition, the local attempts adopts the actually-stored count from the last-written snapshot (fresh-base increment) instead of the stale initial read. Also: the CodeQL fail on this PR is the SAME sha1-as-identifier false positive re-raised because the amend relocated the line — re-dismissed with the standing rationale (deterministic attempt/job/watermark ids, frozen wire contract, not security hashing). Gate 43/43, CI-style ruff check+format green. Note on the earlier review-bot comment backlog: every finding in those seven bot reviews maps to an already-landed round (stuck-processing → r4 stale recovery; cancellation → r2/r4; enqueue fall-through data loss → r5 re-raise; finally-reset done-overwrite → r4+r6 skip_if_terminal; failed_permanent park race + jobs_parked metric → r7; attempts TOCTOU bound → this round).

ran-taig:
@claude Please re-review — round 9 amended in (ce7e949c). The processing-claim overwrite finding is valid and applied, with one deviation from the prescribed detection: your last_written.get("status") != "processing" check cannot detect a suppressed claim on the STALE-RECLAIM path (the initial snapshot's status is already "processing", so a skipped write leaves the check passing and synthesis proceeding). Implemented instead: _set_state now returns bool (True=wrote, False=skip_if_terminal suppressed), the processing transition passes skip_if_terminal=True, and a falsy claim returns None early inside the same try/except — a concurrent run that finished (or parked) the window between the initial fetch and the claim owns the result; no overwrite of done with processing, no re-synthesis whose failure path could park a correctly-completed window. attempts adoption from last_written happens only after a successful claim. All 43 existing tests exercise the True path and stay green unmodified; honest note: no NEW dedicated test for the suppressed-claim corner in this round — if you want it pinned, the injection seam from test_pending_reset_guard_uses_the_merge_base_fetch adapts directly and I'll add it next round. CI-style ruff check+format green.

ran-taig:
@claude Please re-review — round 10 amended in (8e0af102). Valid High, applied as prescribed: enqueue over a failed_permanent job now RAISES instead of returning doc_id — the round-8 skip guard combined with the route's watermark advance meant a plugin resubmit over a parked window got 200+watermark, pruned its buffer, and the window was silently lost with no re-synthesis path. Now: route 500s, watermark never advances, plugin keeps its buffer, and the failure is visible every tick until an operator intervenes (masked events remain in the job doc for recovery). The comment documents why re-opening the job is not an option (this racy path would hand out a fresh attempts budget). Gate 43/43, CI-style ruff check+format green. Honest note: no new dedicated test this round for the raise-over-parked path; it mirrors round-5's raise semantics whose tests cover the route-level 500/no-watermark behavior — happy to pin the parked variant next round if you want it.

ran-taig:
@claude Please re-review — round 11 amended in (bfdfbc44). All four applied: (High) the route's fire-and-forget synthesis is now bounded by a semaphore sized to INTERVIEW_SWEEP_CONCURRENCY via a _bounded_process wrapper — queued tasks just wait; jobs are durable and the sweep remains the backstop. (Medium) failed_permanent enqueue now raises the new typed InterviewJobPermanentlyFailedError, which the route maps to 409 — permanence gets a distinct status in access logs instead of blending into transient 500s, and a future-smarter plugin can stop resubmitting. (Medium) a double-check read runs immediately before the pending upsert (not a CAS — documented as narrowing, not closing, the non-atomic window); a failed re-check raises, preserving the round-10 invariant that enqueue returns normally only when the window is durably owned server-side. (Low) the sweep summary now splits jobs_skipped (None = no-op/concurrently-claimed) from jobs_retried (synthesis genuinely failed and reset), and run_interview_schedule merges the new key. Gate 43/43 (one summary key-set test updated), CI-style ruff check+format green (formatter reflowed one line post-edit; re-verified).

ran-taig:
@claude Please re-review — round 12 amended in (6208289c). Both applied: (1) synthesis exceptions now return a {status: failed_transient} sentinel instead of None — the job WAS claimed and attempts incremented, so the sweep counts it as jobs_retried; None stays reserved for true no-ops (missing/done doc, concurrent-ownership skip). The sentinel is explicitly excluded from the done-write branch so it converges to pending via the finally as before. Eight failure-path test assertions updated to the sentinel; the genuine no-op assertions kept is-None (one initially over-flipped and reverted — the direct-processor fresh-skip case). (2) run_interview_schedule zero-inits all five jobs_* keys before the sweep try-block, so the summary schema is identical whether the sweep succeeds or raises. Gate 43/43, CI-style ruff check+format green.

ran-taig:
@claude Please re-review — round 13 amended in (18fdb2b6). Both applied as prescribed: (1) the enqueue double-check now adopts the freshest attempts from the recheck snapshot before the upsert — a concurrent processor's increment between the two reads is no longer overwritten by the stale first-read value (the exact reset the double-check was added to narrow); (2) process_interview_job's front gate treats failed_permanent like done (terminal no-op, returns None) — an already-parked job is counted jobs_skipped by the sweep instead of being re-counted jobs_parked every pass. Gate 43/43 with zero test changes needed; CI-style ruff check+format green.

ran-taig:
@claude Please re-review — round 14 amended in (ddb2398c). Both applied as prescribed: (1) "partial" is no longer terminal in the async path — it's excluded from the done-write so the finally converges it to pending and the sweep re-drives it (idempotent: written rows dedup as duplicate_attempt, failed rows get a fresh attempt); the sweep counts only "committed" as jobs_done, so partial now lands in jobs_retried. A persistently-partial window is bounded by interview_job_max_attempts like any failing job. (2) the park write now respects _set_state's skip signal — when a concurrent run completed the job between fetch and park, the processor returns None (sweep counts jobs_skipped) instead of miscounting a completed window as failed_permanent. Gate 43/43 with zero test changes; CI-style ruff check+format green.

ran-taig:
@claude Please re-review — round 15 amended in (bf4d9a39). Applied as prescribed: one shared _synthesis_sem now lives in interview_service beside INTERVIEW_SWEEP_CONCURRENCY and gates BOTH synthesis entry points — the route's fire-and-forget _bounded_process imports it, and the sweep uses it instead of its local instance — so the global cap holds at N even when the hourly sweep overlaps in-flight route tasks (previously 2N). Module-level Semaphore is loop-lazy on the supported Python versions, so import-time construction is safe. Gate 43/43, CI-style ruff check+format green (one line reflowed by the formatter, re-verified).

ran-taig:
@claude Please re-review — round 16 amended in (bbb54acc). Applied as prescribed: pure rename _synthesis_sem → synthesis_sem across the service and the route import/usage — the semaphore is intentionally shared, so the private-marker cross-module coupling is gone (plus the import-sort ruff fix the rename triggered). Gate 43/43, CI-style ruff check+format green.

ran-taig:
## ✅ Wet-test verdict: PASS (GCP stack, branch @ bf4d9a39*)

Full pipeline suite + async-specific assertions against a real backend (fake LLM providers), unmodified plugin modules:

| Assertion | Result |
|---|---|
| Phase-1 pipeline suite through the **async default** (happy path, torn-line, no-prune-on-refusal, kill-9, 500-cap drain, dedup) | **25/25** — one harness assertion converted to a poll (accepted-fast semantics: memories land seconds after the tick) |
| Job-doc lifecycle on real runs | every job `done`, `attempts=1`, per-job `memories_written` recorded |
| Parked window resubmit | route returns **409** |
| Sweep drain (seeded pending jobs) | `jobs_processed:2, jobs_done:2` |
| Stale-processing reclaim (11-min-old claim) | reclaimed → `done`, attempts 1→2 (fresh-base increment), memories written |
| Escape hatch `INTERVIEW_ASYNC_SUBMIT=false` | inline `committed`, `memories_written:3`; reverted cleanly |

One false alarm during testing was **operator error, documented for honesty**: an initial stale-reclaim failure traced to my seeded `processing_started_at` being 80 minutes in the *future* (local-vs-UTC date slip) — a future timestamp is correctly non-stale; re-seeded properly, the reclaim passed first try.

\* VM ran the round-15 tip; rounds 16+ since are a symbol rename + metrics-only changes with no behavioral delta to these assertions.

ran-taig:
@erni-a

erni-a (approved):

PR #668, opened 2026-07-29
ci: adopt the org's shared Claude review pipeline

Eldad-Caura:
Sixth and last of the migrations, behind the merged pilot memclawd#106. Held back until
caura-ai/.github#3 landed, because this repo could not migrate behaviour-preservingly without it.

Net **93 insertions, 386 deletions** — a 98-line caller replaces 382 lines of workflow plus a
97-line `claude_pr_review.sh`.

### Why this repo waited

It is the org's only **public** repo, and its local pipeline carried a gate none of the private
five needed: a fail-closed runtime org-membership check. Its stated reason was that a review
consumes `ANTHROPIC_API_KEY` and burns real budget, so fork PRs must not be able to spend it.

The shared pipeline gated the *comment* paths on write access but left the `pull_request` path open
to any author, and had no input to add a gate. Copying the caller template as-is would have
silently dropped that protection on the one repo where an untrusted author can actually open a PR.
caura-ai/.github#3 added it to the shared workflow instead, so every future public repo inherits it.

**Nothing is passed for the gate here, on purpose.** The callee's `author-gate` defaults to `auto`,
which reads the repository's visibility and enables the gate on public repos. Deriving it means a
repo that later goes public is covered without anyone remembering to set a flag — the journey this
repo already made once.

### Behaviour-preserving — each claim checked, not assumed

| | |
|---|---|
| `EXTRA_PROMPT` | carried **byte-identically**, md5 `7ff0f841d062` both sides |
| `REVIEW_PROMPT` | **omitted** — byte-identical to the shared default (md5 `f6f0d288604e`) |
| `skip-paths-jq` | **omitted** — this repo's filter already *is* the org default (md5 `4336db30c5b8`) |
| CLI pin | still `2.1.159`, now via the callee's default |
| `branches` | `[main, dev]`, unchanged |
| job rename | this repo requires **no** status checks at all, so the rename cannot strand a merge |

The four gates with no knob — `synchronize`, draft PRs, `dependabot[bot]`/`renovate[bot]`, and
`max-files` at 30 — are all enforced by the shared `pre-checks`. It also **fixes a bug every local
copy had**: `/files` defaults to `per_page=30`, so a PR with 30 or more changed files reported
exactly 30, `30 -gt 30` was false, and the oversized-PR guard never fired for the PRs it existed
to catch.

### What to expect on this PR specifically

Because the membership gate is now active here, the automated review will **skip** on a PR whose
author is not a *public* member of the org — `GITHUB_TOKEN` can only resolve public membership, so
a maintainer with private org membership sees their own PR skipped. That is unchanged from this
repo's previous behaviour, where the auto-review never fired and a review had to be requested by
comment.

Separately, and as on the other five migrations: a comment-triggered run cannot work until this
merges. An `issue_comment` event runs the workflow from the **default branch** (still the
pre-migration structure, which invokes `.github/scripts/claude_pr_review.sh`) while checking out
**this** branch, where that script is deleted — so it exits 127. Not a defect; it resolves on merge.


Eldad-Caura:
Closing and reopening to run the migrated pipeline from the PR head.

Eldad-Caura:
## Blocked — do not merge. This cannot work as written, and merging it would remove review from this repo entirely.

The reopen run came back `failure` with **zero jobs**, named after the workflow file rather than
`Claude Code Review` — the signature of a workflow GitHub refused to load, not a job that ran and
failed.

**Root cause is a platform constraint, not this YAML.** `caura-ai/.github` is **private**; this is the
org's only **public** repo. GitHub's docs for the exact access setting that repo has
(`access_level: organization`, "Accessible from repositories in the organization") state:

> Access is allowed only from private repositories.

So a public repository cannot resolve a reusable workflow **or** a composite action stored in a
private repository, under any of the available settings. The five migrations that passed
(memclawd#106, enterprise#943, memclaw-onprem#38, caura-test-automation#236,
openclaw-fleet-tester#29, caura-ops#251) are all private repos, which is why the identical `uses:`
reference works there and only here.

Nothing in the caller fixes this — `actionlint` is clean and the callee needs only `extra-prompt`.
The `author-gate` work in caura-ai/.github#3 was necessary but not sufficient.

**Why merging would be actively harmful:** an unresolvable `uses:` means no run is created at all, so
this repo would go from "reviews on demand" to "no reviews, silently". This repo requires no status
checks, so nothing would have blocked the merge.

### The options, all requiring a decision above this PR

1. **Make `caura-ai/.github` public.** Standard for a `.github` repo and the smallest change — all six
   callers then resolve. It discloses the review prompt, the pipeline, and the MemClaw recall shape
   (the `memclaw.net` endpoint and the `code-review` fleet id). No secrets live in the repo; they are
   GitHub secrets. Per-repo blurbs live in the callers, so they are unaffected.
2. **Move the pipeline to a new public repo** and repoint all six callers. Same disclosure, more work,
   but keeps `caura-ai/.github` private for anything else it holds.
3. **Leave this repo on its local pipeline.** Keeps its gate and discloses nothing, at the cost of the
   one public repo staying un-migrated and free to drift — it is already behind on the model bump,
   the failure-logging fix, and the `per_page=100` fix.

Marked as draft so it cannot be merged by accident. Reopening is a one-liner once the call is made.

Eldad-Caura:
Closing. Not abandoned — blocked on a platform constraint this PR cannot route around.

`caura-memclaw` is **PUBLIC**; `caura-ai/.github` is **PRIVATE**. A public repo cannot consume a
reusable workflow or composite action from a private one: GitHub answers "Access is allowed only from
private repositories." The symptom is nasty — the run completes as `failure` with **zero jobs**, named
after the workflow file path, so it looks like a syntax error rather than a permissions wall.

The `v1.0.0` release did not change this. Visibility is a property of the repo, not the ref.

Merging as-is would have deleted this repo's working `claude_pr_review.sh` (-97) and replaced the
workflow with a caller that cannot resolve (+93/-289) — **silently removing code review from the
public repo**. That is why it sat as a draft rather than being merged and reverted.

The branch is kept, so this reopens cleanly if `caura-ai/.github` is ever published or a
pipeline-only public repo is created.

### What this repo runs today, and where it now differs

Its local pipeline still works, and it keeps the two controls that matter most here — the runtime
org-membership gate (checked against `orgs/caura-ai/members/{user}`, not the webhook-frozen
`author_association`) and the max-files cap. The CLI is pinned to the same `2.1.159`.

What it does **not** have, now that the shared pipeline has moved on:

| gap | consequence |
|---|---|
| **no `--max-budget-usd`** | a runaway review can bill without bound — the shared pipeline caps at $10 |
| no `--tools` allow-list, no `--bare` | reviewer keeps shell, file-write and network tools while holding `GH_TOKEN`; the membership gate blocks fork PRs, so this is narrowed, not closed |
| `claude-sonnet-4-6`, not `claude-sonnet-5` | $3/$15 per MTok against $2/$10 introductory — currently the *more* expensive of the two |
| no docs-only skip | pays for markdown- and yaml-only pull requests |
| thin failure diagnostics | posts "check workflow logs" without capturing the CLI's stdout — the exact defect `fail_review` was written to fix |

One thing it is **not** exposed to: the `nounset` bug fixed in `caura-ai/.github`. That was introduced
by the CLI smoke test, and this script's failure path never expands `${RESULT:0:2000}`, so there is
nothing to trip.

The spend ceiling is the gap worth acting on independently of the visibility question — it is five
lines, needs no shared pipeline, and this is the one public repo where an unbounded review is billed.
Tracked separately rather than left implied by a closed PR.

PR #670, opened 2026-07-29
fix(core-api): reserved-id guard as a storage-boundary chokepoint + episode-path regression test

ran-taig:
## Context
On eToro on-prem, bare-\`agent_id=\"main\"\` (the reserved firehose identity) episode writes were observed landing **despite \`RESERVED_AGENT_ID_POLICY=reject\`**, suggesting the reserved-id guard (#505) was bypassed by the episode-persist path.

## Investigation — no bypass
Traced **every** persist path at the deployed version: REST \`/memories\` (incl. the plugin's episode auto-persist, \`context-engine.ts\` → \`POST /memories memory_type=episode\`), \`/ingest\`, STM, crystallizer, interviewer — **all funnel through \`memory_service.create_memory\` / \`create_memories_bulk\`, both guarded.** No unguarded writer exists.

The 353 bare-\`main\` episodes leaked **only during a window when \`reject\` was transiently unset** (deploy/cutover churn). They **stopped the moment stable \`reject\` came up** (0 in the following 8h while the guard fired 350–560×/hr). Reproduced on erni: bare-\`main\` episode → **409 under reject**, **201 under warn**; \`main-<install_id>\` → **201** either way. So the guard already covers the episode path — the defect was reject **durability**, not a code bypass (handled separately, on-prem installer).

## This PR — defense-in-depth
Assert the same guard at **\`storage_client.create_memory\` / \`create_memories\`** — the single boundary **every** memory insert crosses — so no present or future caller can persist a bare-\`main\` row by reaching storage directly and skipping the service guard.
- Only exact \`ALWAYS_RESERVED_AGENT_IDS\` (\`{\"main\"}\`) reject.
- De-collapsed \`main-<install_id>\` (#507) and named ids pass untouched.
- Reuses \`enforce_reserved_write_id\` (lazy import; no new log path, no import cycle).

## Tests
- **\`tests/test_reserved_write_storage_chokepoint.py\`** (hermetic, \`_post\` stubbed): bare-\`main\` rejected at \`create_memory\` + \`create_memories\` under reject; \`main-<install_id>\`/named/\`mcp-agent\` pass; \`warn\` does not block.
- **\`tests/test_api_memories.py::test_reserved_main_episode_rejected_under_reject\`** (integration): \`POST /memories memory_type=episode agent_id=\"main\"\` → **409** under reject; \`main-<install_id>\` episode → **201**. Locks the plugin episode path to the guard so a refactor can't quietly reopen it.

### DIFF
--- a/core-api/src/core_api/clients/storage_client.py
+++ b/core-api/src/core_api/clients/storage_client.py
@@ -20,6 +20,30 @@ from core_api.config import settings
 logger = logging.getLogger(__name__)
 
 
+def _reject_reserved_write_id(agent_id: str | None) -> None:
+    """Storage-boundary reserved-id backstop. No-op under ``allow``/``warn``;
+    only under ``reject`` does it raise ``ReservedAgentIdError`` for a reserved
+    id (the service maps it to HTTP 409).
+
+    Gated on ``reject`` on purpose: this runs on the normal service→storage
+    path, and the service layer has *already* called ``enforce_reserved_write_id``
+    (which logs one ``reserved_agent_write`` event under warn/reject). Calling it
+    again here would emit a duplicate log on every write, inflating the counter
+    ops uses to decide when to flip warn→reject. Under ``reject`` the service
+    path 409s before reaching storage, so this only fires for a *direct-storage*
+    caller that skipped the service guard — exactly the bypass we defend against,
+    and the single log/raise there is correct. ``main-<install_id>`` and named
+    ids are never reserved, so they pass.
+    """
+    from core_api.config import settings
+
+    if settings.reserved_agent_id_policy != "reject":
+        return
+    from core_api.services.agent_identity import enforce_reserved_write_id
+
+    enforce_reserved_write_id(agent_id)
+
+
 # Retry policy (F5 + the 2026-06-11 connect-timeout incident) lives in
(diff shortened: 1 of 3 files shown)


ran-taig:
@claude

ran-taig:
@claude

ran-taig:
@erni-a

erni-a (dismissed):

ran-taig:
Addressed the double-log review (Option 1): `_reject_reserved_write_id` now no-ops under `allow`/`warn` and only raises under `reject`.

Rationale: on the normal service→storage path the service layer already calls `enforce_reserved_write_id` (which logs one `reserved_agent_write` under warn/reject); re-calling it here duplicated that log on every write and inflated the warn→reject counter. Gating on `reject` removes the duplicate while preserving the defensive hard-block: under `reject` the service path 409s before storage, so the backstop only fires for a direct-storage caller that skipped the service guard — where a single log/raise is correct. Docstring updated to reflect the no-op-under-allow/warn semantics. Existing tests unchanged and still valid (warn → proceeds; reject → raises before `_post`).

ran-taig:
@erni-a

erni-a (approved):

PR #673, opened 2026-07-30
ci(review): cap review spend with --max-budget-usd

Eldad-Caura:
This repo's review pipeline ran the CLI with **no spend ceiling**, so one runaway exploration could
bill without bound — and this is the org's one **public** repo, which is where that matters most.
Default is `$10.00` per review.

### Why the membership gate wasn't already enough

It stops a fork pull request from triggering a review at all, which is the right control for *who*
can spend. It does nothing about *how much* a legitimate member's large pull request costs. Sibling
repos on the org's shared pipeline have billed **$6.11 on a two-file diff**. The ceiling is the
missing half of that control, not a replacement for it.

It's a **runaway guard, not a budget target**: the CLI checks it between turns, so a run can overshoot
by about one turn, and on a single hung turn it never fires at all — the job timeout is the only bound
there. Sized above observed spend on purpose, because the expensive reviews are the ones that find
real defects; capping near the average would truncate exactly the runs worth paying for.

### The failure branch had to change with it

Otherwise the ceiling would have been the *least* diagnosable outcome in the script. It discarded the
CLI's stdout and posted "check workflow logs" above a log that held nothing — and budget exhaustion
exits `1` exactly like a crash, so hitting the ceiling would have looked like a broken pipeline.

It now logs the first 2000 chars of stdout (`VAR=$(cmd)` keeps it even when cmd fails, and claude
reports auth, quota and budget failures there rather than on stderr) and branches on the
machine-readable marker to name the ceiling and the spend.

### Verified, not assumed

**Flag support** — `--max-budget-usd` is in `claude --help` on `2.1.159`, the version this workflow
pins in both install steps.

**Validator**, driven through the real script under bash 5.2.21 (what `ubuntu-latest` runs):

```
accept:  10.00  10  .5  2.50
reject:  0  00  0.0  .0  0.00  .00  .  abc  10,00  $10  1.  -1  1e3
```

An empty value takes the `10.00` default — bash `:-` treats unset and empty alike, so that's intended,
not a hole.

**All three runtime paths**, with a stubbed CLI and `gh`:

| scenario | result |
|---|---|
| budget exhausted | posts *"reached the $7.50 spend ceiling after $7.61 without finishing"*, exit 1 |
| ordinary crash | posts the exit code, logs the payload to the job log |
| success | posts the review unchanged — happy path untouched |

(The $7.61-against-$7.50 in that first row is the documented between-turns overshoot, not a bug.)

### Two deliberate non-changes

`jq -e` rather than `grep -q` for the marker: grep stops reading on a match, the upstream `printf`
takes SIGPIPE, and `pipefail` then turns a **match** into a non-zero pipeline.

**No `RESULT=""` declaration**, unlike `caura-ai/.github`. That guard exists there because a CLI smoke
test calls the failure helper before `RESULT` is assigned, and under bash 5.2 `${RESULT:0:2000}` on an
unset variable aborts. Every `$RESULT` reference in this script follows the assignment, so there is
nothing for `nounset` to trip on — I checked rather than copying the fix across.

No workflow change: both invocation sites pass env explicitly, so the script-side default applies.

### What this does not do

It doesn't bring over the rest of the shared pipeline's hardening — `--bare`, the `--tools`
allow-list, the zero-exit `is_error` guard, or the docs-only skip. Those are a separate change, and
this repo can't consume the shared pipeline at all while it is public and `caura-ai/.github` is
private (see #668). This is the one gap worth closing independently of that.

### DIFF
--- a/.github/scripts/claude_pr_review.sh
+++ b/.github/scripts/claude_pr_review.sh
@@ -20,9 +20,34 @@
 #   REVIEW_PROMPT     review instructions + output format
 # Optional env:
 #   MODEL             model id (default: claude-sonnet-4-6)
+#   MAX_BUDGET_USD    per-review spend ceiling, passed to --max-budget-usd (default 10.00)
 set -euo pipefail
 
 MODEL="${MODEL:-claude-sonnet-4-6}"
+
+# A ceiling so one runaway review cannot bill without bound. The reviewer reads the repo
+# across turns to judge a diff, which is what makes it useful and also what makes an
+# unbounded run possible; sibling repos on the org's shared pipeline have billed $6.11 on a
+# two-file diff. The org-membership gate above stops a fork pull request from triggering a
+# review at all, but it does not bound what a member's large pull request costs.
+#
+# A runaway guard, not a budget target. The CLI checks the ceiling BETWEEN turns, so a run
+# can overshoot it by roughly one turn, and on a single hung turn it never fires at all —
+# the job timeout is the only bound there. Sized above observed spend on purpose: the
+# expensive reviews are the ones that find real defects, so capping near the average would
+# truncate exactly the runs worth paying for.
+MAX_BUDGET_USD="${MAX_BUDGET_USD:-10.00}"
+# Shape, then value. Shape does not require a leading digit, so `.50` is accepted the way the
+# CLI accepts it, while `.`, `1.`, `10,00`, `$10`, `-1` and `1e3` are rejected. Value is
+# checked arithmetically rather than with a second pattern: a zero ceiling is accepted by the
+# CLI and makes every review fail on its first turn — which reads as a broken pipeline rather
+# than a bad setting — and spelling "zero" as a regex means enumerating 0, 00, 0.0, .0, 0.00
+# and .00, where the no-leading-digit forms are the easy ones to miss.
+if ! [[ "$MAX_BUDGET_USD" =~ ^[0-9]*\.?[0-9]+$ ]] \
+   || ! awk -v v="$MAX_BUDGET_USD" 'BEGIN { exit !(v + 0 > 0) }'; then
+  echo "::error::MAX_BUDGET_USD must be a positive decimal number, got '${MAX_BUDGET_USD}'" >&2
+  exit 1
+fi
 
 post() { gh api "repos/${REPO}/issues/${PR_NUMBER}/comments" -f body="$1" >/dev/null; }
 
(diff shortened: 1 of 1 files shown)


Eldad-Caura:
### The ceiling is in, but this PR surfaced something bigger

`claude-review` **skipped** on this PR, and the reason is not the change:

> **Claude Code Review** — skipped: PR author 'Eldad-Caura' is not a public member of the 'caura-ai' org

The membership gate is working exactly as written. The precondition isn't met:

```
GET /orgs/caura-ai/public_members/Eldad-Caura  ->  404
GET /orgs/caura-ai/public_members/erni-a       ->  404
```

Both maintainers' org membership is **private**. Rule 1 calls
`orgs/caura-ai/members/{user}` with `GITHUB_TOKEN`, which resolves only *public* members — 204 for
visible, 404 for private-or-not-a-member. The same call with a personal token returns **204**, which
is why this is easy to miss: it looks fine when you check it by hand.

The consequence is that **reviews on this repo are being skipped for every human PR**, not just this
one. Spot-checking recent history, #524 skipped for `erni-a` and #672 for `caura-deploy-bot[bot]`, both
with the same reason. The gate hasn't been protecting spend — it's been suppressing the reviews.

The workflow's own comment predicted it: *"Members must still make their org membership public."*

### Fix, in order of preference

1. **Make the membership public** — Organization → People → your row → *Public*. One toggle per
   person, no code, and it keeps the gate's fail-closed property intact. Each member has to do their
   own; nobody can set it for them.
2. Fall back to `author_association` in `MEMBER`/`OWNER` when the API says 404. Weaker: the comment
   above rule 1 documents that the payload field is frozen at PR-creation time, which is why the
   runtime call was chosen.
3. A `read:org` PAT so private members resolve. I'd avoid it — a long-lived token in a public repo's
   workflow is a worse trade than making a membership public.

Only (1) is yours to do, and it also matters beyond this repo: the shared pipeline's `author-gate`
defaults to `auto`, which reads repository visibility. Every migrated repo is private, so it is
currently off and this cannot bite there — but if any of them is ever made public, `auto` turns the
gate **on** and it would silently skip everything for the same reason.

### On this PR

The ceiling itself is unaffected and the change stands on its own — `CI`, `CodeQL`, `dco`, and the
security scan are all green, and I verified all three runtime paths against a stubbed CLI rather than
relying on a live review. But I want to be straight about it: **this repo cannot currently give this
change a live review**, so the verification behind it is the stubbed harness and the bash 5.2 validator
table in the PR body, not a run of the real thing. Making the membership public and commenting
`@claude` here would produce that live run.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #577, opened 2026-07-19
feat(clients): send a User-Agent header identifying client and version

Souptik96:
## Summary

The Python and TypeScript clients now send a fixed `User-Agent` header on every
request — `memclaw-client-python/<version>` and `memclaw-client-ts/<version>` —
so the server can attribute traffic to a specific client and version. The version
is single-sourced from each package's own metadata so it can't drift.

## Related Issue

Closes #553

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that changes existing behavior)
- [ ] Documentation update
- [ ] Refactor / internal cleanup
- [ ] Other:

## How Has This Been Tested?

**Python** (`clients/python`, venv + `pip install -e ".[dev]"`, Python 3.13, ruff 0.15.22, pytest 9.1.1):

```
$ ruff check .
All checks passed!

$ ruff format --check .
6 files already formatted

$ pytest tests/ -v
...
tests/test_client.py::test_user_agent_header_is_sent PASSED               [ 91%]
tests/test_client.py::test_version_falls_back_when_package_not_installed PASSED [100%]
============================= 12 passed in 0.17s ==============================
```

**TypeScript** (`clients/typescript`, `npm install`, Node v20.18.1, TS ^5.4):

```
$ npx tsc          # strict NodeNext, zero errors
(exit 0)

$ npm test         # tsc && node --test dist/index.test.js
# tests 10
# pass 10
# fail 0
ok 10 - sends a versioned User-Agent header identifying the client
```

New tests added: Python asserts the outgoing request carries
`User-Agent: memclaw-client-python/<version>` and that the version read falls back
to `0.0.0+dev` when distribution metadata is missing (mocked
`PackageNotFoundError`); TypeScript asserts the stub `fetch` receives a
`memclaw-client-ts/<version>` User-Agent.

## Checklist

- [x] I have read [CONTRIBUTING.md](../CONTRIBUTING.md)
- [x] I have added tests that cover my changes (or explained why none are needed)
- [x] `ruff check` and `ruff format --check` pass
- [ ] `mypy` passes — N/A: `mypy` isn't part of the client packages' tooling; CONTRIBUTING's `mypy` target is the backend (`core-api/src/ core-storage-api/src/`), which this PR doesn't touch.
- [x] `pytest` passes locally
[...]

PR #675, opened 2026-07-30
ci(review): gate the on-open review on author_association, not the members API

Eldad-Caura:
**The on-open review has been skipping every human pull request in this repo.** Reviews only ever ran
when someone commented `@claude`.

### The bug

Rule 1 called `orgs/caura-ai/members/{user}` at runtime. That endpoint answers **for its caller**, and
`GITHUB_TOKEN` sees only *public* members — it returns 404 for a private member exactly as for a
stranger. Every maintainer's caura-ai membership is private:

```
GET /orgs/caura-ai/public_members/Eldad-Caura  ->  404
GET /orgs/caura-ai/public_members/erni-a       ->  404
```

So the gate skipped, and said so on the PRs: #673 *"PR author 'Eldad-Caura' is not a public member"*,
#524 the same for `erni-a`, #672 for `caura-deploy-bot[bot]`.

This is easy to miss two ways. Checking the endpoint by hand returns **204**, because a personal token
can always see your own membership. And the `@claude` path kept working throughout — it already read
`author_association` — so the pipeline never looked broken.

### The comment that justified it was wrong

It claimed a private member's `author_association` surfaces as `NONE`. Live payloads from this repo:

| PR | author | `author_association` |
|---|---|---|
| #673 | Eldad-Caura (private member) | **MEMBER** |
| #524 | erni-a (private member) | **MEMBER** |
| #577 | Souptik96 (outsider) | FIRST_TIME_CONTRIBUTOR |

Membership *visibility* does not affect the field, and it still discriminates outsiders correctly. So
requiring public org membership to get a code review was never a deliberate policy — it fell out of
that wrong belief.

### The fix

Rule 1 now reads `author_association`, accepting `MEMBER` and `OWNER` — the same field and the same
values `claude-retrigger` already uses. One mechanism for one policy; two ways of deciding "is this an
org member" is precisely the drift this pipeline exists to remove.

**Behaviour otherwise preserved.** `COLLABORATOR` stays excluded (the policy is "our org", stricter
than repo write access). Bots aren't `MEMBER`, so they stay skipped regardless of rule 4. An absent
field skips, so the gate remains fail-closed.

**The trade, stated:** `author_association` is computed when the event fires, not when the review runs,
so a member removed from the org in between is still honoured — reopening recomputes it. The
alternative reflected review time but didn't work at all, and retrigger already accepts this trade.

### Verified

Gate behaviour across all nine values GitHub can send:

```
MEMBER, OWNER                                          -> review
COLLABORATOR, CONTRIBUTOR, FIRST_TIME_CONTRIBUTOR,
FIRST_TIMER, NONE, MANNEQUIN, unset                    -> skip
```

Full-repo `actionlint` output is byte-identical to `main` — five pre-existing findings, none
introduced. (The `HTTP_CODE` warning moves from line 266 to 292 only because my comment shifted the
line numbers.)

### Also in here

Three comments repeated the `NONE` claim — the file header, rule 1, and the retrigger step. All
corrected, because the false claim is what produced the bug; leaving it in place would invite the same
"fix" again. Drops the now-unused `ORG` env var.

### One thing this does not fix

`caura-ai/.github`'s shared pipeline has the **same** API-based `author-gate`. It is dormant today —
`auto` disables it on private repos and all six consumers are private — but making any of them public
would reproduce this exact silent skip. Its input description is at least honest about the limitation
("an org member visible to GITHUB_TOKEN — in practice a PUBLIC member"), so it is a latent trap rather
than an active bug. Worth the same change, separately.

### DIFF
--- a/.github/workflows/claude_code_review.yml
+++ b/.github/workflows/claude_code_review.yml
@@ -1,7 +1,7 @@
 # Claude Code Review
 #
-# Runs automatically when a PR is opened/reopened by a public member of
-# the ``caura-ai`` organization. Re-trigger after pushing changes by
+# Runs automatically when a PR is opened/reopened by a member of the
+# ``caura-ai`` organization. Re-trigger after pushing changes by
 # commenting "@claude" on the PR (also restricted to org members).
 #
 # Why org-only: this workflow consumes ``ANTHROPIC_API_KEY`` and burns
@@ -10,13 +10,25 @@
 # members makes the policy explicit and stops fork PRs from triggering
 # silent-failure runs.
 #
-# Membership is checked at runtime against ``orgs/caura-ai/members/{user}``
-# (returns 204 for visible members, 404 otherwise). The ``pull_request``
-# event's ``author_association`` field is captured at PR-creation time
-# and never refreshes, so a member whose visibility was PRIVATE when
-# they opened a PR would be permanently locked out — the live API call
-# avoids that. Members must still make their org membership public for
-# the API to return 204.
+# Membership is read from the event payload's ``author_association``,
+# accepting ``MEMBER`` and ``OWNER``. Both the on-open review and the
+# ``@claude`` retrigger use that one field, so the two paths cannot
+# disagree about who counts as a member.
+#
+# This is a correction. The gate previously called
+# ``orgs/caura-ai/members/{user}`` at runtime, on the belief that
+# ``author_association`` reported ``NONE`` for members whose org
+# membership is private. It does not — a private member reports
+# ``MEMBER``, verified on live payloads here. The API call, meanwhile,
+# answers for its caller, and ``GITHUB_TOKEN`` sees only PUBLIC members:
+# it returned 404 for every maintainer, so the on-open review was skipped
+# on every human pull request while retriggers kept working. Requiring
+# public org membership to get a code review was not a deliberate policy;
+# it was a consequence of that wrong belief.
+#
+# What the payload field genuinely costs: it is computed when the event
+# fires, not when the review runs, so a member removed from the org in
+# between is still honoured. Reopening a pull request recomputes it.
 
 name: Claude Code Review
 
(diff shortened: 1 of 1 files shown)


Eldad-Caura:
**Converted to draft — this fix does not work, and CI caught it on its own PR.**

The gate ran and skipped, with the new reason string:

> **Claude Code Review** — skipped: PR author 'Eldad-Caura' is not a caura-ai org member
> (author_association: **CONTRIBUTOR**)

Not `MEMBER`. So `author_association` does **not** report org membership for a private member on a
`pull_request` payload, and the premise of this PR is wrong.

### Where I went wrong

I verified `author_association` by reading the REST API:

```
gh api repos/caura-ai/caura-memclaw/pulls/673 --jq .author_association   ->  MEMBER
github.event.pull_request.author_association (in CI, #675)               ->  CONTRIBUTOR
```

Two different sources for the same field, same user, same repo — and the workflow reads the second.
The REST call is evaluated for the authenticated caller, and my token can see my own private
membership; the webhook payload is not. I verified against a context that isn't the one that runs,
which is the same error as testing shell behaviour on bash 3.2 when CI runs 5.2.

**The comment I deleted was substantively right.** It said the payload field doesn't reflect private
membership; it was wrong only about the exact value (`CONTRIBUTOR`, not `NONE`). I removed a correct
warning because I'd "disproved" it with the wrong instrument. That's worse than leaving it alone, and
the reason it was there is now obvious.

What does hold is that the `@claude` path works for private members — confirmed by the maintainer's
own experience, and that path gates on the `issue_comment` payload, not the `pull_request` one. So the
two event types genuinely differ here. I'm not going to assert why without evidence.

### Where that leaves the actual problem

The on-open review is still skipped for every human pull request, and neither the old gate nor this
one fixes it. Two viable routes:

1. **Make org membership public** (Organization → People → your row → *Public*). The original API gate
   then works as designed, with no code change at all. I argued against this earlier on the strength of
   the verification that just fell over — I withdraw that. The disclosure is a public members list and
   an org badge on your profile; the real risk is spear-phishing target-listing, and it is modest here
   because this repo is public and both maintainers already author and merge in it, so the affiliation
   is already in public commit history.
2. **Gate on repo permission instead** — `GET /repos/{owner}/{repo}/collaborators/{user}/permission`,
   which resolves access granted *via* org membership regardless of visibility, and which `GITHUB_TOKEN`
   can read for its own repo. Two caveats: it widens the policy from "org member" to "has write access
   here", which pulls in outside collaborators that this workflow deliberately excludes; and I would
   have to verify what `GITHUB_TOKEN` actually returns **in CI**, not locally, since that is exactly
   the distinction I just got wrong.

Accepting `CONTRIBUTOR` is not an option — on a public repo that would let anyone with a merged commit
trigger paid reviews.

I'd take (1): no code, no new failure mode, and it restores the gate the way it was designed. If you'd
rather not, say so and I'll build (2) with a CI probe first and no guessing.

The spend ceiling in #673 is unaffected by any of this and still stands on its own.

Eldad-Caura:
Closing. The premise was wrong and the actual fix has landed elsewhere — org membership visibility.
`orgs/caura-ai/members/{user}` now resolves for the maintainer who changed it, so the original gate
works as designed and no code change is needed.

Keeping this closed rather than reworked, because what it proposed was worse than what it replaced: it
deleted a warning that was substantively correct.

The record, so this is not rediscovered:

```
pull_request payload   github.event.pull_request.author_association  ->  CONTRIBUTOR  (private member)
issue_comment payload  github.event.comment.author_association       ->  MEMBER       (private member)
REST API               GET /repos/../pulls/N  .author_association    ->  MEMBER       (caller-dependent)
members API            GET /orgs/../members/U with GITHUB_TOKEN      ->  404          (private member)
```

Two traps in there. The `pull_request` and `issue_comment` payloads disagree for the same person, which
is why `@claude` kept working while the on-open review did not. And the REST API's value is computed
for the authenticated caller, so a personal token reports `MEMBER` where CI sees `CONTRIBUTOR` — it
cannot be used to validate a gate that reads the payload. That is the mistake that produced this PR.

The operational requirement is real and worth stating plainly: this gate requires **public** org
membership. A member whose org membership is private is skipped exactly like a stranger, silently, with
a green check.

PR #676, opened 2026-07-30
docs(review): record the measured membership values on the review gate

Eldad-Caura:
Comments only — but the wrong value in them cost a full round-trip, so it is worth writing down what
was actually measured.

The gate's comments claimed a private org member's `author_association` surfaces as `NONE`. It does
not. On the strength of "disproving" that, I proposed swapping the runtime membership API for the
payload field; CI then skipped that change's own pull request and reported the real value:
**`CONTRIBUTOR`** — unusable as a gate on a public repo, since anyone with a merged commit earns it.

### What was observed, same private member, same repo

```
github.event.pull_request.author_association    ->  CONTRIBUTOR
github.event.comment.author_association         ->  MEMBER
GET /repos/{repo}/pulls/{n} .author_association ->  MEMBER
```

Three things follow, and all three are now in the file:

- **The two payloads disagree.** That is why `@claude` retriggers kept working for private members
  while the on-open review skipped them — the retrigger path reads the comment payload.
- **The REST value is computed for whoever asks.** A personal token reports `MEMBER` where CI sees
  `CONTRIBUTOR`, so validating this gate with a local `gh` call proves nothing. That is precisely the
  mistake behind the bad proposal, and the comment now says so.
- **The operational requirement is stated at the top**, not implied in passing: a member whose org
  membership is private is skipped exactly like a stranger — silently, behind a green check — so each
  member has to set their own membership to *Public*.

The retrigger comment carried the same claim in the *opposite* direction: it warned that private
members appear as `NONE` on the very field that path depends on, when that is the one place the field
does resolve. Corrected, with a note not to "align" the two gates onto it.

### Verification

Zero non-comment lines changed. `actionlint` output identical to `main` (five pre-existing findings),
YAML parses.

This pull request is also the end-to-end check that the on-open review works now that membership is
public — if a verdict appears below from `claude-review` rather than a skip notice, the gate is fixed.

### Still outstanding

The gate requires *public* membership from **every** author, and not everyone on the team has flipped
it yet — those authors' pull requests will keep being skipped silently. Worth a nudge; nobody can
change it on someone else's behalf.

### DIFF
--- a/.github/workflows/claude_code_review.yml
+++ b/.github/workflows/claude_code_review.yml
@@ -10,13 +10,34 @@
 # members makes the policy explicit and stops fork PRs from triggering
 # silent-failure runs.
 #
-# Membership is checked at runtime against ``orgs/caura-ai/members/{user}``
-# (returns 204 for visible members, 404 otherwise). The ``pull_request``
-# event's ``author_association`` field is captured at PR-creation time
-# and never refreshes, so a member whose visibility was PRIVATE when
-# they opened a PR would be permanently locked out — the live API call
-# avoids that. Members must still make their org membership public for
-# the API to return 204.
+# OPERATIONAL REQUIREMENT: every member who wants their pull requests
+# reviewed must set their caura-ai org membership to **Public**
+# (Organization -> People -> their row -> Public). Nobody can set it on
+# their behalf. A member whose membership is private is skipped exactly
+# like a stranger — silently, behind a green check. This is not a
+# preference; it is what the gate below can actually see.
+#
+# Membership is checked at runtime against ``orgs/caura-ai/members/{user}``.
+# That endpoint answers for its CALLER, and ``GITHUB_TOKEN`` sees only
+# PUBLIC members: 204 for a public member, 404 for a private member and
+# for a non-member alike.
+#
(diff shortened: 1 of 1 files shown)


Eldad-Caura:
The gate is fixed, and this run is the proof. The skip reason moved:

```
before   skipped: PR author 'Eldad-Caura' is not a public member of the 'caura-ai' org   (rule 1)
now      skipped: no source files to review (only tests/docs/config)                     (rule 6)
```

Rule 1 runs first and returns immediately, so reaching rule 6 means membership now resolves — the
on-open review is no longer blocked. This pull request is yaml-only, so rule 6 correctly declines to
pay for it; that skip is the design working, not the bug.

No verdict will appear here for that reason. The next pull request touching a source file gets a real
review, which is the first time that has been true for a human-authored PR in this repo.

@erni-a ready for review — comments only, zero executable lines changed, `actionlint` identical to
`main`. Worth flagging that the requirement documented here applies per-author: any member whose
caura-ai membership is still private will keep being skipped at rule 1, silently, behind a green
check. It has to be set by each member on their own account.

erni-a (approved):

PR #677, opened 2026-07-30
feat(search): org flag to disable query-time entity/graph retrieval

ran-taig:
## Summary

Adds **`search.entity_retrieval`** (bool, default `true`) — one org setting that blocks entity lookup and graph search on the read path, so every read resolves through keyword/semantic search alone.

There was no way to turn entity retrieval off. `search.graph_retrieval` only gates graph *expansion* inside `ParallelEmbedAndEntityBoost`, and `ClassifyQuery` never consults it — so the `ENTITY_LOOKUP` short-circuit (which sets `skip_embedding` + `skip_scored_search` and bypasses vector scoring entirely) fired regardless. `graph_max_hops: 0` doesn't help either: the classifier still runs entity FTS, takes hop-0 seeds and short-circuits.

```bash
curl -X PUT "$API/api/v1/settings?tenant_id=$TENANT" -H "X-API-Key: $KEY" \
  -d '{"search": {"entity_retrieval": false}}'
```

## Related Issue

N/A

## Type of Change

- [ ] Documentation update
- [ ] Other:

## What changed

Entity retrieval enters the read path at **two independent points**, so a single gate isn't enough. Both are now gated on the same flag:

| File | Gate |
|---|---|
| `pipeline/steps/search/classify_query.py` | **1** — entity block skipped: no entity FTS, no graph expansion, no `ENTITY_LOOKUP` short-circuit. Query falls through to temporal → recent-context → keyword/semantic |
| `pipeline/steps/search/parallel_embed_entity_boost.py` | **2** — no `GRAPH_HOP_BOOST`, with a log reason distinct from the CAURA-698 over-broad decline so ops can tell a deliberate disable from the precision heuristic |
| `services/memory_service.py` | **3** — the deprecated `_search_memories_legacy` path honours it too, so the switch can't be silently a no-op if `_USE_PIPELINE_SEARCH` is ever flipped |
| `services/organization_settings.py` | `DEFAULT_SETTINGS` + `_LEAF_TYPES` + `ResolvedConfig.entity_retrieval` |
| `config.py` | `ENTITY_RETRIEVAL_ENABLED` env default — fleet-wide switch for on-prem without editing every tenant |
| `mcp_server.py`, `routes/memories.py` (×2), `services/recall_service.py` | `entity_retrieval=config.entity_retrieval`, alongside the existing `graph_expand` |

**Defaults are unblocked at every layer** — tenant unset → env default (`true`) → an absent `ctx.data` key means enabled, so internal callers and existing tests are unaffected.

**Read-side only.** Entity extraction, linking (`entity_linking.auto_entity_linking_enabled` is the write-side switch) and relation inference keep building the graph, so re-enabling needs no backfill. Precedence: `entity_retrieval=false` supersedes `graph_retrieval` and `graph_max_hops`, which only bound expansion depth *after* a match.

Gate 3 uses a `_no_entity_boost()` coroutine rather than branching around the `gather`, keeping the legacy path's cancel-on-error structure byte-identical.

## How Has This Been Tested?

```bash
python -m pytest tests -m unit -q          # 2181 passed
python -m pytest tests -m "not integration and not benchmark" -q
ruff check core-api/src/ && ruff format --check core-api/src/
cd core-api && uv run mypy src/            # 189 files, clean
```

- **2181 unit tests pass**; 3665 pass across all non-DB tests. The 5 remaining failures (`tests/test_events/test_factory.py` ×4, `tests/test_no_direct_db_outside_storage.py`) **pre-exist on `main`** — verified by stashing this change and re-running.
- 15 new tests in `tests/test_entity_retrieval_flag.py`: both gates, the legacy path, `TEMPORAL` / `RECENT_CONTEXT` / `KEYWORD` / `SEMANTIC` routing with the flag off, flag-on regression, absent-key default, and settings resolution + env fallback + type validation.
- Six MCP config stubs gained the new attribute (they construct `_FakeConfig` / `SimpleNamespace` configs by hand).
- Smoke: new `test_entity_retrieval_flag` in `scripts/smoke_test.py` — a memory reachable *only* through the graph (content shares no wording with the query; linked to a *related* entity) must be found with the flag on and absent with it off. The setting is restored in a `finally`.

## Checklist

- [x] `mypy` passes
- [x] `pytest` passes locally (pre-existing failures noted above)
- [x] No docs to update — org settings aren't enumerated in `docs/` or `README.md`; the flag is documented inline at its definition
- [x] `CHANGELOG.md` is generated by release-please from the Conventional Commit (`feat(search):`), so no manual entry

## Additional Notes

**Result sets legitimately change for entity queries when the flag is off.** `ENTITY_LOOKUP` bypasses the `min_similarity` floor (`vec_sim` is `None`, so `PostFilterResults` skips). With the flag off those rows pass through the cosine gate like everything else — intended semantics, and what makes the smoke assertion sound.

**Explicit entity APIs are deliberately untouched** — `GET /entities`, `GET /graph`, `memclaw_entity_get`. Those are direct entity fetches, not query-time retrieval. Happy to add a strict mode that closes them too if reviewers want it.

**Unrelated pre-existing bug spotted nearby:** `scripts/smoke_test.py` `test_settings` PUTs `{"graph_expand": False}`, which isn't a valid settings key — `_check_keys` 422s it, so that "Settings PUT: 200" check is failing on every smoke run. Left alone to keep this PR scoped; the fix is the nested `{"search": {"graph_retrieval": false}}` shape.

### DIFF
--- a/core-api/src/core_api/config.py
+++ b/core-api/src/core_api/config.py
@@ -90,6 +90,12 @@ class Settings(BaseSettings):
     gemini_api_key: str | None = None
     entity_extraction_provider: str = "openai"  # none | fake | openai | anthropic | openrouter | gemini
     entity_extraction_model: str = "gpt-5.4-nano"
+    # Default for the ``search.entity_retrieval`` org setting: query-time entity
+    # lookup + graph search. A tenant override wins; this is the fleet-wide
+    # fallback so an operator can disable entity/graph reads on a whole box
+    # (env / compose override) without editing every tenant's settings.
+    # Read-side only — the write path keeps building the entity graph.
+    entity_retrieval_enabled: bool = True
     use_llm_for_memory_creation: bool = True
     sentry_dsn: str = ""  # Set to enable Sentry error tracking
     redis_url: str = ""  # e.g. redis://localhost:6379/0. Empty = in-memory fallback.
--- a/core-api/src/core_api/mcp_server.py
+++ b/core-api/src/core_api/mcp_server.py
@@ -699,6 +699,7 @@ async def memclaw_recall(
             top_k=capped_top_k,
             recall_boost=config.recall_boost,
             graph_expand=config.graph_expand,
+            entity_retrieval=config.entity_retrieval,
             tenant_config=config,
             search_profile=agent_profile,
             readable_tenant_ids=_get_readable_tenants() or None,
--- a/core-api/src/core_api/pipeline/steps/search/classify_query.py
+++ b/core-api/src/core_api/pipeline/steps/search/classify_query.py
@@ -72,7 +72,22 @@ class ClassifyQuery:
         graph_max_hops: int = search_params["graph_max_hops"]
         top_k: int = search_params["top_k"]
 
-        tokens = extract_entity_tokens(query)
+        # ``search.entity_retrieval`` org setting (default True — absent key on
+        # any internal caller means "enabled", so behaviour is unchanged). When
+        # off, skip the entity block entirely: no entity FTS, no graph
+        # expansion, and no ENTITY_LOOKUP short-circuit, so the query falls
+        # through to the temporal / recent-context / keyword / semantic cascade
+        # below. ``ParallelEmbedAndEntityBoost`` reads the same flag and skips
+        # hop-boosting, making this the single switch for query-time entity and
+        # graph retrieval. Deliberately independent of ``graph_expand``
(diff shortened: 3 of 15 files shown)


ran-taig:
@claude

ran-taig:
@claude

ran-taig:
@claude

ran-taig:
@erni-a

erni-a (approved):

PR #679, opened 2026-07-31
fix(insights): declare numpy so discover keeps clustering

ran-taig:
## What broke

`insights_service._query_discover` imports numpy directly for its k-means step, but **numpy was never a declared core-api dependency** — it arrived transitively via `pgvector`. #605 bumped pgvector 0.4.2 → 0.5.0, which made numpy optional. numpy dropped out of `core-api/uv.lock`, then out of the built image, and discover silently started taking its `except ImportError` branch: flat pattern analysis instead of vector clustering.

This was found while analyzing insight quality on the eToro on-prem deployment. Every nightly insights run on all three tenants was degraded from **Jul 28** onward, and nothing surfaced it — the fallback logs a warning, but the run still reports success and writes a full set of findings.

## Why CI didn't catch it

Two independent blind spots, one guard added for each:

1. **The test skipped itself.** `TestNumpyKmeans::test_basic_clustering` opened with `try: import numpy / except ImportError: pytest.skip(...)`. When numpy vanished, the test skipped and CI stayed green. It now imports numpy unconditionally, plus a new `test_numpy_is_installed` that **fails** rather than skips.
2. **pytest structurally cannot catch this.** CI installs core-api *editable* (`uv pip install -e core-api/[dev]`), while the image installs from the *frozen lock export* — so a module present only transitively imports fine in the test job while being absent from the image. The `image-version` job already exists for exactly this class of bug (the version stamp "silently shipped for weeks" for the same reason), so it now also asserts `import numpy` inside the built image. No extra build cost — the image is already built in that job.

## Changes

| File | Change |
|---|---|
| `core-api/pyproject.toml` | Declare `numpy>=2.0,<3` — the code imports it directly, so it owns the declaration. Comment records the pgvector history. |
| `core-api/uv.lock` | Re-locked: `numpy 2.5.1` added (plus a stale project version-string catch-up, 2.24.1 → 2.26.0). |
| `tests/test_insights_service.py` | Fail-not-skip numpy guard; removed `try/except ImportError` from the clustering test. |
| `.github/workflows/ci.yml` | Assert directly-imported runtime deps are present in the built image. |

## Verification

- `uv export --frozen --no-dev` now emits `numpy==2.5.1` into the image requirements.
- Built the amd64 runtime image from this branch and ran inside it:
  - `import numpy` → `numpy 2.5.1`
  - the real `_numpy_kmeans` correctly separates two synthetic clusters (labels `0`/`1`, centroids `(2,2)`)
- `pytest tests/test_insights_service.py -k NumpyKmeans` passes; `uv lock --check` clean; `ruff` clean.

## Notes / follow-ups (not in this PR)

- **eToro prod is still degraded** — the box runs on-prem `v2.11.13`; restoring clustering there needs a release + cutover, tracked separately.
- **The fallback still reports success.** A degraded discover run is indistinguishable from a healthy one in the audit row. Making that visible (audit-row stat or alert) is a deliberate follow-up, intentionally kept out of this fix.
- Worth a look at whether other services rely on transitively-present modules they import directly; `core-storage-api` also lost numpy in the same bump, though nothing there imports it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -288,3 +288,18 @@ jobs:
             echo "::error::baked version '$BAKED' != pyproject '$EXPECTED' — the version stamp drifted from the source of truth."
             exit 1
           fi
+
+      # Same class of bug as the version stamp above, and equally invisible to
+      # pytest: CI installs core-api editable, so a module that is only present
+      # transitively still imports fine in the test job while being absent from
+      # the built image. numpy was exactly this — it arrived via pgvector <0.5.0,
+      # and when 0.5.0 made it optional the insights discover mode silently
+      # degraded to its non-clustered fallback in production for days (the
+      # fallback logs a warning but the run still reports success). Assert the
+      # modules the runtime code imports directly are actually in the image.
+      - name: Assert directly-imported runtime deps are present in the image
+        run: |
+          docker run --rm core-api-versiontest python -c "import numpy; print('numpy', numpy.__version__)" || {
+            echo "::error::numpy is missing from the core-api image — insights discover mode would silently fall back to non-clustered analysis. It must stay declared in core-api/pyproject.toml (see services/insights_service.py::_numpy_kmeans)."
+            exit 1
+          }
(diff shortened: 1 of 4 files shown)


ran-taig:
@claude

ran-taig:
@erni-a

erni-a (approved):

PR #669, opened 2026-07-29
feat(search): pluggable reranking component (MiniLM cross-encoder), ships dark

arkash20:
## What
Adds a **second-stage reranking component** — `common/ranking/` — mirroring the embedding component one-for-one (Protocol / registry / service / providers), selected by `RANK_PROVIDER` exactly like `EMBEDDING_PROVIDER`. A new `RerankResults` pipeline step sits at the **rerank seam** (between `ExecuteScoredSearch` and `PostFilterResults`) and reorders the candidate pool by the ranker's scores.

## Providers
| `RANK_PROVIDER` | backend | notes |
|---|---|---|
| `noop` (default) | identity | true no-op on first-stage order → **ships dark** |
| `local` | in-process `sentence-transformers` CrossEncoder | **MiniLM-L6 (22M)** |
| `fake` | deterministic overlap | tests |
| `remote` | *(reserved)* | future TEI/Cohere-shaped GPU sidecar via `RANK_BASE_URL` |

## Why MiniLM-L6
Grounded in the A50 quality/latency spike (`benchmark/a50_rerank_latency/RESULTS.md`):
- **Quality:** MiniLM-L6 (22M) reranked **≈ bge-reranker-base (278M)** on our data — within 1–4pp hit@k, even beating bge @1.
- **Latency:** MiniLM at small pools ≈ **50–150ms CPU**; bge in-process is **1–5s** (not viable) → the big model belongs behind a future `remote` GPU provider.

## Ships dark
`RANK_PROVIDER=noop` is the default and a **true identity** on first-stage order, so behaviour is unchanged until a deployment/tenant opts in (same guarantee as A49/`score_formula`). Every failure path — misconfig, timeout past `RANK_TIMEOUT_SECONDS`, provider down, length-contract violation — **degrades to first-stage order**. Recall never fails because rerank did. `sentence-transformers` stays lazy-imported (already an optional dep of the local embedder) → **no new hard dependency**.

## Note on `noop`
Unlike the design sketch, `noop` returns position-descending scores (a true identity) rather than `similarity`: first-stage rows are ordered by the composite `score`, not `similarity`, so returning similarity would silently reorder the pool when boosts diverge — not ship-dark.

## Tests
- `tests/test_ranking_component.py` — protocol conformance, noop true-identity, fake overlap, service degrade (empty / misconfig / length-contract violation).
- `tests/pipeline/test_rerank_results.py` — step reorders by ranker score, noop keeps first-stage order, skips on entity-lookup / empty pool.

[...]

PR #680, opened 2026-08-01
feat(search): remote rerank provider (sidecar) + opt-in local build

arkash20:
Completes the reranking component (#669) so it fits **both** deployment shapes behind the same `RANK_PROVIDER` flag — the right tool per environment.

## `remote` provider (staging/prod) — recommended
`common/ranking/providers/remote.py`: HTTP `/rerank` to a **separate sidecar** (self-hosted TEI reranker / GPU bge, or a Cohere-style endpoint). Keeps **torch out of core-api**, mirroring how you already serve the TEI *embedder*.
- TEI-native contract `{query, texts} → [{index, score}]`; also parses the Cohere `{results:[{index, relevance_score}]}` wrapper. Scores re-projected to **input order** (caller sorts).
- Registry LRU-caches the httpx pool per `(base_url, api_key, model)` with background `aclose()` on eviction (mirrors the OpenAI embedding cache). `remote` with no base URL → `ValueError` → service **degrades to first-stage order**.

## `local` build opt-in (local dev)
- `core-api/Dockerfile` gains `WITH_RERANK_LOCAL` (default `0`) that installs `torch + sentence-transformers` **outside the frozen lock** — the **default image stays lean and prod-unaffected**.
- `docker-compose.rerank-local.yml` overlay builds core-api with it + flips `RANK_ENABLED`/`RANK_PROVIDER=local`. The reranker analogue of `--profile embed-local`, needed because **the TEI sidecar has no arm64 image** — so in-process is the only reranker path on Apple Silicon.
  ```
  docker compose -f docker-compose.yml -f docker-compose.rerank-local.yml up --build
  ```

## Net
| Environment | `RANK_PROVIDER` | Model runs |
|---|---|---|
| Local dev (arm64 Mac) | `local` (via overlay) | in-process in core-api |
| Staging / prod | `remote` | `/rerank` sidecar, core-api lean |

Same pipeline step, same `RANK_ENABLED` kill-switch — just the right backend per env. All still **dark by default**.

## Tests
`tests/test_ranking_remote.py` (httpx `MockTransport`): request shape, TEI + Cohere response parsing, input-order re-projection, empty short-circuit, HTTP-error degrade, registry no-URL guard. **Additionally wet-tested against a real stdlib `/rerank` server** (real httpx round-trip → correct reorder). `.env.example` documents both providers + the overlay.

## Follow-ups (not in this PR)
- Enterprise: stand up a TEI **rerank sidecar** on staging (mirror `staging-memclaw-tei`), set `RANK_BASE_URL`+`RANK_ENABLED`+`RANK_PROVIDER=remote` in core-api Helm — then the **real-TEI wet test** (the one thing MockTransport/stdlib can't cover) runs against live infra.
- A/B `remote`/`local` vs `noop` on the benchmark.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -122,14 +122,27 @@ OPENAI_API_KEY=
 # (zero cost — no candidates built, no model call). Set false to disable
 # reranking in an incident without touching provider config.
 # RANK_ENABLED=false
-# Which ranker runs when enabled. Options: noop | local | fake
-#   local = in-process sentence-transformers CrossEncoder (MiniLM). NOTE:
-#   requires torch + sentence-transformers in the core-api image (~+1-2GB)
-#   and loads the model in-process — the first few searches after start
-#   fall back to first-stage order while the model warms.
+# Which ranker runs when enabled. Options: noop | local | remote | fake
+#   local  = in-process sentence-transformers CrossEncoder (MiniLM). Needs torch
+#            + sentence-transformers in the core-api image (~+2GB) — NOT in the
+#            default image; enable via the WITH_RERANK_LOCAL build-arg. For LOCAL
+#            dev just use the overlay (torch has no arm64 TEI image, so in-process
+#            is the only path on Apple Silicon):
+#              docker compose -f docker-compose.yml \
+#                -f docker-compose.rerank-local.yml up --build
+#   remote = HTTP /rerank sidecar (self-hosted TEI reranker, typically GPU bge, or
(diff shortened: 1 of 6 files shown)


arkash20:
@Eldad-Caura please approve — CI green. Adds the `remote` rerank provider (sidecar, recommended for staging/prod) + opt-in `local` build for Mac dev. Same flags, dark by default. Verified against MockTransport + a real stdlib /rerank server; the real-TEI wet test lands at sidecar-deploy time.

Eldad-Caura (approved):

PR #681, opened 2026-08-02
feat(insights): filter input noise — opaque payloads, title-dedup, discover window

ran-taig:
## Problem

Follow-up to the eToro insights quality analysis (and #679). The insights engine's input is the raw memory firehose, and on a busy fleet that means:

- **~1/3 of all insight citations point at repetition noise** (measured over 4.6K citations from every post-Jun-16 insight on the eToro deployment).
- One family alone — **content-free reasoning artifacts** (`[{"type":"thinking","thinking":"","thinkingSignature":"…"}]`, an empty thinking block with an encrypted blob) — is 4.3% of 30-day episodes but drew **23.9% of citations**: nothing to analyze, but near-identical embeddings form the tightest clusters discover can find. The engine has been correctly reporting this about its own input nightly for weeks.
- The discover sample ("newest 200") **spans under a day** on a busy tenant (~700 distinct titles/day) — nightly runs cluster yesterday's activity and re-discover the same shapes.

Notably, **excluding episodes by type would be wrong**: measured across 141K 30-day episodes, 63% are substantive work, and the noise is defined by *repetition and opacity*, not by memory type.

## Changes (input-side only; nothing in storage is deleted or modified)

| Change | Where | Modes |
|---|---|---|
| Opaque-payload guard: structural content-prefix predicate (no title regexes) | `_insights_scope_filters` | all 6 — an unreadable row can't evidence anything |
| Title-dedup: one exemplar per exact title (newest), annotated `dup_count`/`first_seen`, rendered by the prompt formatter as `[repeats: Nx, first seen: date]` | `_insights_dedup_stmt` + formatter | patterns, stale, failures, discover — **deliberately not** contradictions/divergence, where same-titled disagreement is the evidence |
| Discover window: sample spread (`ORDER BY random()` — genuinely run-varying; a deterministic hash order would be a fixed permutation per corpus state) over trailing  `INSIGHTS_DISCOVER_WINDOW_DAYS = 30` instead of newest-first; patterns/failures/stale get their own scan bounds (`INSIGHTS_PATTERNS_WINDOW_DAYS = 30`, `INSIGHTS_FAILURES_WINDOW_DAYS = 90`, `INSIGHTS_STALE_WINDOW_DAYS = 90` — stale's deliberately wider than its 30-day age threshold, so it reports the 30-90d "recently became stale" band instead of perpetually re-reporting the ancient tail) — the dedup subquery has no inner LIMIT, so unbounded patterns would scan the tenant's whole active history for 50 deduped rows (prod: 21K rows/167ms windowed vs 44K/243ms and growing) | discover only | window start computed on the core-api clock (same pattern as the stale thresholds) |

Correctness details: NULL/empty titles fall back to row-id as dedup key (no collapse); `dup_count` is computed over the *filtered* set so it means "N matching rows in this window"; each mode keeps its own result ordering; `window_start` omitted → the ORIGINAL pre-dedup query shape (identical results and cost — no dedup, no annotations, no full-corpus window scan), so an older core-api works against a newer storage and vice versa; failures sorts with `dup_count` as secondary key so frequent patterns aren't cut by LIMIT below one-off rows.

## Validation

- **Wet tests** (6 new in `test_ph5b_insights_storage.py`, against real Postgres): opaque exclusion across all reads incl. contradictions; dedup exemplar choice + annotations; NULL/empty-title survival; same-title supersede pair still visible to contradictions; window bounding + naive-datetime 422 guard (created_at is timestamptz) + legacy fallback; mode ordering preserved under dedup. 30/30 pass.
- **Formatter unit tests** (3 new): annotation rendering, singleton/legacy rows unchanged, cluster representatives annotated. 38/38 pass in the two core-api insights files.
- **Against production data (read-only)**: predicate catches 97.6% of the opaque family; the new discover shape spans **31 distinct days instead of 1**, residual noise titles drop to ~10% of the sample; ~130ms on the busiest tenant. `ruff` clean.

## Explicitly out of scope

- Upstream write suppression (why agents persist empty encrypted thinking blocks at all) — separate track.
- The pending-status leak, novelty guard, and fallback observability from the same analysis — separate PRs.
- eToro deploy: the box needs a release cutover (which would also pick up #679's numpy fix).

### DIFF
--- a/core-api/src/core_api/clients/storage_client.py
+++ b/core-api/src/core_api/clients/storage_client.py
@@ -1849,8 +1849,18 @@ class CoreStorageClient:
         )
 
     async def insights_query_failures(
-        self, *, tenant_id: str, fleet_id: str | None, agent_id: str, scope: str, max_memories: int
+        self,
+        *,
+        tenant_id: str,
+        fleet_id: str | None,
+        agent_id: str,
+        scope: str,
+        max_memories: int,
+        window_start: datetime | None = None,
     ) -> list[dict]:
+        """``window_start`` bounds the title-dedup scan (see
+        ``INSIGHTS_FAILURES_WINDOW_DAYS``); ``None`` keeps the legacy
+        unbounded read."""
         return await self._post(  # type: ignore[return-value]
             "/insights/failures",
             {
@@ -1859,6 +1869,7 @@ class CoreStorageClient:
                 "agent_id": agent_id,
                 "scope": scope,
                 "max_memories": max_memories,
+                "window_start": window_start.isoformat() if window_start else None,
             },
             read=True,
         )
@@ -1873,7 +1884,12 @@ class CoreStorageClient:
         thirty_days_ago: datetime,
         fourteen_days_ago: datetime,
         max_memories: int,
+        window_start: datetime | None = None,
     ) -> list[dict]:
+        """``window_start`` bounds the title-dedup scan (see
+        ``INSIGHTS_STALE_WINDOW_DAYS`` — wider than the 30-day age threshold
+        rows must exceed to qualify); ``None`` keeps the legacy unbounded
+        read."""
         return await self._post(  # type: ignore[return-value]
             "/insights/stale",
             {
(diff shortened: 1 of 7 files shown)


ran-taig:
@claude

ran-taig:
@claude

ran-taig:
@claude

ran-taig:
@claude

ran-taig:
@claude

ran-taig:
@claude

ran-taig:
@claude

ran-taig:
@erni-a

erni-a (approved):

PR #682, opened 2026-08-02
feat(mcp): migrate core-api to MCP Python SDK v2, serving both protocol eras

Eldad-Caura:
> ### Ready for review — dual-stack, not single-stack
>
> This PR previously carried a "do not merge" banner on the grounds that it speaks only `2026-07-28` and would drop the `initialize` handshake for older clients. **That premise was wrong, and it's now measured rather than assumed.** Details below.

Adopts `mcp>=2,<3` for core-api. The server keeps answering `2025-11-25` clients while gaining the `2026-07-28` surface.

## The blocking premise, retested

`mcp` 2.0 routes **by era** inside `StreamableHTTPSessionManager`, keyed on the `MCP-Protocol-Version` header: a value in the handshake set — or *no header at all*, which is exactly what an old client's opening `initialize` looks like — reaches the legacy transport; anything else reaches the modern single-exchange one. Both eras share one handler registry, so the same 12 tools answer on both. There is no flag to set, and none to unset.

Measured against this branch:

| request | result |
|---|---|
| `initialize` @ `2025-11-25` | **200**, echoes `protocolVersion: 2025-11-25` (not upgraded to latest) |
| `tools/list` (legacy dialect) | **200**, full 12-tool catalogue |
| `tools/call` (legacy dialect) | **200**, real dispatch |
| `tools/call` @ `2026-07-28` | **200** |

So there is no compatibility break to build around. What remained was making sure it *stays* that way.

## Keeping it true

Nothing else in the suite speaks the legacy dialect, so a later change pinning a single era would break every pre-2026 client **silently**. `tests/test_mcp_dual_stack.py` is that guard.

Two things make it worth trusting rather than just green:

- **A negative control.** The same `initialize` declaring `2026-07-28` must be *refused*. Without it the test would still pass if both eras collapsed onto one transport — proving nothing about routing.
- **The guard was falsified before being trusted.** Emptying `HANDSHAKE_PROTOCOL_VERSIONS` makes the legacy handshake fail with `-32602`, confirming the test actually goes red when era-routing breaks.

The test builds its own app rather than reusing the mounted one, for reasons both hit the hard way and written into the module docstring: `StreamableHTTPSessionManager.run()` may be called once per instance and `mcp_server` builds one app at import, so sharing it means whichever of this file and `test_mcp_server_discover.py` ran second dies on the second `run()`; and since `streamable_http_app()` reassigns the manager on the wrapped lowlevel server, the original is snapshotted and restored or this file breaks the other test by collection order. The lifespan is entered in the test body because a fixture yielding across the anyio task group exits the cancel scope from a different task than entered it.

## What the SDK gave us for free

`server/discover`, `resultType`, `ttlMs`/`cacheScope`, `-32020` header validation, `subscriptions/listen` and MRTR all ship in `mcp` 2.0.0 — three planned action items collapse into a dependency bump plus config.

The production delta is **~20 lines**, because core-api imports four symbols from `mcp` and uses the decorator API, which v2 preserves:

- `mcp.server.fastmcp.FastMCP` → `mcp.server.MCPServer`. That module is **deleted** in v2 with no shim — a genuine hard break, just a narrow one.
- `call_tool` takes the new `context` parameter and forwards it; return type is now `CallToolResult | InputRequiredResult`. Telemetry unchanged.
- `stateless_http` / `json_response` / `streamable_http_path` / `transport_security` **moved** from the constructor to `streamable_http_app()` — not removed.
- `cache_hints` added for SEP-2549. `ttl_ms` 5 min (ratified 2026-08-03); `scope` stays `"private"`.

Dependency: `mcp>=1.28.1,<2` → `mcp>=2,<3`, lock regenerated with `--upgrade-package mcp` only. **Note v2 pulls in `httpx2`/`httpcore2` alongside the existing `httpx` and drops `httpx-sse`** — worth a look if image size or HTTP-client consistency matters.

## Advertise only what we serve

`MCPServer.__init__` wires prompt/resource handlers unconditionally and `get_capabilities()` derives advertisement from registered methods — so `server/discover` was claiming prompts AND resources (including `resources.subscribe`) for surfaces that only ever return empty lists. `2026-07-28` makes that consequential: discover is a MUST and is how clients decide what to call. Dropping those handlers makes it honest; an unadvertised method then answers `-32601`.

## A pre-existing test defect, surfaced rather than caused

Both `tools/list` baseline guards (`test_mcp_token_budget.py`, `test_tool_descriptions_regression.py`) serialized **without** `by_alias=True` — asserting on Python field names instead of the wire, so they'd fail on a harmless rename while letting a genuine wire change through. Both now use `by_alias=True`. The fixture is regenerated for the one real wire delta, `meta` → `_meta`; descriptions and schemas are byte-identical, so the prompt-token budget is unchanged.

`mcp_types` v2 exposes snake_case attributes; construction by camelCase alias still works and `model_dump(by_alias=True)` still emits camelCase, so **production code and the wire format are unaffected** — the rest is assertion churn.

## Deliberately untouched

The `"fastmcp"` entry in `common/structlog_config.py`'s reroute list. It's a no-op now, but `common/` is vendored into the enterprise repo under the identical-policy re-vendor bot, so editing it there widens this change's blast radius for no functional gain — the sibling `"mcp"` entry already routes the SDK's loggers, confirmed by JSON-formatted `mcp.server.*` lines in the test run.

## Not covered here

- MRTR / `InputRequiredResult` is unused (`server/discover` itself is covered through `MCPAuthMiddleware` by `test_mcp_server_discover.py`).
- Nothing in `memclawd`.

## Verification

- Full suite **4072 passed**, 3 skipped, 2 xfailed, 3 xpassed, **0 failed**.
- `ruff check` and `ruff format --check` clean on every changed file. Note `origin/main` itself reports pre-existing `ruff check` errors under ruff 0.16.0, unrelated to this branch.
- Dual-stack test verified passing in **both collection orders** against `test_mcp_server_discover.py`.

### DIFF
--- a/core-api/pyproject.toml
+++ b/core-api/pyproject.toml
@@ -32,7 +32,7 @@ dependencies = [
     "openai",
     "python-dateutil>=2.8,<3",
     "httpx>=0.27,<1",
-    "mcp>=1.28.1,<2",
+    "mcp>=2.1.1,<3",
     "argon2-cffi>=23.1",
     "cryptography>=42.0",
     "python-jose[cryptography]>=3.3",
--- a/core-api/src/core_api/app.py
+++ b/core-api/src/core_api/app.py
@@ -246,7 +246,7 @@ async def lifespan(app):
     # imported AFTER that call (slowapi / mcp_server below, uvicorn by the
     # server) — so the import-time pass no-ops for them (it logs a "rerouting
     # was a no-op" warning) and their records never reach the JSON/GCP handler.
-    # Most consequentially, FastMCP's "Error executing tool ..." tool-error
+    # Most consequentially, the MCP SDK's "Error executing tool ..." tool-error
     # lines were invisible in prod logs. The re-route is idempotent, so this
     # post-import re-run from the ASGI lifespan startup safely routes them.
     reroute_third_party_loggers()
@@ -1067,7 +1067,7 @@ if _os.getenv("TESTING") == "1":
 
     app.include_router(testing_router, prefix="/api/v1")
 
-# Mount at /mcp; FastMCP's internal Route("/") handles the canonical /mcp/.
+# Mount at /mcp; the SDK app's internal Route("/") handles the canonical /mcp/.
 # Bare /mcp (no trailing slash) doesn't match Mount's regex, so the parent
 # router would issue a 307 — streaming MCP clients (e.g. Anthropic's
 # remote-MCP integration) hang on the initialize handshake when a redirect
--- a/core-api/src/core_api/mcp_server.py
+++ b/core-api/src/core_api/mcp_server.py
@@ -19,7 +19,7 @@ from uuid import UUID, uuid4
 
 import httpx
 from fastapi import HTTPException
-from mcp.server.fastmcp import FastMCP
+from mcp.server import CacheHint, MCPServer
 from mcp.server.transport_security import TransportSecuritySettings
 from mcp.types import CallToolResult, TextContent
 from pydantic import Field, ValidationError
@@ -220,7 +220,7 @@ def _as_error_result(envelope: str) -> CallToolResult:
     """
     return CallToolResult(
         content=[TextContent(type="text", text=envelope)],
-        isError=True,
+        is_error=True,
     )
 
 
@@ -780,14 +780,14 @@ async def _no_db():
     yield None
 
 
-# ── FastMCP instance ──
+# ── MCPServer instance ──
 
 
-class _InstrumentedFastMCP(FastMCP):
-    """FastMCP that records one capability-usage sample per tool call.
+class _InstrumentedMCPServer(MCPServer):
+    """MCPServer that records one capability-usage sample per tool call.
 
     Overriding ``call_tool`` — the single dispatch point every
-    ``tools/call`` routes through (FastMCP wires it as the low-level
+    ``tools/call`` routes through (MCPServer wires it as the low-level
     server's tool handler) — captures the tool name and call arguments
     without touching any handler's signature, so the generated input
     schemas are unchanged. This is the MCP-side adoption emitter; its REST
@@ -800,7 +800,7 @@ class _InstrumentedFastMCP(FastMCP):
     right semantics for adoption (the capability was invoked).
     """
 
-    async def call_tool(self, name, arguments):  # type: ignore[override]
+    async def call_tool(self, name, arguments, context=None):  # type: ignore[override]
         # PERMANENT rename alias (2026-08-14): tools are listed as caura_*,
         # but memclaw_* calls are accepted forever — saved prompts, keystone
         # rules, and published tutorials quote the old names, and breaking
@@ -812,7 +812,7 @@ class _InstrumentedFastMCP(FastMCP):
         t0 = time.perf_counter()
         status = "ok"
         try:
-            return await super().call_tool(name, arguments)
+            return await super().call_tool(name, arguments, context)
         except Exception:
             status = "error"
             raise
@@ -829,8 +829,15 @@ class _InstrumentedFastMCP(FastMCP):
             )
 
 
-mcp = _InstrumentedFastMCP(
+mcp = _InstrumentedMCPServer(
     name=f"Caura v{VERSION}",
+    # v2 added a dedicated ``version``, surfaced to clients as
+    # ``_meta["io.modelcontextprotocol/serverInfo"].version`` on every result.
+    # Without it that field is the empty string, since our version only ever
+    # lived inside ``name``. Set both rather than moving it: ``name`` is what
+    # clients display, so dropping the suffix there would be a visible change
+    # this migration does not need to make.
+    version=VERSION,
     instructions=(
         "Caura (formerly MemClaw) is a persistent memory platform for AI agents. "  # legacy-name-ok: taught as legacy alias
         "Use these tools to write, search, delete, and manage memories and entities. "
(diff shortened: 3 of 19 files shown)


Eldad-Caura:
@claude Review requested. This is intentionally a **draft** — the branch is single-stack `2026-07-28` and must not merge before the dual-stack lands, so `pre-checks` skipped the on-open review. Retriggering here since the retrigger path has no draft gate.

Most of the diff is test churn from `mcp_types` v2 switching to snake_case attributes, and the review prompt already excludes `tests/`. The production surface worth your attention is small:

- `core-api/src/core_api/mcp_server.py` — the `FastMCP` → `MCPServer` swap, the `call_tool` override now taking and forwarding `context`, and the four transport kwargs moving to `streamable_http_app()`.
- The `cache_hints` block, where `scope="private"` is deliberate but the reasoning is the opposite of what you might expect — the tool surface is caller-invariant, so `"public"` is defensible too. Worth checking that comment for accuracy; an earlier version of it asserted the surface was trust-gated, which is false.
- `core-api/uv.lock` — v2 pulls in `httpx2`/`httpcore2` alongside the existing `httpx` and drops `httpx-sse`.

Eldad-Caura:
Verified the review finding on the new transitive deps. **No code change needed** — but two of its supporting facts are wrong, and the one that is right is right for a different reason. Recording it here so nobody re-litigates.

**"Three brand-new PyPI packages all published 2026-07-24" — incorrect.**

| package | first ever upload | releases |
|---|---|---|
| `httpx2` | 2026-05-11 | 13 |
| `httpcore2` | 2026-05-11 | 13 |
| `truststore` | **2022-03-10** | 16 |

`truststore` is a four-year-old package by Seth Larson (PSF security developer-in-residence). The 0.10.4 wheel pinned in our own lock is dated `2025-08-12` — a year before the claimed date.

**"Confirm httpx2/httpcore2 are the official encode/httpx v2, not third-party forks" — neither.** They are authored by **Tom Christie**, httpx's creator, but published under the **pydantic** org (`github.com/pydantic/httpx2`), not `encode`. Legitimate provenance, unusual home. Worth knowing; not a blocker.

**"truststore silently changes TLS behaviour" — correct, via a different mechanism.** Not `inject_into_ssl()`, which truststore's own docs explicitly forbid libraries from calling. `httpx2/_config.py:39-40` uses `truststore.SSLContext(ssl.PROTOCOL_TLS_CLIENT)` as its **default** SSL context, so httpx2 verifies against the OS trust store rather than a bundled CA set.

**The actionable risk does not materialise, for two independent reasons:**

1. `python:3.12-slim` — our base image at every stage — ships `ca-certificates 20250419` with a populated `/etc/ssl/certs/ca-certificates.crt` (224 KB, verified by running the image). The OS trust store is present, so no `ca-certificates` line or `TRUSTSTORE_DISABLE=1` is needed.
2. core-api does not use the MCP SDK's **client** at all — no `mcp.client` or `ClientSession` imports. `httpx2` is installed as a transitive dep but sits on no outbound path here. core-api's own HTTP still goes through `httpx` 0.28.1, untouched by this PR.

**Genuine residue worth carrying forward:** this becomes live the moment anything of ours uses the SDK's client side — the broker's upstream path is the obvious candidate. At that point the OS-trust-store default is a real deployment property to check per image, not just here. Noting it against that work rather than this PR.

Eldad-Caura:
@claude

Eldad-Caura:
Applied — the finding was real, and it was the sharpest kind: the function's own comment promised "warn-and-continue rather than assert: ... a rename on a patch bump should not take the service down," but only the *method key* was guarded. `mcp._lowlevel_server._request_handlers` was traversed unguarded at import, so a rename of either private attribute raises `AttributeError` during module import and kills the service at startup — the exact outcome the per-method warning exists to prevent. `mcp>=2,<3` accepts any 2.x, including one that renames these.

Now both failure modes warn and continue, with the attribute path getting its own guard and an early return.

Verified rather than assumed: deleting `_lowlevel_server` and calling `_drop_unserved_handlers()` directly logs

```
mcp capability tightening: could not reach the low-level request handlers
(SDK renamed _lowlevel_server/_request_handlers) — skipping; server/discover
may now over-advertise prompts and resources we do not serve
```

and returns without raising. Before the fix that path raised.

Full suite still 4072 passed / 0 failed; `ruff check` and `ruff format --check` clean.

Note the same private path is used in `tests/test_mcp_dual_stack.py` to snapshot and restore the session manager — deliberately left unguarded there, since a test *should* fail loudly if the SDK moves an internal it depends on.

Eldad-Caura:

All checks green, and the one Claude review finding (unguarded private-attribute traversal at import) is applied and verified.

Worth knowing before you look: **this PR's "do not merge — single-stack" banner was wrong**, and that is now measured rather than argued. `mcp` 2.0 routes by protocol era inside the session manager, so one endpoint serves both `2025-11-25` (handshake) and `2026-07-28` (enveloped) with no flag involved. On this branch, `initialize` at `2025-11-25` returns 200 echoing that version, and `tools/list` / `tools/call` work on the legacy dialect alongside the modern one. The compatibility work the PR was blocked on does not need building.

What that leaves is keeping it true, since nothing else in the suite speaks the legacy dialect — hence `tests/test_mcp_dual_stack.py`, which carries a negative control and was itself falsified (breaking era-routing makes it go red) before being trusted.

erni-a (dismissed):

Eldad-Caura:
@claude

Eldad-Caura:
Verified against the locked MCP 2.2.0 low-level handler registry: it registers `subscriptions/listen`, but it does **not** register `resources/subscribe` or `resources/unsubscribe`. Adding those names to `_UNSERVED_METHODS` would therefore produce missing-handler warnings on every startup and would not tighten the advertised resource capability.

The modern `subscribe` capability is derived from `subscriptions/listen`; the legacy resource subscription handlers are absent in this SDK path. I left the tuple unchanged and corrected the misleading `resources.subscribe` wording in the production comment, test comment, and commit message in 36102959.

So the review found a real documentation ambiguity, but not a capability-removal bug.

Eldad-Caura:
@erni-a pleasr approve

erni-a (approved):

PR #685, opened 2026-08-02
fix(ci): pin common/'s ruff contract and lint-gate it

### LINKED ISSUE #684: common/ has no pinned ruff config, so its effective lint rules drift with the ruff version
**Corrected after further investigation.** The original title claimed "21 ruff check errors on main." That figure is real but config-dependent and I stated it misleadingly — under the rule set this repo actually configures, `common/` is **clean**. The genuine defect is narrower and more durable: `common/` has no pinned ruff config at all, so whatever rules apply to it are inherited from ruff's built-in defaults and change when ruff is upgraded.

## Measured

`common/`, 88 files, ruff 0.16.0:

| invocation | `ruff check` | `ruff format --check` |
|---|---|---|
| bare (no `--config`) | **21 errors** | 7 files would reformat |
| `--isolated` (ruff defaults) | **21 errors** — identical codes | 7 files would reformat |
| `--config core-api/pyproject.toml` | **0 errors** | 49 files would reformat |

Bare invocation is byte-identical to `--isolated`. There is no `ruff.toml`, no repo-root `pyproject.toml`, and no `common/pyproject.toml`, so ruff finds no config by walking up from these files and falls back to **built-in defaults** — confirmed by `--show-settings`, which reports `linter.line_length = 88` for `common/structlog_config.py` rather than the 110 configured in `core-api/pyproject.toml`.

## Why this is the actual bug

Ruff 0.16.0's default enabled rule set is very large (YTT, ASYNC, BLE, TRY, S, DTZ, G, …), a substantial expansion over the historical `E4`/`E7`/`E9`/`F` default. So the rules governing `common/` today are simply *whatever ruff ships as default in the installed version* — a set nobody here chose, which grows on every upgrade.

The two `RUF100` "unused `noqa` (non-enabled: `E402`)" hits are the tell. That is not code rot; it is a `noqa` that was meaningful under one effective rule set and became dead under another, with no config change in between. Any ruff bump can add or retire rules across this directory silently.

Note the direction of the numbers: `common/` is de facto an **88-column tree** — only 7 of 88 files drift from ruff defaults, whereas 49 would be reformatted under core-api's 110. It has been maintained to defaults, just never told to stay there.

## The cross-repo constraint is narrower than I first thought

From `caura-memclaw-enterprise/scripts/vendored_files_manifest.json`:

- `policy=identical` — **exactly 3 files**: `common/__init__.py`, `common/serve.py`, `common/structlog_config.py`. These must stay byte-identical across repos, and `ci.yml:141-150` already format-checks precisely these at `--isolated`. That gate is correctly scoped.
- `policy=manual` — `common/events/*` (6 files). Reformatting creates reconciliation work, not a CI break.
- The remaining ~79 files are not vendored and carry no cross-repo formatting constraint.

## Options

**1. Pin the contract (recommended).** Add a `common/ruff.toml` with an explicit `select`/`ignore` and `line-length = 88`, matching how the tree is already maintained and the `--isolated` vendoring gate. Then add `ruff check common/` to CI at that config and fix whatever the *chosen* rule set reports. Fixes the root cause — rule drift — without touching formatting beyond the existing 3-file gate. Cost: someone has to choose the rule set.

**2. Align `common/` with the rest of the repo.** Apply core-api's config. Lint is already clean under it (0 errors), but it reformats **49 of 88 files** and breaks byte-parity on the 3 `identical` files, requiring a coordinated enterprise change. Fights the vendoring contract for little gain.

**3. Document and close.** State in `ci.yml`, beside the vendored-format step, that `common/` is deliberately defaults-formatted and not lint-gated. Cheapest, leaves the drift in place.

Recommendation is **1**. The absence of a pinned contract is the defect; the 21 errors are just its current symptom, and they will be a different 21 after the next ruff bump.

## How this surfaced

Running ruff at a wider scope than CI uses while verifying PR #682, then having to determine whether the failures belonged to that branch. They did not — but the reason took three rounds of triangulation to establish, which is itself evidence of the footgun.

Eldad-Caura:
Fixes #684.

## The problem

`common/` had **no ruff config above it** — no `ruff.toml`, no root `pyproject.toml`, nothing in `common/`. So ruff walked up from each file, found nothing, and fell back to its **built-in defaults**. The rule set governing shared code that every service imports was therefore "whatever the installed ruff version ships as default," and it changes on upgrade.

Confirmed with `--show-settings`: it reported `linter.line_length = 88` (ruff's default) for `common/structlog_config.py` rather than the 110 configured for core-api, and a bare `ruff check common/` was byte-identical to `--isolated`.

Ruff 0.16.0's default set (YTT, ASYNC, BLE, TRY, S, DTZ, G, …) reported 21 findings here that nobody had enabled — including two dead `noqa: E402` directives, meaningful under an older default set and silently dead under the new one. **That drift is the defect.** The specific findings were only its current symptom, and would have been a different set after the next bump.

## The fix

`common/ruff.toml` mirrors core-api'"'"'s lint rules, so shared code is held to the same standard as its consumers — but sets `line-length = 88` rather than 110.

That difference is deliberate and load-bearing. Three files here are vendored into `caura-memclaw-enterprise` under `policy=identical` and are format-checked at ruff defaults via `--isolated` for byte-parity. This tree has been maintained at 88 throughout: **at 88 only 7 of 88 files would reflow; at 110 it would be 49.** Raising it would churn most of the directory and break that parity for no benefit — `E501` is ignored in both configs, so `line-length` affects only `ruff format` and the 88/110 split costs nothing on the lint side.

CI gains one step, `ruff check common/ --config common/ruff.toml`. The config is named explicitly rather than relying on discovery, for the same reason the neighbouring vendored step uses `--isolated`: the gate should not change meaning if config resolution ever does.

## Clearing the 16 findings the pinned rules report

- **6 auto-fixed** — 4 `I001` import order, 2 `RUF100` dead `noqa`.
- **8 `RUF002`/`RUF003` ambiguous unicode** in comments and docstrings. I replaced **only** U+00D7 (`×`→`x`) and U+2013 (en dash→hyphen). **U+2014 (em dash) is not flagged by ruff and is this codebase'"'"'s prose style, so it is untouched** — a blanket dash replacement would have mangled it. Worth noting `core-api/src` has zero `RUF002`/`RUF003` hits, so `common/` was the outlier here, not the rule being unreasonable.
- **2 `B027`** on `EventBus.start`/`stop` — marked `noqa` rather than decorated. Their docstrings define them as optional hooks ("In-process buses treat this as a no-op"), so adding `@abstractmethod` would force every subclass to implement them. That is a contract change, not a lint fix.

## Reviewer notes

- `common/events/{base,pubsub}.py` are `policy=manual` in the enterprise vendoring manifest, so those two edits create reconciliation work there — not a CI break. **None of the three `policy=identical` files are touched.**
- I also updated the neighbouring vendored-format step'"'"'s comment, which said `--isolated` pins defaults "even if a root ruff config is ever added." One now exists nearby, so the comment would have read as stale.
- The import reordering touches four `common/models` files, which is why the full suite matters here rather than just the lint gate.

## Verification

- `ruff check common/ --config common/ruff.toml` — clean.
- All pre-existing ruff scopes still clean; the vendored `--isolated` gate still passes (3 files already formatted).
- **4051 passed, 3 skipped, 1 xfailed, 3 xpassed, 0 failed.**
- `actionlint` reports the same single pre-existing finding as `origin/main`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -129,6 +129,15 @@ jobs:
         # of that blind spot close together or it just drifts again.
         run: uv run ruff check core-storage-api/tests/
 
+      - name: Run Ruff linter (common)
+        # ``--config`` is explicit on purpose. common/ had NO ruff config above it
+        # (no ruff.toml, no root pyproject.toml), so ruff fell back to built-in
+        # DEFAULTS here — meaning the rule set was "whatever the installed ruff
+        # version ships," and it changed under us on the 0.16.0 bump. common/ruff.toml
+        # pins it; naming the config keeps this step immune to future config
+        # discovery changes, the same reasoning as ``--isolated`` below. See #684.
+        run: uv run ruff check common/ --config common/ruff.toml
+
       - name: Run Ruff formatter check (core-api)
         run: uv run ruff format --check core-api/src/
 
@@ -144,9 +153,12 @@ jobs:
         # Enterprise CI format-checks the vendored copies at ruff DEFAULTS (88 cols),
         # so gate them here too or a merge turns enterprise CI red until a reflow
         # lands (full story: the vendoring note in common/structlog_config.py's
-        # docstring). ``--isolated`` pins default behavior even if a root ruff
-        # config is ever added; deliberately NOT all of common/ — non-vendored
-        # files there are not held to the 88-col contract.
+        # docstring). ``--isolated`` pins default behavior regardless of any ruff
+        # config in scope — which now includes common/ruff.toml (added in #684),
+        # so this step is deliberately unaffected by it. Still NOT all of common/:
+        # only the policy=identical files owe byte-parity to enterprise. Note
+        # common/ruff.toml also sets line-length = 88, so the two agree today;
+        # ``--isolated`` is what guarantees they cannot silently diverge.
         run: uv run ruff format --check --isolated common/__init__.py common/structlog_config.py common/serve.py
 
       - name: Guard against raw-string MCP error returns (audit B4 / CAURA-000 #147)
--- a/common/embedding/constants.py
+++ b/common/embedding/constants.py
@@ -53,10 +53,10 @@ EMBEDDING_RETRY_DELAY_S: float = float(os.environ.get("EMBEDDING_RETRY_DELAY_S",
 # tunable controls both the LLM and the embedding pools. Without an
 # explicit limits arg the SDK rides httpx's default (100 max / 20
 # keepalive) which saturates under bulk-write storm fan-out (16
-# concurrent writes × 10 enrichment calls = 160 concurrent LLM
+# concurrent writes x 10 enrichment calls = 160 concurrent LLM
 # requests per process, well over the keepalive budget). See
 # ``common/llm/constants.py`` for full rationale.
-from common.env_utils import (  # noqa: E402
+from common.env_utils import (
     clamp_keepalive,
     read_float_env,
     read_int_env,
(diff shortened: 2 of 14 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #696, opened 2026-08-03
fix(common): converge common/ruff.toml with enterprise's stricter config

Eldad-Caura:
Follow-up to #685. **Land this before caura-memclaw-enterprise's companion PR** — that repo's drift checker compares against OSS `main`, so its PR stays red until this merges.

## Why

#685 pinned `common/`'s ruff contract by mirroring core-api's rules. It then emerged that **enterprise had already solved the same problem independently and earlier**, in `e6d12110` ("deps: take the ruff 0.15.22 → 0.16.0 bump", #898) — same root cause, same fix shape. Its config even quantifies the expansion: **59 → 413 default rules**.

The two configs disagreed, and enterprise's is stricter. It keeps a short global `ignore` and uses targeted per-file-ignores *with a written reason for each*, where OSS globally ignored 18 more rules:

`ASYNC109, B007, B023, B905, E402, E701, PTH103/111/120, RUF005/006/034/059, SIM103/105/108, UP042, UP047`

`select` was already identical on both sides.

## What this does

Adopts enterprise's shape as the shared config, so **one file governs `common/` in both repos** and enterprise can classify it `identical` in its vendoring manifest — after which the drift checker guarantees they cannot diverge again.

Two deliberate details:

- **`target-version = "py312"`, not enterprise's `py313`.** `common/` runs on py312 here and py313 there, so the shared target must be the **lower** bound, or pyupgrade suggests rewrites that break this repo.
- **The per-file-ignores are the union across both trees**, which are *not* the same file set — enterprise's `common/` also contains `email_service.py` and `license/*`. An entry with no matching file is inert, so each repo carries a few patterns that only matter in the other. That is the price of a single shared file, and beats two configs drifting apart. Called out here because it will look odd in review.

## Consequences under the newly-enforced rules

- **Restored the two `noqa: E402` that #685 removed** as `RUF100` "unused". They were unused only relative to OSS's looser ignore list — `E402` is enforced under the shared config, so both late imports need them back. Removing them was right for the config at the time and wrong for this one.
- **Dropped the two `# noqa: B027`** #685 added to `EventBus.start`/`stop`. The shared config per-file-ignores `events/base.py` with the same reasoning, which is the cleaner expression of it — and enterprise had already written that entry.
- **`SIM108` in `_parse_temporal`: applied the ternary.** Both branches are simple assignments to the same name with no interleaved comments, so it is behaviour-identical. Ruff marks `SIM108` unsafe in general, so I applied and read this by hand rather than trusting `--fix`. OSS previously ignored `SIM108` globally; fixing it is the honest way to meet the stricter rule instead of re-adding the ignore and partially undoing the convergence.
- **`UP042`/`UP047`/`ASYNC109` are per-file-ignored with reasons, not fixed.** They would rewrite base classes and public generic signatures on widely imported shared types — its own deliberate decision, not something to bundle into a tooling change. That is the call enterprise already made, in the same words.

## Follow-up left deliberately

Enterprise's `structlog_config.py` entry notes that `SIM105`/`PTH120`/`PTH103` are real findings whose fix belongs upstream *here*. Doing that would let the entry disappear, but it is a code change rather than a tooling one, so it is out of scope.

## Verification

- `ruff check common/` clean; all pre-existing ruff scopes clean; vendored `--isolated` gate still passes.
- `common/ruff.toml` verified **byte-identical** to the enterprise copy via `cmp`.
- **4068 passed, 3 skipped, 1 xfailed, 3 xpassed, 0 failed** — including 176 targeted temporal/enrichment tests, since `_parse_temporal` changed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -130,13 +130,21 @@ jobs:
         run: uv run ruff check core-storage-api/tests/
 
       - name: Run Ruff linter (common)
-        # ``--config`` is explicit on purpose. common/ had NO ruff config above it
-        # (no ruff.toml, no root pyproject.toml), so ruff fell back to built-in
-        # DEFAULTS here — meaning the rule set was "whatever the installed ruff
-        # version ships," and it changed under us on the 0.16.0 bump. common/ruff.toml
-        # pins it; naming the config keeps this step immune to future config
-        # discovery changes, the same reasoning as ``--isolated`` below. See #684.
-        run: uv run ruff check common/ --config common/ruff.toml
+        # common/ had NO ruff config above it (no ruff.toml, no root
+        # pyproject.toml), so ruff fell back to built-in DEFAULTS here — the rule
+        # set was "whatever the installed ruff version ships," and it changed
+        # under us on the 0.16.0 bump. common/ruff.toml pins it (#684).
+        #
+        # DO NOT add ``--config common/ruff.toml``. #685 did, reasoning that
+        # naming the config made the step immune to discovery changes; it instead
+        # broke every per-file-ignore silently. Those patterns are relative to the
+        # config's own directory under discovery (so ``llm/retry.py`` means
+        # ``common/llm/retry.py``), but ``--config`` resolves them against the CWD
+        # instead — from the repo root nothing matches, all per-file-ignores
+        # become inert, and the step fails on exactly the findings they exempt.
+        # Discovery is also what enterprise's equivalent step uses, which matters
+        # because the config is byte-identical across both repos (#696).
+        run: uv run ruff check common/
 
       - name: Run Ruff formatter check (core-api)
         run: uv run ruff format --check core-api/src/
--- a/common/embedding/constants.py
+++ b/common/embedding/constants.py
@@ -56,7 +56,7 @@ EMBEDDING_RETRY_DELAY_S: float = float(os.environ.get("EMBEDDING_RETRY_DELAY_S",
 # concurrent writes x 10 enrichment calls = 160 concurrent LLM
 # requests per process, well over the keepalive budget). See
 # ``common/llm/constants.py`` for full rationale.
-from common.env_utils import (
+from common.env_utils import (  # noqa: E402 — intentional late import, see module docstring
     clamp_keepalive,
     read_float_env,
     read_int_env,
--- a/common/enrichment/service.py
+++ b/common/enrichment/service.py
@@ -44,10 +44,7 @@ def _parse_temporal(val: str) -> datetime | None:
     """Parse a temporal string to UTC datetime, returning None on failure."""
     try:
         dt = dateutil_parser.parse(val, fuzzy=False)
-        if dt.tzinfo is None:
-            dt = dt.replace(tzinfo=UTC)
-        else:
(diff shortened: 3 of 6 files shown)


Eldad-Caura:
CI failure fixed, and both review findings are refuted by that same CI run. Detail below.

## Root cause: `--config` was silently disabling every per-file-ignore

The 9 failures were *exactly* the 9 findings the per-file-ignores exist to exempt — UP042 x3, UP047 x2, ASYNC109 x2, B027 x2. Nothing else. That is the signature of the entries not applying at all.

Per-file-ignore patterns are relative to **the config file's own directory** under normal discovery, so `"llm/retry.py"` means `common/llm/retry.py`. But `--config` resolves them against the **CWD** — run from the repo root, nothing matches, every entry goes inert, and the step fails on precisely what they exempt.

`--config common/ruff.toml` was my addition in #685, on the reasoning that naming the config made the step immune to discovery changes. That reasoning was wrong in a way that only shows up once per-file-ignores exist. Fixed by dropping the flag:

```
uv run ruff check common/        # discovery — what enterprise's step also uses
```

Reproduced both directions locally:

| invocation | result |
|---|---|
| `ruff check common/ --config common/ruff.toml` | 9 errors (reproduces CI) |
| `ruff check common/` | **All checks passed** |

Matching discovery also matters now that the config is byte-identical across both repos — enterprise runs `uv run ruff check common/`, so a divergent invocation here would mean the same file behaves differently in the two repos. Added a comment at the step saying not to re-add the flag, with the reason.

## Both findings do not reproduce

Neither appears among the 9 CI errors, and neither reproduces locally when targeted directly:

**Missing `__init__.py` F401/E402 per-file-ignore** — `ruff check common/ --select F401` → **All checks passed**. `F401` appears **zero** times in the failing run. The finding also reasons across repos: it cites `common/events/__init__.py` having bare re-exports without `__all__`, but `events/*` is `policy=manual` in enterprise's vendoring manifest, so that file legitimately differs between the two trees — the enterprise copy is not evidence about this one.

**Missing `known-first-party` after `[lint.isort]` removal** — `ruff check common/ --select I001` → **All checks passed**. Zero I001 in the failing run. I had also measured this before pushing: identical finding counts with and without the `[lint.isort]` section. Mechanically it holds up — the config lives *in* `common/`, so `from common.*` resolves as first-party without the declaration. Enterprise runs this same config, with `"I"` selected and no isort section, and is clean.

I have not applied either. Happy to add both as belt-and-braces if you would rather carry them, but as written they suppress findings that do not occur, and the isort section would be a byte-difference from the enterprise copy that `check_vendored_drift.py` would then fail on.

## Verification after the fix

- `uv run ruff check common/` (verbatim CI command, from repo root) — clean.
- `core-api/src/`, `core-storage-api/src/`, `core-storage-api/tests/` — clean; vendored `--isolated` gate passes.
- `actionlint` at parity with `main` (same single pre-existing finding).
- Also confirmed the 7 format-dirty files in OSS `common/` are **none of the 9 vendored ones**, so enterprise's stricter `ruff format --check common/` step is unaffected by them.

Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please pprove

erni-a (approved):

Issue #687, opened 2026-08-02
search does not return a freshly-written memory that exactly matches the query


Eldad-Caura:
## Symptom

A memory written through the broker is retrievable by `memory_list` within ~39 seconds, but `memory_search` does **not** return it — while returning *other, older* rows from the same tenant for the same query. Observed repeatedly on staging; prod runs the same code and the same reader/writer topology.

Measured with a probe that polled both legs each iteration:

```
t+1s    listed=false  searchable=false
t+28s   listed=false  searchable=false
t+39s   listed=TRUE   searchable=false     ← in the cloud (matches the broker's 30s WAL drain)
…
t+115s  listed=true   searchable=false     ← still not searchable
```

`memory_list` records carry `created=`, which confirms the write was always prompt — five markers, each persisted within ~1–2 minutes of its run:

| content marker | `created=` |
|---|---|
| `broker-vm-e2e-da8d7d98…` | 09:00:01Z |
| `broker-vm-e2e-5e2adca5…` | 11:45:26Z |
| `broker-vm-e2e-4a913d9f…` | 13:08:33Z |
| `broker-vm-e2e-695665ff…` | 13:17:25Z |

The sharpest form: a search whose query is **byte-identical to a row's entire `content`** returns `Found 2 memories` where neither hit is that row — both are older rows with different content.

## Impact

**Read-your-own-write does not hold.** An agent that saves a memory cannot retrieve it via search for at least ten minutes. For a product whose value is agents sharing common ground, that is a meaningful gap, and it is invisible to callers — nothing errors.

It also silently defeats e2e coverage: a save-then-search assertion passes on *prior* runs' rows, so a broken write path would still look green. That is how this was found.

## Ruled out

Each of these was checked and is **not** the cause:

- **Replication lag.** The reader service is backed by a separate AlloyDB read pool, but `gcp.alloydb.node.postgres.replay_lag` measured **1–2 ms** on both staging and prod, flat across the entire window including the exact minutes of the failing probes.
- **Index build.** `search_vector` is populated by a row-level trigger on INSERT (`001_initial_schema.py`), so FTS is live the moment the row commits.
[...]

PR #697, opened 2026-08-03
fix(search): keep FTS-only rows past the legacy cosine gate (#687)

Eldad-Caura:
Closes the coverage gap behind #687, and fixes a real bug the new test found on its first run.

## The contract

Under `deployment_mode=deferred` a memory is stored with `embedding IS NULL` and `metadata.embedding_pending=True`; a worker fills the vector seconds (prod) to minutes (staging) later. Three mechanisms exist so the row stays discoverable in that window:

1. CAURA-594's admission predicate — `or_(embedding IS NOT NULL, fts_guard)`
2. CAURA-679's FTS-only `similarity` for NULL-embedding rows
3. the pipeline post-filter exempting them from the cosine gate

**The legacy search path never got #3.**

## The bug

```python
rows = [r for r in rows
        if r.get("vec_sim") is None or float(r["vec_sim"]) >= _min_similarity]
```

The comment above it justified the `is None` branch as defensive, because "the scored_search SQL currently enforces `Memory.embedding IS NOT NULL`". **CAURA-594 ended that invariant.** And as the same comment conceded, the storage layer coerces a NULL-embedding row's cosine to `0.0` rather than NULL — so the defensive branch never fires for those rows and `0.0 >= threshold` drops them.

The guard documented its own defeat. It says it is there *"so that FTS-only rows aren't silently gated out by the cosine threshold if either invariant ever changes"* — and that is exactly what happened.

Fix: add the `has_embedding is False` exemption so the legacy path genuinely mirrors the pipeline post-filter it claims to mirror. `has_embedding` is the authoritative NULL signal precisely because `vec_sim == 0.0` is ambiguous with a genuinely orthogonal embedding.

## The missing test

Coverage of this contract stopped at the storage layer. `core-storage-api/tests/test_integration.py` pins it well — but it POSTs `embedding=None` **straight to core-storage-api, bypassing core-api**. And `grep -l embedding_pending` across the suites hits only write-pipeline tests: **nothing paired a pending embedding with a recall.** That is the hole #687 shipped through.

The new test writes through the real core-api path with the embedding deferred, then asserts recall returns the row. It is parameterised over both search implementations because `search_memories` dispatches on `_USE_PIPELINE_SEARCH` — covering only the default would have left the other path unpinned. **That parameterisation is what found the bug:** `[pipeline]` passed and `[legacy]` failed on the very first run, with the log showing `path=scored-search row_count: 1` — the SQL returned the row and the legacy path threw it away.

Its preconditions assert the row really has a NULL embedding and a populated `search_vector` that FTS-matches the query, so it cannot pass vacuously against an embedded row, and cannot fail misleadingly on a schema built by `Base.metadata.create_all` without migration 001's `search_vector` trigger (the fixtures do exactly that; CI's `alembic upgrade head` is what supplies it).

## Verification

| Check | Result |
|---|---|
| `tests/pipeline/` + `tests/test_integration_search.py` | **120 passed** |
| Inverting the new clause | `[legacy]` fails again, nothing else — the test earns its keep |
| `mypy src/` (core-api) | no issues in 189 files |
| `ruff check` (both files) | clean |
| `ruff format --check` | source clean; the test file's pre-existing drift is at lines 67–290, untouched by this change |

Run against a local pgvector 0.8.6 with `alembic upgrade head`, mirroring CI.

## What this does not fix

**#687's production symptom is on the pipeline path** (`_USE_PIPELINE_SEARCH = True`), and that path *passes* the new test under every shape I could reproduce locally: bare, the full MCP-recall argument set (`caller_agent_id`, `fleet_ids`, `recall_boost`, `graph_expand`, `entity_retrieval`), and against 25 competing embedded rows — the crowding scenario CAURA-679's own comment warns about, verified non-vacuous by asserting the competitors really embedded.

So the remaining divergence there is environmental — `tenant_config`, `search_profile`, or the rank provider — not pipeline logic. That narrowing is itself a result, and it stays tracked in #687. This PR fixes the defect that *is* in code and installs the test that will catch the next one.

### DIFF
--- a/core-api/src/core_api/services/memory_service.py
+++ b/core-api/src/core_api/services/memory_service.py
@@ -3437,13 +3437,26 @@ async def _search_memories_legacy(
     async with per_tenant_storage_slot("storage_search", tenant_id):
         rows = await sc.scored_search(search_data)
 
-    # Post-filter by min_similarity, then trim to top_k. The `vec_sim is None`
-    # branch is defensive: the scored_search SQL currently enforces
-    # `Memory.embedding IS NOT NULL` and the storage layer coerces None → 0.0,
-    # so this branch is never taken today. Left in place — mirrored in the
-    # pipeline post_filter step — so that FTS-only rows aren't silently gated
-    # out by the cosine threshold if either invariant ever changes.
-    rows = [r for r in rows if r.get("vec_sim") is None or float(r["vec_sim"]) >= _min_similarity]
+    # Post-filter by min_similarity, then trim to top_k.
+    #
+    # `has_embedding is False` is the clause that matters, and it was missing:
+    # the invariant the old comment here relied on ("the scored_search SQL
+    # enforces `Memory.embedding IS NOT NULL`") stopped holding at CAURA-594,
+    # which admits NULL-embedding rows that FTS-match via
+    # `or_(embedding IS NOT NULL, fts_guard)`. The storage layer coerces those
+    # rows' cosine to 0.0 rather than NULL — as that comment itself noted — so
+    # the `vec_sim is None` branch never fires for them and `0.0 >= threshold`
+    # gated out exactly the FTS-only rows CAURA-679 exists to keep discoverable
+    # during the deferred-embed window. `has_embedding` is the authoritative
+    # NULL signal for precisely this reason. Now genuinely mirrors the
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #699, opened 2026-08-03
test(search): pin the FTS-only candidate-window bug (#687)

Eldad-Caura:
Reproduces #687 locally. **Marked `xfail(strict=True)` rather than fixed** — both viable fixes change core ranking, and I'd rather you pick the tradeoff than have me guess it. Reasoning below.

## The mechanism

The scored search fetches `top_k * SEARCH_OVERFETCH_FACTOR` candidates ordered by score. A NULL-embedding row scores on `fts_score` alone, while embedded rows near the query score far higher — so it never survives the `LIMIT`.

`PostFilterResults` exempts NULL-embedding rows from the cosine gate, and #697 added the same exemption to the legacy post-filter. **Both run downstream of the cut**, so the mechanism CAURA-594/CAURA-679 built for exactly this window is unreachable.

Measured on prod:

| | |
|---|---|
| tenant size | 791 memories |
| embedded row scores | 0.35 – 0.39 |
| FTS-only row score | **0** (reported) |
| candidate window | 10 (`top_k=5` × factor 2) |
| agent `search_profile` | **empty** — every value from constants |

So this is code, not configuration.

## Why every existing test missed it

They run in tenants of a handful of rows, where a 10-candidate window excludes nothing — **including the test I added in #697**. That's the honest lesson: that test pinned the contract at the right layer but in conditions where the bug cannot appear.

The reproduction controls everything except the window: competitors carry the *query's own* embedding, so they legitimately outrank on cosine, and their content omits the token, so the subject row is the **only** FTS match and **is** exempt from the cosine gate. It fails with:

> FTS-only row cut by the candidate window (#687): it is the ONLY row whose content contains 'zqxjvbn…', its search_vector FTS-matches, and it is exempt from the cosine gate — but 10 higher-cosine rows fill the top_k(5) * SEARCH_OVERFETCH_FACTOR(2) window first. search returned 5 row(s), none of them the subject.

## Why I didn't fix it here

**The cheap fix doesn't work.** Flooring the FTS-only similarity to `min_similarity` fails on the numbers: min is **0.3**, the measured embedded noise floor is **0.35–0.39**. The row still loses. There is no constant that works, because `ts_rank_cd/(1+ts_rank_cd)` is not on a comparable scale to cosine — a row whose entire content *is* the query scores ~0.05.

That leaves two real options:

1. **Normalise `fts_score` onto the cosine scale.** Principled, and fixes the root scale mismatch — but it reprices *every* query. This repo measures such choices (`SIMILARITY_BLEND` cites a LoCoMo sweep), so it wants a benchmark run, not a judgement call.
2. **Reserve candidate and result slots for FTS-only rows.** Blast radius bounded to the transient deferred window, and arguably what users want (read-your-own-write) — but it spans the SQL, the pipeline post-filter, *and* the legacy path, and it displaces one result during the window.

Shipping either unbenchmarked into the path that defines the product's recall quality seemed like the wrong call to make alone.

`strict=True` so the marker cannot rot: the moment a fix lands, this fails loudly and forces its removal.

## Verification

| Check | Result |
|---|---|
| `tests/pipeline/` + `tests/test_integration_search.py` | 120 passed, 1 xfailed |
| Test genuinely reproduces | fails with the message above when the marker is removed |
| `mypy src/` (core-api) | no issues, 189 files |
| `ruff check` / `format --check` | 1 pre-existing issue each, identical to `origin/main`; all reformat hunks are ≤ line 448, my addition starts at 492 |

Run against local pgvector 0.8.6 with `alembic upgrade head`, mirroring CI — which matters here: the fixtures build the schema with `Base.metadata.create_all`, so without alembic the `search_vector` trigger from migration 001 is absent and any FTS test fails for the wrong reason.

### DIFF
--- a/tests/test_integration_search.py
+++ b/tests/test_integration_search.py
@@ -8,6 +8,7 @@ scoring blend, entity matching — end-to-end via the service layer.
 """
 
 import hashlib
+import uuid
 from datetime import datetime, timedelta, timezone
 
 import pytest
@@ -18,6 +19,7 @@ from core_storage_api.services.postgres_service import get_session
 from core_api.clients.storage_client import get_storage_client
 from core_api.constants import (
     FRESHNESS_DECAY_DAYS,
+    SEARCH_OVERFETCH_FACTOR,
 )
 from common.models.memory import Memory
 from common.embedding import fake_embedding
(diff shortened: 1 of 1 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #700, opened 2026-08-03
fix(search): reserve candidate and result slots for FTS-only rows (#687)

### LINKED ISSUE #687: search does not return a freshly-written memory that exactly matches the query

Eldad-Caura:
Fixes #687 — option 2 of the three we weighed, plus removal of #699's `xfail(strict=True)` marker.

## Why the exemptions never fired

CAURA-594 admits a NULL-embedding row when it FTS-matches; CAURA-679 scores it on `fts_score` alone so the blend can't bury it. Neither helps it survive the two places that cut by **rank**:

1. storage's candidate `LIMIT` (`top_k * SEARCH_OVERFETCH_FACTOR`)
2. the head slice trimming to the caller's `top_k`

`fts_score` is `ts_rank_cd/(1+ts_rank_cd)` — single digits of a percent for one term — against embedded rows scoring far higher. Measured on prod: **791 memories, embedded rows 0.35–0.39, the FTS-only row 0, a 10-row window.** The row was gone before the cosine-gate exemptions (pipeline's since CAURA-679, legacy's since #697) could run.

**No constant floor fixes this:** `min_similarity` is 0.3, *below* the measured embedded noise floor.

## The change

A small capped allowance in all three places — guaranteeing *discoverable*, not *ranked highly*:

- **`postgres_service`** — a dedicated candidate branch for `embedding IS NULL AND search_vector @@ query`, ordered by FTS relevance and separately capped, unioned into the scored CTE. Rows enter at their own true (low) score; the outer `ORDER BY` still places them last.
- **`post_filter_results`** and **`_search_memories_legacy`** — promote at most `FTS_ONLY_RESERVED_RESULTS`, and only when the head contains none already, displacing the same number of weakest results.

Blast radius is bounded twice: by the caps, and because NULL-embedding rows exist only until the backfill lands (~seconds on prod).

## The bug this nearly shipped with

My first version unioned the two selects directly. That let SQLAlchemy hoist the main branch's `ORDER BY … LIMIT` to the compound statement, **changing the candidate set for embedded rows** — two unrelated graph-fanout tests silently flipped xfail → xpass. Wrapping each operand in its own subquery fixes it, and I only caught it by diffing the xfail/xpass profile against a clean tree. Worth knowing if anyone touches this union again.

## Verification

| Check | Result |
|---|---|
| Full suite (mine) | **4057 passed, 1 xfailed, 3 xpassed** |
| Full suite (clean `main` baseline) | 4054 passed, 2 xfailed, 3 xpassed |
| Delta | exactly the 3 added tests + #699's marker removed; **same 3 pre-existing xpasses** |
| Both halves load-bearing | setting either constant to 0 reproduces the bug |
| Both search paths | saturated test parameterised over `_USE_PIPELINE_SEARCH`; both pass |
| Reservation bounds | unit test: promotes at most the cap, no-ops when the head already has one, untouched for ordinary result sets, result length unchanged |
| mypy | clean — core-api (189 files) and core-storage-api (74) |
| ruff check / format | clean on every file touched; verified against in-place baselines, since comparing via `/tmp` copies misreads isort |

Run against local pgvector 0.8.6 with `alembic upgrade head` (the fixtures use `Base.metadata.create_all`, so without alembic the migration-001 `search_vector` trigger is absent and any FTS test fails for the wrong reason).

## Not done here

**Normalising `fts_score` onto the cosine scale** — option 1. It's the deeper fix and would likely improve keyword recall generally, but it reprices every query and belongs behind a benchmark run rather than my judgement. Worth its own issue.

Also worth a follow-up: nothing tells a caller their just-written memory isn't searchable yet. `embedding_pending` is already in the write response metadata, but undocumented — the cheap half of option 3.

### DIFF
--- a/core-api/src/core_api/constants.py
+++ b/core-api/src/core_api/constants.py
@@ -537,6 +537,7 @@ FTS_BOOST_MAX_TOKENS = 3  # queries with more meaningful tokens than this stay a
 FTS_BOOST_SPECIFICITY_RATIO = 0.4  # strict >; at N=2 this means >=1 specific token triggers boost
 SIMILARITY_BLEND = 0.85  # base_score = SIMILARITY_BLEND * similarity + (1 - SIMILARITY_BLEND) * weight (raised from 0.75 — LoCoMo sweep showed +13pp recall)
 SEARCH_OVERFETCH_FACTOR = 2  # fetch top_k * N candidates from storage, trim to top_k after min_similarity filter — gives post-filter headroom
+FTS_ONLY_RESERVED_RESULTS = 1  # #687: result slots held for rows that FTS-match but are not embedded yet (transient deferred-embed window); 0 restores the pre-#687 head-slice behaviour
 # A26: recall_count is bumped for every RETURNED row, used or not (see
 # TrackRecalls + memory_increment_recall), and feeds recall_boost back into the
 # rank score — a self-reinforcing "returned → boosted → returned" loop with no
--- a/core-api/src/core_api/pipeline/steps/search/post_filter_results.py
+++ b/core-api/src/core_api/pipeline/steps/search/post_filter_results.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 from core_api.pipeline.context import PipelineContext
 from core_api.pipeline.step import StepOutcome, StepResult
 from core_api.pipeline.steps.search.retrieval_types import RetrievalStrategy
+from core_api.search_trim import trim_reserving_fts_only
 
 
 class PostFilterResults:
@@ -37,6 +38,24 @@ class PostFilterResults:
         # Trim to the user-requested top_k (storage returned top_k * overfetch_factor)
         final_top_k = ctx.data.get("final_top_k")
         if final_top_k is not None:
-            filtered = filtered[:final_top_k]
+            filtered = trim_reserving_fts_only(filtered, final_top_k, _is_fts_only)
         ctx.data["filtered_rows"] = filtered
         return None
+
+
+def _is_fts_only(row) -> bool:
+    """True for a row storage admitted on FTS alone, with no embedding yet.
+
+    Same test as the cosine-gate exemption above, which has used this form since
+    CAURA-679 — kept identical so the gate and the reservation cannot disagree
+    about what "FTS-only" means.
+
+    ``row`` is a ``SimpleNamespace``, hence ``getattr``; the legacy path in
+    ``memory_service`` holds dicts and reads the same field with ``.get``. The
+    value is always a real ``bool``, never ``None``: storage computes it as
+    ``Memory.embedding.is_not(None)`` — a non-nullable SQL boolean — and
+    ``execute_scored_search`` carries it onto the namespace with a ``True``
+    default. So truthiness and ``is False`` cannot diverge on any reachable
+    input here.
+    """
+    return not getattr(row, "has_embedding", True)
--- a/core-api/src/core_api/search_trim.py
+++ b/core-api/src/core_api/search_trim.py
@@ -1,0 +1,61 @@
+"""Head-slice trim that reserves result slots for FTS-only rows (#687).
+
+Lives at the package root, importing only :mod:`core_api.constants`, so both the
+pipeline post-filter step and the legacy search path can use it without an import
+cycle. One implementation on purpose: the two paths trim independently, and when
+this logic was duplicated a boundary bug had to be fixed in both copies.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Callable
+from typing import Any
+
+from core_api.constants import FTS_ONLY_RESERVED_RESULTS
+
+
+def trim_reserving_fts_only(
+    rows: list[Any],
+    top_k: int,
+    is_fts_only: Callable[[Any], bool],
+) -> list[Any]:
+    """Trim ``rows`` to ``top_k``, keeping one FTS-only row if any exist.
+
+    #687: exempting FTS-only rows from the cosine gate is not enough to make them
+    reachable. They score on ``fts_score`` alone — single digits of a percent for
+    one term — so with enough embedded rows above them a plain head slice drops
+    them, having already survived storage's candidate LIMIT only because that
+    reserves slots for them too.
+
+    The reservation is deliberately minimal: it promotes at most
+    ``FTS_ONLY_RESERVED_RESULTS`` rows, displacing the same number from the tail
+    of the head — the weakest results — and only when the head contains none
+    already. It never reorders anything. The population is transient: a row is
+    FTS-only just until the deferred embed backfill lands, after which it competes
+    on cosine normally.
+
+    ``is_fts_only`` is passed in because the two callers hold different row
+    shapes: the pipeline has objects (attribute access), the legacy path dicts.
+
+    Never consumes the whole head. A ``top_k=1`` caller (valid input —
+    ``schemas.py`` allows ``ge=1``) asked for their single best match, and
+    answering with only an FTS-pending stub in its place is worse than not
+    surfacing the stub at all. #687's contract is that such a row is
+    *discoverable*, which storage's candidate reservation still provides; it was
+    never that the row outranks the best result.
+    """
+    head = rows[:top_k]
+    if FTS_ONLY_RESERVED_RESULTS <= 0 or top_k <= 0:
+        return head
+    if any(is_fts_only(r) for r in head):
+        return head
+    # The ``- 1`` is what stops the promotion taking every slot: it keeps
+    # ``top_k - len(promoted)`` in the slice below at 1 or more, so at least one
+    # row that earned its place on score always survives.
+    budget = min(FTS_ONLY_RESERVED_RESULTS, top_k - 1)
+    if budget <= 0:
+        return head
+    promoted = [r for r in rows[top_k:] if is_fts_only(r)][:budget]
+    if not promoted:
+        return head
+    return head[: top_k - len(promoted)] + promoted
--- a/core-api/src/core_api/services/memory_service.py
+++ b/core-api/src/core_api/services/memory_service.py
@@ -78,6 +78,7 @@ from core_api.schemas import (
     MemoryOut,
     MemoryUpdate,
 )
+from core_api.search_trim import trim_reserving_fts_only
 from core_api.services.entity_extraction_worker import process_entity_extraction
 from core_api.services.entity_tokens import extract_entity_tokens
 from core_api.services.governance_gate import (
@@ -3457,7 +3458,14 @@ async def _search_memories_legacy(
         or r.get("vec_sim") is None
         or float(r["vec_sim"]) >= _min_similarity
     ]
-    rows = rows[:_top_k]
+    # #687: the same reservation the pipeline post-filter applies — exempting
+    # FTS-only rows from the cosine gate above does not make them reachable,
+    # because they score on fts_score alone and a plain head slice drops them once
+    # enough embedded rows sit above. Shared with that path via
+    # `trim_reserving_fts_only` so the two cannot drift; rows here are dicts, so
+    # the predicate reads `has_embedding` with `.get` (matching the gate above)
+    # where the pipeline uses attribute access.
+    rows = trim_reserving_fts_only(rows, _top_k, lambda r: r.get("has_embedding") is False)
 
     # Build results from storage API response
     memory_ids = [row.get("id") for row in rows if row.get("id")]
--- a/core-storage-api/src/core_storage_api/services/postgres_service.py
+++ b/core-storage-api/src/core_storage_api/services/postgres_service.py
@@ -235,6 +235,22 @@ def _entity_uf_union(parent: dict[UUID, UUID], rank: dict[UUID, int], a: UUID, b
         rank[ra] += 1
 
 
+# #687: how many candidate slots the scored search reserves for rows that
+# FTS-match but have no embedding yet. Deliberately small: the guarantee needed
+# is "discoverable", not "ranked highly", and the population is transient — rows
+# carry a NULL embedding only until the deferred backfill lands. Set to 0 to
+# restore the pre-#687 behaviour where such rows are cut by the candidate LIMIT.
+#
+# This also floors what the result layer can promote: core-api's
+# ``FTS_ONLY_RESERVED_RESULTS`` (currently 1) reserves result slots only among
+# rows that already reached it, and this branch is the only supply it is
+# guaranteed — the main branch carries FTS-only rows just when they earn a slot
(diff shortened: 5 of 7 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@cladue

Eldad-Caura:
Both findings from the previous round addressed — one applied, one respectfully declined with evidence. Re-pushed as a single amended commit (`94d6d43`), rebased onto current `main`.

### Applied — cross-package coupling at `_FTS_ONLY_RESERVED_CANDIDATES`

Documented at the definition site. I did narrow the wording from the suggested version, because the strict form (`must be >= FTS_ONLY_RESERVED_RESULTS`) isn't quite what holds: FTS-only rows also reach the pool via the **main** branch when they earn a slot on score, so `3` isn't a hard ceiling on how many are available to promote — it's the floor on what the result layer is *guaranteed*. The comment says that, and notes the constraint is documented rather than enforced because core-storage-api must not import core-api.

While moving it I also found the constant had been inserted **between** the `# Cached at import time…` comment and the `_MEMORY_VALID_FIELDS` that comment describes — so that comment read as though it documented the new constant. Relocating above the block reunites them. Thanks for pointing at that line; it surfaced a second, unrelated defect in my own diff.

### Declined — tightening `_is_fts_only` to `is False`

The suggestion was to change `post_filter_results._is_fts_only` from truthiness to strict identity to match the legacy path. I checked this against the code and it would make things worse:

1. **`_is_fts_only` is an exact extraction of the gate 14 lines above it**, at `post_filter_results.py:34` — `not getattr(row, "has_embedding", True)`. That gate is **pre-existing on `main`** (it has carried the clause since CAURA-679; confirmed with `git show origin/main:…`), and this PR did not introduce it. Changing only the helper would leave the reservation and the cosine gate *in the same file* disagreeing about what "FTS-only" means. That's a worse split than the cross-file idiom difference being objected to — each file is currently internally consistent, and the idiom difference is forced by row type (objects here via `getattr`, dicts in the legacy path via `.get`).

2. **The two forms cannot diverge in practice.** `has_embedding` originates as `Memory.embedding.is_not(None)` — a non-nullable SQL boolean, since `IS NOT NULL` never yields NULL — and reaches core-api as `has_embedding=row.get("has_embedding", True)` (`execute_scored_search.py:175`). So it is always a real `bool`; there is no `None`/`0`/`""` value for the operators to differ on. The change is a no-op on every reachable input.

3. On the claim that memory_service's comment says the paths "match" when they don't — the comment reads: ``` `has_embedding is False` is the flag (matching the gate above); the pipeline's row objects use the same field. ``` "The gate above" is the gate in *that same file*, which does use `is False`, and the pipeline does use the same *field*. Both clauses are accurate as written; the comment doesn't claim the two paths use the same operator.

Unifying both paths on one idiom is defensible as its own cleanup, but it would change the pre-existing CAURA-679 cosine gate's behaviour on a value that can't occur, which doesn't belong in a fix PR for #687.

### Verification on the rebased tip

|---|---|---|---|
| `main` (6cc0f67) | 4081 | 2 | 3 |
| this branch (94d6d43) | 4084 | 1 | 3 |

The delta is exactly accounted for: #699's `xfail(strict=True)` repro becomes 2 passing parametrized cases (pipeline + legacy), plus 1 new unit test — so +3 passed and −1 xfailed. The 3 xpasses are the same three pre-existing names in both columns, which is the check that caught the earlier near-miss where a bare `union()` silently changed the candidate set for embedded rows.

`ruff check` and `ruff format --check` clean, run separately and from **inside** each project (core-storage-api is `line-length = 110`; a root-level invocation passes while CI fails). `mypy` clean on both.

Eldad-Caura:
@claude

Eldad-Caura:
Round 2 — both Medium findings were real and are fixed, plus the Low suggestion. Re-pushed as a single amended commit (`cc23bc9`), rebased onto `main` after #705.

### The small-`top_k` displacement was a genuine bug

Confirmed before touching it, rather than from reading. With `top_k=1` and an FTS-only row below the cutoff, `keep = max(0, top_k - len(promoted))` reached 0 and the head was replaced outright:

```
top_k=1: [Row(100,emb=False)]   best-embedded-kept=False   ← the bug
top_k=2: [Row(0,emb=True), Row(100,emb=False)]   best-embedded-kept=True
```

`top_k=1` is valid input (`schemas.py`: `ge=1`), so a caller asking for their single best match got only an embed-pending stub instead. Promotion is now bounded by `min(FTS_ONLY_RESERVED_RESULTS, top_k - 1)`, leaving at least one earned slot standing:

```
top_k=1: [0E]  best_kept=True
top_k=2: [0E, 100F]  best_kept=True
```

Framing this against #687's contract: the promise is that such a row is *discoverable*, which storage's candidate reservation still delivers — never that it outranks the best result. At `top_k=1` those two goals are in direct conflict and the best match has to win.

### The cross-package invariant is now enforced, not just documented

You were right that a comment won't stop drift. `tests/pipeline/test_search_pipeline.py` now imports both constants and asserts `_FTS_ONLY_RESERVED_CANDIDATES >= FTS_ONLY_RESERVED_RESULTS`. The root suite has both packages on `pythonpath`, so this needs no new plumbing.

I kept the comment's weaker wording deliberately: `3` is the floor on what the result layer is *guaranteed*, not a hard ceiling, because FTS-only rows also reach the pool via the main branch when they earn a slot on score. The test asserts the direction that actually matters.

### Duplication — taken, and it was the root cause here

The trim now lives once in `core_api/search_trim.py`, taking the FTS-only predicate as a parameter; the pipeline holds row objects and the legacy path dicts, which was the only real difference between the two copies. Your observation that the duplication is *why* this boundary bug existed twice is exactly right — the fix landed once instead of twice. The module sits at the package root importing only `constants`, because `memory_service` defers its pipeline imports to avoid a cycle.

### Both new tests confirmed load-bearing

Mutated to watch them fail for the stated reason, inverting rather than deleting:

| mutation | result |
|---|---|
| budget widened back to `min(…, top_k)` | `assert [100] == [0]` |
| storage constant set to `0` | `assert 0 >= 1` |

### Verification

|---|---|---|---|
| `main` (8351077) | 4084 | 2 | 0 |
| this branch (cc23bc9) | 4088 | 1 | 0 |

+4 is exactly the added tests (2 parametrized paths, the bounded-reservation case, the cross-service invariant); −1 xfail is #699's marker removed. Now that #705 has landed `xfail_strict = true`, those zero xpasses are enforced rather than eyeballed. `ruff check` and `ruff format --check` clean per project, run separately and from inside each project; `mypy` clean on both.

### One earlier finding I'm still declining

From round 1: tightening `_is_fts_only` to `is False` while the cosine gate 14 lines above it keeps truthiness. `has_embedding` is a non-nullable SQL boolean reaching core-api as `row.get(..., True)`, so the two forms cannot diverge on any reachable input, and changing only the helper would split the file's own convention. I've documented that invariant on `_is_fts_only` instead so the next reader doesn't have to re-derive it.

Eldad-Caura:
@claude

Eldad-Caura:
Round 3 — the docstring finding was fair and is fixed (`992e959`, rebased onto `main` after #706).

You read it the way a future maintainer would, which is the way that counts. My intent was provenance — where the value on the namespace comes from — but written next to a `getattr` body, "arriving as `row.get(..., True)`" reads as a claim about *this* function's access pattern, which is wrong. Rewritten to separate the two:

- `row` is a `SimpleNamespace`, hence `getattr`; the legacy path holds dicts and reads the same field with `.get`.
- The value is never `None` because storage computes it as `Memory.embedding.is_not(None)` (non-nullable SQL boolean) and `execute_scored_search` carries it onto the namespace with a `True` default.

That second half is the load-bearing part, since it's the reason truthiness and `is False` cannot diverge here — the point behind declining the round-1 suggestion to change only this predicate. Stating it plainly beats making a future reader re-derive it.

Docstring-only change: full suite unchanged at 4088 passed / 1 xfailed / 0 xpassed, `ruff check` + `format --check` clean, `mypy` clean on 190 files.

Eldad-Caura:
@claude

Eldad-Caura:
Round 4 — both Low findings fixed (`0b61ab6`). One I resolved differently than suggested, and the `uv.lock` one turned out to be worth a closer look.

### Stale comment — real, and my own doing

Correct catch. The comment said "which is what bounds `keep` below", but `keep` stopped existing when I simplified the return to `head[: top_k - len(promoted)]`. A comment asserting something about a variable that isn't there is exactly the defect class I should be catching in my own diff.

I didn't take the suggested replacement verbatim, though, because it isn't accurate either: `budget` is *assigned* on the line the comment sits above, not bounded "below" it. What the `- 1` actually buys is a floor on the slice. So:

```python
# The ``- 1`` is what stops the promotion taking every slot: it keeps
# ``top_k - len(promoted)`` in the slice below at 1 or more, so at least one
# row that earned its place on score always survives.
budget = min(FTS_ONLY_RESERVED_RESULTS, top_k - 1)
```

### `uv.lock` — dropped from the branch entirely, not bumped

Checking the three values rather than two:

| | version |
|---|---|
| `pyproject.toml` on `main` | 2.27.0 |
| `pyproject.toml` on this branch | 2.27.0 |
| `uv.lock` on `main` | **2.24.1** |
| `uv.lock` on this branch | 2.26.0 |

So `main` is *already* out of sync with its own `pyproject.toml`, and the drift is not this branch's. More to the point: **this PR has no reason to touch `uv.lock` at all** — the change was incidental churn from a lock regeneration while setting up a local environment, and it has been in the diff since the first push.

Bumping it to 2.27.0 would smuggle an unrelated version edit into a search fix, on a line release-please owns. So I reverted the file to `main`'s content instead; `uv.lock` is now absent from the diff, which is down to 7 files, all load-bearing. The pre-existing `main` lock-vs-pyproject drift is left alone — it's real, but it isn't mine and it wants its own change.

### Verification

Comment-only plus a file revert: full suite unchanged at **4088 passed / 1 xfailed / 0 xpassed**. `ruff check` and `ruff format --check` clean on both projects, run separately and from inside each. `mypy` clean.

Eldad-Caura:
@claude

Eldad-Caura:

Claude review is clean (✅ no issues found) and all findings from the four earlier rounds are handled. CI green across the board: CI, pre-checks, dco, CodeQL (python + js/ts), Security Scan, and the image-version check.

Summary of what changed across review rounds, for context:
- Round 2 fixed a real `top_k=1` bug — the reservation could displace the caller's single best result outright — and enforced the cross-service constant invariant with a test rather than a comment.
- The trim logic was extracted into one shared helper (`core_api/search_trim.py`) used by both the pipeline and legacy paths, since the duplication was why that boundary bug existed in two copies.
- Round 4 dropped an incidental `uv.lock` edit that had no business in a search fix; the diff is 7 files, all load-bearing.

Verified locally against pgvector with `alembic upgrade head`: full suite **4088 passed / 1 xfailed / 0 xpassed** vs `main` at 4084 / 2 / 0 — the delta is exactly the four added tests plus #699's `xfail(strict=True)` marker removed, as that marker required.

erni-a (approved):

PR #705, opened 2026-08-04
test(search): make three flaky xfail/xpass search tests deterministic

Eldad-Caura:
## What

Three integration tests carried `xfail(reason="Fake embeddings produce low similarity — search returns empty")` and **xpassed** in a full run. This makes them deterministic and removes the last xpass from the suite.

They matter because the xfail/xpass profile is the signal that caught #700's near-miss — where a bare `union()` silently changed the candidate set for embedded rows, invisible to every test written for it and visible only as an xpass count of 3 vs 1 against clean `main`. Noise in that signal is expensive.

**The stated reason was a misdiagnosis in both cases.**

## `tests/test_p3_3_graph_fanout.py` — wrong fixture, not low similarity

These wrote setup through the `db` fixture, which opens a transaction and rolls it back at teardown — per its own docstring, *"session.commit() flushes data without committing the outer transaction"*. `search_memories` runs on a **separate connection**, so it never saw those rows. Storage logged `row_count: 0`: they failed at the SQL layer, nowhere near the similarity gate.

They xpassed in a full run only because `assert len(results) >= 1` was satisfied by **other tests' committed rows** — `tenant_id` is a single session-wide constant and 13 files commit into it via `sc`. So even when green, they asserted nothing about fan-out.

Now they write via `sc` — the fixture documented for precisely this ("use this instead of `db.add()` when the data needs to be visible … e.g. search") — using the bulk insert and bulk link helpers, and assert on **their own memory ids** instead of a bare count.

They also get a tenant of their own. A unique query token is **not** sufficient isolation, and this is measured rather than assumed: `fake_embedding` is a bag-of-words hash, not a near-orthogonal one, and an unrelated committed row (`"User prefers dark mode in the editor"`) scores cosine **0.3341** against a random token — above the 0.3 `min_similarity` gate. A foreign row can still take a top_k slot.

## `tests/test_visibility.py` — simply a stale marker

Already wrote via `sc`, and xpasses in isolation on a clean database (3/3 runs). Nothing flaky; the marker was obsolete. Dropped.

## So it cannot recur silently

- **`xfail_strict = true`** in `pytest.ini`. An xfail that starts passing is now a failure rather than a green XPASS nobody reads. Verified by probe — re-adding a stale marker turns that test `FAILED`. Both remaining markers were checked against the setting: one is already `strict=True`, and the other (dedup-vs-visibility) genuinely xfails, reproducibly and in isolation. **This is the one change here with suite-wide reach — flagging it explicitly; happy to drop it if you'd rather it landed on its own.**
- **Corrected the `tenant_id` docstring**, which claimed *"Unique tenant ID per test module"* while returning one session-wide constant. That is the misconception these tests were built on, and 70 tests across 11 files read it.

## Verification

Both rewritten assertions confirmed load-bearing by mutation — inverting, not deleting: linking 4 of 5 memories fails the set-equality check; searching at `top_k=3` fails the top_k assertion. Restored, both green.

I also confirmed the graph-boost path these tests exist to cover **does** fire, rather than assuming it — `entities.search_vector` is populated by migration 001's trigger (`'populartopic…':1`), and `fts_search_entities` matches the entity, so the boost and its `GRAPH_MAX_BOOSTED_MEMORIES` cap are genuinely exercised. The helper docstring records that this depends on the DB having been migrated: `Base.metadata.create_all` alone leaves the column NULL and would skip the boost silently. The raw `UPDATE entities SET search_vector` that came out was redundant with that trigger.

|---|---|---|---|
| `main` | 4081 | 2 | 3 |
| this branch | 4084 | 2 | **0** |

The +3 are these three tests now passing for real. `core-storage-api/tests/` 103 passed. Root `tests/` is not in CI's ruff scope, so nothing was reformatted there.

## Noted, deliberately not in scope

- `tests/test_p2_semantic_dedup.py:374` has the same `db`-write-then-storage-read trap; its "100 memories" / "1000 memories" latency labels measure whatever is committed in the shared tenant. It is `@pytest.mark.benchmark` and CI excludes benchmarks, so it misleads local numbers only.
- 64 of 128 tests that request `db` never use it, which makes them *look* transactionally isolated while committing through the service layer. That is the leftover-row generator behind the old xpasses. A suite-wide cleanup is a bigger change than this fix.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,4 +1,11 @@
 [pytest]
+# An xfail that starts passing is a failure, not a silent pass. Without this,
+# a stale marker keeps reporting green as XPASS and the xfail/xpass profile
+# stops being a usable signal — which is how three markers went stale here,
+# two of them masking fixtures that wrote to a rolled-back transaction and
+# asserted against other tests' leftovers. Both remaining markers were checked
+# against this setting; use `strict=False` on a marker that genuinely needs it.
+xfail_strict = true
 asyncio_mode = auto
 asyncio_default_fixture_loop_scope = session
 asyncio_default_test_loop_scope = session
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -270,7 +270,15 @@ def storage_http(_patch_storage_client):
 
 @pytest.fixture
 def tenant_id():
-    """Unique tenant ID per test module."""
+    """The run's shared tenant ID — one constant for the whole session.
+
+    NOT unique per test or per module: ``TENANT_ID`` is evaluated once, when this
+    module is imported. Rows any test commits through ``sc`` therefore stay
+    visible to every later test under this id, so an assertion like
+    ``len(results) >= 1`` can be satisfied by another test's data. Where a test
+    needs real isolation, generate its own ``f"test-tenant-…-{uuid4().hex[:8]}"``
+    (keep the ``test-tenant-`` prefix — session teardown sweeps on it).
+    """
     return TENANT_ID
 
 
--- a/tests/test_p3_3_graph_fanout.py
+++ b/tests/test_p3_3_graph_fanout.py
@@ -9,11 +9,13 @@ from uuid import UUID
 
 import pytest
 
+from common.embedding import fake_embedding
 from core_api.constants import (
     GRAPH_HOP_BOOST,
     GRAPH_MAX_BOOSTED_MEMORIES,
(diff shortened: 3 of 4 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #706, opened 2026-08-04
docs(write): document embedding_pending and the strong-mode opt-out

Eldad-Caura:
## Why

A write returns `metadata.embedding_pending: true` when the row was stored without an embedding — but nothing documented it, so a caller had no way to know its memory isn't semantically searchable yet, or what to do about it. The failure mode is quiet and expensive: the search comes back empty, the agent concludes the write failed, and writes it again.

## What's actually true

Verified end to end with a real write before documenting any of it, rather than reading it off the code:

| deployment | `write_mode` | response `metadata` |
|---|---|---|
| `deployment_mode=inline` (OSS local) | `fast` | `{enrichment_pending: True}` |
| embedding deferred (SaaS-like) | `fast` | `{enrichment_pending: True, embedding_pending: True}` |
| embedding deferred | `strong` | `{}` |

Three things follow, and all three are now stated:

- The flag means the memory **is** reachable by keyword and via `GET /memories`, but won't match a *semantic* search until `core-worker` backfills — **~15–20s on production, over 10 minutes on staging**, whose backfill is slower.
- `write_mode="strong"` embeds inline regardless of deployment mode, and is the supported read-your-own-write opt-out. It costs the embedding provider call on the request path — exactly what fast mode's sub-2s p99 visibility SLA exists to avoid — so it's a per-write choice, not a default to flip.
- **`fast` does not unconditionally defer.** It follows the deployment's embedding mode, so a deployment that embeds inline (the OSS local default) never sets the flag at all. Worth stating precisely, since the obvious summary of this feature is wrong.

## Where

- `MemoryOut.metadata` — both pending flags, what each implies, and the opt-out.
- `MemoryCreate.write_mode` — promoted from a bare `#` comment to a `Field` description, so it now appears in the OpenAPI schema.
- The **`memclaw_write` ToolSpec description**, which is the source of truth for MCP `tools/list`. Worth flagging for reviewers: the handler docstring in `mcp_server.py` is *not* what agents read — editing only that would look right and change nothing. The docstring now says so, and carries the maintainer-facing detail.
- README's async-write note, which said "for a moment" — off by a factor of ~30 on prod, and far more on staging.

Tool-surface tokens are paid on every agent call, so the agent-facing wording is deliberately compact: **4885 → 4959** cl100k for `tools/list`, against the 5200 ceiling in `test_mcp_token_budget.py`. The regenerated fixtures and `plugin/tools.json` diff only in `memclaw_write`'s description and its `write_mode` param — no incidental churn.

## Also: the documented regeneration path didn't exist

`tests/fixtures/README.md` has pointed at `capture_baselines.py` for a while, but the script was missing — so refreshing a baseline meant hand-rolling the capture and matching its serialisation by trial (`indent=2`, `ensure_ascii=False`, trailing newline; anything else rewrites all three files as a whole-file diff). Added it, and it prints the token count so a budget regression surfaces locally instead of in CI. Also corrected that README's stale **5000**-token ceiling to point at the constant (it's 5200), and noted that `plugin/tools.json` has a separate generator.

## Verification

Full suite **4084 passed / 2 xfailed / 0 xpassed** — identical to `main`, no test delta. `ruff check` and `ruff format --check` clean, run separately and from inside `core-api`; `mypy` clean.

I skipped `/simplify` here: the change is documentation wording plus one dev-tooling script, with no behavioural logic. Every property asserted in the new prose was confirmed by the probe above rather than by reading.

## Follow-up

The broker's `memory_save` tool description needs the same note; it lives in the private `memclawd` repo, so it's a separate PR.


Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #708, opened 2026-08-04
fix(rerank): distinguish permanent config faults from transient failures

### LINKED ISSUE #704: Reranking: remote provider doesn't chunk to the sidecar's batch cap, and a permanent 4xx is indistinguishable from a transient timeout
Two related defects in the reranking component (#669, #680), found while operating a self-hosted TEI reranker sidecar with `RANK_PROVIDER=remote`. Together they make a pure **configuration** error present as "reranking just isn't helping" — recall never breaks, so there is no failure anyone would notice.

Filing as one issue because the second defect is what hides the first, and fixing observability first makes the chunking work verifiable.

## 1. `remote.py` sends the whole candidate pool in one request, ignoring the sidecar's cap

`rerank_results.py` slices `rows[:RANK_CANDIDATE_LIMIT]` (default **50**) and `RemoteReranker.rank()` posts all of them as a single `{query, texts}` body.

But a TEI reranker's default `--max-client-batch-size` is **32**. So with stock config on both sides, any search returning more than 32 candidates is rejected outright:

```
HTTP 413 {"error":"batch size 33 > maximum allowed batch size 32","error_type":"Validation"}
```

Boundary confirmed against a real TEI CPU deployment: 32 succeeds, 33+ rejected. (413 on TEI 1.7; newer TEI returns 422 for the same validation error.)

The operator-side fix is to raise the sidecar's cap, but nothing in the library enforces or documents that coupling, and the default-vs-default combination is broken. Since a cross-encoder scores each `(query, text)` pair **independently**, chunk-then-merge-by-score is semantically exact here and would make the failure structurally impossible regardless of either side's config.

**Caveat for whoever implements it:** the chunks must be issued **concurrently**. Measured warm latency for a 50-text pool against CPU MiniLM-L6 is ~285ms p50, against a `RANK_TIMEOUT_SECONDS` default of 0.5s. Two *sequential* chunks would blow that deadline and convert a silent 413 into a silent timeout — same user-visible outcome, harder to diagnose.

## 2. `_service.py` treats a permanent config error exactly like a transient one

`RemoteReranker.rank()` calls `resp.raise_for_status()`, and the 413 is then caught by the intentionally-broad handler in `common/ranking/_service.py`:

```
WARNING  Ranking attempt 1/1 failed
ERROR    Ranking failed after 1 attempt(s); keeping first-stage order
```

Degrading to first-stage order is correct and I'm not proposing to change it. The problem is that a **permanent, config-class 4xx** — which will fail identically on every future request until someone changes a flag — is reported at the same severity, through the same message, and consumes the same retry budget as a timeout or a 5xx.

Consequences in practice:

- The log line names no status code, no response body, and no provider URL, so the actual cause (`batch size 33 > maximum allowed 32`) never surfaces. It has to be rediscovered by hand from the sidecar's own logs.
- A misconfiguration is indistinguishable from load. We hit *both* failure modes on the same deployment within 24 hours — one a batch-cap rejection, one genuine cold-start timeouts — and the log output was identical.
- Retrying a 413 is pure waste; it cannot succeed.

### Suggested shape

- Classify by status: `413`/`404`/`422` (and arguably all 4xx except `429`) are permanent → log at `error` **with the status code, response body, and provider base URL**, and skip the retry budget. Timeouts, `429`, and 5xx stay transient at `warning`.
- Consider logging the candidate-pool size on a permanent failure; it's the single most useful datum for the cap mismatch and is already in hand at the call site.

Happy to send a PR for either half if the approach sounds right — the classification half is small and self-contained, and it's the piece that would have turned a multi-hour investigation into a single log line.

Eldad-Caura:
Closes #704 (the classification half; chunking stays open — see Scope below).

## The problem

Reranking degrades to first-stage order on any failure, by design — recall must never break. That posture is unchanged here. What was missing is the distinction between the two *reasons* it can fail, which want opposite handling.

A permanent, configuration-class fault was reported exactly like a transient one:

```
WARNING  Ranking attempt 1/1 failed
ERROR    Ranking failed after 1 attempt(s); keeping first-stage order
```

No status code, no response body, no endpoint. So a misconfiguration was indistinguishable from load, and the actual cause never reached the log.

**The failure that motivated this:** a TEI reranker admits **32** texts per request by default, but core-api sends up to `RANK_CANDIDATE_LIMIT` (**50**) in ONE call — so every search with a larger pool came back `413`. `raise_for_status()` turned that into a generic exception, the broad handler swallowed it, and the search quietly served first-stage order. Recall never broke, so it read as *"reranking isn't helping"*. It took hours and the sidecar's own logs to find — twice.

## The change

New `common/ranking/errors.py` with `PermanentRankError`. Providers, which know their own transport, classify; `_service.py` stays transport-agnostic and only reacts to the type (it already dispatched on `TimeoutError`, so no new concept there). **Anything not explicitly classified stays transient**, so a new failure mode can never silently stop retrying.

- **`remote.py`** — 4xx except `408`/`429` is permanent; those two are about *when* the call happened, not what it contained. 5xx and network errors keep the existing retry path untouched. A `200` whose body isn't a ranked list is also permanent: that means `RANK_BASE_URL` points at something that doesn't speak `/rerank`. The message carries endpoint, status, candidate count and the service's own words, plus a specific hint for 413 — the batch-cap coupling isn't guessable from the status code.
- **`local.py`** — a missing `sentence-transformers` is permanent by construction. The default image ships without torch, so it's the likely result of flipping `RANK_PROVIDER=local`, and it was being retried at warning level.
- **`.env.example`** — documents the `RANK_CANDIDATE_LIMIT` ↔ sidecar-cap constraint where the value is actually set, so it's preventive rather than post-mortem.

### Log-once, not log-per-search

A permanent fault recurs on every search by definition, so the full ERROR fires **once per condition** and repeats drop to DEBUG — reusing the log-once posture this module already applies to provider misconfiguration. Otherwise ERROR volume would track traffic rather than the fault, and alerting on ERROR rate would page on QPS.

The dedup key comes from the exception (`"remote:413"`), deliberately **not** the message — that embeds a per-request response body and would defeat the dedup. A success clears it, so a fixed-then-regressed sidecar reports again. `record_failure()` still fires every occurrence, so the degraded trip-wire keeps counting.

### One thing worth knowing

The retry-skip is a **no-op at the shipped default** of `RANK_RETRY_ATTEMPTS=1` — one attempt happens either way. For a default deployment the deliverable is the ERROR level and the actionable detail. The tests raise the budget so the distinction is observable at all; there's a comment saying so, because "simplifying" it back to the default would turn every budget assertion into a tautology.

## Scope

Classification only. **Chunking** the candidate pool to the sidecar's cap — which would make the batch-cap fault structurally impossible — stays open in #704. The two are complementary rather than alternatives: `401`/`403`, a wrong endpoint, and the wrong-shape case are permanent regardless of chunking. Chunking arguably *depends* on this, since the sidecar's cap isn't discoverable from config and would otherwise have to be learned from a 413.

## Verification

23 tests pass. The meaningful check — reverting just the two providers, leaving the service layer in place:

```
5 failed, 18 passed
  FAILED test_413_raises_permanent_error_with_actionable_detail
  FAILED test_permanent_failure_is_not_retried_and_logs_error
  FAILED test_200_that_is_not_a_ranked_list_is_permanent
  FAILED test_permanent_error_logs_once_then_debug_until_next_success
  FAILED test_local_provider_missing_dependency_is_permanent
```

Exactly the 5 classification tests, while the transient ones — 503, 429 and 408 each still spending the full budget — keep passing. They assert *preserved* behaviour rather than restating the change.

`ruff check common/` + the gated scopes clean, `ruff format` clean, `mypy` clean (189 source files).

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -146,6 +146,11 @@ OPENAI_API_KEY=
 # Per-call scoring deadline (seconds); over this, degrade to first-stage order.
 # RANK_TIMEOUT_SECONDS=0.5
 # Max candidates re-scored per search; rows past the cap keep first-stage order.
+# With RANK_PROVIDER=remote the whole pool goes in ONE /rerank request, so the
+# sidecar must admit this many per call — a TEI reranker allows only 32 unless
+# started with --max-client-batch-size. If it doesn't, every search with more
+# candidates than the sidecar's cap is rejected and silently keeps first-stage
+# order (look for "Ranking failed permanently" in the core-api logs).
 # RANK_CANDIDATE_LIMIT=50
 
 # -- LLM enrichment ---------------------------------------------------------
--- a/common/ranking/__init__.py
+++ b/common/ranking/__init__.py
@@ -12,6 +12,7 @@ Public entrypoint: :func:`common.ranking.get_ranking`.
 from __future__ import annotations
 
 from common.ranking._service import get_ranking
+from common.ranking.errors import PermanentRankError
 from common.ranking.protocols import RankCandidate, RankProvider
 
-__all__ = ["RankCandidate", "RankProvider", "get_ranking"]
+__all__ = ["PermanentRankError", "RankCandidate", "RankProvider", "get_ranking"]
--- a/common/ranking/_service.py
+++ b/common/ranking/_service.py
@@ -24,6 +24,7 @@ from common.ranking.constants import (
     RANK_RETRY_DELAY_S,
     RANK_TIMEOUT_SECONDS,
 )
+from common.ranking.errors import PermanentRankError
 from common.ranking.protocols import RankCandidate
 
 logger = logging.getLogger(__name__)
@@ -68,6 +69,52 @@ _stats = _RankStats()
 # One-shot misconfiguration log dedup, keyed on provider name — a bad
 # ``RANK_PROVIDER`` would otherwise log once per search.
 _misconfiguration_logged: set[str] = set()
+
+# Same log-once posture for permanent provider faults, keyed on
+# ``PermanentRankError.key`` (a stable backend+condition string, NOT the
+# message — that embeds a response body and varies per request). A permanent
+# fault recurs on every search by definition, so logging it in full each time
+# would make ERROR volume a function of traffic rather than of the fault.
+# ``_stats.record_failure()`` still fires every time, so the degraded
+# trip-wire keeps counting occurrences.
+#
+# Entries are cleared per BACKEND on that backend's next success, never
+# wholesale: one process serves many tenants and holds a rank provider per
+# ``(base_url, api_key, model)``, so a healthy tenant's success must not
+# re-arm — or suppress — a different tenant's broken sidecar. Clearing
+# globally would put ERROR volume right back on a traffic curve, this time
+# driven by unrelated tenants.
+_permanent_logged: set[str] = set()
+
+# Bound it. Each entry is one (backend, condition) pair, so the steady state is
+# tiny — but a provider whose scope is per-instance means a long-lived process
+# that rotates tenant rank config accumulates a dead entry per retired backend,
+# the same leak ``_registry.py`` caps its ranker cache at 32 to avoid. On
+# overflow we drop the whole set rather than track recency: the only cost is
+# that a still-broken backend may report once more than strictly needed, which
+# is the safe direction to fail.
+_PERMANENT_LOGGED_MAX = 256
+
+
+def _permanent_scope(provider: object) -> str:
+    """Dedup namespace for ``provider`` — see ``PermanentRankError.key``.
+
+    Falls back to the provider name for providers that don't declare a scope
+    (``noop``/``fake`` never raise permanently, and a third-party provider
+    without one still gets coarse per-name dedup rather than none).
+    """
+    scope = getattr(provider, "dedup_scope", None)
+    if scope:
+        return str(scope)
+    return str(getattr(provider, "provider_name", "unknown"))
+
+
+def _clear_permanent_for(provider: object) -> None:
+    """Re-arm the ERROR for one backend after it succeeds again."""
+    prefix = f"{_permanent_scope(provider)}|"
+    _permanent_logged.difference_update(
+        {key for key in _permanent_logged if key.startswith(prefix)}
+    )
 
 
 def _resolve_provider_name(tenant_config: object | None) -> str:
@@ -131,6 +178,11 @@ async def get_ranking(
                     f"ranker returned {len(scores)} scores for {len(candidates)} candidates"
                 )
             await _stats.record_success()
+            # Re-arm THIS backend's permanent-fault ERROR: if it was fixed and
+            # later regresses, the next fault reports in full rather than
+            # staying silent at DEBUG forever. Scoped, so a healthy tenant
+            # can't re-arm a still-broken one.
+            _clear_permanent_for(provider)
             return scores
         except TimeoutError:
             logger.warning(
@@ -139,6 +191,28 @@ async def get_ranking(
                 RANK_RETRY_ATTEMPTS,
                 RANK_TIMEOUT_SECONDS,
             )
+        # A configuration-class fault (see common/ranking/errors.py): it fails
+        # identically next attempt, so stop instead of spending the turn's
+        # latency to reach the same answer. ERROR, not warning, because the
+        # search still succeeds on first-stage order — this log line is the
+        # only symptom the fault has. Deduped per condition; see
+        # ``_permanent_logged``.
+        except PermanentRankError as exc:
+            await _stats.record_failure()
+            if exc.key in _permanent_logged:
+                logger.debug("Ranking permanently failing (already reported): %s", exc)
+            else:
+                if len(_permanent_logged) >= _PERMANENT_LOGGED_MAX:
+                    _permanent_logged.clear()
+                _permanent_logged.add(exc.key)
+                logger.error(
+                    "Ranking failed permanently, not retrying; keeping "
+                    "first-stage order. This recurs until the configuration "
+                    "changes; further occurrences log at DEBUG until the next "
+                    "success. %s",
+                    exc,
+                )
+            return None
         # Intentionally broad: any provider error degrades to first-stage.
         except Exception:
             logger.warning(
--- a/common/ranking/errors.py
+++ b/common/ranking/errors.py
@@ -1,0 +1,55 @@
+"""Ranking failure classification.
+
+Every rerank failure degrades to first-stage order — that is the component's
+whole posture and it does not change here. What this adds is the distinction
+between the two *reasons* a rerank can fail, because they want opposite
+handling:
+
+* **Transient** (timeout, 429, 5xx, connection reset): the next attempt might
+  succeed. Retry it, log at ``warning``. This is the default for anything not
+  explicitly classified, so a new failure mode never silently stops retrying.
+* **Permanent** (:class:`PermanentRankError`): a configuration-class fault that
+  will fail *identically* on every subsequent call until a human changes
+  something — a batch cap below the candidate limit, a base URL pointing at a
+  service that doesn't speak ``/rerank``, a rejected API key. Retrying spends
+  the turn's latency budget to reach the same failure, so the service layer
+  stops immediately and logs at ``error`` with the provider's detail.
+
+The split matters because the two are indistinguishable from the outside: both
+serve first-stage order and neither breaks recall, so a misconfiguration looks
+exactly like load. Providers raise the specific type; the service layer stays
+transport-agnostic and only reacts to it.
+"""
+
+from __future__ import annotations
+
+
+class PermanentRankError(Exception):
+    """A rerank failure that will recur until the configuration changes.
+
+    Raise this instead of a generic exception when the provider can tell the
+    fault is not worth retrying. The message is surfaced verbatim in the
+    service layer's ``error`` log, so it should carry everything an operator
+    needs to act on without opening the provider's own logs.
+
+    ``key`` identifies the *condition* for log-deduplication. It is required —
+    there is no default, because every plausible default is wrong: the message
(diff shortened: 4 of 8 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
## Round 1 — High applied (it was a real bug I introduced), suggestion applied

### ✅ High — dedup was process-global, not per-backend

Verified the premise before changing anything: `common/ranking/_registry.py:41` caches `RemoteRanker` in an LRU keyed on **`(base_url, api_key, model)`**, and per-tenant `rank_base_url` overrides are supported. So several live sidecars per process is the normal case, and both halves of the finding hold:

1. `_permanent_logged.clear()` on **any** success — healthy tenants succeed constantly, so it fired near-continuously and re-armed the ERROR for a still-broken tenant on nearly every one of its own searches. That put ERROR volume right back on a traffic curve, driven by *unrelated* tenants: exactly what the module comment I wrote claims to prevent.
2. `key=f"remote:{resp.status_code}"` had no backend identity, so two tenants on two different broken sidecars both returning 413 collided — the first logged fault silently suppressed the second.

Fixed by scoping both sides. Providers now expose `dedup_scope` (`remote:{base_url}`, `local:{model}` — matching each one's registry cache key), keys are `{scope}|{condition}`, and recovery clears only that backend's keys via `_clear_permanent_for(provider)`. Providers without a scope fall back to the provider name, so a third-party provider gets coarse dedup rather than none.

Two tests, both of which fail against the previous commit:

| Test | Asserts |
|---|---|
| `test_dedup_is_scoped_per_backend_not_process_wide` | two broken sidecars → **two** ERRORs, one naming each |
| `test_one_backend_recovering_does_not_re_arm_another` | 3 successes on a healthy backend leave the broken one deduped |

```
against the previous (unscoped) commit:  2 failed, 14 passed
```

### ✅ Suggestion — `key` is now required

Agreed, and for a sharper reason than the footgun: *every* plausible default is wrong. The message is the only candidate and it embeds a per-request response body, so defaulting to it breaks the dedup **silently** rather than loudly. Now a keyword-only required arg — a missing key is a `TypeError` at the call site. Docstring rewritten to state both properties a key needs (stable, backend-scoped) and to point at `dedup_scope`.

### Incidental find while writing the tests

`_RankStats` is also module-global, and its "degraded: N consecutive failures" trip-wire fires an **extra** ERROR once enough failures accumulate across a session — which made bare ERROR-count assertions order-dependent. The autouse fixture now isolates both `_permanent_logged` and `_stats` per test, and the count assertions filter for the permanent-failure message specifically. Verified order-independent by running the dedup tests isolated, paired, and in the full file.

### Note on the push

`--force-with-lease` initially rejected: the branch had been rebased onto a newer `main` (now including #706). I replayed onto the new base (`187b5b5`) rather than force-overwriting, and re-ran the gates there.

**Gates:** 25 ranking tests + 43 including the search pipeline, `ruff check common/` and the gated scopes clean, format clean, `mypy` clean (189 files), DCO signed. No `uv.lock` drift.

Eldad-Caura:
@claude re-review — the High is applied (dedup now scoped per backend via a new `dedup_scope` provider property; `key` is required), plus two new tests that fail against the previous commit.

Eldad-Caura:
## Round 2 — all three applied

### ✅ Medium — `dedup_scope` now mirrors the registry cache key in full

Correct, and it was an inconsistency with my own stated invariant: `errors.py` says the scope must match what makes two backends distinct, then I used `base_url` alone while `_get_or_create_remote_ranker` keys on `(base_url, api_key, model)`. So the two collisions the reviewer names — same URL with a per-tenant `rank_model`, and same URL+model behind different `rank_api_key` — were both live.

Now `remote:{base_url}:{model}:{api_key_fp}`. Took the reviewer's optional suggestion on `api_key`, with a change: it's a **sha256 fingerprint, not the raw key**. The scope isn't logged today, but a credential has no business sitting in a string that exists to be logged about, and a future `logger.debug(exc.key)` shouldn't be able to leak it.

Two tests:
- `test_dedup_scope_separates_same_url_different_model` — same URL, different model → two ERRORs
- `test_dedup_scope_fingerprints_the_api_key` — the raw token never appears in the scope, and differing keys still produce differing scopes

### ✅ Low — a 200 with a non-JSON body is permanent

Real gap: `resp.json()` raising `JSONDecodeError` fell through to the generic transient path and got the full retry budget, even though a misrouted `RANK_BASE_URL` returning an HTML page yields the identical body every time. That's the same wrong-endpoint fault as the not-a-ranked-list check immediately below it, so it now raises `PermanentRankError` with key `…|invalid-json`. Caught as `ValueError` (which `json.JSONDecodeError` subclasses) so it also covers httpx's own decode paths.

`test_200_with_non_json_body_is_permanent` asserts one call, not three.

### ✅ Low — stale `local.py` module docstring

Fixed; it still promised `ImportError`.

### Verification

All three fail against the previous commit:

```
3 failed, 16 passed
  FAILED test_dedup_scope_separates_same_url_different_model
  FAILED test_dedup_scope_fingerprints_the_api_key
  FAILED test_200_with_non_json_body_is_permanent
```

28 ranking tests, 46 including the search pipeline. `ruff check common/` + gated scopes clean, format clean, `mypy` clean (189 files), DCO signed, no `uv.lock` drift.

Eldad-Caura:
@claude re-review — all three round-2 findings applied (dedup scope now covers base_url + model + api-key fingerprint; non-JSON 200 classified permanent; local.py docstring corrected), each with a test that fails against the previous commit.

Eldad-Caura:
## CodeQL fixed + both round-3 Lows applied — all checks green

### 🔴 CodeQL: `py/weak-sensitive-data-hashing` (high) — fixed at the root

```
common/ranking/providers/remote.py:76
Sensitive data (password) is used in a hashing algorithm (SHA256) that is
insecure for password hashing, since it is not a computationally expensive
hash function.
```

Fair call. My intent was a cache discriminator, not credential storage, but the rule is right that a truncated fast digest of a token is guessable if it ever leaks — and the previous round's suggestion to fold `api_key` into the scope is what introduced it.

Rather than suppress it or reach for a KDF (absurd for a per-instance discriminator), **the credential is gone from the scope entirely.** `dedup_scope` is now an opaque per-**instance** counter. That's the same discriminator by construction: the registry builds exactly one ranker per `(base_url, api_key, model)`, so *distinct instances are distinct backends*. No config value — least of all the secret — enters the string. The operator still gets the URL: it's in the message.

Which conveniently resolves both round-3 findings too:

### ✅ Low — prefix collision
Gone by construction. `remote:{int}` is a fixed shape, so no tenant-supplied `rank_model` containing `|` can make one scope a literal prefix of another. No escaping scheme needed.

### ✅ Low — unbounded `_permanent_logged` growth
Now capped at 256. A per-instance scope makes this *more* necessary, not less — retired backends would otherwise leave dead entries behind, the same leak `_registry.py` caps its ranker cache at 32 to avoid. On overflow the set is dropped wholesale rather than tracking recency: the only cost is a still-broken backend reporting once more than strictly needed, which is the safe direction.

### One test premise changed with it

`test_success_re_arms_the_permanent_error` previously used a *second* instance to recover. Under per-backend scoping that correctly does **not** re-arm the first — so it now drives a single ranker whose sidecar flips `413 → 200 → 413`. The cross-backend case is asserted separately by `test_one_backend_recovering_does_not_re_arm_another`, and `test_dedup_scope_carries_no_credential_but_still_separates_keys` pins both properties: no secret in the scope, and two credentials still separate.

### Green

| Check | |
|---|---|
| CodeQL | ✅ pass — **0 open alerts** on this PR |
| Analyze (python / js-ts) | ✅ pass |
| CI | ✅ pass |
| Security Scan (bandit + pip-audit) | ✅ pass |
| dco | ✅ pass |

28 ranking tests, 46 including the search pipeline. `ruff check common/` + gated scopes clean, format clean, `mypy` clean (189 files), `bandit -r common/ranking/` clean, no `uv.lock` drift.

Eldad-Caura:
@claude re-review — CodeQL's high-severity `py/weak-sensitive-data-hashing` is resolved by removing the credential from the dedup scope entirely (now an opaque per-instance counter), which also closes both round-3 Lows: fixed-shape scope can't prefix-collide, and `_permanent_logged` is bounded. All checks green, 0 open alerts.

Eldad-Caura:
## Final round applied

### ✅ Low — `dedup_scope` docstring asserted an invariant that isn't true

Correct, and worth fixing precisely because it was a **comment claiming a property the code doesn't have**: I wrote that the registry builds one ranker per config triple for the process lifetime, but `_remote_ranker_cache` is LRU-bounded at 32, so a process cycling through more backends can evict and rebuild one — and the rebuilt instance gets a fresh scope, making a still-broken backend re-report once.

Took option (a) as recommended. The docstring now states the tradeoff rather than a false invariant: the instance is a *proxy* for the config, not a permanent name for it; a re-report on eviction errs toward saying too much about a real fault, the same direction `_PERMANENT_LOGGED_MAX` overflow errs in. It also records why option (b) was declined — deriving the scope from config would hold across eviction, but only by reintroducing a value-derived key over tenant-supplied strings, which is exactly where the credential and delimiter problems came from.

### ⏭️ Suggestion — malformed per-candidate response

Out of scope: that path is unchanged by this diff (`item["index"]` / `float(score)` on a well-formed list). It's a real robustness gap but a different one, and it belongs with the chunking work in #704 rather than widening a classification PR further.

---

Ready for review. Final state: **CodeQL green with 0 open alerts**, CI / Analyze (python + js-ts) / Security Scan / DCO all pass, 28 ranking tests (46 including the search pipeline), `ruff` + `mypy` + `bandit` clean, no `uv.lock` drift.

Four rounds in, the findings went High → Medium → Low → Low, and three of them were real bugs I'd introduced rather than style notes — the process-global dedup, the incomplete scope, and the credential hash. Worth saying explicitly since the shape of the fix changed materially because of them.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #709, opened 2026-08-04
fix(rerank): chunk the remote candidate pool so the sidecar's cap can't reject it

### LINKED ISSUE #704: Reranking: remote provider doesn't chunk to the sidecar's batch cap, and a permanent 4xx is indistinguishable from a transient timeout

Eldad-Caura:
Closes #704. Completes the pair with #708: that made a batch-cap rejection **loud**, this makes it **not happen**.

## The problem

`RemoteRanker.rank()` posted the whole candidate pool as ONE `/rerank` request, up to `RANK_CANDIDATE_LIMIT` (50). A TEI reranker admits **32** per request unless started with `--max-client-batch-size`, so the **default pair was broken out of the box**: every search returning more than 32 candidates was rejected and silently served first-stage order.

An operator could fix it by raising the sidecar's cap — but nothing in the library enforced the coupling, and stock-vs-stock didn't work.

## The change

The pool is split into chunks of **`RANK_REMOTE_MAX_BATCH`** (new, default **32** = TEI's own default) and issued **concurrently**. The two settings are independent from here: `RANK_CANDIDATE_LIMIT` is a quality/latency choice, `RANK_REMOTE_MAX_BATCH` is purely what the sidecar accepts per call.

### Why splitting is invisible in the result

A cross-encoder scores each `(query, text)` pair **independently** — a candidate's score doesn't depend on what shares its batch — so chunk-then-merge returns exactly what one large call would. Not an approximation.

That property is load-bearing and does **not** hold for a listwise reranker that scores a set jointly. The docstring says so, because putting one behind this provider would silently change rankings.

### Two deliberate design points

**Concurrent, not sequential.** The service layer wraps the whole call in one `RANK_TIMEOUT_SECONDS` deadline (0.5s) while a warm 50-candidate rerank already measures ~285ms — sequential chunks would blow it, converting a silent batch-cap rejection into a silent *timeout*: same symptom, harder to diagnose. `gather(return_exceptions=True)` so every chunk is awaited before any is inspected; a bare `gather` propagates the first failure and leaves siblings running unobserved.

To be precise about what this buys: it removes **client-side serialisation**, not server-side compute. One replica still scores the same number of pairs, so budget against roughly the unchunked figure plus a round-trip, not N× it.

**Any chunk failing fails the whole rank.** Scoring the failed chunk's candidates `0.0` would bury real results at the bottom and call it success — worse than not reranking. A `PermanentRankError` is surfaced over a transient sibling: a cap set too high is permanent regardless of what blip another chunk hit, and misreporting it transient would spend the retry budget re-earning the same 413.

## Blast radius is smaller than it looks

The default `top_k` of 5 with 2× overfetch yields a **10-candidate pool** → single-request fast path → byte-for-byte identical on the wire to before. Splitting begins around `top_k >= 17`.

**Operational note:** a deployment already running its sidecar with `--max-client-batch-size 64` should set `RANK_REMOTE_MAX_BATCH=64` to keep single-request behaviour at every `top_k`.

Also corrected: #708's 413 hint advised lowering `RANK_CANDIDATE_LIMIT`, which is no longer the remedy since pool size no longer determines request size. It now names `RANK_REMOTE_MAX_BATCH` and its live value.

## Verification

Each mechanism was reverted individually to confirm the matching test actually guards it:

| Reverted | Result |
|---|---|
| chunking entirely | all new tests fail |
| chunked but **sequential** | the split/order test fails on its peak-in-flight assertion — **and only that one** |
| `raise failures[0]` instead of the permanent-wins scan | the permanent-over-transient test fails |

That last one matters: in an earlier draft the test was **vacuous** — `gather` preserves order and chunk 0 happened to hold the 413, so the naive implementation satisfied it. The handler now keys the status off chunk *size*, making the short tail chunk the permanent one. Removing that shared-counter handler also fixed a one-off flake: **45 runs of the file under deliberate CPU contention, zero failures.**

`ruff check common/` + gated scopes clean, format clean, `mypy` clean (190 files), `bandit` clean, no `uv.lock` drift.

## Not in scope

`RANK_MODEL` still defaults to `cross-encoder/ms-marco-MiniLM-L-6-v2`, which hf.co answers with a 307 to the canonical `…MiniLM-L6-v2` (established in caura-memclaw-enterprise#972). Harmless today, worth a one-line follow-up rather than widening this PR.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -146,12 +146,15 @@ OPENAI_API_KEY=
 # Per-call scoring deadline (seconds); over this, degrade to first-stage order.
 # RANK_TIMEOUT_SECONDS=0.5
 # Max candidates re-scored per search; rows past the cap keep first-stage order.
-# With RANK_PROVIDER=remote the whole pool goes in ONE /rerank request, so the
-# sidecar must admit this many per call — a TEI reranker allows only 32 unless
-# started with --max-client-batch-size. If it doesn't, every search with more
-# candidates than the sidecar's cap is rejected and silently keeps first-stage
-# order (look for "Ranking failed permanently" in the core-api logs).
 # RANK_CANDIDATE_LIMIT=50
+# With RANK_PROVIDER=remote the pool is split into concurrent /rerank requests
+# of at most this many texts, so this and RANK_CANDIDATE_LIMIT above are
+# independent — that one is a quality/latency choice, this is purely what the
+# sidecar will accept per call. The default matches TEI's own
+# --max-client-batch-size default. Set it higher than your sidecar actually
+# allows and requests come back rejected (loudly: "Ranking failed permanently"
+# in the core-api logs, naming this variable).
+# RANK_REMOTE_MAX_BATCH=32
 
 # -- LLM enrichment ---------------------------------------------------------
 USE_LLM_FOR_MEMORY_CREATION=true
--- a/common/ranking/constants.py
+++ b/common/ranking/constants.py
@@ -67,3 +67,17 @@ RANK_RETRY_DELAY_S: float = float(os.environ.get("RANK_RETRY_DELAY_S", "0.5"))
 # Rows beyond this cap keep their first-stage order, appended after the
 # re-ranked head.
 RANK_CANDIDATE_LIMIT: int = int(os.environ.get("RANK_CANDIDATE_LIMIT", "50"))
+
+# Largest batch the ``remote`` sidecar accepts in ONE /rerank request. The
+# provider splits the candidate pool into chunks of this size and issues them
+# concurrently, so RANK_CANDIDATE_LIMIT is no longer bounded by what the sidecar
+# admits. The coupling moves rather than vanishes, though: the pool size now
+# sets FAN-OUT, ``ceil(RANK_CANDIDATE_LIMIT / this)`` concurrent requests per
+# search (2 at the defaults).
+#
+# Default 32 = TEI's own ``--max-client-batch-size`` default, so stock-vs-stock
+# works. Raise it to match a sidecar started with a larger cap and a full pool
+# goes over in one request again. Cohere-shaped endpoints admit far more per
+# call, so 32 is very conservative there — worth raising to at least
+# RANK_CANDIDATE_LIMIT if that is the backend.
+RANK_REMOTE_MAX_BATCH: int = int(os.environ.get("RANK_REMOTE_MAX_BATCH", "32"))
--- a/common/ranking/providers/remote.py
+++ b/common/ranking/providers/remote.py
@@ -21,12 +21,13 @@ order — the caller (RerankResults) owns the sort.
 
 from __future__ import annotations
 
+import asyncio
 import itertools
 import logging
 
 import httpx
 
-from common.ranking.constants import RANK_TIMEOUT_SECONDS
+from common.ranking.constants import RANK_REMOTE_MAX_BATCH, RANK_TIMEOUT_SECONDS
 from common.ranking.errors import PermanentRankError
 from common.ranking.protocols import RankCandidate
 
@@ -73,10 +74,16 @@ class RemoteRanker:
         api_key: str = "",
         model: str = "",
         timeout: float = RANK_TIMEOUT_SECONDS,
+        max_batch: int = RANK_REMOTE_MAX_BATCH,
     ) -> None:
         self._base_url = base_url.rstrip("/")
         self._model = model
         self._instance_id = next(_instance_seq)
+        # Guard a non-positive override: negative yields no chunks at all (and
+        # so no scores), 0 blows up the range step below. Coerced rather than
+        # rejected because no real deployment reaches it — but note 1 means one
+        # HTTP request per candidate.
+        self._max_batch = max(1, max_batch)
         headers = {}
         if api_key:
             # Bearer for Cohere / token-gated TEI; harmless for open TEI.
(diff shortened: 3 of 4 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
## Round 1 — declining the semaphore, with reasoning

All checks green. One Low finding: unbounded fan-out via `asyncio.gather`.

**The finding is technically accurate and I'm not disputing the arithmetic** — `RANK_CANDIDATE_LIMIT=500` with `RANK_REMOTE_MAX_BATCH=1` does create 500 chunk coroutines from one search. But the named harm doesn't follow, for three reasons:

**1. The pool already is the bound.** The client is built with `httpx.Limits(max_connections=100)` (`remote.py:99`). Requests past 100 queue *inside* httpx rather than opening sockets — so "connection-pool exhaustion" is the mechanism that contains this, not the damage from it. A semaphore would be a second throttle in series with one that already works.

**2. The deadline bounds the damage window.** `_service.py` wraps the whole `rank()` in one `RANK_TIMEOUT_SECONDS` (0.5s). A pathological fan-out queues, misses the deadline, and degrades to first-stage order — the designed behaviour for an overloaded sidecar. It fails safe rather than piling up.

**3. A semaphore can't be sized correctly here.** The right cap is a function of the sidecar's own concurrency (`--concurrency`, replica count), which the client has no way to know. Any constant we pick is either so high it never binds or so low it silently serialises chunks — and serialising is precisely the failure this PR exists to avoid, since it converts a batch-cap rejection into a silent timeout.

A `/simplify` altitude pass over this diff reached the same conclusion independently ("do not add a semaphore: httpx's `max_connections=100` already bounds it and the shipped default is 2 requests").

### What I did take from it

The finding points at something real: fan-out is *documented* but not *visible*. `constants.py` now names it explicitly — `RANK_CANDIDATE_LIMIT` sets `ceil(limit / RANK_REMOTE_MAX_BATCH)` concurrent requests per search, 2 at the defaults — so the coupling is at least discoverable at the place it's configured.

If you'd rather have a runtime signal than prose, the cheap version consistent with this PR's theme is a **one-time WARNING when a rank splits into an unusually high number of chunks**, naming both knobs — observability rather than throttling, ~6 lines. Happy to add it; I've left it out rather than expand a clean PR on a Low where two analyses disagree.

### Green

| Check | |
|---|---|
| CI | ✅ pass |
| CodeQL | ✅ pass |
| Analyze (python / js-ts) | ✅ pass |
| Security Scan (bandit + pip-audit) | ✅ pass |
| claude-review | ✅ pass |
| dco | ✅ pass |

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #710, opened 2026-08-04
feat(write): honour per-item write_mode on the bulk path

Eldad-Caura:
Cloud half of adding a read-your-own-write opt-out to the **bulk** write surface. Prerequisite for the broker change that follows — the broker writes via `/api/v1/memories/bulk`, and that endpoint had no `write_mode` concept at all, so broker-side passthrough alone would have shipped a field the cloud silently drops.

## What was missing

`write_mode` existed only on `MemoryCreate`. Neither `BulkMemoryItem` nor `BulkMemoryCreate` carried it, and the bulk path gated inline embedding purely on `settings.inline_embedding` — `_resolve_write_mode` never ran for bulk. On a deferring deployment, *every* bulk-written memory waited for the background backfill before becoming semantically searchable, with no opt-out.

`write_mode="strong"` on an item now embeds that item inline.

**Per item, not per batch**, deliberately: callers coalesce writes from unrelated sources into one request — the broker's durable queue does exactly this — so a batch-level flag would let one caller's opt-in put a provider call on the request path for everyone else's items, or be lost among them. `tests/test_bulk_write_mode.py` pins the mixed-batch case at `[False, True, False]`.

## Scope boundaries — stated in the field description, not left to be discovered

- **Embedding only.** Single-write `strong` also enriches inline; here enrichment defers either way. Honouring the enrich half would put N semaphored LLM calls on a fan-in request — a much larger commitment than one batched embed.
- **Not routed through `_resolve_write_mode`.** That also consults `tenant_config.default_write_mode` and escalates `_STRONG_TYPES`, which would add a request-path provider call for callers who never asked. Only an explicit per-item opt-in counts. *(Worth a reviewer's opinion: closing that asymmetry is defensible, but it's a latency change for strong-defaulting tenants that nobody requested, so I left it and documented it.)*
- **`str`, not `Literal`.** This model deliberately avoids schema constraints that reject a whole batch over one item's field, so an unrecognised value (`"stm"` copied from a single-write payload) degrades to fast instead of 422-ing every sibling. A `Literal` here would have fought that documented policy.

## Two failure modes handled rather than inherited

**A failed opt-in embed does not fail the request.** Those rows were going to defer anyway, so falling through to the background re-embed costs only the latency they'd have had — where raising would let one opt-in item take down a batch of items that never asked for inline embedding. Inline deployments keep the pre-existing contract: the embed is how every row there gets its vector, so a timeout is still 504 and other errors still propagate.

**The opt-in embed gets its own tighter deadline** — `BULK_STRONG_EMBED_TIMEOUT_SECONDS = 8.0` instead of borrowing the required path's 30s. A required embed can justify 30s because the alternative is a 504; an opportunistic one would spend it and then land on the outcome it already accepted for free, with the latency charged to the whole batch. The 8s is bounded on both sides and both bounds matter: it must stay **above** `EMBEDDING_GATE_TIMEOUT_SECONDS` (5.0), which sits deliberately below callers' deadlines so a saturated gate reads as attributable backpressure rather than an anonymous timeout. A unit test pins both bounds.

No double work: an inline-embedded item is excluded from `reembed_batch` by the existing `embeddings[orig_idx] is None` gate and reaches contradiction detection with its vector in hand.

## One thing I built, then deliberately removed

I had this also setting `metadata.embedding_pending` on deferred bulk rows, for parity with the single path. I dropped it after checking what clears it: **nothing does, on this path.** core-worker clears the flag via `metadata_patch` (its docstring: that exists "so a read-after-success returns clean state instead of confusingly claiming the row is still pending"), but bulk backfills in-process through `_reembed_memories_bulk` → core-api's `update_embedding`, which sends only `{embedding, tenant_id}`.

Every deferred bulk row would have claimed to be pending forever, including after it was embedded and searchable — worse than no signal at all, given #706 documented that flag as the read-your-own-write indicator and `plugin/tools.json` tells agents not to re-write on it. `BulkItemResult` returns no metadata either, so it bought nothing at write time. Making it honest means teaching `update_embedding` to patch metadata, which is its own change.

Also corrects two comments this change falsified: `insights_service`'s note that "`BulkMemoryItem` carries no `write_mode` field", and the `MemoryOut.metadata` note I added in #706, which described single-write `strong` in terms that don't hold for bulk.

## Verification

|---|---|---|---|
| `main` | 4088 | 1 | 0 |
| this branch | 4096 | 1 | 0 |

+8 is exactly the added tests, with **no change to any existing test's behaviour** — worth stating, since this touches a write path much of the suite exercises.

Every new assertion confirmed load-bearing by mutation, inverting rather than deleting:

| mutation | result |
|---|---|
| drop the strong subset | opt-in + mixed-batch cases fail |
| make the opportunistic embed fatal | fallback case fails |
| remove the timeout→504 mapping | inline-deployment case fails |
| raise the opt-in budget to the required cap | bounds test fails (`assert 30.0 < 30.0`) |

`ruff check` and `ruff format --check` clean, run separately and from inside `core-api`; `mypy` clean on 190 files.

## Known follow-ups, not in scope here

- A `strong` item that turns out to be a **content duplicate** still pays a discarded embed — the embed runs before the hash-dedup lookup. Pre-existing ordering for inline deployments, newly reachable on deferring ones, and the fix is a reorder of an existing hot path.
- The MCP `memclaw_write` batch branch doesn't forward a top-level `write_mode` into items. The param is accurately documented "(single only)", so nothing is misleading today, but per-item mode isn't reachable from that surface.

### DIFF
--- a/core-api/openapi.broker.json
+++ b/core-api/openapi.broker.json
@@ -293,6 +293,18 @@
               }
             ],
             "title": "Weight"
+          },
+          "write_mode": {
+            "anyOf": [
+              {
+                "type": "string"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "description": "Per-item write mode. 'strong' embeds this item inline, so it is searchable as soon as the write is persisted rather than after the background backfill — at the cost of an embedding provider call on the request path. Any other value behaves as 'fast': the item follows the deployment's embedding mode. On a deployment that already embeds inline this has no effect, since every item is embedded inline anyway. Narrower than MemoryCreate.write_mode: on the bulk path 'strong' affects only the embedding — LLM enrichment defers either way — and the tenant's default_write_mode is not consulted, so only an explicit per-item opt-in embeds inline.",
+            "title": "Write Mode"
           }
         },
         "required": [
--- a/core-api/src/core_api/config.py
+++ b/core-api/src/core_api/config.py
@@ -357,9 +357,11 @@ class Settings(BaseSettings):
     def _validate_timeout_ordering(self) -> "Settings":
         # Local import avoids a circular: constants → common.constants,
         # but this file is imported by constants.py's dependents.
+        from common.embedding.constants import EMBEDDING_GATE_TIMEOUT_SECONDS
         from core_api.constants import (
             BULK_EMBEDDING_TIMEOUT_SECONDS,
             BULK_ENRICHMENT_TOTAL_TIMEOUT_SECONDS,
+            BULK_STRONG_EMBED_TIMEOUT_SECONDS,
         )
 
         if self.request_timeout_seconds < BULK_ENRICHMENT_TOTAL_TIMEOUT_SECONDS:
@@ -420,6 +422,39 @@ class Settings(BaseSettings):
                 f"bulk_request_timeout_seconds "
                 f"({self.bulk_request_timeout_seconds}s) so the storage-phase "
                 f"cap fires before the umbrella."
+            )
+        if EMBEDDING_GATE_TIMEOUT_SECONDS >= BULK_STRONG_EMBED_TIMEOUT_SECONDS:
+            # The opportunistic ``write_mode="strong"`` embed must be able to
+            # outwait the concurrency gate. The gate timeout sits deliberately
+            # below callers' deadlines so a saturated gate surfaces as
+            # attributable backpressure rather than an anonymous caller timeout;
+            # invert that and every gate-saturated strong write is reported as a
+            # generic embed failure instead. Checked here because the gate value
+            # is read from an env var at import, so the ordering is an operator's
+            # to break, not a constant's.
+            raise ValueError(
+                f"EMBEDDING_GATE_TIMEOUT_SECONDS ({EMBEDDING_GATE_TIMEOUT_SECONDS}s) must be < "
+                f"BULK_STRONG_EMBED_TIMEOUT_SECONDS ({BULK_STRONG_EMBED_TIMEOUT_SECONDS}s) so a "
+                "saturated embedding gate stays attributable on the opportunistic "
+                "bulk strong-embed path. Lower the gate timeout, or raise "
+                "BULK_STRONG_EMBED_TIMEOUT_SECONDS (keeping it under "
+                "BULK_EMBEDDING_TIMEOUT_SECONDS)."
+            )
+        if BULK_STRONG_EMBED_TIMEOUT_SECONDS >= BULK_EMBEDDING_TIMEOUT_SECONDS:
+            # The other half of the bound the message above already tells the
+            # operator to respect. The derived default clamps here, so only an
+            # explicit env override can reach this — and it must not, twice over:
+            # an opportunistic embed allowed to run as long as a required one stops
+            # being opportunistic and can hold a whole batch for that budget, for
+            # one item's opt-in; and the additive ordering check above assumes the
+            # embed phase never exceeds BULK_EMBEDDING_TIMEOUT_SECONDS, so letting
+            # it through would silently invalidate that proof rather than just
+            # degrade this path.
+            raise ValueError(
+                f"BULK_STRONG_EMBED_TIMEOUT_SECONDS ({BULK_STRONG_EMBED_TIMEOUT_SECONDS}s) must be < "
+                f"BULK_EMBEDDING_TIMEOUT_SECONDS ({BULK_EMBEDDING_TIMEOUT_SECONDS}s): the "
+                "opportunistic bulk strong-embed budget cannot exceed the required embed cap, "
+                "which the storage-phase ordering check above is proved against."
             )
         return self
 
--- a/core-api/src/core_api/constants.py
+++ b/core-api/src/core_api/constants.py
@@ -22,6 +22,10 @@ from common.constants import (  # noqa: F401
     VECTOR_DIM,
 )
 
+# The embedding concurrency-gate timeout, which the bulk strong-embed budget
+# below must stay ordered above — see BULK_STRONG_EMBED_TIMEOUT_SECONDS.
+from common.embedding.constants import EMBEDDING_GATE_TIMEOUT_SECONDS
+
 # Re-export memory-vocabulary constants from common.enrichment (CAURA-595).
 # common is the source of truth so core-api and core-worker stay in sync.
 from common.enrichment.constants import (  # noqa: F401
@@ -34,6 +38,7 @@ from common.enrichment.constants import (  # noqa: F401
     SERVER_RESERVED_MEMORY_TYPES,
     MemoryType,
 )
+from common.env_utils import read_float_env
 
 # Re-export LLM provider constants from common.llm (CAURA-595).
 from common.llm.constants import (  # noqa: F401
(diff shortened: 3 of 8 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@claude

Eldad-Caura:
CI fix pushed (`4ba4c62`), rebased onto `main` after #708/#709.

The failure was the **frozen v1 broker OpenAPI baseline**, not the code: `BulkMemoryItem` is reachable from `POST /api/v1/memories/bulk`, one of the four paths in that contract, so adding a field made the committed `core-api/openapi.broker.json` stale. Regenerated.

I checked it clears the *second* gate too rather than assuming, since "freeze v1" deserves more than a regenerate — the diff is **12 insertions, 0 deletions**, nothing added to `required`, and `oasdiff breaking --fail-on ERR` against main's baseline run locally with the same pinned check reports **no breaking changes**. `CONTRACT_VERSION` deliberately not bumped: per the generator's own docstring that's for intentionally shipping a new broker major, not an optional additive field.

Re-verified on the new base (#709 added tests, so the numbers moved):

|---|---|---|---|
| `main` (c987649) | 4105 | 1 | 0 |
| this branch (4ba4c62) | 4113 | 1 | 0 |

+8 is exactly the added tests. I measured that baseline from a detached checkout of `origin/main` — my first attempt used `git stash`, which was a no-op now the change is committed, and silently reported the branch's own number as the baseline.

Eldad-Caura:
@claude

Eldad-Caura:
Good finding, and it lands on the exact defect class I'd been watching for elsewhere in this change: a comment asserting a property that's true of the *default* and not of the *configured* set. Fixed in `e511109`.

You're right that `EMBEDDING_GATE_TIMEOUT_SECONDS` is `read_float_env` — I read that function while picking the 8.0 and still wrote the bound as though both sides were literals this repo controls. They aren't: the gate value is an operator's, so the ordering was mine to enforce, not to assert.

Taking your suggested approach rather than an alternative I'd considered (deriving the budget from the gate at import). The validator is better here: deriving would silently change a value the operator didn't ask for, where `_validate_timeout_ordering` is the house mechanism for exactly this and fails loudly at startup, consistent with the other cross-constant timeout orderings in that method.

```python
if EMBEDDING_GATE_TIMEOUT_SECONDS >= BULK_STRONG_EMBED_TIMEOUT_SECONDS:
    raise ValueError(...)   # names both values, and says which to move
```

The error message names both values and both remedies (lower the gate, or raise the budget while keeping it under `BULK_EMBEDDING_TIMEOUT_SECONDS`), so the operator isn't left to rediscover the reasoning.

The constant's comment now points at the validator instead of claiming an invariant nothing checked, and records *why* the ordering matters — invert it and every gate-saturated strong write is reported as a generic embed failure, losing the backpressure attribution the gate's deliberate sub-caller-deadline placement exists to provide.

Pinned by a test, and confirmed load-bearing rather than assumed: making the guard unreachable (`>= … + 1e9`, so both names stay referenced and the failure isn't a build error) fails with `DID NOT RAISE ValueError`.

Suite 4114 passed / 1 xfailed / 0 xpassed against main at 4105 / 1 / 0. Broker OpenAPI baseline still current and still non-breaking; `ruff check` + `format --check` clean; `mypy` clean on 190 files.

Eldad-Caura:
@claude

Eldad-Caura:
Right, and this one matters more than the finding it replaced — my round-1 fix was itself the bigger hazard. A check that bricks startup for deployments that tuned the gate is worse than the mislabelled-failure problem it was guarding against, and "the only fix is a code change + redeploy" is exactly the wrong property for a feature nobody opted into. Fixed in `924e499`.

I applied **both** of your options rather than picking one, because they solve different halves:

- **Derived default** (`_default_strong_embed_timeout`: `min(required, max(8.0, gate + 3.0))`) — the ordering now holds *however* the operator tuned the gate, so no install can be bricked by the upgrade. This is the part that removes the hazard.
- **`read_float_env` override** — a deployment that needs a specific budget gets a config-only lever instead of a code change. This is the part that keeps it operable.

The validator stays, but its job shrinks to catching an *explicitly conflicting override*, which is still worth failing loudly on.

One case worth calling out: a gate raised past `BULK_EMBEDDING_TIMEOUT_SECONDS` clamps rather than grows. That config is incoherent on its own terms — the gate would outlive the very embed it gates — so the validator reports it rather than the constants quietly papering over it.

Both new assertions confirmed load-bearing by mutation:

| mutation | result |
|---|---|
| neuter the gate term in the derivation | `assert 8.0 > 10.0` — "gate=10.0 must leave room to outwait the gate" |
| make the validator guard unreachable | `DID NOT RAISE ValueError` |

Worth admitting: my first attempt at that first mutation silently didn't apply — `ruff format` had wrapped the line I was string-replacing, and I hadn't asserted the replacement matched, so I briefly read "10 passed" as "the test is fine" when it meant "the mutation never happened". Redone with an assertion on the match.

Suite 4115 passed / 1 xfailed / 0 xpassed. Broker OpenAPI baseline current and still non-breaking; `ruff check` + `format --check` clean; `mypy` clean on 190 files.

Eldad-Caura:
@claude

Eldad-Caura:
Comment fix pushed. Both were fair.

**The misplaced comment** was collateral from `ruff check --fix` reordering my new imports: the isort autofix moved `common.embedding.constants` above the block the comment described, stranding it. Restored onto `common.constants`, and the new import now carries its own label saying what it's for. My fault for not re-reading the diff after letting the autofix touch the import block.

**Insights not using `write_mode`** — deliberate, not an oversight, but you're right that it's worth stating rather than leaving implied.

Two reasons to leave it:

1. **It wouldn't restore what was lost.** The comment there records that the only behavioural delta between the strong and fast pipelines was the inline `CheckSemanticDuplicate` step — and bulk skips semantic dedup entirely, regardless of `write_mode`. So setting `strong` on those items would restore the inline *embed*, not the dedup, i.e. a partial fix that makes the comment's own account of the regression less true rather than more.
2. **It's a latency change to an internal flow that nobody asked me to reprice.** `_persist_findings` writes a batch per evolve cycle; opting it in puts an embedding provider call on that request path. That's a product call about the insights flow, not a consequence of adding the mechanism.

The mechanism now exists, so it's a one-kwarg change whenever that call is made. I've flagged it to Eldad rather than deciding it inside a PR about the bulk path — same reasoning as leaving `tenant_config.default_write_mode` unconsulted here.

Suite 4115 passed / 1 xfailed / 0 xpassed; comment-only change, so unchanged from the last push. ruff, mypy, and the broker OpenAPI gates all clean.

Eldad-Caura:
@claude

Eldad-Caura:
@claude

Eldad-Caura:
Round 3 fixed in `cd4f668` — you were right that the upper bound was documented and not enforced.

The validator checked only that the opt-in budget outlives the concurrency gate, while its own error message told the operator to keep it *under* `BULK_EMBEDDING_TIMEOUT_SECONDS`. The derived default clamps there, so only an explicit env override could reach it — but it must not, and the second reason you gave is the one that made me treat this as more than a degradation:

1. An opportunistic embed allowed to run as long as a required one **stops being opportunistic** and can hold a whole batch for that budget on one item's opt-in.
2. The pre-existing storage-phase additive ordering check is *proved against* the embed phase never exceeding `BULK_EMBEDDING_TIMEOUT_SECONDS`. Letting an override through wouldn't just weaken this path — it would silently invalidate that proof.

Both bounds are now enforced in `_validate_timeout_ordering`, and the error message names the constraint the check is protecting. Confirmed load-bearing: making the new guard unreachable fails with `DID NOT RAISE ValueError`.

Suite 4116 passed / 1 xfailed / 0 xpassed. ruff, mypy, and both broker OpenAPI gates clean.

**Follow-up now stacked:** #711 adds `write_mode="strong"` to the insights persist call — the product question you raised on the last round. Eldad made that call, so it's implemented, based on this branch since the field is inert without it.

Unrelated, found while running these suites and worth knowing since it isn't a code issue: `tests/test_api_keystones.py::test_set_then_list_round_trip` writes `_keystones` docs under tenant `default`, which the conftest cleanup never sweeps (it only deletes `test-tenant-%`). After ~15 local full-suite runs it had accumulated 197 rows and the test's own doc fell off the listing it asserts against. CI never sees it because each run gets a fresh database. Clearing those rows restored 17/17. Same family as the isolation problems #705 fixed; noting it rather than widening this PR.

Eldad-Caura:
@claude

Eldad-Caura:
@claude

Eldad-Caura:
Real bug, High is the right severity, and it was mine — round 2's fix broke at the far end of the range that round 3's check then policed. Fixed in `3ea3ca0`.

Reproduced before touching it:

```
required cap = 30.0
  gate= 26.9 -> budget=  29.9   strict gate<budget<required: True
  gate= 27.0 -> budget=  30.0   strict: False   ← unstartable
  gate= 29.9 -> budget=  30.0   strict: False   ← unstartable
```

`min(required, ...)` clamps *onto* the bound, and the validator rejects `>=`. So every gate in `[27, 30)` produced a config that could not start — for an operator who had raised only `EMBEDDING_GATE_TIMEOUT_SECONDS`, to a value still legitimately below the cap. Exactly the failure the derivation exists to prevent, reintroduced one round later at the opposite end.

The derivation now guarantees the bound the validator actually enforces — `gate < budget < required`: the full margin when it fits, otherwise the midpoint of the remaining room, which is strictly inside for any `gate < required`. A gate at or above the cap still returns the cap, because no valid budget exists there and the validator should say so.

```
  gate= 27.0 -> 28.5   strict: True
  gate= 29.9 -> 29.95  strict: True
  gate= 30.0 -> 30.0   strict: False  (correctly rejected — incoherent config)
```

End-to-end: a gate of 28.0 now constructs `Settings()` without raising.

**Your point about the test is the part worth dwelling on.** It asserted `got <= required` and swept only to `gate=25.0` — so it passed for two review rounds while the boundary band was broken. `<=` was the bug: it encoded the clamp I had written instead of the invariant the validator enforces. It now asserts the strict bound and sweeps `26.9 / 27.0 / 28.0 / 29.0 / 29.99`, and is confirmed load-bearing — restoring the clamp-to-cap fails with `assert 30.0 < 30.0`.

Also folded #711 in, since it merged into this branch and the convention here is one commit per open PR. Verified the squash is content-identical to the two-commit state plus this fix, and nothing of #711's was lost.

Suite 4117 passed / 1 xfailed / 0 xpassed. `ruff check` + `format --check` clean per project, `mypy` clean, broker OpenAPI baseline current and non-breaking.

Eldad-Caura:
@claude

Eldad-Caura:
@erni-a pleae approve

erni-a (approved):

PR #711, opened 2026-08-04
feat(insights): persist findings with write_mode="strong"

Eldad-Caura:
Your call from #710's review, implemented.

> **Base is `feat/bulk-write-mode-inline-embed` (#710), not `main`** — the field does nothing without that PR's mechanism. GitHub will retarget this to `main` once #710 merges. The diff here is one commit.

## Why

An insight exists to be recalled, and the obvious next action after an evolve cycle is to search for what it produced. Persisted at the deployment's default write mode, a finding lands in the deferred-embed window — so a semantic search straight after the cycle returns nothing, which reads as *"no insights were written"*.

This restores the inline-embed half of what the pre-#29 serial path had. That path passed `write_mode="strong"` on every `MemoryCreate`; the migration to `create_memories_bulk` dropped it because `BulkMemoryItem` had no such field. It has one now, so this is the field that comment block was written to apologise for finally getting used.

## What it does *not* restore, stated rather than implied

`BulkMemoryItem.write_mode` governs the embedding and nothing else. The strong pipeline's other behavioural delta is the inline `CheckSemanticDuplicate` step, which has **no bulk equivalent at any `write_mode`** — so that stays dropped.

That's fine for insights, for the reason already documented directly below in that comment: the supersede-priors block transitions any prior insight for this focus/scope/agent to `outdated` before this persist runs, which covers the cross-run "same insight regenerated" case. The reason to spell it out is that *"strong" here is not the same promise as "strong" on the single-write path*, and a future reader who assumes parity would be wrong.

## Cost, and why opting in is safe

One batched embedding call per evolve cycle moves onto the request path, capped by `BULK_STRONG_EMBED_TIMEOUT_SECONDS` (8s default, derived from the embedding concurrency gate) and **falling back to the background backfill if the provider fails**. So the worst case is the latency this path already had — not a failed persist. That fallback, from #710, is what makes this an opt-in rather than a reliability trade.

## Verification

The test captures the items at the `create_memories_bulk` boundary and asserts every one carries the mode. Confirmed load-bearing rather than assumed — flipping the value to `"fast"`:

```
assert ['fast', 'fast'] == ['strong', 'strong']
```

|---|---|---|---|
| #710 head (base) | 4116 | 1 | 0 |
| this branch | 4117 | 1 | 0 |

+1 is exactly the added test. `ruff check` and `ruff format --check` clean from inside `core-api`; `mypy` clean on 190 files; broker OpenAPI baseline unaffected.

### DIFF
--- a/core-api/src/core_api/services/insights_service.py
+++ b/core-api/src/core_api/services/insights_service.py
@@ -911,17 +911,29 @@ async def _persist_findings(
     # write_mode behaviour change vs the pre-#29 path
     # -----------------------------------------------
     # The previous serial path passed ``write_mode="strong"`` on every
-    # ``MemoryCreate``. ``BulkMemoryItem`` now carries a ``write_mode`` too, but
-    # it governs only whether the embedding runs inline, and these items don't
-    # set it — so the bulk path still drops the strong mode override in the sense
-    # that matters below. The only behavioural delta between the strong and
-    # fast pipelines is the inline ``CheckSemanticDuplicate`` step (see
-    # ``core_api/pipeline/compositions/write.py``); everything else
-    # (embed, enrich, exact-dedup, write, schedule background tasks) is
-    # identical. The post-write fire-and-forget tasks — entity extraction,
-    # async contradiction detection, deferred enrichment — are still
-    # scheduled per memory by the bulk path's ``ScheduleBackgroundTasks``-
-    # equivalent loop, so contradiction detection coverage is intact.
+    # ``MemoryCreate``. The items below set it again, so the inline-embed half of
+    # that is restored: a finding is semantically searchable the moment this
+    # persist returns rather than after the deferred backfill.
+    #
+    # It is only that half. ``BulkMemoryItem.write_mode`` governs the embedding
+    # and nothing else — the inline ``CheckSemanticDuplicate`` step (see
+    # ``core_api/pipeline/compositions/write.py``) has no bulk equivalent at any
+    # write_mode, so the strong pipeline's one other behavioural delta stays
+    # dropped here regardless. That is fine for insights, for the reason spelled
+    # out below; the point is that "strong" on this path is not the same promise
+    # as "strong" on the single-write path, and reading it as such would be wrong.
+    #
+    # Everything else (enrich, exact-dedup, write, schedule background tasks) was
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
This finding is a false positive, and the cause is mechanical: the review read this PR against **`main`**, not against its base. #711 is stacked on `feat/bulk-write-mode-inline-embed` (#710), which is where the mechanism lives. Every statement in the finding is true of `main` and false of this branch.

Verified on this branch's actual base rather than argued:

```
$ python -c "from core_api.schemas import BulkMemoryItem; ..."
  field present: True
  constructed value: 'strong' (not dropped)
  BULK_STRONG_EMBED_TIMEOUT_SECONDS = 8.0

$ grep -n 'items\[i\].write_mode' core-api/src/core_api/services/memory_service.py
1209:        embed_indices = [i for i in valid_indices if items[i].write_mode == "strong"]
```

And the same checks against `origin/main`, which reproduce exactly what the finding describes: `BULK_STRONG_EMBED_TIMEOUT_SECONDS` — **0 occurrences**. So the reasoning was sound given what it read; it just read the wrong tree.

The decisive point is the prediction in the finding itself:

> the added test … that asserts `i.write_mode == "strong"` on captured `BulkMemoryItem`s would raise `AttributeError` if actually executed, since the field is never set

It does execute, and it passes — `1 passed, 63 deselected`. It also fails correctly when the behaviour is inverted (`write_mode="fast"` → `assert ['fast', 'fast'] == ['strong', 'strong']`), so it is not passing vacuously. Full suite on this branch: **4117 passed / 1 xfailed / 0 xpassed**, against the base branch at 4116 / 1 / 0. A no-op change could not produce that delta, and an `AttributeError` could not produce a green suite.

No code change from this round. The `extra="ignore"` mechanism the finding describes is real and would have made a silent no-op — it's the right thing to have checked — it just isn't reachable here, because the field exists on the base.

**One process note worth more than the finding:** the review doesn't honour a non-`main` base, so any stacked PR will read as "the mechanism doesn't exist". The PR body states the base in the first line, but that isn't enough to change what gets diffed. Worth knowing before the next stacked pair — the alternative is to hold the dependent PR until its base merges, at the cost of not being able to review the two together.

PR #712, opened 2026-08-04
feat(bench): LoCoMo rerank A/B harness

Eldad-Caura:
`BENCHMARKS.md` names LoCoMo/LongMemEval as the instrument we publish against, but ships **no runner** — so every rerank question so far has been answered by throwaway scripts. This commits the one for reranking, and updates that section to say what is and isn't bundled.

## What it measures

One question: **does the cross-encoder order results better than first-stage similarity alone?** It scores the *same* candidate pool two ways —

| | |
|---|---|
| baseline | embedding cosine, top `--pool` (what first-stage ranks by) |
| treatment | that same pool, reordered by `POST {rerank-url}/rerank` |
| truth | each QA item's `evidence` turn ids, binary relevance |

Because both arms score one pool, **recall@pool is identical by construction** — asserted at runtime — so any difference is ordering, which is all a reranker does. Reports nDCG@5/@10, MRR, P@5 with paired-bootstrap CIs and an exact sign test, overall and per LoCoMo category.

Stdlib only, no cloud dependencies — runs against `docker compose` as readily as a deployed stack.

## Design decisions worth knowing

- **Per conversation, never pooled.** Each LoCoMo conversation is one agent's memory and a question only concerns its own. Pooling all ten would measure a task the product never performs.
- **Category 5 excluded** (adversarial/unanswerable) — no evidence turn exists to retrieve, so those items measure answer *abstention*, not ranking.
- **Queries where no evidence reaches the pool are skipped and counted, not scored as zero.** First-stage failed outright there; a reranker cannot invent the row. Scoring them would add a shared zero to both arms and dilute the comparison. The skip count prints so the conditional-vs-unconditional distinction stays visible — on the run below it's 400/1531, i.e. the headline +0.29 is conditional on retrieval succeeding; unconditional it's ≈ +0.215.
- **Speakers kept in the turn text** — who said something is part of what makes a turn the right answer to *"when did X do Y"*.

## Why commit a harness at all

A first attempt at this A/B used a ~55-document fixture and returned a **null result**: nDCG@5 +0.008, CI spanning zero, 8 queries better / 9 worse. That was a **ceiling effect** — cosine alone already scored 0.85–0.97 there, so there was nothing left to win — and it would have been easy to misread as *"reranking doesn't help"* and drop the feature.

On LoCoMo, where 600-turn conversations put first-stage at 0.50, the same reranker against the same sidecar is worth **+0.29 nDCG@5**. Corpus size decided the answer, not the reranker. That's the argument for committing a harness with a real corpus rather than leaving this to fixtures.

## Verified

Against a live TEI `bge-m3` embedder and MiniLM reranker — 10 conversations, 5,882 turns, 1,531 questions, **n=1,131 scorable**:

| metric | first-stage | reranked | Δ | 95% CI | better/worse |
|---|---|---|---|---|---|
| nDCG@5 | 0.5019 | 0.7938 | **+0.2920** | [+0.269, +0.315] | 648 / 96 |
| nDCG@10 | 0.5598 | 0.8159 | +0.2561 | [+0.236, +0.276] | 675 / 112 |
| MRR | 0.4927 | 0.7807 | +0.2880 | [+0.266, +0.311] | 647 / 117 |
| P@5 | 0.1496 | 0.2055 | +0.0559 | [+0.049, +0.063] | 325 / 39 |

Multi-hop gains most (**+0.3812**) from the worst baseline (0.328) — the pattern you'd hope for. Open-domain is the weakest evidence (n=41; its nDCG@5 sign test doesn't reject at p=0.12 even though the CI excludes zero).

That table was produced by running **this exact file**, unmodified, fetched from the branch — not by the prototype it replaces (which reported +0.2915 on the same setup).

Also unit-checked offline: metric helpers against hand-computed values (including nDCG's ideal-capped-at-k case and order sensitivity), and `report()` confirmed to flag a synthetic true effect while **not** flagging a null one. `ruff check` + `format` clean.

## Caveat stated in the module docstring, the README section, and here

This **isolates the reranker**: the baseline is pure cosine, whereas the real first-stage also applies freshness decay, weight blending and recall boost. It answers *"is the cross-encoder worth its latency"* — not *"what is the end-to-end pipeline delta"*.

### DIFF
--- a/BENCHMARKS.md
+++ b/BENCHMARKS.md
@@ -77,10 +77,34 @@ so you can reproduce the methodology against your own MemClaw instance:
 6. **Latency** — measure p50/p95 of `POST /api/v1/search` under your expected
    concurrency against a warm cache.
 
-> A turnkey harness isn't bundled in this repo yet — the datasets are large and
-> publicly hosted, and the runner is being prepared for open release. Until
-> then the steps above describe the exact methodology. Operator-scale guidance
-> and caveats live in [`docs/performance.md`](docs/performance.md).
+> The end-to-end accuracy/token harness isn't bundled yet — the datasets are
+> large and publicly hosted, and the runner is being prepared for open release.
+> Until then the steps above describe the exact methodology. Operator-scale
+> guidance and caveats live in [`docs/performance.md`](docs/performance.md).
+
+### Reranking, specifically
+
+One piece **is** bundled:
+[`scripts/benchmark_rerank_locomo.py`](scripts/benchmark_rerank_locomo.py)
+answers "does second-stage reranking order results better than first-stage
+similarity alone?" on LoCoMo. It scores the *same* candidate pool two ways —
+embedding cosine vs that pool reordered by a `/rerank` sidecar — against each
+question's `evidence` turns, so recall is identical by construction and the
+only variable is ordering. Reports nDCG@5/@10, MRR and P@5 with paired
+bootstrap confidence intervals and a sign test, overall and per LoCoMo
+category. Stdlib only, no cloud dependencies:
+
+```bash
+python scripts/benchmark_rerank_locomo.py \
+    --dataset locomo10.json \
+    --embed-url http://localhost:8080 \
+    --rerank-url http://localhost:8081
+```
+
+Note this isolates the reranker: the baseline is pure cosine, whereas the real
+first-stage also applies freshness decay, weight blending and recall boost. It
+answers "is the cross-encoder worth its latency", not "what is the end-to-end
+pipeline delta".
 
 ## How MemClaw compares
 
--- a/scripts/benchmark_rerank_locomo.py
+++ b/scripts/benchmark_rerank_locomo.py
@@ -1,0 +1,425 @@
+"""
+Rerank A/B on LoCoMo — does the cross-encoder order results better than
+first-stage similarity alone?
+
+Isolates the reranker so that ORDER is the only variable:
+
+    corpus     one LoCoMo conversation's dialogue turns (~600 of them)
+    baseline   embedding cosine, top --pool          (what first-stage ranks by)
+    treatment  that same pool, reordered by POST {rerank-url}/rerank
+    truth      the QA item's ``evidence`` turn ids   (binary relevance)
+
+Both arms score the SAME pool, so recall@pool is identical by construction —
+asserted at runtime. Any difference is ordering, which is all a reranker does.
+
+Retrieval is per-conversation, matching LoCoMo's setup: each conversation is
+one agent's memory and a question only ever concerns its own conversation.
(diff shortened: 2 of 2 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
## Verified against the dataset — the mapping is correct, and now it says why

Exactly the right thing to flag, and the "please verify rather than confirmed bug" framing was well-judged. I checked empirically against the published `locomo10.json` rather than against a summary, by sampling QA items per category value and reading the questions. Each category is unmistakable from its own contents:

```
category 1 (n=282)   "Where did Caroline move from 4 years ago?"      ev=2   -> multi-hop
                     several items carry TWO evidence turns
category 2 (n=321)   "When did Caroline go to the LGBTQ support group?"      -> temporal
                     every question is "When did ...", every answer a date
category 3 (n=96)    "Would Caroline likely have Dr. Seuss books on her
                      bookshelf?"                                            -> open-domain
                     needs world knowledge beyond the transcript
category 4 (n=841)   "What did the charity race raise awareness for?" ev=1   -> single-hop
                     largest bucket, simplest, one evidence turn, direct lookup
category 5 (n=446)   every answer is literally None                          -> adversarial
```

So `{1: multi-hop, 2: temporal, 3: open-domain, 4: single-hop}` is right for this file. Two independent cross-checks agree:

- **Category 5 anchors the sequence** — all-`None` answers can only be adversarial, and both candidate orderings put adversarial at 5.
- **The bucket sizes only fit one way.** Single-hop is by far the most numerous in LoCoMo (841 here) and open-domain the least (96). The suggested ordering would make open-domain the 841-item bucket and single-hop 282, which contradicts both the sizes and the question text above.

The proposed `1: single-hop, 2: multi-hop, 3: temporal, 4: open-domain` is indeed widely repeated in write-ups — it's a reasonable prior, it just isn't what this file contains. Category 2 being uniformly *"When did…"* + a date answer settles it on its own.

**What I took from the finding:** the file asserted the mapping without showing its evidence, which is precisely why it invited correction — and a wrong "fix" here would silently relabel every category row without erroring. The dict now carries the per-category signatures above plus a `DON'T "correct" this to …` note and instructions to re-check by sampling QA items rather than consulting a summary. Same treatment we gave a model-id that kept getting "fixed" back in a sibling repo.

No behaviour change; `ruff check` + `format` clean.

Eldad-Caura:
@claude re-review — verified the category mapping empirically against locomo10.json (it is correct as written; category 2 is uniformly "When did…"+date, category 5 is all-None) and recorded the evidence in the file so it doesn't get "corrected" back.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #713, opened 2026-08-05
test(keystones): give each keystone test its own tenant

Eldad-Caura:
Fixes the test-isolation bug I hit while running suites for #710 and flagged on that PR.

## What was wrong

All 17 tests in `tests/test_api_keystones.py` shared tenant `default`. Keystones are `documents` rows written through the API's **own committed transaction**, so the per-test session rollback never reaches them — every run left rules behind. Measured: **7 per full-suite run**, unbounded.

That isn't merely untidy, because the listing is capped at `KEYSTONE_MAX_RESULTS` (50) and ordered `weight DESC, updated_at DESC`. Once 50+ rules outrank a freshly-written one, the new rule isn't in the response at all and `test_set_then_list_round_trip` fails — **including in isolation**, which makes it read as a code regression with nothing in the diff to explain it. That's exactly how it presented: 197 rows under `default`, a test failing alone, mid-way through unrelated work.

CI never sees it (fresh DB per run), which is why it went unnoticed.

## The fix

- **Each test gets its own tenant** via `get_test_auth(_ks_tenant())`. The helper already accepted an override and merely defaulted to `"default"`; nothing about these tests needs the shared tenant. The `test-tenant-` prefix is load-bearing — it's what the teardown sweeps on.
- **`documents` added to the teardown's table list**, so those rows are cleaned rather than diluted. Without it the new per-test tenants would still pile up, just harmlessly.

## Two assertions that had let this hide

- `test_list_keystones_empty` asserted only `isinstance(resp.json(), list)` — true of a tenant holding 197 rules — while its docstring claimed *"a tenant with no keystones"*. Now `== []`, which the per-test tenant makes honest.
- The round-trip test now asserts the response wasn't truncated. Truncation and "my rule is missing" are different failures; without this the first presents as the second.

## Reproduction — and two attempts that didn't work

Worth recording, because the naive reproductions pass and would have "confirmed" a wrong fix:

| attempt | result |
|---|---|
| 70 rows at the same weight | **passes** — equal weights order newest-first, so the new rule still lands inside the cap |
| 55 rows at a *lower* weight | **passes** — they sort below it and change nothing |
| 55 rows at weight 100 (what `"high"` stores as) | **fails** on the original assertion |

The third is what the full suite's mix of `weight="high"` writers produces. Against that poisoned database: old code fails, fixed code passes 17/17.

## Verification

| | residual keystone rows after one full suite run | suite |
|---|---|---|
| `main` | 7 | 4117 passed / 1 xfailed / 0 xpassed |
| this branch | **0** | 4117 passed / 1 xfailed / 0 xpassed |

Identical counts, so no test's behaviour changed. Root `tests/` isn't in CI's ruff scope, so nothing was reformatted.

I skipped `/simplify` here: a mechanical call-site substitution plus one teardown entry, no new logic or control flow, and every property asserted above was confirmed by the reproduction rather than by reading.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -147,6 +147,14 @@ async def _setup_schema(_engine):
             "memories",
             "audit_log",
             "agent_activity_digests",
+            # Keystones and other docs live here. Written through the API's own
+            # committed transaction, so the per-test session rollback never
+            # reaches them — without this they survive every run. That is not
+            # only untidy: the keystone listing is capped at 50 and ordered by
+            # weight, so accumulated rules eventually push a freshly-written one
+            # out of the response and a round-trip test fails for reasons that
+            # look nothing like accumulated state.
+            "documents",
         ):
             try:
                 await conn.execute(
--- a/tests/test_api_keystones.py
+++ b/tests/test_api_keystones.py
@@ -55,6 +55,23 @@ def _author_headers(headers: dict, agent_id: str) -> dict:
     return {**headers, "X-Agent-ID": agent_id}
 
 
+def _ks_tenant() -> str:
+    """A tenant of this test's own.
+
+    Keystones are stored as ``documents`` rows and are NOT rolled back by the
+    per-test session — the API writes them through their own committed
+    transaction — so a shared tenant accumulates every keystone every run has
+    ever written. That matters here specifically because the listing is capped at
+    ``KEYSTONE_MAX_RESULTS`` (50) and ordered ``weight DESC, updated_at DESC``, so
+    a freshly-written rule is NOT guaranteed to appear once enough
+    higher-weight rules exist: a round-trip assertion starts failing, in isolation
+    too, and reads as a code regression rather than accumulated local state.
+
+    The ``test-tenant-`` prefix matters — the conftest teardown sweeps on it.
+    """
+    return f"test-tenant-ks-{_uid()}"
+
+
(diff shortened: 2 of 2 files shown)


Eldad-Caura:
@claude

Eldad-Caura:

CI is green and the Claude review came back with no issues, but I want to be straight about what that verdict is worth here: it reports *"Both changed files are under `tests/`, which is excluded from review per the instructions."* So it's a non-review rather than a substantive pass — it carries no signal about whether the change is correct. Worth knowing before leaning on it.

What the change actually rests on is the reproduction, which I'd point at instead:

| poison under tenant `default` | old code | fixed |
|---|---|---|
| 70 rows, same weight | passes | passes |
| 55 rows, *lower* weight | passes | passes |
| 55 rows at weight 100 (what `"high"` stores as) | **fails** on the original assertion | **17/17 pass** |

The first two rows are the reason I don't trust a "reproduced it" claim cheaply here — equal weights order newest-first so the new rule stays inside the 50-row cap, and lower-weight rows sort below it. Only 50+ *higher*-weight rules push a fresh one out of the response, which is what the full suite's `weight="high"` writers accumulate.

And the measurable outcome: residual keystone rows after one full suite run go **7 → 0**, with the suite identical at 4117 passed / 1 xfailed / 0 xpassed, so no test's behaviour changed.

erni-a (approved):

PR #714, opened 2026-08-05
feat(bench): first-stage blend A/B on LoCoMo, and share the harness plumbing

Eldad-Caura:
Extends #712 one stage earlier, so the `fts_score` scale question from [HANDOFF-2026-08-04](HANDOFF-2026-08-04-fts-score-scale-normalisation.md) is finally measurable — **and measured**.

## Why #712 couldn't answer it

#712 holds a cosine-selected pool fixed and A/Bs a reranker's ordering, so recall is constant *by construction*. Correct for judging a reranker; wrong for a first-stage change, which moves **which rows are retrieved**. Verified rather than assumed: that script contains no reference to `fts`, `fts_score`, `FTS_WEIGHT`, `similarity_blend` or `ts_rank_cd`.

## What this measures

The keyword half of the production blend, `similarity = (1 - FTS_WEIGHT) * vec_sim + FTS_WEIGHT * fts_score`, by rescaling `ts_rank_cd` before saturation (`--k`; `k=1` is today's `r/(1+r)`), ranking the **whole corpus** in both arms so recall can move. Truth is each QA item's `evidence` turn ids. Same paired bootstrap CI + sign test as #712.

## The numbers

Full LoCoMo, **n=1531** scorable questions (FTS matching on 257), MiniLM-L6-v2, `FTS_WEIGHT=0.3`. Delta vs today; `*` = bootstrap CI excludes zero.

| k | overall recall@5 | fts-matched-only recall@5 | matched-only MRR |
|---|---|---|---|
| 1 | +0.0000 | +0.0000 | +0.0000 |
| 3 | +0.0035\* | +0.0211\* | +0.0080\* |
| 6 | +0.0042\* | +0.0250\* | +0.0185\* |
| 10 | +0.0075\* | +0.0445\* | +0.0357\* |
| 20 | +0.0113\* | +0.0672\* | +0.0692\* |

**Recall does not move the wrong way.** It improves at every `k > 1`, significantly on all five metrics, with **zero** queries getting worse on recall@5/@10/@20.

## Three limits I don't want quoted away

- **It does not locate an optimum.** The effect rises monotonically across the entire range tested — so this says more keyword weight helps *on LoCoMo*, not that `k=20` is right. Picking a production `k` needs an argument this benchmark doesn't supply.
- **Weak instrument for this particular change.** LoCoMo questions are paraphrases, so only ~17% lexically match any turn; the other 83% are a mathematical no-op that dilutes the overall delta. Hence the separate `fts-matched-only` block. Production queries skewing more keyword-shaped would show a larger effect — which is what `_adaptive_fts_weight` already assumes.
- **Blend in isolation.** No freshness decay, weight blending, recall boost, or candidate LIMIT. Not the end-to-end pipeline delta.

All three are in `BENCHMARKS.md` too, so the numbers aren't quotable without them.

## Design decisions worth a reviewer's eye

- **Real Postgres for `ts_rank_cd`.** Reimplementing Postgres FTS in Python would measure a different function, and that function's output scale *is* the question. This costs #712's stdlib-only property, so it's confined to a `TEMP` table that never touches application tables, and mirrors production exactly (bare `to_tsvector('english', content)` per migration 001, `plainto_tsquery('english', q)`).
- **`k=1` asserts a delta of exactly zero** on every metric *and* subgroup — the treatment formula *is* the baseline there, so any movement means the harness is manufacturing differences. Passing.
- **Index breaks score ties**, so both arms order equal-scoring docs identically and a tie can't masquerade as a win.

## The refactor, and how it was verified

Shared plumbing moved to `scripts/_locomo_bench.py` (loader, embed client, nDCG/MRR/recall, paired stats) rather than copied. `benchmark_rerank_locomo.py` imports it and **loses 144 lines**, including its own inline bootstrap.

I can't run that harness here — the TEI reranker has no arm64 image, which the repo's own compose overlay notes. So I verified the refactor by **equivalence** instead: same synthetic scored rows, same seed, through `main`'s `report()` and the rewired one →

```
stdout identical: True
returned dict identical: True
```

No test imports either script, and `scripts/` is outside CI's ruff scope. Both harnesses parse, `--help`, and reject bad `--k` / `--fts-weight`.

### DIFF
--- a/BENCHMARKS.md
+++ b/BENCHMARKS.md
@@ -106,6 +106,48 @@ first-stage also applies freshness decay, weight blending and recall boost. It
 answers "is the cross-encoder worth its latency", not "what is the end-to-end
 pipeline delta".
 
+### First-stage retrieval, specifically
+
+[`scripts/benchmark_blend_locomo.py`](scripts/benchmark_blend_locomo.py) is the
+companion one stage earlier: it varies the **first-stage** score, so which turns
+are retrieved at all is what moves, and **recall** is the headline rather than an
+invariant. It A/Bs the keyword half of
+
+```
+similarity = (1 - FTS_WEIGHT) * vec_sim + FTS_WEIGHT * fts_score
+```
+
+by rescaling `ts_rank_cd` before saturation (`--k`; `k=1` is today's
+`r/(1+r)` formula), ranking the *whole* corpus in both arms. Reports
+recall@5/@10/@20, nDCG@10 and MRR with the same paired bootstrap CI and sign
+test, overall, per category, and — separately — over just the queries where
+`plainto_tsquery` matched anything, since the rest are a mathematical no-op that
+only dilutes the average.
+
+Unlike the rerank harness it needs a Postgres, because `ts_rank_cd` is the thing
+under test and reimplementing Postgres FTS in Python would measure something
+else. Any empty database works; it uses a `TEMP` table and never touches
+application tables.
+
+```bash
+python scripts/benchmark_blend_locomo.py \
+    --dataset locomo10.json \
+    --embed-url http://localhost:8080 \
+    --pg-dsn postgresql://memclaw:changeme@localhost:5433/memclaw \
+    --k 6
+```
+
+`k=1` asserts a delta of exactly zero on every metric — a self-check that the
+harness isn't manufacturing differences the change cannot cause.
+
+Two limits worth knowing before quoting a number from it. Only ~17% of LoCoMo
+questions lexically match any turn (they are paraphrases), so the overall delta is
+diluted by a large majority of queries the change cannot affect — read the
+`fts-matched-only` block for the effect on queries it applies to. And the measured
+effect rises monotonically with `k` across the range tested, so this benchmark can
+say whether more keyword weight helps on LoCoMo, but it does **not** locate an
+optimum; don't read the largest `k` as the best one.
+
 ## How MemClaw compares
 
 For a single chatbot, the public-benchmark leaders (MemClaw, Mem0, Zep) cluster
--- a/scripts/_locomo_bench.py
+++ b/scripts/_locomo_bench.py
@@ -1,0 +1,187 @@
+"""Shared pieces for the LoCoMo benchmark harnesses.
+
+Extracted from ``benchmark_rerank_locomo.py`` when the first-stage blend harness
+(``benchmark_blend_locomo.py``) needed the same dataset loader, embedding client
+and paired statistics. The two harnesses measure different stages and keep their
+own docstrings, arms and metrics; only the parts that are genuinely stage-agnostic
+live here.
+
+Stdlib only, deliberately — the rerank harness runs against ``docker compose``
+with no installed dependencies, and this must not take that away from it.
+"""
+
+from __future__ import annotations
+
+import ast
+import collections
+import json
+import math
+import statistics
+import urllib.error
+import urllib.request
+
+CATEGORIES = {1: "multi-hop", 2: "temporal", 3: "open-domain", 4: "single-hop"}
+
+
+# ------------------------------------------------------------------------ http
+
+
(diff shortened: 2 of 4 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
Critical and correct — the rerank benchmark was fully broken. Fixed in `8937033`: `_post` is imported back from the shared module.

Also dropped a now-unused `CATEGORIES` import that `ruff --select F401` flagged alongside it. I deliberately did **not** take the tempting next step of switching `main()`'s hardcoded `("single-hop", "multi-hop", "temporal", "open-domain")` to `CATEGORIES.values()` — their orders differ, so that would silently reorder #712's report sections as a side effect of a lint fix.

**The miss is worth more than the bug.** I claimed this file was verified, resting on `--help` plus a `report()` equivalence test. Neither could ever have caught it: `--help` only proves the module *imports*, and an undefined global raises at **call** time, which the equivalence test never reaches because it exercises `report()`, not `rerank()`. I'd written off running the harness end to end because the TEI reranker has no arm64 image — then treated the cheap checks as a substitute for the expensive one, instead of looking for a cheaper way to do the real thing.

There was one, and it took about 20 lines of stdlib:

```python
class H(BaseHTTPRequestHandler):
    def do_POST(self):   # TEI-shaped /rerank; order need not be good, only well-formed
        ...{"results": [{"index": i, "score": overlap} ...]}
```

Both harnesses now run **end to end** locally — the rerank one against that stub plus real TEI embeddings, the blend one against Postgres. The rerank run produces a complete per-category report, so this class of breakage can't survive again.

Two things worth flagging for the repo, not just this PR:

- `ruff check --select F821` over these three scripts catches exactly this, and is now clean. But **`scripts/` is outside CI's ruff scope**, so nothing in CI would have caught it either — #712 and this PR both sit in an unlinted directory.
- No test imports either script, so the test suite is silent on them too. The end-to-end stub is currently the only thing that exercises `rerank()` at all.

Re-verified after the fix: `report()` still byte-identical to `main`'s on the same synthetic rows and seed; the blend harness's `k=1` self-check still exact across all six subgroups; `F821`/`F401` clean.

Eldad-Caura:
@claude

Eldad-Caura:
Right — I simplified that query to a plain single-table `SELECT` and left the comment describing the earlier drafted version. Fixed in `46b988d`.

Checking your claim also turned up something adjacent: the `COALESCE(..., 0.0)` was **unreachable**, so it's gone too. Verified rather than reasoned:

```
ts_rank_cd(to_tsvector('english','totally unrelated text'),
           plainto_tsquery('english','zzz nomatch'))  ->  0     (is_null = f)
ts_rank_cd(NULL::tsvector, plainto_tsquery('english','x'))      is_null = t
```

So a non-match scores a genuine `0`, and NULL arises only from a NULL tsvector — which `to_tsvector` of a non-null text cannot produce, and the insert always supplies one. Keeping the COALESCE implied a NULL case that doesn't exist; the comment now records why, so it doesn't get re-added.

The corrected comment describes the actual mechanism: `ORDER BY t.i` holds row order to `texts`, and the zero comes from `ts_rank_cd` itself.

Re-verified: `k=1` self-check still exact on every metric and subgroup, `F821`/`F401` clean across all three scripts.

---

Unrelated to this PR but worth recording here, since it's the first evidence of this session's merges running for real: `main` at `ccec1e6` is now **deployed to staging** (enterprise `Deploy to Staging` run 31006304782, all 10 jobs green, OSS checkout verified at `ccec1e6d38098640f6d6b4272bc6084674592a56`), and the full staging e2e suite was dispatched against it — broker contract canary, broker VM e2e, dist-install e2e all **PASS**, nightly green on 8/9 jobs with the last still running. The VM e2e's payload is byte-identical to every run back to 2026-08-02, so nothing in #700/#706/#710/#711/#713 moved it.

Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #721, opened 2026-08-05
fix(embedding): chunk bulk embeds to the backend's batch cap, and report bulk-only outages

Eldad-Caura:
## The premise changed once I checked prod

The task was "make the embedding fallback audible — one warning when a bulk embed fails and cascades to per-item, so this can never again be invisible."

**The fallback was already audible.** `memory_service.py:2080` logs at ERROR with a full traceback, and the traceback ends in:

```
openai.APIStatusError: Error code: 413 - {'message': 'batch size 50 > maximum allowed batch size 32', 'code': 413, 'type': 'Validation'}
```

87 of those in prod over 30 days, structured JSON, with `dd.trace_id`. The batch size, the cap, and the diagnosis were all already in our own logs. Nobody was reading them. So adding a second log line would have been the one thing guaranteed not to help.

Checking the codebase turned up what was actually missing.

## Finding 1 — the real fix is chunking, and the sibling module already shipped it

`common/ranking` hit **the same 413, from the same sidecar**, and fixed it in #709: `RANK_REMOTE_MAX_BATCH` (default 32, explicitly "TEI's own `--max-client-batch-size` default") chunks the candidate pool so the cap *cannot* reject it. `tests/test_ranking_remote.py` even asserts against `batch size 51 > maximum allowed batch size 32`.

The embedding provider had no equivalent. `OpenAIEmbeddingProvider.embed_batch` sent `input=texts` verbatim, and callers send up to `BULK_MAX_ITEMS` (100) or `EMBEDDING_REEMBED_BATCH_SIZE` (50).

So this adds `EMBEDDING_REMOTE_MAX_BATCH` (default 32) and chunks inside the provider. **This makes the bug impossible rather than merely audible** — and it fixes it for every deployment, not just the one where someone remembered the server flag.

That includes the shipped OSS stack: `docker-compose.yml` starts TEI with only `--model-id` and `--auto-truncate`, so **any bulk write over 32 items 413s today** for anyone following `docs/local-embedder.md`.

Sequential, not the reranker's concurrent `gather` — `_call_gated` holds one `EMBEDDING_MAX_CONCURRENCY` slot for the whole call, so fanning out inside it would multiply real provider concurrency by the chunk count and defeat a cap that exists because of the 2026-07-27 pool exhaustion. Bulk callers arm 30s where the reranker had 0.5s, so there's no need.

## Finding 2 — the aggregate trip-wire was structurally unable to fire

`_EmbeddingStats` is process-wide and shared by every call path. core-api serves search constantly, so every successful *query* embed resets `consecutive_failures`. A 100%-broken bulk path therefore never accumulates a streak.

Verified against prod: `"Embedding service degraded"` never fired **once** in 30 days, while staging fired it repeatedly for an unrelated condition.

Fixed with a bulk-only streak that only a bulk success clears, reported at WARNING with the batch size. What it adds over the existing ERROR is the **consecutive count** — "N failures with no bulk success between" distinguishes systematically-broken from flaky, which a stream of independent tracebacks cannot. It is explicitly *not* the per-occurrence detector; that's the existing ERROR plus the Datadog monitor in enterprise#980.

## Scope note

Chunking is wider than the task as written. I judged it in scope because the goal was "this can never again be invisible" and chunking means it never happens — and because leaving OSS without the fix its own sibling module already has is the kind of asymmetry that bites the next person. Say the word and I'll split it.

## Known limitation, documented in code

The streak is not keyed on the provider, so in a multi-tenant process one tenant's healthy bulk success clears another's broken backend — the same masking, one level down. Keying it properly means keying all of `_stats` (ranking's `_permanent_scope` is the model), which is a wider change. The docstring says so rather than implying a guarantee the code doesn't give.

## Tests

Four new tests, **each verified to fail without its fix**:

| Test | Claim |
|---|---|
| `test_embed_batch_splits_requests_above_max_batch` | 50 texts (the incident's exact size) → `[32, 18]`, nothing over the cap |
| `test_embed_batch_at_or_below_cap_makes_one_request` | fast path preserved — chunking costs nothing at ≤32 |
| `test_embed_batch_preserves_input_order_across_chunks` | `result[i]` still belongs to `texts[i]` (callers zip positionally) |
| `test_bulk_failure_reports_even_while_single_embeds_succeed` | interleaves a successful single embed after every bulk failure — the shape that defeated the shared streak — and asserts `consecutive_failures == 0` as a *precondition* so the test can't silently stop exercising the bug |

Also reviewed and applied: dropped a dead `batch_size` fixture parameter, and `_CappedProvider` now subclasses the real `FakeEmbeddingProvider` so the stand-in actually satisfies the `EmbeddingProvider` protocol.

## Corrections made during review

Two comments I'd written asserted things that turned out to be false, and are fixed:
- "keeps log volume off the traffic curve … bounded number of times" — it's a 10x reduction, linear not bounded.
- A justification citing Vertex behaviour — there is no Vertex embedding provider; `_platform.py` rejects it outright.

## Checks

`ruff check common/` (discovery, no `--config`) · `ruff check core-api/src/` · `ruff format --check --isolated` — clean.
Full `pytest tests/ -m "not benchmark"`: **4090 passed**, and the 26 failures + 8 errors are byte-identical to pristine `main` (no local Postgres). Zero regressions. `core-api/uv.lock` untouched.

### DIFF
--- a/common/embedding/_service.py
+++ b/common/embedding/_service.py
@@ -31,6 +31,10 @@ class _EmbeddingStats:
     Three consecutive failures fire a single ERROR log (per cycle) so a
     sustained provider outage shows up loudly without spamming once-per-
     request. Reset on the next success.
+
+    Bulk calls additionally keep their OWN streak, which only a bulk
+    success clears — see :meth:`record_failure` for the outage the shared
+    streak structurally cannot see.
     """
 
     def __init__(self) -> None:
@@ -38,27 +42,94 @@ class _EmbeddingStats:
         self.successes = 0
         self.last_failure_time = 0.0
         self.consecutive_failures = 0
+        self.consecutive_bulk_failures = 0
         self._lock = asyncio.Lock()
 
-    async def record_success(self) -> None:
+    @staticmethod
+    def _is_report_point(streak: int) -> bool:
+        """Report on first detection (3) then every 10th (13, 23, 33, …).
+
+        Loud enough to alert on a fresh outage, quiet enough not to spam
+        during a sustained one — a 10x reduction, NOT a fixed bound:
+        reports still grow linearly with a condition that recurs on every
+        request, so this caps the slope, not the total. The next success
+        resets the streak, so a later outage re-fires from 3.
+        """
+        return streak >= 3 and (streak - 3) % 10 == 0
+
+    async def record_success(self, *, bulk: bool = False) -> None:
+        """*bulk* additionally clears the bulk-only streak."""
         async with self._lock:
             self.successes += 1
             self.consecutive_failures = 0
+            if bulk:
+                self.consecutive_bulk_failures = 0
 
-    async def record_failure(self) -> None:
+    async def record_failure(self, *, bulk_batch_size: int | None = None) -> None:
+        """Count a failure and report at :meth:`_is_report_point`.
+
+        Passing *bulk_batch_size* marks this as a failed BULK call: it
+        advances a second streak that only a bulk success clears, and
+        reports it separately.
+
+        That second streak exists because the shared one cannot see a
+        bulk-only outage. Every path through this module shares one
+        process-wide ``_stats``, so a service that also serves search
+        traffic records a success per query embed, which resets
+        ``consecutive_failures`` and holds it under the threshold no
+        matter how many bulk calls fail in between. Not hypothetical:
+        prod TEI rejected 100% of bulk embeds for 30+ days on a
+        batch-size cap and "Embedding service degraded" never fired once,
+        because query embeds kept clearing the streak.
+
+        What the bulk report adds over the callers' own per-batch ERROR
+        (which already carries the provider's traceback) is the CONSECUTIVE
+        count: N failures with no bulk success in between distinguishes
+        "systematically broken" from "occasionally flaky", which a stream
+        of independent tracebacks cannot. It is not the per-occurrence
+        detector — the callers' ERROR is — so WARNING, not ERROR: by
+        contract this cascades into the per-item fallback, which usually
+        still persists correct embeddings. Degradation, not data loss.
+
+        *bulk_batch_size* is the size the CALLER asked for, which is not
+        necessarily the size of any single request — the provider splits
+        oversized requests to its own backend's cap. The message says so
+        rather than naming a number: this class is provider-agnostic (Fake
+        and Local never chunk; the OpenAI provider's cap is 32 self-hosted
+        but 2048 hosted), so quoting one constant here would be wrong for
+        most backends and would send an operator to re-tune a knob their
+        deployment does not use — the same misdirection in the opposite
+        direction. The size is still worth reporting: it identifies which
+        caller is affected (bulk write at 100 vs re-embed at 50) without
+        expanding a traceback.
+
+        LIMITATION worth knowing: the streak is process-wide and NOT keyed
+        on the provider, so in a multi-tenant process one tenant's healthy
+        bulk success clears a different tenant's broken backend — the same
+        masking, one level down. ``common/ranking``'s ``_permanent_scope``
+        is the per-backend answer; applying it here means keying all of
+        ``_stats``, which is a wider change than this. Do not read the
+        streak as per-backend.
+        """
         async with self._lock:
             self.failures += 1
             self.consecutive_failures += 1
             self.last_failure_time = time.monotonic()
-            # Fire on first detection (failure 3) and then every 10
-            # failures thereafter (13, 23, 33, …). Loud enough to alert
-            # on a fresh outage, quiet enough not to spam during a
-            # sustained provider degradation. The next success resets
-            # the counter, so a subsequent streak re-fires from 3.
-            if (
-                self.consecutive_failures >= 3
-                and (self.consecutive_failures - 3) % 10 == 0
-            ):
+            if bulk_batch_size is not None:
+                self.consecutive_bulk_failures += 1
+                if self._is_report_point(self.consecutive_bulk_failures):
+                    logger.warning(
+                        "Bulk embedding failing: %d consecutive bulk call(s) "
+                        "failed (requested batch=%d), cascading to the "
+                        "per-item fallback. Embeddings may still be correct; "
+                        "batching is not. The provider already splits a "
+                        "request to its own backend's cap, so a batch-size "
+                        "rejection is unlikely — look at auth, network, "
+                        "quota and the callers' timeouts.",
+                        self.consecutive_bulk_failures,
+                        bulk_batch_size,
+                    )
+            if self._is_report_point(self.consecutive_failures):
                 logger.error(
                     "Embedding service degraded: %d consecutive failures (total: %d/%d)",
                     self.consecutive_failures,
@@ -169,6 +240,14 @@ async def get_embeddings_batch(
     degraded-provider trip-wire fires consistently with the single-embed
     paths (``get_embedding`` / ``get_query_embedding``).
 
+    A failed ``embed_batch`` also advances a bulk-only failure streak that
+    single-embed successes cannot reset — see ``record_failure`` for why
+    that is needed and for the incident behind it.
+
+    Note this is the SERVICE layer: a request larger than the backend's
+    accepted batch is split inside the provider, which is the only layer
+    that knows its own cap, so *texts* has no length limit imposed here.
+
     Two error shapes are explicitly accounted for:
 
     1. ``ValueError`` from ``get_embedding_provider`` — env-var
@@ -176,10 +255,11 @@ async def get_embeddings_batch(
        ``SEND_DIMENSIONS`` mismatch). Used to propagate as an unhandled
        exception that bulk callers caught generically but bypassed
        ``_stats.record_failure``, so the trip-wire never fired under
-       sustained misconfig. Now records a failure and re-raises.
+       sustained misconfig. Now records a failure and re-raises. Counted
+       as a plain failure, NOT a bulk failure — the batch never ran.
     2. Any provider-side exception from ``embed_batch`` — auth, HTTP
-       client errors, Vertex quota, etc. Same record_failure + re-raise
-       contract as before.
+       client errors, provider quota, a batch-size cap. Re-raises as
+       before; now also counted as a bulk failure.
     """
     provider_name = _resolve_provider_name(tenant_config)
     # Provider construction and embed dispatch are wrapped in *separate*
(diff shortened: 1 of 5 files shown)


Eldad-Caura:
@claude please review

Eldad-Caura:
Both findings were real. Fixed, but the High one not the way the prompt suggested — the suggested patch has a side effect worth naming.

## High — cancellation skips the trip-wire: confirmed, and worse than described

Verified the constants rather than trusting the arithmetic:

| | |
|---|---|
| `OPENAI_REQUEST_TIMEOUT_SECONDS` | 25.0 (per request) |
| `BULK_EMBEDDING_TIMEOUT_SECONDS` | 30.0 |
| **`BULK_STRONG_EMBED_TIMEOUT_SECONDS`** | **8.0** |
| worst case, chunked 100 texts | 100.0s |

The review reasoned against the 30s deadline. The tighter one is **8s** — and at 25s per request, *one* request already blows through it. So this accounting gap **predates this PR** on the strong-embed path; chunking widens it to the 30s inline path as well. That makes it more worth fixing, not less.

## Why not `except (Exception, asyncio.CancelledError)`

That records a bulk failure for *every* cancellation, including process shutdown and client disconnect. This codebase deliberately relies on the opposite — `memory_service.py:2078` documents `CancelledError` inheriting `BaseException` precisely so "shutdown-path cancellations still propagate" rather than being treated as provider failures. Applying it verbatim would inflate the streak on every deploy and scale-down.

I took the same catch (as `except BaseException`) but justified the cost instead of leaving it implicit, and documented three things the prompt asked to "verify":

- **Cancellation still propagates** — the `raise` is unconditional.
- **The `await` in the handler cannot be re-interrupted.** Every critical section in `_EmbeddingStats` is await-free, so `asyncio.Lock.acquire()` takes its fast path and never yields. There is no suspension point for a pending cancellation to be re-delivered at. If one ever were, we land back on today's behaviour — stat missed, cancellation still propagating — never worse.
- **The false-positive cost is bounded**: 3 consecutive failures are needed to report, the next bulk success clears the streak, and on the shutdown path the counter dies with the process anyway.

New test `test_bulk_failure_counted_when_caller_deadline_cancels_us` pins it — it **fails with `except Exception` and passes with `BaseException`** (verified by reverting the one line).

## Low — misleading `batch=` : fixed, taking the first option

Agreed, and the stale hint was the worse half. `"Check the provider's batch-size cap first"` is advice for a bug this PR makes unreachable. The message now reports both numbers and rules the cap out explicitly:

> `Bulk embedding failing: 3 consecutive bulk call(s) failed, cascading to the per-item fallback. Embeddings may still be correct; batching is not. Requested batch=100, split into requests of at most 32 — so this is NOT a batch-size cap; look at auth, network, quota and the callers' timeouts.`

`bulk_batch_size` is kept: it still identifies *which* caller is affected (bulk write at 100 vs re-embed at 50) without expanding a traceback.

## Checks

`ruff check common/` · `ruff check core-api/src/` · `ruff format --check --isolated` — clean.
Full `pytest tests/ -m "not benchmark"`: **4091 passed**; the 26 failures + 8 errors are byte-identical to pristine `main` (no local Postgres). Zero regressions. `uv.lock` untouched. Single commit, amended.

Eldad-Caura:
@claude please re-review

Eldad-Caura:
Applied. Good catch — this was a real regression I introduced for hosted deployments, and the fix uses a discriminator the codebase already trusts.

## What changed

The cap is now **derived from the backend** rather than applied flat, in `OpenAIEmbeddingProvider.__init__` where `base_url` already lives — so the registry needs no change and no new plumbing:

- **`base_url` set** → self-hosted OpenAI-compatible server whose admission limit we cannot see → `EMBEDDING_REMOTE_MAX_BATCH` (32, TEI's own default).
- **`base_url` unset** → hosted OpenAI, documented at 2048 inputs/request → new `EMBEDDING_HOSTED_MAX_BATCH` (2048).

This is the same discriminator the registry already uses to decide `send_dimensions` (it *raises* if `base_url` is set with `send_dimensions=True`), so "is this a TEI-like backend" is an established question in this module, not a new inference.

## Two constants rather than one, deliberately

32 is a **guess about an opaque sidecar**; 2048 is a **documented fact about a known API**. Collapsing them into one number would mean either re-introducing the 4x round-trip cost on hosted, or weakening the self-hosted default to something unsafe. Both are env-overridable, so a hosted deployment that wants smaller requests (rate-limit shaping, latency smoothing) still has a config-only lever — which the "pass `None` to mean no chunking" variant would not have given.

I kept `max_batch` injectable (now `int | None`, defaulting to derive) purely so tests can exercise chunking without a 33-text fixture. The registry still never passes it, which is why it stays out of the cache key.

## Tests

Two added, both fail without the change:

- `test_hosted_openai_is_not_chunked_at_the_self_hosted_cap` — 100 texts, no `base_url`, asserts **exactly one** request.
- `test_self_hosted_base_url_is_chunked_at_the_conservative_cap` — 100 texts with `base_url`, asserts the `[32, 32, 32, 4]` split.

## Checks

`ruff check common/` · `ruff check core-api/src/` · `ruff format --check --isolated` — clean.
Full `pytest tests/ -m "not benchmark"`: **4093 passed**; 26 failures + 8 errors byte-identical to pristine `main` (no local Postgres). Zero regressions. `uv.lock` untouched. Single commit, amended.

Eldad-Caura:
@claude please re-review

Eldad-Caura:
Medium fixed. It was a real defect and I introduced it in the previous round — worth naming plainly.

## The Medium is a self-inflicted regression from the last fix

When I made the cap backend-dependent (32 self-hosted / 2048 hosted, previous round), the log message kept hardcoding `EMBEDDING_REMOTE_MAX_BATCH`. So on hosted OpenAI it would have claimed *"split into requests of at most 32"* for a request that **was never split at all** — and `Fake` / `Local` never chunk either. That is precisely the misdirection the sentence was added to prevent, pointed the other way.

Took option **(a)**: drop the number, keep the qualitative framing.

> `Bulk embedding failing: 3 consecutive bulk call(s) failed (requested batch=100), cascading to the per-item fallback. Embeddings may still be correct; batching is not. The provider already splits a request to its own backend's cap, so a batch-size rejection is unlikely — look at auth, network, quota and the callers' timeouts.`

Not option (b): adding `max_batch` to the `EmbeddingProvider` protocol would widen a shared interface — implemented by four providers, two of which have no such concept — to improve one log line. The qualitative form says the true thing for every backend.

Also corrected two docstrings that named the constant, and dropped the now-unused import (`ruff` would have caught it as F401 — `common/` is lint-gated).

## Low (partial results discarded) — declining, with reasoning

Agreed on the mechanics, and I'll take the review's own framing: correctness is unaffected, and the systematic failure this PR targets fails on chunk 1 anyway, so there is no prefix to save in the case that motivated the change.

Against that, preserving the prefix means either a new exception type carrying partial results or callers that accept partial batches — a contract change across `get_embeddings_batch` and both bulk callers, to save re-embedding ≤3 chunks on a transient mid-batch blip that the per-item fallback already handles correctly. Wrong trade for this PR. Worth revisiting if bulk-embed latency ever shows up as a real cost.

## Note for whoever merges

This branch was rebased onto #714 via the GitHub button mid-review; I re-synced, re-applied on the new base, and re-ran everything there. Still a single commit.

## Checks

`ruff check common/` · `ruff check core-api/src/` · `ruff format --check --isolated` — clean.
Full `pytest tests/ -m "not benchmark"`: **4093 passed**; 26 failures + 8 errors byte-identical to pristine `main` (no local Postgres). Zero regressions. `uv.lock` untouched.

Eldad-Caura:
@claude please re-review

Eldad-Caura:
One fixed (and it was worse than reported), one answered with measurements instead of an assumption, one confirmed intentional.

## Low — `max_batch` validation: fixed, and the impact was silent data loss

The review called this "unvalidated". It's sharper than that — I checked what the values actually do:

| value | behaviour |
|---|---|
| `max_batch=0` | `ValueError: range() arg 3 must not be zero`, deep in the loop |
| **`max_batch=-1`** | **`range(0, n, -1)` is empty → returns ZERO embeddings for n texts, no error** |

The second one matters: `memory_service` zips the result onto rows positionally, so a negative cap silently drops every embedding in the batch. That's data loss with no exception anywhere.

Now validated at construction, matching the `truncate_to_dim` guard six lines above it — which exists for exactly this reason ("direct construction … bypasses that path … fail fast at construction instead"). The env path was already safe (`read_int_env` rejects non-positive); direct construction from a script or test was not.

Added `test_openai_provider_init_rejects_non_positive_max_batch`, parametrized over `0` and `-1`.

## Medium — gate hold time: measured rather than assumed

Fair challenge to the docstring's "~10ms" claim. Here is the real distribution, from **300 prod TEI request logs over 2 days** (the sidecar's own `total_time`):

| p50 | p95 | p99 | max |
|---|---|---|---|
| 4.1ms | 10.2ms | 29.1ms | 59.5ms |

So four sequential chunks cost **0.04s at p95** and **0.24s at the worst single request ever observed** — against the tightest budget of 8s (`BULK_STRONG_EMBED_TIMEOUT_SECONDS`). That's ~33x margin at worst-observed. The "~10ms" assumption is accurate (p95 = 10.2ms). Caveat: `total_time` is server-side and excludes network RTT, but the call is in-VPC.

On the gate itself, I'd argue sequential is still right, and for the reason the finding raises. The gate's invariant is *"no more than `EMBEDDING_MAX_CONCURRENCY` concurrent provider calls"*. Sequential chunking **preserves that exactly** — one in-flight request per slot — and only extends hold time. Concurrent chunking would **breach** it, multiplying real provider concurrency by the chunk count, which is precisely the 2026-07-27 connection-pool exhaustion the gate was built to stop. Trading a bounded 40ms hold-time increase for a cap violation is the wrong direction.

I have not run the load test the prompt asks for — that belongs in the loadtest repo against a loaded sidecar, and I didn't want to gate this on it. The measurements above are what I can honestly stand behind. If you'd rather hold the merge for a load test, say so.

## Low — partial results: confirmed intentional

Yes, deliberate. The systematic failure this PR targets (a batch-size cap) fails on chunk 1, so there is no successful prefix to preserve in the motivating case. Saving it otherwise means either a new exception carrying partial results or callers that accept partial batches — a contract change across `get_embeddings_batch` and both bulk callers, to avoid re-embedding ≤3 chunks on a transient mid-batch blip that the per-item fallback already handles correctly. Wrong trade here; worth revisiting if bulk-embed cost ever shows up as a real number.

## Note

Branch was rebased onto #722 mid-review (second time — #714 before). Re-synced, re-applied, re-verified on the new base. Still one commit.

## Checks

`ruff check common/` · `ruff check core-api/src/` · `ruff format --check --isolated` — clean.
Full `pytest tests/ -m "not benchmark"`: **4095 passed**; 26 failures + 8 errors byte-identical to pristine `main`. Zero regressions. `uv.lock` untouched.

Eldad-Caura:
@claude please re-review

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #722, opened 2026-08-05
fix(search): put fts_score on the cosine scale (#687)

Eldad-Caura:
Closes the deeper half of #687. #700 made an unembedded FTS-matching row *reachable*; it could not make a lexical match **rank** like one.

## The defect

`fts_score = ts_rank_cd/(1+ts_rank_cd)` saturated a modal single-term match to **0.0909**, while cosine on the same corpus measures **0.35–0.39**. Blended as if comparable, the `FTS_WEIGHT = 0.3` this repo declares delivered an **effective ~0.09** — keyword relevance structurally under-weighted on *every* query, not just the deferred-embed ones #687 was filed about.

Root cause: migration 001 builds `search_vector` as a bare `to_tsvector('english', content)` with no `setweight`, so every lexeme is weight D = 0.1.

## The fix, and why k=6 is derived rather than picked

`(k*r)/(1+k*r)` with `FTS_RANK_SCALE = 6.0`. Solving `k*0.1/(1+k*0.1) ≈ 0.37` lands a modal match in the cosine range, bringing the effective weight to **0.292** against the nominal **0.300**. `FTS_WEIGHT` remains the *preference* between signals; this is the *unit conversion* that makes the preference mean anything.

## Measured before shipping

Via #714's harness — full LoCoMo, **n=1531**, `FTS_WEIGHT=0.3`. `*` = paired bootstrap CI excludes zero:

| k | overall recall@5 | fts-matched-only recall@5 | matched-only MRR |
|---|---|---|---|
| 1 | +0.0000 | +0.0000 | +0.0000 |
| 3 | +0.0035\* | +0.0211\* | +0.0080\* |
| **6** | **+0.0042\*** | **+0.0250\*** | **+0.0185\*** |
| 10 | +0.0075\* | +0.0445\* | +0.0357\* |
| 20 | +0.0113\* | +0.0672\* | +0.0692\* |

Recall improves at every `k > 1`, significantly on all five metrics, with **zero queries regressing** on recall. The sweep rises monotonically across the whole range, so it confirms direction but **does not locate an optimum** — 6 is the derivation's answer, not the sweep's, and that's recorded beside the constant so nobody bumps it to 20 citing the benchmark.

Two honesty caveats a reviewer should hold me to: the calibration is **at the mode**, not distributional — `ts_rank_cd`'s cover-density penalty isn't a constant factor, so a spread-out multi-term match still lands below the band (arguably correct, but "effective ≈ nominal" describes the typical query). And LoCoMo questions are paraphrases, so only ~17% lexically match anything; the matched-only column is the effect on queries the change can reach.

## It had to go in four places — the fourth is the interesting one

- `resolve_search_profile` — the pipeline's `search_params`
- `_search_memories_legacy` — its own, sent **flat** rather than nested
- **`_SEARCH_PARAM_KEYS` in the storage router** — an allowlist that rebuilds `search_params` from flat keys and **silently drops anything unnamed**. Adding the key to both core-api builders was *not enough*: the legacy path kept the old ranking until it was named here. **The test caught that, not review.**
- `postgres_service`, which reads it and defaults to **1.0** when absent — so a storage revision deployed ahead of core-api keeps today's ranking rather than changing it. Deploys aren't atomic across the two services and the safe direction for the gap is "unchanged".

Threaded as a per-tenant knob rather than a literal, matching `candidate_pool_size` / `score_formula`: `fts_rank_scale` joins `_SEARCH_PROFILE_RULES` bounded `(1.0, 20.0)`. Floor is 1.0 — the pre-#687 formula — so a tenant can revert but can't weaken keyword relevance below where it's always been.

## Hot-path cost: nil, verified not assumed

The compiled SQL renders `ts_rank_cd` **18 times before and after** — the scale multiplies an existing expression object rather than duplicating the call. `k` binds once and the SQL text is byte-identical for any `k`, so one plan and one prepared statement regardless of per-tenant override. EXPLAIN ANALYZE over 4,795 candidate rows: **36.1 ms scaled vs 37.1 ms unscaled**.

## Verification

|---|---|---|---|
| `main` | 4117 | 1 | 0 |
| this branch | 4124 | 1 | 0 |

+7 is exactly the added tests, with **no existing test's behaviour changed** — the signal that matters when a change reprices every query.

Each assertion confirmed load-bearing by mutation:

| mutation | result |
|---|---|
| revert the scale to 1.0 | cosine-scale test fails (`0.3 <= 0.0909`) |
| drop the key from the pipeline builder | **pipeline** case fails only |
| remove it from the storage allowlist | **legacy** case fails only |
| widen the bounds floor to 0.0 | bounds test fails (`DID NOT RAISE`) |

## Also corrects #714

That PR merged one commit earlier and describes `k=1` as "today's formula" in `BENCHMARKS.md` and three places in the harness — false the moment this lands. Fixed, and the harness's `--k`/`--fts-weight` defaults now come from `core_api.constants` rather than hand-mirrored literals, since the constant's instruction to re-derive it was otherwise silently unenforceable.

## Deliberately out of scope, recorded for follow-up

- **`setweight` is not an alternative.** Measured: uniform `setweight` is exactly a constant multiply capped at ×10 (ladder ×1/×2/×4/×10), and `ts_rank_cd(weights[])` rejects weights > 1.0 — so **no configuration lands in the target band** (needs ×6). A **title/content** weight split would change *relative* ranking and is a genuine opportunity, needing its own benchmark.
- `_SEARCH_PARAM_KEYS` is *also* missing `candidate_pool_size` and `score_formula`, so the legacy rollback lever already reverts two other ranking features. Own diff.
- `fts_score` renders `ts_rank_cd` twice per row (~14% of the query), but the naive subquery hoist gets flattened by the planner; a real fix needs an optimization barrier and its own plan-shape review.

### DIFF
--- a/BENCHMARKS.md
+++ b/BENCHMARKS.md
@@ -117,8 +117,9 @@ invariant. It A/Bs the keyword half of
 similarity = (1 - FTS_WEIGHT) * vec_sim + FTS_WEIGHT * fts_score
 ```
 
-by rescaling `ts_rank_cd` before saturation (`--k`; `k=1` is today's
-`r/(1+r)` formula), ranking the *whole* corpus in both arms. Reports
+by rescaling `ts_rank_cd` before saturation (`--k`, defaulting to what
+production ships; `k=1` is the **pre-#687** `r/(1+r)` formula), ranking the
+*whole* corpus in both arms. Reports
 recall@5/@10/@20, nDCG@10 and MRR with the same paired bootstrap CI and sign
 test, overall, per category, and — separately — over just the queries where
 `plainto_tsquery` matched anything, since the rest are a mathematical no-op that
@@ -138,7 +139,9 @@ python scripts/benchmark_blend_locomo.py \
 ```
 
 `k=1` asserts a delta of exactly zero on every metric — a self-check that the
-harness isn't manufacturing differences the change cannot cause.
+harness isn't manufacturing differences the change cannot cause. Since #687 shipped
+`FTS_RANK_SCALE`, `k=1` is a *historical* arm: an A/B from it measures the delta
+since before that change, not a delta against what production runs today.
 
 Two limits worth knowing before quoting a number from it. Only ~17% of LoCoMo
 questions lexically match any turn (they are paraphrases), so the overall delta is
--- a/core-api/src/core_api/constants.py
+++ b/core-api/src/core_api/constants.py
@@ -537,6 +537,33 @@ GRAPH_MAX_BOOSTED_MEMORIES = 50  # cap on memories receiving graph boost (preven
 ENTITY_LOOKUP_MAX_MATCHES = 20
 
 FTS_WEIGHT = 0.3  # blend: (1 - FTS_WEIGHT) * vector + FTS_WEIGHT * keyword
+# Scale applied to ts_rank_cd BEFORE the saturating map, putting fts_score on the
+# same scale as cosine so FTS_WEIGHT above bites in effect and not just in name
+# (#687). Division of labour: FTS_WEIGHT is the *preference* between the two
+# signals; this is the *unit conversion* that makes the preference meaningful.
+#
+# The input to the derivation: migration 001 builds ``search_vector`` as a bare
+# ``to_tsvector('english', content)`` with no ``setweight``, so every lexeme is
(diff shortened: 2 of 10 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #723, opened 2026-08-05
fix(search): deliver every scoring knob to the SQL on both search paths

Eldad-Caura:
## The bug

`candidate_pool_size` (A49) and `score_formula` (A50) applied on the **pipeline search path only**.

Three places had to agree, and each was invisible to the other two:

| | builds | sent both keys? |
|---|---|---|
| `resolve_search_profile` | pipeline's nested `search_params` | ✅ |
| `_search_memories_legacy` | its own payload, keys **flat** | ❌ |
| storage `/scored-search` route | rebuilt `search_params` from a hardcoded **allowlist** when no nested dict was present | ❌ |

So on the legacy path both features silently vanished, and the documented `_USE_PIPELINE_SEARCH = False` rollback lever (`memory_service.py`) reverted two ranking features as a side effect. #722 was the third time that allowlist bit — and a test caught it, not review, because the allowlist was the one registration point **no client-boundary assertion can see**: the nesting happens server-side.

## The fix

Not "add the two names to the allowlist". `_search_memories_legacy` was the only non-test flat-key sender left in the repo — every storage test already posts nested `search_params` — so it now sends nested like the pipeline. That makes the allowlist dead code:

- **deleted** `_SEARCH_PARAM_KEYS` and the flat rebuild;
- the route now **requires** `search_params`, via a new `_require_dict` in the shared `routers/_validation.py` (422). Fail-closed rather than defaulting, because `memory_scored_search` reads six of those keys with **indexed** access — so a payload without them is malformed, and 4xx at the edge beats the KeyError 500 it would otherwise raise from inside the session.

Three registration points become two.

### `top_k` is deliberately not nested

Storage reads the SQL LIMIT as `search_params.get("top_k", top_k)`, so a `top_k` in the nested dict **shadows** the overfetched body-level value. Nesting it would have silently narrowed the legacy candidate window from `top_k * SEARCH_OVERFETCH_FACTOR` to `top_k`. Measured, `top_k=3`:

```
[legacy]   body top_k=6  search_params top_k=None  => SQL LIMIT=6   ✅ overfetch intact
```

## Second instance of the same class, found on the way

The legacy payload sent the per-memory boost factors as the **value of `boosted_memory_ids`** and never sent `memory_boost_factor` at all. The SQL gates the entire entity boost on `if boosted_memory_ids and memory_boost_factor` (`postgres_service.py:1342`), so it saw a falsy factor map and **skipped the whole stack** — graph expansion and hop boosts were computed on every legacy search and then discarded at the storage boundary. Measured before the fix:

```
[pipeline] ids={...} factors={UUID(...): 1.5}  => entity_boost_applies=True
[legacy]   ids={...} factors=None              => entity_boost_applies=False
```

`tests/test_entity_retrieval_flag.py` asserted the buggy shape (`boosted_memory_ids == {str(mid): 1.3}`) at the client boundary, which is why it passed while the SQL skipped the boost. Updated to assert both keys.

## Verification

Tests are parametrised over **both** paths and assert against the dict `PostgresService.memory_scored_search` actually receives — not the client boundary, which cannot see this.

Every test was watched **failing on the legacy path for the stated reason** before the fix (3 quiet knobs at defaults and at tuned values, the boost pair, the route guard); pipeline cases pass throughout, since that path was never broken.

- `tests/` — **4144 passed**, 3 skipped, 1 xfailed, 0 xpassed. Baseline measured on a detached `origin/main` checkout: **4140** → +4 is exactly this PR's new cases.
- `core-storage-api/tests/` — **104 passed** (baseline 103 → +1).
- `ruff check` / `ruff format --check` clean at all of CI's exact scopes, plus `--select F821` over `scripts/`.
- `core-api/openapi.broker.json` regenerated: **no diff**.

`/simplify` run: 9 findings applied (reuse the shared `_require_dict` rather than an 8th inline guard; consolidated a rationale that had been duplicated across 4 files; merged 8 integration cases into 4; corrected a comment that overclaimed which keys it covered). 4 skipped as out-of-diff or behaviour-affecting — see "Deliberately not in this PR".

1. **`SEARCH_OVERFETCH_FACTOR` is inert on the pipeline path** — a live defect this PR's `top_k` reasoning uncovered. `resolve_search_profile` puts the un-overfetched `top_k` in `search_params`; `execute_scored_search` overfetches only the body-level value; storage prefers the nested one. Measured: `caller top_k=3, factor=2 → SQL LIMIT=3`, so `PostFilterResults` has none of the headroom its comment says it has. Widening it changes recall on every query, so it wants its own benchmark and review — same treatment #722 got.
2. **Extracting one shared `search_params` builder.** Two hand-built dicts remain, and they have *already* drifted: the A47 `tenant_config.default_search_profile` merge exists only in `resolve_search_profile`, so a per-tenant default never reaches the legacy path. Restructuring both builders is outside this diff and behaviour-affecting.
3. **Adopting the dead `ScoredSearchRequest` schema** (`schemas.py:260`) — it already declares this endpoint's contract, including `search_params` as required, and has zero references. Adopting it is the deeper fix for the guard, but it changes the 422 body shape and touches every field parse in a hot handler.

### DIFF
--- a/core-api/src/core_api/services/memory_service.py
+++ b/core-api/src/core_api/services/memory_service.py
@@ -43,6 +43,7 @@ from core_api.constants import (
     BULK_ENRICHMENT_CONCURRENCY,
     BULK_ENRICHMENT_TOTAL_TIMEOUT_SECONDS,
     BULK_STRONG_EMBED_TIMEOUT_SECONDS,
+    CANDIDATE_POOL_SIZE,
     CHUNKING_THRESHOLD_CHARS,
     CLASSIFIER_DEPRECATED_MEMORY_TYPES,
     CRYSTALLIZER_SHORT_CONTENT_CHARS,
@@ -67,6 +68,7 @@ from core_api.constants import (
     OPENAI_EMBEDDING_MODEL,
     RECALL_BOOST_CAP,
     RECALL_DECAY_WINDOW_DAYS,
+    SCORE_FORMULA,
     SEARCH_OVERFETCH_FACTOR,
     SIMILARITY_BLEND,
 )
@@ -101,6 +103,13 @@ logger = logging.getLogger(__name__)
 # rollback is needed, flip these to False and ship a hotfix — do NOT re-introduce
 # env-level configuration, since that caused prior silent divergence between
 # deployments and the default code path.
+#
+# A ``_USE_PIPELINE_SEARCH = False`` hotfix must be cut from a core-api at or
+# after the commit that moved the legacy search path onto nested
+# ``search_params``. Storage no longer accepts the flat scoring keys an older
+# core-api would send, so flipping the lever on a stale build 422s every search
+# instead of degrading it. Rolling storage back is not the fix — a storage
+# revision predating that commit reads nested params fine.
 _USE_PIPELINE_WRITE = True
 _USE_PIPELINE_SEARCH = True
 
@@ -3412,6 +3421,8 @@ async def _search_memories_legacy(
     _graph_max_hops = sp.get("graph_max_hops", GRAPH_MAX_HOPS)
     _similarity_blend = sp.get("similarity_blend", SIMILARITY_BLEND)
     _fts_rank_scale = sp.get("fts_rank_scale", FTS_RANK_SCALE)
+    _candidate_pool_size = sp.get("candidate_pool_size", CANDIDATE_POOL_SIZE)
+    _score_formula = sp.get("score_formula", SCORE_FORMULA)
 
     # Temporal hint
     temporal_window = _extract_temporal_hint(query)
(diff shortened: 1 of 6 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
Thanks — the observation is **confirmed**, and it was already scoped out of this PR deliberately (see "Deliberately not in this PR" item 1 in the description). I found it while reasoning about why `top_k` must stay out of the legacy nested dict, then measured it rather than inferring it:

```
[pipeline] caller top_k=3 factor=2  body top_k=6  search_params top_k=3     => SQL LIMIT=3  ✗
[legacy]   caller top_k=3 factor=2  body top_k=6  search_params top_k=None  => SQL LIMIT=6  ✓
```

Two things worth adding to the finding:

1. **The #700 FTS reservation does not compensate.** It adds a separately capped `union` branch (`_FTS_ONLY_RESERVED_CANDIDATES`), it does not widen the main `LIMIT` — so the narrow window stands. It also means `test_fts_only_row_survives_a_saturated_candidate_window` passes either way, which is part of why this stayed invisible.
2. **Your option (a) is the right one.** `top_k` is a query limit, not a scoring knob, so it does not belong in `search_params` at all; option (b) leaves the shadowing mechanism in place for the next knob. That is what I have recommended in the handoff.

Not folding it into this PR: it is not introduced here, and widening the candidate pool changes recall on **every** query on the active path — that wants its own `benchmark_blend_locomo.py` run and plan-shape review, the same treatment #722 got. Landing it here would also make this PR's "no behaviour change beyond the two keys reaching the SQL" claim false.

I have amended the `top_k` comment so it no longer reads as if both builders were safe — it now states plainly that the pipeline builder does not avoid this today, and that it is measured and being fixed separately. Comment-only change; no code or test behaviour differs, so I have not re-triggered review for it.

Eldad-Caura:

All checks green (CI, dco, CodeQL, Security Scan). The `claude-review` verdict found **no correctness issues in the changed lines**; its one "Additional Observation" is a pre-existing defect on the *pipeline* path that I had already measured and scoped out of this PR deliberately — answered [here](https://github.com/caura-ai/caura-memclaw/pull/723#issuecomment-5197344918), with the follow-up handed off.

Verified locally: `tests/` 4144 passed (baseline measured on a detached `origin/main` checkout: 4140 → +4 is exactly this PR's new cases), `core-storage-api/tests/` 104 (baseline 103 → +1), ruff clean at all of CI's exact scopes, broker OpenAPI regenerated with no diff. Each new test was watched failing on the legacy path for its stated reason before the fix.

erni-a (approved):

PR #724, opened 2026-08-06
fix(embedding): let bulk callers pass their budget so a slow provider is attributable

Eldad-Caura:
The last of the follow-ups from #721.

## The problem, and why it was structural

Callers enforce their embed deadlines by **cancelling** the call. A `CancelledError` says only "time ran out" — not which layer ate it. The provider call itself was unbounded, so a slow backend surfaced as the caller's anonymous timeout with nothing naming the embed.

At the strong-embed budget that isn't just untidy:

```
BULK_STRONG_EMBED_TIMEOUT_SECONDS   8s   ← the caller's budget
OPENAI_REQUEST_TIMEOUT_SECONDS     25s   ← what ONE provider request may take
```

That budget could **never** be the cap that fired. The module already applies the opposite principle to its own concurrency gate — `EMBEDDING_GATE_TIMEOUT_SECONDS` sits *"deliberately BELOW the callers' deadlines … so the gate could attribute the failure"* — the provider call was simply never brought under it.

## The change

`get_embeddings_batch` takes an optional keyword `budget_s`. Given it, the provider call is bounded at `budget_s - EMBEDDING_BUDGET_MARGIN_S`, so the inner cap fires first and raises `TimeoutError` **from inside the embed layer**. Omitting it keeps the previous behaviour — right for the two auto-chunk call sites that have no deadline of their own to sit under.

One margin constant rather than a subtraction per call site, so the ordering survives an operator retuning the budgets.

This **narrows the window rather than closing it**: a caller can still cancel for its own reasons, so the `BaseException` catch that counts those stays.

## Cost worth naming

Seven test doubles for `get_embeddings_batch` needed updating, and **four tests failed loudly until they did**. A keyword-only argument on a widely-mocked function has real blast radius even when production callers are unaffected. I updated the doubles to mirror the real signature rather than loosening them to `**kwargs`, so the next signature change fails the same way instead of silently passing.

## On the test

The first version of the attribution test **passed with the fix reverted** — both caps raise `TimeoutError`, so the exception type can't tell them apart. It was worthless and I caught it by reverting.

It now asserts on **elapsed time**: an inner budget 25x tighter than the caller's deadline, failing if the caller's is what stopped it.

```
with fix:     0.87s   ✅
without fix:  5.70s   ❌  "took 5.70s, which means the caller's 5s deadline was what stopped it"
```

A second test pins that omitting `budget_s` imposes no cap of its own.

## Context from the load test

Worth recording since it sizes the margin: driving staging TEI from inside the VPC at the gate's own max concurrency (16), a 100-item write's chunked round trips measured **p95 0.394s, worst 0.428s**. A 1s margin is generous against that without meaningfully shortening the useful budget.

That same run answered #721's open load-test question, and refutes the premise of its Medium finding — chunking does **not** cost ~4x hold time, because per-request latency scales with request size:

| concurrency | unchunked p95 | chunked p95 | ratio |
|---|---|---|---|
| 1 | 0.106s | 0.174s | 1.64x |
| 4 | 0.264s | 0.333s | 1.26x |
| 8 | 0.543s | 0.567s | 1.04x |
| **16** | 0.988s | **0.926s** | **0.94x** |

At the concurrency the gate actually permits, chunking is *faster* — smaller requests interleave better across instances. Zero errors at every level. Staging TEI's shape (min 2 / max 3 / concurrency 10) is identical to prod's, so this is a faithful proxy.

## Checks

`ruff check common/ core-api/src/` · `ruff format --check` — clean.
Full `pytest tests/ -m "not benchmark"`: **4108 passed**; 26 failures + 8 errors byte-identical to pristine `main` (no local Postgres). Zero regressions. `uv.lock` untouched.

### DIFF
--- a/common/embedding/_service.py
+++ b/common/embedding/_service.py
@@ -15,6 +15,7 @@ from collections.abc import Awaitable, Callable
 
 from common.embedding._registry import get_embedding_provider
 from common.embedding.constants import (
+    EMBEDDING_BUDGET_MARGIN_S,
     EMBEDDING_GATE_TIMEOUT_SECONDS,
     EMBEDDING_MAX_CONCURRENCY,
     EMBEDDING_RETRY_ATTEMPTS,
@@ -227,7 +228,10 @@ def _resolve_provider_name(tenant_config: object | None) -> str:
 
 
 async def get_embeddings_batch(
-    texts: list[str], tenant_config: object | None = None
+    texts: list[str],
+    tenant_config: object | None = None,
+    *,
+    budget_s: float | None = None,
 ) -> list[list[float]]:
     """Generate embeddings for multiple texts in a single API call.
 
@@ -247,6 +251,21 @@ async def get_embeddings_batch(
     Note this is the SERVICE layer: a request larger than the backend's
     accepted batch is split inside the provider, which is the only layer
     that knows its own cap, so *texts* has no length limit imposed here.
+
+    *budget_s* is the caller's OWN deadline, and passing it is what keeps a
+    slow backend attributable. Callers enforce their budgets by cancelling
+    us, and a cancellation says only "time ran out" — not which layer ate
+    it. Given the budget, this bounds the provider call at
+    ``budget_s - EMBEDDING_BUDGET_MARGIN_S`` so the inner cap fires first
+    and raises ``TimeoutError`` from HERE, naming the embed as the thing
+    that overran. Without it the provider is free to sit for
+    ``OPENAI_REQUEST_TIMEOUT_SECONDS`` (25 s), which already exceeds the 8 s
+    ``BULK_STRONG_EMBED_TIMEOUT_SECONDS`` outright — that budget could never
+    be the one to fire. Same principle the concurrency gate applies; see
+    ``EMBEDDING_GATE_TIMEOUT_SECONDS``.
+
+    Omitting it keeps the previous behaviour, which is right for callers
+    that have no deadline of their own to be under.
 
     Two error shapes are explicitly accounted for:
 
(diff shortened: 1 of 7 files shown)


Eldad-Caura:
@claude please review

Eldad-Caura:
Both fixed. The High is a genuinely good catch — my change made an existing invariant insufficient without touching it.

## High — the margin can pre-empt the gate

Exactly right, and worth spelling out *why* the existing check doesn't cover it: `EMBEDDING_GATE_TIMEOUT_SECONDS < BULK_STRONG_EMBED_TIMEOUT_SECONDS` still holds in the broken configuration. The new cap sits **inside** that window, so "gate < budget" can be true while `budget - margin` lands under the gate:

```
gate 5s  <  strong budget 8s          ← existing check: passes
        budget - margin = 4s          ← margin 4s: the cap fires FIRST
```

Then a gate-saturation event is reported as a generic "embed exceeded its budget" and the gate's dedicated warning never runs — inverting the attribution the gate check exists to guarantee, silently. My change made a sufficient invariant insufficient without touching it.

`_validate_timeout_ordering` now also requires `BULK_STRONG_EMBED_TIMEOUT_SECONDS - EMBEDDING_BUDGET_MARGIN_S > EMBEDDING_GATE_TIMEOUT_SECONDS`. Defaults leave real room: `8 - 1 = 7 > 5`.

Two tests, following the file's existing pattern:
- an invariant assertion alongside `test_opportunistic_embed_budget_sits_between_the_gate_and_the_required_cap`
- `test_startup_rejects_a_margin_that_pre_empts_the_gate`, which sets the margin exactly to the headroom. Its docstring records that the gate check still *passes* there — that's the whole reason it's a separate check. **Verified it fails with the new validator stubbed out.**

## Medium — duplicated `30.0`

Agreed, and it stung a bit: two literals that must stay equal is the same silent-drift class this PR exists to remove, reintroduced by my own copy-paste. Both now read `BULK_EMBEDDING_TIMEOUT_SECONDS`, so the "matches the hot-path bulk embed" comment is true by construction rather than by coincidence.

## Checks

`ruff check common/ core-api/src/` · `ruff format --check` — clean.
Full `pytest tests/ -m "not benchmark"`: **4110 passed**; 26 failures + 8 errors byte-identical to pristine `main` (no local Postgres). Zero regressions. `uv.lock` untouched. Single commit, amended.

Eldad-Caura:
@claude please re-review

Eldad-Caura:
@erni-a pleaes approve

erni-a (approved):

PR #725, opened 2026-08-06
fix(search): let SEARCH_OVERFETCH_FACTOR reach the SQL LIMIT

Eldad-Caura:
Follow-up to #723, which found this. `claude-review` on that PR independently flagged it at **High** severity.

## The bug

`SEARCH_OVERFETCH_FACTOR` was **inert on the pipeline search path** — the active one (`_USE_PIPELINE_SEARCH = True`).

Three pieces, each correct in isolation:

1. `resolve_search_profile` puts the caller's **unmultiplied** `top_k` into `search_params`.
2. `ExecuteScoredSearch` multiplies only a **local** variable, then sends that as the top-level `top_k`.
3. `postgres_service` read the LIMIT as `sp.get("top_k", top_k)` — preferring the nested one.

Measured with a spy, caller `top_k=3`, factor 2:

```
[pipeline] body top_k=6  search_params top_k=3     => SQL LIMIT=3   ✗
[legacy]   body top_k=6  search_params top_k=None  => SQL LIMIT=6   ✓
```

So `PostFilterResults` ran with none of the headroom its own comment says the overfetch provides. The #700 FTS reservation does **not** compensate — it adds a separately capped `union` branch, it does not widen the main `LIMIT`, which is also why `test_fts_only_row_survives_a_saturated_candidate_window` passes either way.

### What it cost users

The candidate window is ordered by `score`, which blends `weight` into `similarity` — so rows **below** the cosine floor can legitimately fill the whole window and then be discarded by the `min_similarity` post-filter, with nothing behind them to backfill.

The new test builds exactly that: 3 rows at cosine 0.28 / weight 1.0 outranking 4 rows at cosine 0.45 / weight 0.0. Before this change the pipeline path returned **0 of 3** requested results. After: **3**.

## The fix, at both ends

- **`ExecuteScoredSearch`** no longer sends `top_k` inside `search_params`. It stays in `ctx.data` — `ClassifyQuery` (`:73`, `:199`) and the `_ALLOWED_OVERRIDES` merge both read it — so this is a projection at the boundary, not a removal at the source.
- **`postgres_service`** takes the LIMIT from the `top_k` parameter only. This is the end that actually makes it correct, and it is skew-safe in the useful direction: an older core-api that still nests a `top_k` now gets the wider window it always meant to ask for.

Either end alone fixes all four deploy-skew combinations. The strip is there so the payload says what we mean, not for skew.

Also restores diagnostic mode's `max(top_k, 50)`, defeated by the same shadowing — though that mode turns out to be **half-wired** and unreachable (`SearchRequest.diagnostic` is declared but never read by a route; nothing writes `diagnostic_results`). Noted in a comment for whoever reconnects it, including that the branch skips `final_top_k` so `TrackRecalls` would bump `recall_count` for all 50 rows.

## Why no ranking benchmark, unlike #722

**This change cannot reorder results — it can only append.** The previous candidates were the top-N by score; the wider set is a superset, and every row that joins ranks *below* all of them, so it cannot displace anything. After filtering and trimming, the old results are retained in the same order and the new ones fill positions the post-filter had emptied.

That is categorically different from #722, which changed the score itself and therefore reordered everything. My handoff had said this wanted a LoCoMo run; on inspection `benchmark_blend_locomo.py` explicitly holds the candidate LIMIT out of scope ("Those are held out on purpose"), so it is the wrong instrument. The evidence here is the append-only argument plus the full suite: **4162 passed** with every existing ranking assertion (freshness ordering, expired-lower, recall-boost decay, blend) unchanged.

`TrackRecalls` reads the **trimmed** `filtered_rows`, so `recall_count` bumps — and the A26 feedback loop — are unaffected on every reachable path.

## Verification

- Three tests, each parametrised over both paths, each watched failing **on the pipeline path only** before the fix, and re-verified as still failing after the `/simplify` restructure: payload delivery, the SQL `LIMIT` actually applied (counts rows storage returned), and the user-visible starvation above. Legacy passes throughout — it was already correct.
- `tests/` **4162 passed**, 3 skipped, 1 xfailed, 0 xpassed. Baseline measured on a detached `origin/main` checkout: **4156** → +6 is exactly this PR's cases.
- `core-storage-api/tests/` 104, unchanged. Ruff clean at all of CI's exact scopes; `F821` over `scripts/` clean; broker OpenAPI regenerated with no diff.

`/simplify`: applied 6 findings — most usefully, a comment #723 left in `_search_memories_legacy` that described the old storage read and called this fix "pending", and the `_top_k` local, which named the LIMIT as if it came from `search_params` like every other `_`-prefixed local in that block. Both were exactly the misreading this change exists to remove.

## Follow-up, deliberately not here

Stripping one key does not make the nested dict scoring-knobs-only: `min_similarity` and `graph_max_hops` are core-api-local too and still ride along (storage reads neither). The generalisation is to project an explicit allowlist shared with the legacy builder — which would also let `test_scoring_knobs_reach_the_sql_on_both_search_paths` assert set *equality* instead of per-key presence. Left out because several tests construct partial `search_params`, so a strict projection needs its own pass.

### DIFF
--- a/core-api/src/core_api/pipeline/steps/search/execute_scored_search.py
+++ b/core-api/src/core_api/pipeline/steps/search/execute_scored_search.py
@@ -62,11 +62,32 @@ class ExecuteScoredSearch:
             top_k = top_k * SEARCH_OVERFETCH_FACTOR
 
         # Build the request payload for the storage client.
+        #
+        # ``top_k`` is stripped: it is the candidate-window LIMIT, not a scoring
+        # knob, and it travels as its own request parameter below. It stays in
+        # ``ctx.data["search_params"]`` because ClassifyQuery and the
+        # ``_ALLOWED_OVERRIDES`` merge above both read it — this is a projection
+        # at the boundary, not a removal at the source.
+        #
+        # Storage ignores a nested ``top_k`` now, so this strip is not what makes
+        # the LIMIT correct; it is here so the payload says what we mean. Removing
+        # only this one key does NOT make the dict scoring-knobs-only, though:
+        # ``min_similarity`` and ``graph_max_hops`` are core-api-local too and
+        # still ride along (storage reads neither). Projecting an explicit
+        # allowlist shared with the legacy builder is the generalisation, and is
+        # deliberately left out of this change.
+        #
+        # The diagnostic branch's widening to 50 was defeated by the same
+        # shadowing and is restored here — but note that mode is half-wired:
+        # ``SearchRequest.diagnostic`` is never read by a route and nothing writes
+        # ``diagnostic_results``, so no caller reaches it. Whoever reconnects it
+        # should know the branch skips ``final_top_k``, so PostFilterResults does
+        # not trim and TrackRecalls would bump ``recall_count`` for all 50 rows.
         search_data: dict = {
             "tenant_id": data["tenant_id"],
             "query": data["query"],
             "embedding": embedding,
-            "search_params": sp,
+            "search_params": {k: v for k, v in sp.items() if k != "top_k"},
             "top_k": top_k,
             "recall_boost_enabled": recall_boost_enabled,
         }
--- a/core-api/src/core_api/services/memory_service.py
+++ b/core-api/src/core_api/services/memory_service.py
@@ -3478,18 +3478,10 @@ async def _search_memories_legacy(
     # allowlist that dropped whatever it did not name. Full rationale sits with
     # the deleted code, on the storage ``/scored-search`` route.
     #
-    # ``top_k`` is deliberately NOT in this dict. Storage reads the SQL LIMIT as
-    # ``search_params.get("top_k", top_k)``, so a ``top_k`` here would shadow
-    # the overfetched body-level value below and narrow the candidate window to
-    # the caller's unmultiplied top_k — dropping the post-filter headroom that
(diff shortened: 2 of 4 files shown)


Eldad-Caura:
@claude

Eldad-Caura:

All checks green (CI, dco, CodeQL, Security Scan) and `claude-review` came back **✅ No issues found** — it independently traced both search paths, confirmed the strip never mutates `ctx.data["search_params"]` (so `ClassifyQuery` and the override merge still read it), confirmed the A49 `candidate_pool_size` branch is unaffected, and confirmed both rolling-deploy directions land on the wider window.

The headline: on the active search path the SQL was limiting to the caller's `top_k`, not the overfetched one, so `PostFilterResults` had no headroom. In the constructed case in the new test that meant **0 of 3 requested results**; it returns 3 after the fix.

No ranking benchmark, deliberately — widening the candidate window can only *append* (every previously returned row still outranks the rows that join it), so unlike #722 this cannot reorder. Reasoning and the measurements are in the description.

`tests/` 4162 passed vs a baseline of 4156 measured on a detached `origin/main` — +6 is exactly this PR's cases. All three new tests were watched failing on the pipeline path only, before the fix and again after the `/simplify` restructure.

erni-a (approved):

PR #726, opened 2026-08-06
fix(embedding): key degraded-provider stats per backend, not per process

Eldad-Caura:
Closes the limitation documented in #721 and carried through #724.

## The masking, one level up

`_EmbeddingStats` was a single module-level object shared by every call path **and every tenant**. One process holds an embedding provider per `(api_key, model, base_url, send_dimensions, truncate_to_dim)` — the registry caches up to 32, explicitly because *"the same api_key can host multiple simultaneous providers, e.g. real OpenAI for one tenant and a local TEI sidecar"*.

So a healthy backend's success cleared a broken backend's streak, and the broken one never reached the reporting threshold.

That is precisely the masking the bulk streak was added to fix, moved up a level:

| | masked by |
|---|---|
| #721 (fixed) | single-embed successes hiding a broken **bulk path** |
| **this** | one tenant's healthy backend hiding **another tenant's** broken one |

## The change

Stats are keyed on the backend — `provider_name` + `model`, both already on the `EmbeddingProvider` protocol. When provider construction fails there is no object to take an identity from, so the resolved provider **name** is the scope; every tenant misconfiguring the same provider then shares one streak, which is right.

The map is bounded at 256 with clear-on-overflow — same shape and same reasoning as `common/ranking`'s `_PERMANENT_LOGGED_MAX`. Ranking reached this conclusion first for its log-once dedup, and its note on why clearing must be per-backend rather than wholesale applies verbatim.

`_run_with_retry` now takes the backend's stats rather than reaching for a global; that's what let the single-embed paths be keyed at all.

## The test that matters

Drives a broken backend and a healthy one **alternately** — three failures, each separated by a success on the *other* backend — and asserts the broken streak survives and still reports.

Verified it **fails when every scope is collapsed to one**, i.e. against the old shared object. That's the point of it.

## A helper that immediately earned itself

Two existing tests changed shape, not intent: they reset `_stats_by_scope` instead of swapping a global (matching the `_reset_misconfig_dedup` idiom already in the file), and assert through `_only_stats()`, which fails loudly if a test touches two backends without meaning to.

It caught one straight away. `test_bulk_success_rearms_the_bulk_report` used a bare stand-in for the "recovered" provider, which now keys as a **different backend** — so the recovery would have landed on a fresh streak and the test would have passed while proving nothing. Its double subclasses the real fake now, and the comment says why that matters here beyond protocol conformance.

## Checks

`ruff check common/ core-api/src/` · `ruff format --check --isolated` — clean.

Full `pytest tests/ -m "not benchmark"`: **4098 passed**, and I compared against a **same-environment** pristine baseline rather than an older number — 39 failures on both, with `comm` confirming **zero introduced**. (The 39 vs the 26 quoted on earlier PRs is a freshly-created local Postgres, identical before and after.) `uv.lock` untouched.

### DIFF
--- a/common/embedding/_service.py
+++ b/common/embedding/_service.py
@@ -38,7 +38,12 @@ class _EmbeddingStats:
     streak structurally cannot see.
     """
 
-    def __init__(self) -> None:
+    def __init__(self, label: str = "unknown") -> None:
+        # Which backend these numbers belong to. Carried on the instance so
+        # every log line this object emits names its own backend — stats can
+        # be keyed perfectly and still be unactionable if the alert reads the
+        # same for all 256 of them.
+        self.label = label
         self.failures = 0
         self.successes = 0
         self.last_failure_time = 0.0
@@ -75,8 +80,8 @@ class _EmbeddingStats:
 
         That second streak exists because the shared one cannot see a
         bulk-only outage. Every path through this module shares one
-        process-wide ``_stats``, so a service that also serves search
-        traffic records a success per query embed, which resets
+        ``_EmbeddingStats`` PER BACKEND, so a service that also serves
+        search traffic records a success per query embed, which resets
         ``consecutive_failures`` and holds it under the threshold no
         matter how many bulk calls fail in between. Not hypothetical:
         prod TEI rejected 100% of bulk embeds for 30+ days on a
@@ -104,13 +109,10 @@ class _EmbeddingStats:
         caller is affected (bulk write at 100 vs re-embed at 50) without
         expanding a traceback.
 
-        LIMITATION worth knowing: the streak is process-wide and NOT keyed
-        on the provider, so in a multi-tenant process one tenant's healthy
-        bulk success clears a different tenant's broken backend — the same
-        masking, one level down. ``common/ranking``'s ``_permanent_scope``
-        is the per-backend answer; applying it here means keying all of
-        ``_stats``, which is a wider change than this. Do not read the
-        streak as per-backend.
+        Both streaks are per BACKEND — see ``_stats_by_scope``. An earlier
+        version of this class was process-wide, which meant one tenant's
+        healthy bulk success cleared a different tenant's broken backend:
+        the same masking this method exists to fix, one level up.
         """
         async with self._lock:
             self.failures += 1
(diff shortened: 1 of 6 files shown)


Eldad-Caura:
@claude please review

Eldad-Caura:
The High is right, and it's the sharpest kind of finding — my fix reproduced the bug it was fixing, for the commonest case.

## What I got wrong

`provider_name` + `model` names a **config type**, not a backend. Two tenants on the same model behind different `api_key`s, or two self-hosted sidecars both serving `bge-m3` at different `base_url`s, produce the identical scope string and mask each other. That is precisely the cross-tenant masking this PR exists to prevent — I'd moved it one step sideways rather than fixing it.

The root cause was that I copied `common/ranking` **halfway**: I took `_permanent_scope`'s shape but not the thing that makes it work. Ranking asks the provider for its own scope, because only the provider can identify its own backend.

## The fix

Providers now declare `dedup_scope`, and `_stats_scope` reads it — exactly as `_permanent_scope` does.

- **OpenAI: per INSTANCE**, via an opaque counter. The registry caches one provider per full config tuple `(api_key, model, base_url, send_dimensions, query_instruction, truncate_to_dim)`, so live instances *are* distinct backends. And no credential or endpoint enters the string — ranking's comment already makes this point: hashing a secret is a fast-hash-of-a-secret, which CodeQL rightly flags.
- **Fake and local: by model**, matching their per-model instance caches.

Accepted cost, deliberate and identical to ranking's: a cache eviction rebuilds the provider with a fresh scope, so a still-broken backend restarts its streak and re-reports once. Errs toward saying too much about a real fault.

## Tests, including one that pins your exact scenario

`test_dedup_scope_distinguishes_backends_sharing_a_model` builds two providers with **identical name and model** at different `base_url`s and asserts distinct scopes — plus that no secret or endpoint leaks into the key. **Verified it fails under the `name|model` scheme I shipped**, which the earlier tests did not catch, because their doubles happened to differ by model.

The cross-backend masking test still fails when scopes are collapsed to one.

## A test the fix exposed as unrealistic

`_bulk_provider` was handing out a **fresh instance per lookup**. Under per-instance scoping that spreads one backend's history across several scopes — and it was never faithful anyway, since the registry caches one provider per backend and "recovery" is that instance answering again, not a new object appearing. It now returns a single instance with a `recovered` switch.

## Checks

`ruff check common/ core-api/src/` · `ruff format --check --isolated` — clean.
Full `pytest tests/ -m "not benchmark"`: 4099 passed, zero failures introduced against a same-environment pristine baseline. Re-synced and re-verified after the branch was rebased onto #725 mid-review. Single commit.

Eldad-Caura:
@claude please re-review

Eldad-Caura:
Right again, and it's the half I'd have been annoyed to discover in an incident: stats keyed perfectly, alert that reads the same for all 256 of them.

Both messages now name their backend:

```
Embedding service degraded [openai model=bge-m3 at http://tei:80/v1]: 3 consecutive failures (total: 3/3)
Bulk embedding failing [openai model=bge-m3 at http://tei:80/v1]: 3 consecutive bulk call(s) failed (requested batch=50), ...
```

## The name is deliberately not the scope

Worth being explicit, because it looks like duplication. `dedup_scope` is an opaque counter precisely so no credential can enter the key — which makes it a correct key and an unreadable alert. So providers expose a separate `backend_label` carrying model and endpoint, and `_stats_label()` reads it alongside `_stats_scope()`.

That's the same split `common/ranking` already makes: keep the URL out of `dedup_scope`, put it in the message. `base_url` is now retained on the OpenAI provider for that purpose only — non-secret, and the single most actionable identifier for a self-hosted backend. The api_key appears in neither.

The label falls back to the provider name when construction failed outright, which is the only identity available then.

## Test

The cross-backend test now also asserts the report names `broken-model` and does **not** mention `healthy-model` — so the two halves are pinned together: the streak survives, and the alert says whose.

## Checks

`ruff check common/ core-api/src/` · `ruff format --check --isolated` — clean.
Full `pytest tests/ -m "not benchmark"`: **4105 passed**, zero failures introduced against the same-environment pristine baseline. Single commit.

Eldad-Caura:
@claude please re-review

Eldad-Caura:
Both applied.

## Low — redundant name resolution

`_resolve_provider_name` was running three times per request on the two single-embed paths. Resolved once at the top of each and reused for `_stats_scope` and `_stats_label`.

## Suggestion — `_STATS_SCOPE_MAX` colliding with the provider cache

This one's the better catch, and I'd missed the implication of the number entirely: **256 is exactly `_registry._OPENAI_CACHE_MAX`**. Equal caps mean a deployment running at the provider cache's own ceiling wipes the stats map *routinely* — resetting streaks for backends that are still live and possibly still broken. That is precisely the case the overflow clear should never touch; the "safe direction to fail" argument only holds for churn past what the registry retains.

Raised to **512**, so every cached backend fits with headroom and the clear is reached only under genuinely pathological churn. The comment now says why it must stay above the registry's cap rather than merely being "bounded".

## Checks

`ruff check common/ core-api/src/` — clean. **4105 passed**, zero failures introduced against the same-environment pristine baseline. Single commit.

Eldad-Caura:
@claude please re-review

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #727, opened 2026-08-06
fix(search): declare the scored-search wire contract once, in common/

Eldad-Caura:
The follow-up #725 named. `search_params` crosses core-api → core-storage-api as a bare dict, and which keys belong in it was written down **nowhere** — which is why the set has now been wrong in both directions:

- keys the SQL **needed** went missing on one path — `candidate_pool_size`, `score_formula` (#723);
- keys the SQL **never reads** travelled anyway, and one of them (`top_k`) collided with a same-named request parameter and silently became the candidate-window LIMIT, defeating the overfetch on the active path (#725).

## The contract, declared once

`common/constants.py` now holds `SQL_SCORING_PARAM_KEYS` (the nine keys `memory_scored_search` reads) and `SQL_SCORING_REQUIRED_KEYS` (the six it reads positionally).

**Why `common/` and not `core_api/constants.py`** — I had it in core-api first, on the reasoning that both consumers live there. That was wrong, and the comment I wrote to justify it ("storage cannot import this") was factually false: `common/constants.py` is documented as *"DB-query constants shared between core-api and core-storage-api"*, and `postgres_service.py:47` already imports twelve names from it, **including ones used inside `memory_scored_search` itself**. What drifts here is the key set against the SQL — a two-service concern — and `common/` is where this repo puts those. It is also not in enterprise's `vendored_files_manifest.json` (only 9 files are), so there is **no re-vendor cost**.

## Both builders project through it

Neither hand-lists the set now, so a core-api-local knob cannot reach the SQL by accident:

- **Pipeline** projects out of a genuine superset — `ctx.data["search_params"]` also carries `top_k`, `min_similarity` and `graph_max_hops`, which `ClassifyQuery` and `PostFilterResults` read and storage does not. Projecting *at the boundary* rather than removing them at the source keeps them available to those steps.
- **Legacy** projects strictly (`_resolved_scoring[k]`, no `if k in`) — it resolves every knob a few lines above, so a key declared and not resolved is a mistake worth a `KeyError`, not a silent omission at the SQL.

## The route now enforces what #723 only described

#723's comment said a payload missing the positionally-read keys is malformed and should 4xx at the edge — but the guard only checked the dict was non-empty, so an *incomplete* one still reached the SQL and surfaced as a `KeyError` 500 from inside the session. It is now a 422 naming exactly which keys are missing.

## Verification

Every new test watched failing for its own reason:

| mutation | result |
|---|---|
| pipeline sends the whole dict | `surplus=['graph_max_hops', 'min_similarity', 'top_k']` |
| drop `candidate_pool_size` from the declaration | `storage reads but core-api does not declare: ['candidate_pool_size']` |
| remove the route's required-key check | partial-payload 422 test fails |

- `tests/` **4170 passed**, 3 skipped, 1 xfailed, 0 xpassed. Baseline measured on a detached `origin/main` checkout: **4166** → +4 is exactly this PR's cases.
- `core-storage-api/tests/` **105** (baseline 104 → +1).
- Ruff clean at all of CI's exact scopes, including `ruff check common/` and the `--isolated` vendored-file check; `F821` over `scripts/` clean; broker OpenAPI regenerated with no diff.

### On the source-scan test

`test_sql_scoring_param_keys_matches_what_storage_reads` is what makes this a *single* registration point rather than a fourth list to maintain: both services import the tuple, but nothing **derives** storage's reads from it, since the SQL builder names each key inline — which is the readable way to write it.

It round-trips the storage source through `ast.unparse` rather than grepping it. That is not incidental: the storage source discusses these keys in prose as well as reading them, and my first version — a regex over the raw text — scored `top_k` as a read, out of the comment explaining why #725 **stopped** reading it. `ast.unparse` drops comments, so what is left is only code.

## What this does not fix

Adding a knob end-to-end still touches `_SEARCH_PROFILE_RULES` (`organization_settings.py:1086`), and forgetting it is the one step no test catches: `validate_search_profile` copies unknown keys through unvalidated and unclamped on the agent-profile path, while a tenant default 422s. Collapsing the remaining lists into one knob table (default + type + range + `sql: bool`) is the real end state, but it restructures both resolvers and the settings-validation table, so it is not this change.

Also still open and separate: `_search_memories_legacy` never merges the A47 `tenant_config.default_search_profile`, so a rollback to the legacy path silently drops tenant-wide search defaults — a divergence in the *values* resolved, not the keys delivered.

### DIFF
--- a/common/constants.py
+++ b/common/constants.py
@@ -398,3 +398,44 @@ SINGLE_VALUE_PREDICATES: frozenset[str] = frozenset(
 # Diverging values would silently produce different archive footprints
 # across the two deployment modes.
 LIFECYCLE_STALE_ARCHIVE_WEIGHT: float = 0.3
+
+
+# ── Scored-search wire contract (#723 / #725) ──
+# The scoring knobs core-api nests under ``search_params`` on the way to
+# core-storage-api's ``/memories/scored-search``, and that
+# ``PostgresService.memory_scored_search`` reads back out. One declaration for
+# both ends: core-api's two search-path builders project through it, so only
+# these keys cross, and storage requires the ones it reads positionally.
+#
+# It lives here, not beside core-api's default VALUES, because what drifts is
+# the key set against the SQL — a two-service concern — and this module is the
+# shared home for exactly that.
+#
+# The set has been wrong in both directions. Keys the SQL needed went missing on
+# one path (``candidate_pool_size`` / ``score_formula``, #723). And keys the SQL
+# never reads travelled anyway, one of which — ``top_k`` — collided with a
+# same-named request parameter and silently became the candidate-window LIMIT,
+# defeating the overfetch on the active path (#725).
+SQL_SCORING_PARAM_KEYS: tuple[str, ...] = (
+    "fts_weight",
+    "freshness_floor",
+    "freshness_decay_days",
+    "recall_boost_cap",
+    "recall_decay_window_days",
+    "similarity_blend",
+    "candidate_pool_size",
+    "score_formula",
+    "fts_rank_scale",
+)
+# The subset read with INDEXED access in the SQL builder, i.e. no server-side
+# default: a payload missing any of these is malformed, and the route rejects it
+# rather than letting it surface as a KeyError 500 from inside the session. The
+# remainder are read with ``sp.get(key, default)`` and may be omitted.
+SQL_SCORING_REQUIRED_KEYS: tuple[str, ...] = (
+    "fts_weight",
+    "freshness_floor",
+    "freshness_decay_days",
+    "recall_boost_cap",
+    "recall_decay_window_days",
+    "similarity_blend",
+)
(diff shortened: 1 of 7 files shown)


Eldad-Caura:
@claude

Eldad-Caura:

All checks green (CI, dco, CodeQL, Security Scan) and `claude-review` returned **✅ No issues found**. It independently verified the four things this change rests on:

- `SQL_SCORING_PARAM_KEYS` / `SQL_SCORING_REQUIRED_KEYS` exactly match what `memory_scored_search` reads, indexed vs `.get`;
- `ResolveSearchProfile` always populates the full nine, so the pipeline projection never silently drops a knob in production;
- the legacy path resolves all nine before indexing, so the strict projection carries no `KeyError` risk;
- the projection filters only the outbound payload — `ctx.data["search_params"]` is untouched for the steps that read it.

It also confirmed the storage-side required-key check closes a real gap: a non-empty-but-partial `search_params` previously reached a `KeyError` 500 instead of a 422.

Verified locally: `tests/` 4170 vs a 4166 baseline measured on a detached `origin/main` (+4 = this PR's cases), `core-storage-api/tests/` 105 vs 104, ruff clean at all of CI's exact scopes including `common/`, broker OpenAPI regenerated with no diff. Each new test was watched failing for its own reason via mutation — listed in the description.

Worth knowing for review: `common/constants.py` is **not** in enterprise's `vendored_files_manifest.json` (only 9 files are), so this carries no re-vendor cost.

erni-a (approved):

PR #728, opened 2026-08-06
refactor(search): one knob table instead of four registration points

Eldad-Caura:
The last item from the #723 → #725 → #727 thread. A search tuning knob had to be registered in **four** places, and every omission was silent in a different way:

| forgotten | symptom |
|---|---|
| a search-path builder | the knob applies on one path only (#723) |
| the wire-contract tuple | a core-api-local key reaches the SQL — `top_k` became the candidate-window LIMIT (#725) |
| `_SEARCH_PROFILE_RULES` | **accepted unvalidated and unclamped** on the agent path, while the tenant-default path rejects it as unknown |

## One table

`common/constants.py` now holds `SEARCH_KNOBS` — per knob: type, bounds, and whether it crosses the wire. Everything else derives:

- `SQL_SCORING_PARAM_KEYS` = knobs flagged `sql`
- `SQL_SCORING_REQUIRED_KEYS` = knobs flagged `sql_required` (storage's indexed reads)
- `_SEARCH_PROFILE_RULES` — **deleted**; both validators read the table directly

Its third `default` slot went with it: `None` for all twelve entries, so the branch consuming it was unreachable. The log line that said *"using default"* now says *"dropping key"*, which is what actually happens.

## Why the table carries no defaults

Three of them aren't constants — `fts_weight` is query-adaptive, `top_k` and `min_similarity` fall back to the caller's request. A `default` column would need a sentinel and the resolvers would stay hand-written regardless, which is precisely the shape of the dead slot just removed. So the resolver is pinned by a test instead.

## Tests: two in, two out

- **`test_resolve_search_profile_emits_every_declared_knob`** — asserts the resolver emits exactly the declared set. This is what makes a forgotten resolver loud, and it is the only *direct* cover for the three knobs that never reach the wire (`top_k`, `min_similarity`, `graph_max_hops`).
- **`test_sql_flags_match_how_storage_reads_each_knob`** — pins `sql` / `sql_required` against storage's actual reads, indexed vs `.get`.

The two source-scan tests #727 added are **deleted, not weakened**: they asserted the tuples against storage, and the tuples are now derived from the flags this test pins. Same property, one place.

Each verified by mutation, with the offending key named every time:

```
declare a knob, skip the resolver  → declared but not resolved: ['brand_new_knob']
flip sql_required → sql            → indexed-but-not-required: ['similarity_blend']
drop a knob from the resolver      → declared but not resolved: ['graph_max_hops']
```

## Verification

- `tests/` **4172 passed**, 3 skipped, 1 xfailed, 0 xpassed — identical to the `origin/main` baseline I measured on a detached checkout, as expected: two tests added, two superseded ones removed.
- `core-storage-api/tests/` 105, unchanged. Ruff clean at all of CI's exact scopes including `common/`; broker OpenAPI regenerated with no diff.

## Not addressed — now stated rather than implied

`SearchProfileUpdate` (`schemas.py:540`) and the `memclaw_tune` MCP signature still enumerate their own subset with their own bounds, and **they have already drifted**: `graph_max_hops` is `(0, 5)` in the table and `ge=0, le=3` there, so the same knob accepts 5 from a tenant default and rejects it from the REST ingress. That predates this change. Deriving those from the table would alter a public request schema and needs a decision on which bound is correct, so I have flagged it in the table's own comment rather than silently picking one.

Also still open: the two resolvers remain separate, and `_search_memories_legacy` never merges the A47 `tenant_config.default_search_profile`, so tenant-wide search defaults reach the pipeline path only. Extracting one shared resolver is the largest remaining win here and would let the new resolver test pin both paths at once.

### DIFF
--- a/common/constants.py
+++ b/common/constants.py
@@ -1,4 +1,13 @@
-"""DB-query constants shared between core-api and core-storage-api."""
+"""Constants shared between core-api and core-storage-api.
+
+Mostly DB-query values both services need to agree on. Also ``SEARCH_KNOBS``
+and the wire contract derived from it: storage reads only the derived key
+tuples, but the knob table is one declaration on purpose — splitting the
+bounds into core-api and the flags into here would put the knob NAME in two
+files, which is the drift it exists to remove.
+"""
+
+from typing import NamedTuple
 
 # ── Memory liveness ──
 # The statuses that mean "this memory is live". Broader than the literal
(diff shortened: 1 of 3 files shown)


Eldad-Caura:
@claude

Eldad-Caura:

All checks green and `claude-review` returned **✅ No issues found**, having cross-checked that all 12 knobs, types and bounds are preserved exactly against the removed `_SEARCH_PROFILE_RULES` and the storage-side reads, and confirmed the `"using default"` → `"dropping key"` log fix describes pre-existing dead code rather than a behaviour change.

Its one Low finding is the `graph_max_hops` divergence this PR already calls out in the table's comment — `(0, 5)` here vs `ge=0, le=3` in `SearchProfileUpdate` and `memclaw_tune`. **I have deliberately not changed it**, because closing it means choosing a ceiling and the two directions are not equivalent:

- lowering the table to 3 narrows what a tenant-wide default may hold, and would start rejecting an existing stored 4 or 5;
- raising the schemas to 5 widens the agent-facing surface additively, but `graph_max_hops` bounds graph-expansion depth, so that is a performance decision.

Two of the three surfaces say 3, which is weak evidence for 3 being intended, but not enough for me to pick on your behalf inside a refactor whose whole claim is that it changes no values. Happy to do either as a one-line follow-up once you say which.

Verified locally: `tests/` 4172 passed — identical to the `origin/main` baseline I measured on a detached checkout, which is the expected result for two tests added and two superseded ones removed. `core-storage-api/tests/` 105 unchanged, ruff clean at all of CI's scopes including `common/`, broker OpenAPI no diff. Each new test watched failing by mutation with the offending key named.

erni-a (approved):

PR #729, opened 2026-08-06
refactor(search): one resolver for both search paths, and A47 reaches legacy

Eldad-Caura:
The last structural item from the #723 → #725 → #727 → #728 thread. Branched off `main` after #728 merged, so **not** a stacked PR.

## The bug

The knob resolution ladder — agent profile → tenant-wide default (A47) → global constant — was written out **twice**: in `ResolveSearchProfile` and inline in `_search_memories_legacy`. The copies had diverged: only the pipeline merged `tenant_config.default_search_profile`.

So a tenant-wide search default applied on the active path alone, and the per-tenant `fts_rank_scale = 1.0` revert **documented for #687** did nothing on the path you would flip to in an emergency. Measured, with the legacy path resolving its own ladder:

```
[tenant_default-legacy]  storage saw {'fts_rank_scale': 6.0, 'candidate_pool_size': 0, 'score_formula': 0}
                         where the tenant asked for {3.0, 25, 1}
```

## The change

`resolve_search_params()` is the single ladder; both callers use it. The legacy path drops from twelve resolved locals to two and projects its wire payload straight off the resolved dict. The pipeline step becomes the `ctx.data` adapter it always was — 72 lines to 27 — and keeps its class, because the runner emits per-step `step_ms` and resolving inline would delete `resolve_search_profile` from the prod latency breakdown.

Because a knob is now resolved in one place, `test_resolve_search_params_emits_every_declared_knob` covers **both** paths at once.

## Scoped to resolution — this does NOT make the paths interchangeable

I want to be explicit, because my first draft of the docstring claimed the rollback lever now "reproduces" ranking and that is **false in the other direction**. `ClassifyQuery` writes per-strategy `search_param_overrides` that `ExecuteScoredSearch` merges over the resolved dict — TEMPORAL and RECENT_CONTEXT both force `freshness_floor` and `freshness_decay_days` — and that is pipeline-only. So on a query with a temporal hint the **pipeline overrides a tenant's freshness knobs where legacy honours them**: a second divergence, opposite in direction to the one fixed here.

The remaining divergences are enumerated in the handoff. The one to know about: **`readable_tenant_ids` is not forwarded to the legacy path at all** — the dispatcher's call omits it and the signature lacks it — so cross-tenant reads silently narrow to the home tenant there. Also legacy-only-missing: the hard `date_range` filter, the `ENTITY_LOOKUP` short-circuit, rerank, STM injection, and `TrackRecalls`' A26 `caller_agent_id` gate (legacy increments unconditionally and synchronously).

## Verification

- `tests/` **4174 passed**, 3 skipped, 1 xfailed, 0 xpassed. Baseline measured on a detached `origin/main` checkout: **4172** → +2 is exactly the new source axis.
- `core-storage-api/tests/` 105, unchanged. Ruff clean at all of CI's exact scopes; broker OpenAPI regenerated with no diff.
- The A47 fix rides the existing both-paths delivery test as a third source axis (`defaults` / `agent` / `tenant_default`) rather than a separate test — which covers all three quiet knobs through the tenant rung instead of the one knob my first version checked. Verified by mutation, above. The knob-set test still fails with the key named when a knob is dropped from the resolver.

`/simplify` found the false equivalence claim above, folded my standalone test into the existing one, and cut a war story I had told at four sites down to one. It also confirmed the module-level hoist of `validate_search_profile` creates no cycle (`organization_settings` imports nothing from `core_api.services`) and costs 1.37 ms of a 539 ms import.

## Follow-ups

- Seven sibling `resolve_config` imports in `memory_service.py` are still function-local; this change's module-level import is the proof they were never cycle-avoidance. Left alone to keep the diff readable.
- Still the next duplication worth collapsing: `SearchProfileUpdate` and the `memclaw_tune` signature enumerate their own knob subset with their own bounds, and `graph_max_hops` has already drifted — `(0, 5)` in `SEARCH_KNOBS` vs `le=3` there.

### DIFF
--- a/core-api/src/core_api/pipeline/steps/search/resolve_search_profile.py
+++ b/core-api/src/core_api/pipeline/steps/search/resolve_search_profile.py
@@ -1,22 +1,15 @@
-"""ResolveSearchProfile — resolve per-agent search profile with fallback to constants."""
+"""ResolveSearchProfile — adapt ``ctx.data`` to the shared knob resolver.
+
+The resolution ladder itself (agent profile → tenant default → constant) lives in
+``memory_service.resolve_search_params``, because the legacy search path needs the
+identical answer.
+"""
 
 from __future__ import annotations
 
-from core_api.constants import (
-    CANDIDATE_POOL_SIZE,
-    FRESHNESS_DECAY_DAYS,
-    FRESHNESS_FLOOR,
-    FTS_RANK_SCALE,
-    GRAPH_MAX_HOPS,
-    MIN_SEARCH_SIMILARITY,
-    RECALL_BOOST_CAP,
-    RECALL_DECAY_WINDOW_DAYS,
-    SCORE_FORMULA,
-    SIMILARITY_BLEND,
-)
 from core_api.pipeline.context import PipelineContext
 from core_api.pipeline.step import StepResult
-from core_api.services.memory_service import _adaptive_fts_weight
+from core_api.services.memory_service import resolve_search_params
 
 
 class ResolveSearchProfile:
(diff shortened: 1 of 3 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #731, opened 2026-08-07
feat(search): index memories.title, at the same weight as content

Eldad-Caura:
> Reworked from the `setweight` A/B split after Eldad's call: **same weight for title and content**. No calibration question remains, and the diff shrank to one migration plus its tests.

## The bug

`memories.title` has **never been in the tsvector**. Migration 001 builds `search_vector` from content alone, so a memory whose title carries the distinguishing words — and enrichment writes a title on *every* enriched memory — cannot be found by FTS on those words at all.

```
query "pool sizing", title "Postgres connection pool sizing",
terms absent from content:      ts_rank_cd = 0.0000   ← unreachable, not ranked low
```

## What 034 does

- tokenises title and content **together**, so every lexeme keeps the weight D it has always had
- widens the trigger to **`UPDATE OF content, title`** — a title-only edit otherwise leaves a stale index, and that is the async-enrich path's *normal* case, not an edge one
- backfills existing rows

## Why one weight, not `setweight` A/B

The split multiplies every content rank by 4 (D=0.1 → B=0.4), which forces `FTS_RANK_SCALE` to be re-derived, and it moves the modal *enriched* row a long way — titles are LLM summaries that reuse the content's salient terms, so most rows match both fields and the weights add:

| | content-only | title-only | title-echo |
|---|---|---|---|
| today | 0.1000 | **0.0000** | 0.1000 |
| title A / content B | 0.4000 | 1.0000 | 1.5429 |
| **this migration** | **0.1000** | **0.1000** | 0.2250 |

Equal weight keeps a content-only match at **exactly its historical 0.1000**, so `FTS_RANK_SCALE` stays 6.0 and no bound, default or calibration moves anywhere in the stack — that is why this touches three files instead of seven.

The only rows that score differently are ones that **could not be found at all** before, plus rows whose title repeats a content term (0.1 → 0.225). That second effect isn't a field preference: it's the boost `ts_rank_cd` gives *any* repeated term, exactly as if the content had said the word twice.

Preferring a title hit over a content hit would be a relevance claim, and nothing here can test it — `benchmark_blend_locomo.py`'s corpus is LoCoMo dialogue turns, which have no titles. So the benchmark this change was expected to need cannot measure it, and the change is scoped to avoid needing it.

## Backfill safety

Committed batches inside `autocommit_block`, the shape migration 012 uses. A single-transaction rewrite would hold the startup advisory lock past Cloud Run's 240 s probe, get SIGKILLed and restart from zero — the non-converging crash loop that took out six writer boots on 2026-06-16. Resumable via the `IS DISTINCT FROM` predicate. Verified over 21k rows, round-tripped down and up, zero left unbackfilled.

A `GENERATED ALWAYS AS (...) STORED` column would make the stale-vector class unrepresentable and produces a byte-identical vector — noted in the migration, not done here, because converting needs DROP + ADD COLUMN: a full table rewrite under ACCESS EXCLUSIVE plus a GIN rebuild. The moment for it is a future migration already rewriting `memories`.

## Verification

- `tests/` **4180** vs a 4176 baseline on `origin/main` — exactly the 4 new tests. `core-storage-api/tests/` 105. Ruff clean at every CI scope.
- Mutation-verified: **downgrading to 033** fails the title test with `scored 0.0` and the reindex test with `stale`.
- Four tests: title is searchable at all; a title match scores **the same** as a content match (this is what pins the no-preference decision); content-only scoring is untouched at 0.1 raw / 0.375 scored; a title-only UPDATE reindexes.
- Three test sites that hand-wrote a content-only `search_vector` are deleted in favour of the trigger — left alone they would have overwritten the title and silently un-tested the change.

### DIFF
--- a/core-storage-api/src/core_storage_api/database/migrations/versions/034_memories_search_vector_title_weighting.py
+++ b/core-storage-api/src/core_storage_api/database/migrations/versions/034_memories_search_vector_title_weighting.py
@@ -1,0 +1,173 @@
+"""Index ``memories.title`` in ``search_vector``, at the same weight as content.
+
+Revision ID: 034
+Revises: 033
+Create Date: 2026-08-07
+
+``title`` has never been in the tsvector. Migration 001's trigger builds it as
+``to_tsvector('english', coalesce(NEW.content, ''))``, so a memory whose title
+carries the distinguishing words — and the LLM enrichment writes a title on every
+enriched memory — **cannot be found by FTS on those words at all** unless the
+content happens to repeat them. Measured against real Postgres, query
+"pool sizing" against title "Postgres connection pool sizing" with those terms
+absent from the content: ``ts_rank_cd`` 0.0000 before, 0.1000 after. Unreachable,
+not merely ranked low. That is the whole reason for this change.
+
+SAME WEIGHT, DELIBERATELY. The two fields are concatenated and tokenised
+together, so every lexeme keeps the weight D it has always had. ``setweight``
+with title A / content B was the other candidate and is NOT what shipped: it
+multiplies every content rank by 4 (D=0.1 -> B=0.4), which forces
+``FTS_RANK_SCALE`` to be re-derived, and it moves the modal ENRICHED row a long
+way — titles are LLM summaries that reuse the content's salient terms, so most
+rows match in both fields, and there the weights add. Measured:
+
+                              content-only   title-only   title-echo
+    today                        0.1000        0.0000       0.1000
+    title A / content B          0.4000        1.0000       1.5429
+    this migration               0.1000        0.1000       0.2250
+
+Keeping one weight means **a content-only match scores exactly what it scored
+before** — 0.1000, so ``FTS_RANK_SCALE`` stays 6.0 and no bound, default or
+calibration moves anywhere in the stack. The only rows whose score changes are
+ones that could not be found at all before, plus rows whose title repeats a
+content term (0.1 -> 0.225). That second effect is not a field-weighting
+decision: it is the ordinary boost ``ts_rank_cd`` gives ANY repeated term, the
+same as if the content itself had said the word twice.
+
+Preferring the field split would be a relevance claim — that a title hit beats a
+content hit — and nothing available here can test it: the LoCoMo benchmark
+corpus is dialogue turns with no titles.
+
+WHY THE TRIGGER ALSO CHANGES ITS FIRE CONDITION. 001 declared
+``BEFORE INSERT OR UPDATE OF content``. With ``title`` in the vector, an update
+that changes only the title would leave a stale ``search_vector`` behind — the
+enrichment path writes the title, and on the async-enrich flow it lands after the
+row already exists, which is the common case rather than an edge one.
+``OF content, title`` covers both.
+
+A ``GENERATED ALWAYS AS (...) STORED`` column would make the stale-vector class
+unrepresentable — no ``UPDATE OF`` list to fall out of sync — and produces a
+byte-identical vector. Not done here: converting needs DROP + ADD COLUMN, a full
+table rewrite under ACCESS EXCLUSIVE plus a GIN rebuild, strictly worse than this
+migration. The moment for it is a future migration already rewriting ``memories``.
+
+BACKFILL. Existing rows keep the content-only vector until next written, so their
+titles stay unsearchable — the bug persists for them, but nothing they could
+already do gets worse, because content lexemes and their weight are unchanged.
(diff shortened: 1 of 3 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
Reworked to equal weight per your call — force-pushed as `cdfbad8`.

Measuring the options first turned out to matter: **concatenating the two fields is strictly better than `setweight` with equal weights**. Both give the same relative behaviour, but concatenation leaves a content-only match at exactly its historical `0.1000`, so `FTS_RANK_SCALE` stays 6.0 — whereas both-at-weight-B would have been 0.4 and still forced the 6.0 → 1.5 re-derivation for no benefit.

That collapsed the diff from seven files to three: the migration and its tests. Gone entirely are the `FTS_RANK_SCALE` change, the storage-side absent-key default, the `SEARCH_KNOBS` floor, and the retraction of the "1.0 reproduces pre-#687" claim — all of which were only needed because the A/B split rescaled content.

One thing I'd flag as the residual behaviour change: a row whose title repeats a content term goes 0.1 → 0.225. That isn't a field preference — it's the ordinary boost `ts_rank_cd` gives any repeated term — but it is not literally nothing, and it will affect most enriched rows since titles summarise content.

Eldad-Caura:
@claude

Eldad-Caura:
Fixed the High finding from the last review, force-pushed as `b3109dc`.

The batching loop was **O(n²)**: `search_vector IS DISTINCT FROM <expr>` has no supporting index — the GIN index serves `@@`, not equality — so every pass re-scanned and re-evaluated a growing prefix of already-rebuilt rows. That is the same blow-up the backfill exists to avoid, just relocated from lock duration into CPU. I'd fixed the lock half earlier (committed batches via `autocommit_block`) and left this half in place.

Now a keyset cursor on the primary key: `AND id > :last ORDER BY id`, with `RETURNING m.id` advancing the cursor in the same statement, so the backfill is one forward pass. The `IS DISTINCT FROM` predicate stays as a *filter* rather than a cursor — a row with no title tokenises identically either side of this migration, so it needs no rewrite and this skips it.

One thing that bit on the way: `exec_driver_sql` with `%(name)s` placeholders fails here — the bind is asyncpg, whose paramstyle is positional `$1`. Switched to `sa.text` with `:last` so SQLAlchemy compiles it per-driver, and `CAST(:last AS uuid)` because the first pass binds NULL and asyncpg needs the type.

Verified: migration round-tripped down and up over 21,633 rows across ~5 batches, zero left unbackfilled. `tests/` 4180 (baseline 4176), storage 105, ruff clean.

Eldad-Caura:
@claude

Eldad-Caura:

All checks green on `b3109dc`, and `claude-review` came back **✅ No issues found** after the rework.

**What this does:** `memories.title` has never been in the tsvector, so a memory whose title carries the distinguishing words — and enrichment writes one on every enriched memory — scored `ts_rank_cd 0.0000`. Unreachable, not ranked low. Migration 034 tokenises title and content together, widens the trigger to `UPDATE OF content, title` (the async-enrich path writes the title *after* the row exists, so this is the normal case, not an edge one), and backfills.

**Equal weight, per your call.** That is what keeps the diff to three files: a content-only match stays at exactly its historical `0.1000`, so `FTS_RANK_SCALE` stays 6.0 and no bound, default or calibration moves. The `setweight` A/B version I first pushed rescaled every content rank by 4 and moved the modal enriched row from 0.375 to ~0.70 — measured, and not something any corpus here could adjudicate, since the LoCoMo benchmark is dialogue turns with no titles.

**Two rounds of review earned their keep.** The first caught that my "scoring-neutral" claim was false for the modal row. The second caught the backfill still being O(n²) — `IS DISTINCT FROM` has no supporting index, so each pass re-scanned a growing prefix. It is a keyset cursor on the primary key now, one forward pass.

**One residual behaviour change, stated plainly:** a row whose title repeats a content term goes 0.1 → 0.225. That is not a field preference — it is the boost `ts_rank_cd` gives any repeated term — but since titles summarise content it will touch most enriched rows.

Verified: round-tripped down and up over 21,633 rows across ~5 batches, zero unbackfilled; `tests/` 4180 vs a 4176 baseline measured on a detached `origin/main`; `core-storage-api/tests/` 105; ruff clean at every CI scope. Downgrading to 033 fails the title tests with `scored 0.0`.

Worth knowing at merge time: this is the first migration in a while that rewrites rows in `memories`, so the storage deploy will run it during startup under the advisory lock. It is batched and resumable specifically so a large table cannot crash-loop the boot.

erni-a (approved):

PR #732, opened 2026-08-08
fix(search): move 034's backfill out of the startup path

Eldad-Caura:
> **Fixes a blocked staging deploy caused by #731.** Staging was never down — Cloud Run kept the previous revision serving — but the new storage revision could not become healthy, so nothing rolled forward.

## What happened

Migration 034 backfilled `memories.search_vector` inside the alembic startup hook. Every instance boots, serialises on the migration advisory lock, and Cloud Run's startup probe kills the container after **240 s**:

```
09:30:46  Running upgrade 033 -> 034 …
09:34:37  Running upgrade 033 -> 034 …
09:39:16  Running upgrade 033 -> 034 …
09:43:06  Running upgrade 033 -> 034 …
09:47:16  Running upgrade 033 -> 034 …          ← never completes
ERROR: Container failed to become healthy. Startup probes timed out after 4m
```

**Committed batches were supposed to make this converge, and they don't.** The keyset cursor lives in process memory, so every restart resumes from the top and re-evaluates the `IS DISTINCT FROM` predicate — one `to_tsvector` per row — across an ever-growing prefix of rows it has already rebuilt. The quadratic cost I removed from *within* a run came back *across* restarts, which is strictly worse: within a run it was merely slow; across restarts it cannot finish.

That is the same failure mode 034's own docstring described and claimed to prevent. The batching addressed lock duration, which was never the binding constraint — wall-clock against a fixed probe was.

## The fix

**034 is now DDL only — 0.28 s.** The backfill moves to `scripts/backfill_034_search_vector.py`, which is what the house convention said in the first place: DDL in the migration, non-DDL backfills in one-off scripts.

The script takes its window bounds from the **primary key index alone, with no predicate**, so choosing the next slice is an index scan rather than a scan that re-evaluates `to_tsvector` over rows already handled. The predicate then applies *inside* the window, so untitled rows — whose vector 034 does not change — cost a comparison and no write. Resumable via `--from-id`, prints its cursor, idempotent.

Verified locally: 924 rows across 74 windows in **0.78 s**; a re-run reports `0 need rebuilding`.

## Why a partial backfill is safe

This is what makes out-of-band viable at all, and it falls directly out of the equal-weight decision on #731: a row not yet rebuilt keeps its content-only vector and **scores exactly as it always did**, because content lexemes and their weight are identical either side of 034. It has simply not gained title search yet. There is no window in which two rows are scored on different scales — which would not have been true of the `setweight` A/B version.

## Deploy sequence

1. Merge → storage deploys, 034 applies instantly, new and updated rows index their titles from the trigger.
2. Run the script once against staging, then prod, when convenient.

Staging's DB is already in exactly this state: the trigger applied (it commits before the backfill), the backfill partial, alembic still at 033 because the migration never completed. Re-running 034 is idempotent — `CREATE OR REPLACE` / `DROP … IF EXISTS`.

## Verification

`tests/` **4181**, `core-storage-api/tests/` 105, ruff clean at every CI scope, `--select F821` over `scripts/` clean.

The tsvector expression now exists in two places, so a test asserts they are **byte-identical** — a divergence would have the script write vectors the trigger would never produce, invisible until a search misses. Mutation-verified: changing `coalesce(m.title, '')` to `coalesce(m.title, ' ')` in the script fails it with both expressions printed.

### DIFF
--- a/core-storage-api/src/core_storage_api/database/migrations/versions/034_memories_search_vector_title_weighting.py
+++ b/core-storage-api/src/core_storage_api/database/migrations/versions/034_memories_search_vector_title_weighting.py
@@ -51,28 +51,47 @@ byte-identical vector. Not done here: converting needs DROP + ADD COLUMN, a full
 table rewrite under ACCESS EXCLUSIVE plus a GIN rebuild, strictly worse than this
 migration. The moment for it is a future migration already rewriting ``memories``.
 
-BACKFILL. Existing rows keep the content-only vector until next written, so their
-titles stay unsearchable — the bug persists for them, but nothing they could
-already do gets worse, because content lexemes and their weight are unchanged.
-It runs in COMMITTED BATCHES inside ``autocommit_block``, the shape 012 uses:
-this holds the startup advisory lock, and a single-transaction rewrite of a
-multi-million-row table would exceed Cloud Run's 240 s startup probe, get
-SIGKILLed, roll back every batch and restart from zero — the non-converging crash
-loop that took out six writer boots on 2026-06-16. Committed batches make it
-resumable, and the ``IS DISTINCT FROM`` predicate skips rows already done.
+BACKFILL IS OUT OF BAND — ``scripts/backfill_034_search_vector.py``, run once
+after deploy. It is NOT in this migration, and that is not a style preference:
+
+An earlier version of 034 backfilled here, in committed batches, and it took
+staging down's deploy path on 2026-08-08. Every instance boots, blocks on the
+migration advisory lock, and Cloud Run's startup probe kills the container after
+240 s; the leader never finishes, so ``Running upgrade 033 -> 034`` appears at
+09:30, 09:34, 09:39, 09:43, 09:47 and never completes. Committed batches were
+supposed to make that converge, and they do NOT: the keyset cursor lives in
+process memory, so each restart resumes from the top and re-evaluates the
+``IS DISTINCT FROM`` predicate — one ``to_tsvector`` per row — across an
+ever-growing prefix of already-rebuilt rows. The quadratic cost moves from
+within a run to across restarts. (Cloud Run kept the previous revision serving
+throughout, so it was a blocked deploy, not an outage.)
+
+The house rule already said this: DDL belongs in the migration, non-DDL backfills
+belong in one-off scripts. What is left here is instant.
+
+``downgrade()`` is symmetric and equally data-free: it restores the content-only
+trigger and rewrites nothing. Reverting rows is the same volume of work as the
+forward pass and would reintroduce the same startup-probe failure. Run
+``scripts/backfill_034_search_vector.py --revert`` after downgrading if the split
+matters — until then, rows the forward backfill already rebuilt keep matching on
+title terms while rows written after the downgrade do not.
+
+A PARTIAL BACKFILL IS SAFE, which is what makes out-of-band viable and is a
+direct consequence of the equal-weight choice above. A row not yet rebuilt keeps
+its content-only vector, and content lexemes and their weight are identical
+either side of this migration — so it scores exactly as it always did and merely
+has not gained title search yet. There is no window in which two rows are scored
+on different scales.
 """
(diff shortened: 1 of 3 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
Both findings fixed, force-pushed as `be770df`.

**Medium — downgrade no longer reverts backfilled rows.** Correct, and the fix has to be out-of-band for the same reason the forward pass does: reverting is the same volume of work, so putting it back in `downgrade()` would reintroduce the startup-probe failure this PR exists to fix. So the script gained `--revert`, and 034's docstring now states the asymmetry explicitly rather than leaving it implicit.

Round-tripped on real data: forward → `--revert` (938 rewritten) → `--dry-run` confirms 938 need rebuilding again → forward (938) → clean.

**Suggestion — the diagnostic counts cost two extra whole-table `to_tsvector` passes.** Also correct; on a multi-million-row table they cost about as much as the rebuild. The upfront count is now `--dry-run` only, and the final one is gone — the loop visits every id, so its own tally is the answer, and rows written mid-run were handled by the trigger anyway.

The drift guard now covers **both** expressions, parametrised over forward and revert. Mutation-verified: changing `_CONTENT_ONLY` to drop its `coalesce` fails the revert case while the forward case still passes.

`tests/` 4182, storage 105, ruff clean at every CI scope, `--select F821` over `scripts/` clean.

Eldad-Caura:
@claude

Eldad-Caura:
Fixed the Low — `af19eb4`.

`_BATCH = 5000` was only ever consumed by `_rebuild()`, which this PR deletes; removed. Worth noting ruff would not have caught it: `F841` covers unused *locals*, not module-level constants, so this was a genuine review catch rather than something a lint pass would have found.

Re-verified after the removal: migration round-trips 033 ↔ 034, the backfill script runs clean against the result, `tests/` 4182, storage 105, ruff clean at every CI scope.

Eldad-Caura:
@claude

Eldad-Caura:
@erni-a pleasr approve

erni-a (approved):

Eldad-Caura:

All checks green on `af19eb4`, and `claude-review` came back **✅ No issues found** after three rounds that each caught something real: a missing downgrade revert path, two redundant whole-table scans, then a dead constant.

**This fixes a deploy I broke.** #731's migration backfilled inside the alembic startup hook; every storage instance boots, serialises on the migration advisory lock, and Cloud Run's probe kills it at 240 s, so `Running upgrade 033 -> 034` repeated at 09:30/09:34/09:39/09:43/09:47 and never finished. Committed batches did not save it — the cursor lives in process memory, so each restart resumed from the top and re-evaluated one `to_tsvector` per row over a growing prefix. The quadratic cost moved from within a run to across restarts, which is worse.

Staging stayed up throughout — Cloud Run kept the previous revision serving 100%, reader healthy, and the loop stopped when the deploy gave up.

**034 is now DDL-only (0.28 s)** and the backfill is `scripts/backfill_034_search_vector.py`, which is what the house convention said all along. Window bounds come from the PK index with no predicate, so picking a slice never re-evaluates `to_tsvector` over finished rows; the predicate applies inside the window, so untitled rows cost a comparison and no write. Resumable, idempotent, `--revert` for after a downgrade.

**Merge order when you're ready:**
1. Merge → storage deploys, 034 applies instantly, new/updated rows index their titles from the trigger.
2. I run `scripts/backfill_034_search_vector.py` against staging, then the four e2e suites.

Staging's DB is already mid-way through exactly this state — trigger applied, backfill partial, alembic still at 033 — and all of it is benign because equal weighting means an un-rebuilt row scores exactly as it always did. Re-running 034 is idempotent.

Verified: `tests/` 4182, `core-storage-api/tests/` 105, ruff clean at every CI scope, `--select F821` over `scripts/` clean, migration round-trips 033 ↔ 034, script round-trips forward → `--revert` (938 rows) → forward → clean.

erni-a (approved):

PR #733, opened 2026-08-08
ci: adopt the org's shared review pipeline

Eldad-Caura:
Replaces this repo's **297-line local copy** of the review pipeline with a **70-line caller**. The skip rules, the org-membership gate, the reviewer scripts and the prompt now come from `caura-ai/.github`, which every other repo in the org already uses. This was the last repo still carrying its own copy.

## Nothing bespoke had to be preserved

Verified byte-identical to the shared defaults before switching:

| local | shared default |
|---|---|
| `MAX_FILES: 30` | `max-files: 30` |
| source-file `jq` filter | **byte-for-byte** the org default |
| `@anthropic-ai/claude-code@2.1.159` | `cli-version` default |
| org-membership gate | `author-gate: auto` → on for public repos |

The local `REVIEW_PROMPT` was an older copy of the shared template, so inheriting the current one is an upgrade rather than a compromise.

## What this fixes

The local file-count call omitted `per_page`:

```
FILE_COUNT=$(gh api "repos/${REPO}/pulls/${PR_NUMBER}/files" --jq 'length')
```

GitHub's `/files` endpoint **defaults to 30**, so `FILE_COUNT` could never exceed 30 and the `-gt 30` oversized-PR guard **has never once fired here**. The shared `pre-checks` passes `per_page=100`. This was a live bug in all seven copies of the pipeline — this is the last one.

## What this gains

- **`claude-capture`** — findings a maintainer declined become review memory on merge
- **`/remember`** — record a curated note for future reviews
- **bot-PR skip** for dependabot and renovate
- the **maintained prompt** — the local copy still carried a `Fix All Prompts` section removed upstream in v1.3.0, which duplicated every per-issue prompt and cost 20–25% of each comment

## `id-token: write`

Granted even though this repo does **not** run the Gemini reviewer. GitHub validates a reusable workflow's permissions graph **statically at load time**, regardless of any callee job's `if:` — so withholding it makes GitHub refuse to load the *entire* workflow, a `startup_failure` that would take the Claude review down too. Harmless here: nothing requests the token, and `mode=gemini-review` is refused outright on a public repository.

## Unchanged

The author gate behaves exactly as before — `auto` derives from repository visibility, this repo is public, so PRs from anyone who isn't a **public** org member are skipped. Note the long-standing operational cost, which has bitten this repo: `GITHUB_TOKEN` resolves only *public* members, so a member whose org membership is private is skipped silently behind a green check.


Eldad-Caura:
@claude

Eldad-Caura:
Verified rather than assumed, since the finding asked for exactly that — and it splits in two.

**The `@claude` check was not dropped. It moved.** In the shared workflow the caller no longer owns that filter; the callee's `claude-retrigger` job does:

```yaml
claude-retrigger:
  if: >-
    github.event_name == 'issue_comment'
    && contains(github.event.comment.body, '@claude')
```

`/remember` and `@gemini` are gated the same way on their own jobs. On a comment carrying no command, every callee job evaluates its `if:` to false:

| job | comment with no command |
|---|---|
| `pre-checks`, `claude-review`, `gemini-review`, `claude-capture` | skip (wrong event) |
| `claude-retrigger`, `gemini-retrigger`, `claude-remember` | skip (command string absent) |

**The cost concern does not hold.** This caller job has no steps — it is a bare `uses:` — so it never checks out the repo or installs the CLI; those happen inside callee jobs that skip. And skipped jobs allocate no runner, so there are no Action minutes and no API spend. What a command-less comment produces is a workflow run whose jobs all skip.

**The reason it is written this way is deliberate**, and documented upstream in `docs/claude-code-review-caller.yml`: naming the review commands in each caller duplicates the callee's trigger conditions across six repos, which is precisely the drift the shared pipeline exists to remove. A caller that hardcodes `@claude` silently stops working the day a new command is added.

**Taking the second half of the suggestion, though** — the finding offered it explicitly, and it is right that a reader of this file alone is left wondering where the check went. Added a comment at the gate stating that the filter lives on the callee's jobs, and why duplicating it here would be wrong. Amended and force-pushed; no functional change.

erni-a (dismissed):

Eldad-Caura:
Confirmed and fixed — good catch, this one was real.

`.github/scripts/claude_pr_review.sh` had exactly two call sites, `claude_code_review.yml:277` and `:378`, and this PR removes both. Verified nothing else in the tree references it (`grep -rl` returns zero hits outside the file itself) and that no workflow globs `.github/scripts/` for linting, so removing it cannot break a lint step. It was the only file in that directory, which now drops out of the tree entirely. Deleted.

Also correcting something on my side: an earlier push of this branch accidentally included `docs/beacon-memories-count-bug-response-2026-07-26.md`, an unrelated untracked file swept in by a careless `git add -A`. It has been removed from the commit and restored to untracked, unmodified. The PR is now exactly two files: the caller rewrite and this deletion.

erni-a (approved):

PR #734, opened 2026-08-09
fix(search): ship the 034 backfill inside the image

Eldad-Caura:
> Prerequisite for running the 034 backfill as a Cloud Run job. Found while setting that job up.

## The problem

`scripts/backfill_034_search_vector.py` landed in the repo's top-level `scripts/`, which **the storage Dockerfile does not copy** — it takes `common/` and `core-storage-api/src/` only:

```
COPY common/ /app/common/
COPY core-storage-api/src/ /app/core-storage-api/src/
```

So a Cloud Run job on the deployed image cannot reach it, and that job is the one way this is meant to run against staging and prod.

## The fix

Moved to `core_storage_api.scripts.backfill_034_search_vector`, next to the existing `backfill_embeddings.py` and `preflight_012.py` — that package already exists for exactly this, and both siblings are `python -m`-invocable one-offs.

Also switched from a `--dsn` argument to `get_engine()`, the way those siblings do. A job then **inherits the VPC connector, service account and secret bindings of the deployment it runs in**, rather than needing a DSN handed to it — which matters because AlloyDB is only reachable over that connector.

```bash
gcloud run jobs create staging-backfill-034 \
  --image <deployed core-storage-api image> \
  --command python \
  --args -m,core_storage_api.scripts.backfill_034_search_vector \
  --vpc-connector <same as service> --service-account <same> --set-secrets <same>
gcloud run jobs execute staging-backfill-034 --wait
```

## Unchanged

Window bounds from the PK index with no predicate; the predicate applied inside the window on the write; `--from-id` to resume; `--dry-run` as the only whole-table count; `--revert` for after a downgrade.

## Verification

Exercised **as a job would invoke it** — `python -m core_storage_api.scripts.backfill_034_search_vector` against a real database, round-tripping forward (966 rows) → `--revert` → forward → `0 need rebuilding`.

`tests/` 4182, `core-storage-api/tests/` 105, ruff clean at every CI scope — note the module is now under `core-storage-api/src/`, so unlike its previous home it is both `check`- and `format`-gated.

The byte-identical drift guard follows it to the new path. It survived `ruff format` rewrapping the file, which would have broken its regex silently had the test not asserted the match itself — the same class of trap as a string-replace mutation that quietly stops applying.


Eldad-Caura:
@claude

Eldad-Caura:
@claude

Eldad-Caura:
@erni-a pleaes approve

erni-a (approved):

Eldad-Caura:
@claude

PR #735, opened 2026-08-09
ci: restore the local review workflow, modernised, after the shared pipeline could not load

Eldad-Caura:
## This repo has not been reviewed since 2026-08-08

caura-memclaw#733 replaced the local review workflow with a caller into the org's shared pipeline. Every run since has failed instantly — **6 of 6, zero jobs created, no logs**:

```
run 31304400763 -> jobs created: 0
run 31304399944 -> jobs created: 0
run 31279132213 -> jobs created: 0
run 31278978009 -> jobs created: 0
run 31278973492 -> jobs created: 0
run 31278893982 -> jobs created: 0
```

[#734](https://github.com/caura-ai/caura-memclaw/pull/734) went unreviewed.

## Why, and why it can't be fixed in place

A component — reusable workflow **or** composite action — in a **private** repository can be used only by **private** repositories. GitHub's docs state it plainly: *"Access is allowed only from private repositories."*

`caura-ai/.github` is private. This repo is public. So `uses: caura-ai/.github/...@v1` can never resolve here, and an unresolvable component fails *before* any job exists — surfacing as "This run likely failed because of a workflow file issue", which reads like malformed YAML. That's why it went unnoticed for a day.

The six repos that migrated earlier are **all private**, which is exactly why the pattern looked proven. This was the first public caller.

Ruled out first: the caller is `actionlint`-clean; its inputs and secrets all exist in `v1`'s `workflow_call` contract; it grants every permission the callee jobs request; and `v1` resolves fine.

### 1. Security: the reviewer was unrestricted, on a public repo

This is the reason a `git revert` would have been the wrong move. The old script invoked the CLI as:

```bash
claude --print --model … --max-budget-usd …
```

No `--bare`, no `--tools`. So the reviewer ran with the **full built-in toolset** and auto-discovered `.mcp.json` and `CLAUDE.md` **from the checked-out PR tree** — in a job holding `ANTHROPIC_API_KEY` and a write `GH_TOKEN`. Now:

```bash
claude --print --model … --bare --tools "Read,Grep,Glob" --max-budget-usd …
```

`--tools` leaves a prompt injection in the diff nothing to act through. `--bare` closes the two paths the allow-list cannot reach — `--tools` names only *built-in* tools, so a PR-supplied `.mcp.json` would otherwise hand the reviewer tools outside it entirely, and `--bare` also skips `CLAUDE.md` auto-discovery so a reviewed tree can't steer its own review. Both flags exist in the pinned `2.1.159`.

### 2. Shorter prompt, shorter verdicts

The `## 🚀 Fix All Prompts` section is gone. It asked for one combined block per severity **repeating every per-issue prompt** — most of what made these comments long. Replaced by the shared prompt's current constraints: one prompt per issue, nothing after the last issue, and no tool-call or JSON-edit syntax inside prompt blocks.

### 3. A dead assignment

`HTTP_CODE=$(gh api …)` was never read — only the exit status mattered, and it held the PR *number*, not an HTTP code. `actionlint` flagged it once the file was linted.

## Also restored

#733 deleted `.github/scripts/claude_pr_review.sh` (141 lines) as well, so the workflow alone would have been broken. Both come back.

The header now records why this copy is deliberate, so it doesn't get "consolidated" a second time. Sharing with the org is still possible — it needs the pipeline in a **public** repo, which is a decision about publishing, not a technical dead end.

## Checks

`actionlint`, `shellcheck`, `bash -n` — all clean.

### DIFF
--- a/.github/scripts/claude_pr_review.sh
+++ b/.github/scripts/claude_pr_review.sh
@@ -1,0 +1,156 @@
+#!/usr/bin/env bash
+#
+# Shared Claude PR review.
+#
+# Fetches a pull request's diff from the GitHub API, reviews it with Claude, and
+# posts the result as a PR comment. BOTH the on-open `claude-review` job and the
+# `@claude` `claude-retrigger` job call this same script, so the two paths are
+# guaranteed identical — historically they diverged, and the on-open path (the
+# claude-code-action agent) never actually received the diff: it tried to fetch
+# it via `gh`, hit the headless permission wall, and the workflow then stamped a
+# meaningless "No issues found". Feeding the diff on stdin to `claude --print`
+# is the mechanism that demonstrably works.
+#
+# Required env:
+#   REPO              owner/name
+#   PR_NUMBER         pull request number
+#   ANTHROPIC_API_KEY Anthropic API key
+#   GH_TOKEN          token with pull-requests:write / issues:write
+#   EXTRA_PROMPT      repo context, prepended to the review prompt
+#   REVIEW_PROMPT     review instructions + output format
+# Optional env:
+#   MODEL             model id (default: claude-sonnet-5)
+#   MAX_BUDGET_USD    per-review spend ceiling, passed to --max-budget-usd (default 10.00)
+set -euo pipefail
+
+MODEL="${MODEL:-claude-sonnet-5}"
+
+# A ceiling so one runaway review cannot bill without bound. The reviewer reads the repo
+# across turns to judge a diff, which is what makes it useful and also what makes an
+# unbounded run possible; sibling repos on the org's shared pipeline have billed $6.11 on a
+# two-file diff. The org-membership gate above stops a fork pull request from triggering a
+# review at all, but it does not bound what a member's large pull request costs.
+#
+# A runaway guard, not a budget target. The CLI checks the ceiling BETWEEN turns, so a run
+# can overshoot it by roughly one turn, and on a single hung turn it never fires at all —
+# the job timeout is the only bound there. Sized above observed spend on purpose: the
+# expensive reviews are the ones that find real defects, so capping near the average would
+# truncate exactly the runs worth paying for.
+MAX_BUDGET_USD="${MAX_BUDGET_USD:-10.00}"
+# Shape, then value. Shape does not require a leading digit, so `.50` is accepted the way the
+# CLI accepts it, while `.`, `1.`, `10,00`, `$10`, `-1` and `1e3` are rejected. Value is
+# checked arithmetically rather than with a second pattern: a zero ceiling is accepted by the
+# CLI and makes every review fail on its first turn — which reads as a broken pipeline rather
+# than a bad setting — and spelling "zero" as a regex means enumerating 0, 00, 0.0, .0, 0.00
+# and .00, where the no-leading-digit forms are the easy ones to miss.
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
Verified all three. **Two are real and now fixed; the Critical is a false positive.**

### ✅ Fixed — `MAX_FILES` enforced nothing (High)

Correct, and worse than it reads. The endpoint returns **30 per page** and `MAX_FILES` is **30**, so `-gt 30` was unreachable for every PR ever opened here — the cap has never fired once. Both calls are now `--paginate`d and summed across pages.

The second call mattered independently: unpaginated, a PR whose first 30 files were docs would be skipped as "no source files" with source changes on page two.

Summing rather than a bare `length` because `--paginate --jq` applies the filter per page and emits one number per page:

```
$ printf '30\n30\n12\n' | awk '{s+=$1} END {print s+0}'
72
```

### ✅ Fixed — synchronize skip comment (Medium)

Correct, and it's about to get worse rather than being long-standing noise. It hasn't fired because the **org-membership check returns first** (line 186) for an author whose membership isn't public, pre-empting the synchronize branch at line 201 — which is why there are **0** `skipped: push to PR branch` comments in this repo's history despite 47 skip comments overall. Now that membership resolves as public, every push would start stacking them.

The synchronize branch emits `notify=false` and the comment step gates on `!= 'false'`, so the other skips — which each explain a one-off decision — still comment.

### ❌ Not applied — the org-membership check does not crash (Critical)

The finding rests on "GitHub Actions bash steps run with `-eo pipefail` by default". That holds only when `shell: bash` is set **explicitly**. Neither step sets it, and there are no job- or workflow-level `defaults`:

```
step='Evaluate skip conditions' shell=<unset>
step='Log skip reason'          shell=<unset>
job-level defaults: None
workflow defaults: None
```

With `shell` unset, GitHub uses `bash -e {0}` — errexit, **no pipefail**. Without pipefail a pipeline's status is the *last* command's, and `awk` exits 0 regardless of what `gh` did. So the assignment succeeds, `MEMBERSHIP_STATUS` is `404`, and the `elif` branch runs exactly as intended.

Production agrees: this repo has posted skip comments naming a non-public member, which is only reachable *through* that branch.

Worth keeping in mind if anyone adds `shell: bash` to this step later — that single line would make the finding correct.

---

Also fixed in the same push: grouping the synchronize branch's three writes under one redirect, since adding `notify` tripped SC2129.

`actionlint`, `shellcheck`, `bash -n` clean.

Eldad-Caura:
@erni-a ready for approval — this one unblocks merges repo-wide.

Automated review has been dead here since #733 merged (6/6 runs failed instantly, zero jobs, no logs), so nothing has been reviewed for about 14 hours. This restores the local workflow and its script.

All checks green: CI, CodeQL, dco, Security Scan, Analyze (python + js/ts). The review verdict's two real findings are fixed — the `MAX_FILES` cap had never fired because the files endpoint pages at 30 and the limit was 30, and the synchronize skip would have started commenting on every push — and the Critical was verified as a false positive (it assumes `-eo pipefail`, which only applies when `shell: bash` is set explicitly; it isn't here). Reasoning and evidence are in the comment above.

Two behaviours already confirmed live on this PR: the restored workflow loaded and reviewed its own diff, and the synchronize skip posted no comment.

erni-a (approved):

PR #736, opened 2026-08-09
refactor(search): derive SearchProfileUpdate from the knob table

Eldad-Caura:
Item #2 of the post-thread list: close the last registration point for a search knob.

## Why

`SearchProfileUpdate` wrote out its nine fields and their `ge`/`le` by hand, duplicating `SEARCH_KNOBS`. That duplication had already produced a real defect — `graph_max_hops` capped at 3 here while the table allowed 5, so a tenant-wide `default_profile` could hold a graph depth no per-agent profile could set. #730 settled the value; this removes the mechanism that produced it.

## What

The model is built by `create_model` over the table, selected by a new **`agent_tunable`** flag on `SearchKnob`. `AGENT_TUNABLE_KEYS` derives from it exactly as the two SQL tuples already do.

That flag also records, in the one place it matters, **why the split is 9 of 12**: `fts_rank_scale`, `candidate_pool_size` and `score_formula` are tenant-level A/B levers held at their global defaults, not per-agent tuning. Previously that was an undocumented omission a reader could plausibly "fix".

## What stays hand-written, and why

`memclaw_tune`'s MCP signature. It declares its knobs as **named parameters**, and that signature *is* the tool schema shipped to clients in `plugin/tools.json` — `create_model` cannot express it. So it becomes the last copy, and a test pins it: both the parameter set **and the human-readable ranges in each description**, because a bound that moves in the table while the description still advertises the old one is a silently wrong instruction to an agent, not merely stale prose.

The previous bounds test would be tautological against a derived model, so it is **retargeted rather than kept**: one cheap assertion that the derivation still yields a usable model (a botched comprehension would produce the wrong field set or drop constraints entirely), and the real guard moved onto the MCP signature.

## Verification

Mutation-verified in both directions:

```
widen graph_max_hops in the table →
  memclaw_tune documents graph_max_hops as '0-3.' but the knob table says (0, 5)
drop agent_tunable from top_k →
  memclaw_tune exposes ['top_k'] differently from the knob table
```

`tests/` **4183**, `core-storage-api/tests/` 105, ruff clean at every CI scope. **`plugin/tools.json` and the broker OpenAPI both regenerate byte-identical** — the derivation reproduces the previous model exactly, so no client-visible change.

### DIFF
--- a/common/constants.py
+++ b/common/constants.py
@@ -451,25 +451,30 @@ class SearchKnob(NamedTuple):
     # payload omitting it is malformed and the route rejects it rather than
     # letting it surface as a KeyError 500 from inside the session.
     sql_required: bool = False
+    # Exposed on the agent-facing tuning surface (``SearchProfileUpdate``, and the
+    # ``memclaw_tune`` MCP tool). False for the A/B knobs, which are held at their
+    # global defaults until the offline comparison validates them and are flipped
+    # per TENANT via ``search.default_profile``, not per agent.
+    agent_tunable: bool = False
 
 
 SEARCH_KNOBS: dict[str, SearchKnob] = {
     # ── core-api-local: resolved here, never sent to storage ──
-    "top_k": SearchKnob(int, (1, 20)),
-    "min_similarity": SearchKnob(float, (0.1, 0.9)),
+    "top_k": SearchKnob(int, (1, 20), agent_tunable=True),
+    "min_similarity": SearchKnob(float, (0.1, 0.9), agent_tunable=True),
     # Ceiling 3, matching the agent-facing ingress (``SearchProfileUpdate`` and
     # the ``memclaw_tune`` MCP signature). It read 5 here until 2026-08-07 while
     # both of those said 3, so a tenant-wide default could hold a depth no agent
     # profile could ever set. Depth drives graph expansion cost, so 3 is the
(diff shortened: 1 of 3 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
Fixed — `99fb7e7`.

`create_model`'s overloads take `**field_definitions: Any | tuple[Any, Any]`, and mypy would not match a comprehension it inferred as `dict[Any, tuple[Any, None]]`. Hoisted the fields into an explicitly annotated `_PROFILE_FIELDS: dict[str, Any]`, which matches. No errors at the annotation sites either — `SearchProfileUpdate` as `type[BaseModel]` is fine in `body: SearchProfileUpdate` and at the `mcp_server` construction site.

**My miss: I wasn't running mypy locally at all.** Ruff and the test suites, yes; mypy no — so this class of failure could only surface in CI. CI runs `cd core-api && uv run mypy src/` and the same for core-storage-api; both are now clean locally (190 and 76 files), and I've added them to my pre-push routine.

Re-verified: mypy both projects clean, `tests/` 4183, storage 105, ruff clean at every CI scope.

Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #737, opened 2026-08-09
ci(review): give this repo's reviewer the shared code-review memory

Eldad-Caura:
Six of the org's seven repos read learned guidance before reviewing and write a maintainer's declined findings back afterwards, so a finding judged wrong once stops being raised again. This repo was the exception — and not by decision: it keeps a **local copy** of the pipeline (public repo, private `.github`, and a component in a private repository can only be used by private repositories), and that copy predates the memory feature entirely.

The practical cost: the reviewer here has been re-raising things already settled, on the OSS repo where outside contributors see it.

## Recall

Inlined into `claude_pr_review.sh` rather than sourced from a `review_lib.sh` — this copy is deliberately two standalone files, and a library holding one function is structure without a second caller.

Same recall the shared pipeline runs: the query is built from the changed paths so guidance is relevant to *this* diff, recalled bullets are entity-escaped so one cannot close the `<review_guidance>` wrapper and escape into the prompt body, and the wrapper frames them as data. The instructional half of that defence is soft, and the residual risk is accepted for a curated internal fleet — worth revisiting if it ever ingests external content.

## Capture

New script plus a job on merge. `merged == true`, not merely `closed`: a PR closed *without* merging carries no accepted judgement, and learning from an abandoned branch teaches the reviewer a decision nobody stood behind.

The extractor runs with **every tool removed** and `--bare`, because a comment thread on a public repo is attacker-influenced input exactly as a diff is.

Notes are tagged `· claude ·` even though this repo runs one reviewer — the **fleet is shared** with repos that run two, and an untagged note from here would be the only one no per-reviewer filter could account for. Attribution is a constant here rather than inferred from reviewer markers, which is what the shared pipeline needs and this doesn't.

## Trigger changes

`closed` had to join the trigger list for capture to see a merge, so `pre-checks` now excludes it explicitly — otherwise merging a PR would start a fresh review of it. `claude-review` is additionally gated on `pre-checks` succeeding, so it can't fire on a merge either way.

## It lands DARK, deliberately

`MEMCLAW_AGENTS_KEY` is an **org** secret with **`private`** visibility, and a public repo cannot read it. So today it resolves to `''` here:

- recall returns immediately, leaving the review byte-for-byte unchanged
- capture logs `capture is dark (no MEMCLAW_AGENTS_KEY) — skipping` and exits 0

Turning it on is **a repo-level secret and no code change**. Shipping it dark makes the config step independently verifiable rather than bundling a behaviour change with a credential change.

> [!NOTE]
> Adding the key here means a live MemClaw write credential on a public repo. Fork PRs never receive secrets, so a fork cannot reach it; same-repo PRs mean push access already. That's the same posture `ANTHROPIC_API_KEY` has here today.

`actionlint`, `shellcheck`, `bash -n` clean. Verified the dark path actually logs rather than failing silently.

### DIFF
--- a/.github/scripts/claude_pr_capture.sh
+++ b/.github/scripts/claude_pr_capture.sh
@@ -1,0 +1,204 @@
+#!/usr/bin/env bash
+#
+# Capture DECLINED review findings from a just-merged pull request into the shared code-review
+# MemClaw fleet, so future reviews stop re-raising findings a maintainer already judged wrong.
+# The write half of the loop; recall in claude_pr_review.sh is the read half.
+#
+# WHY THIS FILE EXISTS HERE RATHER THAN BEING INHERITED
+#
+# This repo is PUBLIC and caura-ai/.github is PRIVATE, and GitHub allows a component in a private
+# repository to be used only by private repositories — so this repo cannot consume the org's
+# shared pipeline and keeps its own copy. See the workflow header. The fleet is shared even
+# though the code is not: notes written here are recalled by the other six repos and vice versa.
+#
+# DARK AND SILENT without MEMCLAW_AGENTS_KEY. That is the state to expect at the moment this
+# lands: the org secret has `private` visibility, which a PUBLIC repo cannot read, so until a
+# repo-level secret exists this exits immediately having done nothing.
+#
+# BEST-EFFORT THROUGHOUT: a post-merge job must never red-X an already-merged pull request, so
+# every failure path warns and exits 0.
+#
+# Injection threat model, bounded: the comment thread is fed to the CLI as stdin DATA under a
+# fixed prompt, with every tool removed. The notes it extracts are later recalled inside the
+# review script's <review_guidance> wrapper, which entity-escapes them and frames them as data.
+# Only a maintainer-merged pull request triggers this at all.
+#
+# Required env:
+#   REPO              owner/name
+#   PR_NUMBER         pull request number
+#   ANTHROPIC_API_KEY Anthropic API key
+#   GH_TOKEN          token with issues:read and pull-requests:read
+# Optional env:
+#   MEMCLAW_AGENTS_KEY   internal-agents tenant key (empty => dark no-op)
+#   MEMCLAW_API_URL      default https://memclaw.net
+#   CODE_REVIEW_FLEET_ID default code-review
+#   MODEL                default claude-sonnet-5
+#   MAX_BUDGET_USD       per-invocation ceiling (default 2.00)
+set -euo pipefail
+
+MODEL="${MODEL:-claude-sonnet-5}"
+MAX_NOTES=5
+MAX_NOTE_CHARS=600
+MAX_THREAD_CHARS=60000
+# This repo runs ONE reviewer, so attribution is a constant rather than something the extractor
+# has to infer from `<!-- caura-reviewer: ... -->` markers as the shared pipeline does. The tag
+# is still written, because the fleet is shared with repos that DO run both and an untagged note
+# from here would be the only one no per-reviewer filter could account for.
+REVIEWER="claude"
+
+if [ -z "${MEMCLAW_AGENTS_KEY:-}" ]; then
+  echo "::notice::Declined-finding capture is dark (no MEMCLAW_AGENTS_KEY) — skipping"
+  exit 0
+fi
+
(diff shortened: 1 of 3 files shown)


Eldad-Caura:
Both High findings applied — the first is the one that mattered, and the second I had already reasoned about upstream and then failed to carry across.

### ✅ Maintainer authority is now enforced, not inferred

The strongest finding in this review. On a public repo anyone can comment, and the extractor had only a login and a tone to judge authority by — so a stranger writing a confident *"that finding is wrong because X"* could have it captured as a maintainer's decision and written into a fleet **every repo in the org recalls from**. That is a write path into shared memory from an untrusted commenter.

The thread now carries `author_association` per comment and the prompt may treat only `[OWNER]`, `[MEMBER]` or `[COLLABORATOR]` as a decline, explicitly instructed not to infer authority from tone or confidence.

Verified the values against this repo rather than assuming them:

```
Eldad-Caura=MEMBER   github-actions[bot]=CONTRIBUTOR
```

That detail matters: the reviewer bot is `CONTRIBUTOR`, so its verdicts stay in the thread as the findings *being* declined while never counting as the decline itself.

### ✅ Truncation now keeps the tail

Correct, and confirmed: `gh api` returns comments oldest-first, so keeping the first 60000 characters dropped the end — where a decline, being a *reply*, always is.

Slightly galling: I raised this exact point upstream when the same file grew Gemini verdicts (*"truncation takes the tail, so what gets lost is the late decline rather than the verdict it answers"*) and then head-truncated in the port anyway.

Also moved the "was this reviewed at all" gate to run on the **full** thread before truncation, so a long discussion cannot read as a PR nobody reviewed.

### ❌ Not changed — `--tools ""` is the CLI's documented empty allow-list

The concern is right that a silent failure here would be serious, so it is worth answering with evidence rather than assurance:

```
$ claude --help
  --tools <tools...>   Specify the list of available tools from the built-in set.
                       Use "" to disable all tools, "default" to use all tools, ...
```

`""` is documented as *disable all*, not *unrestricted*. I have put that quote in the file, because the reasoning for trusting it should not have to be rediscovered.

Kept the deliberate absence of a run-time probe: this job exits 0 on every failure, so a probe could only downgrade capture to the no-op it already becomes, and the review path — which runs on every PR and fails loudly — is where a changed CLI surfaces first. The comment now says to re-read it when bumping the pin.

---

`actionlint`, `shellcheck`, `bash -n` clean.

Eldad-Caura:
@claude

Eldad-Caura:
Both applied, and the first is a better catch than its Medium suggests.

### ✅ COLLABORATOR narrowed to OWNER/MEMBER

Confirmed against this repo's own workflow, which excludes it in as many words:

```
# ``COLLABORATOR`` (non-org members granted direct repo write
# access via Settings → Collaborators) is INTENTIONALLY
# excluded — the policy is "members of our org only"
```

So capture was **more permissive than the review trigger, for a more durable effect** — the retrigger spends one review, a captured note becomes standing guidance recalled by every repo sharing the fleet. The bar has to be at least as high as that gate, not lower. Narrowed to `[OWNER]` / `[MEMBER]`, with a comment recording that widening it means widening the gate first.

### ✅ jq guard added — though the named trigger does not reproduce

Worth splitting, because half the finding checks out and half does not.

The mechanism you inferred is half right. Bash substring expansion **is** byte-based in a C locale, so `${NOTE:0:600}` can split a multi-byte character:

```
$ LC_ALL=C bash -c 'S="a—b"; printf %q "${S:0:2}"'
$'a\342'          # half an em-dash
```

But jq **accepts** that byte sequence and exits 0:

```
$ jq -n --arg c "$(printf 'decline \342\200')" '{content:$c}' ; echo $?
0
```

So the specific failure does not occur with this jq. The finding still stands on its other leg, which is the one I have acted on: the file promises *every failure path warns and exits 0*, and a bare command substitution under `set -e` inside that loop would abort the script and take the remaining notes with it. A promise the code cannot keep is worth closing whether or not today's jq is what breaks it. The comment records both the reasoning and the negative result, so nobody re-derives it.

Verified the guard skips one note and lets the loop finish rather than aborting.

---

`actionlint`, `shellcheck`, `bash -n` clean.

Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #739, opened 2026-08-10
perf(search): halve ts_rank_cd renders in the scored-search SQL

Eldad-Caura:
## What

`fts_score` was `x / (1 + x)`, where `x` is the `ts_rank_cd(search_vector, plainto_tsquery(...))`
expression — not a bound value. SQLAlchemy inlines an expression at every site that names it, so
writing it twice made Postgres evaluate the ranking function twice.

The algebraically identical form names it once:

```
x / (1 + x)  ==  1 - 1/(1 + x)
```

Compiling the real `memory_scored_search` statement: **`ts_rank_cd(` 18 renders → 9**, and
`plainto_tsquery(` 27 → 18.

## Measured

On a 31,446-memory corpus with genuine multi-term matches, scoring + ordering + limiting, medians
of 7 runs per query:

| query | matching rows | two-render | one-render | delta |
|---|---|---|---|---|
| `case never use` | 11,505 | 92.0 ms | 56.4 ms | **38.7%** |
| `case never project` | 7,222 | 59.6 ms | 37.2 ms | **37.7%** |
| `case never must` | 5,581 | 47.2 ms | 29.9 ms | **36.6%** |
| `case never convent` | 4,016 | 34.6 ms | 22.2 ms | **36.0%** |
| `case never requir` | 1,875 | 19.1 ms | 13.0 ms | **31.8%** |

Read that as the gain on the **FTS scoring component**. The full search also pays six pgvector
distance computations per row, so end-to-end is smaller. Prior attempts to measure this locally
were meaningless: the local corpus's realistic multi-term queries match zero rows, and its only
matching single term returns 3,397 rows that all score the identical 0.375.

## What this does NOT do

It does not get to one render, and the code no longer claims it does. `similarity` names
`fts_score` in both branches of its CASE, `score` inlines `similarity`, and both are output columns
of the scored CTE — four live references remain, times the UNION'd reserved branch.

Getting to 1 means projecting `similarity`/`vec_sim` once in an inner derived table, and it **does**
need a real optimisation barrier: with references still live, Postgres flattens a plain subquery and
re-substitutes. That is a separate change to the hot path with its own plan-shape review. (An
earlier note in this thread claimed no barrier was ever needed — true of the factor in isolation,
false of the statement it composes into.)

## Not bit-identical

The two forms disagree by at most one ULP — measured `max |a - b| = 5.55e-17` over every matching
row of that corpus, not the `0.000e+00` recorded earlier. The old zero was measured where every
matching row scored the same single value, so it had no rank variety to expose the difference.
5.55e-17 is ~4 orders of magnitude below the tightest tolerance anything here asserts (`1e-12`) and
far below any score gap that could reorder results.

## Tests

Two levels, because the factor being clean is not the same as the query being clean:

- the saturating factor in isolation renders `ts_rank_cd` once, plus a control proving the naive
  form really does render twice (so the count-of-one assertion can't pass vacuously);
- **the assembled statement's render counts**, captured on the way to `session.execute` and
  compiled — no DB. This is what catches someone adding a fifth reference to `fts_score`. Reverting
  the fix fails it with `renders ts_rank_cd 18 times, expected 9`.

`_rank_and_score` in the integration suite carried its own copy of the old expression while its
docstring claimed to use the production one; it now matches.

## Verification

- `core-storage-api/tests/`: **114 passed** (105 on `origin/main` + 9 new)
- `tests/`: **4183 passed, 3 skipped, 1 xfailed** — identical to the `a5a55f2` baseline
- mypy clean on both projects; ruff check/format clean at all four CI scopes
- `plugin/tools.json` and the broker OpenAPI regenerate byte-identical

### DIFF
--- a/core-storage-api/src/core_storage_api/services/postgres_service.py
+++ b/core-storage-api/src/core_storage_api/services/postgres_service.py
@@ -251,6 +251,47 @@ def _entity_uf_union(parent: dict[UUID, UUID], rank: dict[UUID, int], a: UUID, b
 _FTS_ONLY_RESERVED_CANDIDATES = 3
 
 
+def _saturate_rank(scaled_rank: Any) -> Any:
+    """Map a scaled ``ts_rank_cd`` onto ``[0, 1)``, naming the rank ONCE.
+
+    ``x / (1 + x)`` is the obvious way to write this and is what shipped until
+    now. It is also a performance bug: ``x`` is not a bound value but the
+    ``ts_rank_cd(search_vector, plainto_tsquery(...))`` expression itself, and
+    SQLAlchemy inlines an expression at every site that names it, so writing it
+    twice makes Postgres evaluate the ranking function twice. The
+    algebraically identical
+
+        x / (1 + x)  ==  1 - 1/(1 + x)
+
+    names ``x`` once, halving the renders contributed by this factor.
+
+    WHAT THIS DOES AND DOES NOT BUY. Compiling the real ``memory_scored_search``
+    statement, ``ts_rank_cd(`` goes from **18 renders to 9** (and
+    ``plainto_tsquery(`` 27 -> 18). NOT to 1: ``similarity`` names ``fts_score``
+    in both branches of its CASE, ``score`` inlines ``similarity``, and both are
+    output columns of the scored CTE — so four live references remain, times the
+    UNION'd reserved branch. Getting to 1 means projecting ``similarity`` and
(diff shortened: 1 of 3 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@erni-a ready for approval — review came back clean (no findings) and all checks are green: CI, dco, CodeQL (python + js/ts), Security Scan, core-api version bake, pre-checks. Single commit, DCO signed off.

erni-a (approved):

PR #745, opened 2026-08-10
ci(review): allow a manual capture backfill, mirroring the shared pipeline's caller

Eldad-Caura:
`claude-capture` only fires on a merge, so there was no way to exercise it on demand. That matters now for two reasons.

## 1. It makes the credential verifiable

`MEMCLAW_AGENTS_KEY` isn't set on this repo yet, so recall and capture are both dark. **Recall** is verifiable the moment it's added — it runs on every review and logs whether it injected anything. **Capture** wasn't: confirming it needed a real merge that happened to carry a declined finding, which isn't something you can arrange in order to check a secret.

The shared pipeline's caller has had this dispatch for exactly that reason; this brings the local copy in line.

## 2. Backfill

Capture didn't exist here until #737, so every decline in this repo's history is unrecorded. #735, #737 and #744 each carry findings a maintainer declined **with a stated reason** — precisely what the extractor looks for. One dispatch per PR recovers them.

## Details

`required: true` on the input, because the only failure mode of an empty value is a confusing one: the script would request `issues//comments`, take the 404, warn, and exit 0 looking exactly like a PR with nothing to capture.

**Only `claude-capture` answers a dispatch**, and that falls out of the existing gates rather than needing a new one — `pre-checks` is gated on `pull_request`, `claude-review` needs it to have succeeded, and `claude-retrigger` needs `issue_comment`. So a dispatch can't start a review, which would otherwise be an expensive surprise given a backfill names an already-merged PR.

`PR_NUMBER` reads `github.event.pull_request.number || inputs.pr_number`, in that order, so a merge never picks up a stale dispatch input.

`actionlint` clean.

### DIFF
--- a/.github/workflows/claude_code_review.yml
+++ b/.github/workflows/claude_code_review.yml
@@ -103,6 +103,20 @@ on:
     types: [opened, reopened, synchronize, closed]
   issue_comment:
     types: [created]
+  # Manual backfill for claude-capture, mirroring the shared pipeline's caller. Two uses: it makes
+  # capture verifiable the moment MEMCLAW_AGENTS_KEY is added — the on-merge path cannot be tested
+  # on demand, since it needs a real merge that happens to carry a declined finding — and it can
+  # extract declines from pull requests that merged BEFORE capture existed here.
+  #
+  # `required: true` because the only failure mode of an empty value is a confusing one: the script
+  # would request `issues//comments`, get a 404, warn, and exit 0 looking like a repo with nothing
+  # to capture.
+  workflow_dispatch:
+    inputs:
+      pr_number:
+        description: "Merged PR number to capture declined findings from"
+        required: true
+        type: string
 
 permissions:
   contents: read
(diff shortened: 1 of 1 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

Eldad-Caura:
@claude

Eldad-Caura:
Declining this finding. The "merged PR only" invariant is about the AUTOMATIC path, not the manual one, and I do not think it should be extended to `workflow_dispatch`.

The reason the on-merge job checks `merged == true` is that it fires by itself on every closed pull request, so without the check it would capture from abandoned branches nobody decided anything on. A `workflow_dispatch` is a different trust context: it requires repo write access, and a maintainer types a specific PR number into it deliberately. Someone with write access who points the backfill at an unsettled discussion has made a choice, in the same way that someone typing `/remember` can record something wrong — the pipeline does not try to stop that either.

Adding the check would also cost an API call on every backfill to enforce something the caller already knows, and would block pointing the backfill at a thread on purpose to check that capture works at all, which is exactly what it was added for.

Stating the counter-argument fairly, because it is not baseless: a note written from an unmerged discussion is guidance drawn from a decision that was never ratified, and that note is recalled by every repo sharing the fleet. I am accepting that risk because the manual path is permission-gated and deliberate, not because the risk is zero. If backfill ever becomes something scheduled or automated rather than typed by hand, this finding becomes correct and the check should be added then.

PR #746, opened 2026-08-10
chore(deps): drop the uv versioning-strategy that measurement showed is inert

Eldad-Caura:
## What

Reverts the `versioning-strategy: increase-if-necessary` added to the `uv` entry in #740, and moves the durable rule to the file header.

## Why — #740 was falsified by its own first real run

#740 shipped with a stated test: the setting should kill the eight individual widening PRs and leave lock-only PRs behind. The test ran, and it failed.

The timing matters, because it is easy to read the evidence backwards:

| when | what |
|---|---|
| 01:41:47Z | scheduled Monday `uv` run produces #738 — **old config** |
| 09:01:46Z | #740 merges |
| 09:34:55Z | config-change re-run closes #738, opens #743 — **new config** |

So the scheduled run everyone was waiting on tested the config *before* the change. The only run that exercised `increase-if-necessary` is the one that produced #743, and #743's manifest edits are identical in kind to #738's:

- `fastapi>=0.140.0,<1` → `>=0.141.1,<1` — 0.141.1 was already allowed
- same shape for `alembic`, `cachetools`, `ruff`, `ddtrace`, `google-cloud-pubsub`
- 56 requirement lines across the four services
- the eight PRs it was meant to retire (#645, #646, #650, #691, #692-#695) were untouched, last updated 07-27 / 08-04

GitHub's options reference *does* list `uv` among the ecosystems supporting `versioning-strategy` ("Supported by: `bundler`, `cargo`, `composer`, `helm`, `mix`, `npm`, `pip`, `pub`, and `uv`"), so this is a gap in dependabot-core, not a mistake in the config. Reverting to the default restores honest behaviour: the file no longer claims to fix something it does not fix.

## Why the rule moved to the header

The most valuable line in the old comment was the warning that the `pip` entry must never take this option — and it sat inside the `uv` entry, 36 lines from the edit it exists to prevent. On a lock-less entry the setting is worse than inert: dependabot-core skips the manifest edit when the new version already satisfies the range and no lockfile exists, so there is nothing left to change and **no PR is filed at all**. That is a silent loss of update coverage, and the warning now lives where both entries can see it.

The header also records why `lockfile-only` is not the answer here: it is honoured, but it drops any update needing a manifest change, most specs in the uv services are capped, and this repo is public — these `>=` floors are part of the contract with OSS consumers, so a raised floor is not purely churn the way it is inside a private service.

Constraint-widening PRs are now documented as a known, accepted cost rather than a solved problem.

## Notes

- Config is 7 lines **shorter** than before #740; the point-in-time census numbers and the nine PR references are deliberately kept in this PR body rather than the config, where nothing fails when they go stale.
- No behaviour change beyond restoring the Dependabot default. All six entries and all ten groups verified intact after the edit.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -12,6 +12,21 @@
 # The uv services share ONE multi-directory entry for the same reason: four separate
 # entries emitted four near-identical group PRs every week.
 # Mirrors the caura-memclaw-enterprise grouping strategy.
+#
+# NO entry here sets `versioning-strategy`, and that applies to all of them:
+#   * `increase-if-necessary` is IGNORED for `uv` — verified against a live run on
+#     2026-08-10 (#740, reverted), which still raised floors for releases the ranges
+#     already allowed. GitHub documents `uv` as supporting it, so this is a
+#     dependabot-core gap; re-verify before believing it is fixed.
+#   * On an entry with NO committed lock — the `pip` entry below — it is worse than
+#     inert: dependabot-core skips the manifest edit when the new version already
+#     satisfies the range and no lockfile exists, leaving nothing to change, so those
+#     updates stop being reported at all.
+#   * `lockfile-only` IS honoured, but it drops any update needing a manifest change.
+#     Most specs in the uv services are capped, so a real major outside a cap would
+#     vanish silently. This repo is public and those `>=` floors are part of the
+#     contract with OSS consumers, so a raised floor is not purely churn here.
+# Net effect: constraint-widening PRs are a known, accepted cost in this repo.
 
 version: 2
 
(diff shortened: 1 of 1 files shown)


Eldad-Caura:
@claude please review.

Focus on the comment text rather than the one-line config revert. The block this replaces was reverted precisely because it asserted a mechanism that measurement then falsified, so the thing worth scrutinising is whether the replacement over-claims in the same way.

Specifically, please check:

1. Every factual claim in the new header block. In particular the dependabot-core claim — that on a lock-less entry the manifest edit is skipped when the new version already satisfies the range and no lockfile exists, so no PR is filed. I read that from `python/lib/dependabot/python/update_checker/requirements_updater.rb` (`skip_pyproject_update?`), and the `uv` ecosystem reuses that class via a bare alias. Is that a fair reading, and is "no PR is filed at all" too strong?
2. Whether `lockfile-only IS honoured` is stated too confidently. My basis is that `updated_requirements` opens with an unconditional `return requirements if update_strategy.lockfile_only?`, ahead of all per-file branching — but I have not run it.
3. Placement: the durable rule moved from the `uv` entry to the file header so it governs the `pip` entry too. Does anything in the header now contradict the per-entry comments further down?
4. Anything in the new text that will silently rot — I deliberately moved the point-in-time counts and PR numbers into the PR body instead, but tell me if something rot-prone survived in the config.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #748, opened 2026-08-10
ci: remove the Dependabot auto-merger revoked on 2026-08-09

Eldad-Caura:
## What this started as, and what it turned out to be

The task was a one-line comment fix: the header claimed *"Majors arrive as separate, ungrouped PRs (different branch names)"*, which is false — `dependabot.yml` defines five `*-majors` groups, so majors arrive grouped.

Reading the file to fix that surfaced something bigger. **This repo still carried the pre-revocation auto-merger.**

| repo | job | behaviour |
|---|---|---|
| caura-ops | `flag-ready` | verifies CI, applies a label |
| caura-memclaw-enterprise | `flag-ready` | verifies CI, applies a label |
| **caura-memclaw (this repo)** | **`automerge`** | **mints an App token, `gh pr merge --squash`** |

That job was carve-out #2 of the never-auto-merge policy. The carve-out was **revoked on 2026-08-09**; the two sibling repos were rewritten at the time and this one was missed — so a policy that no longer exists stayed implemented, in the public repo.

## It never merged anything — for two accidental reasons

1. `permissions:` granted `contents`/`pull-requests`/`checks`/`statuses` but **not `actions: read`**, the scope the rollup poll needs to resolve check runs. This is the same omission that silently no-op'd both siblings for their entire lifetimes.
2. The App was never added as a review-bypass actor on `main` — `bypass_pull_request_allowances` is empty — which the deleted header itself listed as a `PREREQUISITE`. The merge would have been refused regardless.

Evidence, rather than inference: every merged Dependabot PR here was merged **by hand**, including #649 whose branch `dependabot/uv/core-api/uv-minor-patch-bad095c31f` matched the gate exactly. Run [31391240657](https://github.com/caura-ai/caura-memclaw/actions/runs/31391240657) went green against #743 today at 13:06, and #743 is still open.

So it was armed but jammed. Anyone "fixing" the permissions the way we fixed the siblings would have switched on auto-merging in a public repo.

## Why deleted rather than rewritten to `flag-ready`

The label that job exists to write has **no reader here**:

- `ci-verified` is absent from `.github/labels.yml`, which `labels-sync.yml` treats as the source of truth — a port needs a new label entry plus a sync, not just a workflow
- the only workflow reading PR labels is `stale.yml`, and only `status/*`

A ported version would write a label nothing consumes, and would have to **keep `pull_request_target`** to do it (Dependabot's `pull_request` token can't label). Deleting instead removes this repo's **last `pull_request_target` trigger** — `claude_code_review.yml` uses plain `pull_request` — which on a public repo is a real reduction in attack surface. The backlog pressure the carve-out targeted is already handled structurally: `dependabot.yml` caps each ecosystem at two PRs per run.

## Verified safe to remove

- `main`'s required status checks are exactly `["CI","dco"]` — this was never a required check, so removing it cannot block PRs
- nothing in the tree references the file, its job name, or the label
- `DEPLOY_APP_ID` / `DEPLOY_APP_PRIVATE_KEY` remain in use by `release-please.yml`, so the org secrets are still needed
- all remaining workflow YAML and `dependabot.yml` re-validated after the change
- `CONTRIBUTING.md:91` ("Branch protection requires CI green, DCO check green, and ≥1 maintainer approval before merge") becomes unconditionally true — the carve-out was its one exception

## The three added lines

With the workflow gone, this repo would hold **no in-tree evidence** that the carve-out ever existed or was revoked — which is precisely the drift that let the broken copy survive here for weeks after the siblings were fixed. The note sits in the `dependabot.yml` header, next to the existing cross-repo line and directly above the `*-minor-patch` groups the carve-out selected on, so the next person syncing from `caura-ops` sees it.

I did not touch the logic. Your read was right that it was correct: the majors groups genuinely `skipped` today (runs 31388948139 for the actions group, 31388881249 for `uv-majors`), because the allowlist excludes them by name regardless of grouping.


Eldad-Caura:
@claude please review.

This began as a one-line comment fix and became a deletion, so the scope change is the main thing to check.

1. **Is deleting right, versus porting the sibling repos' `flag-ready` job?** My argument: the `ci-verified` label has no reader in this repo (absent from `.github/labels.yml`, which `labels-sync.yml` treats as the source of truth; the only label-reading workflow is `stale.yml`, `status/*` only), and a ported version would have to keep `pull_request_target` to write a label nothing consumes. Deleting removes this repo's last `pull_request_target` trigger. Is there a consumer I have missed?

2. **Did I leave anything dangling?** I grepped for `dependabot-automerge`, `ci-verified` and `automerge` and found no references outside the deleted file, and confirmed `DEPLOY_APP_ID`/`DEPLOY_APP_PRIVATE_KEY` are still used by `release-please.yml`. Please check docs, templates and scripts independently.

3. **Is removal safe against branch protection?** `main` requires exactly `["CI","dco"]`, so the deleted workflow was never a required check. Worth a second pair of eyes — getting this wrong would block every PR.

4. **The three added lines in `dependabot.yml`.** They assert a policy (carve-out revoked 2026-08-09, do not port back). Is that the right place and the right length, or is a commit message the better home? I put it in-tree because the absence of any record here is what let this repo keep the broken copy after the siblings were fixed.

I deliberately did NOT change the workflow logic — the allowlist correctly excluded majors groups by name, which today's skipped runs confirm.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

Eldad-Caura:
All checks green: 8 pass, 4 skipping, no failures. Both required checks (`CI`, `dco`) pass.

@erni-a over to you when you get a chance — flagging one caveat about how much the automated review is worth here.

The `claude-review` verdict came back "✅ No issues found", but the `pull_request` run classified the diff as *"no source files to review (only tests/docs/config)"* and the comment-triggered run returned a bare pass without engaging the four specific questions I raised. For a change that **deletes a workflow**, that is a thin signal, so the assurance really comes from the explicit checks rather than the reviewer:

- `main`'s required status checks are exactly `["CI","dco"]` — the deleted workflow was never required, so this cannot block PRs
- no references to `dependabot-automerge`, `automerge`, or `ci-verified` anywhere outside the deleted file
- `DEPLOY_APP_ID` / `DEPLOY_APP_PRIVATE_KEY` still consumed by `release-please.yml`, so the org secrets stay needed
- `bypass_pull_request_allowances` on `main` is empty, so no stray review-bypass privilege is left behind
- all remaining workflow YAML plus `dependabot.yml` re-validated after the change

The judgement worth a human read is the scope decision: this deletes rather than porting the siblings' `flag-ready` job, on the grounds that `ci-verified` has no reader in this repo and a port would have to retain `pull_request_target` to write a label nothing consumes. If you would rather keep parity across the three repos, that is the call to push back on.

Not merging — ready for your review.

erni-a (approved):

PR #749, opened 2026-08-10
test: stop 107 tests advertising isolation they don't have

Eldad-Caura:
## The signal that was wrong

`tests/conftest.py`'s `db` fixture is a per-test session inside a transaction that
is **rolled back** at teardown. Requesting it reads as "this test is transactionally
isolated."

Almost everything in `tests/` writes through the service layer instead, which commits
on its own connections — so that rollback never touches those rows. **107 test
signatures had drifted into requesting `db` and never using it**: a false isolation
claim, plus a connection, transaction and rollback bought for nothing.

Nothing is lost by dropping it. The autouse `_patch_storage_client`
(`tests/conftest.py:185`) already depends on `_engine` and `_setup_schema`, so every
test gets schema setup either way.

## 73 were direct; 34 were hidden

Five helpers take `db` and never touch it — `_seed_governance`, `_seed_memories`, two
different `_insert_memory`, and `_insert_entity`. `helper(db, ...)` looks like a use
of the name while the parameter goes nowhere, so their callers had to keep requesting
a session that was discarded. Those helpers and their **46 call sites** lose the
argument too.

Worth knowing for anyone repeating this: **seven different `_insert_memory` helpers
exist across `tests/`, and some genuinely use `db`** (e.g. `test_redistribute.py:88`).
Matching by bare name would have broken them, so this was scoped per file.

26 tests still request `db`, and all 26 genuinely use it.

## A guard, so it cannot silently come back

`tests/test_fixture_hygiene.py` asserts two things:

1. no **helper** may take `db` without using it, and
2. no **test** may request `db` without using it.

(1) is what makes (2) sufficient — without it, forwarding to a helper that drops the
parameter reads as a use, which is exactly how 34 of the 107 stayed hidden from the
first sweep. Verified by mutation: reintroducing a dead `db` on a test trips (2),
reintroducing one on a helper trips (1).

## No speedup, and none claimed

The fixture costs **~0.111 ms per use**, so 73 removals is **~0.008 s** against a
~210 s suite. This is hygiene, not performance. (The 258.7 s → 255.7 s I first
observed is noise, ~375× larger than the real saving.)

## The concrete damage: the dedup benchmarks measured an empty table

`tests/test_p2_semantic_dedup.py`'s benchmarks seeded via `db.add()` + `db.flush()` —
into the transaction that rolls back — while the query under test reached Postgres on
a separate connection. **Measured: seed 100, the query sees 0.**

So both `test_latency_100_memories` and `test_latency_1000_memories` timed an empty
table, met their 10 ms and 50 ms budgets comfortably, and said nothing. They were also
using the run's **shared** tenant, whose size is a running total of whatever every
other test left behind rather than the stated count.

They now commit their corpus under their own tenant, **assert it is actually visible
before timing anything**, and clean up after. That assertion is load-bearing, not a
sanity check — it is precisely what was missing. Real numbers, and they scale now:

| corpus | mean | median | max |
|---|---|---|---|
| 100 memories | 3.54 ms | 2.44 ms | 22.57 ms |
| 1000 memories | 4.56 ms | 4.31 ms | 6.22 ms |

## Two conftest comments asserted the belief being removed

- `conftest.py:164` said `per-test rollback is the primary isolation`. It isn't — the
  end-of-session `LIKE 'test-tenant-%'` sweep is what removes service-layer writes.
- `_reset_hooks`'s docstring said tests "that use the `db` fixture" bypass the app
  lifespan, which described a minority even before this change.

## Incidental, in blocks already being edited

13 in-body imports of `_find_semantic_duplicate` made redundant by hoisting it; one
unused `MagicMock`. And the benchmark's "p99" is now `max` — at 20 samples the
99th-percentile index *is* the last element, so it was a max wearing a percentile's
name.

## Verification

- `tests/`: **4185 passed, 3 skipped, 1 xfailed** — the 4183 baseline plus the 2 new
  guard tests. Confirmed under random ordering, not just a pinned seed.
- `core-storage-api/tests/`: **114 passed**, unchanged
- mypy clean on both projects; ruff clean at all four CI scopes
- `plugin/tools.json` and the broker OpenAPI regenerate byte-identical
- `tests/` is outside CI's ruff scope, so no formatter could tidy up after a bulk
  edit — the rewrite was AST-guided and every changed line was read. Per-file ruff
  counts are unchanged everywhere except `test_p2_semantic_dedup.py`, which improves
  from 6 findings to 1.

- **Making `tenant_id` unique per test** — the actual root fix, since the shared
  tenant is the leftover-row generator. It looks like a one-line conftest change and
  measured green in a scratch run, but it changes isolation semantics for 72 tests
  across 11 files and deserves its own PR and its own verification.
- **`memories.supersedes_id` has no index**, so its `ON DELETE SET NULL` self-FK
  fires a full table scan per deleted row — measured 1.9 s to delete 1000 rows from a
  28k-row table, growing linearly. That affects production deletes, not just tests.
- **Three more tests that assert against their own simulation** rather than the code
  they name: `test_visibility.py:678` (benchmarks a local Python reimplementation and
  concludes "<1% overhead" about the real SQL), `test_redistribute.py:125-255`, and
  `test_p0_3_recall_boost.py:225`. Same defect class as the benchmark fixed here.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -161,7 +161,12 @@ async def _setup_schema(_engine):
                     text(f"DELETE FROM {table} WHERE tenant_id LIKE 'test-tenant-%'")
                 )
             except Exception:
-                pass  # Best-effort cleanup; per-test rollback is the primary isolation
+                # Best-effort, and it is the PRIMARY isolation for anything
+                # written through the service layer — not a backstop behind the
+                # ``db`` fixture's rollback. Most tests here write via ``sc``,
+                # which commits on its own connections, so that rollback never
+                # sees those rows and this sweep is what removes them.
+                pass
         # memory_entity_links doesn't have tenant_id — clean via memory join
         try:
             await conn.execute(
@@ -360,9 +365,11 @@ from tests._mcp_test_helpers import mcp_env, parse_envelope, strip_latency  # no
 def _reset_hooks():
     """Ensure hooks are wired for integration tests.
 
-    In production, hooks are configured at app startup (lifespan).  Tests that
-    use the ``db`` fixture bypass the app lifespan, so we wire hooks here to
-    guarantee audit logging behaves identically to the running server.
+    In production, hooks are configured at app startup (lifespan). Tests that
+    exercise the services directly bypass that lifespan, so we wire hooks here
+    to guarantee audit logging behaves identically to the running server. This
+    is autouse because that covers nearly every test in this tree, not only the
+    ones holding a ``db`` session.
     """
     from core_api.services.audit_service import log_action
     from core_api.services.hooks import ServiceHooks, configure_hooks, reset_hooks
--- a/tests/pipeline/test_search_pipeline.py
+++ b/tests/pipeline/test_search_pipeline.py
@@ -25,7 +25,7 @@ _SEED_CONTENTS = [
 ]
 
 
-async def _seed_memories(db, count: int = 3) -> list[MemoryOut]:
+async def _seed_memories(count: int = 3) -> list[MemoryOut]:
     """Insert test memories via the legacy write path and return them."""
     from core_api.services.memory_service import create_memory
 
@@ -460,11 +460,11 @@ async def test_parallel_embed_gather_has_timeout():
 
 
 @pytest.mark.asyncio
-async def test_pipeline_search_returns_results(db):
+async def test_pipeline_search_returns_results():
     """Pipeline search path returns MemoryOut results for seeded memories."""
     from core_api.services import memory_service
 
-    seeded = await _seed_memories(db, count=2)
+    seeded = await _seed_memories(count=2)
     tid = seeded[0].tenant_id
 
     original = memory_service._USE_PIPELINE_SEARCH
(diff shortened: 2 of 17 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@erni-a ready for approval — all 9 checks green (CI, dco, CodeQL python + js/ts, Security Scan, core-api version bake, dependabot config, pre-checks). Single commit, DCO signed off.

One thing worth flagging before you approve: **the automated review skipped this diff as out of scope** — `claude-review` reports `skipping` with "no source files to review (only tests/docs/config)", since this repo excludes `tests/` from review. So the bot's "No issues found" reflects an empty review scope, not an assessment of the change. Human review is the only substantive review available here.

What stands in its place: a 4-agent `/simplify` pass, which found a real gap in my own first cut — my detector counted `helper(db, ...)` as a use of the fixture, hiding 34 of the 107 cases behind five helpers that take `db` and never touch it. Those are fixed, and the new guard test asserts both halves so the same blind spot can't recur (mutation-tested in both directions).

Highest-risk part to eyeball: the bulk edit was AST-guided text surgery across 17 files, because `tests/` is outside CI's ruff scope and no formatter could tidy up after it. Per-file ruff counts are unchanged everywhere except `test_p2_semantic_dedup.py`, which improves 6 findings → 1.

erni-a (approved):

PR #750, opened 2026-08-11
test: interleave the pipeline-vs-legacy latency paths to stop the flake

Eldad-Caura:
## The threshold wasn't the problem; the ordering was

`test_pipeline_overhead_bounded_relative_to_legacy` timed **ten legacy writes, then
ten pipeline writes**, and compared medians against a 1.5x ceiling. It failed at
1.545x on 2026-08-09 — a 3% overshoot — and passed on rerun with identical code.

Whichever block ran second met the worse conditions: runner drift, plus caches gone
cold after the first block's work. That was always pipeline. Measured over 20 paired
replicates locally:

| shape | ratio CV | worst replicate | margin to 1.5x |
|---|---|---|---|
| separate blocks (old) | **10.5%** | **1.403** | 7% |
| interleaved (new) | **4.9%** | 1.175 | 28% |

The block shape reached 1.403 on an *idle laptop* in 20 replicates, which is entirely
consistent with 1.545 on a shared runner. There was also a concrete cold-cache
artefact: in block form the pipeline's first write averaged **20.1 ms vs 14.6 ms** for
its second, because its warmup had happened ~10 writes earlier. Interleaved, that
first write is 14.2 ms.

Widening the ratio would have left the bias in place.

## What I got wrong on the first attempt

An earlier version of this change also moved the gate to **min-vs-min**, arguing that
scheduling noise is strictly additive so the minimum is the cleanest estimator. A
`/simplify` review measured that and it was wrong twice over:

- **Latency climbs ~28% from the first write to the tenth**, because the dedup query
  scans the tenant's rows rather than an ANN index (`EXPLAIN ANALYZE` shows
  `Index Scan` + `Sort`, no HNSW). So per-write cost grows with rows already written,
  and `min` lands on the first iteration — argmin was `i=0` in 15 of 20 replicates. It
  measures the **smallest candidate set**, not the least noise.
- **It bought no stability anyway** once interleaved: CV 4.50% (min) vs 4.71%
  (median) across 10 cross-process runs. min was only better in the *block* shape —
  i.e. it was compensating for the variance interleaving already removes.
- And it went blind to the likeliest regression class here. A per-candidate cost added
  to dedup: **0% detected by a min gate, 90-100% by the median.** An intermittent
  slowdown on 70% of writes: 2.9% vs 92.9%.

So the statistic is unchanged. **min is now printed as a diagnostic only**, and the
comment no longer claims noise is the dominant source of within-run spread.

## Also: this test leaked rows on every run, forever

Its tenants used a `test-lat-` prefix, but conftest's session teardown sweeps
`test-tenant-%`. The rows are committed through the service layer, so nothing else
removed them — **every run leaked 22 rows permanently, and 4,334 had accumulated in
one dev database.** Renamed to `test-tenant-lat-*`; verified the sweep now collects
them (0 rows remain after a run).

## Verification

- **8 consecutive local runs**: median ratio **1.00–1.10x** against the 1.5x gate
- `tests/`: **4185 passed, 3 skipped, 1 xfailed** (unchanged); `core-storage-api/tests/`: **114 passed**
- mypy clean both projects; ruff clean at all four CI scopes; generated artifacts byte-identical
- Timing boundary preserved: `_make_input` is still evaluated at the call site, so
  `MemoryCreate` construction stays outside the timed region exactly as before

## Known residual, in the safe direction

Within each interleaved pair the second write measures ~3.8pp slower, and pipeline
runs second — so the ratio is if anything *overstated*. Recorded in the comment.

The review made a strong case that this test is in the wrong tier entirely, and I
agree, but it is a CI-policy change rather than a test fix:

- **It is the only wall-clock ratio test that gates CI.** An identically-shaped
  assertion (`assert ratio < 3.0`) in `tests/test_p5_p6_benchmarks.py:96` is
  `@pytest.mark.benchmark` and therefore deselected. **But nothing ever runs
  `-m benchmark`** — the tier holds 8 tests and is write-only, so marking this one
  without also adding a non-gating scheduled job would trade a flake for silent zero
  coverage.
- `tests/test_p1_contradiction.py:606` and `:646` are pure microbenchmarks whose only
  assertion is a wall-clock number, marked `integration`, gating CI, burning 100k
  timed iterations per run.
- **A timing-free formulation exists**: spy on `get_storage_client` and assert the
  pipeline issues no more storage round-trips than legacy, plus assert
  `result.step_count`. The 1.5x gate permits ~16 ms of slack on a ~33 ms floor, which
  is ~500-1000x the per-step glue it names, so it cannot resolve that invariant at all.
- This test also bundles a durable field-equivalence guard (17 columns + embeddings +
  metadata) with the timing gate, so a runner hiccup and a real equivalence break
  present as the same failure.

### DIFF
--- a/tests/pipeline/test_pipeline_latency.py
+++ b/tests/pipeline/test_pipeline_latency.py
@@ -22,8 +22,12 @@ ITERATIONS = 10
 
 # Two separate tenants so both paths can write the same content
 # without triggering cross-path dedup
-LEGACY_TENANT = f"test-lat-legacy-{uuid.uuid4().hex[:8]}"
-PIPELINE_TENANT = f"test-lat-pipe-{uuid.uuid4().hex[:8]}"
+# ``test-tenant-`` prefix deliberately: conftest's session teardown sweeps on
+# exactly that pattern, and these rows are committed through the service layer
+# so nothing else removes them. Under the old ``test-lat-`` prefix every run
+# leaked 22 rows permanently — 4,334 had accumulated in one dev database.
+LEGACY_TENANT = f"test-tenant-lat-legacy-{uuid.uuid4().hex[:8]}"
+PIPELINE_TENANT = f"test-tenant-lat-pipe-{uuid.uuid4().hex[:8]}"
 FLEET_ID = "test-fleet"
 AGENT_ID = "test-agent"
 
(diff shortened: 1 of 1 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #751, opened 2026-08-11
perf(db): index the FK columns that reference memories.id / entities.id

Eldad-Caura:
## The defect

PostgreSQL enforces a foreign key from the **referencing** side, once per deleted
parent row. An unindexed referencing column therefore turns every delete into a full
scan of that table. Three were missing one:

| referencing | → target | on delete | index |
|---|---|---|---|
| `memories.supersedes_id` | `memories.id` | SET NULL | **none** |
| `relations.evidence_memory_id` | `memories.id` | SET NULL | **none** |
| `memory_entity_links.entity_id` | `entities.id` | CASCADE | **none usable** |

The third is easy to miss and was **not** in the original report — the PK is
`(memory_id, entity_id)`, so the memories-side FK is covered by the leading column
while `entity_id` is not. A btree cannot serve a trailing column as a prefix.

## Measured

Deleting 4,334 rows from a 30,801-row `memories` table, one statement, `EXPLAIN ANALYZE`:

| FK trigger | before | after |
|---|---|---|
| `memories_supersedes_id_fkey` | **15,343 ms** | **8.7 ms** |
| `memory_entity_links_memory_id_fkey` | 77 ms | 29.7 ms |
| `relations_evidence_memory_id_fkey` | 17 ms | 7.4 ms |

The `supersedes_id` trigger was **98.7% of the statement** at 3.5 ms per deleted row —
and only **87 of 30,801 rows** had a non-NULL `supersedes_id`. The cost is the scan, not
the matches. Linear in table size for a fixed batch (prod's ~92k rows ≈ 3× worse),
quadratic for a fixed-*fraction* purge where the batch grows with the table.

**Not a test-only concern.** `memory_purge_soft_deleted` (batched, via the CAURA-656
retention fanout), `purge_tenant_data` and `purge_fleet_data` all delete in bulk — and
the RI scan is **not tenant-scoped**, so deleting a tenant's children first doesn't
shrink it: every other tenant's rows get scanned once per deleted parent.

## The part worth reviewing: the invalid-index guard

An interrupted `CREATE INDEX CONCURRENTLY` leaves the index in `pg_index` with
`indisvalid = false`. `IF NOT EXISTS` then **skips** it, and the planner refuses to use
it — so the migration stamps green while leaving an index that is permanently useless
*and* still charges write overhead on every insert. Nothing later revisits it.

This is reachable in production because `CONCURRENTLY` waits on concurrent transactions
with no upper bound, and Cloud Run kills the container at the 240 s startup probe.

**Reproduced end to end, on the real table:**

1. Forced an abort with `lock_timeout` behind an open writer → `indisvalid=false` ✓
2. Ran a bare `CREATE INDEX CONCURRENTLY IF NOT EXISTS` → `NOTICE: relation
   "ix_memories_supersedes_id" already exists, skipping`, **still invalid** ✓
3. Ran this migration → guard dropped and rebuilt it; all three `indisvalid = t` ✓

**An earlier draft of this PR had no guard and its docstring claimed `IF NOT EXISTS`
made the migration idempotent.** That claim was false, and it was the stated reason no
guard was added. Migrations 005/007/026 all carry this guard — they are exactly the
three that index large pre-existing tables. 035 now matches them.

## A guard so the next one fails at PR time

`test_every_fk_referencing_column_is_index_leading` walks SQLAlchemy model metadata (no
DB) and asserts every FK's referencing column is usable as an index prefix. **It is what
found the `memory_entity_links` case.** Mutation-tested: removing that index fails the
test naming `memory_entity_links(entity_id) -> entities.id`.

Declaration-level by design — it catches a new FK added without an index, and can't
catch an index declared but never migrated. The CONCURRENTLY guard and chain tests cover
the migration side.

## Verification

- `tests/`: **4186 passed, 3 skipped, 1 xfailed** (4185 baseline + 1 new guard)
- `core-storage-api/tests/`: **114 passed**
- mypy clean; ruff clean at all four CI scopes (incl. `common/`); artifacts byte-identical
- **downgrade round-trip**: 035 → 034 drops all three, → 035 recreates; re-running
  `upgrade` is a clean no-op
- Build cost is not the risk: `CREATE INDEX CONCURRENTLY` on `supersedes_id` took
  **17.9 ms** for 26,927 rows (only 87 non-NULL, 200 kB index) — ~61 ms extrapolated to
  prod. The unbounded part is CONCURRENTLY's *wait*, which is what the guard covers.

## Notes for the reviewer

- Each `CREATE` is a **literal one-liner on purpose**: `test_no_plain_create_index_on_large_tables`
  regex-scans this file and matches neither a split string nor an interpolated `{name}`.
  A tidier loop would silently drop this migration out of guard coverage.
- Which also means **007 and 026 are currently outside that guard** (their f-string index
  names don't match `\w+`). Worth fixing the regex separately, along with a test asserting
  every CONCURRENTLY migration carries the `indisvalid` guard — nothing enforces that
  pattern today, which is how this PR could omit it and still pass CI.
- `test_single_head` hardcodes the expected head, so every migration PR edits a string in
  a test. It could derive it (assert exactly one head, plus a contiguous numeric run)
  without losing coverage. Left alone here to keep this PR to one idea.
- `relations.evidence_memory_id` is currently unmeasurable (that table is near-empty in
  dev) and included deliberately: it is written by a live co-occurrence inference path so
  it grows pairwise, the index costs one btree entry per insert, and at prod scale an
  unindexed version projects to ~5.7 s per 500-row retention batch.

### DIFF
--- a/common/models/entity.py
+++ b/common/models/entity.py
@@ -47,6 +47,14 @@ class Relation(Base):
     __table_args__ = (
         Index("ix_relations_from", "from_entity_id"),
         Index("ix_relations_to", "to_entity_id"),
+        # For DELETEs on ``memories``, not reads here — the referencing side of
+        # a SET NULL FK. Partial because the RI check never looks for NULL.
+        # See migration 035.
+        Index(
+            "ix_relations_evidence_memory",
+            "evidence_memory_id",
+            postgresql_where=text("evidence_memory_id IS NOT NULL"),
+        ),
     )
 
 
@@ -60,3 +68,11 @@ class MemoryEntityLink(Base):
         ForeignKey("entities.id", ondelete="CASCADE"), primary_key=True
     )
     role: Mapped[str] = mapped_column(Text, nullable=False)
+
+    __table_args__ = (
+        # The PK ``(memory_id, entity_id)`` covers the memories-side FK on its
+        # leading column, but a btree cannot serve ``entity_id`` as a prefix —
+        # so deleting an entity scanned this whole table, across every tenant.
+        # See migration 035.
+        Index("ix_memory_entity_links_entity_id", "entity_id"),
+    )
--- a/common/models/memory.py
+++ b/common/models/memory.py
@@ -106,6 +106,17 @@ class Memory(Base):
         ),
         Index("ix_memories_valid_range", "ts_valid_start", "ts_valid_end"),
         Index("ix_memories_subject_entity", "subject_entity_id"),
+        # For DELETEs, not reads — grep will find no query using it. This is the
+        # referencing side of a SET NULL self-FK, which PostgreSQL enforces once
+        # per deleted parent row. Partial because the RI check only ever looks
+        # for a non-NULL match, and almost every row here is NULL (232 kB -> 16 kB).
+        # Created CONCURRENTLY in migration 035, which carries the measurements;
+        # declared here so reflection / autogen round-trip against the live schema.
+        Index(
+            "ix_memories_supersedes_id",
+            "supersedes_id",
+            postgresql_where=text("supersedes_id IS NOT NULL"),
+        ),
         Index("ix_memories_recall_count", "recall_count"),
         Index("ix_memories_tenant_fleet", "tenant_id", "fleet_id"),
         # Backs the cursor-paginated list path (``list_by_filters`` +
--- a/core-storage-api/src/core_storage_api/database/migrations/versions/035_fk_indexes_for_memory_deletes.py
+++ b/core-storage-api/src/core_storage_api/database/migrations/versions/035_fk_indexes_for_memory_deletes.py
@@ -1,0 +1,134 @@
+"""Index the FK columns that reference ``memories.id`` / ``entities.id``.
+
+Deleting a parent row makes PostgreSQL enforce every foreign key that POINTS AT
+it, and enforcement needs an index on the REFERENCING side. Three were missing
+one, so each deleted row cost a full scan of the referencing table:
+
+    memories.supersedes_id           -> memories.id   SET NULL   no index
+    relations.evidence_memory_id     -> memories.id   SET NULL   no index
+    memory_entity_links.entity_id    -> entities.id   CASCADE    no usable index
+
+The last is subtle: ``memory_entity_links`` has PK ``(memory_id, entity_id)``, so
+``memory_id`` is covered as the leading column but ``entity_id`` is not — a btree
+cannot serve a trailing column as a prefix.
+
+Measured on a 30,801-row ``memories`` table, deleting 4,334 rows in one statement
+(``EXPLAIN ANALYZE``), before and after this migration:
+
+    Trigger memories_supersedes_id_fkey:        15343 ms  ->  8.7 ms
+    Trigger memory_entity_links_memory_id_fkey:    77 ms  -> 29.7 ms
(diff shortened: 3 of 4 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
Applied — and verified rather than assumed, because if the planner could not use a partial index for RI enforcement this would have silently reintroduced the 15-second delete.

Measured on the real table, same 1000-row delete, full vs partial:

| | index size | `memories_supersedes_id_fkey` trigger |
|---|---|---|
| full | 232 kB | 2.145 ms / 1000 calls |
| **partial** | **16 kB** | **1.848 ms / 1000 calls** |

And the plan confirms it directly — `EXPLAIN` on the RI query shape (`UPDATE ONLY memories SET supersedes_id = NULL WHERE supersedes_id = $1`) gives `Index Scan using ix_memories_supersedes_id` with the partial index in place. So it is used, and it is 14.5x smaller.

Two extensions beyond the suggestion:

- **`relations.evidence_memory_id` gets the same treatment.** It is nullable and also a SET NULL FK, so the identical argument applies; it went 8192 bytes partial.
- **`memory_entity_links.entity_id` deliberately stays a full index.** It is NOT NULL (half the PK), so a `WHERE ... IS NOT NULL` predicate would exclude nothing and only add noise. Confirmed via `information_schema`.

Also worth noting the write-side win the suggestion mentions is real: a plain btree *does* index NULLs, so before this every memory insert paid for an entry the RI check could never match.

Model declarations updated to match, and confirmed no autogen drift — `alembic check` reports nothing about any of the three indexes, and the DB's `indexdef` matches the declared predicate exactly.

Re-verified after the change: 3 indexes valid, 2 partial, downgrade/upgrade round-trip clean, `tests/` 4186 passed / 3 skipped / 1 xfailed, storage 114 passed, mypy + ruff clean at all four CI scopes.

Eldad-Caura:
@claude

Eldad-Caura:
@erni-a ready for approval — review clean on the amended commit ("✅ No issues found", posted after the re-tag), all 8 checks green: CI, dco, CodeQL python + js/ts, Security Scan, core-api version bake, pre-checks. Single commit `fc4be7a`, DCO signed off.

The one review finding (partial index) is applied and answered above with measurements.

**Highest-risk part to eyeball, since it is a prod migration:** the invalid-index guard in `upgrade()`. An interrupted `CREATE INDEX CONCURRENTLY` leaves `indisvalid = false`, `IF NOT EXISTS` then skips it, and the planner refuses to use it — so without the guard a deploy stamps green while leaving a permanently useless index that still charges write cost. I reproduced that whole chain locally (forced abort via `lock_timeout` behind an open writer → confirmed invalid → confirmed bare `IF NOT EXISTS` skips it → confirmed this migration drops and rebuilds it to valid). Migrations 005/007/026 carry the same guard.

Build cost is not the concern: the `supersedes_id` index took 17.9 ms for 26,927 rows and is 16 kB partial. `CONCURRENTLY`'s unbounded *wait* is the real exposure, which is what the guard covers.

Two things I flagged in the PR body as out of scope but worth someone's time: **007 and 026 are currently outside the CONCURRENTLY guard's regex** (their f-string index names don't match `\w+`), and nothing enforces that CONCURRENTLY migrations carry the `indisvalid` guard at all — which is exactly how this PR could have omitted it and still passed CI.

erni-a (approved):

PR #752, opened 2026-08-11
feat(memory): unified contradiction-model schema + read serialization (A55)

arkash20:
## What & why

Introduces the schema for the **unified contradiction-handling model**, which today is flattened onto `memories.status` + `memories.supersedes_id`. This PR is **additive and read-only** — *nothing populates the new fields yet* (that lands in a follow-up), so behaviourally it is a **no-op**. It unblocks the follow-up work (populating conflict records) and measuring against the STALE contradiction benchmark.

## Schema — migration `035` (on top of `034`)

- **`memories`** — three new columns:
  - `confidence` (REAL, nullable) — advisory confidence in the claim; guards the model's invariant *"weak evidence must not delete strong"*.
  - `is_inferred` (BOOL, default `false`) — system-materialised vs directly stated.
  - `scope` (JSONB, default `'{}'`) — structured validity qualifiers (role/task/location).
  - All nullable/defaulted with constant defaults → **metadata-only** column adds (no table rewrite).
- **`memory_conflicts`** (new) — one row per detected candidate conflict: relationship / diagnosis / evidence_strength / action + confidences + audit_reason. Classification is kept **separate** from applied state (the effect still lives on the memory row). Enum-like columns are `TEXT` + `CHECK` so the vocabulary can grow with a one-line change during iteration.
- **`memory_derivations`** (new) — lineage edges (inferred memory ← source) enabling revalidation when a source is later corrected/deleted.
- **No backfill.**

## Serialization (system-populated, read-only on the API)

- Storage `MEMORY_FIELDS` + core-api `MemoryOut` carry the three fields → recall/get/list surface them.
- SQLite backend parity (DDL columns + `scope` JSON-decode).
- **Public write schemas intentionally untouched** — these are system-only, never agent-supplied.

## Tests

- Storage HTTP round-trip + legacy-default; SQLite parity; migration-chain head bumped `034 → 035`.

## Wet test

- Full alembic chain `001 → 035` via the real runner (up / down / up) — single head `035`.
- `mypy` clean (storage 77 + core-api 190); `ruff` clean on all `src/` + `common/` + storage `tests/`.
- Storage suite **116 passed**; ORM e2e round-trip + FK-CASCADE verified.
- Top-level suite: zero new failures vs baseline (a set of FTS/entity/insights integration tests fail identically with and without this change under the local `create_all` harness because they depend on migration-built triggers — proven by a stash-and-rerun comparison).

## Deploy safety

Additive expand migration; nothing writes the fields yet. Metadata-only column adds; two new empty tables. Only operational note: adding the two FKs (`memory_conflicts → memories.id`) takes a brief lock on `memories` to install the referential trigger (child tables empty → no data scan).

### DIFF
--- a/common/models/__init__.py
+++ b/common/models/__init__.py
@@ -13,6 +13,8 @@ from common.models.fleet import FleetCommand, FleetNode
 from common.models.idempotency import IdempotencyResponse
 from common.models.lifecycle_audit import LifecycleAudit
 from common.models.memory import Memory
+from common.models.memory_conflict import MemoryConflict
+from common.models.memory_derivation import MemoryDerivation
 from common.models.skill_factory import ForgeRejectedFingerprint, SessionTrace
 
 __all__ = [
@@ -33,6 +35,8 @@ __all__ = [
     "IdempotencyResponse",
     "LifecycleAudit",
     "Memory",
+    "MemoryConflict",
+    "MemoryDerivation",
     "MemoryEntityLink",
     "Relation",
     "SessionTrace",
--- a/common/models/memory.py
+++ b/common/models/memory.py
@@ -2,7 +2,17 @@ import uuid
 from datetime import datetime
 
 from pgvector.sqlalchemy import Vector
-from sqlalchemy import DateTime, Float, ForeignKey, Index, Integer, Text, func, text
+from sqlalchemy import (
+    Boolean,
+    DateTime,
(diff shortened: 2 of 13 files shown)


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #753, opened 2026-08-11
feat(broker): gate the four memory operations the broker already calls

Eldad-Caura:
## The gap

`core-api/openapi.broker.json` exists so that a breaking change to a broker-facing
endpoint fails CI. **It covered 4 operations. The broker calls 8.**

Once memclawd's MCP dispatcher was wired to serve `memory_recall`, `memory_list`,
`memory_update` and `memory_delete` — all five capability interfaces are live in
`newCloudDispatcher` today — the broker started depending on:

| | | broker caller |
|---|---|---|
| `GET` | `/api/v1/memories` | `cloud.Client.ListMemories` (`memory_list`) |
| `GET` | `/api/v1/memories/{memory_id}` | `cloud.Client.GetMemory` (`memory_recall`) |
| `PATCH` | `/api/v1/memories/{memory_id}` | `cloud.Client.UpdateMemory` (`memory_update`) |
| `DELETE` | `/api/v1/memories/{memory_id}` | `cloud.Client.DeleteMemory` (`memory_delete`) |

The baseline was never widened, so a breaking change to any of them **passed the gate
silently**. This is the same failure mode as the #723–#736 series, one layer out: the
code grew a dependency and nothing declared it.

## Method-level, not path-level

`BROKER_OPERATIONS` is now `path -> methods`. `/api/v1/memories` also serves POST and
DELETE that the broker never calls — gating the whole path item would fail this gate on
unrelated changes to those operations, which is how a gate gets trained into noise.

The missing-operation preflight checks methods too. A path that survives while the
operation the broker calls is renamed is exactly as breaking as losing the path, and the
old path-only check sailed straight past it. Verified by adding a `PUT` the app doesn't
serve:

```
gen_broker_openapi: broker operation(s) missing from core-api OpenAPI: PUT /api/v1/memories
```

## Additive — `CONTRACT_VERSION` stays v1

This widens what the gate *protects*; it changes no API shape. Verified three ways:

- the four pre-existing path items are **byte-identical**, nothing removed
- `info` unchanged (`v1`)
- **the real gate passes**: I ran the exact CI oasdiff image *and digest* against main's
  baseline → `No breaking changes to report`, exit 0 (4 ops → 8 ops)

## Docs, because a list in a comment is what went stale

`ci.yml`'s comment enumerated four endpoints while the broker had grown to eight. It now
points at `BROKER_OPERATIONS` rather than restating it.

`core-api/scripts/README.md` gains the **cross-repo obligation** that caused this gap:
nothing in this repo can detect that the broker started calling a new cloud endpoint, so
a new `internal/cloud` client method must add its row here in the same change.

## Same gap in the other repo — not fixed here

Enterprise has its own per-service broker baselines. As of 2026-08-11 they cover
`/installs/{claim,heartbeat,policy/stream,register}` and `POST /api/v1/audit-log`, but
**not** `POST /api/v1/installs/leave` or `POST /api/v1/installs/commands/ack`, which the
broker calls from `internal/cloud/leave.go` and `internal/cloud/command_ack.go`. That
needs a separate enterprise PR; recorded in the README so it isn't lost.

## Verification

- `tests/`: **4187 passed, 3 skipped, 1 xfailed**; `core-storage-api/tests/`: **116 passed**
- mypy clean both projects; ruff clean at all four CI scopes; `plugin/tools.json` byte-identical
- `gen_broker_openapi.py` and `--check` both clean (8 operations)
- `actionlint` on `ci.yml`: **1 finding before my change, 1 after** — pre-existing, not
  introduced. (My first comparison showed 0-vs-1 and was wrong: actionlint skips files
  outside a workflows path, so the baseline run had silently linted nothing.)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -74,10 +74,13 @@ jobs:
           echo "$VIRTUAL_ENV/bin" >> $GITHUB_PATH
 
       # --- Broker↔cloud API-versioning gate (Gate C / Phase 1, deliverables A+B) ---
-      # core-api/openapi.broker.json is the FROZEN v1 contract for the four
-      # broker-facing gateway endpoints (POST /api/v1/memories/bulk,
-      # POST /api/v1/search, GET /api/v1/health, GET /api/v1/version) that
-      # memclawd calls against Caura cloud. It is a subset of the full ~91-path
+      # core-api/openapi.broker.json is the FROZEN v1 contract for the
+      # broker-facing gateway operations that memclawd calls against Caura
+      # cloud. The authoritative list is BROKER_OPERATIONS in
+      # core-api/scripts/gen_broker_openapi.py — deliberately not duplicated
+      # here, because this comment listed four endpoints while the broker had
+      # grown to eight, and a list in a comment is exactly what goes stale.
+      # It is a subset of the full ~91-path
       # OpenAPI surface, with info normalized to the frozen contract version so
       # it changes ONLY when a broker endpoint's shape changes. Regenerate with
       # `python core-api/scripts/gen_broker_openapi.py` and commit the result;
--- a/core-api/openapi.broker.json
+++ b/core-api/openapi.broker.json
@@ -757,6 +757,231 @@
         "title": "MemoryType",
         "type": "string"
       },
+      "MemoryUpdate": {
+        "properties": {
+          "content": {
+            "anyOf": [
+              {
+                "maxLength": 10000,
+                "minLength": 1,
+                "type": "string"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "title": "Content"
+          },
+          "entity_links": {
+            "anyOf": [
+              {
+                "items": {
+                  "$ref": "#/components/schemas/EntityLinkIn"
+                },
+                "type": "array"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "title": "Entity Links"
+          },
+          "expires_at": {
+            "anyOf": [
+              {
+                "format": "date-time",
+                "type": "string"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "title": "Expires At"
+          },
+          "memory_type": {
+            "anyOf": [
+              {
+                "$ref": "#/components/schemas/MemoryType"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "description": "Memory type. Auto-classified by LLM if omitted. Valid values: fact, episode, decision, preference, task, intention, plan, commitment, action, cancellation."
+          },
+          "metadata": {
+            "anyOf": [
+              {
+                "additionalProperties": true,
+                "type": "object"
+              },
+              {
+                "type": "null"
+              }
+            ],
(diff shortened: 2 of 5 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
Good catch on the untyped response — it's the same failure this PR is about, so it mattered. Verified: `GET /api/v1/memories/{memory_id}` has `"schema": {}` for its 200, so oasdiff can't detect a breaking response change to it. Listing it protects path, method, params and 422, but **not** the response shape.

**The suggested fix would have made it worse, though.** That route hand-builds a ~20-field `JSONResponse` (`routes/memories.py:519`), and FastAPI *skips* response validation when a handler returns a Response object — so attaching a `response_model` would document a guarantee the code doesn't make. A contract asserting an unenforced shape is worse than one admitting it's untyped. Typing it properly means refactoring the handler to return a model, which changes serialization on a live endpoint and belongs in its own PR.

So instead of prose caveat or false schema, the caveat is now a **guard** — `tests/test_broker_contract.py`:

- `_UNTYPED_200` is an explicit allowlist with a reason per entry.
- Fails if a **new** operation enters the contract untyped.
- Also fails if an allowlisted entry has **since become typed**, so the allowlist can't quietly outlive the problem.

Both directions mutation-tested: removing `GET /memories/{memory_id}` from the allowlist reports it as untyped; adding the already-typed `POST /search` reports it as stale.

Worth noting `GET /health` and `GET /version` were **already** untyped in the contract before this PR — the audit surfaced that too, and they're allowlisted with reasons rather than silently tolerated.

A second test pins the baseline to `BROKER_OPERATIONS`, so a hand-edited baseline fails in a local run rather than only on the runner.

Re-verified: `tests/` 4189 passed / 3 skipped / 1 xfailed, storage 116, mypy clean both, `--check` clean at 8 operations.

Eldad-Caura:
@claude

Eldad-Caura:
@erni-a ready for approval — review clean on the amended commit ("✅ No issues found"), all 8 checks green, single commit `4a43a7e`, DCO signed off.

The review's one finding (untyped 200 on `GET /memories/{memory_id}`) is addressed and answered above: the naive `response_model` fix would have documented a shape FastAPI doesn't enforce for that hand-built `JSONResponse`, so the untyped set is now an explicit allowlist guarded both ways by `tests/test_broker_contract.py`.

**Worth eyeballing:** the four added operations are additive only — I ran the exact CI oasdiff image *and digest* against main's baseline (`No breaking changes to report`, exit 0), and the four pre-existing path items are byte-identical.

**Known limitation stated plainly:** this gates path / method / params / request shape for the four new operations, and the response shape for three of them. `GET /memories/{memory_id}`'s response is *not* pinned (its route hand-builds a ~20-field JSONResponse) — allowlisted with the reason rather than papered over. Typing it properly is a live-endpoint serialization change and wants its own PR.

Also recorded in the README, not fixed here: enterprise's broker baselines don't cover `POST /api/v1/installs/leave` or `POST /api/v1/installs/commands/ack`, which the broker calls — same gap, other repo.

erni-a (approved):

PR #754, opened 2026-08-12
feat(memory): contradiction engine seam + arch flag (A55 Phase 1)

arkash20:
## What & why

Introduces a **single seam** over contradiction detection behind a flag, so the scattered trigger sites route through one entry point and the legacy arch can be retired later. Detection today lives in `contradiction_detector.py` and is fired from **8+ scattered sites**; this consolidates them.

**Behaviour-preserving and OFF by default** — when ON, the engine delegates to the same detector, so this PR is a **no-op** until the follow-up (Phase 2 / A55 1d) adds the `memory_conflicts` record write behind the ON path.

## Changes

- **Flag:** `settings.contradiction_engine_enabled` (default `False`).
- **`services/contradiction/`** — `ContradictionEngine.evaluate_async` facade (`Trigger` enum: WRITE/EMBED/UPDATE/BULK/REEMBED → Path A; ENTITY → Path C) + `run_contradiction_detection` flag-router. The router is a **sync function returning an awaitable**, so the legacy path keeps the exact schedule-time semantics of the old direct `detect_contradictions_async(...)` calls.
- **All 12 trigger sites** now route through the seam: consumer back-channel ×2, `ScheduleBackgroundTasks` ×2, `entity_extraction_worker` ×1, `memory_service` ×7. No direct legacy detector calls remain outside detector/engine/dispatch.

## Tests / guardrails

- **routing-parity** — for every trigger, flag ON reaches the same detector fn with the same args as flag OFF.
- **golden differential** — freezes today's detection *outcomes* (`status`/`supersedes_id`) for 7 scenarios (RDF both directions, semantic ±, Path C forward ±, retraction) into a committed fixture; the engine must reproduce it byte-for-byte.
- **trigger-coverage** — structural guard: no module bypasses the seam, and every `Trigger` kind is wired (catches a future dropped/bypassing site).
- Two observability fixtures repointed to the detector source symbol — **all assertions unchanged**.

## No-degradation evidence

Full top-level suite: failure set **byte-for-byte identical to the pre-existing baseline** (create_all integration tests only) — **zero regressions**. mypy clean (193 files), ruff + format clean.

Design + preservation table + Excalidraw current-vs-engine diagram: `benchmark/a55/subtasks/10-…` / `11-…` (local).

### DIFF
--- a/core-api/src/core_api/config.py
+++ b/core-api/src/core_api/config.py
@@ -291,6 +291,12 @@ class Settings(BaseSettings):
     log_file: str = ""
     # Default False: standalone=True bypasses tenant auth, so it must be an explicit opt-in.
     is_standalone: bool = False
+    # A55 contradiction-engine seam. False (default) => legacy detector call
+    # sites (today's behaviour, unchanged). True => route contradiction
+    # detection through ContradictionEngine.evaluate_async. Phase 1 the two
+    # paths are behaviourally identical (the engine delegates to the same
+    # detector); the flag exists so the old arch can be retired later.
+    contradiction_engine_enabled: bool = False
     crystallizer_enabled: bool = True
     crystallizer_stale_days: int = 180
     crystallizer_dedup_sample_size: int = 1000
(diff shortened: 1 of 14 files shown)


arkash20:
@Eldad-Caura please approve

arkash20:
@Eldad-Caura please approve

erni-a (approved):

PR #755, opened 2026-08-12
fix(capture): mirror the shared pipeline's diagnostics, effort and decline test

Eldad-Caura:
`caura-ai/.github` shipped **v1.5.2** today after two days of chasing an empty extraction. This ports what that produced — **adapted, not copied**, since this repo's copy is deliberately not a fork.

## Ported

**The thread diagnostic.** One notice with the thread length and surviving comment count, logged *before* the verdict gate. Its absence is what cost the shared pipeline two days: every round reasoned about "the thread" as read from a maintainer's laptop, and nothing established that the job sees the same thing. It need not — this runs as `github.token`, and `author_association` is **viewer-dependent**. If a comment is missing, `[]` is the *correct* answer to what the model was handed.

**`--effort max`, with a non-blocking probe.** The empty extraction upstream turned out to be a collapse in *reasoning*, not input — the run that worked generated thousands of output tokens, the ones that didn't generated two characters. This invocation never passed `--effort`, so it ran at the CLI's default. The probe omits the flag and warns if the pinned CLI lacks it; passing an unsupported flag is worse, since the CLI rejects the whole invocation and capture extracts *nothing* rather than merely extracting badly.

**The empty-extraction log.** `No declined findings` was the only silent failure path left in this file — every other one already logs its payload — and it's ambiguous between three outcomes needing different responses.

**The decline test.** Two questions: did the code change in response, and did the maintainer **concede** the finding is correct. The old wording said to ignore findings *"fixed, applied, or otherwise addressed"* — and a maintainer writing "already satisfied" reads as *addressed*, so the most valuable case (a finding refuted because its premise was false) was being instructed away.

## Deliberately **not** ported — the reasons matter more than the list

**The jq trust filter.** Upstream moved authority enforcement out of the prompt into a `select` on `author_association`. Porting it here would **break capture outright**: its bot condition requires a `<!-- caura-reviewer: -->` marker, and this repo's verdicts don't carry one — *checked, not assumed* — so every verdict would be filtered out and there'd be nothing left to decline.

The prompt-based OWNER/MEMBER control here is correct for a **public** repo and deliberately narrower than upstream's OWNER/MEMBER/COLLABORATOR, matching this repo's own retrigger gate. Upstream also never demonstrated the move fixed anything.

**`MAX_THREAD_CHARS`.** Upstream raised it to 90000 because two reviewers put twice the automated text in a thread. This repo runs one; 60000 is right for it.

## Worth knowing

**The shared drift checker cannot see this flag.** It extracts the CLI flag set from the literal invocation, and `--effort` now sits in a conditional array on both sides — so it reports "in sync" whether or not either repo passes it. That's a blind spot introduced upstream when the probe was added, not a pre-existing one.

## Measured, not assumed

- The prompt was rendered from the assembled variable to confirm the escaped quotes survive.
- Both repos' capture currently reports `[]` on recent merges — but the caura-memclaw threads I checked (#751, #753) contain **no declines at all**; every maintainer reply is "Applied" or "Good catch, verified". So `[]` is correct there and says nothing about breakage.

`bash -n` and `shellcheck` clean. DCO signed off.

### DIFF
--- a/.github/scripts/claude_pr_capture.sh
+++ b/.github/scripts/claude_pr_capture.sh
@@ -78,6 +78,26 @@ THREAD=$(gh api "repos/${REPO}/issues/${PR_NUMBER}/comments" --paginate \
   exit 0
 }
 
+# WHAT THE MODEL IS ACTUALLY BEING GIVEN, measured rather than assumed. Mirrored from the shared
+# pipeline, where its absence cost two days: every round of diagnosing an empty extraction reasoned
+# about "the thread" as read from a maintainer's laptop, and nothing established that the job sees
+# the same thing. It does not necessarily — this runs as `github.token`, and author_association is
+# VIEWER-DEPENDENT. If a comment is missing here, [] is the CORRECT answer to what the model was
+# handed, and every theory about prompts and models is aimed at the wrong half of the job.
+#
+# Logged BEFORE the verdict gate below, so it still reports on a thread that bails there: "no review
+# verdict" and "the thread arrived empty" are indistinguishable from outside, and telling them apart
+# is the point.
+#
+# Counted at RECORD BOUNDARIES rather than by matching the header anywhere, because a comment BODY
+# can contain a line shaped like one — most plausibly on a pull request where someone pasted a
+# thread excerpt, which is exactly when this number gets read. `jq -r` prints each record's own
+# trailing newline plus its own, so a real header always follows a blank line; a mid-body quote does
+# not. `awk` also prints 0 and exits 0 on no match, where `grep -c` exits 1 and would abort a job
+# whose entire contract is to warn and exit 0.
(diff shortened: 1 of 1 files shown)


Eldad-Caura:
@erni-a — ready for approval.

Clean review, all checks green including CI, DCO, CodeQL and the security scan.

This mirrors `caura-ai/.github` v1.5.2 into this repo's local copy of capture — adapted rather than copied, since the two are deliberately not forks. The part worth a second look is what I **declined** to port: upstream's jq trust filter would have broken capture here outright, because its bot condition requires a `<!-- caura-reviewer: -->` marker that this repo's verdicts do not carry. Checked rather than assumed. The prompt-based OWNER/MEMBER control here stays, and it is correctly narrower than upstream's.

Also flagged in the PR body: the shared drift checker cannot see the `--effort` flag either way, since it reads the literal invocation and the flag now sits in a conditional array on both sides. That blind spot came in upstream with the probe.

erni-a (approved):

PR #756, opened 2026-08-12
CAURA-704 feat(doc): mint a memory carrying doc content on every doc …

eyal-bl:
…write

The two stores are not cross-searched: `memclaw_recall` never returns documents, and a document row embeds only `data["summary"]` — never its body. So a document's CONTENT was unreachable by meaning; you could only find it if you already knew its `collection` + `doc_id`. The companion "reference memory" the skill docs tell agents to write by hand (plugin/skills/memclaw/SKILL.md) had no server-side implementation at all: no hook, no event, and routes/documents.py imported nothing from memory_service.

Every doc write now mints a `fact`-typed memory whose content is the document body VERBATIM (no prefix, no wrapper), so the body becomes reachable by the paths that read `memories.content` directly. `memclaw_recall` and the recall brief benefit immediately — neither truncates content.

Deliberate scope boundaries, all deferred to the follow-up:

- INSIGHTS IS UNCHANGED. It truncates every memory to 500 chars via `sanitize_content`'s default (core_api/utils/sanitize.py) — a hardcoded function default with no flag or per-memory override — so insights still sees only the first 500 chars of a doc body. Raising that needs its own per-run token budget and is a separate change.
- NO create-vs-update branch and no lookup of previously minted rows. Rewrite behaviour instead falls out of `write_mode="fast"`: `CheckExactDuplicate` 409s an identical body (idempotent re-syncs self-dedupe), while an edited body lands with `metadata["near_duplicate_of"]` pointing at its predecessor. Nothing outdates the older rows yet, so a doc edited N times leaves N active memories — `near_duplicate_of` is the thread a reconciliation pass follows.
- NO schema or storage-api change: no migration, no new storage filters. `source_uri` (`memclaw-doc://<collection>/<doc_id>`) is written as provenance only; nothing queries it yet.
- No size limit is added to doc writes themselves. Doc `data` is unbounded today and stays that way; an over-cap body is SKIPPED rather than truncated, because a truncated body reads as complete.

Skips: `skills` (preserves the staged -> active approval lifecycle — a recallable memory would route around it), `_`-prefixed system collections, bodyless docs (which is what keeps bulk structured records like `{"plan": "business", "seats": 40}` from minting), and bodies over MAX_CONTENT_LENGTH. `data["summary"]` is NOT required: it gates doc INDEXING, not minting, so a summary-less doc is invisible to `op=search` yet its body is still recallable.

Two details worth reviewer attention:

- `DOC_MEMORY_TYPE` is `MemoryType.FACT`, deliberately not "reference" — `reference` is not a MemoryType and the `memories_memory_type_check` CHECK constraint (migration 013) would reject it. `fact`'s taxonomy description covers this case explicitly ("reference material, taxonomies, API/schema documentation ... Includes agent-authored documentation").
- Attribution prefers the real doc writer's agent_id. This is not cosmetic: `memclaw_insights` defaults to `scope="agent"` and filters on `agent_id`, so a row attributed to the `memclaw-doc-indexer` service fallback is invisible to that agent's default insights run. The fallback applies only when the caller has no identity at all (REST with no gateway-stamped X-Agent-ID).

The derivation rule lives in `services/doc_indexing.py` beside `resolve_embed_source` so the MCP and REST surfaces cannot drift; a test asserts neither surface constructs a spec directly. Both call sites wrap the mint in try/except on top of `safe_sync_doc_memory`'s own non-raising contract — a test proved that without it a raising mint turns a successful doc write into an error envelope, the one outcome this must never cause.

Also narrows `collection` to a local `write_collection: str` in the MCP write branch, which incidentally fixes a pre-existing mypy error on the adjacent `resolve_embed_source` call (348 -> 347 tree-wide).

Verification: 55 new tests, all passing; zero new mypy errors; ruff check and format clean. Full suite shows 500 failures on this branch AND on clean main (pre-existing local DB schema drift — `audit_log.seq` missing), with +55 passing here, so no regressions. The wet test against the dev stack has not been run yet.

## Type of Change

- [ ] Documentation update
- [ ] Other:

## Checklist

- [ ] `mypy` passes
- [ ] `pytest` passes locally

### DIFF
--- a/core-api/src/core_api/agent_ids.py
+++ b/core-api/src/core_api/agent_ids.py
@@ -28,6 +28,19 @@ INSIGHTER_AGENT_ID = "memclaw-insighter"
 # route trust gate, so this is future-proofing for any gated re-route.
 INSIGHTER_TRUST_LEVEL = 3
 
+# Fallback identity for the memory minted from a document write, used ONLY when
+# the caller carries no agent identity at all (the REST ``POST /documents`` path
+# without a gateway-stamped ``X-Agent-ID``; the MCP path always has one). Same
+# reasoning as ``INSIGHTER_AGENT_ID``: a stable registerable service identity
+# rather than the anonymous default, and self-registered per tenant on first use
+# via ``get_or_create_agent``.
+#
+# Prefer the real doc writer's agent_id whenever there is one. Attribution is not
+# cosmetic here: ``memclaw_insights`` defaults to ``scope="agent"``, which filters
+# ``Memory.agent_id == agent_id``, so rows attributed to this service identity are
+# invisible to every real agent's default insights run.
+DOC_INDEXER_AGENT_ID = "memclaw-doc-indexer"
+
 # Bare ``"main"`` is the OpenClaw plugin's *unset* default agent_id: when an
 # operator never sets ``MEMCLAW_AGENT_ID`` every install collapses onto this one
 # shared identity (the eToro "firehose"). Unlike ``DEFAULT_AGENT_ID``
--- a/core-api/src/core_api/constants.py
+++ b/core-api/src/core_api/constants.py
@@ -171,6 +171,37 @@ MAX_CONTENT_LENGTH = 10000
 CHUNKING_THRESHOLD_CHARS = 2000  # content above this triggers auto-chunking
 MAX_QUERY_LENGTH = 5000
 
+# ── Doc-derived memories ──
+# ``memclaw_doc(op="write")`` mints a memory carrying the document body so the
+# body becomes reachable by MEANING (recall / recall-brief read
+# ``memories.content`` directly). Docs themselves are only searchable via their
+# ``data["summary"]`` embedding, and ``memclaw_recall`` never returns documents.
+#
+# The memory content is the doc body VERBATIM, so the cutoff is simply the
+# memory schema ceiling: mint whenever the body fits in a memory at all.
+# Over-cutoff bodies are SKIPPED rather than truncated — a truncated body reads
+# as complete and would produce confidently wrong downstream conclusions.
+DOC_MEMORY_MAX_CHARS = MAX_CONTENT_LENGTH
+
+# NOTE: there is deliberately no ``DOC_MEMORY_TYPE``. Doc-derived memories pass
+# no ``memory_type`` at all, so the enrichment classifier assigns one per
+# document — a decision record becomes ``decision``, a runbook ``rule``, an
+# incident writeup ``episode``. Pinning every document to a single type would
+# throw that away. (And ``reference`` was never an option: it is not a
+# MemoryType, and the ``memories_memory_type_check`` CHECK constraint from
+# migration 013 would reject it.)
+
+# Provenance only — written to ``memories.source_uri`` as
+# ``memclaw-doc://<collection>/<doc_id>``. Nothing queries it yet (that would
+# need a storage-side filter + index); it exists so a later reconciliation
+# mechanism has a stable key to match a doc to the memories minted from it.
+DOC_MEMORY_URI_SCHEME = "memclaw-doc"
+
+assert DOC_MEMORY_MAX_CHARS <= MAX_CONTENT_LENGTH, (
+    "DOC_MEMORY_MAX_CHARS must not exceed MemoryCreate.content max_length "
+    f"({MAX_CONTENT_LENGTH}) — an over-cap spec would fail schema validation."
+)
+
 # ── Tool surface bookkeeping ──
 # Tool descriptions live inline in `core_api/tools/memclaw_*.py` spec
 # modules (the SoT). Nothing else should hold a copy.
--- a/core-api/src/core_api/mcp_server.py
+++ b/core-api/src/core_api/mcp_server.py
@@ -1638,6 +1638,12 @@ async def memclaw_doc(
                     return _with_latency(
                         _error_response("INVALID_ARGUMENTS", "op=write requires 'data'."), t0
                     )
+                # ``collection`` is guaranteed non-empty here by the op-level
+                # guard above (write is not in the {list_collections, search}
+                # exemption set), but that guard is too far away for the type
+                # checker to narrow through. Rebind so the derivation helpers
+                # below take a plain ``str``.
+                write_collection: str = collection or ""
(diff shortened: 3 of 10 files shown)


eyal-bl:
@claude

eyal-bl:
Thanks — went through all four. Summary: **1 accepted-with-a-decision, 1 declined on evidence, 2 noted.** Verified each against the code rather than taking them at face value.

---

### 1. Synchronous work on the doc-write path — **correct, accepted, deliberately not changed yet**

You're right, and I've measured it. `write_mode="fast"` defers *only* enrichment; `ParallelEmbedEnrich`, `CheckExactDuplicate`, `DetectNearDuplicate` and `WriteMemoryRow` all still run awaited inline. On a VM against real providers:

```
write_mode=fast  embedding_ms=151  dedup_lookup_ms=9  storage_ms=16
total_ms=210     enrichment_pending=True
```

So ~**210 ms** added per doc write, plus `resolve_config` and the `get_or_create_agent` upsert. Your point that the agent upsert runs on *every* write (not just first-use) is also correct — it was made unconditional to fix a ghost-agent bug found in wet testing, and the cost wasn't revisited.

**Keeping it inline for now**, and documented in the module docstring with the measurements. Reasoning:

- The handlers are `async def` and every wait is on I/O, so the event loop keeps serving other requests — the 210 ms lands on the calling request's response, not on server throughput.
- `track_task` is `asyncio.create_task`: same process, **no queue, no durability, no retry**, and `cancel_all_tasks()` cancels in-flight tasks on shutdown. Backgrounding would hide the latency but let a doc write return `200` while its memory never materialises — a write landing during a deploy would silently lose it. Today a `200` guarantees the memory exists, which is the stronger property.
- For interactive one-doc-at-a-time writes, 210 ms isn't worth trading that away.

Revisit trigger recorded in the code: bulk doc sync. At 210 ms each, 10k docs ≈ 35 min of added wall-clock. The cheapest first win there is dropping the redundant `get_or_create_agent` (on the MCP path `enforce_fleet_write` has already registered the caller) before reaching for a durable queue.

---

### 2. structlog instead of stdlib logging — **declined; the convention is the opposite**

```
logging.getLogger(__name__) in core_api/services/ : 32 files
structlog                                        : 2 files
```

Stdlib logging *is* the house convention here, and these modules follow it. Output is already structured JSON (`{"logger": "core_api.services.doc_memory", "severity": "INFO", ...}`) because the app bridges stdlib into the configured renderer.

The secondary point has merit — `%s` formatting puts `collection`/`doc_id` in the message rather than as queryable fields — but converting just these two modules to structlog kwargs would make them the outliers against 32 peers. Worth a separate repo-wide decision, not a change in this PR.

---

### 3. Hardcoded `visibility="scope_team"` — **noted; the suggested fix isn't implementable**

`common/models/document.py` has **no visibility or ACL column** — only `tenant_id` and `fleet_id` (which *is* threaded through to the memory). So there's no document access scope to inherit.

The underlying question is fair though: a doc with `fleet_id=None` becomes a `scope_team` memory with no fleet, and whether that's the right exposure deserves a deliberate answer. Leaving as-is; flagging for the owner as a design question rather than guessing.

---

### 4. Module-level `assert` stripped under `-O` — **technically correct, following house precedent**

Right that `assert` is compiled out under `-O`/`PYTHONOPTIMIZE`. Noting that `common/enrichment/constants.py` uses module-level asserts for exactly this class of invariant in four places (lines 151, 159, 167, 171), so this follows an established pattern rather than deviating from one. Happy to convert to an explicit `raise` if we'd rather move away from that pattern — ideally for all five together.

eyal-bl:
@claude

eyal-bl:
@claude

eyal-bl:
Both new points verified — **first is correct and I'd have agreed with the fix; declining it anyway for simplicity. Second is correct and sharper than the earlier version.**

---

### 1. Auto-chunk guard starves the tenants who need the feature most — **correct; keeping the guard anyway, deliberately**

Your premise checks out, and it's a better read of the code than my own comment had. When chunking fires, the parent row gets:

```python
"content":    data.content,      # the FULL verbatim body
"source_uri": data.source_uri,   # our memclaw-doc:// provenance
```

So chunking would **not** break the feature — the parent stays recall-reachable with the body byte-for-byte. Which means my guard was never protecting correctness; it was avoiding N+1 rows per doc. And your framing is right: trading that cosmetic concern for a functional gap hits precisely the tenants who enabled chunking *because* they write long content.

**Decision: keeping the guard for now, for simplicity.** `auto_chunk_enabled` defaults to `False` and no tenant here has it on, so the gap is currently theoretical, and one-row-per-doc keeps the provenance key unambiguous while doc-version reconciliation is still unbuilt.

What I've changed is the honesty of the comment — it now states the tradeoff explicitly rather than implying the guard prevents harm:

> KNOWN TRADEOFF, chosen deliberately for simplicity: skipping means a tenant with `auto_chunk_enabled=True` gets NO recall-reachable memory for any body over `CHUNKING_THRESHOLD_CHARS` — and those are exactly the tenants who opted in because they write long content, so the feature is weakest where it would matter most. Letting chunking run would in fact work (the parent row carries the verbatim body and this `source_uri`), so this guard buys one-row-per-doc rather than correctness. Revisit together with the doc-version reconciliation pass, which has to handle multiple rows per `source_uri` anyway.

Your option (a) — a per-write chunk opt-out — would need a new `MemoryCreate` field threaded into the branch at `memory_service.py:727`, since `auto_chunk_enabled` is tenant-config only. Deferring rather than widening the shared write path in this PR.

---

### 2. Unindexed doc writes previously had zero embedding cost — **correct, and a sharper point than last round's**

Confirmed: a doc written without `data["summary"]` made **zero** embedding calls (`resolve_embed_source` returns `None`, so `get_embedding` is never reached). Now it always makes one, because the memory body is embedded.

So for that class of write it isn't just slower — a write path that had **no embedding-provider dependency at all** now has one. That's a distinct and more serious framing than "adds ~210 ms", and I'd only documented the latency.

Decision unchanged (see the previous round: inline, measured, revisit trigger recorded for bulk sync), but the new dependency is worth naming in the docstring rather than leaving as latency-only. Noting it here so the tradeoff is on the record.

Eldad-Caura (approved):

Eldad-Caura (approved):

PR #759, opened 2026-08-12
feat(memory): contradiction classification logic — L1/L2/L3 (A55 1d)

arkash20:
## What & why

The unified contradiction model's **decision layer** — L1 (relationship) → L2 (diagnosis) → L3 (action) — as **pure, dormant modules**. This is the classification brain for A55 1d, landing *ahead* of the write slice that wires it onto the engine-ON path.

**Behaviour-neutral by construction:** nothing calls these in production yet, and the legacy detector / `CONTRADICTION_PROMPT` / verdict path is untouched. So this PR provably changes nothing.

## The three layers

- **`relationship.py` (L1)** — `classify_relationship` / `extract_relationship`. Deterministic RDF exact-value short-circuit (no LLM); otherwise the detector's `CONTRADICTION_PROMPT` extended **on a copy** with a relationship instruction, plus a derivation fallback from the model's existing `non_conflict_reason` taxonomy. → one of the 7 `RELATIONSHIPS` + confidence.
- **`diagnosis.py` (L2)** — `extract_diagnosis` + the combined `classify()`. **One LLM call** yields both the relationship and the diagnosis; most diagnoses derive from `same_subject` / `non_conflict_reason`. → one of the 6 `DIAGNOSES`.
- **`resolution.py` (L3)** — `resolve()` maps `(relationship × diagnosis × evidence) → action + audit_reason`, applying the **safety invariants**: inference must not overturn an explicit fact; a probabilistic relationship softens (`downweight`); low confidence flags (`mark_disputed`).

## No-degradation

`resolve()` only **recommends** an action for the future `memory_conflicts` record — it does **not** apply any `status`/`supersedes_id` effect (that stays with the detector). So current RDF/semantic/Path-C outcomes are unchanged; `test_temporal_change_maps_to_supersede_matching_today` anchors the alignment. The golden differential test (separate) remains the gate when the write slice wires this in.

## Tests

**56 cases** — L1 21, L2 17, L3 18. mypy clean (196 files), ruff + format clean.

## Next slice (not in this PR)

The **write**: a storage endpoint + client to persist a `memory_conflicts` row and set `memories.confidence`/`is_inferred`/`scope`, wire `classify()`+`resolve()` into the engine resolver on the ON path (additive, status/supersedes untouched), + a `.dev` wet test with the flag on.

### DIFF
--- a/core-api/src/core_api/services/contradiction/diagnosis.py
+++ b/core-api/src/core_api/services/contradiction/diagnosis.py
@@ -1,0 +1,148 @@
+"""L2 — diagnosis (cause) classification + the combined L1+L2 classifier (A55).
+
+L2 answers *why* a candidate pair differs: correction / temporal_change /
+scope_difference / entity_mismatch / write_error / unresolved. It builds on L1
+(``relationship``) and the model's existing ``non_conflict_reason`` / same_subject
+signals, so most diagnoses are derivable; the model can also emit ``diagnosis``
+explicitly via the extended prompt.
+
+``classify()`` is the combined entry the engine resolver will call — ONE LLM call
+yields both the L1 relationship and the L2 diagnosis. Like L1 this is dormant
+(nothing wires it into production yet) and never mutates the legacy prompt/verdict.
+See benchmark/a55/subtasks/12-L1-relationship-logic.md (L2 section in 13-…).
+"""
+
+from __future__ import annotations
+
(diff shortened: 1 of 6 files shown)


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #761, opened 2026-08-12
fix(embedding): set per-phase httpx timeouts, matching the LLM client

Eldad-Caura:
## What

Replaces the bare-float `timeout` on the embedding provider's `AsyncOpenAI` client with an explicit `httpx.Timeout`, mirroring what `common/llm/providers/openai.py` already does.

## Why

`AsyncOpenAI(timeout=<float>)` looks like it sets the whole budget, but it leaves **httpx's default 5 s connect phase** in place. On Cloud Run with a VPC connector in `all-traffic` egress mode, outbound connects intermittently exceed 5 s — so calls failed as `httpcore.ConnectTimeout` well inside the nominal 25 s request budget.

This is not a new diagnosis. The LLM client took exactly this fix, and `common/llm/constants.py:110-113` names the two symptoms it was meant to cure:

> ...handlers ("pubsub handler raised; nacking for redelivery" + "Entity extraction failed" — the latter permanently skips entity links for that memory).

The embedding client was simply missed, so the same trickle continued there.

## Why it turned into 321 alerts

The worker's `provider.embed(...)` in `core-worker/consumer.py` is the one Pub/Sub handler that can propagate an exception — `handle_enrich_request` falls through to a heuristic and core-api's handlers use `track_task`. When it propagates, `common/events/pubsub.py:749-758` nacks with `"ack_deadline_seconds": 0`, i.e. redeliver immediately, and there is no delivery-attempt counter or dead-letter topic anywhere in the code. Staging logged 8,720 occurrences and sent **321 alert notifications** off roughly two dozen distinct failures.

Worth being precise about scope: **this bounds the per-attempt connect, not the redelivery loop.** A dead-letter topic and a retry policy on the staging subscriptions are still wanted, and by design those live in infra rather than here (`pubsub.py:7-14`). This removes the cause; that would cap the blast radius.

## Compatibility

Only connect gains headroom (15 s, env-tunable). `read` and `write` keep `OPENAI_REQUEST_TIMEOUT_SECONDS`, and `pool` tracks it unless explicitly decoupled via `EMBEDDING_HTTPX_POOL_TIMEOUT_SECONDS` — so **no existing deployment sees a shorter budget than the bare float gave it**, including one running `OPENAI_REQUEST_TIMEOUT_SECONDS=60`. Same shape and naming as the LLM-side constants.

## Test

`test_timeout_sets_connect_phase_not_bare_float` asserts the client is constructed with an `httpx.Timeout` whose connect phase beats httpx's 5 s default, and that read/write/pool still carry the request budget. It pins the two providers together so they cannot drift apart again.

Confirmed it fails without the fix:

```
E       AssertionError: expected httpx.Timeout so the connect phase is explicit, got float
```

51 passed in that module; 78 passed across the embedding suites. `ruff check` and `ruff format --check` clean on the touched files.

> Deliberately **not** included: raising `EMBEDDING_REMOTE_MAX_BATCH` from 32 to 128. I had that on my list, but the constant's own comment is right — the local `docker-compose` stack starts TEI without `--max-client-batch-size`, so TEI's own default of 32 applies and a code-default change would cause 413s there. That belongs in the prod/staging deployment env, where TEI is explicitly launched with 128.

Found while triaging the prod/staging error-alerts backlog.

### DIFF
--- a/common/embedding/constants.py
+++ b/common/embedding/constants.py
@@ -96,6 +96,27 @@ EMBEDDING_HTTPX_MAX_KEEPALIVE_CONNECTIONS: int = clamp_keepalive(
     max_keepalive_var="EMBEDDING_HTTPX_MAX_KEEPALIVE_CONNECTIONS",
 )
 
+# Per-phase connect timeout, mirroring ``OPENAI_HTTPX_CONNECT_TIMEOUT_SECONDS``
+# on the LLM side. Passing a bare float to ``AsyncOpenAI(timeout=...)`` looks
+# like it sets the whole budget but leaves httpx's DEFAULT 5 s connect phase in
+# place. On Cloud Run with a VPC connector in ``all-traffic`` egress mode every
+# outbound call intermittently exceeds 5 s, which surfaces as
+# ``httpcore.ConnectTimeout`` well inside the nominal request budget. The LLM
+# client took this fix; the embedding client kept the bare float, so the same
+# trickle continued here — visible as the staging
+# "pubsub handler raised; nacking for redelivery" loop, where the worker's
+# embed call is the only handler that can propagate. Only connect/pool get the
+# headroom; the read phase stays governed by ``OPENAI_REQUEST_TIMEOUT_SECONDS``.
+EMBEDDING_HTTPX_CONNECT_TIMEOUT_SECONDS: float = read_float_env(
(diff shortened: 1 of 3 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #762, opened 2026-08-12
CAURA-716 fix(memory): honour agent_provided_fields on the inline enrich path

eyal-bl:
## Problem

`_schedule_enrich_or_inline` computes `agent_provided_fields` — which enrichment columns the caller set explicitly, from Pydantic's `model_fields_set` — and forwards it to the deferred/worker branch. **The inline (OSS) branch dropped it**, inferring caller intent instead by comparing the stored value against the schema default:

```python
if mem.get("memory_type") == "fact"   and enrichment.memory_type: ...
if mem.get("status")      == "active" and enrichment.status:      ...
if mem.get("weight")      == 0.5      and enrichment.weight:      ...
```

The dropped forward carried a comment asserting the gates were equivalent — *"there's no observable behavioural delta between the gates today."* There is one, and it's silent.

That comparison can't distinguish **"caller pinned this to the default"** from **"caller said nothing."** And `fact` / `active` are the defaults *precisely because* they're the neutral choices — which is exactly why a caller pins them.

## Evidence (live VM, real provider)

A write passing `memory_type="fact", status="active"` INSERTs correctly, then gets rewritten seconds later:

| | `memory_type` | `status` | `weight` |
|---|---|---|---|
| at insert | `fact` | `active` | `0.5` |
| +8s | **`decision`** | **`confirmed`** | `0.9` |

`weight` is correct there — not pinned, so enrichment rightly owns it. Both pinned fields lost.

Found while wet-testing #756 (CAURA-704), where the lost `status="active"` made doc-derived memories invisible to `insights focus=patterns` (selects `status='active'` only) — the document never reached the reflection pass.

## Fix

Forward the list; gate on it when present. A small `_agent_pinned` helper routes all five override fields (`memory_type`, `weight`, `status`, `ts_valid_start`, `ts_valid_end`) through one decision so they can't drift apart again.

The value-vs-default comparison is **retained as the fallback** when no list is passed (`agent_provided_fields is None`), so behaviour is unchanged for every existing caller. `reference_datetime` is still deliberately not forwarded — the inline path resolves relative dates against the enrichment call's own clock, a separate concern.

## Scope note for reviewers

This is in the write path every memory traverses. The behavioural delta is precisely: *a caller who explicitly passed one of the five override fields now keeps their value on the inline path* — which the deferred path already does and the dropped comment already claimed. **No caller gains or loses enrichment it was previously getting.**

## Verification

- **14 new tests. 10 fail against pre-fix source**; the other 4 assert deliberately-unchanged behaviour (legacy fallback, deferred path, the `model_fields_set` mechanism) and pass on both sides.
- Full suite: **526 failed / 3688 passed** here vs **526 failed / 3674 passed** on clean `main` — identical failure count (pre-existing local DB schema drift), **+14 passing**.
- **0 new mypy errors** (343 both sides). `ruff check` / `format` clean.

### DIFF
--- a/core-api/src/core_api/services/memory_service.py
+++ b/core-api/src/core_api/services/memory_service.py
@@ -1914,22 +1914,33 @@ async def _schedule_enrich_or_inline(
     those columns on every redelivery.
     """
     if settings.inline_enrichment:
-        # NOTE: ``agent_provided_fields`` and ``reference_datetime``
-        # are intentionally NOT forwarded to ``_enrich_memory_background``.
-        # The inline path pre-dates these concepts and uses an
-        # equivalent gate by reading the row's current value vs the
-        # schema default (``if mem.get("memory_type") == "fact" and
-        # enrichment.memory_type:`` etc., see the body of
-        # ``_enrich_memory_background``). When the agent set
-        # ``memory_type="rule"`` the column already reads ``"rule"``
-        # not ``"fact"``, and the inline gate skips.
+        # CAURA-716: ``agent_provided_fields`` IS now forwarded here.
         #
-        # The two paths converge on the same agent-wins outcome via
-        # different mechanisms; unifying them on
-        # ``agent_provided_fields`` is a reasonable cleanup but out of
-        # scope for CAURA-595 — there's no observable behavioural
-        # delta between the gates today.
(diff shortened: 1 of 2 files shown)


eyal-bl:
@claude

Eldad-Caura (approved):

erni-a (approved):

PR #764, opened 2026-08-12
feat(memory): conflict-record write machinery — storage + resolver (A55 1d)

arkash20:
## What & why

The machinery to persist the L1/L2/L3 classification as a `memory_conflicts` row. **Additive and dormant** — nothing calls the resolver in production yet (the detector wiring is a following slice behind its own flag), and it writes only the classification **record**; the memory-row `status`/`supersedes_id` effect stays with the detector, unchanged. The golden differential test is unaffected.

## Storage
- `postgres_service.memory_conflict_record` — insert a `memory_conflicts` row, validating enum-like fields against the model vocab (bad value → **400**, not a DB CHECK 500).
- `POST /memories/conflicts` + `_MEMORY_CONFLICT_FIELDS`.
- `storage_client.record_memory_conflict`.

## Resolver — `services/contradiction/resolver.py`
`record_conflict(new_memory, candidate, …)`:
1. `classify()` (L1 relationship + L2 diagnosis)
2. `resolve()` (L3 action + audit_reason; `is_inferred` from either memory gates destructive actions)
3. build payload (+ `evidence_strength`) → `record_memory_conflict()`

Best-effort: a storage failure returns `None` (logged), never breaks the write path.

## No-degradation
Record-only — no `status`/`supersedes_id` mutation. The RDF-pair test asserts `exact_value/temporal_change → supersede`, matching today's RDF effect.

## Tests
Storage round-trip + bad-enum 400; resolver RDF→supersede, LLM→classification, inferred→invariant downgrade, storage-failure swallowed. **Storage suite 118 passed; core-api contradiction suite 71 passed.** mypy + ruff clean.

## Next (final sub-slice, not here)
Wire `record_conflict` into the detector where it already holds both memories (RDF/semantic/Path-C), behind a new `contradiction_write_conflict_record` flag (default OFF), avoiding the double-LLM on the semantic path; new test that flag-ON writes a row + golden proves status/supersedes unchanged; then a `.dev` wet test.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #765, opened 2026-08-12
fix(crystallizer): match the /lifecycle-candidates contract; drop inline remediation

Eldad-Caura:
Follow-up to #763, which fixed the same class of bug against a different endpoint.

## Scope note

The task was to fix `_remediate_missing_embeddings`. While verifying it I found that **four** crystallizer consumers of `GET /memories/lifecycle-candidates` disagree with it — each in a different way, and each failing toward a reassuring value. Three are fixed here. The fourth needs a storage-side change and is tracked separately (see the end).

I fixed the three rather than leaving known-broken lines beside the one I was sent to fix; a separate PR would touch the same 25 lines and a reviewer would rightly ask why.

## The endpoint

`core-storage-api/.../routers/memories.py:812-819` returns exactly two keys, both lists of bare UUID **strings**:

```python
return {
    "expired_still_active": [str(r[0]) for r in expired],
    "stale_low_weight": [str(r[0]) for r in stale],
}
```

## The consumers

| Function | Read | Result | This PR |
|---|---|---|---|
| `_check_stale_memories` | `stale_memories` | wrong key → **0 for every tenant** | fixed |
| `_check_expired_still_active` | right key, but `r.get("id")` on strings | **`AttributeError`** whenever a row exists | fixed |
| `_remediate_missing_embeddings` | `missing_embeddings` | key absent → **permanent no-op** | deleted |
| `_check_short_content` | `short_content` | key absent → **0 for every tenant** | tracked separately |

Verified by probe, not by reading — against the endpoint's exact return shape:

```
_check_expired_still_active: RAISED AttributeError: 'str' object has no attribute 'get'
_check_stale_memories:       {'count': 0, 'affected_ids': []}
_check_short_content:        {'count': 0, 'affected_ids': []}
_remediate:                  update_embedding never called — no-op confirmed
```

The expired-check failure is the nastiest: `run_crystallization` catches per-check exceptions into `hygiene[name] = {"error": True}`, so that check **errored precisely when it had something to report** and looked healthy when it didn't.

## The decision: delete rather than repair the remediation

Repair already exists, and is better than anything this function could become. `core_worker.backfill` pages `GET /memories/null-embedding-ids` — keyset-paginated, tenant-scoped, deliberately ids-only as defence-in-depth on the unauthenticated storage API — and publishes one `EMBED_REQUESTED` per row, inheriting the consumer's per-tenant concurrency, retry and DLQ wiring, idempotent under restart.

This corrects a premise in the task: a row-level source **does** exist (`memory_list_null_embedding_rows` behind that endpoint). That makes the "add an endpoint" option cheaper than assumed — and makes deletion clearly right, because the machinery is already built and well-designed. Reimplementing it synchronously would duplicate all of it *and* issue inline embedding calls into the same process-wide gate that write traffic already oversubscribes, so a hygiene report could stall the write path it reports on.

Surfacing the count is this service's job — `_check_missing_embeddings`, fixed in #763. Acting on it is `python -m core_worker.cli backfill-embeddings --tenant-id X`.

**Nothing claimed the remediation worked.** One caller, no report field, and no doc or frontend mentions it (searched `*.py`, `*.ts`, `*.tsx`, `*.md` across the project tree; every other "remediat" hit is the unrelated `governance_remediation`). So there is no stale promise to retract. `update_embedding` keeps two callers in `memory_service`, so the deletion orphans nothing.

Worth stating plainly: after this there is **no automated repair** for rows that fell through — but there never was, since the function was a no-op. The right follow-up is scheduling the worker backfill, not resurrecting inline remediation.

## Tests

Three regressions, each confirmed failing before the change:

```
test_expired_check_counts_string_ids_without_raising  AttributeError: 'str' object has no attribute 'get'
test_stale_check_reads_stale_low_weight               assert 0 == 1
test_inline_remediation_is_gone                       assert not True
```

Plus a **contract pin** asserting the endpoint's key names *and* that it yields `str(r[0])` rather than dicts. That pin is what makes this class non-recurring, and it is deliberately not a mock-based test: a hand-written mock proves nothing here, because whoever wrote the original mocks would have written them from the same wrong assumption as the callers. It correctly passes both before and after — it is a guard, not a regression.

320 passed across the crystallizer, lifecycle and embedding selections. `ruff check` and `ruff format --check` clean. Rebased onto `main` after #764 landed, with tests re-run after the rebase.

## Deliberately not in this PR: `_check_short_content`

Same bug, but it cannot be fixed on the core-api side alone. The query it needs already exists and is **orphaned** — `memory_find_short_content` (`postgres_service.py:2792`) has no caller anywhere in the repo, so someone wrote it for this check and never wired it up. Fixing it means adding `short_content` to the `/lifecycle-candidates` response and moving `CRYSTALLIZER_SHORT_CONTENT_CHARS` into `common/constants.py` so both services can share it — a cross-service change with its own review surface, kept separate rather than smuggled in here. Note it carries the same `r.get("id")`-on-strings shape bug, so it would raise rather than count once data flows.


Eldad-Caura:
@claude please re-review — force-pushed to complete the fix.

The review correctly caught that `_check_short_content` had the same dead-key bug and that the PR was incomplete. It is now fixed in the same change rather than deferred.

Taking option 1 (wire it up) rather than option 2 (remove the check), because the query it needs already existed and was **orphaned**: `memory_find_short_content` (`postgres_service.py:2792`) had no caller anywhere in the repo — written for this check and never exposed. Removing the check would have discarded working server-side logic.

Changes since your review:
- `GET /memories/lifecycle-candidates` now returns `short_content` alongside the two sibling lists, so the caller stays at one round-trip.
- `_check_short_content` reads that key and treats the values as bare UUID strings, fixing the same `r.get("id")` shape bug you saw in the expired check.
- `CRYSTALLIZER_SHORT_CONTENT_CHARS` moved to `common/constants.py` (still re-exported from `core_api.constants`, so no importer changed) — both services must agree on the bound, since core-api rejects shorter writes at the quality gate and storage now lists existing rows below it.
- Two more tests: the short-content regression (fails with `assert 0 == 1` before), and one asserting the threshold is shared rather than duplicated. The contract pin now also requires `"short_content"` in the endpoint source, so it failed before this change too.

All four consumers of that endpoint now match it. Ran the exact CI gates (`ruff check` on the four paths, `ruff format --check` on the three) plus the crystallizer, lifecycle, quality-gate and reports suites — 165 passed. Rebased onto main after #766.

Note the earlier CI failure on this branch was an unrelated flake in `tests/test_reports.py::test_report_internal_group_durable_filter` — it passes locally on both this branch and main, and passed on CI re-run with no code change.

Eldad-Caura:
On the remaining Low finding — *automatic remediation removed with only a manual replacement*. Fair concern, and worth answering with evidence rather than waving at it. Three points:

**1. Nothing automatic was lost, because it never ran.** The deleted function read `candidates.get("missing_embeddings", [])` from an endpoint that returned only `expired_still_active` and `stale_low_weight`, so `missing` was always `[]` and it returned at the guard clause on every single pass. Confirmed by probe before deleting — `update_embedding` was never called. So there is no behavioural regression here: the "always-on" path was always-off.

**2. The common case already self-heals through the queue, not through this function.** `schedule_background_tasks.py:121-126` publishes `EMBED_REQUESTED` whenever `embedding is None` in deferred mode (retrying in-process when `inline_embedding` is True). That is exactly why prod logs *"Storing memory without embedding; deferred backfill scheduled"* — 322 times in the last 48h. Transient provider failures on the normal write path are already retried, with the consumer's retry and DLQ behind them. The reviewer's *"gaps from transient provider failures may now persist indefinitely"* would be true if that publish didn't exist; it does.

**3. The real gap is narrower and elsewhere.** What genuinely never gets scheduled is three specific paths that bypass the publish entirely: the auto-chunk parent (`memory_service.py:480-598` and `:742-867`), `update_memory` on a content change (`:2597-2601`, which keeps the *previous* content's vector — non-NULL, so invisible to any NULL-based backfill), and the atomic-fact fan-out (`:2412`, silent `continue`). Those are real and tracked separately; resurrecting a no-op wouldn't have caught them either, since it read a key that didn't exist.

Where I agree: there is no *scheduled* sweep for rows that fell through, and there should be. That is a Cloud Run job or cron around the existing `core_worker.backfill` — infrastructure, with its own review surface — rather than something to add to this diff. Spun out as follow-up work.

CI is green (8 pass, 0 fail). The earlier failure on this branch was an unrelated flake in `tests/test_reports.py::test_report_internal_group_durable_filter`: it passes locally on both this branch and main, and passed on CI re-run with no code change.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #766, opened 2026-08-12
feat(memory): wire conflict-record write into the detector, flag-gated (A55 1d)

arkash20:
## What & why

Completes the **conflict-record side of A55 1d**: when `contradiction_write_conflict_record` is **ON** (default **OFF**), `_detect` additionally persists a `memory_conflicts` classification record for each confirmed conflict. **Additive** — it never touches the `status`/`supersedes_id` effect, so retrieval is unchanged.

## How

- **config:** `contradiction_write_conflict_record` (default `False`).
- **`_detect`:** collects `(candidate, kind, confidence)` pairs at the RDF + semantic confirm sites and — behind the flag, *after* the effect is applied — calls `record_detected_conflicts` (deferred import to avoid the detector↔resolver import cycle).
- **resolver:**
  - `record_conflict(..., result=…)` skips re-classification when the caller already has it.
  - `record_conflict_from_verdict` records a detector-confirmed conflict **without a second LLM call** — aligned to the supersede effect (`exact_value` / `temporal_change` / `supersede`).
  - `record_detected_conflicts` dispatches `rdf` → classify (deterministic, no LLM) vs `semantic` → verdict.

## No-degradation

- Flag-ON test: records the RDF pair **and** preserves the `outdated`+`supersedes` effect.
- Flag-OFF test: records nothing but still applies the effect (identical to today).
- Broad contradiction sweep: **497 passed**, only the known pre-existing create_all `ph5b` failure — **zero new failures**. Golden differential unchanged. mypy + ruff clean.

## Validation still to do (with the flag)

A `.dev`/local **wet test with both flags ON** — write two contradicting memories, confirm a `memory_conflicts` row lands and status/supersedes are unchanged. Belongs to the flag-rollout step.

## Explicitly out of scope (follow-ups)

- Path C (entity-aware) record wiring + richer semantic-conflict classification.
- Populating `memories.confidence`/`is_inferred`/`scope` → **A59**.

### DIFF
--- a/core-api/src/core_api/config.py
+++ b/core-api/src/core_api/config.py
@@ -297,6 +297,11 @@ class Settings(BaseSettings):
     # paths are behaviourally identical (the engine delegates to the same
     # detector); the flag exists so the old arch can be retired later.
     contradiction_engine_enabled: bool = False
+    # A55 1d — when True, the detector additionally writes a memory_conflicts
+    # classification record for each confirmed conflict (via the resolver).
+    # Additive: it never changes the status/supersedes effect, so retrieval is
+    # unaffected. Default False; enable to start populating conflict records.
+    contradiction_write_conflict_record: bool = False
     crystallizer_enabled: bool = True
     crystallizer_stale_days: int = 180
     crystallizer_dedup_sample_size: int = 1000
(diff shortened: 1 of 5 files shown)


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #767, opened 2026-08-12
feat(embeddings): nightly sweep that re-embeds NULL-embedding rows

Eldad-Caura:
Closes the gap #765 left behind. That PR deleted `_remediate_missing_embeddings` — correctly, since it read a key its endpoint never returned and had never re-embedded a row — but the review rightly noted it left no periodic self-healing path.

## What actually needed fixing

Normal writes already recover on their own: `schedule_background_tasks` publishes `EMBED_REQUESTED` whenever the embedding is `None`, which is what produces the ~322/48h *"Storing memory without embedding; deferred backfill scheduled"* lines in prod. So the gap was never "transient provider failures go unretried" — it is **rows that never got that publish at all**, with nothing periodically looking for them.

## Reuses the existing machinery

`core_worker.backfill` already does this job properly, so nothing here reimplements it. It pages `GET /memories/null-embedding-ids` (keyset-paginated, tenant-scoped, deliberately ids-only as defence-in-depth on the unauthenticated storage API), publishes one `EMBED_REQUESTED` per row so the provider calls inherit the consumer's per-tenant concurrency, retry and DLQ wiring, and is idempotent under restart — a redelivery re-queries and sees only rows still NULL.

The shape follows the established lifecycle cron pattern exactly:

```
core-operations tick (wall-clock aligned)
  -> POST /admin/lifecycle/fanout/embed-backfill
    -> one message per active org
      -> core-worker consumer -> run_embedding_backfill(tenant_id=org)
```

Per-org is not a preference — the storage endpoint refuses un-scoped calls. And because the page loop runs in the consumer rather than inside the HTTP request, it is not bounded by the 120s Cloud Run budget that the nightly cross-link call currently runs into.

## Two decisions worth reviewing

**Registered in core-worker, not through `common.events.lifecycle_handlers`.** The shared adapter protocol is implemented by core-api as well, and core-api cannot run the worker's backfill — routing this through there would have meant adding a method core-api could only `raise NotImplementedError` on. The cost is ~15 lines of audit bookkeeping this handler does itself instead of inheriting from `_run_action`.

**Gated on `embed_backfill_enabled`, default OFF.** The topic, durable subscription and DLQ are Terraform-provisioned — the bus only auto-creates *broadcast* subscriptions — so until infra lands, a nightly fire would publish into a topic nothing consumes. Gating the *registration* rather than no-opping inside the tick also keeps `scheduler.is_healthy` meaningful: an unregistered task has no runtime slot that could look dead.

**→ To turn this on:** provision `memclaw.lifecycle.embed-backfill-requested` + subscription + DLQ, then set `EMBED_BACKFILL_ENABLED=true` on core-operations.

## Rate and timing

Default hour is **04:00, not 02:00**. Every other lifecycle tick defaults to 02:00, and that window is already congested enough that the cross-link call hits the request ceiling nightly — while this sweep feeds the same capacity-constrained embedding backend those ticks compete for. `max_inflight` is 25 rather than the library default of 100, since the sweep shares `EMBED_REQUESTED` with live deferred writes and a large backlog in front of them delays real traffic.

## Observability

The handler logs `scanned` / `published` / `skipped_missing` per org and writes them to the lifecycle audit row. Worth stating plainly: **"how many rows are unembedded" was previously unanswerable** without connecting to AlloyDB directly, because the coverage endpoint is internal-ingress only and no metric carried the number.

It also finalises the audit row on **both** paths. The fanout pre-creates it as `pending` specifically so that a row which never advances reads as a publish failure; leaving it pending after a *consumer* failure would make a failed sweep indistinguishable from an undelivered message. The failure path records `failed` with the error and re-raises, so Pub/Sub still redelivers.

## What this does NOT fix

So it is not mistaken for full coverage: `update_memory` on a content change keeps the **previous** content's vector (`memory_service.py` ~:2597). Those rows are non-NULL, so a NULL-based sweep cannot see them — mis-embedded rather than unembedded, and they need their own fix. The auto-chunk parent and atomic-fact fan-out paths do land as NULL and are caught here.

## Tests

Six in `tests/test_embedding_backfill.py` (CI-gated), each confirmed failing before the change:

- sweeps the event's `org_id`, with the conservative `max_inflight`
- logs the per-org counts
- ack-drops a malformed payload instead of nack-looping it to the DLQ
- the topic is actually subscribed in `register_consumers` — a handler defined but never subscribed is silently orphaned, since the Pub/Sub backend spawns pull loops from the registry as it stands at `start()`
- finalises the audit row `success` with stats, and `failed` + re-raise on error
- the action is registered in `_ACTION_PUBLISHERS` (an unregistered action 404s and the tick just logs-and-returns, so the sweep would silently never run) and the topic string is pinned, since it is infra's contract too

Plus four in `core-operations/tests/` for the registration gating and the fanout path, and a `docs/component-registry.yaml` entry — the AST drift gate in `tests/test_component_registry.py` caught the missing entry, exactly as designed.

**4,295 passed** in the root suite, 44 in core-operations. `ruff check` clean on all five packages; `ruff format --check` clean on the three CI-gated paths.

> Heads-up: `core-operations/tests/` is not run by CI at all today — those four tests are ungated, along with the 44 existing ones covering the whole cron-scheduling surface. Tracked separately; it is a one-step workflow addition and they all pass.


Eldad-Caura:
@claude please review — the previous run never produced a verdict.

The `claude-review` job failed with a transient `API Error: 529 Overloaded` (run 31605563024) rather than on anything in the diff, so this PR currently has no review. `CI` itself passed. Re-requesting a real verdict.

Eldad-Caura:
@claude

Eldad-Caura:
@claude please re-review — all four findings from the two reviews are fixed.

**From the first review:**
- **High, invalid audit status.** Confirmed: `_VALID_STATUSES = {"in_progress", "success", "failure"}` in `core-storage-api/.../routers/lifecycle_audit.py:21`, and the shared `_run_action` uses `"failure"`. Fixed. Worth naming why my test missed it: it asserted `"failed"` — the same wrong literal the handler used — so it passed while the real PATCH would have 422'd. It now pins against the route's own allow-list rather than a string I chose.
- **Medium, `fleet_id` silently ignored.** Confirmed: `run_embedding_backfill` has no fleet parameter and `/memories/null-embedding-ids` filters by tenant only, while the manual trigger route does forward `fleet_id`. Now **refused** rather than plumbed — proceeding would re-embed a whole org for an operator who asked for one fleet, and silently exceeding a requested blast radius on a write-generating path is the worse failure. Plumbing fleet through would mean changing the storage endpoint and service method too, which is a feature rather than a fix. The scheduled fanout never sets `fleet_id`, so only deliberate manual triggers reach it.

**From the second review (both new, not duplicates):**
- **Medium, unguarded failure-path PATCH.** Correct, and it is the structural version of the first review's finding — I had fixed the 422 trigger but left the hazard. The PATCH is now guarded so a storage blip cannot replace the sweep's exception with its own; the real cause propagates and the bookkeeping failure only logs.
- **Low, missing `in_progress` transition.** Added, matching `_run_action`. A per-org sweep is not instant, so without it an operator cannot distinguish "actively sweeping" from "never picked up".

Four new tests, each confirmed failing before its fix: the fleet refusal (asserts the sweep is not awaited), the mask-prevention (asserts the sweep's error propagates when the PATCH also raises), and the `in_progress` → `success` sequence. The two pre-existing audit tests were updated for the extra PATCH and now assert on the finalising call.

340 passed across the embedding/backfill/lifecycle/crystallizer/registry selections. `ruff check` and `ruff format --check` clean. Rebased on main.

Note the `claude-review` job skips on `synchronize`, which is why this is a tag rather than a silent re-push.

Eldad-Caura:
@claude please re-review — both findings and the suggestion are addressed.

**Medium, manual trigger bypasses the infra gate and fails silently.** Verified the crux rather than taking it on faith: `PubSubEventBus.publish` carries an explicit comment that it fires into the client's batch queue and does **not** block on the returned future, so a "topic not found" reaches only the SDK's background thread. A trigger before infra lands really does return 200 with an `audit_id` whose row sits at `pending` forever.

I did **not** add a second gate in core-api, and want to be explicit about why: `embed_backfill_enabled` is a core-operations setting, so mirroring it here would create two sources of truth for "is this on" that can skew — the failure mode being an operator who flips one and reasonably concludes the feature is live. Instead the precondition is now named where someone would actually look, at the action registration, including the required order (provision topic + subscription + DLQ, *then* flip the flag) and why the error is invisible. Happy to add a hard gate if you'd rather; it just seemed worse than documenting a window that closes when Terraform lands.

**Low, stale registry topic count.** Correct — `topics: 7` against 8 in code, and the inline list omitted `embed-backfill-requested`. Both fixed. Worth noting this one has no gate: `tests/test_component_registry.py` drift-checks `cron_ticks` but not `event_topic_families`, which is why the count could go stale silently. I left that asymmetry alone rather than widen the gate mid-PR.

**Suggestion, hard-coded `EMBED_BACKFILL_MAX_INFLIGHT`.** Taken — it is the kind of knob you want during an incident without a deploy. Now `settings.embed_backfill_max_inflight` on `core_worker.config.Settings`, default 25, with a validator rejecting 0 (which would make the publish semaphore block forever — a hang rather than an error, the same trap `per_tenant_storage_write_concurrency` already guards). New test confirms it raises; verified failing without the validator.

392 passed across the embedding/backfill/lifecycle/crystallizer/registry/consumer selections. `ruff check`, `ruff format --check` and `mypy` all clean on core-worker and core-api. Rebased on main.

Eldad-Caura:
@erni-a pleasr approve

erni-a (approved):

PR #768, opened 2026-08-12
ci: run and lint core-operations, closing an ungated tree

Eldad-Caura:
`core-operations/tests/` was never run by CI. The workflow had exactly two pytest steps — `uv run pytest tests/` and `uv run pytest core-storage-api/tests/` — and a repo-wide grep for "core-operations" across `.github/workflows/` returned nothing.

That left the whole cron-scheduling surface ungated: **41 tests** across `test_app.py`, `test_tasks.py` and `test_scheduler.py`, covering wall-clock alignment, per-job `*_run_at_hour` threading, the fanout POST paths, and the scheduler's failure/hot-loop guards.

This is the exact failure mode the `core-storage-api/tests/` ruff step already documents:

> The test tree was unlinted for as long as it was un-run; both halves of that blind spot close together or it just drifts again.

So both halves close here: `ruff check`, `ruff format --check`, and pytest.

## Three things that made this smaller than the task assumed

**No `PYTHONPATH` needed.** I expected to add `core-operations/src` to a path somewhere. It turns out `core-operations/pytest.ini` already sets its own `pythonpath = src ..` (`src` for `core_operations`, `..` for `common`), and pytest resolves *that* file as the configfile when invoked with this path — it does **not** inherit the root `pytest.ini`. Confirmed from pytest's own header:

```
rootdir: .../memclaw/core-operations
configfile: pytest.ini
```

Setting `PYTHONPATH` on the step would have been inert, so it is deliberately absent and the comment says why. (For contrast, the root `tests/` step gets `core_worker` from the root `pytest.ini`'s `pythonpath`, not from its `env:` block — that env var is for the migration/openapi steps.)

**The format check is safe to gate.** The task flagged a risk that `core-operations` might have no ruff config and fall back to 88 columns, demanding a reflow. It has its own: `core-operations/pyproject.toml` declares `line-length = 110`, matching core-api and core-storage-api rather than `common/`'s 88. Both `src/` and `tests/` are already clean under it (`9 files already formatted`), so `format --check` is included rather than deferred.

**No database.** These tests touch none, so nothing is added to the Postgres setup. Postgres is already up for the earlier steps in that job, but this step uses it not at all and runs in well under a second — cheaper than a separate job that would re-do the uv install.

## The gate is not decoration

I checked it actually catches the regression it exists for. Dropping `delay_provider` from one registration — which would make that job fire an immediate boot-time tick and drift:

```
E   AssertionError: lifecycle-insights is not wall-clock aligned
FAILED core-operations/tests/test_app.py::test_all_lifecycle_jobs_registered_and_wall_clock_aligned
1 failed, 40 passed
```

Restoring it returns to `41 passed`.

## Verification

Each new step's command run verbatim in a clean environment (`env -u PYTHONPATH`), mirroring CI:

| Step | Result |
|---|---|
| `ruff check core-operations/src/ core-operations/tests/` | All checks passed! |
| `ruff format --check core-operations/src/ core-operations/tests/` | 9 files already formatted |
| `pytest core-operations/tests/ -v --tb=short` | 41 passed in 0.36s |

`actionlint` reports **one** finding on this file — SC2086 at line 70, the pre-existing `Install dependencies` step — both before and after this change, so these additions introduce none. Worth noting: CI does not currently run actionlint on this repo, which is why that one has survived.

YAML validated by parse, and the three steps confirmed present in the job with no empty `env:` blocks.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,10 +67,17 @@ jobs:
         run: uv python install ${{ env.PYTHON_VERSION }}
 
       - name: Install dependencies
+        # core-operations is installed for its own ruff/mypy/pytest steps
+        # below. Before it was listed here those steps passed only because
+        # core-api's dependency list happens to be a superset of
+        # core-operations' runtime deps — an undeclared coupling that would
+        # have broken confusingly the first time core-operations grew a
+        # dependency core-api lacks. Resolution cost of adding it: one extra
+        # package (119 vs 118).
         run: |
           uv venv .venv
           source .venv/bin/activate
-          uv pip install -e core-storage-api/[dev] -e core-api/[dev]
+          uv pip install -e core-storage-api/[dev] -e core-api/[dev] -e core-operations/[dev]
           echo "$VIRTUAL_ENV/bin" >> $GITHUB_PATH
 
       # --- Broker↔cloud API-versioning gate (Gate C / Phase 1, deliverables A+B) ---
@@ -149,6 +156,13 @@ jobs:
         # because the config is byte-identical across both repos (#696).
         run: uv run ruff check common/
 
+      - name: Run Ruff linter (core-operations)
+        # Paired with the test step below: an unlinted tree drifts, and this one
+        # was both unlinted and unrun. core-operations/pyproject.toml declares
+        # its own line-length = 110, matching core-api and core-storage-api
+        # rather than common/'s 88, so the format check below is safe to gate.
+        run: uv run ruff check core-operations/src/ core-operations/tests/
+
       - name: Run Ruff formatter check (core-api)
         run: uv run ruff format --check core-api/src/
 
@@ -157,6 +171,9 @@ jobs:
 
       - name: Run Ruff formatter check (core-storage-api tests)
         run: uv run ruff format --check core-storage-api/tests/
+
+      - name: Run Ruff formatter check (core-operations)
+        run: uv run ruff format --check core-operations/src/ core-operations/tests/
 
       - name: Run Ruff formatter check (vendored common/ files at ruff defaults)
         # This list mirrors the policy=identical entries of caura-memclaw-enterprise's
@@ -196,6 +213,12 @@ jobs:
 
       - name: Run mypy (core-storage-api)
         run: cd core-storage-api && uv run mypy src/
+
+      - name: Run mypy (core-operations)
+        # core-operations/pyproject.toml carries a full [tool.mypy] block and
+        # mypy as a dev dep, same as its two siblings above; only the CI step
+        # was missing.
+        run: cd core-operations && uv run mypy src/
 
       - name: Enable pgvector extension
         run: PGPASSWORD=changeme psql -h localhost -U memclaw -d memclaw -c "CREATE EXTENSION IF NOT EXISTS vector;"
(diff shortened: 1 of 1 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@claude please re-review — both findings fixed.

**Medium, no mypy step.** Correct and worth having: core-operations declares a full `[tool.mypy]` block and mypy as a dev dep, mirroring both siblings, and only the CI step was missing. Added `Run mypy (core-operations)` next to the other two. It passes clean on 5 source files, and I confirmed it is not decoration — widening a tick's return type to `int` fails it with a missing-return error plus an incompatible-argument error at the `scheduler.register` call.

**Low, core-operations never installed / incidental dependency overlap.** Also correct, and the sharper of the two — it is the kind of coupling that works until it doesn't, then fails confusingly. `core-operations/[dev]` is now installed alongside its siblings, so the dependency is declared rather than inherited by luck. I verified the three packages co-resolve before changing it: **119 packages with core-operations against 118 without**, so exactly one addition and no conflict. The install step now carries a comment explaining why it is load-bearing for the three steps below.

Unchanged and still verified: `actionlint` reports one finding on this file (SC2086 at line 70, the pre-existing install step) both before and after, so these additions introduce none; the test step still needs no `PYTHONPATH` because `core-operations/pytest.ini` supplies `pythonpath = src ..` and pytest resolves that as the configfile for this path; and dropping `delay_provider` from a registration still fails the test step with "lifecycle-insights is not wall-clock aligned".

Worth noting for this PR specifically: `claude-review` skips on a `.yml`-only diff via the workflow's skip-paths filter, which is why the original open produced no verdict and why this is a tag.

Eldad-Caura:
@claude please re-review — both findings fixed, and the coverage one turned out sharper than it looked.

**Low, coverage floor declared but never enforced.** Correct, and my first attempt at fixing it *reproduced the same bug one level up*. Adding `--cov` alone collects coverage but still does not enforce the floor: coverage discovers config relative to the CWD and there is no root `pyproject.toml`, so `fail_under = 50` in `core-operations/pyproject.toml` stays ignored when pytest runs from the repo root. The step now passes `--cov-config=core-operations/pyproject.toml`.

Verified both directions rather than trusting the output: with the flag the run prints `Required test coverage of 50.0% reached. Total coverage: 79.63%`, and temporarily raising the floor to 99 fails with `Coverage failure: total of 80 is less than fail-under=99`. Without the flag, neither line appears at all — which is exactly how this went unnoticed.

**`core-operations` missing from the build step.** Added; verified it builds clean (`core_operations-1.0.0` wheel + sdist).

**One thing you should know, because it is bigger than this PR.** The same defect affects the two siblings. The main "Run tests" step also runs from the repo root with `--cov=core-api/src/core_api --cov=core-storage-api/src/core_storage_api` and no `--cov-config`, and both packages declare `fail_under = 50` — so **neither floor is enforced either**. Demonstrated: running 3 tests with those exact `--cov` flags passes with no "Required test coverage" line, when coverage across all of core_api from 3 tests is obviously far below 50%.

I did not fix that here on purpose. One pytest run covers two packages, so it needs either a root coverage config (which the repo currently lacks *deliberately* — `common/ruff.toml` exists because there was no root config) or a split into two steps, and turning enforcement on might reveal a floor that is currently unmet. Silently lowering the number to make CI pass would be worse than the status quo, so it wants a measured decision rather than a drive-by. Tracked separately with the measurement steps written down.

`actionlint` still reports only the pre-existing SC2086 at line 70, before and after.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #770, opened 2026-08-12
perf(contradiction): batch the semantic per-candidate LLM judge (A61)

arkash20:
## The cost bug

The semantic contradiction path fired **one `complete_json` (gpt-nano) call per candidate**, up to the candidate cap (20). So each newly-embedded memory triggered ~20 nano judge calls — and prod (~442 memories/hr, ~86% one tenant's own agent fleet) was making **~10.8k OpenAI calls/hr** on only ~56 searches + 67 writes (~90 calls/op) → **~$250 every 2–3 days**. Not a loop — the *designed* per-candidate fan-out × memory volume.

## The fix — one batched judge call

- **`_llm_contradiction_check_batch`**: a single `complete_json` listing the NEW statement + all candidates; returns one raw judgment per candidate, **aligned to input order**; a missing/malformed entry defaults to a **safe non-contradiction** (never fabricates or drops one).
- **`_detect` Path 2**: a single candidate keeps the direct per-candidate call; **≥2 candidates (the prod cost driver) go through the batched call**. Both feed the **same** `_judge_contradiction` gates (Gate 1 cross-subject, Gate 2 non_conflict_reason, `_pick_older` direction) → verdict/status/supersede semantics unchanged (the golden differential test still passes).
- The batched prompt also emits per-candidate **`relationship`/`diagnosis`**, so the A55 engine-ON conflict-record write can reuse it with **no second LLM call** — **one call serves both** the flag-off (legacy detector) and flag-on (engine) arms.

**~N× fewer LLM calls per memory (up to ~20×).** Cheaper stopgap if needed: drop the candidate cap 20→5. Key-splitting is attribution only, not cost.

## No-degradation evidence

- **Deterministic:** 169 contradiction tests pass, incl. the **golden differential** (status/supersede byte-identical). Batching changes only *how raws are produced*, not the gates.
- **Real-LLM wet test:** seeded 4 similar "Dan lives in Tel Aviv…" facts, wrote the contradicting "Dan lives in Haifa now." Logs show the batch path exercised (`candidates_initial=2/3`); detection fired correctly — all 4 Tel Aviv → `conflicted`, Haifa → `supersedes_id` set. Same effect, one batched call.
- mypy + ruff clean.

A rigorous accuracy A/B (batch vs per-call over a labeled corpus) is regression/STALE-bench territory (A56); the local bench is too noisy run-to-run to certify small deltas from a single pair of runs.

Refs: **A61** (shared task list).

### DIFF
--- a/core-api/src/core_api/services/contradiction_detector.py
+++ b/core-api/src/core_api/services/contradiction_detector.py
@@ -475,15 +475,19 @@ async def _detect(
             len(candidates) if candidates else 0,
         )
         if candidates:
-            # Fire all LLM checks concurrently instead of serially
-            tasks = [
-                asyncio.wait_for(
-                    _llm_contradiction_check(content, c.get("content", ""), tenant_config),
-                    timeout=10.0,
+            # A61 — batch the LLM judge. A single candidate keeps the direct
+            # per-candidate call; multiple candidates (the prod cost driver, up
+            # to 20) are judged in ONE batched call instead of N. Both feed the
+            # SAME _judge_contradiction gates below. ``judged`` is
+            # (verdict, confidence, raw|None).
+            if len(candidates) == 1:
+                _v, _c = await _llm_contradiction_check(
+                    content, candidates[0].get("content", ""), tenant_config
                 )
-                for c in candidates
-            ]
-            results = await asyncio.gather(*tasks, return_exceptions=True)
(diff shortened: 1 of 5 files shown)


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #771, opened 2026-08-12
perf(contradiction): batch the Path C entity-aware per-candidate LLM judge (A61)

arkash20:
## What

Finishes **A61** (the OpenAI-cost bug). #770 batched the Path A semantic per-candidate LLM fan-out — the measured cost driver. This PR does the same for the remaining fan-out: **Path C's forward entity-overlap detection**.

Path C's forward loop issued **one LLM judge call per candidate**:
- `_llm_entity_aware_contradiction_check` for candidates with resolved entity context on both sides,
- `_llm_contradiction_check` (base) for the rest,

fanning out up to `_ENTITY_LINKS_DETECTION_FETCH_MAX_CANDIDATES` calls per newly-written memory. For a popular subject that is a large per-write cost — the same shape #770 fixed on Path A.

## How

Collapse the fan-out into **at most two batched calls**:
- a new `_llm_entity_aware_contradiction_check_batch` covering the entity-aware group, each candidate carrying its **own** resolved entity context so the authoritative same-subject grounding is preserved per candidate;
- the existing #770 `_llm_contradiction_check_batch` for the base group.

Everything downstream is unchanged:
- judge-selection logic (entity-aware vs base) is byte-for-byte the same;
- each raw judgment runs through `_judge_contradiction`, so the safety gates are identical to the per-candidate path;
- `results` stays aligned to `candidates` and keeps the `(verdict, confidence)` / `Exception` shape the existing loop handles — a whole-batch failure defaults its group to *skipped*, never fabricated;
- a **single candidate keeps the direct call** (mirrors the Path A `len == 1` special case).

Path C forward detection is **status-only** (no engine conflict-record wiring — that lives in Path A), so behavior is identical whether `contradiction_write_conflict_record` is on or off. Serves both the flag-off and flag-on arms.

## Tests

- Updated the three multi-candidate Path C tests (caura_130 preflight-cap, caura_131 detection-cap + a1_17 matched-set) from per-candidate `call_count == len(cands)` assertions to batched-call assertions.
- Added `test_a61_multi_candidate_entity_aware_batches_once_and_flags` — three entity-aware candidates → one batched call, contradiction effect still applied.
- Full unit suite green (2366 passed); no new mypy errors in the changed file; ruff clean.

### DIFF
--- a/core-api/src/core_api/services/contradiction_detector.py
+++ b/core-api/src/core_api/services/contradiction_detector.py
@@ -1462,6 +1462,136 @@ async def _llm_entity_aware_contradiction_check(
     )
 
 
+# A61 — batched entity-aware judge. Path C's forward-detection loop
+# previously fanned out one ``_llm_entity_aware_contradiction_check`` call
+# per candidate (mirroring the Path A semantic fan-out that #770 batched).
+# For a high-fanout subject (a popular entity referenced by many memories)
+# that is up to ``_ENTITY_LINKS_DETECTION_FETCH_MAX_CANDIDATES`` LLM calls
+# per newly-written memory. This variant judges the whole entity-aware
+# group in ONE ``complete_json`` call, carrying each candidate's OWN
+# resolved entity context so the authoritative same-subject grounding is
+# preserved per candidate.
+BATCH_ENTITY_AWARE_CONTRADICTION_PROMPT = """\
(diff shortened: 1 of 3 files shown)


arkash20:
@Eldad-Caura please approve — CI is green (all checks passing).

Eldad-Caura (approved):

PR #772, opened 2026-08-12
fix(events): declare the embed-backfill subscription in the manifest

Eldad-Caura:
Found while preparing to provision the Pub/Sub topic for #767 — before the flag was flipped, so nothing broke.

## The gap

#767 added `bus.subscribe(Topics.Lifecycle.EMBED_BACKFILL_REQUESTED, ...)` to core-worker's `register_consumers()`, but never added it to the generator's hand-maintained list. So `common/events/events_manifest.json` does not record that core-worker consumes that topic.

That matters more than a stale file, because of what reads it. From the generator's own docstring:

> It is the contract that lets the enterprise repo's `check_pubsub_provisioning.py` fail CI when OSS adds a consumed topic without the matching Terraform subscription — **the gap that took staging down when `memclaw.lifecycle.insights-requested` shipped unprovisioned.**

An incomplete manifest means that guard is blind to the topic, so the enterprise side could ship without a subscription and nothing would object. The exact class of incident the manifest exists to prevent.

## Why nothing caught it

There are two mechanisms, and this fell between them:

- **Lifecycle** topics are captured *dynamically*, by invoking the shared registration helpers in `common.events.lifecycle_handlers` against a recording bus. Anything registered through a helper is picked up automatically.
- **Directly-registered** subscribes can't be captured that way (OSS CI doesn't install the worker package), so they're hand-listed in `_MEMORY_DIRECT` and cross-checked against the consumer files by `test_memory_direct_matches_consumer_files`.

That check's regex was `bus\.subscribe\(\s*Topics\.Memory\.([A-Z_]+)` — **Memory only**. A *lifecycle* topic registered *directly* matched neither mechanism. #767 created the first such case, because it registers in core-worker deliberately: the work lives in `core_worker.backfill`, and core-api — which implements the same lifecycle adapter protocol — cannot run it.

## The fix

1. Add the topic to the direct list, and regenerate the manifest with `scripts/gen_events_manifest.py` (it says "do not edit by hand", so I didn't).
2. Widen the guard's regex to `Topics\.(Memory|Lifecycle)\.`, closing the seam for the next person.
3. Rename `_MEMORY_DIRECT` → `_DIRECT_SUBSCRIBES`. It no longer holds only memory-pipeline topics, and a name asserting otherwise is precisely the drift that caused this.

## Verified the guard now bites

Removing the new entry while leaving the `bus.subscribe` in place — i.e. reproducing #767's exact omission:

```
AssertionError: core-worker: _DIRECT_SUBSCRIBES in scripts/gen_events_manifest.py is out of
sync with core-worker/src/core_worker/consumer.py.
Found ['memclaw.lifecycle.embed-backfill-requested', 'memclaw.memory.embed-requested', ...],
expected ['memclaw.memory.embed-requested', 'memclaw.memory.enrich-requested'].
```

Before this change, that same state passed.

The only manifest change is one added line:

```diff
     "core-worker": [
       "memclaw.lifecycle.archive-expired-requested",
       "memclaw.lifecycle.archive-stale-requested",
+      "memclaw.lifecycle.embed-backfill-requested",
```

86 events tests pass; `ruff check` clean.

## What follows

With the manifest correct, the enterprise `pubsub-provisioning` check will now require a matching Terraform subscription — which is the next step and lives in caura-memclaw-enterprise: topic, subscription and DLQ for `memclaw.lifecycle.embed-backfill-requested`, staging first.

### DIFF
--- a/common/events/events_manifest.json
+++ b/common/events/events_manifest.json
@@ -12,6 +12,7 @@
     "core-worker": [
       "memclaw.lifecycle.archive-expired-requested",
       "memclaw.lifecycle.archive-stale-requested",
+      "memclaw.lifecycle.embed-backfill-requested",
       "memclaw.lifecycle.purge-soft-deleted-requested",
       "memclaw.memory.embed-requested",
       "memclaw.memory.enrich-requested"
--- a/scripts/gen_events_manifest.py
+++ b/scripts/gen_events_manifest.py
@@ -38,16 +38,23 @@ MANIFEST_PATH = (
     / "events_manifest.json"
 )
 
-# Memory-pipeline subscriptions are registered by each service's own
-# ``register_consumers()`` (core-api/src/core_api/consumer.py and
-# core-worker/src/core_worker/consumer.py). They are listed explicitly here
-# rather than captured dynamically because OSS CI does not install the worker
-# package, so it cannot be imported in this generator. This set is stable; if
-# you add a direct ``bus.subscribe`` in a service's register_consumers(), add it
-# here too.
-_MEMORY_DIRECT: dict[str, list[str]] = {
+# Subscriptions registered by each service's own ``register_consumers()``
+# (core-api/src/core_api/consumer.py and core-worker/src/core_worker/consumer.py)
+# rather than by a shared helper in ``common.events.lifecycle_handlers``. They are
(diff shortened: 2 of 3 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #773, opened 2026-08-12
fix(reports): anchor working-on lane keywords to word starts

Eldad-Caura:
## What

`working_on` lane keywords now match at a **word start** instead of anywhere in the string.

## The bug is bigger than the flake that surfaced it

Lane matching runs *before* the memory-type fallback, so any substring hit silently outranks the type. The matcher was a bare `k in text`, and the collisions are ordinary English:

| content | contains | wrong lane |
|---|---|---|
| customer **feedback** reviewed | `dba` | Operating |
| **ownership** / **partnership** | `ship` | Building |
| **install** the CLI | `stall` | Operating |
| **distrust** between teams | `trust` | Governing |

A `decision` saying "customer feedback reviewed" was filed under Operating rather than Governing. Checked the full keyword list against `/usr/share/dict/words` — `"ship"` alone matches **1044** words it was never meant to (airship, leadership, relationship…).

## How it surfaced

`tests/test_reports.py` seeds a `uuid4().hex` id into content. `d`, `b` and `a` are all hex digits, so ~0.145% of ids contain `"dba"` — an Operating keyword. That is the ~0.6%-per-run flake behind `test_report_internal_group_durable_filter` failing on [CI run 31634…](https://github.com/caura-ai/caura-memclaw/actions/runs/31636233722) with `Governing 1` instead of `2` (2/1/0 → 1/1/1). The identical commit passed on re-run with no code change.

## The fix

One precompiled pattern per lane, each alternative anchored `\b` at the word **start** but deliberately **not** at the word end. The trailing half is what would break things: the list relies on deliberate prefixes, so `\bcomplian\b` would stop matching "compliance". Verified the intended prefixes still match — `complian`→compliance, `reliab`→reliability, `build`→building, `audit`→auditing, and the hyphenated `on-call`.

**Trade accepted:** prefixed forms no longer match (`rebuild`, `redeploy`, `unhealthy`, `unreliable`). They fall through to the memory-type lane — the intended default for text the heuristic can't place, and a better failure than matching inside arbitrary words. Flagging it rather than burying it, since it's a real behaviour change.

## Two changes, deliberately

Either alone would stop the flake; they fix different things.

1. **`reports.py`** — the product behaviour above.
2. **`tests/test_reports.py`** — seeds a decimal-only unique suffix instead of conftest's hex `uid()`, so these assertions no longer depend on random data at all. Digits cannot spell a lane keyword. Only this module asserts on lanes, so nothing else is affected.

## Verification

The new test fails without the fix. Reverting only the matcher line, leaving everything else in place:

```
AssertionError: {'Building':  {'count': 1, ...},
                 'Governing': {'count': 0, 'items': []},
                 'Operating': {'count': 2, ...}}
```

— all three `decision` rows pulled out of Governing. With the fix: 3/0/0.

Full suite **4355 passed**, 3 skipped, 1 xfailed. `mypy` clean. `ruff check` and `ruff format --check` clean on `core-api/src/`.

### DIFF
--- a/core-api/src/core_api/routes/reports.py
+++ b/core-api/src/core_api/routes/reports.py
@@ -28,6 +28,7 @@ from __future__ import annotations
 
 import asyncio
 import logging
+import re
 from collections.abc import Awaitable
 from datetime import UTC, datetime, timedelta
 from typing import Any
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

Eldad-Caura:

@erni-a

Context on the two earlier CI failures here: both were the same transient `uv python install 3.13` download error (HTTP/2 "refused stream" from the python-build-standalone release asset), not the code — the same failure hit enterprise #1051 in the same window. Green on re-run with no change.

erni-a (approved):

PR #775, opened 2026-08-13
fix(memory): null the vector when a content-change re-embed fails

Eldad-Caura:
## The bug

`update_memory` wrote the re-embedded vector only `if new_embedding is not None`:

```python
patch["content"] = data.content
if new_embedding is not None:          # <-- the guard
    patch["embedding"] = new_embedding
```

`get_embedding` returns `None` on **failure** rather than raising. So on any embed failure the guard omitted the key entirely, and the row kept the **previous content's vector** while its content was replaced.

That row is wrong twice over:

- **Recall ranks it against text it no longer holds.** Nothing errors — the PATCH returns 200 and the row looks healthy.
- **The column is non-NULL**, so neither the async-embed worker nor the nightly NULL-embedding sweep (caura-enterprise#1051, live as of today) can ever see it. Mis-embedded rather than unembedded, and therefore permanent.

`get_embedding`'s own docstring states the contract the call site was breaking:

> *"a transient OpenAI/Vertex hiccup returns `None` rather than raising, so write paths can persist rows with `embedding=NULL` and let the async-embed worker backfill"*

## Why nulling is safe

Both `None` paths are genuine failures — a registry `ValueError` from `_resolve_provider_or_degrade` (env misconfiguration) and exhausted retries in `_run_with_retry`. There is **no** "embeddings intentionally disabled" path that returns `None`, so this cannot wipe a vector a tenant meant to keep.

## The null actually reaches the column

The whole fix depends on this, so it was traced rather than assumed:

| layer | behaviour |
|---|---|
| `storage_client.update_memory` | passes the patch through as the JSON body |
| `PATCH /memories/{id}` route | no `exclude_none` — body forwarded untouched |
| `memory_update` | filters on `hasattr(Memory, key)` **only**, not on the value |

So `{"embedding": None}` becomes `SET embedding = NULL`. The sibling `patch["supersedes_id"] = None` a few lines above relies on the same behaviour.

## Verification

The regression test asserts the property that actually matters — that the row becomes **visible to the sweep** — using `missing_embeddings` from `/memories/embedding-coverage`, the same count the sweep keys on. Not "did we call the embedder", which would pass either way.

Confirmed it fails without the fix. Reverting only the guarded write, leaving everything else in place:

```
tests/test_update_memory_embedding.py:76: assert 0 == 1
FAILED test_content_change_with_failed_embedding_nulls_the_vector
1 failed, 1 passed
```

`0 == 1` is exactly the defect: the row stayed non-NULL and invisible. The companion test pins the success path, and passes either way by design.

Full suite **4357 passed**, 3 skipped, 1 xfailed. `mypy` clean; `ruff check` and `ruff format --check` clean on `core-api/src/`.

## Scope

This is the third of the three embedding-loss write paths, and the only one the sweep could not repair. The other two (auto-chunk parent, atomic-fact fan-out) land as `NULL` and are already swept; they're untouched here.

### DIFF
--- a/core-api/src/core_api/services/memory_service.py
+++ b/core-api/src/core_api/services/memory_service.py
@@ -2645,8 +2645,19 @@ async def update_memory(
     patch: dict = {}
     if content_changed:
         patch["content"] = data.content
-        if new_embedding is not None:
-            patch["embedding"] = new_embedding
+        # Write the embedding unconditionally — including None. ``get_embedding``
+        # returns None only on failure (exhausted retries, or a degraded /
+        # misconfigured provider), and its contract is that callers "persist rows
+        # with ``embedding=NULL`` and let the async-embed worker backfill".
+        #
+        # Guarding this on ``is not None`` omitted the key instead, which left the
+        # PREVIOUS content's vector on a row whose content had just changed. That
+        # is wrong twice over: the row is silently mis-embedded (recall ranks it
+        # against text it no longer holds, with no error anywhere), and because
+        # the column is non-NULL neither the async-embed worker nor the nightly
+        # NULL-embedding sweep can ever see it — so it stays wrong forever. NULL
+        # is the honest state and the one the existing repair paths look for.
+        patch["embedding"] = new_embedding
         patch["content_hash"] = _content_hash(tenant_id, mem.get("fleet_id"), data.content)
         # P1-2: Clear stale contradiction/supersession state on content change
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

Eldad-Caura:

@erni-a

erni-a (approved):

PR #776, opened 2026-08-13
feat(observability): make the unembedded-row backlog measurable

Eldad-Caura:
## The gap

"How many live memories still have no embedding" had **no answer outside the VPC**. Both storage services are internal-ingress, no metric carried the number, and the only route to it was an AlloyDB Auth Proxy session plus a hand-written `COUNT`.

So in practice nobody measured it — including while turning the nightly embed-backfill sweep on today. The sweep now runs against a backlog of unknown size, and there's no way to tell whether it drains.

## What this adds

| layer | |
|---|---|
| `memory_embedding_coverage_by_tenant` | one `GROUP BY` with a `FILTER`'d count |
| `GET /memories/embedding-coverage-all` | storage-api, cross-tenant |
| `GET /admin/lifecycle/embedding-coverage` | core-api, **admin-gated** |
| `embedding-coverage` tick | hourly, logs totals + per-tenant detail |

**One query, not a loop** over the existing per-tenant helper. The loop would be N round-trips against the pool that serves live traffic, and it can't produce a consistent snapshot — tenants counted early and late see different states, so the totals need not add up.

**The log line is the metric.** There's no metrics client anywhere in this stack; observability is structlog into Datadog logs. So the tick logs rather than emitting a gauge, matching how everything else here is observed.

**Registered unconditionally**, not behind `embed_backfill_enabled` — the count matters most when the sweep is *off*, because that's when nothing is draining the backlog. Hourly rather than daily so the curve distinguishes "the sweep drained it" from "it never grew"; a single daily sample taken near the sweep can't.

## The admin gate is load-bearing

The route is deliberately cross-tenant, and the admin key resolves to `tenant_id=None` — so there's no tenant scope to fall back on. Without `enforce_admin`, every tenant's row counts would be readable by any caller. Counts and tenant ids only; no memory content crosses the boundary.

Verified by removing the gate — `test_admin_route_rejects_non_admin` then reports `DID NOT RAISE`, so the test discriminates rather than passing on a route that merely happens to work.

## Notes for review

- Per-tenant log lines are **capped at 20**, rows ordered worst-first so the cap drops tenants nobody would act on. The cap **announces itself** when it bites — a silent truncation would read as "only these tenants have gaps".
- Coverage tests null a row through the **content-change** path, not the write path: writes embed in a background task, so a monkeypatch there races the assertion. (Found the hard way — the first version of the test failed for exactly that reason.)
- Both drift guards fired and were satisfied rather than bypassed: `test_app.py`'s expected-job set, and `docs/component-registry.yaml`, which `tests/test_component_registry.py` enforces 1:1 against `scheduler.register` calls in both directions.

## Verification

Root suite **4361 passed**, core-operations **49 passed**. `mypy` clean on core-api, core-storage-api and core-operations. `ruff check` + `ruff format --check` clean on every CI-linted scope.

The local core-storage-api suite errors on an unprovisioned `memclaw_storage` database **both with and without this change** — pre-existing, and CI creates that database fresh.

## Unrelated, noticed in passing

`core-operations/.coverage` is not gitignored, so a stray coverage artifact sits untracked in the tree and `git add -A` picks it up. Not touched here to keep this PR focused.

### DIFF
--- a/core-api/src/core_api/clients/storage_client.py
+++ b/core-api/src/core_api/clients/storage_client.py
@@ -852,6 +852,14 @@ class CoreStorageClient:
             params["fleet_id"] = fleet_id
         return await self._get("/memories/embedding-coverage", **params) or {}
 
+    async def get_embedding_coverage_all(self) -> dict:
+        """Embedding coverage for every tenant (operator view, one query).
+
+        No ``tenant_id``: this is the cross-tenant aggregate, so callers must
+        gate it on the admin key. Counts only.
+        """
+        return await self._get("/memories/embedding-coverage-all") or {}
+
     async def get_type_distribution(
         self,
         tenant_id: str,
--- a/core-api/src/core_api/routes/lifecycle.py
+++ b/core-api/src/core_api/routes/lifecycle.py
@@ -236,6 +236,42 @@ async def fanout_lifecycle_action(
     return {"action": action, "published": published, "failed": failed}
 
 
+@router.get("/admin/lifecycle/embedding-coverage")
+async def embedding_coverage_all_tenants(
+    auth: AuthContext = Depends(get_auth_context),
+) -> dict:
+    """How many live memories are still unembedded, per tenant.
+
+    This is the standing answer to "is the embed-backfill sweep keeping up",
+    and it lives here because that sweep is what acts on the number. Before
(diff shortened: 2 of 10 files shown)


Eldad-Caura:

@erni-a

erni-a (approved):

PR #777, opened 2026-08-13
test: give every test its own tenant so committed rows stop leaking between them

Eldad-Caura:
## What

`tests/conftest.py`'s `tenant_id` fixture returned **one constant for the whole run**. Most tests here write through `sc`, which commits on its own connections, so the `db` fixture's rollback never sees those rows and nothing removes them until the session-end sweep. Every committed row therefore stayed visible to every later test, and an assertion like `len(results) >= 1` could be satisfied by a **different test's data** — passing for the wrong reason, or failing only when the run order changed.

This is the root cause #749 worked around by removing 107 unused `db` requests.

The fixture now mints `test-tenant-<uid()>` per test, reusing the existing `uid()` helper and keeping the prefix `_setup_schema`'s sweep matches on.

## Scope

Requested by **72 test functions (81 items) across 11 files**. Distinct tenants per run go from **~95 to ~176** — it roughly doubles rather than going 1 → 72, because 94 bespoke `test-tenant-*` ids already existed across 34 files.

Verified safe before changing anything: **no fixture consumes `tenant_id` at all** (only tests and helpers), so `ScopeMismatch` was never possible.

## Guards

Two in `tests/test_fixture_hygiene.py`: the fixture must return a distinct id per call, and it must keep the swept prefix. Both call the fixture's underlying function **directly** rather than comparing ids across two test invocations — that alternative needs module-level state, which would be per-worker if this suite ever runs parallel, letting both cases pass vacuously. A guard that cannot fail is worse than no guard.

Both were confirmed to **fail on the reverted fixture, for their stated reason**, before and after a restructure.

## Also fixed

- **Swept `audit_chain_head`**, which was never in the teardown tuple. It takes one genesis row per tenant, so a per-test tenant leaves one per test. A local database held **9,186** stale `test-tenant-` rows there from earlier runs; this run cleared it to **0**.
- **Four comments that asserted the old shared-tenant behaviour** and would otherwise have taught the wrong model: `test_p3_3_graph_fanout.py`, `test_integration_search.py`, `test_p2_semantic_dedup.py`, and the sweep comment in `conftest.py` that called itself the primary isolation.
- **Deleted the local `tenant_id` override** in `test_p3_1_entity_resolution.py` — byte-identical to the conftest fixture, and its docstring claimed to override a "session-level fixture" that no longer exists.

## Measurement

Re-measured after rebasing onto `af9e2fb7` (the earlier baseline went stale when 23 commits landed mid-work):

| | passed | skipped | xfailed |
|---|---|---|---|
| baseline (`origin/main` af9e2fb7) | 4377 | 3 | 1 |
| this branch | **4379** | 3 | 1 |

Delta is exactly **+2**, the two new guards. `core-storage-api/tests/` 118 passed. mypy clean on both `src/` trees. ruff `check`/`format` clean at CI's scopes; the pre-existing findings under `tests/` are an identical multiset to `origin/main`.

One flake appeared in a contended run — `test_api_interview.py::test_schedule_next_window_resumes_from_watermark` returned 504. It does not use this fixture (it mints its own tenant via `get_test_auth`), and the 504 is an `asyncio.wait_for` timeout. Isolated, it passes 5/5 at **~49–55 s against a 90 s timeout** — inherently tight on a loaded machine. Unrelated to this change, but worth knowing.

## Deliberately not addressed

`get_test_auth()` still returns the shared `"default"` tenant across **172 call sites in 25 files**, and `"default"` does not match the sweep's `LIKE 'test-tenant-%'` — so those API-test rows persist across runs **permanently**. Same defect one layer over; too large for this change.

Separately, `tests/pipeline/test_write_pipeline.py:20` uses a `test-pipeline-` prefix the sweep never matches while genuinely persisting rows, and three `tests/pipeline/` modules use a bare `test-tenant` (no trailing hyphen) that also fails the pattern. Generalising the prefix guard into an AST scan over every tenant literal would catch these, but needs those files fixed first.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -68,6 +68,11 @@ TEST_DB_URL = os.environ.get(
     "postgresql+asyncpg://memclaw:changeme@127.0.0.1:5432/memclaw",
 )
 
+# NOT what the ``tenant_id`` fixture returns — that mints a fresh id per test. This
+# is one id per RUN, kept for the modules whose module-level seed helpers need a
+# tenant without taking a fixture argument. None of them relies on cross-test
+# sharing, so they could each mint their own; threading a tenant through those
+# helpers is the follow-up that would let this go away.
 TENANT_ID = f"test-tenant-{uuid.uuid4().hex[:8]}"
 FLEET_ID = "test-fleet"
 AGENT_ID = "test-agent"
@@ -94,7 +99,7 @@ def get_admin_headers() -> dict:
 
 
 def uid() -> str:
-    """Short unique suffix to avoid duplicate-content 409s across test runs."""
+    """Short unique suffix — for distinct content (409s) and distinct tenant ids."""
     return uuid.uuid4().hex[:8]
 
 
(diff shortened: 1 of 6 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #778, opened 2026-08-13
feat(api): type the memory-get 200 body, so the frozen contract actually pins it

Eldad-Caura:
## What

`GET /api/v1/memories/{memory_id}` entered the frozen broker contract in #753 with `"schema": {}`, so oasdiff pinned its path, method, params and 422 but **not** its response shape. It was allowlisted in `tests/test_broker_contract._UNTYPED_200` with the reason: the handler returned a hand-built 29-field `JSONResponse`, and FastAPI **skips response validation entirely** for a handler that returns a `Response` object — so attaching a `response_model` alone would have documented a guarantee the code does not make, which is worse than untyped.

The handler now returns a plain dict against a new `MemoryDetailResponse`, so the shape is enforced rather than asserted, and the allowlist entry is gone. That assertion fails on a *stale* entry by design, which is what removed it.

## The timestamps are `str`, and that is the point

Typing them `datetime` would not merely annotate — it re-serialises. **Measured**: pydantic v2 renders `2026-08-13T18:52:48.040997+00:00` as `...040997Z`. That is a wire change for every consumer, not something to take on as a side effect of adding a schema.

Worth knowing, because it is now pinned rather than accidental — the fleet already emits **both** forms for the same row:

| endpoint | `created_at` |
|---|---|
| `POST /api/v1/memories` (via `MemoryOut`, typed `datetime`) | `...040997Z` |
| `GET /api/v1/memories/{id}` (this route, pass-through) | `...040997+00:00` |

Aligning them is a deliberate API decision with its own migration story. `str` here documents exactly what ships today.

## Verified byte-equivalent, not assumed

Captured the response before and after the change: key order identical (all 29, same sequence), key set identical, per-key types identical, `created_at` still `+00:00`. The only value differences were genuinely per-run — write timings inside `metadata`, and the content-derived fake embedding.

## Three guards, each shown to fail without its fix

`tests/test_memory_get_serialization_contract.py` pins the two properties that are easy to break here and that yield perfectly valid JSON either way: field order (asserted at import time **and** against a live response) and the `+00:00` timestamp form.

The key-order list is **written out longhand rather than derived from the model**. My first draft compared the response to `MemoryDetailResponse.model_fields` and *could not fail* — FastAPI builds the response order from the model, so reordering fields moved both sides together and the test passed. Caught by reordering two fields and watching it pass. Both assertions now fail on that mutation; the timestamp assertion fails when a field is typed `datetime`.

## Cost, stated plainly

The endpoint now pays pydantic validation of 29 fields per request where it previously serialised straight to JSON. That is the price of the guarantee — a single-row read, so small in absolute terms, but not zero.

## Measurement

| | passed | skipped | xfailed |
|---|---|---|---|
| baseline (`origin/main`) | 4379 | 3 | 1 |
| this branch | **4382** | 3 | 1 |

Exactly **+3**, the new guards. `core-storage-api/tests/` 118 passed. mypy clean on both `src/` trees. ruff `check`/`format` clean at CI's scopes. `gen_broker_openapi.py --check` up to date; `plugin/tools.json` unchanged.

`core-api/openapi.broker.json` regenerated: **+286/-1**, the single removal being the empty `schema: {}` stub. Nullability mirrors the `memories` table plus the two computed fields that are `None` for an un-embedded row, so a NULL column cannot turn a 200 into a 500 via `ResponseValidationError`.

## Notes

- `GET /health` and `GET /version` remain in `_UNTYPED_200`; both were untyped before #753 and are trivial, stable handshake payloads.
- Several other routes in this file still return `JSONResponse` (bulk, STM, ingest). They are outside the broker contract, so they are not in scope here, but they have the same untyped-response property if they are ever added to it.
- `/simplify` was **not** run on this change. It qualifies under the repo threshold (new model, restructured return path, comments asserting behavioural properties), so flagging the omission rather than implying otherwise. Build, lint, mypy, both test trees, the contract check, and a read of the full diff were all done.

### DIFF
--- a/core-api/openapi.broker.json
+++ b/core-api/openapi.broker.json
@@ -440,6 +440,289 @@
         "title": "HTTPValidationError",
         "type": "object"
       },
+      "MemoryDetailResponse": {
+        "description": "Full detail for one memory: row fields, entity links, embedding stats.",
+        "properties": {
+          "agent_display_name": {
+            "anyOf": [
+              {
+                "type": "string"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "title": "Agent Display Name"
+          },
+          "agent_id": {
+            "title": "Agent Id",
+            "type": "string"
+          },
+          "content": {
+            "title": "Content",
+            "type": "string"
+          },
+          "content_hash": {
+            "anyOf": [
+              {
+                "type": "string"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "title": "Content Hash"
+          },
+          "created_at": {
+            "anyOf": [
+              {
+                "type": "string"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "title": "Created At"
+          },
+          "deleted_at": {
+            "anyOf": [
+              {
+                "type": "string"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "title": "Deleted At"
+          },
+          "embedding_preview": {
+            "anyOf": [
+              {
+                "items": {
+                  "type": "number"
+                },
+                "type": "array"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "title": "Embedding Preview"
+          },
+          "embedding_stats": {
+            "anyOf": [
+              {
+                "additionalProperties": true,
+                "type": "object"
+              },
+              {
+                "type": "null"
+              }
+            ],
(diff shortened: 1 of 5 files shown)


Eldad-Caura:
@erni-a please approvce

erni-a (approved):

PR #780, opened 2026-08-13
feat(memory): the memory_relations table behind memory_link (step 1 of the graph)

Eldad-Caura:
## What, and what this deliberately is not

`memory_link` promises a typed relation between two **memory** ids, and nothing in the schema could hold one — `relations` is entity↔entity, `memory_entity_links` is memory↔entity, and `memories.supersedes_id` is memory→memory but single-valued and written only by the contradiction detector. That is why the tool has been accepted-but-unadvertised since memclawd #130.

This adds **the store and nothing else.** Nothing consumes it yet: no storage endpoint, no core-api route, no `cloud.Client` method, no dispatcher wiring, and `memory_link` stays unadvertised. memclawd's `TestAdvertisedToolsAreAllServable` enforces that pairing, so the cloud side must exist before the tool can be re-advertised — this is the prerequisite, landed on its own rather than as part of a cross-repo change.

## Design, decided rather than inferred

The six relation types are `contracts/mcp-tools.md` §8 verbatim; Eldad confirmed the set rather than letting it be read off a tool description.

**`supersedes` is not stored here.** It reuses `memories.supersedes_id` so the detector and the API write one field instead of two stores that can disagree. A CHECK constraint makes that structural — a `supersedes` row cannot be inserted at all. The consequence is a real asymmetry the tool will have to document: `supersedes` is 1:1 and a second one *overwrites*, while a second `elaborates` is an additional row.

**One directed row per link, never two.** Three of the five stored types are semantically symmetric, and writing both directions creates pairs that must stay in sync plus half-edges when one is deleted. Symmetry is a read concern, and all per-type knowledge lives in one constant (`SYMMETRIC_RELATION_TYPES`) instead of spreading into the schema.

**Soft delete leaves the rows.** `memories` is soft-deleted routinely and that is reversible, so cascading would make un-delete lossy. The read path filters `deleted_at IS NULL` — free, since it already joins `memories` for endpoint content — and `ON DELETE CASCADE` covers the terminal case when the purge sweep hard-deletes the row. No new retention machinery.

## Verified rather than asserted

**Model and migration produce byte-identical schemas.** Applied 037 to a scratch database and diffed `\d memory_relations` against the `create_all` result: identical across columns, defaults, all four indexes, both CHECKs, and both CASCADE FKs. Tests build the schema with `create_all` while prod runs alembic, so a divergence there stays invisible until production.

**Seven tests exercise the guarantees against a real database** — a comment claiming a constraint exists is worth nothing until something has watched it refuse a write:

- `supersedes` rejected by CHECK
- self-link rejected by CHECK
- duplicate `(tenant, from, type, to)` rejected by the natural key
- hard-deleting an endpoint cascades the link away
- **soft-deleting an endpoint does *not***

One test pins a **gap** on purpose: the reversed pair of a symmetric type *is* accepted by the natural key. That is exactly why the write path will have to check it — the schema does not make symmetry idempotent, and this documents that where someone would otherwise assume it does. If a later change makes the schema enforce it, that test should fail and be replaced rather than deleted.

## Notes on the migration

Indexes are plain, not `CONCURRENTLY`: the table is created in the same migration, so it is empty and unlocked. `test_no_plain_create_index_on_large_tables` scopes its requirement to large pre-existing tables — the case that crashed six storage-writer boots on 2026-06-16.

Both endpoint columns get their own index because the natural key leads with `tenant_id` and therefore serves neither as a prefix. `test_every_fk_referencing_column_is_index_leading` fails at PR time otherwise, and an unindexed FK referencing column is what cost 15.3 s of a 15.5 s bulk delete before migration 035.

`test_single_head` expectation moved 036 → 037.

## Measurement

| | result |
|---|---|
| `tests/` | 4382 → **4389** (+7, exactly the new tests) |
| `core-storage-api/tests/` | **118** on a fresh database |
| mypy | clean on both `src/` trees |
| ruff check / format | clean at CI's scopes |

One local-only wrinkle worth recording: running `core-storage-api/tests/` against a database where the root `tests/` tree had already run `create_all` gives 106 `DuplicateTableError`s, because `create_all` made `memory_relations` while `alembic_version` still said 036. It is not reachable in CI — that job gets its own fresh Postgres service, and CI runs `alembic upgrade head` before `pytest tests/`, so alembic is always ahead. Flagging it so nobody debugs it as a code defect.

## Remaining chain for `memory_link`

1. storage-api endpoint (upsert + list), with the reversed-pair check for symmetric types
2. core-api route
3. `supersedes` path — `PATCH /memories/{id}/status` already has the full authz stack and already reaches a storage layer supporting `supersedes_id` / `unset_supersedes` / `expected_supersedes_id`; it simply forwards none of them. Use the CAS gate so an API write cannot clobber the detector's pointer.
4. memclawd: `cloud.Client` method, `CloudLinker` capability, `newCloudDispatcher` wiring, a real input schema, then re-advertise in `tools/list`
5. a `BROKER_OPERATIONS` row here

### DIFF
--- a/common/models/__init__.py
+++ b/common/models/__init__.py
@@ -15,6 +15,7 @@ from common.models.lifecycle_audit import LifecycleAudit
 from common.models.memory import Memory
 from common.models.memory_conflict import MemoryConflict
 from common.models.memory_derivation import MemoryDerivation
+from common.models.memory_relation import MemoryRelation
 from common.models.skill_factory import ForgeRejectedFingerprint, SessionTrace
 
 __all__ = [
@@ -36,6 +37,7 @@ __all__ = [
     "LifecycleAudit",
     "Memory",
     "MemoryConflict",
+    "MemoryRelation",
     "MemoryDerivation",
     "MemoryEntityLink",
     "Relation",
--- a/common/models/memory_relation.py
+++ b/common/models/memory_relation.py
@@ -1,0 +1,135 @@
+"""Typed memory-to-memory relations — the store behind the ``memory_link`` tool.
+
+Distinct from the two link tables that already exist, and it is worth being explicit
+about why none of them could serve this:
+
+* ``relations`` is entity <-> entity.
+* ``memory_entity_links`` is memory <-> entity.
+* ``memories.supersedes_id`` is memory -> memory but SINGLE-VALUED, and is written by
+  the contradiction detector.
+
+``memory_link`` promises a typed relation between two MEMORY ids, which is none of
+those, and it was accepted-but-unadvertised until this table existed (memclawd #130).
+
+WHY ``supersedes`` IS NOT STORED HERE. Eldad's call (2026-08-13): ``supersedes``
+reuses ``memories.supersedes_id`` rather than becoming a row here, so the contradiction
+detector and the API write one field instead of two stores that can disagree. The
+consequence is a real asymmetry the tool has to document: ``supersedes`` is 1:1 and a
+second one OVERWRITES, while a second ``elaborates`` is an additional row. The CHECK
+constraint below makes that structural rather than a convention — a ``supersedes`` row
+cannot be written here at all.
+
+WHY ONE DIRECTED ROW, never two. Three of the five stored types are semantically
+symmetric (``contradicts``, ``alternative_to``, ``related_to``), and the tempting
+implementation is to write both directions. That creates pairs which must stay in sync,
+and deleting one leaves a half-edge. So exactly one row is stored, exactly as the caller
+stated it, and symmetry is a READ concern: queries for a symmetric type match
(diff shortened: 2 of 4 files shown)


Eldad-Caura:
Closing as prohibitively stale, not as unwanted.

**787 commits behind main** with conflicts, last touched 2026-08-16. Its migration is `037_memory_relations.py`; main is now at `052`, so the revision number is dead and the chain would have to be re-derived against a substantially different schema.

Nothing depends on it: no open issue asks for `memory_link`, and nothing on main references the table. The "what this deliberately is not" reasoning in this body — why `relations`, `memory_entity_links` and `supersedes_id` each fail to express a typed memory↔memory edge — is the durable part and stays readable here.

Reopen when the graph is actually scheduled; the rebase cost and the redo cost are about equal now.

PR #781, opened 2026-08-14
feat(memory): storage endpoints for the memory_link graph (step 2)

Eldad-Caura:
**Stacked on #780** — based on `feat/memory-relations`, so the diff shows only this step. Retarget to `main` once #780 merges.

## What

An idempotent upsert and a both-directions listing on the table #780 adds: `POST|GET /memories/{memory_id}/relations`.

Still nothing agent-facing — no core-api route, no dispatcher wiring, and `memory_link` stays unadvertised. memclawd's `TestAdvertisedToolsAreAllServable` gates that on the full chain existing.

## Idempotency is the substance

A repeated link returns the existing row rather than 409-ing. `relation_add` (the entity equivalent) documents why at length: it was a plain INSERT once, and the IntegrityError cluster that followed became 5xx storms with rows committed while the caller retried. An agent re-issuing `memory_link` must not reproduce that.

**Symmetric types also match the reversed pair.** `(A contradicts B)` and `(B contradicts A)` are one claim but two rows under the natural key — the schema accepts both, deliberately pinned by #780's tests — so the collapse has to happen in the service. The first caller's direction is preserved; the reversed call returns the existing row rather than re-pointing it. Asymmetric types are **not** collapsed, since a reversed `depends_on` is a different claim.

A race is left open and documented: two concurrent calls writing opposite directions of the same symmetric link can both pass the check. It's benign — a duplicate edge, not corruption — on an agent-initiated, low-rate write. Closing it properly means a partial unique index over `(tenant_id, least(from,to), greatest(from,to), relation_type)` scoped to the symmetric types, at the cost of a more awkward `ON CONFLICT` target. Worth doing if duplicates are ever observed; not worth pre-empting.

## Other behaviour

- `supersedes` → **422 naming `memories.supersedes_id`**, not a 500 from the CHECK constraint. Unknown types and self-links are 422 for the same reason: validated before touching the database.
- Both endpoints must exist, be live, and belong to the tenant, so a link can't be used to probe whether another tenant's memory id exists — flat 404 either way.
- The listing returns both directions unconditionally and omits links whose far endpoint is soft-deleted. That filter is the only thing stopping a soft-deleted memory reappearing through its graph, since soft delete deliberately leaves the edge intact.

## Three bugs the tests caught, one root cause

`metadata` is reserved on a declarative class — it resolves to `Base.metadata`, the SQLAlchemy `MetaData` object — which is why the mapped attribute is `metadata_`. That collision bit in three separate places, each failing only at **execute** time with `AttributeError: 'MetaData' object has no attribute '_bulk_update_tuples'`, a long way from its cause:

1. `.values(**data)` with the wire key `"metadata"`;
2. the `on_conflict_do_update` `set_` key — the ORM attribute name `"metadata_"` leaks into SQL as a column name, so it must be keyed by the **Column object**;
3. binding Python `None` to a JSONB column stores JSON `null`, which is not SQL NULL — so the COALESCE would have treated an omitted field as a real value and blanked what an earlier link recorded. `metadata` is now omitted entirely when absent.

Only (1) was visible at first; (2) and (3) surfaced one at a time as each fix uncovered the next. All three sites carry a comment, because the failure mode gives no hint about the cause. Worth noting the endpoint would have "worked" for the common path with (3) unfixed — it only loses data on a relink.

## Measurement

| | result |
|---|---|
| `core-storage-api/tests/` | 118 → **129** (+11, the new tests) on a fresh database |
| `tests/` | 4389, unchanged |
| mypy | clean |
| ruff check / format | clean at CI's scopes |

## Remaining chain

3. core-api route
4. `supersedes` path — `PATCH /memories/{id}/status` already has the full authz stack and already reaches a storage layer supporting `supersedes_id` / `unset_supersedes` / `expected_supersedes_id`; it forwards none of them. Use the CAS gate so an API write can't clobber the detector's pointer.
5. memclawd: `cloud.Client` method, `CloudLinker`, `newCloudDispatcher` wiring, a real input schema, then re-advertise in `tools/list`
6. a `BROKER_OPERATIONS` row here

### DIFF
--- a/common/models/__init__.py
+++ b/common/models/__init__.py
@@ -15,6 +15,7 @@ from common.models.lifecycle_audit import LifecycleAudit
 from common.models.memory import Memory
 from common.models.memory_conflict import MemoryConflict
 from common.models.memory_derivation import MemoryDerivation
+from common.models.memory_relation import MemoryRelation
 from common.models.skill_factory import ForgeRejectedFingerprint, SessionTrace
 
 __all__ = [
@@ -36,6 +37,7 @@ __all__ = [
     "LifecycleAudit",
     "Memory",
     "MemoryConflict",
+    "MemoryRelation",
     "MemoryDerivation",
     "MemoryEntityLink",
     "Relation",
--- a/common/models/memory_relation.py
+++ b/common/models/memory_relation.py
@@ -1,0 +1,135 @@
+"""Typed memory-to-memory relations — the store behind the ``memory_link`` tool.
+
+Distinct from the two link tables that already exist, and it is worth being explicit
+about why none of them could serve this:
+
+* ``relations`` is entity <-> entity.
+* ``memory_entity_links`` is memory <-> entity.
+* ``memories.supersedes_id`` is memory -> memory but SINGLE-VALUED, and is written by
+  the contradiction detector.
+
+``memory_link`` promises a typed relation between two MEMORY ids, which is none of
+those, and it was accepted-but-unadvertised until this table existed (memclawd #130).
+
+WHY ``supersedes`` IS NOT STORED HERE. Eldad's call (2026-08-13): ``supersedes``
+reuses ``memories.supersedes_id`` rather than becoming a row here, so the contradiction
+detector and the API write one field instead of two stores that can disagree. The
+consequence is a real asymmetry the tool has to document: ``supersedes`` is 1:1 and a
+second one OVERWRITES, while a second ``elaborates`` is an additional row. The CHECK
+constraint below makes that structural rather than a convention — a ``supersedes`` row
+cannot be written here at all.
+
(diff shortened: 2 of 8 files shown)


Eldad-Caura:
Closing as prohibitively stale, not as unwanted.

**792 commits behind main** with 4 conflict hunks, last touched 2026-08-16. It is stacked on #780, which is closing for the same reason.

The feature never landed and nothing is waiting on it: no open issue references `memory_link` or `memory_relations`, and nothing on main expects the endpoints (the single `memory_link` grep hit on main is `test_memory_linked_to_both_keeps_best`, an unrelated entity-relation-weights test that passes).

The design reasoning in this body and #780's stays readable here. If the graph is scheduled, reopening costs about what rebasing 792 commits would, and the migration would need renumbering regardless — see #780.

PR #786, opened 2026-08-15
feat(embeddings): make stale vectors detectable via embedding provenance

Eldad-Caura:
## The last embedding failure with no detector

`memories` carried `embedding` and `content_hash`, but nothing recording **which** content a vector was computed from. So a vector left over from earlier content was byte-identical to a correct one — `embedding IS NOT NULL` says only that *something* was embedded, never *what*.

That left mis-embedded rows undetectable. The nightly NULL sweep can't see them (the column is non-NULL), and recall silently ranks them against text they no longer hold. #775 stopped new ones being created; it could not find existing ones, and nothing could have.

Migration **037** adds `memories.embedded_content_hash`, making staleness a query:

```sql
embedding IS NOT NULL
AND embedded_content_hash IS DISTINCT FROM content_hash
```

`IS DISTINCT FROM` rather than `<>` — with `<>` every NULL-provenance row evaluates to NULL and silently drops out of the `WHERE`, which is the same disappearing-evidence failure this column exists to end.

## Stamped in the storage layer, with one deliberate exception

Provenance is set in storage rather than at each call site, so paths can't drift: `_filter_memory_fields` covers both inserts (single + bulk), and `memory_update` derives it when a patch carries embedding and content_hash together.

`memory_update_embedding` is the exception — it takes the hash as a **parameter** and never reads it off the row. The worker fetches content, embeds, then writes back. A content PATCH landing inside that window means the row has moved on, and copying its current hash at write time would stamp the **new** hash onto a vector computed from the **old** text — recording the row as freshly embedded at the precise moment it became stale. Callers that know what they encoded pass it; the rest leave it NULL, because unknown provenance is honest and a wrong hash is not.

## Three states, deliberately not collapsed

| state | meaning | repairable by the sweep? |
|---|---|---|
| `missing` | no vector | ✅ yes |
| `stale` | wrong vector | ❌ **no** — non-NULL, invisible to it |
| `unknown_provenance` | embedded pre-037 | n/a — undetermined, not damaged |

Conflating unknown with stale would report the whole historical corpus as broken; conflating it with fresh would assert a correctness nobody verified. It drains on its own as rows are rewritten.

## Two guards caught real mistakes — including one that didn't

`test_single_head` pinned the head at 036; updated to 037 with a chain link, which is the deliberate-update step it exists to force.

**`test_no_plain_create_index_on_large_tables` did not catch the first draft.** It regex-scans raw `op.execute("CREATE INDEX …")` only, and the draft used `op.create_index(...)` — the Alembic API form, which emits a plain in-transaction `CREATE INDEX` on `memories` and carries exactly the defect the guard exists to prevent (a plain build took an AccessExclusive lock and crashed 6 storage-writer boots on 2026-06-16).

So this PR does both: the migration now builds **CONCURRENTLY** in an autocommit block, **and** the guard is extended to flag the API form on large tables. Verified by reverting to the broken form and watching it fail with a clear message.

## Verification

**Migration** — ran the real Alembic chain against a fresh database: head reaches `037`, column and partial index created, detector predicate executes, and `downgrade → re-upgrade` is clean.

**Detection** — verified by breaking it:

- stale counter forced to zero → **1 test fails**
- insert provenance removed → **2 tests fail**

Both restore green, so the tests discriminate rather than passing on shape.

The stale test creates a genuinely stale row the way the bug did — rewriting `content` + `content_hash` while leaving `embedding` untouched — and asserts it lands in `stale`, **not** `missing`. That distinction is the whole point.

Suites: root **4372 passed**, core-operations **50 passed**. `mypy` clean on all three packages; `ruff check` + `format --check` clean on every CI scope, including `common/` at its own line-length 88.

## Also

Adds `.coverage` to `.gitignore` — an untracked artefact that `git add -A` swept into a commit twice during this work. Unrelated to the feature; drop the line if you'd rather keep this PR pure.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,3 +52,8 @@ reports/
 !core-storage-api/uv.lock
 !core-worker/uv.lock
 !core-operations/uv.lock
+
+# Coverage data files. Untracked-but-present artefacts kept getting swept
+# into commits by `git add -A` (twice during the 2026-08 embedding work).
+.coverage
+.coverage.*
--- a/common/models/memory.py
+++ b/common/models/memory.py
@@ -42,6 +42,25 @@ class Memory(Base):
     )
     title: Mapped[str | None] = mapped_column(Text)
     content_hash: Mapped[str | None] = mapped_column(Text)
+    # Provenance for ``embedding``: the ``content_hash`` of the text the
+    # vector was actually computed from. Without it a vector left over
+    # from earlier content is byte-identical to a correct one, so a
+    # mis-embedded row is undetectable — ``embedding IS NOT NULL`` says
+    # only that *something* was embedded, never *what*.
+    #
+    # Staleness is then expressible:
+    #     embedding IS NOT NULL
+    #     AND embedded_content_hash IS NOT NULL      -- provenance known
+    #     AND embedded_content_hash IS DISTINCT FROM content_hash
+    #
+    # NULL means "provenance unknown", NOT "stale": every row written
+    # before migration 037 has no recorded hash, and calling those stale
+    # would report the entire historical corpus as damaged.
+    #
+    # Which is why the second line is load-bearing rather than redundant:
+    # ``NULL IS DISTINCT FROM <hash>`` is TRUE, so dropping it silently
+    # folds every unknown row into the stale count.
+    embedded_content_hash: Mapped[str | None] = mapped_column(Text)
     # Per-attempt idempotency token (CAURA-602). Server-derived from
     # ``X-Bulk-Attempt-Id + ":" + index`` on the bulk path; NULL for
     # single-write and pre-rollout rows. The partial unique index
--- a/core-api/src/core_api/clients/storage_client.py
+++ b/core-api/src/core_api/clients/storage_client.py
@@ -807,14 +807,22 @@ class CoreStorageClient:
         memory_id: str,
         tenant_id: str,
         embedding: list[float],
+        embedded_content_hash: str | None = None,
     ) -> dict | None:
         # ``tenant_id`` (the row's home tenant) scopes the by-id embedding
         # write so a worker/backfill can't overwrite a foreign tenant's
         # vector.
-        return await self._patch(
-            f"/memories/{memory_id}/embedding",
-            {"embedding": embedding, "tenant_id": tenant_id},
-        )
+        #
+        # ``embedded_content_hash`` is the hash of the text this caller
+        # embedded. Pass it whenever the content is in hand: it is what makes
+        # a later content change detectable as a stale vector. Omitting it
+        # records provenance as unknown rather than guessing — storage will
+        # NOT infer it from the row, because the row can move between the
+        # caller's read and this write.
+        payload: dict[str, Any] = {"embedding": embedding, "tenant_id": tenant_id}
+        if embedded_content_hash is not None:
+            payload["embedded_content_hash"] = embedded_content_hash
+        return await self._patch(f"/memories/{memory_id}/embedding", payload)
 
     async def update_memory_entities(
         self,
--- a/core-api/src/core_api/routes/lifecycle.py
+++ b/core-api/src/core_api/routes/lifecycle.py
@@ -267,6 +267,8 @@ async def embedding_coverage_all_tenants(
             "total_active": coverage.get("total_active"),
             "missing_embeddings": coverage.get("missing_embeddings"),
             "tenants_with_missing": coverage.get("tenants_with_missing"),
+            "stale_embeddings": coverage.get("stale_embeddings"),
+            "unknown_provenance": coverage.get("unknown_provenance"),
         },
     )
     return coverage
--- a/core-api/src/core_api/services/memory_service.py
+++ b/core-api/src/core_api/services/memory_service.py
@@ -2041,7 +2041,16 @@ async def _reembed_memory(
                 )
             )
             return
-        await sc.update_embedding(str(memory_id), tenant_id, embedding)
+        # Record WHICH text this vector came from. ``content`` is the string
+        # just embedded above, so hashing it here — rather than letting
+        # storage read the row's current hash — keeps the record accurate
+        # even if a content PATCH landed while we were embedding.
+        await sc.update_embedding(
+            str(memory_id),
+            tenant_id,
+            embedding,
+            embedded_content_hash=_content_hash(tenant_id, mem.get("fleet_id"), content),
+        )
         logger.info("Background re-embed succeeded for memory %s", memory_id)
     except (TimeoutError, ValueError, RuntimeError, OpenAIError, GoogleAPIError):
         logger.exception("Background re-embed error for memory %s", memory_id)
@@ -2235,7 +2244,14 @@ async def _reembed_memories_bulk(
             )
             continue
         try:
-            await sc.update_embedding(str(memory_id), tenant_id, embedding)
+            # Same provenance stamp as the single-row path above: hash the
+            # text we embedded, not whatever the row says now.
+            await sc.update_embedding(
+                str(memory_id),
+                tenant_id,
+                embedding,
+                embedded_content_hash=_content_hash(tenant_id, mem.get("fleet_id"), content),
+            )
         except Exception:
             # Broad match for the same reason as the outer batch-call
             # except: httpx-layer errors, pool exhaustion, auth, etc.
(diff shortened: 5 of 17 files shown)


Eldad-Caura:
Confirmed and fixed — the finding is correct.

The index predicate omitted `embedded_content_hash IS NOT NULL`, so it did not match the detector it was built for. `NULL IS DISTINCT FROM <hash>` is TRUE, so the index covered every unknown-provenance row — on day one that is the entire pre-migration table, which makes a "partial" index a full-table one and inverts the reason for making it partial at all.

The same wrong formula was in the migration docstring and the model comment, where it would have told a reader that unknown rows are stale — the exact conclusion the three-bucket design exists to prevent, and the shape any ad-hoc query copied from it would have taken. Both corrected.

Demonstrated on a seeded table with one row of each kind (correct / stale / unknown / missing):

| predicate | rows matched |
|---|---|
| detector, and the corrected index | **1** (the genuinely stale row) |
| previous index predicate | **2** (also swept in the unknown row) |
| unknown-provenance rows | **1** |

The over-match is exactly the unknown count. Index definition re-read from `pg_indexes` after a fresh Alembic run to confirm Postgres stored the corrected predicate.

**One deviation from the suggested fix:** applied in 037 rather than a new 038. 037 is not merged — it is in this PR — so history stays append-only either way, and knowingly shipping a wrong index in order to add a second migration correcting it would leave every deployment briefly carrying the full-table index. If it had already merged, 038 would have been right.

Root suite still 4372 passed; mypy and ruff clean.

Eldad-Caura:
@claude please re-review — the index-predicate fix is pushed and the review job skips on synchronize, so the amended commit hasn't been looked at.

Changed since the last review: `embedded_content_hash IS NOT NULL` added to the partial index predicate so it matches the detector, plus the same correction to the formula in the migration docstring and the model comment. Applied in 037 rather than a new 038 because 037 is unmerged — reasoning in the comment above.

Eldad-Caura:
Both findings confirmed and fixed.

**1. `!=` vs NULL-safe comparison — real, and worse than it reads.** `content_hash` is nullable, so `<>` against a NULL yields NULL and `COUNT(*) FILTER` drops it. Demonstrated on a seeded row (embedding set, provenance *known*, `content_hash` NULL):

| bucket | count |
|---|---|
| stale, via `<>` | **0** ← vanished |
| stale, via `IS DISTINCT FROM` | **1** |
| unknown | 0 |
| missing | 0 |
| total_active | **1** |

Counted in the denominator and in no defect bucket. Now `is_distinct_from`, matching the migration's index predicate.

Worth naming: this is the **same ORM-vs-index drift as the previous round, in the opposite direction**. Last time the index was too broad; this time the ORM was. One rule expressed in two languages, and nothing checks they agree — which is why both slipped past a green suite.

**2. Per-tenant endpoint parity — fixed.** Added `memory_count_stale_and_unknown_embeddings` (fleet-aware, same predicate) and the two keys to `GET /embedding-coverage`. The new test asserts the two endpoints **agree** for a tenant, not merely that both have the keys — absence looking like zero was the actual failure mode.

Both verified by reverting them: restoring `!=` fails 1 test, dropping the per-tenant fields fails 2.

Root suite **4374 passed**; mypy and ruff clean.

@claude please re-review — the review job skips on synchronize.

Eldad-Caura:
Confirmed and fixed — this was the important one.

`handle_embed_request` is the production write path under `DEPLOYMENT_MODE=deferred`, which is what prod core-api runs. It already had `request.content_hash` (it reads it for the cache lookup) but its PATCH sent only `{tenant_id, embedding, metadata_patch}`. So **every row embedded in production would have sat at `unknown_provenance` permanently**, leaving the detector blind to the exact path responsible for almost all embedding.

My earlier audit covered core-api's re-embed paths and the storage layer. core-worker has its **own** storage client and I missed it.

Two implementation notes where I diverged from the suggested prompt:

- **Sent as `embedded_content_hash`, not `content_hash`.** The prompt floated `content_hash` to trigger the `memory_update` stamping branch. That would be wrong — `content_hash` is a real column, so the PATCH would *write* it, overwriting the row's hash with a value from an event that may already be stale. The explicit key sets provenance alone.
- **Kept it on `PATCH /memories/{id}`** rather than switching to the dedicated `/embedding` route. That route replaces `metadata` wholesale, which would clobber the keys the `metadata_patch` merge exists to preserve.

The hash is right on both branches: a fresh embed encodes `request.content`; a cache hit reuses a vector found *by* that same hash. Absent on the event → key omitted → provenance stays NULL.

Verified by reverting the forwarding — two tests fail on `KeyError: 'embedded_content_hash'`. Coverage: fresh-embed, cache-hit, outbound key name (asserts `content_hash` is **not** in the body), and omit-when-unknown.

---

**Worth flagging separately: `core-worker` has no CI coverage at all.** `grep -rn "core-worker" .github/workflows/` returns **zero** matches — not installed, not linted, not type-checked, not tested by any workflow. That is why this reached review instead of CI, and it's the same gap #768 closed for core-operations. Filed as its own task rather than expanded into this PR.

Suites: root **4374**, core-operations **50**, core-worker **70 passed** with one pre-existing failure (`test_post_gives_up_after_max_attempts_on_connect_timeout` — confirmed still failing with all local changes stashed; it exercises `_post` retry, untouched here). ruff + mypy clean on core-worker.

@claude please re-review — the review job skips on synchronize.

Eldad-Caura:
Both findings addressed, plus the suggestion. Also rebased onto main so this PR is now gated by #787.

**1. `embedded_content_hash` added to `MEMORY_FIELDS`.** The column was stamped on writes but exposed by no read path — an operator could see that N rows are stale and never *which*, or why a given row was classified that way.

**2. The "matches exactly" claim was wrong a second time.** The index omits the detector's `status IN LIVE_MEMORY_STATUSES`, so it is broader than the query it serves.

I corrected the **claim** rather than the predicate, deliberately. `LIVE_MEMORY_STATUSES` is a Python tuple in `common/constants.py`; copying its members into an immutable migration would put one rule in two places with nothing checking they agree — the exact failure this column exists to expose, and one this PR has now hit **three** times. A SQL literal would rot silently the first time that tuple changes.

The cost of leaving it out is bounded and measured. The index is already restricted to rows whose vector disagrees with their content — near zero in a healthy corpus whatever the status — so the status term adds little selectivity. On a deliberately pathological seed (every row stale, statuses even) it covered 3002 rows against the 1002 wanted, and `EXPLAIN` confirms Postgres still uses it for the narrower query, applying status as a filter on top:

```
Index Scan using ix_memories_stale_embedding on memories
  Index Cond: (tenant_id = 't3')
  Filter: (status = ANY ('{active,confirmed,pending}'))
```

A size question, not a correctness one. Happy to add the literal if you'd rather have the smaller index and accept the drift risk.

**Suggestion taken.** The per-tenant route folds from three sequential queries into one `memory_embedding_coverage_for_tenant` returning all four counts. Beyond the round trips, separate statements can't see a consistent snapshot — a concurrent write could leave the buckets failing to add up to the total they're reported against.

**Rebased onto main.** #787 landed, so this PR's core-worker changes are now covered by the CI that PR added — ruff, format, mypy and 71 tests all pass on a tree that had none when I first wrote them.

Root suite 4374 passed; mypy clean on core-api, core-storage-api and core-worker.

@claude please re-review.

Eldad-Caura:
Fixed — and correctly this time.

`memory_update` overrode an explicitly-supplied `embedded_content_hash` when the same patch also carried `content_hash`, contradicting the "explicit caller value wins" rule `_filter_memory_fields` documents on the insert path. Latent (core-worker sends provenance *without* `content_hash`, so nothing hits all three today), but it fails in the bad direction: replacing a caller's hash with the row's new one stamps a genuinely stale write as freshly embedded — the exact inversion this column exists to catch. Now derived only when the caller hasn't named it.

**My first regression test for it was worthless, and the control caught that.** The patch omitted `embedding`, so the derivation branch never fired and the assertion held with or without the fix. With `embedding` included the control discriminates properly: unconditional override → fails; guard → passes.

Also: the previous push failed CI on `ruff format --check (core-storage-api)`. My own local run had printed `1 file would be reformatted` and I formatted the wrong file. Fixed; every CI format scope is clean now.

Worth noting this PR is now gated by #787 — the core-worker steps ran here:

```
Run Ruff linter (core-worker)          All checks passed
Run Ruff formatter check (core-worker) files already formatted
Run mypy (core-worker)                 no issues found in 11 source files
Run core-worker tests                  71 passed
```

Those are this PR's own core-worker changes, on a tree that had zero coverage when I wrote them.

All checks green. Root suite 4375 passed.

Eldad-Caura:
@erni-a

erni-a (approved):

PR #787, opened 2026-08-16
ci: run core-worker's lint, types and tests

Eldad-Caura:
## core-worker had no CI at all

```
$ grep -rn "core-worker" .github/workflows/
(no matches)
```

Never installed, linted, type-checked, or run — by any workflow. While being the **production write path in deferred/SaaS mode** (`DEPLOYMENT_MODE=deferred`, which is what prod core-api runs). It owns `handle_embed_request` — essentially all production embedding — plus the embed-backfill sweep consumer and the enrich consumer.

Same gap #768 closed for core-operations, on a service with more production surface. This adds the package to the install step, plus ruff check, ruff format --check, mypy, and the test suite.

## Two things had already rotted behind it

**1. A test has been failing on a clean checkout.** `test_post_gives_up_after_max_attempts_on_connect_timeout` asserted a hardcoded `3`, but the connect-phase budget was deliberately raised to `CONNECT_PHASE_MAX_ATTEMPTS = 5` after the 2026-06-16 prod incident — *"3 attempts (~0.6s of backoff) is too short to ride out a cold start"*.

The production code is right; the test was stale. It now asserts **the constant itself** rather than a literal, so it moves with the budget instead of rotting again. Diagnosed before wiring the suite in, so the new step doesn't land red.

**2. caura#786** threaded embedding provenance through every write path *except* core-worker's `update_memory_embedding` — which would have left every production row permanently at `unknown_provenance`. Caught by human review. No automated check could have caught it, because none ran.

## Each gate verified to actually gate

A step that passes on broken input is worse than no step, so each was broken deliberately:

| injected | result |
|---|---|
| unused import | `ruff check` **FAILS** ✓ |
| bad return type | `mypy` **FAILS** ✓ |
| mangled spacing | `ruff format --check` **FAILS** ✓ |
| `assert False` | `pytest` **FAILS** ✓ |

All restored to green afterwards.

## Wiring notes

- **No `PYTHONPATH`,** deliberately. `core-worker/pytest.ini` sets `pythonpath = src ..` and pytest resolves it as the configfile for that path rather than inheriting the root `pytest.ini`, so an env var would be inert. Confirmed by running the suite with `PYTHONPATH` unset — **67 passed**.
- **No database.** The conftest pins `EVENT_BUS_BACKEND=inprocess`.
- **`--cov-config` is load-bearing**, same as core-operations: core-worker declares `fail_under = 50` and coverage discovers config relative to the CWD, so without the flag the run collects coverage while silently ignoring the floor. Measured before gating: **70.20%**, and the run prints *"Required test coverage of 50.0% reached"*.
- **Line-length 110** for core-worker, matching its siblings rather than `common/`'s 88 — so the format check is safe to gate.
- `actionlint`: the same single pre-existing SC2086 (unquoted `$GITHUB_PATH`) as `main` — one finding on each, unchanged. The line number shifts only because this adds comment lines above it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -74,10 +74,20 @@ jobs:
         # have broken confusingly the first time core-operations grew a
         # dependency core-api lacks. Resolution cost of adding it: one extra
         # package (119 vs 118).
+        #
+        # core-worker likewise, and for a sharper reason: it had NO CI at all —
+        # not installed, linted, type-checked or tested by any workflow — while
+        # being the production write path in deferred/SaaS mode. It owns
+        # ``handle_embed_request`` (essentially all production embedding), the
+        # embed-backfill sweep consumer and the enrich consumer. The cost of
+        # that gap is not hypothetical: caura#786 shipped embedding provenance
+        # through every write path EXCEPT core-worker's, which would have left
+        # every production row permanently at ``unknown_provenance``; it was
+        # caught by human review because no automated check covered the tree.
         run: |
           uv venv .venv
           source .venv/bin/activate
-          uv pip install -e core-storage-api/[dev] -e core-api/[dev] -e core-operations/[dev]
+          uv pip install -e core-storage-api/[dev] -e core-api/[dev] -e core-operations/[dev] -e core-worker/[dev]
           echo "$VIRTUAL_ENV/bin" >> $GITHUB_PATH
 
       # --- Broker↔cloud API-versioning gate (Gate C / Phase 1, deliverables A+B) ---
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
Good catch — fixed. The "Build packages" step's own comment makes the argument: core-operations is there because "a broken pyproject/build-backend would otherwise surface only in its Docker build." core-worker ships as a package with its own pyproject and had nothing in this workflow exercising any of it, so it belongs for the identical reason.

Confirmed `uv build` succeeds for it before wiring the step (`core_worker-2.28.0` sdist + wheel) rather than assuming — the whole point of this PR is not adding steps that land red.

For the record, the previous run proved the four original steps genuinely execute in CI, not just locally:

```
Run Ruff linter (core-worker)             All checks passed
Run Ruff formatter check (core-worker)    files already formatted
Run mypy (core-worker)                    Success: no issues found in 11 source files
Run core-worker tests                     67 passed
                                          Required test coverage of 50.0% reached
                                          Total coverage: 70.20%
```

@claude please re-review — the review job skips on synchronize.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #792, opened 2026-08-17
fix(enrichment): stop dropping atomic facts when their embedding fails

Eldad-Caura:
## The loss

The atomic-fact fan-out **dropped facts outright** when their embedding failed. Both exits in the loop landed before the write:

```python
except (TimeoutError, ValueError, ...):
    logger.warning("atomic-fact embed failed ... (skipping this fact)")
    continue
if child_embedding is None:
    continue          # ← silent
...
await sc.create_memory({...})   # never reached
```

This is data loss, not a repairable gap. Every recovery path in the system keys off a row with `embedding IS NULL`; none of them can see a row that was never written.

**The two arms are not symmetric, which is why this went unnoticed.** `get_embedding` *returns* None once its retry budget is exhausted rather than raising (`common/embedding/_service.py::_run_with_retry`), so under the embedding-gate saturation this path actually meets, the **silent** arm is the common one. It carried no attribution at this call site, so nothing tied a lost fact to a parent, a path, or a cause — the rate was unmeasurable by construction. (It was not globally silent: `_run_with_retry` logs its own terminal error. What was missing was *whose* fact was dropped and why.)

**core-worker already documents the invariant this restores.** Its consumer asserts:

> The auto-chunk parent and atomic-fact fan-out paths do land as NULL, so this does catch them.

True of auto-chunk. False of the fan-out until this PR.

## Changes

- **Persist the child with `embedding=None`** instead of skipping it.
- **Stamp `metadata.embedding_pending = True`.** That field is public API, not bookkeeping — `MemoryOut.metadata` documents it, agents are told to read it, and core-worker clears it when the vector lands. Without it an unembedded child is indistinguishable from a fully-embedded row to every consumer.
- **Schedule the re-embed via `_schedule_embed_or_reembed`** rather than leaving it to the nightly sweep. This is the part I'd most like reviewed: the sweep is the *floor*, not the mechanism. `embed_backfill_enabled` defaults to **False**, so a deployment that hasn't enabled it would strand these rows indefinitely — which is how ~430 memories were stranded in the 2026-07-27 incident this module already carries a postmortem for. Deferred mode publishes `EMBED_REQUESTED` (Pub/Sub owns retry/backoff/DLQ, paced by the consumer's per-tenant slots); inline mode retries in-process with the failure-fallback backoff. Publishing into a tier that just saturated sounds like piling on, but that's the case the postmortem already adjudicated — the incident was caused by *in-process* retry storms, and the durable path is the paced option.
- **Add a WARNING attributing the degradation to this path and parent.** The hourly coverage tick already reports the global backlog; what was missing was which path produced it.

`fanout_unembedded` is counted only **after** the write succeeds, so the summary can never claim a pending repair for a row that was never persisted.

## Tests

7 cases, parametrized across both failure arms: the fact is persisted, flagged `embedding_pending`, and gets a re-embed scheduled with `is_failure_fallback=True`.

**All 6 contract cases were confirmed to FAIL on unfixed code.** The 7th is a control — the happy path still attaches real vectors and schedules nothing — and passes either way by design.

Worth flagging: one test initially passed on unfixed code **for the wrong reason**. `all()` over an empty sequence is `True`, and the broken path writes no children, so the `embedding_pending` assertion was vacuously satisfied by exactly the bug it existed to catch. It now asserts the child count before inspecting them.

`ruff check`, `ruff format --check`, `mypy` (198 files) clean; full root suite **4383 passed**.

## Not in this PR

The sibling `update` path (`memory_service.py` ~:2696) persists `embedding=None` deliberately and also relies on the sweep with no scheduled re-embed. The same durable-handoff argument applies there if you want the invariant uniform — but it's a separate path with its own reasoning, and I didn't want to widen an incident fix.

### DIFF
--- a/core-api/src/core_api/services/memory_service.py
+++ b/core-api/src/core_api/services/memory_service.py
@@ -2462,25 +2462,47 @@ async def _enrich_memory_background(
             parent_visibility = mem.get("visibility") or "scope_team"
             parent_weight = patch.get("weight") or mem.get("weight") or 0.5
             fanout_created = 0
+            fanout_unembedded = 0
             for fact in atomic_facts:
                 fact_content = fact.content
+                # A failed embed must NOT skip the fact. Both exits here used
+                # to ``continue`` BEFORE ``create_memory``, so the child row
+                # was never written at all and the fact was lost outright —
+                # nothing downstream could repair what does not exist. Persist
+                # unembedded instead, exactly as the auto-chunk parent insert
+                # does, and hand the vector off to the normal recovery path
+                # below.
+                #
+                # The two arms are not symmetric in how often they fire.
+                # ``get_embedding`` RETURNS None once its retry budget is
+                # exhausted rather than raising (see
+                # ``common/embedding/_service.py::_run_with_retry``), so under
+                # the gate saturation this path actually meets, the None arm is
+                # the common one. It was not silent globally — ``_run_with_retry``
+                # logs its own terminal error — but nothing here attributed the
+                # loss to a parent, a fact, or this code path.
+                child_embedding: list[float] | None = None
                 try:
                     child_embedding = await get_embedding(fact_content, tenant_config=tenant_config)
                 except (TimeoutError, ValueError, RuntimeError, OpenAIError, GoogleAPIError):
                     logger.warning(
-                        "atomic-fact embed failed for memory %s (skipping this fact)",
+                        "atomic-fact embed raised for memory %s; persisting the fact unembedded",
                         memory_id,
                         exc_info=True,
                     )
-                    continue
-                if child_embedding is None:
-                    continue
                 child_ch = _content_hash(tenant_id, fleet_id, fact_content)
                 child_meta = {
                     "parent_memory_id": str(memory_id),
                     "source": "atomic_fact_fanout",
                     "retrieval_hint": fact.retrieval_hint or "",
                 }
+                if child_embedding is None:
+                    # ``embedding_pending`` is public API, not bookkeeping:
+                    # ``MemoryOut.metadata`` documents it, agents are told to
+                    # read it, and core-worker clears it when the vector
+                    # lands. Without it a fan-out child is indistinguishable
+                    # from a fully-embedded row to every consumer.
+                    child_meta["embedding_pending"] = True
                 # Intentionally NOT wrapped in ``per_tenant_storage_slot``
                 # (CAURA-602 follow-up): this site runs inside
                 # ``_enrich_memory_background``, a fire-and-forget task
@@ -2495,7 +2517,7 @@ async def _enrich_memory_background(
                 # storage-pool occupancy, revisit by giving the task
                 # its own deadline first.
                 try:
-                    await sc.create_memory(
+                    child = await sc.create_memory(
                         {
                             "tenant_id": tenant_id,
                             "fleet_id": fleet_id,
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@claude please re-review — the amended commit restructures the block you flagged.

Your Low finding was correct and it was a defect in this PR's own safety net, so I've addressed both halves rather than just the counter:

1. **`fanout_unembedded` now increments only once the re-embed is actually queued**, not whenever the embedding is None. Previously the summary WARNING could assert "re-embed scheduled for each" for a child that never got one.
2. **A child persisted without a usable id now logs a distinct `logger.error`** naming the parent and the raw response, and is deliberately NOT folded into the counter — that state is strictly worse than the counted one (unembedded with no repair queued, recoverable only by the nightly sweep, and only where it's enabled).
3. **The post-write block moved OUTSIDE the `try`.** You were right that a failure in the recovery step would surface as `"atomic-fact create_memory failed"` for a row that was in fact written — mislabelling the one log an operator would use to decide whether the fact exists. The `except` now covers the write alone and `continue`s, so there is nothing to schedule when there is no row.

Added `test_missing_child_id_is_reported_not_counted_as_scheduled`, which drives `create_memory` returning `{}` and asserts the rows are still written but nothing is scheduled.

Suite is 4384 passing; ruff check, ruff format --check and mypy (198 files) clean.

Eldad-Caura:
@claude final pass please — your Medium is fixed, and one test-coverage tradeoff needs a second opinion.

**Fixed as you described.** `tracked_task` now receives the CHILD's id, not the parent's. Both the coroutine and the wrapper get the same `child_uuid`, matching every other call site in the file. Comment at the call site records why, since the two ids are adjacent and easy to swap back.

**Disclosure on how it's verified.** I first tried to assert the label directly by stubbing `tracked_task` — that does not work, and the reason is worth stating: `tracked_task` is `async def`, so at the call site it only *creates* a coroutine, and under a no-op `track_task` its body never runs. A stub is never invoked, so the assertion was silently unfalsifiable. I removed it rather than keep a test that cannot fail.

What the tests assert instead: the re-embed is scheduled against the child rows and **not** against the parent id (read from `metadata_.parent_memory_id`), and each child gets its own distinct target. That covers the substantive property — a re-embed pointed at the parent would re-embed the wrong content and leave the child unrepaired. **The `tracked_task` label itself is verified by inspection only**, not by test. If you think that gap is worth closing, the honest way is a test that actually awaits the wrapper, and I'd rather you tell me it's needed than add unfalsifiable coverage.

Also in this push: rebased onto current `main` (`8d2bad46`, PR #793 landed mid-review) and re-ran everything on the new base — ruff, ruff format, mypy (198 files), and the 8 tests.

Verification state: **7 of the 8 tests fail against `origin/main`**, confirmed by checking out the pre-fix file rather than by assumption; the 8th is a control asserting the happy path still attaches real vectors and schedules nothing, and passes either way by design. Full suite 4384 passing.

Eldad-Caura:
@erni-a ready for review.

All checks green, single commit, rebased onto current `main`. Three automated review rounds, each of which found something real — the last one a content-leak I introduced, so it's worth knowing what changed:

1. **Round 1** — the re-embed scheduling gap was silently swallowed. `fanout_unembedded` incremented whenever the embedding was None, but scheduling was gated on having a usable child id, so the summary could claim "re-embed scheduled for each" when none was. Now counted only once the repair is actually queued, with a distinct `logger.error` for the un-schedulable case. Also moved the post-write block outside the `try`, which was mislabelling recovery-step failures as `"create_memory failed"` for rows that were in fact written.
2. **Round 2** — `tracked_task` was labelled with the PARENT id while the coroutine targeted the child, so a failed child re-embed would have been filed against the parent and the row needing repair left untraceable.
3. **Round 3** — my new `logger.error` interpolated the whole `create_memory` response with `%r`. In prod that response is the created row, so it would have written raw fact content, and any PII in it, into an ERROR log. Now logs the response KEY SET instead, which is what actually diagnoses the failure and carries no content. I re-audited every log line this PR adds: only UUIDs and that key list are interpolated.

**What the change does:** persists the fact unembedded instead of dropping it, stamps `metadata.embedding_pending = True` (public API — agents are told to read it, core-worker clears it), and schedules a durable re-embed rather than waiting on the nightly sweep. That last point is the one I'd most like your eye on: `embed_backfill_enabled` defaults to **False**, so leaning on the sweep alone would strand these rows on any deployment that hasn't enabled it.

**One coverage gap, stated plainly:** the `tracked_task` label is verified by inspection, not by test. I tried to assert it and found the assertion was unfalsifiable — `tracked_task` is `async def`, so under a no-op `track_task` its body never runs and a stub is never invoked. I removed it rather than keep a test that cannot fail. The tests do assert the substantive property: the re-embed targets the child rows and not the parent.

Verification: 7 of the 8 tests were confirmed to FAIL against `origin/main` by checking out the pre-fix file, not by assumption. The 8th is a control that passes either way by design. Full suite 4384 passing; ruff, ruff format, mypy (198 files) clean.

erni-a (approved):

PR #794, opened 2026-08-17
fix(extraction): drop malformed items instead of the whole extraction

Eldad-Caura:
Closes the class that #788 fixed one instance of.

## The exposure

#788 made `Mention.cluster_id` nullable after a null value on mention 10 of 15 destroyed a whole prod extraction. **Seven sibling non-optional fields carried the identical exposure** — `canonical_name`, `entity_type`, `role`, `from_entity`, `relation_type`, `to_entity`, `surface` — because `ExtractedGraph(**raw)` validated the payload atomically. Nothing enforces the schema we send (`strict=False` on OpenAI, accepted-and-ignored on gemini/vertex), so a `"role": null` reproduced the incident exactly.

Measured against `origin/main` with a stub provider — one null `role` on the second of two entities:

```
before:  entities ['anna bergstrom']   relations 0    ← regex fallback's guess
after:   entities ['anna']             relations 1
```

The LLM's real extraction was discarded and `_fake_extract`'s capitalised-phrase heuristic answered instead. And because `_do_extract` pins a stable CRC32 seed *so that retries reproduce identical output*, the primary's retry budget was spent on a call that could not succeed.

## Why the boundary and not the fields

Defaulting every field loses on three counts:

- "every field must have a default" is an unenforceable convention — the next required field silently reopens the hole, and nothing fails loudly
- a default turns a malformed item into a silently **accepted** one: `entity_type=""` passes `_is_valid_entity` (length ≥ 2 + blocklist) and persists as a real entity row; `role=""` breaks `_reattach_subject_discriminators`, which keys on `role == "subject"`
- the repo already ran this experiment — `common/enrichment/schema.py` *is* the defaults-everywhere design, and `_validate_enrichment` still needed a hand-rolled item-by-item loop, because `AtomicFact.content` is legitimately required

A counted drop beats quiet pollution.

## The fallback decision, which is the subtle part

A total loss still raises, so the fallback chain gets its turn — an empty graph would be indistinguishable from a legitimately entity-free memory. But the predicate is keyed on **entities/relations with no surviving entities**, not on all three lists:

- `mentions` has no consumer that persists anything, so a surviving mention must not suppress the fallback — the worker early-returns on an entity-less graph, losing the extraction with no audit row
- conversely a malformed mention **alone** must not trigger it, or the regex heuristic invents entities the model correctly reported as absent

Both directions are tested, and both fail with the naive three-list predicate — verified by reverting just that line.

Raises `ProviderResponseShapeError` rather than bare `ValueError`, so the remaining #788 follow-up (classify shape failures as non-retryable in `common/llm/retry.py`) has one type to key on rather than a bare `ValueError` that shared infra can't safely catch.

## Logging

Counts only, never items — the payload is memory content plus any PII in it. Flat `key=value` with a stable leading slug (`entity_extraction_items_dropped`), matching the `contradiction_detector` skip-log convention so `grep` works regardless of renderer.

`kept` is read off the parsed graph, **not** re-derived from `raw`. An earlier draft computed `len(raw.get(f))`, which raises `TypeError` on a non-list field — *inside the salvage branch*, discarding the graph it had just rescued. Caught in review; there's now a test for the non-list case.

## Tests — 22

Sibling-field cases assert the good item survives while showing the atomic parse raises on the same payload. `_GRAPH_ITEM_MODELS` coverage is pinned by a test rather than a comment, because a list field absent from the registry is never walked and silently defaults to empty — no runtime behaviour reports that.

## Noted, not done

`common/enrichment/service.py:186` — `EnrichmentResult(**raw)` is still an atomic parse of provider output. Every field is defaulted and most are pre-sanitised, but `summary` and `pii_types` aren't, so a `summary: 5` loses the whole enrichment to `fake_enrich`. Same shape, one level up. Recorded in `_parse_graph_lenient`'s docstring so the next person chasing this class doesn't read the cross-reference and conclude enrichment is already safe.

### DIFF
--- a/core-api/src/core_api/services/entity_extraction.py
+++ b/core-api/src/core_api/services/entity_extraction.py
@@ -4,13 +4,33 @@ import logging
 import re
 import zlib
 
-from pydantic import BaseModel
+from pydantic import BaseModel, ValidationError
 
 from core_api.config import settings
 from core_api.protocols import LLMProvider
 from core_api.providers._retry import call_with_fallback
 
 logger = logging.getLogger(__name__)
+
+
+class ExtractionShapeError(ValueError):
+    """Provider output could not yield a usable graph.
+
+    Subclasses ``ValueError`` so ``call_with_retry``'s broad ``except`` still
+    routes it to the fallback chain unchanged, while giving the pending #788
+    follow-up — classifying shape failures as non-retryable in
+    ``common/llm/retry.py`` — one narrow type to mark rather than a bare
+    ``ValueError`` that shared infra cannot safely single out.
+
+    Deliberately NOT ``common.llm.providers.ProviderResponseShapeError``, which
+    is a transport-layer error: its contract is ``(provider, content,
+    parsed_type)`` and it STORES and renders ``content`` (1 KiB) in ``__str__``.
+    Passing this payload through it would put memory content, and any PII in
+    it, into every traceback that error reaches. This condition is a domain
+    one — the response parsed fine, it just held nothing usable — so it carries
+    diagnostic counts only and never the payload.
+    """
+
 
 EXTRACTION_PROMPT = """\
 Extract named entities, their relations, and surface-form mentions from the following memory content.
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@claude re-review please — your High is fixed, but not the way the prompt suggested, and the reason matters.

You were right that `ProviderResponseShapeError(<single string>)` raises `TypeError`: the signature is `(provider, content, parsed_type)`. Masked by `call_with_retry`'s broad `except`, so the fallback still ran — but the type was wrong, which defeated the whole point of raising it (giving the pending #788 retry-classification follow-up something to key on).

**I did not satisfy that constructor, because doing so would reintroduce a leak I fixed in #792.** `ProviderResponseShapeError.__init__` STORES `content` (truncated to 1 KiB) and renders it in `__str__`. The `content` here is the provider's response — i.e. memory text and any PII in it — so passing it would put that into every traceback and log line the error reaches. That is a transport-layer contract being borrowed for a domain condition: the response parsed fine, it just held nothing usable.

Instead: a narrow local `ExtractionShapeError(ValueError)`. Subclassing `ValueError` keeps today's fallback behaviour byte-identical, gives the follow-up one precise type, and carries diagnostic counts only with no payload slot at all.

Added `test_extraction_shape_error_is_a_valueerror_and_carries_no_payload`, and it exists specifically because of the gap your finding exposed: the end-to-end fallback tests pass **regardless of the raised type**, since retry catches broad `Exception`. They could never have caught this. The new test asserts the ValueError relationship, that counts survive in the message, and that the error has no `content` attribute.

Suite 4407 passing; ruff, ruff format, mypy (198 files) clean. All checks green on the amended commit.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #795, opened 2026-08-17
fix(enrichment): tolerate a junk summary or pii_types

Eldad-Caura:
Closes the follow-up #794's docstring pointed at.

## The exposure — narrower than the note implied

`EnrichmentResult` is built from unenforced LLM output at three sites, and the caller recovers from a ValidationError by falling through to `fake_enrich`. So one junk field discarded a whole good enrichment — title, tags, weight, timestamps, atomic facts and all — and replaced it with the heuristic.

I fuzzed every field rather than trusting the earlier note. `_validate_enrichment` **already** guards `memory_type`, `status`, `weight`, `title`, both timestamps, `retrieval_hint`, `atomic_facts`, `contains_pii`, `business_relevance`; `tags` has a model-level validator. Exactly two were unguarded, and every junk shape broke them:

| field | shapes that raised |
|---|---|
| `summary` | `5`, `None`, `{...}`, `[...]` |
| `pii_types` | `"email"`, `[1, 2]`, `{...}`, `["email", None]` |

`pii_types` was *partially* handled by `raw.get("pii_types") or []` — which covers falsy values but not a truthy bare string or a list with non-string members.

## Fixed on the model, not in the service

Mirrors the `_normalize_tags` validator already in that file. Only **one** of the three construction sites routes through `_validate_enrichment`, so the model is the layer that actually covers all of them, and a future fourth site inherits the guard.

`summary` **discards** a non-str rather than stringifying it: the field is user-visible and lands in `metadata["summary"]` verbatim, so blank reads as "no information" while `"{'a': 1}"` reads as prose the model never wrote.

`pii_types` **wraps** a bare string rather than dropping it — but not for the reason I first wrote. My initial comment claimed this "feeds the governance PII gate, where losing a declared type is the unsafe direction." **That was wrong**, and review caught it: the gate keys on `contains_pii` alone (`governance_decision.py:66`, `governance_remediation.py:62`), so this cannot change a drop/flag decision. The real reason is audit evidence — a compliance review reading `categories: []` on a flagged row cannot tell "no type declared" from "type lost to normalisation". Same choice, corrected rationale.

## Deliberately NOT extended to `title` / `retrieval_hint`

My first draft covered them, and that was worse than useless. `_validate_enrichment` rewrites both to `str` **before** a model validator runs, so the validator can never see a non-str value there — listing them advertises a protection that never fires.

The consequence is documented rather than fixed:

```
title={'a': 1}  ->  "{'a': 1}"
title=None      ->  "None"
```

A junk title still persists as its repr. That's a display bug, not a crash, and fixing it means moving truncation onto the model too — out of scope for a crash fix, but it's now stated in the source instead of implied away.

## Also

Updates the cross-reference in `entity_extraction.py` that said this outer parse was still atomic. That note is what pointed here; leaving it would send the next reader after a gap that no longer exists.

## Tests — 12

**9 confirmed to fail against `origin/main`.** The 3 that pass are genuine controls: a clean payload, plus the two `pii_types` shapes the existing `or []` already handled.

The non-dict shape guard (CAURA-651) is **not** re-tested here — `tests/test_vertex_response_shape.py::TestEnrichmentValidatorRejectsNonDict` already covers it with a stronger message assertion, so duplicating it with a bare `pytest.raises(ValueError)` would have been weaker coverage wearing the same name.

`ruff check common/` and `core-api/src/` clean, `mypy` clean (198 files), full suite **4419 passing**. `common/enrichment/` is not in the enterprise vendored-files manifest, so there's no drift gate to re-sync.

## Noted, not done

`AtomicFact` has no equivalent guard — not an oversight: the service's item-walk pre-cleans facts into str-only dicts before the model sees them. Recorded in the validator docstring.

The remaining LLM-output parse boundaries all live in `core-api/src/core_api/services/` (`dedup_judge`, `crystallizer`, `business_classifier`, `contradiction_detector`, `insights`, `interview`, `agent_digest`, `evolve`). A review sweep found them all already item-tolerant, but that's where the same audit would go next.

### DIFF
--- a/common/enrichment/schema.py
+++ b/common/enrichment/schema.py
@@ -112,3 +112,65 @@ class EnrichmentResult(BaseModel):
     retrieval_hint: str = ""
     atomic_facts: list[AtomicFact] | None = None
     llm_ms: int = 0
+
+    @field_validator("summary", mode="before")
+    @classmethod
+    def _coerce_summary(cls, raw):
+        """Defensive normalisation, same reason as ``_normalize_tags`` above.
+
+        This model is built from unenforced LLM output, and the caller recovers
+        from a ValidationError by falling through to ``fake_enrich`` — so one
+        junk field discarded a whole good enrichment: title, tags, weight,
+        timestamps, atomic facts and all. ``summary`` was the last field
+        nothing guarded (``_validate_enrichment`` coerces the rest, ``tags``
+        is handled above), and a ``summary`` of ``5`` / ``None`` / ``{...}`` /
+        ``[...]`` raised.
+
+        SCOPED TO ``summary`` DELIBERATELY. ``title`` and ``retrieval_hint``
+        look like they belong here and do not: ``_validate_enrichment`` already
+        rewrites both to ``str`` BEFORE this validator runs, so listing them
+        would advertise a protection that never fires. Note the consequence
+        left in place — a junk ``title`` still persists as its repr
+        (``{'a': 1}`` becomes ``"{'a': 1}"``, ``None`` becomes ``"None"``).
+        That is a pre-existing display bug, not a crash, and fixing it means
+        moving truncation onto the model too; kept out so this stays a crash
+        fix.
+
+        A non-str value is DISCARDED rather than stringified: ``summary`` is
+        user-visible and lands in ``metadata["summary"]`` verbatim, so blank
+        reads as "no information" while ``"{'a': 1}"`` reads as prose the model
+        never wrote.
+        """
+        if raw is None:
+            return ""
+        return raw if isinstance(raw, str) else ""
+
+    @field_validator("pii_types", mode="before")
+    @classmethod
+    def _coerce_pii_types(cls, raw):
+        """Drop non-string entries instead of failing the whole enrichment.
+
+        ``_validate_enrichment`` normalises falsy values via
+        ``raw.get("pii_types") or []``, which leaves the two shapes the LLM
+        actually produces: a bare string (``"email"``) and a list with
+        non-string members (``[1, 2]``). Both raised.
+
+        A bare string is WRAPPED rather than dropped. Note what this does and
+        does not affect: the PII gate keys on ``contains_pii`` alone
+        (``governance_decision.py`` / ``governance_remediation.py``), so this
+        cannot change a drop/flag decision. It is the audit trail and row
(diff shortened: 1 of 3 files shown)


Eldad-Caura:
@erni-a ready for review. All checks green, automated review clean.

**What it fixes:** `EnrichmentResult` is built from unenforced LLM output and its caller recovers from a ValidationError by falling through to `fake_enrich` — so one junk field discarded a whole good enrichment (title, tags, weight, timestamps, atomic facts) and replaced it with the heuristic.

**The exposure was narrower than the note from #794 implied.** I fuzzed every field rather than trusting it: `_validate_enrichment` already guards nine fields and `tags` has a model validator. Exactly two were unguarded — `summary` (every junk shape broke it) and `pii_types` (the `or []` covers falsy values but not a truthy bare string or a list with non-string members).

Two things worth your eye, both places where review corrected me:

1. **My first comment asserted a governance-gate safety property that does not exist.** I wrote that `pii_types` "feeds the governance PII gate, where losing a declared type is the unsafe direction." The gate keys on `contains_pii` alone (`governance_decision.py:66`, `governance_remediation.py:62`) — this field cannot change a drop/flag decision. The wrapping choice is still right, but for audit evidence, not gate safety. I verified that myself before rewording, and fixed the same claim where it had been duplicated into the test.

2. **My first draft also covered `title` and `retrieval_hint`, which was worse than useless.** `_validate_enrichment` rewrites both to `str` BEFORE a model validator runs, so the coverage could never fire while implying titles were protected. They are not: a junk title still persists as its repr (`{'a': 1}` → `"{'a': 1}"`, `None` → `"None"`). I scoped the validator to `summary` and documented that consequence in the source rather than papering over it. Fixing it properly means moving truncation onto the model too — deliberately out of scope for a crash fix, and your call whether it's worth its own change.

Verification: 9 of 12 tests confirmed to FAIL against `origin/main`; the 3 that pass are genuine controls. I did not re-test the non-dict shape guard — `test_vertex_response_shape.py::TestEnrichmentValidatorRejectsNonDict` already covers it with a stronger message assertion, and duplicating it behind a bare `pytest.raises(ValueError)` would have been weaker coverage wearing the same name.

`common/enrichment/` is not in the enterprise vendored-files manifest, so there is no drift gate to re-sync.

erni-a (approved):

PR #796, opened 2026-08-17
feat(llm): let a caller declare which failures are not worth retrying

Eldad-Caura:
The last follow-up from #788 — but **not** the way #788 framed it, and that difference is the substance of this PR.

## The waste

`call_with_retry` catches broad `Exception` and spends its whole budget on every failure, including ones that cannot succeed on a second attempt. Entity extraction pins `seed` to a CRC32 of its prompt *specifically so retries reproduce byte-identical output* — so a shape failure there is guaranteed to recur. The retry costs a paid LLM call plus a 1s sleep, per provider, to learn nothing.

Measured on the real path, stubbed provider, total-loss payload:

```
before:  4 provider calls   (2 attempts × 2 providers)
after:   2 provider calls   (1 attempt  × 2 providers)
```

## Why the follow-up as written would have been harmful

#788 recorded this as *"classify `ValidationError` as non-retryable in `common/llm/retry.py`"*. Implementing that literally would have been a regression.

**A retry is only pointless when the call is deterministic, and determinism comes from the caller.** I checked every `complete_json` caller: **only entity extraction pins a seed.** Enrichment and the other ~10 consumers are unseeded, so a re-ask genuinely may return parseable output and their retries earn their place. A global classification would have stripped a useful retry from all of them to save one wasted call on the single deterministic one.

So this is **opt-in**: a `non_retryable` tuple defaulting to `()`. `isinstance(exc, ())` is `False` for every exception, which makes the default a true no-op rather than a behaviour change — that matters on a module this widely consumed.

Keyed on exception **type**, not on seeded-ness, because a seeded call still has non-deterministic failure modes: a network timeout must be retried no matter how fixed the seed is. Seeding is what makes a *shape* failure deterministic; the type is what separates shape from transport.

## Scope of the skip

It bounds retries **within** a provider only. `call_with_fallback` still advances to the fallback provider, which is the whole point — a deterministic failure means "this model cannot parse it", not "no model can", and the fallback is a different model. It's the same reasoning that made a total loss raise rather than return empty in #794: skipping the wasted re-ask must not cost the provider hop.

Entity extraction opts in with `(ExtractionShapeError, ValidationError)`. It is currently the only caller that can honestly make the claim.

## Tests — 7

- the declared type runs **once**
- an undeclared transport failure still uses its **full** budget (the classification must be narrow, not a blanket fail-fast)
- the default is a **no-op** — the guard for every caller this PR doesn't touch
- subclasses are covered, via `isinstance` semantics
- the happy path after a retryable failure is unchanged
- **the fallback provider still gets its turn** after a non-retryable primary
- the heuristic still answers when both providers fail the same way

`core-worker`'s suite (71 tests) also exercises this module and passes; full root suite **4426 passing**. `ruff check common/` + `core-api/src/` clean, `mypy` clean (198 files).

Note on the diff: `common/llm/retry.py` is +48 / −0 — I deliberately did not run `ruff format` on it, since `common/` uses an 88-col config and this file isn't in the format-check scope, so formatting it would have reflowed unrelated code into the diff.

## Honest assessment of the value

Modest, and smaller than when #788 filed it. After #794 only a *total*-loss payload raises from that path, so these events are rarer than they were. What this buys is one fewer paid call and ~1s less latency per occurrence, plus the determinism reasoning now living in code where the next person can see why a global classification would be wrong.

### DIFF
--- a/common/llm/retry.py
+++ b/common/llm/retry.py
@@ -34,6 +34,7 @@ async def call_with_retry(
     max_attempts: int = LLM_RETRY_ATTEMPTS,
     base_delay: float = LLM_RETRY_DELAY_S,
     timeout: float | None = None,
+    non_retryable: tuple[type[BaseException], ...] = (),
 ) -> T:
     """Call *coro_fn* with retry and linear backoff.
 
@@ -49,6 +50,31 @@ async def call_with_retry(
         Base delay in seconds; actual delay = ``base_delay * attempt_number``.
     timeout:
         Optional per-attempt timeout in seconds.
+    non_retryable:
+        Exception types that must NOT be retried within this provider — the
+        first occurrence raises immediately. Empty by default, so every
+        existing caller keeps its current behaviour unchanged.
+
+        OPT-IN, and deliberately not a global classification. The obvious
+        version of this — "a ValidationError can't succeed on retry, so never
+        retry one" — is WRONG for most callers here. It is only true when the
+        call is deterministic, and determinism comes from the CALLER: only
+        entity extraction pins a ``seed`` (a CRC32 of its prompt, precisely so
+        retries reproduce byte-identical output). Every other caller,
+        enrichment included, is unseeded, so a re-ask genuinely may return
+        parseable output and its retry is worth having. Classifying globally
+        would strip a useful retry from ~10 services to save one wasted call
+        on the single deterministic one.
+
+        Keyed on exception TYPE rather than on seeded-ness because a seeded
+        call still has non-deterministic failure modes — a network timeout
+        must be retried no matter how fixed the seed is. Seeding is what makes
+        a SHAPE failure deterministic; the type is what separates shape from
+        transport.
+
+        Note this bounds retries WITHIN a provider only. ``call_with_fallback``
+        still advances to the fallback provider afterwards, which is correct:
+        that is a different model and may well parse what this one mangled.
 
     Raises the last exception if all attempts are exhausted.
     """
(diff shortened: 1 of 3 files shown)


Eldad-Caura:
@erni-a ready for review. All checks green, automated review clean.

**The headline is that #788's own framing of this follow-up was wrong, and I did not implement it as written.** #788 recorded it as "classify `ValidationError` as non-retryable in `common/llm/retry.py`". Doing that literally would have been a regression: a retry is only pointless when the call is *deterministic*, and determinism comes from the caller. I checked every `complete_json` caller — **only entity extraction pins a seed**. Enrichment and the other ~10 consumers are unseeded, so a re-ask genuinely may return parseable output and their retries earn their place. A global classification would have stripped a useful retry from all of them to save one wasted call on the single deterministic one.

So it is opt-in: a `non_retryable` tuple defaulting to `()`, which is a true no-op (`isinstance(exc, ())` is False for every exception). That matters on a module this widely consumed — nothing changes for any caller that does not ask for it.

Keyed on exception type rather than on seeded-ness, because a seeded call still has non-deterministic failure modes: a timeout must be retried no matter how fixed the seed is. Seeding makes a SHAPE failure deterministic; the type separates shape from transport.

**Measured saving on the real path** (stubbed provider, total-loss payload): 4 provider calls → 2. The skip is within a provider only — the fallback still gets its turn, which is the same reasoning that made a total loss raise rather than return empty in #794, and there are two tests specifically pinning that.

Worth a glance: `common/llm/retry.py` shows +48/−0 because I deliberately did NOT run `ruff format` on it. `common/` uses an 88-col config and this file is not in the format-check scope, so formatting it would have reflowed unrelated code into the diff.

**Honest assessment of value:** modest, and smaller than when #788 filed it — after #794 only a total-loss payload raises from that path, so these events are rarer. What it buys is one fewer paid call and ~1s less latency per occurrence, plus the determinism reasoning now living in code where the next person can see why the obvious version is wrong. If you would rather not carry a new parameter on shared infra for that, closing this is a defensible call and I would not argue.

core-worker's suite (71 tests) also exercises this module and passes; full root suite 4426.

erni-a (approved):

PR #798, opened 2026-08-17
fix(enrichment): discard a junk title instead of persisting its repr

Eldad-Caura:
## What

`_validate_enrichment` coerced the LLM's `title` with a bare `str()`:

```python
raw["title"] = str(raw.get("title", ""))[:80]
```

So a non-string reached storage as its Python repr — `{"a": 1}` became the
string `"{'a': 1}"`, and `None` became `"None"`. Both render as a title the
model never wrote.

This was deliberately left out of #795 to keep that a crash fix; #795's
`_coerce_summary` docstring recorded it as a known consequence. That
paragraph is now stale and is removed here.

## Why discard rather than stringify

Same reasoning `_coerce_summary` gives for `summary`, and it composes with a
convention one layer down: `core_worker/consumer.py::_build_patch` skips an
empty `title` **specifically** so it cannot clobber a previously-stored good
one. Discarding is therefore a storage no-op that preserves a valid title,
where a repr would overwrite it on redelivery.

## Why the guard stays in `_validate_enrichment`

An earlier note suggested moving truncation onto the model. Kept it in the
service instead:

- `retrieval_hint` — `title`'s nearest sibling, 28 lines below — already
  isinstance-guards there with the identical shape. Moving only `title` would
  split one concern across two files and invalidate the `_coerce_summary`
  docstring, which explains the two fields *as a pair*.
- The 80-cap is enforced only at this boundary; `fake_enrich` does not pass
  through it, so moving the cap onto the model would start truncating
  heuristic titles for the first time — a behaviour change riding inside a
  coercion fix.
- The only untrusted construction site, `EnrichmentResult(**raw)`, is the last
  line of `_validate_enrichment`. The other production sites are trusted
  internal literals.

This is not a special case layered on shared infrastructure — `title` was the
odd one out, the single field in a sequential normalisation pass still using a
bare `str()` while every neighbour isinstance-guards.

## Verification

The new test was run against the unfixed code and **fails** on all four junk
shapes with the exact repr symptom:

```
FAILED ...test_junk_title_is_not_persisted_as_its_repr[number]
FAILED ...test_junk_title_is_not_persisted_as_its_repr[null]
FAILED ...test_junk_title_is_not_persisted_as_its_repr[object]
FAILED ...test_junk_title_is_not_persisted_as_its_repr[array]
```

(via `git checkout origin/main -- common/enrichment/service.py`, not
`git stash`, which is a no-op once the fix is committed.)

- Root suite: **4432 passed, 3 skipped, 1 xfailed**
- `ruff check common/ tests/...` clean; vendored-file format check clean.
  Deliberately did **not** run `ruff format` on `common/enrichment/` — CI's
  format scope there is only `__init__.py`, `structlog_config.py`, `serve.py`,
  and reformatting would reflow unrelated code.

The 80-char cap is left covered by the existing
`test_p4_1_llm_fallback.py::TestValidateEnrichment::test_title_truncated_to_80`
rather than duplicated, with a pointer comment following this file's own
convention.

### DIFF
--- a/common/enrichment/schema.py
+++ b/common/enrichment/schema.py
@@ -128,13 +128,9 @@ class EnrichmentResult(BaseModel):
 
         SCOPED TO ``summary`` DELIBERATELY. ``title`` and ``retrieval_hint``
         look like they belong here and do not: ``_validate_enrichment`` already
-        rewrites both to ``str`` BEFORE this validator runs, so listing them
-        would advertise a protection that never fires. Note the consequence
-        left in place — a junk ``title`` still persists as its repr
-        (``{'a': 1}`` becomes ``"{'a': 1}"``, ``None`` becomes ``"None"``).
-        That is a pre-existing display bug, not a crash, and fixing it means
-        moving truncation onto the model too; kept out so this stays a crash
-        fix.
+        normalises both to ``str`` BEFORE this validator runs, so listing them
+        would advertise a protection that never fires. Both discard a
+        non-string there rather than stringify it, for the same reason.
 
         A non-str value is DISCARDED rather than stringified: ``summary`` is
         user-visible and lands in ``metadata["summary"]`` verbatim, so blank
--- a/common/enrichment/service.py
+++ b/common/enrichment/service.py
@@ -107,7 +107,14 @@ def _validate_enrichment(raw: dict, llm_ms: int) -> EnrichmentResult:
     except (TypeError, ValueError):
         weight = 0.7
     raw["weight"] = max(0.0, min(1.0, weight))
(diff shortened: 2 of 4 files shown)


Eldad-Caura:
@erni-a ready for review — all checks green, Claude auto-review came back clean.

Small enrichment coercion fix: a non-string `title` from the LLM was being persisted as its Python repr (`{"a": 1}` → the string `"{'a': 1}"`, `None` → `"None"`). Now discarded instead, matching what `_coerce_summary` already does for the sibling field.

The new test was run against the unfixed code and fails on all four junk shapes, so it's a real gate rather than one that passes either way.

erni-a (approved):

PR #799, opened 2026-08-17
fix(core-api): add the missing capability gates on three write routes

Eldad-Caura:
First cluster from the 2026-08-14 OSS/platform audit: **H-12, H-13, H-15** — three write-shaped routes that skipped capability gates their own neighbours already applied.

Grouped because they're one bug repeated, not three unrelated fixes, and they're the same shape as the 2026-06-11 findings already covered in `tests/test_route_authz_gaps.py` — so the tests land beside them.

## H-13 · `POST /fleet/commands` — two bugs in one, and the most serious

The target tenant comes from the **request body** (`body.tenant_id or auth.tenant_id`) and was never checked:

```python
tenant_id = body.tenant_id or auth.tenant_id   # caller names the tenant
cmd = await sc.create_command({"tenant_id": tenant_id, ...})
```

So any authenticated caller could set `body.tenant_id` to a victim tenant and queue commands into their fleet. The **GET sibling immediately below has always called `enforce_tenant`** — the write was the weaker half of the pair. It also had no `enforce_read_only`, so a `capabilities={'read'}` credential could do the same.

Both gates added. Order matters: resolve the tenant from the body first, *then* enforce against the resolved value.

## H-12 · `PATCH /agents/{agent_id}/trust` — no `enforce_read_only`

The adjacent fleet-reassignment route had it; this one didn't. So a read-only credential could rewrite the control this file's own comment calls *"the master key to the whole ladder"*.

**`enforce_usage_limits` is deliberately NOT added**, unlike that neighbour. This is the route you reach for to **demote** a misbehaving agent, so an over-quota tenant must still be able to take trust away — quota state must not stand between an operator and a mitigation. A test pins this so nobody "completes" the fix by adding it.

## H-15 · `PUT /settings` — a hand-rolled check that covered half the cases

It tested `auth.is_demo` directly, which caught the demo sandbox but **not** a credential minted read-only by construction. So a viewer/reporting key could rewrite tenant settings — which matters because settings carry security-relevant toggles: `require_agent_approval` governs whether new agents start quarantined.

Replaced with `enforce_read_only`, which covers demo *and* capabilities together. That's precisely why write-shaped endpoints are meant to call the gate rather than test individual flags. Since the branch was **replaced rather than dropped**, a regression test pins that demo is still refused.

## Tests

6 added to `tests/test_route_authz_gaps.py`. **4 fail without this change**, verified by reverting the three route files:

```
FAILED test_fleet_command_cannot_be_queued_into_another_tenant
FAILED test_fleet_command_rejects_a_read_only_credential
FAILED test_agent_trust_rejects_a_read_only_credential
FAILED test_settings_rejects_a_read_only_credential
```

The other 2 pass either way **by design**, and I'd rather say so than imply all 6 detect the bug — they're pins on decisions, not detectors:

- `test_agent_trust_still_works_when_over_usage_limits` — pins the deliberate omission
- `test_settings_still_refuses_the_demo_sandbox` — pins that replacing the branch didn't lose coverage

The cross-tenant test also asserts the victim's queue is still empty afterwards, not just that the call 403'd.

## Checks

Full suite **4434 passed, 3 skipped, 1 xfailed** (`-m "not benchmark"`, against a local Postgres with the same credentials CI uses). `ruff check` and `ruff format --check` clean as separate steps on all four files. The 108 tests specific to the three touched routes (`test_api_fleet`, `test_api_agents`, `test_api_settings`, `test_agent_admin_authz`, `test_agent_identity_guard`, plus heartbeat/settings-invalidation) all pass.

Note: `ruff check` reports a pre-existing `I001` on `tests/test_route_authz_gaps.py:248` (inline imports in a 2026-06-11 test). It's present on `main` and untouched here — CI's ruff scope evidently doesn't include `tests/`, and reordering it would be unrelated churn in a security PR.

## Remaining highs in this audit

This is 3 of 8 security highs. Still open: **H-11** (perimeter boots wide open when `GATEWAY_SHARED_SECRET` is unset), **H-14**/**H-16** (client-trusted identity — `POST /search` and `GET /stm/notes` let an agent credential read a peer's scoped memories), **H-17** (`POST /stm/promote` bypasses the LTM write gates — note the 2026-06-11 pass fixed its read-only/usage gates, so this is the deeper set), **H-18** (fast write mode skips LLM governance). Plus 10 correctness highs.

### DIFF
--- a/core-api/src/core_api/routes/agents.py
+++ b/core-api/src/core_api/routes/agents.py
@@ -46,6 +46,16 @@ async def patch_agent_trust(
     auth: AuthContext = Depends(get_auth_context),
 ):
     """Update an agent's trust level (and optionally fleet)."""
+    # A read-only credential must not move the trust ladder. This gate was
+    # missing while the neighbouring fleet-reassignment route had it, so a
+    # capabilities={'read'} key could rewrite the very control the comment
+    # below calls the master key.
+    #
+    # NOT gated on ``enforce_usage_limits``, unlike that neighbour, and
+    # deliberately: this is the route you reach for to DEMOTE a misbehaving
+    # agent. An over-quota tenant must still be able to take trust away, so
+    # quota state must not stand between an operator and a mitigation.
+    auth.enforce_read_only()
     auth.enforce_tenant(tenant_id)
     # Trust changes are the master key to the whole ladder — an agent must not
     # be able to PATCH its own (or a peer's) trust_level to self-promote.
--- a/core-api/src/core_api/routes/fleet.py
+++ b/core-api/src/core_api/routes/fleet.py
@@ -949,10 +949,18 @@ async def create_command(
 ):
     """Queue a command for a fleet node."""
     sc = get_storage_client()
-    # We need to verify node exists and get its tenant_id
-    # The storage client get_node takes tenant_id + node_name, but we have node_id
-    # Create the command via the storage client
+    # Queueing a command is a write, and the tenant it lands in comes from the
+    # REQUEST BODY. Both gates are required, and the order matters: resolve the
+    # target tenant first, then enforce against the resolved value.
+    #
+    # Without ``enforce_tenant`` any authenticated caller could set
+    # ``body.tenant_id`` to a victim tenant and queue commands into their fleet
+    # — the GET sibling immediately below has always enforced this, so the write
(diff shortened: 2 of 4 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #802, opened 2026-08-18
fix(core-api): require GATEWAY_SHARED_SECRET in production

Eldad-Caura:
**H-11** from the 2026-08-14 OSS/platform audit — the fail-open perimeter. Third cluster, after #799 (capability gates) and #801 (read-path identity).

## The bug

The `X-Tenant-ID` auth path (`auth.py`, "Path 4") carries **no credential of its own**. It trusts the gateway to have authenticated the caller and injected the identity headers. Its perimeter check is:

```python
gw_secret = settings.gateway_shared_secret
if gw_secret and not hmac.compare_digest(request.headers.get("x-gateway-secret") or "", gw_secret):
    raise HTTPException(401, detail="Direct access to this service is not permitted.")
```

The comment above it says *"No-op when unset (OSS / standalone / dev)"*, and that's accurate — but the consequence is sharper than "no-op" suggests. **An unset secret does not weaken this path, it disables it.** Anyone able to reach core-api directly — its public `run.app` URL, a sidecar, anything inside the VPC — becomes any tenant by setting a header.

And nothing required the secret in production. A deployment that never configured it, or silently dropped it in a config change, served a fully open perimeter that looks identical to a working one from the outside.

## The fix

Production now refuses to boot without it, alongside the existing `SETTINGS_ENCRYPTION_KEY` / `ADMIN_API_KEY` / `IS_STANDALONE` guards it sits with.

**Why refuse to boot rather than 401 the path per request:** a silently-open perimeter is indistinguishable from a working one from outside, which is precisely how it reaches production unnoticed. A service that won't start gets caught at deploy. A request-time guard would also be dead code once this exists, since production can no longer reach that state.

## The extraction, and why it's in scope

The guards moved from inline in `lifespan` into `_validate_production_settings`. Behaviour is unchanged for the four existing guards.

This is here because the block was **untestable in place** — reaching it meant driving the entire startup sequence (Sentry, credential bridging, provider init, then `mcp_lifespan`). Which is why a block whose entire job is preventing unsafe production boots **had no tests at all**. Shipping a new security guard into that same untested hole seemed worse than extracting it.

## Tests

10 added in `tests/test_production_startup_guards.py`, covering **all five** guards rather than only the new one — the four pre-existing ones had no coverage before this.

**2 fail without this change** (verified by neutering only the new guard):

```
FAILED test_production_requires_the_gateway_shared_secret
FAILED test_an_empty_gateway_secret_is_refused_too
```

Two cases worth calling out specifically:

- `test_an_empty_gateway_secret_is_refused_too` — `""` is the shape a missing env var actually takes in a container, not `None`. A truthiness check covers both, but only a test says so.
- `test_a_secretstr_wrapped_default_is_still_caught` — exercises the `get_secret_value()` unwrap branch in the dangerous-defaults loop. A wrapped default would otherwise never compare equal to the bad value, silently bypassing the guard.

Non-production environments are unaffected and still boot bare, pinned by a parametrised test over `development` and `sandbox`.

## Checks

Full suite **4448 passed, 3 skipped, 1 xfailed** (`-m "not benchmark"`, local Postgres with CI's credentials). `ruff check` / `ruff format --check` clean as separate steps. `import core_api.app` verified — the extraction left `lifespan` intact.

## Remaining highs

5 of 8 security highs now covered. Still open: **H-17** (`POST /stm/promote` bypasses the LTM write gates — the June 2026 pass fixed its read-only/usage gates, so this is the deeper set: agent approval/trust, fleet-write policy, reserved memory types, write metering, rate limiter) and **H-18** (fast write mode on the default inline deployment never enforces LLM governance — PII-drop, non-business drop, `keep_private` silently skipped). Plus the 10 correctness highs (H-01 to H-10).

### DIFF
--- a/core-api/src/core_api/app.py
+++ b/core-api/src/core_api/app.py
@@ -115,6 +115,80 @@ class SecurityHeadersMiddleware:
         await self.app(scope, receive, send_with_headers)
 
 
+def _validate_production_settings(app_settings) -> None:  # type: ignore[no-untyped-def]
+    """Refuse to boot a production deployment that is missing a safety control.
+
+    Extracted from ``lifespan`` so these guards are reachable by tests. Buried in
+    the lifespan they were untestable in practice, which is why a block whose
+    entire job is to prevent unsafe production boots had no tests of its own.
+
+    No-op outside ``ENVIRONMENT=production``.
+    """
+    if app_settings.environment != "production":
+        return
+    if app_settings.is_standalone:
+        raise RuntimeError(
+            "IS_STANDALONE=true is not allowed in production. "
+            "Set IS_STANDALONE=false for production deployments."
+        )
+    if not app_settings.settings_encryption_key:
+        raise RuntimeError(
+            "SETTINGS_ENCRYPTION_KEY must be set when ENVIRONMENT=production. "
+            'Generate one with: python -c "from cryptography.fernet import Fernet; print(Fernet.generate_key().decode())"'
+        )
+    _dangerous = {
+        "jwt_secret": "change-me-in-production",
+    }
+    for var, bad_val in _dangerous.items():
+        val = getattr(app_settings, var, None)
+        # A SecretStr field wraps its value, so compare against the unwrapped
+        # one or the guard is silently bypassed — `SecretStr("x") == "x"` is
+        # False. Kept generic on purpose: nothing in ``_dangerous`` is SecretStr
+        # today (the only ones are platform_llm_api_key /
+        # platform_embedding_api_key), so this exists for whatever gets added
+        # next. The example that used to be here, postgres_password, no longer
+        # exists — core-api dropped its DB engine.
+        if hasattr(val, "get_secret_value"):
+            val = val.get_secret_value()
+        if val == bad_val:
+            raise RuntimeError(f"{var.upper()} must be changed from default for production")
+    if not app_settings.admin_api_key:
+        raise RuntimeError("ADMIN_API_KEY must be set for production")
+    if not app_settings.gateway_shared_secret and not app_settings.memclaw_api_key:
+        # What production actually requires is A PERIMETER — not specifically the
+        # gateway one. ``MEMCLAW_API_KEY`` is the other way to have one: when it
+        # is set, auth.py's "Path 2" either authenticates the request against
+        # that key or raises 401 for everything else, so "Path 4" below it is
+        # UNREACHABLE and there is no header-trust surface left to protect. That
+        # is the documented network-exposed OSS pattern, and such a deployment
+        # legitimately sets ENVIRONMENT=production for JSON logging and Sentry.
+        # Demanding a gateway secret it has no gateway for would be a boot
+        # failure with no security value.
+        #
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
Addressed — the finding is correct, and I verified the mechanism rather than taking it on faith.

Inside `if mclaw_key:` in `auth.py`, a request either matches the key or hits an unconditional `raise HTTPException(401, "Invalid API key.")` before Path 3/4 are ever reached. So with `MEMCLAW_API_KEY` set, **Path 4's header-trust surface is unreachable** and there is nothing left for a gateway secret to protect. My guard would have broken a documented deployment pattern (the `auth.py` comment calls it "optional, for network-exposed OSS") that legitimately sets `ENVIRONMENT=production` for JSON logging and Sentry.

Took option (1). The guard now requires **a perimeter**, not specifically the gateway one:

```python
if not app_settings.gateway_shared_secret and not app_settings.memclaw_api_key:
```

I preferred this to the escape-hatch flag in option (2) because a `GATEWAY_SHARED_SECRET_NOT_REQUIRED=true` setting is a control whose only function is to disable a control — it would be set once during a frustrated deploy and never revisited, and it cannot distinguish "we use MEMCLAW_API_KEY instead" from "we have no perimeter at all." Deriving the answer from what is actually configured has no such failure mode. The error message now names both options so an operator hitting it knows there are two ways out.

Tests updated: 12 now, with `test_memclaw_api_key_alone_is_an_acceptable_perimeter` and `test_gateway_secret_alone_is_an_acceptable_perimeter` pinning both valid shapes, and `test_empty_strings_count_as_unset` covering `""` on both since that is what a missing env var looks like in a container.

**Two unrelated notes:**

1. This branch was based on `main` before #801 merged, which is why GitHub asked for a rebase. Rebased onto `9beab138` and force-pushed with lease. My fault — I fetched at worktree-creation time and not again before pushing.

2. **`main` is currently red, independently of this PR.** `tests/test_embedding_coverage_admin.py::test_aggregate_is_worst_first` fails in a full-suite run but passes in isolation — an order-dependent failure. I confirmed it by checking out plain `9beab138` and running the suite there: `1 failed, 4442 passed`. Not introduced by this branch, and worth its own issue.

Eldad-Caura:
@claude

Eldad-Caura:
Fixed. Verified first — `grep -c postgres_password core-api/src/core_api/config.py` is `0`, and the only `SecretStr` fields today are `platform_llm_api_key` and `platform_embedding_api_key`, neither of which is in `_dangerous`.

I kept the comment generic rather than swapping in one of those two, since naming a field that isn't in `_dangerous` would be the same trap one rename later. It now says what the branch is actually for — `SecretStr("x") == "x"` is False, so an unwrapped comparison silently bypasses the guard — notes that nothing in `_dangerous` is `SecretStr` today so the branch exists for whatever gets added next, and records that `postgres_password` is gone because core-api dropped its DB engine (so nobody re-adds it looking for the old example).

For the record: I carried that comment over verbatim during the extraction, so the staleness predates this PR — but I moved it, so it's mine to fix.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #804, opened 2026-08-18
fix(core-api): make /stm/promote pay the LTM write gates

Eldad-Caura:
**H-17** from the 2026-08-14 OSS/platform audit. A promote **is** an LTM write, but it reached long-term memory having paid only `enforce_read_only` and `enforce_usage_limits` — so the STM door into LTM was cheaper than the front door at `POST /memories`.

The June 2026 pass gave promote those two gates and bound its `agent_id` to the authenticated identity. It did not bring the route to parity. Missing, all now applied in the write path's order:

| gate | why it matters here |
|---|---|
| `_reject_reserved_memory_type` | promote takes `memory_type` **from the body**, so a caller could land a server-reserved type the front door refuses |
| agent-approval policy | `require_agent_approval` was never consulted |
| `trust_level == 0` quarantine | a quarantined agent could still reach LTM |
| `enforce_fleet_write` | fleet-write policy unevaluated |
| `check_and_increment(tenant, "write")` | promotes were unmetered |

## Two things worth review attention

**`resolve_write_agent` REPLACES the bare `broker_owned_agent_id` call rather than joining it.** That helper is what `POST /memories` uses, and it subsumes the old call: it runs the same ownership degradation internally (line 33) *and* adds an `_owned_by_other_install` post-check, while returning the agent row the trust gate needs. Calling both would have done the ownership work twice.

Two existing broker tests patched `stm.broker_owned_agent_id`, a seam this removes. They now patch `agent_service.broker_owned_agent_id` — the real gate one level down — so they still exercise the boundary end-to-end through the new call path rather than through a mock of my own code.

**Metering is not the same as `enforce_usage_limits`**, which is why both belong. That gate refuses a write when the org is *already* over its plan cap; `check_and_increment` is what makes the write *count toward* it. Without metering a tenant could promote without limit and never trip the check.

## Deliberately not included

The `@write_limit` rate limiter `POST /memories` carries. It's slowapi's `limiter.limit`, which needs a `Request` in the endpoint signature that this route lacks. Threading one in unverified risked breaking the route for the least security-critical of the five gaps — worth a follow-up rather than a guess here.

## Tests

2 added, **both fail without this change**:

```
FAILED test_promote_rejects_a_server_reserved_memory_type
FAILED test_promote_refuses_a_quarantined_agent
```

Full suite **4443 passed, 3 skipped, 1 xfailed**. `ruff check` / `ruff format --check` clean as separate steps.

**One open thread I could not close and am not hiding:** I also wrote a happy-path test asserting an approved agent's promote returns 200, and it returned **422** — from something downstream of these gates, not from them (the quarantine test passing proves the gates themselves are ordered correctly). I removed the test rather than assert a status I don't understand or paper over it with a weaker assertion. Worth someone establishing what a legitimate promote actually requires — it may want an existing STM note — since the absence of a happy-path test here is a real gap in this PR's coverage.

## Audit status

7 of 8 security highs covered. **H-18** remains (fast write mode on the default inline deployment never enforces LLM governance — PII-drop, non-business drop, `keep_private` silently skipped), plus the 10 correctness highs H-01–H-10.

### DIFF
--- a/core-api/src/core_api/routes/stm.py
+++ b/core-api/src/core_api/routes/stm.py
@@ -7,13 +7,32 @@ import logging
 from fastapi import APIRouter, Depends, HTTPException, Query
 from pydantic import BaseModel
 
+from common.enrichment.constants import SERVER_RESERVED_MEMORY_TYPES
 from core_api.auth import AuthContext, get_auth_context
 from core_api.config import settings
-from core_api.services.agent_service import broker_owned_agent_id
+from core_api.services.agent_service import enforce_fleet_write, resolve_write_agent
+from core_api.services.usage_service import check_and_increment
 
 logger = logging.getLogger(__name__)
 
 router = APIRouter(tags=["stm"])
+
+
+def _reject_reserved_memory_type(memory_type: str | None) -> None:
+    """Twin of ``memories._reject_reserved_memory_type``, for the promote door.
+
+    Deliberately a small duplicate rather than an import of that module's
+    private helper: importing across route modules to reach a ``_``-prefixed
+    function invites a circular import for no real gain. The source of truth
+    both share is ``SERVER_RESERVED_MEMORY_TYPES``, so the two cannot disagree
+    about WHICH types are reserved — only about wording.
+    """
+    if memory_type is None or memory_type not in SERVER_RESERVED_MEMORY_TYPES:
+        return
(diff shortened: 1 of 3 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #806, opened 2026-08-18
fix(core-api): apply LLM governance on the inline fast-write path

Eldad-Caura:
**H-18** — the last of the eight security highs from the 2026-08-14 OSS/platform audit. PII-drop, non-business-drop and `keep_private` were computed and then **silently discarded for most writes**.

## Why this hid so well

Fast mode **does** run two governance steps — `GovernanceScanContent` (deterministic pattern scan) and `BusinessPersonalPregate` (opt-in pre-gate) — so the pipeline looks governed. Neither can see the LLM signal: `contains_pii` / `business_relevance` exist only on the enrichment result.

And fast mode has no `GovernanceDecision` step **correctly**: it *always* defers enrichment (per `ParallelEmbedEnrich`'s docstring), so `ctx.data["enrichment"]` doesn't exist at pipeline time and the step would no-op. Adding it there is the obvious wrong fix — I tried it, it changed nothing, and `test_governance_decision_wired_into_strong_only` pins its absence on purpose.

Fast mode is instead governed by **post-write remediation**, `remediate_after_enrichment`. That was wired **only** into `consumer.py`, which runs when the *worker* PATCHes enrichment back:

| deployment | enrichment runs | remediation ran? |
|---|---|---|
| deferred (Pub/Sub) | worker | ✅ via `consumer.py` |
| **inline (the default)** | in-process, `_enrich_memory_background` | ❌ **never** |

No publish → no consumer → no remediation. This adds the inline counterpart so both modes converge on the same code.

## Why it's gated rather than unconditional

Strong mode enforces the same policy synchronously via `GovernanceDecision`. Running remediation there too would **double-apply** — duplicate audit rows and a second drop attempt on a row policy already acted on.

Strong mode's call site cannot reach this branch today: it's guarded by `not settings.inline_enrichment`. So `run_governance_remediation` is defence against that guard being relaxed, not a live need — but threading it keeps the invariant local to the code instead of resting on a condition two files away.

Failures are caught and logged rather than raised. The row is already persisted and enrichment already landed, so a governance failure must not take the rest of the background fan-out (extraction, Path A) down with it. It logs at `exception` level, which the ops error alerting already watches.

## Tests

4 added: inline-fast remediates; strong mode does **not** double-remediate; a deferred deployment leaves it to the worker; a remediation failure doesn't break the chain.

**Honest note on their strength.** Without this change they fail with `TypeError: unexpected keyword argument` — which proves the signature is new but **not** that the old behaviour was broken. An opt-in flag has no "before" state to pass. The behavioural claim rests on there being no call to `remediate_after_enrichment` anywhere on the inline path beforehand, which is an argument rather than a test. If a reviewer wants a genuine before/after, it would need a test that drives a real fast-mode write end-to-end with governance enabled and asserts the row is dropped — worth doing, and more than I could verify here.

One implementation detail worth flagging: `inline_enrichment` is a derived **property** on `Settings` with no setter, so the tests patch it on the class rather than the instance.

## Checks

Full suite **4461 passed, 3 skipped, 1 xfailed**. `ruff check` / `ruff format --check` clean as separate steps. Rebased onto `main` immediately before pushing.

## Audit status

**8 of 8 security highs now covered** — #799 (H-12/13/15), #801 (H-14/16), #802 (H-11), #804 (H-17), this (H-18). The 10 correctness highs H-01–H-10 remain, including H-03, the confirmed prod recall bug.

### DIFF
--- a/core-api/src/core_api/pipeline/steps/write/schedule_background_tasks.py
+++ b/core-api/src/core_api/pipeline/steps/write/schedule_background_tasks.py
@@ -54,6 +54,9 @@ class ScheduleBackgroundTasks:
                             tenant_config,
                             agent_provided_fields=_agent_provided_enrichment_fields(data),
                             reference_datetime=getattr(data, "reference_datetime", None),
+                            # H-18: nothing applies the LLM governance verdict on
+                            # an inline deployment — see ``_schedule_enrich_or_inline``.
+                            run_governance_remediation=True,
                         ),
                         "background_enrichment",
                         memory_id,
--- a/core-api/src/core_api/services/memory_service.py
+++ b/core-api/src/core_api/services/memory_service.py
@@ -1159,7 +1159,10 @@ async def create_memories_bulk(
     # -- Deterministic governance gate (eToro). Runs BEFORE embeddings +
     # content-hash so masked content flows through dedup + storage, and dropped
     # items never get embedded/enriched/written. The LLM free-form signal is
-    # applied post-persist via the enriched-consumer remediation (deferred bulk).
+    # applied post-persist instead: by the enriched-consumer remediation on a
+    # deferred deployment, and by the per-row ``remediate_after_enrichment`` in
+    # the fan-out loop below on an inline one (H-18 — the inline half used to be
+    # missing, so the verdict was computed and discarded).
     governance_errors: dict[int, str] = {}
     gov_pii = tenant_config.governance_pii
     if gov_pii.enabled:
@@ -1682,6 +1685,7 @@ async def create_memories_bulk(
         # already enqueued them (and re-running would double-bill the
         # LLM provider for entity extraction + enrichment).
         from core_api.services.contradiction import Trigger, run_contradiction_detection
+        from core_api.services.governance_remediation import remediate_after_enrichment
 
         # CAURA-595: per-row enrich publishes when deployment_mode is deferred.
         defer_enrich_publish = (
@@ -1704,6 +1708,40 @@ async def create_memories_bulk(
                             mem_data["memory_type"],
                         ),
                         "entity_extraction",
+                        mem_id,
+                        data.tenant_id,
+                    )
+                )
+            if enrichments[orig_idx] is not None:
+                # H-18, bulk/ingest half. This path enriches SYNCHRONOUSLY
+                # above and persists ``contains_pii`` / ``business_relevance``
+                # into the row's metadata, then had nothing act on them — the
+                # same defect as the single-write inline path, on a second
+                # entry point. The deterministic gate near the top of this
+                # function already rejected pattern-detectable PII pre-write;
+                # what was discarded here is the LLM's free-form judgement.
+                #
+                # A non-None enrichment is exactly "an LLM verdict exists for
+                # this row", which can only happen under
+                # ``settings.inline_enrichment`` — so it is also mutually
+                # exclusive with ``defer_enrich_publish`` below, where the
+                # worker's consumer owns remediation instead. Bulk never runs
+                # ``GovernanceDecision`` (it does not go through the write
+                # pipeline), so there is no strong-mode double-apply to guard
+                # against here as there is in ``_schedule_enrich_or_inline``.
+                track_task(
+                    tracked_task(
+                        remediate_after_enrichment(
+                            {
+                                "id": mem_id,
+                                "content": items[orig_idx].content,
+                                "tenant_id": data.tenant_id,
+                                "agent_id": data.agent_id,
+                                "metadata_": mem_data["metadata_"],
+                            },
+                            tenant_config,
+                        ),
+                        "governance_remediation",
                         mem_id,
                         data.tenant_id,
                     )
@@ -1896,6 +1934,7 @@ async def _schedule_enrich_or_inline(
     *,
     agent_provided_fields: list[str] | None = None,
     reference_datetime: datetime | None = None,
+    run_governance_remediation: bool = False,
 ) -> None:
     """Enrichment counterpart of :func:`_schedule_embed_or_reembed`.
 
@@ -1933,7 +1972,7 @@ async def _schedule_enrich_or_inline(
         # ``reference_datetime`` is still not forwarded — the inline path
         # resolves relative dates against the enrichment call's own clock, and
         # that is a separate concern.
-        await _enrich_memory_background(
+        enriched_row = await _enrich_memory_background(
             memory_id,
             content,
             tenant_id,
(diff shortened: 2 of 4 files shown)


Eldad-Caura:
## Both findings verified and fixed — with a deeper fix than either suggested, plus a third defect the review surfaced indirectly

Pushed as `b98c77bc` (amend + force-with-lease).

### The two findings were real, and one fix closes both

I verified both premises against the source before touching anything:

- **M1 (unconditional GET)** — confirmed. `remediate_after_enrichment` early-returns at `governance_remediation.py:44-46` when both features are disabled, and `resolve_config` is cache-first with a 5-min TTL (`organization_settings.py:1160-1167`). The GET was genuinely wasted on the default path.
- **M2 (stale replica)** — confirmed. `get_memory` → `_get(..., read=True)` by default (`storage_client.py:332`), which routes to `_read_prefix`/`_read_http` when `CORE_STORAGE_READ_URL` is set (`storage_client.py:124-136`).

Rather than reorder the block and add a `read=False` flag, I took the alternative the M2 prompt raised at the end — **make the row available instead of re-fetching it**. `_enrich_memory_background` now returns the verdict-bearing view, assembled from its own frame:

```python
governed_row = {
    "id": str(memory_id), "content": content, "tenant_id": tenant_id,
    "agent_id": agent_id, "metadata_": meta,
}
```

Every field is either an argument to that function or `meta`, which it just built and PATCHed. That kills **both** findings at once and is strictly stronger than the suggested fixes:

- **zero** storage round-trips, not "one, skipped when governance is off" — so governance-enabled tenants stop paying it too;
- no replica read **exists**, so there is no staleness window to reason about rather than a re-routed read.

It also survives a fan-out failure: `governed_row` is assigned right after the PATCH and returned even if the atomic-fact loop later raises, so an unrelated failure can't decide whether the PII policy runs.

### Third defect: `business_relevance` was never persisted on this path

Checking whether the fix actually closed H-18 end to end, it did not. `remediate_after_enrichment:102` keys the non-business branch on `md.get("business_relevance") == "personal"` — but `_enrich_memory_background` wrote `contains_pii`/`pii_types` into metadata and **never wrote `business_relevance`**. The synchronous path (`memory_service.py:1471`) and core-worker (`_ENRICHMENT_METADATA_FIELDS`) both persist it; the inline path didn't.

So non-business **DROP** and **KEEP_PRIVATE** — two of the three behaviours H-18 names — were dead on inline deployments no matter how a tenant configured them, and the two deployment modes disagreed about what an enriched row contains. Fixed alongside.

### Also applied, from a quality pass

- **`resolve_config(tenant_id)` was redundant** — `_schedule_enrich_or_inline` already receives that exact `ResolvedConfig` as its `tenant_config` parameter (both call sites pass `ctx.tenant_config`). Dropped the re-resolve, the import and the comment defending it. Worth noting the TTL cache expires by *insertion*, so the "guaranteed cache hit" I'd have been relying on wasn't guaranteed across the enrichment LLM call.
- **Removed the local `try/except Exception`.** It duplicated `tracked_task`, which already wraps this coroutine and both logs *and* persists a `BackgroundTaskLog` row — swallowing locally suppressed that. Its stated rationale was also wrong: extraction and Path A are scheduled as their own `track_task` calls, not chained behind this, so nothing could cascade. Now unguarded, matching `consumer.py`.
- Trimmed ~30 lines of comment that restated the same rationale in three places.

### Tests: 11, and a real before/after

The previous round's tests were weak — an opt-in flag has no "before" state, so they only failed with `TypeError`. These drive the **real** `_enrich_memory_background` and the **real** `remediate_after_enrichment` against a fake storage client and a real `ResolvedConfig` (not a `SimpleNamespace`, so the accessors' own defaulting is exercised).

**7 of the 11 fail against `dcc8dfd3`**, the previous revision of this branch — the personal-memory drop, the keep_private downgrade, the `business_relevance` persistence, the no-extra-read property, the stale-replica case, the PII drop, and the failure-propagation contract. That is the before/after the last round was missing.

Full suite: **4472 passed**, 3 skipped, 1 xfailed. All CI ruff scopes clean.

### Known gap, deliberately not fixed here — #808

Atomic-fact child rows are created inside `_enrich_memory_background` *before* remediation runs and carry no governance metadata, so a PII parent that policy drops leaves children holding the same content. Inline-only — core-worker doesn't fan out. Closing it means moving remediation inside `_enrich_memory_background` ahead of the fan-out, which reorders that function and deserves its own tests. Filed as #808 rather than expanded into a security fix.

Eldad-Caura:
@claude — both Medium findings fixed, plus a third defect they led me to (`business_relevance` was never persisted on the inline path, leaving non-business drop/keep_private dead). Please re-review `b98c77bc`.

The two findings are closed by removing the re-read rather than reordering/re-routing it: `_enrich_memory_background` now returns the verdict-bearing row assembled from its own frame, so there is no second storage GET and no replica read at all. Worth a careful look at that return-value contract and at the removal of the local `try/except` in favour of the enclosing `tracked_task`.

Eldad-Caura:
## High finding confirmed and fixed — the bulk/ingest path had the same gap

Pushed as `e7adf9c8`. You were right, and I'd missed it: `remediate_after_enrichment` appeared in exactly two places in core-api — `consumer.py` and my new single-write call — so every inline bulk write discarded the verdict.

Verified before fixing:

- `create_memories_bulk` enriches synchronously under `settings.inline_enrichment` and persists `contains_pii`/`business_relevance` into each row's metadata.
- Only the `defer_enrich_publish` branch published anything for the worker's consumer to remediate. The inline branch had no counterpart.
- The path's own comment documented the gap without naming it — it said the LLM signal is "applied post-persist via the enriched-consumer remediation **(deferred bulk)**". Corrected.

**Why it looked governed:** the deterministic gate at the top of that function does run and does reject pattern-detectable PII pre-write. What was discarded is the LLM's free-form judgement — the same distinction as on the single-write path.

### The fix

Remediation is scheduled per row in the fan-out loop, alongside the existing extraction/contradiction tasks, gated on `enrichments[orig_idx] is not None`. That condition is precisely "an LLM verdict exists for this row", which can only hold under inline enrichment — so it is also mutually exclusive with `defer_enrich_publish`, where the worker owns remediation.

One difference from the single-write path worth flagging: bulk does **not** go through the write pipeline, so it never runs `GovernanceDecision`. There is no strong-mode double-apply to guard against, and therefore no `run_governance_remediation`-style flag here — the verdict's existence is the whole condition.

### Tests

Four new, against the **real database**, asserting the row's actual `deleted_at` rather than that a mock was called:

- PII drop soft-deletes a bulk-created row
- non-business drop does too (which also needs `business_relevance` to have been persisted)
- **clean-verdict control** — so neither drop test can pass for an unrelated reason
- **deterministic-gate control** — pattern-detectable PII is still refused outright, with no row to remediate

Both drop tests fail without this change; the two controls pass either way, which is what makes them controls. Content in the drop tests is deliberately benign so it clears the deterministic scan and the drop is attributable only to the LLM verdict.

Full suite: **4476 passed**, 3 skipped, 1 xfailed. All CI ruff scopes clean.

Eldad-Caura:
@claude — High finding fixed in `e7adf9c8`: the bulk/ingest inline path now applies the governance verdict too. Please re-review.

Worth focusing on the gate condition `enrichments[orig_idx] is not None` in the fan-out loop — my claim is that it is exactly equivalent to "inline-enriched", and therefore mutually exclusive with `defer_enrich_publish`. If that equivalence does not hold on some path, the fix either misses rows or double-remediates ones the worker will also handle.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #809, opened 2026-08-18
digest: treat reserved agent ids consistently (drop __health_check__; stop mislabeling main-<hex> as 'main')

ran-taig:
## What

Fixes two agent-attribution bugs in the agent-activity digest, both found while investigating an eToro report where **`__health_check__` showed as the most-active agent** and a **`main`** row appeared. Root cause is shared: the leaf noise-filter (`passes_noise_filter`) and the family rollup (`resolve_parent`) didn't agree on what a *reserved* agent id is.

## Bug 1 — a health-check probe topped the leaderboard

`__health_check__` is a write-path smoke test that persists ~200 `memclaw-smoke-<epoch>` **`fact`** rows/day (2,657 on etoro0; recalled 2 times ever). It passed every gate:
- type `fact` → survives the `episode` durable filter
- titles `memclaw-smoke-*` → don't match the heartbeat/health title regex
- id `__health_check__` → wasn't in `RESERVED_FIREHOSE_AGENTS`

**Fix:** add it (and the `__name__` reserved-system convention, via a new shared `is_reserved_agent`) to the leaf filter; add `memclaw-smoke` to the title regex as belt-and-suspenders.

## Bug 2 — a real agent mislabeled as the retired `main` firehose

The `main` row was **not** the firehose (bare `main` has had zero writes since `RESERVED_AGENT_ID_POLICY=reject`). It was a rollup artifact: a real unconfigured-identity agent **`main-e5366d79a926`** (the default `main-<hex>` id, fleet `yoniclaw-fleet`, 51 cohesive writes of genuine SEO/ASO work) whose parent `resolve_parent` walked to `main` — because bare `main` is still a known id and the child starts with `main-`. The firehose *leaf* exclusion was bypassed by the *rollup*.

**Fix:** `resolve_parent` skips reserved ids as prefix candidates and rejects a reserved parent from the structured `agent:main:subagent:` form too. The agent now stays under its **own** id and is shown honestly (per decision: show `main-<hex>` agents under their own id rather than suppress them — they're attributable and did real work).

## Design

`is_reserved_agent()` is the single predicate both surfaces use, so leaf-filtering and rollup finally agree. `RESERVED_FIREHOSE_AGENTS` also feeds `routes/reports.py`'s server-side `exclude_agent_ids`, so the **live Fleet Status dashboard** now drops the probe's 2.6k `fact`s from its durable corpus too. No change for legitimate families (`cmoclaw-*`, etc.).

## Testing

- `tests/test_report_corpus.py` — 9 pass (added: `is_reserved_agent` predicate; probe dropped as noise; reserved id never a rollup parent incl. structured form; legit rollup unaffected).
- `tests/test_agent_digest.py` — 21 pass (no regression).
- `ruff check` clean.

Grounded on read-only prod queries against `oc-memclaw-prod` (etoro0), incl. the actual generated `agent_activity_digests` rows.

### DIFF
--- a/core-api/src/core_api/services/report_corpus.py
+++ b/core-api/src/core_api/services/report_corpus.py
@@ -16,10 +16,22 @@ import re
 # generator so "day"/"week" mean the same thing on both surfaces.
 PERIOD_DAYS: dict[str, int] = {"day": 1, "week": 7}
 
-# Episodic activity-log type(s) + the unattributed firehose agent excluded so the
-# report reflects durable, decision-bearing per-agent work.
+# Episodic activity-log type(s) excluded so the report reflects durable,
+# decision-bearing per-agent work.
 NON_DURABLE_TYPES = ("episode",)
-RESERVED_FIREHOSE_AGENTS = ("main",)
+# Reserved / automated write-path identities that are never a real agent: the
+# legacy unattributed ``main`` firehose, and ``__health_check__`` (a monitoring
+# probe that writes ~200 ``memclaw-smoke-*`` facts/day). These must appear in a
+# report neither as a ROW nor as a rollup PARENT (see :func:`resolve_parent`) —
+# otherwise a real agent's work gets mislabeled under a reserved id. Note
+# ``routes/reports.py`` also passes this tuple to storage as a server-side
+# ``exclude_agent_ids`` list, so its entries must stay exact agent ids.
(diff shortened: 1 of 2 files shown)


ran-taig:
@claude

ran-taig:
@erni-a

erni-a (approved):

PR #821, opened 2026-08-18
fix(contradiction): abstain instead of guessing when no LLM answered

Eldad-Caura:
Found by sweeping every `fake_fn=` site for the same shape as #807 — a dev stub
standing in on a production failure path. There are 15; most are fine. **This is
the severe one.**

## The chain

1. The no-LLM fallback was `_fake_contradiction_check` — a **negation-word
   heuristic** whose own docstring said *"Simple heuristic for testing"*.
2. Its verdict is not advisory. `detect_contradictions` acts on the **boolean
   alone**: confidence is unpacked as `_confidence` and only logged, because A4
   #13's confidence-weighted veto is still deferred. The one live threshold,
   `RETRACTION_CONFIDENCE_THRESHOLD`, gates retraction, not marking.
3. So it sets the older memory's `status = "conflicted"`, which **demotes it out of
   semantic recall**.
4. Two of the four fallback sites also hardcoded `same_subject: True` for *every*
   candidate — directly against the batch prompt's own rule: *"Set true ONLY when
   subject_a and subject_b refer to the SAME real-world entity"* and *"If
   same_subject is false, contradicts MUST be false."*

A heuristic labelled "for testing" could degrade recall in a product whose job is
recall.

## The fix: split what `call_with_fallback` conflates

`call_with_fallback` invokes `fake_fn` for two different situations, and they are
not the same — which is the same one-thing-two-purposes shape as #305, #306 and
#807.

- **Explicitly configured `fake` provider** → keeps the heuristic. It is the only
  end-to-end coverage of the detect → mark pipeline without an API key, and
  `test_p1_contradiction.py::test_semantic_contradiction_with_fake_llm` depends on
  it. A blanket abstain would have silently deleted that coverage.
- **A configured REAL provider that failed** → abstains.

`fake` only, **not `none`**: "none" asks for the feature to be off. That is not my
invention — `entity_extraction.py` already draws the same line, returning an empty
graph for `none` and calling its heuristic for `fake`.

The abstain reuses the module's own documented encoding rather than a new one:
`(False, _CONF_FALLBACK)` is exactly what `_judge_contradiction` returns for a
malformed response, and the rubric already calls 0.50 "conservative-default".

The batch path returns an **empty dict** per candidate, not `_align`'s
`{"contradicts": False}`. That shape is non-empty, so it misses the malformed
branch and scores `_CONF_CLEAN` — verified: `0.90` vs `{}`'s `0.50`. An abstain
presenting as a confident clean verdict is inert today and stops being inert the
moment A4 #13 gates on confidence.

## Two claims I had wrong, and corrected

My first draft of these docstrings asserted two things that the code contradicts.
Flagging them because the PR is *about* unverified claims:

| I wrote | Actually |
|---|---|
| "`memory_scored_search` excludes `conflicted` rows" | There's an `or_` **carve-out** ([postgres_service.py:1653](core-storage-api/src/core_storage_api/services/postgres_service.py:1653)) retaining a conflicted row that's an exact lexical match, un-demoted. The damage is to semantic/vector recall, not lookup-by-phrasing. |
| "Nothing ever re-evaluates that status" | Path C's retraction judge **does** restore `conflicted` → `active`. Narrowly: canonical direction only, resolved entities both sides, ≥0.90 confidence, tenant kill-switchable. No *sweep* exists. |

The cost is now stated precisely, in one place, with the tests cross-referencing it
rather than restating it — so the two can't drift.

## Verification — by mutation, not just by running the tests

| mutation | result |
|---|---|
| `_deliberate_fake_provider` → always `True` (the original bug) | **8 failed** |
| batch abstain → `{"contradicts": False}` (the 0.90 shape) | **3 failed** |
| entity-aware pairwise site → back to the heuristic | **1 failed** |

That last one matters: the two **entity-aware** sites had *no* abstain coverage —
every existing entity-aware test patches `call_with_fallback` or the batch function
outright, so `fake_fn` was never reached. Either could have been reverted with the
whole suite green. They're also the sites Path C uses.

Batch assertions pin what the **consumer** derives (`_judge_contradiction(entry) ==
(False, _CONF_FALLBACK)`) rather than the raw dict shape — that's where the
0.90-vs-0.50 distinction actually lives.

`test_the_heuristic_really_would_have_flagged_this` is a deliberate vacuity guard:
it proves the test inputs are heuristic-positive, so the abstain assertions can't
pass on bland input. It passes either way by design.

## Two existing tests changed, and why

- `test_fake_provider_triggers_heuristic_via_fallback` → renamed
  `test_missing_credentials_abstains_instead_of_guessing`. It set provider `openai`
  with **no API key** — the production-outage path — and asserted `verdict is
  True`. It pinned the defect.
- Added `test_deliberate_fake_provider_still_exercises_the_heuristic`: same inputs,
  opposite expectation. That pair is the whole point of the split.

- root suite: **4484 passed**, 3 skipped, 1 xfailed
- `ruff check core-api/src/` + `format --check` + `ruff check common/` clean;
  `mypy` clean (198 files)

## Not in this PR

The four moderate sweep findings — `_crystallize_fake` (writes a copy of the
highest-weight memory as a "crystallized insight"), `_fake_insights` (persists a
finding headlined *"Fake insight for testing"*), `_fake_rule` (a generic rule at
confidence 0.6), and `_fake_recall` (truncated memory text as a summary) — all want
a provenance marker following `fake_enrich`'s pattern, where `llm_ms=0` is actually
consumed downstream. One follow-up PR.

### DIFF
--- a/core-api/src/core_api/services/contradiction_detector.py
+++ b/core-api/src/core_api/services/contradiction_detector.py
@@ -7,7 +7,9 @@ P1 fixes:
 
 Multi-provider support:
 - Vertex AI, OpenAI, Anthropic, OpenRouter via provider layer
-- Automatic fallback chain: configured provider -> fallback -> heuristic
+- Automatic fallback chain: configured provider -> fallback -> abstain
+  (heuristic only when the ``fake`` provider was explicitly asked for; see
+  ``_deliberate_fake_provider``)
 """
 
 import asyncio
@@ -17,6 +19,7 @@ import uuid as _uuid
 from datetime import datetime
 from uuid import UUID
 
+from common.provider_names import ProviderName
 from core_api.cache import cache_set_nx
 from core_api.clients.storage_client import get_storage_client
 from core_api.config import settings
@@ -1006,7 +1009,8 @@ async def _llm_contradiction_check(
     Uses the standard 3-tier fallback chain:
     1. Try the configured provider (with retry)
     2. Try the configured fallback provider (via resolve_fallback)
-    3. Fall back to negation-word heuristic
+    3. Abstain — or, when the operator ASKED for the fake provider, the
+       negation-word heuristic. See :func:`_deliberate_fake_provider`.
     """
     provider_name = (
         tenant_config.entity_extraction_provider if tenant_config else settings.entity_extraction_provider
@@ -1021,7 +1025,7 @@ async def _llm_contradiction_check(
     return await call_with_fallback(
         primary_provider_name=provider_name,
         call_fn=_do_check,
-        fake_fn=lambda: (_fake_contradiction_check(new_content, old_content), _CONF_FALLBACK),
+        fake_fn=_pairwise_fake_fn(provider_name, new_content, old_content),
         tenant_config=tenant_config,
         service_label="contradiction",
         model_attr="entity_extraction_model",
@@ -1123,13 +1127,7 @@ async def _llm_contradiction_check_batch(
     return await call_with_fallback(
         primary_provider_name=provider_name,
         call_fn=_do_check,
-        fake_fn=lambda: [
-            {
-                "same_subject": True,
-                "contradicts": _fake_contradiction_check(new_content, c.get("content", "")),
-            }
-            for c in candidates
-        ],
+        fake_fn=_batch_fake_fn(provider_name, new_content, candidates),
         tenant_config=tenant_config,
         service_label="contradiction_batch",
         model_attr="entity_extraction_model",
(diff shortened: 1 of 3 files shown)


Eldad-Caura:

This is the severe finding from sweeping all 15 `fake_fn=` sites for the #807 shape. The no-LLM contradiction fallback was a negation-word heuristic labelled "for testing" in its own docstring, and its verdict is acted on: `detect_contradictions` marks the older memory `conflicted` on the boolean alone, which demotes it out of semantic recall. Two sites also hardcoded `same_subject: True` for every candidate, against the batch prompt's own rule.

The fix splits what `call_with_fallback` conflates — a deliberately-requested `fake` provider keeps the heuristic (it's the only end-to-end coverage of detect→mark without an API key), a real provider that failed abstains. Same one-field-two-purposes shape as #305/#306/#807.

Three things worth your eye in review:

1. **I had two claims wrong and corrected them in-flight.** I'd written that `memory_scored_search` excludes conflicted rows and that nothing ever restores the status. Neither holds: there's an exact-lexical-match carve-out that keeps a conflicted row un-demoted, and Path C's retraction judge does restore `conflicted`→`active` (narrowly). The PR body has the details. Given this PR is *about* unverified claims, I'd rather flag that than bury it.
2. **The two entity-aware sites had zero abstain coverage** — every existing entity-aware test patches `call_with_fallback` or the batch fn outright, so `fake_fn` was never reached. Either could have been reverted with the suite green. Now mutation-verified.
3. **`test_fake_provider_triggers_heuristic_via_fallback` was pinning the defect** — provider `openai`, no API key, asserting `verdict is True`. Renamed and inverted, with a companion test asserting the deliberate-`fake` path still works. That pair is the split.

The four moderate sweep findings (`_crystallize_fake`, `_fake_insights`, `_fake_rule`, `_fake_recall`) want provenance markers following `fake_enrich`'s pattern — one follow-up PR, not this one.

erni-a (approved):

PR #822, opened 2026-08-18
fix(fallbacks): a no-LLM stand-in must not retire the data it replaces

Eldad-Caura:
Two more `fake_fn` sites from the sweep that produced #821. **I classified these as
"moderate — wants a provenance marker" and I was wrong.** Reading their callers
rather than the fallbacks themselves, both destroy real data.

## Why these are not cosmetic

Both callers **retire the previous data as part of persisting**, so a manufactured
result doesn't just add noise — it replaces content that then isn't recoverable.

**`crystallizer_service`** creates one memory per returned fact and then **archives
every source memory in the cluster**. `_crystallize_fake` returns a *verbatim copy*
of the cluster's highest-weight memory. So an LLM outage archived N memories and
left one duplicate of one of them behind, having synthesised nothing — the other
N−1 memories' content is not in the survivor, and `archived` sits outside
`LIVE_MEMORY_STATUSES`.

**`insights_service`** transitions the tenant's prior insights to `outdated`
*before* persisting, guarded by `if findings:`. `_fake_insights` returns one
finding, so an outage retired real insights in favour of one headlined
*"Fake insight for testing"*.

## The fix is to take the path that already existed

Neither caller needed new logic. Both already had a correct "nothing came back"
branch:

- `if not extracted: continue` — before anything is created or archived.
- the empty-findings short-circuit, whose own comment says outdating priors when
  there is nothing to persist "would leave the user with neither".

So the outage path now returns empty and lets those fire.

Applies #821's split: an explicitly configured `fake` provider **keeps** its
stand-in, because that is the intent and it is how the create+archive and
supersede+persist paths get exercised without an API key. A configured real
provider that failed abstains.

## `deliberate_fake_provider` is now shared

It moves from `contradiction_detector` to `common/llm/retry.py`, beside the
`call_with_fallback` whose two invocation contexts it exists to distinguish, and
re-exported through the `providers/_retry` shim. Three callers need it now; leaving
it private would have meant three copies of the FAKE-vs-NONE policy, which is the
same mistake as the six `SortHeader`s in caura-ops#308.

## Verification

Tests assert the **destructive consequence**, not the return value, and run through
the real entrypoints (`_crystallize_cluster`, `_run_llm_analysis`) so they'd still
catch a caller that stopped honouring the empty case:

```
FAILED tests/test_no_llm_fallbacks_preserve_data.py::test_crystallizer_returns_nothing_on_an_outage
FAILED tests/test_no_llm_fallbacks_preserve_data.py::test_insights_returns_no_findings_on_an_outage
2 failed, 4 passed
```

Each has a **control** proving the stand-in is genuinely non-empty (so the abstain
assertions can't pass on bland input), plus a **counterpart** proving the deliberate
`fake` path still returns it. Those four pass either way by design — they're pins,
not gates.

- root suite: **4490 passed**, 3 skipped, 1 xfailed
- `ruff check core-api/src/ common/`, `ruff format --check core-api/src/`, `mypy` —
  all verified by exit code (my first pass reported clean off a piped `tail`'s exit
  status and hid a formatting failure)

## Still open from the sweep

- `_fake_rule` — persists a fabricated generic rule at confidence 0.6 as a
  `memory_type="rule"` memory. Same treatment; separate PR since it writes rather
  than retires.
- `_fake_recall` — returns truncated memory text as a recall *summary*. Read path,
  never persisted, so it wants an honest marker in the text like `_fake_report`'s
  "(LLM unavailable; unsynthesized)", not abstention.
- `documents.py:161-162` defaults missing timestamps to `datetime.min` — year 1,
  which readers sort and display.

### DIFF
--- a/common/llm/retry.py
+++ b/common/llm/retry.py
@@ -117,6 +117,31 @@ async def call_with_retry(
     raise last_exc  # type: ignore[misc]
 
 
+def deliberate_fake_provider(provider_name: str | None) -> bool:
+    """Which of ``call_with_fallback``'s two ``fake_fn`` contexts is this?
+
+    ``fake_fn`` is invoked for two situations that are NOT the same, and callers
+    that conflate them ship a dev stub's output to production:
+
+    * The operator explicitly configured ``fake`` — dev or CI. The stub IS the
+      intent; it is often the only way a pipeline gets end-to-end coverage without
+      an API key.
+    * A REAL provider was configured and every attempt failed, or its key is
+      missing so it resolved to ``FakeLLMProvider``. That is an outage or a
+      misconfigured deployment, not a request for made-up output.
+
+    ``fake`` ONLY, deliberately not ``none``: "none" asks for the feature to be
+    off, which is an abstain. ``entity_extraction`` draws the same line — empty
+    graph for ``none``, heuristic for ``fake``.
+
+    Callers that persist their result should branch on this and take their own
+    "nothing came back" path in the outage case. See ``contradiction_detector``
+    (abstains rather than guessing a verdict that marks memories ``conflicted``),
+    ``crystallizer_service`` and ``insights_service``.
+    """
+    return provider_name == ProviderName.FAKE
+
+
 async def call_with_fallback(
     primary_provider_name: str,
     call_fn: Callable[..., Coroutine[Any, Any, T]],
--- a/core-api/src/core_api/providers/_retry.py
+++ b/core-api/src/core_api/providers/_retry.py
@@ -7,6 +7,6 @@ sites (``core_api.services.*``, tests) working unchanged.
 
 from __future__ import annotations
 
-from common.llm.retry import call_with_fallback, call_with_retry
+from common.llm.retry import call_with_fallback, call_with_retry, deliberate_fake_provider
 
-__all__ = ["call_with_fallback", "call_with_retry"]
+__all__ = ["call_with_fallback", "call_with_retry", "deliberate_fake_provider"]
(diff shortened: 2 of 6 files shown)


Eldad-Caura:
@erni-a ready for review — all checks green. Claude's one Low finding (three stale docstring references to the pre-move `_deliberate_fake_provider` name) is fixed and minimized. Review jobs show skipped above because that re-run was a force-push; the amend was docstrings only.

**I had these two classified as "moderate — wants a provenance marker" and that was wrong.** Reading the callers rather than the fallbacks: both retire the previous data as part of persisting, so a stand-in result destroys content.

- `crystallizer_service` creates a memory per returned fact and then **archives every source memory in the cluster**. `_crystallize_fake` returns a verbatim copy of the highest-weight memory — so an outage archived N memories and left one duplicate, having synthesised nothing.
- `insights_service` transitions prior insights to `outdated` **before** persisting, guarded by `if findings:`. So an outage retired real insights in favour of one headlined "Fake insight for testing".

Neither needed new logic — both already had a correct empty-result branch (`if not extracted: continue`, and the empty-findings short-circuit whose own comment says outdating priors with nothing to persist "would leave the user with neither"). The fix is to take them.

`deliberate_fake_provider` moves to `common/llm/retry.py` beside the `call_with_fallback` whose two invocation contexts it distinguishes. Three callers need it now; leaving it private would have meant three copies of the FAKE-vs-NONE policy — the same mistake as the six `SortHeader`s in caura-ops#308.

Tests assert the destructive consequence rather than the return value, through the real entrypoints, each with a control proving the stand-in is genuinely non-empty so the abstain assertions can't pass vacuously.

erni-a (approved):

PR #824, opened 2026-08-18
feat(usage): route usage metering through a service hook

Eldad-Caura:
OSS half of caura-ai/caura-enterprise#83 ("usage counters never increment"). **On its own this changes no behaviour** — it creates the seam the platform side needs. Counters do not start incrementing until the follow-up lands.

## What was actually wrong

`usage_service` was three no-op stubs with no way for a platform deployment to substitute an implementation, so the write paths and the platform's per-period counters were two subsystems that never met. Measured on a stack that had served traffic for a day:

```
 writes | searches | recalls | count
      5 |        3 |       0 |    25     ← every row identical, 25 distinct orgs
```

…while the same database held **3,180** memories. `recalls = 0` everywhere is the cleanest tell.

They now delegate to `ServiceHooks.usage_meter`. `ServiceHooks` is the existing platform→core injection point (`audit_log` already uses it, wired at `app.py:272`) and is the **only** one — the providers registry is a closed name→class factory and there are no entry points — so this completes a pattern rather than inventing one. With no hook wired, behaviour is exactly as before: allowed, unlimited, nothing recorded. OSS standalone must run without a billing plane.

## Two things found while doing it

**`check_and_increment`'s first parameter was named `org_id`, and no caller has ever passed one.** All ~21 call sites pass a tenant id, including the two importing that name specifically (`routes/stm.py:216`, `routes/memories.py:870`). Harmless while the body ignored the value — a meter trusting the name would have keyed those writes on the wrong entity. `org_id` is a real, distinct field on `AuthContext`, so this was not a harmless synonym.

**`check_and_increment` and `check_and_increment_by_tenant` were identical** once the parameter was corrected. Now an alias assignment rather than a duplicated body: the long name appears on five import lines, all aliasing it to the short one, so no invocation anywhere spells it.

## Why the result is not an enforcement decision

Nothing in core-api reads `allowed`, and that is **by design**. Enforcement arrives out-of-band: the platform computes "over plan" from the persisted counters and stamps `x-org-read-only`, which `AuthContext.is_read_only` reads and `enforce_usage_limits()` turns into a 403 at ~22 write routes. The module docstring says so, so a platform implementer does not build enforcement into a return value that is discarded.

That is also what makes **fail-open** correct on hook failure — this is on the write path of every metered route, and a metering backend that is down must cost a count, not a customer write. The comment names the condition that would invalidate it: the moment `allowed` gains a reader, fail-open becomes "meter down ⇒ every tenant unlimited".

Failure logging is **throttled** (first, then every 100th) rather than one traceback per write — same shape as `audit_queue`'s drop counter, so a meter outage isn't also a log-volume incident.

Also documented why this is a second counter and not a change to `capability_usage`: that path buffers in memory, loses counts on process death by its own docstring, and appends rows to be SUMmed rather than upserting a period total. Right for adoption analytics, unfit for a plan cap.

## Tests — 8, with an honest note

Most exercise a **new capability** and fail against the old code with `TypeError: unexpected keyword argument 'usage_meter'`. That shows the wiring is new, not that old behaviour was broken — the old behaviour (no-op) was correct for OSS, and the defect was the *absence of a seam*, which has no "before" to regress against. I'd rather say that than claim a before/after I don't have.

The one test that must pass identically either way, and does, is the OSS-standalone pin. The throttle test does have a real before/after: without it, 30 failures produce 30 log records instead of 1.

Full suite: **4492 passed**, 3 skipped, 1 xfailed. `ruff check` + `ruff format --check` clean across every CI scope.

## Next

The platform-side wiring (an enterprise implementation registered at startup) is the follow-up that makes counters actually move.

### DIFF
--- a/core-api/src/core_api/services/hooks.py
+++ b/core-api/src/core_api/services/hooks.py
@@ -15,12 +15,25 @@ from __future__ import annotations
 import logging
 from collections.abc import Awaitable, Callable
 from dataclasses import dataclass
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from core_api.services.usage_service import UsageCheckResult
 
 logger = logging.getLogger(__name__)
 
 # Type alias for the audit hook signature.
 AuditHook = Callable[..., Awaitable[None]]
 # Signature: (*, tenant_id, agent_id, action, resource_type, resource_id, detail) -> None
+
+# Type alias for the usage-metering hook signature.
+# Signature: (*, tenant_id, operation, count) -> UsageCheckResult | None
+#
+# ``type`` (PEP 695) rather than a plain assignment: it is evaluated lazily, so
+# the ``usage_service`` import stays under TYPE_CHECKING and the cycle never
+# forms. Worth it over ``Any`` — spelling the return type is what makes a
+# platform-supplied implementation checkable at all.
+type UsageHook = Callable[..., Awaitable[UsageCheckResult | None]]
 
 
 @dataclass
@@ -32,6 +45,11 @@ class ServiceHooks:
     """
 
     audit_log: AuditHook | None = None
+    # METERING, not enforcement — the name is deliberate. See the
+    # ``usage_service`` module docstring for why, and for the implementation
+    # guidance (enqueue and return ``None``; report counters only if that costs
+    # no round-trip — this hook runs on the write path).
+    usage_meter: UsageHook | None = None
 
 
 _hooks = ServiceHooks()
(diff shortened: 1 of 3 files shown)


Eldad-Caura:
@erni-a — ready for approval. Review verdict clean (`✅ No issues found`); CI, CodeQL, security scan and DCO all green.

Worth knowing before merging: **this changes no behaviour on its own.** With no hook wired it is byte-identical to today — allowed, unlimited, nothing recorded — which is what OSS standalone must keep doing. It creates the seam; counters only start moving once the platform-side implementation is registered at startup, which is the follow-up.

The one thing in here that isn't purely additive is the parameter rename: `check_and_increment`'s first argument was named `org_id` and every one of the ~21 call sites passes a tenant id. All calls are positional so nothing breaks, but it is worth a `grep 'check_and_increment(org_id='` in the enterprise repo before merge — I couldn't check that from here.

erni-a (approved):

Eldad-Caura:
Closing my own caveat: I checked the enterprise repo and `check_and_increment` appears **nowhere** in it — no positional calls, no keyword calls, no imports. The rename cannot break a cross-repo caller because there are none. Nothing to verify before merge.

PR #825, opened 2026-08-18
fix(evolve): a disabled or failed provider must not fabricate a rule

Eldad-Caura:
Closes the sweep behind #821, #822 and #823. This was the site I backed out of #823
because it turned on a decision — **Eldad's call: `none` means abstain.**

## The problem

`_fake_rule` returns a generic *"verify information against the most recent source"*
rule at confidence 0.6, and the caller persists it as a `memory_type="rule"` memory.

`rule` is in `SERVER_RESERVED_MEMORY_TYPES`: the enrichment classifier may not emit
it, and a hallucinated one is demoted to the default type. So the reserved types
carry the implicit provenance *"an internal flow authored this deliberately"* rather
than *"a model guessed"*. A fabricated rule borrows that authority — plausible
enough to survive review, attributable to nobody.

## The fix

Abstaining returns `None`, which fails the caller's `isinstance(raw, dict)` check and
yields `("llm_failed", None)`, so no rule is written. That path already existed for a
provider that raised; this routes the no-provider case to it.

`none` is not a request for a stub — only an explicit `fake` provider is, matching
how `entity_extraction` already splits them and what #821 established.

## The test change, and a correction to what I told you

`ResolvedConfig.enrichment_provider` falls back to `entity_extraction_provider`, and
the suite runs with that set to `none`. So tests using `resolve_config` were
inheriting `none` as their way of getting a stub.

**Exactly one test depended on it** — `test_evolve_generate_rule_returns_valid_structure`.
It now builds `ResolvedConfig({"enrichment": {"provider": "fake"}})` and asks
explicitly; `_generate_rule` reads only `enrichment_provider` and `enrichment_model`
off the config, so that is sufficient and faithful.

I earlier reported **five** tests. That count came from a tree that also carried the
recall change from #823 and was wrong — on clean `main` it is one. Correcting it
here because it was the whole reason this got deferred.

## Verification

New test pins the decision, and fails behaviourally against unfixed code — it
collects and runs, so this is not an import artefact:

```
FAILED tests/test_evolve_service.py::test_evolve_generates_no_rule_when_provider_is_none
1 failed, 53 passed
```

The updated existing test passes either way by design: with `provider="fake"`, old
and new code both reach `_fake_rule`. It is a pin, not a gate.

- root suite: **4492 passed**, 3 skipped, 1 xfailed
- `ruff check`, `ruff format --check`, `mypy` — all verified by exit code

## One claim I removed

My first draft asserted rules "are recalled and applied as policy rather than read
as facts". Nothing in recall boosts or prioritises `rule` today — the only grep hit
for that idea was my own comment. The reserved-type provenance is the argument that
actually holds, and the docstring now records the negative finding too, so the next
reader doesn't have to re-check it.

That is the third claim in this sweep I had to verify down rather than up. It is
also the pattern the whole sweep was about: a marker or a type that *looks* like it
means something to a reader, where nothing actually reads it.

### DIFF
--- a/core-api/src/core_api/services/evolve_service.py
+++ b/core-api/src/core_api/services/evolve_service.py
@@ -381,8 +381,33 @@ async def _adjust_weights(
 # -- Rule generation ----------------------------------------------------------
 
 
+def _skip_rule() -> None:
+    """No-LLM rule generation: produce nothing.
+
+    ``None`` fails the caller's ``isinstance(raw, dict)`` check, so it returns
+    ``("llm_failed", None)`` and no rule is written. That path already exists for a
+    provider that raised; this routes the no-provider case to it too.
+
+    Why this one matters: the result is persisted as a ``memory_type="rule"``
+    memory, and ``rule`` is in ``SERVER_RESERVED_MEMORY_TYPES`` — the enrichment
+    classifier is forbidden from emitting it, and any hallucinated ``rule`` is
+    demoted to the default type. So the reserved types carry the implicit provenance
+    "an internal flow authored this deliberately", not "a model guessed". A
+    fabricated one borrows that authority: ``_fake_rule`` returns a generic "verify
+    information against the most recent source" rule at confidence 0.6, plausible
+    enough to survive review and attributable to nobody.
+
+    (Nothing in recall boosts or prioritises ``rule`` today — I checked before
+    claiming it. The reserved-type provenance is the actual argument.)
+    """
+    logger.warning("evolve: no LLM — no rule generated")
+    return None
+
+
 def _fake_rule() -> dict:
-    """Placeholder rule for fake/test LLM provider."""
+    """Placeholder rule for an explicitly configured ``fake`` provider. TEST/DEV
+    ONLY — the production fallback abstains via ``_skip_rule``, because this output
+    is persisted as a ``rule``-type memory."""
     return {
         "condition": "When encountering a similar situation",
         "action": "Verify information against the most recent source before acting",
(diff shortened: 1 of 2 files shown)


Eldad-Caura:

Closes the `fake_fn` sweep (#821, #822, #823, this). Per your call, `none` means abstain: it asks for the feature off, not for a stub — only an explicit `fake` provider gets one, matching how `entity_extraction` already splits them.

The site: `_fake_rule` returns a generic rule at confidence 0.6, persisted as a `memory_type="rule"` memory. `rule` is in `SERVER_RESERVED_MEMORY_TYPES`, so the classifier may not emit it — the reserved types carry the provenance "an internal flow authored this deliberately", and a fabricated one borrows that authority. Abstaining returns `None`, which fails the caller's existing `isinstance(raw, dict)` gate, so no rule is written.

**Two corrections I owe you on this one:**

1. I said five tests depended on `none` granting a stub. It's **one** — `test_evolve_generate_rule_returns_valid_structure`. My count came from a tree that also carried #823's recall change. That miscount was the entire reason this got deferred out of #823, so it's worth naming.

2. My first draft docstring claimed rules "are recalled and applied as policy rather than read as facts". **I couldn't substantiate it** — nothing in recall boosts or prioritises `rule`, and the only grep hit for the idea was my own comment. Replaced with the reserved-type argument, which does hold, and I left the negative finding in the docstring so the next reader doesn't re-check it.

That's the third claim in this sweep I had to verify down rather than up — which is the pattern the sweep was about: a marker or type that looks meaningful to a reader where nothing actually reads it.

erni-a (approved):

PR #826, opened 2026-08-18
fix(documents): a NULL timestamp must serialise, not 500 the read

Eldad-Caura:
Last item from the sentinel half of the sweep behind #821/#822/#823/#825. **I logged
it as cosmetic; it is a real bug.**

## Two problems, and the guard only addressed the one that couldn't happen

`documents.created_at` / `updated_at` carry `server_default=now()` but were never
declared `nullable=False` — `001_initial_schema`, and I checked that no later
migration tightened it. So a NULL is representable.

`DocOut` required a `datetime`, and `_dict_to_out` guarded with:

```python
created_at=d.get("created_at", datetime.min),
```

`.get(key, default)` substitutes only when the **key is absent**. For the case the
schema actually permits — **key present, value `None`** — the default never fires,
`None` reaches a non-optional `datetime`, and pydantic raises *inside the read
route*. Verified:

```
DocOut(..., created_at=None) -> REJECTED None -> ValidationError
```

`_dict_to_out` has six call sites, three of them list endpoints, so one such row
could take a whole page down rather than just its own record.

And where the default *did* fire it was wrong on its own terms: `datetime.min` is
year 1 **and naive**, in a timezone-aware field, handed to an API client with no way
to read it as anything but a real timestamp. It also sorts before every genuine row.

## The fix

Make the response model match the data model: `datetime | None`.

That **strictly widens** what the endpoint can return. Rows with NULL timestamps
could not be served at all before, so no client can be relying on non-null for them;
rows with real timestamps are unaffected. A 500 becoming `null` is an improvement for
every caller.

The NULL case is logged, since it means a row written outside the normal insert path
and should be zero in practice. Identifiers only — `id` and `collection`, never
`data`, which is caller-supplied document content.

## Verification

4 of 5 tests fail against unfixed code, behaviourally:

```
FAILED tests/test_document_null_timestamps.py::test_a_null_created_at_serialises_as_null
FAILED tests/test_document_null_timestamps.py::test_both_timestamps_null_serialises
FAILED tests/test_document_null_timestamps.py::test_a_missing_key_is_null_not_year_one
FAILED tests/test_document_null_timestamps.py::test_the_null_case_is_logged_with_identifiers_only
4 failed, 1 passed
```

The 5th is the control — a normal row, unchanged and silent — and passes either way
by design.

One test asserts the log carries the id but **not** the document's `data`, so the
shape-not-payload rule is gated rather than just intended.

- root suite: **4505 passed**, 3 skipped, 1 xfailed
- `ruff check`, `ruff format --check`, `mypy` — verified by exit code

## Follow-up not taken here

The columns arguably *should* be `NOT NULL` — every insert path sets them via
`server_default`. Tightening them is a migration against a live table and a separate
decision; this PR makes the API honest about what the current schema allows.

### DIFF
--- a/core-api/src/core_api/routes/documents.py
+++ b/core-api/src/core_api/routes/documents.py
@@ -143,14 +143,36 @@ class DocOut(BaseModel):
     collection: str
     doc_id: str
     data: dict
-    created_at: datetime
-    updated_at: datetime
+    # NULLABLE because the columns are. ``documents.created_at`` / ``updated_at``
+    # carry ``server_default=now()`` but were never declared ``nullable=False``
+    # (001_initial_schema, never tightened since), so a NULL is representable. This
+    # model previously required a ``datetime``, which meant such a row could not be
+    # serialised at all — the read raised ValidationError and the route 500'd. No
+    # client can be relying on non-null for those rows, because they were never
+    # served; ``null`` strictly widens what this endpoint can return.
+    created_at: datetime | None
+    updated_at: datetime | None
 
 
 # ── Helpers ──
 
 
 def _dict_to_out(d: dict) -> DocOut:
+    # ``.get(key, datetime.min)`` covered only the MISSING-KEY case and left the
+    # present-but-None one, which is the one the schema actually permits — so the
+    # default never fired where it mattered and the route 500'd instead. And where
+    # it did fire it served year 1 (naive, in a tz-aware field) to a caller with no
+    # way to recognise it as a sentinel. Pass the value through and let ``null``
+    # mean unknown.
+    if d.get("created_at") is None or d.get("updated_at") is None:
+        # Identifier only, never ``data`` — that is caller-supplied document content.
+        logger.warning(
+            "document_timestamp_null id=%s collection=%s created_at_null=%s updated_at_null=%s",
+            d.get("id"),
+            d.get("collection"),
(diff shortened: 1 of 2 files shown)


Eldad-Caura:

Last item from the `fake_fn` sweep's sentinel half. **I logged it as cosmetic and it's a real 500.**

`documents.created_at`/`updated_at` carry `server_default=now()` but were never declared `nullable=False`, so NULL is representable. The guard was `d.get(key, datetime.min)` — which only substitutes when the **key is absent**. The case the schema actually permits is **key present, value None**, where the default never fires and `None` hits a non-optional `datetime`: pydantic raises inside the read route. `_dict_to_out` has six call sites, three of them list endpoints, so one such row could take down a whole page rather than just its own record.

Making the response `datetime | None` strictly widens what the endpoint returns — those rows couldn't be served at all before, so nothing can be depending on non-null for them. A 500 becoming `null` is an improvement for every caller.

One test asserts the new warning carries the row id but **not** the document's `data`, so the shape-not-payload rule is gated rather than just intended.

Noted as a follow-up rather than done here: the columns arguably want `NOT NULL`, since every insert path sets them. That's a migration against a live table and a separate call; this PR just makes the API honest about what the schema currently allows.

erni-a (approved):

PR #827, opened 2026-08-18
feat(db): documents.created_at / updated_at NOT NULL

Eldad-Caura:
Follow-up to #826, which made the API honest that these columns could be NULL. This
closes the hole at the source — they describe row lifecycle, not optional user data.

**Migration against a live table.** Merging runs it on prod at service startup, so
the verification section below is the part worth reading.

## Backfill first, because a failed migration stalls the deploy

`SET NOT NULL` fails if the table contains a NULL, and migrations run in the FastAPI
lifespan startup hook — so a failure means the new revision never becomes ready.
Cloud Run keeps the old one serving, which fails safe, but it fails.

The backfill derives from the row where it can: a NULL `created_at` takes
`updated_at` (an upper bound on it), and vice versa. When **both** are NULL there is
nothing to derive from — `id` is `gen_random_uuid()`, so it carries no time — and
`now()` is the only available value. That asserts the row was created at migration
time, which is **false**; it's chosen because the alternative is refusing to migrate.
The migration says so, and says how to identify those rows afterwards (they share a
near-identical timestamp with each other and with the migration run).

## Lock safety, because `documents` is a declared large table

It's in this repo's own `large_tables` set. A bare `SET NOT NULL` full-scans under
AccessExclusive — the shape that crashed 6 storage-writer boots on 2026-06-16. So:

```
ADD CONSTRAINT ... CHECK (col IS NOT NULL) NOT VALID   -- AccessExclusive, O(1)
VALIDATE CONSTRAINT ...                                -- ShareUpdateExclusive:
                                                       -- scans WITHOUT blocking
                                                       -- reads or writes
ALTER COLUMN ... SET NOT NULL                          -- PG12+ skips the scan
DROP CONSTRAINT ...                                    -- now redundant
```

Each step in its own transaction via `autocommit_block` — **that is the whole
point**. Run them inside one transaction and the AccessExclusive from step 1 is held
across the scan in step 2, reproducing exactly the blocking the pattern avoids.

## New guard, and two things mutation caught in it

The existing `test_no_plain_create_index_on_large_tables` covers CREATE INDEX only;
nothing stopped the naive `SET NOT NULL`. Added
`test_no_plain_set_not_null_on_large_tables`, catching raw SQL, the
`op.alter_column(..., nullable=False)` API spelling, and the **f-string column form
`ruff format` produces**.

That last one mattered. My first regex used `\w+` for the column, so once the
formatter collapsed the statements into f-strings, the guard went **silent on this
very migration** — passing by invisibility rather than on its merits. That is the
documented blind spot which leaves 007 and 026 outside the CREATE INDEX guard.

Then the second: the guard could be satisfied by a **docstring** that merely named
`autocommit_block` while the code didn't use it. Stripping the real one still
passed. It now strips docstrings and comments before checking — line-based, because
rebuilding from `tokenize` re-spaces the source (`op . get_context ( )`) and splits
f-strings on 3.12+, either of which silently breaks a substring check.

Six mutations, all caught, small tables correctly allowed:

| mutation | result |
|---|---|
| naive raw SQL on `memories` | caught |
| `op.alter_column("documents", ..., nullable=False)` | caught |
| f-string column, naive | caught |
| strip real `autocommit_block` (docstring still names it) | caught |
| strip real `VALIDATE CONSTRAINT` (docstring still describes it) | caught |
| same statement on a small table | allowed |

## Verified against real Postgres, not asserted

Seeded a row for each NULL combination at revision 037, then ran the migration:

| row | before | after |
|---|---|---|
| `a-normal` | 01-01 / 02-01 | unchanged |
| `b-null-created` | NULL / 03-01 | **03-01** / 03-01 |
| `c-null-updated` | 04-01 / NULL | 04-01 / **04-01** |
| `d-both-null` | NULL / NULL | both `now()`, **identical** |

`d-both-null` getting one timestamp rather than two milliseconds apart is why the
backfill orders `created_at` first.

Also confirmed: both columns `is_nullable=NO`; helper constraints dropped (0 remain);
a NULL insert is now rejected; the default insert path still works; downgrade and
re-upgrade both clean. And a simulated **mid-migration crash** — leaving a stale
constraint behind, then re-running — succeeds, which is what `DROP CONSTRAINT IF
EXISTS` is there for.

- root suite: **4506 passed**, 3 skipped, 1 xfailed
- `ruff check` on `core-storage-api/src/`, `core-api/src/`, `common/`; `ruff format
  --check` on both src scopes — all verified by exit code

## Note

#826's `datetime | None` on `DocOut` stays. It's now defensive rather than load-bearing,
and I'd rather the API tolerate a shape the DB no longer produces than re-tighten it
in the same breath.

### DIFF
--- a/core-storage-api/src/core_storage_api/database/migrations/versions/038_documents_timestamps_not_null.py
+++ b/core-storage-api/src/core_storage_api/database/migrations/versions/038_documents_timestamps_not_null.py
@@ -1,0 +1,92 @@
+"""``documents.created_at`` / ``updated_at`` NOT NULL.
+
+Both columns were created with ``server_default=now()`` but WITHOUT
+``nullable=False`` (001_initial_schema, never tightened since), so a NULL has
+always been representable even though every insert path sets them. OSS #826 made
+the API honest about that — ``DocOut`` required a ``datetime``, so a NULL row
+raised ValidationError inside the read route and 500'd it, across six call sites
+including three list endpoints. This closes the hole at the source, which is where
+it belongs: the columns describe row lifecycle, not optional user data.
+
+BACKFILL FIRST, because ``SET NOT NULL`` on a table containing a NULL fails — and
+migrations here run in the FastAPI lifespan startup hook, so a failed one means the
+new revision never becomes ready and the deploy stalls (Cloud Run keeps the old
+revision serving; it fails safe, but it fails).
+
+The backfill derives from the row where it can: a NULL ``created_at`` takes
+``updated_at``, which is an upper bound on it, and vice versa. When BOTH are NULL
+there is nothing to derive from — ``id`` is ``gen_random_uuid()`` (v4), so it
+carries no time — and ``now()`` is the only available value. That asserts the row
+was created at migration time, which is false; it is chosen because the alternative
+is refusing to migrate. Such rows are identifiable afterwards by sharing a
+near-identical timestamp with each other and with this migration's run.
+
+LOCK SAFETY. ``documents`` is one of this repo's declared large tables (see
+``test_no_plain_create_index_on_large_tables``), and a bare ``SET NOT NULL`` takes
+an AccessExclusive lock while it full-scans to verify — blocking writes for the
+duration and holding the migration advisory lock, which is the shape that crashed
+6 storage-writer boots on 2026-06-16 (migration 025). Instead:
+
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #828, opened 2026-08-18
feat(usage): durable per-tenant counters behind the usage meter

Eldad-Caura:
Platform half of caura-ai/caura-enterprise#83 — the half that makes counters actually move. #824 added the `ServiceHooks.usage_meter` seam; this is the implementation wired into it, plus the storage it writes to.

## Why the counts are written here, not sent to the platform

The counts must originate in core-api: no interceptor upstream can see `len(body.items)` on a bulk write, the `if auth.tenant_id:` admin skip at every metered call site, or the idempotent replay that deliberately consumes no quota. (`middleware/request_observation.py` already *is* that interceptor and still can't.)

But **core-api has no route to a platform service in any environment** — its only `PLATFORM_*` settings are LLM config, and the dependency runs the other way (`platform-admin-api` holds `CORE_API_URL`). So the meter writes through the storage client core-api already has, and the platform reads `tenant_usage_counters` directly. Both schemas live in one database, so that costs it no network hop and inverts no boundary. This is option A from the discussion on #83.

## Why a second table rather than changing `capability_usage`

That one is adoption analytics and says so: it buffers in memory, its docstring notes counts *"are lost if the process dies"*, and it appends rows to be SUMmed with no unique constraint. Right for "which capabilities get used", wrong for a plan cap — a lost flush is a customer under-billed, and an unbounded append makes every cap check a growing SUM.

`tenant_usage_counters` upserts additively: one row per `(tenant_id, operation, period_start)`, incremented with `ON CONFLICT DO UPDATE SET count = count + excluded.count`.

**Row per operation, not column per operation.** The platform's `enterprise.usage_counters` has `writes`/`searches`/`recalls` columns — which is why `insights` and `evolve`, both metered in core-api, have nowhere to go there.

## Buffered, and why that's safe for billing

The meter coalesces in a dict and flushes every 15s rather than putting a round-trip in front of every write — the same conclusion CAURA-628 reached for audit and `capability_usage` for analytics. What makes it safe here is that the flush is an **additive upsert**:

- a crash loses only the seconds buffered, rather than corrupting a total
- two instances flushing the same period both land
- a **failed flush returns the counts to the buffer** instead of dropping them
- shutdown flushes

## Verified against a real database, not just mocks

- migration applies to head on a fresh database (037 → 038)
- **50 concurrent `+1` plus one `+20` on the same key summed to exactly 70** — the read-modify-write bug this shape exists to avoid would have lost nearly all of them

## Tests

11 unit tests covering the three properties that make buffered counting safe — coalescing preserves the **count** (a bulk write of 20 is not 1), a failed flush **returns** counts, shutdown **flushes** — plus the period key (including a non-UTC instant landing in the right month) and the hook signature `usage_service` invokes by keyword.

`test_single_head` bumped 037 → 038, with a comment saying not to compute it: the pin is what catches two branches both claiming the same `down_revision`.

Full suite: **4516 passed**, 3 skipped, 1 xfailed. `ruff check` + `ruff format --check` clean across every CI scope.

## Next

Counters start moving when this deploys. Making the platform **read** them — replacing or projecting `enterprise.usage_counters` — is the remaining enterprise-side step on #83.

### DIFF
--- a/common/models/__init__.py
+++ b/common/models/__init__.py
@@ -16,6 +16,7 @@ from common.models.memory import Memory
 from common.models.memory_conflict import MemoryConflict
 from common.models.memory_derivation import MemoryDerivation
 from common.models.skill_factory import ForgeRejectedFingerprint, SessionTrace
+from common.models.tenant_usage_counter import TenantUsageCounter
 
 __all__ = [
     "Agent",
@@ -40,4 +41,5 @@ __all__ = [
     "MemoryEntityLink",
     "Relation",
     "SessionTrace",
+    "TenantUsageCounter",
 ]
--- a/common/models/tenant_usage_counter.py
+++ b/common/models/tenant_usage_counter.py
@@ -1,0 +1,88 @@
+"""Durable per-tenant, per-period usage counters (billing-grade).
+
+The exact counterpart to :mod:`common.models.capability_usage`, and separate
+from it on purpose. That table is adoption analytics: core-api buffers counts
+in memory and flushes on a short interval, so anything since the last flush is
+lost if the process dies, and rows are appended to be SUMmed with no unique
+constraint. Both are right for "which capabilities get used" and wrong for
+anything a plan cap is computed from — a lost flush is a customer under-billed
+or over-served, and an unbounded append makes every cap check a growing SUM.
+
+This one upserts: one row per ``(tenant_id, operation, period_start)``,
+incremented atomically with ``ON CONFLICT DO UPDATE SET count = count +
+excluded.count``. No buffering, no loss window, bounded rows, and a period
+total that is a single-row read.
+
+ROW PER OPERATION rather than a column per operation. The platform's
+``enterprise.usage_counters`` has ``writes``/``searches``/``recalls`` columns,
+which is why ``insights`` and ``evolve`` — both metered in core-api — have
+nowhere to go there. Keying the operation as data takes a new one without a
+migration.
+
+Cross-tenant like ``capability_usage``, and for the same reason: it holds
+counts and a ``tenant_id`` grouping dimension, never memory content. RLS is not
+enabled on it. The platform reads it directly to compute plan limits — both
+schemas live in one database, so that costs no network hop.
+
+Written via ``ServiceHooks.usage_meter``, so an OSS standalone deployment with
+no meter wired records nothing and enforces nothing. See
+``core_api.services.usage_service``.
+"""
+
+from __future__ import annotations
+
+from datetime import datetime
+
+from sqlalchemy import BigInteger, DateTime, Index, Text, text
+from sqlalchemy.orm import Mapped, mapped_column
+
+from common.models.base import Base
+
+
+class TenantUsageCounter(Base):
+    __tablename__ = "tenant_usage_counters"
+
+    id: Mapped[int] = mapped_column(BigInteger, primary_key=True, autoincrement=True)
+    # Tenant the usage is attributed to. Grouping dimension, not an RLS key.
+    tenant_id: Mapped[str] = mapped_column(Text, nullable=False)
+    # One of core-api's metered operations: write | search | recall | insights
+    # | evolve. Stored as data so a sixth needs no schema change.
+    operation: Mapped[str] = mapped_column(Text, nullable=False)
+    # Start of the billing period, UTC. Supplied by the caller (month-truncated)
+    # rather than derived here, so a row cannot land in a period other than the
+    # one the caller believes it is counting, and a backfill can target a closed
+    # period explicitly.
+    period_start: Mapped[datetime] = mapped_column(
+        DateTime(timezone=True), nullable=False
+    )
+
+    count: Mapped[int] = mapped_column(
+        BigInteger, nullable=False, server_default=text("0")
+    )
+    created_at: Mapped[datetime] = mapped_column(
+        DateTime(timezone=True), server_default=text("now()"), nullable=False
+    )
+    updated_at: Mapped[datetime] = mapped_column(
+        DateTime(timezone=True), server_default=text("now()"), nullable=False
+    )
+
+    __table_args__ = (
+        # Names and columns must match migration 039 exactly, so a
+        # model-created schema (create_all / autogenerate) agrees with the
+        # migration-created one.
+        #
+        # This unique index is the ``ON CONFLICT`` target. Without it the
+        # upsert has nothing to bind to and concurrent writers quietly create
+        # duplicate rows for one period — the append-and-SUM shape this table
+        # exists to avoid.
+        Index(
+            "uq_tenant_usage_counters_key",
+            "tenant_id",
+            "operation",
+            "period_start",
+            unique=True,
+        ),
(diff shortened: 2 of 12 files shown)


Eldad-Caura:
Both confirmed and fixed in `480a9712`.

**The `KeyError` was real and exactly as described.** Verified before fixing:

```
r = {'tenant_id':'t','operation':'write'}     # no period_start
isinstance(r['period_start'], datetime)       -> KeyError
```

That line sits *outside* the `try/except` below it, so a row omitting the key produced the 500 the coercion block exists to prevent. Now `r.get("period_start")`, with a comment saying why the subscript was wrong, and a regression test asserting 422-not-500 for a row with the key absent.

**Import ordering fixed too** — worth noting it appeared on the rebase, not in my original push: `routers/__init__.py` gained other entries meanwhile and my line ended up in the wrong alphabetical slot.

All CI ruff scopes clean; 12 tests in the new file pass. Rebased on current `main` (`79ccad84`).

Eldad-Caura:
@claude — both findings fixed in `480a9712`. Please re-review.

Eldad-Caura:
Round 2: **1 confirmed and fixed, 1 fixed on a wrong premise, 1 rejected.** Plus CI was red for an unrelated reason and is fixed here too. All in `2fed8b99`.

---

### ✅ Medium — cancellation during a flush drops the batch. Real, and reproduced.

Exactly as described. Written as a test before touching the fix, against the unmodified code:

```
FAILED test_a_shutdown_landing_inside_a_flush_does_not_drop_the_batch
AssertionError: the cancelled batch was never re-sent — stop() found an empty buffer
assert 1 == 2
  where 1 = <AsyncMock name='mock.increment_tenant_usage'>.await_count
```

`await_count == 1` is the whole bug: `stop()` cancelled the loop mid-flush, the batch was already swapped out of `self._counts`, `except Exception` did not fire for `CancelledError`, and `stop()`'s final flush then found an empty buffer and made no call at all.

Fixed by moving the restore into a `finally`, so no exit from `flush` can leave without putting the batch back:

```python
sent = False
try:
    written = await get_storage_client().increment_tenant_usage(rows)
    sent = True
    return written
except Exception:
    logger.exception(...)
    return 0
finally:
    if not sent:
        async with self._lock:
            ...
```

I took this over `shield` and over the "don't cancel while flushing" variant deliberately: cancellation does not only arrive from `stop()` — an enclosing `wait_for`, or event-loop teardown, delivers the same thing — so the guarantee belongs in `flush`, which is the only frame that knows it is holding a batch. Reworking `stop()` would have closed one caller's window and left the rest.

Two things I made explicit in the comment while I was there, because the fix quietly depends on both:

* the restore's own `async with self._lock` is an await point, and it only completes during a propagating cancel because the lock is uncontended and `Lock.acquire` takes a non-yielding fast path. That in turn holds only because `flush` is never re-entered — the loop task and `stop()` are its only callers, and `stop()` joins the task first. A third caller has to re-check it.
* `_run`'s own `except Exception` deliberately still does not catch `CancelledError`. Widening it would swallow the cancellation, loop back to the sleep, and hang `stop()` forever. Left alone on purpose.

**One thing your fix list prompted that wasn't in it.** `stop()` had no deadline, while both of its neighbours in the same shutdown chain take `timeout=5.0`:

```python
shutdown_steps.append(audit_queue.stop(timeout=5.0))
shutdown_steps.append(capability_usage_agg.stop(timeout=5.0))
shutdown_steps.append(usage_meter.stop())          # <- the odd one out, added by this PR
```

The storage client waits 120s on a read. Blocking that long past Cloud Run's grace period loses these counts to SIGKILL *anyway* and strands `event_bus.stop()`, `cancel_all_tasks()` and the storage client's own close behind it. `stop()` now takes the same `timeout=5.0` and logs what it drops — the trade `audit_queue`'s docstring already spells out. Also pinned by a test, which without the fix does not fail so much as hang pytest indefinitely.

---

### ⚠️ Low — `__all__` ordering. Fixed, but not for the stated reason.

The entry was genuinely out of place and is now after `"SessionTrace"`. But it could not have failed a sorted-`__all__` lint rule: **`RUF022` is in the `ignore` list of both** `common/ruff.toml` and `core-storage-api/pyproject.toml`. `ruff check common/` was passing before this change and passes after. Consistency fix only.

---

### ❌ Low — migration line length. Rejected; the limit is 110, not 100.

Line 85 is **102 characters**, and `core-storage-api/pyproject.toml` sets `line-length = 110`. `E501` is ignored on top of that, so neither gate applies:

```
$ ruff format --check .../039_tenant_usage_counters.py
1 file already formatted
$ ruff check .../039_tenant_usage_counters.py
All checks passed!
```

The adjacent `op.execute` is wrapped because it is genuinely over 110, not for symmetry. Leaving line 85 as-is rather than churning the file.

---

### 🔴 Unrelated, and the actual reason CI was red

`Run migrations` failed, which skipped every downstream step including the tests — so the round-1 green was never re-established:

```
UserWarning: Revision 038 is present more than once
ERROR: Multiple head revisions are present for given argument 'head'
```

#827 landed `038_documents_timestamps_not_null` on main while this was open, so both files claimed revision `038`. Renumbered this one to **039** on top of 038, with the two prose references updated.

`TestMigrationChain` should have caught this at PR time and did not. `_load()` keys the chain by revision id, so a duplicate **overwrote** rather than forked — `heads` came back as the expected single `{"038"}` and every downstream assertion read a chain that looked fine. It now asserts each revision is declared exactly once and names both files:

```
AssertionError: revision '038' is declared twice: 038_collision_probe.py and
038_documents_timestamps_not_null.py — renumber the newer one onto the current head
```

Verified by reintroducing the collision as a throwaway file; the guard fires and the chain tests fail. Removed afterwards.

Re-verified on a real database rather than the shared dev one (`DATABASE_URL`, since `POSTGRES_DB` is not read by these settings): the full chain applies `037 -> 038 -> 039` on a fresh database, both indexes and the unique upsert target land, `documents.created_at/updated_at` come out `NOT NULL` proving 038 ran ahead of it, and `downgrade -1` returns cleanly to 038 with the table dropped.

---

**Suite: 4520 passed, 3 skipped, 1 xfailed.** `ruff check` and `ruff format --check` clean across all nine CI scopes.

Eldad-Caura:
@claude — all three findings addressed in `2fed8b99` (two fixed, one rejected with evidence), plus the duplicate-`038` migration collision that was failing CI. Please re-review.

Eldad-Caura:

Review came back clean on `2fed8b99`. Two rounds handled: round 1 was a `KeyError`→500 in the storage endpoint plus import ordering; round 2 was a cancellation window where a shutdown landing inside a flush dropped the buffered batch (reproduced with a failing test first, fixed by restoring in a `finally`), one consistency nit, and one finding rejected with evidence — the line-length limit here is 110, not 100.

Also fixed the duplicate revision `038` that was failing CI after #827 landed on main: this migration is now `039`, and `TestMigrationChain` now catches duplicate revision ids, which it previously could not.

All checks green. Suite 4520 passed, 3 skipped, 1 xfailed.

erni-a (approved):

PR #829, opened 2026-08-19
feat(usage): read half of tenant_usage_counters for the platform

Eldad-Caura:
`#828` made the counters move. Nothing consumes them yet — this is the read the enterprise platform needs. The enterprise-side half of caura-ai/caura-enterprise#83 (swapping `enterprise.usage_counters` for this) is a follow-up PR against that repo.

### The read is an endpoint here, not a cross-schema join in the platform

`#828`'s docstrings said the platform would read `tenant_usage_counters` directly, since `enterprise.*` and `public.*` share one database, "so that costs it no network hop". They do share a database — but that is not the boundary this codebase keeps:

* `platform-storage-api`'s purge method documents that it **"only owns the `enterprise.*` schema"**, and that OSS `public.*` data is reached through core-storage-api instead (CAURA-689).
* `platform-admin-api` already holds a `core_storage_client` and requires `CORE_STORAGE_API_URL` in production — that is how the org purge reaches OSS data today.

So `tenant_usage_counters` keeps one owner, and one HTTP hop on a dashboard read is cheaper than two services agreeing on a table's shape by convention. All three docstrings carrying the old claim are corrected; the model's still read *"the platform reads it directly"* until this commit.

### The org/tenant mismatch it bridges

The platform bills an **organisation**; the meter only ever knows a **tenant**, and an org owns several — 411 tenants across 446 orgs in the enterprise dev database, several orgs with more than one. So `POST /tenant-usage/query` takes a tenant list and sums across it, leaving org→tenants resolution on the platform side where `enterprise.tenants` lives. Same split `purge_org_data(tenant_ids, ...)` already uses, and the only one that keeps this service deployable standalone.

`periods=N` means the newest N periods **that have data** — matching what the legacy `list_usage_history` LIMIT did — rather than N calendar months, since the table has no row for a month nobody used.

Operation names pass through as stored rather than being mapped onto writes/searches/recalls: `insights` and `evolve` are metered by core-api and have no column in that triple, so mapping here would discard counts the write path already pays for.

### Two scans down to one

The obvious shape — a subquery over `tenant_usage_counters` to pick the newest N periods, then filter totals to them — reads the table twice:

```
Sort → HashAggregate → Hash Join
  ├─ Seq Scan on tenant_usage_counters
  └─ Hash → Limit → Sort → HashAggregate
       └─ Seq Scan on tenant_usage_counters      ← second pass
```

Neither pass can prune by `period_start`, because the composite index leads on `tenant_id` with `operation` in the middle — so each one costs the named tenants' whole history. Ranking the **aggregate** with `dense_rank` reads it once instead, over a row set bounded by periods × operations, and Postgres caps it further with a `Run Condition` on the rank:

```
Subquery Scan → WindowAgg (Run Condition: dense_rank() <= 6)
  └─ Sort → HashAggregate
       └─ Seq Scan on tenant_usage_counters      ← only pass
```

`dense_rank` rather than `row_number` is what keeps whole periods: every row of a period ties for one rank.

### Verification

7 integration tests against a live PostgreSQL, each killing a different plausible-wrong query: the sum spanning an org's tenants, a non-member excluded, `insights`/`evolve` surviving, an unknown tenant reading empty rather than 404, and an empty tenant list rejected rather than meaning "every tenant".

Two pin the windowing separately, because neither subsumes the other — a row-`LIMIT` bug is invisible in data with one row per period, and a calendar-window bug is invisible in data with no gaps. Confirmed the first fails against the naive `LIMIT periods`:

```
FAILED test_periods_means_periods_not_rows
assert ['2026-08-01T00:00:00+00:00'] == ['2026-08-01…', '2026-07-01…']
  Right contains one more item: '2026-07-01T00:00:00+00:00'
```

Also fixes a stale `migration 038` reference left behind by #828's renumber to 039.

**Suites:** 4520 passed / 3 skipped / 1 xfailed, plus 125 passed in the core-storage-api integration suite. `ruff check` and `ruff format --check` clean across every CI scope.

### DIFF
--- a/common/models/tenant_usage_counter.py
+++ b/common/models/tenant_usage_counter.py
@@ -21,8 +21,9 @@ migration.
 
 Cross-tenant like ``capability_usage``, and for the same reason: it holds
 counts and a ``tenant_id`` grouping dimension, never memory content. RLS is not
-enabled on it. The platform reads it directly to compute plan limits — both
-schemas live in one database, so that costs no network hop.
+enabled on it. The platform reads it through core-storage-api's
+``POST /tenant-usage/query`` — NOT by joining across schemas, even though both
+live in one database; see that router's docstring for why.
 
 Written via ``ServiceHooks.usage_meter``, so an OSS standalone deployment with
 no meter wired records nothing and enforces nothing. See
--- a/core-api/src/core_api/services/usage_meter.py
+++ b/core-api/src/core_api/services/usage_meter.py
@@ -15,8 +15,13 @@ calling the platform. core-api has no route to a platform service in any
 environment — its only ``PLATFORM_*`` settings are LLM config, and the
 dependency runs the other way (platform-admin-api holds ``CORE_API_URL``).
 Introducing core → platform for metering would invert that. The platform reads
-``tenant_usage_counters`` instead; both schemas live in one database, so that
-costs it no network hop.
+``tenant_usage_counters`` instead, through core-storage-api's
+``POST /tenant-usage/query``.
+
+CORRECTION: this docstring previously said the platform would read the table
+directly, "so that costs it no network hop". They do share a database, but
+reading across schemas is not the boundary this codebase keeps — the rationale
+is in that router's docstring.
 
 BUFFERED, NOT PER-WRITE
 -----------------------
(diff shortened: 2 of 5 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #830, opened 2026-08-19
fix(embedding): reserve capacity so write bursts can't starve recall

Eldad-Caura:
## Why

`EMBEDDING_MAX_CONCURRENCY` is **priority-blind**: document/bulk embeds and the latency-critical search-side query embed draw from one pool, so a burst of background writes holds every slot and starves live recall.

Not hypothetical — from the 5-day prod error map (2026-08-14→19):

| | |
|---|---|
| **2026-08-18 17:48–18:08Z** | bulk write burst (~1.3 k memories in 2 h vs ~5/h baseline) |
| **118** | `Query embedding failed after 2 attempts` → user-visible search degradation |
| **TEI at the time** | 2 active instances, ~5 of 10 container concurrency, `pending_queue` 0, ~3 ms inference, 42.9 k requests served |

The backend had headroom the whole time. **The queue was at our own gate, not at TEI** — which is why adding capacity isn't the fix (and can't be: L4 is capacity-blocked in us-central1, so the 3→8 quota bump was denied).

## What

Reserve a slice of the cap that only query embeds may occupy:

- **`EMBEDDING_QUERY_RESERVED_SLOTS`** — default proportional (`max(1, cap // 4)`) so it tracks whatever cap a deployment sets. Clamped to leave background ≥1 slot, and it *warns* on clamp rather than silently applying it (matching `clamp_keepalive`'s documented rationale).
- Background work additionally passes a tighter gate sized `cap - reserved`; query embeds draw on the **full** cap.
- **Total in-flight is still bounded by `EMBEDDING_MAX_CONCURRENCY`** — the reservation *partitions* the cap rather than adding to it, so the backend-protection invariant the cap exists for is unchanged.

**Deadlock-free by arithmetic, not by luck:** background takes its own gate *before* the shared one, and the reserved slice is the exact difference between the caps — so even with queries holding every reserved slot, the shared slots still free equal the background gate's size. Query embeds only ever take the shared gate, so they can never queue behind background work. `AsyncExitStack` unwinds both slots on **any** escape (including `CancelledError`); leaking one per abandoned wait would erode the background budget to zero and strangle writes.

## The non-obvious part: two interactive callers use `get_embedding`

`get_embedding` defaults to `background=True` (ingest is the dominant caller) but takes an opt-out, because two **interactive** callers go through it and must not be demoted:

- `POST /documents/search` (`routes/documents.py`) — **raises 503** when the embedding is `None`
- the `caura_doc op=search` MCP tool (`mcp_server.py`)

Left on the background budget, document search would be throttled behind exactly the write floods this change exists to survive — turning a bulk ingest into user-facing search 503s. The same applies to a synchronous write path (`EMBED_ON_HOT_PATH=true`), which this module **cannot** detect for itself: `common/embedding` deliberately has no dependency on service config, so only the call site knows whether anyone is waiting.

## Tests

Four guarantees, in `tests/test_embedding_concurrency_gate.py`:
1. a query embed completes against a background flood **sized to the full cap** (the starvation guarantee)
2. background never exceeds `cap - reserved`
3. total never exceeds the shared cap across both classes
4. the search opt-out is honoured

I verified each **fails against a priority-blind gate** rather than passing either way — (1) initially didn't, because I'd sized the flood to the background budget instead of the full cap, leaving free slots regardless; that's fixed and noted in the test docstring.

Suites: 55 passed / 1 skipped across the embedding tests. `ruff check common/` and the touched core-api paths clean; added `common/` lines kept ≤88 per that tree's convention.

## Scope note

This is the **in-process** reservation. Priority ordering is the one property that composes across instances without coordination (it's relative, not an absolute budget), so this is the right layer — but the aggregate fleet-wide cap (`16 × N` instances vs TEI's ~30) remains a separate, unsolved problem, unchanged by this PR.

### DIFF
--- a/common/embedding/_service.py
+++ b/common/embedding/_service.py
@@ -12,9 +12,11 @@ import logging
 import os
 import time
 from collections.abc import Awaitable, Callable
+from contextlib import AsyncExitStack
 
 from common.embedding._registry import get_embedding_provider
 from common.embedding.constants import (
+    EMBEDDING_BACKGROUND_MAX_CONCURRENCY,
     EMBEDDING_BUDGET_MARGIN_S,
     EMBEDDING_GATE_TIMEOUT_SECONDS,
     EMBEDDING_MAX_CONCURRENCY,
@@ -246,6 +248,8 @@ _misconfiguration_logged: set[str] = set()
 # stack.
 _gate: asyncio.Semaphore | None = None
 _gate_loop: asyncio.AbstractEventLoop | None = None
+_bg_gate: asyncio.Semaphore | None = None
+_bg_gate_loop: asyncio.AbstractEventLoop | None = None
 
 
 def _concurrency_gate() -> asyncio.Semaphore:
@@ -261,7 +265,25 @@ def _concurrency_gate() -> asyncio.Semaphore:
     return _gate
 
 
-async def _call_gated[T](make_call: Callable[[], Awaitable[T]]) -> T:
+def _background_gate() -> asyncio.Semaphore:
+    """The tighter cap that only document/bulk embeds are held to.
+
+    Sized at ``EMBEDDING_BACKGROUND_MAX_CONCURRENCY`` so
+    ``EMBEDDING_INTERACTIVE_RESERVED_SLOTS`` of the shared cap stay reachable by
+    search-side query embeds during a write flood. Same per-event-loop
+    rebinding rationale as :func:`_concurrency_gate`.
+    """
+    global _bg_gate, _bg_gate_loop
+    loop = asyncio.get_running_loop()
+    if _bg_gate is None or _bg_gate_loop is not loop:
+        _bg_gate = asyncio.Semaphore(EMBEDDING_BACKGROUND_MAX_CONCURRENCY)
+        _bg_gate_loop = loop
+    return _bg_gate
+
+
+async def _call_gated[T](
+    make_call: Callable[[], Awaitable[T]], *, background: bool
+) -> T:
     """Run *make_call* holding a concurrency slot.
 
     The slot is acquired per ATTEMPT, not per retry sequence, so a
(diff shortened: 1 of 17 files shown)


Eldad-Caura:
Both review findings verified and fixed, plus the CI failure (which was caused by this PR).

**High — inline synchronous write on the background budget: confirmed and fixed.** Line 119 sits in the `elif not defer_embedding:` branch, so a caller is genuinely blocked on the HTTP response (`deployment_mode="inline"` — the OSS default — or `write_mode="strong"`, always inline). Now passes `background=False`. Good catch: my own docstring named this call site and I hadn't updated it.

**Low — stale flag name: confirmed and fixed.** `config.py:54` documents `embed_on_hot_path` as the *legacy* name replaced by `deployment_mode`. The docstring now points at `deployment_mode="inline"` / `write_mode="strong"` and names the actual call site (`parallel_embed_enrich.py`) instead of the dead env var.

**CI failure was mine.** `test_b4_doc_search_no_embedding_sets_iserror` returned `INTERNAL_ERROR` instead of `UPSTREAM_ERROR` because the `_no_vector` test double is positional-only — my new keyword-only `background` raised `TypeError`, which the handler mapped to `INTERNAL_ERROR`. Widened both `_no_vector` stubs in that file to accept the parameter, with a comment explaining why. Local: 143 passed / 1 skipped across the embedding, MCP, write-path, docs and backfill suites; `ruff check` clean.

I also checked the other `get_embedding` test doubles (`test_mcp_doc.py`, `test_backfill_embeddings.py`, `test_documents_rest_doc_memory.py`) — those already tolerate the kwarg, which matches CI having passed them.

Single amended commit, force-pushed with lease.

Eldad-Caura:
@claude

Eldad-Caura:
Verified and fixed — and I went further than the two flagged sites, because this is the **third** review cycle to find the same class of miss.

**Confirmed both.** `routes/documents.py:304` raises **HTTP 502** on `None` and `mcp_server.py:1817` returns `UPSTREAM_ERROR`, so a caller is blocked in each. Both now pass `background=False`.

**Two more the review didn't flag.** Rather than patch only what was reported, I enumerated *every* production `get_embedding(` call site (9 real ones) and classified each. Two additional interactive callers were still on the deferred budget:

- `memory_service.py:649` — inline create; the request awaits it before responding
- `memory_service.py:2856` — synchronous update; the caller awaits the re-embed

Both fixed. That's all four currently-known interactive callers, so a fourth cycle shouldn't find another. The five genuinely-deferred sites (backfills, entity-extraction worker, background re-embed, atomic-fact children, the storage CLI) correctly keep the default.

**Verification against a measured baseline.** `tests/test_mcp_doc.py` had two positional-only `get_embedding` doubles that couldn't accept the kwarg; widened them. I then ran the full suite both with and without my changes: unmodified `origin/main` fails **26 / errors 9**, and this branch fails **26 / errors 9** — the same files (`test_events/*`, `test_gemini_provider`, `test_provider_protocols`, `test_ingest_*`), which are pre-existing and unrelated. **This PR introduces zero new failures.** `ruff check` clean on `common/` and `core-api/src`.

### Deliberately not in this PR

The recurrence is a design signal, not bad luck: with `background=True` as the default, *forgetting* silently demotes a user-facing path, so every new call site inherits the bug. The robust fix is to make the parameter **required** (omission becomes a `TypeError`) — and, since synchronous writes need protection too, to reframe the reservation as *interactive vs deferred* rather than *query vs background*.

I prototyped exactly that and measured the cost: it breaks **~18 additional test doubles** across the suite that don't accept the parameter. That's a legitimate refactor but a poor fit for an incident fix — it would triple this diff and bury the actual behaviour change. Happy to do it as an immediate follow-up; flagging it explicitly rather than leaving the sharp edge undocumented.

Eldad-Caura:
@claude

Eldad-Caura:
Verified and fixed. This one was my mistake in reasoning, not just an oversight: I'd explicitly decided "bulk is always background, leave `get_embeddings_batch` alone" — but it's also called on synchronous paths, so that was wrong.

**Fixed as prescribed:** `get_embeddings_batch` takes a keyword-only `background: bool = True` threaded to both `_call_gated` calls, and the three synchronous callers now pass `background=False`:

- `memory_service.py:1242` — `create_memories_bulk` under `inline_embedding` / `write_mode="strong"`. Its own comment says *"this is the only place a row gets its vector, so a failure fails the request"* — definitively user-blocking.
- `memory_service.py:546` and `:816` — auto-chunk children of a synchronous create, matching the parent embed already fixed to `background=False`.

`_reembed_batch_via_provider` (`:2211`) correctly keeps the default — confirmed it's the genuine background re-embed.

**Both entrypoints are now exhaustively enumerated**: 9 `get_embedding` callers + 4 `get_embeddings_batch` callers, every one classified. The set of call sites is closed, so a further round can only surface a *mis*-classification, not a missed site.

**Verification:** I ran the full suite against a measured baseline. Unmodified `origin/main`: **26 failed / 4507 passed / 9 errors**. This branch: **26 failed / 4507 passed / 9 errors** — identical, same pre-existing files (`test_events/*`, `test_gemini_provider`, `test_provider_protocols`, `test_ingest_*`). Zero new failures. `ruff check` clean on `common/` and `core-api/src`.

### Standing recommendation

Four review rounds have each found a different call site inheriting the permissive default. The individual fixes are now complete, but the *shape* still invites round five for any call site added later: `background=True` as a default means forgetting silently demotes user-facing work. Making the parameter **required** on both entrypoints removes the failure mode entirely (omission becomes a `TypeError`). Measured cost: ~18 test doubles need to accept the parameter. I've kept it out of this PR to avoid tripling an incident fix, but it's a small, mechanical follow-up and I'd recommend taking it.

Eldad-Caura:
Folded in the root-cause fix, and fixed the CI failure — which was **my** miss, not tests: the failing step was `Run Ruff formatter check (core-api)`, not `Run tests`. I'd run `ruff check` but not `ruff format --check`. They're separate gates and I know it; I've now run **every** CI ruff scope locally (check *and* format, plus the `--isolated` 88-col gate for vendored `common/` files) and all pass. The formatter only collapsed my own over-wrapped calls — no unrelated churn.

### `background` is now REQUIRED on both entrypoints

No default, on `get_embedding` **and** `get_embeddings_batch`. Omitting it is a `TypeError`, so a new call site cannot silently inherit the permissive value. All **13** production call sites are classified explicitly:

**Interactive** (`background=False`) — a caller is blocked:
- `POST /documents/search`, `caura_doc op=search`
- `POST /documents`, `caura_doc op=write`
- inline/strong memory create, synchronous update, the inline pipeline embed
- bulk create under `inline_embedding`/`strong`, and both auto-chunk child batches

**Deferred** (`background=True`) — nobody is waiting:
- background re-embed (single + batch), atomic-fact children, entity-extraction worker, entity-embedding backfill, the storage CLI backfill

Also renamed `EMBEDDING_QUERY_RESERVED_SLOTS` → `EMBEDDING_INTERACTIVE_RESERVED_SLOTS` and reframed the docs as **interactive vs deferred**. The old name was actively misleading once synchronous writes were protected too — and that query-centric framing is part of why four rounds each missed a different write path. The constant is new in this PR, so the rename costs nothing downstream.

### Cost, measured rather than estimated

My earlier "~18 test doubles" estimate was from a broader prototype and turned out pessimistic. Actual: **6 stub signatures** widened to accept the parameter, and **20 direct test call sites** (in `test_embedding_retry.py` / `test_bulk_write.py`, which exercise the entrypoints directly) given an explicit value — `background=False` there, so those gate tests keep their existing single-gate concurrency semantics.

### Verification

Full suite run against the baseline both ways: unmodified `origin/main` → **26 failed / 4507 passed / 9 errors**; this branch → **26 failed / 4507 passed / 9 errors**, identical files (`test_events/*`, `test_gemini_provider`, `test_provider_protocols`, `test_ingest_*` — all pre-existing and unrelated). **Zero new failures.** Every ruff gate green.

Eldad-Caura:
@claude

Eldad-Caura:
Both items addressed.

**Low — inaccurate comment: confirmed and fixed.** You're right that the two branches differ: under `settings.inline_embedding` the handler raises (504 on timeout, else re-raise), but a *deferred* deployment with a `write_mode="strong"` item logs and falls through to the backfill path. The comment now states that explicitly and keeps the reason `background=False` is correct either way — the caller synchronously awaits the batch in both cases.

**Suggestion — internal helpers: taken, and it caught a real gap.** `_call_gated` and `_run_with_retry` did still default to `background=False`, which is the same silent-inheritance shape the public API just removed — so an internal caller added against them could quietly get interactive priority. Both now require it. That surfaced one genuine implicit caller: `get_query_embedding` had been relying on the default, and now says `background=False` explicitly with the reason. Every internal and public call site is now explicit; I re-scanned to confirm none are left.

**Verification.** Making the internal helpers required broke five `_call_gated(...)` calls in the pre-existing gate tests — a real contract change, so those now pass `background=False` (preserving their single-gate semantics). Results:

- Full suite: **26 failed / 4507 passed / 9 errors** — identical to unmodified `origin/main`, same pre-existing files. Zero new failures.
- Every CI ruff gate run locally: `check` on all 8 scopes and `format --check` on all 5, plus the `--isolated` 88-col vendored gate. All green. (Last round's CI failure was the core-api *formatter* gate, which I'd skipped — not repeating that.)
- Re-confirmed the reservation tests aren't vacuous: with the reservation neutralized, **3 of them fail**; with it in place all 10 pass.

Eldad-Caura:
@claude

Eldad-Caura:
CI is green and the Claude review came back clean (no issues found) after six cycles. @erni-a please approve for merge.

**What this fixes:** the embedding concurrency cap was priority-blind, so a bulk write burst could hold every slot and starve interactive embeds. On 2026-08-18 that produced 118 failed query embeddings — user-visible search degradation — while TEI itself sat at ~3 ms inference with spare capacity. The queue was at our own gate, not the backend, and adding backend capacity isn't available (L4 is capacity-blocked in us-central1, so the quota bump was denied).

**Approach:** reserve a slice of the existing cap for interactive work. Total in-flight stays bounded by `EMBEDDING_MAX_CONCURRENCY` — the reservation partitions the cap rather than adding to it, so the backend-protection invariant is unchanged. Deadlock-freedom follows from the arithmetic, and `AsyncExitStack` guarantees both slots unwind on any escape including cancellation.

**Why it took six rounds, and what closed it:** the first five each found a different call site silently inheriting a permissive default. Patching them individually was never going to converge, so `background` is now **required** on both public entrypoints *and* both internal helpers — omission is a `TypeError` — with all 13 production call sites explicitly classified. The constant was also renamed to `EMBEDDING_INTERACTIVE_RESERVED_SLOTS`, since the original query-centric name was part of why the write paths kept getting missed.

**Verification:** full suite run against a measured baseline — unmodified `main` and this branch both report 26 failed / 4507 passed / 9 errors on the same pre-existing files, so zero new failures. All eight `ruff check` scopes and all five `ruff format --check` scopes pass, plus the `--isolated` 88-col vendored-`common/` gate. The four reservation tests were each confirmed to fail against a priority-blind gate rather than passing either way.

Note for after merge: this only reaches prod on the next promote, and it's an in-process reservation — the fleet-wide aggregate cap (`per-process × instances` vs TEI's ~30) remains a separate open problem, unchanged by this PR.

erni-a (approved):

PR #831, opened 2026-08-19
CAURA-717 feat(enrichment): V2.2 memory-type taxonomy — deprecate int…

eyal-bl:
…ention/commitment/cancellation, sharpen action-vs-decision

Refines the ENRICHMENT_PROMPT memory-type section based on a full-pool Opus 4.8 relabeling of eToro's 5,251-row enrichment corpus with a V2.2 taxonomy (drafted after a 999-row diagnostic pilot). The changes fall into two categories: reducing the label set from 10 to 7, and tightening the boundary between execution-deeds and deliberation-choices.

Changes to the classifier surface (prompt-only; no DB schema change):

1. `intention`, `commitment`, and `cancellation` folded into the appropriate adjacent types. At full-pool Opus scale (N=5,251 under V2.2) all three collapse to 0 rows — every case fit cleanly into `plan` (structured goals / long-range targets), `action` (deed execution), `task` (standing instruction / pending work), or `decision` (documented change of direction). The three enum entries are retained for read-compat with historical rows (~55 in eToro); `CLASSIFIER_DEPRECATED_MEMORY_TYPES` grows from `{"semantic"}` to `{"semantic", "intention", "commitment", "cancellation"}` and the existing CAURA-701/702 demotion infrastructure (prompt filter, `_validate_enrichment` in single-write, `create_memories_bulk` + `update_memory` in bulk/update, and the crystallizer per-fact guard) picks them all up automatically.

2. `decision` rewritten around the deliberation markers rather than a verb list. A decision now requires at least one visible marker of deliberation — reasoning ("because…"), comparison of alternatives ("X over Y"), or group/meeting framing ("the team concluded") — because the V2.1 phrasing invited bare approval-deeds ("approved config X", "paused loop Y") to be mislabelled as decisions instead of actions. The V2.1 signal-phrase list ("opted for", "selected") was removed to make room for the new deliberation-oriented cues.

3. `action` verb list expanded to explicitly own the execution of a choice: `approved`, `signed off`, `rejected`, `cancelled` are now named alongside `deployed`/`merged`/`sent`. An "Action vs decision — the tiebreaker" resolver section was added in the same style as the existing "Action vs episode vs fact" section: if the content is just verb + object with no visible deliberation, it's action.

4. `plan` widened to include long-range organizational goals and targets (the endpoint a roadmap points at, even when the steps aren't spelled out) — so "The goal is $100M ARR by 2029" has a principled home instead of being crammed into `fact`.

5. `preference` and `task` gained "Look for" cue lists parallel to the ones `decision` and `action` already had — surfaces V2.1 was under-picking because they lacked explicit trigger phrases.

6. Contrastive example block grew from 10 to ~18 lines, covering the new plan/preference/decision/action pairs. The example prompt-length ceiling in `test_a8`, `test_a9`, and `test_enrichment_prompt` was raised from 1500 to 1800 words to match.

7. `crystallizer_service.py` had its own hardcoded LLM-prompt vocabulary listing the deprecated slugs — those three plus `semantic` are dropped from the enumerated set. The existing per-fact demotion in `_run_crystallization` (already added in CAURA-701) reads `CLASSIFIER_DEPRECATED_MEMORY_TYPES` at runtime, so no additional demotion code is needed on this path.

New test file `tests/test_caura717_new_deprecations.py` — 14 unit tests
+ 3 integration tests. Pins each of the three new deprecated slugs by name so an accidental removal from the frozenset fails a test whose CAURA ticket names the exact PR to check. The existing CAURA-702 test file iterates over `CLASSIFIER_DEPRECATED_MEMORY_TYPES` generically and covers the bulk/update demotion for the new slugs automatically.

Validation:

- Full-pool relabeling (N=5,251) with V2.2 completed with 0 errors and 0 rows landing on any deprecated slug. Opus 4.8 self-consistency measured at every 50-row checkpoint stayed constant across the run.
- Overall nano↔Opus agreement on the 7 comparable classes is 45.1%, matching the 45.5% observed on the 1,000-row pilot — the same distillation signal at 5× scale. `episode` reclassification is the primary source of disagreement (18.3% agreement rate — nano's episode bucket is 2,569 rows vs Opus's 649).
- 62 unit tests pass locally against the new prompt. Integration tests (bulk/update demotion) require Postgres+pgvector and run in CI.

No schema, migration, or storage-layer changes. Historical rows tagged `intention`/`commitment`/`cancellation` (and `semantic`) remain untouched; the recall filter still lists every stored type so they stay queryable. New writes go to the successor type. Historical-row backfill is out of scope and can be tracked as a follow-up if needed.

Full write-up: /Users/eyal/code/memtype-fine-tune/prompt_v22_draft.md

## Type of Change

- [ ] Documentation update
- [ ] Other:

## Checklist

- [ ] `mypy` passes
- [ ] `pytest` passes locally


eyal-bl:
@claude

eyal-bl:
Thanks for the review. Both findings are real bugs but they're out of scope for CAURA-717 and belong in a follow-up.

**On the crystallizer prompt at line 65** — the fact that the vocabulary list still offers `"outcome"` and `"rule"` predates this PR. It's a CAURA-699 leak that CAURA-701 also didn't close on this code path; my change on that line was scoped to dropping the three deprecated slugs, not auditing what other slugs remain. I agree it's inconsistent with the CAURA-699 policy — that inconsistency has just been sitting there since CAURA-699 shipped.

**On `_crystallize_cluster` at ~line 427** — that fallback is a separate code path I never touched. Bundling a policy fix there into a taxonomy-deprecation PR would dilute the ticket and make the diff harder to review historically.

Filing both as a follow-up (CAURA-NNN) with its own tests so the CAURA-699 enforcement is finally uniform across the crystallizer's LLM-facing prompt and its per-fact fallback. Not merging that into this PR.

Everything else — the constants change, the prompt updates, the deprecation flow, the tests — should stand on its own merits.

eyal-bl:
@claude

eyal-bl:
Round 2 response — accepting both findings this time.

**#1 (Medium — crystallizer prompt duplicates the vocabulary):** you're right, and this framing is more compelling than the round-1 argument. The manual edit this PR needed on line 65 IS the drift risk, and the fix uses the same pattern `common/enrichment/_prompts.py` already established. Refactored `CRYSTALLIZATION_PROMPT` to build its offered vocabulary from `MEMORY_TYPES_WRITE` at import time. As a side effect, this also closes the earlier CAURA-699 leak (round-1 finding on `outcome`/`rule`) since `MEMORY_TYPES_WRITE` already excludes both `SERVER_RESERVED_MEMORY_TYPES` and `CLASSIFIER_DEPRECATED_MEMORY_TYPES` — the two concerns collapse into a single refactor.

**#2 (Low — stale doc comments):** fixed. Updated the `_CLASSIFIABLE_TYPES` comment to name all four deprecated types (`semantic` from CAURA-701 + `intention`/`commitment`/`cancellation` from CAURA-717) and removed the "commitments" references from the `status` field guidance.

The `_crystallize_cluster` fallback (round-1 finding #2) still lives in its own PR — that's a separate function I never touched here, and the reserved-type-guard fix belongs in the same follow-up that hardens the crystallizer against all classes of policy violations.

All 69 relevant unit tests pass locally, OpenAPI baseline stays current (schema description strings unchanged), and force-pushed.

eyal-bl:
@claude

eyal-bl:
Fair point — accepting this. My refactor did narrow the offered vocabulary (prompt now excludes `outcome`/`rule`) while leaving the two post-LLM guards checking the full 14-value enum, so the asymmetry the review names is real. To be candid: the underlying leak (server-reserved types bypassing at the crystallizer path) is pre-existing since CAURA-699, but my refactor made the asymmetry visible enough that closing it here is cleaner than deferring.

Fixes:
- **`_do_crystallize` parse loop** now checks `MEMORY_TYPES_WRITE` (not `MEMORY_TYPES`), coercing any reserved OR deprecated slug to `DEFAULT_MEMORY_TYPE` at parse time. Removed the now-unused `MEMORY_TYPES` import.
- **`_run_crystallization` post-parse guard** widened to demote `SERVER_RESERVED_MEMORY_TYPES` alongside `CLASSIFIER_DEPRECATED_MEMORY_TYPES` — defense-in-depth, in case a future edit relaxes the parse-time guard.
- **`test_caura717_new_deprecations.py`** gained three unit tests: crystallizer prompt no longer offers deprecated slugs by name, prompt no longer offers reserved slugs by name, and `MEMORY_TYPES_WRITE` membership matches the intended 7-slug set for the guard.

76 relevant unit tests pass locally, OpenAPI baseline current, force-pushed.

This closes the CAURA-699 leak on the crystallizer path in the same PR, so the follow-up ticket I was going to file now becomes empty — no separate PR needed.

eyal-bl:
@claude

Eldad-Caura (approved):

Issue #813, opened 2026-08-18
H-03: ENTITY_LOOKUP short-circuit skips vector scoring entirely — null similarity, top_k under-fill, unreachable memories


Eldad-Caura:
Finding **H-03** of the 2026-08-14 OSS/platform engineering audit, rated **high**. Filed now so the ten correctness highs are trackable — they previously existed only inside the audit artifact. The eight security highs from the same audit are all fixed and merged (#799, #801, #802, #804, #806).

**Area:** core-api · **Surfaced by:** bugs-search-pipeline

## Verified still present

Checked against `main` at `a6c6547a` on 2026-08-18 before filing:

`ENTITY_LOOKUP_MAX_MATCHES = 20` at `constants.py:570`, and `classify_query.py:151-152` still returns `skip_embedding=True, skip_scored_search=True`.

This is the one already observed in production.

## Mechanism

[...]

PR #832, opened 2026-08-19
fix(search): stop entity_lookup answering a query it cannot fill

Eldad-Caura:
Fixes the H-03 recall bug from the 2026-08-14 audit — the one already observed in production. Premise re-verified on `main` before changing anything; the search pipeline had not moved since the issue was filed.

## The bug

`ClassifyQuery` short-circuited to `ENTITY_LOOKUP` whenever entity FTS matched 1..20 entities **and** graph expansion linked at least *one* memory. That plan sets `skip_embedding` + `skip_scored_search`, so `ExecuteScoredSearch` and `RerankResults` SKIP outright and `PostFilterResults` skips on the strategy name — no vector score, no FTS rank, no freshness, no `min_similarity`.

The short-circuit doesn't re-rank, it **replaces** scoring. On "at least one row" that produced exactly the three reported symptoms: `top_k` under-fill (3 linked memories answering `top_k=20` with 3 rows), `similarity=None` on every row, and memories unreachable at *any* `top_k` — the pool is cut to `GRAPH_MAX_BOOSTED_MEMORIES` before content loads, so the relevant memory could be dropped before `top_k` was even applied.

Note the perverse shape: the existing `>20` guard declines when the match is **broad**, so small loose matches hijacked the query while broad ones fell through to the real scorer.

## The fix

The short-circuit now requires the entity pool to fill the caller's `top_k`. Below that, the query falls through to the scored cascade and the entity signal survives as a boost — the hops go to `ParallelEmbedAndEntityBoost`, which applies them over real scores. Entity-linked memories still rank high, now alongside everything else and with a readable similarity.

Two things deliberately **not** done:

- **Not tightened at the FTS layer.** The any-token OR is a prior fix (A7) that stops `Helios telescope` hiding `Helios Robotics`. The bug is the exclusivity, not the recall.
- **Not the two larger alternatives** the issue floats. Retiring the exclusive route entirely, and gating on match *precision*, both stay open — the precision gate needs `fts_search_entities` to return names or rank instead of bare UUIDs, i.e. a core-storage-api contract change.

## The part that would not have been free

Gating only *after* the load would have made the newly-common path pay for rows it discards: a full `load_memories_by_ids` (rows carry the 1536-dim embedding and the tsvector), one of only four per-tenant storage permits, and — worst — `_storage_slot_acquired` left set, so the scored search that *does* run would have slipped past the CAURA-602 bulkhead unmetered. The bulkhead would have been protecting the call that didn't matter and not the one that did.

So `_collect_memories` bails before the load when the pool cannot fill `top_k`. Sound one-directionally: the row build can never yield more rows than `memory_boost` has entries. The converse isn't decidable there — the load applies visibility filters and drops rows whose memory didn't come back — so the post-load check stays as a second gate for a pool that looked adequate and returned short. That second case is the "matched but produced no filtered rows" fall-through the C10 comment already describes, and the pre-load exit is what keeps it rare, as that comment claims.

## Tests

Three new cases: the production symptom (and that the load is *not* issued for it), boost survival, and an adequate pool still taking the exclusive route.

Boost survival is the subtle one. The `>20` over-broad decline deliberately sets `entity_match_declined` to **suppress** hop-boosting, because N ≫ `GRAPH_MAX_BOOSTED_MEMORIES` siblings at hop 0 degenerate into an arbitrary-50 lottery. This new under-fill decline must **not** set it, or falling through would be a downgrade rather than a fix. The two declines log distinct phrases so ops can grep them apart.

Existing fixtures were updated rather than relaxed: several paired a one-memory pool with `top_k` 5 or 10 — precisely the buggy shape. Each now asks for what its fixture links, via a `top_k` seam on the ctx builders, so every test keeps its own subject (dict-shaped `expand_graph`, payload forwarding, the `>20` boundary, slot accounting, the org flag) instead of asserting the bug. Two notes: the `top_k` bump in the `>20` decline test was a no-op — that path clears `matched_ids` before `_collect_memories` runs, as the test's own `assert_not_awaited` proves — and is reverted; and the c10 sentinel tests need the load to actually happen, so they ask for one row rather than ten.

Verified by reverting: dropping the pre-load exit alone fails the "load must not be awaited" assertion while the strategy assertion still passes (the layering working); dropping both fails the strategy and boost assertions.

## Checks

4553 passed, 3 skipped, 1 xfailed. `ruff check` and `ruff format --check` clean on `core-api/src/`. No storage-side change, so the storage suite is untouched. Rebased onto current `main` (`2cc741ee`, which brought CAURA-717) and the suite re-run after.

## Follow-ups this surfaced, not in scope here

- `get_memory_ids_by_entity_ids` is issued twice per fall-through — once in `ClassifyQuery`, again in `_entity_boost_via_storage`. `precomputed_hops` already suppresses the FTS and `expand_graph` re-runs; extending that channel to carry the links would close it.
- `_expand_per_fleet` returns `{}` when every `expand_graph` call fails, and `{} is not None`, so the boost step takes the precomputed branch and silently applies no boost. Pre-existing, but it is the mechanism this fix leans on.
- `RerankResults` reads `r.similarity` by direct attribute access and `PostFilterResults` reads `ctx.data["raw_rows"]`, neither of which the entity path populates — so those skip guards are crash guards, not just cost guards. Worth declaring a candidate-row type before anyone attempts the full merge.

Refs #813

### DIFF
--- a/core-api/src/core_api/pipeline/steps/search/classify_query.py
+++ b/core-api/src/core_api/pipeline/steps/search/classify_query.py
@@ -1,10 +1,13 @@
 """ClassifyQuery — classify incoming query into a retrieval strategy.
 
 Examines the query tokens against the entity full-text index.  When entity
-matches are found the step short-circuits to an *entity_lookup* strategy
+matches are found the step MAY short-circuit to an *entity_lookup* strategy
 (graph-expanded, scored by hop distance) so downstream embedding and scored
-search can be skipped.  Otherwise the query is routed to keyword or semantic
-search based on the adaptive FTS weight.
+search can be skipped — but only when the linked-memory pool can fill the
+caller's ``top_k``.  That route replaces scoring rather than re-ranking it, so
+an under-filled pool falls through instead and the entity hits are applied as a
+hop boost over real scores (H-03).  Otherwise the query is routed to keyword or
+semantic search based on the adaptive FTS weight.
 """
 
 from __future__ import annotations
@@ -144,7 +147,25 @@ class ClassifyQuery:
                         slot_acquired_marker=ctx.data,
                     )
 
-                    if filtered_rows:
+                    # H-03: the short-circuit REPLACES scoring rather than
+                    # re-ranking it, so it is only defensible while the entity
+                    # pool can answer the request on its own. Below top_k it made
+                    # the result strictly worse than never matching an entity:
+                    # memories outside the link set were unreachable at ANY top_k,
+                    # and the pool is itself cut to GRAPH_MAX_BOOSTED_MEMORIES
+                    # before content loads, so the relevant memory could be
+                    # dropped before top_k was even applied.
+                    #
+                    # ``filtered_rows`` is already ``rows[:top_k]``, so this can
+                    # only ever be equality — the real gate is the pre-load pool
+                    # check in _collect_memories, and this re-checks it because
+                    # visibility filtering there can still return short. Keep the
+                    # two in step if _collect_memories ever gains an overfetch
+                    # bound, or this silently becomes always-true.
+                    #
+                    # The truthiness term is not redundant: it keeps a zero-row
+                    # pool from short-circuiting were top_k ever 0.
+                    if filtered_rows and len(filtered_rows) >= top_k:
                         plan = RetrievalPlan(
                             strategy=RetrievalStrategy.ENTITY_LOOKUP,
                             matched_entity_ids=matched_ids,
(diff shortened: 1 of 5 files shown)


Eldad-Caura:
@claude — confirmed and fixed in `452e62b3`.

You found the log lying in precisely the case the post-load re-check exists for, which makes it the one decline an operator would most need to tell apart. Walked the three paths to be sure:

| pre-load pool | load | `filtered_rows` | `_entity_pool_size` | old message | correct? |
|---|---|---|---|---|---|
| empty | skipped (`not memory_boost`) | `[]` | unset | "no linked memories" | ✅ |
| `< top_k` | skipped (pre-load exit) | `[]` | set | "under-filled (pool N)" | ✅ |
| `>= top_k`, trimmed after | **ran** | **non-empty** | unset | "no linked memories" | ❌ |

Now three mutually-exclusive branches with three messages: post-load shortfall names the surviving row count and blames visibility filtering, pre-load under-fill names the pool, and only a genuinely empty pool gets the empty message. They cannot overlap — the pre-load exit returns `[]`, so a non-empty `filtered_rows` proves the load ran.

**On the test**, one thing worth flagging because it changes what the test is worth: asserting on the log alone would have passed for the wrong reason, since the pre-load path also declines. So the fixture links 5 memories and has storage return only 2 — pre-load pool adequate, load awaited, shortfall appearing only afterwards — and the test asserts `load_memories_by_ids.assert_awaited_once()` so it cannot pass by accidentally taking the pre-load exit. That required giving the fixture a `loaded_ids` seam; before it, links and loaded rows were always the same set, so no test could express "storage dropped some", which is part of why this path had no coverage.

Reverting the branch fails the test with the old message visible in the assertion output.

4554 passed. `ruff check core-api/src/` and `ruff format --check core-api/src/` clean.

Eldad-Caura:

Two review rounds, both handled in `d798534c`; CI green. The core change is small — gate the `ENTITY_LOOKUP` short-circuit on the entity pool being able to fill `top_k` — and both review findings were about the *logging* around the new decline paths rather than the gate itself.

What to know when you read it:

- The fix has two tiers on purpose. `_collect_memories` bails **before** the storage load when the pool provably can't fill `top_k`; `execute` re-checks **after**, because visibility filtering can still trim an adequate-looking pool. The pre-load exit isn't an optimisation — without it the newly-common path would spend a per-tenant storage permit on rows it discards and leave `_storage_slot_acquired` set, which would have let the real scored search slip past the CAURA-602 bulkhead unmetered.
- Both rounds found the same class of bug one state apart: an empty result is ambiguous (never loaded / loaded and wholly filtered), and I twice claimed a decline-message set was exhaustive when it wasn't. It's now four states with four messages, and there's a test asserting the original "no linked memories" message is still reachable, so the fix removed the inaccuracy rather than relocating it.
- Existing fixtures were updated, not relaxed: several paired a one-memory pool with `top_k` 5 or 10, which is exactly the buggy shape the PR exists to reject. Each now asks for what its fixture links via a `top_k` seam on the ctx builders, so every test keeps its own subject.

4556 passed, 3 skipped, 1 xfailed. `ruff check core-api/src/` and `ruff format --check core-api/src/` clean. Rebased onto current `main` (`2cc741ee`) with the suite re-run after.

Three follow-ups are written up in the PR description rather than folded in — the duplicated `get_memory_ids_by_entity_ids` round-trip, `_expand_per_fleet` silently disabling the boost when it returns `{}`, and the undeclared candidate-row shapes that make the skip guards crash guards. Happy to file any of them as issues if you'd rather they were tracked.

erni-a (approved):

Issue #818, opened 2026-08-18
H-08: Forge cron wires a 3-arg poison checker into the 1-arg seam — every cron cluster distill dies with TypeError


Eldad-Caura:
Finding **H-08** of the 2026-08-14 OSS/platform engineering audit, rated **high**. Filed now so the ten correctness highs are trackable — they previously existed only inside the audit artifact. The eight security highs from the same audit are all fixed and merged (#799, #801, #802, #804, #806).

**Area:** core-api · **Surfaced by:** bugs-core-services

## Verified still present

Checked against `main` at `a6c6547a` on 2026-08-18 before filing:

`cron_handler.py:133` still defines `async def _check(tenant_id, fleet_id, fingerprint)`; `_make_poison_checker()` is called at `:242` and passed as `poison_checker=` at `:257`. The seam is still `PoisonChecker = Callable[[str], Awaitable[bool]]` (`forge_service.py:101`), invoked as `await poison_checker(fingerprint.fp)` (`:571`). The arity mismatch is unambiguous — this raises `TypeError` on every call.

## Mechanism

[...]

PR #833, opened 2026-08-19
fix(forge): give the cron's poison checker the shape the distill seam calls

Eldad-Caura:
Fixes H-08 (#818): a total, silent outage of the production Forge mining path. Premise re-verified on `main` at `2cc741ee` before changing anything.

## The bug

`run_forge_distill` declares `PoisonChecker = Callable[[str], Awaitable[bool]]` and `_distill_cluster` calls `await poison_checker(fingerprint.fp)` — **one** argument. The cron built its checker from `_make_poison_checker()`, whose inner `_check` required **three**: `(tenant_id, fleet_id, fingerprint)`.

So every eligible cluster raised `TypeError` at the poison check — which sits *after* the LLM distill call. `run_forge_distill`'s broad `except Exception` (correctly there so one bad cluster can't abort a tick) counted each as `skipped_io_error`. For any tenant with `skills_factory` enabled: `candidates_written` always 0, the LLM spend paid and thrown away per cluster, the poison cooloff never consulted, and the audit row reporting **success**.

## Why the wrong shape was there

The three-arg form is the gate evaluator's — `skill_lifecycle` calls `await poison_checker(tenant_id, fleet_id, fingerprint)`. But that path is already served: `promote_pending_candidates` is handed `make_db_poison_checker()` from `skill_promoter`, and `cron_handler` already imports it. So `_make_poison_checker`'s only consumer was the one-arg distill seam, and its gate-evaluator shape served nobody — it was a near-duplicate of `make_db_poison_checker` wired into the wrong seam.

## The fix

Close over `tenant_id` / `fleet_id` and take only the fingerprint. That matches the seam, drops the duplication, and mirrors `scripts/forge_dry_run.py`'s `_wire_poison_checker(db, tenant, fleet)` — which is precisely why the dry-run CLI was unaffected while production was dead.

## Why no test caught it, and what now does

`tests/test_forge_cron_tick.py` asserts `candidates_written == 3` — from a `ForgeRunResult` the test fabricates while `run_forge_distill` is stubbed. The cron adapter's entire job is wiring, and the only suite covering it never invoked what the wiring produced. A shape mismatch was therefore invisible everywhere: the service suites inject their own correct callables, and the CLI wires its own.

Two tests, both calling the injectable through the same seam the service uses:

- `_make_poison_checker(...)` accepts one argument and forwards the closed-over identifiers — wrong identifiers would silently consult another tenant's cooloff rows, which is a quieter bug than the `TypeError`.
- **The regression guard:** run the real cron tick, capture the callable actually passed to `run_forge_distill`, and await it with a single fingerprint.

Reverting the fix fails the second with the production error verbatim:

```
TypeError: _make_poison_checker.<locals>._check() missing 2 required positional arguments: 'fleet_id' and 'fingerprint'
```

## Checks

4552 passed, 3 skipped, 1 xfailed. `ruff check` and `ruff format --check` clean on `core-api/src/`. No storage-side change.

I skipped `/simplify` deliberately: this is a signature plus one call site, with no new control flow or abstraction, so it sits below the threshold. The docstring's factual claims (that `make_db_poison_checker` serves the gate path, that the CLI wires the one-arg shape) were each verified against the code rather than asserted.

## Follow-up this surfaced, not in scope

The broad `except Exception` in `run_forge_distill` is what made this silent for months. It's load-bearing — one malformed cluster must not abort a tick — but it buckets programming errors as `skipped_io_error`, so an arity bug is indistinguishable from a storage hiccup in the counters an operator reads. Separating `TypeError`/`AttributeError` from genuine I/O there would turn this class of bug from silent into loud, and is worth its own change.

Refs #818

### DIFF
--- a/core-api/src/core_api/services/forge/cron_handler.py
+++ b/core-api/src/core_api/services/forge/cron_handler.py
@@ -125,12 +125,26 @@ def _make_memory_fetcher(tenant_id: str):
     return _fetch
 
 
-def _make_poison_checker():
+def _make_poison_checker(tenant_id: str, fleet_id: str | None):
     """Adapt :func:`is_fingerprint_poisoned` (storage-backed) to the
-    ``(tenant, fleet, fp) → bool`` shape the gate evaluator expects.
+    ``PoisonChecker`` seam ``run_forge_distill`` actually calls:
+    ``(fingerprint) → bool``, with the tenant and fleet closed over.
+
+    H-08: this used to take ``(tenant_id, fleet_id, fingerprint)`` — the shape
+    ``evaluate_auto_gates`` wants — while its only caller feeds
+    ``run_forge_distill``, which invokes ``await poison_checker(fingerprint.fp)``
+    with one argument. Every cluster therefore raised ``TypeError`` after the LLM
+    distill call, ``run_forge_distill``'s broad ``except Exception`` counted it as
+    ``skipped_io_error``, and the tick reported success having written nothing.
+
+    The gate-evaluator shape was never needed here: ``promote_pending_candidates``
+    is handed :func:`make_db_poison_checker` (skill_promoter), which is the
+    three-arg wrapper for that path. Closing over the identifiers instead mirrors
+    ``scripts/forge_dry_run.py``'s ``_wire_poison_checker``, which is why the
+    dry-run CLI was never affected.
     """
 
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #834, opened 2026-08-19
fix(forge): make a wiring bug fail CI, and stop it reading as storage trouble

Eldad-Caura:
Follow-up to H-08 (#818). That bug merged as a one-line arity fix; this addresses **why it survived ~2.5 months in production**.

## What let it hide

`run_forge_distill`'s cluster loop ends in a broad `except Exception` that counts everything as `candidates_skipped_io_error` and continues. The handler has to stay — one malformed cluster must not abort a tick — but it meant a `TypeError` from a wrongly-shaped injectable was indistinguishable from a storage hiccup in the only counters an operator reads. The tick returned success with zero candidates written, every day, and the counters said storage was flaky.

## Two changes, at different depths

**1. Programming errors get their own bucket.** `_PROGRAMMING_ERRORS = (TypeError, AttributeError)` is caught ahead of the broad handler at all three injected seams and counted as `candidates_skipped_internal_error`. Still `continue` — these aren't *always* deterministic — but they no longer masquerade as I/O.

Scoped deliberately: `KeyError` is excluded (the dict subscripts here are on dicts this module builds itself) and `ValueError` is excluded (validators raise it on legitimately bad *data*, and the LLM-shaped case already has `DistillParseError`). Widening the set would re-blur the line it exists to draw.

Plus a run-level ERROR line when a run wrote **nothing** and hit programming errors. Per-cluster tracebacks already existed during H-08; what was missing was one line saying the run as a whole was broken. Guarded on zero writes so one bad cluster among successes never pages anyone — there's a test for that negative case, because an alert that cries wolf is worse than no alert.

**2. The same class of bug now fails CI.** This matters more, and it came out of reviewing part 1.

`cron_handler`'s five injectable factories had **no return annotations**, so the closures they returned were implicitly `Any` and satisfied any seam. They now carry the alias they must satisfy (`-> PoisonChecker` etc.), matching what sibling `skill_promoter` already did.

Annotations alone were **inert**, though: `[tool.mypy.overrides]` sets `ignore_errors = true` for `core_api.services.*`, so the whole services layer is exempt from type checking — the actual reason H-08 was statically undetectable. So this adds a narrower override re-enabling it for `core_api.services.forge.*`, which is free today: that subpackage is already mypy-clean (0 errors).

Verified with CI's own command (`cd core-api && mypy src/`) rather than in isolation — my first attempt at proving this was a single-file invocation that resolved the alias to `Any` and reported nothing, which is exactly the trap the override creates. Re-introducing the 3-arg checker now gives:

```
src/core_api/services/forge/cron_handler.py:159: error: Incompatible return value type
(got "Callable[[str, str | None, str], Coroutine[Any, Any, bool]]",
 expected "Callable[[str], Awaitable[bool]]")  [return-value]
```

and the fixed shape is clean. **Note this constrains future forge changes to stay type-clean** — a deliberate tightening, scoped to one subpackage.

## Surfacing

The counter reaches all four places the other buckets do, not just one: the service's `forge_run:` INFO summary, the cron stats dict, and both dry-run CLI outputs (JSON + human). A counter that exists only in a dataclass isn't observability.

Corrected while here: the cron stats comment claimed these buckets make "audit rows actionable". They don't — `lifecycle_audit` reduces the whole dict to `candidates_written + promoted` and stores that single int. The buckets reach the structured **log**, which is what an alert can key on. That comment was wrong before this change too, for five buckets.

The new field is deliberately **not** defaulted: a construction site that forgets it would report "no internal errors", the same silent-zero failure the counter exists to end. Better a loud `TypeError` at construction. Three test literals updated accordingly.

## Checks

4562 passed, 3 skipped, 1 xfailed. `ruff check` / `ruff format --check` clean on `core-api/src/`; mypy clean including the newly-checked forge subpackage. Each half verified by reverting it: dropping the except branch fails all four new tests, dropping the summary fails only the summary test.

`/simplify` found real problems in my first draft, worth noting since they shaped the result: the counter had landed in 1 of 4 enumeration surfaces; the dataclass default was justified by naming two callers that don't construct the type; a comment claimed `KeyError` was covered when it wasn't; and the guard carried a dead `eligible and` term. All fixed.

## Not done — the biggest remaining question

A tick that errors on every cluster **still returns success**, so `lifecycle_audit` writes `status="success"`. Worse, that false success arms the 23h dedup gate, so an operator re-running the fanout to reproduce gets `skipped: recent_success` and no run at all — the false success obstructs its own diagnosis. Raising under the same guard would flip the durable verdict, at the cost of a Pub/Sub nack and redelivery. That's a behaviour change rather than an observability one, so it's deliberately out of scope and worth its own decision.

Refs #818

### DIFF
--- a/core-api/pyproject.toml
+++ b/core-api/pyproject.toml
@@ -166,6 +166,10 @@ module = [
 ]
 ignore_errors = true
 
+[[tool.mypy.overrides]]
+module = "core_api.services.forge.*"
+ignore_errors = false
+
 [tool.coverage.run]
 source = ["src"]
 branch = true
--- a/core-api/src/core_api/services/forge/cron_handler.py
+++ b/core-api/src/core_api/services/forge/cron_handler.py
@@ -36,7 +36,12 @@ from typing import Any
 
 from core_api.clients.storage_client import get_storage_client
 from core_api.services.forge.forge_service import (
+    CandidateWriter,
     ForgeConfig,
+    LlmFn,
+    MemoryFetcher,
+    PoisonChecker,
+    StatusChecker,
     run_forge_distill,
 )
 from core_api.services.forge.poison import is_fingerprint_poisoned
@@ -107,7 +112,7 @@ async def _resolve_auto_promote_clean(org_id: str) -> bool:
 # ── Injectable factories ──────────────────────────────────────────
 
 
-def _make_memory_fetcher(tenant_id: str):
+def _make_memory_fetcher(tenant_id: str) -> MemoryFetcher:
     """Bulk-load ``memories.content`` by id via core-storage-api.
 
     Mirrors the dry-run CLI's fetcher. NULL-safe (storage coerces a NULL
@@ -125,7 +130,7 @@ def _make_memory_fetcher(tenant_id: str):
     return _fetch
 
 
-def _make_poison_checker(tenant_id: str, fleet_id: str | None):
+def _make_poison_checker(tenant_id: str, fleet_id: str | None) -> PoisonChecker:
     """Adapt :func:`is_fingerprint_poisoned` (storage-backed) to the
     ``PoisonChecker`` seam ``run_forge_distill`` actually calls:
     ``(fingerprint) → bool``, with the tenant and fleet closed over.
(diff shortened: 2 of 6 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

Issue #811, opened 2026-08-18
H-01: Python/TS SDK recall() reads `supporting_memories`, a key the server never returns — memories always empty


Eldad-Caura:
Finding **H-01** of the 2026-08-14 OSS/platform engineering audit, rated **high**. Filed now so the ten correctness highs are trackable — they previously existed only inside the audit artifact. The eight security highs from the same audit are all fixed and merged (#799, #801, #802, #804, #806).

**Area:** clients/SDK · **Surfaced by:** bugs-plugin-clients, discrepancies

## Verified still present

Checked against `main` at `a6c6547a` on 2026-08-18 before filing:

`caura_client/models.py:55` still does `data.get("supporting_memories")`, and `clients/typescript/src/index.ts:148` reads the same key. A grep of `core-api/src/` for `supporting_memories` returns **nothing** — the server never emits it.

**Path correction:** the audit cites `clients/python/src/memclaw_client/models.py`; post-rename it is `clients/python/src/caura_client/models.py`.

## Mechanism

RecallResult.from_dict extracts memories from data['supporting_memories'], but core-api POST /api/v1/recall returns {'query','summary','memory_count','memories','items','recall_ms'} (recall_service.summarize_memories builds this dict; the route returns it verbatim). A repo-wide grep shows 'supporting_memories' exists ONLY in the two SDKs and their tests — no server code emits it. Failure: any Python SDK user calling mc.recall('q') against a live deployment gets result.supporting_memories == [] on every call, silently losing all supporting memories while summary still works. The SDK test (clients/python/tests/test_client.py:99) mocks the wrong fixture ({'summary': 'S', 'supporting_memories': [...]}), so the suite passes while the live contract is broken.

## Suggested fix

Read data.get('memories') or data.get('items') (keep 'supporting_memories' as a legacy fallback), and fix the test fixtures to mirror the live response shape.

---

*Location cited by the audit: `clients/python/src/memclaw_client/models.py:55`. Description and suggested fix are the audit's own text; the verification section above is a fresh check.*

Eldad-Caura:
Fixed by #835, which landed on `main`. The SDKs now read recall memories from `memories`/`items` rather than the invented `supporting_memories` key.

Closing: this stayed open only because the PR said `Refs` rather than `Fixes`.

PR #356, opened 2026-06-14
feat: add @caura/memclaw-client TypeScript SDK

arkash20:
## What

The TypeScript/JavaScript client for MemClaw, in `clients/typescript/` — a faithful port of the [Python client](https://github.com/caura-ai/caura-memclaw/tree/main/clients/python).

```ts
import { MemClaw } from "@caura/memclaw-client";
const mc = new MemClaw("mc_xxx", { tenantId: "my-team", agentId: "my-agent" });
await mc.write("Q3 revenue target is $4M.");
console.log((await mc.recall("Q3 revenue target")).summary);
```

- `write`/`search`/`recall`/`health`, typed interfaces, typed errors (`AuthError`/`NotFoundError`/`MemClawApiError`), `.raw` on every result.
- **Zero runtime deps** — native `fetch` (Node 18+), ESM, built with `tsc` (repo convention).
- Rich `package.json` metadata; README funnel gains a TypeScript snippet.

## Name

Ships as **`@caura/memclaw-client`**: the `plugin/` package already owns `@caura/memclaw`, and unscoped `memclaw` is taken on npm by an unrelated project. (Parallels PyPI `memclaw-client`.)

## CI / verification

- New **Client (TypeScript) CI** workflow: `npm test` (= `tsc` build + `node --test`) on changes under `clients/typescript/`. Tests mock `fetch` (no network). **Verified locally: 9/9 pass, build clean.**
- **Wet-tested against a live local stack** (core-api): `health()` and `write()` round-tripped successfully (real 201 + memory id); a 503/401 from the server mapped correctly to typed errors. `search`/`recall` reached the endpoint with valid auth but the local stack's embedding service was down (503) — a server/env condition, not a client issue; their response parsing is covered by the unit tests.
- Root CI untouched (root pytest only runs `tests/`).

## Publishing (maintainer action — not done here)

Adds `publish-npm-client.yml` (npm publish on a `memclaw-client-ts-v*` tag or manual dispatch, with provenance). One-time setup (`@caura` scope access + `NPM_TOKEN`) is documented in the workflow header. I did **not** publish.

PR #352, opened 2026-06-14
feat: add memclaw-client Python SDK

arkash20:
## What

A thin Python client SDK for MemClaw, packaged as **`memclaw-client`** for PyPI. Lives in `clients/python/`.

```python
from memclaw_client import MemClaw
mc = MemClaw("mc_xxx", tenant_id="my-team", agent_id="my-agent")
mc.write("Q3 revenue target is $4M.")
print(mc.recall("Q3 revenue target").summary)
```

- `write` → `POST /api/v1/memories`, `search` → `POST /api/v1/search`, `recall` → `POST /api/v1/recall`, `health` → `GET /api/v1/health`.
- Single dependency (`httpx`), typed (ships `py.typed`), context-manager, typed exceptions (`AuthError`/`NotFoundError`/`MemClawAPIError`), `.raw` on every result.
- Rich PyPI metadata (description, keywords, classifiers, project URLs).
- README gains a **Python client** quickstart funnel (`pip install memclaw-client`).

## Why

From the OSS SEO/adoption review (Gap 1 — install funnel): evaluating MemClaw previously meant `git clone` + Docker, with no `pip install` one-liner. Competitors' growth is built on exactly that. This gives the lowest-friction path for talking to a deployment.

## Naming

`memclaw` is already taken on PyPI **and** npm by unrelated projects, so the package ships as **`memclaw-client`** (import `memclaw_client`). The TS/npm client (`@caura/memclaw`) is a planned follow-up.

## CI / verification

- New **Client (Python) CI** workflow runs `ruff` + `pytest` on changes under `clients/python/`. Tests are fully mocked with `httpx.MockTransport` (no network/DB). **Verified locally: 10 passed, ruff clean.**
- The root CI (core services) is untouched — root pytest only runs `tests/`, so this doesn't affect the backend suite.

## Publishing (maintainer action — not done here)

Adds a `publish-python-client.yml` workflow using PyPI **trusted publishing** (OIDC, no token), triggered by a `memclaw-client-v*` tag or manual dispatch. One-time setup (create the PyPI project + trusted-publisher + `pypi` environment) is documented in the workflow header. I did **not** publish — that's a maintainer action requiring the org's PyPI access.

PR #835, opened 2026-08-19
fix(sdk): read recall memories from the key the server actually sends

Eldad-Caura:
Fixes H-01 (#811). Premise re-verified on `main` before changing anything.

## The bug

Both SDKs parsed recall's memories out of `supporting_memories`. **The server never sends that key.** `POST /api/v1/recall` returns `{query, summary, memory_count, memories, items, recall_ms}` — built by `recall_service.summarize_memories` and pinned server-side by `tests/test_c4_recall_items_alias.py::_EXPECTED_TOP_LEVEL_KEYS`.

So every `recall()` against a live deployment returned an **empty memory list** while `summary` kept working — a silent, total loss of the supporting memories, in both the Python and TypeScript clients.

## One correction to the issue's suggested fix

It proposes keeping `supporting_memories` as a legacy fallback. It isn't legacy:

```
git log -S'supporting_memories' --all -- core-api/ core-storage-api/   # → nothing, ever
```

The key first appears in the SDK commits themselves (#352 Python, #356 TypeScript). It was invented client-side and mirrored between the two clients; no server version ever emitted it. A fallback would be dead code that re-implies a contract which never existed — the exact confusion that caused this — so it's dropped rather than kept.

What **is** kept is the field name: `RecallResult.supporting_memories` and TS `supportingMemories` are published API, and renaming them would break callers for no benefit. Only the wire key was ever wrong.

Reads `memories`, falling back to `items` — the server aliases the identical list under both keys deliberately, so consumers written against `/search`'s shape see the same thing.

## Why CI was green through all of this

The tests mocked the invented shape: `json={"summary": "S", "supporting_memories": [...]}`. A hand-written fixture asserting the SDK's own misunderstanding can't catch a contract mismatch — it encodes it. That's the part worth fixing properly, not just the one-line read.

Both suites now build fixtures from a single helper mirroring the real key set, with a comment naming `summarize_memories` and the server-side test that pins it, so the next person editing either client has a source of truth to check against.

Four cases per SDK:
- the live shape;
- `items` alone;
- an empty result staying empty rather than raising;
- a guard asserting a body carrying **only** `supporting_memories` yields no memories — so nobody "fixes" a future report by reinstating the phantom key.

Verified by reverting the one-line read in each client: three of four fail in each, and the guard comes back with `[Memory(id='ghost')] == []`.

## Checks

Python SDK 114 passed, `ruff check src tests` clean. TypeScript SDK 14 passed, `tsc` clean. Both SDK CI workflows are path-filtered to `clients/**`, so both run here. No server-side change, so the main suite is untouched.

`/simplify` skipped deliberately — two one-line wire-key reads plus fixtures, no new logic. The comments' factual claims were verified against history rather than asserted.

## Noticed in passing, not fixed here

`clients/typescript/package-lock.json` on `main` records `version: 0.1.0` while `package.json` says `1.0.0`, so any `npm install` regenerates it. I reverted that incidental churn out of this PR rather than ship an unvalidated lockfile change (it needs Linux-accurate `npm ci` validation). Worth a one-line commit on its own.

Refs #811

### DIFF
--- a/clients/python/src/caura_client/models.py
+++ b/clients/python/src/caura_client/models.py
@@ -52,5 +52,23 @@ class RecallResult:
 
     @classmethod
     def from_dict(cls, data: dict[str, Any]) -> RecallResult:
-        memories = [Memory.from_dict(m) for m in (data.get("supporting_memories") or [])]
+        """Build from a ``POST /api/v1/recall`` body.
+
+        The wire key is ``memories``; the server aliases the identical list under
+        ``items`` as well, for consumers written against ``/search``'s shape, so
+        either is accepted.
+
+        H-01: this used to read ``supporting_memories`` — a key the server has
+        never emitted in any commit. It was invented in this SDK and mirrored into
+        the TypeScript one, so every ``recall()`` returned an empty list while
+        ``summary`` kept working, and the test mocked the invented shape so the
+        suite stayed green against a broken contract.
+
+        The ATTRIBUTE keeps the name ``supporting_memories``: that is published
+        API and renaming it would break callers. Only the wire key was wrong.
+        """
+        raw = data.get("memories")
+        if raw is None:
+            raw = data.get("items")
+        memories = [Memory.from_dict(m) for m in (raw or [])]
         return cls(summary=data.get("summary"), supporting_memories=memories, raw=data)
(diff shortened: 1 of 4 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #837, opened 2026-08-19
fix(lifecycle): a tick broken by a wiring bug must not report success

Eldad-Caura:
Third and last piece of the H-08 (#818) follow-up. #833 fixed the arity bug; #834 made the failure class distinguishable and loud. Both left the **verdict** lying: a tick where every cluster raised `TypeError` returned normally, so the shared lifecycle runner wrote `status="success"` with zero candidates written.

## Why the verdict mattered more than it looks

`has_recent_lifecycle_success` gates re-runs on a recent **SUCCESS**. So the false success also blocked the operator's next attempt for the 23h dedup window — someone re-curling the fanout to reproduce H-08 would have got `{"skipped": true, "reason": "recent_success"}` and no run at all. **The false verdict obstructed its own diagnosis.**

## Why not just raise

The runner reaches `status="failure"` only via an exception, and then re-raises so the bus nacks. That conflates two questions — "did it succeed?" and "should it be retried?" — and for a deterministic wiring bug the honest answers differ. A plain raise gets the audit row right, then buys max-delivery-attempts worth of identical re-runs, each paying for an LLM call per cluster before failing the same way.

So: `PermanentOpError`. The runner records the same `failure` row and **acks**. Named to match `common.ranking.errors.PermanentRankError` — the same idea one layer over — so one word covers one concept and a grep for either finds both.

Three details that came out of review rather than the first draft:

- **One write, not two.** Both failure classes flow through a single guarded audit write with `permanent = isinstance(...)` deciding only the tail. Two near-identical blocks would have drifted on exactly the axis this change exists to make trustworthy.
- **`stats={"terminal": True}` on the terminal row.** Without it the two classes are byte-identical rows, and "nothing more will happen, a human must act" lives only in a log line — unqueryable, unalertable. Mirrors how the dedup gate disambiguates two meanings of `success` via `stats={"skipped": True}`.
- **Acking requires the row to have been written.** If the audit update fails there is no durable record — the entire justification for acking — so that case falls through to the raise and lets redelivery try again. One wasted retry beats a row stuck `in_progress` forever.

## Scope, stated in the docstring

Only the lifecycle runner honours this. Handlers registered directly on the bus (`core_worker.consumer`) get no special treatment — neither bus inspects the exception type — and they own their own audit rows, so they should write `failure` and return instead. Deliberately **not** exported from `common.events`: it had zero consumers there, and exporting it advertised a bus-level contract the bus doesn't implement. Also documented: prefer a plain raise for cheap idempotent ops, where a nack costs nothing and the machinery isn't worth it.

## The guard, and what it doesn't cover

Raised when the mining half wrote nothing **and** hit programming errors. `promote_result.promoted` is deliberately not consulted, so a tick that promoted earlier candidates and then failed to mine still fails — that promotion work survives in the log line but not the audit row. A wiring bug that silences mining is the more important signal, and it's what went unnoticed for months.

Worth flagging: an earlier draft of this PR **claimed a promotion carve-out the code did not implement**. `/simplify` caught it; the comment now describes what the code does. That's the highest-yield review category in this repo and I'd rather note that it happened than quietly fix it.

## Tests

Five verdict cases on the tick (fails on wrote-nothing-plus-internal-errors; not on internal-errors-alongside-a-write; not on a genuinely quiet tick; not on io_errors alone) and four on the runner (failure row written, ack rather than re-raise, no success row so the dedup gate stays open, re-raise when the row couldn't be written). Plus one on the adapter seam the raise crosses — pure wiring, which is precisely what went unwatched in H-08, where the cron tests stubbed it.

Each behaviour verified by reverting it: dropping the ack path re-raises; dropping the stats marker fails the durable-record assertion; dropping `and recorded` acks a failure that was never recorded.

## Checks

4571 passed, 3 skipped, 1 xfailed. `ruff check core-api/src/ common/`, `ruff format --check`, and `mypy src/` all clean.

## Follow-ups this surfaced

- `_OpFn` is `Callable[..., Awaitable[int]]`, so an op has exactly one int of expressiveness and "I failed and a retry won't help" has nowhere to go but the exception channel. A result object (`OpOutcome(count, verdict, detail)`) is the principled shape, but it rewrites seven op closures across two services — not this PR.
- `core_worker.consumer.handle_embed_backfill_request` already hand-rolls "failed, don't retry" for its `fleet_id` guard, because it owns its own audit row. Two shapes for one concept is fine while the reason differs; if a third appears, that's the moment for the `OpOutcome` refactor.
- That same call site writes its audit row unguarded, unlike every other audit write in the codebase. Pre-existing, unrelated, worth a one-liner.

Refs #818

### DIFF
--- a/common/events/base.py
+++ b/common/events/base.py
@@ -49,6 +49,30 @@ class CircularPublishChainError(RuntimeError):
     """
 
 
+class PermanentOpError(RuntimeError):
+    """A lifecycle op failed in a way redelivering it cannot fix.
+
+    Raise this instead of a generic exception when the op can tell the fault is
+    not worth retrying — a wiring or shape bug rather than a flaky dependency.
+    The lifecycle runner records the ``failure`` row exactly as it would for any
+    other exception, marks it ``stats={"terminal": True}``, and then ACKS.
+
+    Named to match :class:`common.ranking.errors.PermanentRankError`, which is the
+    same idea one layer over; keeping one word for one concept means a grep for
+    either finds both.
+
+    SCOPE: only ``lifecycle_handlers``' shared runner honours this. Handlers
+    registered directly on the bus (see ``core_worker.consumer``) get no special
+    treatment — neither bus inspects the exception type, so raising it there nacks
+    like anything else. Such a handler owns its own audit row and should write the
+    failure row and return, rather than raise.
+
+    Prefer a plain raise when the op is cheap and idempotent. Suppressing a retry
+    is only worth it where redelivery costs something real; for archive/purge a
+    nack is nearly free, so the extra retry is better than the extra machinery.
+    """
+
+
 class EventBus(ABC):
     """Abstract event bus. Concrete implementations: `InProcessEventBus`,
     `PubSubEventBus`.
--- a/common/events/lifecycle_handlers.py
+++ b/common/events/lifecycle_handlers.py
@@ -28,7 +28,7 @@ from typing import Protocol
 
 from pydantic import ValidationError
 
-from common.events.base import Event
+from common.events.base import Event, PermanentOpError
 from common.events.factory import get_event_bus
 from common.events.lifecycle_archive_request import (
     LifecycleArchiveRequest,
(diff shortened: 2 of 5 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

Issue #812, opened 2026-08-18
H-02: Embed back-channel reads from the read replica; replica lag ack-drops the sole Path A contradiction trigger


Eldad-Caura:
Finding **H-02** of the 2026-08-14 OSS/platform engineering audit, rated **high**. Filed now so the ten correctness highs are trackable — they previously existed only inside the audit artifact. The eight security highs from the same audit are all fixed and merged (#799, #801, #802, #804, #806).

**Area:** core-api · **Surfaced by:** gap-embed-consumer-seam

## Verified still present

Checked against `main` at `a6c6547a` on 2026-08-18 before filing:

`core-api/src/core_api/consumer.py` still calls `sc.get_memory(str(payload.memory_id))` with no `read=False`, and `CoreStorageClient._get` still defaults to `read=True` (routing to `_read_prefix`/`_read_http` when `CORE_STORAGE_READ_URL` is set). Both ack-drop branches are intact.

## Mechanism

[...]

PR #838, opened 2026-08-19
fix(consumer): route the embed/enrich back-channel read-backs to the writer

Eldad-Caura:
Fixes H-02 (#812). Premise re-verified on `main` before changing anything.

## The bug

`handle_memory_embedded` reads the memory back with `sc.get_memory(...)`, whose `_get` defaults to `read=True` — the **read-replica** pool when `CORE_STORAGE_READ_URL` is set. But this is a read-your-write: the worker PATCHes the embedding and publishes `Topics.Memory.EMBEDDED` immediately after, and Pub/Sub delivery is typically sub-second, so the read races replication.

Under lag it lands in one of two **ack-drop** branches — row not yet replicated (read as "deleted"), or row visible but the embedding PATCH not (read as "should not happen"). Both ack, so the event never redelivers.

On the deferred-embed path this back-channel is the **sole** Path A contradiction trigger, and when a tenant has enrichment disabled no `ENRICHED` event ever fires to cover for it — so the drop is unconditional. Outcome: contradictory memories are never marked outdated, both stale and fresh stay `active`, and recall keeps serving the stale one. Silently.

## Both handlers, not just the one filed

The issue names `handle_memory_embedded`. **`handle_memory_enriched` has the identical pattern** — same unqualified `get_memory`, same ack-drop, same PATCH-then-publish producer — and per `memory_service.py` those two are the *pair* that own deferred-path contradiction detection. Fixing one would have left the other racing, so both are routed.

## The fix

`get_memory` grows `read: bool = True` (default preserved for its eight other callers) and both back-channel reads pass `read=False`.

Not a new idea in this client: the write-path dedup gate already passes `read=False` with a comment citing exactly this staleness. And a writer read cannot miss, because the producer awaits the PATCH — its own committed transaction storage-side — before publishing, as its own comment says ("the embedding is durable in storage either way").

**The comments were the other half of the bug.** Both ack-drop branches asserted causes that were false under lag — "row was deleted" and "should not happen" — so the code documented an impossibility it was hitting routinely. Against the writer those become true, and they now say why.

The missing-embedding branch is also escalated to **ERROR**: its only remaining causes are real invariant violations, and the cost is a silently-missed contradiction check. It still acks — a retry reads the same writer row and reaches the same branch.

## Tests

The existing enriched test already pinned the call shape, so it failed on the change; it now pins `read=False` and carries the reason. Two new cases: the embedded handler routes to the writer, and a missing embedding surfaces at ERROR rather than hiding at warning.

Verified by reverting: dropping `read=False` fails both routing tests; leaving the log at warning fails the escalation test.

## Checks

4573 passed, 3 skipped, 1 xfailed. `ruff check` / `ruff format --check` clean on `core-api/src/`.

`/simplify` skipped deliberately: one keyword argument, two call sites, and comment corrections — no new logic or control flow. The comments' load-bearing claims (producer ordering, per-PATCH commit, and that these two handlers are the only deferred-path triggers) were each read out of the code rather than assumed.

## Noticed, not changed

`contradiction_detector.py` has three more `get_memory` read-backs (lines ~200, ~1964, and a `refreshed = ...` at ~1981) that may be read-your-writes on the write path. I didn't touch them — they're outside this issue's mechanism and each needs its own check of whether a preceding write is involved. Worth a look if someone wants the sweep.

Refs #812

### DIFF
--- a/core-api/src/core_api/clients/storage_client.py
+++ b/core-api/src/core_api/clients/storage_client.py
@@ -467,8 +467,16 @@ class CoreStorageClient:
             _reject_reserved_write_id(item.get("agent_id"))
         return await self._post("/memories/bulk", data)
 
-    async def get_memory(self, memory_id: str) -> dict | None:
-        return await self._get(f"/memories/{memory_id}")
+    async def get_memory(self, memory_id: str, *, read: bool = True) -> dict | None:
+        """Fetch one memory by id.
+
+        ``read=False`` routes to the WRITER. Pass it for a read-your-write — a
+        read-back of a row this request (or an event's producer) just wrote, where
+        replica lag would make the row or the freshly-PATCHed column invisible.
+        Same reasoning as the write-path dedup gate below, which passes it for the
+        same reason.
+        """
+        return await self._get(f"/memories/{memory_id}", read=read)
 
     async def get_memory_for_tenant(self, tenant_id: str, memory_id: str) -> dict | None:
         return await self._get(f"/memories/{memory_id}", tenant_id=tenant_id)
--- a/core-api/src/core_api/consumer.py
+++ b/core-api/src/core_api/consumer.py
@@ -79,11 +79,16 @@ async def handle_memory_enriched(event: Event) -> None:
         return
 
     sc = get_storage_client()
-    memory = await sc.get_memory(str(payload.memory_id))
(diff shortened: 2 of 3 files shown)


Eldad-Caura:
@erni-a pleaer approve

erni-a (approved):

Issue #814, opened 2026-08-18
H-04: Exact-dedup is check-then-insert with no DB uniqueness; duplicates then wedge the gate with MultipleResultsFound


Eldad-Caura:
Finding **H-04** of the 2026-08-14 OSS/platform engineering audit, rated **high**. Filed now so the ten correctness highs are trackable — they previously existed only inside the audit artifact. The eight security highs from the same audit are all fixed and merged (#799, #801, #802, #804, #806).

**Area:** core-api · **Surfaced by:** bugs-write-pipeline

## Verified still present

Checked against `main` at `a6c6547a` on 2026-08-18 before filing:

`ix_memories_content_hash` is created in `001_initial_schema.py:60` **without** `unique=True`, and `memory_find_by_content_hash` still uses `scalar_one_or_none()`. Both halves of the finding stand.

## Mechanism

The 409 exact-duplicate contract is enforced only by a GET (find_by_content_hash) followed later by a plain INSERT (WriteMemoryRow -> memory_add, no ON CONFLICT; ix_memories_content_hash in migration 001 is NON-unique). Two concurrent writes of identical content by the same agent both pass the GET and both insert, producing two live rows with the same (tenant_id, fleet_id, agent_id, content_hash). In strong mode the window between the GET and the INSERT includes the semantic-dup storage roundtrip plus an up-to-10s LLM judge call, making the race easy to hit under fleet retries. Duplicates are also minted with zero concurrency by the auto-chunk path, the atomic-fact fanout (memory_service.py:2482 inserts children with content_hash and no dedup check), and the update path (separate finding). Once >=2 live rows share a hash, storage's memory_find_by_content_hash uses .scalar_one_or_none() (postgres_service.py:916), which raises sqlalchemy MultipleResultsFound -> storage route 500 -> httpx raise_for_status -> the pipeline runner marks the step FAILED -> core-api returns 500 'Memory write pipeline failed unexpectedly'. From then on EVERY write of that content by that agent 500s forever instead of 409ing; there is no code path that heals the state short of deleting a duplicate.

## Suggested fix

Add a partial unique index on (tenant_id, COALESCE(fleet_id,''), agent_id, content_hash) WHERE deleted_at IS NULL and map the insert conflict to 409 in memory_add; change memory_find_by_content_hash to .limit(1) + .first() (or scalars().first()) so pre-existing duplicates degrade to a 409 instead of a 500.

---

[...]

PR #839, opened 2026-08-19
fix(storage): duplicate content hashes must 409, not 500 forever

Eldad-Caura:
First half of H-04 (#814) — the half that stops the outage. Premise re-verified on `main` before changing anything.

## The wedge

Nothing in the schema enforces one live row per `(tenant_id, fleet_id, agent_id, content_hash)` — `ix_memories_content_hash` is non-unique in migration 001 *and* in the ORM model. Once two live rows share a hash, `memory_find_by_content_hash` raised `MultipleResultsFound` → storage 500 → `raise_for_status` → write-pipeline step FAILED → core-api 500 "Memory write pipeline failed unexpectedly".

**Permanently.** From then on every write of that content by that agent 500s instead of 409ing, and no code path heals it short of deleting a row by hand.

## Both lookups, not just the one filed

The issue cites `memory_find_by_content_hash`. **`memory_find_duplicate_hash` — the UPDATE path's dedup gate — ended in the same `scalar_one_or_none()`**, so content updates to a duplicated value 500'd identically. Both are fixed.

## The fix

Return a row instead of raising: ordered oldest-first with a `LIMIT`, so duplicates degrade to the 409 the contract always intended. Oldest because that's the row which legitimately won the contract — and because a *stable* answer beats whatever the planner returns. The previous query had no `ORDER BY` at all, so even the non-duplicate path was only accidentally deterministic.

`memory_find_by_content_hash` takes `LIMIT 2` rather than 1 so it can notice the duplicate and log a warning naming the tenant, agent, hash prefix and kept row. **That log is the instrument this problem lacked**: duplicates were invisible until one wedged a write, and nobody can add the unique index this table should have had without first knowing how many exist.

## Why duplicates exist at all — this isn't a rare race

Not only the check-then-insert race (whose window in strong mode spans a semantic-dup roundtrip *plus* an up-to-10s LLM judge). The **auto-chunk child payloads are built with a `content_hash` and no preceding dedup lookup**, so that path mints duplicates with zero concurrency; same for the atomic-fact fanout. Any deployment that auto-chunks repeated content likely has duplicates already — which is why the wedge needed healing first.

The partial unique index on `(tenant_id, COALESCE(fleet_id,''), agent_id, content_hash) WHERE deleted_at IS NULL`, plus mapping the insert conflict to 409. That's the half that *prevents* new duplicates, and it can't land safely yet:

- migrations run in the storage service's lifespan startup, so a failing `CREATE UNIQUE INDEX` is a **failed deploy** (the old revision keeps serving);
- given the auto-chunk path above, duplicates are likely already present, so it would fail;
- resolving them means mutating production rows under a keep/delete policy nobody has chosen.

The warning this PR adds is what turns that from a guess into a measurement.

## Test

Two live rows with the same hash, then the dedup GET: 200 with the older row rather than a 500. Reverting the fix reproduces the production error exactly — `sqlalchemy.exc.MultipleResultsFound: Multiple rows were found when one or none was required`.

The test also asserts both inserts really landed, so if a unique constraint ever arrives it fails loudly rather than passing vacuously on a dead premise.

## Checks

126 storage tests passed (125 baseline + 1), 4573 core-api passed, 3 skipped, 1 xfailed. `ruff check` and `ruff format --check` clean on `core-storage-api/src/` and `core-storage-api/tests/`.

Refs #814

### DIFF
--- a/core-storage-api/src/core_storage_api/services/postgres_service.py
+++ b/core-storage-api/src/core_storage_api/services/postgres_service.py
@@ -956,6 +956,40 @@ class PostgresService:
     # B) Content hash / dedup
     # ------------------------------------------------------------------
 
+    @staticmethod
+    def _warn_duplicate_content_hash(
+        *,
+        tenant_id: str,
+        fleet_id: str | None,
+        agent_id: str | None,
+        content_hash: str,
+        kept_id: UUID,
+        path: str,
+        excluded_self: bool = False,
+    ) -> None:
+        """Report live rows that violate the uniqueness the schema never enforced.
+
+        H-04: shared by both dedup gates deliberately. They hit the same anomaly,
+        and a warning emitted from only one of them would undercount the duplicate
+        population — which is the number the partial unique index decision rests
+        on. One call site for the message also means the two cannot drift.
+        """
+        logger.warning(
+            "duplicate live rows share a content_hash — dedup returns the oldest",
+            extra={
+                "tenant_id": tenant_id,
+                "fleet_id": fleet_id,
+                "agent_id": agent_id,
+                "content_hash": content_hash[:12],
+                "kept_memory_id": str(kept_id),
+                # Which gate saw it, so the two are separable in a log query.
+                "dedup_path": path,
+                # True → the count excludes the row being updated, so the real
+                # group is one larger than what this gate could see.
+                "excluded_self": excluded_self,
+            },
+        )
+
     async def memory_find_by_content_hash(
         self,
         tenant_id: str,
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@claude — confirmed and fixed in `863dee2a`. Good catch: it lands on a claim this PR made ("stable across calls"), and the tie case is the *common* one here rather than an edge.

**Verified the premise rather than reasoning about it.** `created_at` is `server_default=text("now()")`, Postgres fixes `now()` for the whole transaction, and the auto-chunk path — the zero-concurrency duplicate source this PR cites — inserts all its children in a single `create_memories` call. I inserted three rows in one transaction and checked: **one distinct `created_at` value**. So duplicates minted that way tie by construction, and `ORDER BY created_at` alone left the pick to the plan.

Both lookups now order by `(created_at, id)`. Ordering by `id` isn't chronological, but among rows of the same instant there's no chronology to preserve — only a stable answer to give.

The new test asserts the tie *exists* (fetches each row's `created_at`, requires one distinct value) rather than assuming it, so it can't pass on ordinary chronological ordering and prove nothing.

## One thing I want to be straight about

**The test does not fail when the tie-break is removed.** I checked, expecting it to, and it passed. On a table this small Postgres returns heap order, which is incidentally stable and insertion-ordered, so the unordered form satisfies the assertions too. Instability is plan-dependent and no fixture can force it on demand.

So I'm not claiming "verified by reverting" for this one. The test is an assertion about the intended outcome — ties exist, the answer is stable, and it's the tie-break winner — not proof the previous form was broken. The correctness argument rests on the `ORDER BY` being total. The docstring now says that explicitly, so nobody later reads it as a regression guard it isn't.

I considered a white-box test compiling the statement and asserting the `ORDER BY` has two keys. I left it out: it would pin the fix reliably but break on any refactor of how the statement is built, and it tests the implementation rather than a behaviour. Happy to add it if you'd rather have the guard than the flexibility.

127 storage tests passed (125 baseline + 2), 4573 core-api passed. `ruff check` and `ruff format --check` clean on `core-storage-api/src/` and `core-storage-api/tests/`.

Eldad-Caura:
@claude — fixed in `3d031fd2`. Fair and consistent: the warning went on one of the two gates that wedge, and reporting from one of two **undercounts the duplicate population the unique-index decision rests on** — which is the whole reason the log exists.

`memory_find_duplicate_hash` now takes `LIMIT 2` for the same reason the read path does (one extra row is what makes the anomaly visible) and reports it.

Two things I did beyond the literal ask:

**The message moved into one `_warn_duplicate_content_hash` helper** rather than being copied. Two near-identical log blocks drift, and these two are the only readers of a signal that has to stay *comparable* between them — if one gains a field or changes wording, the counts stop lining up.

**Two fields added:** `dedup_path`, so the gates are separable in a log query; and `excluded_self`, because this gate excludes the row being updated — so `>1` here means at least two *other* rows share the hash, i.e. the real group is one larger than what this gate can see. Without that flag the two gates would report inconsistent-looking counts for the same underlying group, which would be worse than no signal.

**Unlike round 1, this one is verified by reverting:** removing the warning fails the new test with `the update-path gate resolved a duplicate silently`.

## Noticed while testing, not changed

`GET /memories/duplicate-hash` passes only `tenant_id` + `content_hash` + `exclude_id` to the service method, which also supports `fleet_id` and `agent_id`. So over HTTP that gate queries `fleet_id IS NULL` and **cannot see fleet-scoped duplicates at all** — the "legacy tenant+fleet+content scope, which silently collides cross-agent" that the service's own comment warns about.

I found this because my first version of the test created fleet-scoped rows and the route returned `None`. Pre-existing and separate: widening the route changes dedup scope, so it wants its own decision rather than riding along here. The test is written to the shape the route actually supports, with that reason stated inline so nobody reads the missing fleet as sloppiness.

128 storage tests passed (125 baseline + 3), 4573 core-api passed. `ruff check` and `ruff format --check` clean on `core-storage-api/src/` and `core-storage-api/tests/`.

Eldad-Caura:
@erni-a pleaser appaove

erni-a (approved):

Issue #815, opened 2026-08-18
H-05: Entity-link failure after the row commits aborts the pipeline: 500, no background tasks, retries 409 forever


Eldad-Caura:
Finding **H-05** of the 2026-08-14 OSS/platform engineering audit, rated **high**. Filed now so the ten correctness highs are trackable — they previously existed only inside the audit artifact. The eight security highs from the same audit are all fixed and merged (#799, #801, #802, #804, #806).

**Area:** core-api · **Surfaced by:** bugs-write-pipeline

## Verified still present

Checked against `main` at `a6c6547a` on 2026-08-18 before filing:

`write_memory_row.py` still calls `sc.create_memory(...)` and then loops `sc.create_entity_link(...)` with **no** `try`/`except` between them.

## Mechanism

WriteMemoryRow first commits the memory row via sc.create_memory (line 80), then loops sc.create_entity_link with NO error handling (lines 84-95). entity_links are caller-supplied UUIDs with no upstream existence validation (routes/memories.py write path has none), so a nonexistent entity_id triggers an FK violation -> storage 500 -> httpx.HTTPStatusError; any transient storage error does the same. The runner catches it, marks the pipeline FAILED and breaks BEFORE ScheduleBackgroundTasks runs, and _create_memory_pipeline raises 500 (memory_service.py:404-406). Net effect: (1) the caller gets a 500 for a write that was actually persisted; (2) in deferred-embedding deployments (fast mode, deployment_mode=deferred) the row has embedding=NULL and _schedule_embed_or_reembed is never scheduled, so the memory is permanently invisible to semantic search; (3) enrichment/entity-extraction backfill is also never scheduled; (4) a client retry of the same content now hits CheckExactDuplicate and gets 409 'Duplicate memory exists: <id>' pointing at the degraded orphan, so the caller can neither re-write nor repair it.

## Suggested fix

Wrap the entity-link loop (and anything after create_memory) in try/except that logs and degrades (persist the write result, record link failures in metadata), or validate entity_ids before create_memory; at minimum still run ScheduleBackgroundTasks when the row was committed so the embed/enrich backfill fires.

---

*Location cited by the audit: `core-api/src/core_api/pipeline/steps/write/write_memory_row.py:84`. Description and suggested fix are the audit's own text; the verification section above is a fresh check.*

erni-a:
[...]

PR #840, opened 2026-08-20
fix(write): a committed row must not be abandoned by a failed entity link

Eldad-Caura:
Fixes H-05 (#815). Premise re-verified on `main` before changing anything.

## The bug

`WriteMemoryRow` commits the memory row, then wrote entity links with no error handling. `entity_links` are caller-supplied UUIDs with no upstream existence check, so one nonexistent id is an FK violation → storage 500 → the runner marked the step FAILED and broke **before** `ScheduleBackgroundTasks`:

- the caller got a **500 for a write that had persisted**;
- the embed and enrichment backfill were never scheduled — in a deferred-embed deployment the row kept `embedding=NULL`, permanently invisible to semantic search;
- a retry hit `CheckExactDuplicate` and **409'd pointing at that same degraded row**.

Neither writable nor repairable.

## The fix

Per-link `try/except` that logs ERROR and continues. Per-link rather than one `try` around the loop — one bad id must not discard the valid links beside it — and that's sound because `entity_add_entity_link` opens its own session per call, so an `IntegrityError` on link 1 doesn't poison link 2's transaction.

ERROR rather than the WARNING the audit hook below uses: a dropped link is user-visible data loss, and a link that was never created leaves **no row for `GET /entities/broken-links` to find afterwards**. The log is the only trace, which is also why the count now rides the existing `memory_write_latency` line beside the `*_pending` flags — `success` alone can't show it, because the write succeeded.

## The response had to change too

`_memory_to_out` was echoing the **requested** links. Degrading without touching that would have reported links that don't exist — a quieter wrong answer than the 500 it replaces.

Both callers of the persist pipeline now go through one `_memory_out_with_created_links` helper. **Both deliberately:** the auto-chunk fall-through (0–1 facts extracted) runs the same pipeline and returns separately, so fixing only the main path would have left one caller lying. A shared helper means a third caller can't drift the same way.

## Tests

Four, each failing without the fix with the real `ForeignKeyViolationError` rather than a mocked raise:

1. ghost link → 201 not 500, response reports the link as absent;
2. a good link still reported — guards the truthful echo from *under*-reporting, which would trade a loud failure for a quiet one;
3. **ghost first, then a real link → the valid one survives.** The only test that pins the per-link design; a refactor to one `try`, or to a batch call that aborts on first violation, passes everything else;
4. `ScheduleBackgroundTasks` still runs — the actual harm.

Worth noting how (4) ended up: two earlier versions asserted on `_schedule_embed_or_reembed` and **passed vacuously**. That branch is gated on `embedding is None`, so it never fires in an inline-embedding environment, and forcing deferred mode still missed because `defer_embedding` additionally requires no cached embedding. So it now asserts the **step ran** — "the step after a committed write runs" is the invariant, and it holds in every configuration.

## Checks

4577 passed, 3 skipped, 1 xfailed. `ruff check` / `ruff format --check` clean on `core-api/src/`.

## Two follow-ups worth filing, both surfaced by `/simplify`

**`entity_bulk_upsert_links` already does this, server-side, in one roundtrip.** I verified it: it opens a session per item *precisely* so "an FK violation on item N would otherwise roll back items 0..N-1", and returns a per-slot error — same semantics as this loop, plus idempotent and N→1 roundtrips. Adopting it would put the degradation logic in the layer that owns and tests it. Not done here because it needs the router's 500-item cap handled against an unbounded `entity_links`, and it pairs naturally with the next item.

**The bulk and legacy write paths may discard `entity_links` entirely.** Both pass them *inside* the `create_memory` payload, and `_filter_memory_fields` keeps only real columns — `Memory` has no `entity_links` column — while still echoing them to the caller. Same genre as this bug and strictly quieter. I read the path but did **not** prove it end-to-end, so it wants its own issue and test rather than a claim here.

Refs #815

### DIFF
--- a/core-api/src/core_api/pipeline/steps/write/write_memory_row.py
+++ b/core-api/src/core_api/pipeline/steps/write/write_memory_row.py
@@ -12,6 +12,11 @@ from core_api.services.hooks import get_hooks
 
 logger = logging.getLogger(__name__)
 
+#: Per-link ERROR lines emitted per request before falling back to the single
+#: summary below. ``entity_links`` has no schema bound, so this is what keeps
+#: log volume from tracking caller input.
+_MAX_LINK_ERROR_LOGS = 5
+
 
 class WriteMemoryRow:
     @property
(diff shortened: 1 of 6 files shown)


Eldad-Caura:
@claude — the finding is right and it's fixed in `8b759b64`, but the remedy needed one correction and it turned out to require a change a layer down.

**Why not narrow the catch.** Letting other exception types propagate would **reintroduce H-05**: a transient storage 5xx would abort the pipeline and strand the committed row again, which is precisely the bug this PR exists to fix. So the split belongs in the *classification*, not the control flow — degrade always, report differently.

**Why that wasn't possible in core-api alone.** An FK violation from a caller-supplied id escaped `entity_add_entity_link` as a bare `IntegrityError` → **500**. So core-api saw `HTTPStatusError(500)` for *both* a bad id and a dead storage service. No core-api-side narrowing could have separated them — the information wasn't there.

So `POST /entities/links` now maps it to **409**, using the same `ValueError` → 409 convention `create_entity` already uses in that router. That's the right layer regardless: a nonexistent `memory_id`/`entity_id` is a client error and should never have been a 500. The composite-PK case (re-linking an existing pair) lands there too, which is a conflict rather than a server fault.

With that, core-api classifies **4xx as permanent** (caller's input, one link) and **everything else — 5xx, connect, timeout — as transient** (ours to chase). Both still degrade. The `permanent` flag rides both the per-link log and the summary, so the outage case you were worried about — links endpoint down, every link dropped, 201 returned — is now a distinct, alertable signal rather than N identical data-loss events.

**Log volume, your second suggestion:** taken. `entity_links` has no schema bound, so a caller sending a thousand bad ids would have emitted a thousand ERROR lines — log volume tracking caller input is its own denial of observability. Capped at 5 per-link lines, plus one summary that fires when the cap truncates **or when any failure was transient**, so the cap can never be the thing that hides an outage.

**Not taken — surfacing failures in the response body.** It's a real gap, and I'd flagged it myself. But every existing "not fully settled" signal here is an advisory `metadata` key, and those are stamped *before* the insert, while link failures are only known *after*. Doing it in-idiom means either a second roundtrip or response metadata that diverges from what's stored — a contract decision rather than a review fix. Happy to do it if you want the field.

Two new tests, each verified by reverting:
- a bad id classifies `permanent=True` with `status_code=409` — reverting the route mapping fails it with `assert False is True`;
- a simulated 503 across 7 links classifies transient, caps at 5 per-link logs, and still emits the summary with `transient=7` — reverting the cap fails it.

4579 core-api passed, 128 storage. `ruff check` / `ruff format --check` clean across `core-api/src/`, `core-storage-api/src/` and `core-storage-api/tests/`.

Eldad-Caura:
Round 2 finding accepted and fixed in `8cb9da62`.

**It was real, and the test output proves it.** `exc.orig` went into the `ValueError`, and `create_memory_entity_link` turns `str(e)` into the 409 `detail` — so a caller who guessed a bad `entity_id` got back the constraint name, the table name, and their own id echoed out of the driver:

```
entity link rejected: ... Key (entity_id)=(0dcd377e-...) is not present in table "entities".
```

The raw text is now logged server-side (where it is actually useful for diagnosis) and the caller gets a fixed sentence naming the two possible causes. That is the same line `entity_add` already drew a few hundred lines above, so this is one convention rather than two.

**Third new test, verified by reverting:** `test_link_to_ghost_entity_409s_without_leaking_driver_text` asserts the 409 body contains neither the offending uuid, nor `violates`, nor `memory_entity_links`. With the sanitisation reverted it fails on the uuid assertion, with the driver's `is not present in table "entities"` visible in the response — the failure output above is from that control run.

Gates: **4579** core-api passed / 3 skipped / 1 xfailed, **129** storage (128 + the new test). `ruff check` and `ruff format --check` both clean across `core-api/src`, `core-storage-api/src`, `core-storage-api/tests` — run as separate gates; only the two touched files were formatted.

Eldad-Caura:
@erni-a — ready for your review. Two rounds of Claude review are handled and CI is green on `8cb9da62`.

**What this fixes (H-05, #815):** `WriteMemoryRow` committed the memory row and then wrote entity links with no error handling. `entity_links` are caller-supplied UUIDs with no upstream existence check, so one nonexistent id made storage 500 and broke the pipeline *before* `ScheduleBackgroundTasks` — the caller got a 500 for a write that had persisted, embed/enrichment backfill never ran (in a deferred-embed deployment the row stayed `embedding=NULL` and invisible to semantic search), and a retry 409'd against that same degraded row. Neither writable nor repairable.

**Round 1** (Medium — bare `except Exception` too broad): the catch stays broad on purpose, because letting other exception types propagate would reintroduce the very bug this PR fixes. The split moved into the *classification* instead — 4xx is the caller's input, everything else is ours — which required `POST /entities/links` to start mapping an FK violation to 409, since core-api previously saw `HTTPStatusError(500)` for both a bad id and a dead storage service.

**Round 2** (Low — raw DB text in the response): the 409 `detail` was quoting the driver, handing callers the constraint name, the table name and their own id. Raw text now goes to the log; the caller gets a fixed sentence.

**Gates:** 4579 core-api passed / 3 skipped / 1 xfailed; 129 storage. `ruff check` and `ruff format --check` clean as separate gates. DCO signed.

Three new tests, each confirmed to fail with its own fix reverted — including the round-2 one, where the control run showed the driver's `is not present in table "entities"` in the response body.

Not merging — that's yours.

erni-a (approved):

PR #841, opened 2026-08-20
fix(write): the server-internal write paths must consult a dedup lookup

Eldad-Caura:
Refs #814 — this is the **first half**, and #814 stays open deliberately: the unique index is the second half and lands separately (see *Why this must land first*).

## The bug

Three write paths attach a `content_hash` to every row they insert and then never look it up:

| path | function |
|---|---|
| auto-chunk children, pipeline handler | `_handle_auto_chunk_from_ctx` |
| auto-chunk children, legacy handler | `_create_memory_legacy` |
| atomic-fact fanout | `_enrich_memory_background` |

The **public bulk path already dedups** — `existing_hashes` for the cross-batch case, `seen_hashes` for the in-batch one — and the single-write path has `CheckExactDuplicate`. These three had neither, so the same document re-chunked, or an LLM emitting the same fact twice, minted a fresh row every time.

That is why prod carries duplicates **with no concurrency involved at all**: `18 duplicate groups / 19 surplus rows / 110,057 live rows`, measured read-only against prod.

`_auto_chunk_request_id()` is not a substitute: it mints a fresh UUID per item per call, so `ix_memories_attempt_unique` sees every re-run as a brand-new attempt. It makes a *replayed* batch idempotent, which these callers never do.

## The fix

`_live_duplicate_hashes` + `_drop_duplicate_facts`, applied at all three sites, reusing the existing `bulk_find_by_content_hashes` primitive. Scope is `(tenant, fleet, agent, content_hash)` over live rows — the same scope `memory_find_by_content_hash` uses, so a child is judged by the rule the single-write gate would apply to it.

Two kinds of duplicate, handled separately because only one is something a constraint could ever catch:

- **already live** — content exists from an earlier call;
- **repeated inside this batch** — both rows are in the *same* insert, so no unique index can collapse them. They must be dropped before the statement runs.

## Where it sits, and why that matters

On the auto-chunk paths the dedup runs on the **facts**, inside the `len(facts) > 1` branch, **before the parent insert and before the batch embed**:

- `child_count` in the parent's metadata counts children that *will exist*, not chunks the chunker proposed. Deduping after the parent insert would have made that field permanently wrong with nothing to correct it.
- **a dropped child never costs an embedding.** The embed is the expensive part of this path and it is a batch call, so paying to vectorise text we're about to discard would have been the dominant cost of deduping at all.
- inside the branch, not above it: dedup can reduce the count to 1 or 0, and falling out of the branch on that would write the whole document as a single memory — a different row than either outcome the caller asked for.

The fanout skips at the top of its loop for the same embed reason. Its skip is deliberately *unlike* the two `continue`s inside its embed block: those exit **after** deciding `child_embedding` so a failed embed still persists the fact; this one decides the fact should not be persisted at all.

## Also: the H-04 sibling in the batch lookup

`memory_bulk_find_by_content_hashes` keeps one row per hash via a dict comprehension, **unordered** — so on a pre-existing duplicate group the winner was whatever the plan emitted last. That could disagree with `memory_find_by_content_hash`, which returns the oldest after #839, leaving two dedup paths pointing callers at different rows for the same content.

Now ordered `(created_at, id)` ascending, first-wins. `id` breaks the tie because ties are the **common** case here, not an edge: `created_at` is `server_default=now()`, fixed for a whole transaction, and auto-chunk inserts all its children in one call.

## Scope: what this does NOT close

**This PR does not eliminate duplicate content-hash groups in prod.** The gate is check-then-insert, so two concurrent runs can both pass it and both insert — two overlapping redeliveries enriching the same parent, or two racing auto-chunk requests for the same document. No application-level check can prevent that; only a constraint can.

What it removes is the source that needs **no concurrency at all** to fire, which is the source prod's 18 groups actually came from. The unique index in #814's second half closes the race. Read a green here as *"duplicates now require a race"*, not *"duplicates are impossible"*.

## Why this must land first

`memory_add_all`'s `ON CONFLICT` target is pinned to `ix_memories_attempt_unique`, and a statement can only name one. So a content-hash violation would **not** be swallowed — it would raise `IntegrityError` and abort the **whole batch**.

Adding the unique index first would therefore convert today's silent duplication into a 500 on a live path: both auto-chunk handlers batch-insert their children, and the fanout would 409 per child. Hence two PRs, minting paths first.

## Tests

21 new, **each confirmed by reverting the specific thing it covers**:

- `_drop_duplicate_facts` — live drop, in-batch collapse, first-occurrence survives, unhashable content kept, all-dropped returns empty, tenant/fleet scoping. Removing either branch fails exactly its own two.
- **Wiring at all three sites independently.** Removing site 1's call fails 5 tests, site 2's fails 2, the fanout's fails 2. Site 2 gets its own harness rather than being assumed to match site 1 — on two of this audit's last three findings the filed location was only half the problem.
- `child_count == 1` for three chunks of which two are live, and `embedded == [["fresh"]]` — the two assertions that pin the placement.
- The batch lookup returns the **oldest** of a duplicate group, plus the new warning fires on a duplicate group and stays quiet on a healthy lookup.

  The ordering test had to be **rewritten, because my first version of it was a bad test.** It inserted the group oldest-first — which meant the *unordered* path returned the oldest by accident, since the resolver keeps the first row it reaches and heap order is insertion order. Measured: **5 of 5 runs passed with the `order_by` reverted**; an earlier single observed failure had just been luck.

  It now ages the rows in reverse of write order, so heap order and `created_at` order disagree and the unordered path reaches the *newest* row first, and it spans 10 independent groups rather than one — one group can flip on a plan change, ten cannot agree by chance. Measured: passes 3/3 with the fix, fails 5/5 without.

One existing test needed a mock — `test_embed_stability` used a bare `MagicMock` storage client and the fanout now makes one more call on it.

## Gates

- **4597** core-api passed / 3 skipped / 1 xfailed (4579 baseline + 18)
- **132** storage passed (129 baseline + 3)
- DCO signed

## Noted while here, not fixed

`memories.metadata` is `json` in the live schema (migration 001) while `common/models/memory.py` declares `JSONB`, so the column type differs between the migration-chain schema and any schema built from that metadata. Found while writing the second half; it needs its own issue rather than a drive-by cast.

### DIFF
--- a/core-api/src/core_api/services/memory_service.py
+++ b/core-api/src/core_api/services/memory_service.py
@@ -130,6 +130,131 @@ def _auto_chunk_request_id() -> str:
     return f"auto-chunk:{uuid4()}"
 
 
+async def _live_duplicate_hashes(
+    sc,
+    *,
+    tenant_id: str,
+    fleet_id: str | None,
+    agent_id: str,
+    hashes: list[str],
+) -> set[str]:
+    """Which of ``hashes`` already have a LIVE row, in dedup scope.
+
+    The server-internal write paths — auto-chunk children (both the pipeline
+    and legacy handlers) and the atomic-fact fanout — attach a ``content_hash``
+    to every child and then insert it without ever consulting a dedup lookup.
+    The public bulk path does consult one (``existing_hashes`` +
+    ``seen_hashes`` in ``create_memories_bulk``), and the single-write path has
+    ``CheckExactDuplicate``; these three had neither. That is why prod carries
+    duplicate content-hash groups with no concurrency involved at all: the same
+    document re-chunked, or an LLM emitting the same fact twice, minted a fresh
+    row every time.
+
+    ``_auto_chunk_request_id()`` cannot substitute for this. It mints a fresh
+    UUID per item per call, so the attempt-idempotency index
+    (``ix_memories_attempt_unique``) sees every re-run as a brand-new attempt —
+    it makes a *retried* batch idempotent only when the caller replays the same
+    ids, which these callers never do.
+
+    Scope is ``(tenant, fleet, agent, content_hash)`` over live rows, matching
+    ``memory_find_by_content_hash`` exactly, so a child is judged a duplicate by
+    the same rule the single-write gate would apply to it.
+
+    WHAT THIS DOES NOT CLOSE. It is check-then-insert, so two concurrent runs can
+    both pass the check and both insert — two overlapping redeliveries enriching
+    the same parent, or two racing auto-chunk requests for the same document.
+    Nothing here can prevent that; only a unique constraint can, and adding one
+    is the second half of #814. So this closes the source that needs NO
+    concurrency to fire — which is the source prod's duplicate groups actually
+    came from — and leaves the race for the index. Do not read a green here as
+    "duplicates are impossible"; read it as "duplicates now require a race".
+    """
+    if not hashes:
+        return set()
+    existing = await sc.bulk_find_by_content_hashes(
+        tenant_id,
+        hashes,
+        fleet_id=fleet_id,
+        agent_id=agent_id,
+    )
+    return set(existing)
+
+
+def _drop_duplicate_facts(
+    facts: list[dict],
+    *,
+    tenant_id: str,
+    fleet_id: str | None,
+    live_hashes: set[str],
+    source: str,
+) -> list[dict]:
+    """Return the chunker's facts minus the ones already recorded.
+
+    Drops two kinds of duplicate, which have to be handled separately because
+    no index can collapse the first:
+
+    * **already live** — the content exists from an earlier call, per
+      ``live_hashes``.
+    * **repeated within this batch** — two children of the SAME call carrying
+      identical content. A unique index cannot resolve this one: the conflict is
+      between two rows of a single INSERT, so it has to be collapsed before the
+      statement runs.
+
+    Applied to the FACTS, before the parent insert and before the batch embed,
+    rather than to the finished child payloads. Two things fall out of that
+    ordering, and both matter more than the slightly earlier call site:
+
+    * the parent's ``child_count`` is the number of children that will exist,
+      not the number the chunker proposed — otherwise this change would make
+      that field silently wrong;
+    * a dropped child never costs an embedding. The embed is the expensive part
+      of this path, and it is a batch call, so paying for text we are about to
+      discard would be the dominant cost of deduping at all.
+
+    Dropping rather than collapsing-and-reporting is deliberate: these children
+    are derived rows, not caller-submitted items. Nothing upstream holds an
+    index into them — both auto-chunk handlers discard ``create_memories``'
+    return value entirely — so there is no per-item result to point at a
+    survivor, and the parent memory is what the caller gets back either way.
+    """
+    kept: list[dict] = []
+    seen: set[str] = set()
+    dropped_live = 0
+    dropped_repeat = 0
+    for fact in facts:
+        content = fact.get("content")
+        if not content:
+            # Outside the dedup contract: an unhashable fact cannot collide.
+            kept.append(fact)
+            continue
+        content_hash = _content_hash(tenant_id, fleet_id, content)
+        if content_hash in live_hashes:
+            dropped_live += 1
+            continue
+        if content_hash in seen:
+            dropped_repeat += 1
+            continue
+        seen.add(content_hash)
+        kept.append(fact)
+
+    if dropped_live or dropped_repeat:
+        # INFO, not WARNING: a re-chunked document hitting this is the feature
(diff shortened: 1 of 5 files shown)


Eldad-Caura:
Both round-1 findings accepted and applied.

**1 — the batch lookup resolved duplicate groups silently.** More than a consistency nit, and the shared helper says why itself:

> shared by both dedup gates deliberately. They hit the same anomaly, and a warning emitted from only one of them would **undercount the duplicate population** — which is the number the partial unique index decision rests on.

That count is exactly what the second half of #814 is sized against, so every duplicate group visible only through this path was missing from it. Worth noting the ordering fix is what made the warning possible at all: before it, this path did not know *which* row it kept, so it had nothing honest to report.

It now emits the same warning with `dedup_path="bulk_find_by_content_hashes"`, keeping the two gates separable in a log query.

Two tests, not one: a duplicate group warns, **and a healthy lookup does not** — a warning that fired on every lookup would make the signal useless for counting anything. Removing the call fails the first.

**2 — check-then-insert does not close the race.** Correct, and no code change can fix it here; only a constraint can. So it is now stated where an engineer will actually hit it, in `_live_duplicate_hashes`' docstring, in the terms on-call needs:

> this closes the source that needs NO concurrency to fire — which is the source prod's duplicate groups actually came from — and leaves the race for the index. Do not read a green here as "duplicates are impossible"; read it as "duplicates now require a race".

I've also made that explicit in the PR description. To be unambiguous about the split: **this PR does not eliminate duplicate content-hash groups in prod.** It removes the zero-concurrency source. The unique index in #814's second half is what makes them impossible, and #814 stays open for it.

**Gates:** 4597 core-api passed / 3 skipped / 1 xfailed, 132 storage (129 baseline + 3). mypy clean. `ruff check` and `ruff format --check` clean as separate gates.

Eldad-Caura:
One correction to my own comment above, before this goes to review again — pushed in `7a04391a`.

**The ordering test I described as verified was not a real guard.** I re-ran its control five times and it passed 5/5 with the `order_by` reverted. The single failure I'd observed earlier was luck, not a signal.

The cause is my own refactor: switching from a dict comprehension (*last* wins) to an explicit loop (*first* wins) means that with the group inserted oldest-first, the unordered path already reaches the oldest row first — so it returns the right answer for the wrong reason, and reverting the fix changed nothing.

Rewritten: the rows are now aged in reverse of write order, so heap order and `created_at` order disagree and the unordered path reaches the **newest** row first — the wrong answer. And across 10 independent groups rather than one, because a single tied group can flip on a plan change while ten cannot agree by chance. Measured: **passes 3/3 with the fix, fails 5/5 without.**

This is the same failure mode as the one I hit on the migration side of #814 an hour earlier, where a 4-row tie-break test only failed 1 run in 3. Worth stating for anyone reading these as a pattern: on this table, "insert rows in an order that makes the wrong answer look right" is the default outcome, not the exception.

Suites re-run on the rewrite: **4597** core-api, **132** storage. Both ruff gates and mypy clean.

Eldad-Caura:
@erni-a — ready for your review. Round-1 review handled and CI green on `7a04391a`.

**What it fixes:** three server-internal write paths — auto-chunk children on both handlers, and the atomic-fact fanout — attached a `content_hash` to every row and never looked it up. The public bulk path already dedups (`existing_hashes` + `seen_hashes`) and the single-write path has `CheckExactDuplicate`; these three had neither. That is why prod carries `18 duplicate groups / 19 surplus rows` over 110k live rows **with no concurrency involved at all**.

**Scope, stated plainly:** this does *not* make duplicates impossible. It is check-then-insert, so a race can still produce one. It removes the source that needs no race at all — the one prod's groups came from. The unique index in #814's second half closes the remainder, and #814 stays open for it.

**Two things changed since the last automated review**, both from its findings:
1. `memory_bulk_find_by_content_hashes` now emits the same duplicate-group warning the single-row gate does. Not cosmetic — the shared helper's docstring says a warning from only one gate *undercounts the duplicate population*, and that count is what sizes the index decision.
2. The concurrency limitation is now documented in `_live_duplicate_hashes` and in the PR description.

**And one correction I made on my own initiative:** the ordering test I had claimed was verified turned out not to be a real guard — it passed 5/5 with the fix reverted, because my refactor made the resolver first-wins and the group was inserted oldest-first, so the unordered path returned the right answer by accident. Rewritten to age rows in reverse of write order across 10 groups; now passes 3/3 with the fix and fails 5/5 without. Details in the comment above.

**Gates:** 4597 core-api passed / 3 skipped / 1 xfailed; 132 storage. mypy clean (199 files). `ruff check` and `ruff format --check` clean as separate gates. DCO signed. All CI checks green (`claude-review: SKIPPED` is expected — the review job skips on `synchronize`).

21 new tests, each confirmed to fail with its own fix reverted.

Not merging — that's yours.

erni-a (approved):

PR #842, opened 2026-08-20
feat(storage): one live memory per (tenant, fleet, agent, content_hash)

### LINKED ISSUE #814: H-04: Exact-dedup is check-then-insert with no DB uniqueness; duplicates then wedge the gate with MultipleResultsFound

Eldad-Caura:
Fixes #814 — the second half, on top of #841. #841 closed the paths that minted duplicates with **no concurrency at all**; this adds the constraint that makes them impossible, and turns the resulting rejection into the 409 the dedup contract already promises.

## Why a constraint is needed at all

`ix_memories_content_hash` is **non-unique** and covers only `(tenant_id, content_hash)`, so nothing in the schema has ever enforced the dedup contract the write path advertises. #839 stopped the wedge — two live rows sharing a hash made `scalar_one_or_none()` raise `MultipleResultsFound` → storage 500 → every subsequent write of that content 500ing forever — but it did so by *tolerating* duplicates. #841 removed the source that needed no race. What remains is the race, and only a constraint closes it: every application-side gate here is check-then-insert.

Measured against prod **before** writing the migration, because a unique index over existing data either applies or fails the deploy — there is no third outcome:

```
duplicate_groups: 18   surplus_rows: 19   worst_group: 3   live_rows: 110,057
```

The `GROUP BY` in that query is character-for-character the key below, so the count describes *this* index rather than an approximation of it.

## Migration 040

Cleanup first in the normal transaction, then the index in an `autocommit_block`. Migrations run in the FastAPI lifespan startup hook, so a failed one is a failed deploy — Cloud Run keeps the old revision serving. Same ordering rationale as 038.

- **The survivor is the oldest per group**, ordered `(created_at, id)` — exactly what `memory_find_by_content_hash` returns after #839, so the row kept is the row the dedup gate would already have handed callers. Anything else silently re-points live lookups at a different row.
- Surplus rows are **soft**-deleted and stamped `metadata.deduped_by_migration`, so all 19 stay findable: `SELECT id FROM memories WHERE metadata ->> 'deduped_by_migration' = '040'`.
- `CREATE UNIQUE INDEX CONCURRENTLY` inside `autocommit_block`, with `_drop_invalid` first. For a *unique* index the invalid-index hazard is worse than 035's measured 547× slower plan: **an invalid unique index does not enforce uniqueness either**, so without that step the migration stamps green while the constraint it exists to add is silently absent.
- The `CREATE ... INDEX ... ON memories` prefix is on **one line with the index name spelled literally**, so `test_no_plain_create_index_on_large_tables` actually covers it — its `\w+` for the name matches neither a split prefix nor an interpolated constant, which is how 007 and 026 ended up outside its coverage while doing the right thing. Verified: removing `CONCURRENTLY` fails that test.

## Two things the migration got wrong until a test caught them

Both found by writing the cleanup test — which is the point of having written it. The cleanup had gone green **twice** against local databases having soft-deleted *zero* rows, because neither had a duplicate to find.

1. **`COALESCE could not convert type jsonb to json`.** The live `metadata` column is `json` (migration 001) while `common/models/memory.py` declares `JSONB`, so the type differs between the migration-chain schema and any schema built from that metadata. Casting the *column* (not the result) makes one statement correct against both.
2. **JSON `null` is not SQL NULL.** `COALESCE` passes it straight through, and `'null'::jsonb || '{...}'::jsonb` does not fail — it **wraps both in an array**, which drops the marker (`->> 'key'` on an array is NULL) and replaces the row's metadata with `[null, {...}]`. Reachable: SQLAlchemy's JSON type stores a Python `None` as JSON `null` unless the column sets `none_as_null`, which this one does not. Now guarded with `jsonb_typeof(...) = 'object'`.

## The rejection must be a 409, not a 500

`memory_add` was a bare `add` + `flush`. With the index in place a duplicate insert raises, and unhandled that is a 500 — the caller reads "Memory write pipeline failed unexpectedly" for an ordinary race, with nothing said about which row now owns the content.

| layer | behaviour |
|---|---|
| storage service | `DuplicateContentHashError`, re-SELECTing the winner so the message names it. Matched on the **index name**, not `IntegrityError` broadly — `ix_memories_attempt_unique` can fire here too and means something different. |
| `POST /memories` | maps exactly that to 409 |
| storage client | raises a typed `DuplicateMemoryError`, **not** an `HTTPException` — nothing else in `core_api/clients/` raises a FastAPI type, and the app-level `httpx.HTTPStatusError` handler documents why it can't do this job either (it deliberately re-raises 4xx as "a bug in OUR request shape", which a legitimate duplicate is not) |
| `WriteMemoryRow`, `_create_memory_or_409` | translate to 409, matching `CheckExactDuplicate` |
| atomic-fact fanout | deliberately does **not** translate — it runs in a fire-and-forget task with no HTTP contract, so a 409 would go nowhere and would abort the facts after it. A refusal there means the fact is already recorded, which is the outcome it wants |

## The end-to-end test that caught a bug the mocked ones could not

`POST /memories` twice with the same hash, against the real index and a real session.

The first version of `memory_add` rolled back and re-SELECTed on the **same** session, and `get_session` holds an explicit `session.begin()` — so the lookup raised:

```
InvalidRequestError: Can't operate on closed transaction inside context manager
```

**Every duplicate insert would have 500'd** — precisely the failure this change exists to remove, shipped inside the fix for it. Nothing mocked could have found it: the unit tests stub the client and the service, so the session semantics never run.

The handler now wraps the whole `async with`, so the failed session has unwound before the lookup, and `_describe_content_hash_winner` opens its own session and **takes no session parameter**, so it cannot be handed the dead one again.

## The `created_at` tie deserves its own note

A tie-break test that passes for the wrong reason is the failure mode here. `created_at` is `server_default=now()`, fixed for a whole transaction, and auto-chunk inserts all its children in one call — so **ties are the common case, not an edge**.

A single tied group cannot falsify the tie-break: which tie a sort emits first is plan-dependent and not controllable. Measured at **2 of 3 runs passing with the `id` reverted**. Across 20 independent groups an unordered tie would have to pick the lowest id in all 20. Measured: passes 3/3 with the fix, **fails 3/3 without** (13, 8 and 16 of 20 groups wrong).

## Tests

33 new, each confirmed by reverting the thing it covers.

- **Cleanup:** survivor is the oldest; the marker merges into existing metadata rather than replacing it, across all three metadata shapes (SQL NULL, JSON null, object with keys); rows outside the key are untouched, including the `COALESCE`'d NULL-fleet group; a re-run does not re-stamp resolved rows; and the index **builds** after the cleanup where it raised before it.
- **409:** the winner's id survives the trip; a malformed or non-JSON body still yields a usable message rather than raising inside the error path; only 409 is translated (400/404/422/500/503 propagate, so a storage outage is not relabelled as a duplicate); the write pipeline returns 409 rather than the 500 (reverting shows exactly `HTTPException(500, 'Memory write pipeline failed unexpectedly')`); the fanout treats a refusal as recorded **and** still writes the fact after it; and a second agent may write the same content.

Two of #841's tests now take a `without_content_hash_index` fixture: they fabricate duplicate groups to assert what the *lookups* do, and that state is no longer creatable while the constraint stands — which is the constraint working. The fixture's teardown runs the migration's own `CLEANUP_SQL` before rebuilding, because a test that fabricated duplicates has by definition left the table un-indexable.

## The bulk path gets the same treatment

`memory_add_all`'s `ON CONFLICT` arbitrates **one** index — `ix_memories_attempt_unique` — so a content-hash violation there is not swallowed: it aborts the whole multi-row INSERT. Left alone that would be a 500, meaning the same duplicate answers 409 or 500 depending only on which endpoint the caller used.

A second arbiter isn't available (a statement names one) and per-row upserts would trade the batch's single roundtrip for N, so **the batch fails as a unit and says so** — honest, because nothing was written. There's a test asserting exactly that: after a refused batch, the non-duplicate item in it is confirmed absent.

**The auto-chunk children need something different from a 409.** The parent row is already committed when the children go in, so a 409 there would report failure for a write that partly persisted and leave the parent childless — the H-05 shape (#815). Both call sites therefore route through `_insert_children_or_degrade`, which swallows **only** `DuplicateMemoryError` and logs at WARNING with the dropped count. Degrading loses nothing in that case: the constraint refuses the batch only when the content is already stored, so those children already exist. Every other failure still propagates.

Because the batch is all-or-nothing, a refusal *can* drop children that were not duplicates — hence the WARNING and the count. The parent's `child_count` will overstate what exists, and that line is the only record of the gap.

Reachability, for the record: this is a **race**, not a routine outcome. #841 gave the children path an in-batch collapse and a live-hash lookup, and `/memories/bulk` dedups via `existing_hashes` + `seen_hashes` before it calls storage. A duplicate reaches the INSERT only when two concurrent writes of the same content interleave.

## Gates

- **4620** core-api passed / 3 skipped / 1 xfailed
- **141** storage passed
- DCO signed

## Deploy note

The cleanup mutates 19 rows in prod on first boot of the new revision. They are soft-deleted, not removed, and all 19 are recoverable by the `metadata.deduped_by_migration = '040'` marker. If the index build is interrupted (Cloud Run's 240 s startup probe vs `CONCURRENTLY`'s unbounded wait on concurrent transactions), `_drop_invalid` clears the half-built index on the next attempt rather than leaving a non-enforcing one behind.

### DIFF
--- a/common/models/memory.py
+++ b/common/models/memory.py
@@ -148,6 +148,32 @@ class Memory(Base):
                 "deleted_at IS NULL AND client_request_id IS NOT NULL"
             ),
         ),
+        # Enforces the dedup contract the write path advertises: one LIVE row
+        # per (tenant, fleet, agent, content_hash). Created ``CONCURRENTLY`` in
+        # migration 040 with the same key and predicate; declared here so
+        # reflection / autogen round-trip against the live schema, and so the
+        # suites that build a schema from this metadata rather than from the
+        # migration chain (``tests/conftest.py`` uses
+        # ``Base.metadata.create_all``) exercise the constraint too.
+        #
+        # ``ix_memories_content_hash`` above is NOT a substitute: it is
+        # non-unique and keyed on ``(tenant_id, content_hash)`` only — it makes
+        # the lookup fast, it never made it correct.
+        #
+        # ``agent_id`` is in the key because two agents recording identical
+        # content are two independent observations, which is the same scope
+        # ``memory_find_by_content_hash`` dedups on. ``COALESCE(fleet_id, '')``
+        # for the reason 007 needs it: PostgreSQL treats NULLs as distinct, so
+        # without it fleetless rows would escape the constraint entirely.
+        Index(
+            "uq_memories_live_content_hash",
+            "tenant_id",
+            func.coalesce(text("fleet_id"), ""),
+            "agent_id",
+            "content_hash",
+            unique=True,
+            postgresql_where=text("deleted_at IS NULL AND content_hash IS NOT NULL"),
+        ),
         Index("ix_memories_valid_range", "ts_valid_start", "ts_valid_end"),
         Index("ix_memories_subject_entity", "subject_entity_id"),
         # For DELETEs, not reads — grep will find no query using it. This is the
--- a/core-api/src/core_api/clients/storage_client.py
+++ b/core-api/src/core_api/clients/storage_client.py
@@ -19,6 +19,38 @@ from core_api.config import settings
 
 logger = logging.getLogger(__name__)
 
+_DUPLICATE_FALLBACK_DETAIL = "Duplicate memory exists"
+
+
+class DuplicateMemoryError(Exception):
+    """Storage refused a memory insert: a live row already holds this content.
+
+    Only raisable since migration 040 gave ``(tenant, fleet, agent,
+    content_hash)`` a partial unique index. Before it, an insert of duplicate
+    content succeeded and quietly created a second row.
+
+    Carries storage's ``detail`` verbatim, which names the winning row's id — an
+    agent told "duplicate" without being told *which* row cannot use the one it
+    should have got. Callers that own an HTTP contract map this to 409, matching
+    ``CheckExactDuplicate``: that gate answers the duplicate it can see before
+    the write, this answers the race it cannot, and callers should not have to
+    tell the two apart.
+    """
+
+
+def _storage_detail(response: httpx.Response) -> str:
+    """Storage's ``detail`` string, or a safe stand-in.
+
+    Defensive about the body because this runs on an error path: a 409 whose body
+    is not JSON, or is JSON without ``detail``, must not turn a clean 409 into a
+    JSONDecodeError from inside the exception handler.
+    """
+    try:
+        detail = response.json().get("detail")
+    except (ValueError, AttributeError):
+        return _DUPLICATE_FALLBACK_DETAIL
+    return detail if isinstance(detail, str) and detail else _DUPLICATE_FALLBACK_DETAIL
+
 
 def _reject_reserved_write_id(agent_id: str | None) -> None:
     """Storage-boundary reserved-id backstop. No-op under ``allow``/``warn``;
@@ -442,7 +474,23 @@ class CoreStorageClient:
         # ({"main"}) reject; the de-collapsed main-<install_id> form (#507) and
         # every named id pass untouched.
         _reject_reserved_write_id(data.get("agent_id"))
-        return await self._post("/memories", data)
+        try:
+            return await self._post("/memories", data)
+        except httpx.HTTPStatusError as exc:
+            # Migration 040's unique index can now REJECT this insert, which
+            # storage maps to 409. Recognised here, at the single boundary every
+            # memory insert funnels through, so no caller has to know that a
+            # storage 409 is a thing this endpoint does.
+            #
+            # A TYPED error rather than ``HTTPException``: nothing else in
+            # ``core_api/clients/`` raises a FastAPI type, and the app-level
+            # ``httpx.HTTPStatusError`` handler documents the reason it cannot do
+            # this job either — it deliberately re-raises 4xx as "a bug in OUR
+            # request shape", which a legitimate duplicate is not. Callers that
+            # own an HTTP contract translate it; see ``WriteMemoryRow``.
+            if exc.response.status_code != 409:
+                raise
+            raise DuplicateMemoryError(_storage_detail(exc.response)) from exc
 
     async def create_memories(self, data: list[dict]) -> list[dict]:
         """Bulk-insert memories with per-attempt idempotency (CAURA-602).
@@ -465,7 +513,21 @@ class CoreStorageClient:
         # Defense-in-depth reserved-id chokepoint (see create_memory).
         for item in data:
             _reject_reserved_write_id(item.get("agent_id"))
-        return await self._post("/memories/bulk", data)
+        try:
+            return await self._post("/memories/bulk", data)
+        except httpx.HTTPStatusError as exc:
+            # Same translation as ``create_memory``, for the same reason — and it
+            # has to be here too, not only there: migration 040's constraint can
+            # abort a whole batch, and ON CONFLICT cannot arbitrate a second
+            # index, so the bulk route answers 409 as well. Without this the two
+            # endpoints would disagree about what a duplicate looks like to a
+            # core-api caller.
+            #
+            # NOTE the batch is all-or-nothing: a 409 here means NOTHING in the
+            # batch was written, unlike the per-item outcomes a success returns.
+            if exc.response.status_code != 409:
+                raise
+            raise DuplicateMemoryError(_storage_detail(exc.response)) from exc
 
     async def get_memory(self, memory_id: str, *, read: bool = True) -> dict | None:
         """Fetch one memory by id.
--- a/core-api/src/core_api/pipeline/steps/write/write_memory_row.py
+++ b/core-api/src/core_api/pipeline/steps/write/write_memory_row.py
@@ -5,7 +5,9 @@ from __future__ import annotations
 import logging
 import time
 
-from core_api.clients.storage_client import get_storage_client
+from fastapi import HTTPException
+
+from core_api.clients.storage_client import DuplicateMemoryError, get_storage_client
 from core_api.pipeline.context import PipelineContext
 from core_api.pipeline.step import StepResult
 from core_api.services.hooks import get_hooks
@@ -82,7 +84,23 @@ class WriteMemoryRow:
             "visibility": data.visibility or "scope_team",
         }
         storage_t0 = time.perf_counter()
-        memory = await sc.create_memory(memory_data)
+        try:
+            memory = await sc.create_memory(memory_data)
+        except DuplicateMemoryError as exc:
+            # Migration 040's unique index rejected the insert. ``CheckExactDuplicate``
+            # ran earlier in this same pipeline and found nothing, so reaching here
+            # means a concurrent writer committed the same content in between — the
+            # one duplicate case a check-then-insert gate cannot see.
+            #
+            # 409, the same code and shape that gate raises, because it is the same
+            # answer: the content is already stored, here is the row. Without this
+            # the step would be marked FAILED and the caller would get "Memory
+            # write pipeline failed unexpectedly" — a 500 for a completely ordinary
+            # race, and one that says nothing about which row to use instead.
+            #
+            # Nothing has been committed by THIS request at this point, so unlike
+            # everything below, raising here is correct rather than a strand.
+            raise HTTPException(status_code=409, detail=str(exc)) from exc
         timings["storage_ms"] = round((time.perf_counter() - storage_t0) * 1000)
 
         # H-05: the row above is COMMITTED, so everything after it degrades rather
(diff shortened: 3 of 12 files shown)


Eldad-Caura:
Finding accepted and fixed. I had documented this gap in the PR description and argued it belonged in a follow-up; on the severity question the review is right and I was wrong — this PR introduces the constraint, so it owns the fallout, and leaving the two endpoints disagreeing about what a duplicate looks like is not a defensible state to ship.

One factual correction to the write-up, which changes the frequency but not the severity: the children insert **is** exercised on every synchronous auto-chunk write, but a *conflict* there is not. #841 gave that path both an in-batch collapse and a live-hash lookup, so a duplicate only reaches the INSERT by a genuine race between two concurrent writes of the same content. Same for `/memories/bulk`, which dedups via `existing_hashes` + `seen_hashes` before it ever calls storage. Reachable, worth fixing — but a race, not a routine outcome.

## What changed

**The 409 mapping, extended to the bulk path** — `memory_add_all` catches the content-hash violation (matched on index name, so `ix_memories_attempt_unique` keeps its own meaning), the bulk route maps it to 409, and `create_memories` in the storage client translates it the same way `create_memory` does. Raw driver text goes to the log, not the response — the same line #840 drew.

I took option 1 from your prompt rather than option 2. A second ON CONFLICT arbiter isn't available (a statement names one), and per-row upserts would trade the batch's single roundtrip for N on a hot path. So the batch fails as a unit and says so — which is honest, because nothing was written. There's a test asserting exactly that: after a refused batch, the *non-duplicate* item in it is confirmed absent.

**And a deeper fix than the one suggested.** Your prompt asked whether the auto-chunk children inserts "need equivalent handling". They need *different* handling: the parent row is **already committed** when the children go in, so mapping to 409 there would hand the caller a 409 for a write that partly persisted and leave the parent childless — the H-05 shape (#815) this audit already has a fix pattern for.

So both children call sites now route through `_insert_children_or_degrade`, which swallows *only* `DuplicateMemoryError` and logs at WARNING with the dropped count. Degrading loses nothing in that specific case: the constraint refuses the batch only when the content is already stored, so the children this call would have written already exist. Every other failure still propagates, because for those the rows genuinely are missing — there's a test for that too, since a degrade that hides real loss would be worse than the 500.

The batch being all-or-nothing does mean a refusal can drop children that were *not* duplicates. That is why the log is a WARNING carrying the count: the parent's `child_count` will overstate what exists, and that line is the only record of the gap.

## Tests

7 new, each confirmed by reverting the specific handler it covers:

- bulk batch with a duplicate → 409, with no driver text and no offending value in the body;
- a refused batch **writes nothing** — the fresh item in it is verified absent afterwards;
- a refused child batch does **not** strand the committed parent, and leaves a trace;
- any other child-insert failure still raises;
- an empty child list makes no storage call;
- the bulk client translates 409 and still propagates 503.

Controls: removing `memory_add_all`'s catch fails the two storage tests; making the children insert raise fails the strand test; removing the bulk-only client translation fails its test (I re-ran that one with a bulk-specific anchor after my first attempt accidentally patched the single-row block and tested the wrong path).

**Gates:** 4620 core-api passed / 3 skipped / 1 xfailed, 141 storage. mypy clean. `ruff check` and `ruff format --check` clean as separate gates.

The PR description's "one known interaction, not fixed here" section is now removed, since it is fixed here.

Eldad-Caura:
@claude — the High is fixed in `c56a9196`, please take another pass.

The bulk path now answers 409 the way the single-row path does, and the auto-chunk children **degrade** instead, because the parent is already committed there and a 409 would strand it. Reasoning and the seven new tests are in the comment above; the two things most worth a second opinion:

1. **Swallowing `DuplicateMemoryError` at the children sites.** I argue it loses nothing — the constraint refuses only when the content is already stored, so those children exist — but it does mean a batch refusal can drop *non-duplicate* children too, since the INSERT is all-or-nothing. That is logged at WARNING with the count rather than being made exact. If you think the parent's `child_count` needs correcting to match, say so.
2. **Failing the batch as a unit** rather than adding a second arbiter or per-row upserts. A statement can only name one arbiter, and per-row would trade one roundtrip for N on a hot path — but it does mean one duplicate rejects the whole batch.

Eldad-Caura:
Round-2 finding accepted and fixed in `d0e91a04` — and it was reachable, so I fixed all the sites rather than just mine.

**Reproduced before changing anything.** Two writes with `fleet_id: ""` *did* get the index's 409, but the winner lookup missed the live row:

```
first insert : 200 -> stored fleet_id = ''
second insert: 409
detail       : Duplicate memory exists but is no longer live; retry the write
```

Reachable because `fleet_id` is `str | None` with no empty-string normalisation anywhere on the write path, so `''` is stored literally. And the symptom is worse than a misleading message — it tells the caller to **retry**, which can only 409 again.

**Worse than you put it, in one respect.** My own comment on that line *asserted* the lookup agreed with the index's `COALESCE` grouping. It did not. A comment claiming a property the code lacks is the failure mode most worth catching, so the fix is one shared predicate — `_content_hash_fleet_scope` — instead of three hand-written ones that can drift from each other or from the constraint they describe.

**Applied to all content-hash lookups**, per your "consider fixing all three call sites if the empty-string case is reachable":

| site | why it matters |
|---|---|
| `_describe_content_hash_winner` | the misleading 409 above |
| `memory_find_by_content_hash` | the gate — disagreeing with the constraint means a clean "no duplicate" followed by a rejected INSERT |
| `memory_find_duplicate_hash` | its sibling, same shape |
| `memory_bulk_find_by_content_hashes` | **a fourth site you didn't list** — the batch lookup #841's auto-chunk dedup depends on. An assertion in my edit script caught it: I expected 2 matches for the gate pattern and got 3. |

`memory_find_semantic_duplicate` deliberately unchanged — it isn't keyed on `content_hash`, so this index doesn't govern it.

**Two more tests**, parameterised over both fleet shapes: the 409 names the live winner and never says "no longer live". Reverting the predicate fails the empty-string case and passes the null one — precisely characterising the bug.

**Adjacent finding, not fixed here** (it needs its own change): `memory_add_all`'s post-conflict re-query has the same mismatch against migration 007's `ix_memories_attempt_unique`, which also keys on `COALESCE(fleet_id, '')`. It branches on `fleet_id is not None`, and its comment claims the re-query window "matches the unique index's scope … exactly" — which it doesn't for the `''`/NULL pair. Same class, different constraint, so it belongs with 007's domain rather than bolted on here.

**Gates:** 4620 core-api passed / 3 skipped / 1 xfailed, 143 storage. mypy clean. `ruff check` and `ruff format --check` clean as separate gates.

Eldad-Caura:
@erni-a please apporve

erni-a (approved):

PR #843, opened 2026-08-20
fix(crystallizer): a crashed run must not disable crystallization forever

### LINKED ISSUE #817: H-07: Crystallization permanently wedges: uncaught HTTPException leaves the report stuck 'running' with no staleness cutoff
Finding **H-07** of the 2026-08-14 OSS/platform engineering audit, rated **high**. Filed now so the ten correctness highs are trackable — they previously existed only inside the audit artifact. The eight security highs from the same audit are all fixed and merged (#799, #801, #802, #804, #806).

**Area:** core-api · **Surfaced by:** bugs-aux-services

## Verified still present

The per-fact handler is still `except (SQLAlchemyError, ValueError, GoogleAPIError)` and the outer is `except (SQLAlchemyError, httpx.HTTPError, ValueError, RuntimeError)`. Neither includes `HTTPException`, and the module never mentions it.

## Mechanism

run_crystallization() short-circuits when a 'running' report row exists (line 93-95) and there is NO staleness cutoff anywhere (storage-side report_find_running filters only status=='running', postgres_service.py:8728-8743; nothing ever resets a stuck row). Inside _run_crystallization, create_memory() is called per extracted fact (line 317) with the per-fact except clause catching only (SQLAlchemyError, ValueError, GoogleAPIError) — fastapi.HTTPException is NOT caught. create_memory deterministically raises HTTPException(409) on semantic duplicates (memory_service.py:894-898, threshold SEMANTIC_DEDUP_THRESHOLD=0.95, on by default): crystallized facts are by construction near-verbatim merges of cluster members that are themselves >=0.95-similar and still ACTIVE at create time (sources are archived only after the create loop). The fake-LLM fallback (_crystallize_fake) even returns a source memory's content VERBATIM (cosine 1.0). The escaping HTTPException also passes the outer except at line 181 — (SQLAlchemyError, httpx.HTTPError, ValueError, RuntimeError) does not include HTTPException — so run_crystallization raises without ever reaching sc.update_report (line 194). Result: the report row stays status='running' forever, and EVERY future run (manual POST /crystallize, nightly lifecycle trigger via lifecycle_audit.crystallize) returns the stale report id at line 93 and silently does nothing. Crystallization is permanently disabled for that tenant until an operator hand-edits the analysis_reports row. Secondary: if two racing first calls both create reports, report_find_running's scalar_one_or_none() raises MultipleResultsFound → 500s on every subsequent call.

## Suggested fix

Wrap everything between create_report and update_report in try/finally that marks the report 'failed' on any exception; add HTTPException to the per-fact and outer except clauses (409 duplicates are an expected outcome, count and continue); add a staleness cutoff (e.g. started_at > now()-interval '1 hour') to report_find_running.

---

*Location cited by the audit: `core-api/src/core_api/services/crystallizer_service.py:330`. Description and suggested fix are the audit's own text; the verification section above is a fresh check.*

Eldad-Caura:
Fixes #817. Re-located against `main` at `f718790b` before changing anything — the issue's line numbers had drifted.

## The bug

`run_crystallization` short-circuits on whatever `find_running_report` returns, and that lookup matched on `status == 'running'` and **nothing else**. So a report row left `running` by a crashed run suppressed every later run — manual POST and the nightly lifecycle trigger alike — permanently, until an operator edited the row by hand.

The escape route was an uncaught `HTTPException`:

| handler | catches |
|---|---|
| per-fact | `(SQLAlchemyError, ValueError, GoogleAPIError)` |
| outer | `(SQLAlchemyError, httpx.HTTPError, ValueError, RuntimeError)` |

`HTTPException` is in neither, so a 409 from `create_memory` propagated past both and aborted the run **before** `update_report`.

That 409 is expected **by construction**: a crystallized fact is a near-verbatim merge of cluster members that are ≥0.95 similar to each other and still ACTIVE at create time (sources are archived only *after* the create loop), so the semantic gate fires on it. Verified stronger than the issue states — the fake-LLM path returns a member's content **verbatim**, so the *exact* content-hash gate fires too, and that one doesn't need semantic dedup enabled at all.

## The bug, preserved in the dev database

While diagnosing this, `analysis_reports` locally held:

| status | count | detail |
|---|---|---|
| completed | 16 | newest `2026-08-18 05:11:42` |
| **running** | **1** | started `05:11:42.810` — wedged for two days |
| failed | 10 | all from today — this guard working |

A run crashed on 2026-08-18 and crystallization had been dead there ever since. **The three `test_api_crystallizer` tests had been green for two days precisely because the endpoint was short-circuiting on that orphaned row** — passing while the feature did nothing.

## The fix

1. **A guard** around everything between `create_report` and `update_report`, marking the report `failed` and re-raising. `except BaseException`, not `Exception`: a `CancelledError` wedges the row exactly as thoroughly as a bug and is *more* likely on the scheduled path. The marking write is best-effort and its failure must not replace the original exception.
2. **The per-fact handler catches `HTTPException`.** 409 → `duplicate_facts`, anything else → `failed_facts`. Counted, not just logged, because a run whose every fact was rejected now *completes* with `new_memories: 0` — otherwise indistinguishable from "found nothing to crystallize".
3. **The outer handler gains `HTTPException`** as a backstop, preserving the contract that a crystallization failure is non-blocking.
4. **A staleness cutoff** (`REPORT_RUNNING_STALE_AFTER`, 1h) in `report_find_running`. Part 1 stops *new* orphans; this recovers rows already in the table, and is the backstop for the one case part 1 can't cover — the failure-marking write itself failing. A ceiling on a plausible run, not a timeout: nothing is cancelled, and the worst case of getting it wrong is a second report row.

## And the run moves off the request

Fixing the abort forced this. `POST /crystallize` awaited the whole run and then answered `status="running"` — a claim that was never true, since the run had already finished by the time the response was written.

It only looked fine because the run aborted in about a second. With the 409 handled, the run does its actual work and **does not fit in a request**. Measured on the local ~19k-row database: **1.23s aborting on main, past the request timeout (504) when completing.** The endpoint's test file went from **226s of timeouts to 1.55s** once the run was scheduled rather than awaited.

So `run_crystallization` is split:

- `_reserve_report` — find-or-create, awaited, returns the id
- `start_crystallization` — reserves, then **publishes** the run to its own topic (see below). Both routes use it. The response is now true, and `GET /crystallize/reports` is how a caller follows the run — which the response shape always implied
- `run_crystallization` — unchanged semantics, still awaits. The nightly lifecycle trigger keeps using it: no request budget, and it wants the finished result
- `_execute_crystallization` — the guarded body, for an already-reserved row

`/crystallize/all` gains more from this than the single route: it fans out, so awaiting each run made the request cost the **sum** of them.

## Also fixed — the issue mentions it only in passing

`report_find_running` used `scalar_one_or_none()`, which raises `MultipleResultsFound` on two matching rows — a 500 on *every* call rather than a wedge. Two rows is reachable: two racing first calls each create one and nothing in the schema prevents it. Now newest-first + `LIMIT 1`, the same degrade #839 applied to the content-hash lookups. Newest rather than oldest, because this answers "is a run in flight".

## The run is queued, not scheduled in-process

Review round 1 raised this and it changed the design. A fire-and-forget `asyncio.Task` assumes the process keeps scheduling it after the response is flushed. Checked against the deploy config:

- `staging-memclaw-core-api` has `--cpu-boost` but **not** `--no-cpu-throttling`; prod has it, but only so the ddtrace writer isn't starved between requests — and it keeps CPU available while an instance *lives*, not the instance alive for a multi-minute task.
- **The in-process event bus dispatches with `asyncio.create_task`** — so `track_task` and "publish" are the same mechanism in that mode.

`EVENT_BUS_BACKEND=pubsub` is what the deployed services run, so publishing there is a real queued delivery with retry and a DLQ. Under the in-process bus (OSS standalone default) it stays an asyncio task, which is fine: a self-hosted container isn't subject to request-scoped CPU throttling.

**A dedicated topic**, not the existing `CRYSTALLIZE_REQUESTED` — that handler needs a `lifecycle_audit` row an API request has no id for, and it **dedups on a 24h window**, so a manual "crystallize now" would be silently skipped after a successful nightly run.

The payload carries the **already-reserved** `report_id` and the consumer executes that row. A consumer that reserved its own would leave the caller polling an id nothing ever finishes — this bug, reintroduced by its own fix.

## ⚠️ Deploy prerequisite (enforced)

`events_manifest.json` is regenerated, which is the contract `caura-memclaw-enterprise`'s `check_pubsub_provisioning.py` reads to **fail CI** when OSS consumes a topic with no matching Terraform subscription — per its docstring, "the gap that took staging down when `memclaw.lifecycle.insights-requested` shipped unprovisioned."

**So this needs a Terraform subscription for `memclaw.lifecycle.crystallize-on-demand-requested` before it can deploy.** If it did ship unprovisioned the failure is degraded rather than wedged: the message is never delivered, the report stays `running`, and the 1-hour cutoff recovers it.

## Behaviour change to be aware of

`POST /crystallize` and `POST /crystallize/all` no longer wait for the run. They return the report id immediately with `status="running"` — the same response shape as before, now accurate. Any caller that treated the 200 as "analysis complete" needs to poll `GET /crystallize/reports` instead. Nothing in-tree did (the nightly path uses `run_crystallization` directly, which still awaits), but it's worth a second pair of eyes.

## Tests

22 new, each confirmed by reverting the specific thing it covers.

- **Crystallizer (10):** a 409 from either gate is counted and the run continues; a non-409 counts separately; a duplicate doesn't abandon the facts after it; an escaping exception marks the report `failed`; the original exception is re-raised unchanged; a failure to mark failed doesn't mask the real cause; a `CancelledError` is caught too; a clean run still completes; and a **genuinely** running report still short-circuits — fixing the wedge must not turn the lock off.
- **Storage (6):** a fresh running report still blocks; a stale one no longer does; one just *inside* the cutoff still does (both sides of the boundary, so the cutoff can't drift into cancelling live work); two running rows return one instead of raising; a completed row never blocks; another fleet's row doesn't block.
- **Queueing (6):** `start_crystallization` publishes rather than running inline; it doesn't publish a second run when one is in flight; the consumer executes the **reserved** report rather than re-reserving; `run_crystallization` still **awaits**, so splitting the entry points didn't quietly make the nightly path fire-and-forget; and a cancelled marking write doesn't mask the original cause.

One process note. An earlier version of this PR scheduled the run in-process, and its test asserted "returns while the run is still blocked". Reverting that fix made the test **hang** rather than fail — the awaiting implementation blocked forever on a run that never returns — so it needed an `asyncio.wait_for` bound. That version is gone (the run is published now), but the lesson stands and is why the current queueing tests assert on the publish call rather than on timing: a test that hangs on regression is nearly as useless as one that passes.

**Controls:** neutering the guard fails 2, removing the per-fact handler fails 4, dropping the cutoff fails 1, restoring `scalar_one_or_none` fails 1, running inline instead of publishing fails 1, re-reserving in the consumer fails 1, and narrowing the inner handler to `Exception` fails 1.

## Gates

- **4634** core-api passed / 3 skipped / 1 xfailed
- **149** storage passed
- DCO signed

## Noted while here, not fixed

`report_type` is accepted by `GET /reports/running` and sent by the storage client, but **`analysis_reports` has no such column** — dropped by the route, and dropped again by `_filter_fields` on the way in. Harmless while there's one report type, and entirely fiction. `report_get_latest_completed` also ignores `fleet_id`, so `GET /reports/latest` can answer with another fleet's report. Both want their own change; neither is part of this wedge.


Eldad-Caura:
Both findings accepted and fixed in `d5444cd9`. The Medium was right, and checking it properly changed the design.

## The Medium — verified in the deploy config, not assumed

You asked me to confirm the runtime model rather than take it on trust. Doing that turned up two things:

- **`staging-memclaw-core-api` gets `--cpu-boost` but NOT `--no-cpu-throttling`.** Prod does have it — but only so the ddtrace writer isn't starved between requests, and it keeps CPU available while an instance *lives*, not the instance alive for a multi-minute task once no request is in flight.
- **Worse: the in-process event bus dispatches with `asyncio.create_task`.** So `track_task` and "publish to the bus" are literally the same mechanism in that mode — which means "just publish it" is not automatically an improvement, and I had to check where the bus is real.

`EVENT_BUS_BACKEND=pubsub` is what the deployed services run (alongside `DEPLOYMENT_MODE=deferred`), so publishing there **is** a queued delivery with retry and a DLQ. Under the in-process bus (OSS standalone default) it stays an asyncio task — acceptable there precisely because a self-hosted container isn't subject to request-scoped CPU throttling.

So the run is now published, not scheduled. The staleness cutoff remains the backstop for a message that never gets delivered at all.

## Why a dedicated topic rather than the existing `CRYSTALLIZE_REQUESTED`

I tried to reuse it. Its handler can't serve an API trigger:

- it needs a `lifecycle_audit` row to report progress into, which an API request has no id for — inventing one would file manual runs into the record of the scheduled sweep;
- it **dedups on a 24-hour window** (`has_recent_lifecycle_success`). Right for a nightly fanout; wrong for a person pressing "crystallize now", who'd be silently skipped because last night's run succeeded.

Hence `memclaw.lifecycle.crystallize-on-demand-requested`, with a payload carrying the **already-reserved** `report_id`. That detail is the subtle one: the consumer must execute the row the API reserved. A consumer that reserved its own would leave the caller polling an id nothing ever finishes — this bug, reintroduced by its own fix. There's a test asserting the consumer never calls `create_report`.

## The deploy prerequisite is enforced, not a footnote

`events_manifest.json` is regenerated, and that's the point: it's the contract `caura-memclaw-enterprise`'s `check_pubsub_provisioning.py` reads to **fail CI** when OSS consumes a topic with no matching Terraform subscription — per its own docstring, "the gap that took staging down when `memclaw.lifecycle.insights-requested` shipped unprovisioned."

So **this PR requires a Terraform subscription for the new topic before it can deploy**, and the enterprise check will say so rather than letting it ship silently. Worth knowing the failure mode if it did ship unprovisioned: the message is never delivered, the report stays `running`, and the 1-hour cutoff recovers it — degraded, not wedged.

## The Low

Correct, and worse than cosmetic: the comment beside that write promises the original exception survives, and `except Exception` broke exactly that promise for a `CancelledError` from the write itself. Now `BaseException`, matching the outer handler for the same reason, with a test that fails when reverted to `Exception`.

## Tests

Three more, each falsified: `start_crystallization` publishes rather than running inline (an inline await fails it); the consumer executes the reserved report rather than re-reserving (re-reserving fails it); and a cancelled marking write doesn't mask the original cause (`except Exception` fails it).

**Gates:** 4634 core-api passed / 3 skipped / 1 xfailed, 149 storage. mypy clean. `ruff check` and `ruff format --check` clean at CI's exact scopes — note `common/` is only format-checked for three named files with `--isolated`, so I wrapped my `common/` additions to that directory's 88-char convention by hand rather than reformatting files CI doesn't police.

Eldad-Caura:
@claude — the Medium is fixed in `d5444cd9`, please take another pass.

The fire-and-forget task is gone: the run is published to a dedicated topic and executed by a consumer. Reasoning and the config I checked are in the comment above. Three things most worth a second opinion:

1. **The dedicated topic.** I couldn't reuse `CRYSTALLIZE_REQUESTED` because its handler needs a `lifecycle_audit` row and dedups on 24h. If you think an API trigger *should* share the nightly fanout's dedup, say so — that's a product call I made as an engineering one.
2. **The reserved-`report_id` contract.** The publisher creates the row and the consumer must finish *that* row. I've tested that the consumer never calls `create_report`, but the invariant lives in a comment rather than a type.
3. **The in-process fallback.** Under `EVENT_BUS_BACKEND=inprocess` this is still `asyncio.create_task`, i.e. exactly what you flagged. I argue that's acceptable because self-hosted containers aren't CPU-throttled per request — but it does mean the fix is deployment-conditional, which is worth challenging.

Eldad-Caura:
All three accepted and fixed in `ffdd39e6`. Worth noting the pattern: every finding across both rounds has been in the *machinery* added to deliver the fix, not in the wedge fix itself — which has now survived two rounds untouched.

## Medium — the missing `ValidationError` guard

Verified before fixing: `_run_action` does catch it and drop, and its comment names the prior incident — an unhandled exception on a bad payload nacks the delivery and redelivers forever / DLQs (audit M16).

So this is the cost of bypassing the shared wrapper, made concrete. The handler now has its own guard with the same drop-and-log shape. The realistic trigger is exactly the one you named: a rolling deploy where publisher and consumer briefly disagree on the schema.

## Medium — idempotency. The sharpest of the three.

`EventBus.subscribe` documents at-least-once, and redelivery here isn't hypothetical: **this is the multi-minute run that made moving off the request necessary in the first place**, so it can outlive the subscription's ack deadline and be redelivered while the first attempt is still going. Re-running would repeat every LLM call and overwrite a report that had already finished.

`execute_reserved_report` now reads the row via `sc.get_report` (which already existed) and short-circuits on a terminal status.

**One thing I've stated plainly in the code rather than let the fix imply:** this is *not* a lock. Two deliveries can still overlap while the row is `running`, because a status read cannot exclude an in-flight run. What it prevents is the case that actually corrupts a result — clobbering a `completed`/`failed` report with a second full run. If you think the overlap window needs closing too, that wants a real claim (a CAS on status, or an advisory lock), which is a different change from this one.

## Low — report shape

Correct. `analysis_reports.crystallization` had a different shape depending on which branch ran, and a consumer reading `duplicate_facts` shouldn't have to know that. Both keys added to the fallback.

## Tests

Five more, each falsified: removing the idempotency check fails 2 (one per terminal status), removing the `ValidationError` guard fails 1, dropping the fallback keys fails 1.

Also the positive cases, so the short-circuit can't quietly turn every *first* delivery into a no-op: a still-`running` row **is** executed, and a vanished row is dropped rather than re-reserved.

**Gates:** 4641 core-api passed / 3 skipped / 1 xfailed, 149 storage. mypy clean. `ruff check` and `ruff format --check` clean at CI's exact scopes. Events manifest in sync.

Reminder on the deploy prerequisite, since it's enforced rather than advisory: this needs a Terraform subscription for `memclaw.lifecycle.crystallize-on-demand-requested` in `caura-memclaw-enterprise`, or `check_pubsub_provisioning.py` will fail there.

Eldad-Caura:
@claude — round-2 findings fixed in `ffdd39e6`, please take another pass.

Both Mediums and the Low are addressed; details above. The one place I stopped short of what a stricter reading might want: the idempotency fix prevents clobbering a **finished** report, but does not prevent two deliveries overlapping while the row is still `running` — a status read can't exclude an in-flight run. Closing that needs a real claim (CAS on status, or an advisory lock). I judged that out of scope for a wedge fix and said so in the code rather than leaving the guard looking stronger than it is; push back if you disagree.

Eldad-Caura:
Both accepted and fixed in `544727d9`. These two are worth naming precisely, because they are **this audit's own findings recurring inside its fix.**

## Medium — the phantom `running` row

The row is reserved *before* the publish, so a transient Pub/Sub error left a report nothing would ever execute, blocking every later trigger for that tenant until the staleness cutoff expired. One network blip buying an hour of "crystallization does nothing" — which is H-07's own shape, reached through the fix for H-07.

The publish is now guarded: the report is marked `failed` before the error is re-raised, so the caller gets an inspectable report *and* the next trigger is free to start. The marking write is best-effort and must not mask the publish error — same rule, and same test, as the run's own guard.

## Medium — reading the replica. This is H-02 (#812) again.

Verified rather than assumed:

- `_get` defaults `read=True`;
- line 535 of the storage client documents `read=False` as "Pass it for a read-your-write";
- `find_by_content_hash` passes it explicitly, commented *"stale replica data here would…"*.

The check **is** a read-your-write — just across a *delivery* boundary rather than a request one. The status it reads was written by the **previous delivery of the same message**, so a lagging replica answers `running` for a finished report and the entire multi-minute LLM run happens again. Precisely the duplication the check was added to prevent, defeated by the read it depends on.

`get_report` now takes a `read` flag (default `True` — the display paths legitimately want the replica) and the check passes `read=False`.

## Tests

Three more, each falsified: neutering the publish guard fails 1, reverting to the replica read fails 1, the marking write masking the publish error fails 1.

**Gates:** 4644 core-api passed / 3 skipped / 1 xfailed, 149 storage. mypy clean. `ruff check` and `ruff format --check` clean at CI's exact scopes. Events manifest in sync.

## An observation worth more than these two fixes

Across three rounds, **all seven findings have been in the machinery added to deliver the fix — none in the four-part wedge fix itself**, which has survived all three rounds untouched.

And two of the seven were textbook instances of patterns *this same audit already filed*: H-02 (replica lag on a read-your-write) and H-05 (a committed row abandoned by a later failure). I reintroduced both in new code, in the PR fixing a sibling of them.

That says these are codebase-wide shapes rather than isolated bugs, and they'd be better served by something enforceable — a lint for `_get` on a read-your-write path, or a checklist item for "reserved a row before an operation that can fail" — than by finding them one PR at a time. Happy to file that as its own issue.

It also strengthens an argument I raised earlier and will restate now that there's more evidence: the plain inline-with-504 option would have shipped the wedge fix with none of this surface area. The Pub/Sub path is the better end state, but it is a materially bigger change than #817 required, and the review history is the receipt.

Eldad-Caura:
@claude — round-3 findings fixed in `544727d9`, please take another pass.

Both were correct and both were instances of patterns this audit already filed (H-02 replica-lag read-your-write; H-05 row abandoned by a later failure), reintroduced in new code. Details above.

Since three rounds have now found seven issues **all in the delivery machinery and none in the wedge fix**, the specific thing I'd value on this pass: is the on-demand path carrying more risk than the bug it fixes? If your read is yes, say so plainly — reverting to the inline call and shipping only the four-part wedge fix is still available, and I'd rather take that than keep hardening a mechanism the issue never asked for.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #844, opened 2026-08-20
fix(storage): org hard-purge must reach every tenant-scoped table

### LINKED ISSUE #819: H-09: Org hard-purge misses six tenant-scoped tables — tenant data survives 'permanent' deletion
Finding **H-09** of the 2026-08-14 OSS/platform engineering audit, rated **high**. Filed now so the ten correctness highs are trackable — they previously existed only inside the audit artifact. The eight security highs from the same audit are all fixed and merged (#799, #801, #802, #804, #806).

**Area:** core-storage · **Surfaced by:** bugs-storage

## Verified still present

All six tables have **zero** mentions anywhere in `core-storage-api/.../postgres_service.py`, so they are in neither `_PURGE_TENANT_TABLES` nor `count_tenant_data`. Confirmed they exist as live tables (`session_traces`, `recall_event`, `recall_candidate`, `agent_activity_digests`, `forge_rejected_fingerprints`, `capability_usage`, `audit_chain_head` all present in the database).

## Mechanism

purge_tenant_data() (backing core-api POST /admin/org/purge-data, the documented 'hard side of organization deletion') deletes only the tables listed in _PURGE_TENANT_TABLES + _PURGE_ORG_KEYED_TABLES. Six tenant_id-scoped tables created by later migrations are absent and have no ON DELETE CASCADE path from any purged parent, so their rows survive an org hard-delete: session_traces (migration 021 — carries goal_phrase, memory_ids, signals_summary), recall_event/recall_candidate (027 — recall_event.query_text is raw user query content; recall_candidate has FK only to recall_event), agent_activity_digests (029 — 'narrative' is an LLM summary of the tenant's memories), forge_rejected_fingerprints (020), capability_usage (023), and audit_chain_head (025). Failure scenario: operator runs POST /admin/org/purge-data for tenant T → response reports per-table counts and success, but SELECT * FROM session_traces/recall_event/agent_activity_digests WHERE tenant_id='T' still returns the tenant's content — data the endpoint promises to permanently destroy is retained indefinitely. count_tenant_data() (line 2232, backing the CAURA-696 deletion preview) mirrors the same list, so the 'what will be deleted' preview also under-reports. memory_conflicts/memory_derivations are fine (CASCADE from memories); lifecycle_audit is documented as intentionally retained, but none of the six above are mentioned anywhere as intentional.

## Suggested fix

Append session_traces, recall_event (recall_candidate cascades), agent_activity_digests, forge_rejected_fingerprints, capability_usage, and audit_chain_head to _PURGE_TENANT_TABLES (children-before-parents), and mirror them in count_tenant_data. Document tenant_suppression/lifecycle_audit as intentionally retained.

---

*Location cited by the audit: `core-storage-api/src/core_storage_api/services/postgres_service.py:333`. Description and suggested fix are the audit's own text; the verification section above is a fresh check.*

Eldad-Caura:
Fixes #819.

## The bug

`purge_tenant_data` backs `POST /admin/org/purge-data`, documented as the hard side of organization deletion. It deletes only what `_PURGE_TENANT_TABLES` lists, and that tuple never grew as later migrations added tenant-scoped tables. **None of the missing ones has an ON DELETE CASCADE path from a purged parent**, so their rows survived: the endpoint reported per-table counts and success while the tenant's content was still queryable by `tenant_id`.

`count_tenant_data` — the CAURA-696 deletion preview — iterates the same tuples, so the "what will be deleted" answer under-reported by exactly the same set. It needs no separate change; one edit fixes both.

## The list is seven, not six

Derived from the live schema rather than from the issue, which is how the discrepancy surfaced:

| table | what survived |
|---|---|
| `recall_event` | raw user query text (027). `recall_candidate` has no `tenant_id` and rides its CASCADE |
| `session_traces` | `goal_phrase`, `memory_ids`, `signals_summary` (021) |
| `agent_activity_digests` | `narrative` is an LLM summary **of** the tenant's memories (029) |
| `forge_rejected_fingerprints` | (020) |
| `capability_usage` | (023) |
| `audit_chain_head` | the tamper-evident chain head for `audit_log`, which was already purged (025) |
| **`tenant_usage_counters`** | **not in the issue** — migration 039 postdates the audit that filed it |

`audit_chain_head` is listed immediately after `audit_log` and deliberately not retained: with the log rows gone the head points at nothing, and leaving it would make a later chain verification fail against a tenant that no longer exists.

## Retained, now recorded as such

A new `_RETAINED_TENANT_TABLES` names the tenant-scoped tables deliberately kept, so their absence reads as a decision rather than as the oversight this issue is:

- **`tenant_suppression`** — the suppression flag itself. Wiping it would *un-suppress* a tenant as part of deleting them, which is the opposite of what a purge is for.
- **`lifecycle_audit`** — the operational audit trail, including the row recording the hard-delete being performed (already documented on `_PURGE_ORG_KEYED_TABLES`).

## The sibling the issue doesn't mention

**`purge_fleet_data` has the same gap.** Exactly three of the new tables carry their own `fleet_id` — `session_traces`, `agent_activity_digests`, `forge_rejected_fingerprints` — so a single fleet's teardown was leaving them behind in a *shared* tenant, which is the case that purge exists for.

The other four are tenant-wide with no `fleet_id` and are correctly absent: removing them for one fleet would delete another fleet's rows.

## The test that matters longest

`test_every_tenant_scoped_table_is_purged_or_deliberately_retained` derives its expectation from `information_schema`. Every `tenant_id` table must have exactly one of three fates:

1. listed in `_PURGE_TENANT_TABLES` — deleted directly;
2. reachable by ON DELETE CASCADE from a purged parent;
3. listed in `_RETAINED_TENANT_TABLES` — a recorded decision.

**A future migration that adds a tenant-scoped table now fails this test instead of silently reintroducing the bug** — which is precisely how `tenant_usage_counters` slipped past the audit.

A second test stops the retained list becoming a way to hide a gap: it must not overlap the purge list, and must not name tables that don't exist.

Then end-to-end: a row in each newly covered table is counted by the preview, reported deleted by the purge, and actually gone afterwards; and the fleet purge removes the doomed fleet's rows while leaving a **sibling** fleet's intact — over-reach being the failure that matters for a purge used on a shared tenant.

**Controls:** removing the six tenant tables fails 2 tests and the assertion names all six; removing the three fleet tables fails 2.

The end-to-end inserts build themselves from `information_schema` (required columns only, filled by type) so a future NOT NULL doesn't quietly turn them into tests of nothing. Two columns need fixed values because a CHECK restricts them — `session_traces.outcome_label` and `capability_usage.transport` — held in a small map rather than a constraint parser.

## Gates

- **148** storage passed
- **4620** core-api passed / 3 skipped / 1 xfailed
- DCO signed

## One note on the suite run, worth recording

Five crystallizer tests failed on the first pass with `MultipleResultsFound` from `report_find_running`. **Not caused by this change** — this diff adds table names to two tuples and touches nothing crystallizer-related. The dev database had accumulated **three** `running` `analysis_reports` rows for one tenant, which `scalar_one_or_none()` on `main` cannot answer. Clearing those orphans, all eleven pass.

Worth flagging because it's an unplanned live reproduction of the secondary finding in #817/#843: two-or-more running rows is reachable in ordinary use — these came from this session's own crystallizer runs — and until #843 lands it 500s *every* crystallizer call for the affected tenant.

### DIFF
--- a/core-storage-api/src/core_storage_api/services/postgres_service.py
+++ b/core-storage-api/src/core_storage_api/services/postgres_service.py
@@ -340,11 +340,55 @@ _PURGE_TENANT_TABLES: tuple[str, ...] = (
     "agents",
     "fleet_nodes",
     "audit_log",
+    # The tamper-evident chain head for ``audit_log`` (migration 025). Listed
+    # right after it, and NOT retained: with the log rows gone the head points at
+    # nothing, and leaving it behind would make a later chain verification fail
+    # against a tenant that no longer exists.
+    "audit_chain_head",
     "documents",
     "analysis_reports",
     "dedup_reviews",
     "background_task_log",
     "idempotency_responses",
+    # ── H-09 (OSS #819): tenant-scoped tables added by later migrations that
+    # this tuple never grew to include. None has an ON DELETE CASCADE path from
+    # any table above, so their rows SURVIVED an org hard-delete — an endpoint
+    # documented as permanently destroying the tenant's data reported success
+    # while the content below was still queryable by tenant_id.
+    #
+    # Ordering is free among these: every one was verified to have no foreign key
+    # at all, so none can trip another's DELETE. Derived from the live schema
+    # rather than from the issue, which is how ``tenant_usage_counters`` (added
+    # after the audit was written) turned up.
+    #
+    # ``recall_event`` carries the raw user query text (migration 027).
+    # ``recall_candidate`` is NOT listed: it has no ``tenant_id`` of its own and
+    # rides the ON DELETE CASCADE from ``recall_event``, the same way
+    # ``memory_entity_links`` rides ``memories``.
+    "recall_event",
+    # ``goal_phrase``, ``memory_ids``, ``signals_summary`` (migration 021).
+    "session_traces",
+    # ``narrative`` is an LLM summary OF the tenant's memories (migration 029) —
+    # derived content, but still the tenant's content.
+    "agent_activity_digests",
+    "forge_rejected_fingerprints",
+    "capability_usage",
+    "tenant_usage_counters",
+)
+
+# Tenant-scoped and deliberately NOT purged, recorded here so their absence
+# reads as a decision rather than the oversight H-09 was:
+#
+# ``tenant_suppression`` — the suppression flag itself. Wiping it would
+# UN-suppress a tenant as part of deleting them, which is the opposite of what
+# a purge is for; the row is a few bytes of policy, not tenant content.
+#
+# ``lifecycle_audit`` (``org_id``-keyed) — the operational audit trail,
+# including the row recording the hard-delete being performed. See
+# ``_PURGE_ORG_KEYED_TABLES``.
+_RETAINED_TENANT_TABLES: tuple[str, ...] = (
+    "tenant_suppression",
+    "lifecycle_audit",
 )
 # OSS keys these by ``org_id``, which equals the tenant id in the
 # single-key-per-tenant OSS model (CAURA-654). ``lifecycle_audit`` is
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #846, opened 2026-08-21
fix(api): the read params the plugin advertises must reach the query

### LINKED ISSUE #820: H-10: Plugin caura_list advertises scope/written_by/weight_min/weight_max; REST GET /memories silently ignores all four
Finding **H-10** of the 2026-08-14 OSS/platform engineering audit, rated **high**. Filed now so the ten correctness highs are trackable — they previously existed only inside the audit artifact. The eight security highs from the same audit are all fixed and merged (#799, #801, #802, #804, #806).

**Area:** plugin · **Surfaced by:** bugs-plugin-clients, discrepancies

## Verified still present

`plugin/src/tool-definitions.ts` still declares `written_by` (:213), `weight_min` (:216) and `weight_max` (:217). `core-api/src/core_api/routes/memories.py` declares none of them as query params (the only `scope` matches there are prose about `scope_agent` visibility). FastAPI drops undeclared query params silently.

## Mechanism

The plugin's caura_list schema (lines 206-226) exposes scope ('agent'|'fleet'|'all'), written_by, weight_min, weight_max — mirroring the MCP caura_list tool (mcp_server.py:2274-2296 accepts all four). But the plugin dispatch stringifies every param into query args of GET /memories, whose FastAPI signature (routes/memories.py:186-215) declares none of scope/written_by/weight_min/weight_max; FastAPI drops undeclared query params silently. Failures: (1) caura_list {agent_id:'me', written_by:'agent-b'} — written_by is dropped and the route treats agent_id as the AUTHOR filter, so the agent gets its own memories instead of agent-b's; (2) caura_list {agent_id:'me', scope:'all'} — scope dropped, results stay author-filtered to 'me' instead of tenant-wide; (3) weight_min/weight_max never filter anything. caura_stats has the identical defect: its 'scope' param (tool-definitions.ts:288) doesn't exist on GET /memories/stats (routes/memories.py:326-335), so scope='fleet'/'all' aggregation semantics are silently lost.

## Suggested fix

Either translate plugin params to the REST contract (written_by → agent_id author filter, drop/implement scope) and remove weight_min/weight_max/scope from the plugin schema, or add these params to the REST routes to match the MCP surface.

---

*Location cited by the audit: `plugin/src/tool-definitions.ts:466`. Description and suggested fix are the audit's own text; the verification section above is a fresh check.*

Eldad-Caura:
Fixes #820.

## The bug

`plugin/src/tool-definitions.ts` declares `scope`, `written_by`,
`weight_min` and `weight_max` on `caura_list`, and `scope` on
`caura_stats` — faithfully mirroring `plugin/tools.json`, which is
generated from the MCP tool specs. Both then dispatch to REST routes
that declared **none** of them, and FastAPI drops an undeclared query
param silently.

So the calls never failed. They answered a different question:

* `caura_list {agent_id:'me', written_by:'agent-b'}` — `written_by`
  dropped, `agent_id` doubles as the author filter, so the caller got
  its OWN memories back, presented as agent-b's;
* `caura_list {agent_id:'me', scope:'all'}` — `scope` dropped, results
  stay author-filtered to the caller instead of tenant-wide;
* `weight_min` / `weight_max` never filtered anything.

A plausible-looking populated list is why this went unnoticed.

## Making `scope` real needed its ladder too

`GET /memories` now accepts all four params and `GET /memories/stats`
accepts `scope`. Adding `scope` without its trust gate would have traded
a wrong-data bug for a worse one: REST has no trust ladder on reads, so
`scope=all` would have been a trust-1 tenant-wide read on a surface
where MCP requires trust >= 2.

`_resolve_read_fleet_gate` therefore moves out of `mcp_server` into
`agent_service` as `resolve_read_fleet_gate`, and the MCP tools and the
REST routes now call one implementation. **The executable body is
unchanged** — verified by comparing the ASTs with docstrings stripped;
only the docstring's first paragraph is edited, to name its four callers
instead of two. `mcp_server` keeps a `_resolve_read_fleet_gate` alias,
the same pattern already used there for `_require_trust`.

`scope` is opt-in. Omitted, both routes behave byte-for-byte as before,
so the dashboard and every existing caller are untouched — and no caller
can regress on a param that was being discarded a commit ago.

## Two gaps in `/memories/stats` that this surfaced

Stats never received hardening its list sibling already had:

* **No fleet-read gate at all.** `GET /memories` has called
  `enforce_fleet_read` all along; stats took `fleet_id` and aggregated
  it ungated, so a trust-1 agent could read the breakdown of a fleet
  whose rows the list route would have refused to show it. Now gated
  under the identical condition.
* **`agent_id` was impersonatable.** In `memory_stats_breakdown` it is
  both the author filter *and* the visibility identity, so an agent
  credential naming a peer learned that peer's private per-type/status
  counts — the widening `GET /memories` closes by preferring the
  gateway-authenticated agent. Stats cannot borrow that fix as-is: with
  one knob doing both jobs, forcing it to the caller would silently
  narrow the historical "no `agent_id` -> team/org-wide" aggregate. The
  conflict is rejected instead, which closes the leak and leaves every
  legitimate call — omitted, or naming yourself — working.

## Tests

`tests/test_api_read_scope_params.py`, 21 tests. Sixteen fail on
`origin/main` with only the source reverted; the five that pass either
way are deliberate "do not break this" guards (legacy stats shape,
naming yourself) and structural checks, and the one non-discriminating
ladder test says so in its docstring.

The last three are the guard this class needed. `GET` tools are matched
against `plugin/tools.json`, so a param a tool advertises but its route
does not read now fails a test instead of returning a wrong answer; the
excuse list for genuinely-not-query params is itself checked for
staleness, and every plugin-exposed tool must be classified as
query-dispatch or body-dispatch, so a new tool cannot escape the check
by being new. Same shape as the purge coverage guard in #844.

`tests/_mcp_test_helpers.py::stub_storage_client` now also patches the
`agent_service` alias of `get_storage_client`. Without it the moved gate
resolves the real ASGI-bridged client and five existing `caura_list`
tests read live storage instead of their stub.

## Not included

The plugin needs no change — its params now reach a route that reads
them. One residue is left deliberately: `enrichBody` injects
`MEMCLAW_FLEET_ID` when the caller named no fleet, so `caura_list
{scope:'all'}` from a fleet-configured install still narrows to that
fleet. That is plugin-side param population, identical in effect to
passing `fleet_id` explicitly to the MCP tool, and separable from making
the REST contract honest.

### DIFF
--- a/core-api/openapi.broker.json
+++ b/core-api/openapi.broker.json
@@ -1491,7 +1491,7 @@
     },
     "/api/v1/memories": {
       "get": {
-        "description": "List memories with filtering, sorting, and pagination.\n\n**Visibility scoping:** When ``agent_id`` is provided, the caller can see\ntheir own ``scope_agent`` memories plus all ``scope_team``/``scope_org``\nmemories. When ``agent_id`` is omitted, ``scope_agent`` memories are hidden\n(safe default). This means memory types like ``insight`` that are typically\ncreated with agent-scoped visibility will only appear when ``agent_id`` is\npassed.\n\n**Pagination:** Both cursor-based and offset-based pagination are supported.\nCursor pagination (via ``cursor``) is recommended for large datasets and only\nworks with ``sort=created_at&order=desc``. Offset pagination works with any\nsort order.",
+        "description": "List memories with filtering, sorting, and pagination.\n\n**Visibility scoping:** When ``agent_id`` is provided, the caller can see\ntheir own ``scope_agent`` memories plus all ``scope_team``/``scope_org``\nmemories. When ``agent_id`` is omitted, ``scope_agent`` memories are hidden\n(safe default). This means memory types like ``insight`` that are typically\ncreated with agent-scoped visibility will only appear when ``agent_id`` is\npassed.\n\n**Author filter:** ``written_by`` selects the authoring agent. It is separate\nfrom ``agent_id`` (the visibility identity) so a caller can ask for a peer's\nauthored rows without claiming that peer's identity; ``agent_id`` remains the\nauthor filter when ``written_by`` is omitted. Filtering by author only ever\nnarrows the caller's existing visibility — it cannot surface a row an\nunfiltered list would have hidden.\n\n**Scope:** ``scope`` is optional and mirrors the MCP / plugin ``caura_list``\ncontract, including its trust ladder (see ``_resolve_scoped_read``). Omit it\nand the route behaves exactly as before.\n\n**Pagination:** Both cursor-based and offset-based pagination are supported.\nCursor pagination (via ``cursor``) is recommended for large datasets and only\nworks with ``sort=created_at&order=desc``. Offset pagination works with any\nsort order.",
         "operationId": "list_memories_api_v1_memories_get",
         "parameters": [
           {
@@ -1540,6 +1540,43 @@
                 }
               ],
               "title": "Agent Id"
+            }
+          },
+          {
+            "description": "Opt-in read scope, mirroring the MCP/plugin contract. 'agent' = your own memories (trust >= 1); 'fleet' = cross-agent within a fleet (own fleet trust >= 1, a different fleet trust >= 2); 'all' = tenant-wide (trust >= 2). Omit for the historical behaviour, where `agent_id` is the author filter.",
+            "in": "query",
+            "name": "scope",
+            "required": false,
+            "schema": {
+              "anyOf": [
+                {
+                  "pattern": "^(agent|fleet|all)$",
+                  "type": "string"
+                },
+                {
+                  "type": "null"
+                }
+              ],
+              "description": "Opt-in read scope, mirroring the MCP/plugin contract. 'agent' = your own memories (trust >= 1); 'fleet' = cross-agent within a fleet (own fleet trust >= 1, a different fleet trust >= 2); 'all' = tenant-wide (trust >= 2). Omit for the historical behaviour, where `agent_id` is the author filter.",
+              "title": "Scope"
+            }
+          },
+          {
+            "description": "Author filter. Distinct from `agent_id`, which also serves as the visibility identity; falls back to `agent_id` when omitted.",
+            "in": "query",
+            "name": "written_by",
+            "required": false,
+            "schema": {
+              "anyOf": [
+                {
+                  "type": "string"
+                },
+                {
+                  "type": "null"
+                }
+              ],
+              "description": "Author filter. Distinct from `agent_id`, which also serves as the visibility identity; falls back to `agent_id` when omitted.",
+              "title": "Written By"
             }
           },
           {
@@ -1657,6 +1694,42 @@
                 }
               ],
               "title": "Run Id"
+            }
+          },
+          {
+            "in": "query",
+            "name": "weight_min",
+            "required": false,
+            "schema": {
+              "anyOf": [
+                {
+                  "maximum": 1,
+                  "minimum": 0,
+                  "type": "number"
+                },
+                {
+                  "type": "null"
+                }
+              ],
+              "title": "Weight Min"
+            }
+          },
+          {
+            "in": "query",
+            "name": "weight_max",
+            "required": false,
+            "schema": {
+              "anyOf": [
+                {
+                  "maximum": 1,
+                  "minimum": 0,
+                  "type": "number"
+                },
+                {
+                  "type": "null"
+                }
+              ],
+              "title": "Weight Max"
             }
           },
           {
--- a/core-api/src/core_api/mcp_server.py
+++ b/core-api/src/core_api/mcp_server.py
@@ -31,7 +31,6 @@ from core_api.auth import get_admin_key
 from core_api.clients.storage_client import KeystoneUpsertPayload, get_storage_client
 from core_api.constants import (
     DEFAULT_SEARCH_TOP_K,
-    DEFAULT_TRUST_LEVEL,
     EVOLVE_OUTCOME_TYPES,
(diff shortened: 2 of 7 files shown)


Eldad-Caura:
Both observations checked. One applied, one declined with evidence. Also fixed the CI failure (unrelated to the review).

**Applied — inverted weight range.** `weight_min > weight_max` now returns 422 instead of an empty page. Agreed, and it's more than a UX nicety here: an empty page reads as "no such memories", which is the same silently-wrong-answer shape this PR exists to remove. Covered by `test_an_inverted_weight_range_is_rejected`, confirmed failing with the validation removed.

**Declined — threading the agent dict through to collapse the three lookups.** The premise that they're redundant doesn't hold, in two separate ways:

1. **The second read is deliberate.** `resolve_read_fleet_gate`'s own docstring states that every branch returns a `min_level` matching the access it grants *so that* a stale read can only under-grant, and that `require_trust` re-reads trust to reject a caller demoted in between — "`get_agent` is uncached, keeping even that transient divergence to a single async context switch (do NOT add caching — it would widen it)". Passing the already-fetched dict to `require_trust` is that caching, and it widens exactly the TOCTOU window the comment was written to close.

2. **The third read is load-bearing, not redundant.** `resolve_read_fleet_gate` short-circuits `scope='agent'` to `(1, fleet_id)` **without inspecting the fleet it was handed** (`agent_service.py:350-351`) — correct, since scope='agent' pins the author filter to the caller. That makes `enforce_fleet_read` the only thing refusing `scope=agent&fleet_id=<another fleet>` from a trust-1 caller. Removing it to save a round-trip would open a hole.

Since that non-redundancy is invisible from the call site — three lookups genuinely *look* like two too many — I've pinned it: `test_scope_agent_still_cannot_name_another_fleet` fails if `enforce_fleet_read` is ever dropped from this path, and its docstring records why the round-trip is not surplus.

Worth noting the cost is narrower than three-per-scoped-read: only `scope='fleet'` reaches the gate's `get_agent` at all, and `enforce_fleet_read` only runs when a `fleet_id` is actually present.

**CI fix (not from the review).** `GET /memories` is one of the eight frozen broker operations, so `core-api/openapi.broker.json` was stale. Regenerated. Every added param is optional, and I ran the same pinned `tufin/oasdiff` digest CI uses against main's baseline locally: *"No breaking changes to report"*.

Gates: ruff check + ruff format (separately), mypy clean at 199 files, broker baseline current, and the full `tests/` run is 4651 passed with 16 pre-existing failures that reproduce identically on unmodified `origin/main` (all search/FTS/entity-relation, none touching list or stats).

Eldad-Caura:
Correcting the suite numbers in my previous comment, and one real thing they were hiding.

**The bug the recount found.** `tests/test_cross_tenant_audit_surfaces.py` calls `list_memories()` **directly as a function**, bypassing FastAPI's dependency resolution — so any parameter it doesn't pass arrives as the `Query(...)` default *object*, not the default value. My `weight_min > weight_max` check then compared two `Query` instances:

```
TypeError: '>' not supported between instances of 'Query' and 'Query'
```

Fixed by passing the four new params explicitly at that call site, with a comment recording why every param must be enumerated there. It is the only direct-call site of either route in the suite (`memory_stats` has none). Worth flagging as a live example of the hazard: `scope`'s `Query` default is also `is not None`, so a direct call would have taken the scope branch — the same trap with a quieter failure.

**The numbers, measured rather than inferred.** I wrote "4651 passed with 16 pre-existing failures". That was arithmetic on an earlier run, not a measurement, and it was wrong in both halves. Back-to-back full runs, same DB, `-p no:randomly`:

| | passed | failed |
|---|---|---|
| unmodified `origin/main` | 4627 | 17 |
| this branch | 4650 | 17 |

The failing **sets are identical** — `comm` in both directions is empty — so no test that passes on main fails here. 4650 − 4627 = 23, exactly the new tests. All 17 are search/FTS/entity-relation and reproduce on unmodified main; one of them is flaky run-to-run, which is what produced my "16".

The other gates are unchanged: ruff check + ruff format (separately), mypy clean at 199 files, broker baseline current, oasdiff reports no breaking change.

Eldad-Caura:
@erni-a ready for review — all gates green.

Auto-review verdict was **"No issues found"** with two non-blocking observations: one applied (inverted weight range now 422), one declined with evidence (the three agent lookups aren't redundant — one is a documented TOCTOU re-read, the other is the only thing refusing `scope=agent&fleet_id=<another fleet>`, now pinned by a test).

Closes #820.

**What to look at, if you look at one thing:** `scope` is opt-in on both routes. Omitted, `GET /memories` and `GET /memories/stats` behave exactly as before — that's what keeps the dashboard and every existing caller unaffected, and no caller can regress on params that were being silently discarded until this commit.

**The judgement call worth a second opinion** is in `memory_stats`: `agent_id` there is *both* the author filter and the visibility identity, so I could not copy the fix `GET /memories` uses (prefer the authenticated agent, ignore the param) — that would have silently narrowed the historical "no `agent_id` → team/org-wide" aggregate. I rejected the conflict instead. Legitimate calls (omitted, or naming yourself) are untouched, and both are covered by tests.

Signals:

- Broker contract: baseline regenerated (`GET /memories` is one of the eight frozen ops); every added param optional; `oasdiff breaking` vs main — *"No breaking changes to report"*.
- Full suite: **4650 passed** vs **4627** on unmodified `origin/main`, difference = the 23 new tests. 17 failures on both, identical sets — all pre-existing local search/FTS failures.
- 23 tests in `tests/test_api_read_scope_params.py`; 17 confirmed failing with the source reverted, and the six that don't discriminate say so in their own docstrings.

Not merging — that's yours.

erni-a (approved):

PR #848, opened 2026-08-21
fix(embedding): the concurrency gate must cover every caller

Eldad-Caura:
`EMBEDDING_MAX_CONCURRENCY` only bounds the embedding backend if every caller passes through the semaphore. **One did not.**

core-worker's deferred-embed consumer resolves the platform singleton and calls `provider.embed` on it — a path that never traverses the gate `get_embedding` / `get_embeddings_batch` / `get_query_embedding` apply. A control experiment makes it unambiguous:

| | peak in-flight | gate entries |
|---|---|---|
| core-worker `provider.embed()` | **10** | **0** |
| core-api `get_embedding()` (control) | 2 | 10 |

*(cap forced to 2, 10 concurrent callers, identical stub on both arms)*

## Why it hasn't hurt yet — and why that's not reassuring

`PubSubEventBus._pull_loop` awaits `_dispatch_all` once per message inside a `for`, so a batch drains **sequentially** and one instance holds roughly one embed at a time. That is a property of the loop's shape, not a limit anyone chose. Parallelising it for drain throughput reads as a pure win while removing the only thing bounding this path.

**And the code claimed the bound was real.** `max_messages`' comment said it *"Caps `workers x max_messages` concurrent embed calls across the deployed pool"* — a cap that never existed, sitting on the exact knob someone optimising that loop would read first. `per_tenant_concurrency`'s docstring inherited the same batch-implies-fan-out premise. Both corrected here.

## What changed

- **`call_embedding_gated`** — the gate made public (was `_call_gated`), *renamed rather than aliased* so the module has one name for one behaviour and the never-nest warning lives on the function everything actually calls.
- **core-worker takes it**, `background=True`.
- **A gate timeout propagates and nacks** rather than acking a memory embedding-less. Saturation is transient, so Pub/Sub redelivery is the right backpressure for a queue consumer — unlike core-api, which degrades to `None` because an HTTP caller is waiting.

### Why the gate is at the call site, not inside the provider

Pushing it into the provider would close the hole for all callers without anyone remembering — but it deadlocks. With `EMBEDDING_PROVIDER=openai` and no per-tenant key, `_registry` hands back **the platform singleton itself**, so `get_embedding` → `call_embedding_gated` → `provider.embed` → gate #2. The outer hold occupies a shared slot while the inner acquire waits for one; once `EMBEDDING_MAX_CONCURRENCY` callers are inside, none proceed and each burns the full timeout. (`OpenAIEmbeddingProvider.embed_query` also calls `self.embed`, nesting again.) Documented on the function.

## Tests — both confirmed failing without the fix

1. **Unit** — 8 concurrent `handle_embed_request` against a background cap of 2 never exceed 2 in flight. Ungated it reports `peak in-flight 8 exceeded the gate cap 2`.
2. **AST guard** (`tests/test_embedding_calls_are_gated.py`) — fails on any ungated `.embed()` outside `common/embedding`, naming the offender (`consumer.py:188: .embed()`). A reviewer cannot see that a new `.embed()` is ungated; this can. #830 spent six review cycles, five finding a *different* inherited call site, learning to make this class of mistake structurally impossible rather than auditing for it. Mirrors the existing `test_no_direct_db_outside_storage.py`.

## Honest limits

- **`background=True` buys this process nothing.** The gate is per process and core-worker has no interactive embeds, so the reserved interactive slots are unusable there and the worker is capped at the background budget instead of the full one. It's still right for the resource that *is* shared — the backend pool core-api's search also hits. The comment says this outright rather than implying a local benefit.
- **The gate is not per tenant.** One tenant's backfill flood can hold the whole background budget. core-api wraps its embed in a per-tenant slot; core-worker has no equivalent. Out of scope, now stated in the code.
- **Currently inert in production.** Real per-instance embed concurrency is ~1, so this is preventive. The test manufactures concurrency the production path can't yet produce — don't read it as evidence the gate binds today.

## Verification

- `core-worker/tests/` **72 passed** (baseline on unmodified `main`: 71)
- `tests/test_embedding_concurrency_gate.py` **10 passed** after the 12-reference rename; new guard passes; `test_no_direct_db_outside_storage.py` unaffected
- `ruff check` clean on all 6 CI scopes; `ruff format --check` clean on all 5 plus the `--isolated` 88-col vendored gate
- Both new tests verified to **fail** with the fix reverted

### DIFF
--- a/common/embedding/__init__.py
+++ b/common/embedding/__init__.py
@@ -13,6 +13,10 @@ Public surface:
   falls back to :func:`get_embedding`'s symmetric behaviour. Backwards-
   compatible with the prior ``core_api.services.embedding`` module.
 * :func:`get_embedding_provider` — factory.
+* :func:`call_embedding_gated` — runs a caller-supplied embed under the
+  process-wide concurrency gate, for the one caller that holds a provider
+  directly instead of going through the entrypoints above. Never nest it
+  inside them; see its docstring for why that deadlocks.
 * :func:`init_platform_embedding` / :func:`get_platform_embedding` —
   platform-tier singleton, initialised once at service startup from
   ``PLATFORM_EMBEDDING_*`` env vars.
@@ -39,6 +43,7 @@ from common.embedding._platform import (
 )
 from common.embedding._registry import get_embedding_provider
 from common.embedding._service import (
+    call_embedding_gated,
     get_embedding,
     get_embeddings_batch,
     get_query_embedding,
@@ -53,6 +58,7 @@ __all__ = [
     "EmbeddingProvider",
     "FakeEmbeddingProvider",
     "InstructionAwareEmbedder",
+    "call_embedding_gated",
     "fake_embedding",
     "get_embedding",
     "get_embedding_provider",
--- a/common/embedding/_service.py
+++ b/common/embedding/_service.py
@@ -281,7 +281,7 @@ def _background_gate() -> asyncio.Semaphore:
     return _bg_gate
 
 
-async def _call_gated[T](
+async def call_embedding_gated[T](
     make_call: Callable[[], Awaitable[T]], *, background: bool
 ) -> T:
     """Run *make_call* holding a concurrency slot.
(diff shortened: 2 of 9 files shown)


Eldad-Caura:
Thanks — declining the one finding, with evidence, because applying it would **break CI**.

**`consumer.py:188` overlong statement — not applying.**

The suggestion is to wrap the call across three lines. I originally wrote it exactly that way; `ruff format` collapsed it. Verified just now:

- `line-length = 110` (root `pyproject.toml:68`, inherited by core-worker)
- the line is **104 characters** — under the limit
- `ruff check core-worker/src/core_worker/consumer.py` → `All checks passed!` (no E501)

And applying the wrap actively fails the formatter gate:

```
$ ruff format --check core-worker/src/ core-worker/tests/
1 file would be reformatted, 19 files already formatted

$ ruff format --check --diff core-worker/src/core_worker/consumer.py
-        embedding = await call_embedding_gated(
-            lambda: provider.embed(request.content), background=True
-        )
+        embedding = await call_embedding_gated(lambda: provider.embed(request.content), background=True)
```

So the wrapped form is what `Run Ruff formatter check (core-worker)` would reject. The premise "likely exceeding the project's configured line-length lint rule" is the part that doesn't hold — 104 < 110.

The comparison to `get_embeddings_batch` in `common/embedding/_service.py` is also not quite like-for-like: `common/` has its own `ruff.toml` at `line-length = 88`, so identical code formats differently in the two trees. Same reason those calls are wrapped there and this one isn't here.

Leaving the line as-is. Readability point is fair in the abstract, but it's the formatter's call in this tree, and CI enforces it.

Eldad-Caura:
@erni-a

erni-a (approved):

PR #849, opened 2026-08-21
fix(contradictions): a detection lock must not outlive the run it guards

### LINKED ISSUE #816: H-06: Contradiction idempotency locks (1h TTL, never released) suppress the documented re-detection after updates
Finding **H-06** of the 2026-08-14 OSS/platform engineering audit, rated **high**. Filed now so the ten correctness highs are trackable — they previously existed only inside the audit artifact. The eight security highs from the same audit are all fixed and merged (#799, #801, #802, #804, #806).

**Area:** core-api · **Surfaced by:** bugs-core-services

## Verified still present

Both locks are still set via `cache_set_nx` (`contradiction_detector.py:50,57`) and there is **no `cache_delete`** anywhere in that module — the keys are never released.

## Mechanism

detect_contradictions_async acquires `contradiction:path_a:{memory_id}` with SET NX and a 3600s TTL (lines 39, 42-50) and returns immediately if the key exists (lines 192-194). The key is never deleted — grep shows no cache_delete of `contradiction:path_a`/`contradiction:path_c` anywhere — and it is keyed on memory_id alone, not on the triggering event or content hash. Consequence 1 (realistic, tenant-visible): update_memory explicitly clears stale supersession state and re-fires detection on content change ("P1-2: Re-check contradictions after content update", memory_service.py:2833-2850, Trigger.UPDATE with the new embedding), but if the memory was written (and detected) less than an hour earlier — the normal case for a quick correction — the lock is still held from the write-time detection, so the UPDATE-triggered run exits at line 192 with skipped=True and the edited content is never checked; since update_memory also reset status to 'active' and supersedes_id to None (memory_service.py:2663-2666), a memory edited into direct contradiction with another active memory stays unflagged, and no later trigger re-fires (updates neither publish EMBEDDED/ENRICHED events nor schedule re-embeds when the inline embed succeeds). The identical problem exists on Path C: update-triggered entity re-extraction fires Trigger.ENTITY (entity_extraction_worker.py:443) which detect_contradictions_by_entities_async skips via _acquire_path_c_lock (line 1878). Consequence 2: the lock is acquired BEFORE detection runs and kept even when detection throws (the `except Exception` at 212-213 does not release it), so one transient LLM/storage failure blocks all re-detection for that memory for a full hour with no retry. The lock's documented purpose is only deduping back-to-back ENRICHED/EMBEDDED deliveries. Requires Redis to be configured (cache_set_nx fails open without it), which is the SaaS deployment shape.

## Suggested fix

Release the lock in a finally block after detection completes (or on failure), and/or key the lock on (memory_id, content_hash or trigger-generation) so an UPDATE with new content is not deduped against the write-time run.

---

*Location cited by the audit: `core-api/src/core_api/services/contradiction_detector.py:192`. Description and suggested fix are the audit's own text; the verification section above is a fresh check.*

Eldad-Caura:
Fixes #816.

## The bug

The A4 #14 idempotency lock exists for one job: the ENRICHED and
EMBEDDED back-channels both fire Path A for the same memory, and
whichever loses the race should skip. It was keyed on `memory_id`
alone, taken *before* detection, never released, with a 1h TTL. Two
consequences follow, and they are independent.

**An edit inside the hour was never checked.** `update_memory` clears
supersession state and re-fires detection on a content change — the
documented "P1-2: Re-check contradictions after content update". But a
memory corrected shortly after it was written, which is the normal
shape of a quick correction, still held the write-time lock, so the
UPDATE run exited at the lock and the edited text was never examined.
The same update resets `status` to `active` and `supersedes_id` to
`None`, so a memory edited into direct contradiction with another
active memory stays unflagged — and nothing re-fires later, because an
update publishes no EMBEDDED/ENRICHED event and schedules no re-embed
when the inline embed succeeds. Path C has it too, via the entity
re-extraction an update triggers.

**One transient failure blocked re-detection for an hour.** The lock is
taken before detection and was kept when detection threw. A single LLM
or storage blip suppressed every later trigger for that memory, with
nothing scheduled to retry.

## The fix, and the part that is easy to get wrong

Both locks are now keyed on `(memory_id, content fingerprint)`, so an
edit is simply a different lock, and released whenever a run did not
reach a verdict.

**Releasing in a `finally` — the issue's first suggestion, read
literally — would have destroyed the deduplication the lock exists
for.** The two back-channel deliveries are frequently further apart
than one detection takes, so releasing on success just lets the second
delivery re-run every LLM judgement. A run that concluded therefore
KEEPS its lock for the full TTL; only the content key lets a genuine
edit through. `test_a_completed_run_keeps_its_lock` pins that so the
literal reading cannot be reintroduced.

Content is a safe key here because enrichment does not write it — the
routing table covers metadata only (`summary`, `tags`,
`retrieval_hint`, `contains_pii`), so both back-channels publish the
same string for a given version and still collapse to one run.

**"Concluded" is not "reached the end of the function".** Path C has
three early returns after the lock — no candidates, and two preflights
that can drop every candidate — which are *results*, not failures, and
the commonest outcomes at that. Marking completion once, early, would
have kept the lock on a throw in the judging loop, leaving half of the
second consequence unfixed; marking it once, late, would have released
the lock on the most common path of all and destroyed the dedup
entirely. Each legitimate exit is marked instead.
`test_path_c_releases_the_lock_when_it_throws_mid_detection` fails
under the early-marker variant; the Path C dedup tests fail under the
late one.

**Both paths fingerprint the row they will actually examine, not the
caller's copy of it.** `_detect` reads `new_memory`; the `content`
parameter is only what the caller believed the content was, and the
back-channel consumers pass the write-time payload, which a mid-flight
update leaves stale. Keying on that would put a lock on text nobody
looked at, and would make the fix depend on every future caller keeping
the two in step. So both paths now resolve the row first and key on it,
which also means a soft-deleted row never takes a lock at all.

The cost is one `get_memory` on a duplicate delivery that would
previously have been refused at the lock. It falls only on callers that
do not already pass `new_memory` — and the duplicate-heavy path, the two
back-channel consumers, does pass it. Against the alternative branch,
which runs a candidate search plus an LLM judgement per candidate, a
single-row GET is the cheaper half by a wide margin.

## Ownership on the release

Adding a release added a way to lose the lock to someone else, so the
release is compare-and-delete on a per-acquisition token, never a bare
`DEL`. A run whose work outruns the 3600s TTL no longer owns the key —
by then it may have expired and been re-taken — and deleting it there
would hand a SUCCESSOR a duplicate detection, which is worse than the
bug being fixed. `cache_delete_if` runs the compare and the delete as
one server-side step; GET-then-DEL from the client would leave exactly
the window it is meant to close.
`test_a_late_release_cannot_take_a_successor_s_lock` expires the key
mid-run, lets a second run take it, and fails if the first run's
`finally` removes it.

## Scope

Both arches are covered: `ContradictionEngine` is a facade that
delegates to these same two functions, so the flag makes no difference.
`cache_delete_if` is new; like its neighbours it is total (a missing
Redis or a failed EVAL returns False rather than raising), which is what
keeps the never-raises contract the Pub/Sub consumers depend on to ack.
`test_cache_delete_if_is_total` pins that, since the release runs inside
a `finally`, and a second test pins that a non-matching value is not
deleted and that the compare+delete is a single round trip.

## Tests

`tests/test_h06_contradiction_lock.py`, 14 tests. All of them fail on
`origin/main` with the source reverted, but that number overstates the
signal: several fail only because the symbols they reference do not
exist there, not on behaviour.

The evidence that they discriminate is separate, and stronger — each of
the six designs rejected along the way was actually applied, and in
every case exactly the one test guarding it failed while the rest
stayed green:

| rejected design | test that caught it |
|---|---|
| release on success (`finally`, unconditional) | `test_a_completed_run_keeps_its_lock` |
| mark the run concluded once, early | `..._throws_mid_detection` |
| key on the caller's `content` parameter | `..._not_the_caller_copy` |
| release with a bare `DEL`, no ownership | `..._successor_s_lock` |
| let a null content reach the fingerprint | `..._still_gets_a_usable_key` |
| treat an empty row content as a missing one | `..._is_a_real_value_not_a_missing_one` |

Locks are modelled with an in-memory SETNX / compare-and-delete store
rather than by patching the `_acquire_*` helpers, so the assertions run
against the real key derivation and the real ownership check — the two
things that changed.

`tests/test_a4_14_back_channel_idempotency.py` is updated for the new
two-argument helpers, and its Path C skip test's docstring corrected:
that path does now fetch the row before the lock.

### DIFF
--- a/core-api/src/core_api/cache.py
+++ b/core-api/src/core_api/cache.py
@@ -78,6 +78,42 @@ async def cache_delete(key: str) -> bool:
         return False
 
 
+# Release-if-owner. GET-then-DEL from the client would be two round trips with a
+# window in between, which is the whole bug this guards against; the compare and
+# the delete have to be one atomic step, so they run server-side.
+_DELETE_IF_SCRIPT = """
+if redis.call('get', KEYS[1]) == ARGV[1] then
+    return redis.call('del', KEYS[1])
+end
+return 0
+"""
+
+
+async def cache_delete_if(key: str, expected: str) -> bool:
+    """Delete ``key`` only while it still holds ``expected``. Returns True iff
+    this call deleted it.
+
+    For releasing a ``cache_set_nx`` lock. An unconditional ``DEL`` is the
+    classic distributed-lock footgun: a holder whose work outruns the lock's
+    TTL no longer owns the key, and deleting it then destroys a SUCCESSOR's
+    lock rather than its own. Pairing SET NX with a per-acquisition token and
+    releasing through this compare-and-delete makes a release a no-op once the
+    lock has moved on.
+
+    Same fail-quiet contract as its neighbours: no Redis, or a failed EVAL,
+    returns False rather than raising. Callers release inside ``finally``
+    blocks on paths whose never-raises property is load-bearing.
+    """
+    r = await _get_redis()
+    if r is None:
+        return False
+    try:
+        return bool(await r.eval(_DELETE_IF_SCRIPT, 1, key, expected))
+    except Exception:
+        logger.debug("cache_delete_if failed for key=%s", key, exc_info=True)
+        return False
+
+
 async def cache_set_nx(key: str, value: str, ttl: int) -> bool:
     """Atomic SET-if-not-exists with TTL. Returns True iff we newly set
     the key (i.e. the caller "owns" the lock); False if the key already
--- a/core-api/src/core_api/services/contradiction_detector.py
+++ b/core-api/src/core_api/services/contradiction_detector.py
@@ -13,13 +13,14 @@ Multi-provider support:
 """
 
 import asyncio
+import hashlib
 import logging
 import time
 import uuid as _uuid
 from datetime import datetime
 from uuid import UUID
 
-from core_api.cache import cache_set_nx
+from core_api.cache import cache_delete_if, cache_set_nx
 from core_api.clients.storage_client import get_storage_client
 from core_api.config import settings
 from core_api.constants import SINGLE_VALUE_PREDICATES
@@ -41,22 +42,82 @@ logger = logging.getLogger(__name__)
 _DETECTION_LOCK_TTL_SECONDS = 3600
 
 
-async def _acquire_path_a_lock(memory_id) -> bool:
+def _content_fingerprint(content: str) -> str:
+    """Short stable digest of the text a detection run examined.
+
+    Part of the lock key (H-06). Keyed on ``memory_id`` alone, the lock also
+    deduped runs that were checking DIFFERENT text: ``update_memory`` clears
+    supersession state and re-fires detection on a content edit, but a memory
+    corrected within the TTL — the normal case for a quick fix — still held the
+    write-time lock, so the edited content was never checked and no later
+    trigger re-fired. Including the content makes an edit a different lock.
+
+    Hashed rather than embedded: memory content is unbounded and Redis keys are
+    values. Truncated because this only has to separate versions of ONE
+    memory's text, not resist collision across a corpus.
+    """
+    return hashlib.sha256(content.encode("utf-8")).hexdigest()[:32]
+
+
+def _path_a_lock_key(memory_id, content: str) -> str:
+    return f"contradiction:path_a:{memory_id}:{_content_fingerprint(content)}"
+
+
+def _path_c_lock_key(memory_id, content: str) -> str:
+    return f"contradiction:path_c:{memory_id}:{_content_fingerprint(content)}"
+
+
+def _lock_token() -> str:
+    """A fresh per-acquisition owner token — see ``_release_lock``."""
+    return _uuid.uuid4().hex
+
+
+async def _acquire_path_a_lock(memory_id, content: str, token: str) -> bool:
     """Try to acquire the Path A (semantic + RDF) idempotency lock for
-    ``memory_id``. Returns True iff this caller owns the lock and should
-    proceed; False if another caller already holds it and we should skip.
+    ``memory_id`` at this ``content``. Returns True iff this caller owns the
+    lock and should proceed; False if another caller already holds it and we
+    should skip.
 
     Lock is keyed per-path so Path A and Path C run independently for
-    the same memory.
+    the same memory, and per-content so an edit is not deduped against the
+    run that checked the previous text. ``token`` identifies THIS acquisition
+    so the release can tell its own lock from a successor's.
     """
-    return await cache_set_nx(f"contradiction:path_a:{memory_id}", "1", _DETECTION_LOCK_TTL_SECONDS)
+    return await cache_set_nx(_path_a_lock_key(memory_id, content), token, _DETECTION_LOCK_TTL_SECONDS)
 
 
-async def _acquire_path_c_lock(memory_id) -> bool:
+async def _acquire_path_c_lock(memory_id, content: str, token: str) -> bool:
     """Try to acquire the Path C (entity-overlap) idempotency lock for
-    ``memory_id``. Returns True iff this caller owns the lock; False
-    means another caller already ran detection for this memory."""
-    return await cache_set_nx(f"contradiction:path_c:{memory_id}", "1", _DETECTION_LOCK_TTL_SECONDS)
+    ``memory_id`` at this ``content``. Returns True iff this caller owns the
+    lock; False means another caller already ran detection for this memory at
+    this content."""
+    return await cache_set_nx(_path_c_lock_key(memory_id, content), token, _DETECTION_LOCK_TTL_SECONDS)
+
+
+async def _release_lock(key: str, token: str) -> None:
+    """Drop a detection lock so the next trigger can retry.
+
+    Released ONLY when a run did not reach a verdict — it threw, or it exited
+    before examining anything (a missing / soft-deleted row). A run that
+    concluded keeps its lock for the full TTL, INCLUDING the "no candidates"
+    and "preflight dropped them all" outcomes, which are results rather than
+    failures. Collapsing the back-to-back ENRICHED and EMBEDDED deliveries is
+    the entire point of the lock: releasing on success would let the second
+    delivery re-run every LLM judgement whenever it landed more than one
+    detection apart from the first.
+
+    Compare-and-delete on ``token``, never a bare DEL. A run whose work outran
+    the 3600s TTL no longer owns the key: by then the lock may have expired and
+    been re-taken by a NEW run, and an unconditional delete would take that
+    successor's lock away and hand it a duplicate detection — the exact
+    guarantee this lock exists to provide. Releasing by token makes a late
+    release a no-op instead.
+
+    ``cache_delete_if`` is total — it swallows both a missing Redis and a
+    failed EVAL — so this cannot break the never-raises contract that lets the
+    Pub/Sub consumers ack unconditionally.
+    """
+    await cache_delete_if(key, token)
 
 
 def _parse_dt(value) -> datetime | None:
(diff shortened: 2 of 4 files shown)


Eldad-Caura:
Both points checked. The Suggestion was right and is applied; the Low is declined, because acting on it would undo the Suggestion.

**Applied — key on the row that will be examined, not the caller's copy.** You're right, and it's sharper than "worth a comment": `content` was being read by *nothing* in `detect_contradictions_async` except my two lock lines — `_detect` reads `new_memory`. So the parameter had quietly become the lock key and nothing else, while the text actually examined came from somewhere else.

The divergence is reachable, not just theoretical: both back-channel consumers pass `content=payload.content` (the write-time payload) together with `new_memory=<row fetched now>`, so an update landing between the write and the delivery makes them disagree.

Rather than document the invariant, I removed it. Both paths now resolve the row first and fingerprint `new_memory["content"]`, with the parameter as a fallback for a row with no content. Path A therefore mirrors Path C structurally, and a soft-deleted row no longer takes a lock at all — the "gone row" case is now resolved before the acquire rather than released after it. `test_path_a_keys_on_the_row_that_will_be_examined_not_the_caller_copy` fails if the key goes back to the parameter.

**Declined — passing content into Path C from the entity-extraction caller.** The premise is accurate (the fetch did move ahead of the lock, and duplicate deliveries now pay a GET they previously skipped), but the proposed remedy reintroduces exactly the defect your Suggestion asks me to remove. `process_entity_extraction` does hold a `content`, but it is the copy handed to it when extraction was *scheduled*; Path C detects against the row it fetches. An update landing in between would have it lock a fingerprint for text it did not look at — caller-copy keying, one path over.

On the cost itself:

- It falls only on callers that don't already pass `new_memory`. The duplicate-heavy path — the two back-channel consumers — passes it, so Path A adds no GET there at all.
- What it buys is a single-row `get_memory`. What it guards is `find_entity_overlap_candidates` plus an LLM judgement per surviving candidate. On a retry storm the skipped branch is the cheap one by orders of magnitude, and it is precisely the branch that gets cheaper the more duplicates there are.
- The observed duplicate rate is the one recorded in `test_a4_14`'s header — Path C completing about twice per memory.

If storage-api load ever does become the binding constraint here, the answer is a cheap `memory_id`-only in-flight lock in front of the content-keyed one, not trusting a caller's copy. That's more moving parts than this bug justifies today.

**Also in this push:** rebased onto `main` (#848 landed). CI was green on the previous head; re-running.

Gates: 20 tests green across the two lock files; full suite 4661 passed with 16 pre-existing failures whose set is identical to unmodified `main`; ruff check + format separately; mypy clean at 199 files.

Eldad-Caura:
@claude

Eldad-Caura:
Round 2: both applied. The Medium was a bug I introduced with the fix, and it deserved the re-tag.

**Applied — fencing token on the release.** Correct, and the important part is that this hazard did not exist before this PR: nothing ever released, so nothing could release someone else's lock. Adding a release added the way to lose one.

`_acquire_*` now sets a per-acquisition uuid instead of the literal `"1"`, and `_release_lock` compares-and-deletes on it via a new `cache_delete_if`. That runs as a single Lua `EVAL` rather than GET-then-DEL from the client: a client-side compare would leave exactly the window it is meant to close.

`test_a_late_release_cannot_take_a_successor_s_lock` reproduces the scenario directly — it expires the key mid-run, lets a second run acquire it, then lets the first run's `finally` fire. With the ownership check the successor's lock survives; with a bare `DEL` it does not, and I confirmed that by simulating the bare `DEL` in the lock model: exactly that one test failed, the other twelve stayed green.

`cache_delete_if` keeps the fail-quiet contract of its neighbours (missing Redis or failed EVAL → `False`, never raises), which is what preserves the never-raises property the Pub/Sub consumers ack on. Two tests pin it: totality, and that a non-matching value is not deleted.

**Applied — the null-content hole.** `examined = new_memory.get("content") or content` is now `... or ""`. You're right about the severity being in the silence rather than the crash: the `AttributeError` lands in the outer handler and the memory just never gets checked on that trigger. `test_a_row_with_no_content_still_gets_a_usable_key` fails without the coercion.

**Correction to my last comment.** I wrote that 9 of 10 tests fail on `origin/main`. With the new tests it is 13 of 13 — but that number overstates the signal, and did before too: several fail there only because the symbols they reference don't exist yet, not on behaviour. The real evidence is that each rejected design was applied and caught by exactly one test:

| rejected design | test that caught it |
|---|---|
| release on success (`finally`, unconditional) | `test_a_completed_run_keeps_its_lock` |
| mark the run concluded once, early | `..._throws_mid_detection` |
| key on the caller's `content` parameter | `..._not_the_caller_copy` |
| release with a bare `DEL`, no ownership | `..._successor_s_lock` |
| let a null content reach the fingerprint | `..._still_gets_a_usable_key` |

Also rebased onto `main` again (#850 landed).

Gates: 23 tests green across the two lock files; full suite 4665 passed, 16 pre-existing failures whose set is identical to unmodified `main`; ruff check + format separately; mypy clean at 199 files.

Eldad-Caura:
@claude

Eldad-Caura:
Round 3: applied. Good catch — my round-2 fix reintroduced the divergence round 1 removed.

`examined = new_memory.get("content") or content or ""` treats an empty-but-present `content` as a missing one and falls through to the caller's copy. Confirmed against the consumer: `_detect` reads `new_memory.get("content", "")` (`contradiction_detector.py:387`), so for an empty row it examines `""` while the lock would have been keyed on the caller's stale string — exactly the "fingerprint the text detection will ACTUALLY examine" guarantee the comment above the line claims.

Now `is None`, per your suggestion:

```python
raw_content = new_memory.get("content")
examined = raw_content if raw_content is not None else (content or "")
```

`test_an_empty_row_content_is_a_real_value_not_a_missing_one` writes a row with `content=""`, passes a different non-empty string as the caller's `content`, and asserts the acquired key is the fingerprint of `""`. Putting the `or` chain back fails exactly that one test and nothing else.

Path C needs no equivalent change: it has no caller fallback, so its `or ""` collapses `None` and `""` to the same `""` that `_detect` will read.

That makes six rejected designs now, each applied and each caught by exactly one test:

| rejected design | test that caught it |
|---|---|
| release on success (`finally`, unconditional) | `test_a_completed_run_keeps_its_lock` |
| mark the run concluded once, early | `..._throws_mid_detection` |
| key on the caller's `content` parameter | `..._not_the_caller_copy` |
| release with a bare `DEL`, no ownership | `..._successor_s_lock` |
| let a null content reach the fingerprint | `..._still_gets_a_usable_key` |
| treat an empty row content as a missing one | `..._is_a_real_value_not_a_missing_one` |

Gates: 24 tests green across the two lock files; full suite 4670 passed, 16 pre-existing failures whose set is identical to unmodified `main`; ruff check + format separately; mypy clean at 199 files.

Eldad-Caura:
@claude

Eldad-Caura:
@erni-a ready for review — verdict clean, all gates green.

Closes #816. Four review rounds; the last came back **"No issues found"**, with one remaining observation (the extra Path C `get_memory` on a duplicate delivery) that I declined in round 1 and that the reviewer agrees is "not a defect".

**Worth knowing before you look:** the fix itself — key the lock on content, release only when a run didn't reach a verdict — was unchanged from the first push. All three earlier findings were in the machinery around it, and one of them was a bug the fix *introduced*: the release started as a bare `DEL`, which would let a run outliving the 1h TTL delete a **successor's** lock. That's now a compare-and-delete on a per-acquisition token, via a new `cache_delete_if` that runs the compare and the delete in one Lua `EVAL`.

**The judgement call worth a second opinion** is what "concluded" means. A run that reaches a verdict keeps its lock for the full TTL — releasing on success would destroy the deduplication the lock exists for, since the two back-channel deliveries are frequently further apart than one detection takes. Path C's three "no candidates / preflight dropped everything" exits are therefore *conclusions*, not failures, and are marked as such individually rather than by one marker at the top or bottom of the body. Both of the obvious simplifications there are wrong in opposite directions, and each has a test.

Signals:

- Full suite **4670 passed**; 16 failures, set identical to unmodified `main` (all pre-existing local search/FTS).
- 24 tests across the two lock files. Six rejected designs were each applied and caught by exactly one test — table in the PR body.
- Rebased onto `main` twice as #848 and #850 landed.

Not merging — that's yours.

erni-a (approved):

PR #850, opened 2026-08-21
fix(embedding): a gate timeout is capacity, not a provider failure

Eldad-Caura:
Follow-up to #848. Same theme — making the gate mean what it says — but an independent behaviour.

## The amplifier

The gate raised a bare `TimeoutError`, which `_run_with_retry`'s broad `except Exception` cannot distinguish from "the backend returned an error". So a caller that merely **queued** got retried — and each retry queues another waiter for another full `EMBEDDING_GATE_TIMEOUT_SECONDS`.

**The retry path fed the saturation it was reacting to**, and the caller waited ~11s (5 + 1 backoff + 5) to be told what attempt one already knew.

### Measured, not theorised

On staging under a 1400-item/tenant write burst:

| | |
|---|---|
| Peak gate timeouts | **835/min** (23:16), **906/min** (23:18) |
| `Embedding attempt 2/2 failed` | **outnumbered** `attempt 1/2 failed` |

Attempt 2 failing *more* than attempt 1 is what you expect when the retry is part of the problem rather than a second chance at it. *(That sample was `--limit`-truncated — treat the ordering as the signal, not the ratio.)*

Alongside it, ~94 `Embedding service degraded [openai model=BAAI/bge-m3]: N consecutive failures` — while the backend was serving at **~7ms**. The backend was fine; we were queueing on ourselves and then blaming it.

## The fix

`EmbeddingGateTimeout` carries the distinction. It **subclasses `TimeoutError`**, so every existing handler is unaffected — a gate timeout has always surfaced as one (`parallel_embed_entity_boost`, `parallel_embed_enrich`), and callers that only need "no vector" need no change. Only code that must tell saturation from failure catches it.

- **`_run_with_retry` returns on the first one** instead of retrying. Nothing is lost: the provider never saw the call, so there's no partial work to reconcile, and `None` still means what it meant — NULL for the backfill sweep, or a 503 upstream. It just arrives in ~5s, without a second slot-wait.
- **Neither `_run_with_retry` nor `get_embeddings_batch` records it as a provider failure.** That streak drives the "Embedding service degraded" ERROR an operator reads to decide the backend is unhealthy.
- **The gate's warning stopped claiming otherwise.** It said *"degrading as provider failure"* — accurate about the old behaviour, now the opposite of what happens.

## Why not just widen the retry backoff?

Backoff doesn't help: the cost isn't the delay between attempts, it's that each attempt **occupies a waiter slot for the full timeout**. More attempts = more concurrent waiters on an already-saturated gate, regardless of spacing.

## Tests — all three confirmed failing without the fix

1. A gate timeout is **attempted once** (reverted: `gate timeout was retried 2 times`)
2. A genuine provider error is **still retried** `EMBEDDING_RETRY_ATTEMPTS` times — so the narrowing can't silently disarm retries
3. A gate timeout leaves the degraded streak at **zero**, with a control proving a real fault still advances it (so the assertion is about classification, not a dead counter)

Also corrected the now-stale docstrings in the gate suite that asserted the old "treated as a provider failure" contract.

## Verification

- All 15 `tests/*embed*.py` files: **185 passed, 1 skipped**, including `test_embedding_retry.py` — the file most exposed to a retry-behaviour change
- `ruff check` clean on all 6 CI scopes; `ruff format --check` clean on all 5 plus the `--isolated` 88-col vendored gate
- Rebased onto `main` after #848 merged; its `_call_gated` → `call_embedding_gated` rename is reflected here

## Scope

Does **not** add backoff, jitter, or a circuit breaker. `EMBEDDING_RETRY_DELAY_S` is still linear and jitter-free, and the degraded breaker is still log-only (`last_failure_time` is written and never read). Both are real, both are separate.

### DIFF
--- a/common/embedding/__init__.py
+++ b/common/embedding/__init__.py
@@ -17,6 +17,10 @@ Public surface:
   process-wide concurrency gate, for the one caller that holds a provider
   directly instead of going through the entrypoints above. Never nest it
   inside them; see its docstring for why that deadlocks.
+* :class:`EmbeddingGateTimeout` — raised when the concurrency gate, not the
+  backend, is what stopped an embed. A ``TimeoutError`` subclass, so
+  existing handlers are unaffected; catch it only to tell saturation apart
+  from provider failure.
 * :func:`init_platform_embedding` / :func:`get_platform_embedding` —
   platform-tier singleton, initialised once at service startup from
   ``PLATFORM_EMBEDDING_*`` env vars.
@@ -43,6 +47,7 @@ from common.embedding._platform import (
 )
 from common.embedding._registry import get_embedding_provider
 from common.embedding._service import (
+    EmbeddingGateTimeout,
     call_embedding_gated,
     get_embedding,
     get_embeddings_batch,
@@ -55,6 +60,7 @@ from common.embedding.providers.fake import (
 )
 
 __all__ = [
+    "EmbeddingGateTimeout",
     "EmbeddingProvider",
     "FakeEmbeddingProvider",
     "InstructionAwareEmbedder",
--- a/common/embedding/_service.py
+++ b/common/embedding/_service.py
@@ -252,6 +252,24 @@ _bg_gate: asyncio.Semaphore | None = None
 _bg_gate_loop: asyncio.AbstractEventLoop | None = None
 
 
+class EmbeddingGateTimeout(TimeoutError):
+    """No slot freed within ``EMBEDDING_GATE_TIMEOUT_SECONDS``.
+
+    A distinct type because this is a CAPACITY signal, not a provider
+    fault — the provider never saw the call. Conflating the two is not
+    cosmetic: it makes the retry path amplify the very saturation it is
+    reacting to, since every retry queues another waiter for another full
+    gate timeout, and it credits the provider's degraded-streak with
+    failures it did not cause.
+
+    Subclasses :class:`TimeoutError` so the existing handlers keep working
+    unchanged — a gate timeout has always surfaced as one (see
+    ``parallel_embed_entity_boost``, ``parallel_embed_enrich``), and callers
+    that only want "the embed didn't happen" still need no changes. Only
+    code that must tell saturation apart from failure catches this.
+    """
+
+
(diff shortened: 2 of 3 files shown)


Eldad-Caura:
Confirmed and fixed — good catch, this was a real regression I introduced.

**Verified the exact sequence before fixing.** `record_failure()` is reached only by the loop's terminal fallthrough, so:

1. attempt 1 raises a real provider error → `except Exception` logs it, sets `last_exc`, sleeps, continues — **does not** record
2. attempt 2 raises `EmbeddingGateTimeout` → my branch returned `None` immediately
3. the terminal `record_failure()` is never reached, so the genuine attempt-1 failure vanishes

Pre-PR that sequence exhausted the loop and did record it. And as you note, it's the co-occurrence case that matters most: a provider both erroring *and* slow enough to saturate the gate is what an outage looks like — precisely when the degraded signal must not go quiet.

**Fix**: `last_exc` is set *only* by the provider-failure branch, so it is already the exact predicate for "the backend misbehaved earlier in this call" — no new state needed. The timeout branch now records before returning, and says so in the log line:

```python
except EmbeddingGateTimeout:
    logger.warning(..., " (an earlier attempt failed for a provider reason; counting that)"
                        if last_exc is not None else "")
    if last_exc is not None:
        await stats.record_failure()
    return None
```

The timeout *itself* is still not counted — that part was the point of the PR.

**New test** pins your scenario directly: `test_earlier_provider_failure_survives_a_later_gate_timeout` — attempt 1 `RuntimeError`, attempt 2 `EmbeddingGateTimeout`, asserts `consecutive_failures == 1` and that attempt 2 happened at all. Confirmed failing with the guard removed:

```
AssertionError: the real provider failure from attempt 1 was dropped when attempt 2 hit the gate
```

Amended and force-pushed. Now 4 tests in the class, all verified failing without their fix; `tests/*embed*.py` → **186 passed, 1 skipped**; ruff check + format clean on all CI scopes.

Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #851, opened 2026-08-21
fix(embedding): blank text is a caller error, not a backend outage

Eldad-Caura:
Found while investigating why `prod-memclaw-tei` was returning 413s. It is not a capacity problem, and it rewrites what the 2026-08-18 incident was.

## The evidence

```
openai.APIStatusError: Error code: 413 -
  {'message': 'Input validation error: `inputs` cannot be empty',
   'code': 413, 'type': 'Validation'}
```

| | |
|---|---|
| TEI 413s, 7 days | **274** — 272 inside 2026-08-18 **17:00–18:59Z** |
| user-visible `Embedding service unavailable` | **138** |
| request body size | ~1,509 bytes — nowhere near any size limit |
| blank-text guard before this PR | **none** |

## Three faults compounded on every blank query

1. **The request was sent.** No backend can embed an empty string, so the round trip was always wasted.
2. **It was retried.** The error is deterministic — attempt 2 could only fail identically. That's why 274 requests came from ~136 logical calls.
3. **The `None` surfaced as "Embedding service unavailable"** — 138 times, while the backend answered in **~7 ms**.

The third is the expensive one. **That window was recorded as an embedding *capacity* incident and attributed to gate starvation.** It had no gate timeouts — because the cause was never capacity. The error message sent everyone after a healthy component.

## The fix

`_is_blank` guards `get_embedding` and `get_query_embedding` before the provider is even resolved. It returns `None` — the contract callers already handle — so nothing downstream changes shape. What changes:

- zero backend requests instead of two
- the log names the caller, not the provider
- whitespace counts as blank, not just `""`; a lone space produced the identical 413

The provider-degraded streak is deliberately **untouched**, for the same reason a gate timeout no longer touches it (#850): the backend never saw the call, so counting it there points the operator at the wrong component.

`memory_service` also stops flattening two situations into one error. A blank query and a dead embedder both arrive as `None`; it now tells them apart:

```python
if not query or not query.strip():
    raise ValueError("Search query must not be blank")
raise ValueError("Embedding service unavailable")
```

## Explicitly not changed: `get_embeddings_batch`

A blank element there may or may not poison the whole request. **I could not verify it** — staging TEI runs internal-only ingress, and a control request with *normal* text 404s from outside exactly as a blank one does, so the experiment says nothing either way. I'm not guessing from a broken harness.

Guarding it properly means one vector per input position with `None` for blanks, and the signature is `list[list[float]]` with no slot for that — callers zip results onto rows positionally. That's a contract change and belongs in its own commit.

The confirmed fault is the single-text path: **every 2026-08-18 traceback goes through `embed_query`.**

## Tests

15 tests, **13 fail without the guard**:

- every blank form (`""`, `" "`, `"\t"`, `"\n"`, …) rejected for both entrypoints, asserting the provider was **never awaited**
- real text still embeds — the control that catches a guard swallowing too much
- a blank leaves the degraded streak at zero
- `_is_blank` agrees with the backend on the boundary (`" a "` is not blank)

Full run: `264 passed, 1 skipped` across every `tests/*embed*.py` and `tests/*search*.py`. `ruff check` + `format --check` clean on `common/`, `core-api/src/`, the new test file, and the `--isolated` 88-col vendored gate.

### DIFF
--- a/common/embedding/__init__.py
+++ b/common/embedding/__init__.py
@@ -13,6 +13,10 @@ Public surface:
   falls back to :func:`get_embedding`'s symmetric behaviour. Backwards-
   compatible with the prior ``core_api.services.embedding`` module.
 * :func:`get_embedding_provider` — factory.
+* :func:`is_blank_text` — the predicate the entrypoints above use to reject
+  unembeddable input. Public so callers can apply the SAME definition of
+  blank when they need to answer differently (e.g. a 400 rather than a
+  degraded result) instead of re-implementing it.
 * :func:`call_embedding_gated` — runs a caller-supplied embed under the
   process-wide concurrency gate, for the one caller that holds a provider
   directly instead of going through the entrypoints above. Never nest it
@@ -52,6 +56,7 @@ from common.embedding._service import (
     get_embedding,
     get_embeddings_batch,
     get_query_embedding,
+    is_blank_text,
 )
 from common.embedding.protocols import EmbeddingProvider, InstructionAwareEmbedder
 from common.embedding.providers.fake import (
@@ -72,5 +77,6 @@ __all__ = [
     "get_platform_embedding",
     "get_platform_init_errors",
     "get_query_embedding",
+    "is_blank_text",
     "init_platform_embedding",
 ]
--- a/common/embedding/_service.py
+++ b/common/embedding/_service.py
@@ -395,6 +395,16 @@ async def call_embedding_gated[T](
         return await make_call()
 
 
+def is_blank_text(text: str) -> bool:
+    """Whether *text* carries nothing for a model to encode.
+
+    Whitespace counts as blank, not just the empty string: a lone space is
+    equally unembeddable and reaches the backend as the same validation
+    error.
+    """
+    return not text or not text.strip()
+
+
 def _resolve_provider_name(tenant_config: object | None) -> str:
     """Tenant override first, else ``EMBEDDING_PROVIDER`` env, else ``"fake"``."""
     if tenant_config is not None:
@@ -713,6 +723,24 @@ async def get_embedding(
     ``background=True`` for deferred work nobody is waiting on. See
     ``EMBEDDING_INTERACTIVE_RESERVED_SLOTS``.
     """
+    # Blank text is a CALLER problem, not a backend one, and the backend
+    # says so: TEI answers ``413 Input validation error: `inputs` cannot be
+    # empty``. Prod took 274 of those in one week — 272 inside the
+    # 2026-08-18 17:00-18:59 incident — because nothing checked, each one
+    # was retried (the error is deterministic, so the retry could only fail
+    # again), and the ``None`` that came out was reported to users as
+    # "Embedding service unavailable" while the backend was serving
+    # normally at ~7 ms.
+    #
+    # Returning ``None`` keeps the documented contract — callers already
+    # handle it — but now it costs zero backend requests instead of two,
+    # and the log names the real cause instead of blaming the provider.
+    if is_blank_text(text):
+        logger.warning(
+            "Embedding skipped: blank text, nothing to encode (no request sent)"
+        )
+        return None
+
     provider_name = _resolve_provider_name(tenant_config)
     provider = await _resolve_provider_or_degrade(tenant_config, "Embedding")
     if provider is None:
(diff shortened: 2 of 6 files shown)


Eldad-Caura:
Both confirmed and fixed — and verifying the first turned up something worse than the finding.

## 1. Blank query still surfaced as 503 — fixed with a type, not a message

You're right, and it defeated the PR's own stated purpose. Every search path funnels `ValueError` into `HTTPException(503)`, so "Search query must not be blank" still landed in the 5xx rate that pages an on-call.

`BlankQuery` is now a **type**:

```python
class BlankQuery(ValueError): ...
```

It subclasses `ValueError` deliberately — any handler that doesn't know it keeps today's behaviour instead of escaping as a 500 — and the two that do answer **400**.

## …and your question about the pipeline path was the important part

You asked me to verify whether `_search_memories_pipeline` shared the fault. **It does, and it's the only path that runs.**

- `_USE_PIPELINE_SEARCH = True`
- prod logged **zero** `"legacy search path invoked"` in the incident window

I initially misread this as "the fix is on a dead path", because `grep _get_or_cache_embedding` *inside* `memory_service.py` shows one caller — in the legacy function. The live caller is a **cross-file import**:

```python
# pipeline/steps/search/parallel_embed_entity_boost.py:26
from core_api.services.memory_service import BlankQuery, _get_or_cache_embedding
```

So the pipeline arm produced all 138 responses. Both arms are now fixed — the legacy one too, so a hotfix flipping `_USE_PIPELINE_SEARCH` back doesn't silently lose the distinction. The `BlankQuery` arm is ordered *before* the `ValueError` arm in both, since it subclasses it and `except` matches top-down.

## 2. Duplicated predicate — fixed

`is_blank_text` is now public and exported from `common.embedding`, and the call site imports it rather than re-implementing `not query or not query.strip()`. Same precedent as `call_embedding_gated` in #848.

## Tests

Three added on top of the original 15, all verified failing without their fix:

- **blank query → 400 through the live pipeline step** (reverted: `AssertionError: blank query answered 503; a 503 pages an on-call for a healthy backend`)
- **genuine provider failure → still 503** — the control that catches a downgrade applied too widely
- `BlankQuery` remains catchable as `ValueError`

`385 passed, 1 skipped` across all embedding, search and pipeline suites. `ruff check` + `format --check` clean on `common/`, `core-api/src/`, both test files, and the `--isolated` 88-col gate — the formatter did catch me on the new class's blank lines.

Amended and force-pushed; rebased on `3559f823`.

Eldad-Caura:
@claude

Eldad-Caura:
Took the placement suggestion — `BlankQuery` now sits below the flag block, leaving `_USE_PIPELINE_WRITE` / `_USE_PIPELINE_SEARCH` contiguous again.

On the batch finding: **the recovery mechanism described isn't what happens**, and checking it made the gap look *worse*, not better — though still correctly out of scope.

**There is no per-item `get_embedding` fallback.** `memory_service.py:1548` does:

- `inline_embedding=True` → **`raise`** (or 504 on timeout). The request fails outright; nothing is retried per item.
- deferred → logs `"bulk inline embed ... failed; deferring to backfill"` and leaves `embeddings[i] = None`

So the real recovery is the **deferred backfill**, and that path goes through core-worker's `handle_embed_request` → `call_embedding_gated(lambda: provider.embed(...))` — the **raw provider**, which does **not** traverse `get_embedding` and therefore does **not** get this guard (see #848 for why that call site holds the provider directly). A blank reaching the backfill would 413, nack, redeliver, and 413 again.

**And it is reachable, narrowly.** `schemas.py:43` is `content: str = Field(min_length=1, ...)`, which rejects `""` but **not** `" "`. Whitespace-only content passes validation, and `is_blank_text` counts it as blank precisely because TEI rejected a lone space identically in the incident.

So the honest state: `""` cannot reach the batch path, `" "` can, and whether TEI 413s on `" "` specifically I **cannot verify** — staging TEI is internal-ingress only and a *control* request with normal text 404s from outside exactly as a blank one does. I'm not going to guess from a harness I've already proven broken.

Still keeping it out of this PR, for the reason in the commit message rather than the one in the finding: guarding it means one vector per input position with `None` for blanks, and the signature is `list[list[float]]` with no slot for that — callers zip results onto rows positionally. That's a contract change.

What I'd file as the follow-up is narrower and better than filtering inside `get_embeddings_batch`: **reject whitespace-only `content` at the schema edge**, so it never reaches either the batch path or the backfill. One validator, no contract change, and it closes the input where it enters rather than three layers down.

`385 passed, 1 skipped`; ruff check + format clean; rebased on latest `main` and force-pushed.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

Issue #852, opened 2026-08-21
Auto-chunk writes compute the LLM governance verdict and then discard it (parent and children)


Eldad-Caura:
Found while fixing #808. Not caused by it, and not fixed by it — #808's fix moves governance ahead of the atomic-fact fan-out inside `_enrich_memory_background`, and the auto-chunk branch never reaches that function at all.

## The gap

`create_memory` routes large content to the auto-chunk branch (`memory_service.py`, the `len(data.content) > CHUNKING_THRESHOLD_CHARS and tenant_config.auto_chunk_enabled` condition). That branch:

1. runs `build_enrichment_pipeline()` — which contains `GovernanceScanContent` (the deterministic pattern gate) and `ParallelEmbedEnrich` + `MergeEnrichmentFields`, but **not** `GovernanceDecision`;
2. calls `_handle_auto_chunk_from_ctx`, which writes the parent and every chunk child directly through the storage client;
3. schedules **only** `process_entity_extraction`.

It never calls `_schedule_enrich_or_inline`, so `remediate_after_enrichment` never runs either. `GovernanceDecision` lives in `build_strong_write_pipeline` only, and the post-write remediation is reached from the two enrichment paths and the bulk path — none of which this branch takes.

So the LLM's free-form verdict is **computed, persisted, and then never acted on**. `MergeEnrichmentFields` writes `contains_pii`, `pii_types` and `business_relevance` into `fields["metadata"]`, and `_handle_auto_chunk_from_ctx` copies that straight into `parent_metadata`. The row ends up carrying `contains_pii: true` and staying live, with every chunk child alongside it.

## Why it matters more than it looks

This is the same defect as H-18 (#806) — a write path that computes the verdict and discards it — on a third entry point that #806 did not cover. Two things make it worse than the #808 case it was found next to:

- the **parent** leaks too, not only derived rows;
- the branch is selected *by content length*. It is the path for large pasted documents, which is exactly the shape of input most likely to carry personal data.

The deterministic pre-write gate still runs, so regex/Luhn/entropy-detectable PII never persists here either. What is lost is the same free-form judgement H-18 was about.

## Scope

Affects any deployment with `chunking.auto_chunk_enabled` and either governance policy enabled. Independent of `deployment_mode`: the branch runs before the mode-dependent write pipelines are chosen.

## Why it is not folded into #808

[...]

Issue #847, opened 2026-08-21
Audit findings recur inside the PRs that fix them: propose a lint for the reintroduced shapes and schema-derived guards for sibling drift


Eldad-Caura:
The 2026-08-14 audit's findings keep reappearing — including inside the PRs that fix them. Seven of the ten correctness highs are now merged (#814, #815, #817, #819, and #820 in flight as #846), and across those five PRs the same shapes recurred often enough that catching them one PR at a time looks like the wrong strategy.

There are two distinct recurrences, and they want different remedies.

## Shape A — a fixed bug class reappears in NEW code

* **H-02** (replica read on a read-your-write) was reintroduced in #843's new delivery code and caught in review.
* **H-05** (a row reserved/committed before a later fallible step) was reintroduced in #843's new delivery code and caught in review.
* **H-04**'s `scalar_one_or_none` shape turned up a third time during #844's suite run — as a live `MultipleResultsFound` against the dev database, not a code read.

None of these is a mistake about the *specific* bug that was filed. They are the same reasoning error made in a new place, by someone who had just fixed the previous instance. That is what a lint or a checklist is for, not a code review.

Worth noting #843's own history: three review rounds produced seven findings, **every one in the new delivery machinery and none in the four-part fix itself**. The fix was fine; the scaffolding it needed attracted the findings.

## Shape B — a hardening lands on one of a pair and never reaches the sibling

This one is sharper, because in each case the two sites are adjacent and obviously related:

* **H-09 / #844.** `purge_tenant_data` was filed. `purge_fleet_data` had the identical gap for the three fleet-scoped tables and was not filed — found only by deriving the table list from `information_schema` instead of from the issue.
* **H-10 / #846.** `GET /memories` prefers the gateway-authenticated agent over the `agent_id` query param, with a comment explaining that otherwise an agent credential could widen its view by naming a peer; and it calls `enforce_fleet_read`. `GET /memories/stats` — the sibling, in the same file, aggregating the same rows — had **neither**. A trust-1 agent could read another fleet's breakdown that the list route would have refused, and could learn a peer's private per-type counts.

In both cases the sibling was found by deriving the invariant from the schema or the code rather than from the issue text. That has now held on every high in this audit.

## What would actually help

For **Shape A**, the cheap version is a review-checklist item; the durable version is a lint. Candidates, in rough order of value:

1. a replica/`_get`-style read on a path that has just written in the same request;
2. `scalar_one_or_none()` where two rows are reachable in ordinary operation;
[...]

PR #853, opened 2026-08-21
fix(governance): a dropped memory must not leave its children behind

### LINKED ISSUE #808: Atomic-fact children escape the governance drop on inline deployments
Found while fixing H-18 (#806). Not introduced by that PR — #806 makes it *reachable* by
making governance actually run on the inline path, where before it never ran at all.

## The gap

`_enrich_memory_background` (`core-api/src/core_api/services/memory_service.py`) fans out
atomic-fact **child memories** derived from the parent's content, then #806 applies the
governance verdict to the **parent only**, in the caller.

So with `governance_pii.action = "drop"` and an LLM verdict of `contains_pii: true`:

1. The enricher extracts N atomic facts from the parent's content.
2. Each is written as its own row via `sc.create_memory(...)` — a direct storage call
   that bypasses the write pipeline, so neither `GovernanceScanContent` nor
   `GovernanceDecision` sees it.
3. `child_meta` is `{parent_memory_id, source, retrieval_hint}` (+ `embedding_pending`) —
   it carries no `contains_pii` / `pii_types` / `business_relevance`.
4. Control returns to `_schedule_enrich_or_inline`, which soft-deletes the **parent**.

The parent is gone; the children, derived from the same text, survive with no governance
metadata and no audit trail linking them to the drop. Same for
`governance_non_business.disposition = "keep_private"`: the parent is downgraded to
`scope_agent` while the children keep the parent's original `visibility`
(`parent_visibility` is read from the parent row and passed through).

## Scope

**Inline deployments only.** core-worker does not implement atomic-fact fan-out
(`_ENRICHMENT_UNROUTED_FIELDS` documents `atomic_facts` as deliberately unrouted), so no
children exist on a deferred deployment. The deterministic pre-write gate still protects
both modes against regex/Luhn/entropy-detectable PII — this is specifically about the
LLM's free-form judgement.

## Why it wasn't folded into #806

It needs a design decision rather than a line: either govern children individually, or
make the parent's verdict cascade to its fan-out, or move remediation to before the
fan-out and skip it entirely on a drop. The third is probably right — a row policy says
must not exist shouldn't spawn derived rows first — but it reorders
`_enrich_memory_background` and deserves its own tests.

## Suggested fix

Apply the verdict inside `_enrich_memory_background` immediately after the enrichment
PATCH lands and before the atomic-fact loop, returning early when
`remediate_after_enrichment` reports a drop. That kills the derived rows at the source and
keeps a single governance site per write.

Eldad-Caura:
Fixes #808.

## The bug

`_enrich_memory_background` fans the enricher's `atomic_facts` out into
a child memory per claim, each one made of text lifted from the parent.
The LLM governance verdict was applied by the CALLER, after that
function returned — by which time every child already existed.

With `governance_pii.action = "drop"` and `contains_pii: true`, the
parent was soft-deleted and the children survived it: same content, no
governance metadata, and no audit row tying them to the drop. The audit
log records one deletion; the tenant's data still holds the text.

`keep_private` had the matching hole. The parent was downgraded to
`scope_agent`, but `parent_visibility` is read off the row fetched at
the top of the function, before remediation touched it — so the
children were written at the ORIGINAL visibility, re-publishing exactly
the content the policy had just made private.

Inline deployments only: core-worker does not implement the fan-out
(`_ENRICHMENT_UNROUTED_FIELDS` lists `atomic_facts` as deliberately
unrouted), so a deferred deployment has no children to leak. The
deterministic pre-write gate still covers pattern-detectable PII in
both modes — this is the LLM's free-form judgement.

## The fix: enrich, then govern, then derive

Remediation moves out of `_schedule_enrich_or_inline` and into
`_enrich_memory_background`, immediately after the enrichment PATCH and
before anything derived from the row exists. On a drop it returns,
which also skips the entity extraction scheduled at the end —
entities mined out of dropped content are the same leak in another
table.

`remediate_after_enrichment` now returns a `RemediationOutcome` rather
than a bare bool, because a caller that goes on to create derived rows
needs more than "should I stop?": `keep_private` reports the new
visibility so the children inherit the downgrade instead of the value
`mem` still holds. Not re-read from storage — `get_memory` can route to
the read replica, the H-02 shape the `governed_row` assembly directly
above already documents.

Surviving children now carry the parent's `contains_pii` / `pii_types` /
`business_relevance`. They are extracted from the parent's content, so a
finding about that content is a finding about them; without it a child
reads as clean to an audit query while being made of the flagged text. A
drop never reaches this code, so it only ever labels rows the policy
allowed to live.

## Why the function now has two `try` blocks

The governance step sits BETWEEN them, deliberately, and that is the
only reason for the split.

H-18 established that a governance failure must propagate: the
enclosing `tracked_task` turns it into a `BackgroundTaskLog` row, and
swallowing it "would only downgrade an unenforced governance policy to
a bare log line".
`test_remediation_failure_surfaces_to_the_task_tracker` pins it.

`RuntimeError` is in the exception tuple `_enrich_memory_background`
catches, so simply moving the call inside that `try` would have
silently broken this — the easy way to stop the fan-out on failure is
also the way that loses the failure. Raising past both blocks keeps the
property AND gives the fail-safe ordering: a policy that could not be
applied is not followed by rows it might have forbidden.

The second block's handler logs "fan-out error" rather than
"enrichment error" so the two phases stay tellable apart.

## Not included

`_handle_auto_chunk_from_ctx` has the same class of gap and is filed
separately as #852: the auto-chunk branch computes the verdict
(`MergeEnrichmentFields` writes it into the metadata it persists) and
then never applies it, to the parent or the children, because it never
reaches `_schedule_enrich_or_inline` and `build_enrichment_pipeline`
carries no `GovernanceDecision`. It is a different enforcement point —
that path has the verdict PRE-persist — and bundling two enforcement
designs into one PR is what #847 is about.

## Tests

`tests/test_h808_governed_fanout.py`, 11 tests. A plain revert gives no
signal here: the file imports `RemediationOutcome`, so on `main` it
fails at collection rather than on behaviour. Each pre-fix behaviour was
therefore applied to the fixed tree instead, and in every case exactly
the intended tests failed:

| simulated pre-fix behaviour | tests that caught it |
|---|---|
| verdict does not stop the derivation | `..._creates_no_children`, `..._schedules_no_entity_extraction` |
| `keep_private` does not cascade | `..._cascades_to_the_children` |
| children carry no verdict | `..._carry_the_parents_governance_verdict` |
| governance failure swallowed | both failure tests **and** H-18's `test_remediation_failure_surfaces_to_the_task_tracker` |
| drop returns the pre-drop snapshot | `..._is_reported_as_gone` |
| no guard on the flag/config pair | `..._fails_loudly_and_early` |

The "governance failure swallowed" row is the useful one: tripping H-18's
own pre-existing test confirms that property is genuinely preserved by the
two-block structure, rather than passing by accident.

Two of the eleven guard against over-applying the fix — an untouched parent
still passes its own visibility down, and with
`run_governance_remediation=False` (strong mode, already governed
synchronously) remediation is not called at all.

`test_governance_remediation.py` is updated for the new return type, and
`consumer.py` reads `.dropped`.

## Review round

Two Lows, both applied. A drop now returns `None` rather than the
pre-drop snapshot — `is not None` is the reading a caller will assume
means "still there", and handing back a dict for a row that was just
soft-deleted invites it. And `run_governance_remediation=True` without
`governance_config` now raises up front instead of failing opaquely on
an attribute of `None` two modules away: the parameters are
independent, and the caller-side "if you add a third call site" warning
describes exactly the circumstance in which they would be set
inconsistently.

### DIFF
--- a/core-api/src/core_api/consumer.py
+++ b/core-api/src/core_api/consumer.py
@@ -104,7 +104,7 @@ async def handle_memory_enriched(event: Event) -> None:
     # counterpart to the synchronous GovernanceDecision step. ``resolve_config``
     # tolerates a None session (cache-first; cold-miss opens its own).
     gov_cfg = await resolve_config(payload.tenant_id)
-    if await remediate_after_enrichment(memory, gov_cfg):
+    if (await remediate_after_enrichment(memory, gov_cfg)).dropped:
         # Row was dropped by policy — skip contradiction detection on it.
         logger.info(
             "memory-enriched: row dropped by governance policy; ack",
--- a/core-api/src/core_api/services/governance_remediation.py
+++ b/core-api/src/core_api/services/governance_remediation.py
@@ -16,6 +16,7 @@ eventually-consistent here (≈ enrichment-deferral latency).
 from __future__ import annotations
 
 import logging
+from dataclasses import dataclass
 from typing import Any
 
 from core_api.clients.storage_client import get_storage_client
@@ -33,17 +34,34 @@ from core_api.services.governance_gate import (
 logger = logging.getLogger(__name__)
 
 
-async def remediate_after_enrichment(memory: dict, cfg: Any) -> bool:
+@dataclass(frozen=True)
+class RemediationOutcome:
+    """What remediation did to the row — enough for a caller to follow it.
+
+    ``dropped`` alone was sufficient while the only caller had nothing left to
+    do afterwards. A caller that goes on to create rows DERIVED from this one
+    needs the rest: ``keep_private`` downgrades the row's visibility, and a
+    derived row that keeps the pre-downgrade visibility re-publishes exactly
+    the content the policy just made private (#808).
+    """
+
+    dropped: bool = False
+    visibility: str | None = None
+    """The row's new visibility when remediation changed it; ``None`` when it
+    was left alone."""
+
+
+async def remediate_after_enrichment(memory: dict, cfg: Any) -> RemediationOutcome:
     """Apply LLM-signal governance to a fast-mode row after enrichment landed.
 
-    Returns ``True`` if the row was dropped (the caller should then stop further
-    processing of it). No-op + ``False`` when governance is disabled or the
-    signals are clean.
+    Returns what was done. A caller that only needs "should I stop?" reads
+    ``.dropped``; one that creates derived rows must also honour
+    ``.visibility``. No-op when governance is disabled or the signals are clean.
     """
     pii_cfg = cfg.governance_pii
     nb_cfg = cfg.governance_non_business
     if not pii_cfg.enabled and not nb_cfg.enabled:
-        return False
+        return RemediationOutcome()
 
     md = memory.get("metadata_") or memory.get("metadata") or {}
     content = memory.get("content") or ""
@@ -54,7 +72,7 @@ async def remediate_after_enrichment(memory: dict, cfg: Any) -> bool:
         # A malformed enriched-event payload without an id would otherwise
         # soft-delete "None" and stamp resource_id="None" on every audit row.
         logger.warning("governance: remediate_after_enrichment called with memory missing 'id'; skipping")
-        return False
+        return RemediationOutcome()
     memory_id = str(raw_id)
     sc = get_storage_client()
 
@@ -78,7 +96,7 @@ async def remediate_after_enrichment(memory: dict, cfg: Any) -> bool:
             )
             await sc.soft_delete_memory(memory_id)
             logger.info("governance: dropped fast-mode memory %s (pii)", memory_id)
-            return True
+            return RemediationOutcome(dropped=True)
         # mask/flag: the LLM gives no offsets to redact a free-form span, and in
         # fast mode the row is already persisted — so a "mask"-configured tenant
         # can only be flagged here. Keep the action truthful (flag), but record
@@ -113,7 +131,7 @@ async def remediate_after_enrichment(memory: dict, cfg: Any) -> bool:
             )
             await sc.soft_delete_memory(memory_id)
             logger.info("governance: dropped fast-mode memory %s (non-business)", memory_id)
-            return True
+            return RemediationOutcome(dropped=True)
         if nb_cfg.disposition == "keep_private":
             await sc.update_memory(memory_id, tenant_id, {"visibility": "scope_agent"})
             await emit_governance_audit(
@@ -123,4 +141,7 @@ async def remediate_after_enrichment(memory: dict, cfg: Any) -> bool:
                 detail=nonbusiness_audit_detail(ACTION_NB_KEEP_PRIVATE, content, "fast"),
                 resource_id=memory_id,
             )
-    return False
+            # Reported so a caller creating derived rows can carry the
+            # downgrade to them instead of publishing the same content wider.
+            return RemediationOutcome(visibility="scope_agent")
+    return RemediationOutcome()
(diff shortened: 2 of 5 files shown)


Eldad-Caura:
Both Lows applied.

**A drop now returns `None`.** You're right that this is a latent trap rather than a live bug — the sole caller discards the value — but `is not None` is exactly the reading the next caller will assume means "the row is still there", and handing back a dict for a row that was just soft-deleted invites it. `None` now means one thing on every branch: there is no live governed row. The docstring says so, and notes the value is diagnostics only and not a governance hook. `test_a_dropped_parent_is_reported_as_gone` fails if the snapshot comes back.

**The flag/config pair now fails fast.** `run_governance_remediation=True` with `governance_config=None` raises `ValueError` before any work, rather than surfacing an `AttributeError` on `None.governance_pii` from two modules away. Worth more than the tidiness: the caller-side ⚠ note about adding a third call site describes precisely the circumstance in which someone sets the flag and forgets the config, so the opaque failure would land on the person least placed to read it. Raised at the top of the function rather than at the governance step so a misconfigured call site doesn't first burn an LLM enrichment. `test_the_flag_without_a_config_fails_loudly_and_early` covers it.

I chose the guard over the type-level alternative you offered: an overload would express it, but the two parameters are read at different points and a runtime check is what actually fires on the path that matters.

Both were verified the same way as the rest of this PR — the pre-fix behaviour applied to the fixed tree, checking that exactly the intended test catches it. That makes six simulated designs, table in the PR body.

Gates: 22 tests green across the two governance files; full suite **4681 passed**, 16 pre-existing failures whose set is identical to unmodified `main`; ruff check + format separately; mypy clean at 199 files. CI was green on the previous head; re-running.

Eldad-Caura:
@erni-a ready for review — all gates green.

Closes #808. **This is the last of the ten audit correctness highs.** Verdict found no correctness bugs; the two Lows are applied and covered.

Not re-tagging `@claude` this round: unlike #849, neither fix restructured a lifecycle step — one changed a return value nothing consumes, the other added a precondition guard. Say the word if you'd rather have another pass.

**Worth knowing before you look:** this fix reorders a background task, and the reordering is the whole change. `_enrich_memory_background` now runs enrich → govern → derive, with the governance step deliberately sitting **between two `try` blocks**. That is not stylistic. H-18 established that a governance failure must reach the task tracker rather than become a log line, and `RuntimeError` is in the exception tuple this function catches — so putting the call inside the `try` (the obvious way to stop the fan-out on failure) silently loses the failure. Raising past both blocks keeps the propagation *and* the fail-safe ordering. I only found that because H-18's own test failed.

**The judgement call worth a second opinion:** `remediate_after_enrichment` now returns a `RemediationOutcome` instead of a bool. The alternative was to re-read the parent row after remediation to learn its new visibility — which would have walked into the replica-lag trap the code immediately above it already documents (H-02). Three call sites; two needed a one-word change.

**A related gap I deliberately left out**, filed as #852: the auto-chunk branch computes the same LLM verdict and never applies it — to the parent *or* the children — because it never reaches this code path and `build_enrichment_pipeline` carries no `GovernanceDecision`. It is arguably worse than what this PR fixes (the parent leaks too, and the branch is selected by content length, so it is the path for large pasted documents). It needs a different enforcement point, and bundling two of those is what #847 is about.

Signals:

- Full suite **4681 passed**; 16 failures, set identical to unmodified `main` (pre-existing local search/FTS).
- 22 tests across the two governance files. A plain revert gives no signal here — the new file imports a new symbol, so on `main` it fails at collection, not on behaviour. Instead each of six pre-fix behaviours was applied to the fixed tree; every one failed exactly the intended test. Table in the PR body. The best of them trips H-18's *pre-existing* test, which is what proves that property survives the reordering.

Not merging — that's yours.

erni-a (approved):

PR #854, opened 2026-08-21
fix(embedding): a full backend must be told once, not six times

Eldad-Caura:
Groundwork for bounding embedding demand across instances, which is the
last open item of this workstream. The concurrency cap is per process, so
`cap x instances` is what actually reaches the backend and no instance can
observe that number. Measuring it first changed what the fix should be.

## The arithmetic, prod, 2026-08-21

| | scaling | concurrency | in flight |
|---|---|---|---|
| core-api | minScale 10, maxScale 200 | cap **16** | **160** at the scale it runs, 3,200 at the ceiling |
| core-worker | minScale 2, maxScale 20 | 1/instance (the pull loop drains sequentially) | **2** now, 20 at the ceiling |
| **TEI (supply)** | minScale 2, **maxScale 3** | containerConcurrency 10 | **20 warm, 30 absolute** |

**~8x oversubscribed at the scale prod runs at; ~107x at the configured
ceiling.** Instance count was observed between 9 and 12 across a week, so
the ceiling figure describes a state prod has never entered.

## Why this is not a token bucket, and not fair shares

**Nothing has ever pushed back.** In the 7 days to 2026-08-21 TEI served
**345,802 requests with zero 5xx and zero 429** — every non-2xx was one
of the 270 blank-input 413s that #851 fixed.

Meanwhile *our* gate rejected work that backend would have served: 1,915
gate timeouts on 08-17 alone, with TEI sitting at roughly half of a single
instance's container concurrency. Raising the cap 8 -> 16 was the
2026-08-21 change and it was in that direction on purpose.

So fair shares (30 / 10 = **3**) would optimise for a coincident burst
that has never happened, at the cost of the case that happens
constantly. Demand is bursty and not coincident across instances, and a
generous per-process cap plus a hard aggregate signal beats a small
per-process cap. A shared Redis bucket buys a coordination dependency in
the hot path to enforce a limit nothing is reaching.

**What bounds the aggregate is the backend's own 429.** It is the only
signal carrying cross-instance information — it can only fire once the
*other* instances have taken the capacity. Rejected work already lands on
the durable EMBED_REQUESTED path and drains at core-worker's sequential
rate, which is the demand smoothing. Two things had to be true for that
to work, and neither was.

## Fault 1 — a 429 was a provider failure

It was retried (answering "you are full" with more load) and counted
against the streak behind `Embedding service degraded [<backend>]`,
pointing an operator at a service behaving exactly as intended. Same
class as #850, one layer out: our gate says *this process* is at its cap,
a 429 says *the shared backend* is.

`EmbeddingBackendBusy` now names it, raised at `call_embedding_gated` —
the one choke point every embed passes through since #848 — so no call
site can forget the classification.

Narrow on purpose: **429 only**. Cloud Run returns it for "no instance
available" and TEI when its queue is full, both unambiguously "later". A
503 is equally what a real outage looks like, so it stays a fault:
retried, counted, loud.

## Fault 2 — one embed was up to six requests

`AsyncOpenAI` was constructed without `max_retries`, so the SDK applied
its default of 2 (three HTTP requests per call) *beneath* a service layer
already retrying `EMBEDDING_RETRY_ATTEMPTS` times.

The multiplication was invisible: SDK retries happen below the gate, so
they take no slot, emit no log line, and count as one call in every stat
we keep. Two concrete harms beyond tidiness:

1. It inverted the response to saturation — 6x the load at the exact
moment capacity ran out.
2. It broke the timeout arithmetic other budgets derive from.
`OPENAI_REQUEST_TIMEOUT_SECONDS` is per *request*, so three is 75 s
inside a call believed capped at 25 s — and
`BULK_STRONG_EMBED_TIMEOUT_SECONDS` (8 s) was sized against the 25 s
figure.

`EMBEDDING_PROVIDER_MAX_RETRIES=0` pins it. Nothing is lost: genuine
provider errors are still retried, now holding a gate slot with the
attempt counted and logged.

**These ship together deliberately.** Without the pin, "not retried"
would be false — the SDK would have already sent the 429 three times
before this layer saw it.

## Also corrected

- `get_embeddings_batch`'s docstring credited **both** bulk callers with a
per-item retry. Only `_reembed_batch_via_provider` has one;
`create_memories_bulk` re-raises inline or leaves `embedding=NULL` for the
backfill. Durable either way, but there is no second attempt in that
request.
- The cap's own comment pointed at "raising TEI maxScale/containerConcurrency"
as the real fix. It is not available: L4 GPUs are unavailable in
us-central1 and the 3 -> 8 quota request was **denied**, not deferred.
- "fleet" as a word for a set of Cloud Run services.

## What this does not do

It is not a hard aggregate cap. Demand is still `cap x instances` with no
coordination; what changes is that the backend can now say no and be
believed. A real ceiling means routing the remaining synchronous
non-interactive embeds through core-worker — an API-contract change for
`POST /documents`, and not justified by anything measured here.

## Verification

`4739 passed, 3 skipped, 1 xfailed` — full suite, zero failures. All 12
CI ruff scopes clean.

**5 of the 8 new tests fail against unfixed source** (429 retried, streak
advanced, classifier absent, batch streak advanced, `max_retries == 2`).
The other 3 are controls that must pass both ways: a 500 is still retried
and counted, a 503 is still a fault, and an earlier genuine failure in the
same call is still recorded — the regression the same fix introduced when
it was made one layer up in #850.

### DIFF
--- a/common/embedding/__init__.py
+++ b/common/embedding/__init__.py
@@ -25,6 +25,11 @@ Public surface:
   backend, is what stopped an embed. A ``TimeoutError`` subclass, so
   existing handlers are unaffected; catch it only to tell saturation apart
   from provider failure.
+* :class:`EmbeddingBackendBusy` — the same distinction for the SHARED
+  backend: raised on a 429, meaning capacity ran out across every calling
+  instance rather than in this process. Not a ``TimeoutError`` (the
+  refusal is immediate). Catch it to tell aggregate saturation apart from
+  a provider fault; the concurrency cap is per process and cannot see it.
 * :func:`init_platform_embedding` / :func:`get_platform_embedding` —
   platform-tier singleton, initialised once at service startup from
   ``PLATFORM_EMBEDDING_*`` env vars.
@@ -51,6 +56,7 @@ from common.embedding._platform import (
 )
 from common.embedding._registry import get_embedding_provider
 from common.embedding._service import (
+    EmbeddingBackendBusy,
     EmbeddingGateTimeout,
     call_embedding_gated,
     get_embedding,
@@ -65,6 +71,7 @@ from common.embedding.providers.fake import (
 )
 
 __all__ = [
+    "EmbeddingBackendBusy",
     "EmbeddingGateTimeout",
     "EmbeddingProvider",
     "FakeEmbeddingProvider",
--- a/common/embedding/_service.py
+++ b/common/embedding/_service.py
@@ -270,6 +270,54 @@ class EmbeddingGateTimeout(TimeoutError):
     """
 
 
+class EmbeddingBackendBusy(Exception):
+    """The backend refused the call because it is at capacity (HTTP 429).
+
+    The same class of signal as :class:`EmbeddingGateTimeout`, one layer
+    out: our gate says "this PROCESS is at its cap", a 429 says "the
+    SHARED backend is". Both are capacity, neither is a fault, and
+    retrying either deepens it.
+
+    Worth naming separately because a 429 is the only evidence any single
+    process can ever have about aggregate demand. The concurrency cap is
+    per process, so ``cap x instances`` is what actually arrives at the
+    backend and no instance can observe that number — but a 429 can only
+    happen once the other instances have taken the capacity, so it
+    reports the aggregate directly. It is the signal an aggregate cap
+    would have to be built on, which is why it must not be swallowed into
+    the generic provider-failure path (where it was: a 429 counted
+    against the degraded-provider streak, and was retried).
+
+    Deliberately NOT a ``TimeoutError`` subclass, unlike its sibling. A
+    gate timeout is one — we really did wait — and existing handlers were
+    already written for that. A refusal is immediate, and callers that
+    special-case timeouts should not silently inherit this.
+    """
+
+
+def _is_backend_busy(exc: BaseException) -> bool:
+    """Whether *exc* is the backend declining for capacity (HTTP 429).
+
+    Duck-typed on the status code rather than importing the provider SDK:
+    this is the service layer, it is shared by every provider, and it has
+    no ``openai`` import today. ``openai.APIStatusError`` exposes
+    ``status_code`` directly and ``httpx.HTTPStatusError`` exposes it via
+    ``.response``, so both shapes are covered without either dependency.
+
+    Only 429 counts, and the narrowness is the point. Cloud Run returns
+    429 for "no instance available" and TEI returns it when its queue is
+    full, so a 429 unambiguously means "at capacity, try later". A 503 is
+    the ambiguous one — it is equally what a genuine outage looks like —
+    and classifying it as capacity would suppress the degraded-provider
+    ERROR during a real one. A 503 therefore stays a provider failure:
+    retried, counted, and loud.
+    """
+    code = getattr(exc, "status_code", None)
+    if code is None:
+        code = getattr(getattr(exc, "response", None), "status_code", None)
+    return code == 429
+
+
 def _concurrency_gate() -> asyncio.Semaphore:
     """The process-wide cap on concurrent provider calls.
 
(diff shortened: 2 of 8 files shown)


Eldad-Caura:
Review finding confirmed and fixed, plus one comment the change made stale.

## `read_int_env` rejecting its own default — real, and worse than the warning

Verified in `common/env_utils.py:50`: the floor was `if value < 1`, hardcoded. So with `default=0`:

- `EMBEDDING_PROVIDER_MAX_RETRIES=0` printed `must be >= 1; falling back to 0` — a warning about a correct value.
- **And it arrived via the fallback path, not by being accepted.** That is the part worth fixing rather than silencing: if the default were ever changed to 2, an explicit `=0` would have silently become 2, and the knob would quietly stop being able to disable SDK retries — the exact failure it exists to prevent.

Took option 1. The floor is now a keyword-only `minimum: int = 1`, so **every existing caller is byte-for-byte unchanged**, and the garbage-input guard is untouched. Enterprise has no vendored `env_utils.py` and no `read_int_env` callers (checked), so there is no cross-repo edge.

4 tests added — 3 fail against the unfixed helper (`TypeError: unexpected keyword argument 'minimum'`); `test_minimum_defaults_to_one` passes both ways on purpose, as the backward-compatibility guard.

## Also in this push

`core_worker/consumer.py`'s comment on the nack said "a gate timeout raises `TimeoutError`". Two saturation classes reach that line now, and the comment was asserting the behaviour, so it needed to name both — and to say that the nack is what makes the queue the demand smoothing rather than just a transport.

## Verification

`4743 passed, 3 skipped, 1 xfailed` — full suite, zero failures. All 12 CI ruff scopes clean. Rebased on `70a357f7`, single commit, force-pushed.

Eldad-Caura:
@claude please re-review — the `read_int_env` finding is fixed with a keyword-only `minimum` (existing callers unchanged), and the core-worker nack comment now names both saturation classes.

Eldad-Caura:
Both handled. The finding is right, and it's the sharper of the two reviews.

## "the work defers instead" — wrong, and wrong in the worst place

Confirmed: deferral is the **caller's** property, not this layer's. A write's `None` persists as `embedding=NULL` for the backfill sweep; a batch's fans out to EMBED_REQUESTED. A query embed has neither — the search just fails with a 503. So a shared message in `_run_with_retry` cannot claim it, and `background` is no proxy either: a synchronous write is `background=False` and *does* still land in the backfill.

Dropped the claim rather than parameterising it. The line now states what this layer decided and stops:

> `… — not retrying; a refusal is capacity, and retrying it deepens the shortage`

Added the substantive complement as a test rather than pinning log text: **`test_the_interactive_path_also_stops_at_one`** — `get_query_embedding` under a 429 makes exactly one request, returns `None`, and records no failure. That covers the asymmetry the review identified, as behaviour.

## `exc_info` / original error — taken, partly

Added the provider's own text (`Backend said: %s`). Agreed on the reasoning: this has never fired, so the first time it does the question is "was that really a 429", and the answer has to be in the line that alerts.

Not `exc_info`, deliberately — a genuine capacity event fires this per refused call, and the precedent for volume is 1,915 gate timeouts in one hour. The chain is still attached via `raise … from exc` for anything that wants the traceback.

## Other providers — checked, and one real gap outside this PR

- `providers/local.py` — in-process sentence-transformers, no HTTP, nothing to retry.
- `providers/fake.py` — pure function.
- `common/ranking/providers/remote.py:95` — bare `httpx.AsyncClient`. httpx does **no** automatic retries by default, so no hidden multiplier.
- **`common/llm/providers/openai.py:95` — `openai.AsyncOpenAI` with no `max_retries`.** Same latent 3x. Out of scope here (this PR is the embedding path and the LLM path has its own retry/budget story), but it is real and worth its own change.

So the embedding side is complete with this fix.

## Verification

`4744 passed, 3 skipped, 1 xfailed` — full suite, zero failures. All 12 CI ruff scopes clean. **6 of the 9 tests now fail against unfixed source**, including the new interactive-path one.

One note for the record: an earlier run of this suite showed two failures (`test_stats_excludes_soft_deleted`, `test_schedule_sweep_processes_pending_jobs`). Both pass in isolation and **did not reproduce on a full re-run** — order-dependent, in code paths unrelated to anything here. Flagging rather than calling them pre-existing, which I did not prove.

Eldad-Caura:

Two Claude review cycles, both findings fixed and answered in-thread: `read_int_env` rejecting its own intended default (fixed with a keyword-only `minimum`, existing callers byte-for-byte unchanged), and a log line that claimed work was deferred on a path where nothing defers.

All 8 checks green. `4744 passed, 3 skipped, 1 xfailed` on the full suite; all 12 CI ruff scopes clean; 6 of the 9 new tests fail against unfixed source. Single commit, rebased on `70a357f7`.

Short version of what it does: the concurrency cap is per process, so `cap x instances` is what reaches TEI and no instance can observe that number — a 429 is the only signal that can, because it can only fire once the *other* instances have taken the capacity. This makes that signal usable: a 429 stops being a retried provider fault, and one embed stops being up to six HTTP requests. Measured context is in the PR body — TEI served 345,802 requests in 7 days with zero 429s and zero 5xx, which is also why this is not a token bucket.

erni-a (approved):

Issue #856, opened 2026-08-21
Deferred deployments: the auto-chunk parent is never embedded or enriched, and nothing backfills it


Eldad-Caura:
Found while establishing the deployment-mode scope of #852. Orthogonal to governance — filing separately.

## The gap

On the multi-fact auto-chunk branch (`_handle_auto_chunk_from_ctx`, `len(facts) > 1`), the parent row is written and then **nothing is scheduled except entity extraction**:

```
$ # every scheduling/backfill call reachable in the function, by AST
scheduling/backfill calls present:  build_persist_pipeline, get_embeddings_batch, process_entity_extraction
absent: _schedule_embed_or_reembed, _schedule_enrich_or_inline,
        publish_memory_embed_request, publish_memory_enrich_request,
        remediate_after_enrichment
```

(`build_persist_pipeline` is the 0-1-fact fall-through, a different sub-path. `get_embeddings_batch` embeds the *children*.)

In an **inline** deployment that is fine — `ParallelEmbedEnrich` already produced both the embedding and the enrichment synchronously, so the parent is complete when it is written.

In a **deferred** deployment it is not. The auto-chunk branch never sets `resolved_write_mode`, so `ParallelEmbedEnrich` takes its `else` arm:

```python
defer_embedding = (not settings.inline_embedding) and cached_embedding is None
defer_enrichment = not settings.inline_enrichment
```

Both defer. The parent is therefore inserted with `embedding: None` and no LLM-derived fields — and the two mechanisms that would normally repair that are both in `ScheduleBackgroundTasks`, which this branch never runs.

## Nothing marks it or retries it

- `metadata["enrichment_pending"]` is set by `MergeEnrichmentFields` only when `resolved_write_mode == "fast" and enrichment is None`. This branch leaves `resolved_write_mode` unset, so the marker is absent — and `enrichment_pending` is the only signal core-worker clears, and the only one `MemoryOut` documents for agents.
- `_schedule_embed_or_reembed` (the CAURA-594 shim that publishes `EMBED_REQUESTED` in deferred mode) is called from `ScheduleBackgroundTasks`, `_create_memory_legacy`, and the bulk re-embed failure fallbacks. None of them cover this path.
- There is no reconciliation sweep over unenriched or unembedded rows.

[...]

Issue #855, opened 2026-08-21
The legacy write path (_USE_PIPELINE_WRITE=False rollback lever) has no content governance


Eldad-Caura:
Found while fixing #852, and deliberately kept out of that PR.

`_USE_PIPELINE_WRITE = False` is a documented emergency lever:

> Hardcoded pipeline flags. Intentionally not read from env/Settings: the legacy write/search paths are deprecated and scheduled for removal. **If an emergency rollback is needed, flip these to False and ship a hotfix** […]

Flipping it routes every write through `_create_memory_legacy`, which has **no content governance at all**:

| gate | pipeline path | `_create_memory_legacy` |
|---|---|---|
| `GovernanceScanContent` (deterministic PII/PCI/secrets: mask / drop / flag) | yes, in `build_enrichment_pipeline` | **absent** |
| `BusinessPersonalPregate` (opt-in early business/personal reject) | yes, fast + strong | **absent** |
| `GovernanceDecision` (LLM free-form PII + business/personal) | strong, and auto-chunk after #852 | **absent** |
| `remediate_after_enrichment` (post-write, fast/inline) | via `_schedule_enrich_or_inline` | **never called** |

The one thing it does is copy `enrichment.contains_pii` / `pii_types` into `metadata` — a label, with nothing acting on it. It doesn't even record `business_relevance`.

So a tenant with `governance.pii.action = "drop"` gets a write path that stores the credit-card number the deterministic scanner would have rejected. The lever's whole point is to be reachable under pressure, which is the worst moment to discover a policy stopped applying.

## Why this is not #852

#852 is one branch of the live path computing a verdict and discarding it. This is a *different* path having no gates to begin with. Adding `GovernanceDecision` alone here would be worse than nothing — it would make the path look governed while the deterministic backstop, which is the fail-closed one, still isn't there.

## Note on severity

This is latent, not live: `_USE_PIPELINE_WRITE` is a hardcoded `True` with no env override, so nothing reaches `_create_memory_legacy` today except tests that call it directly. Filing it as a property of the rollback lever rather than a production leak.

## Options

1. **Document the caveat on the lever.** `_USE_PIPELINE_SEARCH` already carries one ("must be cut from a core-api at or after the commit that moved the legacy search path onto nested `search_params`"); the write lever's caveat — *rolling back drops all content governance* — is not written down anywhere. Cheapest, and it puts the fact where whoever flips the switch will read it.
[...]

PR #857, opened 2026-08-21
fix(governance): the auto-chunk branch must apply the verdict it computes

### LINKED ISSUE #852: Auto-chunk writes compute the LLM governance verdict and then discard it (parent and children)

Eldad-Caura:
Fixes #852.

## The bug

The auto-chunk branch runs `build_enrichment_pipeline`, which ends in `MergeEnrichmentFields` — so `contains_pii`, `pii_types` and `business_relevance` are sitting in `fields["metadata"]` before the first row is written, and `parent_metadata` copies them onto the parent. Nothing then acted on them: `GovernanceDecision` lives only in `build_strong_write_pipeline`, and `remediate_after_enrichment` is reached from the two enrichment paths and the bulk path, none of which this branch takes.

Four distinct losses, not one:

| policy | what happened |
|---|---|
| `pii.action="drop"` / `non_business.disposition="drop"` | parent **and** every chunk child persisted |
| `keep_private` | parent kept its visibility, and so did the children cut out of its text |
| `flag` / `mask` | row was marked (`MergeEnrichmentFields` does that much) but **no audit row** was written |
| any | children carried no verdict at all — `metadata_` was exactly `{parent_memory_id, source}` |

That last row is the shape #808 just fixed for atomic facts, on the other fan-out.

The deterministic gate is unaffected — `GovernanceScanContent` sits in `build_enrichment_pipeline` after `LoadTenantConfig`, so pattern-detectable PII was already masked or rejected here. What was lost is the free-form judgement, same as H-18 (#806) on a third entry point.

## The fix: decide before deriving

`GovernanceDecision` runs as the **first statement** of `_handle_auto_chunk_from_ctx`, ahead of the chunking call. Ahead, not merely before the insert: chunking is an LLM round-trip a rejected write shouldn't pay for, and deciding before anything is derived is the fail-safe order.

Three things fall out of that placement, and all three are deliberate:

- **The `keep_private` downgrade cascades for free.** The step mutates `data.visibility` in place, and both the parent payload and every child payload read `data.visibility or "scope_team"`. Nothing has to be threaded through by hand — but it only works if the step runs before either payload is built.
- **The 0-1-fact fall-through is covered.** That sub-path hands the same context to `build_persist_pipeline`, which has no `GovernanceDecision` of its own, on a branch that never sets the `resolved_write_mode` that `ScheduleBackgroundTasks` keys post-write remediation on.
- **A `drop` costs no chunking call.**

It's a composition (`build_auto_chunk_governance_pipeline`), not a bare `GovernanceDecision().execute(ctx)`, so that *"which write paths enforce the LLM verdict?"* stays answerable by reading `compositions/write.py`. That's the question that got the wrong answer when this branch was written; a test now pins the exact set of compositions carrying the step.

It is **not** appended to `build_enrichment_pipeline` — that composition is shared with the extract-only branch (`persist=False`), which writes nothing and returns a preview of content the caller already holds. A 422 there would refuse a request that leaks nothing.

### Why the `governance.failed` check matters

The runner re-raises `HTTPException` (the 422 a `drop` raises) but **catches everything else** and returns `failed`. Without the explicit check, a governance step that blew up — an audit backend outage is the realistic one — returns as an ordinary result and the write proceeds ungoverned: the exact defect being fixed, one level up. Not theoretical: it fired during development against a stale test fake (below).

### Children inherit the verdict, via a shared helper

`_inherit_governance_signals` replaces the inline loop #808 added for atomic facts and is now called from both fan-outs. Shared rather than repeated, because that repetition is the shape #847 is about — this is the second of the pair, and it didn't get in #808 what the first one did.

## A stale test fake, surfaced by the new check

`test_child_dedup_minting_paths.py` built `tenant_config` as a `SimpleNamespace` carrying the two attributes the handler happened to read when those tests were written. `GovernanceDecision` reads `governance_pii`/`governance_non_business`, so it raised `AttributeError` inside the step — which the new check correctly turned into a refused write rather than an ungoverned one. Replaced with the real `ResolvedConfig`; governance left unconfigured, so the step SKIPs and those dedup tests observe exactly what they always did.

## Tests

`tests/test_h852_governed_auto_chunk.py`, 14 tests. A plain revert gives no signal (the file imports `build_auto_chunk_governance_pipeline`, so on `main` it fails at collection, not on behaviour). Each rejected or pre-fix design was applied to the **fixed** tree instead, and in every case exactly the intended tests failed:

|---|---|
| no gate at all (the pre-fix branch) | 9 of the 14 |
| children inherit nothing | `..._carry_the_parents_governance_verdict`, `..._disabled_enforces_nothing` |
| gate inside the `len(facts) > 1` branch | `..._costs_no_chunking_call`, `..._fall_through_is_governed_too` |
| `governance.failed` ignored | `..._failed_governance_step_refuses_the_write` |
| step appended to the shared enrichment pipeline | `..._only_the_persisting_compositions_apply_the_llm_verdict` |
| inherit unconditionally rather than by presence | `..._does_not_invent_a_verdict` |

Four of the fourteen guard against over-applying the fix: an enabled policy with nothing to act on writes unchanged; a disabled policy enforces nothing; an uncertain LLM signal (`llm_ms == 0`) records the uncertainty and lets the write through; a parent with no verdict keys yields children with only their provenance keys.

## Scope, and two things deliberately not in here

**Inline deployments.** In a deferred deployment `ParallelEmbedEnrich` skips the LLM on this branch, so there is no verdict to apply — the gate records `governance_llm_uncertain` and the write proceeds. Same scope as #808.

**`_create_memory_legacy` — filed as #855.** Its auto-chunk branch has the same gap, but so does the rest of it: no `GovernanceScanContent`, no `GovernanceDecision`, no `BusinessPersonalPregate`, and `remediate_after_enrichment` never called. It is ungoverned end to end rather than leaking one verdict, and it is unreachable (`_USE_PIPELINE_WRITE` is a hardcoded `True`; the lever exists to cut a hotfix build). Wiring a whole governance stack into a deprecated path is a different change from applying a verdict on a live one.

**Deferred auto-chunk parents are never enriched at all — filed as #856.** Noticed while establishing the deployment-mode scope above, and orthogonal to governance.

## Gates

`ruff check` + `ruff format --check` on `core-api/src/` clean and run separately; `mypy` clean; broker OpenAPI baseline up to date (8 operations, no contract change — no route or parameter is touched).

### Full suite, and one flake worth naming

| run | result |
|---|---|
| `origin/main` (70a357f7) ×2 | 4731 passed, 0 failed |
| this branch | 4743 passed, **2 failed** |
| this branch, rerun | 4745 passed, 0 failed |
| this branch, new test file ignored | 4731 passed, 0 failed |

4731 + 14 = 4745, so the counts line up.

The one run with failures hit `test_api_interview.py::test_schedule_queues_once_then_respects_pending_and_dueness` and `test_interview_async_submit.py::test_schedule_sweep_processes_pending_jobs`. They did not reproduce, and they are not reachable from this change:

- `interview_service.py` writes memories only through `create_memories_bulk` (line 553). Nothing in the interview path calls `create_memory`, so nothing reaches `_handle_auto_chunk_from_ctx` or the new composition.
- The first of the two sorts *before* `test_h852_governed_auto_chunk.py` in collection order, so the new file cannot have polluted it.
- The observed shape — `jobs_processed >= 1` passing while `jobs_done == 0` — is the outcome `process_pending_interview_jobs` itself documents for a race: *"a concurrent run claimed the job between the sweep's fetch and our claim"*, counted as `jobs_skipped`.

Reporting it rather than rounding to "green": one run in three did fail, and the reason it did is load-sensitivity in a shared module-level `synthesis_sem`, not this diff.

### DIFF
--- a/core-api/src/core_api/pipeline/compositions/write.py
+++ b/core-api/src/core_api/pipeline/compositions/write.py
@@ -38,6 +38,24 @@ def build_enrichment_pipeline() -> Pipeline:
     )
 
 
+def build_auto_chunk_governance_pipeline() -> Pipeline:
+    """The LLM verdict gate for the auto-chunk branch (#852).
+
+    ``GovernanceDecision`` cannot simply be appended to
+    ``build_enrichment_pipeline``: that composition is shared with the
+    extract-only branch (``persist=False``), which writes nothing and returns a
+    preview of content the caller already holds. Rejecting there would refuse a
+    request that leaks nothing. So the step is applied only on the branch that
+    goes on to persist.
+
+    A composition rather than a bare ``GovernanceDecision().execute(ctx)`` call
+    so that "which write paths enforce the LLM verdict?" stays answerable by
+    reading this module — the question that got the wrong answer when the
+    auto-chunk branch was written.
+    """
+    return Pipeline("write_auto_chunk_governance", [GovernanceDecision()])
+
+
 def build_persist_pipeline() -> Pipeline:
     """Only for persist=True, non-chunked memories."""
     return Pipeline(
--- a/core-api/src/core_api/services/memory_service.py
+++ b/core-api/src/core_api/services/memory_service.py
@@ -150,6 +150,32 @@ def _auto_chunk_request_id() -> str:
     return f"auto-chunk:{uuid4()}"
 
 
+# The LLM governance verdict, as ``MergeEnrichmentFields`` / the enrichment
+# worker record it on the parent's metadata.
+_GOVERNANCE_SIGNAL_KEYS = ("contains_pii", "pii_types", "business_relevance")
+
+
+def _inherit_governance_signals(child_meta: dict, parent_meta: dict) -> None:
+    """Copy the parent's governance verdict onto a row derived from its content.
+
+    Derived rows — atomic facts (#808), auto-chunk children (#852) — are built
+    out of the parent's text, so a finding about that content is a finding about
+    them. Without the copy a child reads as clean to every later consumer (an
+    audit query, any future remediation pass) while being made of the flagged
+    text.
+
+    Only ever called on rows the policy allowed to live: a ``drop`` verdict
+    stops both fan-outs before they reach their children.
+
+    Shared by the two fan-outs rather than repeated, because that repetition is
+    exactly the shape #847 is about — the first of the pair got this in #808 and
+    the second did not.
+    """
+    for key in _GOVERNANCE_SIGNAL_KEYS:
+        if key in parent_meta:
+            child_meta[key] = parent_meta[key]
+
+
 async def _live_duplicate_hashes(
     sc,
     *,
@@ -692,7 +718,36 @@ async def _create_memory_pipeline(data: MemoryCreate) -> MemoryOut:
 
 async def _handle_auto_chunk_from_ctx(data: MemoryCreate, ctx: object) -> MemoryOut:
     """Auto-chunking branch using pipeline context enrichment results."""
+    from core_api.pipeline.compositions.write import (
+        build_auto_chunk_governance_pipeline,
+        build_persist_pipeline,
+    )
     from core_api.services.ingest_service import _chunk_content
+
+    # #852: apply the LLM's free-form verdict, which this branch had computed
+    # and then discarded. It belongs HERE rather than at the one call site so
+    # that the enforcement travels with the function that does the writing.
+    #
+    # First statement in the function, ahead of the chunking call, for two
+    # reasons: a rejected write should not pay for the chunking LLM round-trip,
+    # and a policy that could not be applied must not be followed by rows it
+    # might have forbidden.
+    #
+    # The step mutates in place — ``fields["metadata"]`` for the PII flag,
+    # ``data.visibility`` for a ``keep_private`` downgrade — and both the parent
+    # payload and every child payload below read those. That is what makes the
+    # downgrade cascade rather than needing to be threaded through by hand, and
+    # it is why this runs before either payload is built.
+    governance = await build_auto_chunk_governance_pipeline().run(ctx)
+    if governance.failed:
+        # Not optional, and not merely tidy: the runner CATCHES a non-HTTP
+        # exception and returns ``failed`` instead of raising. Without this
+        # check a governance step that blew up would return here as an ordinary
+        # ``None`` result and the write would proceed ungoverned — the exact
+        # defect this function is being fixed for, reintroduced one level up.
+        # (An ``HTTPException`` — the 422 a ``drop`` policy raises — is
+        # re-raised by the runner and never reaches this line.)
+        raise HTTPException(status_code=500, detail="Memory governance pipeline failed unexpectedly")
 
     sc = get_storage_client()
     fields = ctx.data["memory_fields"]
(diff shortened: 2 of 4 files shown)


Eldad-Caura:
@erni-a ready for review.

All gates green, Claude review clean on the first pass ("No issues found"), nothing amended since — so no re-review was triggered.

Worth knowing before merging:

- **One flaky-test caveat, not caused by this PR.** One of three full-suite runs on this branch hit two interview-scheduler tests; they did not reproduce, `origin/main` was green twice, and the interview path cannot reach the changed code (`interview_service.py` writes only via `create_memories_bulk`). Filed as #858 with the captured assertions. The numbers table is in the PR body.
- **Two adjacent findings filed, not fixed here:** #855 (the `_USE_PIPELINE_WRITE=False` rollback lever routes to a path with no content governance at all) and #856 (deferred deployments never embed or enrich the auto-chunk parent, and nothing backfills it).
- **`test_child_dedup_minting_paths.py` is touched** — its `tenant_config` stand-in was missing the attributes the new gate reads. Swapped for the real `ResolvedConfig` with governance unconfigured, so those tests observe exactly what they did before.

erni-a (approved):

PR #859, opened 2026-08-21
fix(llm): one attempt must be one request

Eldad-Caura:
Follow-up to #854, which found the same unpinned `max_retries` on the
embedding client. The question worth answering first was whether that
conclusion transfers, because the LLM path has a real retry policy of its
own and pinning to 0 would be wrong if it did not. It does — and the
evidence here is stronger.

## The LLM path retries twice, in two nested layers

`common/llm/retry.py` is deliberate: `call_with_retry` takes
`LLM_RETRY_ATTEMPTS` attempts with linear backoff, and
`call_with_fallback` runs it **twice** — once for the primary provider,
once for the tenant-resolved fallback. So the real request count was the
product of three layers:

```
call_with_fallback     2 providers
  x call_with_retry    LLM_RETRY_ATTEMPTS = 2
    x the SDK          DEFAULT_MAX_RETRIES = 2  ->  3 requests
  = up to 12 HTTP requests for one logical call, where the code reads as 4
```

Invisible at every layer that could have caught it. SDK retries emit no
log line, so this — from the test run against the unpinned client —

```
common.llm.retry: test-full-budget attempt 1/2 failed (RateLimitError: 429)
```

is **one** logged attempt covering **three** real requests. Every stat,
dashboard and spend report counted it as one.

## Three harms, and the first is the one that decides it

**1. It overrode callers that had explicitly opted out of retrying.**
`recall_service.py:214` passes `max_attempts=1` — documented in
`call_with_fallback` as "pass 1 on latency-sensitive read paths (e.g.
recall) to fail fast to the fallback provider instead of retrying a
slow/hung primary". It got three tries anyway, with exponential backoff,
inside a 15 s budget chosen to make the fast failure possible. Measured:

| | requests |
|---|---|
| `max_attempts=1`, before | **3** |
| `max_attempts=1`, after | **1** |
| `LLM_RETRY_ATTEMPTS=2`, before | **6** |
| `LLM_RETRY_ATTEMPTS=2`, after | **2** |

**2. It broke ceilings whose derivation was written down.**
`config.py:76` sizes the inline cap as "35s leaves headroom for
nano-class LLM tail latency (typical p95 ~6-12s, plus 2 retries x 1s
linear backoff)". That counts one request per attempt. At three requests
of `openai_request_timeout_seconds` (25 s) each, one attempt is up to 75 s
— over the 35 s ceiling, and over `BULK_ENRICHMENT_TOTAL_TIMEOUT_SECONDS`
(30 s), on the **first** attempt rather than the third.

**3. Its backoff contradicted the stated policy.** `constants.py:41`
chose linear delay because "a multi-second backoff would push the request
past timeout" — while the SDK applied exactly that underneath, inside the
budget that choice was protecting.

## The cost, stated rather than glossed

The SDK honours `Retry-After` and jitters; `call_with_retry` does neither.
That matters more here than for embedding, because this one client class
serves three **hosted** APIs with real per-org rate limits (OpenAI,
Anthropic, OpenRouter) rather than a self-hosted backend.

It is still the right trade. Callers wrap each attempt in
`asyncio.wait_for` at 10-15 s (`dedup_judge`, `recall_service`), which
cancels a respectful `Retry-After` wait long before it completes — so the
compliance was mostly theoretical, while the multiplier was not. The real
fix is teaching `call_with_retry` about `Retry-After`, where it is visible
and inside the budget accounting, not keeping a second retry layer no
caller can see. Worth a follow-up.

## Scope

Only `OpenAILLMProvider` needed it. Checked the rest: `gemini.py` uses
`genai.Client` and `vertex.py` uses `GenerativeModel` (different SDKs,
their own policies — not touched here), `fake.py` makes no calls.

`LLM_PROVIDER_MAX_RETRIES` uses the keyword-only `minimum=0` that #854
added to `read_int_env`, so an explicit `0` in a manifest is *accepted*
rather than arriving via the fallback path.

## Verification

`4762 passed, 3 skipped, 1 xfailed` — full suite, zero failures. All 12
CI ruff scopes clean.

**All 4 new tests fail against the unpinned client** (3, 3, 6 and
`max_retries == 2` against expected 1, 1, 2 and 0). They assert at the
HTTP boundary via a `MockTransport`, answering 429 — the status the SDK
considers retryable — because the multiplier lives *below* the provider,
which is exactly what made it invisible. Mocking the provider would have
proved nothing.

### DIFF
--- a/common/llm/constants.py
+++ b/common/llm/constants.py
@@ -9,6 +9,7 @@ from __future__ import annotations
 import os
 
 from common.env_utils import read_float_env as _read_float_env
+from common.env_utils import read_int_env
 
 # ── Provider model defaults ──────────────────────────────────────────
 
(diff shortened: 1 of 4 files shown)


Eldad-Caura:

All 9 checks green. Claude review clean — though worth noting it opened with "comments-only change here", which isn't right: `max_retries=LLM_PROVIDER_MAX_RETRIES` in `common/llm/providers/openai.py` is the functional line. It does name and clear all three source files, so it did look, but the one-line behaviour change is the thing to eyeball.

What it does, and why it isn't just a copy of #854: the LLM path has its own real retry policy, so pinning to 0 needed justifying rather than assuming. It turned out to have **two** nested layers — `call_with_retry` at `LLM_RETRY_ATTEMPTS=2`, run twice by `call_with_fallback` (primary + fallback) — so the SDK's default made the count a three-way product: **up to 12 HTTP requests where the code reads as 4**.

The finding that decides it: `recall_service.py:214` passes `max_attempts=1` specifically to fail fast to the fallback provider instead of retrying a slow primary, and got **3** tries anyway with exponential backoff inside a 15s budget. A caller explicitly opted out and was silently overridden.

Measured at the HTTP boundary, before -> after: `max_attempts=1` 3 -> 1; `LLM_RETRY_ATTEMPTS=2` 6 -> 2; `max_retries` 2 -> 0. All 4 new tests fail against the unpinned client.

Accepted cost is stated in the PR body rather than glossed: we lose the SDK's `Retry-After` compliance, which matters more here than for embedding because this client serves three hosted APIs. The follow-up is to teach `call_with_retry` about it — where it's visible and inside the budget accounting.

`4762 passed, 3 skipped, 1 xfailed`; all 12 CI ruff scopes clean; single commit on `e1732c1f`.

erni-a (approved):

PR #860, opened 2026-08-21
fix(auto-chunk): the deferred parent must be completed, not abandoned

### LINKED ISSUE #856: Deferred deployments: the auto-chunk parent is never embedded or enriched, and nothing backfills it

Eldad-Caura:
Fixes #856.

## The bug

`_handle_auto_chunk_from_ctx` has two exits, and only one of them completes the row.

The **0-1-fact fall-through** hands its context to `build_persist_pipeline`, whose `ScheduleBackgroundTasks` step carries a branch commented *"Strong mode (or no mode set)"* — this branch's exact case, since it never sets a `resolved_write_mode`. That branch publishes the enrich request and the embed request whenever either value is missing.

The **multi-fact exit** builds its rows by hand and never runs that step. It scheduled nothing but entity extraction.

Inline deployments never noticed — `ParallelEmbedEnrich` produced both values synchronously, so there was nothing to repair. Deferred deployments defer both, and the parent was inserted with `embedding: None` and no LLM-derived fields, with nothing ever coming back for it.

Measured by driving `_create_memory_pipeline` in both modes:

| | inline | deferred (before) | deferred (after) |
|---|---|---|---|
| parent embedding | present | **`None`** | `None`, backfill scheduled |
| parent title | set | **`None`** | `None`, backfill scheduled |
| `embedding_pending` | absent | **absent** | `True` |
| `enrichment_pending` | absent | **absent** | `True` |
| backgrounded tasks | none | **none** | `enrich_or_publish`, `embed_or_publish` |
| children | 2, embedded | 2, embedded | 2, embedded |

So the row was permanently unembedded and unenriched: no title, summary, tags, classified type or temporal fields, and `embedding IS NULL`, hence unreachable by semantic recall. The children were fine, which is the shape of the symptom — a document whose chunks retrieve while the body holding them does not.

It wasn't marked either. `MemoryOut.metadata` documents an absent flag as *"that stage ran inline"*, so the row actively claimed the opposite of the truth, and `enrichment_pending` is the only flag core-worker clears.

## The fix

Schedule the same two backfills the sibling exit already gets, on the same conditions, and set the two markers it gets for free from `write_memory_row` / `MergeEnrichmentFields`.

`_enrichment_backfill_needed` lifts the step's four-clause condition rather than approximating it. All four matter: an inline deployment already enriched, and a tenant with enrichment disabled or on the `none` provider will never be enriched by anyone — claiming a pending pass for either would be a marker that never clears and a publish nothing consumes.

**Not reusing `ScheduleBackgroundTasks`**: it reads `ctx.data["memory"]`, which this branch never populates, and it also schedules Path A contradiction detection, which auto-chunk parents have never had and which is not this fix's to add. Restructuring the step so the two could share the pair would pull its `embedding is None` / `else: contradiction` arms apart for one caller's benefit. A test pins the two exits at parity instead — comparing the `tracked_task` labels each site uses — which is the property that was actually violated.

## The governance flag, which is not optional here

`_schedule_enrich_or_inline` carries a warning: a **third** call site *"MUST pass `run_governance_remediation=True` unless it has already applied `GovernanceDecision` synchronously for the same write"*. This is that third call site.

The answer is True, and not incidentally: #852 did add `GovernanceDecision` to this branch, but `defer_enrichment` implies `enrichment is None`, which is precisely the input on which that step takes its *"uncertain signal"* branch — it records `governance_llm_uncertain` and enforces nothing. So whenever this line can be reached, the verdict has not been applied.

It's inert in practice (the flag is read only on the inline arm, which `not settings.inline_enrichment` makes unreachable from here). The deferred arm publishes, and `core_api.consumer` calls `remediate_after_enrichment` when the ENRICHED back-channel returns — **which means this also closes the deferred half of #852**: before, a deferred auto-chunk parent was never enriched, so no verdict existed; now one is produced and the consumer applies it.

## Deliberately unchanged

- **Children are still embedded inline**, in one batched call, even in deferred mode. They're complete on insert; only the parent was ever incomplete. `get_embeddings_batch` is typed `-> list[list[float]]` and raises rather than returning gaps, so there is no partial-embed case to mark. A test pins this so changing it later is deliberate.
- **`_create_memory_legacy`'s auto-chunk branch has the identical gap** (verified by AST — it schedules `process_entity_extraction` and nothing else). Left alone per #855: that path is unreachable and ungoverned end to end, and repairing one of its many gaps inside a fix aimed at the live path is the bundling #847 is about.

## Tests

`tests/test_h856_deferred_auto_chunk_backfill.py`, 10 tests. Each rejected or pre-fix design applied to the fixed tree; exactly the intended tests failed:

|---|---|
| no backfill at all (the pre-fix branch) | 8 of the 10 |
| markers set unconditionally | the inline guard + both never-enriched guards |
| `run_governance_remediation` omitted | `..._requests_governance_remediation` |
| enrich gate ignores the tenant's config | both never-enriched guards |
| embed backfill passes no content hash | `..._carries_the_parents_own_content_hash` |
| embed backfill dropped | 6, including the parity guard |

Five of the ten guard against over-applying it: inline schedules nothing and marks nothing; enrichment-disabled and the `none` provider each get the embed half only; entity extraction still fires alongside; the children stay untouched.

One harness note: the recorder capturing scheduler arguments is deliberately a `def`, not an `async def`. These tests close each scheduled coroutine rather than awaiting it, so a replacement whose body only runs on await never runs — a test that asserts nothing while looking like it does. It caught itself on the first run.

## Gates

`ruff check core-api/src/` and `ruff format --check core-api/src/` clean, run separately. `mypy` clean (199 files). Full suite **4755 passed, 0 failed** — `main` at the time carried #857's 14 new tests (4731 + 14 = 4745), plus these 10.

Rebased onto `58e62ae4` after #859 landed; the adjacent suites (auto-chunk, governance, dedup, embed-off-hot-path, write pipeline) re-run green on the new base.

### DIFF
--- a/core-api/src/core_api/services/memory_service.py
+++ b/core-api/src/core_api/services/memory_service.py
@@ -150,6 +150,24 @@ def _auto_chunk_request_id() -> str:
     return f"auto-chunk:{uuid4()}"
 
 
+def _enrichment_backfill_needed(enrichment, tenant_config) -> bool:
+    """Does this row still owe an LLM enrichment pass?
+
+    The condition ``ScheduleBackgroundTasks`` applies on its strong / no-mode
+    branch, lifted so the auto-chunk handler — which never runs that step — can
+    ask the same question rather than approximate it. All four clauses matter:
+    an inline deployment already enriched, and a tenant with enrichment off or
+    on the ``none`` provider will never be enriched by anyone, so claiming a
+    pending pass for either would be a marker that never clears.
+    """
+    return bool(
+        enrichment is None
+        and not settings.inline_enrichment
+        and tenant_config.enrichment_enabled
+        and tenant_config.enrichment_provider != "none"
+    )
+
+
 # The LLM governance verdict, as ``MergeEnrichmentFields`` / the enrichment
 # worker record it on the parent's metadata.
 _GOVERNANCE_SIGNAL_KEYS = ("contains_pii", "pii_types", "business_relevance")
@@ -798,6 +816,18 @@ async def _handle_auto_chunk_from_ctx(data: MemoryCreate, ctx: object) -> Memory
         parent_metadata["auto_chunked"] = True
         parent_metadata["child_count"] = len(facts)
         parent_metadata["write_latency_ms"] = round((time.perf_counter() - t0) * 1000)
+        # #856: in a deferred deployment ``ParallelEmbedEnrich`` skipped both
+        # provider calls, so this row is incomplete. ``MemoryOut.metadata``
+        # documents absent flags as "that stage ran inline", which for this row
+        # is the opposite of the truth — so say so, on the same conditions the
+        # backfills below are scheduled on. The other exit of this function gets
+        # both flags for free: ``write_memory_row`` sets ``embedding_pending``
+        # and ``MergeEnrichmentFields`` sets ``enrichment_pending``.
+        defer_enrichment = _enrichment_backfill_needed(ctx.data.get("enrichment"), tenant_config)
+        if embedding is None:
+            parent_metadata["embedding_pending"] = True
+        if defer_enrichment:
+            parent_metadata["enrichment_pending"] = True
 
         # Auto-chunk parent insert — wrapped in the storage bulkhead
         # like the regular single-write path. Auto-chunk fires two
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #861, opened 2026-08-21
feat(llm): honour Retry-After, or hand the call to the fallback provider

Eldad-Caura:
The follow-up #859 named. Pinning the SDK's internal retries removed an
invisible 3x multiplier, but it also dropped the one thing that layer did
better than us: it honoured `Retry-After`. This rebuilds that in
`call_with_retry`, where it is logged and inside the budget accounting
instead of hidden underneath it.

## The interesting half is the refusal

Honouring the header unconditionally would be worse than ignoring it,
because every budget above this layer is small and fixed:

| | budget |
|---|---|
| `dedup_judge` per-attempt `wait_for` | 10 s |
| `recall_service` per-attempt `wait_for` | 15 s |
| `BULK_ENRICHMENT_TOTAL_TIMEOUT_SECONDS` | 30 s |
| `enrichment_inline_timeout_seconds` | 35 s |

A provider saying "come back in 60 s" is entirely normal. Sleeping that
out inside a 15 s budget spends the whole window asleep and lets the outer
timeout fire with nothing to show.

So a hint over `LLM_MAX_RETRY_AFTER_S` (5 s) **ends the retry loop**, which
is what lets `call_with_fallback` reach the second provider. That is the
right answer to "this provider is rate-limited for the next minute", and
it is a decision the old SDK-level retries structurally could not make:
they had no idea a fallback existed. All they could do was wait.

Measured, primary asking for 60 s:

| | before | after |
|---|---|---|
| primary attempts | 3 | **1** |
| sleeping before the hop | 3 s | **0 s** |
| reaches the fallback | eventually | immediately |

5 s as the cap because it has to leave room for the retry it precedes: the
tightest budget is 10 s and one request can take
`OPENAI_REQUEST_TIMEOUT_SECONDS`.

## Honouring, when it fits

`max(our_backoff, asked)`, never replacement. A hint SHORTER than our own
backoff would have us hammer a provider that is already refusing, so both
floors hold: never sooner than we planned, never sooner than we were told.

Parsing is provider-agnostic — duck-typed on `.response.headers`, because
this module is shared by the OpenAI-compatible, Gemini and Vertex
providers and imports none of them:

- `retry-after-ms` before `Retry-After`. Both are sent, the millisecond
form is the precise one, and the two genuinely disagree at the cap
boundary (5,200 ms vs "5" land on opposite sides).
- Delta-seconds **and** HTTP-date, since RFC 9110 §10.2.3 allows both and
providers send both.
- A date in the past clamps to 0 — clock skew must not become a negative
sleep.
- Anything unparseable is treated as no header at all. A provider sending
a malformed hint must not break the retry it was hinting about.

## Jitter, which is the same fix rather than a drive-by

`BULK_ENRICHMENT_CONCURRENCY` is 10, so a bulk write fires ten enrichment
calls at once. With a purely linear delay, a provider that rate-limits the
batch gets all ten back at the same instant one second later — which is
how a rate limit sustains itself. Both halves of this change are about not
hammering a provider that just said no.

Additive only, and applied AFTER the `max`, so it can never undercut a
`Retry-After`. `random` rather than `secrets` is deliberate: this
schedules a sleep, it does not protect anything.

## Not done here

`call_with_retry` still has no notion of its TOTAL remaining budget — only
the per-attempt `timeout` its callers pass. The 5 s cap is a fixed stand-in
for "small enough to fit any of them". Threading a real deadline through
would let the cap be derived instead of chosen, and would be a better
answer; it needs a signature change across ~10 call sites, so it is worth
its own PR rather than riding along in this one.

## Verification

`4778 passed, 3 skipped, 1 xfailed` — full suite, zero failures. All 12 CI
ruff scopes clean.

Of the 16 new tests, the 9 parsing tests cannot even import against
`main` (`retry_after_seconds` does not exist), and **5 of the remaining 7
fail** there: the header ignored (`[1.0] == [3.0]`), three attempts against
a 60 s hint (`3 == 1`), three primary attempts before the hop
(`['primary','primary','primary','fallback']`), and both jitter bounds.
The 2 that pass are deliberate controls describing behaviour this must NOT
change — a hint shorter than our backoff, and no hint at all.

### DIFF
--- a/common/llm/constants.py
+++ b/common/llm/constants.py
@@ -7,6 +7,7 @@ keeps re-exports for back-compat; new code should import from here.
 from __future__ import annotations
 
 import os
+import sys
 
 from common.env_utils import read_float_env as _read_float_env
 from common.env_utils import read_int_env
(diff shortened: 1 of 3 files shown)


Eldad-Caura:
Both findings confirmed and fixed.

## Malformed `retry-after-ms` masking a valid `Retry-After` — real

`common/llm/retry.py:69` had both parses inside one `try`, so `float(ms)` raising fell to the function-level `except` and returned `None` — discarding a well-formed seconds header sitting on the same response. "Prefer the precise one" quietly meant "ignore the other whenever the provider's ms header is broken", which is the opposite of the intent.

Fixed as suggested: the ms parse gets its own guard and falls through. Preferring the precise header should cost nothing when it turns out to be junk.

## Negative jitter fraction — real, and I moved where it's enforced

`read_float_env` does permit negatives, and a negative fraction makes `random.uniform` return a *subtracting* offset — sleeping less than the provider asked for, which is the one thing the ordering exists to prevent. A comment asserting a guarantee the code doesn't enforce.

Floored in **two** places, deliberately:

- at the point of use in `call_with_retry`, because that is what relies on the invariant — it now holds however the value arrived, including a future caller override or a test patch
- clamp-and-warn where the env var is read, so a bad config is visible to an operator rather than silently corrected

Worth noting how I got there: I first asserted on the constant, which needed `importlib.reload`. It passed alone and **failed under full-suite ordering** — a reference-staleness hazard, not a real defect, but a test that only works in isolation is not a test. Asserting at the point of use is both robust and the better home for the guarantee.

## Also corrected — my own verification claim

The PR body previously said "5 of the remaining 7 fail against unfixed source". That was measured before these two fixes added tests, and re-deriving it exposed two artifacts in my probe harness: a stub returning `None` made the "expect `None`" parsing tests pass trivially, and a BSD-sed pattern (`\?` is unsupported on macOS) silently failed to apply, turning failures into errors.

The body now states only what I actually reproduced — the three behavioural assertions with their exact output, and the two review-fix ones — plus the plain fact that **none** of the 20 tests can import against `main`, since neither new name exists there.

`4782 passed, 3 skipped, 1 xfailed`; all 12 ruff scopes clean; rebased on latest `main`, single commit, force-pushed.

Eldad-Caura:

All 8 checks green. Two Claude review cycles, both findings fixed and answered in-thread: a malformed `retry-after-ms` discarding a valid `Retry-After` beside it, and `LLM_RETRY_JITTER_FRACTION` accepting a negative value that would invert the additive-only guarantee.

The substance: #859 pinned the SDK's internal retries and that dropped its `Retry-After` compliance, so this rebuilds it at the layer that owns retrying. The half worth looking at is the **refusal** — every budget above `call_with_retry` is 10-35s and providers routinely ask for 60s, so honouring a hint unconditionally would spend the whole window asleep and let the outer timeout fire with nothing. A hint over the cap ends the retry loop instead, which is what reaches the fallback provider. Primary asking for 60s goes from 3 attempts and 3s of sleeping to 1 attempt and an immediate hop.

Jitter is in the same change rather than a drive-by: `BULK_ENRICHMENT_CONCURRENCY` is 10, so a rate-limited bulk write brings all ten back in lockstep one second later.

`4782 passed, 3 skipped, 1 xfailed`; all 12 ruff scopes clean; single commit on latest `main`.

One note on sequencing: #862 builds directly on this and is **blocked on it merging** — its stacked base means CI cannot run there. Merging this unblocks it.

erni-a (approved):

PR #862, opened 2026-08-21
feat(llm): give the retry loop a deadline it can actually enforce

Eldad-Caura:
Stacked on #861, which added `Retry-After` handling and capped it with a
fixed `LLM_MAX_RETRY_AFTER_S = 5.0` — a stand-in for "small enough to fit
any caller's budget", because this layer had no idea what its budget was.
Now it can be told, and the cap derived.

## The bug this exposes

The worst case of an LLM call is enforced by parameter VALUES, not by
structure. `recall_service.py:207` writes the arithmetic out:

> "One primary attempt (15s), then one fallback attempt (15s), then
> `_fake_recall` keeps the worst case ~30s"

That holds only while `max_attempts` stays 1. Restore the default — an
entirely reasonable-looking change — and it silently becomes 15 + 1 + 15 =
**31s per provider**, with only a comment saying otherwise. Same class as a
flag whose default degrades correctness: the property lives in a value
that anyone may change, not in a shape that resists it.

Tested as a pair, so neither half is vacuous:

| `max_attempts=2`, `timeout=15` | wall clock |
|---|---|
| no budget | **31s** |
| `budget_s=15` | **≤ 15s** |

## What a declared budget enforces

Three things stop being guesses:

1. **The loop stops when the budget is spent**, instead of starting an
attempt that cannot finish. Reachable with no sleep at all — an attempt
that runs to its own timeout can exhaust the budget by itself.
2. **Each attempt's timeout becomes `min(timeout, remaining)`.** Without
this the budget is advisory, enforced only by whoever cancels from
outside; an attempt could start with 2s left and run for 15.
3. **The `Retry-After` cap is derived** — half of what remains, so
honouring a wait always leaves at least as much time again for the attempt
it precedes. Waiting out a hint and then having no time to use it is the
same wasted request as not waiting at all.

The derived cap can **exceed** `LLM_MAX_RETRY_AFTER_S`, deliberately. That
constant is the answer for callers that declared nothing; a caller
declaring 60s has said it can afford to wait, and second-guessing it would
make the budget decorative.

Per PROVIDER, exactly like `timeout` — `call_with_fallback` forwards the
same value to both, so two providers at 15s is the ~30s worst case
`recall` describes, not one shared 15s.

## Why two call sites and not sixteen

I went looking to thread this everywhere and the code says not to:
**14 of the 16 `call_with_fallback` sites pass no timeout at all.** Giving
them budgets means inventing fourteen SLAs, which is a per-service
conversation, not a refactor.

And the two ceilings that looked derivable are not: both
`enrichment_inline_timeout_seconds` (35s) and
`BULK_ENRICHMENT_TOTAL_TIMEOUT_SECONDS` (30s) bound a **gather of N
concurrent calls**, so no individual call knows its share.

So it is wired at the two sites that already do the arithmetic by hand:

- **`recall_service`** — `budget_s=15.0`. No behaviour change today
(`max_attempts=1` already gives 15s); what changes is that the ~30s
promise now survives someone restoring the default.
- **`dedup_judge`** — `budget_s=10.0`. Per-attempt was its ONLY bound,
which bounds little: 2 providers x 2 attempts x 10s plus backoff is about
**42s worst case, on the write path, inside a 45s request budget**. Now
10s per provider, so ~20s before `_fake_dedup_check`.

Reaching that heuristic sooner is the intended trade, not a regression: a
duplicate check that has already spent 40s has failed at its job either
way, and the confidence floor documented below that call exists precisely
so a heuristic verdict can never sink a legitimate write.

`budget_s <= 0` raises, matching the existing `max_attempts` guard —
silently running one unbounded attempt would hide the misconfiguration
behind exactly the overrun the budget was meant to prevent. `monotonic`,
not `time()`, so an NTP step cannot move a deadline mid-retry.

## Verification

`4805 passed, 3 skipped, 1 xfailed` — full suite, zero failures. All 12 CI
ruff scopes clean.

Being precise about the baseline: `budget_s` does not exist on the parent
commit, so 10 of the 13 new tests fail at fixture level rather than at an
assertion (`retry_mod.time` is not even imported there), and
`test_a_non_positive_budget_is_a_misconfiguration` fails with `TypeError:
unexpected keyword argument`. The demonstration that carries weight is the
31s-vs-15s pair above, plus three controls that pass either way and pin
what must NOT change for the 14 callers with no budget: all attempts still
run, the fixed 5s cap still applies, and the linear backoff is untouched.

### DIFF
--- a/common/llm/retry.py
+++ b/common/llm/retry.py
@@ -18,6 +18,7 @@ from __future__ import annotations
 import asyncio
 import logging
 import random
+import time
 from collections.abc import Callable, Coroutine
 from datetime import UTC, datetime
 from email.utils import parsedate_to_datetime
@@ -104,6 +105,7 @@ async def call_with_retry(
     base_delay: float = LLM_RETRY_DELAY_S,
     timeout: float | None = None,
     non_retryable: tuple[type[BaseException], ...] = (),
+    budget_s: float | None = None,
 ) -> T:
     """Call *coro_fn* with retry and linear backoff.
 
@@ -119,6 +121,30 @@ async def call_with_retry(
         Base delay in seconds; actual delay = ``base_delay * attempt_number``.
     timeout:
         Optional per-attempt timeout in seconds.
+    budget_s:
+        Optional wall-clock budget for this whole call — every attempt plus
+        every delay between them. PER PROVIDER, exactly like *timeout*:
+        ``call_with_fallback`` forwards the same value to the primary and to
+        the fallback, so two providers means two budgets.
+
+        Given one, three things stop being guesses. Each attempt's timeout
+        becomes ``min(timeout, remaining)`` so an attempt cannot overrun the
+        budget; the loop stops once the budget is spent instead of starting
+        an attempt with no time to finish; and the cap on a ``Retry-After``
+        is DERIVED from what is left rather than read from
+        ``LLM_MAX_RETRY_AFTER_S``, which exists precisely because callers
+        without a declared budget give this layer nothing to reason from.
+
+        Why it matters more than tidiness: without it the worst case is
+        enforced by parameter VALUES rather than by structure. ``recall``
+        documents "one primary attempt (15s), then one fallback attempt
+        (15s) ... worst case ~30s" — a guarantee that holds only while
+        ``max_attempts`` stays 1, and that silently doubles if anyone
+        restores the default. A budget makes the same promise hold whatever
+        the attempt count is.
+
+        Omitting it keeps the previous behaviour exactly, which is right for
+        the callers that have no deadline of their own to declare.
     non_retryable:
         Exception types that must NOT be retried within this provider — the
         first occurrence raises immediately. Empty by default, so every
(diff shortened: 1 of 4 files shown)


Eldad-Caura:
⚠️ **CI has not run on this PR, and the single green check is misleading.**

Base is `honour-provider-retry-after` (#861), and every substantive workflow is gated on `branches: [main, dev]` — `ci.yml`, `codeql.yml`, `security-scan.yml`, `claude_code_review.yml`. Only `dco.yml` has no branch filter, which is why it is the one check showing. My mistake in choosing a stacked base; flagging it rather than leaving one green tick to look like a passing build.

Nor does waiting fix it by itself: GitHub's auto-retarget when #861 merges fires `pull_request` with action `edited`, and those workflows listen for `opened`/`reopened`/`synchronize`. It needs a push.

**Sequence, once #861 merges:** I rebase this branch onto `main` and force-push, which retargets the base and triggers the full suite. Not requesting review until that is green.

Rebasing onto `main` now is not an option — this PR modifies the `Retry-After` cap logic #861 introduces, so without it the branch does not build.

Local verification stands in the meantime: `4805 passed, 3 skipped, 1 xfailed`, all 12 CI ruff scopes clean. That is not a substitute for CI.

Eldad-Caura:
CI is now genuinely green: all 11 checks settled, 8 pass — `CI`, CodeQL, both `Analyze` jobs, Security Scan, dco, pre-checks, and the image-version guard. The earlier warning about a misleading single check no longer applies.

Rebased with `git rebase --onto origin/main f834abd8` rather than a plain rebase: #861 was squash-merged as a new SHA, so the branch still carried its original commit and a plain rebase would have replayed the duplicate. Back to a single commit on `main`, re-verified there — `4805 passed, 3 skipped, 1 xfailed`, all 12 ruff scopes clean.

@claude please review — this PR has not been reviewed yet. It was opened against #861's branch, so the review workflow was branch-gated out (`branches: [main, dev]`), and the force-push after retargeting counts as `synchronize`, which skips.

Eldad-Caura:
Confirmed and fixed — and it's mine to own: before this PR every `break` sat inside the `except` block, so `last_exc` was always assigned. The new loop-top guard is the first path that can reach `raise last_exc` with nothing caught.

## Reproduced, not just accepted

Removing the fix and running the new test gives exactly the predicted failure:

```
E       TypeError: exceptions must derive from BaseException
common/llm/retry.py:326: TypeError
```

A type error standing in for a perfectly clear condition, which is the part that matters — an operator seeing that would have no idea the budget was the cause.

## The fix, and one thing added to the suggestion

Synthesises a `TimeoutError` naming the budget, **but only when there is nothing to report**. If an earlier attempt failed for a real reason, that exception is the useful one and running out of time afterwards must not overwrite it — the same precedence the retry paths already apply, and the same lesson as the `last_exc` handling in #850. That has its own test (`test_a_real_failure_outranks_the_synthetic_timeout`), which passes either way by design: it pins that the new branch does not *lose* the behaviour.

I did not add the defensive `if last_exc is None: raise RuntimeError(...)` at the raise site. With every break path now assigning it, that guard would be unreachable — untestable code asserting an invariant the three call sites already hold. The `# type: ignore[misc]` on that line is honest again.

## On the test's determinism

Driven by a clock that jumps between the deadline being computed and the check reading it, rather than by a sub-microsecond `budget_s`. A real tiny budget does trigger it, but only because two `monotonic()` calls happen to differ by ~100 ns — that is a race, and the thing under test is the control flow.

## Also corrected

The commit body opened "Stacked on #861", which stopped being true when #861 merged and this rebased onto `main`. Now reads "Follows #861" — worth a force-push because it is the permanent record.

`4807 passed, 3 skipped, 1 xfailed`; all 12 ruff scopes clean; single commit on `main`.

Eldad-Caura:
@claude please re-review — the `last_exc is None` path now synthesises a `TimeoutError` naming the budget, and only when nothing was caught, so a genuine earlier failure still outranks it.

Eldad-Caura:

All 8 checks green, review clean on the second pass after one Medium finding (a `raise None` reachable from the new loop-top guard, which surfaced as `TypeError: exceptions must derive from BaseException` — fixed and reproduced in a test).

**One correction to the review's summary, because it is the thing worth your eye.** It says both call sites pass `budget_s` equal to their existing `timeout`, "which is a no-op for current `max_attempts` values". True for `recall_service` (`max_attempts=1`, so 15s either way — the value there is that the ~30s promise now survives someone restoring the default). **Not true for `dedup_judge`**, which runs the default 2 attempts: `budget_s=10.0` takes its worst case from roughly 42s across both providers to ~20s. That is a real behaviour change, on the write path.

I believe it is the right one — per-attempt was its only bound, and 42s inside a 45s request budget means a duplicate check that has already failed at its job — and the confidence floor documented directly below that call exists so a heuristic verdict can never sink a legitimate write. But it is a latency/behaviour trade on a write path rather than pure future-proofing, so it deserves a human yes rather than mine.

Scope note: wired at two sites, not sixteen. 14 of the 16 `call_with_fallback` callers declare no timeout at all, and the inline (35s) and bulk (30s) ceilings bound a gather of N concurrent calls rather than one call — so no individual call knows its share. Budgets for those are a per-service SLA conversation, not a refactor.

`4807 passed, 3 skipped, 1 xfailed`; all 12 ruff scopes clean; single commit on `main` (`e3267c4c`).

erni-a (approved):

Issue #858, opened 2026-08-21
Flaky: two interview scheduler tests fail on a job-claim race (1 in 3 full-suite runs)


Eldad-Caura:
Observed **once in three full-suite runs** while measuring #857's blast radius. Filing with the evidence rather than the diagnosis — it did not reproduce, so the mechanism below is inference from the code, not a confirmed root cause.

## What failed

```
FAILED tests/test_api_interview.py::test_schedule_queues_once_then_respects_pending_and_dueness
FAILED tests/test_interview_async_submit.py::test_schedule_sweep_processes_pending_jobs
```

Both in the same run; the other four runs of the same trees (two on `origin/main`, two on the branch) were fully green.

## The captured assertions

`test_schedule_sweep_processes_pending_jobs`:

```
    assert summary["jobs_processed"] >= 1     # passed
>   assert summary["jobs_done"] >= 1
E   assert 0 >= 1
```

`test_schedule_queues_once_then_respects_pending_and_dueness`, on the third scheduler run ("command done + interview fresh → NOT due → nothing queued"):

```
>   assert len(await _interview_commands(client, tenant_id, headers)) == 1
E   AssertionError: assert 2 == 1
```

No exception or error was logged in either.

## Why these two look like the same race

`jobs_processed >= 1` with `jobs_done == 0` means the sweep picked jobs up and none reached `committed`. `process_pending_interview_jobs` names that outcome itself:

> No-op: missing/done doc, **or a concurrent run claimed the job between the sweep's fetch and our claim** — nothing was retried.

which increments `jobs_skipped`. The claim race is plausible here because `synthesis_sem` is a module-level `asyncio.Semaphore(INTERVIEW_SWEEP_CONCURRENCY)` explicitly *"shared with the route's fire-and-forget path"* — so the sweep and any still-running submit task from an earlier test contend on one process-wide gate, and which one claims a job depends on scheduling.

The dueness failure is consistent with the same thing one step earlier: `_advance_watermark` returns early — **without stamping `last_interview_at`** — when `cursor_to <= existing_seq`. A submit whose watermark write lost that race leaves `last_interview_at` unset, so `_is_due` says due and the third scheduler run queues a second command, which is exactly the `2 == 1` seen.

## Why it matters

[...]

PR #863, opened 2026-08-22
ci: gate the rename programme's rule 7 instead of trusting review

Eldad-Caura:
Wave 0 of the sunset programme — the prerequisite the plan implies but does not schedule.

## Why this is first

Hard rule 4 says the do-not-touch list becomes CI. **It is not CI.** The only occurrences of the old brand in `.github/workflows` today are the Postgres role and database — which are sub-wave 3b's own target — plus comments. So rule 7, *mint nothing new carrying the old brand*, is currently enforced by whoever happens to be reviewing.

The programme runs for weeks with ordinary feature work alongside it. Every old-brand name that lands meanwhile adds a redirect somebody keeps forever and spends a rename option already paid for. The gate is a grep; what it prevents is doing Wave 5 twice.

## What it does

`scripts/legacy_name_ratchet.py` counts the lines carrying the old brand **per file**, in the tree being built and in `main`, and fails on any file that went up. Decreases are the point of the programme, so they're reported as progress:

```
No new lines. 47 removed by this change (-47 net).
```

Two design calls worth stating, because both were the difference between a gate that survives the programme and one that gets switched off:

- **Per file, not per repo total.** A total hides an addition behind an unrelated deletion in the same PR — during a programme whose whole shape is "totals go down", which is exactly when nobody looks. A per-file count also keeps a within-file move at net zero, which is most of what waves 3–5 do; a per-line diff would flag every move.
- **Against `main`'s tree, not a committed baseline.** A baseline goes stale, churns on every move, and has to be regenerated by the same person the gate is pointed at. On a `pull_request` the checkout is already the merge commit, so comparing against `main` *is* this PR's contribution, and there is nothing to maintain. Same anonymous `--depth=1` fetch the broker-contract gate already uses.

## The escape hatch, which rule 3 requires

Old names stay readable forever, so legitimate additions exist — a compat alias, a redirect, a test pinning the old wire format. A line carrying `legacy-name-ok` is exempt:

```python
LEGACY_TOOL_ALIAS = "..."  # legacy-name-ok: permanent shim, plugins pinned to v1
```

It exempts **only its own line** — otherwise one marker at the top of a file would silence the file — so every exception lands in the diff with its reason, where a reviewer can disagree. `CONTRIBUTING.md` documents the rule, the marker and the two commands.

## The hole it found in itself

`git grep` exits 1 for "no matches" **and** for a pathspec that matches nothing, both with an empty stderr. So a broken pattern or pathspec reads as a clean tree, and the gate passes every PR from then on without checking anything — a silent no-op, the worst failure available to a gate.

`main()` now refuses to pass when *both* trees come back empty: that is either a broken scan or a finished migration, and both want a human.

A first draft carried a `returncode == 1 and not stderr` guard that looked like it handled this. Measuring the actual exit codes showed a bad ref exits 128, so the clause was unreachable and the fail-closed test was passing for the wrong reason. Removed, and replaced with a guard for the case that is real.

## Tests

`tests/test_legacy_name_ratchet.py`, 13 tests, each building a throwaway git repo and running the real script. Mocking `git grep` would test the arithmetic and none of the part that breaks — output parsing, the tree-ish prefix and case-insensitivity are all git's behaviour.

Every design was simulated against the finished tree; in each case exactly the intended tests failed:

|---|---|
| exempt marker not honoured | both marker tests |
| marker exempts the whole file | `..._exempts_only_its_own_line` |
| repo total instead of per file | 4, incl. `..._not_masked_by_a_deletion_elsewhere` |
| split one field too far | 9 |
| pathspec matches nothing | 10 |
| no both-empty guard | `..._finding_nothing_at_all_fails_rather_than_passing` |

The split-field case is the one worth keeping: it corrupts the path into a fragment of the matched line, so the real file's count never rises and the gate passes while appearing to work.

## Notes

- **The gate failed this PR twice while I wrote it** — first on the script's own pattern constant, then on prose in my comments that mentioned the brand. Both fixed by rewording rather than exempting, which is the behaviour it should encourage. It passes itself now.
- The branch was originally named with the old brand. Renamed, for the same reason.
- Current footprint here: **2,463 lines across 280 files**. `--report` prints the breakdown.
- **This repo only.** Replicating to the other repos is a follow-up, one at a time, per rule 6.

## Gates

`ruff check` clean on both new files. `actionlint` reports one shellcheck finding at `ci.yml:87`, which is pre-existing — it reproduces on `origin/main` — and none in the added job. Full suite **4785 passed, 0 failed**. Rebased onto `bdb20a24` after #861 landed; gate and tests re-verified on the new base.


Eldad-Caura:
@claude re-review please.

The round restructured the parsing rather than tweaking around it, which is the case worth a second pass: output is now read from `git grep -z` (NUL-separated fields) instead of colon-splitting, and the failure report re-greps with a `:(literal)` pathspec.

Both changes came from the same root as the Low you found — a delimiter appearing inside a field — so I would rather have the new parse looked at than assume it closed the class.

Two tests added, each confirmed to fail without its fix. Worth flagging that the first draft of the glob test did **not** fail without the fix: it asserted the offending file was named and its line shown, both true either way, because the bug over-matches (`a[b].py` also matches `ab.py`) rather than under-matching. Rewritten to pin what actually breaks — an innocent file's line must not appear under a guilty file's heading.

Also added `timeout-minutes: 5` per the suggestion.

Eldad-Caura:
@claude re-review please — third round, and the second consecutive one to find something real in this file, which is the signal to keep looking rather than stop.

Applied. The serious half was right: a bare substring test meant prose containing the marker inside a longer word (`legacy-name-okay`) silently exempted a real occurrence — a hole in the gate, and this PR adds a CONTRIBUTING section that writes the marker out repeatedly, which is exactly how that gets typed by accident.

Now a bounded, case-insensitive token: marker followed by end-of-line, whitespace, or the colon of `: <reason>`. Deliberately **not** required to sit in a comment — comment syntax differs across the file types this scans, and JSON has none at all while still being somewhere a compat alias can legitimately live.

Three tests added, each confirmed to fail against the substring check: the longer-word case is not exempt, a differently-cased marker is, and a bare marker with no reason still works (a reason is asked for, not enforced).

Worth a look at the regex boundary specifically — over-matching there reopens the hole, under-matching turns deliberate annotations red.

Eldad-Caura:
Applied — and it is the exact mirror of the previous round: I bounded the marker on the right and not the left, so `somelegacy-name-ok` exempted a line nobody annotated.

Now bounded both sides, and they are genuinely separate checks rather than one tidied up: the lookahead stops `legacy-name-okay`, the lookbehind stops `somelegacy-name-ok`, and either alone leaves the hole open from the other end. One more test, confirmed to fail with the right boundary only. 19 tests.

Three rounds, three real findings, every one in the exemption predicate rather than in the counting the gate exists to do — which is [#847](https://github.com/caura-ai/caura/issues/847) in miniature and the reason each round got re-tagged instead of being called done.

@claude one more look at the predicate, please. If this round comes back clean I will stop.

Eldad-Caura:
@erni-a ready for review.

All gates green. Four review rounds, each one applied — the last found something I want to flag explicitly rather than bury, because it changes what the PR claims.

**The limit, stated rather than closed.** Marking an *existing* line exempt frees a slot in that file, which a new unmarked line can then fill: non-exempt flat, gate passes, a real addition through. That is not fixable by counting, and not a shortcoming of the implementation — legitimately adding one marked alias has the identical per-file signature (non-exempt flat, exempt +1, total +1), because the difference is *which* line carries the marker, not how many do.

So it is reported instead: every newly exempted line is printed with its file, line and reason, **including on the passing path**, which is the path that swap takes. An exemption is the one move that buys headroom, so it is the one that must never be quiet. The docstring and CONTRIBUTING say exactly this, so the guarantee claimed matches the guarantee delivered.

Rounds 2 and 3 were the mirror halves of one bug in the exempt-marker predicate (unbounded right, then unbounded left). Round 1 was the colon-in-path parse. All have tests confirmed to fail without their fix — 21 in total.

**One caveat on my local full-suite run:** 4827 passed, 1 failed — `test_schedule_sweep_processes_pending_jobs`, which is #858. It is not this PR: the diff touches no service code at all (a script, a test file, a CI job, a CONTRIBUTING section). While confirming that I root-caused #858, and my original diagnosis there was wrong — it is accumulated state in the shared local test DB, because those tests mint `t-`-prefixed tenants and the conftest sweep only matches `test-tenant-%`. 0 of 228 affected tenants get cleaned. CI gets a fresh database, which is why CI is green. Written up on the issue; the fix is a one-line prefix change and I would take it next if you want it.

erni-a (approved):

PR #864, opened 2026-08-22
docs(llm): say why the Vertex and Gemini SDK imports are deferred, and enforce it

Eldad-Caura:
I set out to hoist `vertex.py`'s function-level imports to module scope,
which is what the top-level-imports rule asks for. **That would have been a
regression**, so this documents and enforces the exception instead.

## Why hoisting breaks

The deferral is about optional dependencies, not circular imports:

* `google-cloud-aiplatform` is an **extra** for core-worker
  (`vertex = [...]` in its `pyproject.toml`), and only a hard dependency
  for core-api.
* `google-genai` appears **zero times** in core-worker's
  `pyproject.toml` — not required, not optional. core-worker never has it.

Verified rather than assumed, by blocking `vertexai` and
`google.cloud.aiplatform` on the import path:

```
A. import common.llm.providers.vertex  ->  OK (module loads without the SDK)
   class present: True
B. top-level SDK import -> ModuleNotFoundError: No module named 'vertexai'
   => hoisting makes vertex.py unimportable in a no-extra install
```

Deferred, an install without Vertex fails only if it actually CALLS Vertex
— which is what an optional provider should do. Hoisted, the module cannot
be imported at all, and `common/llm/_platform.py` imports it inside a `try`
on the strength of exactly that property.

## Why it read as an oversight

`gemini.py:45` already explains its own lazy import — and cites this file
as the precedent: *"same pattern as VertexLLMProvider with vertexai"*. So
the file everything else points at was the one copy that never said why.
That is the actual defect, and it is why I flagged it as a rule violation
in the first place.

`registry.py` also deliberately omits `vertex` from the top-level provider
imports it does for fake/gemini/openai. Those three placements are one
decision; changing any alone breaks it, which the module docstring now
says.

## A comment cannot stop the next hoist

`tests/test_optional_provider_sdks_stay_optional.py` pins the invariant for
both providers: with the SDK hidden behind a meta-path finder, each module
must still import and still expose its provider class.

Without it the rule argues for the change and nothing argues back until
something fails at deploy time in the one install that lacks the package.
Confirmed to bite: hoisting the import makes
`test_module_imports_without_its_sdk[vertex-without-aiplatform]` fail with
the simulated `ModuleNotFoundError`.

Each case carries a control asserting the blocker really does hide the
package — otherwise the guard would pass no matter where the imports lived.
Raising from `find_spec` rather than falling through is what makes it
absence and not a redirect; falling through would resolve from
site-packages, where these packages do exist in the dev environment.

## Verification

`4811 passed, 3 skipped, 1 xfailed` — full suite, zero failures. All 12 CI
ruff scopes clean. No behaviour change: comments, a docstring, and a new
test.

### DIFF
--- a/common/llm/providers/vertex.py
+++ b/common/llm/providers/vertex.py
@@ -7,6 +7,28 @@ the event loop.
 CAURA-333: ``VertexEmbeddingProvider`` was removed (broken — never passed
 ``output_dimensionality`` to the SDK, so writes failed against pgvector's
 1024-dim column). Only the LLM-side provider remains.
+
+⚠ THE SDK IMPORTS IN THIS MODULE ARE DEFERRED ON PURPOSE, against the
+repo's top-level-imports rule. Stated here because this file is cited as
+the precedent for the pattern (``gemini.py``: "same pattern as
+VertexLLMProvider with vertexai") and was the one copy that never said
+why — so it reads as an oversight and gets re-flagged.
+
+The reason is optional dependencies, not circular imports.
+``google-cloud-aiplatform`` is an EXTRA for core-worker
+(``vertex = [...]`` in its ``pyproject.toml``), and only a hard dependency
+for core-api. Hoisting these to module scope makes
+``common.llm.providers.vertex`` unimportable wherever the extra is absent
+— verified by blocking ``vertexai`` and ``google.cloud.aiplatform`` on the
+import path: the module still imports cleanly today and
+``VertexLLMProvider`` is still reachable, while a top-level import raises
+``ModuleNotFoundError``. Deferred, an install without Vertex fails only if
+it actually CALLS Vertex, which is what an optional provider should do.
+
+``_platform.py`` imports this module lazily too, inside a ``try``, for the
+same reason — and ``registry.py`` deliberately omits it from the top-level
+provider imports it does for fake/gemini/openai. All three are one
+decision; changing any of them alone breaks it.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@erni-a ready for your approval — though the thing worth your eye is what this PR does NOT do.

I set out to hoist `vertex.py`'s function-level imports to module scope, which is what the top-level-imports rule asks for. **That would have been a regression**, so this documents and enforces the exception instead.

`google-cloud-aiplatform` is an EXTRA for core-worker (`vertex = [...]`) and a hard dependency only for core-api; `google-genai` appears **zero times** in core-worker's `pyproject.toml` — not required, not optional. Verified by blocking both on the import path: today the module imports cleanly and `VertexLLMProvider` is reachable, while a top-level import gives `ModuleNotFoundError`. `_platform.py` imports it inside a `try` relying on exactly that property.

The real defect was that `gemini.py:45` already explains its own lazy import **and cites this file as the precedent** ("same pattern as VertexLLMProvider with vertexai") — so the file everything points at was the one copy that never said why. Hence it reading as an oversight, which is how I came to it.

And because a comment would not have stopped me: `tests/test_optional_provider_sdks_stay_optional.py` hides each SDK behind a meta-path finder and asserts the module still imports. Hoisting the import makes it fail with the simulated `ModuleNotFoundError` — checked. Each case carries a control proving the blocker really hides the package, since raising from `find_spec` rather than falling through is what makes it absence rather than a redirect.

If you would rather these providers were restructured so the rule holds without an exception — a separate module per provider under an optional-import boundary, say — that is a design change worth its own PR, and I would not want to smuggle it in here.

All 8 checks green, review clean. `4811 passed, 3 skipped, 1 xfailed`; all 12 ruff scopes clean; no behaviour change. Single commit on `main`.

erni-a (approved):

Issue #865, opened 2026-08-22
The conftest tenant sweep matches almost nothing: 12,714 rows leak per local DB, 0 reclaimed


Eldad-Caura:
Found while root-causing #858. That issue is one instance; this is the class.

## The sweep reclaims almost nothing

`tests/conftest.py` cleans up at end of run with `DELETE FROM {table} WHERE tenant_id LIKE 'test-tenant-%'` across seven tables. Measured on a local `memclaw_fresh` that had been running the suite for a day:

```
  table                       swept prefix   NOT swept
  relations                              0           0
  entities                               0          24
  memories                               0        4142
  audit_log                              0        7414
  audit_chain_head                       0         462
  agent_activity_digests                 0           0
  documents                              0         672
  TOTAL                                  0       12714
```

**Zero rows matched.** Distinct leaked tenants, by prefix:

| tenants | prefix | |
|---|---|---|
| 1089 | `t-` | interview tests — fixed in #866 |
| 132 | `tenant-` | |
| 108 | `test-search-pipe` | |
| 72 | `rep-tenant` | `test_reports.py` |
| 48 | `test-pipeline` | |
| 36 | `self-org` | `test_security_tenant_isolation.py` |
| — | `ent-org-a`/`-b`, `rep-adm1-*`, `default` | |

The prefix is load-bearing — the sweep is the only thing that ever removes rows written through the service layer, because most tests write via a storage client that commits on its own connection and the per-test session rollback never sees them — but it was a convention in a docstring. Nearly every test file picked a prefix that reads well for its own subject instead.

## Why it matters, and why CI has not caught it

CI provisions a fresh Postgres per run, so residue never builds and every test is sound there. A local database accumulates forever, and the failure mode is not "the table is big" — it is that accumulated rows silently change what a test observes:

- **#858:** interview job documents from 228 tenants, and a *cross-tenant* sweep endpoint under test began reading other runs' residue. Went from intermittent to deterministic over a day, and now fails in isolation.
- The conftest already documents the same mechanism for keystones: *"the keystone listing is capped at 50 and ordered by weight, so accumulated rules eventually push a freshly-written one out of the response and a round-trip test fails for reasons that look nothing like accumulated state."*

Any test asserting on a count, a cap, an ordering, or anything cross-tenant is exposed. The symptom is always a failure that looks like a bug in the code under test.

## What #866 does and does not fix

[...]

PR #866, opened 2026-08-22
test(interview): mint tenants the end-of-run sweep can actually see

### LINKED ISSUE #858: Flaky: two interview scheduler tests fail on a job-claim race (1 in 3 full-suite runs)

Eldad-Caura:
Fixes #858.

## Not a race — accumulated state, and my original diagnosis was wrong

I filed #858 as a job-claim race on the shared `synthesis_sem`, inferred from the code because it would not reproduce. It reproduced today — **deterministically, and in isolation**, which the report explicitly said it did not do. That difference is the answer.

`tests/conftest.py` cleans up with `DELETE FROM {table} WHERE tenant_id LIKE 'test-tenant-%'`, and `documents` is in its table list. The two interview files minted tenants as `get_test_auth(f"t-{uid()}")` — a prefix that `LIKE` does not match. So every interview job either file had ever written was still there:

```
interview_jobs   done              286
                 failed_permanent   12
                 processing          2   <- reclaimed by the sweep on every run

tenants with interview jobs:    228
matched by the conftest sweep:    0
```

The endpoint under test is **cross-tenant**: `process_pending_interview_jobs` loops over every tenant taking pending and stale-`processing` jobs. Those two stuck rows qualified on every run, from tenants no test knew about — so `jobs_processed >= 1` passed on residue while `jobs_done`, which counts only rows reaching `committed`, stayed 0. That is exactly the `1 processed / 0 done` shape observed.

CI provisions a fresh Postgres per run, which is why CI stayed green while a local database degraded from "1 in 3" to "every time" over a day of running the suite.

## The fix

34 call sites now use a new `conftest.new_tenant_id()`, and the prefix moves into `SWEEP_TENANT_PREFIX` — used by every minter (`TENANT_ID`, the `tenant_id` fixture, `new_tenant_id`) and by the `DELETE`s, which are parameterised on it instead of repeating the literal.

The prefix was load-bearing while documented as a convention in a docstring. This sweep is the only thing that ever removes rows written through the service layer, because most tests write via a storage client that commits on its own connection where the per-test rollback never reaches. One constant is what makes the two sides agree by construction rather than by everyone remembering.

## Verified by what stops accumulating

Not "the test is green" — it was green in isolation this morning too:

| | before | after |
|---|---|---|
| 3 consecutive runs of the two tests | fails | **3× pass** |
| `interview_jobs` + `interview_watermarks` left behind | ~100 per run, forever | **0** |
| full interview suites (47 tests) | — | pass, leaving 0 rows |
| full suite | 1 failed | **4815 passed, 0 failed** |

## Tests

`tests/test_tenant_sweep_coupling.py`, 4 tests pinning the coupling rather than any call site. Each confirmed to fail against the way it can break:

| simulated break | caught by |
|---|---|
| a minter hardcodes its own prefix | the minter + fixture tests |
| a `DELETE` hardcodes the pattern | the no-literal + binding tests |
| a table added to the sweep with an inline pattern | `..._binds_the_prefix` |
| the fixture diverges from the plain function | `..._fixture_mints_a_sweepable_id` |

Two notes worth keeping:

- The no-literal test **parses rather than greps**, so only code counts. Docstrings and comments here quote the prefix while explaining it, which is the opposite of a problem — written as a grep first, it failed on this change's own documentation.
- The third case is the realistic regression: a table appended to the sweep list with its own inline pattern would leak silently while every other table stayed clean.

## Scope

The interview files only. The same mismatch affects at least `test_reports.py`, `test_security_tenant_isolation.py` and others — **12,714 rows across seven tables were unswept on the measured database, of which 0 matched**. Filed as #865 with the numbers and three options; converting the rest, or making the sweep prefix-agnostic, is a decision about the fixture rather than a flake fix and wants taking as one change.

This PR's guard protects the conftest's internal coupling. It deliberately does not guard call sites — such a check would be red today, for the reason in #865.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -73,7 +73,13 @@ TEST_DB_URL = os.environ.get(
 # tenant without taking a fixture argument. None of them relies on cross-test
 # sharing, so they could each mint their own; threading a tenant through those
 # helpers is the follow-up that would let this go away.
-TENANT_ID = f"test-tenant-{uuid.uuid4().hex[:8]}"
+# The prefix the end-of-run sweep matches on, and the only place it is written.
+# The two sides that must agree — every minter here and the DELETE in
+# ``_setup_schema`` — were previously coupled by a note in a docstring, which is
+# what #858 cost: a file minting its own prefix leaked every row it ever wrote.
+SWEEP_TENANT_PREFIX = "test-tenant-"
+
+TENANT_ID = f"{SWEEP_TENANT_PREFIX}{uuid.uuid4().hex[:8]}"
 FLEET_ID = "test-fleet"
 AGENT_ID = "test-agent"
 
@@ -101,6 +107,24 @@ def get_admin_headers() -> dict:
 def uid() -> str:
     """Short unique suffix — for distinct content (409s) and distinct tenant ids."""
     return uuid.uuid4().hex[:8]
+
+
+def new_tenant_id() -> str:
+    """A tenant id unique to one test AND visible to the end-of-run sweep.
+
+    The prefix is not cosmetic: ``_setup_schema`` cleans with
+    ``tenant_id LIKE 'test-tenant-%'``, so a tenant minted with any other prefix
+    is never reclaimed and its committed rows outlive the run. That is how #858
+    happened — two interview files minted ``t-`` tenants, their job documents
+    accumulated across every local run, and a cross-tenant sweep endpoint under
+    test started reading other runs' residue.
+
+    Defined here rather than repeated at call sites so the prefix and the DELETE
+    that depends on it stay in one file. The ``tenant_id`` fixture is the same
+    value for tests that can take a fixture; this is for the ones that mint
+    several per test.
+    """
+    return f"{SWEEP_TENANT_PREFIX}{uid()}"
 
 
 # ---------------------------------------------------------------------------
@@ -170,7 +194,8 @@ async def _setup_schema(_engine):
         ):
             try:
                 await conn.execute(
-                    text(f"DELETE FROM {table} WHERE tenant_id LIKE 'test-tenant-%'")
+                    text(f"DELETE FROM {table} WHERE tenant_id LIKE :prefix"),
+                    {"prefix": f"{SWEEP_TENANT_PREFIX}%"},
                 )
             except Exception:
                 # Best-effort, and the ONLY thing that ever removes rows written
@@ -186,8 +211,9 @@ async def _setup_schema(_engine):
             await conn.execute(
                 text(
                     "DELETE FROM memory_entity_links WHERE memory_id IN "
-                    "(SELECT id FROM memories WHERE tenant_id LIKE 'test-tenant-%')"
-                )
+                    "(SELECT id FROM memories WHERE tenant_id LIKE :prefix)"
+                ),
+                {"prefix": f"{SWEEP_TENANT_PREFIX}%"},
             )
         except Exception:
             pass
@@ -304,9 +330,10 @@ def tenant_id():
     reason, or failing only when the run order changed. Nothing removes those rows
     mid-run; see the sweep in ``_setup_schema`` for why.
 
-    Keep the ``test-tenant-`` prefix: that sweep matches on it.
+    Keep the ``test-tenant-`` prefix: that sweep matches on it — see
+    :func:`new_tenant_id`, which owns it.
     """
-    return f"test-tenant-{uid()}"
+    return new_tenant_id()
 
 
 @pytest.fixture
--- a/tests/test_api_interview.py
+++ b/tests/test_api_interview.py
@@ -10,7 +10,7 @@ from datetime import UTC, datetime, timedelta
 import pytest
 
 import core_api.services.interview_service as interview_service
-from tests.conftest import get_test_auth, uid
+from tests.conftest import get_test_auth, new_tenant_id, uid
 
 
 # ── helpers ──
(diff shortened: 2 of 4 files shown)


Eldad-Caura:
@claude review please.

The auto-review skipped this one under `pre-checks` condition 6 — *"only tests/config/docs files changed, no source to review"* — which is a fair heuristic that happens to be wrong here: the change is in `tests/conftest.py`, the highest-fan-in file in the suite, and it alters the cleanup semantics every test depends on.

Worth looking at specifically:

- The `DELETE`s are now parameterised on `SWEEP_TENANT_PREFIX` rather than repeating the literal. If that binding is wrong, cleanup silently stops working — which is the exact class of bug this PR fixes, so getting it wrong here would be poor form.
- `SWEEP_TENANT_PREFIX` moved above `TENANT_ID` so the module-level constant can derive from it. Ordering matters at import time.
- 34 mechanical call-site swaps in the two interview files, `get_test_auth(f"t-{uid()}")` → `get_test_auth(new_tenant_id())`. Neither file asserts on tenant shape, but a missed site leaks silently rather than failing.
- The four guard tests pin the coupling rather than any call site; each is confirmed to fail against the way it can break. The no-literal one parses rather than greps deliberately — as a grep it failed on this PR's own documentation.

Verified by measurement rather than a green run: three consecutive runs leave `interview_jobs` and `interview_watermarks` at 0 rows, where each run previously added ~100 permanently.

Eldad-Caura:
@erni-a ready for review — with one caveat I want to be explicit about rather than let a green tick imply.

**This PR has not been substantively reviewed by the bot.** It skipped on open under `pre-checks` condition 6 (*only tests/docs/config changed*), and when I tagged it directly it returned "No issues found" while stating that everything in the diff is *"excluded from review per the task instructions"*. So that checkmark means "did not look", not "looked and found nothing". It rests on my own verification.

Worth knowing there is a real gap behind that: **`tests/conftest.py` cannot get an automated review**, because the exclusion is path-based and it lives under `tests/`. It is the highest-fan-in file in the suite. A one-line carve-out in condition 6 would fix it; I have not made that change here because it belongs with whoever owns that workflow.

What the change is: the two interview files minted `t-`-prefixed tenants while the conftest sweep only deletes `test-tenant-%`, so every interview job document they ever wrote survived every run — 300 documents across 228 tenants locally, and the endpoint under test sweeps cross-tenant, so it began reading other runs' residue. CI provisions a fresh database, which is why CI never saw it.

Verified by what stops accumulating rather than by a green run: three consecutive runs now leave those collections at **0 rows**, where each run previously added ~100 permanently. Full suite 4815 passed, 0 failed.

The prefix now lives in one constant used by every minter and by the `DELETE`s, with four tests pinning that coupling — each confirmed to fail against the specific way it can break.

Scope is the interview files only. The same mismatch leaves **12,714 rows across seven tables unswept, 0 matched**, filed as #865 with the numbers and three options — that one is a decision about the fixture, not a flake fix.

erni-a (approved):

PR #868, opened 2026-08-22
test(retry): stop a faked `asyncio.sleep` escaping the module under test

Eldad-Caura:
test(retry): stop a faked `asyncio.sleep` escaping the module under test

caura#863's CI run failed in `test_no_hint_keeps_the_linear_backoff` with

    assert [0.7359102269208245, 1.0, 2.0] == [1.0, 2.0]

on a PR that touches a shell script, a workflow job and CONTRIBUTING.md.
The test's own captured log names the intruder, so this needed no
guessing:

    WARNING common.http_retry storage_client.POST /memories/similar-candidates:
            ConnectError on attempt 3/5, retrying in 0.74s

That is a `detect_contradictions_async` background call, left running by
an earlier test, retrying a storage-api that does not exist in CI. Its
third backoff came due while the recorder was installed and was reported
as the LLM retry loop's own sleep. It logged before both `llm.retry`
lines, which is why the stray value is at index 0.

## The defect

    monkeypatch.setattr(retry_mod.asyncio, "sleep", fake)

does not patch `retry_mod`'s sleep. `retry_mod.asyncio` IS the one
`asyncio` module object, so that line replaces `asyncio.sleep` for every
importer alive — the code under test, every other module, and every task
already running. `patch("pkg.mod.asyncio.sleep", ...)` resolves the same
shared object. Five test files did this, twelve sites.

Two distinct costs, and the second is the one that has no assertion
watching it:

* **Mis-attribution.** Any task that sleeps during the window lands in
  the recorder. That is this failure.
* **Silencing.** Everyone else's backoff becomes a no-op, so a leaked
  retry loop spins through its remaining attempts at once instead of
  waiting. Nothing anywhere asserts that it did not.

`test_llm_retry_budget` had it worst. It fakes `time.monotonic` the same
way, which hands its fake clock to the whole process, and its fake sleep
ADVANCES that clock. A stray 0.74s there does not add a list entry — it
silently spends 0.74s of the budget the assertions are computing on.

## The fix

Patch the module REFERENCE, not the module CONTENTS: give the importer
under test a private copy of `asyncio` and leave the shared one alone.

    monkeypatch.setattr(mod, "asyncio", scoped(asyncio, sleep=fake))

`tests/_scoped_module.scoped` returns a real `ModuleType` holding a
shallow copy of the original namespace. A copy rather than a stub or a
forwarding proxy because the importers reach for more than `sleep`:
`retry.py` also uses `asyncio.wait_for`, and `memory_service` uses nine
asyncio names. An explicit stub would have to list them, and would break
— silently, in the `wait_for` case — the next time one was added.
Removing the namespace copy in a simulation fails 12 existing tests, so
that property is already covered by tests that exist for other reasons.

## Tests

Six, in the two files whose assertions the recorder feeds. Each pair
pins the property in both directions, because the two ways of getting
this wrong fail in opposite manners:

| test | fails without the fix |
|---|---|
| `..._sleep_taken_elsewhere_is_not_recorded` | yes — records `[0]` |
| `..._sleep_taken_elsewhere_still_really_happens` | yes — returns instantly |
| `..._a_sleep_elsewhere_does_not_spend_the_budget` | yes — clock moves |
| `..._the_real_clock_still_runs_for_everybody_else` | yes — 1000-and-change |
| `..._the_retry_loop_is_still_recorded` | vacuity guard |
| `..._the_retry_loop_does_still_move_it` | vacuity guard |

The last two do not fail against the old code, deliberately. They fail
against the realistic MIS-fix: a fixture scoped to the wrong module
records nothing at all, and every `assert slept == [...]` in both files
would then be asserting on a list no code path can fill. Simulated —
scoping the fixture at `common.http_retry` instead — and the suite takes
its sleeps for real and hangs, so the mis-scope is loud rather than
silent.

## What this does NOT fix

The leaked task itself. `detect_contradictions_async` is a post-commit
fire-and-forget call that outlives the test that started it and spends
~3s retrying an endpoint that is not there; it appears in the warnings
summary as `coroutine 'detect_contradictions_async' was never awaited`.
Filed separately — deciding what the fixture should do about background
tasks is not a flake fix.

Worth recording why it surfaced on #863 in particular. `pytest.ini` sets
`asyncio_default_test_loop_scope = session`, so nothing runs the loop
during a sync test; a block of them accumulates a leaked task's expired
timers and the next async test drains them all at once. #863 added 24
sync subprocess tests, ~1.1s, immediately before this file in
alphabetical order. It did not create the bug — it moved the timing that
exposes it, which is also why the re-run went green.

Root `tests/` is the only Python tree in the repo `ruff check` does not
cover, so none of this was lintable. Left alone here; noted.

### DIFF
--- a/tests/_scoped_module.py
+++ b/tests/_scoped_module.py
@@ -1,0 +1,56 @@
+"""Fake a stdlib function for ONE importer instead of for the process.
+
+``monkeypatch.setattr(some_module.asyncio, "sleep", fake)`` does not patch
+``some_module``'s sleep. ``some_module.asyncio`` IS the one ``asyncio``
+module object, so that line replaces ``asyncio.sleep`` for every importer
+alive — the code under test, every other module, and every task already
+running. The same is true of ``patch("pkg.mod.asyncio.sleep", ...)``: the
+dotted string resolves the same shared object.
+
+That is how caura#863's CI failure happened. ``test_llm_retry_after``
+recorded sleeps into a list and asserted the list was ``[1.0, 2.0]``; it
+got ``[0.7359102269208245, 1.0, 2.0]``, and the test's own captured log
+named the intruder::
+
+    WARNING common.http_retry storage_client.POST /memories/similar-candidates:
+            ConnectError on attempt 3/5, retrying in 0.74s
+
+A post-commit ``detect_contradictions_async`` task, left running by an
+earlier test, retrying a storage-api that does not exist in CI. Its third
+backoff came due while the recorder was installed, so an unrelated task's
+sleep was reported as the retry loop's.
+
+Patching the module REFERENCE instead of the module CONTENTS is the whole
+fix: give the importer under test its own copy, and nobody else's sleep
+can reach the recorder::
+
+    monkeypatch.setattr(mod, "asyncio", scoped(asyncio, sleep=fake))
+    patch.object(mod, "asyncio", scoped(asyncio, sleep=fake))
+
+A real ``ModuleType`` carrying a shallow copy of the original namespace,
+rather than a stub or a forwarding proxy: everything the importer reaches
+for keeps working whether or not this file anticipated it. ``retry.py``
+uses ``asyncio.wait_for`` beside its ``sleep``, and ``memory_service`` uses
+nine asyncio names; an explicit stub would have to list them and would
+break — silently, in the wait_for case — the next time one was added.
+
+The copy is a snapshot. That is fine for stdlib modules, and wrong for a
+module whose attributes a test rebinds later; use ``monkeypatch`` directly
+for those.
+"""
+
+from __future__ import annotations
+
+import types
+
+
+def scoped(real: types.ModuleType, **overrides: object) -> types.ModuleType:
+    """A private copy of *real* with *overrides* applied.
+
+    Assign it over the importing module's own reference, so the override is
+    visible to that importer and to nothing else.
+    """
+    clone = types.ModuleType(real.__name__)
+    clone.__dict__.update(real.__dict__)
+    clone.__dict__.update(overrides)
+    return clone
--- a/tests/test_embed_off_hot_path.py
+++ b/tests/test_embed_off_hot_path.py
@@ -19,6 +19,7 @@ cases for the publish path.
 
 from __future__ import annotations
 
+import asyncio
 import uuid
 from types import SimpleNamespace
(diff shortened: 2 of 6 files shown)


Eldad-Caura:
@erni-a ready for you when you have a moment.

**All checks green.** Full suite run locally at 4826 passed / 0 failed, on a rebase that includes #863 — i.e. with the exact collection adjacency that produced the original failure.

**The tick on `claude-review` means "did not look".** It skipped under pre-checks condition 6, "only tests/config/docs files changed". That is the same gap I raised on #866, and this PR is a second instance of it: the exclusion is path-based, so no change confined to `tests/` can be reviewed automatically, including `tests/conftest.py` and now `tests/_scoped_module.py` — a helper five test files import. I did not re-tag `@claude` here; on #866 that produced "No issues found" alongside a statement that the diff was excluded, which is worse than a skip because it reads as a pass.

**What is worth a human eye, in order:**

1. `tests/_scoped_module.py` — the whole fix is `types.ModuleType` + two `__dict__.update` calls. The property that matters is that it copies the real namespace, not just the overrides; `retry.py` also reaches for `asyncio.wait_for` and `memory_service` for nine asyncio names. I removed the copy in a simulation and 12 existing tests fail, so it is covered — but by tests written for other reasons, which is worth knowing.

2. The two vacuity guards (`..._the_retry_loop_is_still_recorded`, `..._does_still_move_it`) deliberately pass against the old code. They exist for the mis-fix, not the bug: a fixture scoped to the wrong module records nothing, and every `assert slept == [...]` in both files becomes an assertion on a list nothing can fill. I confirmed the mis-scope is loud — point the fixture at `common.http_retry` and the suite starts taking its sleeps for real and hangs.

3. The other three files (`test_embed_off_hot_path`, `test_embedding_retry`, `test_p4_1_llm_fallback`) are prophylactic. Only `test_llm_retry_after` and `test_llm_retry_budget` had assertions a stray sleep could corrupt. I included the rest because leaving them means a foreign backoff loop still gets silently no-op'd, and because half-fixing a twelve-site class is worse than either extreme — but that is a judgement call, and splitting them out is reasonable if you would rather.

**Not fixed here, filed as #869:** the leaked `detect_contradictions_async` task that caused the original symptom. It needs a measurement pass before anyone picks a fix, and what the fixture should do about background tasks is not a flake decision.

erni-a (approved):

PR #870, opened 2026-08-22
chore(clients): 1.0.1 — ship the H-01 recall fix to the registries

erni-a:
The #835 fix is merged but unpublished — PyPI/npm still serve 1.0.0, whose `recall()` returns empty memories for every SDK user (Python and TS). Version bump to 1.0.1 on both clients; the `caura` metapackage and `memclaw-client` shim need no change (`>=1.0.0` resolves forward).

Publish plan after merge: `uv build && uv publish` in clients/python (needs a fresh PyPI token or Trusted Publishing — the Aug 17 token was to be revoked), `npm publish` in clients/typescript (OTP).

Handover follow-up: this closes the gap between §04 'all ten highs merged' and what users actually install.

### DIFF
--- a/clients/python/pyproject.toml
+++ b/clients/python/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "caura-client"
-version = "1.0.0"
+version = "1.0.1"
 description = "Official Python client for Caura (formerly MemClaw; the memclaw_client import and MemClaw class remain permanent aliases) — governed shared memory for AI agent fleets (multi-agent, multi-tenant, MCP-native)."
 readme = "README.md"
 requires-python = ">=3.9"
--- a/clients/python/src/caura_client/__init__.py
+++ b/clients/python/src/caura_client/__init__.py
@@ -34,4 +34,4 @@ __all__ = [
     "DEFAULT_BASE_URL",
 ]
 
-__version__ = "1.0.0"
+__version__ = "1.0.1"
--- a/clients/typescript/package.json
+++ b/clients/typescript/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@caura/memclaw-client",
-  "version": "1.0.0",
+  "version": "1.0.1",
   "description": "Official TypeScript/JavaScript client for Caura (formerly MemClaw) \u2014 governed shared memory for AI agent fleets (multi-agent, multi-tenant, MCP-native).",
   "license": "Apache-2.0",
   "author": "Caura <hello@caura.ai>",


Eldad-Caura:
Verified the premise from the published artifacts rather than from the version numbers, and it is real — and worse than "unpublished".

**PyPI `caura_client-1.0.0-py3-none-any.whl`, `caura_client/models.py:55`:**

```python
memories = [Memory.from_dict(m) for m in (data.get("supporting_memories") or [])]
```

**npm `@caura/memclaw-client@1.0.0`, `dist/index.js:105`:**

```js
const supporting = data?.supporting_memories;
```

Both live artifacts contain the exact line #835 fixed. Uploaded 2026-08-17; the fix landed 2026-08-20. So every SDK user's `recall()` returns an empty memory list today, silently, with `summary` still working.

**The three-file scope is right, and I checked why.** A 1.0.1 of the two leaf packages propagates without republishing the wrappers:

| dist | declares | resolves forward |
|---|---|---|
| PyPI `caura` 1.0.0 | `caura-client>=1.0.0` | yes |
| PyPI `memclaw-client` 0.5.0 | `caura-client>=1.0.0` | yes |
| npm `caura` (in-repo) | `@caura/memclaw-client: ^1.0.0` | yes |

**But the stated blocker is wrong in both halves, and the npm half will fail silently if this is tagged as-is.**

1. **PyPI needs no credential from anyone.** `publish-python-client.yml` uses Trusted Publishing (OIDC, `id-token: write`, environment `pypi`) — there is no token to hand over. What it needs is the one-time PyPI Trusted Publisher entry plus a GitHub environment named `pypi`, then a tag. Worth confirming those exist, since the 2026-08-17 uploads may have been manual.

2. **npm has no `NPM_TOKEN`.** `gh secret list` returns only `ENTERPRISE_DEPLOY_TOKEN` and `MEMCLAW_AGENTS_KEY`. `publish-npm-client.yml` reads `secrets.NPM_TOKEN` for `NODE_AUTH_TOKEN`, so the publish step authenticates as nobody and fails. Also, "npm OTP" does not apply here — CI publishes with an automation token, which is exempt from OTP by design. **This is the actual blocker**, and it is a repo-secret gap, not something in a maintainer's pocket.

3. **The release tags are memclaw-named.** The triggers are `memclaw-client-v*` and `memclaw-client-ts-v*`. Shipping 1.0.1 means creating new `memclaw-*` refs, which is what sunset hard rule 7 forbids. Not a reason to hold the fix — shipping beats purity here — but it should be a deliberate exemption rather than something that happens because someone pushed a tag. Extending the trigger to accept a `caura-client-v*` tag alongside the old one is a two-line change and satisfies rule 3 as well.

Both publish workflows' header comments still say "memclaw-client to PyPI" and instruct creating a PyPI project named `memclaw-client`, while `clients/python/pyproject.toml` now builds `caura-client`. Stale, and misleading to whoever runs this next.

Not blocking the diff, which is correct as written.

Eldad-Caura (approved):

Eldad-Caura:
Merged — and the registries are still serving the broken 1.0.0. Recording that here so this does not read as shipped.

```
repo main 496ba0f8   clients/python  1.0.1    clients/typescript  1.0.1
PyPI      caura-client              1.0.0
npm       @caura/memclaw-client     1.0.0
```

Both publishes are tag-triggered and no tag was pushed, so neither workflow ran. `recall()` still returns an empty list for every SDK user.

**Both publish paths look unproven for the current package names, so I would not assume a tag just works.**

- The last client release tag of any kind is `memclaw-client-v0.4.0` (30 Jul). There has never been a `memclaw-client-ts-v*` tag — **the npm workflow has never run once**.
- The 1.0.0 uploads on 17 Aug were minutes apart (`caura-client` 08:22, `memclaw-client` 08:48) with no matching tag, which is the signature of a manual `twine upload` rather than the workflow.
- The Python workflow last ran when the dist was still named `memclaw-client`. It now builds `caura-client`. **PyPI Trusted Publishers are per-project**, so a publisher registered against `memclaw-client` will not authorise a `caura-client` upload — the OIDC exchange is rejected. Its own header comment still says "Create the PyPI project `memclaw-client`", which suggests that is exactly how it was set up.
- npm has no `NPM_TOKEN` secret at all.

So the likely outcome of pushing the two tags today is that both halves fail, for two different reasons.

**Order that avoids finding this out the slow way:**

1. Check the PyPI Trusted Publisher is registered for project **`caura-client`** (repo `caura-ai/caura`, workflow `publish-python-client.yml`, environment `pypi`), and that the `pypi` GitHub environment exists. Add if missing.
2. Add the `NPM_TOKEN` secret — automation token with publish rights on `@caura`. This also unblocks #873.
3. Then push `memclaw-client-v1.0.1` and `memclaw-client-ts-v1.0.1`, and **verify against the registries afterwards**, not against the workflow's green tick — a `workflow_dispatch` run that publishes nothing can still succeed.
4. Fix the stale header comments in both workflows while you are in there, and consider adding a `caura-client-v*` trigger alongside the old pattern so future releases stop minting `memclaw-*` refs (rule 7).

Related: #873 (bare `caura` on npm unclaimed, same missing credential).

PR #872, opened 2026-08-22
ci(review): stop excluding conftest.py and shared test helpers from review

erni-a:
Handover §06 trap: the reviewer's blanket `tests/` exclusion means `tests/conftest.py` — the highest-fan-in file in the suite — can never get automated review (#866's 'No issues found' actually meant 'did not look'). This carves out `conftest.py` and `tests/_*.py` shared helpers; ordinary test files remain excluded.

### DIFF
--- a/.github/workflows/claude_code_review.yml
+++ b/.github/workflows/claude_code_review.yml
@@ -126,7 +126,11 @@ permissions:
 env:
   EXTRA_PROMPT: "This is a FastAPI + PostgreSQL (pgvector) backend with a TypeScript OpenClaw plugin. Uses requirements.txt (not uv/pyproject.toml). Tests are in tests/ directory."
   REVIEW_PROMPT: |
-    Do NOT review files under tests/ or test_* files. Focus only on source code.
+    Do NOT review files under tests/ or test_* files — EXCEPT conftest.py
+    and tests/_*.py shared fixtures/helpers, which you MUST review like
+    source code: they are the highest-fan-in files in the suite, and the
+    Aug 22 handover recorded a PR where "No issues found" actually meant
+    "diff was excluded". Focus otherwise on source code.
 
     Review this PR for real issues only — not style preferences.
 


Eldad-Caura:
This is the right change but it does not yet close the case it was written for, and this PR is itself the demonstration.

The finding from #866 was that `claude-review` **skipped entirely** — the job never ran, so no prompt was involved. That skip is decided in `pre-checks`, at `claude_code_review.yml:266`:

```
--jq '[.[] | select(.filename | (startswith("tests/") or startswith("test_") or test("^(Dockerfile|helm/|docs/|.*\\.(md|toml|lock|ya?ml)$)")) | not)] | length'
```

It counts files that are *not* tests/docs/config; zero means `reason=no source files to review (only tests/docs/config)` and the whole job is skipped. `tests/conftest.py` and `tests/_scoped_module.py` both match `startswith("tests/")`, so a PR confined to them still counts zero and still skips — with the improved `REVIEW_PROMPT` never reaching a reviewer.

The evidence is in this PR's own checks. All three of the current PRs skipped:

- **#872** (this one) — touches only `.github/workflows/claude_code_review.yml`, a `.yml`, so zero source files. **The PR that fixes the review exclusion was not reviewed, for the reason it exists to fix.**
- **#871** — `tests/test_api_key_prefix_pin.py` only. Skipped.
- **#870** — `pyproject.toml`, `package.json`, `__init__.py`. Skipped.

What this PR *does* fix is real and worth keeping: a PR touching source **and** `conftest.py` previously had the conftest part ignored by the reviewer. That case is now covered.

To close the original case, line 266's filter needs the same carve-out as the prompt — treat `conftest.py` and `tests/_*.py` as source for the purpose of the "is there anything to review" count. Roughly:

```
select(.filename
  | if (test("(^|/)conftest\\.py$") or test("^tests/_[^/]*\\.py$"))
    then false
    else (startswith("tests/") or startswith("test_") or test("^(Dockerfile|helm/|docs/|.*\\.(md|toml|lock|ya?ml)$)"))
    end
  | not)
```

Worth a test of the jq expression against a fixture list of filenames before trusting it — the two carve-out patterns and the existing three branches interact, and this filter is exactly the kind of thing that reads correct and silently over- or under-matches. `tests/_scoped_module.py` (landed in #868, imported by five test files) and `tests/conftest.py` are the two files that should flip from "skip" to "review"; `tests/test_foo.py` and `docs/x.md` should not.

One consequence worth deciding deliberately rather than discovering: with this in place, every conftest-touching PR starts consuming a review run, including trivial ones.

Eldad-Caura (approved):

PR #874, opened 2026-08-22
ci(review): count conftest and shared test helpers as source, not just say so

Eldad-Caura:
ci(review): count conftest and shared test helpers as source, not just say so

#872 told the reviewer to review `conftest.py` and `tests/_*.py`. That
instruction lives in `REVIEW_PROMPT`, which is only read once the review
job runs — and whether it runs is decided earlier, by `pre-checks`
condition 6:

    SRC_FILES=$(gh api ... --jq '[.[] | select(.filename |
      (startswith("tests/") or ...) | not)] | length')
    if [[ "$SRC_FILES" -eq 0 ]]; then skip

Both carved-out patterns start with `tests/`, so a pull request confined
to them still counted zero source files, the job skipped, and the new
prompt was never reached. The half that shipped covers a real case — a PR
touching source AND `conftest.py` no longer has the conftest part ignored
— but the case it was written for stayed open.

This is the other half: those two patterns now count as source.

## Why they earn it

`tests/conftest.py` is the highest-fan-in file in the suite, and
`tests/_scoped_module.py` is imported by five test files. A defect in
either is a defect in everything downstream, which is the argument for
reviewing source in the first place. The exclusion exists so a docs typo
does not spend a review; it was never meant to cover the files most
capable of breaking the suite silently.

Three of this month's PRs demonstrate the gap, and the third is the one
worth stating plainly:

  #866  tests/conftest.py only          -> skipped
  #868  five test files + a new helper  -> skipped
  #872  one workflow .yml               -> skipped

**The pull request removing the exclusion was not reviewed, for the reason
it existed to fix.** A `.yml` is config, so it too counted zero.

Not widened further. Ordinary `test_*.py`, docs, lockfiles, Dockerfile,
helm and other workflow YAML still count zero, which is what keeps a
one-word README change from costing a review run.

## The test runs the workflow's own expression

`tests/test_review_source_filter.py` extracts the `--jq` argument from
`claude_code_review.yml` and executes it against fixture file lists. It
does not restate it.

That is the whole point. A copy pins the copy: the two drift, the test
stays green, and the expression deciding whether any code gets reviewed
goes back to being unverified — which is how this gap survived in the
first place. Extraction also means rewording that step fails loudly here,
with a message saying to fix the regex rather than delete the test.

17 tests. Five fail against the merged state — the four shared-helper
paths and the mixed-count case — and were confirmed to do so by reverting
the workflow hunk and re-running. The other twelve pass either way by
design: they are the do-not-over-widen side, and a carve-out broad enough
to catch the helpers by accident would break them.

`jq` is preinstalled on GitHub runners; the module skips if it is absent
locally rather than failing.

## Consequence to expect

Every conftest-touching pull request now spends a review run, including
trivial ones. That is the intended trade and it is worth naming, since it
is a real cost per PR rather than a one-off.

### DIFF
--- a/.github/workflows/claude_code_review.yml
+++ b/.github/workflows/claude_code_review.yml
@@ -266,8 +266,20 @@ jobs:
           # Paginated for the same reason as the count above, and here the failure is worse than
           # a dead cap: unpaginated, a pull request whose first 30 files are all docs would be
           # skipped as "no source files" even with source changes on page two.
+          #
+          # `conftest.py` and `tests/_*.py` count as SOURCE here, which is the other half of the
+          # carve-out #872 made to REVIEW_PROMPT. Telling the reviewer to review them achieved
+          # nothing on its own: a change confined to those files still counted zero here, so the
+          # job skipped and the prompt was never read. #866 was exactly that, and #872 itself was
+          # too — a lone workflow `.yml` is config, so the PR fixing the exclusion was not
+          # reviewed either. They earn it: they are the highest-fan-in files in the suite
+          # (`tests/_scoped_module.py` is imported by five test files), and a defect in one is a
+          # defect in everything that imports it.
+          #
+          # The expression is pinned by tests/test_review_source_filter.py, which extracts THIS
+          # line from THIS file and runs it — a copy in a test would only prove the copy works.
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@erni-a pleasse approve

erni-a (approved):

Issue #873, opened 2026-08-22
`npm install caura` is in our README, and npm's similarity filter blocks the name against `cuara`


Eldad-Caura:
`npm install caura` is in our README, the name is unclaimed, and we have already lost the equivalent name twice

## The state, as measured 2026-08-23

| name | registry | holder |
|---|---|---|
| `caura` | npm | **nobody — HTTP 404** |
| `caura` | PyPI | us, 1.0.0 (metapackage → `caura-client>=1.0.0`) |
| `memclaw` | npm | `deshrajdry` / "Deshraj Yadav", repo `deshraj/memclaw`, "Memory layer integration for OpenClaw", created 2026-02-06 |
| `memclaw` | PyPI | Emanuilo Jovanovic, repo `emanuilo/memclaw`, v2.0.0 |
| `caura-sdk`, `caura-ai` | PyPI | nobody — HTTP 404 |

Two live instructions point at the empty npm slot:

- `README.md:377` — "`MemClaw` remains a permanent alias of `Caura`, and `npm install caura`…"
- `clients/npm-caura/README.md:9` — `npm install caura`

Both currently fail with 404. That is the worst of the three possible states: not "we haven't shipped it", but "we are actively directing traffic at a name anyone can take", and whoever takes it inherits that traffic plus two README citations pointing at it.

## Why it is unclaimed

`clients/npm-caura/` exists in the repo, is not `private`, is at version 1.0.0, and depends on `@caura/memclaw-client: ^1.0.0`. It has **no publish workflow**. Neither do `clients/caura-meta/` or `clients/memclaw-client-shim/` — three of our five client dists have no automated publish path at all. The two PyPI wrappers exist on the registry only because someone uploaded them by hand on 2026-08-17; the npm one never got that manual step, so it has simply never existed.

## The precedent is our own

`memclaw` on both registries is third-party-held today, by two unrelated parties, and one of them describes itself as a memory layer for OpenClaw — the same category we are in. `pip install memclaw` and `npm install memclaw` install foreign code right now. Those names were lost by not claiming them, and they are not recoverable.

The rebrand plan lists registry names under "unaffected by all scenarios" precisely because they are one-way doors. `caura` on npm is the one such door still open, and the rebrand is about to spend six to eight weeks making the name far more visible.

## Blocked on the same thing as #870

[...]

PR #875, opened 2026-08-22
ci(release): give `caura` on npm a publish path, and stop trusting the tick

Eldad-Caura:
ci(release): give `caura` on npm a publish path, and stop trusting the tick

Two problems, one cause. `clients/npm-caura` has never had a publish
workflow, so the bare `caura` name on npm has never been claimed —
`npm install caura` returns 404 while `README.md:377` and
`clients/npm-caura/README.md:9` both instruct people to run it. And the
TypeScript client's workflow has never run at all: no matching tag has
ever existed, and the 1.0.0 on the registry was uploaded by hand on
2026-08-17.

Neither can publish today regardless: there is no `NPM_TOKEN` secret.
That still needs a maintainer. Everything else is here.

## Why the name matters more than an unshipped package

A documented install command pointing at an unclaimed slot is the worst of
the three states, because whoever claims it inherits the traffic and the
two README citations.

The precedent is ours: the old brand's bare name is third-party-held on
BOTH registries — npm by `deshrajdry`, PyPI by `emanuilo` — and one of the
two describes itself as a memory layer for OpenClaw. Installing either by
that name fetches foreign code today. Both were lost by not claiming them
and neither is recoverable. The rename programme lists registry names
under "unaffected by all scenarios" precisely because they are one-way
doors, and it is about to spend six to eight weeks making this name far
more visible.

## Three guards, each for a failure that already happened

**The credential must exist.** Without the secret, `npm publish`
authenticates as nobody and fails with a 401/ENEEDAUTH that reads like an
npm-side problem. That already cost a debugging round, and the handover it
came through recorded the blocker as "npm OTP" — which does not apply,
since CI publishes with an automation token, exempt from OTP by design.

**The tag must agree with package.json.** A mismatch does not fail: npm
publishes whatever package.json says, under a tag claiming otherwise, and
the release looks done while the registry serves a different version.

**The registry must actually serve it.** This is the one worth having.
#870 merged a version bump to 1.0.1, every check went green, and both
registries carried on serving 1.0.0 — because the publish is tag-triggered
and no tag was pushed. The tick meant "nothing went wrong", not "it
shipped". So the last step asks the registry, with a short retry for
propagation, and fails if the version is not being served.

## Verified locally, because these workflows are unproven

Neither has a successful run to point at, so the parts that could be
exercised without publishing were:

* Built `clients/typescript`, packed both dists, installed them into a
  scratch project, and imported the metapackage. `import { Caura } from
  "caura"` resolves, `Caura` is a class, and `MemClaw === Caura` still
  holds — rule 3's alias survives the re-export, which `export *` would
  silently drop for a default export.
* Confirmed the packed 1.0.1 contains the H-01 fix: the built
  `dist/index.js` reads `data?.memories ?? data?.items`, not the invented
  `supporting_memories` key still live on the registry.
* Exercised the tag/version guard both ways, and both tag prefixes, so the
  legacy pattern still strips correctly.
* `actionlint` clean, which covers shellcheck on the `run:` blocks.

## Names

The old `memclaw-client-ts-v*` tag still triggers — rule 3, and a tag
pattern is exactly what sits in someone's runbook — but new releases use
`caura-client-ts-v*`, because rule 7 is "mint nothing new carrying the old
brand" and every release under the old pattern mints one. The two lines
that must carry the legacy literal are marked `legacy-name-ok:` with their
reason; the rest was reworded, and two of those rewordings improve the
code: the package name is now read from package.json instead of written
out, so the eventual rename lands in one place.

Net effect on the ratchet: **-3 lines**.

## Not in this change

`publish-python-client.yml` needs the same treatment and one correction
that matters: its setup note still says to create a PyPI project named
`memclaw-client`, while `clients/python/pyproject.toml` now builds
`caura-client`. PyPI Trusted Publishers are per-project, so an entry
registered against the old name will not authorise the new one — the
Python half will likely fail for that reason. Separate change, since this
one was scoped to npm.

`clients/caura-meta` and `clients/memclaw-client-shim` also have no
publish workflow; both exist on PyPI only because someone uploaded them by
hand.

Refs #873

### DIFF
--- a/.github/workflows/publish-npm-caura.yml
+++ b/.github/workflows/publish-npm-caura.yml
@@ -1,0 +1,175 @@
+name: Publish caura to npm
+
+# Publishes `clients/npm-caura` — the bare `caura` name on npm, a thin
+# re-export of the scoped TypeScript client.
+#
+# WHY THIS EXISTS AS ITS OWN WORKFLOW. The package had none, so the name was
+# never claimed: `npm install caura` 404s today while `README.md` and
+# `clients/npm-caura/README.md` both instruct people to run it. That is worse
+# than an unshipped package — it is a live instruction pointing at a slot
+# anyone can take, and whoever takes it inherits the traffic and the two
+# README citations.
+#
+# The precedent is ours. The old brand's bare name is third-party-held on BOTH
+# registries — npm by `deshrajdry`, PyPI by `emanuilo` — and one of the two
+# describes itself as a memory layer for OpenClaw. Installing either by that
+# name fetches foreign code today. Both were lost by not claiming them and
+# neither is recoverable; `caura` on npm is the equivalent door still open.
+#
+# ORDERING. This package depends on the scoped client at `^1.0.0`, so that
+# package must exist on the registry at a matching version first — it does
+# (1.0.0, and 1.0.1 once `publish-npm-client.yml` runs). npm does not verify
+# dependencies at publish time, so publishing this against a missing version
+# would succeed and produce a package nobody can install.
+#
+# One-time setup by a maintainer:
+#   1. Add an `NPM_TOKEN` repository secret — an npm automation token with
+#      publish rights. Automation tokens are exempt from 2FA/OTP prompts by
+#      design; an OTP is never entered in CI.
+#   2. Nothing else. The name is unscoped, so no npm org is required.
+#
+# Release: push a tag `caura-npm-v<version>` matching the version in
+# clients/npm-caura/package.json, or run this workflow manually.
+
+on:
+  push:
+    tags:
+      - "caura-npm-v*"
+  workflow_dispatch:
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    permissions:
+      # Both, and `contents: read` is not redundant: declaring a permissions
+      # block sets every scope NOT listed to `none`, so naming only id-token
+      # leaves `contents: none` and `actions/checkout` authenticating with a
+      # token that cannot read the repo.
+      contents: read
+      id-token: write # npm provenance
+    defaults:
+      run:
+        working-directory: clients/npm-caura
+    steps:
+      - uses: actions/checkout@v7
+      - uses: actions/setup-node@v7
+        with:
+          node-version: "20"
+          registry-url: "https://registry.npmjs.org"
+
+      # Checked explicitly because the failure it prevents is a confusing one.
+      # Without the secret, `npm publish` authenticates as nobody and fails
+      # with a 401 or ENEEDAUTH that reads like an npm-side problem rather
+      # than a missing repository secret. This has already cost a debugging
+      # round once.
+      - name: The publish credential must exist
+        env:
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+        run: |
+          if [[ -z "${NPM_TOKEN}" ]]; then
+            echo "::error::NPM_TOKEN repository secret is not set. Add an npm automation token with publish rights; nothing here can publish without it."
+            exit 1
+          fi
+
+      # A tag that disagrees with package.json does not fail — it publishes
+      # whatever package.json says, under a tag claiming otherwise. The
+      # release then looks done and the registry serves a different version.
+      - name: The tag must agree with package.json
+        if: startsWith(github.ref, 'refs/tags/')
+        run: |
+          TAG_VERSION="${GITHUB_REF_NAME#caura-npm-v}"
+          PKG_VERSION=$(node -p "require('./package.json').version")
+          if [[ "$TAG_VERSION" != "$PKG_VERSION" ]]; then
+            echo "::error::tag ${GITHUB_REF_NAME} says ${TAG_VERSION} but package.json says ${PKG_VERSION}"
+            exit 1
+          fi
+          echo "publishing $(node -p "require('./package.json').name")@${PKG_VERSION}"
+
+      # The ORDERING note in the header is the one risk here with no natural
+      # symptom: npm does not resolve dependencies at publish time, so
+      # publishing this against a client version that is not on the registry
+      # SUCCEEDS, and produces a package that fails for the first person who
+      # installs it. Every other risk named in this file has a check; this one
+      # was described and left to the operator, which is the shape of comment
+      # that goes stale and is worth even less than no comment.
+      #
+      # Every dependency is checked, not just the one that exists today, so
+      # adding a second cannot quietly skip it.
+      - name: Every dependency must already be on the registry
+        run: |
+          node -p "Object.entries(require('./package.json').dependencies || {}).map(([n, r]) => n + '@' + r).join('\n')" \
+            > /tmp/deps.txt
+          # Retried on the same schedule as the post-publish check below, and
+          # for the same reason: the expected release order is client then
+          # metapackage, minutes apart, so this lookup runs exactly when the
+          # dependency is newest and least likely to have propagated. A single
+          # attempt would turn npm's own CDN lag into a failed release.
+          while read -r SPEC; do
+            [[ -z "$SPEC" ]] && continue
+            RESOLVED=""
+            for attempt in 1 2 3 4 5; do
+              # `--json`, not the human-readable form. `npm view <spec> <field>`
+              # prints `pkg@ver 'ver'` per match, which parses correctly today
+              # but is presentation output with no stability contract; `--json`
+              # is a string for one match and an array for several, and both
+              # are contractual. Same reason the collaborators lookup below
+              # uses it.
+              #
+              # `|| true` is load-bearing, not defensive noise. Actions runs
+              # steps under `bash -eo pipefail`, so an unresolvable spec makes
+              # `npm view` exit non-zero, pipefail propagates it, and `set -e`
+              # aborts the whole step on attempt 1 — silently, before the
+              # diagnostic below ever prints. That is the exact case this loop
+              # exists for, so without this the retry is decorative.
+              RESOLVED=$(npm view "$SPEC" version --json 2>/dev/null | node -e '
+                let s = "";
+                process.stdin.on("data", d => s += d).on("end", () => {
+                  try {
+                    const j = JSON.parse(s);
+                    const a = Array.isArray(j) ? j : [j];
+                    const last = a[a.length - 1];
+                    // A string, or nothing. npm prints its E404 as VALID JSON
+                    // on stdout when --json is set — an object with an "error"
+                    // key — so JSON.parse succeeding proves nothing about
+                    // whether the package exists. Without this type check that
+                    // error object itself becomes a non-empty RESOLVED, and a
+                    // missing dependency silently PASSES the gate. Verified:
+                    // it did exactly that before the check was added.
+                    console.log(typeof last === "string" ? last : "");
+                  } catch { console.log(""); }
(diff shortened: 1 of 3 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
Applied — the finding was right, and it is the sharpest kind: a comment asserting a property that nothing enforces, sitting in a file where every other named risk did get a check. That asymmetry is exactly what makes it rot.

Added `Every dependency must already be on the registry`, before `Publish`.

Two deliberate differences from the suggested version:

**All dependencies, not just the named one.** Reading `dependencies['@caura/memclaw-client']` specifically would work today and silently skip the second dependency the day one is added — the same shape of gap as the one being fixed. It iterates `Object.entries(dependencies)` instead.

**`npm view "<name>@<range>" version` rather than fetching `versions --json` and doing semver matching in the workflow.** npm already resolves the range; reimplementing that in shell would be a second, worse semver implementation whose disagreements with npm's would only show up at publish time.

Exercised against the live registry, both directions:

```
@caura/memclaw-client@^1.0.0   -> 1.0.0        pass
@caura/memclaw-client@^99.0.0  -> (none)       correctly rejected
caura@^1.0.0                   -> (none)       correctly rejected
```

The third is incidental but worth recording: it re-confirms the bare `caura` name is still unclaimed, which is what this PR exists to fix.

`actionlint` clean (shellcheck over the `run:` blocks). Ratchet unchanged at **-3 net**.

@claude re-review — the review job skips on `synchronize`, so this needs the explicit tag.

Eldad-Caura:
Both Lows applied. The second one is the better catch of the two.

**Retry on the dependency check.** Right, and arguably the wrong way round before: the expected release order is client then metapackage, minutes apart, so that lookup runs exactly when the dependency is newest and *least* likely to have propagated — the one place in the file where lag was most likely, with the only check that had no retry. Same 5×10s schedule as its neighbour now. Confirmed the happy path still resolves on attempt 1, so it costs nothing when the dependency is already there.

**Hardcoded `caura`.** This is drift I introduced in this PR while arguing against drift — the commit message praises deriving the name from `package.json`, and then the new file hardcodes it in two steps. Derived in both now, and `grep` confirms no `caura@` literal remains in any `run:` block.

**On factoring the guards into a composite action — not in this PR, and the reasoning is worth stating rather than just deferring.**

The suggestion is sound, and the evidence for it is strong: the drift you found happened inside a single PR, which is the failure mode duplication is supposed to be judged by.

Against it: the third publish workflow is Python, and it will not share this code — `python -c` and the PyPI JSON API rather than `node -p` and `npm view`, and PyPI's Trusted Publishing has no credential-presence check to make because there is no credential. So the shared helper would have exactly two callers, indefinitely. Parameterising tag prefix and working directory for two callers costs roughly what the duplication costs, and composite actions bring their own edges (`shell:` required per step, `working-directory` not inherited from `defaults`).

The deciding factor is timing rather than design: every SDK user's `recall()` returns an empty list right now, and this is the change that unblocks shipping the fix. I would rather land it and revisit if a third npm package appears. Happy to be overruled — if you would prefer the refactor first, say so and I will do it before merge.

`actionlint` clean. Ratchet unchanged at **-3 net**.

@claude re-review — the job skips on `synchronize`, so this needs the explicit tag.

Eldad-Caura:
Added `verify-npm-credential.yml` — manual-dispatch only, read-only, no `npm publish` in any form.

**Why, concretely.** The registry says the two packages involved were published by two different accounts:

| package | maintainer |
|---|---|
| broker, `@caura-ai` scope | `caura-ai-admin` |
| client, `@caura` scope | `caura` |

The token currently in `NPM_TOKEN` is believed to be the broker's, and the client package was created by a different person. A broker token publishes the broker fine and 403s on the other scope.

**What makes that worth a workflow rather than a warning is that the two halves fail differently.** The unscoped `caura` name is unclaimed, so *any* valid account can take it, while the scoped client needs the right one. So a wrong token does not fail cleanly — it can claim the bare name under one owner while the scoped package stays under another. npm has no simple way to move a package between owners afterwards, which makes that outcome worse than either success or outright failure. Reversing it means asking npm support.

`npm whoami` is fatal here: no answer means invalid or expired, which is itself the finding. Everything after it is advisory, because `npm access` output and flags have moved between npm majors and reporting "could not determine" as "no access" would be worse than saying nothing.

Package names are read from the two manifests rather than written out — same reasoning as the last round, and it means this cannot drift from what the publish workflows actually ship.

One operational note: **it cannot be run until this merges.** `workflow_dispatch` only appears in the Actions UI once the workflow is on the default branch.

Ratchet still **-3 net** — the four old-brand lines this file first introduced were reworded away, and deriving the names from `package.json` removed two of them as a side effect. `actionlint` clean.

@claude re-review — the job skips on `synchronize`.

Eldad-Caura:
Both applied. The High is correct and it was mine — worth spelling out how it got past me, because the mechanism generalises.

**The retry loop didn't retry.** Actions runs steps under `bash --noprofile --norc -eo pipefail`. An unresolvable spec makes `npm view` exit non-zero, `pipefail` propagates it through the pipeline, and `set -e` aborts the step on attempt 1 — silently, before the "not visible yet" diagnostic prints. The sibling step further down already had `|| true`; this one did not. So the loop added specifically to tolerate propagation lag would have died on exactly the lag it was for.

**I reported both paths as verified last round, and that verification was invalid.** I ran them in an interactive shell, which has neither `-e` nor `pipefail`, so I exercised the logic and not the environment. Side by side:

```
$ bash /tmp/dep.sh                                    # my shell — what I tested
attempt 1: not visible yet, waiting
attempt 2: not visible yet, waiting
attempt 3: not visible yet, waiting
exit=0

$ bash --noprofile --norc -eo pipefail /tmp/dep.sh    # what Actions runs
exit=1
```

No output at all. Fixed with `|| true` on the command substitution, and re-verified by executing both paths under the Actions invocation:

```
failure path: retries 3x, then errors            exit=1  ✓
success path: @caura/memclaw-client@^1.0.0 -> 1.0.0, attempt 1  ✓
post-publish loop, unpublished version: retries, then errors   ✓
```

I also parsed every `run:` block across all three workflows under that same shell — 13 blocks, 0 syntax errors — rather than just re-checking the one flagged. If one substitution was unguarded, others could have been.

**The Medium is right and it pre-existed** in `publish-npm-client.yml`: declaring a `permissions:` block sets every unlisted scope to `none`, so naming only `id-token: write` leaves `contents: none` and `actions/checkout` authenticating with a token that cannot read the repo. It has never bitten because the workflow has never run — which is precisely the argument for fixing it before the first real run rather than discovering it mid-release. `contents: read` added to all three jobs, with the reasoning in a comment so nobody "tidies" it away as redundant.

`actionlint` clean. Ratchet **-3 net**.

@claude re-review — the job skips on `synchronize`.

Eldad-Caura:
Correct, and this is the finding I am least happy to have needed. `npm access get status` reports public-vs-restricted **visibility**. The workflow exists to answer "can this token publish these two packages", and it was asking a different question — one that would have returned a clean-looking answer in precisely the case it was written to catch.

That is the same shape as the round-one finding on the sibling file: a step whose comment states a purpose its code does not serve. Three of the four rounds on this PR have been that shape. Worth naming rather than brushing past — it is #847's pattern, and on the evidence it is a blind spot of mine in workflow code specifically, where nothing executes locally and the prose is doing more work than usual.

Replaced with `npm access list collaborators <pkg> --json`, looked up against the account from `npm whoami`. Three properties chosen deliberately:

**An unpublished name is not a failure.** It reports "any valid account may claim it", which is the true answer for the unscoped name and is the thing this PR is trying to change. Treating it as missing permission would make the workflow cry wolf on every run until the name is claimed.

**Uncertainty stays advisory.** npm's access output and flags have moved between majors, so empty or unparseable answers report UNKNOWN. Reporting "could not determine" as "no access" would be worse than saying nothing.

**A definite negative now fails the job.** Previously every path exited green. A diagnostic that finds the blocker and then passes is one nobody reads twice, and the failure message spells out the specific hazard — that a wrong token 403s on the scoped client while still claiming the unclaimed bare name, splitting the packages across two owners with no self-service fix.

Tested the parser against all six shapes npm can produce:

```
{"caura":"read-write","caura-ai-admin":"read-only"}  caura           -> read-write  ✓
{"caura":"read-write","caura-ai-admin":"read-only"}  caura-ai-admin  -> read-only   ✓
{"caura":"read-write"}                               caura-ai-admin  -> NONE        ✓
(empty)                                              caura           -> UNKNOWN     ✓
not json at all                                      caura           -> UNKNOWN     ✓
{}                                                   caura           -> NONE        ✓
```

The two that matter are rows 4 and 5: those are the ones that must not become a false refusal now that a definite negative fails the build.

All 13 `run:` blocks across the three workflows re-parsed under `bash --noprofile --norc -eo pipefail`. `actionlint` clean. Ratchet **-3 net**.

@claude re-review — the job skips on `synchronize`.

Eldad-Caura:
Round 5's finding does not reproduce, but the concern behind it was right and I have fixed that instead.

**Not reproducible as described.** On npm 11.6.2, `npm view <spec> version` for a multi-match range prints one line per match, not an array:

```
@caura/memclaw-client@1.0.0 '1.0.0'
@caura/memclaw-client@1.0.1 '1.0.1'
```

`tail -1 | tr -d "'" | awk '{print $NF}'` yields `1.0.1`, not `]`. Checked against a wider range too (`lodash@^4.0.0`, hundreds of matches) → `4.18.1`. Conveniently this case became real while the PR was open: `^1.0.0` now matches two published versions.

**But the underlying point stands.** That is presentation output with no stability contract, and I was already using `--json` in the collaborators lookup for exactly that reason. Switched for consistency and to stop depending on a format npm is free to change.

**Doing so introduced a worse bug than the one being fixed, which the switch itself surfaced.** `npm view --json` prints its E404 as *valid JSON on stdout*:

```json
{ "error": { "code": "E404", "summary": "No match found for version ^99.0.0" } }
```

So `JSON.parse` succeeds, and my first version returned the error object as `RESOLVED`. Non-empty means found, so **a missing dependency would have silently PASSED the gate** — the exact failure this step exists to prevent, introduced while fixing a cosmetic log line. Caught by re-running the same five-case matrix rather than assuming the change was safe.

Fixed with a type check — a string or nothing:

```
@caura/memclaw-client@^1.0.0        -> '1.0.1'    FOUND
@caura/memclaw-client@1.0.0         -> '1.0.0'    FOUND
@caura/memclaw-client@^99.0.0       -> ''         gate blocks
totally-not-a-real-pkg-xyz@^1.0.0   -> ''         gate blocks
lodash@^4.0.0                       -> '4.18.1'   FOUND
```

One more self-inflicted catch: the explanatory comment I added inside the node script used backticks, which shellcheck reads as command substitution inside a single-quoted string (SC2016). `actionlint` failed; comment reworded without them.

**Context that changes this PR's urgency, not its content:** 1.0.1 shipped to both registries a few hours ago — PyPI `caura-client` 1.0.1 at 06:07 and npm `@caura/memclaw-client` 1.0.1 at 06:19, published manually by the `caura` account. I verified both artifacts contain the H-01 fix. So the user-facing bug is closed. Bare `caura` on npm is **still 404**, which is the part this PR is for.

`actionlint` clean. Ratchet **-3 net**.

@claude re-review — the job skips on `synchronize`.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #876, opened 2026-08-23
docs: stop promising 'npm install caura' — npm blocks the bare name

erni-a:
Registry-vs-README audit (Erni, 23 Aug): README:377 promised `npm install caura` works; the bare-name publish was actually rejected by npm's similarity filter ('too similar to csurf'), so the instruction 404s for every reader. Reworded to the scoped package, with the bare name noted as pending an npm support grant. All other install instructions verified against live registries and match.

### DIFF
--- a/README.md
+++ b/README.md
@@ -376,6 +376,10 @@ console.log((await mc.recall("Q3 revenue target")).summary);
 
 `MemClaw` remains a permanent alias of `Caura`.
 
+The bare `caura` name on npm is held up by the registry's package-name
+similarity filter, so install the scoped package above. On the Python side,
+`pip install caura` works and pulls the same client.
+
 See [`clients/typescript/`](clients/typescript/) for the full client.
 
 ---


Eldad-Caura:
@claude rebased onto main and re-scoped — please re-review; the review job skips on `synchronize`.

**Main solved most of this while the PR sat.** #879 already deleted the false `npm install caura` promise, so the original diff no longer applied. What remains unique here is the *explanation*, which main does not give: why the bare name is unavailable, and what to use instead.

Rewritten so no new line carries the old brand. The first draft appended the explanation onto the existing `` `MemClaw` remains a permanent alias `` line, and the ratchet correctly failed it — `README.md (50 -> 50, count unchanged — the text is new)`. That is #885's new-text detection working, hours after it merged. Rather than mark it exempt, the explanation is now its own paragraph and line 377 is byte-identical to main. Ratchet: `No new lines`. Sentinel: `All 22 protected strings survive`.

**Every install command the README states is now verified live**, which felt obligatory given this PR exists because one of them was false:

| command | |
|---|---|
| `npm install @caura/memclaw-client` | HTTP 200 |
| `pip install caura` | HTTP 200 — metapackage, `requires caura-client>=1.0.0`, author `hello@caura.ai` |
| `pip install caura-client` | HTTP 200 |

The `pip install caura` claim is new in this PR, so I checked it is genuinely ours and not a squatter before adding it.

One thing left deliberately vague: the README says the name is held up by "the registry's package-name similarity filter" without naming the colliding package. Two sources disagree — #876's original audit said `csurf`, the 23 Aug verification pass said `cuara` — and I have verified neither. The user-facing advice is identical either way, so the doc states the effect rather than a cause I cannot stand behind.

Eldad-Caura:
@erni-a please approve

Eldad-Caura (approved):

PR #774, opened 2026-08-13
CAURA-000- chore(rename): update repo references for caura + caura-enterprise

erni-a:
Follow-up to today's renames (**this repo is now caura-ai/caura** — 430 stars carried, redirects live). 36 explicit reference updates across 15 files: badges, clone URLs, pyproject URLs, issue templates, the enterprise-dispatch workflow, publish-workflow notes.

**Untouched by design:** `ghcr.io/caura-ai/caura-memclaw-*` image names (frozen — org-scoped, unaffected by repo renames), CHANGELOG (history), release-please config (release identity).

**Outside this PR:** PyPI trusted publisher for `memclaw-client` must be repointed to repo `caura` before the next client publish (Ran/Erni, PyPI settings).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #877, opened 2026-08-23
docs: four README references that point at things which do not exist

yanki-m:
All four broke on the 2026-08-15 repo rename. None is a legacy name kept on purpose — they are wrong, not old.

| Line | Was | Now |
| --- | --- | --- |
| 61, 132 | `cd caura-memclaw` | `cd caura` |
| 770 | `caura-memclaw-db-1` | `caura-db-1` |
| 1046 | `memclaw/` (tree root) | `caura/` |
| 1267 | `#how-memclaw-compares` | `#how-caura-compares` |

**The quickstart is the one that matters.** Both `cd caura-memclaw` lines follow `git clone https://github.com/caura-ai/caura.git`, so the quickstart dies at its second line for every reader. It passes locally only where a `caura-memclaw -> caura` symlink survived the rename, which is why it has gone unnoticed since 08-15. The 2026-08-19 rebrand audit flagged it as "a functional break".

The others: compose derives its project name from the directory, so a fresh clone gets `caura-db-1` and the `pg_dump` step names a container that will not exist; the heading has read `## How Caura compares` (L415) since the rename, leaving the L1267 anchor dead.

## What this deliberately does not touch

The other 59 occurrences stay. `MEMCLAW_*` envs, the `memclaw` Postgres role and database, the plugin and skill id, the `memclaw-client` / `MemClaw` aliases, the `ghcr.io/caura-ai/caura-memclaw-*` image names (verified against `docker-compose.yml`) and the `/memclaw/` routes are live contracts under hard rule 3 — old names stay readable forever. They come out with the env dual-read and infra waves, not with a prose sweep.

Line 377's `npm install caura` claim is the fifth broken reference and is already fixed by #876; this branch leaves it alone to avoid the conflict.

## Ratchet

`scripts/legacy_name_ratchet.py` reports `No new lines. 4 removed by this change (-4 net).`

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ The fastest way to see Caura work. Standalone mode runs single-tenant with auth 
 
 ```bash
 git clone https://github.com/caura-ai/caura.git
-cd caura-memclaw
+cd caura
 cp .env.example .env && echo "IS_STANDALONE=true" >> .env   # single-tenant, no API key
 docker compose up -d                                        # Postgres + pgvector + Redis + API (~30s)
 
@@ -129,7 +129,7 @@ The fastest path is Docker Compose — one command brings up Postgres + pgvector
 
 ```bash
 git clone https://github.com/caura-ai/caura.git
-cd caura-memclaw
+cd caura
 cp .env.example .env
 ```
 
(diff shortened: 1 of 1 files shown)


erni-a:
Nice catch on the container name, the tree-root label, and the dead anchor — I'd only found the two `cd` lines. The same #774 half-update also exists outside the README (AGENT-INSTALL.md ×2, CONTRIBUTING.md, static/docs/integration-guide.md); #878 now covers exactly those four and leaves the README to this PR, so the two merge cleanly in either order.

erni-a (approved):

PR #883, opened 2026-08-23
ci(review): review a draft PR when it is marked ready

Eldad-Caura:
## What

One word in the trigger list: `ready_for_review` added to `on.pull_request.types`.

## Why

A PR opened as a **draft is never reviewed automatically at all**:

| event | what happens | why |
|---|---|---|
| `opened` (as draft) | skipped | `pre-checks` draft skip, deliberate |
| marked ready → `ready_for_review` | **nothing fires** | not in the trigger list |
| next push → `synchronize` | skipped | `pre-checks` synchronize skip, deliberate |

Each skip is individually correct; together they leave no automatic path to a review. The only route is a manual `@claude` comment, which relies on someone remembering.

## This does not change the drafts-are-opt-in policy

The header documents that policy deliberately — "drafts are unfinished; each review costs real budget". That reasoning is about the *draft state*, and it is preserved exactly: `ready_for_review` fires when the PR **stops** being a draft.

Verified against the gates that read the event payload:

- the synchronize skip tests `github.event.action == 'synchronize'` → no match
- the draft skip tests `github.event.pull_request.draft` → already `false` when the event fires
- `claude-capture` is gated on `github.event.action == 'closed'` → unaffected
- `claude-retrigger` is gated on `issue_comment` → unaffected

No gate changes. The two header comments that described the old behaviour are updated in the same commit so they do not go stale.

## Scope — the gap is org-wide and cannot be fixed centrally

Measured on `origin/main` today. A reusable workflow **cannot declare its own triggers**, so every caller carries its own `on:` block and each needs its own one-line PR — moving the shared pipeline's `v1` tag delivers nothing here:

| repo | `types:` today | status |
|---|---|---|
| `caura` (OSS) | `[opened, reopened, synchronize, closed]` | this PR |
| `caura-onprem-installer` | `[opened, reopened, synchronize]` | [installer#8](https://github.com/caura-ai/caura-onprem-installer/pull/8) |
| `.github` `self-review.yml` + caller template | `[opened, reopened, synchronize, closed]` | separate PR |
| `caura-enterprise` | `[opened, reopened, synchronize, closed]` | still open |
| `caura-ops`, `caura-daemon`, `caura-test-automation` | `[opened, reopened, synchronize, closed]` | still open |

## Testing

`actionlint .github/workflows/claude_code_review.yml` clean. The change is inert until a draft is promoted and cannot affect any currently-working path.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #884, opened 2026-08-23
ci: assert the do-not-touch list survives, not just that it stops growing

Eldad-Caura:
Phase 0.2 of the sunset plan. **This gates the sweeps in Phase 1.3 and Phase 7** — those should not start until something like this exists.

## Why the ratchet is not enough

Hard rule 4 says the do-not-touch list becomes CI. `scripts/legacy_name_ratchet.py` (#863) is half of it, and it is *directional*: it fails a file whose old-brand count goes **up**. A sweep that **deletes** a floor string makes that count go **down**, so the ratchet reports it as progress and passes it.

The two gates answer different questions and neither subsumes the other:

| | question | tripped by |
|---|---|---|
| ratchet | did this change mint a new old-brand name? | a rename wave |
| sentinel | did this change remove a name something depends on? | a prose sweep |

Phase 7 is a prose sweep.

## The sharpest case carries no brand at all

`terraform/datadog-gcp/gcp_alerts.tf:40` in caura-enterprise is a CRITICAL `google_monitoring_alert_policy`, live since 2026-07-27, matching `jsonPayload.message` against a regex alternation of three phrases emitted by `common/embedding/_service.py`:

- `Embedding service degraded` — >=3 consecutive provider failures
- `Embedding concurrency gate timeout` — our own per-process cap is full
- `Embedding backend refused at capacity` — TEI returned 429

The terraform carries its own warning that rewording any of them in OSS silently stops the alert. Nothing on this side enforced it. These are ordinary English with no brand in them, so **no brand-based gate can ever see them** — which is why the list is a list rather than a pattern.

## Why those three are AST-checked and the rest are not

`_service.py` discusses all three phrases in its own comments (`:90`, `:642`, `:735`) as well as emitting them. A plain text search therefore **passes on the exact tree that kills the monitor** — the one where the `logger.error` call is gone and only the commentary about it survives. Verified:

```
$ # reword only the emitted message, leave the three comments intact
$ python3 scripts/do_not_touch_sentinel.py
  common/embedding/_service.py
      'Embedding service degraded' — it survives only in prose — no logging call emits it any more
      breaks: gcp_alerts.tf:40 stops detecting provider failure (>=3 consecutive)
```

A log filter matches emitted messages; a comment emits nothing. The other 19 entries are contracts where the characters are the whole dependency, so a text search is the right check for them.

## What is on the list

22 strings. Three Datadog-matched phrases; the `memclaw-smoke` emitter/matcher pair (rename one and ~200 probe facts/day flood every per-agent report); wire contracts (the `memclaw_` tool-alias shim, the `/api/v1/memclaw` mount, `API_KEY_PREFIX = "mc_"`, the two agent ids, the plugin id and install path, the observability logger name, the org-settings key); published channel names (npm package, interviewer console script and config dir, both release-tag patterns); and the three alembic migrations that carry the brand, which rule 2 says are never edited.

`scripts/do_not_touch_sentinel.py --list` prints all 22 with what each one breaks.

## Proof it bites

Per the house rule that a test passing either way is worth nothing, the parameterised test **removes each string and asserts the gate fails**. A row naming a string its file does not contain, or one that survives its own removal, fails the suite — so a row that protects nothing cannot be added quietly later. All 22 confirmed biting; 35 tests pass.

The gate also fails closed rather than skipping: a listed path that no longer exists is a failure (moving the file is one of the ways the string goes away), an unparseable file raises, and an empty list fails.

## Notes

- The 19 brand-carrying entries each carry `legacy-name-ok`, so the ratchet's exemption report names all 19 on this PR. That is the intended reading: it is the diff where a reviewer can see exactly what is being pinned. Excluding this file from the ratchet instead would leave a hole in that scan.
- No fetch and no base — unlike the ratchet this is not a comparison, so the checkout alone answers it.
- **Out of lane, worth someone's time:** `gcp_alerts.tf:169` is a second log-text monitor of the same shape, matching `memclaw_http_error|memclaw_unavailable` emitted by **caura-ops**. Same trap, different repo, and caura-ops has no gate of any kind. Filed nowhere yet.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -513,3 +513,29 @@ jobs:
         run: |
           git fetch --no-tags --depth=1 origin main
           python3 scripts/legacy_name_ratchet.py --base origin/main
+
+  do-not-touch-sentinel:
+    # The other half of hard rule 4. The ratchet above is directional — it fails a
+    # file whose old-brand count goes UP — so a sweep that DELETES a load-bearing
+    # string reads as progress and passes. This job asserts the same list in the
+    # other direction, and must exist before the Phase 1.3 and Phase 7 sweeps.
+    #
+    # Not every string it guards carries the brand: three are ordinary English
+    # that a CRITICAL production Datadog monitor matches on as a regex substring.
+    # Reword one and the alert stops firing with nothing in the tree to fail.
+    #
+    # No fetch and no base. Unlike the ratchet this is not a comparison — the
+    # question is whether the strings are in the tree being built, so the
+    # checkout alone answers it.
+    name: Do-not-touch sentinel
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+
+      - name: Load-bearing strings survive
+        run: python3 scripts/do_not_touch_sentinel.py
--- a/scripts/do_not_touch_sentinel.py
+++ b/scripts/do_not_touch_sentinel.py
@@ -1,0 +1,390 @@
+#!/usr/bin/env python3
+"""Fail a PR that removes a string something outside this repo still depends on.
+
+Hard rule 4 of the sunset plan: the do-not-touch list becomes CI. Half of it
+shipped as ``scripts/legacy_name_ratchet.py``, which is *directional* — it fails a
+file whose old-brand count goes **up**. A sweep that **deletes** a load-bearing
+string makes that count go **down**, so the ratchet reads it as progress and
+passes it. This is the other half: the same list, asserted in the other
+direction.
+
+The two gates are not redundant and neither subsumes the other. The ratchet
+answers "did this change mint a new old-brand name?"; this one answers "did this
+change remove a name something already depends on?" A rename wave trips the
+first, a prose sweep trips the second, and Phase 7 is a prose sweep.
+
+**The strings here are not all old-brand strings, and that is the point.** The
+three embedding phrases carry no brand at all — they are ordinary English that a
+production Datadog monitor matches on as a regex substring
+(``terraform/datadog-gcp/gcp_alerts.tf:40`` in caura-enterprise, a CRITICAL
+policy that has run since 2026-07-27). Reword one in a tidy-up — or merely
+downgrade the level it is logged at — and the alert stops firing with nothing to
+fail: no test breaks, no count moves, and the only symptom is an alert that never
+arrives. No brand-based gate can ever see that, which is why the list is a list
+rather than a pattern.
+
+Usage::
+
+    scripts/do_not_touch_sentinel.py            # fail on any missing string
+    scripts/do_not_touch_sentinel.py --list     # print the list and exit 0
+"""
+
+from __future__ import annotations
+
+import argparse
+import ast
+import sys
+from dataclasses import dataclass
+from pathlib import Path
+
+REPO_ROOT = Path(__file__).resolve().parents[1]
+
+# A plain substring of the file's text. Right for a name that is a contract:
+# whoever depends on it depends on the characters, not on where they sit.
+LITERAL = "literal"
+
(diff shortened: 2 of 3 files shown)


Eldad-Caura:
All three findings were real and are fixed. Verified each against the actual files rather than synthetically, because two of them turn on what is genuinely in the tree.

## 1. Severity — confirmed, and it was the sharper half of the finding

Correct, and worth restating precisely because it is counterintuitive: **the Datadog filter does not select on severity at all**, so it is tempting to conclude the level does not matter. It matters for a different reason — the level decides whether the line is emitted to Cloud Logging in the first place. Downgrade it and the phrase is intact, the filter would match it, and the line never arrives to be matched.

`Sentinel` now carries `min_level`, and `_log_calls` returns `(level, message)`. Pinned at what is actually there: `error` for `Embedding service degraded`, `warning` for the other two. A **minimum**, not a set — raising a level is harmless.

Verified on the real file, downgrading only the call and touching nothing else:

```
  common/embedding/_service.py
      'Embedding service degraded' — it is emitted at debug, below error —
      the phrase is intact but the line no longer reaches the sink the monitor reads
```

One thing the fix surfaced that the finding did not: `logger.log(level, msg)` names its level in an argument, and the one call of that shape in this repo (`observability.py:97`) picks it with a conditional. There is no static answer, so it counts as an emitter of the phrase but can never satisfy a level floor — unverifiable is reported, not assumed, and the message asks for an explicit `.error()` instead.

## 2. Comment-satisfiable literals — confirmed, and it was live on two entries

This is the same hole I closed for `LOG_MESSAGE`, in the other kind. Measured, not reasoned: with the alias shim **fully deleted** from `mcp_server.py`, bare `memclaw_` still survives at lines 516, 556 and 1408, so the old entry passed a tree with no shim in it. Same shape in `app.py`, where the comment directly above the mount describes the route.

Fixed by pinning functional syntax rather than a bare name, which makes the entry unsatisfiable by prose without needing a runtime mechanism to enforce it:

| entry | was | now |
|---|---|---|
| `mcp_server.py` | `memclaw_` | `name.startswith("memclaw_")` |
| `app.py` | `/api/v1/memclaw` | `prefix="/api/v1/memclaw"` |
| `report_corpus.py` | `memclaw-smoke` | `cache refresh\|memclaw-smoke)` |
| `installer.py` | `memclaw-interviewer` | `".config" / "memclaw-interviewer"` |
| `030_…insighter.py` | `memclaw-insighter` | `WHERE m.agent_id = 'memclaw-insighter'` |
| `012_vector_dim…py` | `MEMCLAW_RUN_DESTRUCTIVE_MIGRATIONS` | `os.environ.get("MEMCLAW_RUN_DESTRUCTIVE_MIGRATIONS"` |

I went with a variant of your option (a) rather than a tokenize-based "outside a comment" check, for two reasons: several entries are in `.ts`, `.json`, `.yml` and `.toml`, where there is no Python tokenizer to reuse; and tokenize cannot separate a docstring from a value string, which is where two of these actually lived. Tightening the literal fixes it at the source and documents which line is the functional one.

**Held in place by a test, not by care.** `test_no_literal_can_be_satisfied_by_a_comment_alone` is parameterised over every `LITERAL` entry and fails if the pinned text appears on any comment-only line — so the next entry someone adds cannot reintroduce this.

The one deliberate exception is `019_tenant_suppression.py`, where the string is only ever in the migration's docstring; there is no functional line to pin because the live topic constant is in `common/events/topics.py`. That entry pins rule-2 immutability rather than a contract, and its `breaks` says so.

## 3. f-strings — fixed

`_static_text` now joins the literal segments of a `JoinedStr`. Agreed on the reasoning: a false failure is the safe direction, but one that reports "the phrase is gone" about a phrase sitting right there would send someone hunting for a deletion that never happened. A phrase broken up *by* an interpolation still does not match, which is correct — a substring filter would not find it either, so the monitor really is broken.

## Tests

60 passing, up from 35. New: the downgrade case, level-raising, `exception`-as-error, the `logger.log` limitation, f-strings, interpolation-splitting, and the comment-satisfiable invariant across all 19 literal entries.

Rebased onto `c47aa6cc` (#883 landed in the meantime).

@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #885, opened 2026-08-23
fix(ratchet): a line moved between two files is not a new name

### LINKED ISSUE #882: legacy-name ratchet: a cross-file move false-positives, and its remedy text invites a false exemption reason
Found by the automated review on [caura-enterprise#1212](https://github.com/caura-ai/caura-enterprise/pull/1212) and **verified here** against `scripts/legacy_name_ratchet.py`. Filed upstream because the script now exists in three repos — this one (canonical), `caura-onprem-installer`, and `caura-enterprise` — and the ports differ from this copy only by formatting. Fixing one copy would turn that into a behavioural divergence, so the fix belongs here and should propagate.

## 1. A cross-file move is reported as a violation (the substantive one)

`main()` computes `grown` from per-file increases:

```python
grown = {p: (base.get(p, 0), n) for p, n in head.items() if n > base.get(p, 0)}
```

Move a branded line from file A to file B and B goes `0 -> 1` while A drops. The repo total is unchanged and nothing new was minted, but the gate fails.

Reproduced in a scratch repo against the current script:

```
### move the brand line from a.py to b.py (repo total unchanged)
This change adds the legacy name in 1 file(s).

  b.py  (0 -> 1)
      2: A = "https://memclaw.net"
exit=1
```

The per-file design is still right — a repo total would hide a real addition behind an unrelated deletion, which is the worse failure. The module docstring already defends per-file counting, but it only claims to keep a **within-file** move at net zero, and that claim is accurate. Cross-file is simply not covered.

**Why it is worth fixing rather than accepting.** The failure text tells the contributor:

> If the addition is deliberate — a compat alias, a redirect, a test pinning the old wire format, all of which rule 3 requires — append `legacy-name-ok: <reason>`

A move is none of those three. So the only way past the gate is to write a reason that is not true, into the one annotation whose entire value is that its reasons are true. The marker's integrity is what makes the "report, don't adjudicate" design defensible, so this quietly erodes the thing the gate depends on. That matters more than the spurious red.

**On the suggested fix.** The review proposed detecting renames with `git diff --find-renames`. I tested that case separately and a whole-file rename does also false-positive, so rename detection would help it — but it would **not** help the case above, where a line moves between two files that both already exist. That is the likelier shape during consolidation waves. Options, roughly in increasing cost:

- **(a)** Add "a line moved from another file" to the accepted reasons in both the docstring and the failure text. Cheapest, keeps the marker honest, still costs a marker per move.
- **(b)** Rename detection via `--find-renames`. Covers whole-file moves only; adds a git call and some state.
- **(c)** Fall back to a repo-total check *only for lines that are byte-identical to one deleted elsewhere in the same change*. Covers both shapes and stays strict about genuinely new text, but it is the most code.

I would start with (a) and treat (c) as the real fix if moves turn out to be common once the infrastructure waves start.

## 2. An unresolvable `--base` prints a traceback (cosmetic)

`_git()` raises a bare `RuntimeError` for git exits other than 0/1, and `main()` never catches it:

```
RuntimeError: git grep ... no-such-ref -- :/ failed (128): fatal: unable to resolve revision: no-such-ref
```

Loud and actionable — it carries the git message and the bad ref, and `test_an_unresolvable_base_fails_loudly` already pins that behaviour — but it reads as a crash rather than a validation error. Catching it in `main()` and printing to stderr before a non-zero exit would be a few lines. Low priority; noting it so it is not rediscovered.

### Not a bug

Datadog's static analysis flagged `LEGACY = "mem" + "claw"` in `tests/test_legacy_name_ratchet.py:27` as string concatenation that should be an f-string. That concatenation is deliberate: it keeps the literal out of a file the gate itself scans. An f-string would reintroduce it and force the test file to carry an exemption marker. False positive — worth a comment on the line if the same scanner ever runs here.

Eldad-Caura:
Closes #882.

## 1. The cross-file move

The per-file count fails a move: the destination goes `0 -> 1` while the source drops, so a change that minted nothing is reported as an addition.

That mattered more than the spurious red because of the remedy the failure text offered. None of the three reasons rule 3 recognises — compat alias, redirect, pinned wire format — describes a move, so **the only way past the gate was to write an exemption reason that was not true**, into the one annotation whose whole value is that its reasons are true. The "report, don't adjudicate" design rests on the marker's integrity, so this eroded the thing the gate depends on.

So a move is now recognised rather than annotated away. A line is a **move** when the repo-wide count of that exact text did not go up: the name existed, in those words, and now sits somewhere else. It is **minted** when the count did go up, which no rearrangement can produce.

Per-file counting is unchanged and still primary — a repo total would hide a real addition behind an unrelated deletion, which is the worse failure and the reason the design is what it is. The move check only ever *excuses* a file the per-file count already flagged; it can never excuse text the repo did not have.

### Why (c) and not the (a) the issue suggested starting with

The issue's own preference was to start with (a) — add "a line moved from another file" to the accepted exemption reasons — and treat (c) as the real fix if moves turned out to be common. **I went to (c), and that is a judgement call worth flagging rather than burying.** Two reasons:

- The sweeps that make moves common are the next thing to run. Phase 7.1 alone is −1,318 lines and Phase 1.2 is archival that moves files wholesale. "If moves turn out to be common" resolves to yes on the current schedule, not eventually.
- (a) pays for every move with a marker on the annotation whose integrity is the issue's actual concern. Adding a fourth, much more common reason makes `legacy-name-ok` mean "this might just be a move", which is most of the way back to the problem.

(a) remains a one-line fallback if that call is wrong — say so and I will swap it.

### Why text comparison and not `--find-renames`

The issue already establishes that rename detection sees a whole-file rename but **not** a line moving between two files that both already exist, which is the likelier shape during a consolidation wave. Text comparison covers both, and it has no similarity threshold to fall off.

That last part is not hypothetical. Agent 3 is currently blocked on this in caura-enterprise: `40-memclaw-envsubst.sh` → `40-caura-envsubst.sh` by `git mv`, with variable renames inside the same file in the same commit. A move-plus-edit is exactly where a similarity threshold is in play. This fix does not care — the three branded lines are byte-identical, so they pair with their own deletions whatever happened to their neighbours. Pinned as `test_a_renamed_file_whose_other_lines_changed_is_still_a_move`.

### The report got more precise too

A move sitting beside a real addition still fails, and now names **only** the added line. Before, the report re-grepped the file and printed every non-exempt line in it, sending the reader to "fix" a line that was already in the tree.

```
  b.py  (0 -> 2)
      3: SNUCK = "memclaw-new"      <- named
                                    <- the moved line is not
```

## 2. The traceback (issue's part 2)

An unresolvable `--base` now exits **2** with the git message on stderr instead of a traceback. A broken invocation is not a rule violation, and the two want different reactions from whoever is reading CI. Exit 1 continues to mean the gate ran and found something.

`test_an_unresolvable_base_fails_loudly` was passing either way before — a traceback also puts the ref on stderr — so it is strengthened to pin the exit code and the absence of a traceback.

## Tests

Seven new or strengthened cases; **all seven confirmed failing against the unfixed script** and passing with it:

```
FAILED test_a_move_between_two_existing_files_is_net_zero          <- the #882 repro
FAILED test_a_whole_file_rename_is_net_zero
FAILED test_a_renamed_file_whose_other_lines_changed_is_still_a_move
FAILED test_a_reindented_move_is_still_a_move
FAILED test_a_move_does_not_launder_an_addition_beside_it
FAILED test_an_unresolvable_base_fails_loudly
```

Plus `test_a_second_copy_of_an_existing_line_is_an_addition` and `test_new_text_still_fails_even_though_something_was_deleted`, which pin that strictness did not move. All 28 pass.

## Propagation

Fixed in the OSS copy because it is canonical. The `caura-onprem-installer` and `caura-enterprise` ports differ by formatting alone, so this should propagate to both once it lands — **the enterprise one is the copy currently red for Agent 3**, so it is not optional. I have not touched either port; flag if you want me to carry them or if that belongs to whoever owns those repos this week.

## Not a bug (from the issue)

Datadog flagged `LEGACY = "mem" + "claw"` in the test file as string concatenation that should be an f-string. It is deliberate — it keeps the literal out of a file the gate scans. Left as-is.


Eldad-Caura:
Both findings were real. Fixed, and the second one changed the shape of the PR rather than just patching it.

## 1. Over-reporting — fixed

Straight bug, and it contradicted the comment sitting directly above it. `text.strip() in minted` tests key membership on a `Counter` and throws the count away, so a file holding three identical lines of which one was minted printed all three.

The report now spends a budget instead:

```python
budget = Counter(minted)
for _, lineno, text, exempt in _grep(None, pathspec=_literal(path)):
    stripped = text.strip()
    if not exempt and budget[stripped] > 0:
        budget[stripped] -= 1
        print(f"      {lineno}: {stripped[:110]}")
```

Pinned by `test_only_the_minted_copies_of_a_repeated_line_are_named`: base holds two copies in two files, both are emptied, three appear in one new file. Two moved, one minted, exactly one line named.

## 2. Coincidental pairing — real, and unfixable by counting

Agreed, and I want to be precise about *why* I took option 1 rather than 2, because it is not the cheaper choice — it is the only honest one.

Pairing by text cannot know the deletion it paired with is the same line. Your example is right: delete a dead `URL = "https://memclaw.net"` in one place, add a genuinely new one in a new service, and the repo total is flat. That is a laundered mint.

But it is not fixable by any count, and **this codebase already has a worked example of that exact impasse**. `_report_new_exemptions` faces the identical structure — marking a line exempt frees a slot a new line can fill, and both cases are identical in every number because the difference is *which* line carries the marker, not how many do. Its author's answer was to report, not adjudicate, and to print on the passing path so the one move that buys headroom can never be made quietly.

The move check has the same shape of limit, so it now gets the same answer. Every excused move is printed, on the passing path, **with the file whose deletion paid for it**:

```
2 line(s) treated as moved rather than added — the same text left
another file in this change. Check the pairing is real and not a coincidence:
  b.py: A = "https://memclaw.net"
      was in: a.py
  newsvc.py: OLD = "memclaw-legacy-thing"
      was in: dead.py

No new lines. 2 removed by this change (+0 net).
```

That second row is your scenario, and the attribution is what makes it catchable: a reviewer seeing `newsvc.py` paired against `dead.py` has found the laundering. Nothing else in CI would have shown it. Your suggestion 2's parenthetical — "at least logging a warning listing which base occurrence each move was attributed to" — is exactly this, and it is what I built; the `--find-renames` half of that suggestion still cannot work, for the reason the issue already established.

Documented in both places a reader looks: the module docstring's "What this does not enforce" section now carries it alongside the exemption swap, and `_minted`'s docstring spells out the boundary.

Pinned by `test_an_excused_move_is_always_named`, which asserts the passing exit code *and* the attribution line.

## On the trade this leaves

Worth stating plainly since it is a deliberate loosening. Before this PR the gate failed every cross-file move, including real ones, and the only way past was an exemption whose stated reason would not be true. After it, the gate passes moves and prints them for review, and a coincidental pair passes with a visible attribution. I think that is the better trade — a gate people can only satisfy by lying is worse than one that is honest about a narrow blind spot — but it is a trade, and if you would rather have the stricter version back, option (a) from the issue is still a one-liner.

## Tests

30 passing, up from 28. Both new tests confirmed failing against the un-fixed reporting and passing with it.

Rebased onto current `main`.

@claude

Eldad-Caura:
Confirmed as a real bug, reproduced before fixing, and fixed. Good catch — it is the case my own test matrix stepped around.

## The bug

Reproduced exactly as described. Base holds one copy in `a.py`; the change deletes it and adds one copy each to `b.py` and `c.py`. True repo-wide growth is **1**:

```
This change adds the legacy name in 2 file(s).

  b.py  (0 -> 1)
      1: URL = "https://memclaw.net"
  c.py  (0 -> 1)
      1: URL = "https://memclaw.net"
exit=1
```

`_minted` was called once per growing file and recomputed `added_repo_wide` from the untouched global counters each time, so both destinations claimed the same budget of 1. Two failures reported for one addition, and one of them a legitimate move — which is precisely the false gate failure this PR exists to remove, reappearing in a variant I had not enumerated.

Worth naming why it slipped through: every move test I wrote has a **single** destination. The multi-destination case is not exotic — it is what a consolidation wave produces when one branded line is pulled into two new modules — so the gap was in the test matrix, not just the code.

## The fix

The charge is a repo-wide quantity, so it is now computed once for the run and spent down as files are visited:

```python
def _mint_budget(head_total, base_total) -> Counter[str]:
    """How many of each text the repo actually gained — the whole charge available."""
    return +Counter({text: n - base_total.get(text, 0) for text, n in head_total.items()})
```

`_minted` takes that shared `budget` and decrements it by what it charges, so a second destination cannot be billed for an addition already accounted for. Paths are iterated in sorted order: when one text lands in several files and the repo gained fewer copies than that, *which* destination is charged is arbitrary, but the count is not, and an arbitrary-but-unstable answer would flake.

Now:

```
This change adds the legacy name in 1 file(s).
  ...
1 line(s) treated as moved rather than added — the same text left
another file in this change. ...
```

One mint, one move, and the move is named with its attribution rather than silently dropped.

## Tests

`test_one_deletion_pays_for_only_one_of_two_destinations` — asserts exactly one file is charged, and that the other is reported as a move. Confirmed failing against the un-fixed `_minted` and passing with it. 31 passing.

**One test I wrote and then deleted, since the discipline cuts both ways.** I added `test_which_destination_is_charged_does_not_flake` to pin the `sorted()` ordering, then checked it against a build with `sorted()` removed — it still passed, because `git grep` already returns paths in a stable order, so two consecutive runs agree regardless. It could not fail, so it was worth nothing and I removed it rather than shipping it as coverage. The `sorted()` stays as defensive determinism; it is simply not something a test can currently distinguish.

Rebased onto `main` (now including the merged sentinel from #884). Both gates green on this branch: `All 22 protected strings survive.` / `No new lines.`

@claude

Eldad-Caura:
Fixed. This one is worth calling out because it broke the feature added to answer the *previous* review round — the attribution was the whole remedy for the coincidental-pairing limitation, and it was pointing at the wrong files.

## `came_from` named bystanders, not sources

Confirmed. It listed every base file that *contained* the text rather than every file whose count of it went **down**. A file holding an unchanged copy throughout contributed nothing to the move and was named as if it had. Worse with the `[:3]` truncation and the sort, exactly as described: an alphabetically earlier bystander pushes the real source behind `(+N more)`.

That inverts the point of the line. It exists so a reviewer seeing an unrelated file has found a laundered mint — and a bystander *is* an unrelated file, so the report manufactured false positives for the one check a human is asked to perform. Noise there is worse than printing nothing, because it trains people to skip it.

Now restricted to files whose count actually dropped:

```python
came_from = sorted(
    p
    for p, c in base_by_file.items()
    if p != path and c[text] > head_by_file.get(p, Counter())[text]
)
```

`head_by_file` threaded through for it. Pinned by `test_a_file_that_kept_its_copy_is_not_named_as_the_source`, which deliberately names the bystander `aaa-bystander.py` so it sorts ahead of the real source and would win the truncation — confirmed failing against the containment version, passing now. 32 tests.

## On the triple-scan suggestion — declining, with a reason

Correct that `counts`, `exempt_counts` and `_texts` each run a full-repo `git grep`, and that `_texts` could subsume the others.

I am not taking it, and not because it is wrong. The scans are milliseconds here — 2,454 matching lines across 281 files, and the job's own comment budgets it as "seconds of work; the only thing that can hang is the fetch." Against that, collapsing three helpers into one restructures every call site in a gate whose failure mode is *passing silently* — the module docstring already devotes a paragraph to how a broken scan is indistinguishable from a clean tree. Trading a real chance of introducing that against no measurable gain is the wrong side of the bet, especially in the same PR as a correctness fix.

Happy to do it as a standalone change if you want it, where it can be reviewed as a refactor on its own terms rather than riding along with behaviour changes.

## State

Three review rounds on this PR, five findings, all real, all fixed:

1. Over-reporting when a repeated line was partly moved — budget-consuming display loop.
2. Coincidental pairing can launder a mint — reported with attribution, documented as the limit it is.
3. Multi-destination double-charging — shared mint budget spent across files.
4. `came_from` naming bystanders — this one.
5. (Round 1) The unresolvable-`--base` traceback — exit 2 with the git message.

Both gates green on the branch: `No new lines.` / `All 22 protected strings survive.`

@claude

Eldad-Caura:
Confirmed, fixed — and fixing it closed a second hole the script's own docstring called unclosable. This is the best finding of the four rounds.

## The gap is real, and it predates this PR

Reproduced, and then checked against `origin/main` to see whether I had introduced it:

```
### WITH my move detection
No new lines.          exit=0
### origin/main (before this PR)
No new lines.          exit=0
```

Both pass. The `n > before` filter is the original per-file comparison, so a same-file swap — drop one branded line, add a different brand-new one, total flat — has never been visible to this gate.

**That does not make it not my problem.** My docstring asserts "Strictness is unchanged for genuinely new text: a new name has no matching deletion to pair with, so its repo-wide count rises and it fails," and you are right that the claim is only true when the file's *total* also rises. A comment asserting a property the code does not enforce is the exact recurring failure this programme keeps hitting. And the text machinery the move check needed is precisely what closes it — it was being short-circuited before it ran.

So the count pre-filter is gone. `_minted` now runs for every path in head, and a file enters `grown` when it has minted text, not when its tally rose. The failure line says so when the count did not move:

```
  f.py  (2 -> 2, count unchanged — the text is new)
      1: URL_NEW = "memclaw-brand-new-service"
```

## The part I did not expect: this closes the exempt-and-add swap

Removing the filter turned `test_the_exempt_and_add_swap_is_reported_even_though_it_passes` red — because the swap is now **caught** rather than merely reported.

That test encoded the original author's conclusion, and the reasoning was sound for what it addressed:

> No count separates that from the legitimate case… Both look identical per file — non-exempt flat, exempt up one, total up one — because the difference is *which* line carries the marker, not how many do.

True of every **count**. Not true of the **text**. A legitimate marked alias adds new *exempt* text and no new non-exempt text, so there is nothing to charge. The swap adds non-exempt text the repo never had, and that is chargeable. Verified all three directions rather than assuming:

| case | before | now |
|---|---|---|
| exempt-and-add swap | passed silently | **caught** |
| legitimately adding one marked alias | passed | passes |
| marking an existing line exempt, nothing added | passed | passes |

All three pinned as tests. The exemption report stays — it is what made the hole visible, and a deliberate alias landing beside unrelated work should still never be silent.

Docstrings corrected in both places that claimed the hole was open, and the module docstring now separates what the count cannot do (both directions, now fixed) from what genuinely remains: the coincidental identical deletion, which is identical in every number *and* every text, so nothing here can decide it and it stays reported-with-attribution.

## Tests

35, up from 32. Four new or rewritten, all confirmed failing with the pre-filter restored and passing without it:

```
FAILED test_the_exempt_and_add_swap_is_caught
FAILED test_a_same_file_swap_for_a_brand_new_name_is_caught
```
plus `test_adding_one_marked_alias_still_passes` and `test_marking_an_existing_line_exempt_still_passes`, which pin that the fix did not overreach — those two are the ones that would break if the charge were computed wrongly.

## Scope note

This PR is now doing more than #882 asked. It went out to fix a false *positive* (a move failing) and has ended up closing two false *negatives* (a same-file swap and the exempt-and-add swap passing). All three come from the same change — comparing text instead of counts — so splitting them would mean shipping the machinery and deliberately not using it. Flagging it rather than letting it pass as scope creep: say the word if you would rather the false-negative half landed separately.

Both gates green: `No new lines.` / `All 22 protected strings survive.`

@claude

Eldad-Caura:
Fixed. My own comment said "flat **or falling**" and the code branched only on rising — so the note contradicted the line directly above it, which is the same shape of defect this programme keeps producing: prose asserting a property the code does not implement.

Three branches now:

```python
if after > before:
    note = ""
elif after == before:
    note = ", count unchanged — the text is new"
else:
    note = ", count fell — the text is new"
```

Reachable, not theoretical — two branded lines removed and one brand-new one added gives `2 -> 1` with a genuine mint:

```
  existing.py  (2 -> 1, count fell — the text is new)
      1: C = "memclaw-brand-new"
```

Pinned by `test_a_falling_count_is_not_described_as_unchanged`, confirmed failing against the two-branch version. 36 tests.

Worth taking seriously despite the Low severity, for the reason the finding gives: this note exists to stop a reader arguing with the arithmetic when a flat count fails. A reader who can see the note is wrong about their own diff stops trusting the whole report — and the report is the only thing standing between the excused-move path and a silent laundered mint.

**Holding this one before propagating.** Eldad has asked me to port this fix to the `caura-enterprise` and `caura-onprem-installer` copies, so a defect landing here would have gone to three repos rather than one. Both ports will carry this version.

Both gates green: `No new lines.` / `All 22 protected strings survive.`

@claude

Eldad-Caura:
Both taken this round, including the scan consolidation I declined last time. The reason for the reversal is worth stating.

## Inconsistent error handling — fixed

Correct, and it is a mis-signal rather than an inconsistency. `head = counts(None)` reaches git the same way the base scan does and can fail the same way, but unguarded it surfaced as a traceback with **exit 1** — and in this script exit 1 means *"ran, and found new names."* A gate that could not run was reporting itself as a gate that failed you.

Both scans are under one guard now, so any git failure exits 2 with the message. Verified against a directory that is not a git repo:

```
exit: 2
stderr: git grep ... failed (128): fatal: not a git repository ...
traceback? False
```

Pinned by `test_a_failing_working_tree_scan_also_exits_two`, confirmed failing with the guard removed from the head scan only.

## Redundant scans — reversing my earlier decline

I declined this last round on the grounds that the scans are milliseconds and restructuring a gate for no measurable gain is the wrong side of the bet. **Two things changed that, and neither is performance.**

First, Eldad has asked me to port this to `caura-enterprise` and `caura-onprem-installer`. Shipping a known duplication into three repos, having been told about it twice, is a different decision from leaving it in one.

Second — and this is the part I got wrong the first time — it was never really about the subprocess count. `counts()`, `exempt_counts()` and `_texts()` each encoded *the same rule* about what a countable line is, in three places that had to agree. That is a correctness surface, not a performance one, and it is the kind that stays silent when it drifts.

So there is one `scan()` per tree returning a `Scan` namedtuple, and the counts derive from it:

```python
def counts(self) -> Counter[str]:
    return Counter({path: sum(c.values()) for path, c in self.by_file.items()})
```

Six full-repo greps per run become **two** — but the point is one definition of "a line that counts" instead of three.

**Verified equivalent, not assumed.** `--report` output against the real 2,454-line tree is byte-identical before and after:

```
$ diff /tmp/before_report.txt /tmp/after_report.txt && echo IDENTICAL
IDENTICAL
```

37 tests pass.

## State

Five review rounds, nine findings, all real, all addressed. The count pre-filter removal in round four also closed two false negatives the script's own docstring had called unfixable.

Both gates green: `No new lines.` / `All 22 protected strings survive.`

This is the version that gets ported to the enterprise and installer copies — as a **semantic** port preserving AST-identity, since those two already differ from OSS by an 88-column reformat and a text diff would misread that as drift.

@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #886, opened 2026-08-23
feat(env): read CAURA_* everywhere the old names are read

Eldad-Caura:
Phase 5.1 of the rebrand (work order item 5.1): every environment **reader**
accepts the new `CAURA_*` name and the old `MEMCLAW_*` one. Purely additive —
old spellings keep working forever, per rule 3.

**Writers are deliberately not in this PR.** The generated installers still
emit the old names; that is item 5.2, which the work order says must land only
after this is deployed.

### The gating defect

`plugin/src/env.ts` filtered the plugin `.env` through `/^MEMCLAW_[A-Z_]+$/`,
so every `CAURA_*` key was dropped before it reached `process.env`. Until that
changed, the new names would have been accepted everywhere and read nowhere —
the rest of the lane would look done while doing nothing. It is now the
exported, unit-tested `isPluginEnvKey`.

That predicate is load-bearing for a **security** reason, not a cosmetic one:
its prefix bound is what stops a plugin `.env` setting `PATH` or
`NODE_OPTIONS`. Tests pin the exclusions, not just the inclusions.

### Two things worth reviewer attention

**1. core-api deliberately does NOT use `AliasChoices`.** The work order
specified it; it is unsafe here. `AliasChoices` resolves to the first alias
that is *defined*, and the empty string counts:

```
CAURA_API_KEY=""  +  MEMCLAW_API_KEY="real-key"   ->   ""
```

`auth.py:376` gates the whole Path-2 perimeter on `if mclaw_key:`, so that
resolves to falsey and **the API-key perimeter drops silently**. The startup
guard at `app.py:157` only catches it when `gateway_shared_secret` is also
unset, so a deploy with a gateway secret boots clean and quietly loses a
perimeter. The trigger is mundane — a deploy template that has started listing
new names with an unfilled `CAURA_API_KEY=`, which is exactly what 5.4 will
cause people to write. Replaced with two plain fields collapsed by a model
validator, giving first-**non-empty**; for a secret, blank and unset mean the
same thing. Pinned by `test_blank_new_name_does_not_disable_the_perimeter`.

**2. `deploy.ts` keeps a looser filter on purpose.** `hasPluginEnvPrefix` is a
bare prefix test because `deploy.ts` uses it to *preserve* an operator's
existing `.env` keys across a redeploy. Reusing the strict predicate there
would have silently deleted anything it rejects (`MEMCLAW_API_KEY2`,
lowercase) from a file we do not own the contents of. The strict/loose split
is intentional and tested.

### Aliases are listed in full, not derived

Every call site spells both names. A `readEnv("RECALL_GATE")` form that built
them from a suffix was rejected: it makes the old names ungreppable, and
grepping for the old brand is how this programme is tracked.

The cost of that choice is 28 hand-written pairs, where a suffix typo
type-checks, passes any test that only sets the new name, and strands the old
name permanently. So a source-scraped guard asserts all 28 share a suffix. It
globs `plugin/src` rather than listing files, and covers the `_readBoolEnv` /
`_readIntEnv` call sites — those are the multi-line ones, where a typo is
hardest to eyeball, and one of them gates cross-agent read isolation.

### Scope note

`docker-compose.yml` interpolates the image tag in a *different process*, so a
code-level dual-read cannot reach it. Without its own nested default an
operator following 5.4-updated docs would set `CAURA_VERSION` and silently get
`:latest`, while core-api reported the pinned version. Verified all three
cases resolve (`CAURA_VERSION`, `MEMCLAW_VERSION`, neither).

`plugin/openclaw.plugin.json` declares both names in `configSchema` — the
plugin's registration object carries an empty schema, so the manifest is the
real contract with the OpenClaw host.

### Verification

- 406 plugin tests, 139 client tests, 56 core-api tests — all pass
- `ruff check` and `ruff format --check` clean at CI's exact scopes
- **Every new test confirmed failing without its fix** (2 TS, 4 core-api, 12
  client), including the planted-typo check on the alias-pair guard
- Ratchet: **-4 net, no unexempted additions**; 10 exemptions, each a compat
  alias or a test pinning an old name
- Sentinel (#884): all 22 protected strings survive

### Out of scope, flagged not fixed

- `tool-definitions.ts` `enrichBody` now honours `CAURA_FLEET_ID` for free, but
  the `caura_list {scope:'all'}` narrowing is a *behavioural* bug (it injects
  `fleet_id` whenever the caller omits it). Fixing that is not additive.
- The three `.env` line parsers (`env.ts`, `deploy.ts`, `index.ts`) still
  differ on quote-stripping. Unifying them would change `deploy.ts` behaviour,
  so it wants its own PR.
- Observed once and not reproduced in 5 runs: `validation.test.ts`
  "tampered signature fails closed" flaked. `verifyCommandSignature` checks
  timestamp freshness *before* the HMAC, so a slow run can return
  `expired_timestamp` where the test expects `invalid_signature`. Untouched by
  this PR.

### DIFF
--- a/clients/python/src/caura_client/interviewer/cli.py
+++ b/clients/python/src/caura_client/interviewer/cli.py
@@ -48,6 +48,23 @@ def _default_agent_id() -> str:
     return f"cc-{getpass.getuser()}@{socket.gethostname()}"
 
 
+def _read_env(*names: str, default: str = "") -> str:
+    """First alias carrying a value, new names before old ones.
+
+    A blank alias never shadows a working one, but a lone ``X=`` still means
+    "empty" rather than falling back to ``default`` — which is what
+    ``os.environ.get(name, default)`` did before the aliases existed.
+    """
+    saw_blank = False
+    for name in names:
+        value = os.environ.get(name)
+        if value:
+            return value
+        if value is not None:
+            saw_blank = True
+    return "" if saw_blank else default
+
+
 def _build_parser() -> argparse.ArgumentParser:
     parser = argparse.ArgumentParser(
         prog="memclaw-interviewer",
@@ -56,11 +73,17 @@ def _build_parser() -> argparse.ArgumentParser:
     sub = parser.add_subparsers(dest="command", required=True)
 
     def common(p: argparse.ArgumentParser) -> None:
-        p.add_argument("--base-url", default=os.environ.get("MEMCLAW_BASE_URL", DEFAULT_BASE_URL))
-        p.add_argument("--api-key", default=os.environ.get("MEMCLAW_API_KEY", ""))
-        p.add_argument("--tenant-id", default=os.environ.get("MEMCLAW_TENANT_ID", ""))
-        p.add_argument("--agent-id", default=os.environ.get("MEMCLAW_AGENT_ID", "") or _default_agent_id())
-        p.add_argument("--fleet-id", default=os.environ.get("MEMCLAW_FLEET_ID", "") or None)
+        p.add_argument(
+            "--base-url",
+            default=_read_env("CAURA_BASE_URL", "MEMCLAW_BASE_URL", default=DEFAULT_BASE_URL),  # legacy-name-ok: rule 3 dual-read alias
+        )
+        p.add_argument("--api-key", default=_read_env("CAURA_API_KEY", "MEMCLAW_API_KEY"))  # legacy-name-ok: rule 3 dual-read alias
+        p.add_argument("--tenant-id", default=_read_env("CAURA_TENANT_ID", "MEMCLAW_TENANT_ID"))  # legacy-name-ok: rule 3 dual-read alias
+        p.add_argument(
+            "--agent-id",
+            default=_read_env("CAURA_AGENT_ID", "MEMCLAW_AGENT_ID") or _default_agent_id(),  # legacy-name-ok: rule 3 dual-read alias
+        )
+        p.add_argument("--fleet-id", default=_read_env("CAURA_FLEET_ID", "MEMCLAW_FLEET_ID") or None)  # legacy-name-ok: rule 3 dual-read alias
         p.add_argument(
             "--projects",
             nargs="+",
@@ -71,7 +94,11 @@ def _build_parser() -> argparse.ArgumentParser:
         p.add_argument(
             "--harness",
             choices=(HARNESS_CLAUDE_CODE, HARNESS_CURSOR),
-            default=os.environ.get("MEMCLAW_INTERVIEWER_HARNESS", HARNESS_CLAUDE_CODE),
+            default=_read_env(
+                "CAURA_INTERVIEWER_HARNESS",
+                "MEMCLAW_INTERVIEWER_HARNESS",  # legacy-name-ok: rule 3 dual-read alias
+                default=HARNESS_CLAUDE_CODE,
+            ),
             help="which agent's transcripts to harvest (default: claude-code; env MEMCLAW_INTERVIEWER_HARNESS)",
         )
         p.add_argument(
@@ -121,7 +148,7 @@ def _projects_root(args: argparse.Namespace) -> Path:
 def _resolve_allowlist(args: argparse.Namespace) -> list[str]:
     if args.projects is not None:
         return list(args.projects)
-    env = os.environ.get("MEMCLAW_INTERVIEWER_PROJECTS", "")
+    env = _read_env("CAURA_INTERVIEWER_PROJECTS", "MEMCLAW_INTERVIEWER_PROJECTS")  # legacy-name-ok: rule 3 dual-read alias
     return [g.strip() for g in env.split(",") if g.strip()]
 
 
--- a/clients/python/tests/test_interviewer_env_dual_read.py
+++ b/clients/python/tests/test_interviewer_env_dual_read.py
@@ -1,0 +1,97 @@
+"""Phase 5.1: the interviewer CLI reads both env spellings.
+
+Rule 3 makes the old names permanent, so every case is asserted in both
+directions — asserting only the new name would pass after someone deleted the
+fallback and stranded every crontab an older installer wrote.
+
+Old names are confined to the constants below so the ratchet has one marked
+line per name rather than one per assertion.
+"""
+
+import pytest
+
+from caura_client.interviewer.cli import _build_parser, _read_env, _resolve_allowlist
+
+OLD_PREFIX = "MEMCLAW_"  # legacy-name-ok: rule 3 — the alias prefix under test
+NEW_PREFIX = "CAURA_"
+
+ALIASED = [
+    ("api_key", "API_KEY", "k-value"),
+    ("tenant_id", "TENANT_ID", "t-value"),
+    ("agent_id", "AGENT_ID", "a-value"),
+    ("fleet_id", "FLEET_ID", "f-value"),
+    ("base_url", "BASE_URL", "https://example.invalid"),
+]
+
+
+SUFFIXES = [suffix for _, suffix, _ in ALIASED] + ["INTERVIEWER_PROJECTS", "INTERVIEWER_HARNESS"]
+
+
+@pytest.fixture
+def clean_env(monkeypatch):
+    for suffix in SUFFIXES:
+        monkeypatch.delenv(NEW_PREFIX + suffix, raising=False)
+        monkeypatch.delenv(OLD_PREFIX + suffix, raising=False)
+    return monkeypatch
+
+
+def _parse(argv):
+    return _build_parser().parse_args(argv)
+
+
+@pytest.mark.parametrize(("dest", "suffix", "value"), ALIASED)
+def test_old_name_still_read(clean_env, dest, suffix, value):
+    clean_env.setenv(OLD_PREFIX + suffix, value)
+    assert getattr(_parse(["status"]), dest) == value
+
+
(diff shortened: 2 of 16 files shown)


Eldad-Caura:
On the High finding — **not reproduced.** The review's own instruction was to
verify with `docker compose config` and replace *if confirmed broken*. I ran
it; it resolves correctly, so I've kept the nested default and added the test
coverage the review correctly noted was missing.

Compose v2 **does** resolve a `${...}` nested inside a default. All four
combinations, from `docker compose config --images`:

| `CAURA_VERSION` | `MEMCLAW_VERSION` | resolved tag |
|---|---|---|
| unset | unset | `:latest` |
| unset | `v0.9.9` | `:v0.9.9` |
| `v2.0.0` | unset | `:v2.0.0` |
| `v2.0.0` | `v0.9.9` | `:v2.0.0` |

Row 2 is the "every existing user" path the finding was about, and row 1 is the
default path — both correct. I also grepped the fully rendered config with both
names unset for a leaked literal `MEMCLAW_VERSION`: **0 occurrences**, so
nothing is being substituted in unparsed.

Why v2 is a safe floor here rather than an assumption:

- `README.md:124` already requires it verbatim — "**Docker Compose v2** (built
  into modern Docker). Confirm with `docker compose version`."
- This file has no `version:` key, i.e. it is the Compose Specification schema
  that only v2 reads.
- Documented usage is `docker compose up -d` (the v2 CLI), on line 2 of the file.

Compose v1 would indeed not handle this, but it is EOL and out of support, and
a v1 user cannot parse this file today regardless of my change.

**What I did change,** because the "isn't covered by any of the new tests" part
was a fair hit and there was no compose test at all:

- `tests/test_env_dual_read.py::TestComposeImageTags` pins that every
  **first-party** `ghcr.io/caura-ai/*` image interpolates the new name with the
  old one as its fallback. Verified failing in both directions that matter —
  flattened to the new name alone, and reverted to the old name alone. It
  deliberately ignores third-party images (pgvector, redis, TEI), which carry
  their own upstream tags.
- A comment at the image marking the nesting as deliberate, naming the v2
  requirement, and saying not to flatten it — since dropping either name is the
  failure mode. The rejected alternatives from the review prompt (a wrapper
  script, or reading only the new name) both break an existing pin: this var is
  how installs pin today.

Worth stating plainly for the record: had it been broken, the finding would have
been right about the blast radius. Every existing install goes through row 2 or
row 1.

Eldad-Caura:
@claude re-review please — force-pushed since the last review, and the review job skips `synchronize`.

Changes since: a compose-tag test plus a comment on the nested default. The High finding was verified not-reproducible (evidence in the comment above); nothing else in the diff moved.

Eldad-Caura:
Fixed the Low, and deferring both Suggestions with reasons.

**Fixed — `plugin/src/deploy.ts:6`.** Correct and the right catch: that line is
a security-boundary comment describing what a *remote* deploy command may write
into the plugin's `.env`, and it had gone stale the moment the filter changed.
Now names `hasPluginEnvPrefix` rather than a prefix literal, so it cannot drift
again the next time the prefix set moves, and it states the property being
enforced. I swept the other two files this PR touches for the same class of
stale filter/allow-list comment — none.

**Deferred — cli.py user-facing strings (Suggestion 1).** Agreed on substance:
`--help` and the `_require_config` error currently teach only the old name,
which is a real discoverability gap. It is deliberately out of scope here.
This PR is work-order item 5.1 (readers); rewriting docs, help text and error
strings to *teach* the new name is item 5.4, which is explicitly scoped to
"rewrite every doc, installer heredoc, workflow and test" across both repos in
one pass. Doing a slice of it here would fork that work and conflict with it.

Recording the exact sites so 5.4 does not have to rediscover them:
`cli.py` module docstring (lines 4-6), `--projects` help (68), `--harness`
help (75), `_require_config` (both messages), `_deny_guidance`. Also
`.env.example:45-55` and `docker-compose.yml:63`, same class.

**Already covered — plugin manifest (Suggestion 2).** The legacy property's
description reads "Accepted forever as an alias for CAURA_AGENT_ID; the new
name wins when both are set", and the new name is listed first, so a form built
off the schema renders them in that order. I'd rather not add a "(legacy)"
label on top: "deprecated" would be actively wrong here — rule 3 makes the old
name permanent, not scheduled for removal.

Re-verified after the amend: 406 plugin / 139 client / 57 core-api tests pass,
ruff clean at CI's scopes, sentinel all 22 strings survive, ratchet **-5 net**
with no unexempted additions.

Eldad-Caura:
@claude re-review please — force-pushed for the deploy.ts security-comment fix; the review job skips `synchronize`.

Eldad-Caura:
Rebased onto #885 and the ratchet is green again. Also: both new Lows are the
deferred-to-5.4 class — reasons below.

**Why the ratchet went red.** #885 landed while this PR was open and changed the
model from "per-file count went up" to comparing line **text**. That is a
strictly better gate — its docstring is explicit that a count is the wrong check
in both directions — but it means a dual-read PR looks maximally suspicious to
it: rewriting `process.env.MEMCLAW_X` into
`readEnv(["CAURA_X", "MEMCLAW_X"])` produces new brand-carrying text on every
line, even where the count is flat or **falls**. My local run had passed against
the pre-#885 script on my branch, which is why I did not see it before pushing.

The fix is the mechanism #885 documents for exactly this: each of those lines is
a compat alias, so each carries `legacy-name-ok` with a reason. 58 markers.
Post-rebase the ratchet reports **"No new lines. 50 removed by this change
(-50 net)."**

Worth stating rather than burying: 58 exemptions is a lot of them to eyeball,
and the gate prints every one on the passing path so a reviewer can. They are
one homogeneous class — `readEnv([new, old])` / `_read_env(new, old)` call sites
plus the two prefix predicates and the compose image tags. If any single one is
not a compat alias, that is a real finding and I'd want to know.

Three things the markers broke and I had to fix, since appended comments change
the text my own source-scraping tests read:

- the `MEMCLAW_KEYSTONES_TOKEN_CAP` default pin — its regex now tolerates a
  comment between the alias array and the default
- the compose-tag test — its `$` anchor now absorbs a trailing YAML comment
- `constants.py` needed a shorter marker: the long form pushed the line past
  ruff's width, and `ruff format` would have wrapped the brand onto a line with
  no marker on it, re-failing the gate in a way that looks unrelated

Re-verified after the rebase: 406 plugin / 139 client / 57 core-api tests pass,
`ruff check` + `format --check` clean at CI's scopes, sentinel all 22 strings
survive, and all four compose interpolation cases still resolve.

**The two new Lows** — `012_vector_dim_1024.py:122,154` and `cli.py:99` — are
both the same class I deferred last round: operator-facing prose that still
teaches only the old name. Consistent with that, deferring to item 5.4. The
code reads both names in every case; only the guidance strings lag, and an
operator who follows them gets a working result.

This class is now the third round it has been raised, which is itself the
argument for doing it as one pass rather than per-PR: the full set includes
`.env.example`, `scripts/`, `docs/`, both `--help` surfaces and several error
messages across two repos. A slice here forks that work and conflicts with it.
Inventory so far, so 5.4 does not rediscover it:

- `cli.py` docstring 4-6, `--projects` help 68, `--harness` help 99,
  `_require_config` (both messages), `_deny_guidance`
- `012_vector_dim_1024.py` pre-flight comment 122, `RuntimeError` 154
- `.env.example:45-55`, `docker-compose.yml:63`

Eldad-Caura:
@claude re-review please — rebased onto #885's text-based ratchet and added the alias markers it requires; the review job skips `synchronize`.

Eldad-Caura:
Split the last Low, because two of its three sites aren't the same thing.

**Fixed — `constants.py` `_resolve_version` docstring.** This one was mine to
fix and I should have caught it: the docstring states a numbered *precedence
list*, and my change made line 1 of it false about the function it documents.
That is stale documentation introduced by this PR, not documentation that
happens to predate the rename. Same category as the `deploy.ts` security
comment last round.

Worded as "``CAURA_VERSION`` env, or its legacy alias below — the new name wins
when both are set", which fixes the precedence claim without adding another
brand-carrying line to mark.

**Still deferred — `cli.py` docstring + `--projects` / `--harness` help.** These
are a catalogue for discovery: incomplete, not incorrect. Nothing in them
asserts a precedence that the code contradicts, and an operator who follows them
gets a working result. Teaching the new spelling in user-facing help across both
repos is item 5.4, in one pass, for the reason this thread demonstrates — it has
surfaced in four consecutive reviews and the full set spans `.env.example`,
`scripts/`, `docs/`, two `--help` surfaces and several error messages.

The line I'm drawing, so it's reviewable rather than just asserted:

- **fix now** — a docstring or comment that became *wrong* about code this PR
  changed (`deploy.ts:6`, `constants.py:74`)
- **defer to 5.4** — prose that is merely *incomplete*, i.e. still lists only
  the old name but claims nothing false (`cli.py` docstring and help text,
  `012_vector_dim_1024.py:122,154`, `.env.example`)

If you'd rather the incomplete ones land here too, say so and I'll fold them
in — it's a scope call, not a technical one, and I'd rather be told than guess
a second time.

Green after the amend: ratchet **-51 net, no new lines**, sentinel all 22
strings survive, ruff clean, 23 core-api tests in the touched suites pass.

Eldad-Caura:
@claude re-review please — force-pushed for the `constants.py` docstring precedence fix; the review job skips `synchronize`.

Eldad-Caura:
@erni-a pleae approve

erni-a (approved):

PR #888, opened 2026-08-23
fix(ratchet): name the line the change added, and group repeated exemptions

Eldad-Caura:
Two report fixes on top of #885. Neither changes what passes or fails. Both change whether the output can be trusted — which, for a gate whose entire product is a diagnostic, is the same class of defect as a wrong verdict.

## 1. The failure report could name a line nobody touched

Found by the review on [installer#11](https://github.com/caura-ai/caura-onprem-installer/pull/11), and it is a regression I introduced in #885.

Spending a text-keyed budget in file order assumes the first occurrence of a text is the new one. **Line order says nothing about that.** A file that already holds the text on line 3 and gains an identical line on line 14:

```
  f.py  (1 -> 2)
      3: T = "memclaw-thing"     <- unchanged line, named
                                 <- line 14, the actual addition, never printed
```

The reader is sent to a line that did not change, hunting a mistake that is not there, by a gate whose whole output is "here is the line you added". Worse than imprecise — before #885 the report printed every matching line in the file, so the real one was at least *among* the output.

**Which physical line a change added is the one question the text cannot answer and git's diff can**, so the report now asks git:

```
  f.py  (1 -> 2)
      14: T = "memclaw-thing"
```

The **decision** still compares text and must: a per-line diff calls every move an addition, which is exactly what #885 fixed. The report has the opposite problem and needs the opposite tool. `_added_lines()` is used for reporting only, returns empty on any failure, and the caller falls back to naming every occurrence of the minted text — too many lines is a nuisance, none is a dead end.

Where several identical lines were **all** added and only some are minted, all are printed and the split is stated:

```
      1: URL = "https://memclaw.net"
      2: URL = "https://memclaw.net"
      3: URL = "https://memclaw.net"
      ^ 3 added lines share this text; 1 new, 2 moved in from elsewhere — identical, so which is which cannot be told apart
```

That replaces `test_only_the_minted_copies_of_a_repeated_line_are_named`, which I added two rounds ago to satisfy an earlier finding. It was asking for false precision: all three are lines the change added, they are byte-identical, and naming one is a guess dressed as a finding.

## 2. The exemption report became unreadable at exactly the size that matters

Reported by Agent 5 from a live Phase 5 PR, and worth acting on rather than filing.

A dual-read wave exempts dozens of lines carrying one identical reason — their PR has **58 markers**, and 5.3 and 5.4 make that worse, not better. At that length the list stops being read, which costs the report the thing it exists for: the exempt-and-add swap it guards against shows up as the **one reason that does not match its neighbours**.

Repeats past four are now collapsed:

```
  6x  rule 3 dual-read alias
      in 1 file(s): many.py
  many.py:7: SNUCK = "memclaw-new"  # legacy-name-ok: headroom, honestly
```

**Grouping cannot hide the hole it guards** — a laundered line carries a reason that does not match the group, so collapsing the group is precisely what surfaces it. Pinned both ways: `test_many_exemptions_sharing_a_reason_are_grouped` and `test_an_odd_reason_out_survives_the_grouping`.

## Tests

40, up from 37. Both new behaviours confirmed failing against the pre-fix reporting.

Gates green: `No new lines.` / `All 22 protected strings survive.`

## Propagation

This goes to `caura-enterprise` and `caura-onprem-installer` once it lands, same as #885 — all three copies are now byte-identical, so it is a straight copy with an AST check rather than a semantic port. [installer#11](https://github.com/caura-ai/caura-onprem-installer/pull/11) is still open and will carry this before it merges; the enterprise copy (#1225, merged) needs a follow-up PR.

### DIFF
--- a/scripts/legacy_name_ratchet.py
+++ b/scripts/legacy_name_ratchet.py
@@ -104,6 +104,12 @@ EXEMPT_RE = re.compile(
 )
 
 
+# Above this many lines sharing one reason, the list is collapsed to a count.
+# Four is where a homogeneous block stops being read line by line; below it the
+# individual lines are still the more useful form.
+_EXEMPTION_GROUP_AT = 4
+
+
 def _git(args: list[str]) -> str:
     """git, with ``grep``'s no-match exit treated as an empty result.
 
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #891, opened 2026-08-23
fix(sentinel): a file the gate cannot read exits 2, not a traceback

Eldad-Caura:
Found reviewing the enterprise port ([enterprise#1228](https://github.com/caura-ai/caura-enterprise/pull/1228)).

`_log_calls` raises `RuntimeError` when a pinned Python file will not parse — deliberately, so an unreadable file is a failed gate rather than a silent skip. But `main()` called `_check` inside a list comprehension with nothing catching it, so the gate crashed with a traceback and Python's default **exit 1**.

**Exit 1 is not free here.** In this script it means the gate *ran* and something load-bearing is missing — the loudest signal it has. A gate that could not run must never be mistaken for a gate that failed you. `main()` now catches it and exits **2**, the same distinction the ratchet already draws for an unresolvable `--base`.

## On the test

Driven through `main()` with a **substituted list** rather than the real one. Only a `LOG_MESSAGE` entry reaches the parser, and not every repo's list has one — the enterprise list has none — so driving it off the real `SENTINELS` would pass vacuously in exactly the copies where this was found. The substituted form makes the case identical in all three copies, which is the property that keeps them diffable.

Confirmed failing with the guard removed.

## Checks

61 tests pass. `All 22 protected strings survive.` / `No new lines.`

Propagates to `caura-onprem-installer` and `caura-enterprise`; the enterprise copy already carries it in #1228, noted there.

### DIFF
--- a/scripts/do_not_touch_sentinel.py
+++ b/scripts/do_not_touch_sentinel.py
@@ -364,7 +364,14 @@ def main() -> int:
         return 1
 
     root = Path(args.root).resolve()
-    failures = [(s, why) for s in SENTINELS if (why := _check(s, root)) is not None]
+    try:
+        failures = [(s, why) for s in SENTINELS if (why := _check(s, root)) is not None]
+    except RuntimeError as exc:
+        # A file this gate cannot read is a gate that did not run, which is not
+        # the same as a gate that found something — and a traceback reads as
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #893, opened 2026-08-23
fix(sentinel): the log message is the first argument, not any of them

Eldad-Caura:
Found reviewing the enterprise port ([enterprise#1228](https://github.com/caura-ai/caura-enterprise/pull/1228)).

`_log_calls` treated **every** positional string argument as candidate message text, so a `%`-substitution *value* could satisfy a `LOG_MESSAGE` entry for a call whose message had changed:

```python
logger.error("Widget renamed: %s", "Widget degraded")
```

still reported `Widget degraded` as emitted. That is a false pass on exactly the regression this kind exists to catch — and the harder one to notice, because the phrase really *is* still in the file, so grepping confirms the wrong answer.

**It matters most here.** The three entries this guards are the embedding phrases matched by the CRITICAL Datadog policy at `gcp_alerts.tf:40`. A rename that moved one of those phrases from the format string into an argument would have passed the gate and killed the alert.

## Why not simply "the first argument"

The obvious fix breaks `logger.log(level, msg)`, where the level takes the first slot — a case added deliberately two rounds ago, with its own test. So it is the first argument, except for `.log`, where it is the second:

```python
index = 1 if node.func.attr == "log" else 0
```

Both cases still pinned: `test_the_level_first_logging_form_is_read_too` and the new `test_a_format_argument_does_not_count_as_the_message`, the latter confirmed failing against the read-every-argument version.

## Note on where this was found

The enterprise list has **no** `LOG_MESSAGE` entries — the monitored log lines are emitted by other repos; enterprise holds the monitors that read them. So the port could not have hit this bug, and the reviewer found it by reading the mechanism rather than by running it. Fixed here first because this is the copy where the kind is actually load-bearing.

## Checks

62 tests pass. `All 22 protected strings survive.` / `No new lines.`

Propagates to the installer; enterprise already carries it in #1228.

### DIFF
--- a/scripts/do_not_touch_sentinel.py
+++ b/scripts/do_not_touch_sentinel.py
@@ -85,11 +85,12 @@ class Sentinel:
 
 # ── the list ─────────────────────────────────────────────────────────────────
 #
-# Every entry carries ``legacy-name-ok`` because this file is itself scanned by
-# the ratchet, and the reason is the same one every time: the line exists to pin
-# a string rule 3 keeps readable forever. Excluding the file from the ratchet
-# instead would leave a hole in that scan, which is the trade its author already
-# refused once for the same reason.
+# Entries whose pinned text contains the old brand carry ``legacy-name-ok``,
+# because this file is itself scanned by the ratchet, and the reason is the same
+# every time: the line exists to pin a string rule 3 keeps readable forever.
+# Entries pinning brand-free text need no marker and correctly have none — the
+# ratchet never sees them. Excluding this file from the ratchet instead would
+# leave a hole in that scan, which is the trade its author already refused once.
 
 SENTINELS: tuple[Sentinel, ...] = (
     # -- Matched by a live production Datadog monitor. Carries no brand. -------
@@ -273,9 +274,11 @@ def _static_text(node: ast.expr) -> str | None:
 def _log_calls(source: str, path: str) -> list[tuple[str | None, str]]:
     """``(level, message)`` for every logging call — level ``None`` if unknowable.
 
-    Positional args only, and every one of them: ``logger.log`` takes the level
-    first, so keying on argument position would miss the message while keying on
-    "any string argument" costs nothing.
+    One argument per call, by position: index 0, or index 1 for ``logger.log``
+    where the level takes the first slot. Reading every positional argument
+    instead — which this did until a %-substitution value was found able to
+    satisfy a check for a message that had been renamed — is the wider net that
+    catches the wrong fish.
     """
     try:
         tree = ast.parse(source)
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
Fixed. The docstring said the opposite of what the code now does, in a file whose whole design rests on prose describing behaviour accurately — so this is the defect class the file itself is most exposed to.

```
One argument per call, by position: index 0, or index 1 for ``logger.log``
where the level takes the first slot. Reading every positional argument
instead — which this did until a %-substitution value was found able to
satisfy a check for a message that had been renamed — is the wider net that
catches the wrong fish.
```

It now records *why* the narrower read is right rather than just what it does, since the previous text was written to defend the opposite choice and someone will eventually wonder which was intended.

The same stale docstring is in the two ports. Enterprise carries it in #1228 (fixing there now); the installer merged as #12 before this landed, so it gets a follow-up — which also has to restore its own module docstring, clobbered by my assembly in #12 taking the OSS head wholesale. Reported separately on #12 and correct: it now claims this gate protects three embedding phrases in a repo that has no application code at all.

62 tests pass. `All 22 protected strings survive.` / `No new lines.`

@claude

Eldad-Caura:
Both taken, and both are the same defect class as everything else this file has attracted — the code and the prose about it disagreeing.

## A mistyped `kind` silently ran the wrong check

`_check` returned early for `LITERAL` and then fell straight into the log-message path, so any other value — a typo when adding an entry — would have been checked as `LOG_MESSAGE`. The entry's author would believe they had pinned a literal, and the gate would pass or fail for reasons unrelated to what they wrote.

Now explicit, and it raises:

```python
if sentinel.kind != LOG_MESSAGE:
    raise RuntimeError(f"unknown kind {sentinel.kind!r} on {sentinel.path}")
```

Not defensive padding: **a gate running the wrong check is worse than one that refuses to run**, which is the distinction this file already draws for a missing file and an unparseable one. It surfaces through the exit-2 path added earlier, so it reads as "could not run" rather than "found something".

Pinned by `test_an_unknown_kind_refuses_to_run`, which passes a plausible typo (`"litteral"`) rather than obvious junk.

## "Every entry carries `legacy-name-ok`" was wrong

Correct, and I wrote that comment before the list had any brand-free entries. Several now correctly have no marker — the `_next/static` greps, `"dd-monitor:true"`, the embedding-phrase filter — because their pinned text contains no old brand and the ratchet never sees them. The comment now says the marker applies to entries containing the old-brand literal, and that brand-free entries need none, so the next person to add one does not go looking for a tag that should not be there.

## Carried to all three copies in the same round

Both changes are in the shared mechanism, so rather than let the copies diverge and chase them separately they went into all three open PRs together: [caura#893](https://github.com/caura-ai/caura/pull/893), this one, and [installer#13](https://github.com/caura-ai/caura-onprem-installer/pull/13). Verified after the fact rather than assumed:

```
OSS         mechanism functions byte-identical to OSS: True   (4 functions)
ENTERPRISE  mechanism functions byte-identical to OSS: True   (4 functions)
INSTALLER   mechanism functions byte-identical to OSS: True   (4 functions)
```

## Checks

63 tests pass. `All 22 protected strings survive.` / `No new lines.`

@claude

Eldad-Caura:
Green on all 10 checks, review clean. Ready to merge.

@erni-a

erni-a (approved):

PR #895, opened 2026-08-23
feat(installers): write CAURA_* into new installs

Eldad-Caura:
Phase 5.2 — the writer half of the env rename, and the follow-up to #886.

The server-generated installer and the interviewer installer now write the new
names into a **new** customer's `.env` file and crontab. **New names only, not
both:** every reader accepts either spelling since #886, so nothing we ship
needs the old name in a file we are creating from scratch, and rule 7 says not
to mint one.

**Existing installs are untouched.** Their `.env` keeps the old names and keeps
working. This changes what a *new* file contains — it is not a migration, and
nothing rewrites a file already on disk.

### Why this can land beside 5.1 rather than after a release

The work order's precondition is "5.1 deployed", which I initially read as
needing a published artifact first. It doesn't, and the reason is worth stating
because it is the thing that makes this PR safe:

| writer | its reader | ships as |
|---|---|---|
| `routes/plugin.py` | `plugin/src/env.ts` | one core-api deploy — the plugin is `"private": true`, never published to npm; the installer curls source from `/api/plugin-source`, which reads core-api's own filesystem |
| `interviewer/installer.py` | `interviewer/cli.py` | one `caura-client` wheel |

At both sites the reader ships in the **same unit** as the writer, so there is
no window where a new-name file meets an old-name reader. Ordering inside the
generated script confirms it: source is fetched at 441-460, the `.env` written
at 470-477, the build runs at 483 — one run.

The one residual, for completeness: the `deploy` command can push env vars to
an install still running older source, where a `CAURA_*` key hits the older
filter and is dropped. That is a silent no-op — existing values persist — and
it self-heals on the next source fetch. #886 already replaced that filter with
`hasPluginEnvPrefix`, which accepts both.

### The security tests needed care, not a rename

Two of them are **negative** assertions:

```python
assert "MEMCLAW_NODE_NAME=$(whoami)" not in script   # injection guard
assert "MEMCLAW_API_KEY=" not in script              # no key when none supplied
```

After renaming the shell variables these would pass **because the old name no
longer appears anywhere in the script**, proving nothing. A rename that quietly
turns an injection test into a tautology is worse than the rename. Both now
name the new variable, and I re-checked the node-name one by removing
`shlex.quote` from the generator and confirming it still fails on a real
injection regression.

`test_api_plugin.py:161` was a related trap — not an assert but a list
comprehension indexed `[0]`, so a missed rename would have surfaced as an
IndexError rather than a clean failure.

### Left alone deliberately

These look renameable and are not — each strands or orphans live installs:

- the plugin's on-disk path `~/.openclaw/plugins/memclaw` *(sentinel-pinned)*
- the interviewer's config dir `~/.config/memclaw-interviewer` *(sentinel-pinned)*
- `CRON_MARKER` — `merge_crontab` matches this literal, so changing it means
  `uninstall` can no longer find an installed entry and a re-install
  **duplicates** rather than replaces. Equally floor-class and **not currently
  sentinel-pinned** — worth adding.
- the shipped console-script name `memclaw-interviewer` *(pinned via
  `pyproject.toml`)*, the `~/.bashrc` CA marker, and the systemd drop-in
  filename

### Scope

`_ENV_KEYS` in `installer.py` is an **allowlist** filtered against the dict
`cli.py` builds, so those two had to flip in the same commit — a mismatch
produces an env file with no exports at all and a cron job that fails auth with
no clue why.

`scripts/wet-test-install.sh` writes to the same customer path, so it moves too
rather than being left to disagree with the real installer.

Two fixture hostnames (`memclaw.dev`, `memclaw.corp.internal`) and one prose
string were cleaned rather than exempted — they are arbitrary test data and
product naming, so marking them as "compat aliases" would have been a false
reason on a gate whose value depends on its reasons being true.

### Deferred to 5.4, unchanged

Operator-facing guidance that still names only the old spelling: `cli.py`
docstring and `--help`, the `agent_identity.py` error message that says "set
MEMCLAW_AGENT_ID", `012_vector_dim_1024.py`'s pre-flight text, `.env.example`,
`README.md`. All are incomplete rather than incorrect — the old name still
works everywhere it is named.

### Verification

- 106 core-api / 139 client / 101 gate-self-tests pass; `ruff check` +
  `format --check` clean at CI's scopes; `bash -n` on the harness
- the new "only new names" test confirmed failing when a belt-and-braces old
  name is planted in the `.env` heredoc
- the injection test confirmed failing when `shlex.quote` is removed
- Ratchet **-115 net, no new lines**; sentinel all 22 strings survive

### DIFF
--- a/clients/python/src/caura_client/interviewer/cli.py
+++ b/clients/python/src/caura_client/interviewer/cli.py
@@ -403,15 +403,17 @@ def _cmd_install(args: argparse.Namespace) -> int:
         print(f"[interviewer] {exc}", file=sys.stderr)
         return 2
 
+    # New installs are written with the new names only. Existing env files keep
+    # their old names and keep working — every reader above accepts both.
     env = {
-        "MEMCLAW_BASE_URL": args.base_url,
-        "MEMCLAW_API_KEY": args.api_key,
-        "MEMCLAW_TENANT_ID": args.tenant_id,
-        "MEMCLAW_AGENT_ID": args.agent_id,
-        "MEMCLAW_FLEET_ID": args.fleet_id or "",
+        "CAURA_BASE_URL": args.base_url,
+        "CAURA_API_KEY": args.api_key,
+        "CAURA_TENANT_ID": args.tenant_id,
+        "CAURA_AGENT_ID": args.agent_id,
+        "CAURA_FLEET_ID": args.fleet_id or "",
     }
     if not args.all_projects:
-        env["MEMCLAW_INTERVIEWER_PROJECTS"] = ",".join(allow)
+        env["CAURA_INTERVIEWER_PROJECTS"] = ",".join(allow)
 
     env_file = installer.env_file_path()
     log_file = installer.log_file_path()
--- a/clients/python/src/caura_client/interviewer/installer.py
+++ b/clients/python/src/caura_client/interviewer/installer.py
@@ -35,12 +35,12 @@ CRON_MARKER = "# memclaw-interviewer (managed)"
 # Only the connection identity is persisted to the env file (0600). Non-secret
 # behavior flags (--harness, --all-projects) ride on the cron command instead.
 _ENV_KEYS = (
-    "MEMCLAW_BASE_URL",
-    "MEMCLAW_API_KEY",
-    "MEMCLAW_TENANT_ID",
-    "MEMCLAW_AGENT_ID",
-    "MEMCLAW_FLEET_ID",
-    "MEMCLAW_INTERVIEWER_PROJECTS",
+    "CAURA_BASE_URL",
+    "CAURA_API_KEY",
+    "CAURA_TENANT_ID",
+    "CAURA_AGENT_ID",
+    "CAURA_FLEET_ID",
+    "CAURA_INTERVIEWER_PROJECTS",
 )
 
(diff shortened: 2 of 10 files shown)


Eldad-Caura:
@erni-a ready for approval — green and the automated review came back clean on the first round.

Phase 5.2, the writer half of the env rename, follow-up to #886 (merged). New installs get `CAURA_*` names only; existing installs are untouched and keep working on the old ones.

Two things worth a human eye rather than a skim:

1. **The negative security assertions.** `assert "MEMCLAW_NODE_NAME=$(whoami)" not in script` would have passed vacuously after the rename — the old name no longer appears in the script at all — turning a shell-injection guard into a tautology. Both negatives now name the new variable, and I re-checked the injection one by removing `shlex.quote` from the generator to confirm it still fails on a real regression.

2. **What was deliberately not renamed.** The plugin's on-disk path, the interviewer config dir, `CRON_MARKER`, the console-script name, the bashrc CA marker and the systemd drop-in filename all stay. Renaming any of them strands or orphans live installs rather than renaming anything. `CRON_MARKER` is the one that is equally floor-class but **not** sentinel-pinned — if it were changed, `uninstall` would stop finding installed entries and a re-install would duplicate rather than replace. Worth pinning separately.

Not merging.

erni-a (approved):

PR #896, opened 2026-08-23
chore(assets): name the demo recording for the brand it should show

yanki-m:
`static/images/memclaw-demo.gif` → `static/images/caura-demo.gif`, plus the one reference at `README.md:48` — the hero animation directly under the title, and the first thing anyone visiting this repo sees.

## What this does not fix

The recording. The terminal title bar in the capture reads:

```
memclaw — governed shared memory for agent fleets
```

The old brand is in the **pixels**, not only the filename. I confirmed that by reading the first frame rather than inferring it from the name — which is why it is worth stating here, since a `git mv` looks like a fix and is not one.

The `alt` text beside it has said *"Caura demo — write, recall, and governed cross-fleet memory in action"* for some time, so the page already described one brand while showing another. This rename makes the filename agree with the alt text; the artwork still disagrees with both.

## What closes it

Re-record the demo with the corrected terminal title and overwrite `caura-demo.gif` in place. No further code change needed — this PR gets the name and the reference out of the way so the re-capture is a one-file commit.

Same shape as #887 (the social preview), where the rename landed first and the artwork followed.

Ratchet: `No new lines. 1 removed by this change (-1 net).`


erni-a (approved):
Reviewed: scoped correctly, no overlap with the overnight lane work, sentinel + ratchet green. Approving and merging per Erni.

PR #899, opened 2026-08-23
fix(ratchet): the exemption report lists only the lines the change wrote

### LINKED ISSUE #894: ratchet: the "newly exempted" report lists a file's pre-existing exemptions too
Found by `caura-ops`'s review while vendoring `scripts/legacy_name_ratchet.py` at `f490c422`, and reproduced against that exact file. Reporting rather than patching: caura-ops holds the script byte-identical so re-vendoring stays a one-liner, and this belongs where the script is maintained.

**Not a gate bug** — pass/fail is unaffected. It degrades the one report that guards the ratchet's known hole.

## What happens

`_report_new_exemptions` computes the per-file delta correctly and prints a correct total, then re-greps the file and prints **every** exempt line it currently holds. In a file that already had exemptions, the old ones are listed as if this change added them.

## Reproduction

```bash
git init -q -b main && git config user.email t@e.com && git config user.name t
B="mem""claw"
printf 'URL = "https://%s.net"\n' "$B" > other.py      # a non-exempt line, so the empty-tree guard stays quiet
cat > f.py <<EOF
A = "${B}-one"  # legacy-name-ok: pre-existing alias one
B = "${B}-two"  # legacy-name-ok: pre-existing alias two
EOF
git add -A && git commit -qm base
printf 'D = "%s-three"  # legacy-name-ok: brand new alias three\n' "$B" >> f.py
git add -A
python3 /path/to/legacy_name_ratchet.py --base HEAD
```

Output:

```
1 line(s) newly exempted in 1 file(s) — check each one is a
compat alias, a redirect or a pinned wire format, and not headroom for a new name:
  f.py:4: D = "memclaw-three"  # legacy-name-ok: brand new alias three
  f.py:1: A = "memclaw-one"  # legacy-name-ok: pre-existing alias one
  f.py:2: B = "memclaw-two"  # legacy-name-ok: pre-existing alias two
```

Header says **1**, list shows **3**. The genuinely new line is printed first — `_added_lines` from #888 is doing its job — but the other two are printed anyway.

## Why it is worth fixing

The script's own docstring explains that a count-based ratchet cannot distinguish "exempted an existing line to buy headroom" from "legitimately added a marked alias", so it reports every new exemption instead of adjudicating — *"no PR should be able to make one quietly"*.

That safeguard is only as good as the list being read. Phase 5's dual-read wave is exactly the change that fills files with legitimate exemptions, and from then on every later PR touching those files reprints the whole accumulated pile. The one line a reviewer needs to see gets buried in lines that are not part of the diff — and a reviewer who learns the list is mostly noise stops reading it, which is the failure the report exists to prevent.

It also makes the output self-contradictory (`1` vs three lines), which invites the reader to distrust the total rather than the list.

## Shape of a fix

`Scan.exempt` is per-file counts only, so `_report_new_exemptions` has nothing finer to filter by and falls back to a re-grep. Tracking exempt lines per `(path, stripped text)` — the way `by_file` already tracks non-exempt lines — would let it diff per text and print only the entries whose count rose, the way `_minted` / `_report_excused_moves` already work on precise deltas.

Two things worth preserving:

- **Identical texts.** If a file gains a second exemption whose text matches an existing one, they cannot be told apart. `_report_excused_moves` already handles that class by printing all and stating the split; the same treatment fits here.
- **The fallback.** `_added_lines` returns an empty set on any git failure and the caller falls back to naming everything. Keeping that is right — too many lines is a nuisance, none is a dead end.

No urgency; the gate is correct today.

Eldad-Caura:
Closes #894.

`_report_new_exemptions` computed its per-file delta correctly, then re-greped
each file and printed EVERY exempt line it currently holds. In a file that
already had exemptions, the old ones were listed as if this change had added
them -- header says 1, list shows 3.

Not a gate bug: pass/fail was unaffected. It degraded the one report that
guards the ratchet's known hole.

## The fix is the pair the offender report already uses

The file's exempt count decides WHICH FILE to talk about; git's own diff decides
which of its lines the change is responsible for. That split already exists
twenty lines below, for the same reason, with the same fallback -- the offender
report filters `_grep` candidates through `_added_lines`. This makes the two
consistent rather than inventing a mechanism.

Picking git's diff over the per-`(path, text)` tally the issue sketched, because
the diff answers the harder case for free: two exempt lines with identical text,
one old and one new, are indistinguishable by text and trivially distinguished
by line number. It also keeps `Scan` unchanged.

An existing line that GAINS a marker is a rewrite, so git calls it an addition
and it stays in scope -- which matters, because that is precisely the headroom
move this report exists to expose.

## The header now counts what it prints

Renamed from "N line(s) newly exempted" to "N exempt line(s) written by this
change", because those are different quantities and the second is the one on
screen. Editing an existing exemption's REASON does not newly exempt anything,
but it is exactly how a true reason becomes a false one, so the line has to be
read. Counting the tally rise would have reintroduced a smaller version of the
same contradiction.

## One hunk in here is not mine, and is worth keeping

`ruff format` rewrapped the `subprocess.run(...)` arguments in `_git`. That line
was already unformatted on `main` -- **CI never format-checks `scripts/`**, only
`core-api/src`, `core-storage-api`, `core-operations`, `core-worker` and three
named files under `common/`. So the canonical copy of this gate had drifted from
the project's own formatter without anything noticing.

Keeping the rewrap rather than reverting it, because it removes a real tax:
caura-enterprise's copy is *already* formatted this way, so that one hunk is the
entire difference between the two files today, and every re-vendor has to
reformat around it. After this and the three port PRs land, all four copies are
byte-identical for the first time -- which is the property the whole
copy-don't-adapt model rests on.

## Verified

- the issue's reproduction, run against both versions: 3 lines before, 1 after
- 42 tests pass; **the two new tests fail against the unfixed script**, and the
  first assertion to trip is the substantive one (a pre-existing reason absent
  from the output), not the header wording
- two existing tests updated for the new header string -- assertion substance
  unchanged, only the literal
- `ruff check` and `ruff format --check` clean

## Not fixed here, and worth its own issue

Filed as #900: the report only fires for files whose exempt count ROSE, so a
change that removes one exemption and adds another in the same file is net zero
and the new one is never printed. The gate still catches the dangerous version
-- removing a marker raises the non-exempt count and `_minted` charges it -- but
the report is silent where it should not be. Opposite direction, and it changes
which files get scanned rather than which lines get printed.

Signed-off-by: eldad-caura <eldad@caura.ai>## Review round 2: the fallback note no longer names a cause

It said `(a diff could not be read — ...)`. `_added_lines` returning empty is
the likely reason, but a diff that read fine and simply did not intersect the
file's exempt lines lands in the same branch, and nothing at that point can tell
them apart. Now reports the observation and names the files:
`(no diff line matched in N file(s): ... — every exempt line they hold is
listed, so some may predate this change)`.

I tested the rename case the review hypothesised. It does **not** reach the
fallback: with the pathspec limited to the new path git cannot pair the rename,
so all three lines report as added, the header says 3 for a delta of 3, and the
output is correct — every one of those lines *was* written to that path by the
change. The wording still needed fixing, because I can show that one candidate
cause behaves well, not that the set of causes is empty.

### DIFF
--- a/scripts/legacy_name_ratchet.py
+++ b/scripts/legacy_name_ratchet.py
@@ -138,7 +138,12 @@ def _git(args: list[str]) -> str:
     # counts occurrences of an ASCII name, so a mangled character elsewhere on
     # the line changes nothing about the match.
     proc = subprocess.run(
-        args, capture_output=True, text=True, check=False, encoding="utf-8", errors="replace"
+        args,
+        capture_output=True,
+        text=True,
+        check=False,
+        encoding="utf-8",
+        errors="replace",
     )
     if proc.returncode == 0:
         return proc.stdout
@@ -411,8 +416,10 @@ def _report_excused_moves(
     print()
 
 
-def _report_new_exemptions(before: Counter[str], after: Counter[str]) -> None:
-    """Name every line this change newly exempted, whether or not it fails.
+def _report_new_exemptions(
+    before: Counter[str], after: Counter[str], base: str
+) -> None:
+    """Name every exempt line this change wrote, whether or not it fails.
 
     The one thing a count-based ratchet cannot decide for itself, and the reason
     this report exists. Marking an existing line exempt lowers a file's
@@ -432,18 +439,21 @@ def _report_new_exemptions(before: Counter[str], after: Counter[str]) -> None:
     It is still printed, on the passing path as well. An exemption is the move
     that buys headroom, and a deliberate one — an alias landing in the same PR as
     unrelated work — should never be made quietly even when it is legitimate.
+
+    Which is why the list is filtered to the lines this change actually wrote,
+    rather than every exempt line the file happens to hold. A file's exempt count
+    is what selects it; the diff is what selects the lines. Re-greping and
+    printing all of them puts a file's whole accumulated pile under a header
+    counting one, sends the reader to lines that are not in the diff, and reads
+    as self-contradictory — which invites distrusting the total rather than the
+    list. Phase 5's dual-read wave is exactly the change that builds those piles,
+    so every later PR touching those files would reprint them. A reader who
+    learns the list is mostly noise stops reading it, and that is the failure
+    this report exists to prevent.
     """
     added = {p: n - before.get(p, 0) for p, n in after.items() if n > before.get(p, 0)}
     if not added:
         return
-
-    total = sum(added.values())
-    print(
-        f"{total} line(s) newly exempted in {len(added)} file(s) — check each one is a"
-    )
-    print(
-        "compat alias, a redirect or a pinned wire format, and not headroom for a new name:"
-    )
 
     # Grouped by reason, because the wall is the failure mode. A dual-read wave
     # exempts dozens of lines carrying one identical reason, and at that length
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@erni-a pleae approve

erni-a (dismissed):

Eldad-Caura:
**Not taking the cache — the premise inverts, and there is a better version of the observation underneath it.**

The duplication is real: for a path in *both* the exemption set and the minted set, `_grep` and `_added_lines` each run twice. But the scenario given for why it matters is the one where it cannot happen.

A path is only in both sets if it gains an exemption **and** mints a name the repo never had. A large dual-read wave that passes mints nothing — `grown` is empty, so the overlap is exactly zero and the cache saves nothing. The overlap is non-empty only on a **failing** PR, and specifically on the exempt-and-add swap, where the file count is small and the run is about to exit 1 anyway. So the cost lands where nobody is waiting on it, and is absent where the review says it would add up.

**The redundancy worth naming is a different one.** `scan(None)` at the top of `main()` already reads every line both reports later re-`grep` for — `_grep` returns `(path, lineno, text, exempt)` and `scan` throws the line number away:

```python
for path, _, text, is_exempt in _grep(tree):
```

Both reports then re-grep per path purely to get that discarded number back. Caching the per-path calls would paper over that; having `Scan` retain line numbers removes it, and drops the `_grep` count from `2 + len(paths_of_interest)` to `2` flat. That is a real simplification, it is the same fix for both reports, and it is a change to `Scan`'s shape rather than to this PR's loop body — so it wants its own PR, and I have not filed one yet because it should be measured first rather than assumed worth the churn.

I did want to confirm the safety claim regardless, since a cache is only "optional cleanup" if it is also correct: both calls read the working tree, which does not move during a run, so caching would have been sound. It is just not buying anything here.

**One thing the review did not flag that I have since called out in the commit body**, in case it helps whoever reads this next: one hunk in this diff is `ruff format` rewrapping `subprocess.run(...)` in `_git`, which is not my change. That line was already unformatted on `main` because **CI never format-checks `scripts/`**. I kept it deliberately — it is the entire remaining difference between this file and caura-enterprise's copy, so keeping it makes all four copies byte-identical once the port PRs land.

Eldad-Caura:
@claude force-pushed — the amend adds the explanation of the `ruff format` hunk to the commit body. No code change since your review.

Eldad-Caura:
**Taken — the note no longer names a cause.** You are right that the branch conflates two things and asserts one of them.

Now:

```
  (no diff line matched in 1 file(s): scripts/foo.py — every exempt line they
   hold is listed, so some may predate this change)
```

Observation, plus the files, plus the consequence for the reader. `_added_lines` returning empty on a git failure is still the likely cause, but a diff that read fine and simply did not intersect the file's exempt lines lands in the same branch — and nothing at that point can distinguish them, so naming one would be a guess printed as a finding. Which is the habit this whole report exists to discourage, so it would have been a bad place to keep it.

**I did test the rename case you hypothesised, and it does not reach the fallback.** Base holds `old.py` with two exempt lines; the change renames it to `new.py` and adds a third:

```
3 exempt line(s) written by this change in 1 file(s) —
  new.py:1: A = "…-one"    # legacy-name-ok: alias one
  new.py:2: B = "…-two"    # legacy-name-ok: alias two
  new.py:3: C = "…-three"  # legacy-name-ok: alias three
```

With the pathspec limited to the new path git cannot pair the rename, so it reports the file as wholly added — and the output is right, not merely lucky: every one of those lines *was* written to that path by this change, and the header's 3 matches the delta of 3.

So the change is on the strength of the argument rather than a demonstrated failure: I can show one candidate cause behaves well, not that the set of causes is empty. That asymmetry is exactly why the message should describe what happened instead of why.

**Not adding a test for the fallback**, and saying so rather than leaving it implied. The only way in is `_added_lines` swallowing a `RuntimeError`, which needs `git diff` to fail on a tree-ish `git grep` already accepted — I could not construct that without monkeypatching the function, at which point the test pins the mock. The pre-existing fallback in the offender report is untested for the same reason.

Note for whoever merges: **installer#15 and enterprise#1234 already landed** carrying the previous revision of this file, so both are one hunk behind until the re-sync PRs go in.

Eldad-Caura:
@claude force-pushed — the fallback note now reports the observation and names the files, instead of asserting a cause. Details in the comment above.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #902, opened 2026-08-23
fix(plugin): the strings the plugin emits still said memclaw

yanki-m:
W1 of the sunset programme — the plugin-emission item. Only strings the plugin **writes out**: to a console, or into an agent's context. Nothing it reads, matches, or is identified by.

| Changed | |
| --- | --- |
| `[memclaw]` log prefix | **70 of 86** occurrences, 16 files |
| `info.name` | the ContextEngine's display label |
| the recall block | told the model its memories "were retrieved from MemClaw for this session" |

`-72` legacy-brand lines.

## The 16 prefixes left alone — and why the ratchet was right

Each of the remaining `[memclaw]` prefixes sits on a line that **also** names something protected: `MEMCLAW_TENANT_ID`, `MEMCLAW_NODE_NAME`, `MEMCLAW_TOOLS`, or the plugin id `"memclaw"` in `plugins.slots.contextEngine`.

My first pass changed those too. The ratchet failed it — and the failure was correct in a way worth recording. The **count fell** on every one of those lines, and the new text-comparison gate flagged them anyway, because editing part of a line that still carries the old name mints new text around a name W4 is about to change regardless. Touching them now means touching them twice.

So the rule I settled on: **change the prefix only where it is the sole occurrence on the line.** The mixed lines belong to W4, with the env dual-read.

## Deliberately untouched, in the same files

| Location | Why |
| --- | --- |
| `context-engine.ts:505` `info.id = "memclaw"` | the plugin identity — kept by the programme's own decision; flipping it can silently lose memory on straggler nodes |
| `context-engine.ts:488` `MemClawContextEngine` | exported class, i.e. public API. Needs an alias, not a rename |
| `context-engine.ts:127` `/\b(?:MemClaw\|Caura) API 409\b/` | matches error text older versions produced |
| `educate.ts` (25) | regexes and markers matching content old plugin versions wrote to **users' disks** |

The distinction throughout: *a string we print is ours to change; a string we match or are known by is a contract.*

## Verification

- `tsc` clean
- Test subset **unchanged against baseline: 174 pass / 41 fail**. Those 41 fail identically on untouched `main` — POSIX path assumptions on a Windows checkout, pre-existing and unrelated. I ran the baseline first precisely so this diff could be judged against it.

## Worth doing separately

The prefix is inlined at all 86 call sites, including twice inside `logger.ts`'s own helpers. Extracting a `LOG_PREFIX` constant would make the next change to it one line instead of a sweep. That is a refactor and did not belong in a rename commit — but it should happen.


erni-a (approved):
Reviewed: scoped correctly, no overlap with the overnight lane work, sentinel + ratchet green. Approving and merging per Erni.

PR #903, opened 2026-08-23
fix(ratchet): git chooses which files the exemption report examines, not the tally

### LINKED ISSUE #900: ratchet: a net-zero exemption change is never reported
Found while fixing #894, deliberately left out of [#899](https://github.com/caura-ai/caura/pull/899) — opposite direction, and it changes which *files* get scanned rather than which *lines* get printed.

## What happens

`_report_new_exemptions` selects files with:

```python
added = {p: n - before.get(p, 0) for p, n in after.items() if n > before.get(p, 0)}
```

A change that **removes one exemption and adds another in the same file** leaves that file's exempt count flat, so the file is never selected and the new exemption is never printed. The report's stated purpose is that an exemption "should never be made quietly even when it is legitimate" — and this is a way to make one quietly.

## What is and is not at risk

**The gate still catches the dangerous version.** Removing a marker raises the file's non-exempt count, and if that text is new to the repo `_minted` charges it. So this is not a way past the ratchet.

What is lost is review visibility: a PR can land a brand-new `legacy-name-ok` reason that no reviewer is pointed at, provided it drops an unrelated one in the same file. During Phase 5's dual-read wave — files with dozens of exemptions being edited repeatedly — a swap like that is unremarkable in a diff and would not be flagged.

## Shape of a fix

Select on **files the change touched** rather than on a rise in the tally, then let #899's per-line filter do the rest. Roughly: union the current `added` keys with the paths from `git diff --name-only <base>`, and keep the rest as-is — the line filter already prints only what the change wrote, so widening the file set does not widen the output for untouched files.

One thing to preserve: the count in the header must keep matching the list, which after #899 it does by construction (it counts what it prints, not the tally delta).

## Why not now

#899 is a filtering fix inside the loop body; this changes the loop's input. Landing them together would make the diff read as one change when they are two, and the revert story for each is different.

No urgency: the gate is correct either way.

Eldad-Caura:
Closes #900. Follows #899, which fixed the other half — that one narrowed the
LINES; this narrows what selects the FILES.

`_report_new_exemptions` examined only files whose exempt count ROSE. Remove one
exemption and add another in the same file and the count is flat, so the file
was never looked at and the new reason was never printed.

The gate still caught the dangerous form: un-marking a line raises the
non-exempt count and `_minted` charges the text. What was lost is review
visibility — a PR could land a brand-new `legacy-name-ok` reason that nobody is
pointed at, provided it dropped an unrelated one in the same file. In a Phase 5
file already carrying dozens of exemptions, a one-for-one swap is unremarkable
in a diff, and "an exemption should never be made quietly" is the entire reason
this report exists.

## The fix

`_changed_paths()` asks git which files moved; candidates are those intersected
with the files that hold an exempt line at all. The risen-tally set is kept only
as the fallback for when git cannot answer, because a selection that is too
narrow beats one that is arbitrary.

## Widening the file set forced a real change to `_added_lines`

It returned an empty set both when the diff failed and when the diff was fine.
That was survivable while only risen-tally files were examined, because such a
file always had something added. Now every touched file holding a marker is a
candidate, including ones the change only DELETED from, or touched nowhere near
their exempt lines — and for those, empty is the correct answer.

So it returns `None` for "could not read" and an empty set for "added nothing",
and the caller treats them oppositely: `None` lists everything (the documented
nuisance-over-dead-end trade), empty prints nothing. Collapsing them would have
been #894 reintroduced through the file-selection door, and worse — it would have
fired on files whose exempt count never moved at all.

A side effect worth having: the fallback note can go back to saying "diff
unreadable", because that branch is now only reachable when the diff genuinely
failed. #899 had to hedge the wording precisely because the two cases were
indistinguishable.

## Verified

- the issue's scenario, both versions: the new reason is invisible before, named
  after
- 45 tests pass. **`test_a_net_zero_swap_of_exemptions_is_still_reported` fails
  against `main`** — that is the bug
- the other two new tests pass either way, because they guard the NEW
  mechanism's failure mode rather than reproducing the old bug. Both were
  confirmed by breaking the property they guard — reverting `if not picked:
  continue` to `picked = here` fails exactly those two and nothing else
- `ruff check` / `ruff format --check` clean; ratchet and sentinel green on this
  change

## The cost, measured rather than asserted

Selecting on the diff means one `git grep` and one `git diff` per touched file
that holds an exemption. On a synthetic dual-read-shaped repo — 80 files, 12
exemptions each, all 80 touched without altering a single exemption, which is
the worst case — the run goes **0.06s -> 1.46s**.

Acceptable: the gate's own two full-tree scans dominate a real run, and CI's job
takes seconds. Stated because it scales linearly with touched-files-holding-
exemptions, so a 500-file Phase 5.4 PR would be several seconds more.

Two ways to buy it back if that ever matters, neither taken here because
1.4s does not justify the churn: have `Scan` retain line numbers, which removes
every per-path `_grep` (both reports re-grep only to recover the number `scan()`
already read and discarded); or replace the per-path `git diff` with one
whole-repo `-U0` parse.

### DIFF
--- a/scripts/legacy_name_ratchet.py
+++ b/scripts/legacy_name_ratchet.py
@@ -198,7 +198,30 @@ def _grep(tree: str | None, pathspec: str = ":/") -> list[tuple[str, str, str, b
     return out
 
 
-def _added_lines(base: str, path: str) -> set[int]:
+def _changed_paths(base: str) -> set[str] | None:
+    """Every path this change touched, per git. ``None`` if the diff failed.
+
+    What selects the files the exemption report talks about. The exempt tally
+    cannot do it: a change that removes one exemption and adds another in the
+    same file is net zero, so a rise-based filter never looks at that file and
+    the new exemption is never printed — the one thing the report exists to stop
+    happening quietly. Asking git which files moved has no such blind spot.
+
+    ``None`` rather than an empty set on failure, because the two mean opposite
+    things to the caller: "could not tell" must fall back to the old tally-based
+    selection, while "nothing changed" must select nothing.
+    """
+    try:
+        return {
+            line
+            for line in _git(["git", "diff", "--name-only", base]).splitlines()
+            if line
+        }
+    except RuntimeError:
+        return None
+
+
+def _added_lines(base: str, path: str) -> set[int] | None:
     """Line numbers this change actually added to ``path``, per git's own diff.
 
     Used for the failure report only, never for the decision. The decision
@@ -212,13 +235,21 @@ def _added_lines(base: str, path: str) -> set[int]:
     reader is sent to a line that has not changed, for a gate whose whole output
     is "here is the line you added".
 
-    Empty on any failure, and the caller falls back to naming every occurrence of
+    ``None`` on failure, and the caller falls back to naming every occurrence of
     the minted text: too many lines is a nuisance, none is a dead end.
+
+    ``None`` rather than an empty set, because once the exemption report selects
+    files by what git says changed rather than by a risen tally, an empty result
+    is a real and common answer — a file this change only DELETED from adds no
+    lines, and every exempt line it still holds predates the change. Collapsing
+    that into the failure case would reprint the whole pile for exactly the files
+    that deserve none of it, which is the bug #894 fixed, reintroduced through
+    the back door.
     """
     try:
         out = _git(["git", "diff", "-U0", base, "--", _literal(path)])
     except RuntimeError:
-        return set()
+        return None
 
     added: set[int] = set()
     lineno = 0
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@erni-a ready for review — green, no red checks, blocked only on approval.

Closes #900, the follow-up I filed while fixing #894. #899 narrowed which *lines* the exemption report prints; this narrows what selects the *files*. Together they close both halves: a net-zero exemption swap is now reported, and a touched file whose markers all predate the change still says nothing.

Two things worth a reviewer's eye:

- **`_added_lines` now returns `None` vs an empty set**, and the two are treated oppositely. That is forced, not cosmetic — once every touched file is a candidate, "added nothing" is a common correct answer that must print nothing, while "could not read the diff" must still list everything.
- **A measured 0.06s → 1.46s** on a synthetic 80-file worst case, reported in the commit body rather than left to be discovered. Two ways to buy it back are named there; I took neither, because 1.4s does not justify the churn.

erni-a (approved):

PR #904, opened 2026-08-23
fix(plugin): caura_list scope='all' spans fleets instead of narrowing to one

Eldad-Caura:
`enrichBody` in `plugin/src/tool-definitions.ts` fills in `fleet_id` from the configured fleet whenever the caller omits it. That default is deliberate — it is what makes ordinary calls fleet-scoped without every agent having to say so. But `scope: 'all'` is a caller explicitly asking to span every fleet, and the default was overriding it.

#886 gave this code path `CAURA_FLEET_ID` for free and deliberately did not touch the narrowing, because fixing it is behavioural and #886 was additive-only. This is that fix.

## The defect

On the two read-enumeration tools the server applies `fleet_id` as a row filter **outside any scope branch** — `memory_list_by_filters` and `memory_stats_breakdown` both do `if fleet_id: Memory.fleet_id == fleet_id` — while their shared ladder `resolve_read_fleet_gate` passes `fleet_id` through untouched for `'all'`.

So a tenant-wide read came back narrowed to a single fleet. Because the filter is a strict equality it also dropped every fleet-less (`fleet_id IS NULL`) row. Nothing in the response said so: the call succeeded and simply answered with less.

## Behaviour matrix, measured on both sides

With a configured fleet `F` and no caller-supplied `fleet_id`. Only the two bold cells move.

| tool | `scope` | before | after |
|---|---|---|---|
| `caura_list` | omitted | `fleet_id=F` | `fleet_id=F` |
| `caura_list` | `'agent'` | `fleet_id=F` | `fleet_id=F` |
| `caura_list` | `'fleet'` | `fleet_id=F` | `fleet_id=F` |
| `caura_list` | `'all'` | **`fleet_id=F`** | **omitted** |
| `caura_stats` | omitted | `fleet_id=F` | `fleet_id=F` |
| `caura_stats` | `'agent'` | `fleet_id=F` | `fleet_id=F` |
| `caura_stats` | `'fleet'` | `fleet_id=F` | `fleet_id=F` |
| `caura_stats` | `'all'` | **`fleet_id=F`** | **omitted** |
| `caura_insights` | any | `fleet_id=F` | `fleet_id=F` |
| `caura_evolve` | any | `fleet_id=F` | `fleet_id=F` |

A caller-supplied `fleet_id` is byte-identical in every row, at every scope: this withholds a default, it does not strip a value.

Two deliberate non-changes worth naming:

- **`scope='agent'` keeps the default.** It also narrows — an agent listing its own memories misses ones it wrote in another fleet, and all fleet-less ones. But `'agent'` does not contradict a fleet-scoped default the way `'all'` does, so it is left as-is rather than folded into a bug fix.
- **`scope='fleet'` keeps the default.** When `F` is the caller's home fleet this is exactly the pin `resolve_read_fleet_gate` would apply anyway; when it is not, the server raises the bar to trust 2, which is the correct gate for a cross-fleet read.

## Why only two of the four tools that accept scope='all'

`caura_insights` and `caura_evolve` deliberately do **not** opt in:

1. Their reads branch on `scope` and ignore `fleet_id` entirely under `'all'` (`_insights_scope_filters`, `evolve_service._filter_by_scope`), so there is no read to widen.
2. Their `fleet_id` doubles as a **write target** — the fleet persisted findings and outcome rules land in, and the key insights supersedes priors under. Withholding it there would relocate a write and strand the priors it was meant to supersede.

I nearly widened the fix to `caura_insights` on the grounds that it looked like the same shape; the `BulkMemoryCreate(fleet_id=...)` in `insights_service` is why that would have been wrong.

## Did any caller depend on the narrowing?

No.

- No in-repo caller: `caura_list` / `caura_stats` are agent-facing tools with no internal call site in `plugin/src`.
- `scripts/trust_matrix_e2e.py` is the only script exercising `scope='all'`, and it drives core-api directly over HTTP with an `X-API-Key` — the plugin is not in its path.
- The server-side MCP `caura_list` has no environment-driven fleet injection at all, so there is no sibling copy of this defect left unowned.

## Tests

`plugin/src/fleet-scope.test.ts` asserts on the **query string actually put on the wire**. An assertion that the tool "returns results" would pass either way, since the narrowed call succeeds — that is the trap this bug sets for its own test.

- The two `scope='all'` tests **fail without the fix**, with `'fleet-from-env'` where `null` is expected.
- The five guard tests were verified load-bearing by breaking what they guard: removing the default outright fails all five; applying the fix without the per-tool opt-in fails exactly the insights/evolve one and nothing else.

It needs its own file because the env vars are read at `env.ts` import time, so they must be set before the module is imported.

Gates: 413/413 plugin tests, `tsc` clean, legacy-name ratchet "No new lines", sentinel "All 23 protected strings survive". `env.ts` and the env-reading paths are untouched.

### DIFF
--- a/plugin/package.json
+++ b/plugin/package.json
@@ -16,7 +16,7 @@
     "pretest": "bash ../scripts/gen-version.sh",
     "build": "tsc",
     "dev": "bash ../scripts/gen-version.sh && tsc --watch",
-    "test": "tsc && NODE_ENV=test node --test dist/educate.test.js dist/tool-definitions.test.js dist/env.test.js dist/transport.test.js dist/health.test.js dist/config.test.js dist/runtime-contract.test.js dist/identity.test.js dist/install-id.test.js dist/validation.test.js dist/reconcile-skills.test.js dist/context-engine.test.js dist/heartbeat.test.js dist/interview-buffer.test.js dist/interview-command.test.js dist/task-trail.test.js dist/keystones.test.js dist/resolve-agent.test.js dist/bootstrap-memoization.test.js",
+    "test": "tsc && NODE_ENV=test node --test dist/educate.test.js dist/tool-definitions.test.js dist/env.test.js dist/transport.test.js dist/health.test.js dist/config.test.js dist/runtime-contract.test.js dist/identity.test.js dist/install-id.test.js dist/validation.test.js dist/reconcile-skills.test.js dist/context-engine.test.js dist/heartbeat.test.js dist/interview-buffer.test.js dist/interview-command.test.js dist/task-trail.test.js dist/keystones.test.js dist/resolve-agent.test.js dist/bootstrap-memoization.test.js dist/fleet-scope.test.js",
     "check:version": "bash -c 'V=$(node -p \"require(\\\"./package.json\\\").version\"); grep -qF \"PLUGIN_VERSION = \\\"$V\\\"\" src/version.ts || (echo \"version.ts out of sync with package.json ($V): run bash ../scripts/gen-version.sh\" >&2 && exit 1) && echo \"check:version ok: $V\"'"
   },
   "devDependencies": {
--- a/plugin/src/fleet-scope.test.ts
+++ b/plugin/src/fleet-scope.test.ts
@@ -1,0 +1,176 @@
+/**
+ * The configured-fleet default is a DEFAULT, not an override.
+ *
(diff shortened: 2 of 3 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #905, opened 2026-08-23
docs: fix four old-brand references that name things which do not exist

Eldad-Caura:
Lane 3 (Phase 5.4 first slice): the docs that are actually served — `static/docs/`, `README.md`, `docs/`.

**The rule applied:** fix prose that is factually **wrong**; defer prose that is merely **incomplete**. ~110 old-brand occurrences examined across 26 files; **4 were wrong**, in 4 files. The rest name something that genuinely still answers to that name and are left for 5.4's wholesale rewrite.

### Fixed

| File | Was | Why it's wrong |
|---|---|---|
| `docs/plugin-upgrade.md` | `grep '^MEMCLAW_API_URL='` ×5 | Since #895 the installer writes a `.env` of only `CAURA_*`. On any current install all five greps return empty and the re-install POSTs blank identity — silently wiping the per-node identity the section exists to preserve. |
| `static/docs/integration-guide.md` | "only `MEMCLAW_*` keys are read" | False since #886 — `isPluginEnvKey` is `/^(?:CAURA\|MEMCLAW)_[A-Z_]+$/`. Reads as a contradiction next to the all-`CAURA_*` `.env` the installer just wrote. |
| `docs/operator-forge-cron.md` | `X-Memclaw-Admin-Token: ${MEMCLAW_ADMIN_TOKEN}` ×2 | Neither header nor variable exists anywhere in the tree. `enforce_admin` tests `is_admin` off the credential from `X-API-Key`. Both examples 401 on every tick. |
| `README.md` | `X-Memclaw-Gateway` | Has never existed in the codebase in any spelling since the initial public release. Real one is `X-Gateway-Secret`. |

The last one is **replaced, not renamed** — minting `X-Caura-Gateway` would move the phantom rather than remove it. The `plugin-upgrade` fix accepts **both** prefixes rather than swapping one for the other, because `deploy.ts` preserves whichever a node's `.env` already had, so both are live in the field. It takes the first **non-empty**, not the first defined: a hand-edited `.env` can carry a present-but-blank key (the §09 empty-string trap), and blank is exactly the failure this snippet cannot afford.

### Verification

The snippet was extracted **verbatim from the rendered doc** and run against fixtures:

```
new-install.env  (CAURA_* only, as the installer writes)  OLD=[]              NEW=[mc_realkey]
legacy.env       (MEMCLAW_* only, pre-rename)             OLD=[mc_legacykey]  NEW=[mc_legacykey]
halffilled.env   (CAURA_ blank + real MEMCLAW_)           NEW=[mc_theonlyrealkey]
```

Empty under the old form, correct under the new, no regression on a pre-rename `.env`. Also `bash -n` clean, and both YAML examples still parse.

Gates: ratchet **no new lines, −9 net** (two dual-read lines carry `legacy-name-ok`); sentinel **all 23 protected strings survive**.

### Deliberately deferred

Every one still names something real, verified against the tree — not skipped for time:

- **`MEMCLAW_*` env family** (guide `.env` block, README table, `MEMCLAW_VERSION`, `MEMCLAW_RUN_DESTRUCTIVE_MIGRATIONS`, `MEMCLAW_INTERVIEWER`, `MEMCLAW_SKILL_TARGETS`, `MEMCLAW_API_KEY`) — all confirmed dual-read; incomplete, and 5.4's to rewrite.
- **Plugin id `memclaw`** — `index.ts:131`, `context-engine.ts:505`. So `plugins.allow`/`entries`/`slots`, `openclaw plugins enable memclaw` are all correct.
- **Gateway RPC `memclaw.status|deploy|educate|allowlist.*`** — all registered in `index.ts`.
- **`[memclaw]` log prefix** — still what the plugin emits.
- **`~/.openclaw/plugins/memclaw`, `~/.claude/skills/memclaw/`, `.memclaw-owned`** — real paths (`paths.ts`, `plugin.py:731/739`, `OWNED_MARKER`). Floor-class: changing them breaks live installs.
- **`/api/v1/memclaw/keystones`** — the live endpoint path.
- **`memclaw.auto_upgrade_enabled`** — real settings key (`organization_settings.py:580`).
- **GHCR `caura-memclaw-core-api` / `-core-storage-api`** — match `docker-compose.yml` exactly. Phase 6, gated on 1.7.
- **`POSTGRES_USER`/`_DB` = `memclaw`, `pg_dump -U memclaw`** — sentinel-protected floor.
- **`memclawd` / `memclaw` CLI / `memclawctl`** — real binaries and harnesses.
- **`memclaw.<domain>.<verb>` events** (`component-registry.yaml`, `operator-forge-cron.md:15`) — real topics, and Lane 5's to flip. Out of this lane.
- **`docs/performance.md:15,72` blog links** — re-probed today: `/blog/memclaw-benchmarks` **and** `/blog/caura-benchmarks` both return 200. The link works, so it is incomplete, not wrong. (`BENCHMARKS.md:165` is the third hit in that unowned item and sits outside this lane.)
- **`memclaw-client` package / `memclaw-interviewer` CLI** (README:584-585) — `pyproject.toml` keeps `memclaw-interviewer` as a permanent console script and ships a `memclaw_client` import shim. Stale wording, not false. Flagged below.
- **`docs/live-memory-pitch/*`, `beacon-…-2026-07-26.md`, `docs/plans/`** — dated internal planning and incident records. Historical, not instructions.

### Worth a second pair of eyes

`README.md:585` says the interviewer CLI is "shipped in the `memclaw-client` package", but the distribution is now `caura-client`; `memclaw-client` on PyPI is frozen at its pre-rename release. The console script and import shim both survive, so nothing is strictly false — but a reader who runs `pip install memclaw-client` gets a frozen build, not the current one. Left alone as 5.4 scope rather than widened here.

Boundaries respected: no edits to `plugin/src` (Lane 4), `common/events/` (Lane 5), `scripts/` or `.github/`. Read-only there to verify what the docs claim.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1192,7 +1192,7 @@ Anything not listed above is internal and may change in any release without a ma
 
 - Python module layout (`core_api.middleware.*`, `core_api.providers.*`, `core_api.pipeline.*`, `core_api.services.*`, `common/*`)
 - Database schema, table names, migration paths
-- Gateway-injected HTTP headers (`X-Memclaw-Gateway`, `X-Tenant-ID`, `X-Agent-ID`, `X-Org-Read-Only`)
+- Gateway-injected HTTP headers (`X-Gateway-Secret`, `X-Tenant-ID`, `X-Agent-ID`, `X-Org-Read-Only`)
 - Most `/api/v1/admin/*` and all `/api/v1/testing/*` routes (the documented exception is `POST /admin/agent-keys/provision`, which is part of the stable identity-bootstrap surface — see the Agents row above)
 - The `core-storage-api` microservice (internal, not user-facing)
 - The plugin's TypeScript module structure
--- a/docs/operator-forge-cron.md
+++ b/docs/operator-forge-cron.md
@@ -42,7 +42,7 @@ http_target:
   oidc_token:
     service_account_email: <core-operations-sa>@<project>.iam.gserviceaccount.com
   headers:
-    X-Memclaw-Admin-Token: ${MEMCLAW_ADMIN_TOKEN}   # see core-api/auth.enforce_admin
+    X-API-Key: ${ADMIN_API_KEY}   # the admin key — see core-api/auth.enforce_admin
 ```
 
 ### Kubernetes CronJob (alternative)
@@ -67,9 +67,10 @@ spec:
                 - |
                   curl -fsS \
                     -X POST \
-                    -H "X-Memclaw-Admin-Token: $MEMCLAW_ADMIN_TOKEN" \
+                    -H "X-API-Key: $ADMIN_API_KEY" \
                     "$CORE_API_BASE_URL/admin/lifecycle/fanout/forge-distill"
               envFrom:
+                # whichever secret you use, it must expose ADMIN_API_KEY
                 - secretRef: { name: memclaw-admin }
           restartPolicy: OnFailure
 ```
--- a/docs/plugin-upgrade.md
+++ b/docs/plugin-upgrade.md
@@ -58,11 +58,24 @@ When upgrading an existing install, read its current `.env` first and pass the v
 
 ```bash
 ENV=$HOME/.openclaw/plugins/memclaw/.env
-URL=$(grep    '^MEMCLAW_API_URL='    "$ENV" | cut -d= -f2-)
-KEY=$(grep    '^MEMCLAW_API_KEY='    "$ENV" | cut -d= -f2-)
-FLEET=$(grep  '^MEMCLAW_FLEET_ID='   "$ENV" | cut -d= -f2-)
-TENANT=$(grep '^MEMCLAW_TENANT_ID='  "$ENV" | cut -d= -f2-)
-NODE=$(grep   '^MEMCLAW_NODE_NAME='  "$ENV" | cut -d= -f2-)
+
+# Read one setting, accepting either prefix: the installer writes CAURA_* into
+# new installs, older installs kept the pre-rename prefix, and a re-deploy
+# preserves whichever the file already had. First NON-EMPTY wins — a key can be
+# present but blank in a hand-edited .env, and blank here silently re-installs
+# the node with no identity at all.
+read_env() {
+  for _p in CAURA MEMCLAW; do  # legacy-name-ok: dual-prefix read, both are live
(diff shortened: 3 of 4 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

Eldad-Caura:

Two things worth knowing before you look:

**No Claude review ran, by design.** `pre-checks` skipped it with *"no source files to review (only tests/docs/config)"* — a docs-only PR doesn't burn review budget. I did not override that with a manual `@claude`, since it's a deliberate cost policy and widening it isn't mine to call. But this PR is not purely prose: `docs/plugin-upgrade.md` gains a shell function that operators copy-paste and run against live gateways. If you'd rather it got the automated pass, commenting `@claude` takes the retrigger path, which bypasses the docs-only skip.

**The substance is four wrong references, not a rename sweep.** ~110 old-brand occurrences were examined; 4 were factually wrong and the other ~106 name things that genuinely still answer to that name (plugin id, RPC methods, log prefix, install paths, the dual-reading env family, the live keystones endpoint). Those are deliberately left for Phase 5.4. The deferred list with the reason each one is still true is in the PR description.

The one I'd read closest is `docs/plugin-upgrade.md`: the old snippet recovered node identity with `grep '^MEMCLAW_API_URL='`, and since #895 the installer writes a `.env` of only `CAURA_*` — so on any current install it returned empty and the re-install silently POSTed blank identity. Proven by running the doc's snippet verbatim against a fixture `.env`: empty under the old form, correct under the new, no regression on a pre-rename file.

erni-a (approved):

Eldad-Caura:
Correction to a number in the description above, recorded rather than quietly edited.

I wrote "~110 old-brand occurrences examined". The measured figure at the merge base `c3ec9aab` is **169 matching lines / 191 raw occurrences across 26 files** in `static/docs/`, `README.md` and `docs/`. My estimate was low; the survey itself covered all 26 files, so the 4-wrong finding and the deferred list are unaffected.

Related, and more useful to whoever picks up 5.4: the handover put `static/docs/integration-guide.md` at **11 hits**. It is **34 matching lines** at the same commit. That is the §00 "treat every count as a floor and go enumerate" failure mode, on the single file the handover names as the highest-value one in the slice. Anyone scoping 5.4 from those numbers should re-measure first — the file is roughly 3× the size the plan assumes.

PR #910, opened 2026-08-23
docs(mcp): scope has no single default, so the SoT descriptions stop naming one

### LINKED ISSUE #908: SoT tool descriptions call scope='agent' the default, which is false on the plugin surface (now disagrees with the param schema)
Follow-up to #906, which fixed the plugin-side half. The registry side is still wrong, and the two now contradict each other in the same tool card.

## The claim

`core-api/src/core_api/tools/caura_list.py:17`

> `scope='agent'` (default) lists your memories at trust ≥ 1

`core-api/src/core_api/tools/caura_stats.py:16`

> `scope='agent'` (default) counts only memories visible to YOU (trust ≥ 1)

## Why it is false on the plugin surface

One string, two surfaces, different defaults. Per `tools/_types.py:51` the spec drives "`/tool-descriptions` output, plugin `tools.json`, and per-tool trust gating" — so this single description is served to both the MCP server and the OpenClaw plugin. Only one of them defaults `scope`:

| surface | how `scope` is defaulted | is "'agent' (default)" true? |
|---|---|---|
| MCP `caura_list` / `caura_stats` (`mcp_server.py:2207`, `:2424`) | `Annotated[...] = "agent"` — real Python default | **yes** |
| plugin `caura_list` / `caura_stats` | no schema default; omitted → param absent from the query string | **no** |

On the plugin path the request reaches `GET /memories` (or `/memories/stats`), where `scope` is an opt-in query param (`Query(default=None)`) and the route only consults the ladder when it was supplied:

```python
author_filter = written_by if written_by is not None else agent_id
if scope is not None:
    author_filter, fleet_id = await _resolve_scoped_read(...)
```

So omitting `scope` from the plugin is a **different request** from `scope='agent'`. It:

- skips the trust-1 gate entirely (`_resolve_scoped_read` → `require_trust` never runs);
- takes its author filter from the `agent_id` **query param** rather than the authenticated credential (`auth.agent_id or agent_id`);
- serves a tenant-wide team/org read when no agent identity is available, where `scope='agent'` 400s (`"scope='agent' requires an agent identity"`).

They do coincide in the common case, because the plugin already fills in `agent_id` from the configured identity. That is what made this survive — it is only visibly wrong at the edges.

## Why it matters now

#906 corrected the plugin's **param** description to say what omitting actually does. It did not touch the registry, so an agent using the plugin now sees both of these at once, for the same parameter:

- tool description (from this registry): *"scope='agent' (default) lists your memories at trust ≥ 1"*
- param schema (`plugin/src/tool-definitions.ts`): *"Optional. Omitted: filtered by agent_id if one is set, with no trust gate — not the same request as 'agent'."*

That is a worse resting state than before. It was out of Lane 4's scope (`plugin/src` only), hence this issue rather than a PR.

## Suggested fix

Reword so the sentence is true on **both** surfaces — describe the trust ladder per value and stop naming a default, since the default is a property of the surface, not of the tool:

> `scope='agent'` lists your memories at trust ≥ 1; `scope='fleet'` reads your OWN fleet at trust ≥ 1, a different fleet at trust ≥ 2; `scope='all'` (tenant-wide) requires trust ≥ 2. Omitting `scope` over REST/plugin filters by `agent_id` without the trust gate; the MCP tool defaults it to `'agent'`.

`plugin/tools.json` is **generated** — `scripts/export_tool_specs.py` dumps the SoT `REGISTRY` and CI fails on drift — so this is one edit in `tools/caura_list.py` + `tools/caura_stats.py` plus a regenerate, not two hand-syncs.

## Do NOT "fix" these — they are correct as written

- `mcp_server.py` `caura_list` / `caura_stats` param descriptions and docstrings: those params really do default to `"agent"`.
- `caura_insights` / `caura_evolve` anywhere: their `scope` travels in a JSON body whose Pydantic model really is `Field(default="agent")`, so calling `'agent'` the default is true on every surface for those two.

The asymmetry is the whole finding: the claim holds wherever `scope` has a binding default, and fails only on the two query-param routes where it does not.

## Adjacent, already fixed in #906, for reference

`caura_list.written_by` claimed it was "ignored when scope='agent'". It is rejected, not ignored — `_resolve_scoped_read` 400s on a `written_by` naming anyone but the caller. The plugin schema is corrected; the registry does not carry that claim.

Eldad-Caura:
Closes #908. Follows #906, which fixed the plugin's param schema and left the registry side contradicting it.

## The problem

`caura_list` and `caura_stats` described `scope='agent'` as "(default)". One description string is served to **both** surfaces — per `tools/_types.py` the spec feeds `/tool-descriptions` *and* the generated `plugin/tools.json` — and only one of them defaults `scope`:

| surface | how `scope` is defaulted | claim true? |
|---|---|---|
| MCP `caura_list` / `caura_stats` | `Annotated[...] = "agent"` — real Python default | yes |
| REST / plugin `GET /memories`, `/memories/stats` | `Query(default=None)`; omitted → absent from the query string | **no** |

On the REST path the route consults the ladder only when `scope` was supplied:

```python
author_filter = written_by if written_by is not None else agent_id
if scope is not None:
    author_filter, fleet_id = await _resolve_scoped_read(...)
```

So an omitted `scope` takes its author filter from `written_by ?? agent_id` (list) or aggregates over `agent_id` (stats), and skips `require_trust` entirely. It coincides with `scope='agent'` in the common case — the plugin fills in `agent_id` from the configured identity — which is why it survived; it diverges only at the edges (no trust gate, author from the param rather than the credential, and a tenant-wide read where `'agent'` would 400).

Since #906 corrected the plugin's **param** description, an agent on the plugin path was seeing the tool description and the param schema disagree about the same parameter. This closes that.

## The fix

Both descriptions now state the trust ladder per **value** and name the per-surface default separately, which is true everywhere the string is served.

## Deliberately not touched — already correct

- `mcp_server.py`'s own param descriptions and docstrings for these two tools: those parameters really do `= "agent"`.
- `caura_insights` / `caura_evolve` anywhere: their `scope` arrives in a request body whose model really is `Field(default="agent")`, so the claim holds on every surface there.

The asymmetry is the whole finding — the claim is true wherever `scope` has a binding default, and false only on the two query-param routes.

## Regenerated artifacts

Per `tests/fixtures/README.md`: `plugin/tools.json` via `scripts/export_tool_specs.py`, and the three fixtures under `tests/fixtures/` via `capture_baselines.py`. **Five** drift gates fail without the regenerate (`test_tools_export_in_sync`, both `tool_descriptions_*_matches_baseline`, `test_mcp_tools_list_content_matches_baseline`, `test_v1_baseline_matches_live_registry`) — verified by running them before regenerating, so the generated files here are mechanical, not hand-edited.

**Token budget:** `tools/list` goes 4928 → 4968 cl100k against `CEILING_TOKENS = 5200`. No bump needed; 232 tokens of headroom remain. Tool-surface tokens are paid on every agent call, so the wording is one clause per tool rather than a full explanation of the split — that lives in a source comment beside each `_DESCRIPTION`.

## Test

`test_query_param_scope_tools_do_not_claim_a_default` asserts neither description contains "(default)", mirroring the plugin-side guard from #906 so the two surfaces cannot silently drift apart again. Confirmed load-bearing by restoring the old wording on `caura_stats` — fails with that exact message.

Gates: 24/24 on the four affected test files, plugin 414/414 (three consecutive runs), `ruff check core-api/src/` and `ruff format --check core-api/src/` clean (run separately), ratchet "No new lines", sentinel "All 23 protected strings survive". Rebased onto `14b7e0cb`.

Three failures in `test_gemini_provider.py` / `test_provider_protocols.py` are **pre-existing** — reproduced on a clean detached `origin/main` worktree, where they fail worse (missing optional deps in that Python). Unrelated to this change.


Eldad-Caura:
@erni-a ready for approval — all checks green, Claude review clean (no issues found). Closes #908. Description wording plus mechanically regenerated artifacts (tools.json + three fixtures) and one guard test; tools/list token count 4928 → 4968 against the 5200 ceiling, no bump needed.

erni-a (approved):

PR #911, opened 2026-08-23
feat(events): bind both topic names so a publisher flip stays lossless

Eldad-Caura:
Step 2 of 6 in the Pub/Sub rename wave. **A runtime no-op by construction** — publishing is byte-for-byte unchanged, and the new subscribe behaviour is off unless an environment opts in.

> **Merge order matters: [caura-enterprise#1242](https://github.com/caura-ai/caura-enterprise/pull/1242) (step 1, EXPAND) must merge first.** This PR's regenerated manifest makes the enterprise cross-repo check require the twin subscriptions. Verified both ways: it passes against the expanded Terraform and fails with 13 precise errors against the unexpanded one. The enterprise check only triggers on `terraform/**` changes, so nothing enforces this order automatically — it is held by hand.

## The design decision

`topics.py` is a `StrEnum` whose members *are* their string values; equality, dict-key hashing, f-string formatting and `topic_path` building all depend on that. Introducing a publish name and a subscribe set therefore needed a decision, not a search-and-replace. Three shapes were possible:

| | |
|---|---|
| **A second enum family** mirroring the first | Duplicates 18 members across two repos by hand, drifts, and doubles what the contract step has to delete. Rejected. |
| **Thread a subscribe set through all 14 call sites** | A diff every future `subscribe()` has to remember to join. The failure it invites is exactly one consumer nobody updated — the undercount that falls outside every lane's scope. Rejected. |
| **Resolve both in the bus** ✅ | Cannot miss a call site; a `subscribe()` added next week inherits it; the contract step is a deletion in one file. |

The enum members are untouched. Two derived functions sit beside them:

- **`publish_name()` returns ONE name, always.** Dual-publishing is the version of this cutover that delivers every event twice to every subscriber holding both names — which, after this step, is all of them. There is deliberately no API here that can express it.
- **`subscribe_names()` returns one name or two.**

Resolving the publish name in the bus too means flipping a family later is one line in `topics.py`, rather than a sweep across every publisher at the riskiest moment of the cutover.

## The two backends need opposite defaults

This is the substantive finding, and it is the §09 question — *which way does the default point* — asked twice with different answers.

**Pub/Sub: OFF by default.** A topic name here is a provisioned subscription. Binding a twin that does not exist yet is not a soft failure — `_pull_loop` treats `NotFound` as permanent, halts, and `is_healthy` turns `/readyz` red. That is the May staging outage precisely. With the flag off the subscription set is byte-identical to before, which is what makes this mergeable and deployable everywhere ahead of any environment's apply.

`EVENT_BUS_DUAL_SUBSCRIBE` demands an explicit affirmative — unset, empty, whitespace and typos all mean off. The realistic failure is a deploy template that has started listing the variable with nothing filled in yet; read as "yes" that 503s every consumer in an unexpanded environment, read as "no" it means the cutover has not started, which step 3 verifies per service anyway. Only the second is recoverable by doing nothing.

**In-process: ON always, no flag.** A name here is a dict key — it costs nothing and cannot fail. Leaving this backend single-name would mean a flipped family silently dispatches to nobody in exactly the standalone and on-prem deployments that never run the Terraform the flag is gated on.

## The broadcast flag has to carry to the twin

A broadcast topic has no durable subscription by design. A twin that is bound but left out of `_broadcast_topics` is treated as an ordinary work queue at `start()`, opening a pull loop against a `<prefix>--<twin>` subscription that was never provisioned and never will be — a red health endpoint on the one topic whose entire design is that it degrades quietly. Covered by `test_broadcast_flag_carries_to_the_twin`.

## One existing assertion had to change

`test_events_manifest_is_well_formed` asserted every topic starts with the outgoing prefix. That is the assertion this rename had to come and edit — and it would have gone quietly vacuous the moment the prefix it names stopped appearing, which is the trap §09 describes. It now checks **shape** plus **closure under the rename**: every consumed topic's twin is listed too. That property means the same thing on both sides of the cutover and after it, and it is what turns "a service may subscribe to the twin" into "the twin subscription must already exist".

## Verification

- 29 new tests, plus the full events suite: **103 passed**, against a baseline of 74 on unmodified `main`. The 12 failures are pre-existing in this local env (`No module named 'google'`) and identical in number and name on `origin/main`.
- Publishing unchanged, asserted directly (`test_no_family_is_flipped_yet`, `test_publish_name_is_the_identity_while_nothing_is_flipped`), so a later flip fails on the one line that says the cutover has not started.
- The payoff property has both halves: a flipped family reaches a dual-bound subscriber, **and** the counterfactual — the same flip with a single-name binding reaches nobody and raises nothing — so the first test cannot pass for the wrong reason.
- The manifest closure assertion was confirmed to fail with the generator's twin line removed.
- `ruff check` / `ruff format --check` at CI's exact scopes, run separately. No lockfile touched.
- Legacy-name ratchet: `No new lines. 1 removed by this change (-1 net).`

## Not in this PR

No publisher is flipped. `FLIPPED_FAMILIES` is empty. Flipping is a separate dispatch gated on observable state — one family at a time, ordered by blast radius, **`audit.*` last**, because those rows are hash-chained and a lost or reordered audit event is the one failure this programme cannot undo.

The enterprise copy of `common/events/` is `manual` policy in `vendored_files_manifest.json` — the re-vendor bot never touches it — so it needs its own PR rather than arriving by vendoring.


Eldad-Caura:
@erni-a please approve

erni-a (approved):

Eldad-Caura:
Drive-by from Lane 4 (#904 / #906 / #910), on the step-4 boundary. Not a defect in this PR — the ordering is documented on `FLIPPED_FAMILIES` and the failure is already proven by `test_a_flipped_family_reaches_nobody_without_the_dual_binding`. One narrow gap I could not talk myself out of:

**The two halves of the invariant live at different scopes.** `FLIPPED_FAMILIES` is module state — flipping a family is one line that deploys to *every* environment at once. `dual_subscribe` is per-environment, read from `EVENT_BUS_DUAL_SUBSCRIBE` and gated on that environment's expand apply. So the precondition is per-environment while the thing depending on it is global, and nothing links them at runtime.

**The residual failure is the silent one.** This PR deliberately made the too-early direction loud — dual-subscribe against an unprovisioned twin is `NotFound`, halts `_pull_loop`, reds `/readyz`. The reverse has no signal at all: `subscribe_names(dual=False)` returns `(current,)`, which does not track the flip, so a flipped family in a flag-off environment publishes to a topic nobody is bound to. No exception, no NotFound, green health, zero delivered. For `audit.*` that is the hash-chained loss the comment calls out as the one thing this programme cannot undo — reached not by flipping audit early, but by flipping any family in an environment whose flag was never set.

**Suggestion, in this PR's own idiom:** have the Pub/Sub bus refuse to start when `FLIPPED_FAMILIES` is non-empty and `dual_subscribe` is false — the combination is never intentional, and failing loudly at construction converts the last silent mode into a loud one. That is the same call the flag parsing already makes ("only the second is recoverable by doing nothing"). In-process needs no guard since it binds both unconditionally.

Cheap version if a hard failure is too blunt for on-prem: log it at error on every `start()` and surface it through `is_healthy`, so at least something is red.

Not my lane and I have not touched `common/events/` — flagging before step 4 rather than after, since this is the step §13 says the programme keeps getting wrong.

PR #912, opened 2026-08-23
docs: teach the CAURA_* env names, with one surviving alias table

Eldad-Caura:
**Phase 5.4 for this repo's docs.** Every place the documentation told a reader to set a `MEMCLAW_*` variable now teaches the `CAURA_*` spelling. This changes **what we teach, not what anything reads** — every old name keeps working.

### Verified each name against its reader first

A doc that teaches a name nothing accepts is worse than one teaching a stale name that works, so nothing was renamed on faith:

| Name taught | Reader | 
|---|---|
| `CAURA_API_URL` `_API_KEY` `_TENANT_ID` `_FLEET_ID` `_NODE_NAME` `_AUTO_WRITE_TURNS` | `plugin/src/env.ts:85-104`, via `readEnv` |
| `CAURA_AUTO_FIX_CONFIG` | `plugin/src/index.ts:406` — **not** env.ts |
| `CAURA_SKILL_TARGETS` | `plugin/src/reconcile-skills.ts:194` |
| `CAURA_RUN_DESTRUCTIVE_MIGRATIONS` | `012_vector_dim_1024.py:133` |
| `CAURA_VERSION` | `docker-compose.yml:72,104` |
| `CAURA_API_KEY` (server gate) | `core-api/config.py`, via `_prefer_the_new_api_key_name` |

**Two of those are not where you would look.** `AUTO_FIX_CONFIG` is read in `index.ts` and `SKILL_TARGETS` in `reconcile-skills.ts`, not in `env.ts` with the rest. Grepping `env.ts` alone — the obvious move — reports both as unsupported and leaves two names un-swept. Worth knowing for the repos still to do.

**Both readers resolve first NON-EMPTY, not first defined.** `readEnv` skips a blank alias; `config.py` deliberately avoids `AliasChoices` for the same reason. That is precisely what makes teaching the new name safe next to a deploy template still carrying the old one, and the alias note now says so.

### One surviving alias table

Per 5.4's "leave exactly one legacy-alias table per repo", the mapping now lives in a single place — the **Legacy spellings** note under the plugin env table — and nowhere else. It states the prefix swap, the first-non-empty rule, and that unprefixed vars (`ADMIN_API_KEY`, `POSTGRES_*`, …) never had a branded spelling. After this, exactly **two** `MEMCLAW_` mentions remain in the swept files, both deliberate alias statements.

### Scope notes, both worth a look

**`.env.example` is included, and it is outside this lane's stated docs scope.** The README section it pairs with now says `CAURA_VERSION`; shipping a template that writes `MEMCLAW_VERSION` next to it would manufacture exactly the doc-versus-artifact mismatch this lane has spent four PRs removing. `POSTGRES_USER`/`POSTGRES_DB` in that file are floor strings and are untouched. Revert that one file if you'd rather keep the lane boundary strict.

**5.4's stated dependency, `5.3`, is still open** — `caura-onprem-installer#17`, "accept the CAURA_* spelling of every name this installer reads". That gate is real but it is about the **installer's own** names in the **installer's** repo: teaching a customer to put `CAURA_*` in `install.conf` before the installer reads it would produce a broken install. Every name in *this* PR is read by *this* repo's code, verified above, so the OSS half is not gated on it. The ONPREM half of 5.4 still is.

### Gates

Ratchet **no new lines, −40 net**, three exempt lines (the two alias statements and one troubleshooting row that legitimately keeps `memclaw.allowlist.fix`, the live RPC method). Sentinel **23/23**.

The exemption markers are HTML comments. Checked that this is safe for the served guide: the customer path is the README's relative link, GitHub-rendered, where HTML comments are hidden — and `/static/docs/integration-guide.md` 404s publicly, so nobody reads it as raw text. Also confirmed the edited troubleshooting row still parses as a 2-cell table row against its 2-cell header.

### Not in scope

Product-name prose (`# Upgrading the memclaw plugin`), the plugin id and its RPC methods, install paths, the `[memclaw]` log prefix, the skills dir. All real names, none of them env vars. Prose renaming is Phase 1.3 / Phase 7, not 5.4.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -42,9 +42,9 @@ ENVIRONMENT=development
 # build from local source instead, run
 # ``docker compose up --build --no-pull``.
 #
-# MEMCLAW_VERSION is tag-only (it expands inline into ``image:`` for
+# CAURA_VERSION is tag-only (it expands inline into ``image:`` for
 # both core-api and core-storage-api). For *digest pinning* — strict
-# supply-chain reproducibility — a single MEMCLAW_VERSION is not
+# supply-chain reproducibility — a single CAURA_VERSION is not
 # enough: the two services have different images and therefore
 # different sha256 digests, so a shared digest would fail to pull on
 # whichever service it doesn't belong to. Pin per-service via a
@@ -52,7 +52,7 @@ ENVIRONMENT=development
 # own ``name:tag@sha256:<digest>``. Look up each service's digest with:
 #     docker buildx imagetools inspect ghcr.io/caura-ai/caura-memclaw-core-api:v1.2.3
 #     docker buildx imagetools inspect ghcr.io/caura-ai/caura-memclaw-core-storage-api:v1.2.3
-# MEMCLAW_VERSION=latest
+# CAURA_VERSION=latest
 
 # -- Database (PostgreSQL + pgvector) ----------------------------------------
 POSTGRES_HOST=127.0.0.1
@@ -69,7 +69,7 @@ POSTGRES_REQUIRE_SSL=false
 ADMIN_API_KEY=
 # Optional: when set, ALL non-admin requests must include this key via X-API-Key.
 # Useful when exposing the API to a network (not just localhost).
-MEMCLAW_API_KEY=
+CAURA_API_KEY=
 
 # -- Embedding provider -----------------------------------------------------
 # Options: openai | local | fake
--- a/README.md
+++ b/README.md
@@ -173,7 +173,7 @@ Anthropic, Gemini, and OpenRouter don't offer embedding APIs here — pair them 
 docker compose up -d
 ```
 
-By default this **pulls the multi-arch images from `ghcr.io`** (`linux/amd64` + `linux/arm64`) on first run — takes ~30 seconds. Subsequent `up` commands re-use the cached image (no registry round-trip, works offline). To pin a specific version, set `MEMCLAW_VERSION=v1.2.3` in your `.env`. To build from local source instead (e.g. when iterating on a fork), run `docker compose up --build --no-pull`.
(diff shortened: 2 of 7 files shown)


Eldad-Caura:
@erni-a ready for your review — all checks green, and this one **did** get an automated review: `claude-review` came back **✅ No issues found**.

Worth noting why, since the last three PRs skipped it: including `.env.example` defeats the docs-only skip filter. So the lane's most substantive change is also the only one that got an automated pass — which is a decent argument for that filter being slightly too broad for prose that ships copy-pasteable config.

Two judgment calls I'd like you to confirm rather than assume:

1. **`.env.example` is outside this lane's stated docs scope.** I included it because the README section it pairs with now teaches `CAURA_VERSION`, and shipping a template that writes `MEMCLAW_VERSION` beside it would create exactly the doc-versus-artifact mismatch this lane has been removing. Happy to drop that one file if you'd rather keep the boundary strict.

2. **5.4's stated dependency 5.3 is still open** (`caura-onprem-installer#17`). I judged that gate to be about the installer's own names in the installer's repo — teaching a customer `CAURA_*` in `install.conf` before the installer reads it would break installs. Every name in this PR is read by this repo's code, each one verified against its reader. The ONPREM half of 5.4 remains gated.

The detail worth carrying to the other repos is in the description: two of the ten names are read outside `env.ts` (`AUTO_FIX_CONFIG` in `index.ts`, `SKILL_TARGETS` in `reconcile-skills.ts`), so the obvious `grep env.ts` reports them as unsupported and silently leaves them un-swept.

erni-a (approved):

Eldad-Caura:
@claude

This branch was rebased onto `main` after your first review, which fired `synchronize` — so the review job skipped and your ✅ verdict now predates the current head (`7472cb55`). Re-reviewing so the verdict matches what would actually merge.

The commit is unchanged in content; only the base moved (main picked up #910).

PR #915, opened 2026-08-24
ci: accept either spelling of the agents key, so the rename is an admin flip

Eldad-Caura:
The `.github` half of 5.4. **It cannot be a rename**, and the tests half is almost entirely a no-op — both for reasons worth reading before the next repo does this.

### Why .github is a dual-accept, not a rename

`MEMCLAW_AGENTS_KEY` is a **live repository secret** — created 2026-08-10, confirmed with `gh secret list`. Rewriting the reference to `secrets.CAURA_AGENTS_KEY` resolves to `''` until an admin creates that secret, and **both scripts are built to go dark and silent on an empty key**. So the obvious rename would have switched off declined-finding capture and learned-guidance recall with nothing red anywhere — the exact silent-degradation shape this programme keeps re-learning.

So the workflow resolves from either secret and exports the new name:

```yaml
CAURA_AGENTS_KEY: ${{ secrets.CAURA_AGENTS_KEY || secrets.MEMCLAW_AGENTS_KEY }}
```

An unset secret is falsy, so **behaviour today is byte-identical** — the existing secret still wins. Adding `CAURA_AGENTS_KEY` flips it with no code change, and deleting the old one afterwards is then safe. **The remaining secret rename is an admin action; this PR makes it a no-risk flip instead of a coupled change.**

The scripts read new-with-old-fallback via `${CAURA_X:-${MEMCLAW_X:-}}` — first **non-empty**, since `:-` treats blank as unset, so an unfilled `CAURA_AGENTS_KEY=` cannot shadow a working old key and take capture dark.

**Exercised all five states** rather than reasoning about them:

```
  unset + unset           -> []        (goes dark)
  new-only                -> [newkey]
  old-only                -> [oldkey]
  blank-new + working-old -> [oldkey]  <- must not go dark
  both                    -> [newkey]
```

`actionlint` clean, both scripts pass `bash -n`, workflow parses as YAML.

### Why the tests half is (correctly) empty

5.4 says "rewrite every doc, installer heredoc, workflow and **test** to teach `CAURA_*`". Taken literally on tests, that **destroys the coverage that proves the old names still work**. `tests/test_env_dual_read.py` says so in its own docstring: *"a test that only proves the new name works would pass just as well after someone deleted the fallback."* Per file:

| File | Why it stays |
|---|---|
| `tests/test_env_dual_read.py` | **Is** the dual-read guard; both names are the assertion. Already marked `legacy-name-ok`. |
| `tests/conftest.py:51` | Scrubs `MEMCLAW_API_KEY` **and** `CAURA_API_KEY` from the env so a dev's real key can't leak into tests. Dropping the old name re-opens that leak. |
| `tests/test_migration_012_safety_gate.py`, `tests/test_api_status.py` | These are the **old-name arm** of two-directional coverage. Flipping them to `CAURA_*` leaves nothing proving the old name opts in. |
| `clients/python/tests/test_interviewer_cursor.py` | Same — the old-name arm for `INTERVIEWER_HARNESS`. |
| `tests/test_agent_identity_guard.py`, `tests/test_preflight_012.py` | Assert **source text** (`agent_identity.py`'s error, migration 012's pre-flight). Source changes first, test follows; both sources are outside this lane. |
| `tests/test_plugin_source_manifest.py` | Refers to `MEMCLAW_TOOLS`, a `plugin/src` identifier — not an env var, and Agent 4's lane. |
| `plugin/src/*.test.ts` | `env.test.ts` is the plugin's dual-read guard and pins the *exclusions* too. Agent 4's lane. |

### Also deliberately untouched

- **`ci.yml`'s `POSTGRES_DB` / `POSTGRES_USER` / `pg_isready -U memclaw`** — floor strings. The sentinel pins that family, and CI's service credentials must keep matching the compose defaults that customers' volumes were created with.
- **Issue templates** ("MemClaw version") — product-name prose, which is the W1 punch list / Phase 1.3, not the 5.4 env family.

### One finding for whoever does the core-storage-api half

`scripts/do_not_touch_sentinel.py:256` pins `os.environ.get("MEMCLAW_RUN_DESTRUCTIVE_MIGRATIONS"` as a **floor string**. So 5.4's instruction to rewrite `012_vector_dim_1024.py`'s pre-flight text collides with the sentinel unless that row moves in the same commit. Ordering trap, not a blocker — but it will fail the gate if hit blind.

Also: `scripts/repro_contradictions_*.py` read `MEMCLAW_API_URL` / `_API_KEY` / `_TENANT_ID` through a **bare** `_env()` with no fallback, so `CAURA_*` genuinely does not work there. Real gap, in `scripts/`, outside this lane.

Ratchet: **no new lines, −17 net**, 8 exempt lines. Sentinel: **23/23**.

### DIFF
--- a/.github/scripts/claude_pr_capture.sh
+++ b/.github/scripts/claude_pr_capture.sh
@@ -11,7 +11,7 @@
 # shared pipeline and keeps its own copy. See the workflow header. The fleet is shared even
 # though the code is not: notes written here are recalled by the other six repos and vice versa.
 #
-# DARK AND SILENT without MEMCLAW_AGENTS_KEY. That is the state to expect at the moment this
+# DARK AND SILENT without CAURA_AGENTS_KEY. That is the state to expect at the moment this
 # lands: the org secret has `private` visibility, which a PUBLIC repo cannot read, so until a
 # repo-level secret exists this exits immediately having done nothing.
 #
@@ -29,8 +29,9 @@
 #   ANTHROPIC_API_KEY Anthropic API key
 #   GH_TOKEN          token with issues:read and pull-requests:read
 # Optional env:
-#   MEMCLAW_AGENTS_KEY   internal-agents tenant key (empty => dark no-op)
-#   MEMCLAW_API_URL      default https://caura.ai
+#   CAURA_AGENTS_KEY     internal-agents tenant key (empty => dark no-op)
+#   CAURA_API_URL        default https://caura.ai
+# Both also accept their pre-rename MEMCLAW_* spelling.  # legacy-name-ok: rule 3 dual-read alias
 #   CODE_REVIEW_FLEET_ID default code-review
 #   MODEL                default claude-sonnet-5
 #   MAX_BUDGET_USD       per-invocation ceiling (default 2.00)
@@ -46,8 +47,15 @@ MAX_THREAD_CHARS=60000
 # from here would be the only one no per-reviewer filter could account for.
 REVIEWER="claude"
 
-if [ -z "${MEMCLAW_AGENTS_KEY:-}" ]; then
-  echo "::notice::Declined-finding capture is dark (no MEMCLAW_AGENTS_KEY) — skipping"
+# Either spelling: the workflow passes CAURA_AGENTS_KEY (itself resolved from
+# whichever secret exists), but a manual run may still export the old name.
+# ``:-`` treats blank as unset, so this is first NON-EMPTY, not first defined —
+# an unfilled CAURA_AGENTS_KEY= must not shadow a working old one and take the
+# capture dark without saying so.
+AGENTS_KEY="${CAURA_AGENTS_KEY:-${MEMCLAW_AGENTS_KEY:-}}"  # legacy-name-ok: rule 3 dual-read alias
+
+if [ -z "$AGENTS_KEY" ]; then
+  echo "::notice::Declined-finding capture is dark (no CAURA_AGENTS_KEY) — skipping"
   exit 0
 fi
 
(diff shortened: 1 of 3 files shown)


Eldad-Caura:
@erni-a ready for your review — all checks green, `claude-review` **✅ No issues found**.

Two things need your decision rather than mine:

1. **There is a follow-up admin action.** This PR makes the agents-key rename safe but does not complete it. `MEMCLAW_AGENTS_KEY` is a live repo secret; adding a `CAURA_AGENTS_KEY` secret with the same value flips the workflow over with no code change, after which the old secret can be deleted. Until then behaviour is byte-identical. I deliberately did **not** do a bare rename — both scripts go dark and silent on an empty key, so that would have switched off declined-finding capture and guidance recall with nothing red.

2. **The tests half of 5.4 is intentionally empty**, and I think 5.4's wording should be corrected rather than followed. "Rewrite every test to teach `CAURA_*`" destroys the coverage proving the old names still work — `tests/test_env_dual_read.py` warns about exactly this in its own docstring, and `tests/conftest.py:51` scrubbing both spellings is a key-leak guard, not brand debt. Per-file reasoning is in the description. If you'd rather I add new-name arms alongside the existing old-name ones, that's a separate and bigger change in core-storage-api's suite.

Two findings for whoever picks up the remaining halves, both in the description: the sentinel pins `MEMCLAW_RUN_DESTRUCTIVE_MIGRATIONS` as a floor string, so rewriting migration 012's pre-flight text will fail the gate unless that row moves in the same commit; and `scripts/repro_contradictions_*.py` read the old names through a bare `_env()` with no fallback, so `CAURA_*` genuinely does not work there.

erni-a (approved):

PR #916, opened 2026-08-24
fix(scripts): the contradiction repros could not read a CAURA_* environment

Eldad-Caura:
Closes the gap flagged in #915. **Unlike the rest of 5.4, this one was genuinely broken** — not stale teaching.

### The bug

All six contradiction repro harnesses resolved config through a bare lookup with no alias:

```python
base = _env("MEMCLAW_API_URL")   # os.environ.get(name), no fallback
```

So an operator who followed the renamed docs and exported `CAURA_API_URL` got:

```
ERROR: $MEMCLAW_API_URL must be set     → exit 2
```

**Confirmed against the pre-change file, not assumed** — loaded both versions side by side with only `CAURA_API_URL` set:

```
operator exports CAURA_API_URL only:
  BEFORE: exit(2) — 'ERROR: $MEMCLAW_API_URL must be set'
  AFTER : https://caura.ai
```

### The fix

Five of the six shared an identical `_env` that exits on a missing value; three of those *also* read an optional tenant id inline. Rather than duplicate the alias rule twice per file, the lookup is one helper and `_env` is the required-wrapper over it:

| | |
|---|---|
| `_env_any(name)` | first non-empty of `name` and its pre-rename spelling |
| `_env(name)` | the same, exiting 2 when neither is set |

`with_entity_links.py` has no `_env` at all — three inline reads with defaults — so it gets the same helper in default-bearing form.

`or` rather than a defined-check throughout, so blank counts as unset and an exported-but-empty `CAURA_*` cannot shadow a working legacy value. Same first-non-empty rule as `env.ts`'s `readEnv` and core-api's config.

### Verified

Every helper, all six files, five states each — **30 assertions, all passing**:

```
new only                -> NEW    OK
old only (legacy)       -> OLD    OK
blank new + old         -> OLD    OK     <- the one that matters
both set                -> NEW    OK
neither                 -> None   OK
```

Plus `py_compile` on all six after every edit.

### One thing worth recording

The mechanical pass through this family **made `with_entity_links.py` worse before it made it better.** Renaming the literals there without adding a fallback would have dropped legacy support entirely — that file was the one with no helper to fix, so the rename landed but the alias didn't. Caught on the diff read, not by a gate: the ratchet counts old names and this *removed* one, so it read as progress.

That is the same asymmetry the do-not-touch sentinel exists for, one category out — a sweep that deletes a working fallback lowers the count and looks like progress. Worth a thought for whoever sweeps the remaining families.

Also: the docstrings deliberately don't spell the old prefix. The mapping sits on the line below, which carries the marker — keeping the exemption off the line a `--help` reader sees.

Ratchet: **no new lines, −29 net**, six exempt lines. Sentinel: **23/23**. `scripts/` is not ruff-gated in CI (checked), so formatting was done by hand — including a two-blank-line fix after the new import block.

### DIFF
--- a/scripts/repro_contradictions_collision.py
+++ b/scripts/repro_contradictions_collision.py
@@ -29,9 +29,9 @@ The verdict logic inspects the entity_links after the writes settle
 and reports which arm fired.
 
 Usage:
-    export MEMCLAW_API_URL=https://caura.ai
-    export MEMCLAW_API_KEY=mc_...
-    export MEMCLAW_TENANT_ID=rantaig-...
+    export CAURA_API_URL=https://caura.ai
+    export CAURA_API_KEY=mc_...
+    export CAURA_TENANT_ID=rantaig-...
     python scripts/repro_contradictions_collision.py
 """
 
@@ -48,8 +48,18 @@ import uuid
 import httpx
 
 
+def _env_any(name: str) -> str | None:
+    """First non-empty of ``name`` and its pre-rename spelling.
+
+    ``or`` rather than a defined-check on purpose: blank counts as unset, so an
+    exported-but-empty ``CAURA_*`` cannot shadow a working legacy value.
+    """
+    legacy = name.replace("CAURA_", "MEMCLAW_", 1)  # legacy-name-ok: rule 3 dual-read alias
+    return os.environ.get(name) or os.environ.get(legacy)
+
+
 def _env(name: str) -> str:
-    val = os.environ.get(name)
+    val = _env_any(name)
     if not val:
         print(f"ERROR: ${name} must be set", file=sys.stderr)
         sys.exit(2)
@@ -85,9 +95,9 @@ def main() -> int:
     ap.add_argument("--verbose", "-v", action="store_true")
     args = ap.parse_args()
 
-    base = _env("MEMCLAW_API_URL").rstrip("/")
-    key = _env("MEMCLAW_API_KEY")
-    tenant = _env("MEMCLAW_TENANT_ID")
+    base = _env("CAURA_API_URL").rstrip("/")
+    key = _env("CAURA_API_KEY")
+    tenant = _env("CAURA_TENANT_ID")
 
     # Generate two distinguishable contexts that share a common given
     # name. Use a uniqued suffix to avoid polluting prior runs.
--- a/scripts/repro_contradictions_diagnostic.py
+++ b/scripts/repro_contradictions_diagnostic.py
@@ -35,8 +35,18 @@ import uuid
 import httpx
 
 
+def _env_any(name: str) -> str | None:
+    """First non-empty of ``name`` and its pre-rename spelling.
+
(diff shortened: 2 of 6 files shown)


Eldad-Caura:
On the review's one suggestion — consolidating the duplicated `_env_any`/`_env` into a shared module. I checked it properly rather than waving it off, and I'm **declining it**, with the drift concern addressed a cheaper way.

**The suggestion is feasible.** `scripts/` does have a precedent for private shared modules — `_locomo_bench.py` and `_f3_wet_test_observer.py`, imported by siblings. So this isn't a novel structure.

**But the cost lands in the wrong place.** There's no `__init__.py`; the idiom is a `sys.path.insert(0, ...)` prologue plus `# noqa: E402` on the import:

```python
sys.path.insert(0, str(Path(__file__).resolve().parent))

from _locomo_bench import (  # noqa: E402  (path shim above must run first)
```

That's three lines of boilerplate and a lint suppression per script, to share six lines — added to standalone debug harnesses whose whole virtue is that you paste one path into a terminal and it runs. It also introduces an import-path failure mode these scripts don't currently have.

**And the duplication predates this PR.** `_env` was already copy-pasted into five of the six, and `_detected` into four. Verbatim duplication is this family's existing convention; restructuring it as a side effect of a compat fix is scope I don't think this change should carry.

**The real concern was drift** — "if the legacy-fallback window needs to be closed later, someone has to remember to touch six files." That's fair, and closing the window is a planned Phase 7 step, so it will happen. Cheaper answer: every fallback site carries the identical marker, so one grep enumerates them with no memory required:

```
$ grep -rln "legacy-name-ok: rule 3 dual-read alias" scripts/
scripts/repro_contradictions_collision.py
scripts/repro_contradictions_diagnostic.py
scripts/repro_contradictions_not_detected.py
scripts/repro_contradictions_proper_noun.py
scripts/repro_contradictions_race.py
scripts/repro_contradictions_with_entity_links.py
```

Six files, six hits. That's what the marker convention is for, and it works across the whole repo rather than just this directory.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

Eldad-Caura:
Correcting my comment above: **I've now applied the consolidation I declined** — #924.

Leaving the decline in the record rather than editing it, because the reasoning is the interesting part of the mistake. I weighed boilerplate cost against tidiness and concluded the boilerplate lost. But the reviewer's point wasn't tidiness, it was **drift**: closing the legacy-fallback window is a planned Phase 7 step, so six copies means six chances to miss one. Import cost is paid once, at review time; drift cost is paid later, by whoever is mid-sweep. Weighed against the right thing, it goes the other way.

The grep I offered here as the cheaper mitigation was also weaker than I claimed. It relies on six files carrying an identical marker string — which is itself a thing that can drift. #924 makes it one file, so the ratchet's exemption report *is* the enumeration and can't fall out of sync.

PR #924, opened 2026-08-24
refactor(scripts): one alias rule for the repro harnesses, not six copies

Eldad-Caura:
Follow-up to #916, **applying the review suggestion I argued against on that PR.** My counter-argument is [in the record there](https://github.com/caura-ai/caura/pull/916#issuecomment-5389559926) and I'll leave it standing — but on reflection it weighed the wrong two things against each other.

I compared boilerplate cost against tidiness. The reviewer's actual point was **drift**: closing the legacy-fallback window is a planned Phase 7 step, so six copies means six chances to miss one. The import cost is paid once, at review time. The drift cost is paid later, by whoever is mid-sweep. That's the trade, and it goes the other way.

### What changed

`scripts/_env_compat.py` holds the alias rule once, in the three shapes callers actually need:

| | when nothing is set |
|---|---|
| `env_any` | `None` — the value is optional |
| `env_required` | exits 2 with a message |
| `env_default` | the supplied default |

The scripts import it through the same `sys.path` prologue they already use for `_locomo_bench` and `_f3_wet_test_observer`, so this follows the directory's existing convention rather than inventing one. (`scripts/` has no `__init__.py` deliberately — these are harnesses you run by path.)

**The marker count is the measure.** Six exempt lines become one:

```
$ grep -rln "legacy-name-ok: rule 3 dual-read alias" scripts/
scripts/_env_compat.py
```

So the ratchet's own exemption report is now the enumeration of where the fallback lives — which is a better answer to the drift concern than the grep I offered on #916, because it can't drift at all.

### One thing my own check caught

`collision.py` and `race.py` import **only** `env_required` — their tenant read is required, so pulling in `env_any` as well would have been an unused import. Found by an AST pass, not by a gate: `scripts/` isn't ruff-gated, so F401 would not have failed CI.

### Verified after the move

From a working directory **outside the repo**, so the path prologue is genuinely exercised rather than accidentally satisfied by `cwd`:

- all six import cleanly
- all six answer `--help` (the real entry path)
- **30 assertions** across new-only, old-only, both, blank-new-beside-a-working-old, and nothing-set

The last case is stricter than before: through `env_required` it now asserts `exit(2)` rather than a `None` return. My first harness run tripped over exactly that and died with `ERROR: $CAURA_API_URL must be set` — the harness was wrong, not the code, and fixing it turned that case into a real assertion.

Ratchet: **no new lines**, one exempt line. Sentinel: **23/23**. Net −26 lines.

### DIFF
--- a/scripts/_env_compat.py
+++ b/scripts/_env_compat.py
@@ -1,0 +1,49 @@
+"""Env lookup shared by the standalone repro/wet-test harnesses.
+
+Every name these scripts read was renamed by the Caura rebrand, and rule 3
+makes the pre-rename spelling permanent — so each read has to try both. That
+rule lives here once rather than in each script, which is the difference
+between closing the legacy window later as one edit and remembering six.
+
+Private, underscore-prefixed module, imported by sibling scripts through the
+``sys.path`` prologue they already use for ``_locomo_bench`` and
+``_f3_wet_test_observer``: ``scripts/`` deliberately has no ``__init__.py``,
+because these are harnesses you run by path, not an importable package.
+
+The three entry points differ only in what they do when nothing is set:
+
+    env_any       -> ``None``            (the value is optional)
+    env_required  -> exits 2 with a message
+    env_default   -> the supplied default
+"""
+
+from __future__ import annotations
+
+import os
+import sys
+
+
+def env_any(name: str) -> str | None:
+    """First non-empty of ``name`` and its pre-rename spelling.
+
+    ``or`` rather than a defined-check on purpose: blank counts as unset, so an
+    exported-but-empty ``CAURA_*`` cannot shadow a working legacy value. That is
+    the same first-non-empty rule the plugin's ``readEnv`` and core-api's
+    settings use, and it is the reason neither uses ``AliasChoices``.
+    """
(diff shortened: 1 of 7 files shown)


Eldad-Caura:
@erni-a ready for your review — all checks green, `claude-review` **✅ No issues found** (and it independently verified the call-site/signature match per file, the `os`/`sys` imports the path shim needs, and that the extracted logic is byte-for-byte preserved).

Worth flagging that this reverses a call I made publicly on #916. I declined this same suggestion there with reasoning I've left standing, then corrected it — the decline weighed boilerplate against tidiness, when the reviewer's actual point was drift: closing the legacy-fallback window is a planned Phase 7 step, and six copies is six chances to miss one.

Concrete measure: six `legacy-name-ok` exempt lines become one, so the ratchet's exemption report is now the enumeration of where the fallback lives, rather than a grep across six files that could itself drift.

Verified from a working directory outside the repo so the `sys.path` prologue is genuinely exercised — all six import, all six answer `--help`, 30 assertions including blank-new-beside-a-working-old and the nothing-set case now asserting `exit(2)` through `env_required`.

erni-a (approved):

PR #925, opened 2026-08-24
fix(events): refuse to publish a flipped family into a name nothing binds

### LINKED ISSUE #913: events: a flipped topic family in a dual-subscribe-off environment publishes into the void, silently
Raised on [#911](https://github.com/caura-ai/caura/pull/911#issuecomment-5389247807) by the Lane 4 session; filed here so it is trackable rather than living in a merged PR's comment thread. **Lane 5's call and lane** — nobody outside it has touched `common/events/`.

**Not a defect in #911.** The ordering is documented on `FLIPPED_FAMILIES`, and the failure is already pinned by `test_a_flipped_family_reaches_nobody_without_the_dual_binding`. The gap is narrower: the invariant is enforced by discipline, and the half that discipline protects is the silent one.

## The two halves live at different scopes

`FLIPPED_FAMILIES` is module state in `topics.py`. Flipping a family is one line that deploys to **every** environment at once. `dual_subscribe` is per-environment, read from `EVENT_BUS_DUAL_SUBSCRIBE` and gated on **that** environment's expand apply.

So the precondition is per-environment while the thing depending on it is global, and nothing links them at runtime.

## Only one direction is loud

| combination | what happens |
|---|---|
| dual on, twin not provisioned | `NotFound` → `_pull_loop` halts → `/readyz` red. **Loud**, deliberately |
| **family flipped, dual off** | `publish_name` returns the renamed topic; `subscribe_names(topic, dual=False)` returns `(current,)` and **never consults `FLIPPED_FAMILIES`**. Publishes to a topic nothing is bound to — no exception, no `NotFound`, green health, zero delivered |

The behaviour is not a surprise: `test_a_flipped_family_reaches_nobody_without_the_dual_binding` asserts exactly this, `assert seen == []`, with the docstring *"the event goes nowhere and nothing raises, which is precisely why this ordering is not optional."* The test proves the hazard exists. Nothing prevents it.

## Why this is the route to the unrecoverable case

The danger to `audit.*` is **not** flipping audit early — that ordering is well understood and audit is scheduled last for exactly that reason.

It is flipping **any** family in an environment whose flag was never set. Staging looks correct, another environment silently drops every message in that family, and the audit rows are hash-chained so the gap cannot be reconstructed after the fact.

## Suggested fix, in #911's own idiom

Refuse to start when `FLIPPED_FAMILIES` is non-empty and `dual_subscribe` is false. That combination is never intentional — checked: even a hypothetical environment provisioned after the cutover would still have `subscribe_names(dual=False)` binding the pre-rename name, so there is no valid state it excludes. Failing at construction converts the last silent mode into a loud one, which is the same call the flag parsing already makes.

In-process needs no guard; it binds both unconditionally.

Softer variant if a hard failure is too blunt for on-prem: log at error on every `start()` and surface through `is_healthy`, so at least something is red.

## Why it is worth doing before step 4

This is the same move as caura-enterprise#1226: replace "safe if everyone follows the order" with "safe by construction". That one converted a permanent judgement call about rollback targets into a revision that refuses to start. This converts a per-environment deployment discipline into a process that will not boot in the wrong combination — and the whole reason it matters is that the current failure has no signal at all.

Eldad-Caura:
Closes #913.

Every other ordering mistake in this rename is loud. Binding a twin that was never provisioned is a permanent `NotFound` that halts the pull loop and reds the readiness probe; naming a family that does not exist raises at import. One combination has no signal at all — a flipped family publishes under its renamed name while a `dual=False` subscriber binds only the current one, so the message lands on a topic nothing is pulling. No exception, green health, zero delivered.

`test_a_flipped_family_reaches_nobody_without_the_dual_binding` already proved it. Nothing prevented it. `PubSubEventBus` now refuses to construct in that state.

## The condition is NOT "is FLIPPED_FAMILIES empty"

#913 proposed keying the guard on a non-empty `FLIPPED_FAMILIES`, on the premise that combined with `dual=False` it is never intentional. I was asked to re-derive that premise rather than take it. **It has one exception, and it is a state this cutover really passes through.**

After the contract step a family's enum members **are** the renamed names. `renamed` is idempotent, so for those topics `publish_name` and `subscribe_names(dual=False)` agree — there is no twin left to bind, and `dual=False` is not merely tolerable but *correct*. Measured:

| topic | `publish_name` | `subscribe_names(dual=False)` | mismatch? |
|---|---|---|---|
| mid-flip (`<brand>.pipeline.…`) | `caura.pipeline.…` | `('<brand>.pipeline.…',)` | **yes** — the hazard |
| contracted (`caura.pipeline.…`) | `caura.pipeline.…` | `('caura.pipeline.…',)` | no — safe and correct |

An emptiness check refuses to start the second row. And `dual=False` is the **default**, so it would take out precisely the standalone and on-prem deployments that never run the Terraform the flag is gated on — converting a silent failure into an outage, in the population least able to diagnose it.

So the guard asks the two functions instead: `unbound_publish_topics(dual=...)` returns the topics whose publish name is absent from their own subscribe set. It fires exactly when a publish would go nowhere, cannot false-positive on an already-renamed family, and needs no premise about intent — agreement is agreement however it arose.

**Both implementations pass the "refuses to construct" test.** Only `test_the_guard_does_not_fire_once_a_family_is_fully_contracted` separates them — I verified by implementing the guard as originally proposed and confirming that test, and only that test, fails.

## At construction, not at `start()`

A publish-only process never calls `start()`, so a check there would leave the write side unguarded — and the write side is the side that loses the messages.

## Live context, corrected

The brief I picked this up from said Deploy Infrastructure had failed three consecutive times, latest 01:33 after the merge, leaving the twins declared but not created. **That reading has expired, and it conflated two branches.** Re-measured from the run history:

- The four failures on `pubsub-caura-topic-twins` (00:56, 01:04, 01:16, 01:24) were all **pre-merge**, on the PR branch.
- The 01:33 failure was on `add-datadog-dbm-collector` — unrelated Datadog work.
- ent#1242 merged 01:28, and the two post-merge `dev` runs (01:45:25, 01:45:33) both **succeeded**.

So the expand step appears to have applied. I am deliberately not asserting the twins exist: that is a claim about a deploy in a specific environment, and the entire argument of this PR is that it must be confirmed per running service rather than inferred from a green branch run. The guard does not depend on the answer either way.

## Scope

The guard only. `FLIPPED_FAMILIES` is untouched and still empty, so this is a **runtime no-op in every environment today** — with nothing flipped, no topic publishes anywhere unbound under either flag value (`test_unbound_publish_topics_is_empty_in_the_shipped_state`). Steps 3–6 remain Lane 5's. **No environment configuration is changed or recommended here** — in particular this PR does not set, or tell anyone to set, `EVENT_BUS_DUAL_SUBSCRIBE`.

The in-process backend needs no guard: `subscribe` binds `subscribe_names(..., dual=True)` unconditionally, so the mismatch is unreachable there for any value of `FLIPPED_FAMILIES` — asserted across the empty, partial and all-families cases.

`common/` is vendored to enterprise, so the guard reaches that tree on the next re-vendor rather than needing a parallel change.

## Tests

Seven new tests in `test_topic_rename_cutover.py`. Verified load-bearing by breaking what each guards:

- Removing the guard entirely → `test_pubsub_bus_refuses_to_construct_when_a_flip_has_nothing_bound` fails, alone.
- Implementing it as the issue proposed → `test_the_guard_does_not_fire_once_a_family_is_fully_contracted` fails, alone.

Gates: full suite **5001 passed**, 3 skipped, 1 xfailed, 0 failed. `ruff check common/` clean (CI's exact invocation — no `--config`, per that file's own warning; `ruff format` deliberately NOT run over `common/`). mypy core-api clean, 199 files. Ratchet "No new lines", sentinel "All 23 protected strings survive", `gen_events_manifest.py --check` up to date. Rebased on `f310d432`.

Note the pre-commit ruff hooks are scoped `^(core-api|core-storage-api)/src/` and so do not cover `common/` or `tests/`; `ruff check common/` under `common/ruff.toml` is the gate that applies to this diff, and CI runs it as its own step.

### DIFF
--- a/common/events/pubsub.py
+++ b/common/events/pubsub.py
@@ -48,7 +48,7 @@ from typing import Any
 from pydantic import ValidationError
 
 from common.events.base import Event, EventBus, EventHandler
-from common.events.topics import publish_name, subscribe_names
+from common.events.topics import publish_name, subscribe_names, unbound_publish_topics
 
 logger = logging.getLogger(__name__)
 
@@ -157,6 +157,28 @@ class PubSubEventBus(EventBus):
         # project, eliminating cross-env message fan-out. See _topic_name().
         self._topic_prefix = topic_prefix
         self._dual_subscribe = dual_subscribe
+        # Refuse to exist in the one combination that fails silently: publishing
+        # a flipped family under its renamed name while binding only the current
+        # one. See ``unbound_publish_topics`` for why the check compares the two
+        # names instead of asking whether ``FLIPPED_FAMILIES`` is empty.
+        #
+        # At construction, not at ``start()``: publishing does not require
+        # ``start()`` (a publish-only process never calls it), so a check there
+        # would leave exactly the write side of the hazard unguarded. Raising
+        # here also means a misconfigured process dies before it can report
+        # itself ready. This is the loud direction of a fault that otherwise has
+        # no signal, so a hard failure is the point rather than a side effect.
+        if unbound := unbound_publish_topics(dual=dual_subscribe):
+            raise ValueError(
+                f"dual_subscribe is off, but {len(unbound)} topic(s) would be "
+                f"published under a name this bus does not bind: {sorted(unbound)}. "
+                "Nothing would be delivered and nothing would raise. Either those "
+                "families were flipped before this environment could bind their "
+                "twins, or this environment's twin subscriptions exist and its "
+                "configuration has not caught up — the two halves are set "
+                "independently, which is what lets them disagree. Resolve which "
+                "one is wrong; do not start a publisher in this state."
+            )
         self._max_messages = max_messages
         self._pull_timeout = pull_timeout
         self._error_backoff = error_backoff
--- a/common/events/topics.py
+++ b/common/events/topics.py
@@ -172,20 +172,24 @@ def family(topic: str) -> str:
 FLIPPED_FAMILIES: frozenset[str] = frozenset()
 
 
-def known_families() -> frozenset[str]:
-    """Every family the topics declared in this module actually use.
+def all_topics() -> tuple[str, ...]:
+    """Every topic declared in this module.
 
     Derived by walking the enums rather than listed, so it cannot drift from
     them, and so the same code yields the right answer in each repo despite the
-    two copies declaring different families.
+    two copies declaring different topics.
     """
-    return frozenset(
-        f
+    return tuple(
+        str(member)
         for attr in vars(Topics).values()
         if isinstance(attr, type) and issubclass(attr, enum.StrEnum)
         for member in attr
-        if (f := family(member))
     )
+
+
+def known_families() -> frozenset[str]:
+    """Every family the topics declared in this module actually use."""
+    return frozenset(f for topic in all_topics() if (f := family(topic)))
 
 
 # A family named here that does not exist would be a SILENT no-op: publish_name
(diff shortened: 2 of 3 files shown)


Eldad-Caura:
@erni-a polease pprove

erni-a (approved):

PR #932, opened 2026-08-24
docs: commit the rebrand sunset plan in-repo

Eldad-Caura:
Phase 0.1 of the sunset plan, and the last P0 task that had not moved.

Until now **neither repo contained any rebrand plan at all**, so the only way to know whether an old-brand string was contract or debt was to ask someone who had been in the room. That is the wrong default generally, and a bad one right now, with the rename being executed in parallel across several repos by people and agents who cannot all ask.

### What it deliberately does not do

**It does not restate the phase board.** That changes daily. A copy committed here would be wrong within a week and authoritative-looking forever. What is committed is the part that does not change: the seven rules, the shape of the floor, and what each CI gate does and does not catch.

**It does not enumerate the floor.** `scripts/do_not_touch_sentinel.py` is the authoritative list, because a list in CI cannot silently rot the way a list in prose does. The doc explains the categories and tells the reader to run `--list`.

### The one exemption

`docs/plans/rebrand-sunset-plan.md:10` mentions the legacy name and is marked `legacy-name-ok`. It is load-bearing: the file has to be findable by exactly the grep it exists to answer. Someone searching this repo for the old name should land on the rules rather than guess. The ratchet reports it, which is the intended outcome — an exemption a reviewer can see.

### A correction it carries

Planning documents have been circulating a claim that Pub/Sub topics, the database and the Cloud Run services are **permanently frozen**. They are not, and all three are being migrated right now. Anyone acting on the frozen assumption would conflict with work in flight, so the doc says so explicitly and describes the sequence for each — because in all three cases the order matters more than the change.

### Gates

- `legacy_name_ratchet.py --base origin/main` → **No new lines**, 1 exemption reported.
- `do_not_touch_sentinel.py` → **All 23 protected strings survive.**

### DIFF
--- a/docs/plans/rebrand-sunset-plan.md
+++ b/docs/plans/rebrand-sunset-plan.md
@@ -1,0 +1,118 @@
+# Rebrand sunset plan — the in-repo reference
+
+**Status:** active · **Last verified against the repos:** 2026-08-24
+**Enforcement:** `scripts/legacy_name_ratchet.py` and `scripts/do_not_touch_sentinel.py`, both required checks
+
+The product was renamed to Caura. This file is the in-repo answer to "is this old-brand
+string a bug, and may I change it?" — because for a large and load-bearing subset of
+them the answer is **no**, and nothing in the code says so at the point of use.
+
+If you grepped this repo for `memclaw` and landed here, that grep worked as intended. <!-- legacy-name-ok: this doc must be findable by the exact grep it exists to answer -->
+Read [The floor](#the-floor) before you change anything you found.
+
+---
+
+## The seven rules
+
+These are the durable part of the plan. Every gate in CI traces to one of them.
+
+| # | Rule | What it means in practice |
+| --- | --- | --- |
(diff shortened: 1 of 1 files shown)


Eldad-Caura:
@erni-a pelease approve

Eldad-Caura:
@erni-a green and ready for review.

All 13 checks settled — 10 pass, 3 skipped, nothing pending or failing. Both rebrand gates included: **Legacy-name ratchet** reports `No new lines` with 1 exemption surfaced, and **Do-not-touch sentinel** confirms all 23 protected strings survive.

One thing worth a reviewer's eye rather than a rubber stamp: the single `legacy-name-ok` exemption at `docs/plans/rebrand-sunset-plan.md:10`. It is deliberate — the file has to be findable by the grep it exists to answer — but it is exactly the kind of annotation that should be read rather than waved through, and the ratchet reports it precisely so someone does.

erni-a (approved):

erni-a (approved):

PR #931, opened 2026-08-24
chore: point CHANGELOGs at GitHub Releases

Eldad-Caura:
## Summary

Both CHANGELOGs duplicated release history that [GitHub Releases](https://github.com/caura-ai/caura/releases) already publishes in full (52 per-component releases). The duplicated bodies carried **482 pre-rename brand lines** and grew with every release.

| file | lines | old-brand lines |
|---|---|---|
| `CHANGELOG.md` | 968 → 185 | 417 → **0** |
| `plugin/CHANGELOG.md` | 201 → 24 | 65 → **0** |

10 insertions, 970 deletions.

## Cut point, and why

Retain every release section **already on the post-rename `caura-ai/caura` URL scheme**, never fewer than the latest release:

- `CHANGELOG.md` — `2.30.0`, `2.29.0`, `2.28.0`
- `plugin/CHANGELOG.md` — `2.17.0`, `2.16.1`

That boundary is where the rename actually falls (`2.27.0` and older point at `caura-ai/caura-memclaw`), so the retained text needs no rewriting and contains no stale links, while each file still answers "what shipped last". It also happens to remove 100% of the old-brand lines in both files — the only one left in the retained range was the preamble's "changes to MemClaw", rewritten here.

## Why this cannot break releases

- Versions come from `.release-please-manifest.json` (`"." : 2.30.0`, `"plugin": 2.17.0`), not from these files.
- `release-type: simple` / `node` **append**; the updater never parses existing content to compute the next version. Verified empirically on the last release commit f38db85d: `CHANGELOG.md | 59 +++++`, **59 insertions, 0 deletions**, inserted at `@@ -6,6 +6,65 @@` — directly below the prose preamble, nothing else touched.
- That same commit proves the updater preserves hand-written content above the first `##` heading, which is where the Releases pointer is placed — so it survives future releases.

## Notes for reviewers

- Full history remains in Releases and in git; nothing is unrecoverable.
- `chore:` (not `docs:`) so this commit stays out of the next generated section.
- Two pre-existing inconsistencies spotted but **not** touched, as they're outside this PR's scope: `.github/PULL_REQUEST_TEMPLATE.md:32` asks contributors to update an `Unreleased` section that has never existed in this file (release-please generates sections from commits); and `plugin/src/context-engine.ts:1240` cites "CHANGELOG line 4940", which refers to *OpenClaw's* changelog, not this one — unaffected.

PR #933, opened 2026-08-24
chore: rename the old brand out of core-api and core-storage-api prose

Eldad-Caura:
Phase 1.3, **slice 1 of 6** — the internal share of the mass rewrite, scoped to the two Python backend trees. Comments, docstrings and package descriptions, plus one user-visible installer title.

## Measured

| scope | before | after |
|---|---|---|
| `core-api/` + `core-storage-api/` (non-exempt) | 135 | 114 |
| repo-wide (non-exempt) | 1456 | 1432 |

Ratchet: **-24 net, no new lines**, 1 exemption. Sentinel: **all 23 protected strings survive.**

## What was deliberately left alone

The remainder in these trees is the designed permanent floor, not missed work:

- **env vars** — `MEMCLAW_API_KEY`, `MEMCLAW_RUN_DESTRUCTIVE_MIGRATIONS`, `MEMCLAW_AGENT_ID`, … Operators set these; `config.py` already dual-reads `CAURA_*` as primary.
- **wire formats** — the `/api/v1/memclaw` mount, the `memclaw_*` tool prefix, the `memclaw-doc://` URI scheme, on-disk install paths, plugin ids.
- **Pub/Sub topics and schema names** — Agent 5 owns that migration; untouched here.
- **landed Alembic migrations** (012, 019, 030) — immutable history.
- **`memclawd` references** — a separate component, named by its own repo.
- **8 of the sentinel's 23 protected strings** live in these trees.

## Two notes for review

**1. `_SKILL_LABELS` carries the only exemption this change writes.** The display value becomes `"Caura"`; the KEY stays `"memclaw"` because it is both the `?skill=` query value and the `static/skills/<id>/` directory name (`_SKILL_FILES` builds the path from it, `_VALID_SKILLS` validates against it). The dict is exploded so the `legacy-name-ok` marker sits on the key's own line and stays under the 110-char limit.

**2. `distill_prompt._SYSTEM_PREAMBLE` is LLM-visible text.** The brand token is pure identity prose, but the same prompt asks the model to keep fingerprint wording (`goal_phrase`, `domain`, `step_skeleton`) stable across re-runs of a cluster. I judged the preamble brand token inert with respect to those output fields, but flagging it rather than assuming.

## Also

`auth.py` now documents `CAURA_API_KEY` rather than `MEMCLAW_API_KEY`, per #929. The attribute it annotates is still `settings.memclaw_api_key`, which holds the resolved value of either (`config.py:378`) — this teaches the preferred name without changing behaviour. Docstring arrow alignment preserved at column 42.

The installer-title test moves with the label it pins — a source change and its pinning assertion cannot land in separate slices without leaving CI red in between.

## Verification

- `scripts/do_not_touch_sentinel.py` → exit 0, all 23 survive
- `scripts/legacy_name_ratchet.py --base origin/main` → exit 0, -24 net
- `ruff check` / `format --check` / `format --check --isolated` at CI's exact scopes → all clean
- `pytest tests/test_install_skill_endpoint.py` → 16 passed; related files 127 passed
- 7 pre-existing `kreuzberg` PDF-ingest failures confirmed identical on a clean `origin/main` tree (module not installed locally) — unrelated to this change

Rebased onto `8688e363` after #931 landed.

### DIFF
--- a/core-api/pyproject.toml
+++ b/core-api/pyproject.toml
@@ -1,7 +1,7 @@
 [project]
 name = "core-api"
 version = "2.30.0"
-description = "MemClaw Core API service"
+description = "Caura Core API service"
 license = "Apache-2.0"
 requires-python = ">=3.12"
 dependencies = [
--- a/core-api/src/core_api/auth.py
+++ b/core-api/src/core_api/auth.py
@@ -29,7 +29,7 @@ class AuthContext:
 
     OSS auth paths:
     1. Admin API key (ADMIN_API_KEY)      → is_admin=True, tenant_id=None
-    2. MemClaw API key (MEMCLAW_API_KEY)  → gates all non-admin access when set
+    2. Caura API key (CAURA_API_KEY)      → gates all non-admin access when set
     3. Standalone mode                     → tenant_id from config, org_role="admin"
     4. X-Tenant-ID header (enterprise)    → tenant_id from header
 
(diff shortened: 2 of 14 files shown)


Eldad-Caura:

All checks green (0 non-passing). Both brand gates passed on CI: **Do-not-touch sentinel** (all 23 protected strings survive) and **Legacy-name ratchet** (-24 net, no new lines). The Claude auto-review came back clean — no issues found.

Scope recap: prose, docstrings and package descriptions in `core-api` + `core-storage-api`, plus one user-visible installer title (`_SKILL_LABELS`) whose pinning test moves with it. The one exemption this PR writes is on the `_SKILL_LABELS` key, which is both the `?skill=` query value and the on-disk `static/skills/<id>/` directory name.

Measured: 135 → 114 non-exempt old-brand lines in the two backend trees.

erni-a (approved):

PR #938, opened 2026-08-24
chore: rename the old brand out of plugin comment prose

Eldad-Caura:
Phase 1.3, **slice 3** — comment and docstring prose in `plugin/src`. 42 lines across 15 files, plus one new sentinel entry. No identifiers, no wire values, no logic.

**Measured: -42 net old-brand lines, no new lines.**

## Only 42 of `plugin/`'s 796 non-exempt lines are safe prose

The exclusions are the interesting part:

| excluded | why |
|---|---|
| **`educate.ts` + `educate.test.ts`** — skipped entirely | The legacy migration surface. Its comments **quote the old headings verbatim** (`"## MemClaw —"`, `auto-added by MemClaw plugin`, the pre-C1 blurb), so rebranding them makes the comments describe strings that don't exist. #936 pinned the seven anchors; the comments *about* them aren't pinned and must keep the old spelling to stay true. |
| `educate.test.ts` fixtures | Legacy **content**, not prose. Renaming leaves the suite green while testing nothing. |
| `heartbeat.ts:492,527` | Comments describing the `.includes("memclaw")` checks — renaming makes them describe a search the code doesn't perform. |
| `config.ts` allowlist comments | "ensure memclaw is in `plugins.allow`" names the plugin id, which is protected. |
| `context-engine.ts:2`, `context-engine.internal.ts:2` | Deferred so `"MemClaw ContextEngine"` moves **with** the class it names rather than drifting from it. |
| 72 wire lines + 18 identifier lines | `MEMCLAW_*` constants, `memclaw://` URIs, `/api/v1/memclaw` routes, `skills/memclaw` paths; identifiers are their own slice. |

## One sentinel entry added (30 → 31)

`reconcile-skills.ts:80` — `OWNED_MARKER = ".memclaw-owned"`. **Same category as the six pinned by #936, and missed there.** It's written into every managed skill dir on users' disks, and `isMemclawOwned()` gates *all* pruning on it:

```ts
if (!isMemclawOwned(join(skillsRoot, slug))) continue; // foreign — leave alone
```

Rename it and every existing dir reads as foreign, so orphans stop being collected — permanently and silently.

Pinned in **this** PR rather than a follow-up because this PR rebrands the prose immediately around it: those comments now say Caura, which gives the next reader one more reason to think the marker should follow them. Flagging that as a deliberate scope addition.

Verified the same way as #936 — sentinel catches the rename (**exit 1**), ratchet passes it (**exit 0**).

## Three judgement calls worth a look

**`OWNED_MARKER_BODY`** text is rebranded. It's written to users' disks but **never read back or compared** (only `existsSync` on the path — grepped both uses), so it's inert. Existing on-disk markers keep the old body harmlessly.

**`isMemclawOwned()` keeps its name** this slice while its docstring now says Caura. That inconsistency is deliberate and resolves in the identifier slice — renaming the function here would mean touching its call sites and tests, which belongs with the rest of the identifier work.

**`tools.ts:7`** described an `"Available MemClaw tools: …"` prompt line — but `prompt-section.ts:34` emits `"Available tools: "` with no brand at all. The comment was **already stale**; corrected to match what's actually emitted rather than rebranding a string that doesn't exist.

## Verification

| check | result |
|---|---|
| `do_not_touch_sentinel.py` | exit 0, all 31 survive |
| sentinel on a renamed marker | **exit 1**, names it |
| `legacy_name_ratchet.py --base origin/main` | exit 0, no new lines, -42 net, 1 exemption (the pinned string) |
| `cd plugin && npm ci && npm run build` (tsc) | exit 0 |
| `cd plugin && npm test` | **414 tests, 414 pass, 0 fail** |

CI type-checks and builds `plugin/` but does not run its test suite, so the 414 above were run locally.

### DIFF
--- a/plugin/src/agent-auth.ts
+++ b/plugin/src/agent-auth.ts
@@ -1,5 +1,5 @@
 /**
- * Agent authentication broker for MemClaw plugin.
+ * Agent authentication broker for Caura plugin.
  *
  * Auto-provisions agent-scoped credentials using the tenant-scoped key
  * as a bootstrap credential. Both kinds share the `mc_` prefix on the
--- a/plugin/src/config.ts
+++ b/plugin/src/config.ts
@@ -317,7 +317,7 @@ function canonicalDir(p: string): string {
  * ``openclaw.json`` — OpenClaw's documented, watched load path for extra
  * skill directories (``docs/tools/skills-config.md``; consumed by
  * ``src/skills/runtime/refresh.ts``). This is how a reconciled *additive*
- * target dir (one MemClaw doesn't own and can't publish as a plugin skill)
+ * target dir (one Caura doesn't own and can't publish as a plugin skill)
  * actually reaches agents.
  *
  * Append-only and idempotent: existing entries are preserved, a dir already
--- a/plugin/src/context-engine.ts
+++ b/plugin/src/context-engine.ts
@@ -1050,7 +1050,7 @@ export class MemClawContextEngine {
    *
    *   1. **Persist** any pre-computed summary (passed in by OpenClaw
    *      when its compaction pipeline ran the summarization step
-   *      before invoking us) as a MemClaw episode memory tagged
+   *      before invoking us) as a Caura episode memory tagged
    *      ``auto-compaction``. This is the long-standing
    *      observability side-effect — surviving summaries get
    *      recallable later via ``caura_recall``.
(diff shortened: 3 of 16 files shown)


Eldad-Caura:

All checks green (0 non-passing) on head `0c3de09f`; no rebase happened, so the clean Claude review corresponds to exactly this commit. Locally: `tsc` build exit 0 and `npm test` 414/414 pass — worth knowing CI builds `plugin/` but does not run its test suite, so that run was local only.

**The one thing that is more than prose,** and the reason to give this a second look: it adds a 31st sentinel entry pinning `OWNED_MARKER = ".memclaw-owned"` (`reconcile-skills.ts:80`). That is the same category as the six #936 pinned, and it was missed there — the marker sits in every managed skill dir on users' disks and the ownership check gates all pruning on it, so renaming it makes every existing dir read as foreign and orphans stop being collected, permanently and silently. Verified: sentinel exits 1 on the rename, ratchet exits 0.

I pinned it in this PR rather than a follow-up because this PR rebrands the comments immediately around it — those comments now say Caura, which gives the next reader one more reason to think the marker should follow. Deliberate scope addition, not an accident.

Note also that `educate.ts` and `educate.test.ts` were skipped **entirely**, on purpose: their comments quote the old headings verbatim, so rebranding them would produce accurate-looking documentation of strings that no longer exist.

erni-a (approved):

PR #940, opened 2026-08-24
perf(llm): log per-call token usage and add contradiction-judge reasoning-effort control

arkash20:
## Why

Aug-2026 prod OpenAI spend ($1,155 on core-api+worker) is ~80% **output** tokens. The A61 call batching (#770/#771) cut daily request count ~5x but the bill barely moved — OpenAI bills tokens, not calls, and the only per-call cost signal in our logs was `llm_ms`. Tracked as **E3** on the shared task checklist.

## What

Two changes, deliberately **without** touching the 20-candidate `/similar-candidates` default (contradiction recall is already the weak spot; cutting candidates trades recall for cost — decision 2026-08-24):

1. **Token-usage logging** — `OpenAILLMProvider.complete_json`/`complete_text` now log `tokens_in` / `tokens_out` / `tokens_reasoning` from `response.usage` on every call, so a spend fix is verifiable in dollars instead of call counts. `tokens_reasoning` (from `completion_tokens_details`) is broken out because reasoning models bill it as output while it never appears in the visible content. Defensive against OpenAI-compatible endpoints that omit `usage`.

2. **`CONTRADICTION_REASONING_EFFORT` knob (default: unset)** — `complete_json` accepts `reasoning_effort` (protocol-wide, accept-and-ignore on fake/gemini/vertex), forwarded only when set. All four contradiction-judge call sites pass the setting through `_judge_effort_kwargs()` — empty kwargs when unset, so the wire request and pre-existing test doubles are untouched by default.

## Wet test (local enterprise stack, real gpt-5.4-nano)

Full rebuild + e2e on 5 contradicting memory pairs, which caught two things unit tests couldn't:

- **Reasoning mode rejects `temperature=0.0`** (400: only default 1 supported) → the provider now drops `temperature` whenever `reasoning_effort` is sent. Without this the judge silently degraded to abstention via the fallback chain.
- **Valid effort values are model-specific** — gpt-5.4-nano takes `none|low|medium|high|xhigh` (no `minimal`). Documented on the setting; an unsupported value 400s every call and abstains.

Verified end-to-end: token fields present on every LLM log line; `effort=low` → judge verdicts still correct (verdict=True 0.90 on planted contradictions), `tokens_reasoning=37–117` per judge call; `effort=none` → correct verdicts, zero reasoning tokens; unset → byte-identical behavior to main.

**Operational note:** on this local workload the *default* (unset) mode already showed `tokens_reasoning=0`, so the right prod move is: deploy the logging first, read the prod `tokens_reasoning` distribution, then decide the knob value from data.

`ruff` / `mypy` / 5028 unit tests green.

### DIFF
--- a/common/llm/protocols.py
+++ b/common/llm/protocols.py
@@ -33,6 +33,7 @@ class LLMProvider(Protocol):
         temperature: float = 0.0,
         seed: int | None = None,
         response_schema: dict | None = None,
+        reasoning_effort: str | None = None,
     ) -> dict:
         """Send a prompt and return a parsed JSON dict.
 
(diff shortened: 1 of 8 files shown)


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #941, opened 2026-08-24
refactor(plugin): collapse the plugin id to one constant and pin both ends

Eldad-Caura:
The plugin id was spelled inline **eleven times** in `config.ts`. That's what made a *partial* rename possible — and the partial rename is the dangerous one.

## The failure this removes

The id keys four fields in the **user's** `~/.openclaw/openclaw.json` — `plugins.allow`, `plugins.entries.<id>`, `plugins.slots.memory`, `plugins.slots.contextEngine` — and the same string is in `openclaw.plugin.json`, which is what OpenClaw's **loader** reads.

Rename the eleven and leave the manifest alone: it type-checks, the whole suite stays green, and the plugin writes `entries.<new>` into a config OpenClaw still keys as `<old>`. The memory slot is never claimed, and per the docstring on `isContextEngineSlotClaimed`, keystone injection silently dies.

**Only the manifest was pinned** — so the sentinel went green on exactly that half. That made the gate *worse than no gate*, because it looked covered.

## What changed

- `config.ts` declares `export const PLUGIN_ID = "memclaw"` once; all eleven sites reference it, including the two property accesses (now `entries?.[PLUGIN_ID]`) and the operator-facing warning that quotes the slot value.
- `config.test.ts` asserts `PLUGIN_ID` equals `openclaw.plugin.json`'s `id`. **This is the only thing in the build that compares them.**
- The sentinel pins `PLUGIN_ID = "memclaw"` next to the manifest entry it belongs with (35 total), so neither end can move alone.

Two comments keep the old literal deliberately: both narrate a historical bug where the code created `["memclaw"]` from an empty allowlist. They describe what a *past version* wrote, so they're only accurate in the old spelling.

## The new test fails without its fix — both directions

| simulated change | sentinel | suite |
|---|---|---|
| rename the **plugin** side only | exit 1 | new test **fails** |
| rename the **manifest** side only | exit 1 | new test **fails**, and `fail 1` — it is the **only** test of 42 that catches manifest drift |
| restored | exit 0 | 42 pass, 0 fail |

The second row is the point: nothing else in the suite notices, which is why this drift class survived until now.

## Side effect

Collapsing eleven literals into one removes **12 old-brand lines** (`-12 net`) without touching behaviour.

## Verification

| check | result |
|---|---|
| `do_not_touch_sentinel.py` | exit 0, all 35 survive |
| `legacy_name_ratchet.py --base origin/main` | exit 0, no new lines, -12 net, 2 exemptions (both pinned strings) |
| `cd plugin && npm run build` (tsc) | exit 0 |
| `cd plugin && npm test` | **415 tests, 415 pass, 0 fail** (was 414; +1 is the new test) |
| `npm run check:version` | ok |
| `tests/test_do_not_touch_sentinel.py` | 89 passed |
| `ruff check` | clean |

The sentinel entry is placed beside the existing manifest-id entry rather than appended, both because that's its natural home and to avoid a tail conflict with #939.

### DIFF
--- a/plugin/src/config.test.ts
+++ b/plugin/src/config.test.ts
@@ -1,7 +1,8 @@
 import { test, describe } from "node:test";
 import assert from "node:assert/strict";
 import { writeFileSync, readFileSync, mkdtempSync, mkdirSync, rmSync } from "fs";
-import { join } from "path";
+import { join, dirname } from "path";
+import { fileURLToPath } from "url";
 import { tmpdir } from "os";
 import {
   autoFixAllowlist,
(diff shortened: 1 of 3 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #943, opened 2026-08-24
docs(npm): make @caura/client the canonical install name

Eldad-Caura:
Closes the `npm install caura` question in #873 — the answer is that it cannot work, and we should stop waiting for it to.

### The blocker is `csurf`, not `cuara`

Reaching the actual rejection took clearing three unrelated walls (a dead `~/.npmrc` token, then 2FA, then a scope-limited granular token). The real error names a package **no planning document mentioned**:

```
403 Forbidden - PUT https://registry.npmjs.org/caura
Package name too similar to existing package csurf
```

| | `csurf` | `cuara` (what every write-up assumed) |
|---|---|---|
| weekly downloads | **435,324** | 2 |
| status | archived + deprecated by its maintainers | unmaintained |

`caura` is Levenshtein distance **2** from `csurf` and shares three of five positions (`c_ur_`). **That block is the filter working, not a false positive** — typosquats on high-traffic packages are what it exists to stop. The unscoped name is not coming back on any timescale worth planning around, so the scoped name becomes canonical.

### What is deliberately NOT renamed

The implementation package keeps the name `@caura/memclaw-client`. It is pinned in `do_not_touch_sentinel.py` with the reason *"the published npm package name changes under everyone who installed it"* — which is correct. It is live at 1.0.1 with real installs, and rule 3 is alias-never-replace.

So every **forward-facing** surface loses the old brand — install command, docs, new installs — while the old package keeps resolving forever. This mirrors PyPI, where `caura` is a metapackage for `caura-client`.

### Two things worth a reviewer's eye

**A README claim this would have falsified.** It described the TypeScript client as having *"zero dependencies"*. The canonical package now has exactly one — the implementation it re-exports — so it now reads *"no third-party dependencies"*, which is what the sentence meant and is still true.

**The four old-brand lines are MOVES, not mints.** An earlier draft of this PR reformatted them (quote style, a caret bump to `^1.0.1`) and the ratchet correctly failed it as three new mints requiring three permanent exemptions. Keeping the text byte-identical to what left `clients/npm-caura` makes the ratchet pair them as moves instead. **Exemptions are forever; a move is free.** `^1.0.0` resolves to 1.0.1 regardless.

### Gates

- `legacy_name_ratchet.py --base origin/main` → **No new lines, 8 removed (−4 net)**, 4 recognised as moves.
- `do_not_touch_sentinel.py` → exit 0, all protected strings survive, including the pinned `@caura/memclaw-client`.

### Not included

Publishing. `publish-npm-client.yml` **has never run and cannot** — its `NPM_TOKEN` secret does not exist, which is why `@caura/memclaw-client@1.0.0` was uploaded by hand. `@caura/client@1.0.0` needs the same manual publish, and `npm deprecate` on the old name pointing at the new one is worth doing at the same time.


Eldad-Caura:
@claude

Eldad-Caura:
@claude re-review please — force-pushed; the review job skips on `synchronize`.

Fixed the stale setup note, and the finding was better than it looked. **Scope access is not a formality here — it is the exact axis that failed during the publish attempt this morning.**

A granular token scoped to `@caura` returned `cannot publish over the previously published versions` for the implementation package (so it *had* publish rights inside the scope) while returning a bare `You may not perform that action with these credentials` for the unscoped name. Same token, same account, opposite answers — and the second message says nothing about scope. It reads like a name problem, which is what sent me chasing the similarity filter for two rounds before it was actually the block.

So the note now records that, rather than just correcting "no npm org is required" to "an org is required". A future maintainer rotating the token gets the failure mode described, not just the requirement.

One thing I did *not* do: mark the new comment lines `legacy-name-ok`. The first draft named the implementation package literally and the ratchet failed it as two new mints. Per the house convention Lane C established — keep the literal brand out of prose entirely, only pinned values carry markers — I reworded it off the literal instead. Two permanent exemptions for a comment would be a bad trade.

Ratchet `No new lines, 8 removed (−4 net)`; sentinel exit 0; actionlint clean.

Eldad-Caura:
@claude

Eldad-Caura:
@claude re-review please — force-pushed; the review job skips on `synchronize`.

**Renamed the workflow.** You raised this twice, at Low then Medium, and the escalation was right — I under-weighted it first time.

`.github/workflows/publish-npm-caura.yml` → **`publish-npm-metapackage.yml`**.

The collision was worse than "two similar filenames". It was **inverted**: the file named `…-client.yml` publishes the *implementation* (`clients/typescript` → the scoped implementation package), while the file named `…-caura.yml` publishes the *client* (`clients/npm-client` → `@caura/client`). Filename and directory pointed at opposite things. Naming the new one for what it publishes — the metapackage that re-exports — removes the ambiguity without depending on either brand word.

The header now says so explicitly, because the next person to hit a publish failure will read the file before the filename.

**Why I only renamed one of the two.** The sibling is pinned in `do_not_touch_sentinel.py:252` by `path=".github/workflows/publish-npm-client.yml"`, so renaming *it* would fail the sentinel and require the list to move in the same PR. Nothing references this file's name, so renaming it is free. The asymmetric fix is the cheap one and it resolves the confusion either way.

**Tag prefix left as `caura-npm-v*`.** You suggested changing it too. It is already distinct from the sibling's `caura-client-ts-v*`, so it does not contribute to the confusion, and changing a tag pattern is the one thing here that could strand a runbook. (Both workflows have in fact never run, so nothing can be stranded today — but that is an argument for it being harmless, not for it being useful.)

Also rebased: #945 merged mid-review and removed old-brand names across 43 files, so my stale base read as re-adding all of them. Ratchet `No new lines, 8 removed (−4 net)`, sentinel exit 0, actionlint clean.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #944, opened 2026-08-24
fix(ratchet): report exempt lines this change removed

Eldad-Caura:
Closes the gap Lane B demonstrated: a change that destroys a dual-read alias passing both required gates green.

### The defect

`legacy-name-ok` means *"the ratchet may ignore this line"*. It has never meant *"this line is protected"* — but the marker **looks** like protection, so lines carrying it were treated as handled. Deleting one **lowers** the file's non-exempt count, which this gate reports as progress.

Lane B's demonstration on then-current `main`: a blanket rename over two plugin files removed the alias deciding which `.env` keys survive a redeploy. Ratchet `-83 net, no new lines`, exit 0. Sentinel exit 0. **All three of the aliases involved already carried the marker.** That redeploy path rewrites the operator's `.env` as a whole file, so the failure mode is data loss, green.

### Why this reports rather than fails

Measured, not assumed. Across the last 400 commits, **seven removed a marked line and none was a genuine removal of protection** — every one reworded or consolidated markers while keeping the alias:

```
-6/+1   "one alias rule for the repro harnesses"     (six markers → one)
-1/+3   "teach the CAURA_* env names, one surviving"  (net +2)
-1/+1   "the log message is the first argument"       (reworded in place)
```

A rule keyed on text identity would have produced **four false positives and zero true ones**. A rule keyed on per-file counts would have failed the consolidation too. Nothing textual separates "consolidated the alias" from "deleted the alias" — that judgement is irreducibly human, and a gate that cries wolf four times teaches people to stop reading it. So this hands a human the fact instead of guessing the verdict.

### Verified against the original demonstration

Same blanket rename, with this change in place:

```
1 exempt line(s) removed by this change. ``legacy-name-ok`` marks a line
the ratchet ignores, NOT a line that is protected — deleting one lowers the
count and reads as progress. Confirm each alias still exists in some form:
  readEnv(["CAURA_DEPLOY_FAILURE_COOLDOWN_HOURS", "MEMCLAW_DEPLOY_FAILURE_COOLDOWN_HOURS"])
No new lines. 83 removed by this change (-83 net).
```

The destroyed alias is named. The `-83 net` progress line is unchanged — both facts are now visible instead of only the flattering one.

### Implementation note

`Scan` gains `exempt_text` alongside `exempt`. The existing counter is keyed by **path**, which is what selects files for the new-exemption report; the removal report has to name the **line**, since "did this alias survive the rewording" cannot be answered from a count. Both come from the same single read per tree — the scan's "read a tree once" contract is preserved.

### Tests

Four added. Confirmed to fail without the fix rather than assumed: removing the call fails two of them (`2 failed, 47 passed`); restored, `49 passed`. The other two guard the false-positive direction — no report when nothing was removed, and no failure when markers are consolidated.

### What this does not change

Anything whose removal breaks an already-installed user still belongs in `do_not_touch_sentinel.py`, which pins exact text and **does** fail. This marker is not a substitute for that and never was — the report says so where it prints.

### DIFF
--- a/scripts/legacy_name_ratchet.py
+++ b/scripts/legacy_name_ratchet.py
@@ -295,6 +295,7 @@ class Scan(NamedTuple):
     by_file: dict[str, Counter[str]]
     total: Counter[str]
     exempt: Counter[str]
+    exempt_text: Counter[str]
 
     def counts(self) -> Counter[str]:
         """Non-exempt lines per file — what the ratchet holds flat."""
@@ -306,14 +307,20 @@ def scan(tree: str | None) -> Scan:
     by_file: dict[str, Counter[str]] = {}
     total: Counter[str] = Counter()
     exempt: Counter[str] = Counter()
+    # Keyed by text, unlike ``exempt`` which tallies per file. Both are needed:
+    # the per-file tally selects which files the new-exemption report examines,
+    # while the removal report has to name the actual line, since "did this alias
+    # survive the rewording" cannot be answered from a count.
+    exempt_text: Counter[str] = Counter()
     for path, _, text, is_exempt in _grep(tree):
         if is_exempt:
             exempt[path] += 1
+            exempt_text[text.strip()] += 1
             continue
         stripped = text.strip()
         by_file.setdefault(path, Counter())[stripped] += 1
         total[stripped] += 1
-    return Scan(by_file, total, exempt)
+    return Scan(by_file, total, exempt, exempt_text)
 
 
 def _mint_budget(head_total: Counter[str], base_total: Counter[str]) -> Counter[str]:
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@claude re-review please — force-pushed, and the review job skips on `synchronize`.

### What was wrong: I committed my own demonstration

The failures were **my mistake, and an instructive one**. To prove this change catches Lane B's scenario I ran their exact destructive sweep — `perl -pi -e 's/memclaw/caura/gi'` over `educate.ts` and `heartbeat.ts` — then tried to undo it with:

```
git add -A          # ← staged the mutation
...
git checkout -- plugin/src/educate.ts plugin/src/heartbeat.ts
git reset
```

**`git checkout -- <path>` restores from the INDEX, not from HEAD.** The mutation was already staged, so it restored the broken version over itself, and `git reset` then left it sitting in the working tree looking clean. The next `git add -A` swept it into the commit.

That is 12 `tsc` errors (`caura_TOOLS`, `caura_API_KEY`, `iscauraFullyConfigured`…) and a sentinel failure — which is exactly what those two gates are for, and they both did their job. Restored with `git checkout origin/main -- <paths>`, naming the source explicitly.

The irony is worth stating plainly: **the PR that adds a guard against blanket renames was broken by a blanket rename**, and the demonstration it exists to justify is the thing that broke it.

### The diff now

```
scripts/legacy_name_ratchet.py    | 56 +++++++++-
tests/test_legacy_name_ratchet.py | 83 +++++++++++++++++
```

Nothing else. `plugin/` is byte-identical to `origin/main`.

- 49 tests pass; confirmed still failing without the fix (removing the call fails 2 of the 4 new ones).
- Ratchet: `No new lines.`
- Sentinel: exit 0.

The demonstration itself is preserved where it belongs — in the PR body and the commit message — rather than in the tree.

Eldad-Caura:
@erni-a please pprove

erni-a (approved):

PR #945, opened 2026-08-24
chore(plugin): rename the MEMCLAW_* constants and [memclaw] prefix — the env keys stay

yanki-m:
The W5 identifier sweep plus the W4 set-site flip, landed together because the ratchet proved they cannot land apart: renaming only two constants left sibling `MEMCLAW_*` tokens on every touched line — mixed lines that would need editing twice.

**−402 lines net.** The largest single cut of the programme so far.

## What changed

- Every exported `MEMCLAW_*` constant in `plugin/src` → `CAURA_*`. These are **TypeScript identifiers** — the env keys the plugin reads are untouched: every `readEnv(["CAURA_X", "MEMCLAW_X"])` list is byte-identical, markers intact (verified: zero unmarked quoted `MEMCLAW_` strings before the sweep, zero after).
- `isMemclaw*` → `isCaura*`, `memclawPromptSection*` → `caura*`, `MemClawContextEngine` → `CauraContextEngine` **with the old name kept as an exported const+type alias** — `dist/` is consumed outside this repo, so the pre-rename class name is public API.
- `[memclaw]` → `[caura]` on every log line, **including the matcher** at `scripts/gateway_integration_test.py` that greps gateway logs for it — prefix and assertion in the same commit, per the one-cluster rule.
- Tests/e2e now **set** `CAURA_*`; the alias-pinning tests that deliberately set `MEMCLAW_*` keep doing so.
- Teaching prose (cli.py docstring, comments, script banners) teaches `CAURA_*`.

## Untouched, deliberately

On-disk sentinel strings: `educate.ts` matchers, `.memclaw-owned`, the engine `info.id`, gateway RPC method names, every quoted legacy env key. Two new exemptions name the **kept plugin id** a user must literally type into `openclaw.json` (`index.ts:484`, `:755`).

## Verification

- `tsc` clean
- Full 19-file plugin suite: **336 pass / 74 fail** vs **335/75 on untouched `origin/main`** measured in the same shell — one net gain, no regressions (the 74 are the known Windows path fixtures)
- `bash -n` / `node --check` / `ast.parse` clean on every touched script
- Ratchet: `No new lines. 402 removed by this change (-402 net).`

### DIFF
--- a/clients/python/src/caura_client/interviewer/cli.py
+++ b/clients/python/src/caura_client/interviewer/cli.py
@@ -1,9 +1,9 @@
 """memclaw-interviewer CLI — run / status / hook.
 
 Config precedence: flags > env > defaults. Env vars:
-  MEMCLAW_API_KEY (required)      MEMCLAW_TENANT_ID (required)
-  MEMCLAW_BASE_URL                MEMCLAW_AGENT_ID (default user@host)
-  MEMCLAW_FLEET_ID                MEMCLAW_INTERVIEWER_PROJECTS (comma-sep globs)
+  CAURA_API_KEY (required)      CAURA_TENANT_ID (required)
+  CAURA_BASE_URL                CAURA_AGENT_ID (default user@host)
+  CAURA_FLEET_ID                CAURA_INTERVIEWER_PROJECTS (comma-sep globs)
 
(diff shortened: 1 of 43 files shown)


Eldad-Caura:
@claude

Eldad-Caura (approved):

PR #946, opened 2026-08-24
feat(llm): attribute the per-call token log to its service label

arkash20:
## Why

Follow-up to #940 (E3), prep for E4's dollar verification. The token log made per-call cost visible, but attributing a line to its **service** (contradiction judge vs enrichment vs dedup vs recall) still required correlating adjacent log lines or clustering by prompt size. The cost-baseline window Erni is about to run needs cleanly attributed data from day one.

## What

- `call_with_retry` publishes its per-tier label (e.g. `contradiction_batch-primary` / `-fallback`) through a new `ContextVar` in `common/llm/call_context.py` — set on entry, reset in `finally` so it never leaks into a sibling task's lines.
- `OpenAILLMProvider.complete_json`/`complete_text` render it on the existing token line: `... tokens_in=N tokens_out=N tokens_reasoning=N service=<label>`.
- Zero call-site, protocol, or provider-signature changes (a parameter would have touched dozens of signatures and every test double). Calls outside the retry layer log `service=-`, keeping the log shape stable for a regex-extracted log-based metric.
- The tier suffix is kept deliberately: a cost regression can then be attributed not just to a service but to a misfiring fallback tier.

## Tests

4 new tests: direct call logs placeholder; label flows end-to-end through `call_with_fallback`; label resets after the call (no leak into subsequent unlabeled calls); reset holds when the call raises. Full suite: 5,046 passed.

### DIFF
--- a/common/llm/call_context.py
+++ b/common/llm/call_context.py
@@ -1,0 +1,28 @@
+"""Ambient label for the LLM call currently in flight (E4-prep).
+
+The per-call token log (E3, ``common/llm/providers/openai.py``) is the
+cost-attribution surface — but the provider is invoked through ``call_fn``
+closures built by ``call_with_fallback`` and never sees the caller's
+``service_label``. Threading a parameter through every provider method,
+protocol, and test double would touch dozens of signatures for one log
+field; a ``ContextVar`` instead rides the await chain from the retry
+layer down to the provider with no call-site changes.
+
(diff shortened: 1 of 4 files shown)


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #947, opened 2026-08-24
CAURA-718 fix(doc): render the whole doc payload into the minted memory

eyal-bl:
CAURA-704 minted the memory from ``data["content"]``, falling back to ``data["body"]``, and skipped the write when neither was present. In production at eToro that meant the feature could never fire.

Their only organic doc feed (``meeting-summaries``) writes ``{url, date, title, source, summary, platform, decisions, action_items, participants, duration_minutes}``. No ``content``. No ``body``. Every write returned 200, minted nothing, and logged the skip at DEBUG — so the feed looked healthy while the feature it was built for never ran once. Verified in prod: six docs written 2026-08-20, zero memories, all six "expected skip: no body".

The field guess was never defensible. ``documents.data`` has NO schema — it is free-form JSONB (``data: dict``, no required keys; even ``{}`` is a legal document), and the tool description says the body "lives wherever the caller puts it (e.g. data['content'])", offering ``content`` as an EXAMPLE rather than a contract. Nothing obliges a producer to use it and nothing validates it. A field-name list cannot be made correct, only longer.

So the memory now carries the WHOLE ``data`` payload rendered as text, with ``summary`` first. There is no body-field requirement: a document mints whenever it carries any usable data.

Why summary first, specifically: ``caura_insights`` truncates memory content to 500 chars (``sanitize_content``'s default), so whatever leads the render is what the reflection pass actually sees. A raw ``json.dumps(data)`` would spend ~250 of those chars on eToro's ``url``/``date``/``title``/``source`` and push ``decisions``/``action_items`` past the cut. Rendering summary-first puts the information-dense field inside every consumer's window — for their real doc shape the entire render is ~370 chars, so insights now sees the decisions and action items rather than the metadata.

Format is ``key: value`` lines rather than JSON because this text is embedded and fed to LLMs, for which braces and quotes are tokens that carry no meaning. ``summary`` is emitted as bare prose (no label) so the prefix reads naturally. String values pass through unmodified; lists/dicts become compact JSON with ``ensure_ascii=False`` so names and non-English text stay legible. ``default=str`` keeps a non-serialisable value from raising, since nothing here may fail the doc write. Empty values are dropped — they waste prefix budget.

Two accepted consequences, both deliberate:

- Bulk structured records (``{"plan": "business", "seats": 40}``) now mint, where the old rule excluded them as a side effect of requiring a body field. A sync writing many small documents therefore produces one memory per document. That is the cost of not guessing.
- "content is byte-identical to a field" becomes "lossless render of the record". Byte-fidelity was only ever a means to *don't truncate, don't paraphrase*, which the render preserves — every non-empty field is present and strings are untouched.

Server-written collections are unaffected: ``ingest-sources``, ``interview_jobs`` and ``skills_rollback`` are written through ``sc.upsert_document`` directly and never reach this code path, so no ingested content is duplicated. ``skills`` and ``_``-prefixed collections keep their skips — minting a skill body would route around the staged -> active approval lifecycle.

Reviewer note, untested in production: the memory content is what ``BusinessPersonalPregate`` classifies, and eToro runs ``non_business.disposition="drop"`` with ``pregate.enabled`` and ``fail_closed=true``. The render now also includes ``participants`` — real names — which pushes the classifier marginally toward "personal", and a drop is swallowed by ``safe_sync_doc_memory`` so the doc write still returns 200. Worth pushing one real meeting doc through the pregate before rollout; excluding ``participants`` from the render sidesteps it at the cost of losing who-was-there from recall.

Also fixes the size cap to measure the RENDER rather than one field — the ``key: `` labels and separators make the render slightly longer, so a doc can be under the cap on its body and over it once rendered.

## Type of Change

- [ ] Documentation update
- [ ] Other:

## Checklist

- [ ] `mypy` passes
- [ ] `pytest` passes locally

### DIFF
--- a/core-api/src/core_api/constants.py
+++ b/core-api/src/core_api/constants.py
@@ -178,10 +178,13 @@ MAX_QUERY_LENGTH = 5000
 # ``memories.content`` directly). Docs themselves are only searchable via their
 # ``data["summary"]`` embedding, and ``caura_recall`` never returns documents.
 #
-# The memory content is the doc body VERBATIM, so the cutoff is simply the
-# memory schema ceiling: mint whenever the body fits in a memory at all.
-# Over-cutoff bodies are SKIPPED rather than truncated — a truncated body reads
-# as complete and would produce confidently wrong downstream conclusions.
+# The memory content is the whole ``data`` payload rendered as text (CAURA-717),
+# so the cutoff is simply the memory schema ceiling: mint whenever the RENDER
+# fits in a memory at all. Note the render is slightly longer than any one field
+# — ``key: `` labels and blank-line separators — so a doc can be under the cap
+# on its body and over it once rendered. Over-cutoff payloads are SKIPPED rather
+# than truncated: a truncated payload reads as complete and would produce
+# confidently wrong downstream conclusions.
 DOC_MEMORY_MAX_CHARS = MAX_CONTENT_LENGTH
 
 # NOTE: there is deliberately no ``DOC_MEMORY_TYPE``. Doc-derived memories pass
--- a/core-api/src/core_api/services/doc_indexing.py
+++ b/core-api/src/core_api/services/doc_indexing.py
@@ -19,18 +19,22 @@ Embed-source contract (``resolve_embed_source``):
   path).
 
 Doc-memory contract (``resolve_doc_memory``):
-- A doc write also mints a memory carrying the document body verbatim,
-  because the two stores are not cross-searched: ``caura_recall``
-  never returns documents, and only ``data["summary"]`` is embedded on
-  the document row. The minted memory is what makes the *body*
-  reachable by meaning.
+- A doc write also mints a memory carrying the document's data, because
+  the two stores are not cross-searched: ``caura_recall`` never returns
+  documents, and only ``data["summary"]`` is embedded on the document
+  row. The minted memory is what makes the *content* reachable by
+  meaning.
+- CAURA-717: the memory carries the WHOLE ``data`` payload rendered as
+  text, ``summary`` first — not one guessed field. See the note above
+  ``_render_doc_data`` for why guessing could not be made correct.
 - Independent of the embed rule: a doc with no ``summary`` is stored
   unindexed (invisible to ``op=search``) yet still mints a memory, so
-  recall can reach the body even when doc-search cannot.
+  recall can reach the content even when doc-search cannot.
 """
 
 from __future__ import annotations
 
+import json
 import logging
 from dataclasses import dataclass
 from datetime import datetime
@@ -41,11 +45,23 @@ logger = logging.getLogger(__name__)
 
 SKILLS_COLLECTION = "skills"
 
-# Field names checked, in order, for the document body. ``content`` is the
-# convention advertised in the ``caura_doc`` tool description ("The full body
-# lives wherever the caller puts it (e.g. data['content'])"); ``body`` is
-# accepted as the obvious synonym.
-_BODY_FIELDS: tuple[str, ...] = ("content", "body")
+# CAURA-717: there is deliberately no list of "body" field names any more.
+#
+# The previous rule looked for ``data["content"]`` then ``data["body"]`` and
+# skipped the mint when neither was present. That was a guess against a store
+# with NO schema: ``documents.data`` is free-form JSONB (``data: dict``, no
+# required keys — even ``{}`` is a legal document), and the tool description
+# says the body "lives wherever the caller puts it (e.g. data['content'])",
+# offering ``content`` as an EXAMPLE rather than a contract.
+#
+# In production at eToro the only organic doc feed writes
+# ``{url, date, title, source, summary, platform, decisions, action_items,
+# participants, duration_minutes}`` — no ``content``, no ``body``. Every write
+# returned 200 and silently minted nothing, so the feature could never fire for
+# the one feed that had real content. Guessing field names cannot be made
+# correct, only longer.
+#
+# The whole ``data`` payload is rendered instead — see ``_render_doc_data``.
 
 
 class InvalidDocIndexingError(ValueError):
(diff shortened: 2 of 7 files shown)


eyal-bl:
@claude

eyal-bl:
Both verified against the code. **One is a real bug — fixed and pushed. One is correct but is the deliberate design decision this PR exists to make.**

---

### Whitespace-only strings not treated as empty — **correct, fixed**

Reproduced exactly:

```python
resolve_doc_memory("notes", "n1", {"summary": "Real summary.", "ws": "   \n\t  ", "real": "kept"})
# -> 'Real summary.\n\nws:    \n\t  \n\nreal: kept'
```

A bare `ws:` followed by blanks — precisely the noise the filter exists to remove, and contradicting the comment two lines above it. It was also inconsistent inside its own function, since `summary` already used `.strip()`.

Extracted the check into `_is_empty()`:

```python
if value is None:
    return True
if isinstance(value, str):
    return not value.strip()
return value == [] or value == {}
```

Deliberately **not** a truthiness check: `0` and `False` are meaningful field values (`duration_minutes: 0`, `archived: false`) and must still render. Added tests for both directions — 4 whitespace cases and 3 falsy-but-real cases — and confirmed the whitespace ones **fail against the pre-fix source**, so they actually pin the bug rather than just passing.

Good catch: my own `test_empty_values_are_omitted_from_the_render` only covered `""`, which is why this got through.

---

### Whole-payload capture widens what becomes recallable — **correct, and deliberate**

Accurate, and it is the central decision of this PR rather than an oversight. It was made explicitly after the production finding and is recorded in the commit message under "Two accepted consequences, both deliberate".

The trade, stated plainly: the previous `content`/`body` whitelist wasn't a privacy control, it was a **guess against a schemaless store**. `documents.data` has no schema — no required keys, `{}` is a legal document — and the tool description offers `data['content']` as an *example* ("the full body lives wherever the caller puts it"), not a contract. That guess silently excluded eToro's only organic doc feed: six documents written 2026-08-20, zero memories, every write returning 200. A field-name list can't be made correct, only longer.

Two things that bound the exposure:

1. **This changes discoverability, not accessibility.** Every one of these fields is already stored in `documents.data` and readable today via `caura_doc op=read`. The change is that they become *semantically* findable. Real — recall surfaces things you didn't know to ask for — but it isn't new data exposure.
2. **Governance already inspects this content.** `GovernanceScanContent` and `BusinessPersonalPregate` run on the memory content before the row is written, so tenant PII policy and business/personal disposition apply to the rendered payload. Note eToro's PII action is `flag`, not `mask` — it annotates rather than redacts.

A per-collection opt-in or deny-list is a reasonable future addition and I'd support it, but it's deliberately out of scope here: adding one now would reintroduce a curated-key list, which is the thing that failed in production.

**One related risk I want visible rather than buried** (also in the commit body): the render now includes `participants` — real names — and that text is exactly what `BusinessPersonalPregate` classifies. eToro runs `disposition="drop"` with `fail_closed=true`, and a drop is swallowed by `safe_sync_doc_memory`, so the doc write still returns 200 while the memory silently doesn't exist. Untested against a real meeting doc. Worth pushing one through the pregate before this reaches that tenant; excluding `participants` from the render sidesteps it at the cost of losing who-was-there from recall.

eyal-bl:
@claude

eyal-bl:
Taking option **1** — and conceding the specific gap you identified, which is fair.

### The documentation gap is real

The docstring lists "two consequences of not guessing, both accepted" and both are about **volume and duplication**:

- bulk structured records now mint, so a many-document sync produces one memory per document
- server-written collections (`ingest-sources`, `interview_jobs`, `skills_rollback`) are unaffected because they bypass this path

Nowhere does it say *"every field in a schemaless payload becomes semantically searchable."* That reasoning lives in the commit body and a PR comment — neither of which a future reader of `doc_indexing.py` will see. The code should carry it. I'll add that note to the `resolve_doc_memory` commentary.

### Confirming it's an accepted, reviewed trade-off

It is — decided explicitly after the production finding, not inherited by accident. Two things bound it, and both belong in that note:

**Discoverability, not accessibility.** Every one of these fields is already stored in `documents.data` and readable today via `caura_doc op=read`. What changes is that they become findable *by meaning*. That is a genuine widening — recall surfaces things you didn't know to ask for, and LLM context is a different exposure surface than an explicit keyed fetch — but it is not new data reaching a new party.

**Governance already inspects this exact text.** `GovernanceScanContent` (PII scan → per-tenant mask/drop) and `BusinessPersonalPregate` both run on the rendered memory content *before* the row is written. So there is already a tenant-configurable control on this path; it simply isn't a hardcoded key list.

### On option 2 (denylist) — declining, but not for the reason I gave last round

I previously waved this off as "reintroducing the curated list that failed." That was glib and I'll correct it: the failure modes are opposite, not identical. An allowlist that misses a key **breaks the feature silently**. A denylist that misses a key **leaks silently but the feature still works**. A denylist is strictly additive protection, so "it's the same mistake" isn't a real argument.

The reasons I'd still leave it out here:

1. **A five-name list is false comfort.** `token`/`secret`/`password`/`api_key`/`ssn` misses `auth_header`, `bearer`, `credentials`, `private_key`, `access_token`, `national_id`, and every domain-specific variant. Shipping it invites people to believe key-level redaction exists when it mostly doesn't.
2. **The real control already exists and is tenant-scoped.** Key-level redaction, if we want it, belongs next to `governance.*` where a tenant can configure it — not as a constant in a derivation helper, where no operator can see or change it.
3. **It doesn't address the concrete instance we actually have.** See below.

If the team wants field-level redaction, I'd rather open it as its own ticket against the governance surface than bolt a partial list into this function.

### The concrete instance of your concern, and it's untested

The sharpest real-world case isn't a hypothetical `api_key` — it's already in the eToro payload this PR was written for. The render now includes `participants`, a list of real names, and that text is exactly what `BusinessPersonalPregate` classifies. eToro runs `non_business.disposition="drop"` with `pregate.enabled` and `fail_closed=true`, and a drop is swallowed by `safe_sync_doc_memory` — so the doc write still returns 200 while the memory silently doesn't exist.

That is the same class of invisible failure this PR exists to fix, arriving through a different door, and it has never been exercised against a real meeting document. Flagging it here as the thing I'd want checked before this reaches that tenant. Excluding `participants` from the render sidesteps it, at the cost of losing who-was-there from recall.

Eldad-Caura (approved):

PR #948, opened 2026-08-24
ci(mypy): put common/ under the type gate, and fix what it was hiding

Eldad-Caura:
## The gap

`common/` is type-checked by **nothing** on `origin/main`.

CI runs mypy four times (ci.yml:233, :236, :242, :248). Each one `cd`s into a
single service and checks that service's own `src/`, so no invocation's path
ever resolves to `common/`. Lint does cover it — `ruff check common/` at :167 —
so the directory looks gated at a glance.

The service runs don't cover it transitively either. 40 files under
`core-api/src` import `common`, but every service sets
`ignore_missing_imports = true` with a `mypy_path` of just its own `src/`, so
`common/` sits outside the search path and each `common.*` import degrades to
`Any` in silence. Measured: `cd core-api && mypy src/` reports **199 source
files** and none of them are `common/`'s 94.

Nor does the hook. `.pre-commit-config.yaml:23-27` deliberately runs no mypy and
defers to CI as *"the authoritative type-check"* — a deferral pointing at a
check that did not include this directory.

## What it was hiding

**12 errors in 7 files**, measured on `origin/main` before anything changed.
All 12 are fixed here; `common/` is clean at the new gate.

Two are worth calling out on their own. `# type: ignore[arg-type]` in
`llm/providers/openai.py` and `# type: ignore[import-untyped]` in
`events/pubsub.py` both name error codes mypy does **not** report at those lines
(an overload failure, and a namespace-package attribute). Neither suppressed
anything. Nobody could have known — nothing had ever run mypy over these files
to say so.

The other ten are declared types too narrow or too loose to verify the code
around them:

| file | was | fix |
|---|---|---|
| `ranking/providers/local.py`, `embedding/providers/local.py` | `self._model = None` unannotated → mypy infers the attribute's type as `None`, so every `.predict`/`.encode` errors against `None` (and in the ranking case the inference flows backwards into the model-loading lambda) | annotate `Any` — what a lazily-imported optional dep can honestly claim |
| `embedding/_service.py` | resolver returned `object \| None` though the registry it calls already returns `EmbeddingProvider`; 3 call sites couldn't see `embed` | narrow to the protocol; the `InstructionAwareEmbedder` dispatch now binds the narrowed provider instead of a bool, so `embed_query` is checked rather than assumed |
| `ranking/providers/remote.py`, `enrichment/service.py` | guards mypy can't follow — a raise on a separately-built failure list; a dict subscript used as both guard and value | narrow at the binding |
| `enrichment/service.py` | heuristic classifier passed bare strings for `memory_type` | use the `MemoryType` members, so the 13th branch someone adds is checked rather than trusted |

No `# type: ignore` is added without a reason on the line, and no existing check
is loosened. The `memory_type` literals were all valid members already — that
one buys future checking, not a bug fix.

## Why `common/mypy.ini`

Sibling of `common/ruff.toml`, and it exists for the same reason that file does:
`common/` has no `pyproject.toml` and neither repo has a root `[tool.mypy]`.
Unlike ruff, **mypy resolves config from the CWD rather than walking up from the
files it checks**, so `--config-file` is required rather than stylistic. Drop the
flag and the step runs at mypy's bare defaults and reports 17 — the 12 plus 5
missing-stub errors for optional deps — so the failure mode is a noisier gate,
not an absent one.

Its settings are the `[tool.mypy]` block all four services already carry,
verbatim and with nothing added, so `common/` is checked on the same terms as the
`src/` trees that import it. `python_version` is py312 — the lower bound of the
two trees — for the reason `common/ruff.toml` pins `target-version` the same way.

## Cross-repo and cross-lane notes

- **`caura-enterprise` gets the same gate in its own PR.** There, `common/`
  measured **clean (0 errors, 33 files)**, so it is a CI-step change only. That
  copy of `common/mypy.ini` will be classified `identical` in
  `vendored_files_manifest.json`, and `check_vendored_drift.py` fails an
  `identical` entry that is missing upstream — **so this PR must land first.**
  (The reverse direction is safe: the gate's unclassified-file check iterates the
  enterprise tree, so an OSS-only new file cannot wedge enterprise CI.)
- **Lane P is mid-cutover in `common/events/`.** This touches
  `events/pubsub.py` — one comment line at :261, correcting that ignore code.
  Lane P's hunks on that file are at ~48/126/156/439/497, so there is no textual
  overlap, but flagging it rather than letting it surface in a merge.
- Only 10 files are vendored into enterprise (4 `identical`, 6 `manual` =
  `events/*`). `events/pubsub.py` is `manual` policy and already divergent
  between the trees, so this change does not need a matching enterprise edit.

## Verification

Run from the repo root with CI's own env (`uv pip install -e` all four services):

- `mypy --config-file common/mypy.ini common/` → clean (94 files); was 12 errors
- `ruff check common/` → clean
- `ruff format --check --isolated common/{__init__,structlog_config,serve}.py` → clean
- all four service mypy steps → unchanged, clean (199 / 83 / 5 / 11 files)
- `pytest tests/` filtered to enrich/embed/rank/pubsub/event/llm → 888 passed, 1 skipped
- the 7 test files exercising `fake_enrich` → 138 passed
- `actionlint` → same one pre-existing finding as `origin/main`, none added
- `ruff check` and `ruff format` run separately and scoped; `ruff format` was not
  run over `common/`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -247,6 +247,37 @@ jobs:
         # along, nothing ever ran it.
         run: cd core-worker && uv run mypy src/
 
+      - name: Run mypy (common/)
+        # common/ was type-checked by NOTHING before this step. Each mypy step
+        # above ``cd``s into one service and checks that service's own src/, so
+        # no invocation's path ever resolved to common/ — while 40 files under
+        # core-api/src alone import it. Those imports were not caught either:
+        # ``ignore_missing_imports = true`` (every service sets it) plus a
+        # ``mypy_path`` of just that service's src/ means an unresolved
+        # ``common.*`` import degrades to Any in silence rather than failing.
+        # ``cd core-api && mypy src/`` reported 199 files and none of them were
+        # common/'s 94.
+        #
+        # Nor did the hook cover it: .pre-commit-config.yaml deliberately runs no
+        # mypy, deferring to "CI is the authoritative type-check" — which pointed
+        # at a check that did not include this directory.
+        #
+        # What it was hiding: 12 errors in 7 files, all fixed in the commit that
+        # added this step. Two were ``# type: ignore[...]`` comments naming an
+        # error code mypy does not report here (``arg-type`` for an overload
+        # failure, ``import-untyped`` for a namespace-package attribute), so they
+        # suppressed nothing — nobody could have known, because nothing ever ran
+        # mypy over these files to say so.
+        #
+        # ``--config-file`` is required, not stylistic: common/ has no
+        # pyproject.toml and the repo root has no [tool.mypy], and mypy resolves
+        # config from the CWD rather than walking up from the files it checks
+        # (ruff does walk up, which is why common/ruff.toml needs no flag). Drop
+        # the flag and this runs at mypy's bare defaults, which reports 17 —
+        # the 12 plus 5 missing-stub errors for optional deps — so the failure
+        # mode is a noisier gate, not an absent one. See common/mypy.ini.
+        run: uv run mypy --config-file common/mypy.ini common/
+
       - name: Enable pgvector extension
         run: PGPASSWORD=changeme psql -h localhost -U memclaw -d memclaw -c "CREATE EXTENSION IF NOT EXISTS vector;"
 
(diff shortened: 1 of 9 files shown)


Eldad-Caura:
@erni-a ready for your approval — all checks green (CI, dco, ratchet, sentinel, CodeQL, security scan) and the Claude review came back clean with no issues found.

One sequencing note for whoever merges: the enterprise half is open as a draft at caura-ai/caura-enterprise#1291 and **cannot go green until this lands**, because `check_vendored_drift.py` fails an `identical` manifest entry that is missing upstream. Once this is merged, that PR just needs marking ready for review.

erni-a (approved):

PR #949, opened 2026-08-25
fix(apm): stop Pub/Sub pull timeouts from reporting as error spans

Eldad-Caura:
Ports the fix already merged as [caura-enterprise #1287](https://github.com/caura-ai/caura-enterprise/pull/1287) to this repo's copy of `common/events/pubsub.py` — which turns out to be where nearly all of the volume actually comes from.

## Problem

An idle Pub/Sub subscriber emits **one error span per empty pull window**, forever.

`PubSubEventBus._pull_loop` calls `subscriber.pull(timeout=...)` and already treats `DEADLINE_EXCEEDED` as the non-event it is:

```python
except gexc.DeadlineExceeded:
    # No messages in the pull window; loop back and try again.
    continue
```

ddtrace's gRPC integration marks the span failed on *any* non-OK status before that `except` ever runs — `contrib/internal/grpc/client_interceptor.py`, `if response_code != grpc.StatusCode.OK` → `span.error = 1` — with no per-status config to opt out. So the application handles the timeout correctly and APM records a failure anyway.

**Measured over 24h**, `@error.type:StatusCode.DEADLINE_EXCEEDED` on `/google.pubsub.v1.Subscriber/Pull`:

| service | env | spans |
|---|---|---|
| core-api | staging | 51,568 |
| core-api | prod | 617 |
| core-worker | prod / staging | 99 / 90 |

An idle deployment is far noisier than a busy one — a pull that returns messages succeeds — which is why staging dwarfs prod. Even so, in prod this was the single largest source of error spans (1,223 of ~3,163), so error rate and any monitor on it were reporting mostly poll timeouts.

## Fix

A trace processor that clears `span.error` for that exact resource + status, and nothing else.

- **Fixes the metric, not just the display.** `SpanAggregator.on_span_finish` chains `user_processors` ahead of sampling and the writer, so the payload the Agent computes APM stats from already carries `error: 0`.
- **The span is kept, not dropped.** A pull that fails for a real reason still arrives as an error, and poll timing stays visible. `error.*` tags are left in place so the timeouts stay queryable on purpose (`@error.type:StatusCode.DEADLINE_EXCEEDED`) rather than merely absent.
- **`StatusCode.CANCELLED` is deliberately untouched** — a pull interrupted by shutdown, which I observed for real alongside the timeouts. It says something true about the process.
- **Registration is additive.** `configure(trace_processors=...)` replaces the user-processor list rather than appending, so a bare call would silently evict any processor registered before us. `install()` reads what's registered and passes it back through.
- **`install()` never raises.** Both the private-aggregator read and the `configure()` call are guarded — a signature change across a major ddtrace bump would otherwise propagate out of `PubSubEventBus.start()` and fail service boot. Correct error rates aren't worth trading a service that starts.
- **Registered inside `start()`'s `topic_count > 0` branch** — the spans come from the pull loop, so a publisher-only bus can't produce one and shouldn't pay for registration (which recreates the trace writer).

There is no gRPC status→error config knob in ddtrace 4.x (nothing like `DD_TRACE_HTTP_SERVER_ERROR_STATUSES` for gRPC; `config.grpc` carries no error-status field), so a user trace processor is the only supported altitude.

## Tests deliberately do not require ddtrace

Worth calling out, because it differs from the enterprise version. CI installs only each service's `[dev]` extra, never `datadog` — so a module-level `pytest.importorskip("ddtrace")` would skip the whole file in the one environment that gates merges, shipping this with no coverage at all.

The filter only touches `.error`, `.resource` and `.get_tag()`, so a stub span covers it. One additional test pins the stub to the real `Span` shape (and asserts `ddtrace.constants.ERROR_TYPE == "error.type"`) whenever the extra happens to be installed — skipped in CI by design, so stub drift gets caught locally rather than never.

16 tests plus that fidelity check: the clear; the trace returned rather than dropped; five must-stay-an-error cases (`PERMISSION_DENIED`, `NOT_FOUND`, `CANCELLED`, Publish timeout, `StreamingPull`); a missing `error.type` tag; `install()`'s idempotency guard; an already-registered processor surviving; a renamed private aggregator; an absorbed `configure()` failure; and a publisher-only bus not registering at all. **Each was confirmed to fail without its fix.**

## Verification

- `16 passed, 1 skipped` for the new file, run in a venv **without** ddtrace — i.e. the CI configuration
- `5075 passed, 4 skipped, 1 xfailed` for the full repo-root suite
- `ruff check common/` clean; `ruff format --check` clean; `mypy --config-file common/mypy.ini common/` reports only two pre-existing `dateutil` stub errors in `common/enrichment/service.py`, none in the new code

### DIFF
--- a/common/events/pubsub.py
+++ b/common/events/pubsub.py
@@ -47,6 +47,7 @@ from typing import Any
 
 from pydantic import ValidationError
 
+from common.events import trace_filters
 from common.events.base import Event, EventBus, EventHandler
 from common.events.topics import publish_name, subscribe_names, unbound_publish_topics
 
@@ -569,6 +570,24 @@ class PubSubEventBus(EventBus):
         # gRPC channel. publish() handles the off-loop construction
         # lazily on the first call (see ``_ensure_publisher``).
         if topic_count > 0:
+            # Inside this branch, not at the top of start(): the spans this
+            # corrects come from the pull loop, so a publisher-only bus can
+            # never produce one and shouldn't pay for the filter — and
+            # registering it calls tracer.configure(), which recreates the
+            # trace writer.
+            #
+            # Registered here rather than at import for the same reason.
+            # Idempotent, never raises, and a no-op when ddtrace isn't
+            # installed. See trace_filters for the measurement that motivated
+            # it. Logged either way so "pull spans are still showing as errors"
+            # is one grep from the answer instead of a guess about whether the
+            # filter ever loaded. The negative branch doesn't name a cause:
+            # False means ddtrace is absent OR registration failed, and
+            # install() warns about the latter itself.
+            if trace_filters.install():
+                logger.debug("Pub/Sub pull-timeout span filter registered")
+            else:
+                logger.debug("Pub/Sub pull-timeout span filter not registered")
             loop = asyncio.get_running_loop()
             # Snapshot the stop generation BEFORE the shielded await.
             # A clean ``stop()`` resets both ``_stopped`` and
--- a/common/events/trace_filters.py
+++ b/common/events/trace_filters.py
@@ -1,0 +1,152 @@
+"""APM span corrections for Pub/Sub's pull loop.
+
+``PubSubEventBus._pull_loop`` calls ``subscriber.pull(timeout=...)``. When no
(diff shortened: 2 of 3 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #953, opened 2026-08-25
fix(usage): bill recalls against the recall counter, flag-gated

arkash20:
## What

Closes canonical **D13** (register MEM-21/N1, found in the 2026-08 AX-audit verification pass): plans have carried separate `searches`/`recalls` limits since the initial schema and the platform hook maps `"recall"` → the recalls counter — but both recall call sites (REST `/recall`, MCP `caura_recall`) passed `"search"`. The recalls counter never moved and the per-plan recall cap could never fire.

- Both call sites now bill via `recall_operation()`.
- **Flag-gated, default OFF** (`METER_RECALL_AS_RECALL`): the recalls counter feeds over-plan enforcement (`_is_over_plan_limits` → `x-org-read-only` → 403 on write routes), so making the cap computable for the first time is a deliberate billing decision, not a deploy side effect. With the flag off, billing is byte-identical to before.
- Regression-guard test asserts the bare `"search"` literal cannot silently return to either recall site.

## Testing

- `ruff` / `mypy` clean; full suite **5157 passed** incl. 5 new tests (`tests/test_d13_recall_metering.py`).
- Local-stack e2e wet test against `tenant_usage_counters`:
  - flag OFF: 5 recalls + 2 searches → all billed `search`, zero `recall` rows (legacy behavior preserved exactly);
  - flag ON: 3 recalls + 1 search → `recall = 3`, `search +1` — the recalls counter moved for the first time.

## Rollout note

Enabling the flag on hosted should be coordinated with billing: free-tier tenants recalling >500/mo would become over-plan at the next evaluation.

### DIFF
--- a/core-api/src/core_api/config.py
+++ b/core-api/src/core_api/config.py
@@ -354,6 +354,13 @@ class Settings(BaseSettings):
     paddle_business_monthly_price_id: str | None = None
     paddle_business_annual_price_id: str | None = None
     use_stm: bool = False
+    # D13 — meter /recall and MCP caura_recall against the "recall" counter
+    # instead of "search". Off by default because the recalls counter feeds
+    # plan enforcement (`_is_over_plan_limits` → x-org-read-only → 403 on
+    # write routes): flipping this makes the per-plan recall cap computable
+    # for the first time, so enabling it is a deliberate billing decision,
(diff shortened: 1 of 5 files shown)


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #954, opened 2026-08-25
fix(entities): make the entities read path work — search filters + relations

arkash20:
## What

Closes canonical **C22 + C23** (register MEM-12/N2+N3) — the two structural bugs behind "trivial to populate, impossible to use well" from the SupportHive v2 and Capability Review field reports.

**C22 — entity search was non-functional.** `GET /entities` declared `search` and `entity_type` since day one but never forwarded them: core-api dropped both on the storage-client call and the storage router didn't accept them, so `?search=foo` silently returned the unfiltered list — while the storage service (`entity_list`) supported both filters all along. Now forwarded through every hop.

**C23 — relations were structurally `[]` since v1.0.0.** `get_entity` read `entity.get('relations', [])` off the with-memories payload, which has never carried a relations key — so REST and MCP showed no relations while /graph showed the edges. Relations now come from the (previously unused) `GET /entities/{id}/relations` storage endpoint, tenant-scoped, and the S5/C14 evidence-visibility filter finally operates on real rows. Endpoint failure degrades to `[]` rather than failing the entity read.

## Testing

- `ruff`/`mypy`/format clean on both services; root suite **5164 passed** (7 new tests in `tests/test_c22_c23_entities_read.py`, incl. grep-guards on both formerly-dropped hops); storage suite 154 passed; S5 scope tests updated to feed relations from the new source and still pass.
- Local-stack e2e wet test on a clean tenant, fresh images: **11/11** — every filter combination (search hit/miss, entity_type, combined AND, fleet_id unchanged), `entity_get` returning real relations with correct fields, a freshly-upserted relation immediately visible (the exact WR-5 repro), and /graph agreeing with entity_get on edge count.
- Dirty-tenant run also exercised: filters behave correctly against 84k pre-existing entities (and surfaced an unrelated pre-existing /graph 500 on full-graph serialization at that scale — known graph-blowup class, untouched here).

### DIFF
--- a/core-api/src/core_api/clients/storage_client.py
+++ b/core-api/src/core_api/clients/storage_client.py
@@ -1371,12 +1371,18 @@ class CoreStorageClient:
         self,
         tenant_id: str,
         fleet_id: str | None = None,
+        entity_type: str | None = None,
+        search: str | None = None,
         limit: int = 100,
         offset: int = 0,
     ) -> list[dict]:
         params: dict[str, Any] = {"tenant_id": tenant_id, "limit": limit, "offset": offset}
         if fleet_id is not None:
             params["fleet_id"] = fleet_id
+        if entity_type is not None:
+            params["entity_type"] = entity_type
+        if search is not None:
+            params["search"] = search
         return await self._get_list("/entities", **params)
 
     async def count_memories_per_entity(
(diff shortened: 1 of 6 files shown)


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #955, opened 2026-08-25
fix(forge): the distill cron never ran, and the dry-run CLI never used an LLM

Eldad-Caura:
> **Production impact, and it is the headline.** The forge distill cron has never been able to run a tick. Started as a dev-script repair (spun out of caura#948); the same broken wiring turned out to be in the production cron handler, so both are fixed here — one bug, one fix, two call sites.

## The bug

Both `cron_handler._wire_llm_fn` and `scripts/forge_dry_run.py::_wire_llm_fn` did:

```python
from common.llm import LLMRequest, call_with_fallback   # LLMRequest does not exist
```

`LLMRequest` has never been exported by `common.llm` and is defined nowhere under `common/`. So the import **always** raised `ImportError`, and each file's `except` clause always ran:

| | what the `except` does | consequence |
|---|---|---|
| **`cron_handler.py`** (production) | `raise RuntimeError("forge_cron: common.llm not importable — the production cron requires a configured LLM provider chain")` | **every tick dies at the wiring step.** The distill cron cannot run at all, and the error blames a missing deployment dependency rather than a wrong import — so the symptom points away from the cause |
| **`forge_dry_run.py`** (CLI) | `return _fake_llm_fn` | **the CLI has never once used a real LLM.** It degraded silently on every invocation, logging a warning about "a packaging variant without the LLM provider chain" |

The fake fallback is a real, documented feature of the CLI. The bug is that it was reached *unconditionally* — which is precisely what kept this invisible.

Both then called `call_with_fallback(request, expecting="json")`, matching no signature it has ever had — the real one is `(primary_provider_name, call_fn, fake_fn, tenant_config=None, *, service_label="", ...)` with no `expecting` parameter. Dead code in the CLI; unreachable in the cron.

**Third defect, CLI only:** `run_forge_distill(db, ...)` passed a session positionally to a keyword-only function → `TypeError: too many positional arguments` on every real invocation. Verified: `inspect.signature(run_forge_distill)` has **zero** positional parameters. A comment in `cron_handler` still described a "now-vestigial first positional arg … kept for CLI / test-call-site compatibility" — which is what the CLI was relying on. Corrected, because a stale comment asserting a parameter exists is how this survived.

## The fix

Rewired both against the real API, following the shape `common/enrichment/service.py:347` already uses: `call_fn` receives the resolved provider and calls `complete_text` (`LlmFn`'s contract is raw text, which `parse_distill_response` parses downstream).

Provider comes from `ENTITY_EXTRACTION_PROVIDER`. Forge has no provider setting of its own, and `call_with_fallback` already resolves per-tenant model overrides against `enrichment_model`, so borrowing the enrichment selector keeps both halves of one write pipeline on one provider. A dedicated `FORGE_PROVIDER` would be new config surface and belongs in its own change, not a repair.

**The two differ deliberately on `fake_fn`**, which `call_with_fallback` always has:

- **CLI → returns the placeholder.** That is its documented promise.
- **Cron → raises.** The tick *persists* what comes back, as skill candidates that promote into `staged`. `deliberate_fake_provider` (`retry.py:380`) documents exactly this line for persisting callers: a configured `fake` provider is an operator asking for a stub; a real provider whose attempts all failed is an outage; *an outage is not a request for made-up output*. Neither may write placeholder skills, so both raise and only the message differs — the two want different operator responses. Same posture as `contradiction_detector`, which abstains rather than guess a verdict.

## Why nothing caught any of it

`scripts/` is covered by neither `ruff check` nor mypy in `ci.yml`. **The cron is worse, because it looks covered:** `core_api.services.forge.*` is deliberately opted *into* the type gate (`ignore_errors = false`), and `cd core-api && mypy src/` still reports `Success: no issues found in 199 source files`.

`common/` sits outside core-api's `mypy_path`, so every `common.*` symbol resolves to `Any` — neither the missing name nor the wrong call signature is reportable. Put the repo root on the path and that same unmodified file reports all three:

```
cron_handler.py:204: error: Module "common.llm" has no attribute "LLMRequest"  [attr-defined]
cron_handler.py:218: error: Unexpected keyword argument "expecting" for "call_with_fallback"  [call-arg]
cron_handler.py:218: error: Need type annotation for "response"  [var-annotated]
```

caura#948 gated `common/` itself. **This is the same gap one layer out — the services still cannot see what they import** — and closing it is its own piece of work, not something to bolt onto a repair. Sizing, measured: OSS root `scripts/` has 45 errors in 8 files, root `tests/` has 487 in 100 files.

## Verification

- **Three regression tests, each confirmed to fail without this commit** (reverted the two source files, re-ran, watched them fail with the production `RuntimeError` verbatim, restored): the CLI's wiring must not return the stub when `common.llm` imports; the cron's wiring must resolve rather than raise; the cron must refuse to emit a stub. The existing tests here exercised `_fake_llm_fn` directly and *deliberately skipped the branch selection* — the exact line the bug lived on.
- Behavioural, end-to-end with `ENTITY_EXTRACTION_PROVIDER=fake`: the CLI's real wiring routes through `call_with_fallback` → `fake_fn` and yields JSON `parse_distill_response` accepts with every required key; the cron raises instead. `_fake_llm_fn` stays awaitable and slug-distinct, so its existing tests are untouched.
- `tests/test_forge_distill.py` → **79 passed**; forge/cron/skill/distill/promote across `tests/` → **420 passed**.
- All four gated mypy runs clean (199 / 83 / 5 / 11). `ruff check` + `ruff format --check core-api/src/` clean, run separately. `common/` gate from #948 still clean.
- No new lint anywhere: pre-existing findings in `scripts/forge_dry_run.py` (EXE001, G201, and a RUF100 `# noqa` that suppresses nothing) and in `tests/test_forge_distill.py` are left alone — unrelated to this repair, and neither path is lint-gated. Counts verified identical to `origin/main`.
- `core-api/uv.lock` untouched. Rebased onto `main` at `3cc6ea8d`; none of the four intervening commits touch these files.

## One judgement call worth flagging

I did **not** narrow the `except ImportError` further. With only real names imported it now fires solely for a genuinely absent package, which is what it was written for. Guarding against a *future* rename would need a runtime `getattr` dance that duplicates what the type checker should do — the durable fix is making `common/` visible to the service mypy runs, above.

### DIFF
--- a/core-api/src/core_api/services/forge/cron_handler.py
+++ b/core-api/src/core_api/services/forge/cron_handler.py
@@ -31,6 +31,7 @@ attribute failures to the right tenant.
 from __future__ import annotations
 
 import logging
+import os
 from datetime import UTC, datetime, timedelta
 from typing import Any
 
@@ -200,11 +201,17 @@ async def _wire_llm_fn() -> LlmFn:
     # Lazy import — ``common.llm`` pulls in vertex/openai SDKs that
     # we don't want loading at core-api startup just for the cron
     # adapter wiring.
+    #
+    # Every name here is one ``common.llm`` actually exports. It used to ask for
+    # ``LLMRequest``, which has never existed in that module, so the import below
+    # always raised and the ``except`` clause below always ran — meaning this
+    # function raised RuntimeError on every tick and the distill cron could not
+    # run at all. Worse, the message blamed a missing provider chain, so the
+    # symptom read as a deployment problem rather than as a wrong import.
     try:
-        from common.llm import (  # type: ignore[import-not-found]
-            LLMRequest,
-            call_with_fallback,
-        )
+        from common.llm import LLMProvider, call_with_fallback
+        from common.llm.retry import deliberate_fake_provider
+        from common.provider_names import ProviderName
     except ImportError as exc:
         raise RuntimeError(
             "forge_cron: common.llm not importable — the production "
@@ -213,10 +220,50 @@ async def _wire_llm_fn() -> LlmFn:
             "dry-run); see scripts/forge_dry_run.py."
         ) from exc
 
+    # Same selector the enrichment write path uses. Forge has no provider
+    # setting of its own, and ``call_with_fallback`` resolves per-tenant model
+    # overrides against ``enrichment_model`` by default, so borrowing the
+    # enrichment provider keeps the two halves of one write pipeline on the same
+    # provider. A dedicated ``FORGE_PROVIDER`` would be a new config surface and
+    # belongs in its own change, not in a repair.
+    provider_name = os.environ.get("ENTITY_EXTRACTION_PROVIDER", ProviderName.OPENAI)
+
+    def _refuse_fake() -> str:
+        """``call_with_fallback``'s last resort, which this caller must not take.
+
+        This tick PERSISTS what the LLM returns, as skill candidates that
+        promote into ``staged``. ``deliberate_fake_provider`` documents the line
+        that matters for persisting callers: a configured ``fake`` provider is an
+        operator asking for a stub, while a real provider whose every attempt
+        failed is an outage — and an outage is not a request for made-up output.
+        Neither case may write placeholder skills here, so both raise; the
+        distinction is kept in the message because the two want different
+        operator responses. Same posture as ``contradiction_detector``, which
+        abstains rather than guessing a verdict.
+        """
+        if deliberate_fake_provider(provider_name):
+            raise RuntimeError(
+                "forge_cron: ENTITY_EXTRACTION_PROVIDER is 'fake'. The fake-LLM "
+                "fallback is intentionally CLI-only (see scripts/forge_dry_run.py) "
+                "— the cron will not mint placeholder skill candidates."
+            )
+        raise RuntimeError(
+            "forge_cron: every configured LLM provider failed for "
+            f"'{provider_name}'. Refusing to persist placeholder skill candidates."
+        )
+
     async def _llm_fn(prompt: str) -> str:
-        request = LLMRequest(messages=[{"role": "user", "content": prompt}])
-        response = await call_with_fallback(request, expecting="json")
-        return response.text
+        async def _call(llm: LLMProvider) -> str:
+            # ``complete_text``, not ``complete_json``: ``LlmFn``'s contract is
+            # raw text, which ``parse_distill_response`` parses downstream.
+            return await llm.complete_text(prompt)
+
+        return await call_with_fallback(
+            primary_provider_name=provider_name,
+            call_fn=_call,
+            fake_fn=_refuse_fake,
+            service_label="forge_cron",
+        )
 
     return _llm_fn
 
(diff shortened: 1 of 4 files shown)


Eldad-Caura:
@claude please re-review the amended commit (`60f75634`) — the CI failure is fixed, and the cause is worth recording.

**It was our change, but the bug wasn't ours.** CI failed on an unrelated test, `test_the_retry_budget_is_a_sum_not_a_product`, with `AttributeError: 'module' object at common.llm.retry has no attribute 'retry'`.

Root cause: `tests/test_forge_cron_tick.py` did `sys.modules.pop("common.llm", None)` **outside** its `patch.dict`. Popping before `patch.dict` takes its snapshot means the snapshot records "absent", so the real module never came back and every later test in the session had to re-import it — and that re-import rebinds the package without rebinding its `retry` submodule attribute, which is exactly what the unrelated `monkeypatch.setattr("common.llm.retry.LLM_RETRY_DELAY_S", …)` could no longer resolve.

The leak predates this PR. It was harmless only because `_wire_llm_fn`'s import **always raised**, so nothing downstream ever completed a re-import of `common.llm`. Fixing the import is what turned a latent leak load-bearing — which is itself a small illustration of the PR's theme.

Isolated it rather than guessing: the failure reproduces with `test_forge_cron_tick.py` + **either** of my two new tests alone, and does not reproduce with my new tests deselected.

The `pop` was also unnecessary — a `sys.modules` entry of `None` is itself what makes `import` raise, and `patch.dict` restores it. Removed the line, and verified the branch is still exercised for the right reason (`ImportError: import of common.llm halted; None in sys.modules`) so the test can't pass vacuously, and that `getattr(common.llm, 'retry')` works again afterwards.

Full `tests/` suite locally: **5182 passed**, and the `common.llm.retry` failure is gone. One unrelated timing test (`test_sweep_recovers_stale_processing_but_not_fresh`, a ~12s sweep test) failed on that run and passes 3/3 in isolation; it touches nothing this PR changes and was not in CI's failure list. Re-running the full suite to confirm it's a flake — will report either way rather than leave it unexplained.

Also rebased onto `main` at `d9bf8250`.

Eldad-Caura:
@erni-a ready for your approval — **CI is green** on the current head `60f75634` (CI, CodeQL, dco, ratchet, sentinel, security scan, image-version), and the Claude review came back clean on the amended commit.

Worth reading before you merge, because the headline is a production outage rather than the dev-script repair this started as:

**The forge distill cron has never been able to run a tick.** `cron_handler._wire_llm_fn` imported `LLMRequest` from `common.llm` — a name that has never existed there — so the import always raised, the `except ImportError` always fired, and every tick died with `RuntimeError: forge_cron: common.llm not importable`. The message blamed a missing deployment dependency, so the symptom pointed away from the cause. The dry-run CLI had the identical import; its `except` returns the stub, so that CLI had never once used a real LLM.

**It was invisible to the type gate, and that is the part worth acting on.** `core_api.services.forge.*` is deliberately opted *into* mypy (`ignore_errors = false`), and `cd core-api && mypy src/` still reports "no issues found in 199 source files" — because `common/` sits outside core-api's `mypy_path`, so every `common.*` symbol resolves to `Any`. Put the repo root on the path and that same unmodified file reports all three errors. caura#948 gated `common/` itself; this is the same gap one layer out and it is not fixed by it. Sized for whoever picks it up: OSS root `scripts/` has 45 mypy errors in 8 files, root `tests/` has 487 in 100 files — a programme, not a PR.

**On the CI failure you spotted:** it was our change, but not our bug. `tests/test_forge_cron_tick.py` leaked `sys.modules` (a `pop` outside its `patch.dict`); the leak was harmless only while `_wire_llm_fn`'s import always raised. Fixing the import made a latent leak load-bearing. Diagnosed by isolation, not guesswork, and fixed at the source. Full `tests/` suite run twice locally: **5182 then 5197 passed**, no failures on the second. One `~12s` timing test failed on the first run only, passes 3/3 in isolation, touches nothing here — a flake, confirmed by re-run rather than assumed.

Three regression tests included, each confirmed to fail without the fix.

erni-a (approved):

PR #958, opened 2026-08-25
test(plugin): make the signature-tamper test deterministic

arkash20:
Fixes the 1-in-16 flake in `verifyCommandSignature > tampered signature fails closed regardless of mode`: the tamper replaced the last hex char with a literal `"0"`, so whenever the genuine (timestamp-dependent) signature already ended in `0`, the "tampered" copy was byte-identical to the original — verification correctly returned true and the test failed. Verified empirically (7/200 sampled timestamps produce sig-ends-in-0) and first observed on #954's rebase run — the suite only started running in CI with #950.

Fix: flip the last char to a guaranteed-different value + assert the tampered signature differs from the original. Validation file passes 8/8; fixed tamper produced 0 identical copies across 3200 sampled timestamps.

### DIFF
--- a/plugin/src/validation.test.ts
+++ b/plugin/src/validation.test.ts
@@ -74,7 +74,15 @@ describe("verifyCommandSignature", () => {
 
   test("tampered signature fails closed regardless of mode", () => {
     const cmd = signedCommand();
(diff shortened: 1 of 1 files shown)


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #959, opened 2026-08-25
fix(mcp): stop _with_latency corrupting non-dict JSON payloads

arkash20:
## What

Closes canonical **C24** (register API-01/N4, from the AX-audit verification pass). `_with_latency` stamped dict payloads inline but appended a plain-text `_latency_ms` line after any other result — for `_serialize`'s list branch that produced `[...]` + trailing junk, which strict JSON parsers reject. This is the likely real source of the field-reported "invalid JSON from /search" (REST serialization is stock FastAPI end-to-end and cannot emit unescaped output; the sighting matches the MCP surface).

Fix: valid-but-non-dict JSON (arrays, bare scalars) returns **unchanged** — an array can't carry the stamp without breaking its shape, and no client ever successfully parsed the stamped form. Dict payloads keep the inline stamp, error envelopes keep the `isError` promotion (regression-guarded), non-JSON prose keeps the suffix.

## Testing

- 6 new unit tests (`tests/test_c24_latency_json_integrity.py`): strict-parse guarantees per payload class, byte-identical array pass-through, isError promotion intact. Full root suite **5186 passed**; ruff/mypy clean.
- Wet: verified against the **shipped container image** (docker-exec into the rebuilt core-api): array untouched + strictly parseable, dict stamped inline, prose suffixed.

### DIFF
--- a/core-api/src/core_api/mcp_server.py
+++ b/core-api/src/core_api/mcp_server.py
@@ -666,6 +666,16 @@ def _with_latency(result: str, t0: float) -> str | CallToolResult:
             if isinstance(data.get("error"), dict):
                 return _as_error_result(payload)
             return payload
+        # C24 — valid JSON that is NOT a dict (``_serialize``'s list branch,
+        # or a bare scalar) must go back UNCHANGED. The old fall-through
+        # appended the plain-text ``_latency_ms`` line after the JSON,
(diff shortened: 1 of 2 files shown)


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #961, opened 2026-08-25
chore(sentinel): retire the observability logger-name pin

Eldad-Caura:
Removes one entry from `scripts/do_not_touch_sentinel.py`. **No rename in this PR** — see
*What this PR is not* below.

## The entry's justification is false

```
Sentinel(
    path="core-storage-api/src/core_storage_api/observability.py",
    text="memclaw.observability",
    breaks="the logger name changes and every log-based filter keyed to it goes quiet",
)
```

Re-measured **2026-08-25**. Every place a "log-based filter" could live is empty:

| surface | state | how it was measured |
| --- | --- | --- |
| Datadog log-type monitors | **0 of 18** — all `query` / `trace-analytics` / `synthetics` / `event-v2` | full monitor listing; offset 18 returns empty, so the list is complete |
| GCP log-based metrics | **0** | `gcloud logging metrics list` → empty |
| GCP log sinks | **2**, both built-in (`_Required`, `_Default`), filtering on `LOG_ID()` only | `gcloud logging sinks list` |
| GCP log-match alert policies | **2 of 6 exist** — but neither is keyed to a logger name | Monitoring API `alertPolicies`, cross-checked against `caura-enterprise/terraform/datadog-gcp/gcp_alerts.tf` |
| Datadog log pipelines · index exclusion filters · log-based custom metrics | `0 active, 0 disabled` · 1 index, `*`, zero exclusions · none | **hand-checked in the Datadog UI on 25 Aug**; see *limits* below |

### The one surface that is not empty, and why it does not save the entry

Two `conditionMatchedLog` alert policies do exist:

- `Prod embedding degraded / gate timeout` — `gcp_alerts.tf:40`
- `caura-ops memory backend unreachable` — `gcp_alerts.tf:169`

Both match on **`jsonPayload.message` text**, neither references a logger field, and
neither targets `core-storage-api` at all — they watch `prod-*-core-api` / `-core-worker`
and `caura-ops-api` / `-worker`. The strings they match on are separately pinned by the
sentinel entries directly above this one, which stay.

The logger name **is** live and queryable — `@logger:memclaw.observability` returns ~50k
records over two days across `core-storage-reader` / `-writer`. The attribute exists.
What does not exist is anything keyed to it.

## Why removing a pin is the safe direction here

The sentinel's whole power is that people believe the list. An entry asserting a
dependency nobody can locate teaches readers that the list contains unchecked claims —
and the cost of that lands on the entries that really are load-bearing. A list that is
shorter and entirely true is a stronger gate than one that is longer and partly
aspirational.

## Limits of this audit — stated deliberately

- **Facets and saved views were not checked.** They break a human's filtering workflow,
  not an alert. That is why this is still the right call, and why whoever does the rename
  should mention them in the release note.
- The three Datadog surfaces in the last table row could not be re-verified from this
  environment (no MCP tool exposes pipelines, exclusion filters or log-based metrics, and
  no API key is available here). They rest on the 25 Aug hand-check. The four rows above
  them were re-measured independently for this PR.

## What this PR is not

**The logger is not renamed.** `observability.py:26` still reads
`logging.getLogger("memclaw.observability")`, and `tests/test_observability.py` still
pins it in five places. That is ordinary debt for a later sweep; bundling it here would
make neither change reviewable.

## Gates

- `do_not_touch_sentinel.py` → `All 34 protected strings survive.` (was 35)
- `legacy_name_ratchet.py --base origin/main` → `No new lines.`, with the expected
  *"1 exempt line(s) removed by this change"* report. The alias it flags is the pin's own
  copy of the string; the original at `observability.py:26` is untouched.
- `pytest tests/test_do_not_touch_sentinel.py tests/test_sentinel_scanner.py` → 153 passed.
  No test hardcodes the entry count.

### DIFF
--- a/scripts/do_not_touch_sentinel.py
+++ b/scripts/do_not_touch_sentinel.py
@@ -188,12 +188,6 @@ SENTINELS: tuple[Sentinel, ...] = (
         breaks="the generated installer writes to a path no installed plugin reads",
     ),
     Sentinel(
-        path="core-storage-api/src/core_storage_api/observability.py",
-        text="memclaw.observability",  # legacy-name-ok: pinned floor string
-        kind=LITERAL,
-        breaks="the logger name changes and every log-based filter keyed to it goes quiet",
-    ),
-    Sentinel(
         path="core-api/src/core_api/services/organization_settings.py",
         text="memclaw.auto_upgrade_enabled",  # legacy-name-ok: pinned floor string
         kind=LITERAL,


Eldad-Caura:
Declining this one, with evidence. Noting first that the review ran on `53ebc278` and the
head is now `e10696f7` (rebased onto `main` after #960 merged) — the PR diff is byte-identical
across the two, so the finding still applies to current content and is answered on its merits.

**The finding's only evidence for the dependency is the entry's own `breaks=` string** —
which is the exact claim this PR audits. Quoting a pin's justification back as proof that
the pin is needed is circular; if that were sufficient, no sentinel entry could ever be
retired, and the list could only grow.

### The two remedies offered, against what this PR actually does

> If it still contains that literal unchanged, restore the removed Sentinel entry

`observability.py:26` **does** still contain it, unchanged, deliberately, and the PR body
says so. **This PR removes the pin, not the string.** Restoring the entry would re-add an
assertion whose named dependant does not exist.

> If the logger name was intentionally renamed/removed as part of this PR, add a new
> Sentinel entry pinning the new logger name

No rename happened here, deliberately — bundling one would make neither change reviewable.
So neither branch of the suggestion applies.

### "removing drift protection … log-based alerting/filtering depends on"

That is the part that was measured, on 2026-08-25, and it is not true:

| surface | state |
| --- | --- |
| Datadog log-type monitors | **0 of 18** — every one is `query` / `trace-analytics` / `synthetics` / `event-v2`; offset 18 returns empty, so the list is complete |
| GCP log-based metrics | **0** |
| GCP log sinks | **2**, both built-in (`_Required`, `_Default`), filtering on `LOG_ID()` only |
| GCP log-match alert policies | **2 exist** — `gcp_alerts.tf:40` and `:169`. Both match on `jsonPayload.message` **text**; neither references a logger field, and neither targets `core-storage-api` at all |
| Datadog pipelines · index exclusions · log-based custom metrics | `0 active, 0 disabled` · 1 index, `*`, zero exclusions · none |

`@logger:memclaw.observability` is a live, queryable attribute (~50k records over two days).
The attribute exists. Nothing is keyed to it. Those are different facts, and the entry
asserts the second one.

### Why removal is the safe direction

`do_not_touch_sentinel.py` is not drift detection over a verified consumer graph — it
asserts that specific strings still **exist**, and its entire power is that people believe
the list. An entry naming a dependant nobody can locate teaches readers that the list
carries unchecked claims, and that cost lands on the entries that really are load-bearing.
A shorter list that is wholly true is a stronger gate than a longer one that is partly
aspirational.

Two limits are stated in the PR body rather than glossed: **facets and saved views were not
audited** (they break a human's filtering workflow, not an alert — which is why this is
still the right call), and the three Datadog rows in the last table row rest on a UI
hand-check rather than an API re-measurement.

Gates on this head: sentinel `All 34 protected strings survive.`, ratchet `No new lines.`,
`dco`, CodeQL, Security Scan, and the ratchet/sentinel required checks all green.

No change applied.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

Eldad-Caura:
**Follow-up: the two limits this PR stated have since been closed, and one supporting claim is a query artifact.**

The PR body flagged facets and saved views as unchecked, and the three Datadog surfaces as
resting on a UI hand-check. Both gaps have now been closed in the Datadog UI (25 Aug). Recording
the completed table here, since the merged body still carries the older caveat:

| surface | result | source |
| --- | --- | --- |
| log-based custom metrics | none exist | UI, 25 Aug |
| custom log pipelines | `0 active, 0 disabled` | UI, 25 Aug |
| index exclusion filters | 1 index, `*` filter, zero exclusions | UI, 25 Aug |
| log-type Datadog monitors | **0 of 18** | re-measured for this PR |
| GCP log-based metrics | **0** | re-measured for this PR |
| GCP log-matched alert policies | 2 exist estate-wide, neither names it | re-measured for this PR |
| **Datadog facets** | **`Showing 0 of 8`** searching `logger` | UI, 25 Aug |
| **Datadog saved views** | **`No matching views found`** | UI, 25 Aug |

Two details make those last two nulls worth more than their face value: the saved-views search was
for the **bare old-brand string** with *"show only my saved views"* **off**, so it covered every
saved view in the org — which clears that surface for **every name in the register**, not just this
one. And eight facets is Datadog's near-default set, so this org has done essentially no log-facet
configuration at all.

### One correction, because it will otherwise be reused

A flat-zero histogram on `@logger.name:"memclaw.observability"` was offered as corroboration. **It
is a query artifact, not a null.** The attribute this service emits is `logger`, which Datadog
exposes as `custom.logger` and queries as `@logger` — there is no `@logger.name`. Same 21-day
window, same org:

```
@logger.name:"memclaw.observability"  →       0
@logger:"memclaw.observability"       → 194,874
```

This does not disturb the conclusion — 194k records prove the name is *live*, not that anything is
*keyed to* it, and "keyed to" is what the retired entry claimed. But if `@logger.name` is used to
clear other names in the register it will return zero for every one of them regardless of truth.
Worth fixing in the query template before the next sweep.

Nothing here changes the merged diff.

PR #963, opened 2026-08-25
feat(api): envelope convergence — dual-emit items + keystones envelope opt-in (C30)

arkash20:
## What

Closes canonical **C30** (register API-02) — second task off the ratified wire contract (**D1**: `items` is canonical).

- **POST /documents/search** dual-emits `results` + `items` (same array) with `count` — the items/results sibling-endpoint split was SupportHive's #1 time-cost bug. `results` is permanent until a separate announced deprecation wave.
- **GET /keystones** gains `envelope=1` opt-in → `{count, items}`; the bare array stays the byte-identical default (plugin session-start fetch unaffected).

Zero behavior change for any existing caller; both changes are additive/opt-in per the contract's rule of the page.

## Testing

- Full suite **5197 passed** (3 new source-guard tests); ruff/mypy/format clean; broker baseline unchanged.
- Wet (fresh image, live stack): documents/search returns `results == items` with coherent `count` on real hits; keystones default confirmed bare-array; `envelope=1` returns `{count, items}`. (Keystone row-seeding via a user token hits the pre-existing authoring trust gate — outside this diff, which only shapes the response after rows are fetched.)

### DIFF
--- a/core-api/src/core_api/routes/documents.py
+++ b/core-api/src/core_api/routes/documents.py
@@ -657,11 +657,18 @@ async def search_documents(
             result_count_by_tenant=counts,
             query_summary=(body.query or "")[:200],
         )
+    # C30 / wire-contract D1 (ratified 2026-08-25): ``items`` is the canonical
+    # list key everywhere — /search always used it, this route used
+    # ``results``, and the mismatch cost the SupportHive builder a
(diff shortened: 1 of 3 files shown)


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #978, opened 2026-08-25
fix(worker,storage): enrichment completion observable in the C25 view (B7)

arkash20:
## What

Closes canonical **B7** (register PERF-02): fast mode's `enrichment_pending` is the pollable completion signal — but the async completion paths cleared only the legacy top-level key, and the C25 read view gives `metadata._system` precedence, so `system_metadata.enrichment_pending` stayed **True forever** on fast-mode rows. Polling could never observe completion; same for `embedding_pending`.

- **Worker enrichment PATCH**: dual-home flag clear + mirrors enrichment fields into `_system` (async-enriched rows now match strong-mode rows' namespaced view). Heuristic-fallback redeliveries keep the both-homes clear.
- **Worker embed PATCH**: dual-home `embedding_pending` clear.
- **core-api async-apply path** (wholesale metadata replace): pops from `_system` too.
- **Storage `metadata_patch` merge**: deep-merges `_system` one level (`jsonb_set` old||new) — the shallow `||` would let any flag-clear REPLACE the stored `_system`, clobbering siblings (`write_latency_ms`, `write_mode`, …). Patches without `_system` keep the exact prior statement.

With this, the fast-mode contract is complete: write returns `system_metadata.enrichment_pending: true` → poll GET/search until it flips false with `summary`/`tags` populated. (The push-event companion is API-15, product-scoped.)

## Testing

- Worker **75 passed** (3 new: dual-home clear, `_system` mirroring, heuristic-fallback both-homes; 3 existing exact-shape assertions updated to the intended new patch shape). Storage **156 passed** incl. 2 new deep-merge tests **against real Postgres** — sibling `_system` keys survive the clear; no phantom namespace on plain patches. Root suite **5217 passed**; ruff/mypy clean across all three services.
- Local stack runs no core-worker, so the full async round-trip can't wet-test locally; the storage deep-merge (the risky part) is verified against the real compose Postgres, and the write-side flag was wet-verified in C25's run.

### DIFF
--- a/core-api/src/core_api/services/memory_service.py
+++ b/core-api/src/core_api/services/memory_service.py
@@ -3069,7 +3069,7 @@ async def _enrich_memory_background(
             patch["status"] = enrichment.status
 
         meta.pop("enrichment_pending", None)
-        # B7×C25 — this path REPLACES metadata wholesale, so clear the
+        # B7 x C25 — this path REPLACES metadata wholesale, so clear the
         # namespaced copy too or the C25 read view stays pending forever.
         if isinstance(meta.get("_system"), dict):
             meta["_system"].pop("enrichment_pending", None)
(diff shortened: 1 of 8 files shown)


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #979, opened 2026-08-25
fix(logging): route warnings through logging so they stop counting as errors

Eldad-Caura:
## Problem

`warnings.warn()` writes straight to stderr, bypassing structlog entirely. Cloud Run derives a log's status from the stream it arrived on, so every Python warning reaches Datadog as **`status:error`**.

**Measured: 1,543 of 1,550 error-status logs from staging core-api over 6h — 99.5% — were a single warning**, at ~257/hour. That service's error rate, Error Tracking, and anything keyed on either were reporting essentially one warning.

This is the same false-error mechanism `common/structlog_config.py`'s own header already documents for uvicorn's supervisor lines, arriving by a different route. That one was fixed by keeping `uvicorn.*` propagating to the structlog root handler — warnings never went through `logging` at all, so they were untouched by it.

## Fix

`logging.captureWarnings(True)` in `configure_logging()`. Warnings go to the `py.warnings` logger, which has no handler of its own and propagates to the root handler configured there — so they emit as JSON at WARNING. **They stay visible; they stop counting as errors.**

Verified by capture rather than by reasoning:

| | stderr | stdout |
|---|---|---|
| without the call | 54 bytes (→ `status:error`) | 0 bytes |
| with it | 0 bytes | JSON, `level: warning`, `logger: py.warnings` |

## The warning itself is not ours to fix

Worth stating plainly, since it's the whole volume: it comes from `mcp.server.fastmcp.server.Settings.lifespan` (mcp 1.28.1), annotated `Callable[[FastMCP[LifespanResultT]], ...]` — a generic forward reference pydantic-settings 2.15.0 can't resolve. We can't fix the annotation from here, so reclassifying it is the fix available to us. It also stays queryable as `status:warning` rather than disappearing.

## An ordering precondition worth knowing about

Only warnings raised **after** `configure_logging()` are captured, which makes each service's `app.py` import order load-bearing. core-api calls `configure_logging()` at `app.py:22` and the mcp import lands at `:38`, so warnings raised while those modules are defined are caught. Moving an import above that call would silently send its warnings back to stderr. Called out in the comment at the call site.

## Tests

Two, and the split is deliberate:

- **`test_configure_logging_captures_warnings`** pins that the call happens. It asserts on `logging._warnings_showwarning` rather than raising a warning, because pytest's warnings plugin wraps each test in a recorder that replaces `warnings.showwarning` — so a `warnings.warn()` in a test is swallowed by the plugin, and `showwarning` reads as the recorder's rather than logging's. Re-enabling capture inside the test to dodge that would only assert that the test called `captureWarnings` — vacuous. **Confirmed to fail without the fix.**
- **`test_py_warnings_logger_emits_warning_status`** covers the other half — that a `py.warnings` record renders as `status:warning` — driven through the logger directly, the same approach as the neighbouring `uvicorn.error` test. This one passes either way by design; it pins the destination, not the routing.

The private-attribute reliance is intentional and documented: it's the flag `captureWarnings` actually toggles, the recorder doesn't save or restore it, and a CPython rename would fail loudly with `AttributeError` rather than silently passing.

## ⚠️ Merge order — this must land BEFORE the enterprise PR

`common/structlog_config.py` is an `identical`-policy vendored file in caura-memclaw-enterprise, whose CI runs `check_vendored_drift.py` against OSS **`main`** demanding byte equality.

- The matching enterprise PR's CI **cannot go green** until this is on OSS `main`.
- In the window between the two merges, **every open enterprise PR fails the drift gate**.

So: merge this, then the enterprise PR immediately after. I've verified the enterprise copy is byte-identical to this branch's and that `check_vendored_drift.py` reports `OK: 12 vendored files checked` against it.

## Verification

- `39 passed` in `tests/test_logging.py`; **`5204 passed`, 4 skipped, 1 xfailed** on the full repo-root suite
- `ruff check common/` clean; `ruff format --check --isolated common/__init__.py common/structlog_config.py common/serve.py` clean (the 88-col vendoring gate)
- `mypy --config-file common/mypy.ini common/` — only two pre-existing `dateutil` stub errors in `common/enrichment/service.py`, none in changed code

### DIFF
--- a/common/structlog_config.py
+++ b/common/structlog_config.py
@@ -34,6 +34,7 @@ from __future__ import annotations
 import logging
 import sys
 import threading
+import warnings
 from typing import Any
 
 import structlog
@@ -96,6 +97,9 @@ _configured_log_file: str | None = None
 # or a library that ran before us.
 _installed_handler: logging.Handler | None = None
 _installed_file_handler: logging.Handler | None = None
+# True only when this module changed warning capture from disabled to enabled.
+# A caller that enabled capture before us still owns that process-global state.
+_installed_warning_capture = False
 # RLock so a custom `warnings.showwarning` hook that re-enters
 # `configure_logging` on the same thread won't deadlock. The mismatch
 # warning below is the only user-visible call site that could plausibly
@@ -441,8 +445,6 @@ def configure_logging(
                 or json_logs != _configured_json_logs
                 or log_file != _configured_log_file
             ):
-                import warnings
-
                 warnings.warn(
                     f"configure_logging re-called with different arguments: "
                     f"({environment!r}, {log_level!r}, json_logs={json_logs!r}, "
@@ -608,7 +610,7 @@ def _reset_for_testing() -> None:
     """
     global _configured, _configured_environment, _configured_log_level
     global _configured_json_logs, _configured_log_file
-    global _installed_handler, _installed_file_handler
+    global _installed_handler, _installed_file_handler, _installed_warning_capture
     with _configure_lock:
         structlog.reset_defaults()
         # Drop any bound contextvars so test fixtures don't leak request-scoped
@@ -630,6 +632,10 @@ def _reset_for_testing() -> None:
         _installed_handler = None
         _installed_file_handler = None
         root.setLevel(logging.WARNING)
+        # Undo only the process-global warning capture installed by this module.
+        if _installed_warning_capture:
+            logging.captureWarnings(False)
+        _installed_warning_capture = False
         # Clear the per-logger WARNING overrides that _configure_logging_impl
         # installs at INFO level. Without this, a test that reconfigures at
         # DEBUG would see root at DEBUG but these four stuck at WARNING —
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
Addressed the review finding: `_reset_for_testing()` now undoes warning capture only when this module changed it from disabled to enabled, so test isolation is restored without clobbering capture owned by an earlier caller. Added order-independent regression tests for both the undo path and the pre-enabled ownership case, with `finally` cleanup for the process-global state.

Eldad-Caura:
@claude

Eldad-Caura:
CI follow-up: the first amended head failed only because mypy does not expose the private `logging._warnings_showwarning` attribute in typeshed. Replaced the direct access in production code with an equivalent `getattr(..., None)` check. Local verification is green: mypy (202 source files), common ruff/format, 41 focused logging tests, and 5,466 full-suite tests.

Eldad-Caura:
@claude

Eldad-Caura:
Addressed the private-stdlib review finding on the new head: ownership is now detected by comparing the public `warnings.showwarning` hook before and after `logging.captureWarnings(True)`, so production startup no longer reads `logging._warnings_showwarning`. This still distinguishes capture installed by this module from capture already owned by another caller. The two reset tests now assert the same behavior through the public hook. Verification: Ruff lint/format, mypy (202 files), 41 focused logging tests, mutation check, and the full suite (5,466 passed, 4 skipped, 1 xfailed).

Eldad-Caura:
@erni-a pelase approve

erni-a (approved):

PR #981, opened 2026-08-25
fix(search): retrieval contract for genuine contradictions — newer value wins (A34)

arkash20:
## Summary

Implements the A34 retrieval contract (first slice of MEM-04): **whenever a result set contains both a superseded row and its successor, the successor ranks immediately above its stale predecessor** — never below it, never dangling unscored at the tail.

This closes the live Hermes/STALE-T2 trap: a query exact-matching the OLD wording ranks the stale `conflicted` row #1 (A31's exact-match exemption deliberately keeps it surfaced/un-penalized) with the correction underneath — and the answer LLM picks the stale value.

## What changed

One reorder pass in `LoadAndSerialize` (`load_and_serialize.py`), covering **both** successor arrival paths:

- **Injected** successors (pulled in by `find_successors`, unscored) — previously appended at the tail; now placed directly above their stale predecessor.
- **Organically recalled** successors that ranked *below* their predecessor — promoted to directly above it. Wet-testing showed this path alone reproduces the bug.

Contract details:
- A successor that already out-ranks its predecessor keeps its earned position.
- Chains (C supersedes B supersedes A) resolve newest-first via recursion; a corrupt supersession *cycle* terminates (placed-before-recurse guard).
- **No new wire fields**: the successor's `supersedes_id` names the loser; the stale row's `status` says why it lost. Injected rows stay unscored (`score`/`score_parts` null — D12 contract preserved).

## Testing

- 7 new regression tests in `tests/test_a34_successor_contract.py` (injected above / mid-list / organic-above keeps rank / organic-below promoted / chain newest-first / multiple stale rows / no-op without successors). Full suite: **5225 passed**.
- **Live wet test** on the local enterprise compose stack (fresh core-api image): seeded the exact evolving-profile pair that failed pre-fix — post-fix the correction ranks #1 directly above the stale exact-match (raw score 0.71 vs 0.80).
- **67-question regression bench** (strong mode, d=0, golden-baseline compare): single-session-user, temporal-reasoning, knowledge-update exactly at baseline; multi-session within thresholds; recall unchanged everywhere. One initial single-session-preference accuracy dip was diagnosed as answer-LLM flake — all 3 failing questions had perfect recall and **zero supersession pairs in their K=20 sets** (the reorder provably never fired), and an immediate same-build re-run scored **12/12 (100%)** at baseline recall (0.9722).

Refs: canonical A34 / MEM-04; ratified wire-contract page (no new fields needed).

### DIFF
--- a/core-api/src/core_api/pipeline/steps/search/load_and_serialize.py
+++ b/core-api/src/core_api/pipeline/steps/search/load_and_serialize.py
@@ -26,6 +26,15 @@ _SCORE_FACTORS = (
 )
 
 
+def _mem_field(memory, key: str):
+    """Field access tolerating both SimpleNamespace and dict Memory shapes."""
+    if hasattr(memory, key):
+        return getattr(memory, key)
+    if isinstance(memory, dict):
+        return memory.get(key)
+    return None
+
+
 def _score_parts(row) -> ScoreParts | None:
     """Build the D12 factor breakdown from a scored row; None when unscored.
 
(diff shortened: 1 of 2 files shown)


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #982, opened 2026-08-25
feat(ratchet): a floor mention is not a compat alias — give it its own marker

Eldad-Caura:
`legacy-name-ok` means two different things and the report cannot tell them apart:
a **compat alias**, which is what rule 3's escape hatch is for, and a **floor
mention** — a permanent name appearing in text, where nothing new is declared at all.

## Measured, fleet-wide, 100% coverage

622 markers across the seven repos:

| kind | count | share |
| --- | --- | --- |
| compat alias | 340 | 55% |
| **floor mention** | **274** | **44%** |
| should be reworded, no marker | 8 | 1% |

**44% of every exemption in the programme is a floor mention**, and today the only
thing separating them from aliases is unstructured prose in the reason text.

`caura-daemon#136` is the shape of it in a single PR: eleven exemptions, four of them
aliases. A reviewer sees eleven undifferentiated entries and stops reading.

## What this does NOT claim — please read this bit

An earlier framing of this work (mine, in the brief I was given) said the mixed marker
was producing **false justifications at scale**, citing the `#882` note about being
forced to *"write a reason that was not true, into the one annotation whose whole value
is that its reasons are true."*

**The data does not support that, and this PR should not be approved on it.** I checked
the audit myself rather than taking it secondhand:

- Only **11 of 622 markers (1.8%)** put alias language on a non-alias line.
- **218 of the 274** floor mentions already say "floor", "pinned" or "frozen" in their
  own reason text.
- Most of the 11 are `taught as legacy alias` on prose lines, which is arguably true
  anyway.

Authors have been drawing this distinction correctly all along. The gate has simply had
nowhere to record it, so it could not count on it. **This is a signal and structure
change, not a corruption fix.** The `#882` risk was real and it did not scale.

What *did* scale is 274 floor mentions filed in the same list as the aliases — because
floor mentions accrue from **editing documentation** rather than from renaming anything.
Both mechanisms re-verified against the current gate:

- **Re-wrapping mints.** Re-flowing a comment that mentions the CLI verb fails the gate
  *while the file's count falls* — the text is new even though the tally dropped.
- **Markers are self-perpetuating.** Stripping a marker and leaving the line otherwise
  identical fails: the previously-exempt text becomes non-exempt text the repo never had.

So that half of the pile grows as docs are maintained, not as the brand is swept.

## The sunset plan already asks for this

`docs/plans/rebrand-sunset-plan.md` already says *"a mention gets reworded, a contract
gets the marker … **mark it, and say which**."* Until now "which" had nowhere to live but
the reason prose. This gives that sentence a mechanism.

## The design

1. **Both markers exempt identically.** The pass/fail decision reads only whether
   `_kind()` returned `None`, never which marker it found. A mislabelled marker cannot
   turn a red build green or a green one red.
2. **Every existing `legacy-name-ok` stays valid and untouched.** No migration. The 52
   pre-existing tests pass **unedited** — the test-file diff is purely additive, zero
   deleted lines. That is the evidence, not the regex.
3. **The new marker is a falsifiable claim.** "This names something on the floor and the
   line cannot be correct without the literal" is checkable in one move: reword it, and
   if it is still correct the marker was false. "This is fine" would not have earned a
   marker.
4. **Reason text stays asked-for, not enforced**, exactly as today.

**Precedence, when both claims are true of one line.** 28 lines across the fleet name a
frozen thing and declare a dual-read at once — e.g. an image tag whose repository name is
permanent while its version is read from either spelling. One line takes one marker, so
precedence is `_KINDS` order (alias first) rather than position on the line: the answer
must not depend on which end the author typed first, and the tie breaks away from
under-counting aliases, which is the expensive direction.

One regex serves both markers with the boundaries **outside** the alternation, so the
bounded-token rule is stated once and cannot drift per marker. `_KINDS` is the single
table; a third kind is one entry plus a paragraph in the failure text (which the table's
comment says out loud).

`Scan` is untouched — `exempt` and `exempt_text` are not split, because `exempt` only
*selects* files (the report re-greps and reads kinds first-hand) and `exempt_text` keys on
text that still contains its own marker.

## What the report now prints

```
10 exempt line(s) written by this change in 3 file(s) —
4 compat alias(es), 6 floor mention(s).

4 compat alias(es) — rule 3's escape hatch, and what this report is for.
Check each is a compat alias, a redirect or a pinned wire format,
and not headroom for a new name:
  ...

6 floor mention(s) — a permanent name in text, not a new declaration.
Check the line cannot be correct without the literal: a pasteable
command, an on-disk path, a served mirror path. If rewording it
would leave it correct, the claim is false and it should carry no
marker at all:
  ...
```

The first line still counts every kind together and is byte-identical to today's. The
split line is skipped when only one kind is present, where it would just restate the
section header.

## Verification

- **70 tests pass**; the test-file diff is purely additive and the 52 pre-existing tests
  are byte-identical.
- New tests fail against the pre-change script. The boundary guards would pass vacuously
  without the feature, so each was validated by **mutating the implementation** instead:
  drop the lookahead, drop the lookbehind, break the two-marker tie by position, compare
  the winning marker's casing raw, stop normalising casing, let a floor marker excuse its
  whole file, print the split line for one kind, order sections by size. **All 8 mutations
  are detected.** Two of them caught assertions of mine that were vacuous — one only
  checked an exit code where the real failure was a line silently dropped from the report,
  and one asserted a reason that the report happened to print anyway as part of the line
  text.
- **60 real commits of `origin/main` replayed** through the old and new scripts, each in
  its own worktree against its own first parent: **exit code identical 60/60**, stdout
  byte-identical on 41. Of the 19 that differ, 18 wrote exemptions (the diff is the added
  section header — no content lost, totals unchanged) and 1 is the deliberate wording
  change in the removed-exemption report.
- Gate passes on its own source, do-not-touch sentinel still passes (34/34), `ruff check`
  and `ruff format --check` clean on both files.
- Runtime unchanged: 0.36s → 0.38s over the full tree. Subprocess and regex call counts
  are identical to before (7 and 10 on a control fixture).

- **Reclassifying the 274 existing floor mentions.** Mechanism first; a mechanism change
  plus a 274-line reclassification is unreviewable.
- **A third `deferred` kind.** 7 markers are parked on another lane's clock ("moves at the
  flip", "on its own rename schedule") and are neither rule 3 nor permanent floor. The
  table now makes adding one a small change, but 7/622 does not earn it in the same diff
  that introduces the mechanism.
- **The fence-line block marker.** `caura-daemon/specs/001-memclaw-broker/fleet-overview.md:605,672`
  carry markers that render visibly inside ```` ```bash ```` fences, and a floor-vs-alias
  marker does not help them — the marked line is a pasteable command that cannot leave the
  fence, the marker cannot be removed without tripping the gate, and the literal is the
  binary name so it cannot be reworded. Needs a fence-level mechanism.
- **Splitting the removed-exemption report by kind.** Its population is the other way round
  — seven removals across four hundred commits — so the wall is hypothetical there. Its
  guidance sentence is widened to be true of both kinds (telling a reader to confirm an
  alias survived would be a false instruction for a floor mention), but the list is not
  split. Noted in the code as worth revisiting if the rate changes.

## Porting

The seven copies are **not** byte-identical — four generations, plus one genuine fork.
`caura` is upstream and this lands here first. Plan in the lane report; the headline is
that `#944` (the removed-exemption report, written after a data-loss change passed both
required gates green) is currently in **`caura` only, 1 of 7**, and is worth landing
downstream ahead of this.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -218,6 +218,36 @@ It has to be the whole token — `legacy-name-okay` in a sentence does not exemp
 anything — but the casing is up to you, and the reason is asked for rather than
 enforced.
 
+### Two markers: which claim are you making?
+
+`legacy-name-ok` is for something new that **bears** the old name — the alias,
+the redirect, the pinned wire format above.
+
+Often the line is not declaring anything. It just **names** something the rename
+will never reach, and cannot be correct without the literal: a command a reader
+pastes, a path on disk, a mirror URL. Use `legacy-name-floor` instead:
+
+```markdown
+| macOS | `~/Library/Application Support/...` | <!-- legacy-name-floor: the app dir name -->
+```
+
+Both exempt the line identically — nothing about pass or fail changes, and
+picking the wrong one cannot turn a build red or green. They are counted apart
+so the aliases stay readable: a documentation sweep can easily add ten mentions
+around one alias, and under a single marker that alias is the eleventh line
+nobody reads.
+
+**When both are true of one line, use `legacy-name-ok`.** Some lines name a
+frozen thing and declare a dual-read at once — an image tag whose repository
+name is permanent while its version is read from either spelling. One line takes
+one marker, and the alias is the claim rule 3 wants eyes on, so it wins.
+
+There is a third case, and it takes no marker at all. Ask what breaks if the old
+spelling is not on that line. If the answer is nothing — it is prose that happens
+to name the thing — **reword it and take no exemption.** That is the test the
+floor marker's claim has to pass: reword the line, and if it is still correct,
+the marker was false.
+
 Marking a line exempt frees a slot in that file, so every new exemption is
 printed in the CI output whether the check passes or not. That is deliberate: it
 is the one move the ratchet cannot adjudicate for itself, so it is the one that
--- a/docs/plans/rebrand-sunset-plan.md
+++ b/docs/plans/rebrand-sunset-plan.md
@@ -132,6 +132,12 @@ is prose that happens to name the thing — **reword it off the literal and take
 exemption.** If a reader has to type the string, or find it on disk, or a machine has to
 match it, that is a contract: **mark it, and say which.**
 
+"Which" is a marker, not a convention in the reason text. `legacy-name-ok` is a contract
+that **bears** the old name — the alias, the redirect, the pinned wire format rule 3
+recognises. `legacy-name-floor` is one that only **names** something the rename will
+never reach. Both exempt the line identically; they are counted apart so that a sweep
+adding ten mentions cannot bury the one alias among them.
+
 Marking a mention is not a harmless extra. Every exemption widens the surface the gate
 no longer watches, and it spends the one annotation whose whole value is that its reasons
 are true.
--- a/scripts/legacy_name_ratchet.py
+++ b/scripts/legacy_name_ratchet.py
@@ -30,10 +30,44 @@ gate is pointed at. On a ``pull_request`` the checkout is already the merge
 commit, so the comparison against ``main`` is exactly this PR's contribution and
 there is nothing to maintain.
 
-**The escape hatch.** A line carrying the marker ``legacy-name-ok`` is not
-counted. Rule 3 keeps old names readable forever, so legitimate additions exist:
-a compat alias, a redirect, a test pinning the old wire format. Put the marker on
-the line with a reason and it lands in the diff where a reviewer will see it.
+**The escape hatch, and the two claims it can make.** A line carrying either
+marker is not counted. The gate cannot tell them apart — the pass/fail decision
+never reads which one is there, and nothing that failed before passes now. They
+differ only in what a reviewer has to check, which is the only thing a marker
+was ever for. Put one on the line with a reason and it lands in the diff.
+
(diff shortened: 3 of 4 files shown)


Eldad-Caura:
@claude — both review findings addressed, force-pushed to `2e50008b`.

**1. Reason bleed across markers (the real one).** Confirmed before fixing: six alias-first doubly-marked lines grouped under `6x  dual-read alias  legacy-name-floor: the command`. You were right that the reason is also the grouping key, so the pollution stops identical claims collapsing — worse than cosmetic.

Fixed by bounding a reason at the **next marker** rather than at end of line, extracted into `_reason()` beside `_kind()` so the two cannot disagree about which marker owns the line. Test added over **both** orders (`alias-first`, `floor-first`); verified it fails against the pre-fix slice in the alias-first case only, which is exactly the asymmetry you identified.

**2. The 622 vs 340+274 gap.** Real, and the arithmetic now closes: the missing 8 are the reword rows — lines that should carry no marker at all. The docstring names all three buckets.

72 tests pass, the test-file diff is still purely additive (52 pre-existing tests byte-identical), 8 implementation mutations all detected, ratchet passes on its own source, sentinel 34/34, ruff clean.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #983, opened 2026-08-25
ci: check the services against common/ instead of against Any

Eldad-Caura:
## What this fixes

`core_api.services.forge.*` is **deliberately opted back in** to the type gate — `ignore_errors = false`, one of few carve-outs from the broad `core_api.services.*` exemption — and `cd core-api && uv run mypy src/` still reported **"no issues found in 200 source files"** while `forge/cron_handler.py` imported `LLMRequest` from `common.llm` (a name that has never existed there) and called `call_with_fallback` with a signature it has never had.

The forge distill cron raised `RuntimeError` on every tick and **had never once run**. The error blamed a missing provider chain, so it read as a misconfiguration. Repaired in #955 — this PR is about why nothing caught it.

**No service's mypy run could see `common/`.** Each step `cd`s into one service and checks that service's own `src/`, and `mypy_path` named only that `src/`. So every `common.*` import resolved to nothing, and `ignore_missing_imports = true` — which every service sets, and needs, for its unstubbed third-party deps — turned that silence into `Any` rather than an error.

A service could sit in a gated tree, with a green gate, and have its use of the shared library checked by nothing at all. #948 put `common/` itself under the gate; this is the same gap one layer out and is not closed by it.

## The change

`mypy_path = "src"` → `mypy_path = "src:.."` in all four service pyprojects. `..` is the repo root, where `common/` lives; the path is resolved relative to the pyproject, so CI, a local `mypy src/` and an IDE all agree.

**Not a `MYPYPATH` env var on the CI steps.** That would leave the local run and the contributor's editor disagreeing with the gate — which is the shape of this failure, not a fix for it.

**Not packaging `common` as a distribution.** That is the more honest description of what `common/` is, and it stays available later; it is not what this PR buys. It would mean relocking every service, changing the Dockerfiles that `COPY` `common/` to `/app/common`, and changing how the runtime resolves the package — a production-adjacent change to obtain a type-check property that one line per service already obtains.

## Proof the gate bites

Re-inserting the historical import into `forge/cron_handler.py`:

```
src/core_api/services/forge/cron_handler.py:39: error: Module "common.llm" has no attribute "LLMRequest"  [attr-defined]
Found 1 error in 1 file (checked 200 source files)
```

Before this change the same file passed.

**And it has already earned itself once, on this branch.** Rebasing onto #989 (WT-2 entity canonicalisation, merged while this was open) surfaced two errors nothing else reported. Its new Phase-1.5 block binds `key = canonical_match_key(...)`, a `str`, inside a function where `key` is already the Phase-1 lookup tuple `(canonical_name, entity_type, fleet_id)` fifty lines up. Harmless at runtime — the earlier loop has finished — but it reads as the same thing and is not. `canonical_match_key` comes from `common/entity_naming.py`, which is new and which no service run could see. Renamed to `match_key`; that is the whole fix.

## Measured cost, before any fix

| service | today | with `common/` visible |
|---|---|---|
| core-api | 0 | 0 |
| core-storage-api | 0 | **34** |
| core-operations | 0 | 0 |
| core-worker | 0 | 0 |

All 34 were in one file — `postgres_service.py`, which imports this repo's SQLAlchemy models from `common.models`. **All are fixed here, nothing is exempted.** Every one was annotation-level; none was a defect in what the code does at runtime:

- **24** — a predicate list or variable inferred `BinaryExpression[bool]` from an `is_()` / `in_()`, then given the wider `ColumnElement[bool]` that a `== <optional>` comparison returns. Declared at the five list sites and above the five `tenant_pred` branches.
- **7** — `Model.__table__`, annotated `FromClause` by the declarative base (no `.delete()`, not accepted by `insert()` / `update()`) though every mapped class here is table-mapped, so at runtime it is always a `Table`. A `_table()` helper says that once instead of at each site. The Core-not-ORM targeting in `entity_embeddings_write_back` is unchanged — `_table(Entity)` **is** `Entity.__table__`, it only narrows the declared type.
- **2 narrowing** — `event_hash` is `Mapped[bytes | None]` (the column is nullable), but a NULL cannot reach the assignment: the branch above compares it against `compute_event_hash`'s `bytes` return, so a NULL row is reported `event_hash_mismatch` and returns first. Asserted, with that reasoning written down. And the entity dedup re-SELECT, which already raises on `None`, now binds the optional to its own name.
- **1** — `ins_stmt` annotated, because `literal_column("xmax")` types as `Any` and mypy will not infer a variable whose type is partly `Any`.

## Changed since the first revision

The first push exempted `postgres_service.py` with `ignore_errors = true` and deferred the 34 to a follow-up, reasoning that a gate arriving alongside a 34-error diff is a gate that gets reverted.

**The review was right and I've dropped that.** A module-level `ignore_errors` re-opens this exact gap for any *new* error in that file — and, as the review noted, that file has no coverage safety net either (it is in `omit`). The review suggested 34 per-line `# type: ignore` comments as the bounded alternative; fixing the underlying declarations turned out to be a **smaller** diff than 34 ignores, and leaves nothing suppressed. Nothing is exempted in this PR now, and there is no follow-up to remember.

## One thing this PR surfaces but does not report

**core-api's 0 above is not coverage.** Its `[[tool.mypy.overrides]] ignore_errors = true` covers `core_api.services.*`, `core_api.pipeline.steps.*`, `routes.fleet`, `routes.memories`, `clients.storage_client`, `mcp_server`, `app` and `auth`.

Neutralise those overrides and core-api goes from **343 errors to 355** with `common/` visible. Sixteen errors in six files that this change surfaces and the gate will not report, because they sit in modules exempt for unrelated reasons — including `services/memory_service.py`, which passes a `str` to `publish_memory_enrich_request(memory_id: UUID)`.

Forge was one of the few carve-outs back in, which is exactly why forge was the one that bit. Recorded here rather than fixed — it is a separate finding, and a separate decision about how much of `core_api.services.*` should stay exempt.

## Checks

- All four service mypy steps and the `common/` step green.
- `tests/` 5284 passed · `core-storage-api/tests/` 156 · `core-operations/tests/` 50 · `core-worker/tests/` 75.
- `ruff check` and `ruff format --check` clean, run separately and scoped.
- `actionlint` output on `.github/workflows/ci.yml` unchanged from `origin/main`.
- `pre-commit run --from-ref origin/main --to-ref HEAD` clean.
- No lockfile touched.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -259,6 +259,32 @@ jobs:
             exit 1
           fi
 
+      # The four steps below check each service against common/ as well as
+      # against its own src/, because every service pyproject now carries
+      # ``mypy_path = "src:.."`` — ``..`` being the repo root, where common/
+      # lives. Before that, ``common.*`` resolved to nothing, and
+      # ``ignore_missing_imports = true`` (which every service sets, and needs)
+      # turned that silence into Any rather than an error. So the step below was
+      # a gate over 200 files that could not see a single one of the ~40
+      # common/ symbols they import.
+      #
+      # What it cost: ``core_api.services.forge.*`` is deliberately opted back
+      # IN to the type gate (``ignore_errors = false``, one of few carve-outs
+      # from the broad ``core_api.services.*`` exemption), and this step still
+      # reported "no issues found in 200 source files" while
+      # ``forge/cron_handler.py`` imported ``LLMRequest`` from ``common.llm`` —
+      # a name that has never existed there — and called ``call_with_fallback``
+      # with a signature it has never had. The forge distill cron raised
+      # RuntimeError on every tick and had never once run; the error blamed a
+      # missing provider chain, so it read as a misconfiguration. Repaired in
+      # #955. Being inside a gated tree proved nothing, and that is the
+      # assumption worth distrusting.
+      #
+      # The setting is in each pyproject rather than an env var here on purpose:
+      # a ``MYPYPATH`` set only on these steps would leave the local
+      # ``mypy src/``, the contributor's IDE, and any future pre-commit hook
+      # disagreeing with the gate — which is the shape of the failure this
+      # closes, not a fix for it.
       - name: Run mypy (core-api)
         run: cd core-api && uv run mypy src/
 
@@ -279,14 +305,20 @@ jobs:
 
       - name: Run mypy (common/)
         # common/ was type-checked by NOTHING before this step. Each mypy step
-        # above ``cd``s into one service and checks that service's own src/, so
-        # no invocation's path ever resolved to common/ — while 40 files under
-        # core-api/src alone import it. Those imports were not caught either:
+        # above ``cd``s into one service and, until ``mypy_path = "src:.."``
+        # landed, checked only that service's own src/ — so no invocation's path
+        # ever resolved to common/, while 40 files under core-api/src alone
+        # import it. Those imports were not caught either:
         # ``ignore_missing_imports = true`` (every service sets it) plus a
-        # ``mypy_path`` of just that service's src/ means an unresolved
-        # ``common.*`` import degrades to Any in silence rather than failing.
+        # ``mypy_path`` of just that service's src/ meant an unresolved
+        # ``common.*`` import degraded to Any in silence rather than failing.
         # ``cd core-api && mypy src/`` reported 199 files and none of them were
         # common/'s 94.
+        #
+        # This step is still the one that OWNS common/. The service steps reach
+        # it only through the imports they happen to follow, so a common/ module
+        # no service imports — or a branch of one none of them exercises — is
+        # covered here and nowhere else. Keep it.
         #
         # Nor did the hook cover it: .pre-commit-config.yaml deliberately runs no
         # mypy, deferring to "CI is the authoritative type-check" — which pointed
--- a/core-api/pyproject.toml
+++ b/core-api/pyproject.toml
@@ -138,7 +138,15 @@ indent-style = "space"
 
 [tool.mypy]
 python_version = "3.12"
-mypy_path = "src"
+# ``..`` is the repo root, where ``common/`` lives. common/ is not an installed
+# distribution — it is PYTHONPATH-mounted (the Dockerfile COPYs it to
+# /app/common), so without the repo root here every ``common.*`` import degrades
+# to Any under ``ignore_missing_imports`` and this service's use of the shared
+# library is checked by nothing. That is what let ``from common.llm import
+# LLMRequest`` — a name that has never existed there — ship in the forge cron and
+# raise on every tick (#955) while ``mypy src/`` reported no issues in 200 files.
+# Relative to this file, so CI, a local ``mypy src/`` and an IDE all agree.
+mypy_path = "src:.."
 explicit_package_bases = true
 warn_return_any = false
 warn_unused_ignores = false
--- a/core-operations/pyproject.toml
+++ b/core-operations/pyproject.toml
@@ -76,7 +76,15 @@ indent-style = "space"
 
 [tool.mypy]
 python_version = "3.12"
-mypy_path = "src"
+# ``..`` is the repo root, where ``common/`` lives. common/ is not an installed
+# distribution — it is PYTHONPATH-mounted (the Dockerfile COPYs it to
+# /app/common), so without the repo root here every ``common.*`` import degrades
+# to Any under ``ignore_missing_imports`` and this service's use of the shared
+# library is checked by nothing. That is what let ``from common.llm import
+# LLMRequest`` — a name that has never existed there — ship in the forge cron and
+# raise on every tick (#955) while ``mypy src/`` reported no issues in 200 files.
+# Relative to this file, so CI, a local ``mypy src/`` and an IDE all agree.
+mypy_path = "src:.."
 explicit_package_bases = true
 warn_return_any = false
 warn_unused_ignores = false
--- a/core-storage-api/pyproject.toml
+++ b/core-storage-api/pyproject.toml
@@ -98,7 +98,15 @@ exclude = ["**/migrations/env.py"]  # late-import pattern; lint-only via per-fil
 
 [tool.mypy]
 python_version = "3.12"
-mypy_path = "src"
+# ``..`` is the repo root, where ``common/`` lives. common/ is not an installed
+# distribution — it is PYTHONPATH-mounted (the Dockerfile COPYs it to
+# /app/common), so without the repo root here every ``common.*`` import degrades
+# to Any under ``ignore_missing_imports`` and this service's use of the shared
+# library is checked by nothing. That is what let ``from common.llm import
+# LLMRequest`` — a name that has never existed there — ship in the forge cron and
+# raise on every tick (#955) while ``mypy src/`` reported no issues in 200 files.
+# Relative to this file, so CI, a local ``mypy src/`` and an IDE all agree.
+mypy_path = "src:.."
 explicit_package_bases = true
 warn_return_any = false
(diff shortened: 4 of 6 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
Reviewed against `2ea15fa4`, which was the head at the time. Force-pushed `077a9224` — the finding is fixed, and I took the stronger of the two options you offered.

**On the `ignore_errors` finding: you're right, and I've removed the override entirely.**

My reasoning for the exemption was that it wasn't a *regression* — nothing had ever checked that file against `common/`, so it stayed where it was. That's true and it was also beside the point. The objection that lands is the one you made: a module-level `ignore_errors` is unbounded forward in time. It exempts errors that don't exist yet, including a fresh `common.*` symbol or signature mismatch — the exact bug this PR exists to catch — and you're right that this file has no coverage safety net either, so there was nothing else watching it.

**I fixed the 34 rather than taking the per-line-ignore option, because the fix turned out to be smaller.** The 34 errors come from far fewer than 34 causes:

- 24 collapse into **10 declarations** — five predicate lists and five `tenant_pred` branches, each inferred `BinaryExpression[bool]` from an `is_()`/`in_()` and then handed the wider `ColumnElement[bool]` a `== <optional>` returns.
- 7 are `Model.__table__` (annotated `FromClause`; a `Table` at runtime), now routed through one `_table()` helper.
- 3 are singletons: two narrowings and one annotation.

So it's ~14 edits against your 34 `# type: ignore` comments, and it leaves nothing suppressed and no follow-up to remember. `postgres_service.py` is fully gated as of this push.

Two things I want to flag rather than have you find:

**1. The `assert` is new code in a hot loop, and I want it challenged if you disagree.** `event_hash` is `Mapped[bytes | None]` — genuinely nullable in the model and the DB. I convinced myself a NULL cannot reach the assignment: the `else` branch above compares it against `compute_event_hash(...)`, which returns `bytes`, so a NULL row is reported `event_hash_mismatch` and returns first. The assert encodes that and the comment says why. If the reasoning is wrong, the right fix is widening `expected_prev` to `bytes | None`, not deleting the assert.

**2. This PR now contains a two-line change to code from #989, which merged while it was open.** Rebasing onto it surfaced two errors nothing else reported: its new Phase-1.5 block binds `key = canonical_match_key(...)`, a `str`, in a function where `key` is already the Phase-1 lookup tuple `(canonical_name, entity_type, fleet_id)`. Harmless at runtime, since the earlier loop has finished. `canonical_match_key` lives in `common/entity_naming.py` — new, and invisible to every service run until this PR. Renamed to `match_key`. I'd rather it were reviewed as part of #989's story than buried here, so: flagging it.

Full suite green: `tests/` 5284, `core-storage-api/tests/` 156, `core-operations/tests/` 50, `core-worker/tests/` 75.

@claude please re-review — the review job skips on `synchronize`, so this push is otherwise unreviewed.

Eldad-Caura:
Reviewed `077a9224`; force-pushed `dffe28ce`. Took the suggestion, and the Low finding is already answered by the run you reviewed.

**On the `_table()` suggestion — applied.** It is now `def _table(model: type[Base]) -> Table`, using this repo's own declarative base from `common.models` rather than `DeclarativeBase`. Verified it does what you wanted:

```
probe.py:8: error: Argument 1 to "_table" has incompatible type "type[NotAModel]"; expected "type[Base]"  [arg-type]
```

The cost is one `# type: ignore[return-value]` on the return, since `Base.__table__` is annotated `FromClause`. That is a fair trade and arguably the point of having a single helper: the `FromClause`-is-really-a-`Table` claim is now made in exactly one place, next to the paragraph explaining it, instead of implied at seven call sites.

I checked the ignore is not one of the dead ones — `--warn-unused-ignores` does not flag it, and removing it produces exactly `[return-value]`:

```
postgres_service.py:204: error: Incompatible return value type (got "FromClause", expected "Table")  [return-value]
```

**On the Low finding — core-api, core-operations and core-worker were checked, and the evidence is the run you reviewed.** All four service mypy steps execute in the `CI` job on every push, and it reported `SUCCESS` on `077a9224`. The absence of source fixes for those three is the finding: with `common/` visible they report **0**, so there was nothing to fix.

Per-service, measured before any change:

| service | today | with `common/` visible |
|---|---|---|
| core-api | 0 | 0 |
| core-storage-api | 0 | **34** |
| core-operations | 0 | 0 |
| core-worker | 0 | 0 |

You were right to ask about core-api's carve-outs specifically, and the honest answer is in the PR body: **core-api's 0 is not coverage.** `[[tool.mypy.overrides]] ignore_errors = true` covers `core_api.services.*`, `core_api.pipeline.steps.*`, `routes.fleet`, `routes.memories`, `clients.storage_client`, `mcp_server`, `app` and `auth`. Neutralise those and core-api goes 343 → 355 with `common/` visible: sixteen errors in six files that this change surfaces and the gate will not report, because those modules are exempt for reasons that predate it. Not fixed here — it is a separate finding and a separate decision about how much of `core_api.services.*` should stay exempt. `services.forge.*` is one of the few carve-outs back **in**, which is why forge was the one that bit.

The carve-outs are untouched by this PR; the only override I changed was the `postgres_service` one, which I added in the first revision and removed in the second.

Eldad-Caura:
@erni-a — green and ready for your approval. One timing note at the bottom.

**Gates, all on `fda2d34f`** (your rebase onto #991): CI, CodeQL (python + js/ts), dco, Security Scan, `core-api` image version bake, Legacy-name ratchet, Do-not-touch sentinel — all SUCCESS.

**Two Claude review rounds, both findings taken:**
1. *Medium* — the first push exempted `postgres_service.py` with `ignore_errors = true`. Right objection: a module-level ignore is unbounded forward in time, so it would have re-opened this exact gap for any *new* error in that file, which also has no coverage safety net. **Dropped.** All 34 errors are fixed; nothing is exempted and there is no follow-up PR to remember.
2. *Suggestion* — type the `_table()` helper as the declarative base rather than `type[Any]`. **Applied**, verified: passing a non-mapped class is now `Argument 1 to "_table" has incompatible type "type[NotAModel]"; expected "type[Base]"`.

**Changed since the last automated review** (which ran on `077a9224`, and the review job skips on `synchronize`): only item 2 — the `_table` signature, one import, and one `# type: ignore[return-value]` on its return. I confirmed that ignore is load-bearing rather than decorative: `--warn-unused-ignores` does not flag it, and removing it produces exactly `[return-value]`. Everything else is identical to what was reviewed.

**Why it's worth landing:** `services/forge/*` is opted *into* the type gate, and `mypy src/` still passed while the forge cron imported a name from `common.llm` that has never existed — it raised on every tick and had never run (#955). No service run could see `common/` at all. This is that fix. Proof it now bites is in the PR body.

**It has already caught something.** Rebasing onto #989 surfaced two errors nothing else reported — its new Phase-1.5 block rebinds `key` (the Phase-1 lookup tuple) to a `str` from `common/entity_naming.py`. Harmless at runtime, renamed to `match_key` here, flagged in the review thread so it isn't buried.

**Tests:** `tests/` 5284 · `core-storage-api/tests/` 156 · `core-operations/tests/` 50 · `core-worker/tests/` 75.

---

**On merge timing.** The lane holding the Pub/Sub cutover reported a live incident earlier today — `subscriptions-per-project` exhausted at 10006/10000 — and argued that *any* merge is risky while that stands, because a rolling Cloud Run deploy transiently raises the subscription count and `_ensure_broadcast_subscription` degrades silently rather than failing startup, so a bad deploy looks clean.

**I have not verified that and it may already be resolved** — #991 ("release ephemeral subscriptions before the slow shutdown steps") is exactly that shape and is now on `main`, underneath this branch. Passing it on rather than deciding: worth a glance at project quota headroom before merging, and nothing here is urgent enough to need it today.

erni-a (approved):

PR #985, opened 2026-08-25
fix(api): rate-limited routes without a Response param 500 on every call

arkash20:
## Summary

**Completes the D14 headers_enabled fix**: `POST /documents` still 500s on every call on current main — the same slowapi bug #984 fixed for `/recall` and `/ingest/commit` (staging outage, live since 2026-08-25 17:51Z).

slowapi with `headers_enabled=True` (D14, #976) injects `X-RateLimit-*`/`Retry-After` into the endpoint's injectable `Response`. A rate-limited handler with no `Response`-annotated parameter raises inside `_inject_headers` on **every** request. #984 covered two of the three broken handlers; `upsert_document` is the third.

> Rebased over #984: originally this PR fixed all three; the `/recall` + `/ingest/commit` halves were dropped in favor of #984's merged fix. What remains is the `/documents` fix plus a structural guard #984's behavioral test doesn't provide.

## What changed

- `response: Response` added to `upsert_document` (with a comment explaining why it must stay).
- New guard `tests/test_d14_rate_limited_response_param.py`: AST-based — every `write_limit`/`search_limit`/`write_bulk_limit`-decorated handler must declare a `Response` param, plus a handler-count floor so a decorator rename can't make it pass vacuously. Unlike a per-endpoint behavioral test, this fails the moment any future rate-limited route ships without the param.
- `tests/test_documents_rest_doc_memory.py` direct-invocation call sites updated to pass a `Response()`.

## Testing

- Full root suite post-rebase: **5248 passed** (includes #984's `tests/test_rate_limit.py`).
- **Wet-verified** on the local enterprise compose stack: `POST /documents` 200 with `x-ratelimit-limit/remaining/reset` + `retry-after` (was 500 unconditionally).

Refs: D14 (#976) follow-up; completes #984.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #990, opened 2026-08-25
feat(api): complete the OpenAPI spec — response schemas + servers block (C33)

arkash20:
## Summary

C33 (register API-10), OSS half. Before: **66 of 104** success responses in `/api/openapi.json` were empty `{}` schemas and the spec had no `servers` block — agents and generated clients got zero response shapes. After: **23 remain**, all deliberate skips (STM routes — dead feature; plugin/skill delivery payloads — script/text; admin internals).

## What changed

- **`core_api/openapi_responses.py`** — 44 spec-only response models attached to 43 endpoints via `responses={200: {"model": ...}}`. Deliberately **not** `response_model=`: that would route responses through Pydantic serialization (coercing values, dropping undeclared keys) — a runtime behavior change. Spec-only attachment moves only the spec; the wire is untouched. Shapes were extracted from the handlers and their storage calls, not guessed; conditional keys carry when-they-appear descriptions (e.g. ingest/preview's four branches, health's degraded/unhealthy keys, keystone `data.agent_id` only when scope=agent).
- **`servers` block** — new `public_api_url` setting (`PUBLIC_API_URL`). Empty (OSS default): no servers block, spec byte-identical to before. Set (`https://memclaw.dev`): spec gains `servers=[{url}]`. The enterprise deploy env rides the C33 enterprise PR.
- **Guards** — `tests/test_c33_openapi_completeness.py`: every annotated endpoint must surface a real schema; empty-success-schema count **ratcheted at ≤23** (new routes can't ship blank); servers behavior asserted both ways.
- **Broker baseline** regenerated — additive only (`HealthResponse`/`VersionResponse` now typed on the two broker-set ops); their `_UNTYPED_200` allowlist entries removed, which that guard itself demanded.

## Testing

- Full root suite: **5252 passed**; mypy/ruff clean.
- **Live wet sweep** on the local enterprise compose stack (fresh image, fresh tenant): 22 endpoints exercised end-to-end — each live JSON response `model_validate`d against its model **and** checked for undocumented top-level keys (the docs must not lie by omission). All PASS. Live spec: exactly 23 empty schemas; no servers block without the env (unit tests cover the env-set path both ways).
- Wet testing this slice also caught the `POST /documents` D14 500 fixed in #985.

Found & filed along the way: **C36 (High)** — `GET/PUT /settings` returns `api_keys` verbatim while the docstring claims masking (separate task, not touched here).

Refs: canonical C33 / register API-10. Enterprise half (bare `/openapi.json` redirect, llms-full.txt renderer, `PUBLIC_API_URL` env, refreshed `openapi.snapshot.json`) follows as a separate PR.

### DIFF
--- a/core-api/openapi.broker.json
+++ b/core-api/openapi.broker.json
@@ -479,6 +479,68 @@
         "title": "HTTPValidationError",
         "type": "object"
       },
+      "HealthResponse": {
+        "properties": {
+          "event_bus": {
+            "description": "ok, unhealthy, or error.",
+            "title": "Event Bus",
+            "type": "string"
+          },
+          "platform_init_errors": {
+            "anyOf": [
+              {
+                "items": {
+                  "type": "string"
+                },
+                "type": "array"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "description": "Only when platform init recorded errors and all dependencies are up (status=degraded).",
+            "title": "Platform Init Errors"
+          },
+          "redis": {
+            "description": "connected, unavailable, or not configured.",
+            "title": "Redis",
+            "type": "string"
+          },
+          "status": {
+            "description": "ok, degraded, or unhealthy (unhealthy is served as 503).",
+            "title": "Status",
+            "type": "string"
+          },
+          "storage": {
+            "description": "connected or unreachable.",
+            "title": "Storage",
+            "type": "string"
+          },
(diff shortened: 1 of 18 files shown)


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #991, opened 2026-08-25
fix(events): release ephemeral subscriptions before the slow shutdown steps

Eldad-Caura:
**Fixes the leak behind tonight's prod incident.** Does not fix the capacity problem underneath it — see the end.

## What happened

`core-api` leaked one ephemeral broadcast subscription **per worker, per unclean shutdown**. It accumulated **6,571** in staging against a live instance count in the low tens and exhausted the **10,000 subscriptions-per-project** cap.

That cap is shared across environments, so the failure surfaced in **prod**:

```
RESOURCE_EXHAUSTED: Your project has exceeded a limit:
(type="subscriptions-per-project", current=10006, maximum=10000)
```

Prod `core-api` could no longer create its own subscription, so cross-process cache invalidation degraded to the 5-minute TTL. The retry churn then pushed the container past its 512 MiB limit.

## The cause is ordering, not the delete

Cloud Run allows **10s** between SIGTERM and SIGKILL. `app.py`'s `shutdown_steps` are awaited **sequentially**, and the first three carry `timeout=5.0` each. On any shutdown where a queue has work to flush, the budget is spent before `event_bus.stop()` is reached — and the delete lived inside `stop()`.

The process is killed, the subscription survives, and its `expiration_policy` holds project quota for a full day. `86400s` is the Pub/Sub minimum, so it cannot be tuned down.

## The natural experiment that confirms it

| service | `event_bus.stop()` position | subscriptions |
|---|---|---|
| `platform-auth-api` | early, behind one suppressed task-cancel | **2–6** |
| `core-api` | **4th**, behind 3 × `timeout=5.0` flushes | **6,571** |

Same library, same TTL, same mechanism. Different position in the shutdown. Three orders of magnitude apart.

## The change

`release_broadcast_subscriptions()` is the delete alone, split out of `stop()` so it can run **first**. It needs only the subscriber client and the recorded paths — no pull loops stopped, no queues flushed, no storage open — so it cannot be starved by the steps that were starving it.

**`stop()` still calls it**, so this is an idempotent hoist, not a move: a bus torn down without the new call behaves exactly as before.

Added to the `EventBus` ABC as a **no-op default**, so a shutdown path can call it without knowing which implementation it holds. `InProcessEventBus` has no external resource to release; overriding it would be inventing work.

## Tests

Both confirmed failing without the change (`AttributeError` on each). They pin the two properties the fix rests on:

- **the release works on a fully live bus**, with no teardown run first — that is the entire point, and a release that closed the subscriber would break the ordering it exists to enable
- **it is idempotent** — `stop()` calls it again on every clean shutdown, and a second delete would log a `NotFound` per topic on every graceful exit, training readers to ignore the one warning that matters

## What this does not do

**Staging still holds ~6,558 subscriptions on one topic against a project-wide cap of 10,000.** This stops the leak. It does not reclaim what has already leaked, and it does not make the cap comfortable — a quota increase or a change to how `org.settings-changed` fans out is still needed, or the next hard staging churn repeats tonight.

## Gates

- `Do-not-touch sentinel` — 34/34
- `Legacy-name ratchet` — no new lines
- `ruff check` / `ruff format` — clean
- `pytest tests/test_events/` — 127 passed

### DIFF
--- a/common/events/base.py
+++ b/common/events/base.py
@@ -122,6 +122,19 @@ class EventBus(ABC):
     async def stop(self) -> None:
         """Drain + shut down. Called on graceful shutdown."""
 
+    async def release_broadcast_subscriptions(self) -> None:
+        """Hand back any per-process broadcast subscriptions. No-op by default.
+
+        Exists on the base class so a shutdown path can call it unconditionally,
+        without knowing which implementation it holds. Only the Pub/Sub bus has
+        anything to release; an in-process bus has no external resource and
+        overriding it would be inventing work.
+
+        Separate from ``stop()`` because it must be runnable FIRST, inside the
+        platform's SIGTERM budget, ahead of teardown that may not finish. See
+        the Pub/Sub implementation for why that ordering is load-bearing.
+        """
+
     @property
     def is_healthy(self) -> bool:
         """True when the bus can still deliver events end-to-end.
--- a/common/events/pubsub.py
+++ b/common/events/pubsub.py
@@ -245,6 +245,13 @@ class PubSubEventBus(EventBus):
         # this process created so ``stop()`` can delete them.
         self._broadcast_topics: set[str] = set()
         self._broadcast_sub_paths: list[str] = []
+        # Short names of subscriptions THIS bus deliberately deleted, so a pull
+        # loop that is still running when the delete lands can tell "I released
+        # this myself" from "this was never provisioned". Both surface as
+        # NotFound on the next pull; only the second is an error worth the loud
+        # halting-the-loop log. Short names because that is what ``_pull_loop``
+        # is given — ``_broadcast_sub_paths`` holds full resource paths.
+        self._released_subscriptions: set[str] = set()
         self._instance_id = uuid.uuid4().hex[:12]
 
     def _spawn_background_task(self, coro: Any) -> asyncio.Task[Any]:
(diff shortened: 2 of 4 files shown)


Eldad-Caura:
@claude Fixed the HIGH finding — force-pushed as `0f4de521`.

You were right, and the mechanism is exactly as described: hoisting the delete ahead of `stop()` leaves the pull loops live when their subscriptions vanish, so every graceful shutdown would have taken the permanent-error path — ERROR log, `_failed_subscriptions` entry, `is_healthy` false. I removed the comment that documented that ordering and then broke the thing it documented.

**I did not take the suggested fix.** Setting `_stopping` early is smaller, but `_ensure_publisher()` raises `RuntimeError("PubSubEventBus is stopped")` off that same flag (`pubsub.py:291`), so it would break the publishes the three flush steps still need to make — trading a spurious error for a real one.

Instead the bus records the short names it deliberately released, and the `NotFound` branch returns quietly for those only:

- **Recorded before the delete, not after** — the loop can observe the deletion the instant it lands, so recording afterwards leaves a window where the loud path still fires.
- **`start()` clears the record** — names are deterministic in a per-object `_instance_id`, so start/stop/start recreates the *same* names; a stale entry would silence a genuine fault on the second run.
- **Scoped to what we deleted** — a subscription that was never provisioned still halts its loop loudly, which is the behaviour the branch exists for.

Two tests pin the pair, and I verified them by reintroducing the exact reviewed state (suppression removed, hoist kept):

```
test_released_subscription_exits_pull_loop_quietly   FAILED  (NotFound: 404 gone)
test_unprovisioned_subscription_still_reports_loudly PASSED
```

The second passing under the mutation is the load-bearing half — it shows the suppression is narrow rather than a blanket swallow of `NotFound`.

Gates: sentinel 34/34, ratchet no new lines, mypy clean over `common/` (97 files), ruff clean, 130 tests in `tests/test_events` pass.

One note on scope, unchanged from the original description: this stops the leak, it does not reclaim what has already leaked, and it does not make the project-wide cap comfortable. Draining the backlog and the quota increase are being handled separately.

Eldad-Caura:
@erni-a plesae approve

erni-a (approved):

PR #993, opened 2026-08-25
feat(common): allow a self-hosted base URL for the chat LLM path

### LINKED ISSUE #992: [Feature] let the chat LLM path use any OpenAI-compatible endpoint
### Problem

The chat LLM path (dedup judge, enrichment, contradiction detection, entity extraction) cannot use any endpoint except each provider's hard coded hosted URL, and it sends a `response_format` that only hosted OpenAI accepts. Both gaps have the same effect: the call fails, `call_with_fallback` walks to the fake provider, and every LLM feature silently runs on heuristics. Nothing in the API or the health output shows that this happened; `common.llm.retry` logs a warning and the write succeeds.

Two things cause it.

**1. The base URLs are constants.** `OPENAI_CHAT_BASE_URL`, `ANTHROPIC_CHAT_BASE_URL`, and `OPENROUTER_CHAT_BASE_URL` are module constants in `common/llm/constants.py`, and `_credentials.py` passes them to `OpenAILLMProvider` explicitly, which also defeats the SDK's own `OPENAI_BASE_URL` variable. The comment above them says the URL is "swapped via tenant-config override", but no tenant-config field carries a URL; only the provider name is swapped, and each name has its own hard-wired URL. The embedding path already has `OPENAI_EMBEDDING_BASE_URL`; the chat path has no equivalent. Further, a service needing a direct line out in my current production environment (and I assume others in my industry) is hard NO due to security and compliance concerns.

**2. The `response_format` shapes are hosted-OpenAI only.** `OpenAILLMProvider.complete_json` sends `{"type": "json_object"}` when the caller passes no schema, and `json_schema` with `strict: false` when it does. Observed on 2.30.0 (I'm running LM Studio, but I suspect Ollama and similar would have the same issue):

- LM Studio (0.4.21) rejects `json_object`: `'response_format.type' must be 'json_schema' or 'text'`.
- Anthropic's OpenAI-compatible endpoint (which `ENTITY_EXTRACTION_PROVIDER=anthropic` uses) rejects both. No-schema callers get `response_format.type: Input should be 'json_schema'`; entity extraction gets `response_format.json_schema.strict: Input should be True`. With a funded key, all 44 chat calls in one write sweep returned HTTP 400 and fell back to the fake provider. So the shipped Anthropic setting cannot complete a single JSON call.

Direct probes of the Anthropic endpoint show what it does accept: `json_schema` with `strict: true` and a closed schema (`additionalProperties: false` on every object). A permissive strict schema is rejected, `type: text` is rejected, and omitting `response_format` works but the model wraps the JSON in a Markdown code fence, which the bare `json.loads` in `complete_json` does not survive.

### Proposed solution

Still going through things, but from what i've gleaned so far, should be straight forward and consistent with what the code already does.

1. Read the three chat base URLs from the environment, with the current values as defaults, the same way `ANTHROPIC_DEFAULT_MODEL` is read two lines below. Keep the literal hosted URL in a separate constant so the provider can tell whether it is talking to `api.openai.com` after an override. Fix the stale comment. Document the variables in `.env.example` next to `OPENAI_EMBEDDING_BASE_URL`.
2. In `complete_json`, send what the endpoint accepts. Hosted OpenAI keeps today's shapes. Any other base URL gets no `response_format` without a schema, and `strict: true` with a closed schema when one is given (a small helper closes each object, lists every property under `required`, and drops `default` and `title`). Strip a Markdown code fence before `json.loads`.

An operator then sets `OPENAI_CHAT_BASE_URL=http://host.docker.internal:1234/v1` and the judge runs on a local model, or sets `ENTITY_EXTRACTION_PROVIDER=anthropic` and it runs on Anthropic. Hosted OpenAI behaviour does not change.

### Alternatives considered

- **The tenant provider name.** It only swaps between the three hosted URLs.
- **The SDK's `OPENAI_BASE_URL`.** Defeated, because the constructor passes `base_url` explicitly.
- **A `sitecustomize` patch of the constants.** Works, but it is not a supported configuration and does not fix the `response_format` half.
- **A native Anthropic provider on the Messages API.** The right long-term shape for that provider, and a larger change. This proposal makes the existing compatible path work first.

### Affected surface

Server configuration / env var

Yes, I can open a PR

### Additional context

Reproduced against Caura 2.30.0 with LM Studio 0.3.x serving `qwen2.5-coder-7b-instruct`, and against `https://api.anthropic.com/v1` with `claude-haiku-4-5-20251001`. A branch with the change, tests, and a CHANGELOG entry is ready; the PR follows this issue. With the patched image, the dedup judge on Haiku 4.5 rejects a paraphrase in the 0.85 to 0.97 band and admits a refinement at confidence 0.90, which the shipped build cannot do from any provider setting except a funded hosted OpenAI key.

### Pre-flight

zznate:
## Summary

Let the chat LLM path (dedup judge, enrichment, contradiction detection, entity extraction) use any OpenAI-compatible endpoint. The three chat base URLs now come from the environment, and `complete_json` sends a `response_format` the endpoint accepts. Hosted OpenAI behaviour does not change.

## Related Issue

Closes #992

## Type of Change

- [ ] Documentation update
- [ ] Other:

## How Has This Been Tested?

New tests in `tests/test_llm_openai_compatible_json_mode.py` build the real provider and swap only its transport, the way `test_llm_provider_sdk_retries.py` does. They pin:

- hosted OpenAI still gets `json_object` without a schema and a non-strict `json_schema` with one;
- any other base URL gets no `response_format` without a schema, and `strict: true` with a closed schema when one is given, including nested `$defs`;
- the schema helper does not modify its input and drops `default` and `title`;
- a reply wrapped in a Markdown code fence parses;
- each `*_CHAT_BASE_URL` variable wins over the default, checked in a fresh interpreter because the constants are read at import time.

```
core-api/.venv/bin/ruff check common/llm tests/test_llm_openai_compatible_json_mode.py   # All checks passed
core-api/.venv/bin/ruff format --check common/llm tests/test_llm_openai_compatible_json_mode.py
core-api/.venv/bin/mypy common/llm/providers/openai.py common/llm/constants.py             # no errors in these files
core-api/.venv/bin/pytest tests/                                                           # 5315 passed, rebased on 71e9981
```

Verified live against 2.30.0 plus this change, built with `core-api/Dockerfile`:

- `ENTITY_EXTRACTION_PROVIDER=anthropic` with `claude-haiku-4-5-20251001`: all chat calls succeed. The dedup judge rejects a paraphrase in the judge band, admits a refinement and a negation at confidence 0.90, and the entity extraction calls complete through the strict schema with no fallback warning in the log. On the shipped build every one of these calls returned HTTP 400 and fell back to the fake provider.
- `OpenAILLMProvider.complete_json` with `base_url=http://localhost:1234/v1` against LM Studio serving `qwen2.5-coder-7b-instruct`: the no-schema path returns a parsed verdict, where the shipped build was rejected with `'response_format.type' must be 'json_schema' or 'text'`.

## Checklist

- [x] `mypy` passes
- [x] `pytest` passes locally

## Additional Notes

- `OPENAI_HOSTED_CHAT_BASE_URL` keeps the literal hosted URL so the provider can tell, after an override, whether it is talking to `api.openai.com`. That is the one endpoint that accepts `json_object`, so it keeps the old shapes and nothing changes for existing deployments.
- The strict-schema helper exists because Anthropic's compatible endpoint requires `strict: true` and rejects any object without `additionalProperties: false`. Pydantic-generated schemas leave both open for fields with defaults. The helper closes them on a copy; the models are untouched, and the client-side Pydantic parse stays the real guard, as the docstring already says.
- `CHANGELOG.md` had no `Unreleased` section; this adds one above 2.30.0.
- A native Anthropic provider on the Messages API would be the deeper fix for that provider. This change makes the existing compatible path work and is the smaller step.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -102,6 +102,19 @@ OPENAI_API_KEY=
 # OPENAI_EMBEDDING_MODEL=BAAI/bge-m3
 # TEI rejects the ``dimensions=`` SDK kwarg — must be false when talking to TEI.
 # OPENAI_EMBEDDING_SEND_DIMENSIONS=false
+
+# -- Self-hosted chat LLM (opt-in) -------------------------------------------
+# The chat path (dedup judge, enrichment, contradiction detection, entity
+# extraction) talks to each provider's hosted URL by default. Each URL can
+# be overridden, so ``ENTITY_EXTRACTION_PROVIDER=openai`` can point at any
+# OpenAI-compatible server: LM Studio, vLLM, or a gateway. Set the matching
+# API key to whatever the server expects (LM Studio ignores it) and name a
+# loaded model in ``ENTITY_EXTRACTION_MODEL``. From inside a container,
+# ``host.docker.internal`` reaches a server on the host.
+#
+# OPENAI_CHAT_BASE_URL=http://host.docker.internal:1234/v1
+# ANTHROPIC_CHAT_BASE_URL=https://api.anthropic.com/v1
+# OPENROUTER_CHAT_BASE_URL=https://openrouter.ai/api/v1
 # Optional: only set when running an instruction-aware model (Qwen3-Embedding,
 # e5-instruct). bge-m3 is symmetric — leave empty.
 # EMBEDDING_QUERY_INSTRUCTION=
--- a/common/llm/constants.py
+++ b/common/llm/constants.py
@@ -19,20 +19,30 @@ GEMINI_DEFAULT_MODEL = os.environ.get(
     "GEMINI_DEFAULT_MODEL", "gemini-3.1-flash-lite-preview"
 )
 
-# OpenAI's chat-completions base URL — used by ``OpenAILLMProvider``
-# (and for openrouter / anthropic compat where the API mirrors OpenAI's
-# shape, with the base URL swapped via tenant-config override).
-OPENAI_CHAT_BASE_URL = "https://api.openai.com/v1"
+# Chat-completions base URLs. ``OpenAILLMProvider`` works against any
+# OpenAI-compatible endpoint by varying ``base_url``; the registry picks
+# the URL from the provider name (``ProviderName``). Each URL can be
+# overridden from the environment, so a provider can talk to a self-hosted
+# server (LM Studio, vLLM, a gateway) without a code change.
+#
+# ``OPENAI_HOSTED_CHAT_BASE_URL`` is the literal hosted endpoint. It stays
+# separate from the overridable value so the provider can tell whether it
+# is talking to api.openai.com, which accepts request shapes that other
+# compatible servers reject.
+OPENAI_HOSTED_CHAT_BASE_URL = "https://api.openai.com/v1"
+OPENAI_CHAT_BASE_URL = os.environ.get(
+    "OPENAI_CHAT_BASE_URL", OPENAI_HOSTED_CHAT_BASE_URL
+)
 
-# Anthropic + OpenRouter base URLs and default models. The
-# ``OpenAILLMProvider`` works against any of these endpoints by varying
-# ``base_url``; the registry picks the right tuple based on
-# ``ProviderName``.
-ANTHROPIC_CHAT_BASE_URL = "https://api.anthropic.com/v1"
+ANTHROPIC_CHAT_BASE_URL = os.environ.get(
+    "ANTHROPIC_CHAT_BASE_URL", "https://api.anthropic.com/v1"
+)
 ANTHROPIC_DEFAULT_MODEL = os.environ.get(
     "ANTHROPIC_DEFAULT_MODEL", "claude-haiku-4-5-20251001"
 )  # Anthropic API requires native model IDs
-OPENROUTER_CHAT_BASE_URL = "https://openrouter.ai/api/v1"
+OPENROUTER_CHAT_BASE_URL = os.environ.get(
+    "OPENROUTER_CHAT_BASE_URL", "https://openrouter.ai/api/v1"
+)
 OPENROUTER_DEFAULT_MODEL = os.environ.get(
     "OPENROUTER_DEFAULT_MODEL", "openai/gpt-5.4-nano"
 )
--- a/common/llm/providers/openai.py
+++ b/common/llm/providers/openai.py
@@ -18,6 +18,7 @@ from __future__ import annotations
 import json
 import logging
 import time
+from urllib.parse import urlsplit
 
 import httpx
 import openai
@@ -89,9 +90,41 @@ def _usage_tokens(response) -> tuple[int, int, int]:
     )
 
 
+_HOSTED_OPENAI_HOST = urlsplit(OPENAI_HOSTED_CHAT_BASE_URL).hostname or ""
+_LOOPBACK_HOSTS = frozenset({"localhost", "127.0.0.1", "::1"})
+
+
 def _is_hosted_openai(base_url: str) -> bool:
-    """True when ``base_url`` is api.openai.com itself, after any override."""
-    return base_url.rstrip("/") == OPENAI_HOSTED_CHAT_BASE_URL
+    """True when ``base_url`` points at api.openai.com, whatever the scheme.
+
+    The decision is by host, not by string: ``http://api.openai.com/v1`` or
+    a trailing slash still count as hosted. A proxy or regional alias on
+    another host does not, on purpose: the request shapes below are chosen
+    for what the *server* accepts, and a proxy that is not api.openai.com
+    has to be assumed to take the strict shapes every compatible server
+    takes. Pointing ``OPENAI_CHAT_BASE_URL`` at such a proxy therefore
+    gives it the non-hosted shapes, which hosted OpenAI accepts as well.
+    """
+    return (urlsplit(base_url).hostname or "").lower() == _HOSTED_OPENAI_HOST
+
+
+def _is_plaintext_off_host(base_url: str) -> bool:
+    """True when ``base_url`` is plain ``http`` to a host that is not loopback."""
+    parts = urlsplit(base_url)
+    return (
+        parts.scheme == "http" and (parts.hostname or "").lower() not in _LOOPBACK_HOSTS
+    )
+
+
+def _allows_null(schema: dict) -> bool:
+    """True when ``schema`` already admits ``null``."""
+    t = schema.get("type")
+    if t == "null" or (isinstance(t, list) and "null" in t):
+        return True
+    return any(
+        isinstance(opt, dict) and opt.get("type") == "null"
+        for opt in schema.get("anyOf", []) + schema.get("oneOf", [])
+    )
 
 
 def _strict_schema(schema: dict) -> dict:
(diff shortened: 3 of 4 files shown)


erni-a:
Thanks for this — it's obvious you ran it against a real self-hosted stack rather than writing to the docs, and the `host.docker.internal` note in `.env.example` is the kind of detail that saves the next person an hour. I've approved the workflows so CI runs.

I validated it against current `main` locally as well. Recording what came back clean, so none of it gets re-litigated in review:

- Applies onto today's `main`; full-suite delta is **zero** (same 16 pre-existing local failures with and without it).
- The module-level `os.environ.get` pattern matches what `constants.py` already does for `GEMINI_DEFAULT_MODEL`, `ANTHROPIC_DEFAULT_MODEL` and friends, so it's consistent with the file rather than a new idiom.
- The override changes only the constructor *default* — `registry.py` still passes `base_url=` explicitly, so per-tenant configuration keeps precedence. That's the right layering.
- `test_complete_json_preserves_json_object_when_no_schema` (the deliberate back-compat pin in `tests/test_a5b_prompt_and_schema.py`) still passes, because hosted OpenAI keeps its old shape. Good — that's the test I expected this to trip.
- `_strict_schema` not mutating its input is pinned by its own test. Appreciated.

Four things before this can land, roughly in order of how much they matter:

**1. The non-hosted `response_format` change needs evidence.** The default Anthropic and OpenRouter base URLs aren't `api.openai.com`, so after this they stop receiving `response_format={"type": "json_object"}` and rely on the prompt plus fence-stripping. Your reasoning is that those endpoints reject `json_object` — if so, those paths were already broken and this fixes them, which is a stronger claim than the PR currently makes. Could you paste the actual error response from Anthropic's compatible endpoint (and LM Studio, if that's where you hit it)? It's the load-bearing premise, and I'd like it in the PR record rather than in reviewers' heads.

**2. `_strict_schema` marks every property `required`.** That's what strict mode demands, but it's a behavioural change for the model, not just the wire: fields that were optional must now be emitted, so the model may invent a value where it previously omitted one. Our enrichment schemas have optional fields (PII types among them) where "absent" and "empty" aren't the same thing. A test that runs a real enrichment schema through `_strict_schema` and shows what happens to its optional fields would settle it.

**3. A plaintext override silently ships the API key.** `OPENAI_CHAT_BASE_URL=http://…` to anything non-loopback sends the provider key in the clear, with nothing in the logs. This is operator configuration rather than caller input, so it isn't a privilege issue — but a one-line warning at startup when the override is non-`https` and non-loopback would be cheap and kind.

**4. `_is_hosted_openai` is a string comparison,** so `http://api.openai.com/v1`, or a regional/proxy alias that is genuinely hosted OpenAI, silently takes the non-hosted path and a different request shape. Worth either normalising the scheme or saying in the docstring that exact-match is intentional.

Two mechanical items, same as the other PR: please **rebase** (this is currently conflicting), and **drop the `CHANGELOG.md` hunk** — that file is generated by release-please from Conventional Commit subjects (`changelog-path` in `release-please-config.json`), so a hand-written entry fights the generated one.

None of the four are objections to the idea. Pointing an OpenAI-compatible provider at LM Studio or vLLM is exactly what a self-hoster should be able to do, and the incompatibility handling is the part most people would have skipped.

zznate:
Rebased onto `main` at `d82eebd` and dropped the `CHANGELOG.md` hunk. The review points are in a second commit, `a7f321b`, so the delta is readable on its own.

**1. Evidence.** Verbatim error bodies, 2026-08-25/26, Caura 2.30.0 as shipped, `ENTITY_EXTRACTION_PROVIDER=anthropic` with a funded key, `claude-haiku-4-5-20251001`:

- every no-schema caller (dedup judge, enrichment, contradiction):
  `Error code: 400 - {'error': {'code': 'invalid_request_error', 'message': "response_format.type: Input should be 'json_schema'", 'type': 'invalid_request_error', 'param': None}}`
- entity extraction, which passes a schema with `strict: False`:
  `Error code: 400 - {'error': {'code': 'invalid_request_error', 'message': 'response_format.json_schema.strict: Input should be True', 'type': 'invalid_request_error', 'param': None}}`

All 44 chat calls in one write sweep returned one of those two, then `call_with_fallback` walked to OpenAI (401, no key) and to the fake provider, with only `common.llm.retry` warnings in the log. Direct probes of `https://api.anthropic.com/v1/chat/completions` with the same key: `json_schema` with `strict: true` and every object closed works; `strict: true` with a permissive `{"type": "object"}` is rejected (`For 'object' type, 'additionalProperties' must be explicitly set to false`); `{"type": "text"}` is rejected (`Input should be 'json_schema'`); no `response_format` works but the reply comes fenced in ```json.

LM Studio 0.3.x serving `qwen2.5-coder-7b-instruct`, same no-schema call: `'response_format.type' must be 'json_schema' or 'text'`.

So the Anthropic and OpenRouter defaults were not receiving a working `json_object` before: on Anthropic every JSON call failed. I have not tested OpenRouter; it advertises OpenAI compatibility and takes the strict shapes, and it also takes the no-`response_format` form, which is what it now gets without a schema.

**2. Optional fields.** You are right that "required" changed model behaviour. Now: a property the source schema left optional is made nullable on the wire (`anyOf: [<schema>, {"type": "null"}]`), and after parsing `_drop_optional_nulls` removes a `null` for such a key, so callers see the same absent-or-present shape as before. A `null` for a field the source schema itself made nullable (`Mention.cluster_id`, `EnrichmentResult.ts_valid_start`) is kept as sent. Tests run `ExtractedGraph.model_json_schema()` and `EnrichmentResult.model_json_schema()` through both helpers and validate the cleaned reply with the models; hosted OpenAI is untouched and returns replies as sent.

**3. Plaintext override.** One warning at construction when the base URL is `http` to a non-loopback host, naming the provider and the URL: the key travels in the clear. Loopback stays silent.

**4. `_is_hosted_openai`.** Decides by host now, case-insensitive, any scheme, trailing slash or not. The docstring says a proxy or alias on another host is non-hosted on purpose: the shapes are chosen for what the server accepts, and hosted OpenAI takes the strict shapes as well, so a proxy loses nothing.

Local run at `a7f321b`: the three provider test files 44 passed; full suite 5389 passed, 16 failed, the same 16 as on `main`.

zznate:
Rebased onto `main` at `38649e1` (head `002c618`, the same two commits). One conflict, in `complete_json`: #1010's `raise_if_truncated` check and this branch's fence-stripped parse touched the same lines; the resolution keeps both in that order, the truncation check first, then `json.loads(_strip_code_fence(content))`.

Local run at `002c618` from a clean worktree: the three provider test files 44 passed; full suite 5470 passed, 16 failed, and the same 16 fail on `main` at `38649e1` in a clean worktree (`test_integration_search`, `test_p3_2_relation_weights`, `test_a7_classifier_recall`, `pipeline/test_search_pipeline`, `test_ph6_entity_linking_storage`), none in the LLM provider suites. ruff, ruff format, and mypy clean on the touched files.

Eldad-Caura:
Maintainer update pushed at 0d6e95c3. I rebased the two original signed contributor commits onto current main unchanged, then added a separate signed one-line fix replacing the Ruff C420 comprehension with dict.fromkeys. Verification: Ruff check passed for tests/, Ruff format --check passed for tests/ (395 files), the 30 focused compatibility tests passed, and both naming gates passed. Waiting for exact-head CI before retriggering review.

Eldad-Caura:
@claude

Eldad-Caura:
@claude

PR #995, opened 2026-08-25
fix(api): bind /recall to the authenticated agent the way /search does

### LINKED ISSUE #994: [Bug] /recall does not bind the caller identity the way /search does
### Caura version

2.30.0, 2.32.0

### Deployment mode

Self-hosted (Docker Compose)

### Operating system

Linux (WSL2 on Windows 11), Docker Desktop

### Python version

3.12

### PostgreSQL version

pgvector/pgvector:pg16

### LLM provider in use

anthropic

### Describe the bug

`POST /api/v1/recall` shares its request body with `POST /api/v1/search`, but it never received the identity binding that #801 gave `/search` (H-14 in the 2026-06-11 audit, linked from #799). Two consequences for an agent credential, that is, a request with `MEMCLAW_API_KEY` set on the server and `X-Agent-ID` bound:

1. **The filter is not checked against the caller.** `/search` returns 403 when `filter_agent_id` names another agent. `/recall` accepts it, runs its trust<2 fleet forcing against the named agent, and uses that agent as the visibility identity.
2. **An omitted filter means tenant-wide visibility.** `/recall` passes `caller_agent_id=body.filter_agent_id` to `search_memories`, so with no filter it passes `None`, which is the identity a tenant credential gets. The trust<2 fleet forcing also sits inside `if body.filter_agent_id:`, so it does not run either. `/search` binds `eff_agent_id = body.filter_agent_id or auth.agent_id` and forces the fleet for a trust-1 agent whether or not the filter is present.

So a trust-1 agent in one fleet reads another fleet's `scope_team` rows through `/recall`, with content, while `/search` refuses the same request. The MCP twin, `caura_recall`, already passes the authenticated `agent_id` as `caller_agent_id` and is not affected. Source: `core-api/src/core_api/routes/memories.py`, `recall_endpoint` (line 1960 on `71e9981`), compared with `search` (line 1658).

### Steps to reproduce

With `MEMCLAW_API_KEY` set so that `X-Agent-ID` is bound, and `KEY` holding that value. `R` is any unique suffix.

```bash
U=http://localhost:8000/api/v1; R=$(date +%s)
c() { curl -s -H "X-API-Key: $KEY" -H "Content-Type: application/json" -H "X-Agent-ID: $1" "${@:2}"; }

# two fleets, one agent each; both agents stay at the default trust level 1
c owner-$R -X POST $U/fleet -d "{\"tenant_id\":\"default\",\"fleet_id\":\"fleet-a-$R\"}"
c peer-$R  -X POST $U/fleet -d "{\"tenant_id\":\"default\",\"fleet_id\":\"fleet-b-$R\"}"

# the owner writes a team-scoped row in fleet a
c owner-$R -X POST $U/memories -d "{\"tenant_id\":\"default\",\"agent_id\":\"owner-$R\",\"fleet_id\":\"fleet-a-$R\",\"visibility\":\"scope_team\",\"content\":\"The canary-$R rollout plan moves the billing service to region eu-west-3 on Friday.\"}"

# the peer registers in fleet b
c peer-$R -X POST $U/memories -d "{\"tenant_id\":\"default\",\"agent_id\":\"peer-$R\",\"fleet_id\":\"fleet-b-$R\",\"visibility\":\"scope_agent\",\"content\":\"registration note for peer-$R\"}"
sleep 3

# 1. the peer, no filter
c peer-$R -X POST $U/search -d "{\"tenant_id\":\"default\",\"query\":\"canary-$R rollout plan\",\"top_k\":5}" | jq '[.items[].content] | map(contains("canary-'$R'")) | any'
c peer-$R -X POST $U/recall -d "{\"tenant_id\":\"default\",\"query\":\"canary-$R rollout plan\",\"top_k\":5}" | jq '{canary: ([.items[].content] | map(contains("canary-'$R'")) | any), fleets: ([.items[].fleet_id] | unique)}'

# 2. the peer names the owner in filter_agent_id
c peer-$R -X POST $U/search -d "{\"tenant_id\":\"default\",\"query\":\"canary-$R rollout plan\",\"top_k\":5,\"filter_agent_id\":\"owner-$R\"}" -o /dev/null -w "%{http_code}\n"
c peer-$R -X POST $U/recall -d "{\"tenant_id\":\"default\",\"query\":\"canary-$R rollout plan\",\"top_k\":5,\"filter_agent_id\":\"owner-$R\"}" -w "%{http_code}\n" | jq -c '[.items[].content] | map(contains("canary-'$R'")) | any'
```

### Expected behavior

`/recall` behaves as `/search` does on the same body:

1. No filter: the trust-1 peer is forced to its own fleet and does not see the fleet-a row. `false` from both calls.
2. Filter naming the owner: 403 from both calls, `filter_agent_id ... does not match the authenticated agent identity`.

A tenant or user credential (no bound agent) keeps full-tenant recall and may still filter by any agent, as it does on `/search`.

### Actual behavior

Observed 2026-08-26 against 2.30.0 with real embeddings (bge-m3):

1. No filter: `/search` returns `false` (five hits, all from the peer's fleet). `/recall` returns `{"canary": true, "fleets": [five different fleet ids]}`: the fleet-a row with its content, and rows from four other fleets in the tenant.
2. Filter naming the owner: `/search` returns `403`. `/recall` returns `200` and `true`.

An owner-side control read of the same query finds the row on both paths, so the empty `/search` result is scope enforcement, not a missing row.

### Relevant logs / config

```shell
Server: `MEMCLAW_API_KEY` set (compose override), `EMBEDDING_PROVIDER=openai` with `OPENAI_EMBEDDING_BASE_URL` at an LM Studio `bge-m3` endpoint, `ENTITY_EXTRACTION_PROVIDER=anthropic`. The result does not depend on the LLM tier; it reproduces with the fake provider too, as long as embeddings are real enough for the query to hit.
```

### Additional context

- #801 added the binding to `/search` and `/stm/notes` (`routes/memories.py`, `routes/stm.py`) and did not touch `recall_endpoint`. Its comment block in `search` describes exactly the two escalations above.
- A PR follows with the same binding on `/recall` and four tests mirroring the `/search` ones in `tests/test_route_authz_gaps.py`.
- Found while building an API conformance harness for a self-hosted assessment; the harness runs an owner-side control read beside every cross-agent read, which is how the `/search` and `/recall` difference showed up.

### Pre-flight

zznate:
## Summary

Bind `POST /recall` to the authenticated agent the way `POST /search` is bound: an agent credential may only filter to itself, and an omitted `filter_agent_id` recalls as the authenticated agent, not as the tenant. A tenant or user credential is unchanged.

## Related Issue

Closes #994

## Type of Change

- [ ] Documentation update
- [ ] Other:

## How Has This Been Tested?

Four tests in `tests/test_route_authz_gaps.py`, next to the `/search` ones from #801 and using the same `as_auth` fixture:

- a peer in `filter_agent_id` is refused with 403 and the same message `/search` gives;
- filtering to the agent's own id is allowed;
- an omitted filter reaches `search_memories` with `caller_agent_id` equal to the authenticated agent and `filter_agent_id` `None` (the search call is patched at the module seam to capture its arguments);
- a tenant credential with no bound agent may still filter by any agent.

```
core-api/.venv/bin/ruff check core-api/src/core_api/routes/memories.py tests/test_route_authz_gaps.py   # All checks passed
core-api/.venv/bin/ruff format --check core-api/src/core_api/routes/memories.py tests/test_route_authz_gaps.py
core-api/.venv/bin/pytest tests/                                                                        # 5304 passed
```

Verified live on a 2.30.0 image with this change cherry-picked, real `bge-m3` embeddings, and `MEMCLAW_API_KEY` set so `X-Agent-ID` binds. The reproduction from the issue now gives the `/search` answers on `/recall`: a trust-1 peer with no filter is forced to its own fleet and does not see another fleet's `scope_team` row, and a filter naming the owner returns 403. An external conformance check that runs an owner-side control read beside the peer read passes on both paths.

## Checklist

- [x] `mypy` passes
- [x] `pytest` passes locally

## Additional Notes

- The change is the `/search` block moved to `/recall`, with the comment saying why, so the two read paths now read the same. The trust<2 fleet forcing and `enforce_fleet_read` run against the effective identity, and `search_memories` gets it as `caller_agent_id`.
- The MCP twin, `caura_recall`, already passes the authenticated `agent_id` as `caller_agent_id` and is unchanged.
- No documentation change was needed: the public API contract for `/recall` already describes the `/search` semantics; this makes the route match it.
- `CHANGELOG.md` had no `Unreleased` section on `main`; this adds one above 2.32.0. If #993 lands first, the two entries merge under the same heading.

### DIFF
--- a/core-api/src/core_api/routes/memories.py
+++ b/core-api/src/core_api/routes/memories.py
@@ -1989,14 +1989,33 @@ async def recall_endpoint(
     """
     # Read endpoint — readable set widening applies (see /search).
     auth.enforce_readable_tenant(body.tenant_id)
+    # Same identity binding as /search, for the same two reasons. An agent
+    # credential may only filter to itself, and the effective identity is
+    # body-first with a fall back to the authenticated agent. Before this,
+    # /recall ran its trust<2 fleet forcing against whatever
+    # ``filter_agent_id`` the caller asserted, and an agent that omitted the
+    # filter passed ``caller_agent_id=None`` to the search, which is the
+    # tenant-wide visibility a tenant credential gets. Either way a trust-1
+    # agent read another fleet's scope_team rows through /recall while
+    # /search refused the same request. The MCP twin (``caura_recall``)
+    # already binds to the authenticated agent.
+    if auth.agent_id and body.filter_agent_id and body.filter_agent_id != auth.agent_id:
+        raise HTTPException(
+            status_code=403,
+            detail=(
+                f"filter_agent_id '{body.filter_agent_id}' does not match the "
+                f"authenticated agent identity '{auth.agent_id}'."
+            ),
+        )
+    eff_agent_id = body.filter_agent_id or auth.agent_id
     if auth.tenant_id:
-        if body.filter_agent_id:
+        if eff_agent_id:
             fleet_id_hint = body.fleet_ids[0] if body.fleet_ids and len(body.fleet_ids) == 1 else None
-            _agent = await get_or_create_agent(body.tenant_id, body.filter_agent_id, fleet_id_hint)
+            _agent = await get_or_create_agent(body.tenant_id, eff_agent_id, fleet_id_hint)
             if not body.fleet_ids and _agent.get("fleet_id") and _agent.get("trust_level", 0) < 2:
                 body.fleet_ids = [_agent["fleet_id"]]
             if body.fleet_ids and len(body.fleet_ids) == 1:
-                await enforce_fleet_read(body.tenant_id, body.filter_agent_id, body.fleet_ids[0])
+                await enforce_fleet_read(body.tenant_id, eff_agent_id, body.fleet_ids[0])
         # D13 — a recall is a recall, not a search: plans meter them separately
         # and the recalls counter never moved because this site (and the MCP
         # twin) billed "search". Flag-gated; see ``recall_operation``.
@@ -2021,7 +2040,8 @@ async def recall_endpoint(
         query=body.query,
         fleet_ids=body.fleet_ids,
         filter_agent_id=body.filter_agent_id,
-        caller_agent_id=body.filter_agent_id,
+        # Visibility identity is the authenticated agent, as in /search.
+        caller_agent_id=eff_agent_id,
         memory_type_filter=body.memory_type_filter,
         status_filter=body.status_filter,
         top_k=body.top_k,
--- a/tests/test_route_authz_gaps.py
+++ b/tests/test_route_authz_gaps.py
@@ -256,10 +256,10 @@ async def test_stm_promote_rejects_peer_agent(client, as_auth, _stm_enabled):
 async def test_delete_audit_attributes_gateway_agent(client, as_auth, sc):
     """A gateway agent credential deleting WITHOUT the agent_id query param
     must be attributed to its verified identity, not None."""
+    from core_storage_api.services.postgres_service import get_read_session
     from sqlalchemy import select
 
     from common.models.audit import AuditLog
-    from core_storage_api.services.postgres_service import get_read_session
 
     tenant = f"tenant-{_uid()}"
     await _seed_agent(sc, tenant, "deleter-agent", 3)
@@ -541,6 +541,92 @@ async def test_a_tenant_credential_may_still_filter_search_by_any_agent(
     as_auth(tenant)  # no agent_id
     resp = await client.post(
         "/api/v1/search",
+        json={
+            "tenant_id": tenant,
+            "query": "anything",
+            "top_k": 5,
+            "filter_agent_id": "agent-b",
+        },
+    )
+    assert resp.status_code == 200, resp.text
+
+
+# ---------------------------------------------------------------------------
+# POST /recall — the sibling of /search never received the H-14 binding
+# ---------------------------------------------------------------------------
+
+
+async def test_recall_cannot_borrow_a_peer_identity_via_filter_agent_id(
+    client, as_auth
+):
+    """``/recall`` ran the H-14 path unbound: ``filter_agent_id`` was both the
+    visibility identity and the subject of the trust<2 fleet forcing, and an
+    omitted filter passed ``caller_agent_id=None``, the tenant-wide identity.
+    A trust-1 agent read another fleet's scope_team rows through ``/recall``
+    while ``/search`` refused the same request.
+    """
+    tenant = f"tenant-{_uid()}"
+    as_auth(tenant, agent_id="agent-a")
+    resp = await client.post(
+        "/api/v1/recall",
+        json={
+            "tenant_id": tenant,
+            "query": "anything",
+            "top_k": 5,
+            "filter_agent_id": "agent-b",
+        },
+    )
+    assert resp.status_code == 403, resp.text
+    assert "does not match the authenticated agent identity" in resp.text
+
+
+async def test_recall_filtering_to_own_agent_id_is_allowed(client, as_auth):
+    tenant = f"tenant-{_uid()}"
+    as_auth(tenant, agent_id="agent-a")
+    resp = await client.post(
+        "/api/v1/recall",
+        json={
+            "tenant_id": tenant,
+            "query": "anything",
+            "top_k": 5,
+            "filter_agent_id": "agent-a",
+        },
+    )
+    assert resp.status_code == 200, resp.text
+
+
+async def test_recall_binds_an_omitted_filter_to_the_authenticated_agent(
+    client, as_auth, monkeypatch
+):
+    """An agent credential that omits ``filter_agent_id`` must still recall as
+    itself, not as the tenant. The search call is the seam: its
+    ``caller_agent_id`` is what decides which scope_agent rows are visible.
+    """
+    from core_api.services import memory_service
+
+    seen: dict = {}
+
+    async def _fake_search(**kwargs):
+        seen.update(kwargs)
+        return []
+
+    monkeypatch.setattr(memory_service, "search_memories", _fake_search)
+    tenant = f"tenant-{_uid()}"
+    as_auth(tenant, agent_id="agent-a")
+    resp = await client.post(
+        "/api/v1/recall",
+        json={"tenant_id": tenant, "query": "anything", "top_k": 5},
+    )
+    assert resp.status_code == 200, resp.text
+    assert seen["caller_agent_id"] == "agent-a"
+    assert seen["filter_agent_id"] is None
+
+
+async def test_a_tenant_credential_may_still_recall_by_any_agent(client, as_auth):
+    tenant = f"tenant-{_uid()}"
+    as_auth(tenant)  # no agent_id
+    resp = await client.post(
+        "/api/v1/recall",
         json={
             "tenant_id": tenant,
             "query": "anything",


erni-a:
Thanks for this — and for including the tenant-credential case in the tests. That's the one people usually leave out, and it's what makes the change reviewable rather than just plausible.

I've verified the reasoning against `main` rather than taking it on faith: `/search` already carries this binding (its own comments describe it as closing "the search side of the BOLA chain"), and the MCP `caura_recall` path resolves identity from the authenticated agent, so the REST `/recall` handler was the outlier. Your change lines it up with both, and the `eff_agent_id` fallback keeps tenant/user credentials behaving as they do today.

Two process notes:

**1. Security-affecting changes.** For anything with a security dimension, we'd rather you go through [private vulnerability reporting](https://github.com/caura-ai/caura/security/advisories/new), or `security@caura.ai` as a fallback — see [SECURITY.md](SECURITY.md). It lets us prepare the patch and a public advisory together, and you get credited in the advisory. No issue at all this time, and I'm happy to credit you in one for this if you'd like it — just letting you know the door exists.

**2. Before this can land:**

- **Rebase onto `main`.** The PR is conflicting right now. Most relevantly, #984 added a `response: Response` parameter to `recall_endpoint` (slowapi needs it to inject rate-limit headers, and its absence was 500ing the route) in the same region you're editing. A fair amount else has landed since you opened this.
- **Drop the `CHANGELOG.md` edit.** That file is generated by release-please from Conventional Commit subjects — see `changelog-path` in `release-please-config.json` — so a hand-written entry will duplicate or conflict with the generated one. Your commit subject already carries it.

I'm approving the workflows so CI runs, and I'll take it through review once it's rebased and green.

One behavioural note worth capturing in the release notes rather than the code: an agent credential that currently omits `filter_agent_id` and relies on tenant-wide recall will start getting fleet-scoped results after this. That's the fix doing its job, but it will look like a change to anyone who built against the old behaviour.

erni-a:
CI has run now, and the one red check is **not your change** — it's the rebase staleness, so don't go chasing it.

The failing gate is *Broker OpenAPI has no breaking change vs main*. It regenerates the broker spec from the branch's code and diffs it against the copy committed on `main`. The last commit to touch that spec is #1005, merged a few hours ago, which your branch predates — so the diff reports properties on `GET /api/v1/health` (`version`, `platform_init_errors`, `unhealthy_dependencies`) as "removed" when in fact your branch simply doesn't have the code that produces them yet. A rebase clears it.

For what it's worth, I validated your change against current `main` locally rather than waiting on that:

- Applies cleanly onto today's `main` (the "conflicting" flag was computed before the recent merges).
- It correctly preserves #984's `response: Response` parameter on `recall_endpoint`, and the binding lands inside `recall_endpoint` rather than being absorbed into `/search`'s existing `eff_agent_id` block — both things I specifically checked, since that region has moved a lot today.
- Ruff (CI's exact scopes), the legacy-name ratchet and the do-not-touch sentinel all clean; 67 tests green across the authz, recall, rate-limit and MCP suites; full-suite delta against `main` is zero.
- Reverting only `memories.py` and keeping your tests fails exactly `test_recall_cannot_borrow_a_peer_identity_via_filter_agent_id` and `test_recall_binds_an_omitted_filter_to_the_authenticated_agent`, while both no-regression tests still pass. That's the right two failing for the right reason — the tests earn their keep.

So it's just the rebase and dropping the `CHANGELOG.md` hunk. If you'd rather not, say the word — `maintainerCanModify` is on, so we can do the rebase on your branch and keep your authorship rather than let a fix like this sit.

zznate:
Rebased onto `main` at `d82eebd` (head `4bb33b8`) and dropped the `CHANGELOG.md` hunk; the commit subject carries the note. #984's `response: Response` parameter on `recall_endpoint` is kept, and the binding sits inside `recall_endpoint` as before.

Local run against this head: `tests/test_route_authz_gaps.py` 28 passed; full suite 5363 passed, 16 failed, the same 16 you see on `main` (`test_atomic_fact_unembedded_persist`, `test_crystallization_must_not_wedge`, `test_embed_off_hot_path` and friends), none in the authz, recall, rate-limit, or MCP suites.

For the release notes, in one sentence: an agent credential that omits `filter_agent_id` and relied on tenant-wide recall now gets fleet-scoped results, the same as `/search`.

Noted on the security process; the next report of this kind goes through private vulnerability reporting first. Credit in an advisory is welcome, thank you.

zznate:
Rebased onto `main` at `38649e1` (head `9733014`, the same single commit, no conflicts). Local run from a clean worktree: `tests/test_route_authz_gaps.py` 28 passed; full suite 5444 passed, 16 failed, the same 16 that fail on `main` at `38649e1` (`test_integration_search`, `test_p3_2_relation_weights`, `test_a7_classifier_recall`, `pipeline/test_search_pipeline`, `test_ph6_entity_linking_storage`), none in the authz, recall, rate-limit, or MCP suites.

Eldad-Caura:
Thank you for this, and I'm sorry it sat for ten days while we shipped our own fix for the same bug without referencing yours. That's on us, not on you.

**Your diagnosis was correct in full.** #994 is one of the better bug reports this repo has received: exact source line references on a named commit, a working reproduction, the right root cause, and the observation that the MCP twin `caura_recall` was already bound and therefore unaffected. You also identified both consequences separately — the unchecked filter *and* the omitted-filter case passing `caller_agent_id=None` — and it's the second one that turned out to matter most, for reasons below.

**Why I'm closing this rather than merging it.** #1268 landed a shared `_resolve_read_identity` used by both `/search` and `/recall`, which covers everything this PR does plus one vector it doesn't: `/recall` parses `SearchRequest`, so `caller_agent_id` is also a caller-settable identity field. Guarding only `filter_agent_id` leaves the same escalation reachable by a different spelling. That's the sole substantive difference — your handling of the two consequences you reported is equivalent to what shipped, and you matched `/search`'s error message exactly.

The shared-resolver shape mattered for a second reason: two routes parsing the same request model had drifted on what its fields mean, and that drift *is* the bug class. One implementation is what stops it recurring, which is why we went that way rather than adding a parallel check in `/recall`.

**Your test is being merged, in #1276.** Our five tests all named an identity explicitly, so the *default* path — no filter, no caller id, the ordinary recall — went unpinned. That's the case that needs no crafted request, and it's the consequence you flagged second. The behaviour is correct on `main`; nothing was holding it there. Your test now does, with attribution to you and to this PR.

Two things you got right that we got wrong, and I'd rather say so than quietly fix them:

**The patch seam.** You patched `search_memories` on `memory_service`. I first patched it on the route module — the binding `/search` uses — not realising `/recall` re-imports the function *inside* the handler and so resolves it at call time from the service module. My version silently no-op'd: the real search ran and the assertions read whatever that call left behind. It passed for the wrong reason until a removal probe caught it. #1276 carries a comment explaining the seam so nobody repeats it.

**Asserting on arguments rather than results.** Your test captures what reaches `search_memories`. Ours asserted result-set parity, which cannot distinguish "bound to the caller" from "happened to return the same rows" — those coincide whenever the caller has nothing hidden from it. Yours is the stronger assertion for an identity contract.

**And your verification method was better than ours.** You cherry-picked onto a 2.30.0 image and reproduced against real `bge-m3` embeddings end to end. We verified with unit tests against a fake provider.

#994 is closed by #1268. If you hit anything else in the read paths, please do open it — this was a real high-severity finding and you found it from the outside.

PR #996, opened 2026-08-25
fix(api): PATCH with an explicit null on a NOT NULL field is 400, not 500

Eldad-Caura:
## The bug

`PATCH /api/v1/memories/{id}` with `{"content": null}` returns **500**. So do four other fields, and they fail in two different ways — which matters, because a fix that handles one failure mode does not handle the other:

| field | result before this PR |
|---|---|
| `content` | `TypeError: 'NoneType' object is not subscriptable` — `data.content[:200]` while building the audit diff, **before any query runs** |
| `memory_type` | asyncpg `NotNullViolationError` |
| `weight` | asyncpg `NotNullViolationError` |
| `status` | asyncpg `NotNullViolationError` |
| `visibility` | asyncpg `NotNullViolationError` |

Every `MemoryUpdate` field is typed `X | None`, because that is how an optional PATCH field is spelled. An explicit `null` therefore validates and reaches the service indistinguishable from a real value — `min_length`, `ge`/`le` and `pattern` bind only the non-None branch of the union — and the "No fields to update" 400 does not fire, because the field *is* set.

Reproduced at the model level:

```
>>> m = MemoryUpdate.model_validate({"content": None})
>>> m.model_fields_set, m.content
({'content'}, None)
>>> m.content[:200]
TypeError: 'NoneType' object is not subscriptable
```

## This is five fields, not one, and that changes what it is

One nullable-vs-NOT NULL mismatch is an oversight. Five, sitting alongside **nine siblings that legitimately support null-as-clear**, is a schema whose nullability was never systematically checked against the model.

- **Nullable columns — null clears them, and must keep doing so:** `title`, `source_uri`, `predicate`, `object_value`, `subject_entity_id`, `ts_valid_start`, `ts_valid_end`, `expires_at`. The `simple_fields` loop writes those NULLs deliberately.
- **`metadata`** already has its own rule — 400 in merge mode, added when null-as-clear was withdrawn there.
- **The five above** are `nullable=False` on `Memory` and got the same `X | None` treatment as the nine that aren't.

I measured this rather than reading it: every optional field was PATCHed with an explicit `null` and the status recorded. Nine returned 200, `metadata` returned 400, `metadata_mode` and `entity_links` returned 200, and five raised.

## The fix

Reject those five with a **400 naming every offending field**, placed immediately after the "no fields to update" check — so a null `content` no longer costs an embedding call and a dedup query on the way to failing.

400 rather than a schema-level 422 to match the `metadata=null` rule a hundred lines further down: same "this null has no valid meaning" decision, already answered that way in the same function.

All offending fields are named in one response. A caller serialising the whole schema sends every null at once; reporting only the first would walk them through the fix one round-trip at a time.

## Reachability

Not only hand-crafted requests. The plugin's update tool skips `undefined` but **forwards `null` verbatim** (`plugin/src/tool-definitions.ts`), so any MCP caller passing a null field reaches this. `MemoryUpdate` also has a second entry point in `mcp_server.py` besides the HTTP route — both funnel through `update_memory`, so the guard covers both.

## Why the list is a literal

`NON_NULLABLE_UPDATE_FIELDS` is written out rather than derived from `Memory.__table__` at import time: schema field names and column names coincide for these five but not in general (`metadata` maps to `metadata_`), so a derivation would need its own mapping table and would fail silently if that drifted.

`test_guard_list_matches_the_model` is what actually holds the two in sync — it asserts the tuple still equals the set of `MemoryUpdate` fields backed by a NOT NULL column. **Add a NOT NULL column with a matching update field and that test fails until it is listed.**

## Why nothing caught it

mypy reports the `content` one:

```
memory_service.py: Value of type "str | None" is not indexable  [index]
```

but `core-api/pyproject.toml` carries `[[tool.mypy.overrides]] ignore_errors = true` for `core_api.services.*`. It is one of **343** errors that exemption hides. The other four are runtime-only — no type checker would have found them, since `patch[attr_name] = None` is perfectly well-typed.

Surfaced while making `common/` visible to the service mypy runs (#983). **This is not a finding from that change** — it is reported with or without it, and is hidden purely by the services exemption.

## Tests

16 new, in `tests/test_patch_null_non_nullable.py`:

- all five rejections, parametrized, each asserting 400 *and* that the message names the field;
- a multi-field request (`content` + `status` + `weight`) naming all three;
- all eight null-as-clear siblings still returning 200 with the column cleared — the guard must not over-reach;
- a normal non-null PATCH of the same fields still working;
- the drift guard against the model.

**The six behavioural tests were confirmed to fail without the fix**, with exactly the errors in the table above — verified by removing the guard and re-running, not assumed.

## Notes for the reviewer

- The import block in the new test file trips `I001` under ruff's defaults, exactly as its sibling `tests/test_c7_metadata_mode_query_param.py` does. The repo-root `tests/` tree is outside every ruff scope (CI's, and pre-commit's `^(core-api|core-storage-api)/src/`) and carries 654 such findings; I matched the surrounding style rather than formatting one file to a config the repo doesn't apply there.
- 16 tests in `tests/` fail on my machine (search/FTS and relation-weight). They reproduce **identically on unmodified `origin/main`** and are unrelated — local Postgres text-search configuration.
- mypy green on all four services and `common/`; `ruff check` and `ruff format --check` clean on every CI scope, run separately.

### DIFF
--- a/core-api/src/core_api/services/memory_service.py
+++ b/core-api/src/core_api/services/memory_service.py
@@ -143,6 +143,18 @@ class BlankQuery(ValueError):
     """
 
 
+# ``MemoryUpdate`` fields whose backing column on ``Memory`` is NOT NULL, so an
+# explicit ``null`` in a PATCH body has no way to be honoured. Kept as a literal
+# rather than derived from ``Memory.__table__`` at import time: the schema field
+# names and column names coincide for these five but not in general (``metadata``
+# maps to ``metadata_``), so a derivation would need its own mapping table and
+# would fail silently if that drifted. ``test_patch_null_non_nullable`` asserts
+# this tuple still matches the model, which is the check that actually holds it
+# in sync — add a NOT NULL column with a ``MemoryUpdate`` field and that test
+# fails until it is listed here.
+NON_NULLABLE_UPDATE_FIELDS = ("content", "memory_type", "weight", "status", "visibility")
+
+
 def _content_hash(tenant_id: str, fleet_id: str | None, content: str) -> str:
     return hashlib.sha256(f"{tenant_id}:{fleet_id or ''}:{content}".encode()).hexdigest()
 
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@erni-a — green and ready for your approval.

**All checks SUCCESS on `1b3ed43d`**: CI, CodeQL (python + js/ts), dco, Security Scan, image version bake, Legacy-name ratchet, Do-not-touch sentinel. **Claude review: no issues found.**

One note on the run history: the first CI attempt failed in `Set up job`, before any of this code ran — GitHub could not download `astral-sh/setup-uv` (`Name or service not known` on its own internal API, three attempts). Infrastructure, not the change. Re-ran the failed job and it passed.

**What it fixes:** `PATCH /api/v1/memories/{id}` with an explicit `null` on any of five fields returned a 500 — `content` with `TypeError: 'NoneType' object is not subscriptable` before any query, and `memory_type` / `weight` / `status` / `visibility` with an asyncpg `NotNullViolationError`. All five are NOT NULL columns that `MemoryUpdate` exposes as `X | None`, so the null validates and reaches the service as if it were a value.

**Worth a moment on scope, since it changes what the bug is:** nine sibling fields legitimately support null-as-clear and had to keep working, so the fix is a named list of the five, not a blanket "reject nulls". That asymmetry is the finding — `MemoryUpdate`'s nullability was never checked against the model's. `test_guard_list_matches_the_model` now fails if the two drift.

**Tests:** 16 new. The six behavioural ones were confirmed to fail without the guard, with exactly the errors above — verified by removing it and re-running, not assumed.

**Reachable from the plugin**, not only hand-crafted requests: its update tool skips `undefined` but forwards `null` verbatim.

erni-a (approved):

PR #997, opened 2026-08-25
fix(events): bound and narrow the broadcast-subscription release

Eldad-Caura:
Two defects in [#991](https://github.com/caura-ai/caura/pull/991), both mine, both already merged to `main`. Neither was found here — they were found by review on the enterprise port of the same change ([caura-enterprise#1316](https://github.com/caura-ai/caura-enterprise/pull/1316)), which is a useful argument for porting a fix rather than letting two copies drift.

## 1. The suppression was scoped too broadly

`_pull_loop` catches `NotFound`, `PermissionDenied` and `InvalidArgument` in one handler. #991 put the "we released this ourselves, exit quietly" check inside that shared handler — but the justification only holds for `NotFound`.

A genuine `PermissionDenied` (an IAM grant changed) or `InvalidArgument` (we built a malformed request) against a subscription this process happens to have released was **silently swallowed**: loop returns, no ERROR, `_failed_subscriptions` untouched, `is_healthy` still true. That defeats exactly the safety net the rest of the branch exists to provide.

Narrowed to `isinstance(exc, gexc.NotFound)`, with a comment saying why the other two are deliberately excluded.

**Why this survived review the first time**, since it's the more useful half: I argued on #991 that the suppression was "narrow rather than a blanket swallow" and cited `test_unprovisioned_subscription_still_reports_loudly` as evidence. That test varies the **name** (never released) while holding the exception at `NotFound`. It constrains one of the two axes. My claim was true along the axis I tested and false along the one I didn't — and I stated it confidently enough that the review took it.

## 2. The delete RPC was unbounded, which defeated the point of the method

`release_broadcast_subscriptions()` exists **solely** to finish inside Cloud Run's 10s SIGTERM budget. The delete inherited the SDK's generated default — `timeout=60.0`, retry deadline also 60.0, **six times the entire budget for one call** — run sequentially, once per subscription. One slow Pub/Sub call and the process is killed mid-release, leaking precisely what the method was added to reclaim. Under dual-subscribe every broadcast topic has two names, so even a single topic is two deletes.

Now `BROADCAST_RELEASE_TIMEOUT_SECONDS = 2.0` per call, **and the deletes fan out concurrently** rather than sequentially — a bounded call still overruns an unbounded budget if you multiply it by the subscription count.

A timed-out delete costs nothing that matters: `expiration_policy` still reaps the subscription. Bounding only trades a slow failure mode for a fast one, which is the whole point.

## Tests — each confirmed failing without its fix

| Test | Mutation that breaks it |
|---|---|
| `test_released_subscription_still_reports_non_notfound_loudly` (parametrized: `PermissionDenied`, `InvalidArgument`) | Restore the shared-handler check — **both** params fail |
| `test_release_bounds_the_delete_rpc` | Remove the explicit `timeout=` |
| `test_release_survives_one_failing_delete` | Fails **only** when both isolation layers go |

That last row is stated as a limitation rather than dressed up: the isolation is deliberately layered (per-call `try/except` **plus** `return_exceptions=True`), so removing either one alone leaves the test green. It's a characterization test for the behaviour, not a single-line regression guard, and pretending otherwise is how the first gap got through.

**Harness note**, carried from a sibling lane and worth knowing generally: same-byte-length mutations can reuse stale `__pycache__` bytecode and report a **false pass**. Every mutation above was run with the cache purged.

## Gates

61 tests in `tests/test_events/test_pubsub_bus.py` · mypy clean over `common/` (97 files) · ruff check and format clean on the changed source · legacy-name ratchet: no new lines · do-not-touch sentinel: 34/34 · `core-api/uv.lock` untouched (hashed before and after every `uv` command).

### DIFF
--- a/common/events/pubsub.py
+++ b/common/events/pubsub.py
@@ -69,6 +69,20 @@ SOURCE_ENV_ATTRIBUTE = "source_env"
 # Pub/Sub minimum for ``expiration_policy``.
 BROADCAST_SUBSCRIPTION_TTL_SECONDS = 86400
 
+# Per-call ceiling on the delete RPC in ``release_broadcast_subscriptions()``.
+#
+# The SDK's generated default is 60s, with a retry deadline also 60s — six times
+# Cloud Run's entire 10s SIGTERM budget for a single call. Since that method
+# exists precisely to finish inside that budget, inheriting the default defeats
+# it: one slow Pub/Sub call and the process is killed mid-release, leaking the
+# subscriptions the method was added to reclaim.
+#
+# A timed-out delete costs nothing that matters. ``expiration_policy`` still
+# reaps the subscription, so bounding the call only trades a slow failure mode
+# for a fast one — which is the entire point. Prefer failing fast and leaving
+# the TTL to clean up over blocking a shutdown that has nine seconds to live.
+BROADCAST_RELEASE_TIMEOUT_SECONDS = 2.0
+
 
 class PubSubEventBus(EventBus):
     """Pub/Sub-backed bus.
@@ -874,7 +888,7 @@ class PubSubEventBus(EventBus):
                 gexc.NotFound,
                 gexc.PermissionDenied,
                 gexc.InvalidArgument,
-            ):
+            ) as exc:
                 # We deleted this one ourselves. ``release_broadcast_subscriptions()``
                 # runs ahead of the rest of teardown so it fits inside the
                 # SIGTERM budget, which means this loop can still be mid-pull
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

Eldad-Caura:
@erni-a Ready for review — all checks green, Claude review clean.

**Context worth having before you read it:** both defects this fixes are mine, shipped in #991 which is already on `main`. Neither was found by review here — they were caught by review on the enterprise port of the same change ([caura-enterprise#1316](https://github.com/caura-ai/caura-enterprise/pull/1316)), which is a decent argument for porting a fix rather than letting two copies drift.

**What's actually wrong today on `main`:**

1. A real `PermissionDenied` (revoked IAM) against a broadcast subscription the process had released exits its pull loop **silently** — no ERROR, `is_healthy` still true. The suppression I added sat in the shared handler for all three exception types when it only justifies `NotFound`.

2. `release_broadcast_subscriptions()` exists *solely* to finish inside Cloud Run's 10s SIGTERM window, and it inherits the SDK's `timeout=60.0` default, sequentially, per subscription. Six times the whole budget for one call — so on a Pub/Sub blip it fails to do the one thing it was added for.

**The thing I'd most like a second opinion on:** whether `BROADCAST_RELEASE_TIMEOUT_SECONDS = 2.0` is the right number. My reasoning is that a timed-out delete costs nothing that matters — `expiration_policy` still reaps the subscription — so the trade is a slow failure for a fast one. But I picked 2.0 from the reviewer's suggestion rather than from measurement, and I have no data on real `delete_subscription` latency under load. If you think it should be lower or higher, say so.

**One honest caveat on the tests**, since I over-claimed on exactly this in #991: two of the three new tests fail cleanly under a single-line mutation. The third, `test_release_survives_one_failing_delete`, only fails when *both* isolation layers are removed, because the isolation is deliberately layered. It's a characterization test, not a regression guard, and I've said so in the PR body rather than presenting all three as equivalent.

No rush from my side. The leak this sits on top of is already fixed and staging is holding.

erni-a (approved):

PR #999, opened 2026-08-26
fix(core-api): raise the uvicorn worker healthcheck timeout to 30s

Eldad-Caura:
## What

- `common/serve.py`: add a `--timeout-worker-healthcheck` passthrough (default
  = uvicorn's own 5) and forward it to `uvicorn.run()`.
- `core-api/Dockerfile`: pass `--timeout-worker-healthcheck 30` in the `CMD`.
- `tests/test_serve.py`: cover both the default and the explicit value.

## Why

uvicorn's multiprocess supervisor pings each worker every 0.5s and `SIGKILL`s
one that has not answered within `timeout_worker_healthcheck`, then respawns it
(`Multiprocess.keep_subprocess_alive`). The default is **5s**. A `SIGKILL`ed
worker never runs lifespan shutdown, so `core-api`'s ephemeral Pub/Sub broadcast
subscriptions are never released and hold **project-wide** quota for 24h. On
staging that is running at ~2,560 kills/hour.

**The flag was not reachable before this change.** `common/serve.py` builds an
argparse parser with a fixed flag list, so a Dockerfile `CMD` carrying
`--timeout-worker-healthcheck` made argparse `exit 2` and **the container would
never have started**. Adding the passthrough is most of the diff. Worth knowing
for anyone who reads the one-line version of this fix and reaches for the CMD
directly.

The launcher keeps uvicorn's default of 5 rather than taking an opinion, because
it is vendored byte-for-byte into `caura-enterprise` under the `identical`
policy — the per-service value belongs next to that service's `--workers`.

30s stays far inside Cloud Run's 120s request timeout, so a genuinely hung
worker is still recycled long before a request could hang behind it.

## Why there is no version pin here — the brief's premise did not survive checking

This PR was scoped as "pin uvicorn, and/or raise the timeout." I verified the
pin half and it does not hold up. Three corrections, all checkable:

**1. `core-api/requirements.txt` does not exist.** The `uvicorn[standard]>=0.34,<1`
line is in the **root** `requirements.txt`, which is a local-dev install path
(`AGENT-INSTALL.md`) and governs no container image. `core-api/pyproject.toml:13`
is the real declaration, and it reads `>=0.51.0,<1`.

**2. uvicorn already cannot float.** Since #362 (2026-06-14) the image installs
frozen from the committed lock:

```
uv export --frozen --no-dev $EXTRAS --no-emit-project ... -o /tmp/reqs.txt
uv pip install --system --no-cache -r /tmp/reqs.txt
```

`core-api/uv.lock` pins `uvicorn 0.51.0` exactly, and the export carries
per-wheel hashes so the install runs in require-hashes mode. An `==` in
`pyproject.toml` would be belt-on-belt and would only add Dependabot noise.

**3. The healthcheck is not a new feature, so it did not arrive on a rebuild.**
It landed in uvicorn **0.37.0** (2025-09-23, *"Add `--timeout-worker-healthcheck`
option"*). Verified directly in the **0.49.0** sources — the version this repo
was locked to *before* the bump — same `timeout_worker_healthcheck: int = 5`
default, and the same body:

```python
if process.is_alive(timeout=self.config.timeout_worker_healthcheck):
    continue
process.kill()  # process is hung, kill it
```

The lock moved `0.49.0 -> 0.51.0` on 2026-07-21 (#606). The `SIGKILL` behaviour
was **already active before that bump**, so it cannot explain a subscription
baseline that moved on 2026-08-24. Something else changed then, and pinning
would have encoded the wrong cause while leaving the actual 5s sensitivity in
place.

The real exposure was never the version. It was the 5s budget. This raises it.

## Relationship to the urgent fix

This is **defence in depth, not the primary fix.** The staging crash-loop is
caused by `staging-memclaw-core-api` being the only one of its four siblings
deployed without `--no-cpu-throttling`; that is
[caura-enterprise#1321](https://github.com/caura-ai/caura-enterprise/pull/1321),
which is on a ~9-10 hour clock against a shared subscription cap and should land
first and independently.

The evidence that it is the CPU flag and not this timeout: `core-storage-writer`
and `core-storage-reader` run **this same launcher** at `--workers 2` on **this
same locked uvicorn 0.51.0**, with CPU always-allocated, and record **zero**
worker kills in an hour where `core-api` records 2,560. The kills are also
traffic-independent — an instance serving 2 requests in an hour was still killed
75 times, steadily through minutes with no requests at all.

## Checks

- `pytest tests/test_serve.py tests/test_logging.py` — 43 passed.
- **New tests confirmed failing without the fix**: reverting only `common/serve.py`
  gives `common.serve: error: unrecognized arguments: --timeout-worker-healthcheck 30`
  plus a `KeyError` on the default assertion. 2 failed / 4 passed before, 6 passed after.
- `ruff check common/` (discovery-based, from repo root, as CI runs it) — clean.
- `ruff format --check --isolated common/__init__.py common/structlog_config.py common/serve.py`
  — 3 files already formatted. This is the gate that protects the enterprise
  re-vendor; longest line in `common/serve.py` is 84 cols, under the 88 limit.
- `ruff check core-api/src/` — clean.
- `pre-commit run --files ...` — all hooks pass.
- `core-api/Dockerfile` `CMD` re-parsed as JSON: valid 16-element array.
- No `uv.lock` in this diff; all four `uv.lock` hashes byte-identical before and
  after every command run.

## Note for reviewers

`common/serve.py` is an `identical`-policy vendored file. Merging this will make
the enterprise copy drift until `revendor-common.yml` re-syncs it (6-hourly, or
on the `oss-main-pushed` dispatch). That is the designed path and it self-heals;
flagging it so the drift is expected rather than alarming.

### DIFF
--- a/common/serve.py
+++ b/common/serve.py
@@ -21,7 +21,8 @@ workers. The supervisor lines then emit as structured JSON with ``status:info``.
 Usage (see each service's Dockerfile ``CMD``)::
 
     python -m common.serve <app_import> --settings <settings_import> \
-        --port <port> [--host H] [--workers N] [--timeout-keep-alive S]
+        --port <port> [--host H] [--workers N] [--timeout-keep-alive S] \
+        [--timeout-worker-healthcheck S]
 
 ``<app_import>`` is an ASGI import string (``core_api.app:app``); it is passed
 to uvicorn as a string so the workers can re-import it. ``<settings_import>`` is
@@ -70,6 +71,23 @@ def _build_parser() -> argparse.ArgumentParser:
     p.add_argument(
         "--timeout-keep-alive", type=int, default=65, dest="timeout_keep_alive"
     )
+    # uvicorn's multiprocess supervisor pings each worker every 0.5s and
+    # SIGKILLs one that has not answered within this many seconds, then
+    # respawns it. A SIGKILLed worker never runs lifespan shutdown, so
+    # whatever its shutdown hooks release stays leaked. Exposed here because
+    # the value was previously unreachable: this parser rejects unknown
+    # flags, so putting --timeout-worker-healthcheck in a service CMD made
+    # argparse exit 2 and the container never started.
+    #
+    # Defaulted to uvicorn's own 5 rather than an opinion of ours: this file
+    # is vendored byte-for-byte into the enterprise repo, so the per-service
+    # value belongs in that service's CMD, next to its --workers.
+    p.add_argument(
+        "--timeout-worker-healthcheck",
+        type=int,
+        default=5,
+        dest="timeout_worker_healthcheck",
+    )
     return p
 
 
@@ -98,6 +116,7 @@ def main(argv: list[str] | None = None) -> None:
         port=args.port,
         workers=args.workers,
         timeout_keep_alive=args.timeout_keep_alive,
+        timeout_worker_healthcheck=args.timeout_worker_healthcheck,
         log_config=None,
     )
 
--- a/core-api/Dockerfile
+++ b/core-api/Dockerfile
@@ -127,4 +127,16 @@ ENTRYPOINT ["/app/docker-entrypoint.sh"]
 # supervisor/parent configures structlog before starting — otherwise its
 # lifecycle lines ("Started parent process", "Received SIGTERM", ...) bypass
 # the JSON pipeline and land in Datadog as status:error. See common/serve.py.
-CMD ["python", "-m", "common.serve", "core_api.app:app", "--settings", "core_api.config:settings", "--host", "0.0.0.0", "--port", "8000", "--workers", "2", "--timeout-keep-alive", "65"]
+#
+# --timeout-worker-healthcheck 30 (uvicorn's default is 5). With --workers 2
+# the supervisor pings each worker every 0.5s and SIGKILLs one that has not
+# answered in time. SIGKILL skips lifespan shutdown, so this app's ephemeral
+# Pub/Sub broadcast subscriptions are never released and hold project-wide
+# quota for 24h — a worker recycle is not free here, and 5s is a tight budget
+# for a Python worker's pong thread (a GC pause or a stalled scheduler is
+# enough). 30s is still far inside Cloud Run's 120s request timeout, so a
+# genuinely hung worker is still recycled well before a request could hang on
+# it. This is defence in depth, not the primary fix for the staging
+# crash-loop; that is the missing --no-cpu-throttling on the Cloud Run
+# service (caura-enterprise#1321).
+CMD ["python", "-m", "common.serve", "core_api.app:app", "--settings", "core_api.config:settings", "--host", "0.0.0.0", "--port", "8000", "--workers", "2", "--timeout-keep-alive", "65", "--timeout-worker-healthcheck", "30"]
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,12 @@
 fastapi>=0.115,<1
-uvicorn[standard]>=0.34,<1
+# Floor is 0.37, not 0.34: common/serve.py passes
+# ``timeout_worker_healthcheck`` to uvicorn.run(), and that parameter landed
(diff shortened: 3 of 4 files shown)


Eldad-Caura:
Good finding — the immediate claim is wrong, but the mechanism behind it is
right and pointed at a real defect. Both parts addressed.

## The kwarg does exist at the version actually resolved

Run against the installed package, as the prompt asked:

```
$ python -c "import uvicorn, inspect; \
    print(uvicorn.__version__); \
    print('run:', 'timeout_worker_healthcheck' in inspect.signature(uvicorn.run).parameters); \
    print('Config:', 'timeout_worker_healthcheck' in inspect.signature(uvicorn.Config.__init__).parameters)"
0.52.1
run: True
Config: True
```

And at the version the **container** actually gets — `core-api/uv.lock` pins
`uvicorn 0.51.0` exactly, installed via `uv export --frozen` + `uv pip install`
in require-hashes mode — `run()` carries `timeout_worker_healthcheck: int = 5,`
and forwards it into `Config`. So no `TypeError`, and the container starts.

## But the review is right twice over, and I have taken both points

### 1. The tests structurally could not have caught it — fixed

Correct: every test in the module monkeypatches `serve.uvicorn.run` with a
`MagicMock`, which swallows any keyword silently. A renamed or dropped uvicorn
parameter would keep CI green and fail **only** in the deployed container.
Added `test_uvicorn_really_accepts_the_healthcheck_kwarg`, which asserts against
the real `inspect.signature` of both `uvicorn.Config.__init__` and `uvicorn.run`.
No mock involved, so it fails loudly in CI if the parameter ever disappears.

### 2. The pin floor was genuinely lying — fixed

This is the part worth having caught. I established the boundary rather than
inferring it, by reading `Config.__init__` at both tags:

| uvicorn | `timeout_worker_healthcheck` |
|---|---|
| 0.36.0 | **absent** |
| 0.37.0 | present, `int = 5` |

matching the changelog entry *"Add `--timeout-worker-healthcheck` option"*
(0.37.0, 2025-09-23).

So `uvicorn[standard]>=0.34,<1` declared a range this launcher **cannot run
on** — anything in [0.34, 0.37) raises `TypeError` on startup. Raised the floor
to `>=0.37` with a comment recording why.

One correction to the finding's framing, which changes its blast radius but not
its validity: **`requirements.txt` does not govern any container image.** Every
service installs frozen from its own `uv.lock`; that root file is the local-dev
install path (`AGENT-INSTALL.md`). `core-api/pyproject.toml`'s floor is already
`>=0.51.0`, i.e. above the introduction version, so the file that governs the
image was never lying. The fix is still correct — the dev path should not lie
either — it just was not a live production hazard.

## On "this is why it got worse silently on a rebuild"

That part does not survive checking, and it is why this PR still contains no
`==` pin. Since `timeout_worker_healthcheck` predates 0.49.0, and the lock moved
`0.49.0 -> 0.51.0` on 2026-07-21 (#606), the SIGKILL behaviour was **already
active before that bump**. 0.49.0 carries the same default of 5 and the same
`is_alive(timeout=...) -> process.kill()` body. And since #362 (2026-06-14) the
image has installed frozen from a hash-locked `uv.lock`, so the version cannot
float between builds at all. The exposure was the 5s budget, not the version.

## Checks after the force-push

- `pytest tests/test_serve.py tests/test_logging.py` — 44 passed.
- `ruff check common/` — clean; `ruff format --check --isolated` on the three
  vendored files — already formatted (this is the gate that protects the
  enterprise re-vendor).
- `pre-commit run --files ...` — all hooks pass.
- No `uv.lock` in the diff; all four lock hashes byte-identical throughout.
- Rebased onto current `origin/main`.

Priority note for anyone reading: this is the durable fix, not the urgent one.
The staging crash-loop is caused by the missing `--no-cpu-throttling` on the
Cloud Run service — caura-enterprise#1321 — which is racing a shared Pub/Sub
subscription cap and should land first.

Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #1000, opened 2026-08-26
ci: type-gate repo-root scripts/

Eldad-Caura:
## The gap

Repo-root `scripts/` is type-checked by **nothing**, for the same structural reason `common/` was before #948: not a service `src/` tree, no `pyproject.toml` of its own, and no root `[tool.mypy]` — so it appears in none of CI's mypy scopes. The pre-commit hooks are narrower still (`files: ^(core-api|core-storage-api)/src/`) and deliberately run no mypy at all.

It is not an idle tree. **`scripts/forge_dry_run.py` carried the other half of the forge cron outage** — the same `from common.llm import LLMRequest` as core-api's `cron_handler`, a name that has never existed there, with an `except` returning a stub so that CLI had never used a real LLM either. #955 repaired both halves and #983 gated the service one. The `scripts/` half has stayed ungated since.

## The gate

`scripts/mypy.ini` + a `Run mypy (scripts/)` CI step. The `[mypy]` block is the one every service pyproject and `common/mypy.ini` already carry, so code moving between `scripts/` and a `src/` tree is checked on the same terms. Two additions, both measured:

- **`mypy_path` over the four service srcs.** Without it, `scripts/`'s ten `core_api.*` imports resolve to nothing and degrade to `Any` under `ignore_missing_imports` — the exact gap #983 closed for the services.
- **`follow_imports = silent`.** Without it the step also reports errors *inside* the service trees (29 from core-api), which have their own gate and their own reviewed exemptions. A step that fails on paths it does not own teaches people to ignore it. `silent` still checks `scripts/` against the real types.

## Is the gate load-bearing?

Clean as landed:
```
$ mypy --config-file scripts/mypy.ini scripts/
Success: no issues found in 29 source files
```

Re-inject the forge signature into `forge_dry_run.py`:
```
scripts/forge_dry_run.py:120: error: Module "common.llm" has no attribute "LLMRequest"  [attr-defined]
Found 3 errors in 1 file (checked 29 source files)
```

That is the bug class this exists for, caught.

## Exemptions

**Six modules are exempt by name, never by glob**, so a new script is gated the day it lands. That is the ratchet #998 established, and the shape #983's review rejected when a glob was proposed. The 33 remaining findings are what an unchecked tree looks like — untyped-dict indexing and loop-variable rebinding in benchmark/repro scripts — and the gate lands without them on purpose: **landing a gate together with its backlog is how gates get reverted.** Each entry carries its error count and codes; delete one once its file is clean.

`gen_events_manifest.py` is **fixed rather than exempted**, so both scripts that import `common.*` — the import shape the forge bug had — are gated from day one. Its two errors were a real contravariance mistake: `Callable[[object], None]` cannot accept a `Callable[[PipelineStorageAdapter], None]`. Annotation-only; verified the module still imports and `_LIFECYCLE_HELPERS` is unchanged.

## Two silent-failure traps, documented at both sites

Both report **success** rather than failing, which is what makes them worth writing down:

| trap | consequence |
|---|---|
| running from `scripts/` instead of the repo root | `mypy_path` resolves against the **CWD**, not the config file. Measured: `core-api/src` catches a bogus symbol imported from `core_api`; `../core-api/src` does not and mypy reports success. |
| omitting `--config-file` | mypy resolves config from the CWD and does not walk up the way ruff does. Without it this runs at bare defaults: 84 errors in 38 files instead of the 29 it owns. |

Note these are the mirror image of the ruff trap in #696 (`--config` breaks per-file-ignores, so ruff must use discovery). **mypy needs the flag; ruff must not have it.** Both are now stated at their step.

## Deliberately not included

- **Repo-root `tests/`** — 526 mypy errors across 108 files, 654 ruff. Real, and wants its own PR.
- **A ruff gate for `scripts/`** — 110 errors, 12 files needing reformat. Also its own PR. (mypy is the gate that would have caught the forge bug; ruff would not have.)

Measured here so the follow-ups are cheap. `actionlint` is clean apart from one pre-existing SC2086 also present on `origin/main`.

3 files changed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -340,6 +340,36 @@ jobs:
         # mode is a noisier gate, not an absent one. See common/mypy.ini.
         run: uv run mypy --config-file common/mypy.ini common/
 
+      - name: Run mypy (scripts/)
+        # Repo-root scripts/ was type-checked by nothing, for the same structural
+        # reason common/ was: not a service src/ tree, no pyproject.toml of its
+        # own, and no root [tool.mypy], so it appeared in no mypy scope above.
+        #
+        # It is not an idle tree. scripts/forge_dry_run.py carried the other half
+        # of the forge cron outage — the same ``from common.llm import
+        # LLMRequest`` as core-api's cron_handler, a name that has never existed
+        # there, with an ``except`` that returned a stub so the CLI had never used
+        # a real LLM either. #955 repaired both halves; #983 gated the service
+        # one. This is the same gate one layer out, and re-injecting that import
+        # confirms it fires:
+        #
+        #   scripts/forge_dry_run.py:120: error: Module "common.llm" has no
+        #   attribute "LLMRequest"  [attr-defined]
+        #
+        # MUST run from the repo root. ``mypy_path`` in scripts/mypy.ini is
+        # resolved against the CWD, not against the config file, so a step that
+        # ``cd``s into scripts/ first would silently degrade every core_api.*
+        # import to Any under ignore_missing_imports — the exact gap #983 closed
+        # for the services. ``--config-file`` is required for the usual reason
+        # (mypy resolves config from the CWD and does not walk up); without it
+        # this reports 84 errors in 38 files at bare defaults instead of the 29
+        # files it owns.
+        #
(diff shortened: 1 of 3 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #1006, opened 2026-08-26
fix(events): claim a slot for the broadcast id instead of rolling a random one

Eldad-Caura:
Makes the broadcast-subscription leak bounded by **how many workers are alive** instead of **how many workers have ever died** — without depending on shutdown running.

## The leak

`core-api` runs `uvicorn --workers 2` (`core-api/Dockerfile:130` via `common/serve.py:95`). uvicorn's supervisor pings each worker over a pipe and **SIGKILLs** any that misses `timeout_worker_healthcheck` (5s default), then spawns a replacement — `Multiprocess.keep_subprocess_alive`, uvicorn 0.51.0:

```python
if process.is_alive(timeout=self.config.timeout_worker_healthcheck):
    continue
process.kill()          # SIGKILL
process.join()
...
process = Process(self.config, self.sockets)
process.start()
self.processes[idx] = process
```

SIGKILL means no lifespan shutdown, so `release_broadcast_subscriptions()` (#991, #997) never runs. The replacement called `uuid.uuid4().hex[:12]` again and created a **brand-new** subscription, orphaning its predecessor's for the full `BROADCAST_SUBSCRIPTION_TTL_SECONDS` (86400 — the Pub/Sub `expiration_policy` minimum, not tunable).

`_ensure_broadcast_subscription` has had an `AlreadyExists` → reuse branch all along. It was **dead code**: `uuid4()` guarantees the name is never reused, so it could never fire. This makes it the expected path.

## Why this is a lock and not a computation

The identity has to satisfy three properties **at once**:

| | stable across respawn | distinct between concurrent workers | distinct across instances |
|---|---|---|---|
| `uuid4()` (today) | ❌ | ✅ | ✅ |
| pid / `SpawnProcess-N` | ❌ | ✅ | ✅ |
| Cloud Run instance id | ✅ | ❌ **the trap** | ✅ |
| bare slot index | ✅ | ✅ | ❌ |
| **flock slot + instance token** | ✅ | ✅ | ✅ |

**Property 2 is the trap.** Two concurrently-live workers on one subscription does not fail loudly — Pub/Sub load-balances between them and each sees roughly half the events. For `org.settings-changed` that is a process serving stale org settings with nothing red anywhere. That is worse than the leak, so stability must never be bought at its expense.

**There is no worker-slot identity in the environment**, so the fix constructs one:

- uvicorn 0.51.0 passes the child only `(config, sockets)` — `_subprocess.py:51`. The slot index `idx` exists solely in the parent's list and is never exported.
- Workers are `multiprocessing.get_context("spawn")` children, so `os.environ` is byte-identical to the parent's. `SpawnProcess-N` and pid both come from monotonic counters and are **never reused** — they fail stability, which is the property that matters.
- Cloud Run injects `K_SERVICE`/`K_REVISION`/`K_CONFIGURATION`, all per-instance; `CLOUD_RUN_TASK_INDEX` is Jobs-only. Nothing per-worker.

`flock` supplies properties 1 and 2 together, as a kernel guarantee rather than a probability: the kernel releases the lock when the holder dies **however** it dies, so a SIGKILLed worker frees its slot with no cooperation from a shutdown path that never ran — and it cannot hand a slot to a newcomer while the incumbent is alive. The instance token supplies property 3, and is deliberately drawn from the same directory the locks live in so the token and the mutual exclusion describe the **same** boundary.

**The fallback direction is the load-bearing decision**: when no slot can be proven ours, take a `uuid4()` and accept the leak. Never reuse a slot we could not claim. A leak is bounded, visible in quota, and reaped in a day; a collision is a silent halving of broadcast delivery.

## Tests — each confirmed failing without its fix

Real spawned processes, matching uvicorn's own supervisor. They have to be: the property under test is that the *kernel* releases a claim on SIGKILL, and a mock cannot exhibit that.

| Mutation | Test that goes red |
|---|---|
| `return f"{token}-{slot}-{os.getpid()}"` (stability removed) | `test_broadcast_slot_is_stable_across_a_sigkilled_respawn` |
| flock failure ignored instead of `continue` (concurrent distinctness removed) | `test_broadcast_slot_is_distinct_between_concurrently_live_workers` **and** `test_broadcast_slot_falls_back_rather_than_sharing_when_exhausted` |
| `return f"{slot}"` (token dropped) | `test_broadcast_slot_is_distinct_across_instances` |
| fallback returned without memoising | `test_identity_is_stable_per_process_even_on_the_fallback_path` |

Mutation 2 killing two tests is not a coupling artifact — both pin real consequences of "never reuse a held slot". `__pycache__` purged before every run.

## Two defects review caught in my own first cut

Recording these because the pattern from #997 repeated — the comment asserting a property was the thing that was wrong.

1. **A module global holding the flock fd, with a comment claiming it kept the claim open.** It did nothing: `os.open` returns a raw `int`, CPython never closes raw fds on GC, and the assignment was overwritten on every call while all the flocks stayed held. Worse than useless — the next reader either trusts a mechanism that does not exist, or deletes the global and adds the "missing" close, which would hand a live slot to a sibling. Replaced with a comment at the `return` stating the real invariant.
2. **The memo did not memoise the failure path.** On fallback, `_slot_id` stayed `None`, so every subsequent bus in the process re-ran the whole claim and minted its *own* `uuid4()` — losing the per-process property the memo exists to provide, and falsifying `start()`'s comment that a start/stop/start cycle recreates the same names. Production escaped only by the `get_event_bus()` singleton, which nothing in this file states or depends on. Reproduced, then fixed by moving the fallback into `_process_broadcast_slot_id` so there is one contract with one owner.

Also from review: the claim is now resolved lazily in `start()` and only when a broadcast topic is registered, so a publisher-only bus does no filesystem work; and one unopenable lock file degrades per-slot instead of abandoning the slots after it.

## What it does not do

- **It does not eliminate the leak, and the PR title should not be read as claiming that.** It converts "one orphan per worker *death*" into "one orphan per *instance* teardown that skips shutdown". The instance token does not outlive the instance, so instance churn still leaks. Given ~130 worker deaths per instance per 25 minutes against instance lifetimes measured in hours, that is a large reduction — but it is bounded, not zero.
- **It does not reclaim what has already leaked.**
- **It does not make `release_broadcast_subscriptions()` redundant.** The two cover disjoint exits: slot-stable naming handles the death where shutdown *cannot* run; the delete handles the exit where it can and where reuse is impossible anyway (the instance is going away and its token never returns). Deleting also drops a backlog no successor will drain.

## Notes

- **`common/events/pubsub.py` is `manual` in the enterprise vendor manifest** — the re-vendor bot will not sync it. A hand-port is needed, and it is not cosmetic: `platform-auth-api` subscribes `Topics.Fleet.POLICY_EVENTS` with `broadcast=True` and also runs `--workers 2`, so it has the identical leak shape.
- A reused subscription keeps its creation-time config. Relevant to the server-side `source_env` `filter` the module docstring floats: a Pub/Sub `filter` is **immutable**, so adding one means changing this name, not updating in place. Noted in the code.

## Gates

`pytest tests/` 5337 passed / 0 failed (`tests/test_events/` 140) · mypy `--config-file common/mypy.ini common/` clean (97 files) · `ruff check` clean on `common/`, `core-api/src/`, `core-storage-api/src/`, `core-storage-api/tests/` · `ruff format --check` clean at CI's scopes (`common/` not blanket-formatted) · do-not-touch sentinel 34/34 · legacy-name ratchet: no new lines · `core-api/uv.lock` untouched (hashed before and after every `uv` command).

### DIFF
--- a/common/events/pubsub.py
+++ b/common/events/pubsub.py
@@ -14,6 +14,15 @@ Terraform can't pre-provision them, so ``_ensure_broadcast_subscription``
 creates one at ``start()`` (with an ``expiration_policy`` backstop) and
 deletes it at ``stop()``.
 
+That per-process id is a claimed *slot* (see ``_claim_broadcast_slot``),
+not a random value. The distinction is the difference between a leak
+bounded by how many workers are alive and a leak bounded by how many
+workers have ever died: a SIGKILLed worker's replacement re-derives the
+same name and reuses the subscription, where a random name orphans one
+per death for a full ``BROADCAST_SUBSCRIPTION_TTL_SECONDS``. Deletion at
+``stop()`` is still load-bearing and is NOT made redundant by this — the
+two cover different exits. See ``release_broadcast_subscriptions``.
+
 Import is lazy so `common.events` can be imported in OSS standalone
 installs that don't have `google-cloud-pubsub` installed.
 
@@ -38,11 +47,16 @@ from __future__ import annotations
 
 import asyncio
 import concurrent.futures
+import contextlib
 import functools
 import json
 import logging
+import os
+import tempfile
+import threading
 import uuid
 from collections import defaultdict
+from pathlib import Path
 from typing import Any
 
 from pydantic import ValidationError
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@claude — force-pushed with the fix for the Medium finding.

## Taken, as written

`_read()` in `_instance_token` now catches `OSError` rather than only `FileNotFoundError`. You were right that it was the one filesystem call in the module that did not degrade to `None`, and right about the consequence: it runs on the startup path via the unguarded call in `start()`, so a `PermissionError` on the token file crashed boot instead of falling back to a random id — the exact opposite of what the docstring promises.

## The call site at `start()` — I did not guard it there, deliberately

You flagged it as unguarded and stopped short of asking for a fix. I agree it needed one, but not at that line.

The guard now lives **inside `_process_broadcast_slot_id`**, wrapping the `_claim_broadcast_slot` call. That function is the one whose docstring promises "always returns an id", so enforcing it there makes the promise true for every caller rather than for the one call site that remembered to wrap it. Guarding `start():921` would leave the next caller exposed and leave the promise resting on a convention. It is `except Exception`, not `BaseException`, so a cancellation during startup still propagates.

It does not replace the widening. Each expected failure is still caught where it happens and named there; the boundary guard is the backstop for the unforeseen one, on a path whose alternative is a service that will not boot.

## The first version of the test was worthless and I nearly shipped it

My initial test asserted `start()` survives an unreadable token. It passed with the narrow `except` restored — because the new boundary guard caught the escape. A test satisfied by either of two layers pins neither, which is the gap #997 called out in `test_release_survives_one_failing_delete`.

So there are now three, each pinning one layer:

| Test | Mutation that breaks it |
|---|---|
| `test_claim_returns_none_rather_than_raising_on_an_unreadable_token` | restore `except FileNotFoundError` |
| `test_identity_survives_an_unexpected_error_in_the_claim` | remove the guard in `_process_broadcast_slot_id` |
| `test_start_survives_an_unreadable_instance_token` | **neither alone** — labelled in-code as a characterization test for the end-to-end contract, not a regression guard for either layer |

Full matrix, `__pycache__` purged before each run — six mutations, each killing exactly its intended test:

```
BASELINE                                                          9 passed
identity varies per process (stability removed)                   1 failed  test_broadcast_slot_is_stable_across_a_sigkilled_respawn
flock failure ignored (concurrent distinctness removed)           2 failed  ..._distinct_between_concurrently_live_workers, ..._falls_back_rather_than_sharing_when_exhausted
token dropped (cross-instance distinctness removed)               1 failed  ..._is_distinct_across_instances
fallback not memoised                                             1 failed  test_identity_is_stable_per_process_even_on_the_fallback_path
_read() narrowed back to FileNotFoundError                        1 failed  test_claim_returns_none_rather_than_raising_on_an_unreadable_token
totality guard removed                                            1 failed  test_identity_survives_an_unexpected_error_in_the_claim
RESTORED                                                          9 passed
```

## Gates

Rebased onto `origin/main` (`94cd4f1d`) immediately before the push, then re-run: `pytest tests/` **5361 passed / 0 failed** · `tests/test_events/` 143 · mypy clean over `common/` (97 files) · `ruff check` and `ruff format --check` clean at CI's exact scopes · do-not-touch sentinel 34/34 · legacy-name ratchet no new lines · `core-api/uv.lock` untouched (hashed before and after every `uv` command).

Eldad-Caura:
On the Low — **the observation is correct, and I am not changing it. Reasoning below rather than a silent pass.**

`_process_broadcast_slot_id()` does `mkdir`, up to `BROADCAST_MAX_SLOTS` `os.open`/`flock` calls, and a small read directly on the loop at the `start()` call site, while `_ensure_broadcast_subscription`'s `_create()` 37 lines below offloads to `self._pull_executor`. That inconsistency is real and accurately described.

**What makes it not worth the change here is that the stall has no victim.**

- It runs inside lifespan startup, before the app serves anything and before any pull loop exists — the executor it would be handed to has not been given work yet, and nothing else is queued on the loop to be delayed.
- The `flock` is `LOCK_EX | LOCK_NB`. There is no blocking wait to inherit: the worst case is 16 non-blocking attempts, a `mkdir`, and one small read against an instance-local tmpfs. That is bounded and sub-millisecond, not an unbounded I/O wait like the RPC in `_create()`, which is what earns that one its executor.
- Memoised per process, so it happens once.

So the two calls look alike but differ in the property the convention exists to protect against: `_create()` can block on the network for seconds; this cannot block at all.

There is also a small argument for leaving it inline. The claim path is the one thing in this change that must be **totally** non-throwing — an escape here crashes boot instead of degrading to a random id, which is the failure the guard inside `_process_broadcast_slot_id` exists to prevent and which has a dedicated test. Exceptions crossing an executor boundary surface differently, and moving a totality guarantee across that boundary to fix a stall with no victim is a worse trade than the inconsistency.

If a future change makes slot claiming genuinely blocking — a network filesystem for `BROADCAST_SLOT_DIR`, or a blocking `flock` — this becomes correct and should be taken then. Recording that here so the reasoning is available rather than rediscovered.

The two suggestions are both fair and both deliberately out of scope:

- **`tempfile.mkstemp` for the scratch file.** Agreed as defence-in-depth. The precondition it guards (instance-local, single-tenant `/tmp`) holds on every current target, and widening this PR to harden against a deployment model that does not exist would enlarge a change that is fixing a live production leak.
- **Deriving `BROADCAST_MAX_SLOTS` from the worker count.** The right shape is a startup warning when configured workers approach the ceiling, not a derived constant — the ceiling has to cover the transient overlap during `restart_all`, where old and new workers are briefly both live, so it cannot equal the worker count. Worth doing; separate change.

No code change in response to this comment.

Eldad-Caura:
@erni-a pelase approve

erni-a (approved):

PR #1009, opened 2026-08-26
fix(llm): cap complete_json output and surface max-tokens truncation clearly

ran-taig:
## Problem

After the Vertex-only cutover ([caura-enterprise#1300](https://github.com/caura-ai/caura-enterprise/pull/1300)), prod entity extraction showed a **~7% total-failure rate** (9 of ~130 in the first 50 minutes): `gemini-2.5-flash-lite` occasionally emits a runaway ~50k-token JSON that the API truncates mid-string. `complete_json` then fails with a `JSONDecodeError` pointing at a ~200KB char offset — a misleading symptom — and the caller falls back to the keyword heuristic (writes unaffected, entity-graph quality degraded).

Prod evidence (core-api, 2026-08-26 08:12–08:32 UTC):
```
entity-extraction-primary attempt 1/2 failed (JSONDecodeError: Unterminated string starting at: line 8851 column 7 (char 203549))
```

## Fix (all three chat providers — no path keeps the gap)

Applies to `VertexLLMProvider`, `GeminiLLMProvider`, **and `OpenAILLMProvider`** (the OpenAI-compatible path serving OpenAI / Anthropic / OpenRouter — same uncapped-JSON gap, added for parity per review):

1. **Cap the output**: `complete_json` now passes an output-token ceiling — `max_output_tokens=LLM_JSON_MAX_OUTPUT_TOKENS` on the Gemini SDKs, `max_completion_tokens` on the OpenAI-compatible API (the same param `complete_text` already uses there). Default **8192**, env-overridable. Legit outputs are far smaller (enrichment ~150 tokens, entity graphs a few thousand), so a runaway now fails fast and cheap instead of burning ~50k output tokens per attempt.
2. **Name the failure**: a shared `raise_if_truncated()` checks the truncation finish reason (Gemini `candidates[0].finish_reason=MAX_TOKENS` / OpenAI `choices[0].finish_reason="length"`) *before* parsing and raises a clear, retryable `ValueError` (provider, model, cap) instead of a parse error deep inside a partial payload. Duck-typed over all three SDKs' response shapes — any missing attribute means "not truncated", so an SDK shape change can never break the happy path.

Genuine parse failures (`finish_reason=STOP`, malformed body) still raise `JSONDecodeError` unchanged, and `call_with_retry`'s existing retry behavior is preserved — a truncated attempt retries the same as before, just with an accurate log line and a fraction of the token spend.

## Testing

- `tests/test_llm_json_truncation.py` (13 tests): truncation raises with context for all three response shapes (Gemini enum, string reason, OpenAI `"length"`), STOP/missing-shape cases don't, all three providers pass the cap into their request config (verified via captured config/kwargs), happy-path parse, and untruncated-bad-JSON still raises `JSONDecodeError`.
- Adjacent suites green: `test_gemini_provider.py`, `test_platform_providers.py`, `test_provider_protocols.py` (125 passed) plus the OpenAI-kwargs-sensitive suites `test_a5a_seed_and_canonical.py`, `test_a5b_prompt_and_schema.py`, `test_credential_bridge.py`, `test_e3_judge_token_cost.py`, `test_extraction_partial_parse.py`, `test_audit_s6_c1_events.py` (62 passed). Ruff clean.

## Notes

- This addresses the truncation symptom; the deeper follow-ups remain open: honoring `model_override` on the platform-singleton path (so e.g. entity extraction can use a stronger model than enrichment) and migrating off the deprecated `vertexai` SDK to google-genai (unlocks gemini-3.x on v1beta1).

### DIFF
--- a/common/llm/constants.py
+++ b/common/llm/constants.py
@@ -164,6 +164,17 @@ if _jitter_requested != LLM_RETRY_JITTER_FRACTION:
 # behaviour without a deploy.
 LLM_PROVIDER_MAX_RETRIES = read_int_env("LLM_PROVIDER_MAX_RETRIES", 0, minimum=0)
 
+# Output-token ceiling for JSON completions (``complete_json``). Legit
+# outputs are small — enrichment ~150 tokens, entity graphs a few
+# thousand — but prod 2026-08-26 saw gemini-2.5-flash-lite emit runaway
+# ~50k-token JSON that the API truncated mid-string, surfacing as a
+# JSONDecodeError deep inside a 200KB partial payload (~7% of entity
+# extractions fell back to the keyword heuristic). The cap makes a
+# runaway fail fast and cheap; the truncation itself is detected via
+# ``finish_reason`` and raised as a clear, retryable ValueError instead
+# of a parse error (see providers/_truncation.py).
+LLM_JSON_MAX_OUTPUT_TOKENS = read_int_env("LLM_JSON_MAX_OUTPUT_TOKENS", 8192)
+
 # Fallback model for OpenAI-compatible providers when the tenant's
 # configured model is not set — env-overridable so on-call can swap to
 # a cheaper / different family without a redeploy.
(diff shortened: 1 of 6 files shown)


ran-taig:
@claude

ran-taig:
@claude

ran-taig:
@erni-a

erni-a (approved):

PR #1010, opened 2026-08-26
fix(settings): mask api_keys in GET/PUT /settings responses (C36)

arkash20:
## Summary

**C36 (High)**: `GET /settings`' docstring promised "(API keys masked)" while the implementation returned the stored bytes verbatim — any tenant-scoped credential (a low-trust agent's included) could read the tenant's provider keys, `PUT` echoed the same unmasked tree, and the skills-inbox settings helper served the same view. Found during C33's schema extraction, verified in code, filed on the canonical list.

## What changed

All at the display chokepoint (`get_settings_for_display` + both `update_settings` return sites) in `organization_settings.py`:

- Every non-empty `api_keys` value is masked to `****<last4>` — enough to identify which key is configured, useless as a credential.
- **Round-trip safety**: the write path treats a masked value as "unchanged" and drops it before validation/merge. This matters because the dashboard sends the whole `api_keys` group when any single key is edited — unedited siblings arrive masked on every save and must not overwrite the stored keys. An explicit empty string still passes through (that's how a key is cleared). No real provider key starts with `****`, so the sentinel can't collide.
- The raw path (`get_raw_settings` / `ResolvedConfig`) is untouched — enrichment/recall/embedding keep resolving real keys; masking is display-only.

## Testing

- 8 unit tests in `tests/test_c36_api_key_masking.py`: mask shape (incl. short keys that would leak through a naive last-4), empty/absent passthrough, masked-sibling drop, whole-group drop, explicit clear, masked PUT response, raw-path integrity.
- Full root suite: **5389 passed**; ruff/mypy clean.
- **Live wet test** (local enterprise stack, fresh image + tenant): PUT a real key → GET returns `****wxyz`, real key absent from the entire response tree → dashboard-style round-trip PUT (masked value + edited sibling in one body) → postgres shows the stored key **byte-identical** and the sibling's new value written through.

Refs: canonical C36.

### DIFF
--- a/core-api/src/core_api/services/organization_settings.py
+++ b/core-api/src/core_api/services/organization_settings.py
@@ -21,6 +21,7 @@ Cross-worker invalidation is tracked as a follow-up (see CAURA-571).
 
 from __future__ import annotations
 
+import hashlib
 import logging
 from dataclasses import dataclass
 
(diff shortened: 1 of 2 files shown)


arkash20:
@Eldad-Caura please approve

Eldad-Caura (dismissed):

arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #1011, opened 2026-08-26
perf(contradiction): sparse batch-judge verdict schema (E4)

arkash20:
## Why

Closes the loop the August cost incident opened (E4 on the shared task list). The batch judge answered with a full 5-field judgment for **every** candidate — ~35 output tokens × 20 candidates ≈ 700/call — when typically 0-4 candidates contradict. That filler was ~80% of the judge's OpenAI bill, and OpenAI bills tokens, not calls (#770/#771's batching cut request count 5× while the invoice never moved; #940/#946's token logging is how this is now measurable).

## What

- Both batch prompts (Path A semantic + Path C entity-aware) answer **sparsely**: full judgment only for contradicting candidates (`hits`), a bare index list for the rest (`clean`). Every candidate is still read and judged — **the 20-candidate coverage is deliberately untouched** — only the answer got shorter.
- `_expand_sparse_batch` replaces the two `_align` closures. Every non-hit expands to `{"contradicts": False}`, byte-identical to the dense parser's non-hit output, so `_judge_contradiction`'s Gate 1 / Gate 2 and `_pick_older` semantics are unchanged. Abstain-on-outage semantics unchanged (`_batch_fake_fn` / `_skip_contradiction_batch` untouched).
- **Dense back-compat**: the pre-E4 per-index shape still parses, so a fallback provider answering in the old format degrades to old behaviour, never to "everything is clean".
- Drops `relationship`/`diagnosis` from the reply: verified unconsumed — `record_detected_conflicts` receives only `(candidate, path, confidence)` and the A55 engine derives its own classification via its own prompts.

## Measured (real gpt-5.4-nano)

| | old dense | new sparse |
|---|---|---|
| output tokens / 20-candidate call (median of 3) | 651 | **94 (−86%)** |
| verdicts on the same batch | 0/4 tp, 1 fp in one run | 0/4 tp, 0 fp — parity, minus the false positive |

(Both contracts miss the planted update-style contradictions — the known pre-existing judge-quality issue, identical before and after; this PR changes cost, not verdict behaviour.)

Wet-tested e2e on the local enterprise stack: sparse batch calls at 93–98 output tokens, planted on-call-handoff contradiction detected (`verdict=True 0.90`) and superseded. At August volumes this is roughly $500–800/month off the $929 output line; the `service=` token logs verify it in prod.

## Tests

12 new tests: sparse expansion, all-clean, omitted-index safety, int-key/malformed tolerance, dense back-compat, Gate 1/Gate 2 on hits, prompt-contract checks, e2e through the batch function. Full suite: 5,392 passed (one pre-existing order-dependent flake in `test_interview_async_submit`, passes in isolation, unrelated file).

### DIFF
--- a/core-api/src/core_api/services/contradiction_detector.py
+++ b/core-api/src/core_api/services/contradiction_detector.py
@@ -1204,7 +1204,14 @@ keys. EVERY candidate index from 0 to N-1 must appear in exactly one of them:
 - "hits": object mapping each contradicting candidate index (as a STRING) to
   its judgment with the three fields above.
 {{"clean": [1, 2, 4], "hits": {{"0": {{"same_subject": true, "non_conflict_reason": "none", "contradicts": true}}, "3": {{"same_subject": true, "non_conflict_reason": "none", "contradicts": true}}}}}}
-If NO candidate contradicts, "hits" is an empty object and "clean" lists every index.
+Calibration: in most batches FEW OR NO candidates contradict — an empty "hits"
+is the normal answer, not a failure to find something. Apply the three rules to
+every candidate independently, exactly as if you were judging that one pair on
+its own. A candidate goes in "hits" ONLY when all three rules pass decisively;
+uncertainty, partial overlap, related-but-compatible claims, or any applicable
+non_conflict_reason put it in "clean". Never add a hit because a batch "should"
+contain one. If NO candidate contradicts, "hits" is {{}} and "clean" lists every
+index.
 """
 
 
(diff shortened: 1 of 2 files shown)


arkash20:
@Eldad-Caura please approve

Eldad-Caura (dismissed):

arkash20:
**Regression-bench validation (67-question LongMemEval sample, d=0 fast mode, local stack, real gpt-5.4-nano) — three runs tell the story:**

| run | Path A positives | Path C positives | outcome |
|---|---|---|---|
| main control (no E4) | 12 | 27 | baseline for comparison (the May golden is stale — main itself fails it) |
| E4 v1 (first sparse contract) | **55 (4.6x)** | **199 (7.4x)** | ❌ real regression: hit-inflation buried memories, recall@20 −8 to −10pp |
| E4 v2 (calibrated, this push) | 28 | 92 | ✅ recall equal-or-better than control in 4/5 categories (multi-session −2.5pp, under threshold) |

The first sparse framing ('list the contradicting ones') made the model hunt for hits. The calibration block fixes the decision boundary: *an empty hits is the normal answer; judge each candidate as if it were the only pair*. Remaining accuracy deltas were traced per-question to answer/eval-stage variance — flipped questions had **gold_rank=1 with identical rank in both runs** (retrieval untouched).

Token saving after calibration: **699 → 53 median output tokens** per 20-candidate call (−92%).

Two follow-ups worth noting for the record: (1) positives are still ~2-3x the old contract — given the old judge's documented under-firing some may be genuine, but conflicted-rate should be watched after the prod roll (the `service=` token logs + PATH_A verdict logs cover this); (2) the golden baseline predates ~850 PRs and fails against unmodified main — it needs re-baselining once this merges.

arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #1012, opened 2026-08-26
chore: retire old-brand names from internal test and build identifiers

yanki-m:
## Summary

Four commits of internal identifier renames on the rebrand's W5 (safe-rewrite) list. Nothing here is on a user's disk, in a published artifact, or on the wire — temp-directory prefixes, test function names, stale docstrings, and the repo-root private package name.

`legacy_name_ratchet.py --base origin/main`: **No new lines. 30 removed by this change (-30 net).**
`do_not_touch_sentinel.py`: **All 34 protected strings survive.**

## Related Issue

Part of the rebrand programme's W5 bucket (see the 2026-08-25 classification). No single issue.

## Type of Change

- [x] Refactor / internal cleanup

## What each commit does

| commit | lines | scope |
|---|---:|---|
| `test(plugin): brand the temp-dir prefixes in six suites` | 10 | `mkdtempSync` prefixes in `config`, `heartbeat`, `install-id`, `interview-buffer`, `interview-command`, `task-trail` |
| `test: retire the old brand from test names and docstrings` | 15 | 8 files under `tests/` |
| `chore(scripts): brand the wet-test clone dir, fix a stale constant name` | 2 | `scripts/wet-test-install.sh` |
| `chore: rename the root private package` | 3 | `package.json` + both lockfile `name` fields |

## Why each is safe

- **Temp-dir prefixes.** Each of the ten prefix strings was verified to appear in exactly one file — nothing asserts on them. The `.openclaw/plugins/` paths on adjacent lines are untouched.
- **Test names.** Verified no workflow, config or doc selects any of the renamed functions by name. The `memclaw_api_key` settings field the two guard tests assert on is unchanged.
- **Stale docstrings corrected to what the code actually does**: `MEMCLAW_TOOLS` → `CAURA_TOOLS` (that is the real export in `plugin/src/tools.ts`), `MEMCLAW_*` → `CAURA_*` in `test_api_plugin` (its assertions already check `CAURA_*`), `(memclaw_server)` → `(mcp_server)`.
- **One asserted value moves**: `test_api_plugin`'s example host `memclaw.example.com` → `caura.example.com`. The assertion four lines below it moves in the same commit.
- **`wet-test-install.sh`**: `CLONE_DIR` only — the string appears nowhere else in the repo. `PLUGIN_DIR` and the `plugins.allow` / `plugins.entries` keys are the real on-disk install and stay.
- **Root `package.json`**: `private: true`, no `workspaces` block, nothing imports it, never published. `clients/typescript`'s published manifest is untouched.

## How Has This Been Tested?

- The six plugin suites: `npx tsc && node --test dist/{config,heartbeat,install-id,interview-buffer,interview-command,task-trail}.test.js` — **86 pass / 13 fail, byte-identical before and after the change**. The 13 are pre-existing failures on the contributor's Windows machine, unrelated to this PR.
- `ruff check` on the eight touched `tests/` files: **0 findings before and after**. All eight `py_compile` clean.
- `bash -n scripts/wet-test-install.sh` clean.
- `npm ci --dry-run --ignore-scripts` exits 0; both JSON manifests parse.

## Checklist

- [x] No new tests needed — this renames test identifiers and comments; no behaviour changes
- [x] `ruff check` passes on the touched files (unchanged finding count)
- [ ] `mypy` — not run locally; scoped by `scripts/mypy.ini` / per-service configs in CI
- [ ] `pytest` — **not run locally.** The root `conftest.py` requires a live Postgres (`memclaw`/`changeme`) that this machine has no access to, and Docker is unavailable. CI runs the full suite.
- [x] No documentation change needed
- [x] Not user-facing — no CHANGELOG entry

## Additional Notes

Two lines went slightly past the original scope, both inside a file already being edited: `test_plugin_source_manifest.py` also carried the pre-rename repo name and one prose line, corrected here rather than left for a two-line follow-up.

Three items originally scoped into this group turned out to need nothing: `smoke_test.py` and `latency_test.py` are already clean, and `forge_dry_run.py`'s one remaining line names the live `lifecycle` Pub/Sub topic, which moves with its family rather than with a sweep.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,11 +1,11 @@
 {
-  "name": "memclaw",
+  "name": "caura",
   "version": "0.0.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
-      "name": "memclaw",
+      "name": "caura",
       "version": "0.0.0",
       "license": "Apache-2.0",
       "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "memclaw",
+  "name": "caura",
   "private": true,
   "version": "0.0.0",
   "description": "Production-grade persistent memory platform for OpenClaw agents. Agents store and retrieve memories via semantic search, with entity/relationship graphs and tenant isolation.",
--- a/plugin/src/config.test.ts
+++ b/plugin/src/config.test.ts
@@ -267,7 +267,7 @@ function _autoFixWithConfig(initial: Record<string, unknown>): {
   result: ReturnType<typeof autoFixAllowlist>;
   cleanup: () => void;
 } {
-  const tmp = mkdtempSync(join(tmpdir(), "memclaw-autofix-test-"));
+  const tmp = mkdtempSync(join(tmpdir(), "caura-autofix-test-"));
   mkdirSync(join(tmp, ".openclaw"), { recursive: true });
   const cfgPath = join(tmp, ".openclaw", "openclaw.json");
   writeFileSync(cfgPath, JSON.stringify(initial), "utf-8");
@@ -403,7 +403,7 @@ function _ensureExtraDirsWithConfig(
   result: ReturnType<typeof ensureExtraSkillDirs>;
   cleanup: () => void;
 } {
-  const tmp = mkdtempSync(join(tmpdir(), "memclaw-extradirs-test-"));
+  const tmp = mkdtempSync(join(tmpdir(), "caura-extradirs-test-"));
   mkdirSync(join(tmp, ".openclaw"), { recursive: true });
   const cfgPath = join(tmp, ".openclaw", "openclaw.json");
   if (initial !== null) {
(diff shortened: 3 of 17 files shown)


yanki-m:
CodeQL flagged `tests/test_api_plugin.py:30` with `py/incomplete-url-substring-sanitization` (1 high).

**It is not a regression from this PR** — the identical alert is already open on `refs/heads/main` for the same file:line with the old hostname. Renaming the value changed the line, so code scanning re-attributed the pre-existing alert as "new in code changed by this pull request".

Fixed properly rather than suppressed, in 8e71cd9, because the analyzer had a point about the test. It was:

```python
assert "CAURA_API_URL=" in script
assert "caura.example.com" in script
```

Two independent substring checks — they pass even if the variable name and the URL land on unrelated lines, so the test never actually verified that the URL is what got assigned. `shlex.quote` leaves this URL untouched (every character is in its safe set), so the generated script carries the assignment verbatim and it can be pinned in one assertion:

```python
assert "CAURA_API_URL=https://caura.example.com" in script
```

Strictly stronger, and no longer a bare-host substring test.

Worth noting separately: four sibling alerts for the same rule remain open on `main` (`test_identity_token.py:177,178`, `core-worker/tests/test_storage_client_auth.py:193`, `test_platform_providers.py:615`). Same shape, same weakness. Out of scope here, but they will each surface the moment their line is touched.

erni-a (approved):

PR #1018, opened 2026-08-26
fix(auth): reject a tenantless enforce_tenant call instead of passing it

Eldad-Caura:
## The defect

`AuthContext.enforce_tenant` gated writes with a bare equality:

```python
if self.tenant_id != requested_tenant:
    raise HTTPException(403, ...)
```

`None != None` is `False`, so a caller whose context has **no tenant of its own** and who **names no tenant** matched — and was authorized with no write scope at all, rather than refused.

That pairing is reachable, not hypothetical:

- The `CAURA_API_KEY` path (auth Path 2) builds `AuthContext(tenant_id=None)` for a valid key sent without an `x-tenant-id` header — a tenantless, non-admin context. This is already documented in `auth.py`'s own path table.
- Several write bodies carry `tenant_id: str | None` (e.g. `routes/fleet.py`, `routes/interview.py`), so a request that simply omits the field arrives here as `enforce_tenant(None)`.

The sibling `enforce_readable_tenant` was never exposed — it tests `requested_tenant not in self.readable_tenant_ids`, and `None` is not in the list, so it already refuses. Only the write-side gate had the hole.

## The fix

Mirror the resolution #987 established for `_require_tenant`: **naming no tenant is a 400** — a request problem, since the credential is authenticated — and every other unmatched case stays a **403**.

```python
if self.is_admin:
    return
if requested_tenant is None:
    raise HTTPException(400, AUTH_TENANT_REQUIRED)   # new
if self.tenant_id != requested_tenant:
    raise HTTPException(403, AUTH_TENANT_MISMATCH)
```

Once `requested_tenant` is guaranteed non-`None`, the existing `!=` already refuses a tenantless caller that names a real tenant (`None != "acme"` → 403), so the only new branch is the null-target guard. The signature widens to `str | None` to match what callers already pass.

New error code `AUTH_TENANT_REQUIRED` (400), distinct from `AUTH_MISSING_TENANT_CONTEXT` (401, "no key/header") — the credential here is authenticated; it just named no target.

### One behavioural change to call out

A tenant-scoped key that passes `None` as the target now gets **400** instead of the previous 403. That is the same request-vs-credential distinction #987 drew, and it is more correct — but it is a status change, so it is flagged rather than buried. No existing test asserted the old 403 (there was no direct `enforce_tenant` coverage at all before this PR), and the two routes with an optional `tenant_id` that reach this call have no route tests.

## Tests

New `tests/test_auth_enforce_tenant.py`, six cases. Two fail against the pre-fix code — confirmed:

```
test_tenantless_caller_naming_no_tenant_is_rejected   pre-fix: no exception raised (the bypass)
test_tenant_key_naming_no_tenant_is_bad_request        pre-fix: 403, not 400
```

The other four (admin bypass, matching tenant, mismatch → 403, tenantless-naming-a-tenant → 403) pass both before and after, pinning that the fix changes nothing else.

## Gates

`ruff check core-api/src/` clean · `ruff format --check core-api/src/` clean · `mypy src/` clean (201 files — the widened signature adds no call-site errors) · legacy-name ratchet and do-not-touch sentinel both green.


Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #1019, opened 2026-08-27
fix(interview): stop the sweep reporting silent failure as an idle tick

Eldad-Caura:
Two ways the interviewer sweep could report success while having done nothing. Both were found chasing an intermittent failure in `tests/test_interview_async_submit.py`; see the honesty note at the bottom about what this does and does not close.

## 1. A failed job sweep is indistinguishable from an idle one

`run_interview_schedule()` zero-inits the five job counters, then:

```python
try:
    jobs_summary = await process_pending_interview_jobs()
except Exception:
    logger.exception("interview schedule: pending-jobs sweep failed")
```

A sweep that raised therefore returned **exactly** what a sweep with nothing to do returns — `jobs_processed: 0` and four more zeros, over a `200`. The zero-init is deliberate and correct (callers index these keys directly), but it means the failure is only visible in the logs.

The hourly core-operations tick POSTs this endpoint. A sweep raising on every tick reads as a quiet queue for as long as nobody goes looking at logs.

Fixed by keeping the `200` — the scheduling half of the summary above is still valid and still worth returning — and adding `jobs_sweep_ok` plus `jobs_sweep_error` so the two states are tellable apart. The counter keys keep their unconditional zero-init, so the schema is unchanged for existing callers.

## 2. The enabled-tenant list was read from the replica

`list_interviewer_enabled_orgs()` took no `read` parameter at all, while siblings on the same client (`get_document`, `get_report`, `find_by_content_hash`) all do. It resolves to `tenants_list_interviewer_enabled()`, which uses `get_read_session()`.

This is not one stale field among many. It is the call that decides **which tenants are examined at all**, so replica lag does not return a tenant something out of date — it removes that tenant, and every node and job under it, from the tick entirely. And by design the summary then reports success, because from the sweep's point of view there was simply nothing there.

The same function already pins the writer 140 lines further down, for a strictly smaller problem:

```python
# read=False → primary, matching the advance path: a stale
# replica cursor would re-issue a consumed window ...
watermark = await sc.get_document(..., read=False)
```

Fixed by threading `read` through the client and the `services/tenants.py` helper, and passing `read=False` at both sweep call sites. Both additions are keyword-only with `read: bool = True`, so every display-path caller is unchanged.

## Tests

New `tests/test_interview_sweep_freshness.py`, five cases. Mutation-checked — each mutation killed by exactly its intended test, `__pycache__` purged before each:

```
BASELINE                                  5 passed
restore the exception swallow             1 failed  ..._failed_job_sweep_is_distinguishable...
sweeps back to the replica                2 failed  ..._both_sweeps_read_..._from_the_writer (both params)
helper silently drops read=               1 failed  ..._service_helper_forwards_read...
RESTORED                                  5 passed
```

The `read=` test is parameterised over **both** sweeps, because only one of the two call sites is on the path the flaky test exercises — fixing one and not the other is exactly the shape of the bug being fixed.

## Gates

`ruff check core-api/src/` clean · `ruff format --check core-api/src/` clean (201 files) · `mypy src/` clean (201 files) · ratchet and sentinel green · `tests/test_interview_async_submit.py` + `tests/test_api_interview.py` + the new file: **48 passed**.

## What this does not close

**The flake itself is not proven fixed, and I would rather say so than imply otherwise.**

It reproduced twice in three early runs, then survived eight further runs under instrumentation without recurring, so I never captured the failing path. What is established: the assertion is `assert 0 >= 1` on `jobs_processed`, which increments once per result unconditionally — so the collected job list was empty, or the sweep raised and was swallowed by the code fixed in §1.

Also worth recording, because it was the prevailing theory and it is wrong: **replica lag cannot be the local cause.** `tests/conftest.py` points `_session_factory` and `_read_session_factory` at the same engine — "reader and writer share the same engine in tests (no replica spun up)". Fix §2 is a real production bug, not the flake's trigger.

What §1 does buy: the next occurrence carries a named error instead of `assert 0 >= 1`, which is the difference between a diagnosable failure and another archaeology session.

### DIFF
--- a/core-api/src/core_api/clients/storage_client.py
+++ b/core-api/src/core_api/clients/storage_client.py
@@ -2651,9 +2651,19 @@ class CoreStorageClient:
             raise RuntimeError("core-storage-api /tenants/agent-digest-enabled returned 404")
         return result.get("org_ids", [])
 
-    async def list_interviewer_enabled_orgs(self) -> list[str]:
-        """Orgs whose ``interviewer.enabled`` setting is True."""
-        result = await self._get("/tenants/interviewer-enabled")
+    async def list_interviewer_enabled_orgs(self, *, read: bool = True) -> list[str]:
+        """Orgs whose ``interviewer.enabled`` setting is True.
+
+        ``read`` defaults to True — the replica is right for anything merely
+        displaying the set. The interviewer sweep passes ``read=False``: it is
+        the entry point that decides which tenants get looked at *at all*, so
+        replica lag there does not delay one record, it drops a whole tenant
+        from the tick and everything under it. A tenant that enables the
+        interviewer and is swept before the row replicates is skipped
+        entirely, with a summary that reports success. Same reason the sweep
+        already pins the writer for its watermark read.
+        """
+        result = await self._get("/tenants/interviewer-enabled", read=read)
         if result is None:
             raise RuntimeError("core-storage-api /tenants/interviewer-enabled returned 404")
         return result.get("org_ids", [])
--- a/core-api/src/core_api/services/interview_service.py
+++ b/core-api/src/core_api/services/interview_service.py
@@ -1096,7 +1096,10 @@ async def process_pending_interview_jobs(limit_per_tenant: int = 20) -> dict:
     """
     sc = get_storage_client()
     now = datetime.now(UTC)
-    tenants = await list_tenants_with_interviewer_enabled()
+    # read=False → primary, same reason as the watermark read below: this call
+    # selects which tenants are swept at all, so replica lag drops a tenant
+    # from the tick entirely rather than merely returning it a stale field.
+    tenants = await list_tenants_with_interviewer_enabled(read=False)
     summary = {
         "tenants": len(tenants),
         "jobs_processed": 0,
@@ -1234,7 +1237,10 @@ async def run_interview_schedule() -> dict:
     """
     sc = get_storage_client()
     now = datetime.now(UTC)
-    tenants = await list_tenants_with_interviewer_enabled()
+    # read=False → primary, same reason as the watermark read below: this call
+    # selects which tenants are swept at all, so replica lag drops a tenant
+    # from the tick entirely rather than merely returning it a stale field.
+    tenants = await list_tenants_with_interviewer_enabled(read=False)
     summary = {
         "tenants": len(tenants),
         "nodes_considered": 0,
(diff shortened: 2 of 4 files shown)


Eldad-Caura:
@claude — taken, force-pushed.

You are right, and the precedent you cite is the right one. `jobs_sweep_error` lands in an HTTP response body, and an exception's own `__str__` is not under our control — a storage or httpx error can carry a host, a URL or a request fragment. `_storage_detail` refuses to forward a raw error body to a caller for exactly that reason, and this was inconsistent with it.

Now `type(exc).__name__` only. The full text was never lost: `logger.exception` two lines above already records it, which is the right place for it. The type name is also the part a dashboard would branch on, so nothing useful to the caller is given up.

The regression test now pins the leak rather than just the happy path — it raises `RuntimeError("db://user:pw@internal-host/secret path=/etc/x")` and asserts both that the field equals `"RuntimeError"` and that `internal-host` and `/etc/x` are absent from it. Mutation-checked: restoring `f"{type(exc).__name__}: {exc}"` fails that test and only that test.

`ruff check` + `format --check` clean · `mypy src/` clean (201 files) · 5/5.

The endpoint is `enforce_admin()`-gated, so this was defence in depth rather than an open leak — but an admin response is still the wrong place to put an arbitrary exception string, and the log already had it.

Eldad-Caura:
CI failure fixed — force-pushed. It was mine, and the gate was right.

`tests/test_fixture_hygiene.py::test_no_helper_takes_the_db_fixture_without_using_it` flagged my spy:

```python
async def _spy(db=None, *, read: bool = True):   # db never touched
```

I had mirrored the real signature of `list_tenants_with_interviewer_enabled(db=None, *, read=True)` out of habit. But the spy never uses `db`, and that gate exists precisely to stop a dead `db` parameter propagating — as its docstring puts it, otherwise "every caller must keep requesting a session that goes nowhere". Copying a signature is not a reason to take a parameter you ignore.

Now `async def _spy(*_a, read: bool = True)`, with a comment saying why it is not named. The real helper still accepts `db` positionally, so nothing about the seam under test changes.

Re-verified after the change:

- `tests/test_fixture_hygiene.py` + the new file: **9 passed** (the gate that failed CI is in that run).
- Mutations still bite — dropping `read=False` from the sweeps still fails 2 tests, restoring gives 5 passed. The signature change did not blunt the spy.
- `ruff check` clean over `core-api/src/` and the test file.

Worth noting the failure was invisible locally: I had run the two interview test files and the new one, not the repo-wide hygiene suite, so the first signal was CI. That is on my verification scope, not on the gate.

arkash20 (approved):

PR #1020, opened 2026-08-27
feat(entities): write the extraction-derived subject back to the memory row (A63)

arkash20:
## Why

First PR of **A63** (judge misses update-style contradictions). `memories.subject_entity_id` is NULL on nearly every row: the write-time triple path (CAURA-123's `EmitMemoryTriple`) only fires on narrow predicate regexes, and the async extraction worker creates entity links without ever writing the subject back. That keeps two shipped mechanisms dormant — the deterministic RDF contradiction path and the A1 #17 cross-subject preflight (which requires the column non-NULL on both sides and so has never fired in practice) — and blocks A63's next step, subject-scoped candidate selection for the contradiction judge.

## What

- **Worker** (`entity_extraction_worker.py`): after link creation — and before the Path C dispatch at the function tail, so the first detection run already sees it — write back the subject under strict conservatism: **exactly one** distinct subject-role entity required (0 or 2+ → `skipped_ambiguous`; a wrong subject is worse than none because A1 #17 treats the column as authoritative); two surface forms resolving to one entity count as one; failures are non-fatal and logged.
- **Storage** (`POST /memories/{id}/subject-entity`): one conditional UPDATE guarded by `subject_entity_id IS NULL` — the triple path's higher-fidelity value is never clobbered, and concurrent deliveries are race-safe without read-modify-write. Column + index already exist (migration-free).
- Observability: `subject_writeback` log line with `outcome=set | kept_existing | skipped_ambiguous`.

## Tested

- 6 worker unit tests (single-subject set, multi-subject skip, no-subject skip, same-entity surface forms, benign already-set, non-fatal failure).
- 3 storage integration tests vs real PG (sets when NULL + visible on read, never clobbers, foreign-tenant no-op).
- Wet test on the local stack: `outcome=set` on real writes, populated `subject_entity_id` on read-back, `skipped_ambiguous n_subjects=4` on a multi-subject row.
- ruff / mypy (core-api + core-storage-api) / full root suite green.

## Next in A63

Predicate canonicalization → subject+predicate candidate selection alongside cosine (the semantically-distant-update miss class, STALE T2) → re-cut the `temporal_supersession` gate with planted-update bench cases. Note: the regression golden was re-baselined today against current main (post-#1011), so those follow-ups get a meaningful gate.

### DIFF
--- a/core-api/src/core_api/clients/storage_client.py
+++ b/core-api/src/core_api/clients/storage_client.py
@@ -588,6 +588,21 @@ class CoreStorageClient:
     async def soft_delete_memory(self, memory_id: str) -> bool:
         return await self._delete(f"/memories/{memory_id}")
 
+    async def set_subject_entity_if_null(
+        self, memory_id: str, tenant_id: str, subject_entity_id: str
+    ) -> bool:
+        """A63 — conditional subject write-back from the extraction worker.
+
+        Storage-side single UPDATE guarded by ``subject_entity_id IS
+        NULL`` (the write-time triple path's value wins). Returns whether
+        the row was actually updated; ``False`` is a benign skip."""
+        result = await self._post(
+            f"/memories/{memory_id}/subject-entity",
+            {"tenant_id": tenant_id, "subject_entity_id": subject_entity_id},
+            read=False,
+        )
+        return bool(result and result.get("updated"))
+
     async def update_memory_status(
         self,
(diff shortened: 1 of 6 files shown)


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #1022, opened 2026-08-27
feat(recall): premise guard — challenge assumptions the memories refute (A64)

arkash20:
## Summary

**A64.** The recall answer LLM complies with questions whose premises its own memories refute — STALE dim2 (false premise) scored **0/15 in every retrieval configuration including the oracle**: the model plans a Portland itinerary while holding the evidence the user left Portland. Retrieval cannot fix this; the answer prompt can.

New org setting **`recall.premise_guard`** (default **off**) injects one benchmark-tuned instruction block into `RECALL_PROMPT`: check whether the question rests on an assumption the memories contradict or no longer support; if so, flag it and answer for the actual current state — **but if the memories do answer the question, answer it** (no abstaining just because a memory is older).

## Evidence (all in `benchmark/a57-recall-experiments-findings.md`)

| run | result |
|---|---|
| STALE T2 (n=15), guard on | overall **31% → 58%** (dim2 0→60%) |
| True-premise control (67-q pair, off vs on) | overall **76% → 76%** — no cost |
| knowledge-update under v2 wording | **93%, fully recovered** (v1 wording had dropped it to 79% via over-abstention — that's why the second clause exists) |

The wording is a measured dial (v1 reached STALE 71% but converted answerable questions into "not enough information"); the code comment requires a fresh control pair before changing it.

## Zero-impact by construction

- Off (the default): the block slot renders empty — `RECALL_PROMPT` is **byte-identical** to pre-A64 (unit-asserted).
- On: opt-in per org via `PUT /settings {"recall": {"premise_guard": true}}`.

## Testing

- 4 unit tests (`tests/test_a64_premise_guard.py`): knob default + coercion, byte-identical off path, guard + anti-abstention clause present when on, clean block concatenation. Full root suite **5435 passed**; core-api suite green; mypy/ruff clean.
- **Wet-verified** on the local enterprise stack (fresh image): `diagnostic.recall_prompt` (D12) shows a clean prompt with the flag off, the settings round-trip flips it live, the guard text appears with the flag on, and the answer hedges instead of complying on a seeded false-premise query.

Refs: canonical A64 (filed from the A57 experiment program).

### DIFF
--- a/core-api/src/core_api/services/organization_settings.py
+++ b/core-api/src/core_api/services/organization_settings.py
@@ -57,6 +57,13 @@ DEFAULT_SETTINGS: dict = {
         "provider": None,
         "model": None,
         "enabled": None,
+        # A64 — premise guard: instructs the recall answer LLM to challenge a
+        # question whose assumption the memories contradict or supersede,
+        # instead of going along with it. Off (None/False) keeps the recall
+        # prompt byte-identical to pre-A64. Evidence: STALE T2 31%->71%
+        # overall with the guard; true-premise control +1.9pp overall
+        # (benchmark/a57-recall-experiments-findings.md).
+        "premise_guard": None,
     },
     "embedding": {
         "provider": None,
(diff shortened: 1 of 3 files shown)


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #1023, opened 2026-08-27
fix(contradiction): updates ARE contradictions — re-cut the temporal gate, let history questions see superseded values (A63)

arkash20:
## Why

Parts 2+3 of **A63** (judge misses update-style contradictions), shipped together because the regression bench proved they only work as a pair.

**The gate bug:** `temporal_supersession` was defined as *'sequential lifecycle states, both true in order — the newer supersedes the older'*. That describes exactly what an **update** is — and in MemClaw, *newer supersedes older* IS the contradiction the product exists to flag (flagging marks the stale row and writes the supersedes chain). Measured: 0/4 planted region migrations flagged, 3/3 runs, both prompt contracts.

## What

1. **Prompts (all four judge prompts):** `temporal_supersession` redefined as *records of past events that stay true as history* ('hired in 2020' / 'promoted in 2023'); explicitly NOT for current-state changes ('lives in X' vs 'lives in Y') — a state change is an update, reason `none`, `contradicts=true`. Also removed the pairwise prompt's 'if in doubt choose temporal_supersession' nudge.
2. **Gate 2 veto:** `temporal_supersession` removed from `NON_CONFLICT_REASONS`. The A/B showed the model answering `contradicts=true` **with** that reason — semantically coherent — and the code veto overriding every one to false. The label stays in the vocabulary and the A55 mappers; it no longer outranks the verdict. Event-record guard held: 0 false positives on planted history batches.
3. **`history_query` retrieval flag (the necessary pair):** with the gate fixed, the judge correctly buries old values — and the bench immediately showed history questions losing them (KU recall −12.5pp): scored search hard-EXCLUDES `outdated` rows and halves `conflicted` ones. New `_extract_history_hint()` detects past-state/change/duration questions ('what was…', 'did I switch…', 'when I just started…', 'how long have I been…', 'how many hours have I spent…'); the pipeline forwards it; `memory_scored_search` skips both protections for exactly those queries. Present-state queries keep both.

## Bench (67-question LongMemEval sample, d=0 fast, vs the 2026-08-27 golden)

| category | golden | this branch |
|---|---|---|
| overall accuracy | 78.7% | **84.4% (+5.7pp)** |
| multi-session | 72.7 / .863 | **90.9 (+18.2) / .950 (+8.7)** |
| temporal-reasoning recall | .864 | **.969 (+10.5pp)** |
| knowledge-update | 90.9 / 1.000 | **100 acc** / recall healed to 1.000 |
| single-session-preference | 83.3 / .972 | 91.7 / .972 |

No thresholds breached on final numbers. Iteration history (first run FAILed on KU recall −12.5pp → per-question diagnosis → retrieval flag → cumulative-duration pattern) lives in the run dirs. Integration-tested vs real PG: an outdated perfect-match row loses to fillers on a normal query, ranks first on a history query.

## Tests

19 history-hint tests (all regressed bench phrasings trigger; 7 present-state phrasings stay quiet), gate-semantics tests updated (`temporal_supersession` + `contradicts=true` now survives, pinned), PG integration test for the demotion/exclusion lift. Full suite 5,453 green; ruff/mypy both services clean.

**Post-merge:** the golden should be re-captured (Arkady's call) — its recall metric credits superseded old values on two present-state questions this change correctly buries.

### DIFF
--- a/core-api/src/core_api/pipeline/steps/search/execute_scored_search.py
+++ b/core-api/src/core_api/pipeline/steps/search/execute_scored_search.py
@@ -137,6 +137,12 @@ class ExecuteScoredSearch:
         if data.get("valid_at"):
             search_data["valid_at"] = str(data["valid_at"])
 
+        # A63 — a history question must see superseded values: tell the
+        # storage side to skip the outdated/conflicted status demotion.
+        if data.get("history_hint"):
+            search_data["history_query"] = True
+            logger.info("execute_scored_search: history query — status demotion lifted")
+
         date_range = data.get("date_range_filter")
         if date_range:
             search_data["date_range_start"] = date_range["start_date"]
--- a/core-api/src/core_api/pipeline/steps/search/extract_temporal_hint.py
+++ b/core-api/src/core_api/pipeline/steps/search/extract_temporal_hint.py
@@ -1,8 +1,10 @@
 """ExtractTemporalHint — auto-detect time scope from query.
 
-Sets two context keys:
+Sets three context keys:
 - ``temporal_window``: soft freshness boost (timedelta or None)
 - ``date_range_filter``: hard WHERE filter (dict or None)
+- ``history_hint``: bool — the query asks about a past state / change /
+  duration, so the scored search must NOT demote superseded rows (A63)
 """
 
 from __future__ import annotations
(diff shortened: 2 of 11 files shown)


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #1024, opened 2026-08-27
feat(contradiction): Path D — basis invalidation in shadow mode (A58)

arkash20:
## Summary

**A58** — the first slice of the Type-II (propagated staleness) program. "Tore ACL, no weight-bearing for six weeks" retires "commutes by bicycle daily" — but the two statements are *simultaneously true*, so Path A's contradiction judge is **right** to say no, and cannot be tuned into catching this without a false-positive explosion. Type-II needs a differently-shaped question ("did the new evidence break the practical basis of the old default?") over a differently-selected candidate set. That is CUP-Mem's bucket-bridge mechanism (STALE, arXiv:2605.06527), ported here onto our existing `(subject_entity_id, predicate)` slot model.

**Shadow mode only — this PR writes nothing.** Verdicts are logged (`path_d_shadow` lines) so the Type-II base rate and precision can be measured on real corpora before A59 builds the enforcing path (`unsafe` status, migration 037).

## Design (per the analysis doc on the A58 canonical row)

1. **Bridge** (1 LLM call): which *other* predicates of this subject may have had their practical basis broken — hard-gated to a `SINGLE_VALUE_PREDICATES` menu with a confidence floor; the paper's precision guardrails ("shared life domain / topic overlap is not enough — name the broken basis") ported verbatim.
2. **Scoped fetch** (0 LLM): `rdf-conflicts` rows per bridged predicate. When that route is empty — predicates are sparsely populated until A63's canonicalization lands — it falls back to Path C's **already-fetched entity-overlap candidates** (`route=overlap`, zero extra storage calls).
3. **Judge** (≤2 LLM calls, CUP-Mem's `invalidation_merge_max_keep`): `INDIRECT_INVALIDATE | WEAK_CHALLENGE | SET_UNKNOWN_CURRENT | NO_OP`, enum verbatim for paper comparability.

**Hook placement matters**: it rides Path C's conclusion, *not* Path A — wet testing showed `subject_entity_id` resolves during entity extraction (A63 #1020), so it is always NULL at Path A time. Runs under Path C's idempotency lock; double-wrapped so no failure can reach Path C's outcome. Flag `basis_invalidation_shadow`, default off (mirrors `contradiction_write_conflict_record`).

## Testing

- 8 unit tests: no-subject gating (no LLM spend), menu hard-gate + confidence floor, judge cap + **storage purity** (asserts `find_rdf_conflicts` is the only storage call), retired-candidate skip, verdict normalization, error isolation, overlap fallback, hook wiring.
- Full root suite **5442 passed** (includes the 887 contradiction tests + the golden differential); mypy clean.
- **Wet test reproduced the paper's canonical bike/ACL case end-to-end** on the local stack: strong-mode writes, subject resolved by extraction, bridge fired, overlap route selected the bicycle memory, judge returned `WEAK_CHALLENGE` (0.78) — *"no weight-bearing for six weeks makes the prior assumption that Dana commutes by bicycle unsafe during recovery"* — and both rows' statuses stayed byte-untouched. +2.7s background, zero user-visible latency.

## Next (not in this PR)

The A58 spike: replay the 15 labeled STALE T2 pairs through the write path with the flag on and measure hit rate / false-positive rate — that number decides A59's build scope (`unsafe` status + write-side state materialization).

Refs: canonical A58; A59; analysis `cupmem_stale_contradiction_analysis.md`.

### DIFF
--- a/core-api/src/core_api/config.py
+++ b/core-api/src/core_api/config.py
@@ -334,6 +334,15 @@ class Settings(BaseSettings):
     # Additive: it never changes the status/supersedes effect, so retrieval is
     # unaffected. Default False; enable to start populating conflict records.
     contradiction_write_conflict_record: bool = False
+    # A58 — Path D (basis invalidation) in SHADOW mode: after Path A, one
+    # bridge LLM call proposes <=3 other predicates of the same subject whose
+    # practical basis the new memory may have broken; <=2 invalidation-judge
+    # calls produce verdicts that are LOGGED ONLY (``path_d_shadow`` lines).
+    # No status is ever written under this flag — it exists to measure the
+    # Type-II base rate + precision on real corpora before A59 builds the
+    # enforcing path (``unsafe`` status). Skipped when the new memory has no
+    # resolved subject_entity_id.
+    basis_invalidation_shadow: bool = False
     crystallizer_enabled: bool = True
     crystallizer_stale_days: int = 180
     crystallizer_dedup_sample_size: int = 1000
--- a/core-api/src/core_api/services/contradiction/basis_invalidation.py
+++ b/core-api/src/core_api/services/contradiction/basis_invalidation.py
@@ -1,0 +1,308 @@
+"""Path D — basis invalidation, SHADOW MODE (A58).
+
+The Type-II gap: a new memory can destroy the practical basis of an OLD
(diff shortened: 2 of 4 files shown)


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #1025, opened 2026-08-27
fix(contradiction): lower the candidate floor to 0.45 so paraphrased updates reach the judge (A63)

arkash20:
## Why

Part 4 of **A63**. Measured through the real local pipeline: four genuine updates, each deliberately worded differently from the fact it supersedes, scored **0.483 / 0.503 / 0.568 / 0.607** cosine against those facts. Every one sits below the 0.70 candidate floor, so Path A returned **zero** candidates for all four (`candidates_initial=0`) and the judge never saw the contradiction it exists to catch. That is the STALE Type-II miss class.

**After: the same probe takes reachability from 1/4 to 4/4.**

## Why it is cost-neutral

The candidate query is `WHERE (1 - distance) >= floor ORDER BY distance LIMIT <window>` — the **window** bounds how many rows the judge sees, not the floor. Lowering it cannot add a 21st candidate; it changes which rows fill the window, and takes the sparse case from 0 to a few. Post-#1011 a clean candidate costs ~1.5 output tokens.

## Also: a no-op trap removed

`/similar-candidates` defaulted `threshold` to a hardcoded `0.7` and `limit` to `20`, while the service defaulted to the shared constants — and **no caller passes either**, so the route's literals were what prod ran and the constants were decorative. *Lowering the constant alone would have changed nothing.* Both now come from named constants (`CONTRADICTION_SIMILARITY_THRESHOLD`, new `CONTRADICTION_CANDIDATE_WINDOW=20`, both env-tunable), with a test pinning the wiring. Same third-place-duplication trap the scored-search route documents at length.

## Bench — reports a FAIL, and I want the reviewer to judge it, not me

67-question sample, d=0, vs the golden captured today on merged main:

| category | Δ accuracy | Δ recall@20 |
|---|---|---|
| single-session-user | **+7.7pp** | +0.0 |
| single-session-preference | **+8.3pp** | +0.0 |
| multi-session | +0.0 | −2.5pp |
| temporal-reasoning | **+14.3pp** | −8.0pp ❌ threshold |
| knowledge-update | −7.2pp | −4.2pp |
| **overall accuracy** | **+4.7pp (78.1 → 82.8)** | |

Per-question analysis of every negative cell:

- **The TR/KU recall dips are all multi-gold aggregation questions** ("total weeks spent reading", "order of airlines flown", "wake-up time Tue/Thu") where answer correctness was **unchanged** (False→False, False→False, True→True) and one even improved rank 2→1. The lost recall is set-recall on rows this change correctly demotes as superseded.
- **The KU accuracy drop is one question with IDENTICAL retrieval** (recall 1.0, rank 2 in both runs) — the answer generator phrased it badly; not a retrieval regression.

So the breach is the bench's recall metric counting superseded memories as required gold — the third time this has happened since supersession started actually working. I have **not** touched the golden or the thresholds to make this pass; filed as a separate finding instead.

## Honest limit found while validating this

With candidates now reaching it, gpt-5.4-nano still answers "clean" on 3/4 of these pairs: it will not connect *"finished the Snowflake cutover"* to *"the warehouse is BigQuery"*, or *"paging duty transferred to Lior"* to *"Oren holds the pager"*. I tried softening #1011's batch calibration; one case flipped — **but only while my probe sentences were quoted verbatim in the prompt**, and the win vanished when I rewrote the same guidance generically. I reverted that edit rather than ship something that scores well only on its own test set. Cross-vocabulary attribute matching is a judge-capability limit, which argues for predicate canonicalization (deterministic RDF, no LLM inference) or a tiered judge — filed as follow-ups.

Full suite 5,465 + storage 160 green; ruff/mypy clean.

### DIFF
--- a/common/constants.py
+++ b/common/constants.py
@@ -10,6 +10,8 @@ files, which is the drift it exists to remove.
 from datetime import timedelta
 from typing import NamedTuple
 
+from common.env_utils import read_float_env, read_int_env
+
 # ── Memory liveness ──
 # The statuses that mean "this memory is live". Broader than the literal
 # ``active`` value: enrichment promotes writes past ``active`` on the normal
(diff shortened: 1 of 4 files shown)


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #1026, opened 2026-08-27
ci: test TypeScript client on supported Node versions

### LINKED ISSUE #973: chore(ci): TypeScript client CI tests only Node 20 (engines says >=18) and uses npm install, not npm ci
### Problem

Two small gaps in `.github/workflows/client-typescript-ci.yml`:

**1. The engine floor is never tested.** `clients/typescript/package.json` declares:

```json
  "engines": { "node": ">=18" }
```

and the README says "uses native `fetch` (Node 18+)". CI runs one version:

```yaml
      - uses: actions/setup-node@v7
        with:
          node-version: "20"
```

Node 18 is the version most likely to break — it is where `fetch` is still experimental and where several `node:test` features landed later than 20. The `npm test` script is `tsc && node --test dist/index.test.js`, so both the build and the runner differ across versions.

**2. Installs are not reproducible.** The workflow runs `npm install`, not `npm ci`, despite `clients/typescript/package-lock.json` being committed. `npm install` may resolve outside the lockfile and rewrite it, so CI can pass against dependency versions no contributor has locally.

### Proposed solution

```yaml
    strategy:
      fail-fast: false
      matrix:
        node-version: ["18", "20", "22"]
    steps:
      - uses: actions/checkout@v7
      - uses: actions/setup-node@v7
        with:
          node-version: ${{ matrix.node-version }}
      - name: Install
        run: npm ci
      - name: Build + test
        run: npm test
```

The two changes are independent — a PR doing only `npm install` → `npm ci` is still worth merging.

### Alternatives considered

- Raising `engines` to `>=20` to match CI: simpler, but drops support the README advertises; testing the claim is the better direction unless the maintainers actually want to drop 18.
- Keeping `npm install` for dependency drift detection: that is what Dependabot is for; CI should be reproducible.

### Affected surface

CI only.

### Good-first-issue notes

One workflow file. The PR's own run verifies it. If the Node 18 job fails, that failure is the useful finding — report it on the PR rather than dropping 18 from the matrix, since it means the published `engines` range is wrong.

mikemikimike:
## Summary / Problem

Closes #973. The TypeScript client workflow only exercised Node 20 even though the package declares Node >=18, and it used `npm install` despite the committed lockfile.

## Changes

- Test the client on Node 18, 20, and 22 with a non-failing matrix.
- Use the matrix Node version in the setup step.
- Use `npm ci` for deterministic lockfile-based installs.

## Tests

- `npm ci` (from `clients/typescript`) — passed.
- `npm test` (from `clients/typescript`) — passed; 14 tests passed.
- `git diff --check` — passed.

## Compatibility / Known limitations

This changes CI coverage and dependency installation only; it does not change the published client API or runtime behavior. The three Node versions will be exercised by the hosted GitHub Actions workflow after the PR is opened.

## Issue link

Closes https://github.com/caura-ai/caura/issues/973

### DIFF
--- a/.github/workflows/client-typescript-ci.yml
+++ b/.github/workflows/client-typescript-ci.yml
@@ -14,6 +14,10 @@ on:
 jobs:
   test:
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        node-version: ["18", "20", "22"]
     defaults:
       run:
         working-directory: clients/typescript
@@ -21,8 +25,8 @@ jobs:
       - uses: actions/checkout@v7
       - uses: actions/setup-node@v7
         with:
-          node-version: "20"
+          node-version: ${{ matrix.node-version }}
       - name: Install
-        run: npm install
+        run: npm ci
       - name: Build + test
         run: npm test


Eldad-Caura:
Thank you for adding the supported Node-version matrix. The Node 18, 20, and 22 jobs pass, but DCO still fails because the functional commit 5e096895 does not contain your Signed-off-by trailer. Please add the sign-off to that commit yourself, rebase onto current main, and force-push the updated branch; we will rerun exact-head CI and review after that.

PR #1027, opened 2026-08-27
fix(client-python): raise health check HTTP errors

### LINKED ISSUE #971: fix(client-python): health() skips _raise_for_status — HTTP errors are returned as data, not raised
### Describe the bug

`Caura.health()` is the only method on the Python client that does not run the shared error mapping, so HTTP failures are returned to the caller as ordinary data instead of raising. A health check that gets a `401` or a `503` looks like a successful probe.

`clients/python/src/caura_client/client.py`:

```python
def health(self) -> dict[str, Any]:
    """Liveness probe (GET /api/v1/health)."""
    response = self._http.get("/api/v1/health")
    return response.json()
```

Every sibling calls `_raise_for_status` first — `_post` (used by `write`/`search`/`recall`), `get_document`, and `submit_interview`. The TypeScript client does not have this gap: its `health()` goes through `request()`, which calls `raiseForStatus()`, so the two official clients disagree on the same endpoint.

Secondary: `response.json()` on a non-JSON body (an HTML 502 from a proxy, say) raises a bare `json.JSONDecodeError` rather than a `CauraAPIError`.

### Steps to reproduce

```python
import httpx
from caura_client import Caura

transport = httpx.MockTransport(
    lambda request: httpx.Response(503, json={"detail": "database unavailable"})
)
mc = Caura("mc_test", tenant_id="t", transport=transport)
print(mc.health())
```

**Expected:** raises `CauraAPIError(503, "database unavailable")`, matching every other method and matching the TypeScript client.
**Actual:** returns `{'detail': 'database unavailable'}`. A caller doing `if client.health():` treats a dead deployment as live.

With `401` the divergence is worse: the other methods raise `AuthError`, `health()` returns the error body.

### Proposed fix

Add the one missing call:

```python
def health(self) -> dict[str, Any]:
    """Liveness probe (GET /api/v1/health)."""
    response = self._http.get("/api/v1/health")
    self._raise_for_status(response)
    return response.json()
```

Tests in `clients/python/tests/test_client.py` following the existing `httpx.MockTransport` pattern: `health()` raising `AuthError` on 401 and `CauraAPIError` on 503, plus the existing success path still passing.

Optionally, guard the non-JSON body in the same PR — but the status check is the actual bug and is fine to fix alone.

### Affected surface

Public client API — `health()` starts raising where it previously returned an error dict. That is the documented contract of every other method, and returning error bodies as success was not a usable behaviour, but it is worth a CHANGELOG note.

### Good-first-issue notes

One-line fix plus two tests. `clients/python/tests/test_client.py` already has `MockTransport` helpers to copy; no running server needed. Run with `cd clients/python && pip install -e ".[dev]" && pytest -q`.

mikemikimike:
## Summary

Closes #971.

`Caura.health()` now uses the Python client's shared HTTP error mapping before decoding the response. This keeps health checks consistent with the other client methods and prevents HTTP failures from being returned as successful data.

## Implementation

- Call `_raise_for_status()` in `clients/python/src/caura_client/client.py`.
- Add mocked regression coverage for 401 authentication errors and 503 API errors while preserving the existing success test.

This changes error responses from returned dictionaries to the documented typed exceptions. Successful responses and the public method signature are unchanged. No live service or API key is required.

## Validation

- `python -m pytest clients/python/tests/test_client.py -q` (18 passed; external pytest plugin autoload disabled because the host has an incompatible global `langsmith` plugin)
- `python -m ruff check clients/python/src/caura_client/client.py clients/python/tests/test_client.py` (passed)
- `git diff --check` (passed)

## Not run / known limitations

- Full pytest suite was not run; this focused client test covers the changed behavior without external services.
- Full mypy reports 3 pre-existing errors in `clients/python/src/caura_client/interviewer/cli.py` for Windows-unavailable `fcntl.flock`, unrelated to this change.
- Repository-wide formatter check reports existing formatting differences in untouched files; no formatter-only changes were included.

### DIFF
--- a/clients/python/src/caura_client/client.py
+++ b/clients/python/src/caura_client/client.py
@@ -113,6 +113,7 @@ class Caura:
     def health(self) -> dict[str, Any]:
         """Liveness probe (GET /api/v1/health)."""
         response = self._http.get("/api/v1/health")
+        self._raise_for_status(response)
         return response.json()
 
     def get_document(
--- a/clients/python/tests/test_client.py
+++ b/clients/python/tests/test_client.py
@@ -163,6 +163,25 @@ def test_health():
     assert make_client(handler).health()["status"] == "ok"
 
 
+def test_health_raises_auth_error():
+    def handler(request):
+        return httpx.Response(401, json={"error": {"message": "invalid key"}})
+
+    with pytest.raises(AuthError) as exc:
+        make_client(handler).health()
+    assert exc.value.status_code == 401
+
+
+def test_health_raises_api_error():
+    def handler(request):
+        return httpx.Response(503, json={"detail": "database unavailable"})
+
+    with pytest.raises(CauraAPIError) as exc:
+        make_client(handler).health()
+    assert exc.value.status_code == 503
+    assert str(exc.value) == "[503] database unavailable"
+
+
 def test_auth_error_parses_envelope():
     def handler(request):
         return httpx.Response(403, json={"error": {"message": "cross-fleet", "details": {"x": 1}}})


Eldad-Caura:
@claude

Eldad-Caura:
@erni-a

erni-a (approved):

PR #1029, opened 2026-08-28
fix(standalone): match a path-converter slug when the body guard skips injection

### LINKED ISSUE #1028: [Bug] Standalone tenant injection breaks skills-inbox reject/quarantine/edit when the slug carries a slash
### Caura version

2.33.1 (`main` at 38649e1, running with #993 and #995 merged in; the middleware is unchanged at 2913e61)

### Deployment mode

Other (please describe in Additional Context)

### Operating system

Linux (WSL2 on Windows 11), Docker Desktop Kubernetes

### Python version

3.12

### PostgreSQL version

pgvector/pgvector:pg16

### LLM provider in use

other / multiple

### Describe the bug

In standalone mode (`IS_STANDALONE=true`) the three skills-inbox actions that take a JSON body, `POST /api/v1/skills-inbox/{slug}/reject`, `/quarantine`, and `/edit`, answer 422 `unknown field 'tenant_id' is not permitted on this request body` whenever the slug carries a slash. Every candidate the factory produces has one (`forge/<name>` from the Forge, `agent/<name>` from an agent write), so on a self-hosted deployment a staged candidate can be approved or deferred through the API and cannot be rejected, quarantined, or edited. It happens with the tenant key and with the admin key, with or without `?tenant_id=`.

Cause, from the source: `StandaloneTenantMiddleware` injects `tenant_id` into a JSON body that lacks it, and the four action models are `STRICT_WRITE_BODY` (`extra="forbid"`, `routes/skills_inbox.py:269-294`, `schemas.py:52`). The middleware has a guard for exactly this, `_body_model_accepts_tenant_id` (`core-api/src/core_api/middleware/standalone_tenant.py:116`), which skips the injection for a route whose body schema has no `tenant_id`; its docstring names the skills-inbox actions. The guard matches the request path against the OpenAPI path templates with one segment per placeholder (`_path_template_to_regex`, `:221`, `{slug}` becomes `[^/]+`). The inbox routes declare `{slug:path}`, and OpenAPI renders that as `{slug}`, so a slug with a slash never matches its template, the guard fails open, and the field is injected into a body that forbids it. A slug without a slash passes body validation, which is how the cause was found.

### Steps to reproduce

Standalone deployment with `MEMCLAW_API_KEY` set; `KEY` holds that value.

```bash
U=http://localhost:8000/api/v1
c() { curl -s -w " [%{http_code}]\n" -H "X-API-Key: $KEY" -H "Content-Type: application/json" "$@"; }

# 1. opt the tenant in
c -X PUT "$U/settings?tenant_id=default" -d '{"skills_factory":{"enabled":true}}'

# 2. a candidate with a slash in its slug (the shape the Forge and agent writes produce); lands status "staged"
c -X POST $U/documents -d '{"tenant_id":"default","collection":"skills","doc_id":"agent/probe-skill-1","data":{"name":"Probe skill 1","slug":"probe-skill-1","description":"Synthetic skill for the report.","domain":"test","kind":"create","source":"agent","content":"# Probe skill 1\n\nSynthetic text."}}'

# 3. reject it
c -X POST "$U/skills-inbox/agent/probe-skill-1/reject?tenant_id=default" -d '{"reason":"probe"}'
# quarantine and edit answer the same; approve (no body) and defer (no body) work

# 4. control: the same body on a slug without a slash passes validation
c -X POST $U/documents -d '{"tenant_id":"default","collection":"skills","doc_id":"probe-skill-2","data":{"name":"Probe skill 2","slug":"probe-skill-2","description":"Synthetic skill for the report.","domain":"test","kind":"create","source":"agent","content":"# Probe skill 2\n\nSynthetic text."}}'
c -X POST "$U/skills-inbox/probe-skill-2/reject?tenant_id=default" -d '{"reason":"probe"}'
```

### Expected behavior

Step 3 answers 200 with `{"slug":"agent/probe-skill-1","previous_status":"staged","new_status":"rejected",...}`, as the route does for a slug without a slash: the middleware leaves the body alone because `RejectRequest` has no `tenant_id`.

### Actual behavior

Step 3:

```
{"detail":[{"type":"extra_forbidden","loc":["body","tenant_id"],"msg":"Extra inputs are not permitted","input":"default"}],"error":{"code":"INVALID_ARGUMENTS","message":"unknown field 'tenant_id' is not permitted on this request body (at 'tenant_id')","details":{"errors":[{"type":"extra_forbidden","loc":["body","tenant_id"],"msg":"Extra inputs are not permitted"}]}}} [422]
```

Step 4 gets past body validation (it then answers 422 `skill 'probe-skill-2' has no fingerprint; cannot poison cluster`, which is a different check: a hand-written candidate has no Forge fingerprint to poison; noted below).

### Relevant logs / config

```shell
IS_STANDALONE=true
MEMCLAW_API_KEY=<set>
# the request above, with the tenant key or the admin key, with or without ?tenant_id=default:
POST /api/v1/skills-inbox/agent/probe-skill-1/reject -> 422 extra_forbidden body.tenant_id
POST /api/v1/skills-inbox/agent/probe-skill-1/quarantine -> 422 extra_forbidden body.tenant_id
POST /api/v1/skills-inbox/agent/probe-skill-1/edit -> 422 extra_forbidden body.tenant_id
POST /api/v1/skills-inbox/agent/probe-skill-1/defer (no body) -> 200
POST /api/v1/skills-inbox/agent/probe-skill-1/approve -> 200
```

### Additional context

Deployment: the compose configuration rendered as Kubernetes manifests on Docker Desktop (one replica of `core-api`, `core-storage-api`, pgvector, redis); the request path is the same as compose.

Sending `tenant_id` in the body yourself does not help: the middleware then leaves the body alone and the model still refuses the field.

A fix with a test that fails before the change is on `zznate/caura`, branch `fix/skills-inbox-standalone-body`: the guard matches the templates in two passes, one segment per placeholder first and then across slashes only when nothing matched, bounded by the literal segments and the anchors. PR to follow.

Side observation from step 4, not part of this report: `reject` requires a Forge fingerprint (`has no fingerprint; cannot poison cluster`), so a hand-authored `staged` candidate can be approved but not rejected. Say if that deserves its own issue.

### Pre-flight

zznate:
## Summary

In standalone mode the tenant middleware's body-injection guard matched OpenAPI path templates one segment per placeholder, so a `{slug:path}` parameter with a slash never matched and `tenant_id` was injected into the skills-inbox action bodies, which forbid extra fields. Every Forge or agent candidate has such a slug, so `reject`, `quarantine`, and `edit` answered 422. The guard now matches in two passes, one segment per placeholder and then across slashes when nothing matched.

## Related Issue

Closes #1028

## Type of Change

## How Has This Been Tested?

`tests/test_standalone_tenant_body_guard.py`, written first: the converter case on a hermetic app (422 before, 200 after), the one-segment control, a body that declares `tenant_id` and still receives it, and the real inbox action path on the application (`_body_model_accepts_tenant_id` returns False). Reproduced on a standalone deployment of 2.33.1: `POST /api/v1/skills-inbox/agent/<slug>/reject` with `{"reason":"x"}` answered 422 `unknown field 'tenant_id'` with the tenant key and the admin key; the same call on a slug without a slash passed body validation.

```
pytest tests/test_standalone_tenant_body_guard.py tests/test_standalone.py tests/test_skills_inbox_routes.py tests/test_unknown_field_rejection.py tests/test_route_authz_gaps.py tests/test_stm_admin_tenant.py tests/test_env_write_new_names.py
150 passed
pytest tests/
16 failed, 5454 passed, 4 skipped, 1 xfailed
```

The 16 failures are the pre-existing ones in `test_integration_search`, `test_p3_2_relation_weights`, `test_a7_classifier_recall`, `pipeline/test_search_pipeline`, and `test_ph6_entity_linking_storage`, the same set as on `main` at 2913e61.

## Checklist

- [x] `mypy` passes (the four errors it reports on this file are missing third-party stubs, identical on `main`)
- [x] `pytest` passes locally (at the baseline above)
- [x] I have updated relevant documentation (docstrings; no user-facing docs name the guard)
- [ ] I have updated `CHANGELOG.md` under the `Unreleased` section (release-please generates it from the commit subject)

## Additional Notes

The second pass is bounded by the literal segments around each placeholder and by the anchors, and it runs only when the first pass matched nothing, so a route the one-segment pass resolves keeps its answer. The alternative, reading the converter from the route objects, is what the guard's docstring already rules out for FastAPI 0.137's `_IncludedRouter`.

### DIFF
--- a/core-api/src/core_api/middleware/standalone_tenant.py
+++ b/core-api/src/core_api/middleware/standalone_tenant.py
@@ -154,11 +154,25 @@ def _body_model_accepts_tenant_id(scope) -> bool:
         return True
     method = scope.get("method", "").lower()
     path = scope.get("path", "")
-    return not any(methods and method in methods and pattern.match(path) for pattern, methods in skip)
+    # Two passes. A placeholder is one path segment first, which is what
+    # the template says. The schema cannot say that a parameter was declared
+    # ``{slug:path}`` and so carries slashes (the skills-inbox actions take a
+    # Forge slug, ``forge/<name>``), so a path no template matched one
+    # segment at a time is matched again with placeholders spanning slashes.
+    # The second pass is bounded by the literal segments around each
+    # placeholder and by the anchors, and it runs only when the first found
+    # nothing, so a route the first pass resolves keeps its answer.
+    for across_slashes in (False, True):
+        for strict, loose, methods in skip:
+            pattern = loose if across_slashes else strict
+            if methods and method in methods and pattern.match(path):
+                return False
+    return True
 
 
-# ``id(app)`` -> tuple of (compiled path regex, frozenset of lowercase methods)
-# for the operations whose JSON body schema has no ``tenant_id`` property.
+# ``id(app)`` -> tuple of (one-segment path regex, across-slashes path regex,
+# frozenset of lowercase methods) for the operations whose JSON body schema
+# has no ``tenant_id`` property.
 # Memoised because it is derived from a schema that cannot change after
 # startup; keyed by app identity so a test that builds a second app is correct
 # rather than merely fast.
@@ -192,7 +206,13 @@ def _no_tenant_id_body_routes(app) -> tuple | None:
                 if "tenant_id" not in resolved["properties"]:
                     methods.add(method.lower())
             if methods:
-                entries.append((_path_template_to_regex(raw_path), frozenset(methods)))
+                entries.append(
+                    (
+                        _path_template_to_regex(raw_path),
+                        _path_template_to_regex(raw_path, across_slashes=True),
+                        frozenset(methods),
+                    )
+                )
         result: tuple | None = tuple(entries)
     except Exception:  # pragma: no cover — never break requests over this
         result = None
@@ -218,10 +238,17 @@ def _resolve_schema(node: dict, components: dict) -> dict | None:
     return node
 
 
-def _path_template_to_regex(template: str):
-    """``/api/v1/memories/{memory_id}`` -> a pattern matching one concrete path."""
+def _path_template_to_regex(template: str, *, across_slashes: bool = False):
+    """``/api/v1/memories/{memory_id}`` -> a pattern matching one concrete path.
+
+    A placeholder matches one segment, or, with ``across_slashes``, any
+    non-empty run of characters including slashes: the shape of a parameter
+    declared with the ``:path`` converter, which the OpenAPI template renders
+    as a plain ``{name}``.
+    """
     parts = re.split(r"(\{[^/}]*\})", template)
-    pattern = "".join("[^/]+" if p.startswith("{") and p.endswith("}") else re.escape(p) for p in parts)
+    placeholder = ".+" if across_slashes else "[^/]+"
+    pattern = "".join(placeholder if p.startswith("{") and p.endswith("}") else re.escape(p) for p in parts)
     return re.compile(f"^{pattern}$")
 
 
--- a/tests/test_standalone_tenant_body_guard.py
+++ b/tests/test_standalone_tenant_body_guard.py
@@ -1,0 +1,93 @@
+"""The standalone middleware's body-injection guard and path converters.
+
+``StandaloneTenantMiddleware`` injects ``tenant_id`` into a JSON body only
+when the route's declared body schema takes the field (SAFE-01). The guard
+matches the request path against the OpenAPI path templates, and OpenAPI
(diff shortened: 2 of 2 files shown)


Eldad-Caura:
Thank you for investigating #1028 and for adding focused coverage. The underlying issue remains valid, but we reproduced a correctness problem in this implementation: the across-slashes fallback treats every placeholder as .+, so an overlapping ordinary route can be mistaken for a path-converter route and required tenant injection can be suppressed. We are closing this implementation while keeping #1028 open for a converter-aware fix with an overlapping-route regression test.

PR #1034, opened 2026-08-28
fix(storage): accept discrete AlloyDB settings

Eldad-Caura:
## Summary

- let core-storage derive `DATABASE_URL` from resolved `ALLOYDB_HOST`, `ALLOYDB_PORT`, `ALLOYDB_USER`, `ALLOYDB_PASSWORD`, and `ALLOYDB_DATABASE` settings
- preserve explicit `DATABASE_URL` precedence and the existing local-development fallback
- fail startup on partial AlloyDB configuration and URL-encode credential/database components

## Why

The sandbox workflow currently has to assemble a complete DSN before deploying core-storage. Cloud Run then stores that DSN as a plain environment value. Supporting discrete settings lets the password remain a Secret Manager binding while non-secret connection fields remain ordinary environment variables.

This is the OSS prerequisite for caura-enterprise #1358; that workflow continues to check out OSS `main`, so this PR must land before the enterprise secret-backed deployment path is enabled.

## Validation

- 7 focused configuration tests pass
- Ruff check and format pass
- mypy passes for `core_storage_api.config`
- configured pre-commit hooks pass
- legacy-name ratchet passes
- negative controls confirmed the derivation and partial-config tests fail against the previous behavior

### DIFF
--- a/core-storage-api/src/core_storage_api/config.py
+++ b/core-storage-api/src/core_storage_api/config.py
@@ -3,9 +3,20 @@
 from __future__ import annotations
 
 from typing import Any, Literal
+from urllib.parse import quote
 
-from pydantic import field_validator
+from pydantic import Field, SecretStr, field_validator, model_validator
 from pydantic_settings import BaseSettings, SettingsConfigDict
+
+LOCAL_DATABASE_URL = "postgresql+asyncpg://memclaw:changeme@localhost:5432/memclaw"  # legacy-name-ok: existing local-dev compatibility default
+
+
+def _alloydb_database_url(*, host: str, port: int, user: str, password: str, database: str) -> str:
+    return (
+        "postgresql+asyncpg://"
+        f"{quote(user, safe='')}:{quote(password, safe='')}@"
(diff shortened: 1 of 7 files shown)


Eldad-Caura:
Claude Medium applied in bd6d00c4: both writer and reader database URLs are now `SecretStr`, excluded from representations and model serialization, and unwrapped only at database/preflight consumers. The regression test covers `repr`, `str`, `dict(settings)`, and `model_dump`; the dual-engine tests were updated for the secret type.

Validation: 8 config tests and 2 dual-engine tests pass; Ruff, mypy on changed production modules, pre-commit, diff-check, and the legacy-name ratchet pass. The full local integration suite could not be used because the existing local test database is partially migrated (`tenant_usage_counters` already exists); CI will run it in its provisioned environment.

Eldad-Caura:
@Codex

Eldad-Caura:
@claude

erni-a (approved):

Eldad-Caura:
@erni-a

erni-a (approved):

PR #1039, opened 2026-08-28
refactor(events): remove the unprovisioned Memory.CREATED topic

Eldad-Caura:
Removes `Topics.Memory.CREATED`, the one declared-but-unprovisioned topic blocking the `memory` family's rename cycle.

## Why remove rather than provision

`memclaw.memory.created` was declared and never wired to anything. Measured against the running project, not the source tree:

| check | result |
| --- | --- |
| topic in GCP, either name, either environment | **absent** — 143 topics enumerated in `alpine-theory-469016-c8` |
| the other four `memory` topics | present in all four env x name combinations, each with a DLQ |
| listed in `common/events/events_manifest.json` | **absent** — the other four appear under both prefixes |
| Terraform subscription | none |
| publisher or subscriber, either repo, non-test | **none** |
| references outside `caura` / `caura-enterprise` | none |

Its only uses were the enum declaration, two docstring examples, and tests that had picked it as an arbitrary sample topic.

Provisioning it would have meant creating 4 topics plus 4 DLQs, their subscriptions and IAM across staging and prod, and then carrying all of it through the rename — for an event nothing emits and nothing reads.

Leaving it in place is what blocks `memory`. A family that declares a topic which does not exist cannot be flipped safely: publishing into a missing topic is silent loss that Pub/Sub reports as success. That is the same defect that disqualifies `pipeline`, recorded in the `FLIPPED_FAMILIES` comment this PR updates.

## Test substitution, and the trap avoided

Tests used `Memory.CREATED` as a stand-in topic for bus mechanics. They now use `Memory.ENRICHED`, chosen because it appears **nowhere else** in those three files.

The obvious choice, `EMBEDDED`, would have been wrong. `test_drain_raises_on_circular_publish_chain` and `test_stop_swallows_circular_chain_runtime_error` each build a deliberate **two-topic** cycle — handler A publishes B, handler B publishes A. Both files already use `EMBEDDED` as the other half of that pair, so substituting it would have collapsed each cycle into a self-loop. Those tests would very likely still pass, while no longer testing the scenario their own comments describe.

## Legacy-name ratchet

Rewriting a line containing `memclaw` makes it new text, so grandfathered lines lose their exemption. Handled two ways, per the ratchet's own guidance that rewording beats marking:

- **Reworded** (no exemption needed): the module docstring now names the enum member without pinning its literal string. This drops a legacy-name occurrence outright, and stops the docstring going stale when `memory` flips.
- **Marked** `legacy-name-ok` (6 lines): assertions in `test_topics.py` and two wire payloads in `test_pubsub_bus.py`. These pin the current wire format on purpose — interpolating the enum instead would make them self-referential and assert nothing about the bytes on the bus.

## Verification

Run at this repo's exact CI scopes, exit codes read directly rather than through a pipe:

- `ruff check common/` — 0, clean
- `ruff format --check --isolated` (the three gated vendored files, plus `common/events/topics.py`) — 0, already formatted
- `scripts/legacy_name_ratchet.py --base origin/main` — 0, the 6 exemptions listed above reported for review
- `scripts/do_not_touch_sentinel.py` — 0, all 34 protected strings survive
- `pytest tests/ -m "not benchmark"` — 5469 passed, 4 skipped, 1 xfailed, 16 deselected

**One unresolved failure — do not merge until it is settled.** The full run on this branch reported `test_interview_async_submit.py::test_schedule_sweep_processes_pending_jobs` as failed (5469 passed, 1 failed).

A full-suite baseline on pristine `origin/main`, same machine and same venv recipe, came back **clean**: 5470 passed, 0 failed. So this cannot be written off as pre-existing, and I am not claiming it is.

What points the other way, for whatever it is worth: the test passes on its own twice against this branch, and the file contains no reference to `Memory.CREATED`, to `__members__`, or to any all-topics iteration, so there is no evident mechanism by which removing an unused enum member reaches it. It also runs ~41s and is a scheduler sweep, which is the shape a timing-dependent test takes.

A second full-suite run on this branch is in progress to separate a flake from a real ordering interaction. I will post the result here either way.

`uv.lock` files hashed before and after every `uv` command and are unchanged; `UV_FROZEN=1` throughout.

## Downstream note

This repo is public, so this removes a public enum member. It was never functional — the topic exists in no environment, so publishing through it would have raised `NotFound`. Flagging it rather than burying it, since it is an API surface change and not purely dead-code removal.

## Coupling

`common/events/topics.py` is vendored into `caura-enterprise` under the `manual` policy, so the two copies move together as one cluster. Enterprise counterpart: caura-ai/caura-enterprise#1373

### DIFF
--- a/common/events/topics.py
+++ b/common/events/topics.py
@@ -6,8 +6,8 @@ that ask a subscriber to do work.
 
 Uses `enum.StrEnum` (Python 3.11+) so members behave like the underlying
 string in every context: equality, dict-key hashing, f-string formatting,
-and Pub/Sub `topic_path` building all see `Topics.Memory.CREATED` as
-the literal `"memclaw.memory.created"`. A plain `(str, enum.Enum)` mix
+and Pub/Sub `topic_path` building all see `Topics.Memory.EMBEDDED` as
+its literal string value. A plain `(str, enum.Enum)` mix
 equates but does NOT format as the value — `f"{M.X}"` returns
 `"M.X"` — which would corrupt any string-formatted use site.
 """
@@ -18,7 +18,6 @@ import enum
 
 
 class Memory(enum.StrEnum):
-    CREATED = "memclaw.memory.created"
     EMBED_REQUESTED = "memclaw.memory.embed-requested"
     EMBEDDED = "memclaw.memory.embedded"
     ENRICH_REQUESTED = "memclaw.memory.enrich-requested"
@@ -93,7 +92,7 @@ class Lifecycle(enum.StrEnum):
 
 class Topics:
     """Namespaced facade so call sites keep the ergonomic form
-    `Topics.Memory.CREATED` instead of importing each inner enum."""
+    `Topics.Memory.EMBEDDED` instead of importing each inner enum."""
 
     Memory = Memory
     Audit = Audit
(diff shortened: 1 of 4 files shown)


arkash20 (approved):

Eldad-Caura:
Follow-up on the unresolved failure flagged in the description, since I said I would post the result either way.

Re-ran the full suite on this branch: **5470 passed, 0 failed**, zero `FAILED` lines (`pytest tests/ -m "not benchmark"`, same machine and venv recipe as before). `test_interview_async_submit.py::test_schedule_sweep_processes_pending_jobs` did not reproduce.

That puts the branch at exactly the pristine `origin/main` baseline count of 5470 passed. Combined with the test passing twice in isolation and the file carrying no reference to `Memory.CREATED`, to `__members__`, or to any all-topics iteration, the evidence points to a flake rather than a regression from this change.

Stated with its limit: one intermittent failure across three branch runs, and none in a single baseline run, is a small sample. This is "no evidence of a regression," not proof of independence. The test is a ~41s scheduler sweep, which is where I would look first if it recurs — worth tracking on its own rather than here, since nothing about it is specific to this change.

PR #1041, opened 2026-08-28
fix(logging): UVICORN_ACCESS_LOG was inert — enforce it where it's undone

Eldad-Caura:
The knob added in caura-ai/caura#1037 **does nothing**. I verified the deployment rather than assuming it: staging has `UVICORN_ACCESS_LOG=false` on every service, current images, a successful deploy — and hours later was still emitting **~2,400 access lines/hour**, from the serving revision.

### Why

uvicorn does not keep the `access_log=False` it is configured with. Its HTTP protocols re-derive the decision **once per connection**:

```python
self.access_log = self.access_logger.hasHandlers()
# uvicorn/protocols/http/{h11,httptools}_impl.py
```

`Config.configure_logging()` acts on the flag by clearing `uvicorn.access`'s handlers and setting `propagate=False`. Our own `_route_third_party_to_root` then runs — at app import, in every worker, *after* uvicorn's `Config` — and sets `propagate=True` to pull uvicorn's lines into structlog. `hasHandlers()` walks ancestors while they propagate, finds the **root** handler, and answers True. Every request is logged again.

Reproduced end to end:

```
after uvicorn access_log=False  -> hasHandlers() = False
after structlog reroute         -> hasHandlers() = True
=> uvicorn protocol sets self.access_log = True   (logs every request)
```

### The fix

Enforce the setting in the module that undoes it. When the line is unwanted, `uvicorn.access` is **silenced** rather than rerouted — no handlers **and** `propagate=False`, which is what makes `hasHandlers()` answer False, because the walk stops at the first non-propagating logger. The snapshot is still taken so `_reset_for_testing` restores.

Two things fall out of putting it here rather than in `serve.py`:

- the parse moves to `access_log_enabled()`, so it lives in **one** place instead of being inlined (`common/env_utils`, where it would otherwise belong, does not exist in the enterprise repo this file is vendored into byte-for-byte)
- it now works for the services that exec the **uvicorn CLI** directly and never run `serve.py` — core-worker, core-operations, platform-operations — which the previous approach structurally could not reach

`serve.py` still passes `access_log=`; it is correct and explicit, just not sufficient alone, and its comment now says so.

### On the tests — this is the part worth reviewing

The original tests asserted `uvicorn.run` was called with `access_log=False`. That was true, and irrelevant: they tested **the call, not the behaviour**, so they passed green while the feature did nothing. There is also a signature guard, which only proved the parameter exists.

The new test asserts the property uvicorn actually reads, with a root handler installed — the exact trap:

```python
assert logging.getLogger("uvicorn.access").hasHandlers() is False
```

Confirmed it **fails without the fix** and passes with it. Also confirmed the snapshot test fails if the snapshot is dropped from the new branch, so neither test passes either way.

They also get a save/restore fixture, because my first cut restored by hand and leaked — clearing `_third_party_logger_original_state` discards the snapshots taken for uvicorn / fastmcp / mcp / slowapi too, leaving the module believing it had never rerouted anything.

### Checks

- `tests/test_logging.py` + `tests/test_serve.py`: **78 passed**
- full suite: 38 failed / 5452 passed here vs **39 failed / 5435 passed on `origin/main`** — the same pre-existing environment failures (pubsub, provider protocols, gemini, ingest), 17 more passing on this branch. None are logging-related.
- `ruff format --check --isolated` clean on both vendored files (the enterprise re-vendor gate), `ruff check` clean, all lines ≤88 cols

### Follow-on

Both files are `identical`-policy vendored, so the enterprise side re-vendors automatically once this merges. No enterprise change is needed — caura-ai/caura-enterprise#1369 already set the env var on the seven `common.serve` services; the three uvicorn-CLI services it deliberately skipped are now covered for free, and their exclusion comment there is worth revisiting separately.

### DIFF
--- a/common/serve.py
+++ b/common/serve.py
@@ -42,12 +42,11 @@ from __future__ import annotations
 
 import argparse
 import importlib
-import os
 from typing import Any
 
 import uvicorn
 
-from common.structlog_config import configure_logging
+from common.structlog_config import access_log_enabled, configure_logging
 
 
 def _load(path: str) -> Any:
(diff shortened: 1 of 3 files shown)


Eldad-Caura:
Fixed — and it was worse than the line numbers suggest: the comment block ended up sitting **directly above** `def access_log_enabled()` with no blank line between them, so it read as that function's header and promised a WARNING the function never emits.

Moved `access_log_enabled()` above the comment rather than editing the comment, so the block adjoins `_THIRD_PARTY_LOGGERS_TO_REROUTE` again — the subject it was written for. No logic change.

```
def access_log_enabled() -> bool:
    ...
    return raw not in {"0", "false", "no", "off"}


# Names of third-party loggers that install their own (typically stderr,
# ... The function logs a WARNING in that case so the caller can
# move the call into an ASGI lifespan startup handler ...
_THIRD_PARTY_LOGGERS_TO_REROUTE: tuple[str, ...] = (
```

Re-confirmed after the move that the regression test still **fails without the fix** and passes with it — a pure reorder can still break a test that depends on import-time ordering, so it is worth re-running rather than assuming. `ruff format --check --isolated` and `ruff check` clean, all lines ≤88 cols, 78 passed across `test_logging.py` + `test_serve.py`.

Also tidied a self-inflicted DCO problem while amending: `-s` on an amend appended a second `Signed-off-by`, and my first dedup left the survivor stranded mid-message instead of as a trailer. It is a single trailing sign-off now.

Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

Eldad-Caura:
@erni-a

erni-a (approved):

PR #1042, opened 2026-08-29
fix(ratchet): pass -z when asking git which paths changed

Eldad-Caura:
`_changed_paths()` called `git diff --name-only` **without `-z`**, so git honoured `core.quotePath` and returned any path holding a non-ASCII byte C-quoted:

```
git diff --name-only   →  "caf\303\251.py"   (C-quoted, quotes included)
git grep -z            →  café.py            (raw — what everything else uses)
```

Every other path in the module is raw, because `_grep()` reads them with `-z` and explains why in its own docstring. A quoted path therefore matches nothing in the exempt tally the selection intersects with, so the file drops silently out of the diff-based selection and falls back to the risen tally alone — reopening, **for exactly those files**, the net-zero exemption-swap blind spot `_changed_paths()` exists to close.

Not a hypothetical class of file here: `_git()` pins UTF-8 rather than trusting the locale precisely because the tree carries accented names in fixtures.

## Measured, with an ASCII control

Same swap in both cases — one marker removed from a file that already held exemptions, one newly marked line added:

| filename | before this PR |
|---|---|
| `ascii.py` | reported ✅ |
| `café.py` | **silently missed** ❌ |

## Report-only

`_changed_paths()` feeds the exemption report, which sits downstream of the pass/fail decision. Verified on this tree: **same exit code** either way. This changes what gets printed, never whether the gate fails.

## Test

`test_a_net_zero_swap_is_reported_under_a_non_ascii_filename` **fails without the `-z`**, while the existing ASCII swap test passes either way — which is why that one could never have caught this.

---

Found by the Claude reviewer on caura-ops#332 and fixed there first. The same missing `-z` was inherited from the canonical copy into every repo carrying `_changed_paths`; this is that one-line change in the copy that PR could not reach.

### DIFF
--- a/scripts/legacy_name_ratchet.py
+++ b/scripts/legacy_name_ratchet.py
@@ -391,12 +391,23 @@ def _changed_paths(base: str) -> set[str] | None:
     ``None`` rather than an empty set on failure, because the two mean opposite
     things to the caller: "could not tell" must fall back to the old tally-based
     selection, while "nothing changed" must select nothing.
+
+    ``-z`` for the same reason :func:`_grep` passes it, and it is load bearing
+    here in a way that is easy to miss. Without it git honours ``core.quotePath``
+    and returns a path holding any non-ASCII byte C-quoted — ``café.py`` comes
+    back as the ten characters ``"caf\\303\\251.py"``, quotes included. Every
+    other path in this module is raw, because ``_grep`` reads them with ``-z``,
+    so a quoted path matches nothing in ``after`` and the file drops silently out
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@erni-a ready for your approval — the Claude review came back clean, all checks green.

One-line fix plus its regression test: `_changed_paths()` now passes `-z` so a non-ASCII path is not C-quoted out of the exemption report's file selection. Report-only, exit code verified unchanged.

erni-a (approved):

PR #1043, opened 2026-08-29
fix(client): ship py.typed for legacy memclaw package

### LINKED ISSUE #970: fix(client-python): memclaw_client legacy shim has no py.typed — types are lost on the legacy import
### Describe the bug

`memclaw_client` is documented as a permanent import alias that "behaves exactly as it did in 0.4.x", but it is not marked as typed, so type checkers treat it as an untyped third-party module while `caura_client` is fully typed.

- `clients/python/src/caura_client/py.typed` exists.
- `clients/python/src/memclaw_client/` contains only `__init__.py` — no `py.typed`.
- `clients/python/pyproject.toml` declares package data for one package only:

```toml
[tool.setuptools.package-data]
caura_client = ["py.typed"]
```

Per PEP 561, inline types are only honoured for a package that ships its own `py.typed` marker. The re-export in `memclaw_client/__init__.py` (`from caura_client import *`) does not inherit the marker from the source package.

### Steps to reproduce

```python
# t.py
from memclaw_client import Caura
reveal_type(Caura)
```

```
$ pip install caura-client
$ mypy t.py
```

**Expected:** `Caura` resolves with full type information, identical to importing from `caura_client`.
**Actual:** mypy reports the module as untyped (`module is installed, but missing library stubs or py.typed marker`) and `Caura` degrades to `Any`. Swapping the import to `caura_client` type-checks correctly.

### Proposed fix

1. Add an empty `clients/python/src/memclaw_client/py.typed`.
2. Extend package data:

```toml
[tool.setuptools.package-data]
caura_client = ["py.typed"]
memclaw_client = ["py.typed"]
```

### Affected surface

Internal (not in the SemVer contract) — packaging metadata only, no runtime behaviour change.

### Good-first-issue notes

Two-line change. Worth confirming the marker actually lands in the built wheel (`python -m build && unzip -l dist/*.whl | grep py.typed`) rather than only in the source tree.

mikemikimike:
## Related issue

Closes #970.

## Background

The legacy `memclaw_client` package re-exports the typed Caura client but did not ship a PEP 561 marker, so type checkers treated imports through the compatibility package as untyped.

## Change

- Add `memclaw_client/py.typed`.
- Include the marker in the `memclaw_client` package data configuration.

## Verification

- `python -m build --wheel --outdir dist` — passed.
- Wheel inspection confirmed both `caura_client/py.typed` and `memclaw_client/py.typed` are packaged.
- `git diff --check` — passed.

The client tests were attempted with pytest but the Windows host has pre-existing environment incompatibilities (`fcntl` is unavailable and pytest temporary-directory setup raises `PermissionError`). No runtime code was changed.

### DIFF
--- a/clients/python/pyproject.toml
+++ b/clients/python/pyproject.toml
@@ -61,6 +61,7 @@ where = ["src"]
 
 [tool.setuptools.package-data]
 caura_client = ["py.typed"]
+memclaw_client = ["py.typed"]
 
 [tool.ruff]
 target-version = "py39"


Eldad-Caura:
Thank you for the contribution. The legacy memclaw_client package this PR adds py.typed to was removed from main in #1244, so there is no longer a package target for this change. Closing this as obsolete after that removal.

PR #1044, opened 2026-08-29
feat(clients): add stats() to the Python and TypeScript clients

### LINKED ISSUE #552: feat(clients): add stats() to the Python and TypeScript clients
### Problem

Both official clients expose only health, recall, search, write, but the server has a read-only stats endpoint (GET /api/v1/stats, core-api/src/core_api/routes/stats.py) useful for dashboards and self-introspection, requiring no elevated trust at default scope.

### Proposed solution

add stats() to clients/python/src/memclaw_client/client.py and clients/typescript/src/index.ts, mirroring the endpoint's query params (scope, memory_type, status, fleet_id, include_deleted); add tests in both existing test files following the mocked-transport patterns; add a README section to each client.

### Alternatives considered

Calling GET /api/v1/stats directly with httpx/fetch: works, but duplicates the auth-header and error-handling logic the clients already encapsulate, and every consumer reinvents it.
- Using the MCP memclaw_stats tool: only available to MCP-connected agents, not to plain SDK consumers (dashboards, scripts, CI).
- Adding the full remaining endpoint surface (list/manage/doc) at once: much larger review burden; stats() is the smallest read-only step toward parity and sets the pattern.

### Affected surface

New surface (not currently exposed)

No, but I'm available to discuss

### Additional context

_No response_

### Pre-flight

cl0ud08:
﻿## Summary

Adds `stats()` to the Python and TypeScript clients, wrapping `GET /api/v1/stats`. Returns a new `Stats` model/interface exposing `tenant_count`, `memory_count`, `agent_count`, plus the raw payload on `.raw`.

## Related Issue

Closes #552.

## Type of Change

- [ ] Documentation update
- [ ] Other:

## How Has This Been Tested?

- `pytest` in `clients/python`: 138 passed. The 7 failures in `test_interviewer_installer.py` / `test_interviewer_platform.py` are POSIX-only (`fcntl`, chmod semantics) and reproduce identically on unmodified `main` on Windows.
- `npm test` in `clients/typescript`: 17/17 pass; `tsc` compiles clean.
- New tests cover: correct path with empty query string, counter parsing, missing keys defaulting to `0`, and a 503 raising `CauraAPIError` with the right status code.

## Checklist

- [x] `ruff check` and `ruff format --check` pass — see tooling note below
- [x] `mypy` passes — see tooling note below
- [x] `pytest` passes locally

## Additional Notes

The issue describes query params (`scope`, `memory_type`, `status`, `fleet_id`, `include_deleted`) that don't exist on the route — `public_stats()` in `core-api/src/core_api/routes/stats.py` has an empty signature. Implemented as a no-arg method matching the route as it stands; flagged this on the issue before starting.

Two deliberate choices, both easy to reverse:

- Counters are deployment-wide (`count_all(tenant_id="")`), not scoped to `tenant_id` like every other client method. Called out explicitly in both docstrings and both READMEs so nobody misreads `memory_count` as their own tenant's count.
- `stats()` calls `_raise_for_status()` although `health()` deliberately doesn't. A wrong `base_url` or a deployment predating this route returns non-2xx, and a bare `.json()` would surface a `ValueError` instead of a `CauraAPIError`.

Tests mock the exact response shape emitted by `stats.py` rather than an assumed one.

**CHANGELOG:** not edited — it's release-please-generated with no `Unreleased` section. The commit message is a conventional commit, so the entry should generate on release. Happy to add one manually if that's wrong.

**Tooling notes**, both pre-existing and unrelated to this change:

- `ruff check` passes clean. `ruff format --check` reports 11 files, including `tests/test_client.py` — all are already unformatted on `main`, verified by checking out upstream's copy and re-running. Not reformatted here to keep the diff scoped; happy to do it as a separate PR.
- `mypy` passes on every file this PR touches. It reports 3 errors in `interviewer/cli.py` (`fcntl.flock` / `LOCK_EX` / `LOCK_NB`), which are Windows-only — that file is untouched here and errors identically on upstream's version.

### DIFF
--- a/clients/python/README.md
+++ b/clients/python/README.md
@@ -48,6 +48,7 @@ with Caura("standalone", tenant_id="default", base_url="http://localhost:8000") 
 | `search(query, top_k=5, ...)` | `POST /api/v1/search` | `list[Memory]` |
 | `recall(query, top_k=5, ...)` | `POST /api/v1/recall` | `RecallResult` |
 | `health()` | `GET /api/v1/health` | `dict` |
+| `stats()` | `GET /api/v1/stats` | `Stats` |
 | `get_document(doc_id, *, collection, ...)` | `GET /api/v1/documents/{doc_id}` | `dict` |
 | `submit_interview(...)` | `POST /api/v1/interview/submit` | `dict` |
 | `close()` | — | `None` |
@@ -76,6 +77,22 @@ Caller-owned keys belong under `metadata` (`mc.write("...", metadata={"tags": [.
 `search()` and `recall()` are unaffected: filter bodies still accept unknown
 fields, deliberately. See
 [api-surfaces.md](https://github.com/caura-ai/caura/blob/main/docs/api-surfaces.md#request-body-contract-writes-are-strict-searches-are-not).
+
+### Deployment stats
+
+`stats()` returns three counters — `tenant_count`, `memory_count`,
+`agent_count` — for the landing-page status bar:
+
+```python
+with Caura("mc_xxx", tenant_id="my-team") as mc:
+    s = mc.stats()
+    print(s.tenant_count, s.memory_count, s.agent_count)
+```
+
+These are **deployment-wide totals, not scoped to your `tenant_id`** — the
+server counts across every tenant on the deployment. The underlying endpoint
+is unauthenticated (no key required to call it directly), but this method
+still sends your API key header for consistency with every other call.
 
 ### Fetching a document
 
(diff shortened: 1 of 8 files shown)


Eldad-Caura:
Thank you for implementing stats() in both clients and adding focused tests. The Python and TypeScript suites pass, but DCO fails because commit 355f7143 does not contain your Signed-off-by trailer. Please amend and sign off that commit yourself, rebase onto current main, and force-push the updated branch; we will rerun exact-head CI and review after that.

PR #1045, opened 2026-08-29
perf(logs): drop the CAURA-132 diagnostics to DEBUG

Eldad-Caura:
`contradiction_detector` has **21 `logger.info` calls and zero `logger.debug`** — it has no debug tier at all, so everything it says ships at production level. Seven of those calls are marked `CAURA-132 diag`, added to answer one question: why the entity-aware judge returned `verdict=False` on populated contexts.

Two of the seven fire **once per (memory × candidate) pair** rather than once per memory, against ~10 candidates per memory. Measured over 6h of prod on 2026-08-29, the seven were **3,767 of this logger's 4,947 lines — 76%**:

| pattern | lines / 6h |
|---|---|
| `PATH_A_SEMANTIC verdict` | 1,872 |
| `PATH_C_DETECTION verdict` | 1,128 |
| `PATH_A_SEMANTIC entry` | 174 |
| `PATH_C_DETECTION entry` | 165 |
| `PATH_C_DETECTION after_a1_17` | 150 |
| `PATH_C_DETECTION context_fetched` | 143 |
| `PATH_C_DETECTION judge_selection` | 135 |

The wet-test they were added for has already run — CAURA-133's test cites *"the wet-test on dev v2.12.1 (CAURA-132 logs)"*. They served their purpose and are now paying rent per candidate.

### What stays

**Demoted, not deleted.** `LOG_LEVEL=DEBUG` brings them straight back, and the question they answer can recur.

Everything that records an **outcome** stays at INFO — the `path_a_completed` / `path_c_completed` summaries (which already carry `n_conflicts`, `n_candidates`, `elapsed_ms`), the contradictions actually found, Path C retracting a Path A verdict, and the preflight fail-open notes. Losing those costs real forensics; losing per-candidate trace lines does not.

One INFO call *inside* a converted block was deliberately left alone: `"Path C preflight skipped all N candidates"` is a decision outcome, not a trace.

### Verified that DEBUG is actually filtered

This is the part I would not skip again. core-api defaults `log_level` to INFO and prod sets no override, and driving a real logger through `configure_logging("production", "INFO")` drops the DEBUG record while the INFO one emits.

That check exists because the `UVICORN_ACCESS_LOG` change earlier this week asserted that a *flag was passed* rather than that *logging stopped* — it shipped green and did nothing in production for a full day. A verbosity change that only reads the source has exactly the same defect.

### Tests

Three layers, and I confirmed each fails when broken rather than assuming:

- tier asserted at all seven call sites (AST, so it survives reformatting) — reverting one demotion **fails**
- a **reverse guard** on the outcome lines, so a later cleanup can't sweep them down too — demoting one outcome line **fails**
- end-to-end filtering through the real logging config

273 contradiction tests pass, 1 skipped. `ruff check` and `ruff format --check` clean.

### Worth knowing before you merge

This target is **#4**, not #1. In a clean post-access-log-fix window, prod's top loggers are:

```
platform_auth_api.routers.auth            2573
core_api.access                           2554   <- a second, app-level access log
mcp.server.streamable_http                 962
core_api.services.contradiction_detector   872   <- this PR
```

`core_api.access` is notable: a separate application-level access log that the uvicorn work did not touch. If the goal is volume rather than this specific module, those two are each roughly 3x the target here — happy to look at them next.

### DIFF
--- a/core-api/src/core_api/services/contradiction_detector.py
+++ b/core-api/src/core_api/services/contradiction_detector.py
@@ -570,7 +570,26 @@ async def _detect(
         # Symmetric to PATH_C_DETECTION entry log; lets us tell apart
         # "Path A ran but found no semantic candidates" from "Path A
         # ran, found candidates, and the LLM judge said no".
-        logger.info(
+        #
+        # All seven CAURA-132 diag sites in this module log at DEBUG, not
+        # INFO. They were added to answer one question — why the entity-
+        # aware judge returned verdict=False on populated contexts — and
+        # that wet-test has since run (see CAURA-133, whose test cites
+        # "the wet-test on dev v2.12.1 (CAURA-132 logs)"). What they cost
+        # in the meantime: two of them fire once per (memory x candidate)
+        # pair rather than once per memory, so on 2026-08-29 the seven
+        # accounted for 3,767 of this logger's 4,947 prod lines in 6h —
+        # 76% — against ~10 candidates per memory.
+        #
+        # DEBUG keeps them one env var away (LOG_LEVEL=DEBUG) rather than
+        # deleting them, because the question they answer can recur. What
+        # stays at INFO is everything that records an OUTCOME: the
+        # path_a/path_c_completed summaries (which already carry
+        # n_conflicts, n_candidates and elapsed_ms), the contradictions
+        # actually found, Path C retracting a Path A verdict, and the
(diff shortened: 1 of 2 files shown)


arkash20 (approved):

Eldad-Caura:
@erni-a

erni-a (approved):

PR #1046, opened 2026-08-29
refactor(ratchet): file permanent names under the floor marker

Eldad-Caura:
Every exemption in this repo carried `legacy-name-ok`, whose report tells a reviewer to check each line is *"a compat alias, a redirect or a pinned wire format"*. Most were not. `legacy-name-floor` exists for the other claim — a line that merely **names** something the rename will never reach and cannot be correct without the literal — and nothing had been filed under it yet. This does that filing.

The split is what makes the alias list readable: rule 3's escape hatch is the thing that needs eyes, and it was diluted by lines nobody needs to re-litigate.

## How each line was decided

By what the line **does**, not by how its reason was worded, with **ties broken toward the alias** — an alias filed as floor is a rule 3 decision nobody looks at, while a floor mention filed as an alias is one extra line in a list already being read. Anything without a clear floor claim was left alone.

The bulk of what moved is `scripts/do_not_touch_sentinel.py`, whose entries pin exact text living elsewhere in the tree: rewording one breaks the sentinel, so nothing is declared there and the literal is load-bearing.

## Behaviour-neutral, and checked rather than assumed

Both markers exempt a line identically, so a swap cannot move a line in or out of the gate:

- exempt and non-exempt tallies **identical** before and after
- ratchet exits **0**
- `do_not_touch_sentinel.py` exits **0**
- the repo's ratchet and sentinel test suites pass

## Reason text shortened with the marker

Under a `legacy-name-floor` marker, `pinned floor string` and `…, which is floor` say nothing the marker has not. Shortening was applied **per claim, never per line** — the reason is the report's grouping key, so shortening only some lines of a group would split one claim across two headings.

## Expect a double report on this PR

Both halves are correct and neither is a failure: the swapped lines are new exempt *text*, so the new-exemption report lists them; the old text is gone, so the removal report lists them too. This was measured before the work started and is the known shape of a marker swap.

### DIFF
--- a/core-api/src/core_api/constants.py
+++ b/core-api/src/core_api/constants.py
@@ -207,7 +207,7 @@ DOC_MEMORY_URI_SCHEME = "caura-doc"
 # recognizer for doc-memory URIs (the reconciliation pass above) must accept
 # these schemes alongside it.
 LEGACY_DOC_MEMORY_URI_SCHEMES = (
-    "memclaw-doc",  # legacy-name-ok: scheme already persisted in customers' memories.source_uri
+    "memclaw-doc",  # legacy-name-ok: compat scheme persisted in customers' memories.source_uri
 )
 
 assert DOC_MEMORY_MAX_CHARS <= MAX_CONTENT_LENGTH, (
--- a/core-api/src/core_api/routes/plugin.py
+++ b/core-api/src/core_api/routes/plugin.py
@@ -693,7 +693,7 @@ _VALID_SKILL_AGENTS = {"claude-code", "codex", "both"}
 # reach either. ``memclaw`` is the default (the operational manual); the
 # opt-in ``company-brain`` posture skill layers on top of it.
 _SKILL_LABELS = {
-    "memclaw": "Caura",  # legacy-name-ok: wire — ?skill= param + on-disk dir
+    "memclaw": "Caura",  # legacy-name-floor: wire — ?skill= param + on-disk dir
     "company-brain": "Company Brain",
 }
 _VALID_SKILLS = frozenset(_SKILL_LABELS)
--- a/core-worker/pyproject.toml
+++ b/core-worker/pyproject.toml
@@ -1,7 +1,7 @@
 [project]
 name = "core-worker"
 version = "2.34.0"
-description = "Caura core async worker — subscribes to memclaw.memory.embed-requested and backfills embeddings" # legacy-name-ok: frozen topic name
+description = "Caura core async worker — subscribes to memclaw.memory.embed-requested and backfills embeddings" # legacy-name-floor: frozen topic name
 license = "Apache-2.0"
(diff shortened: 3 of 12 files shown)


Eldad-Caura:
@claude Fixed the finding — thanks, it was right.

The header comment above `SENTINELS` still told the reader the entries carry `legacy-name-ok` while this PR moved them to `legacy-name-floor`, so the prose contradicted the list directly beneath it.

Rewritten to say what is actually true, and checked per repo rather than asserted: most entries now carry `legacy-name-floor`, and the few still on `legacy-name-ok` pin something that really is a compat alias and say so in their own reason.

I also recorded a coupling the old wording never mentioned, because it bit me while doing this: where an entry's pinned text includes the source line's own trailing comment, the marker is part of the string being matched — so swapping one side without the other stops the pin matching and fails this gate. That is not hypothetical; it happened in caura-enterprise and the sentinel caught it.

Re-verified after the change: exempt and non-exempt tallies unchanged, ratchet exit 0, sentinel exit 0.

Eldad-Caura:
@claude Addressed both Low findings. Restored legacy-name-ok on the active LEGACY_DOC_MEMORY_URI_SCHEMES compat roster and its contract assertion, with the test marker shortened so formatting keeps it beside the literal. Reworded the sentinel header to state that all current sentinel copies are legacy-name-floor because they name exact source strings, even when a source line declares an alias. Sentinel, ratchet, Ruff format/check, mypy, and the direct legacy-scheme invariant pass.

Eldad-Caura:
@claude Two fixes since your last pass, both from your findings.

**The floor claim was wrong on lines whose own reason says the name moves.** You caught `config.py:28` ("live host, moves at the flip") — right, and it was a hole in how I classified: I keyed on "live host" and never treated scheduled-change language as disqualifying. Floor means the rename will never reach it, so a reason saying otherwise settles the question. Swept that rule across every repo and reverted **8** lines in total (7 in caura-enterprise, 1 here) — the Datadog LLM Observability app and staging domain "on its own rename schedule", the Artifact Registry repo "being replaced; retired only in §7", and the `memclawd` repo reference. All back on `legacy-name-ok`.

**The corrected `SENTINELS` comment had a run-on line.** My replacement text joined onto the sentence that followed it, producing a 114-character line — which failed `ruff check` in caura-test-automation, whose CI lints `scripts/` at 100. Rewrapped in every repo, including the differently-worded variant here.

Re-verified after both: exempt and non-exempt tallies unchanged, ratchet exit 0, sentinel exit 0, suites green, and every repo's own lint scope clean.

arkash20 (approved):

Eldad-Caura:
@erni-a Ready for approval: latest Claude review is clean and the full CI suite plus all required checks are green.

erni-a (approved):

PR #1048, opened 2026-08-29
fix(observability): label mounted-app traffic instead of "unmatched"

Eldad-Caura:
I went to cut `core_api.access` volume and came back with the opposite recommendation, plus a defect.

## Don't cut this one

It is **not** the `uvicorn.access` case. This line carries `http_route`, `http_method`, `http_status_code`, `http_duration_ms` **and `tenant_id`**, and its own docstring says it backs a **log-based distribution metric** for per-endpoint API-usage dashboards. Removing or filtering it takes out the metric. Nothing else in the stack carries per-tenant request attribution.

What *is* wrong with it is the label.

## The defect

Only FastAPI's `APIRoute` sets `scope["route"]`. A Starlette `Mount` never does — starlette 0.50 contains no such assignment at all. `app.py` mounts `/mcp` and `/static`, so every request into either fell through to `"unmatched"`, the bucket documented as *"404s, or failures before routing"*.

6h of prod, 2026-08-29:

| | count |
|---|---|
| `unmatched` GET 200 | 4,115 |
| `unmatched` POST 200 | 838 |
| `unmatched` POST 202 | 74 |
| `unmatched` 4xx — the only real 404s | 69 |

**5,027 of 5,099 (98.6%) were successful requests.** That is ~40% of the metric's events unattributable, the entire MCP transport invisible in the per-endpoint dashboard, and an `unmatched` count that looks like a 404 problem and isn't.

## The fix

Fall back to the **mount prefix**. That keeps the cardinality contract — mounts are declared statically in `app.py`, so it adds one bounded label each, never a raw path. Genuine 404s still read `unmatched`, and that label now means what the docstring says.

The prefix is the `root_path` **delta** across the downstream call (captured before, read after), because a `Mount` appends to `root_path`. Subtracting `app_root_path` instead looks equivalent and is not: with the app served under `--root-path /gw`, a genuine 404 gets labelled `/gw`. There is a test for exactly that, and it fails against the `app_root_path` version — I checked rather than reasoned.

## Verification

- Reproduced against **fastapi 0.128 / starlette 0.50**, with and without an app-level root_path, rather than assuming framework behaviour
- Confirmed **no capability-map key** collides with `/mcp` or `/static` (35 keys, 0 hits), so mounted traffic still records no REST capability and the adoption signal is unchanged
- Three new tests, each confirmed to **fail without its fix**: mount labelling (both mounts), 404-still-wins, and app-root_path-does-not-leak
- 20 passed across `test_request_observation.py` + `test_capability_usage.py`; ruff check and format clean

## On volume

**This commit changes volume by zero**, which is not what I was asked for, so to be explicit about the trade: cutting a metric's source to save log lines is the wrong direction, and I would rather say so than quietly ship it.

The one real volume option here is that `/health` and `/version` probes are **12%** of these events and are not "API usage" — arguably they inflate the dashboard as well as the bill. But that is a call for whoever owns the dashboard, not for a log-reduction pass. Happy to do it on request.

### DIFF
--- a/core-api/src/core_api/middleware/request_observation.py
+++ b/core-api/src/core_api/middleware/request_observation.py
@@ -13,11 +13,24 @@ deployments they land in the JSON log file just like any other log line.
 
 Cardinality contract
 --------------------
-``http_route`` is ALWAYS the route template read from
-``scope["route"].path`` (e.g. ``/api/v1/memories/{id}``) or the literal
-``"unmatched"`` when no route matched (404s, or failures before routing).
-It is NEVER the raw request path with path params interpolated — raw paths
-would explode the metric's label cardinality and make it expensive.
+``http_route`` is ALWAYS one of three bounded things, in this order:
+
+1. the route template from ``scope["route"].path``
+   (e.g. ``/api/v1/memories/{id}``) — set by FastAPI's ``APIRoute``;
+2. the mount prefix (``/mcp``, ``/static``) for traffic served by a
+   mounted sub-app, which is NOT a FastAPI route and so never populates
+   ``scope["route"]``;
+3. the literal ``"unmatched"`` — genuine 404s and failures before
+   routing.
+
+It is NEVER the raw request path with path params interpolated — raw
+paths would explode the metric's label cardinality and make it
+expensive. Mount prefixes are safe on that count because mounts are
+declared statically in ``app.py``: one extra label each.
+
+Case 2 was folded into case 3 until 2026-08-29, which made ~40% of prod
+events unattributable and hid MCP traffic from the per-endpoint
+dashboard entirely — see the fallback in ``__call__`` for the numbers.
 
 ``scope["route"]`` only exists AFTER the router has run, so this middleware
 must inspect it once the downstream app returns, not before.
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
Fair, and the pin observation is correct — `requirements.txt` has `fastapi>=0.115,<1` and no direct starlette constraint, so a transitive bump can move both behaviours.

Left the approach as is, and documented why, because I think the existing protection is stronger than a pin:

**It degrades safely.** If either assumption stops holding — `Mount` no longer mutating this scope dict in place, or starting to set `scope["route"]` — the delta comes out empty and the label falls back to `"unmatched"`. That is exactly today's behaviour. There is no path to a *wrong* label or an unbounded one; the worst case is losing the improvement.

**It degrades loudly, in CI.** This is the part I would push back on: `tests/test_request_observation.py` drives real `app.mount()` calls through the real middleware over a real ASGI transport. Nothing about the framework is mocked. So a dependency bump that changes either behaviour turns `test_mounted_subapp_labels_by_mount_prefix` red before it reaches prod — it is a behavioural guard, not a documentation claim.

That is also why a pin would catch *less*. A pin stops the bump; it never tells you which assumption broke, and it silently rots as the pin ages. The test says precisely what changed. Same reasoning as the `test_uvicorn_really_accepts_the_access_log_kwarg` guard added earlier this week, for the same class of risk.

What I did add is the missing part — the code now says all of this explicitly, including a note not to replace those tests with mocks, since that is the one edit that would give the unpinned dependency a silent path back in.

If you would still like starlette pinned directly, happy to do it, but I would rather that be a deliberate dependency decision than a side effect of this fix.

Re-verified after the amend: 20 passed, ruff check and format clean. Comment-only change on top of the reviewed commit.

Eldad-Caura:
@claude

arkash20 (approved):

PR #391, opened 2026-06-14
build(core-*): adopt FastAPI 0.137 — drop the <0.137 caps, relock

Eldad-Caura:
## What
Now that **#364** made core-api's CAURA-602 startup guard read `app.openapi()` (robust to FastAPI 0.137's `include_router` sub-app mounting), the temporary `fastapi<0.137` caps added during the 2026-06-14 incident are no longer needed. This drops them and adopts 0.137 fleet-wide.

- All four `core-*` services: `fastapi>=0.115,<0.137` → **`<1`**.
- Relocked (`uv lock --upgrade-package fastapi`) → **fastapi 0.137.0 + starlette 1.3.1** across all four. The committed locks pin the exact versions; the Dockerfiles install frozen from them.
- `greenlet<3.5.0` kept (platform-wheel guardrail, unrelated).

## Verification
- **CI runs the test suites under 0.137** here — CI installs lock-less (`uv pip install -e …`), so with the caps gone it resolves 0.137 and runs lint/mypy/pytest against it. **This PR's green CI is the gate** (the "verify beyond import" step).
- Earlier proof that the guard holds under 0.137: a linux/amd64 build imported `core_api.app` green on fastapi 0.137.0 (#364).
- If CI surfaces any 0.137 runtime/route-behavior issue beyond the guard, it shows here before merge.

## Rollback
Trivial — revert this PR; the locks return to 0.136.x and the caps come back.

Signed-off-by: eldad-caura <eldad@caura.ai>

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1050, opened 2026-08-29
fix(observability): stop logging successful health and version probes

Eldad-Caura:
Follow-up to #1048, which established that `core_api.access` is worth keeping
because it carries `tenant_id` and backs a per-endpoint API-usage metric. This
removes the one part of its volume that is definitively *not* API usage.

## What and why

Cloud Run probes `/api/v1/health` and `/api/v1/version` on a fixed interval per
instance. Over the 7 days to 2026-08-29 prod logged **37,850** successful
probes, and **100% of them were untenanted** — not one came from a real API
consumer:

| route | tenant | status | 7d count |
|---|---|---|---|
| `/health` | *(none)* | 200 | 32,047 |
| `/version` | *(none)* | 200 | 5,804 |
| `/health` | *(none)* | 503 | 440 |
| `/version` | *(none)* | 405 | 3 |
| `/health` | *(none)* | 405 | 2 |

`/health` was the **third-busiest route** on a dashboard whose purpose is
showing which endpoints customers use.

## Successful probes only

The 440 `/health` 503s in that table are the reason this filters on status
rather than route. A failing probe is precisely what an incident investigation
reaches for; suppressing by route alone would have discarded all 440 to save
1.2% more volume. Non-2xx probes still log.

What this gives up: "are probes arriving at all?" is no longer answerable from
these logs, because silence now means both "healthy" and "nothing is calling
us". Cloud Run reports instance health directly and failures still log, so that
gap is covered elsewhere. It's noted in the code.

## Correcting my own number from #1048

I wrote in the #1048 body that probes were "12% of these events". **That was a
sampling artifact and I was wrong to state it unqualified.** The probe rate is
near-constant while real traffic swings ~8×, so the *share* moves entirely with
the denominator:

| day | total | probe (ok) | share |
|---|---|---|---|
| 2026-08-23 | 377,252 | 5,665 | 1.5% |
| 2026-08-24 | 342,713 | 4,564 | 1.3% |
| 2026-08-26 | 405,511 | 5,145 | 1.3% |
| 2026-08-27 | 255,618 | 5,504 | 2.2% |
| 2026-08-28 | 46,568 | 5,827 | **12.5%** |

My 6h sample landed on 08-28, a quiet day. The honest figure is the absolute
one: **~5.4k lines/day, permanently**. That is a *small* win on volume — the
stronger case for this change is that it stops infrastructure noise from
dominating a customer-usage dashboard.

## The subtle part: which path space?

`_PROBE_ROUTES` holds `/health`, **not** `/api/v1/health`. Since FastAPI 0.137,
`include_router(prefix=...)` mounts the router instead of flattening it, so the
prefix lives in `root_path` and never reaches `http_route`. Verified against the
locked fastapi 0.141.1 / starlette 1.3.1 by driving real requests through the
real `core_api.app`:

```
GET /api/v1/version      -> 200, no event emitted        (suppressed)
GET /api/v1/health       -> 503, http_route='/health'    (failure kept)
GET /api/v1/memories/abc -> 401, http_route='/memories/{memory_id}'
GET /api/v1/nope         -> 404, http_route='unmatched'
```

**The tests build the app with `include_router(prefix=...)`, not
`@app.get("/health")`.** That distinction is the whole point: registering the
route directly on the app produces the label `/health` by coincidence, so such
a test passes whether or not the real prefixed case works. Getting this wrong is
not hypothetical — see below.

Confirmed the two suppression tests fail without the middleware change:

```
FAILED test_successful_probe_emits_no_event[/api/v1/health]
FAILED test_successful_probe_emits_no_event[/api/v1/version]
```

`test_prefixed_non_probe_route_still_logs_its_stripped_label` pins the label
space itself, so a future FastAPI that returns to flattening prefixes turns CI
red rather than silently restoring the probe traffic.

## Also corrected

Three claims in this file that were wrong, two of them mine from #1048:

- pinned versions are **fastapi 0.141.1 / starlette 1.3.1**, not "0.128 / 0.50"
- the dependency range is `fastapi>=0.141.1,<1`, not `>=0.115,<1`
- the cardinality contract's example label was `/api/v1/memories/{id}` — a label
  this metric never emits

## Separately: `_REST_CAPABILITY` is dead and I have not touched it here

While verifying the label space I found that all 35 entries in
`_REST_CAPABILITY` (same file) are keyed `/api/v1/...`, so **none can ever
match** — the REST half of the capability adoption signal has recorded nothing
since FastAPI 0.137 was adopted:

- #353 added the map on 2026-06-14, correct at the time
- #391 adopted FastAPI 0.137 on 2026-06-15, changing the label space

It worked for one day. `tests/test_capability_usage.py:157` registers
`@app.post("/api/v1/recall")` directly on the app, which is why it stays green.
Only the MCP half still records, so the adoption report looks plausible rather
than empty.

I've deliberately left it out of this PR — it's a behavioural change to a
different signal and deserves its own review and its own revert boundary.

### DIFF
--- a/core-api/src/core_api/constants.py
+++ b/core-api/src/core_api/constants.py
@@ -150,6 +150,22 @@ MEMORY_STATUSES_PATTERN = (
 # — return ``0`` / "unreachable" rather than block landing-page hits).
 PROBE_TIMEOUT_SECONDS = 5.0
 
+# Probe route paths, declared here rather than inline in ``routes/health.py``
+# so that exactly one string backs both the route decorator and the
+# access-log suppression in ``middleware/request_observation.py``. Those two
+# have to agree or successful-probe traffic silently returns to the logs, and
+# a shared constant makes disagreement impossible instead of merely
+# detectable — renaming the endpoint moves the suppression with it.
+#
+# These are ROUTER-RELATIVE, which is what ``scope["route"].path`` carries and
+# therefore what the access log's ``http_route`` label is. ``app.py`` serves
+# them under ``include_router(prefix="/api/v1")``, so the URL a caller hits is
+# ``/api/v1/health`` while the label is ``/health``. Keep that distinction:
+# writing the prefixed form here would match nothing.
+HEALTH_PATH = "/health"
+VERSION_PATH = "/version"
+PROBE_ROUTES = frozenset({HEALTH_PATH, VERSION_PATH})
+
 # ── Memory visibility levels ──
 # Named constants are the SoT — ``MEMORY_VISIBILITIES`` and the regex below
 # derive from them so a rename here propagates to membership checks and
--- a/core-api/src/core_api/middleware/request_observation.py
+++ b/core-api/src/core_api/middleware/request_observation.py
@@ -5,18 +5,27 @@ Pure ASGI middleware — matches the ``SecurityHeadersMiddleware`` /
 than ``BaseHTTPMiddleware`` (which wraps in Starlette's task groups and has
 known edge cases around cancellation and streaming responses).
 
-Emits exactly one structured ``http.request`` log event per HTTP request,
+Emits at most one structured ``http.request`` log event per HTTP request,
 carrying the *templated* route, method, status, and wall-clock duration.
 These events back a log-based distribution metric for per-endpoint
 API-usage dashboards on Cloud-Logging deployments; on file-logging
 deployments they land in the JSON log file just like any other log line.
 
+"At most" rather than "exactly": a liveness/readiness probe that SUCCEEDS
+is not logged, because it is infrastructure chatter rather than API usage.
+A probe that FAILS still is. See ``PROBE_ROUTES`` for the routes, the
+rationale, and what that costs.
+
 Cardinality contract
 --------------------
 ``http_route`` is ALWAYS one of three bounded things, in this order:
 
-1. the route template from ``scope["route"].path``
-   (e.g. ``/api/v1/memories/{id}``) — set by FastAPI's ``APIRoute``;
+1. the route template from ``scope["route"].path`` — set by FastAPI's
+   ``APIRoute``. Note this is the template as the ROUTER declares it, with
+   no router prefix on it: a GET of ``/api/v1/memories/abc`` is labelled
+   ``/memories/{memory_id}``. Since FastAPI 0.137,
+   ``include_router(prefix=...)`` mounts the router instead of flattening
+   it, so the prefix lives in ``root_path`` and never reaches this label;
 2. the mount prefix (``/mcp``, ``/static``) for traffic served by a
    mounted sub-app, which is NOT a FastAPI route and so never populates
    ``scope["route"]``;
@@ -62,9 +71,44 @@ from typing import Any
 from starlette.types import ASGIApp as ASGIApplication
 from starlette.types import Receive, Scope, Send
 
+from core_api.constants import PROBE_ROUTES
 from core_api.services.capability_usage import record_usage
 
 logger = logging.getLogger("core_api.access")
+
+# Liveness/readiness probe routes, whose SUCCESSFUL requests are not logged.
+#
+# ``PROBE_ROUTES`` is imported rather than restated because ``routes/health.py``
+# builds its route decorators from the very same constants — one string backs
+# both, so the endpoint cannot be renamed out from under this check. See the
+# note beside it in ``constants.py`` for why the values are router-relative
+# (``/health``, not ``/api/v1/health``).
+#
+# Why drop them: Cloud Run probes each instance on a fixed interval, so
+# these arrive at a near-constant rate no matter what real traffic does —
+# ~5.4k/day in prod, of which 100% over 7 days to 2026-08-29 (37,850 events)
+# were untenanted infrastructure calls, never a real API consumer. As a
+# SHARE of this metric they are wildly unstable for the same reason: 1.3% on
+# a busy day, 12.5% on a quiet one. The stable number is the absolute rate.
+#
+# Why only the successful ones: a probe that returns 200 answers no question
+# this metric exists to answer, but a failing one is exactly what an
+# incident investigation reaches for — prod served 440 ``/health`` 503s in
+# that same week. Suppressing by route alone would have thrown all 440 away
+# to save 1.2% more volume. Non-2xx probes stay.
+#
+# What this gives up: "are probes arriving at all?" can no longer be
+# answered from these logs, since silence now means both "healthy" and
+# "nothing is calling us". Cloud Run reports instance health directly, and a
+# probe that fails still logs, so that gap is covered elsewhere.
+#
+# And the denominator for these two routes is now gone, so a per-route ERROR
+# RATE (failures / total) is meaningless for them — every probe event that
+# survives is a failure, so any such ratio reads 100% the moment one 503
+# lands. For PROBE_ROUTES read the absolute failure COUNT instead. Nothing
+# in-repo derives a rate from this metric today; the point is that an alert
+# added later must not, and an on-call engineer seeing "100% errors on
+# /health" should recognise it as this suppression rather than an outage.
 
 # REST route-template → (capability, op) for the adoption signal. Keyed by
 # (METHOD, templated path) so the same path under different verbs maps to the
@@ -122,7 +166,11 @@ _REST_CAPABILITY: dict[tuple[str, str], tuple[str, str | None]] = {
 
 
 class RequestObservationMiddleware:
-    """ASGI middleware that emits one ``http.request`` event per request."""
+    """ASGI middleware emitting at most one ``http.request`` event per request.
+
+    "At most" because a SUCCESSFUL liveness/readiness probe emits none; see
+    ``PROBE_ROUTES``. Every other request, failed probes included, emits one.
+    """
 
     def __init__(self, app: ASGIApplication) -> None:
         self.app = app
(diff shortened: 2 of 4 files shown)


Eldad-Caura:
Thanks — the concern is the right one to raise, but the premise is wrong on the
pinned version, and the recommendation it leads to is weaker than what the code
now does. Taking both parts separately.

## The `include_router` claim is verified, not assumed

> this is not how FastAPI has historically implemented `include_router` (it
> copies routes and prepends the prefix into the route's own `path`)

Correct about history, and that is exactly what changed. FastAPI 0.137 made
`include_router(prefix=...)` mount the router as an opaque `_IncludedRouter`
(`path=None`, no public `.routes`). This repo already documents the change and
was already bitten by it — `app.py:1030-1037` explains that the route-path
guard there had to move to `app.openapi()` for precisely this reason:

> FastAPI 0.137 changed `include_router(prefix=...)` to mount the router as an
> opaque `_IncludedRouter` (path=None, no public `.routes`), so the prefixed
> paths are no longer top-level `APIRoute.path` entries — which is exactly what
> silently broke this guard when 0.137 shipped.

Verified three independent ways rather than from docs:

**1. The locked versions, driving real requests through the real `core_api.app`:**

```
fastapi 0.141.1 | starlette 1.3.1
GET /api/v1/version      -> 200, no event emitted
GET /api/v1/health       -> 503, http_route='/health'
GET /api/v1/memories/abc -> 401, http_route='/memories/{memory_id}'
GET /api/v1/nope         -> 404, http_route='unmatched'
```

Note the pins are `fastapi>=0.141.1,<1` / `starlette 1.3.1` per `core-api/uv.lock`
— not `requirements.txt`, which is unrelated to this service.

**2. Production, which is the part that makes this not a judgement call.** The
labels prod has been emitting for months carry no prefix:

| http_route | 6h count |
|---|---|
| `/fleet/heartbeat` | 1,945 |
| `/health` | 1,347 |
| `/documents/query` | 1,196 |
| `/documents/{doc_id}` | 662 |

There is no `/api/v1/...` label in prod at all. This isn't a prediction about
library behaviour — it's the observed output of the deployed system.

**3.** The rebased commit also fixes stale version strings in this file that
said "0.128 / 0.50"; that was my error in #1048 and I suspect it fed the review's
doubt. They now read 0.141.1 / 1.3.1.

## On the blast radius — one half is real, and already filed

> would silently break the REST route→capability mapping table ... unless that
> table's keys were also updated

**This is correct, it already happened, and it is not caused by this PR.** All 35
`_REST_CAPABILITY` keys are `/api/v1/...`, so none can match. Timeline:

- #353 (2026-06-14) added the map, correct at the time
- #391 (2026-06-15) adopted FastAPI 0.137

It worked for one day and has recorded nothing since. `tests/test_capability_usage.py:157`
stays green because it registers `@app.post("/api/v1/recall")` directly on the
app instead of via `include_router(prefix=...)` — the label matches there and
only there. This is written up at the bottom of the PR description and has its
own task; I deliberately kept it out of this PR since it changes a different
signal and deserves its own revert boundary.

The dashboards/alerts half of the concern has the same answer: any filter on a
`/api/v1`-prefixed label has been dead since June, independently of this change.
This PR alters no label.

## What I changed anyway

Suggestion (a), a startup assertion, only *detects* drift. Suggestion (b),
deriving from the router, is the better instinct, so I did the strict form of
it: `PROBE_ROUTES` now lives in `constants.py`, and `routes/health.py` builds its
decorators from those same constants:

```python
@router.get(HEALTH_PATH, responses=...)
@router.get(VERSION_PATH, responses=...)
```

One string backs both the endpoint and the suppression, so a rename moves the
suppression with it — drift becomes impossible rather than merely detectable, and
no startup check is needed for that failure mode.

This closed a real gap, and thank you for pushing on it: every test in the file
previously built its own `/health`, so all of them would have stayed green if
`routes/health.py` renamed the endpoint. Two tests now cover it:

- `test_real_health_router_probe_is_suppressed` imports the **production** health
  router and serves it the way `app.py` does, so it exercises the actual
  registration rather than a stand-in.
- `test_probe_constants_are_the_ones_the_router_registers` asserts every
  suppressed path is one the real router registers.

Both were confirmed to fail against injected faults rather than assumed to work:

```
# endpoint renamed /health -> /healthz
FAILED test_successful_probe_emits_no_event[/api/v1/health]

# literal path reintroduced in the decorator, constant left behind
AssertionError: ['/health'] is suppressed from the access log but is not a path
the health router registers (['/healthz', '/status', '/tool-descriptions',
'/version', '/whoami']) — the suppression matches nothing and probe traffic is
being logged
```

And the two original suppression tests fail without the middleware change, so
none of this is vacuous. Rebased onto current main; ruff lint + format pass on
`core-api/src/`.

Eldad-Caura:
@claude re-review please — force-pushed since your last pass. `PROBE_ROUTES` moved to `constants.py` and `routes/health.py` now builds its route decorators from those same constants, plus two new tests covering the real production router.

Eldad-Caura:
Both actionable items fixed. Taking the High separately, since it isn't one for
this PR.

## Class docstring — fixed, and a fair catch

I updated the module docstring to "at most one" and left the class docstring
saying "emits one". Now:

```python
"""ASGI middleware emitting at most one ``http.request`` event per request.

"At most" because a SUCCESSFUL liveness/readiness probe emits none; see
``PROBE_ROUTES``. Every other request, failed probes included, emits one.
"""
```

## The collision suggestion — pinned rather than documented

You're right that `PROBE_ROUTES` matches a bare label with no notion of which
router owns it, so a second router declaring `/health` would have its successful
traffic silently dropped.

I didn't take the "scope it to the health router" option: doing that at runtime
couples the middleware to a specific router and puts an import in a hot path, to
guard a case that doesn't exist yet. But documenting the coupling is too weak —
it's a silent failure, and a comment doesn't catch anything.

So the assumption is now pinned by a test that walks the **real** app and asserts
each probe path is declared exactly once. Getting the real label space required
descending through `_IncludedRouter` via `original_router`, since
`include_router(prefix=...)` mounts opaquely — the same obstacle `app.py:1030`
hit. That's a private attribute, so the assertion is written to be
self-validating: if the walk ever stops descending, the probes are found **zero**
times and the test fails rather than quietly passing.

Confirmed against an injected collision rather than assumed:

```
# second router registered declaring its own /health
AssertionError: '/health' is declared by 2 route(s) in the real app; the
access-log suppression matches this label with no regard for which router owns
it, so anything other than exactly 1 means either the walk stopped working (0)
or a second router is having its successful traffic silently dropped (>1)
```

## `_REST_CAPABILITY` — agreed, already filed, deliberately not here

Your diagnosis is correct in every particular, including that
`test_capability_usage.py` registers `@app.post("/api/v1/recall")` directly on
the app and is green for exactly that reason. I found the same thing while
verifying the label space for this change and wrote it up in the PR description
before your first pass, with the timeline: #353 added the map on 2026-06-14,
#391 adopted FastAPI 0.137 on 2026-06-15. It worked for one day and has recorded
nothing for ~2.5 months.

It has its own task queued. Keeping it out of this PR is deliberate, not an
oversight — it changes a different signal (`record_usage` starts firing for REST
traffic where it currently never does), and it deserves its own review and its
own revert boundary. Bundling a live behavioural change to the adoption pipeline
into a log-volume reduction would make both harder to reason about and to back
out.

Worth stating plainly for anyone reading later: **this PR does not change any
route label.** The labels have been unprefixed since June. Nothing here makes
`_REST_CAPABILITY` worse; the diff only made the existing breakage legible.

Rebased on current main, all checks green before this push.

Eldad-Caura:
@claude re-review please — amended since your last pass: class docstring corrected, plus a test that walks the real app and pins each probe path to exactly one declaring router.

Eldad-Caura:
Good catch, and it's now documented rather than left implicit — I'd covered "are probes arriving at all?" but not the ratio distortion.

Added to the `PROBE_ROUTES` note:

> And the denominator for these two routes is now gone, so a per-route ERROR RATE (failures / total) is meaningless for them — every probe event that survives is a failure, so any such ratio reads 100% the moment one 503 lands. For `PROBE_ROUTES` read the absolute failure COUNT instead. Nothing in-repo derives a rate from this metric today; the point is that an alert added later must not, and an on-call engineer seeing "100% errors on /health" should recognise it as this suppression rather than an outage.

On your option (1), exclude from the error-rate calc: I checked both repos for a consumer before writing that, and there is no dashboard, alert or SLO definition referencing `http.request` or `http_route` outside the Python source — the log-based metric is configured outside version control. So there is nothing to exclude it from yet, which makes (2) the only actionable half today and the reason the note is aimed at whoever adds the first alert.

Rebased onto current main (picked up #1051, #1053). All checks were green on the prior push; this adds a comment block only.

Eldad-Caura:
@erni-a ready for review. Automated review is clean — its remaining note (per-route error-rate distortion for the suppressed routes) is addressed in the code comment, with the caveat that no in-repo consumer of this metric exists to exclude them from.

Summary of what this does: successful `/health` and `/version` probes no longer emit access-log events; failing ones still do. ~5.4k lines/day, and `/health` stops being the third-busiest route on a dashboard meant to show customer API usage. Every one of the 37,850 probe events in the last 7 days was untenanted infrastructure traffic.

Two things worth your eye specifically:
- The suppression list is router-relative (`/health`, not `/api/v1/health`) because FastAPI 0.137 mounts prefixed routers. `routes/health.py` now builds its decorators from the same constants, so a rename moves the suppression with it.
- I found while verifying this that `_REST_CAPABILITY` in the same file has been dead since June for exactly that reason — written up at the bottom of the PR body, and deliberately left for a separate PR.

erni-a (approved):

PR #1051, opened 2026-08-29
refactor(plugin): clean stale legacy-name prose

Eldad-Caura:
## Summary

- clean the 20 PR2 legacy-brand mentions proven safe in plugin prose, comments, test names, and inert fixtures
- rename the block-local isMemClawOrphan variable to isPluginOrphan
- use canonical caura_* names for two deliberately nonexistent tool fixtures
- correct stale test descriptions that contradicted current Caura behavior

## Scope guards

- leaves educate.test.ts lines 695 and 808 untouched; those memclaw_ comments pin pre-rename fixture behavior
- leaves every sentinel-pinned fence tag, heading, backup suffix, ownership marker, install path, plugin id, and skill slug unchanged
- changes no runtime, wire, persisted, or customer-disk contract

## Verification

- plugin TypeScript build passes
- full plugin suite: 425 passed
- affected plugin suites after rebase: 155 passed
- version sync: plugin 2.19.3 is current
- sentinel: All 39 protected strings survive.
- ratchet: No new lines. 20 removed by this change (-20 net).
- report: 752 lines across 162 files
- every CI Ruff check and format scope passes
- every CI mypy scope passes
- one DCO-signed commit

PR #1053, opened 2026-08-29
test(plugin): cover historical phantom-file cleanup

Eldad-Caura:
## Summary

- Adds the regression guard required by the OSS legacy-name review plan before or with PR2.
- Covers cleanup of a historical phantom TOOLS.md under the plural workspaces parent when it carries the old on-disk heading.
- Follow-up to #1051, which was human-merged before this guard could be added.

## Sensitivity proof

Temporarily disabling only the historical-heading predicate makes the new test fail because the phantom file survives. Restoring the predicate makes it pass.

## Validation

- Full plugin suite: 426 passed, 0 failed on the byte-identical tree
- TypeScript build and focused regression test: passed
- Do-not-touch sentinel: all 39 protected strings survive
- Legacy-name ratchet: one pinned compatibility fixture; no new lines
- Plugin version sync: 2.19.3 current
- Focused simplify pass: clean after correcting the marker kind

## Scope

Test-only. No product logic or persisted contract changes.

### DIFF
--- a/plugin/src/educate.test.ts
+++ b/plugin/src/educate.test.ts
@@ -396,6 +396,22 @@ describe("writeEducationFiles", () => {
     assert.ok(!existsSync(join(base, "workspaces", "AGENTS.md")));
   });
 
+  test("cleans up pre-existing phantom legacy TOOLS.md at <baseDir>/workspaces/", () => {
+    const base = tmpBase();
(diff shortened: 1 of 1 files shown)


Eldad-Caura:
@erni-a Ready for approval — Claude review found no issues and all required checks are green.

erni-a (approved):

PR #1056, opened 2026-08-29
fix(observability): re-key the REST capability map to real route labels

Eldad-Caura:
Follow-up to #1050, which surfaced this while verifying the route-label space.

## The bug

Every one of the 35 entries in `_REST_CAPABILITY` was keyed `/api/v1/...`, but
the `http_route` the middleware looks them up by is **router-relative**. Measured
against the real app rather than argued:

```
real labels: 96   map entries: 35
map keys that MATCH a real (method, label) pair:  0 of 35
after stripping the prefix:                      35 of 35
```

So `record_usage` has never fired for REST traffic. The MCP half (via the
`call_tool` wrapper) kept working, which is why the adoption report looked
plausible instead of empty — a silent failure, not a visible one.

## How it broke

- **#353** (`0421ceb0`, 2026-06-14) added the map, prefixed. Correct at the time.
- **#391** (`fdcec535`, 2026-06-15) adopted FastAPI 0.137, which mounts a
  prefixed router instead of flattening it, moving the prefix into `root_path`.

It worked for one day and has recorded nothing for ~2.5 months.

## Why no test caught it

`test_rest_middleware_records_mapped_capability` registered its route as
`@app.post("/api/v1/recall")` — directly on the app. That is the **one shape**
that makes a prefixed key match, because without a router there is no prefix to
move. The test was green the entire time production recorded nothing.

It now builds the app the way `app.py` really does:

```python
router = APIRouter()

@router.post("/recall")
async def recall(request: Request): ...

app.include_router(router, prefix="/api/v1")
```

## The guard that prevents recurrence

`test_every_capability_key_matches_a_real_route` walks the **real** app and
asserts every key corresponds to a real `(method, label)` pair. It catches both
this failure and the mundane one — a route renamed while its capability entry is
left behind, silently switching that capability off.

**On the suggested import-time guard in `app.py`:** I deliberately did not add
one. Recovering the label space requires descending through FastAPI's
`_IncludedRouter` via the private `original_router`, because
`include_router(prefix=...)` mounts opaquely. The existing
`_TIMEOUT_OPT_OUT_PATHS` check can live at import time precisely because
`app.openapi()` is public and prefixed; there is no public API exposing the
label space. Wiring a private attribute into startup would convert a dependency
bump into a **production boot failure**. Failing CI is the right blast radius.

The walk is self-validating: if it ever stops descending, no key matches and the
test fails loudly rather than passing on an empty comparison.

## Confirmed failing before the fix

```
FAILED test_rest_middleware_records_mapped_capability
E       assert 0 == 1

FAILED test_every_capability_key_matches_a_real_route
E       AssertionError: 35 of 35 capability keys match no route on the real app,
        so those capabilities record nothing: [('DELETE', '/api/v1/documents/{doc_id}'), ...]
```

And confirmed working end-to-end against the real `core_api.app` after:

```
POST /api/v1/recall       -> record_usage(capability='recall', transport='rest')
GET  /api/v1/memories/abc -> record_usage(capability='manage', op='read', transport='rest')
GET  /api/v1/health       -> no call (not a capability route)
```

## One semantic change worth flagging

`/keystones` is served under both `/api/v1` and the legacy `/api/v1/memclaw`
alias. Since the label drops the prefix, those two **collapse to a single entry
per verb** — 35 entries become 32. Both aliases already mapped to identical
capabilities so nothing is lost today, but the two URL forms are no longer
distinguishable in this signal. That would matter only if they ever need to
diverge, and it's noted in the code.

## Expect the numbers to move

This turns a signal back on rather than changing one. REST capability usage will
go from zero to real values, so the adoption report's REST/MCP split will shift
noticeably on deploy — that is the fix working, not a regression. Worth knowing
before anyone reads the next report.

### DIFF
--- a/core-api/src/core_api/middleware/request_observation.py
+++ b/core-api/src/core_api/middleware/request_observation.py
@@ -116,52 +116,78 @@ logger = logging.getLogger("core_api.access")
 # tool vocabulary (mcp_server tool names minus the ``caura_`` prefix, and
 # the manage/doc sub-ops) so REST and MCP roll up together in the report.
 #
+# The paths are ROUTER-RELATIVE — ``/memories``, not ``/api/v1/memories`` —
+# because that is what ``scope["route"].path`` carries and therefore what
+# ``http_route`` is matched against below. Same label space as
+# ``PROBE_ROUTES``; the note beside it in ``constants.py`` explains why.
+#
+# They were prefixed until 2026-08-29, and every lookup missed. #353 added
+# this map on 2026-06-14 when ``include_router(prefix=...)`` still flattened
+# the prefix into each route's own path, so the keys were right; #391 adopted
+# FastAPI 0.137 the very next day, which mounts the router instead and moves
+# the prefix into ``root_path``. The REST half of the adoption signal
+# recorded nothing at all from then until this was re-keyed. It failed
+# silently rather than loudly because the MCP half kept working, so the
+# report stayed plausible instead of going empty — and because the test
+# covering this registered its route directly on the app rather than through
+# a prefixed router, which is the one shape that makes a prefixed key match.
+#
+# ``/keystones`` is served twice: under the canonical ``/api/v1`` mount and
+# under the permanent legacy brand-prefixed alias mount beside it in
+# ``app.py``. Since the label drops the prefix, those two now collapse to a
+# single entry per verb rather than the two they needed before. That is only
+# safe because both mounts carry the same capability; the two URL forms are
+# no longer distinguishable in this signal, which is fine for adoption
+# accounting and would not be if they ever diverge.
+#
 # Routes NOT in this map are simply not recorded as capability usage (admin,
 # list/registry, ingest pipeline, health, plugin bootstrap). Extend this when
 # a new capability-bearing route is added — it's the REST half of the
 # transport-agnostic taxonomy; the MCP half is automatic via call_tool.
+# ``tests/test_capability_usage.py`` asserts every key here matches a real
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
Legacy-name ratchet fixed. The map change itself was already net-negative on old-brand names (it deletes four), but two *comment* lines I added named the legacy alias path literally, and the ratchet flags new text carrying the name even when the file's count falls.

The ratchet offers two exemption markers and also says: *"if rewording the line would leave it correct, reword it rather than marking it."* Rewording does leave these correct — `app.py` is the source of truth for the mount paths, so the comment doesn't need to spell one out — so I reworded rather than claiming `legacy-name-floor`. The comment now says "the permanent legacy brand-prefixed alias mount beside it in `app.py`".

```
No new lines. 4 removed by this change (-4 net).
All 39 protected strings survive.
```

Both gates run locally now, plus ruff lint/format and the 28 tests in the two affected files.

Worth noting the earlier CI failures were partly an artifact: this branch has been auto-rebased onto a moving main three times, and one failing run was measuring my pre-reword commit while I was already fixing it. The run before that flagged 15 unrelated files purely because my base predated the brand sweep in #1054 — rebasing cleared those, they were never mine.

Rebased onto current main (now including #1052).

arkash20 (approved):

Eldad-Caura:
@claude re-review please — force-pushed since your last pass, which ran against `e5267ba4` rather than current HEAD `67733733`. The delta is comment rewording only (dropping a literal legacy alias path to satisfy the legacy-name ratchet); no logic, keys, or tests changed. All checks green now.

Issue #1062, opened 2026-08-29
fix(version): whitespace CAURA_VERSION shadows legacy fallback


Eldad-Caura:
## Problem

core-api/src/core_api/constants.py selects the version override with:

    os.environ.get("CAURA_VERSION") or os.environ.get("MEMCLAW_VERSION")

A whitespace-only CAURA_VERSION is truthy, so it prevents a valid MEMCLAW_VERSION from being read. The later strip check rejects the whitespace value and falls through to VERSION file or package metadata instead.

This contradicts the README promise that the first non-empty value wins.

## Reproducer

- Set CAURA_VERSION to three spaces.
- Set MEMCLAW_VERSION to 1.2.3-old.
- Call _resolve_version().

Expected: 1.2.3-old.

Actual: the legacy value is skipped and resolution falls through to the file or metadata chain.

## Coverage gap

tests/test_env_dual_read.py::TestVersionResolution::test_blank_new_name_does_not_shadow_the_old_one covers an empty string, but not whitespace-only input.

## Suggested fix

Normalize each candidate before selecting the first non-empty value, and add a regression case for whitespace-only CAURA_VERSION with a valid MEMCLAW_VERSION.

Tracked separately from the PR7 legacy-name prose and marker cleanup.

PR #1063, opened 2026-08-29
refactor(core): refresh stale legacy-name prose

Eldad-Caura:
## Summary

- use CAURA_VERSION in the explicit override test while preserving the whitespace-only MEMCLAW_VERSION compatibility test with its rule 3 marker
- update the unminted rollback schema example and stale integration database prose
- preserve the legacy Docker ARG and document why a same-line marker would change its default value

## Safety and reviewer checks

- The whitespace test still sets MEMCLAW_VERSION and now clears ambient CAURA_VERSION first, so it exercises the intended legacy branch.
- ARG MEMCLAW_VERSION= remains present; the VER fallback remains unchanged.
- The rollback discriminator is comment-only, has no writer or schema dispatch today, and keeps its column alignment. Phase 3 should still accept a memclaw-prefixed value defensively.
- Difference from the work order: the requested inline Docker marker is intentionally omitted. Docker treats a # that is not at the beginning of a line as instruction content, so adding the marker after ARG MEMCLAW_VERSION= would make the default non-empty and break fallback behavior. See the official Dockerfile reference: https://docs.docker.com/reference/dockerfile/#format
- Separate whitespace-resolution bug filed as #1062.

## Verification

- 58 targeted pytest tests passed
- ruff check core-api/src/ passed
- ruff format --check core-api/src/ passed: 202 files already formatted
- root test-file Ruff lint passed
- ratchet after PR8 rebase: 618 before, 612 after; no new lines; 6 removed. The expected seventh removal is the unsafe Docker marker described above. The original pre-PR8 measurement was 711 to 705, also minus 6.
- sentinel: All 39 protected strings survive.
- DCO sign-off present

### DIFF
--- a/core-api/Dockerfile
+++ b/core-api/Dockerfile
@@ -73,6 +73,8 @@ RUN if [ "$WITH_RERANK_LOCAL" = "1" ]; then \
 # to find (the endpoint silently served "dev"). Read by
 # core_api.constants._resolve_version().
 ARG CAURA_VERSION=
+# Keep the legacy build arg unmarked: Docker treats inline ``#`` text as its
+# default value, so the ratchet's same-line marker would break the fallback.
 ARG MEMCLAW_VERSION=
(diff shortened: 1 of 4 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #1065, opened 2026-08-29
fix(version): honor whitespace-empty overrides

### LINKED ISSUE #1062: fix(version): whitespace CAURA_VERSION shadows legacy fallback

Eldad-Caura:
## Summary

- strip each version environment candidate before choosing the first non-empty value
- preserve canonical CAURA_VERSION precedence when both values are meaningful
- cover both empty and whitespace-only canonical overrides with named parameterized cases

A whitespace-only CAURA_VERSION previously short-circuited the valid MEMCLAW_VERSION fallback, then fell through to the baked or package version.

Closes #1062

## Verification

- regression proved red before the fix: expected 1.2.3-old, received package version 2.28.0
- 22 focused version and status tests passed
- core-api mypy: 202 source files, no issues
- Ruff lint passed for core-api and the changed test file
- core-api formatter check passed: 202 files already formatted
- legacy-name ratchet: no new lines; 612 before and after
- sentinel: All 39 protected strings survive.
- focused /simplify review completed; no remaining findings

### DIFF
--- a/core-api/src/core_api/constants.py
+++ b/core-api/src/core_api/constants.py
@@ -74,7 +74,8 @@ def _resolve_version() -> str:
     Precedence (most to least authoritative):
 
     1. ``CAURA_VERSION`` env, or its legacy alias below — explicit deploy /
-       ad-hoc override. The new name wins when both are set.
+       ad-hoc override. The first non-empty value wins, with the new name
+       checked first.
     2. ``VERSION`` file baked into the image at build time from
        ``pyproject.toml`` (see ``core-api/Dockerfile``). Deterministic and
        independent of installed-package metadata — the prod Dockerfile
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #1066, opened 2026-08-29
fix(storage): require shared-secret authentication

Eldad-Caura:
## Summary

- stop publishing `core-storage-api` from the default Compose stack and bind the development-only port to loopback
- require a fail-closed `X-Storage-Secret` before storage routing and attach it at every core API and worker storage-call choke point
- preserve the one-command quick start with a persistent random secret generated in a private Docker volume, and document the deployment requirement

Non-Compose deployments must set the same non-empty `CORE_STORAGE_SHARED_SECRET` (or mounted `CORE_STORAGE_SHARED_SECRET_FILE`) on storage and every direct caller before deploying this change.

## Validation

- regression test fails on the parent commit: unauthenticated `memories/bulk-get` returns 200 instead of the required 401
- root suite: 5,527 passed, 4 skipped, 1 xfailed
- core-storage-api: 173 passed; core-worker: 78 passed
- all CI Ruff check/format scopes and all six mypy scopes pass
- legacy-name ratchet and do-not-touch sentinel pass
- both Compose files render; rebuilt stack healthy; storage port 8002 has no host mapping; missing secret returns 401 and configured callers return 200
- `uv.lock` files unchanged

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -70,6 +70,10 @@ ADMIN_API_KEY=
 # Optional: when set, ALL non-admin requests must include this key via X-API-Key.
 # Useful when exposing the API to a network (not just localhost).
 CAURA_API_KEY=
+# Internal core-api → core-storage-api credential. Docker Compose generates a
+# random value in a private volume; non-Compose deployments must set the same
+# non-empty value on every storage caller and storage instance.
+# CORE_STORAGE_SHARED_SECRET=
 
 # -- Embedding provider -----------------------------------------------------
 # Options: openai | local | fake
--- a/README.md
+++ b/README.md
@@ -173,25 +173,38 @@ Anthropic, Gemini, and OpenRouter don't offer embedding APIs here — pair them 
 docker compose up -d
 ```
 
-By default this **pulls the multi-arch images from `ghcr.io`** (`linux/amd64` + `linux/arm64`) on first run — takes ~30 seconds. Subsequent `up` commands re-use the cached image (no registry round-trip, works offline). To pin a specific version, set `CAURA_VERSION=v1.2.3` in your `.env`. To build from local source instead (e.g. when iterating on a fork), run `docker compose up --build --no-pull`.
+> **Security requirement:** port 8002 belongs to the internal storage data
+> plane and must never be exposed to the host or any untrusted network. The
+> shipped Compose file does not publish it; `core-api` reaches it only across
+> the private Compose network and authenticates with a random, per-installation
+> secret generated into a private Docker volume.
+> Do not add an `8002:8002` port mapping. If local debugging absolutely requires
+> host access, bind only `127.0.0.1:8002:8002` and keep the
+> `CORE_STORAGE_SHARED_SECRET` check enabled. Only exact `GET /healthz` and
+> `GET /readyz` probes are public; every storage data route requires the secret.
+
+By default this **pulls the multi-arch images from `ghcr.io`** (`linux/amd64` + `linux/arm64`) on first run — takes ~30 seconds. Subsequent `up` commands re-use the cached image (no registry round-trip, works offline). To pin a specific version, set `CAURA_VERSION=v1.2.3` in your `.env`. To build from local source instead (e.g. when iterating on a fork), run `docker compose up --build --pull never`.
 
 To upgrade to a newer image at the same tag (e.g. `:latest` after we cut a new release), run `docker compose pull && docker compose up -d`. Without an explicit `pull`, the local cache wins — there's no silent version drift.
 
 > **Offline / air-gapped operation**: depending on whether the image is already cached locally:
-> - **Image cached, no network**: `docker compose up -d` works as-is — `pull_policy: missing` doesn't try to pull when the image is present. Use `docker compose up --no-pull` if you want to be explicit.
-> - **No local image, no network**: `docker compose up --build --no-pull` (build from source, don't try to pull).
+> - **Image cached, no network**: `docker compose up -d` works as-is — `pull_policy: missing` doesn't try to pull when the image is present. Use `docker compose up -d --pull never` if you want to be explicit.
+> - **No local image, no network**: `docker compose up --build --pull never` (build from source, don't try to pull).
 > - **Strict no-network guarantee** (e.g. an air-gapped pipeline that should never reach `ghcr.io`): drop a `docker-compose.override.yml` setting `pull_policy: never` for both services — Compose then fails fast if the image is absent rather than attempting a pull.
 
 | Service | URL |
 |---|---|
 | Core API (REST + MCP) | http://localhost:8000 |
-| Core Storage API | http://localhost:8002 |
 | PostgreSQL (pgvector) | localhost:5432 |
 | Redis | localhost:6379 |
 
+`core-storage-api` is intentionally reachable only by other services on the
+Compose network; it has no host URL.
+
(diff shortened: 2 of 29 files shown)


Eldad-Caura:
Addressed both review findings in 1cca63dc:

- Kept exact GET /healthz and GET /readyz public for external probes; trailing-slash, HEAD, POST, prefixed health, and every data-plane request still require X-Storage-Secret.
- Split the root ASGI fixtures: the transport injected into CoreStorageClient has no default secret, so the broad suite now exercises production header injection; direct storage-router tests use a separate authenticated raw client.

Validation on the rebased commit: 5,528 root tests, 179 storage tests, 78 worker tests, all six mypy scopes, all CI Ruff check/format scopes, Compose rendering, legacy-name ratchet, do-not-touch sentinel, and an isolated rebuilt Docker stack. The exact parent accepts the unauthenticated bulk-get with 200; this commit returns 401.

Eldad-Caura:
@claude

Eldad-Caura:
Addressed the second review in 0e0363ff:

- Storage-secret rejections now carry a stable, distinct detail. Core API retains its cached Cloud Run token, and the worker neither evicts nor retries for that response; generic identity-layer 401 behavior remains unchanged. New regression tests pin both branches and fail on the prior PR commit.
- Replaced repeated Settings construction with core_worker.config’s existing module singleton. I did not add function-local imports: the repository’s active import rule requires imports at module scope, and core_worker.app loads configuration immediately during startup, so the reported health-path lazy boundary does not exist at HEAD.

Validation after the fixes and final rebase: 5,529 root tests, 186 storage tests, 79 worker tests, all six mypy scopes, exact CI Ruff check and format scopes, Compose rendering, legacy-name ratchet, and do-not-touch sentinel. The identity-token regression tests were also run against the prior PR commit and failed on the old eviction/retry behavior.

Eldad-Caura:
@claude

Eldad-Caura:
Addressed both low-severity findings in ab201a05:

- Moved the exact storage-secret rejection classifier into common/storage_auth.py and switched both callers to it. Behavior is unchanged, and every runtime consumer of that shared module already declares httpx.
- Simplified both Compose storage healthchecks to the public headerless GET /healthz probe. Secret initialization still gates service startup, the secret remains mounted read-only for protected routes, and old-image version skew remains compatible.

Post-change validation: 26 root client/config tests, 19 storage auth/config tests, 16 worker auth/config tests, all six mypy scopes, exact CI Ruff check/format scopes, both Compose renders, legacy-name ratchet, and do-not-touch sentinel. The focused simplify review was clean across callers, security, and deployment behavior.

Eldad-Caura:
@claude

Eldad-Caura:
Addressed the offline-build finding in 84ced573:

- storage-secret-init now has the same local build fallback as core-storage-api.
- Corrected the two remaining README examples from the unsupported --no-pull spelling to --pull never.
- Verified with an isolated real run: Compose built and executed storage-secret-init successfully with --build --pull never, and the one-shot exited 0. Both Compose files render, the ratchet is clean, and all 39 protected strings survive.

I kept the three settings validators local. Although their current bodies match, each is a Pydantic decorator bound to a separate service Settings model; introducing a shared mixin/helper here would add abstraction to the critical fix without changing the security or deployment behavior.

Eldad-Caura:
@claude

Eldad-Caura:
Addressed the CORS-ordering finding in f9bcfb9f:

- Restored CORS as the outermost middleware, with shared-secret authentication immediately inside it and the reader-role filter inside authentication. Configured-origin preflight is handled by CORS, while every actual data request still authenticates before the role filter or a route handler.
- Added regressions for preflight and for the full CORS → auth → reader stack: missing-secret PATCH is a CORS-decorated 401; the same authenticated reader request reaches the filter and is a CORS-decorated 405. Both new tests fail against the pre-fix PR commit and pass now.

Validation: all 187 storage tests pass; the 12 focused auth tests, storage Ruff check/format, storage mypy, diff check, and focused simplify re-review are clean. The configured allowed origin remains unchanged; this does not widen the internal storage data plane to frontend callers.

Eldad-Caura:
@claude

Eldad-Caura:
Addressed the missing-secret observability finding in 67be0d26:

- Storage keeps GET /healthz as liveness, but GET /readyz now returns a generic 503 when its resolved credential is empty; startup logs the exact condition without exposing secret material.
- Both shipped Compose files now use /readyz for the service_healthy dependency gate, so core-api is not released behind unusable storage.
- Core-worker now logs and fails startup before provider initialization, consumer registration, or bus.start when its credential is empty, preventing nack/redelivery/DLQ churn. Core-api logs the missing caller credential; its existing authenticated dependency probe catches both empty and mismatched caller/storage values and returns 503. Storage itself cannot infer a non-empty remote mismatch.

Negative controls: the new storage readiness test returns 200 on the prior code instead of expected 503; the new worker test reports DID NOT RAISE and shows bus startup was previously allowed. Current validation: 189 storage tests, 80 worker tests, 20 core health/startup tests, affected Ruff and mypy scopes, both Compose renders, ratchet, sentinel, diff check, and focused simplify re-review are clean. Managed Cloud Run infrastructure should point an HTTP startup/readiness probe at /readyz; that probe wiring is outside this OSS repository.

Eldad-Caura:
@claude

Eldad-Caura:
Addressed the production startup finding in 96db90fc:

- moved the required storage-secret guard into `_validate_production_settings`, so production now refuses to boot with an empty or whitespace-only `CORE_STORAGE_SHARED_SECRET`
- preserved the existing log-only behavior for development and sandbox
- added regression coverage for empty/space/tab secrets in production and an explicit empty-secret non-production case

Verification: 5,532 passed, 4 skipped, 1 xfailed; the post-rebase focused guard suite passes 15/15; CI-scope Ruff check/format and core-api mypy are clean.

Eldad-Caura:
@claude

Eldad-Caura:
Addressed the all-environment startup consistency finding in 3731213c:

- renamed the guard to `_validate_startup_settings` and made missing or whitespace-only storage credentials fatal in development, sandbox, and production
- moved validation ahead of Sentry/provider/MCP initialization, matching core-worker’s fail-fast behavior
- retained production-only enforcement for the other perimeter settings
- added regression coverage across all three supported environments; the same assertions fail against the prior head with `DID NOT RAISE`
- rebased onto #1074 and updated the auth proof to use a valid tenant-scoped `bulk-get`, preserving the isolated 200-on-parent/401-on-PR negative control

Verification after rebase: root 5,534 passed (4 skipped, 1 xfailed), storage 194 passed, focused startup 17/17, focused auth 13/13, core-api/storage mypy clean, and discovery-based CI-scope Ruff check/format clean.

Eldad-Caura:
@claude

Eldad-Caura:
Fixed the current-head Legacy-name ratchet failure in 1af3070d. Ruff had collapsed an existing two-line call and thereby created novel line text containing the pre-rename API-key field; the test now preserves the exact already-ratcheted legacy-key line while still calling `_validate_startup_settings`. No runtime behavior changed. Post-rebase validation: legacy-name ratchet clean, Ruff check/format clean, startup guards 17/17, one signed commit.

Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #1067, opened 2026-08-29
fix(fleet): require an explicit tenant scope on the command-status update

Eldad-Caura:
## What

`PATCH /fleet/commands/{command_id}/status` took `tenant_id` as an optional body field and passed it to `fleet_update_command_result`, which defaulted it to `None` — and `None` meant *build the UPDATE with no tenant predicate*:

```python
stmt = sql_update(FleetCommand).where(FleetCommand.id == command_id)
if tenant_id is not None:                      # omitted -> no scope at all
    stmt = stmt.where(FleetCommand.tenant_id == tenant_id)
```

The cross-tenant write was what a caller got by **leaving the field out**, addressing any tenant's command by UUID.

## Why it matters

This is the shape of GHSA-xw4x-jwf5-8m9h one endpoint over. That advisory was a fleet-command route that validated the tenant the caller *named* and never checked it against the row it *addressed*. Its fix (`6db0ae57`) landed on the create path, which now checks the node/tenant pair. The status path kept the same hazard in a quieter form: the guard existed, but skipping it was the default. The method's own docstring named the vulnerability class — *"keying on `command_id` alone let any authenticated tenant complete another tenant's command by UUID (cross-tenant BOLA)"* — directly above a parameter that defaulted to exactly that.

Reachability, stated honestly: through core-api this was **not** exploitable by an ordinary tenant. `command_result` sends `tenant_id: auth.tenant_id`, which is `None` only for admin credentials. What this fixes is the *default*: storage authenticates nothing (GHSA-wgvw-28pq-jc36) and trusts the tenant its caller names, so the safe behaviour depended on every present and future call site remembering an optional kwarg. That is defence-in-depth, not an emergency — but it is the exact defect class both advisories came from.

## How

Two changes, one per layer.

**Wire contract** — absent and null are no longer the same request. Absent is a 422; `null` is the admin/unscoped opt-out. This costs the only caller nothing: core-api already sends `{"tenant_id": auth.tenant_id}` on every call, and `httpx`'s `json=` serializes the admin case as an explicit `null`, so the request that was always correct stays correct.

**Type** — `fleet_update_command_result` now takes `tenant_id: str | Unscoped` with no default. Widening the scope has to be spelled `tenant_id=UNSCOPED`, which is greppable and reviewable in a way `=None` never was. Both failure modes now fail loudly, verified rather than asserted:

```
error: Argument "tenant_id" ... has incompatible type "None"; expected "str | Unscoped"  [arg-type]
error: Missing named argument "tenant_id" for "fleet_update_command_result"  [call-arg]
```

A non-string `tenant_id` is also rejected at the edge instead of reaching asyncpg as a bind parameter, where it surfaced as a 500 for what is squarely a malformed request.

## Tests

Seven new tests. **Three fail on the parent commit** — omitted `tenant_id` rejected, the two-tenant attack blocked, non-string rejected. The other four pin behaviour that must *not* change: a wrong tenant matches no row, the owning tenant updates, `null` is still the admin opt-out, and the sentinel reads as its own name in a log line.

Verified locally against pgvector:pg16 — 175 core-storage-api tests, 5541 root tests, `ruff check` / `ruff format --check` at CI's scopes, and `mypy src/` all clean.

## Scope

Deliberately narrow: this is one endpoint and one method. The broader question — that the tenancy invariant is enforced by convention across 187 storage routes and 192 SQL-layer methods with nothing checking it — is a follow-up PR that adds a CI gate and a shrinking allowlist. This lands first so that gate starts with one fewer exception.

Does **not** touch `docker-compose.yml`, storage auth middleware, or the README's Self-Hosted section — that is GHSA-wgvw-28pq-jc36's territory and is being handled separately.

### DIFF
--- a/core-storage-api/src/core_storage_api/routers/fleet.py
+++ b/core-storage-api/src/core_storage_api/routers/fleet.py
@@ -8,7 +8,7 @@ from uuid import UUID
 from fastapi import APIRouter, HTTPException, Request
 
 from core_storage_api.schemas import FLEET_COMMAND_FIELDS, FLEET_NODE_FIELDS, orm_to_dict
-from core_storage_api.services.postgres_service import PostgresService
+from core_storage_api.services.postgres_service import UNSCOPED, PostgresService
 
 router = APIRouter(prefix="/fleet", tags=["Fleet"])
 _svc = PostgresService()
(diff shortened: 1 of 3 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #1068, opened 2026-08-29
chore: annotate permanent legacy names outside plugin/

Eldad-Caura:
Marker sweep B — the last of the annotation phase. **Annotation only:** every changed line is its previous text plus a trailing marker. Proven mechanically, not asserted: 26 lines removed, 26 added, and each added line starts with the exact bytes of the line it replaced.

`612 → 586` lines across 133 files. That is 26 lines, matching the 26 markers added exactly.

## Marker kind follows the ratchet's own split

Not defaulted to one marker. **15 `legacy-name-ok`** where the line *bears* an alias — the README bridging lines, the client import/class aliases, the published tag pattern, the MCP server's own declaration that `memclaw_*` tool names are permanent. **11 `legacy-name-floor`** where it merely *names* a permanent thing — the removed share/unshare tools, the ghcr repository name, the PyPI keyword, the on-disk skill slug, the `HEARTBEAT.md` content predicate, the uninstall step, and one dated wet-test observation.

## The line this sweep exists for

`publish-python-client.yml:18` — the release tag pattern `memclaw-client-v*` is **sentinel-pinned** (`breaks` = "the release tag pattern stops triggering a publish") yet carried no marker, while its npm twin at `publish-npm-client.yml:33` did. A frozen contract read as remaining sweep work. It now carries the twin's exact marker text.

Verified the marker is a YAML comment and not part of the value:

```
trigger tags: ['memclaw-client-v*']
marker leaked into the tag value? False
```

## Twelve lines are deliberately left counted — please read this part

They sit **inside multi-line Python docstrings**, where a trailing `#` is string content rather than a comment. Marking them would exempt the line in the gate *while corrupting the documentation*. Demonstrated rather than assumed — appending a marker to `tests/test_enrichment_inline_timeout.py:4`:

```
(a) gate:    kind=legacy-name-floor      <-- goes quiet
(b) __doc__: "...ParallelEmbedEnrich had a  # legacy-name-floor: dated incident record"
```

Silent, with no red check anywhere. Eight of the twelve are **dated incident records**, whose host names and log tokens this sweep is explicitly required to keep verbatim — so injecting marker text mid-sentence is the precise harm the rule exists to prevent.

| file | line | why bare |
|---|---|---|
| `core-api/src/core_api/routes/plugin.py` | 736 | inside docstring; also a present-tense host claim I could not substantiate as floor |
| `tests/test_api_documents_skills.py` | 4 | inside module docstring |
| `tests/test_storage_client_retry.py` | 3, 87 | inside docstring — dated `memclaw.dev` incident |
| `tests/test_a4_13_path_c_retraction.py` | 265 | inside docstring — dated `memclaw.net` incident |
| `tests/test_enrichment_inline_timeout.py` | 4 | inside module docstring — dated incident |
| `tests/test_query_embedding_stampede_guard.py` | 3 | inside module docstring — dated wet-test |
| `tests/test_caura_000_graph_frontier_cap.py` | 7 | inside module docstring — pins a log token |
| `clients/python/src/caura_client/__init__.py` | 1, 5, 6 | inside module docstring |
| `clients/python/src/memclaw_client/__init__.py` | 3 | inside module docstring |

Two more stay bare for reasons already established in the spec, both confirmed here: `scripts/f3_wet_test_matrix.sh:102` ends in a backslash continuation where a trailing comment is a syntax error that would break auth refresh, and `.github/workflows/publish-docker.yml:149` sits inside a `labels: |` literal block where a comment becomes an OCI label value on every published image. Confirmed no marker leaked into that block.

`scripts/gateway_integration_test.py:724` is also untouched — it is the `HEARTBEAT.md` content predicate frozen by the earlier scripts PR. Its docstring (`:716`) and check label (`:728`) are annotated; the predicate itself stays as it is.

## The formatter cannot strand these markers

Five of the six `core-api` lines exceed that tree's 110-column limit once annotated, which is exactly the condition that stranded a marker in a sibling repo. Checked rather than assumed:

```
ruff format --check core-api/src/   ->  202 files already formatted
```

The long lines are unsplittable string literals and `E501` is ignored in that tree, so the formatter leaves them alone and the markers stay on their literals.

## Checks

| check | result |
|---|---|
| `legacy_name_ratchet.py` | exit 0 |
| `do_not_touch_sentinel.py` | exit 0 — all **39** protected strings survive |
| `ruff check` (CI scopes: core-api, core-storage-api, common, core-operations, core-worker) | pass |
| `ruff format --check core-api/src/` | pass — 202 files |
| `actionlint` on both changed workflows | clean, exit 0 |
| workflows parse; trigger + `images` values unchanged | verified |
| gate tests | **174 passed** |
| python-client tests | **142 passed** |

One anchor had drifted since the spec was written — `contradiction_detector.py` 1901 → 1920. Every other anchor held byte-exact. All were re-derived by content rather than trusted by line number.


Eldad-Caura:
@claude

Applied the `docs/api-surfaces.md` finding. Verified it byte-exact before changing anything rather than taking the description on trust — and the finding was correct:

```
line 34 (mine, before fix)  after last pipe: ' <!-- legacy-name-floor: ... -->'
line 35 (pre-existing)      after last pipe: ''
```

So the file's own convention really is inline-inside-the-cell, and my placement was the odd one out. The marker now sits immediately after the `` `memclaw_share_skill`/`memclaw_unshare_skill` `` literal, matching how the Keystones row on line 35 places its `<!-- legacy-name-ok: taught as legacy alias -->` right after `` `/memclaw/keystones` ``.

Checked after the move:

| check | result |
|---|---|
| line 34 still exempt | yes — `kind=legacy-name-floor` |
| content after final pipe | none, same as line 35 |
| table column count | 30 rows, all 3 columns, 0 mismatched |
| ratchet | 586 lines / 133 files — unchanged |
| sentinel | 39/39 |
| gate tests | 174 passed |

No other change; still annotation-only.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #1069, opened 2026-08-29
feat(storage): gate the tenancy invariant in CI

Eldad-Caura:
> **This PR fixes no vulnerability.** It adds a CI gate that makes the tenancy invariant checkable and stops new violations of the shape. The 120 paths currently on its allowlist — 45 of them in the `id-addressed` backlog — remain exactly as they are today. Do not read "gate added" as "advisories closed." The fixes are going out one endpoint at a time in their own PRs (#1067, #1074, and the `GET /memories/{memory_id}` one next), and each shrinks this list.

## Why

Two advisories this month were the same defect wearing different clothes. **GHSA-xw4x-jwf5-8m9h** validated the tenant a request *named* and never checked it against the row it *addressed*. **GHSA-wgvw-28pq-jc36**'s exploit primitive was smaller still: `POST /memories/bulk-get` made the `tenant_id` filter optional, so omitting it returned any row by id across every tenant.

Neither is a property of one endpoint. The invariant — *every statement that touches a tenant-scoped table is bound to a tenant* — was held up across 189 routes and 192 service methods by convention alone. Nothing read it, so nothing noticed when it lapsed, and a lapse is not a syntax error, not a type error, and not a test failure. It is a row belonging to somebody else.

## What this checks

`scripts/tenant_scope_gate.py` does **not** try to prove that any particular statement is correctly scoped. That is a semantic question, a gate that guesses at semantics produces false positives, and a security gate with false positives gets switched off. It checks three things that are exactly decidable:

| | |
|---|---|
| **Completeness** | Routes come from the **live app**, so one registered in a loop still counts; the walk is cross-checked against `app.openapi()` and raises on disagreement, so a FastAPI internals change breaks the build instead of silently shrinking the surface. Service methods come from `inspect.signature` via `isroutine`, so classmethods cannot hide. |
| **Justification** | Anything without a binding tenant scope must appear in `core-storage-api/tenant_scope_allowlist.json` under one of six categories. |
| **Direction** | `--base origin/main` fails a PR that grows the list — or that lets an entry already on it get *weaker* (`OPTIONAL → NONE`). |

Categories rather than a bespoke sentence per row, because there are 120 rows. `legacy_name_ratchet.py` already carries two markers instead of one for exactly this reason — *"a reviewer facing eleven undifferentiated entries stops reading."*

## Current state (green on trunk)

```
tenant-scope gate: 189 routes + 192 service methods, 261 bound to a tenant, 120 allowlisted.
    6  admin-unscoped
   22  cross-tenant-by-design
   45  id-addressed  <- backlog
    7  no-tenant-data
   40  opaque-body-write
```

The categories are not equally benign, and separating them is the point. Four describe paths correct as they stand. **`id-addressed` is the backlog** — GHSA-wgvw's shape, where knowing a UUID is not the same as being entitled to the row behind it.

**The list has already shrunk three times**, and not by anyone remembering to: when #1067 and #1074 merged, the rebase left rows describing code that no longer matched them, and CI failed here until they were removed. A fixed path cannot quietly stay listed.

## On `readable_tenant_ids`

The brief for this work described it as default-unsafe — 11 parameters where *omitting the argument widens the read to every tenant*. **That is inverted, and I checked all eleven sites.** Every one reads `if readable_tenant_ids: IN(...) else: == tenant_id`, so omitting it **narrows to the home tenant**. It is a widening grant you must pass, not a guard you can forget; making it mandatory would compel the dangerous direction.

What is gated instead is the property that makes omitting it safe: a method taking the grant must also take a binding scope to fall back to. The genuinely default-unsafe parameters were the three `tenant_id: str | None = None` ones, where `None` drops the predicate — the worst was fixed in #1067.

## Tests

**49.** Beyond the accounting, they inject the faults the gate must catch: an unscoped method attached to the real `PostgresService` (end to end through enumeration), a *scoped* one that must **not** need a row, a classmethod, a truncated route walk, a grown allowlist, a weakened entry, an unresolvable `--base`, a lookalike `_require_if_present` helper, and every classifier idiom in both polarities.

Eight review rounds have run against this. Each found something real in the classifier, and the tests are largely the residue of that — including two false-positive directions I introduced while fixing earlier findings and caught by re-measuring trunk rather than by a test.

## Scope

Adds files and one CI step; changes no runtime behaviour. Does not touch `docker-compose.yml`, storage auth middleware, or the README's Self-Hosted section — that is GHSA-wgvw-28pq-jc36's reachability half, handled separately.


Eldad-Caura:
## Review round 1 — one applied, one declined

### ✅ Applied (Medium): allowlist compared only identifiers

Confirmed, and it was worse than reported. Reproduced against the pushed code:

```python
recorded = {"method:memory_admin_list": {"verdict": "OPTIONAL", "category": "admin-unscoped"}}
live     = [Entry("method", "memory_admin_list", "NONE", "no binding tenant parameter")]
check(live, [], recorded)  # -> []      <- green on a silent weakening
```

The review proposed fixing `check()`. That alone would not have closed it. `check()` compares the tree against the *committed* allowlist, so it only sees a regression while the file still remembers the old verdict — an author who weakens a path and then re-runs `--write` moves the record along with the code, and both `check()` and the id-set ratchet go quiet. The base tree is the only copy that still remembers.

So both halves:

- **`check()`** — for every id in both sets, the recorded `verdict` must match the live one. Catches a stale allowlist.
- **`ratchet()`** — for every id in both sets, the verdict may not *weaken* against `--base` (`REQUIRED` > `OPTIONAL` > `NONE`). Catches the re-written allowlist that hides a regression.

Four new tests, `30` total: the OPTIONAL→NONE regression, an unchanged verdict staying quiet, the ratchet failing on a weakening, and strengthening being allowed. Writing them caught a false positive in my first attempt — a baseline row with no recorded `verdict` was being treated as maximally strong and reported as a regression. A missing verdict supports no comparison, so it now makes none, rather than guessing a direction.

### ❌ Declined (High): "GHSA-wgvw-28pq-jc36 remains unpatched"

The premise is correct and the conclusion is right — `POST /memories/bulk-get` is still an optional tenant filter over caller-supplied ids, and this PR catalogs it rather than fixing it. That is deliberate, not an oversight, and the finding is scoped at the wrong PR:

1. **The gate is not the remediation and does not claim to be.** `id-addressed` is documented in the script as *"THIS IS THE BACKLOG… Shrink this category; do not add to it"*, and both entries carry notes naming the advisory. Nothing here reports the vulnerability as resolved.
2. **Landing a gate together with its backlog is how gates get reverted.** That is this repo's own stated position, in `scripts/mypy.ini`: *"These 33 findings are what a tree that has never been type-checked looks like… the gate is landing without them deliberately."* 121 entries is a 47-item security backlog; folding it into the PR that introduces the mechanism makes both unreviewable.
3. **It is a wire-contract change.** Making `tenant_id` required on `bulk-get` changes an endpoint whose docstring deliberately mirrors the `GET /memories/{id}` contract. It needs a caller audit, not a one-line edit.

The precedent is already in this PR's own history: `fleet_update_command_result` was the same shape, and it went out as its own PR (#1067) *before* this one — which is why its two rows are absent from this allowlist rather than being the first thing the gate excuses.

**I am opening the `bulk-get` fix as a separate PR next**, same pattern. When it lands, the two entries disappear from this allowlist the way the fleet ones did — a shrink, which the ratchet permits and reports.

---

Rebased onto `6744a801`. Verified after the rebase: 30 gate tests, 175 core-storage-api tests, 5571 root tests, `mypy --config-file scripts/mypy.ini scripts/`, `ruff check` at CI's scopes. `uv.lock` unchanged. The gate step ran on the previous CI run and reported the same numbers as local — `189 routes + 192 service methods, 260 bound to a tenant, 121 allowlisted`.

Eldad-Caura:
@claude

Eldad-Caura:
## Review round 2 — both applied

Both findings were real. Reproduced each against the pushed code before fixing.

### Subscript/`.get()` ignored receiver and context

```python
async def h(request):
    response = {}
    response["tenant_id"] = "audit"      # a WRITE to an unrelated dict
    return await svc.by_ids(body["ids"])  # no tenant read at all
# classified: REQUIRED
```

This is the direction the module docstring calls the dangerous one, and it was live. A false `NONE` costs one allowlist line somebody deletes; a false `REQUIRED` removes the route from the list altogether, so a real gap is never shown to anyone.

Fixed as suggested — subscripts must be `ast.Load`, and reads are anchored to the request body — with one addition the review didn't ask for: **the anchoring is transitive.** `recall_log_write`'s route does `event = body.get("event")` then `_require(event, "tenant_id")`, and a name-exact anchor would have scored that unscoped. Names derived from the body, iterated to a fixed point, count as the request.

**A note on how this nearly went wrong**, since it bears on how much to trust the seeded list. My first version of the anchoring matched only `ast.Assign` — but `body: dict = await request.json()` is an `AnnAssign`, and **99 of the 107 body bindings in the routers use the annotated form**. The tightening looked correct, passed every synthetic test I had, and moved the allowlist from 121 entries to 167: 46 correctly-scoped routes reclassified as unscoped. Caught by re-running the gate against trunk and asking why the number moved, not by a test. There is now a test pinning the annotated form, and one pinning the nested-object case.

After the fix the trunk numbers are **unchanged** — `260 bound to a tenant, 121 allowlisted`, same distribution — so no seeded entry depended on the false-positive path.

### Classmethods silently excluded from enumeration

Confirmed: `inspect.isfunction` is `False` for a classmethod reached through the class (it is a bound method), so one would have been enumerated never, allowlisted never, ratcheted never. Staticmethods were already caught — `isfunction` is `True` for those on modern Python — but `isroutine` covers all three and is the honest predicate.

You were also right that `test_every_public_service_method_is_enumerated` could not have caught this: it built its expected set with the same predicate, so it reproduced the bug and agreed with itself. It now parses the class body with `ast` — an independent oracle — and there are explicit classmethod and staticmethod cases. I verified both new tests fail with the old predicate restored, rather than assuming.

---

**36 tests** (was 30). Verified after rebasing onto `6d13b630`: 36 gate tests, 175 core-storage-api tests, 5577 root tests, `mypy --config-file scripts/mypy.ini scripts/`, `ruff check`. `uv.lock` unchanged.

The round-1 High (`bulk-get` still unpatched) is being addressed in its own PR, as described in the previous comment — it is written and tested, and I'll link it here when it opens.

Eldad-Caura:
@claude

Eldad-Caura:
## Review round 3 — both applied

Both real, both reproduced against the pushed code first.

### High — ratchet silently no-opped on any git failure

```python
gate.ratchet("no-such-ref-xyz", ALLOWLIST_PATH, {...})  # -> []
```

Empty list means "nothing grew". A bad `--base`, or the `git fetch` step above it failing, produced exactly that — so the only mechanism holding the allowlist flat could be skipped on a green build with nothing printed. That is worse than the gate not existing, because the green is read as evidence.

Fixed by separating the two cases the review identified. `git rev-parse --verify <base>^{commit}` first: if the ref does not resolve, that is a hard error — *"This is not 'nothing grew' — it is 'nobody looked'"* — naming the fetch step as the thing to check. Only then `git show`, and only a stderr that actually says the path is absent counts as "this commit introduces the allowlist". Any other git failure is reported rather than swallowed.

### Medium — inline-guard heuristic credited a raise that a missing tenant never reaches

```python
tenant_id = body.get("tenant_id")
if tenant_id:
    if something_else:
        raise HTTPException(422)   # unreachable when tenant_id is absent
# classified: REQUIRED
```

Searching `ast.walk(sub)` finds a raise anywhere beneath the branch. The raise now has to be a statement of the branch itself — unconditional given the test, which is the weakest claim that is still a claim. The example above is `OPTIONAL` again, and the real `purge_tenant_data` idiom still reads `REQUIRED`.

**Trunk is unchanged after both fixes** — `260 bound to a tenant, 121 allowlisted`, same distribution — so no seeded entry was resting on either loose path. Worth stating explicitly, since a tightening that moved the numbers would mean some of the 121 had been wrong.

---

**38 tests** (was 36). Two added: a garbage `--base` must produce an explicit error, and a raise nested under a second condition must not count as a guard.

Rebased onto `13b320da` — the branch had been rebased upstream while I worked, so I re-synced, confirmed my local commit was a strict superset of the remote one (same author, +93/−12, my round-3 work only), and re-ran everything after: 38 gate tests, 5579 root tests, `mypy --config-file scripts/mypy.ini scripts/`, `ruff check`. `uv.lock` unchanged.

Eldad-Caura:
@claude

Eldad-Caura:
## Review round 4 — applied

Fair, and it is my own regression from round 3. Fixing the silent-no-op finding, I replaced a broad `except CalledProcessError` with a check for `"does not exist"` in `git show`'s stderr — which is English prose git is free to reword and which is translated under a non-C locale. That is the same unstable-signal problem `_ref_exists` was added to avoid, reintroduced a few lines below it.

Replaced with `git cat-file -e <base>:<path>`, as suggested: the question now has a numeric answer, so there is no text to match and nothing to pin `LC_ALL` around. The three cases are now decided independently:

| condition | check | outcome |
|---|---|---|
| base does not resolve | `git rev-parse --verify` | hard error — "nobody looked" |
| path not in base | `git cat-file -e` | `[]` — this commit introduces the allowlist |
| read fails anyway | `git show` exit code | hard error — corrupt object, unreadable store |

Took the `cat-file` option over pinning the locale because pinning would have made the substring match reliable while leaving it a substring match; a git version that rewords the message would still break it.

**39 tests** (was 38), with a direct case for `_path_in_ref` in both directions. Worth noting that this path is not hypothetical on this PR: the allowlist does not exist on `origin/main` yet, so every CI run here takes the "path not in base" branch — it is the branch being exercised right now, not a future one.

Rebased onto `1304e3ee`. Verified after: 39 gate tests, 5580 root tests, `mypy --config-file scripts/mypy.ini scripts/`, `ruff check`. `uv.lock` unchanged.

Eldad-Caura:
@claude

Eldad-Caura:
## Review round 5 — applied, with the premise confirmed and the conclusion corrected

The finding is right about `git fetch`, and I checked rather than reasoning about it. From this branch's own CI log:

```
git fetch --no-tags --depth=1 origin main
From https://github.com/caura-ai/caura
 * branch            main       -> FETCH_HEAD
```

`FETCH_HEAD` and nothing else — no `-> origin/main`. So the fetch is indeed not what creates the ref.

The conclusion doesn't follow, though: **`origin/main` does resolve, and the gate has been comparing against a real ref.** That is verifiable now rather than assumed, because round 3 made an unresolvable base a hard error — run `33281681848` on commit `207e0747` carries that check and passed, printing the normal summary. The ref exists because `actions/checkout` leaves it behind, not because this step fetched it.

Which is still worth fixing, for the reason the finding gives even though its failure mode isn't the live one: "resolves because of how a different action happens to be configured" is not something a gate should rest on, and a checkout-action upgrade that narrows its refspec would turn this step red with a confusing message. Now:

```
git fetch --no-tags --depth=1 origin +main:refs/remotes/origin/main
```

An explicit destination makes the ref this step's own output. The `+` forces the update so a stale `origin/main` from checkout cannot win — worth having on its own, since a stale base is the quieter bug here: the ratchet would compare against an old main and report green.

I left the two pre-existing jobs that use the bare form alone (the broker gate and the legacy-name ratchet). They have the same dependence, and it is real, but it is not this PR's to change and they are not currently failing.

**Unchanged elsewhere**: 39 gate tests, gate green against `origin/main`, `actionlint` reports no new findings (the single SC2086 it flags is pre-existing on `ci.yml:99`), `uv.lock` unchanged.

Eldad-Caura:
@claude

Eldad-Caura:
## Review round 6 — applied

Correct. `startswith("_require")` reads as a naming convention but behaves as a promise about code nobody has written yet: `_require_if_present` or `_require_optional` — names that say plainly they do *not* raise — would have been trusted as proof of scoping, dropping the route off the allowlist and out of the ratchet. The invisible direction again.

Now an enumerated set:

```python
FAIL_CLOSED_GUARDS = frozenset({"_require", "_require_dict", "_require_number"})
```

All three, not just `_require`, because all three are real fail-closed guards in `_validation.py` — matching only `_require` as the prompt's first suggestion proposed would have quietly lost the other two the day someone used one with a tenant key. (Today only `_require` is called with one, at 23 sites; the other two are used for other fields.)

**The set is checked rather than trusted.** A hand-maintained list of names is itself a claim that can rot — rename a helper, or soften one to return instead of raise, and the claim stands over code that no longer backs it. `test_every_fail_closed_guard_exists_and_raises` parses `_validation.py` and asserts every name in the set is really defined there and really contains a `raise`. That is the same move the rest of the gate makes: derive the check from the code instead of asserting it in a comment.

**41 tests** (was 39). Two added — the guard-set integrity check above, and one asserting a `_require_if_present(body, "tenant_id")` call classifies as `NONE`. I verified the second fails with the prefix match restored, rather than assuming it would.

Trunk unchanged: `260 bound to a tenant, 121 allowlisted`, same distribution — so no seeded entry was resting on a lookalike name.

Verified after: 41 gate tests, 5582 root tests, `mypy --config-file scripts/mypy.ini scripts/`, `ruff check`. `uv.lock` unchanged.

---

The round-1 High is now closed by **#1074**, which makes `tenant_id` required on `bulk-get` and applies it in SQL. When that merges, `route:POST /api/v1/storage/memories/bulk-get` and `method:memory_get_memories_by_ids` come out of this allowlist — the gate fails on the stale rows until they do, which is the intended behaviour and the third shrink so far.

Eldad-Caura:
@claude

Eldad-Caura:
## Review round 7 — applied, and it found a real divergence in the routers

Both halves were live false `REQUIRED`s. Reproduced before fixing:

```python
tenant_id = body.get("tenant_id")
if tenant_id and other_flag:          # raises only when the tenant IS present
    raise HTTPException(400)
# was: REQUIRED

tenant_id = DEFAULT_TENANT            # never came from the request
if not tenant_id:
    raise HTTPException(400)
# was: REQUIRED
```

### One correction to the suggested fix

The prompt proposed rejecting compound `BoolOp` tests outright. **That would have misclassified eight correctly-guarded routes** — `if not isinstance(tenant_id, str) or not tenant_id: raise` is the dominant real idiom in these routers. Compoundness is not what separates a guard from a non-guard; **polarity** is, and the two operators differ in exactly the way that matters:

- `or` is true if any branch is → one negative check on the tenant carries the test → **credit**
- `and` needs all branches → a positive tenant branch is false precisely when the tenant is gone → **refuse**
- …except an `and` whose branches are *all* negative scope checks (`if not tenant_id and not readable_tenant_ids: raise`), which does fail closed against having no scope at all → **credit**

Plus the receiver anchoring the review asked for: a bare name counts only as a handler parameter (FastAPI fills those from path/query) or an alias bound from the body.

### What the tightening surfaced

Trunk moved 121 → 124, and each mover was worth reading rather than working around:

- **`POST /fleet/commands`** — `if node is None or node.tenant_id != body.get("tenant_id"): raise`. A genuine guard, and *the fix that closed GHSA-xw4x-jwf5-8m9h*. An absent tenant reads as `None`, which differs from the row's real tenant, so the raise is reached. Handling only `is None` / `== None` would have scored that endpoint unscoped. Now handles `!=` / `is not`.
- **`PATCH /memories/{memory_id}`** — reads its scope with `body.pop("tenant_id", None)` so the key cannot reach the column update. Removing the key afterwards does not make it less of a read; `pop` joins `get`.
- **`POST /memories/quality-metrics`** — **not** a classifier problem. It is genuinely the odd one out:

| route | guard |
|---|---|
| `stats-breakdown` | `if not tenant_id: raise` |
| `daily-durable-counts` | `if not tenant_id: raise` |
| `list` | `if not tenant_id: raise` |
| **`quality-metrics`** | `if not tenant_id and not body.get("readable_tenant_ids"): raise` |

Three sibling stats routes require a binding tenant; the fourth accepts a caller-supplied widening grant in its place — and its own docstring lists `tenant_id` as non-optional, so the code and the doc disagree. Since storage authenticates nothing and takes that grant verbatim from the body, this is the weakest of the four. It is now an allowlist row under `cross-tenant-by-design` with a note saying exactly that, rather than being silently credited as it was before this round. **Flagging it for a maintainer** — aligning it with its siblings is a behaviour change, so not something this PR should make.

Final: `259 bound to a tenant, 122 allowlisted` (was 260/121 — the one row is quality-metrics, newly visible).

**47 tests** (was 41). Six added, covering each idiom in both directions: the `and`-guard that fires only when present, an `and` with a non-scope branch, the compound `or` that must still count, the `!=` pair check, a same-named local from a config default, and the `pop` read.

Verified: 47 gate tests, 5588 root tests, `mypy --config-file scripts/mypy.ini scripts/`, `ruff check`. `uv.lock` unchanged.

Eldad-Caura:
@claude

Eldad-Caura:
## Review round 8 — Medium applied; High is partly overtaken by events

### ✅ Medium: alias tracking had no notion of where a binding was in effect

Real, and reproduced:

```python
value = compute_something()
if not value:                      # guarding something that is not the tenant
    raise HTTPException(400)
value = body.get("tenant_id")      # bound AFTER the guard
# was: REQUIRED
```

Python has no block scope, so a flat name→key map let a later binding retroactively credit an earlier guard. `aliases` is now a list of `(lineno, name, key)` and each guard resolves against the last binding before its own line; a rebinding to something that is not a scope read clears the name rather than leaving it credited.

The docstring says plainly what it still does not do: branches are ignored, so a name bound inside one arm of an `if` counts from that line on. That errs toward crediting, which is the wrong direction — but the alternative is real dataflow analysis, and every handler in this tree binds its scope once, straight-line, before using it. Written down rather than left implicit.

**49 tests** (was 47): the rebound-name case, and the ordinary bind-then-guard order that must keep working.

### ⚠️ High: the two entries it names as live are already gone

The review ran against `4b8d9383`. **#1074 merged in the meantime**, and both `route:POST /api/v1/storage/memories/bulk-get` and `method:memory_get_memories_by_ids` came off this allowlist as a result — that push is `a07e65ef`, before this one. The allowlist is now **120 entries, 45 `id-addressed`** (was 122/47).

Worth noting how they came off, because it is the mechanism working rather than me remembering: after #1074 merged, the rebase left the allowlist describing code that no longer matched it, and CI failed on this PR with

```
::error::method:memory_get_memories_by_ids is in the allowlist but no longer needs to be
::error::route:POST /api/v1/storage/memories/bulk-get is in the allowlist but no longer needs to be
```

A fixed path cannot quietly stay on the list.

The remaining named entries — `DELETE /memories/{memory_id}`, `method:memory_get_by_id`, `method:memory_soft_delete` — are still real, and I am not disputing them. `memory_get_by_id` is being removed right now in the next PR (the `GET /memories/{memory_id}` fix, which deletes the unscoped SQL method and migrates all eleven call sites off the unscoped client method).

On the ask for a tracking mechanism: the tracking is the sequence, not a field. #1067, #1074, and the one in flight each closed one and shrank this list — three shrinks, verifiable in the ratchet's own output. A per-entry owner-and-expiry would be a second list to maintain, and the failure mode of expiry dates is that they pass. What the reviewer is right about is the PR description, which I have updated to say in its own words that **this PR fixes nothing and is not a remediation** — the numbers in it were stale anyway.

Verified after: 49 gate tests, 5590 root tests, `mypy --config-file scripts/mypy.ini scripts/`, `ruff check`. `uv.lock` unchanged.

Eldad-Caura:
@claude

Eldad-Caura:
@claude

Eldad-Caura:
## Rounds 9 and 10 — three applied

### Round 9a: a scope literal counted wherever it appeared in a guard test

```python
config = await load_config()
if not config["tenant_id"]:      # says something about a config file
    raise HTTPException(400)
# was: REQUIRED
```

`_scope_keys_mentioned` matched the bare string anywhere in the subtree. It now goes through `_tenant_key_read`, which requires the receiver to be request-derived — the same anchoring the main classification loop already used. Guards are held to the rule the rest of the classifier follows.

### Round 9b: a negation was looked through instead of pushed inward

```python
if not (tenant_id or other_flag):   # true only when BOTH are falsy
    raise HTTPException(400)
# was: REQUIRED
```

A request with no tenant and a truthy flag sails past that raise. Handing the whole operand to the mention scan finds `tenant_id` and credits it. The negation is now distributed — De Morgan — so `not (A or B)` becomes the `and` case, which already refuses a branch that is not itself a negative scope check. The other direction is preserved and tested: `not (A and B)` is true when either is falsy, so a missing tenant *does* reach the raise and it still counts.

### Round 10: an entry's category contradicted its own note

Correct, and it was mine. I filed `POST /memories/quality-metrics` under `cross-tenant-by-design` — *"legitimately spans tenants and would be wrong if it did not"* — while writing a note saying its three sibling stats routes all require `tenant_id` outright and this one looks like an inconsistency. Those cannot both be true.

The real problem was that the closed set had no honest slot for it. Added one, which the gate's own rule says needs a reviewer's agreement — and a reviewer raised it twice:

> **`grant-in-lieu-of-tenant`** — the binding tenant may be omitted when the caller supplies a `readable_tenant_ids` grant instead. **ALSO BACKLOG.** Storage authenticates nothing and takes that grant verbatim from the body, so accepting it in place of a home tenant is strictly weaker than requiring one.

There is now a `BACKLOG_CATEGORIES` set, so the summary marks all three — `id-addressed`, `blind-spot`, `grant-in-lieu-of-tenant` — rather than the two that happened to be hard-coded.

### Current state

```
tenant-scope gate: 189 routes + 191 service methods, 261 bound to a tenant, 119 allowlisted.
    6  admin-unscoped
    1  blind-spot  <- backlog
   21  cross-tenant-by-design
    1  grant-in-lieu-of-tenant  <- backlog
   43  id-addressed  <- backlog
    7  no-tenant-data
   40  opaque-body-write
```

**#1075 merged**, so `route:GET /api/v1/storage/memories/{memory_id}` and `method:memory_get_by_id` came off — the fourth enforced shrink, and again not by anyone remembering: the rebase left CI failing on both stale rows until they were regenerated. `id-addressed` is down from 47 to 43 across the three fixes.

One new row: `POST /tenant-usage/increment`, filed `blind-spot`. Its guard is real — `if not r.get("tenant_id") or not r.get("operation"): raise` — but `r` reaches the body through `for i, row in enumerate(rows)` and `r = dict(row)`, and the derivation tracker follows subscript/get chains only. I chose to record that rather than teach the tracker loop targets and `dict()` copies: nine rounds have been spent on the inference layer, and the category exists precisely so a human can say "checked, it is fine" without the gate having to prove it.

**53 tests** (was 49). Four added: the negated-or, the negated-and that must still count, a scope literal read off a config, and one read off the body inside a guard.

Verified: 53 gate tests, 5603 root tests, `mypy --config-file scripts/mypy.ini scripts/`, `ruff check`. `uv.lock` unchanged.

Eldad-Caura:
@claude

Eldad-Caura:
## Round 11 applied — `5465fb7c`

**Confirmed and fixed.** The `_categories` block documented six categories while an entry claimed a seventh. Diagnosis was exactly right: in round 10 I added `grant-in-lieu-of-tenant` to `CATEGORIES` and hand-edited the one entry to claim it, without re-running `--write`.

### Regenerated (the fix as suggested)

`--write` rewrites `_categories` from `CATEGORIES` and preserves hand-written `category`/`note` fields, so the repair is one line and all 119 entries survive it untouched:

```
 core-storage-api/tenant_scope_allowlist.json | 1 +
```

### Also fixed the reason it was possible

The finding is a drift between a generated artefact and the source it is generated from. That is the exact defect class this gate exists to catch, and the gate had it in its own file — because `load_allowlist` keeps only `raw["exceptions"]`, so `check()` never sees `_categories` and could not have noticed.

Regenerating alone fixes today's instance and leaves the channel open. So `check_category_doc` now asserts the checked-in block equals `CATEGORIES`, wired into `main` next to the other checks:

```
error: tenant_scope_allowlist.json documents categories that do not match the
script (missing grant-in-lieu-of-tenant). The '_categories' block is generated
from CATEGORIES in scripts/tenant_scope_gate.py; re-run `--write` so the file
explains every category its own entries claim.
```

**Equality rather than "every used category is present."** The weaker property is what you suggested, and it catches this instance. Equality also catches a description edited in the JSON instead of the script, and a category dropped from the script but left documented — and since `--write` generates the block, equality is always satisfiable by the command the error names. The message enumerates which of the three drifts occurred.

### Tests — 3 added, 55 total

You suggested a test; I took it, and put the check in the gate as well so CI fails rather than a reader noticing. Each fails without its fix, verified by reverting:

| test | fails without |
|---|---|
| `test_the_file_documents_every_category_its_entries_claim` | ✅ on the stale committed JSON |
| `test_trunk_is_green` | ✅ on the stale committed JSON — the gate itself now catches it |
| `test_category_doc_drift_is_caught` | ✅ `AttributeError` with the check absent; covers dropped / unknown / reworded |

### Verification

`mypy --config-file scripts/mypy.ini scripts/` → clean, 30 files. 55 gate tests pass. Gate green at 119 exceptions, unchanged. Root suite matches an untouched `main` checkout test-for-test — the 46 local failures (pubsub, providers, gemini, ingest) reproduce identically on `main` and are missing optional deps in my local Python, not this change. `uv.lock` untouched.

Worth noting this is the eleventh round, and the first to find a defect in the *data* rather than the AST inference. That reads like the inference layer has stopped yielding.

Eldad-Caura:
@claude

Eldad-Caura:
## Round 12 applied — `04d5228a`

**Confirmed.** The comment was wrong, and wrong in the direction that matters. Reproduced before touching it:

```
comment example (tenant_id and readable_tenant_ids): ('OPTIONAL', 'reads tenant_id with .get() ...')
both binding     (tenant_id and org_id)            : ('REQUIRED', '')
```

The `and` branch itself is correct — two binding keys give `REQUIRED`. What was wrong is that the comment illustrated it with `readable_tenant_ids`, which is `WIDENING_GRANT` and deliberately outside `BINDING_SCOPE`, so that operand returns an empty set, `all(parts)` is false, and the test is not credited.

Your last paragraph is the important part and I've taken it: the comment did not merely misdescribe the code, it described behaviour that **contradicts** the `grant-in-lieu-of-tenant` category doc, which says the grant is strictly weaker than a binding tenant. Left alone it is an argument, sitting in the source, for weakening the gate. So the fix is the comment, not the behaviour — the corrected text uses `org_id`, and says outright why the grant is excluded and that `quality-metrics` is the live consequence.

### Two tests added — and they do NOT fail on the parent

Flagging this explicitly, because the standing rule here is that a new test must fail on the parent commit, and these do not:

```
=== on the parent commit's script ===
4 passed, 53 deselected
```

They pass either way, because the parent's *code* was already right — only its comment was wrong. They are characterization tests, not proof of a fix, and I'd rather say so than let two green tests read as coverage of something.

What they do hold is the future weakening the bad comment argued for. Verified by mutation — adding `readable_tenant_ids` to `BINDING_SCOPE`:

```
FAILED tests/test_tenant_scope_gate.py::test_the_widening_grant_does_not_stand_in_for_a_binding_scope
1 failed, 3 passed
```

The two are deliberately adjacent and structurally identical, classified opposite ways: `not tenant_id and not org_id` → `REQUIRED`, `not tenant_id and not readable_tenant_ids` → `OPTIONAL`. That pair reads as symmetric and isn't, which is what made the comment plausible enough to write and to review, and nothing pinned it. If the value doesn't justify tests that pass on the parent, say so and I'll drop them and keep the comment fix alone.

### Verification

57 gate tests pass. `mypy --config-file scripts/mypy.ini scripts/` clean, 30 files. Gate green, allowlist byte-identical — no reclassification, as expected from a comment change. `uv.lock` untouched.

Eldad-Caura:
@claude

Eldad-Caura:
## Round 13 applied — `fb8e6c75`

Both findings confirmed. The first is a real gate defect in the dangerous direction — the first one found in thirteen rounds.

---

### Medium: `tenant_ids` trusted as a binding scope — **fixed, and it was live**

Not hypothetical. Measured before changing anything:

```
PostgresService methods taking tenant_ids: 1
  tenant_usage_query    required    also-binding=None
```

Its only caller is `POST /tenant-usage/query`, which does:

```python
...
await _svc.tenant_usage_query(body.tenant_ids, ...)
```

Straight from the request body, no binding tenant anywhere in the call — and the gate was reporting it **REQUIRED**, fully scoped. So the only instance of the parameter is the unsafe one, and it was invisible.

Worth noting it is *weaker* than the `readable_tenant_ids` sites that prompted the comparison: those all carry a binding `tenant_id` to fall back on, so omitting the grant narrows. Here there is nothing to fall back to. The caller names its own scope.

Took your option (b): `tenant_ids` is out of `BINDING_SCOPE`. The reasoning is shape, not provenance — a parameter taking a *list* of tenants cannot be the thing confining a statement to the caller's own; it says which tenants to span, and something upstream has to have earned that span. The gate can't see which, and doesn't get the benefit of the doubt.

Both paths are now carried under `grant-in-lieu-of-tenant`, whose description I widened to cover any caller-supplied tenant list. The route was already listed but under `opaque-body-write` — wrong, since that category describes an insert whose tenant rides in the payload and this is a POST-for-read that inserts nothing.

---

### High: the backlog formalizes unscoped writes — **split by blast radius; not fixing them here**

Two parts, and I disagree with one.

**On "without remediation tracking":** the ratchet *is* the forcing function, and it is not theoretical — it has fired four times during this PR's review. #1067, #1074 and #1075 each fixed one of these paths, and each time the rebase left CI **failing** on the now-stale rows until they were regenerated. `id-addressed` went 47 → 43 that way. The debt is not silently accepted; it cannot be added to, and every payment is enforced rather than remembered.

**On fixing the destructive ones here: declining, deliberately.** The brief for this PR was explicit that the allowlist is seeded so the check goes green on trunk today *without* a 90-item refactor. Folding ~21 behaviour changes into the PR that introduces the gate is how gates get reverted — and it would put a large untested tenant-scope change under the same review as the mechanism meant to police it. The three already landed as separate PRs, which is the shape that works.

**On the part that was right:** the category genuinely flattened two different findings. "Another tenant's row can be read by UUID" and "another tenant's row can be **deleted** by UUID" were one number. Split:

```
   22  id-addressed-read  <- backlog
   21  id-addressed-write  <- backlog, MUTATES
```

Classified by reading each path, not by name — which mattered: `POST /memories/entity-links` **sounds** like a write and calls `memory_get_entity_links_for_memories`, a read. Two automated detectors also disagreed with each other (a raw-SQL scan caught `agent_update_search_profile` but missed `memory_soft_delete`'s ORM mutation; an ORM scan did the reverse), which is why this is a reviewed category and not a fourth inference layer. Every one of the 43 landed in exactly one side; a test asserts none is left on the retired flat name.

**And the split needed teeth.** It creates a number people watch, and any such number can be made to fall the cheap way — relabel a write as a read: same id, same verdict, `check` and the size comparison both stay quiet. The ratchet now refuses that:

```
an unscoped mutating path was relabelled as something milder against origin/main:
      ~ method:already_here: id-addressed-write -> id-addressed-read
    A path leaves the mutating backlog by getting a tenant scope and leaving the
    allowlist, not by being refiled.
```

**On filing tracked issues:** I haven't, because creating issues in this repo isn't mine to do unilaterally. Happy to open one per mutating path, or a single tracking issue, if a maintainer says go.

---

### Tests — 6 added, 61 total

| test | fails without its fix |
|---|---|
| `test_a_plural_tenant_list_is_not_a_binding_scope` | ✅ on parent |
| `test_the_id_addressed_backlog_is_split_by_blast_radius` | ✅ on parent |
| `test_ratchet_fails_when_a_mutating_path_is_relabelled_as_a_read` | ✅ with the guard removed |
| `test_ratchet_allows_a_mutating_path_to_leave_the_list_entirely` | control for the above — proves the guard isn't over-broad |
| `test_the_file_documents_every_category_its_entries_claim` (r11) | ✅ |
| `test_category_doc_drift_is_caught` (r11, updated) | ✅ |

The first test asserts through the real enumeration against the real service — I drafted it against a synthetic class first and threw that away, since classifying a stand-in means reimplementing the rule under test and checking the copy agrees with itself.

### Verification

61 gate tests. `mypy --config-file scripts/mypy.ini scripts/` clean, 30 files. `--base origin/main` (CI's exact invocation) green. Root suite 46 failed / 5553 passed across two consecutive runs — baseline was 46 / 5547 with 55 gate tests, so exactly +6 and no new failures; the pre-existing 46 reproduce on an untouched `main` checkout. `uv.lock` untouched.

Eldad-Caura:
@claude

Eldad-Caura:
## Round 14 — `26cb8eb5`

One applied, one declined with evidence.

---

### Low: `::error::` annotations truncated at the first newline — **fixed, and it mattered more than "Low"**

Correct, and the consequence is worse than cosmetic: every multi-line message here puts the diagnosis on line one and the **remedy** on the continuation lines. Truncation drops exactly the half that says what to do.

Before and after, same failure, `GITHUB_ACTIONS=1`:

```
# before — the annotation stops here, the rest spills into the log
::error::the ratchet base 'no-such-ref' does not resolve, so the allowlist was never compared against anything.
    This is not 'nothing grew' — it is 'nobody looked'. Check that the
    `git fetch --no-tags --depth=1 origin main` step before this one ran.

# after — one workflow command, GitHub renders %0A as the line breaks
::error::the ratchet base 'no-such-ref' does not resolve, ...%0A    This is not 'nothing grew' — it is 'nobody looked'. Check that the%0A    `git fetch ...` step before this one ran.
```

`%` is encoded first — doing it after the newlines would rewrite the `%` of a `%0A` this function had just produced and render the escape as text. There is a test for that specifically.

---

### Medium: `enumerate_methods` downgrades on *any* defaulted binding param — **declining**

Measured first:

```
methods with BOTH a mandatory and a defaulted binding param: 0
```

No method has this shape, so the change moves no entry today. That makes it purely a question of which way to err, and the proposal errs toward **REQUIRED** — the direction this gate must never be wrong in.

The concrete counterexample is in this repo's own schema. `settings` and `lifecycle_audit` key on `org_id`, not `tenant_id`, so:

```python
async def settings_get(self, tenant_id: str, org_id: str | None = None): ...
```

can filter on the **defaulted** parameter and carry the mandatory one for logging. Under the proposal that is REQUIRED and never appears in the allowlist; the caller who forgets the argument gets the unscoped query. That is GHSA-wgvw-28pq-jc36 at the SQL layer — the `bulk-get` defect, which is the advisory this PR exists because of.

**On the route/method asymmetry being an inconsistency:** it is deliberate, because the two are reading different evidence and it does not combine the same way.

- A route is REQUIRED because a guard **raises** on the missing key. Positive proof no request reaches the body without it — so one guard settles the route however many optional reads sit beside it.
- A method is "mandatory" only because a parameter **has no default**. That proves a caller passes *something*; it does not prove the value is used as a predicate.

A guard outvoting optional reads is sound. An argument's presence outvoting a droppable predicate is not.

The rationale was already in `enumerate_methods`' docstring, and it did not prevent the proposal — so it is now a comment at the code, and a test. The test **passes on the parent** (it pins a decision, not a fix), flagged for the same reason as round 12's: prose demonstrably does not hold this and a red test does.

---

### Note on the branch: I force-pushed over a rebase, and repaired it

Being explicit since it touched shared state. Mid-round the branch was rebased upstream from base `424f764c` to `0868e4d6`; my fetch-and-push ran in one command, so I pushed my older-based commit over it instead of stopping. Nothing was lost — `0868e4d6` is PR #1025, still on `main`, and the branch content is a single commit that was intact throughout — but the branch base went backward by one commit until I rebased it forward.

It is now on `455c5dd2` (current `main`), and the tree is byte-identical to the maintainer rebase that landed as `26cb8eb5` — `git diff` between them is empty, so there was nothing left to push. Verified `0868e4d6` is an ancestor of the current HEAD.

---

### Tests — 4 added, 65 total

| test | fails without its fix |
|---|---|
| `test_a_multi_line_error_survives_as_one_annotation` | ✅ on parent |
| `test_annotation_encoding_does_not_eat_its_own_escapes` | ✅ on parent |
| `test_every_gate_error_is_emitted_as_a_single_line` | ✅ on parent |
| `test_a_mandatory_binding_param_does_not_outvote_a_defaulted_one` | ✗ passes on parent — pins the declined change |

### Verification

65 gate tests. `mypy --config-file scripts/mypy.ini scripts/` clean, 30 files. `--base origin/main` green on the new base, allowlist unchanged by the rebase. Root suite 46 failed / 5560 passed — same 46 as every prior round, all pre-existing and reproducing on an untouched `main` checkout. `uv.lock` untouched.

Eldad-Caura:
@claude

Eldad-Caura:
## Round 15 applied — `f7f6dc38`

### Low: duplicate `id`s silently collapse — **fixed**

Confirmed, and it is worse than "a bad hand-edit loses a row". Reproduced against the real loader:

```
entries parsed: 1 -> category kept: no-tenant-data
the more severe one was DISCARDED
```

The dict comprehension keeps whichever comes **last**, so the surviving row is chosen by array order. That is a way straight past the relabel guard added last round: append a second copy of an `id-addressed-write` entry carrying a milder category, and the lookup returns the milder one — the id set is unchanged, the verdict is unchanged, `check` and the ratchet both stay quiet, and the summary's MUTATES count falls without anything being fixed.

`_index_by_id` now refuses it, and `ratchet` parses the **base** copy through the same helper. The base is normally safe by induction — this check runs on every PR, so a duplicate cannot reach `main` once it is in place — but that induction has no base case for what is already there, and a duplicate in the base decides what every comparison is made against.

Surfaced as a gate error rather than a traceback, since a malformed allowlist is the author's to fix:

```
error: tenant_scope_allowlist.json lists method:agent_activity_digest_upsert more than once.
Two rows for one path means the gate silently picks whichever comes last — delete the
duplicate and keep the row a reviewer agreed to.
```

Verified `exit 1` with a duplicate injected, `exit 0` without.

---

### Suggestion: `tenant_id: str = Query(...)` read as defaulted — **documented, deliberately not fixed**

Agreed on the analysis, with one correction to the premise. You wrote that this style "doesn't appear to match the current routers' style" — it is in the tree, four times:

```
audit.py:152      limit: int = Query(default=100_000, ge=1, le=500_000)
documents.py:136  readable_tenant_ids: list[str] | None = Query(default=None)
documents.py:157  readable_tenant_ids: list[str] | None = Query(default=None)
documents.py:180  readable_tenant_ids: list[str] | None = Query(default=None)
```

None is a binding scope — three are the widening grant, which is genuinely optional and correctly read as such. So the cost today is zero rather than merely unlikely.

Not fixing it, and I think that is the right call rather than a deferral: reading the default expression to tell `Query(...)` from `Query(default=None)` is a fourth inference layer whose *only* effect is to move routes toward REQUIRED. Every wrong answer it gave would be a route that silently left the list. Over-listing costs a line someone deletes; under-listing is the advisory. The limitation is now a comment at the code with those four sites named, so the next person weighing it has the measurement rather than the hunch.

---

### Tests — 3 added, 68 total

| test | fails without its fix |
|---|---|
| `test_duplicate_allowlist_ids_are_rejected` | ✅ on parent |
| `test_a_duplicate_in_the_ratchet_base_is_an_error_not_a_pass` | ✅ on parent |
| `test_the_committed_allowlist_has_no_duplicates` | ✗ — a cheap invariant on the committed artefact |

### Verification

68 gate tests. `mypy --config-file scripts/mypy.ini scripts/` clean, 30 files. `--base origin/main` green. Root suite 46 failed / 5563 passed — the same pre-existing 46 as every round, reproducing on an untouched `main` checkout. `uv.lock` untouched.

Eldad-Caura:
@claude

Eldad-Caura:
## Follow-up issues filed, + round 16 applied — `28bf86ac`

### The 21 mutating paths are now tracked — as 12 issues, not 21

Filing one per allowlist row would have created duplicate work items, because the rows are not independent. Resolved each mutating route to its handler through the gate's own operation walk and read what it calls:

**9 route↔method pairs, 1:1** — one fix clears both rows:

| # | path | |
|---|---|---|
| #1080 | `DELETE /memories/{memory_id}` | `memory_soft_delete` |
| #1081 | `PATCH /entities/{entity_id}` | `entity_update` |
| #1082 | `PATCH /reports/{report_id}` | `report_update_completed` |
| #1083 | `PATCH /lifecycle-audit/{audit_id}` | `lifecycle_audit_finalize` |
| #1084 | `PATCH /agents/{agent_id}/search-profile` | `agent_update_search_profile` |
| #1085 | `PATCH /memories/{memory_id}/entities` | `memory_add_entity_links` |
| #1086 | `POST /fleet/commands/ack` | `fleet_ack_commands` |
| #1087 | `POST /memories/dedup-reviews/{review_id}/decision` | `dedup_review_decide` |
| #1088 | `POST /memories/increment-recall` | `memory_increment_recall` |

**All nine read no tenant at all** — checked, not assumed: `"tenant_id" in source` is False for every one of those handlers. So each is a live hole, not a theoretical one.

**2 method-only** (#1089 `agent_reset_search_profile`, #1090 `fleet_delete_commands_for_nodes`) — **not** live holes. Both current callers scope the lookup first: the reset route does `agent_get_by_id(agent_id, body["tenant_id"])`, and the fleet route resolves the node through `fleet_get_node_id(tenant_id=...)`. The guarantee lives in the caller rather than the query, which is a hazard for the next caller, not an open door today. Filing these at the same severity as the nine would have been wrong.

**1 dead code** (#1091 `entity_delete_entity_links`) — a repo-wide search finds no caller anywhere. Fix is deletion, which shrinks the backlog by one at zero behavioural cost.

Worth noting #1084 is self-evidencing: the sibling *reset* route in the same file scopes correctly while the *update* route does not, which is what makes it an oversight rather than a design.

**The references are enforced, not just written.** All 21 entries carry `Tracked in #N`, and `check` now rejects any `id-addressed-write` entry whose note loses it:

```
error: method:agent_reset_search_profile is id-addressed-write with no tracked issue in its
note. Every mutating unscoped path carries a `Tracked in #N` reference — file one and add
it, or fix the path and delete the entry.
```

---

### Round 16

**Medium — ratchet base hardcoded to `origin/main`: fixed, with a caveat on the premise.** The trigger is `branches: [main, dev]`, so you read that correctly. But **no `dev` branch exists on the remote**, and both sibling ratchets in this file (`legacy_name_ratchet.py:678`, the broker OpenAPI check at :139) hardcode `main` the same way — so this is a repo-wide latent gap, not something specific to this step.

Fixed anyway, because it aligns the code with the claim the ratchet already makes in its own docstring — "the comparison is exactly this PR's contribution" — which a hardcoded base quietly breaks for any PR not targeting `main`. Now `github.base_ref`, quoted, keeping the explicit destination refspec. `base_ref` is always set since the workflow is `pull_request`-only; if that changes, the empty refspec fails the fetch and the gate's unresolvable-base check fails too — both loud. I have not touched the two siblings; that is a separate change.

**Low — stale `id-addressed` in `_comment`: fixed.** Both the generator and the regenerated JSON now name `id-addressed-write` / `id-addressed-read` and say which half to clear first.

---

### Tests — 2 added, 70 total

| test | fails without its fix |
|---|---|
| `test_every_mutating_entry_carries_a_tracked_issue` | ✅ on parent |
| `test_a_mutating_entry_without_an_issue_is_rejected` | ✅ on parent |

### Verification

70 gate tests. `mypy` clean, 30 files. `actionlint` — 1 finding, byte-identical to the parent and at line 99, not this step; my step adds none. Gate green, `--write` produces no drift (notes survive regeneration). Rebased onto `e92464af` and re-run there. Root suite 46 failed / 5583 passed — the same pre-existing 46 as every round.

Eldad-Caura:
@claude

Eldad-Caura:
## Round 17 applied — `37ae2d7a`

Both confirmed. The second is an inconsistency I introduced last round, and the diagnosis is exactly right.

---

### Medium: `${{ }}` interpolated into a `run:` body — **fixed**

Correct, and I introduced it in round 16 while fixing something else. Now through `env:`:

```yaml
env:
  BASE_REF: ${{ github.base_ref }}
run: |
  git fetch --no-tags --depth=1 origin "+${BASE_REF}:refs/remotes/origin/${BASE_REF}"
  uv run python scripts/tenant_scope_gate.py --base "origin/${BASE_REF}"
```

Your reasoning for why "not exploitable today" is not a defence is the part worth keeping, so it is in the comment: the pattern gets copied to a context that *is* attacker-controlled. Through `env` the value is data to the shell rather than text pasted into the program. `actionlint` is unchanged at 1 finding — the pre-existing one at line 99, not this step.

---

### Medium: `grant-in-lieu-of-tenant` exempt from the tracked-issue requirement — **fixed, option 1**

You are right, and the cause is precise: I answered "does this need an owner?" with "does this mutate?", and those are different questions. `grant-in-lieu-of-tenant` destroys nothing, so it correctly stayed out of a set named `DESTRUCTIVE_CATEGORIES` — and fell out of the enforcement that set happened to drive. The "ALSO BACKLOG" wording did read as an oversight because it *was* one.

Rather than widen `DESTRUCTIVE_CATEGORIES` (which would make the `MUTATES` summary marker lie about three reads), the two questions are now two sets:

```python
DESTRUCTIVE_CATEGORIES = frozenset({"id-addressed-write"})              # drives the MUTATES marker
TRACKED_CATEGORIES = DESTRUCTIVE_CATEGORIES | {"grant-in-lieu-of-tenant"}  # drives the issue requirement
```

`id-addressed-read` is deliberately outside `TRACKED_CATEGORIES`, and that is now documented rather than incidental: at 22 entries it is the bulk backlog to grind down, and an issue per entry would be 22 issues nobody reads. That is the deprioritisation your option 2 asked to be explicit — applied to the category that actually warrants it.

**Two issues filed** for the three entries, which are two remediation units:

- **#1095** — `POST /tenant-usage/query` + `tenant_usage_query`. The caller names up to 1000 tenants and gets their usage totals back. p1.
- **#1096** — `POST /memories/quality-metrics`. Accepts `readable_tenant_ids` in place of a binding tenant while its three sibling stats routes require one. p2.

All three entries now carry `Tracked in #N`, and the gate enforces it.

---

### Tests — 72 total

| test | on parent |
|---|---|
| `test_every_owned_entry_carries_a_tracked_issue` | ✅ fails |
| `test_owning_a_category_is_not_the_same_question_as_mutating` | ✅ fails |
| `test_an_owned_entry_without_an_issue_is_rejected[grant-in-lieu-of-tenant]` | ✅ fails |
| `test_an_owned_entry_without_an_issue_is_rejected[id-addressed-write]` | passes — already enforced |

One note on how those are written: parametrising over `gate.TRACKED_CATEGORIES` directly evaluated at **collection** time, so against the parent the whole module raised `AttributeError` and 72 tests became one collection error. That satisfies "fails on parent" only by breaking everything, and it would make the file uncollectible against any older gate. Now parametrised over a literal tuple, with `set(_TRACKED) == gate.TRACKED_CATEGORIES` asserted inside the set-comparison test so the list cannot drift. On the parent the module collects, 69 pass, and exactly the 3 intended tests fail.

### Verification

72 gate tests. `mypy` clean, 30 files. `actionlint` 1 finding, identical to parent. Gate green; `--write` is idempotent (byte-identical output, so the notes survive regeneration). Rebased onto `9454aa81` and re-run there.

Root suite 46 failed / 5585 passed. One run in this round reported 47 — the extra was `test_unknown_field_rejection.py`, which my commit does not touch and which passes 21/21 in isolation both here and on an untouched `main` checkout. Order-dependent flake in the full-suite run, same as the 46↔47 fluctuation seen earlier; flagging it rather than quoting the clean number.

Eldad-Caura:
@claude

arkash20 (approved):

PR #1072, opened 2026-08-29
ci(release): assert the python client publishes what its tag claims

### LINKED ISSUE #1071: publish-python-client.yml has neither of the two release guards its npm twin has
`publish-python-client.yml` builds `clients/python` and publishes to PyPI. It has **no tag-vs-manifest guard** and **no post-publish verification**. `publish-npm-client.yml` has both, and each exists because of a specific way a release can look successful and not be.

Found while adding the caura-spelled release tag in #1070. Flagged there rather than fixed, because it is a separate concern.

## 1. A tag that disagrees with `pyproject.toml` publishes silently

The python workflow never reads the tag. `GITHUB_REF_NAME`, `github.ref` and `TAG_VERSION` appear nowhere in the file — it builds from `pyproject.toml` and publishes, so the tag is a pure trigger.

Consequence: push `caura-client-v2.0.0` while `pyproject.toml` says `1.0.1`, and the workflow publishes **1.0.1** with a green tick, under a tag claiming 2.0.0. The release looks done; the registry serves something else.

The npm twin refuses this:

```yaml
- name: The tag must agree with package.json
  if: startsWith(github.ref, 'refs/tags/')
  run: |
    TAG_VERSION="${GITHUB_REF_NAME#caura-client-ts-v}"
    TAG_VERSION="${TAG_VERSION#memclaw-client-ts-v}"
    PKG_VERSION=$(node -p "require('./package.json').version")
    if [[ "$TAG_VERSION" != "$PKG_VERSION" ]]; then
      echo "::error::tag ${GITHUB_REF_NAME} says ${TAG_VERSION} but package.json says ${PKG_VERSION}"
      exit 1
    fi
```

Note it must strip **both** tag prefixes, and after #1070 the python workflow has two spellings too (`caura-client-v*` and the legacy `memclaw-client-v*`), so a port needs both strips or it fails every release under one spelling.

The `if: startsWith(github.ref, 'refs/tags/')` condition matters: this workflow also has `workflow_dispatch`, where there is no tag to compare and the guard must not run.

## 2. A green publish step is not evidence PyPI serves it

The npm twin polls the registry after publishing, with this rationale in the file:

> #870 merged a version bump, every check passed, and both registries went on serving the old build because no tag had been pushed — the tick meant "nothing went wrong", not "it shipped". Ask the registry.

The python workflow ends at `pypa/gh-action-pypi-publish` and asks PyPI nothing. The same failure is available: a dispatch run with no version bump completes green while PyPI keeps serving the previous release.

## What to do

Port both, adapted rather than copied:

- version from `pyproject.toml` via `tomllib` (the pattern `core-api/Dockerfile` already uses)
- distribution name is `caura-client`
- verification against the PyPI JSON API rather than `npm view`
- keep the retry-with-backoff shape; PyPI is not immediately consistent either

## Not a hypothetical

This is the same class as the `memclaw-client-v*` trigger that #1068 found unmarked — a release contract that nothing was asserting. Here nothing asserts that the thing published is the thing the tag named.

Eldad-Caura:
Closes #1071. Ports both release guards from `publish-npm-client.yml`, adapted rather than copied.

## 1. A tag that disagrees with `pyproject.toml` published silently

The workflow never read the tag — it builds from `pyproject.toml` and publishes. So pushing `caura-client-v2.0.0` against a pyproject saying `1.0.1` shipped **1.0.1** with a green tick, under a tag claiming otherwise. The release looked done; PyPI served something else.

```yaml
- name: The tag must agree with pyproject.toml
  if: startsWith(github.ref, 'refs/tags/')
```

**Both prefixes are stripped**, because #1070 gave this workflow two spellings. Stripping only the canonical one would fail every release tagged the legacy way — the same coupling the npm twin documents at its own strip. The `if:` matters too: this workflow also has `workflow_dispatch`, where there is no tag to agree with and the guard must not run.

## 2. A green publish step is not evidence PyPI serves it

The twin carries this check after #870, where a version bump passed every gate while both registries went on serving the old build — the tick meant "nothing went wrong", not "it shipped". This asks the index, with retries, since PyPI is not immediately consistent.

**One deliberate difference from the twin.** The check asks for the *project* JSON and tests membership, rather than requesting the version URL directly. Measured:

```
curl -sf https://pypi.org/pypi/caura-client/99.99.99/json   ->  exit 56
```

PyPI answers that 404 in a way curl reports as **56, "failure receiving network data"** — indistinguishable from a real network fault, so an absent version and a broken connection would have looked identical, and the retry loop would have burned five attempts either way without the log telling you which. Membership-testing separates them.

## Verified by execution, not by reading

| case | result |
|---|---|
| `caura-client-v1.0.1` vs pyproject `1.0.1` | strips to `1.0.1`, **passes** |
| `memclaw-client-v1.0.1` vs `1.0.1` | strips to `1.0.1`, **passes** |
| `caura-client-v2.0.0` vs `1.0.1` | **fails** — "tag says 2.0.0 but pyproject.toml says 1.0.1" |
| `memclaw-client-v9.9.9` vs `1.0.1` | **fails** with the same shape |
| probe, live `1.0.1` | reports served |
| probe, synthetic `99.99.99` | reports not served, loop would retry |

The name and version lookups both resolve against the real file (`caura-client`, `1.0.1`).

## Scope

Workflow-only. No product code, no change to what is built or where it is published, and the legacy tag trigger is untouched.

### DIFF
--- a/.github/workflows/publish-python-client.yml
+++ b/.github/workflows/publish-python-client.yml
@@ -30,6 +30,23 @@ jobs:
       - uses: actions/setup-python@v7
         with:
           python-version: "3.12"
+      # A tag that disagrees with pyproject.toml does not fail — it publishes
+      # whatever pyproject.toml says, under a tag claiming otherwise. The release
+      # then looks done while PyPI serves a different version. Both prefixes are
+      # stripped so the check works on either tag. Skipped on workflow_dispatch,
+      # where there is no tag to agree with.
+      - name: The tag must agree with pyproject.toml
+        if: startsWith(github.ref, 'refs/tags/')
+        run: |
+          TAG_VERSION="${GITHUB_REF_NAME#caura-client-v}"
+          TAG_VERSION="${TAG_VERSION#memclaw-client-v}" # legacy-name-ok: strips the tag pattern kept above
+          PKG_VERSION=$(python -c "import tomllib; print(tomllib.load(open('pyproject.toml','rb'))['project']['version'])")
+          if [[ "$TAG_VERSION" != "$PKG_VERSION" ]]; then
+            echo "::error::tag ${GITHUB_REF_NAME} says ${TAG_VERSION} but pyproject.toml says ${PKG_VERSION}"
+            exit 1
+          fi
+          echo "publishing $(python -c "import tomllib; print(tomllib.load(open('pyproject.toml','rb'))['project']['name'])")==${PKG_VERSION}"
+        working-directory: clients/python
       - name: Build
         run: |
           pip install build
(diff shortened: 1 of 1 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@claude

Both findings taken — they were correct, and the second one caught a real gap between what this PR claimed and what the code did.

**Retry window.** It was the npm twin's flat 5×10s, ported as-is. Now ten attempts with increasing backoff (`sleep $((attempt * 5))`), **225s of waiting in total**, and no sleep after the final attempt so a failure does not idle for nothing. I measured the total rather than estimating it — my first draft comment said "~4.5 minutes" and the arithmetic is 225s, so the comment now states the measured figure.

**Indistinguishable failures — my error to own.** This PR's own description claimed the two outcomes were "separate outcomes". Internally they were: the membership test fixed curl's misleading exit 56. But the loop still printed one message for both and ended on one generic error, so a network fault and an absent version read identically to whoever opens a red run. That gap was real and I had asserted it away.

They now report apart on every attempt, and `LAST_STATE` carries the distinction into the final error so it survives past the per-attempt lines:

```
attempt 3/10: could not reach the PyPI index, retrying
::error::published without error, but after 10 attempts: could not reach the PyPI index (caura-client==1.0.1)

attempt 3/10: index reachable but 1.0.1 is not in it, waiting
::error::published without error, but after 10 attempts: index reachable but 1.0.1 is not in it (caura-client==1.0.1)
```

I kept `curl -sf` and a separate `if !` rather than switching to `-w "%{http_code}"`. With the membership test doing the version check, curl's only remaining job is "did the fetch work", which `-f` answers directly — parsing a status code would add a branch without adding information.

Executed all three paths against a stubbed index rather than reasoning about them:

| path | result |
|---|---|
| version served | exits 0 on attempt 1 |
| index reachable, version absent | ends `index reachable but 1.0.1 is not in it` |
| index unreachable | ends `could not reach the PyPI index` |

`actionlint` still clean. No change to the tag guard, which was not flagged.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #1074, opened 2026-08-29
fix(memories): require a tenant scope on bulk-get, and apply it in SQL

Eldad-Caura:
Closes the remaining half of **GHSA-wgvw-28pq-jc36**. Follow-up to #1067, and the fix the round-1 review of #1069 asked for.

## What was wrong

`POST /memories/bulk-get` took a list of ids and an **optional** `tenant_id`. Omit it and every id you named came back, whichever tenant owned it, from a service that authenticates nothing:

```python
tenant_filter = body.get("tenant_id")          # optional
by_id = await _svc.memory_get_memories_by_ids(uids)   # no tenant in the query at all
...
if mem is None or (tenant_filter is not None and mem.tenant_id != tenant_filter):
```

The route's own docstring justified the optionality — *"to mirror the single-row `GET /memories/{id}` contract"*. That did not make it safe; it meant two endpoints shared a hole.

## Two changes, because the filter was weak in two different ways

**It could be omitted.** `tenant_id` is now a `_require` on the route and a non-defaulted argument on `CoreStorageClient.bulk_get_memories`. The client half was not hypothetical: of its two call sites, `entity_service` passed a tenant and **`crystallizer_service` did not** — with `tenant_id` in scope the whole time, simply never passed. That is what an optional security argument reliably produces.

**It was applied in Python, after the fact.** `memory_get_memories_by_ids` took ids alone and returned whatever they matched; the route then compared `tenant_id` per row and blanked the mismatches. So even on the path that *did* supply a tenant, another tenant's rows crossed the database boundary and were discarded afterwards by a comparison no test pinned. The predicate now lives in the query:

```python
stmt = select(Memory).where(
    Memory.id.in_(memory_ids),
    Memory.tenant_id == tenant_id,
    Memory.deleted_at.is_(None),
)
```

A foreign row is never selected, rather than selected and dropped.

**Unchanged:** the per-id `None`. A foreign id still reads as absent rather than erroring, so this does not become an existence oracle for memory UUIDs.

## Reachability, stated plainly

Through core-api the crystallizer's ids were already tenant-scoped upstream, so on that path this was defence-in-depth rather than a live leak. What made it an advisory is that storage authenticates nothing and trusts the tenant its caller names — so the filter being *skippable* is the whole vulnerability, and "the caller will remember" is not a control. One of the two callers did not remember.

## Tests

Five new. **Two fail on the parent commit**: an omitted tenant is rejected, and `{"ids": []}` is rejected too — the guard runs before the empty-list short-circuit, so the contract doesn't depend on how many ids you happen to send. The other three pin behaviour that must not change: a foreign id reads as absent, a foreign id is indistinguishable from a nonexistent one, and input order is preserved (callers zip the response back).

Three existing tests in `tests/test_storage_bulk_endpoints.py` called the client with no tenant and are updated — they were pinning the vulnerable contract, and their failure is the change working.

Verified against pgvector:pg16 — 180 core-storage-api tests, 5541 root tests, `ruff check` / `ruff format --check` and `mypy src/` for both `core-api` and `core-storage-api` at CI's scopes. `uv.lock` unchanged.

## Interaction with #1069

#1069 adds the CI gate that keeps this class of defect from recurring; its allowlist currently carries `route:POST /api/v1/storage/memories/bulk-get` and `method:memory_get_memories_by_ids` under `id-addressed`, annotated as this advisory. **When this merges, those two rows must come out** — `python3 scripts/tenant_scope_gate.py --write` does it, and the gate fails on the stale entries until someone does, which is the intended behaviour. Same shrink #1067 produced.

Does not touch `docker-compose.yml`, storage auth middleware, or the README's Self-Hosted section — that is the reachability/authentication half of GHSA-wgvw-28pq-jc36 and is being handled separately.

### DIFF
--- a/core-api/src/core_api/clients/storage_client.py
+++ b/core-api/src/core_api/clients/storage_client.py
@@ -1065,19 +1065,21 @@ class CoreStorageClient:
     async def bulk_get_memories(
         self,
         ids: list[str],
-        tenant_id: str | None = None,
+        tenant_id: str,
     ) -> list[dict | None]:
         """Fetch many memories in one round-trip; order matches input ``ids``.
 
-        Missing rows (deleted, nonexistent, or — when ``tenant_id`` is
-        provided — cross-tenant) come back as ``None`` in the same slot
-        rather than being dropped from the list. Lets callers zip the
-        response back to their original id list. Capped at 1000 ids
-        server-side; callers needing more must chunk client-side.
+        Missing rows — deleted, nonexistent, or belonging to another tenant —
+        come back as ``None`` in the same slot rather than being dropped from
+        the list. Lets callers zip the response back to their original id list.
+        Capped at 1000 ids server-side; callers needing more must chunk
+        client-side.
+
+        ``tenant_id`` is required. As an optional argument it was the client
+        half of GHSA-wgvw-28pq-jc36, and one of the two call sites did in fact
+        omit it.
         """
-        payload: dict[str, Any] = {"ids": ids}
-        if tenant_id is not None:
-            payload["tenant_id"] = tenant_id
+        payload: dict[str, Any] = {"ids": ids, "tenant_id": tenant_id}
         return await self._post(  # type: ignore[return-value]
             "/memories/bulk-get", payload, read=True
         )
--- a/core-api/src/core_api/services/crystallizer_service.py
+++ b/core-api/src/core_api/services/crystallizer_service.py
@@ -524,7 +524,7 @@ async def _run_crystallization(
     # deleted) — same per-row "skip missing" semantics the old loop's
     # ``if mem:`` gate provided, just with Nx fewer HTTPs (audit P5).
     memories_by_id: dict[UUID, dict] = {}
-    bulk_rows = await sc.bulk_get_memories([str(mid) for mid in total_ids])
+    bulk_rows = await sc.bulk_get_memories([str(mid) for mid in total_ids], tenant_id=tenant_id)
     for mid, mem in zip(total_ids, bulk_rows):
         if mem is not None:
             memories_by_id[mid] = mem
--- a/core-storage-api/src/core_storage_api/routers/memories.py
+++ b/core-storage-api/src/core_storage_api/routers/memories.py
@@ -441,22 +441,24 @@ async def find_duplicate_hash(
 async def bulk_get_memories(request: Request) -> list[dict | None]:
     """Fetch many memories by id in a single round-trip.
 
-    Body: ``{"ids": ["uuid", ...], "tenant_id": "..." (optional)}``.
+    Body: ``{"ids": ["uuid", ...], "tenant_id": "..."}``.
 
     Returns a list of memory dicts in the **same order** as the input ids,
-    with ``null`` for ids that don't exist (or are soft-deleted, or — if
-    ``tenant_id`` is provided — belong to a different tenant). Order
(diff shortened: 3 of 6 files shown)


Eldad-Caura:
## Review round 1 — applied

Correct, and a fair catch: I rewrote the paragraph explaining that `tenant_id` is required and left the `Body:` line seven lines above it still saying `(optional)`. The two halves of the same docstring contradicted each other, and the stale half is the one an integrator skims first.

```diff
-    Body: ``{"ids": ["uuid", ...], "tenant_id": "..." (optional)}``.
+    Body: ``{"ids": ["uuid", ...], "tenant_id": "..."}``.
```

I also swept for the same wording elsewhere in case the contract was described in more than one place — `core-api/src`, `core-storage-api/src` and `docs/` have no other "optional" near `bulk-get` or `bulk_get_memories`. The client docstring was already rewritten in the original commit.

Re-verified after: 180 core-storage-api tests, `ruff check` / `ruff format --check` and `mypy src/` at CI's scopes. `uv.lock` unchanged.

Eldad-Caura:
@claude

Eldad-Caura:
## Review round 2 — premise accepted, fix declined *for this PR*

The finding is correct. `GET /memories/{memory_id}` declares `tenant_id: str | None = None` and falls through to `memory_get_by_id(memory_id)` — a bare `session.get(Memory, memory_id)` with no tenant predicate. Same defect, same advisory, and this PR's own docstring points straight at it. I am not disputing any of that; it was already on my list, and #1069's allowlist carries it as `route:GET /api/v1/storage/memories/{memory_id}` under `id-addressed` with the note *"Falls back to memory_get_by_id when tenant_id is omitted"*.

**It does not belong in this PR, because it is not the same size of change.** Measured, not estimated:

| | `bulk-get` (this PR) | `GET /memories/{id}` |
|---|---|---|
| storage-client methods | 1 | **2** — `get_memory` and `get_memory_for_tenant` both exist |
| call sites | 2 | **12** |
| call sites already passing a tenant | 1 of 2 | **1 of 12** |

Closing it means migrating ~11 call sites off `get_memory` onto `get_memory_for_tenant`, across `contradiction_detector`, `memory_service`, `consumer` and `routes/testing` — and at several of those the tenant is not currently in scope and has to be threaded in. That is a change with its own blast radius, its own review, and its own chance of breaking an internal path. Folding it into a PR that is currently green and narrowly about one endpoint would make both harder to review and would put a 12-call-site refactor behind a Critical label on a fix that is already complete.

It is the same argument that produced this PR: #1067 went out on its own, then #1069, then this one, each one endpoint at a time — and each shrank the allowlist rather than growing it.

**I will open it as the next PR.** Concretely: make `tenant_id` required on the route, delete the unscoped `memory_get_by_id` in favour of `memory_get_by_id_for_tenant`, migrate the call sites, and drop the row from #1069's allowlist. Two storage-side rows come out of the backlog with it — the route and `method:memory_get_by_id`.

Until then it is not lost or quietly accepted: it is a named row in a ratcheted list that a maintainer has to look at, which is the difference between a backlog and an oversight. That is what #1069 is for, and this endpoint is exactly the kind of thing it exists to keep visible.

No code change in this round. This PR remains: 180 core-storage-api tests, 5541 root tests, ruff + mypy clean at CI's scopes.

Eldad-Caura:
@claude

Eldad-Caura:
@erni-a pelase approve

erni-a (approved):

PR #1075, opened 2026-08-30
fix(memories): require a tenant scope on GET /memories/{id}

Eldad-Caura:
The single-row form of **GHSA-wgvw-28pq-jc36** — and the contract #1074's `bulk-get` optionality was justified by *"mirroring"*. Raised as a Critical on that PR's review; declined there because it is a much larger change, and promised as this one. Follows #1067 and #1074.

## What was wrong

```python
@router.get("/{memory_id}")
async def get_memory(memory_id: UUID, tenant_id: str | None = None) -> dict:
    if tenant_id is not None:
        memory = await _svc.memory_get_by_id_for_tenant(memory_id, tenant_id)
    else:
        memory = await _svc.memory_get_by_id(memory_id)   # bare session.get, no predicate
```

Omit the query parameter and you got the row's whole content, whichever tenant owned it.

## Why this took a whole PR

The client had **two** methods, and the unsafe one was the shorter name:

| | call sites | passing a tenant |
|---|---|---|
| `get_memory(memory_id)` | 11 | **1** |
| `get_memory_for_tenant(tenant_id, memory_id)` | 10 | 10 |

Ten of eleven callers of the unscoped method omitted the tenant. Not because anyone decided to — the argument simply wasn't there to forget. So this deletes the choice: there is now **one** method, `get_memory(memory_id, tenant_id, *, read=True)`, and the unscoped call cannot be written.

**Keeping the name and requiring the argument** — rather than keeping `get_memory_for_tenant` and deleting `get_memory` — was deliberate. The latter reads better in isolation, but `get_memory` is mocked *by name* in 31 test files; renaming it churns all of them, while requiring an argument leaves 54 plain attribute mocks untouched. The `_for_tenant` qualifier only ever existed to distinguish it from the unsafe sibling; with that gone it is noise. I tried it the other way first and measured.

Storage side: `memory_get_by_id` is deleted, and `memory_get_by_id_for_tenant` now filters **in SQL** rather than fetching by primary key and comparing `memory.tenant_id != tenant_id` in Python afterwards — same change #1074 made to `bulk-get`. A foreign row is never selected.

404 still covers "no such memory" and "not yours" alike, so this is not an existence oracle for memory UUIDs.

## Two behaviour changes worth reviewing

**`/testing/set-field`** fetched the row unscoped and compared tenants in Python to raise 403. It now scopes the fetch, so a foreign id is a **404** — one less oracle. Admin credentials (no tenant) still get 403; `memories.tenant_id` is NOT NULL and could never equal `None`, so they were rejected before too.

**A soft-deleted row is now absent from a by-id read.** That was *always* true of the scoped path — two integration tests reached one only because they went through the unscoped path. They now assert the 404 and read `deleted_at` from the `include_deleted` listing, which tests the same two properties (the row is soft-deleted, and a PATCH did not resurrect it).

## Tests

Five new; **one fails on the parent** (the omitted tenant — the vulnerability). The other four pin the contract: a foreign id reads as absent, is indistinguishable from a missing one, own rows return, and soft-deleted rows do not.

The wider churn is the interesting part. The first full run after the signature change gave **104 failures** — call sites and mock signatures that had been passing no tenant. Each was a place the old signature let a caller skip the scope, and every one now names its tenant. That number is the honest measure of what "optional" cost.

Verified after rebasing onto `d6fd10b9` (which includes #1066's storage authentication — no overlap with this change): 199 core-storage-api tests, 5550 root tests, `ruff check` / `ruff format --check` and `mypy src/` for both packages at CI's scopes, broker OpenAPI baseline current. `uv.lock` unchanged.

## Effect on #1069

Removes two more rows from the tenancy-gate allowlist — `route:GET /api/v1/storage/memories/{memory_id}` and `method:memory_get_by_id`, both `id-addressed`. That gate fails on stale rows until they are regenerated, so the shrink is enforced rather than remembered; it will be the fourth.

### DIFF
--- a/core-api/src/core_api/clients/storage_client.py
+++ b/core-api/src/core_api/clients/storage_client.py
@@ -573,19 +573,22 @@ class CoreStorageClient:
                 _storage_detail(exc.response), _storage_duplicate_fields(exc.response)
             ) from exc
 
-    async def get_memory(self, memory_id: str, *, read: bool = True) -> dict | None:
-        """Fetch one memory by id.
+    async def get_memory(self, memory_id: str, tenant_id: str, *, read: bool = True) -> dict | None:
+        """Fetch one memory by id, within ``tenant_id``.
+
+        ``tenant_id`` is required. This method used to take an id alone, and
+        eleven call sites used it — none of them passing a tenant, which is what
+        an optional scoping argument reliably produces. There was a second,
+        correctly scoped ``get_memory``; having both meant the unsafe
+        one was always a shorter name away. Now there is one, and the unscoped
+        call cannot be written rather than merely being discouraged.
 
         ``read=False`` routes to the WRITER. Pass it for a read-your-write — a
-        read-back of a row this request (or an event's producer) just wrote, where
-        replica lag would make the row or the freshly-PATCHed column invisible.
-        Same reasoning as the write-path dedup gate below, which passes it for the
-        same reason.
+        read-back of a row this request (or an event's producer) just wrote,
+        where replica lag would make the row or the freshly-PATCHed column
+        invisible. Same reasoning as the write-path dedup gate below.
         """
-        return await self._get(f"/memories/{memory_id}", read=read)
-
-    async def get_memory_for_tenant(self, tenant_id: str, memory_id: str) -> dict | None:
-        return await self._get(f"/memories/{memory_id}", tenant_id=tenant_id)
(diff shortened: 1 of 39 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #1076, opened 2026-08-30
docs(readme): fix the quickstart write, field paths, and single-agent framing

yanki-m:
## Why

Every curl block in the README was executed verbatim against a standalone stack running the published `ghcr.io` image. The four-command "try it locally" demo — the first thing a new user runs — **fails on its first call**.

### 1. The quickstart write returned 422 (`0e5688b`)

Both examples omitted `agent_id`:

```
{"error":{"code":"INVALID_ARGUMENTS","message":"Field required",
  "details":{"errors":[{"loc":["body","agent_id"],"msg":"Field required"}]}}}
```

**Worth a maintainer's eye:** current source already makes the field optional — `schemas.py:76` declares `agent_id: str | None = Field(default=None)` and standalone fills `DEFAULT_AGENT_ID` — but published images through v2.x still reject the omission. So the README described behaviour source has and `:latest` does not. The examples now pass `agent_id` explicitly, which works against every version; **cutting a release is what makes source and image agree**, and that call is yours.

### 2. `summary` and `tags` were documented one level up from where they live (`0e5688b`)

Six field names were listed in one breath, reading as six siblings. Four are. `summary` and `tags` aren't on `MemoryOut` — enrichment writes them into `metadata` (`memory_service.py:1155, 2001, 3069`) and reads them from there (`agent_digest.py:77`). A client following the old text reaches for `response.tags`, gets `undefined`, and concludes enrichment failed while the data sits in `response.metadata.tags`.

The `enrichment_pending` sentence now names its real path too — `metadata.enrichment_pending` — confirmed by the re-run.

### 3. The keyless quickstart contradicted its own demo (`0e5688b`)

The page offers four commands with "nothing to configure", then searches `"authentication token lifetime"` for a memory that says `"JWT with 15-minute expiry"`. That's a paraphrase jump requiring real embeddings. The note now says plainly that this query is the one needing a provider key.

### 4. "Built for fleets, **not single agents**" (`f214884`)

That line disqualified the reader who has one agent — nearly everyone at line 37 of a README they just opened. Caura serves that case fine; fleets are where it's optimised, not where it's limited. The comparison paragraph at L444 had a softer version of the same problem, conceding single-agent memory to Mem0/Zep/Letta right after a table Caura mostly wins. Both now establish parity at one agent and separation above it.

## Verification

```
# fixed write, run verbatim
PASS  id=43c44675-500b-432e-9e68-f5b34de9588a
metadata: {"write_mode":"fast","enrichment_pending":true,"write_latency_ms":8}
```

Test memories were written to a scratch `agent_id` and deleted individually afterwards.

## One thing I could not verify — please read

Fix 3 is **deliberately non-committal about which keyword query does work.** My first attempt substituted `"JWT expiry"` and claimed it works keyless. Testing returned **zero results** — it would have swapped one broken example for another.

The reason is worth a look independently of this PR: on the test stack, a freshly written, fully embedded memory is retrieved by **no query at all**, while its neighbours rank normally.

| Query | Results | Top similarities |
|---|---|---|
| "Growth plan" | 5 | 0.631, 0.529, 0.482 |
| "deployment" | 5 | 0.523, 0.485, 0.460 |
| **"JWT expiry"** | **0** | — |
| **"authentication token lifetime"** | **0** | — |

Ruled out: soft-delete, visibility, fleet scope, a degenerate embedding (1024 dims, all non-zero, statistically indistinguishable from a retrievable row), and the similarity floor — `min_similarity: 0.0` still returns nothing. The only difference between the retrievable and unretrievable rows is `agent_id`.

That looks like a retrieval issue rather than a documentation one, so it isn't addressed here. **If it reproduces on a clean stack, the keyless quickstart can't be fixed by editing prose at all** — the demo returns nothing regardless of which query is printed. A concrete keyword example belongs in the README once someone with a healthy instance can demonstrate one.

## Suggested follow-ups (not in this PR)

- **A CI job that executes the README's curl blocks.** All of the above sat in the highest-traffic section of the repo and no existing test could see it. Fixing them without this just resets the clock — the `e2e/` harness is already there to hang it on.
- Confirm `title` populates with a real provider key. It returns `null` on the test instance, which has none — expected, but unverified, and the README claims it.
- The README is 1,332 lines; API Reference (236), Upgrading from v1.x (127) and Public API & Stability (109) read like `docs/` material, and `## Telemetry` appears twice (L1224, L1261).

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ Agents write plain text. Caura turns it into searchable, governed, self-improvin
 
 **One loop, three pillars: write, recall, compound** — every interaction makes the next one smarter.
 
-**Built for fleets, not single agents.** Public agent-memory benchmarks (LoCoMo, LongMemEval) measure one agent, one user, one long conversation — the single-chatbot shape. The deployment shape we see in production is the opposite: dozens or thousands of agents working on behalf of a company, sharing what they learn under governance. Caura is architected around that shape from day one — scoped memory, cross-agent outcome propagation, fleet-wide trust tiers — and competes on the axes that compound with agent count: latency, token efficiency, and governance. See [Performance](#performance) for the numbers, or read the [benchmarks write-up](https://caura.ai/blog/caura-benchmarks).
+**Optimized for fleets.** One agent works, and that's where most teams start — nothing below changes for a single-agent setup. What Caura adds is headroom: scoped memory, cross-agent outcome propagation, and fleet-wide trust tiers are there from the first write, and they keep paying off as agents multiply. Public agent-memory benchmarks (LoCoMo, LongMemEval) measure one agent, one user, one long conversation — the single-chatbot shape — so they score the on-ramp rather than the axes that compound with agent count: latency, token efficiency, and governance. That second shape is what we see in production: dozens or thousands of agents working on behalf of one company, sharing what they learn under governance. See [Performance](#performance) for the numbers, or read the [benchmarks write-up](https://caura.ai/blog/caura-benchmarks).
 
 > **In production at eToro (NASDAQ: ETOR):** 300+ AI agents on one governed
 > memory — 26,500+ memories, 1,372 shared skills, 23 ms p50 search.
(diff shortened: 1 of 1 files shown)


Eldad-Caura:
@claude

Eldad-Caura (approved):

Eldad-Caura:
Addressed the low-severity quickstart UX note in c07f0c83. The four-command promise now ends at the successful write, and the search block identifies semantic recall as requiring a real embedding provider before a user runs it. I deliberately did not replace the query with `"JWT"`, because the artifact’s live-stack test could not verify that retrieval result. The PR is consolidated to one signed commit; no additional Claude pass requested for this docs-only low-severity fix.

PR #1077, opened 2026-08-30
feat(crystallizer): subject-local Type-II state materializer, shadow phase (A59)

arkash20:
## Summary

Third route at Type-II staleness — a fact that stops being a safe current default because a **later, non-contradicting** fact broke its practical basis ("commutes by bicycle" → "tore ACL, no weight-bearing six weeks").

Two routes were measured and rejected first:
- **read-side tricks** — recency injection useless (the update sits mid-history), query expansion +20pp on one probe type but net-flat overall;
- **per-write bridge + judge (A58)** — flagged 2 of 15 known cases and sometimes targeted *another subject's* memory, because predicates are sparse and its fallback pool was not subject-local.

This one is **batch-first**: once per night, per **subject**, one LLM call over that subject's own live memories, inside the crystallizer that already sweeps each tenant. The corpus shape is what makes it cheap — subjects hold ~1.4 memories on average and only ~23% hold two or more (measured), so the candidate set is *changed multi-memory subjects only*: `C` calls per tenant-night, zero read-time cost.

The output is deliberately a **predicate-specific successor** naming the current state in the stale row's own vocabulary, not a profile blob — so it reuses retrieval machinery that already exists (successor injection, the A34 ranking contract) instead of adding a read path or a new status.

## Shadow phase — this PR writes nothing

It emits a `type_ii_staleness` report section: counts, rejection reasons, and a bounded sample carrying **the stale text, the supporting text, and the proposed replacement side by side**, so precision can be scored by hand in minutes. That is the direct lesson of the A58 spike, whose verdicts went only to logs in a shape nobody could score.

Deterministic gates, no LLM trust: cited ids must be in that subject's own bundle (cross-subject targeting is impossible by construction), support must **postdate** the stale row (temporal causality), replacement non-empty, confidence floor. **A NULL predicate does not disqualify a proposal** — predicates are sparsely populated in production, which is exactly why A58's rdf route found nothing.

Flag `type_ii_materializer_shadow`, default off. Independent of `auto_crystallize` (it creates no memories), and wrapped so a failure can never abort the surrounding report.

## Provenance

The design shape — predicate-specific successor over profile blob, subject-local bundles, changed-subject watermark, no new status — came from an independent design pass. My amendments: sparse-predicate tolerance, a stable idempotency key (`sha256(subject|stale_id)`) so nightly re-runs can't stack successors, and making the shadow output *scorable* rather than log-only.

## Testing

9 unit tests: subject-local bundling + live-only filter, candidate gating, all five rejection paths, NULL-predicate tolerance, idempotency key stability, auditable sample shape, no LLM spend on single-fact subjects, per-subject failure containment, flag default + phase wiring. Full suite **5559 passed**; ruff + mypy clean.

## Next (not in this PR)

Turn the flag on for one tenant, hand-score ~30 sampled proposals off the report section. Write mode is only built if precision clears the bar — and when it is, my recommendation is to demote sources to `conflicted` (soft, exact-match carve-out preserved) rather than `outdated`, since a wrong call then stays recoverable.

Refs: canonical A59; A58 (#1024); findings `benchmark/a57-recall-experiments-findings.md`.

### DIFF
--- a/core-api/src/core_api/config.py
+++ b/core-api/src/core_api/config.py
@@ -345,6 +345,16 @@ class Settings(BaseSettings):
     # enforcing path (``unsafe`` status). Skipped when the new memory has no
     # resolved subject_entity_id.
     basis_invalidation_shadow: bool = False
+    # A59 — Type-II state materialization, subject-local and batch (runs in the
+    # nightly crystallizer, not per write). SHADOW ONLY at this setting: one
+    # LLM call per CHANGED multi-memory subject proposes which of that
+    # subject's own memories stopped being safe current defaults and what
+    # should stand instead; proposals are validated deterministically and
+    # emitted as an auditable report section. Nothing is written to memories.
+    # Precision is scored by hand off that section before a write mode is
+    # built (the A58 spike could not be scored because its verdicts only went
+    # to logs). Default off.
+    type_ii_materializer_shadow: bool = False
     crystallizer_enabled: bool = True
     crystallizer_stale_days: int = 180
     crystallizer_dedup_sample_size: int = 1000
--- a/core-api/src/core_api/services/crystallizer_service.py
+++ b/core-api/src/core_api/services/crystallizer_service.py
@@ -419,6 +419,12 @@ async def _execute_crystallization(
             except Exception:
                 logger.warning("type_ii shadow phase failed for %s", tenant_id, exc_info=True)
                 type_ii = {"enabled": True, "error": True}
+            # Nested under ``hygiene`` deliberately: ``analysis_reports`` has
+            # FIXED columns (summary/hygiene/health/usage_data/issues/
+            # crystallization), so a NEW top-level report key is silently
+            # dropped by storage and never reaches the API. Nesting needs no
+            # migration and rides the route that already serialises hygiene.
+            hygiene["type_ii_staleness"] = type_ii
 
         if auto_crystallize:
             try:
@@ -456,9 +462,6 @@ async def _execute_crystallization(
                     "info": info,
                 },
                 "hygiene": hygiene,
-                # A59 — audit section for the Type-II shadow sweep (empty
-                # ``{"enabled": False}`` when the flag is off).
-                "type_ii_staleness": type_ii,
                 "health": health,
                 "usage_data": usage,
                 "issues": issues,
--- a/core-api/src/core_api/services/crystallizer_service.py
+++ b/core-api/src/core_api/services/crystallizer_service.py
@@ -26,6 +26,7 @@ except ImportError:
         pass  # type: ignore[misc]
 
 
+from core_api.config import settings
 from core_api.constants import (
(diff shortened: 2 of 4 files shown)


arkash20:
@Eldad-Caura please approve

Eldad-Caura (dismissed):

arkash20:
## Wet test (added after the first push — the PR was not wet-tested when opened)

Rebuilt core-api from this branch on the local enterprise stack with `TYPE_II_MATERIALIZER_SHADOW=1` and real providers, wrote a subject pair, triggered a crystallizer run, and inspected the report through the API.

**It caught a defect the mocked tests could not.** `analysis_reports` has FIXED columns (summary/hygiene/health/usage_data/issues/crystallization) and the report detail route hand-assembles a fixed key list on top of that — so a new top-level `type_ii_staleness` key was computed, logged, then **silently dropped**: invisible to every consumer, which would have made the whole shadow phase unscorable. Fixed by nesting under `hygiene` (no migration, rides existing serialization). Commit `cc8cb5bc`.

**Live result after the fix** — the section returns through `GET /crystallize/reports/{id}` with a hand-scorable sample:

```json
{"mode": "shadow", "subjects_scanned": 1, "subjects_called": 1, "accepted": 1, "rejected": 0,
 "samples": [{
   "stale_content": "Robin... commutes to the office by bicycle every day, about 25 minutes each way.",
   "supporting_content": ["Robin... tore their ACL playing football and cannot bear weight on that leg for six weeks..."],
   "replacement_content": "Robin... is currently unable to bear weight on the injured leg for six weeks...; whether they can commute by bicycle during this period is unknown.",
   "broken_basis": "the later injury prevents bearing weight on that leg, which affects the feasibility of cycling",
   "confidence": 0.72, "predicate": null, "key": "be40866da25ef444"}]}
```

Both source rows kept their statuses across every run — shadow purity holds on real data. Note the proposal declines to invent a replacement commute mode, which is the desired behavior.

**Two real constraints observed, worth knowing before the write phase:**
1. **Same-subject dependency.** In one run the two memories resolved to two *different* `subject_entity_id`s, so neither bundle reached the 2-memory threshold and no call fired. Reach is bounded by entity-resolution quality — A63's canonicalization directly widens it.
2. **Skips what contradiction detection already handled.** In another run Path A/C had already marked the older row `conflicted` with a supersede chain; the live-status filter then correctly skipped the subject. Intended, and it means the two mechanisms do not double-adjudicate.

**Unrelated blocker found while standing the stack up:** current main requires `CORE_STORAGE_SHARED_SECRET` (caura#1066) but the enterprise local `docker-compose.yml` never gained it, so a fresh local core-api crash-loops with `RuntimeError: CORE_STORAGE_SHARED_SECRET is required for core-api`. caura-enterprise `0085e018` fixed staging/prod deploys only. Reported separately — not part of this PR.

arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #1079, opened 2026-08-30
fix(crystallizer): Type-II precision gates — 10% -> 80% on the same corpus (A59)

arkash20:
## Summary

The first real shadow sweep of the A59 materializer (#1077) scored **~10% precision** by hand — and two proposals were *actively harmful*: the "replacement" restated the stale claim ("based in Seattle" after evidence of a move) or asserted the old and new claims together. Written as successors, those would enshrine the error with fresh timestamps — worse than leaving the corpus alone.

This PR fixes that. Measured on the **identical six tenants**:

| | before | after |
|---|---|---|
| accepted proposals | 13 | 5 |
| correct on hand-scoring | 1–2 (~10%) | **4 of 5 (80%)** |
| harmful (restates/merges the stale claim) | **2** | **0** |

## Three changes

1. **Exclude raw transcript turns from subject bundles.** `memory_type` is *not* the discriminator — the nine false positives were typed `fact`/`preference` exactly like the true hits (verified against the DB). Content shape is: a dialogue turn ("User: …", "Assistant: …") is not a claim about the subject's current state, so it can be neither a stale default nor evidence that one broke.
2. **Prompt** now requires the replacement to state what is NO LONGER true, and forbids restating the stale claim or asserting both together.
3. **Deterministic restatement gate** — reject any replacement whose token overlap with the stale row is ≥0.6.

## What the surviving proposals look like

- *"works from home, office is here in the house"* + *"gave the landlord a move-out date, building is being converted"* → *"living situation is in transition; whether they will still be working from home is unknown"*
- *"on shift 4AM–noon every day"* + *"back home by nine most mornings"* → correctly **retires** the shift claim (before the fix, it asserted both)
- *"machine lacks the computing power"* + *"4K multi-angle edits stay smooth"* → correctly retires the limitation

Note the five LongMemEval tenants dropped to `subjects_scanned=0`: **every** subject-bearing row in them is a raw dialogue turn. That confirms the diagnosis rather than hiding it — those corpora contain nothing state-bearing for this mechanism to reason over.

## Still shadow-only

n=5 on one synthetic corpus justifies continuing and nothing more. The write gate is unchanged: sweep a **real** tenant corpus (distilled memories, not benchmark seeds) and score ~30 proposals. My standing reservation for that phase: demote sources to `conflicted`, not `outdated`, so a wrong call stays recoverable.

## Testing

12 unit tests (3 new: transcript exclusion, restatement rejection, genuine replacement still passes). Full suite **5565 passed**; ruff + mypy clean. Re-swept the same six tenants on a rebuilt image — the numbers above are from that live run, read back through `GET /crystallize/latest`.

Refs: A59; follows #1077; findings `benchmark/a57-recall-experiments-findings.md`.

### DIFF
--- a/core-api/src/core_api/services/type_ii_materializer.py
+++ b/core-api/src/core_api/services/type_ii_materializer.py
@@ -37,6 +37,7 @@ from __future__ import annotations
 import hashlib
 import json
 import logging
+import re
 from typing import Any
 
 from core_api.config import settings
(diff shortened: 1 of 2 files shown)


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #1092, opened 2026-08-30
fix(search): preserve lexical matches through candidate limits

Eldad-Caura:
## Why

Follow-up to #1076. The README probe exposed a real retrieval failure: an embedded memory that full-text matched JWT expiry still returned zero results, including with min_similarity 0.0. Strong vector-only rows filled the storage candidate window before core-api could apply its hybrid ranking policy.

## What changed

- Reserve a small, bounded supply of full-text matches before the storage candidate limit, while preserving the existing pending-embedding fallback when lexical scoring is disabled.
- Carry the full-text match signal through storage and both core search implementations, then reserve at most one lexical result at the final trim.
- Allow lexical evidence to relax only the untuned global similarity fallback. Per-request, agent, and tenant similarity floors remain strict.
- Add diagnostics for full-text matches and global-floor bypasses.
- Add an end-to-end regression for the exact README write and JWT expiry query under a saturated candidate window, covering both pipeline and legacy search paths and an explicit strict-floor control.

## Validation

The new end-to-end regression failed on the pre-fix code in both search paths, including the explicit 0.0 retry.

- API retrieval/recall slice: 194 passed
- core-storage-api suite: 199 passed
- core-api mypy: 203 source files clean
- core-storage-api mypy: 84 source files clean
- Ruff check and format: clean for changed production and new test files

One DCO-signed commit, rebased onto current main.


Eldad-Caura:
@erni-a

erni-a (approved):

Eldad-Caura:
Ready to merge: all required checks are green, Claude review found no issues, and @erni-a approved.

erni-a (approved):

PR #1094, opened 2026-08-30
ci(release): harden the caura-sdk PyPI publish path

Eldad-Caura:
## Summary

- fail before checkout when either GitHub OIDC request variable is missing, without minting a token
- require tag-triggered releases to use `caura-sdk-v*` and match `pyproject.toml`
- verify the canonical package version through PyPI's project JSON after upload
- report reachable-but-absent, uncheckable response, and unreachable index as distinct failures

## Why

Without a tag/manifest guard, a `caura-sdk-v2.0.0` tag can publish the manifest's `1.0.0` and still report success. Querying the project JSON avoids PyPI's missing-version response being surfaced by curl as exit 56, which otherwise looks like a network fault.

The audit table credited this workflow with a post-publish check, but `origin/main` had none. This PR fixes the measured workflow rather than silently preserving that stale table entry.

## Verification

- `actionlint .github/workflows/publish-python-sdk.yml`
- extracted guard harness: OIDC present passes; either/both variables missing fail; matching tag passes; mismatched version and wrong prefix fail
- extracted post-publish harness: served and normalized prerelease versions pass; valid-index absence, malformed response, and unreachable index fail with distinct final states
- `python -m build --no-isolation clients/caura-sdk-meta`
- exact CI `ruff check` and separate `ruff format --check` scopes
- `pytest --noconftest tests/test_legacy_name_ratchet.py tests/test_do_not_touch_sentinel.py -q` (174 passed)
- ratchet and sentinel scripts against `origin/main` (no new lines; all 39 protected strings survive)

### DIFF
--- a/.github/workflows/publish-python-sdk.yml
+++ b/.github/workflows/publish-python-sdk.yml
@@ -40,16 +40,83 @@ jobs:
     permissions:
       id-token: write # required for PyPI trusted publishing
     steps:
+      # Trusted publishing has no stored token to check. GitHub exposes both
+      # request variables only when this job can obtain an OIDC credential.
+      # Check the context without minting a live token in this step.
+      - name: The publish credential must be obtainable
+        run: |
+          if [[ -z "${ACTIONS_ID_TOKEN_REQUEST_URL}" || -z "${ACTIONS_ID_TOKEN_REQUEST_TOKEN}" ]]; then
+            echo "::error::No OIDC token-request context, so PyPI trusted publishing cannot authenticate. This job needs 'permissions: id-token: write'; there is no stored token to fall back on."
+            exit 1
+          fi
+          echo "OIDC token-request context is present"
       - uses: actions/checkout@v7
       - uses: actions/setup-python@v7
         with:
           python-version: "3.12"
+      # This package was created under the Caura name and has no legacy tag
+      # history, so tag-triggered releases accept only caura-sdk-v*.
+      - name: The tag must agree with pyproject.toml
+        if: startsWith(github.ref, 'refs/tags/')
+        run: |
+          if [[ "${GITHUB_REF_NAME}" != caura-sdk-v* ]]; then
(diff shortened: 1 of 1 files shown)


Eldad-Caura:
@claude

arkash20 (dismissed):

Eldad-Caura:
Applied in `a6f89ef7`:

- `packaging` is now an explicit direct install alongside `build`.
- The 404 diagnosis is fixed by capturing HTTP status separately from curl transport status.

I did not take the suggested fail-fast behavior: this workflow creates `caura-sdk` on its first successful trusted publication, so a temporarily absent project JSON endpoint is an eventual-consistency case worth retrying. It is now reported precisely as `index reachable but project caura-sdk is not in it`; network failure, other HTTP responses, malformed JSON, and valid-project version absence each have separate states.

The extracted workflow guard passed the served-version and normalized-prerelease cases; every constructed failure case failed with its intended final diagnostic. `actionlint`, package build, exact Ruff gates, 174 ratchet/sentinel tests, and both repository safety scripts are green.

Eldad-Caura:
@claude

Eldad-Caura:
Applied in `74f953bb`: every PyPI index request now has a 5-second connection timeout and 15-second total timeout, so a stalled socket enters the bounded transport-failure retry path.

Declined the non-canonical-version suggestion because the two comparisons intentionally protect different contracts: the tag must exactly match the source manifest, while the registry check must use the PEP 440 version emitted by the build backend. The extracted test proves the complete case: raw `1.0.0-rc.1` passes the tag/manifest guard and the registry check correctly finds canonical `1.0.0rc1`.

The full extracted guard matrix, `actionlint`, package build, exact Ruff gates, 174 ratchet/sentinel tests, and both repository safety scripts pass on the amended commit.

Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #1098, opened 2026-08-30
test(readme): execute keyless quickstart in CI

Eldad-Caura:
## Summary
- make the no-key quickstart search for `JWT expiry`, the keyword path fixed in #1092, while keeping the semantic paraphrase as the provider-enabled follow-up
- wait for Compose health before the documented curls and describe keyless heuristic enrichment accurately
- extract the marked README curl block through a strict localhost-only parser, reconstruct safe curl requests, and require search to return the memory just written
- boot the current-source core-api and core-storage-api images in CI and run that round trip, with isolated Compose controls and guaranteed cleanup

## Validation
- red-before-green: the runner against a fresh keyless stack failed with the old `authentication token lifetime` query (`search did not return the memory`)
- green: the corrected marked block wrote and found its memory against a fresh current-source Compose stack
- `pytest tests/test_readme_quickstart.py tests/test_readme_keyword_recall.py -q` — 12 passed
- `ruff check` + `ruff format --check` on the new script/tests
- `mypy --config-file scripts/mypy.ini scripts/` — clean (30 files)
- `actionlint -ignore SC2086 .github/workflows/ci.yml` — clean; SC2086 is an unchanged repository baseline
- full `/simplify` pass — no remaining findings

Follow-up to #1076 and #1092.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -570,13 +570,30 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: read
+    env:
+      # Local-only tag shared by the two images used for the README Compose
+      # check below. Keeping this job's existing display name avoids changing
+      # the branch-protection status context while broadening its runtime-image
+      # coverage.
+      CAURA_VERSION: readme-ci
+      COMPOSE_PROFILES: ""
+      README_COMPOSE_PROJECT: "readme-quickstart-${{ github.run_id }}"
     steps:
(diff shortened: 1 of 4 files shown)


Eldad-Caura:
Fixed the legacy-name ratchet failure without adding an exemption: the CI job now builds each image under a neutral local tag, reads Compose’s expected service image name from docker compose config, and retags locally before the --no-build startup. Local legacy-name ratchet, actionlint, Ruff, mypy, and the focused README tests (12 passed) are clean.

Eldad-Caura:
@claude

arkash20 (approved):

Issue #1095, opened 2026-08-30
storage: POST /tenant-usage/query totals any tenants the caller names


Eldad-Caura:
`route:POST /api/v1/storage/tenant-usage/query`

Service method: `method:tenant_usage_query`

### The defect

The route takes a caller-supplied list of tenants and totals their usage counters, with no binding tenant anywhere in the call:

```python
class TenantUsageQuery(BaseModel):
    tenant_ids: list[str] = Field(min_length=1, max_length=MAX_TENANTS_PER_QUERY)
    ...

@router.post("/query")
async def query_tenant_usage(body: TenantUsageQuery) -> dict:
    return {"periods": await _svc.tenant_usage_query(body.tenant_ids, ...)}
```

`body.tenant_ids` goes straight into the query. The caller names its own scope, so anyone who can reach the port can read any tenant's per-period, per-operation usage totals by naming them — up to `MAX_TENANTS_PER_QUERY = 1000` at a time.

This is weaker than the `readable_tenant_ids` sites elsewhere in the service: those all carry a binding `tenant_id` to narrow to, so omitting the grant fails closed. Here there is nothing to fall back to.

It was classified `REQUIRED` — fully scoped — by the tenant-scope gate until `tenant_ids` was removed from `BINDING_SCOPE` in caura-ai/caura#1069, on the grounds that a parameter taking a *list* of tenants says which tenants to span rather than which one confines the query.

### The fix

Take a binding `tenant_id` (or `org_id` — the consumer bills an organisation, and an org owns several tenants, which is why the plural exists) and check the requested `tenant_ids` are ones that caller is entitled to, rather than trusting the list. Then delete both allowlist entries.

---

### Why this is filed

`core-storage-api` authenticates no request (GHSA-wgvw-28pq-jc36), and docker-compose publishes it on `0.0.0.0:8002`. Anything reachable there is reachable by anyone who can reach the port.

Tracked by the tenant-scope gate added in caura-ai/caura#1069 under the `grant-in-lieu-of-tenant` category, which the gate flags `<- backlog`. The ratchet means the allowlist can shrink but never grow. **Fixing this means deleting its allowlist entries; CI will fail until they are removed, which is the intended forcing function.**

Eldad-Caura:
## Correcting this issue: the fix I proposed is not implementable here

[...]

PR #1101, opened 2026-08-30
fix(storage): require a binding tenant on quality-metrics

### LINKED ISSUE #1096: storage: POST /memories/quality-metrics accepts a caller-supplied grant instead of a binding tenant
`route:POST /api/v1/storage/memories/quality-metrics`

### The defect

The route accepts a caller-supplied `readable_tenant_ids` grant *in place of* a binding tenant:

```python
if not tenant_id and not readable_tenant_ids:
    raise ...
```

So a request with no `tenant_id` but a `readable_tenant_ids` list naming other tenants passes the guard, and the caller has named its own scope. Storage authenticates nothing and takes the list verbatim from the body, so "a grant was supplied" is strictly weaker than "a tenant was bound".

The inconsistency is what makes this look unintentional rather than designed: its three sibling stats routes (`stats-breakdown`, `daily-durable-counts`, and the third in the same router) require a binding `tenant_id` and treat `readable_tenant_ids` only as a widening grant on top of it — the fail-closed shape used at all eleven other grant sites in the service.

### The fix

Require `tenant_id` the way the sibling stats routes do, and let `readable_tenant_ids` widen from there rather than substitute for it. That makes omitting the grant narrow to the home tenant — fail-closed — and brings the four stats routes into one shape. Then delete the allowlist entry.

---

### Why this is filed

Eldad-Caura:
Closes #1096.

`POST /memories/quality-metrics` guarded with:

```python
if not tenant_id and not body.get("readable_tenant_ids"):
    raise HTTPException(status_code=422, detail="tenant_id is required")
```

so a request carrying **only** a `readable_tenant_ids` list passed it, and the caller named its own scope. `core-storage-api` authenticates no request (GHSA-wgvw-28pq-jc36) and reads that list verbatim from the body, so accepting it in place of a home tenant is strictly weaker than requiring one.

The error message already said `tenant_id is required` while the condition did not enforce it.

### Why this is a defect rather than a design

Measured across the router: of the **eight** routes that forward `readable_tenant_ids` to the service layer, this was the **only** one with that guard shape. The other seven — `stats-breakdown` and `daily-durable-counts` among them — all use `if not tenant_id: raise 422`, treating the grant as a widening *on top of* a binding tenant.

That difference is what makes the grant safe everywhere else: with a tenant always present, forgetting the grant narrows to the home tenant. Fail-closed. Here there was nothing to narrow to.

### The change

One line. `readable_tenant_ids` is untouched and still widens the read across the named siblings.

### Tests

Five, in `core-storage-api/tests/test_quality_metrics_tenant_scope.py`. Deliberately only **one** fails on the parent commit:

| test | on parent |
|---|---|
| `test_a_grant_alone_is_rejected` | ✅ **fails** — this is the defect |
| `test_no_scope_at_all_is_rejected` | passes |
| `test_own_tenant_is_returned` | passes |
| `test_the_grant_still_widens_on_top_of_a_tenant` | passes |
| `test_omitting_the_grant_narrows_to_the_home_tenant` | passes |

The four that pass either way are there on purpose: the cheap way to satisfy the first test is to break the grant, and those four make that impossible to do quietly. Stating it explicitly so the ratio does not read as four tests that forgot to assert anything.

### Verification

`core-storage-api` suite: 10 failed / **194** passed / 7 errors — byte-identical failures and errors to pristine `origin/main` (10 / **189** / 7), so the delta is exactly the five new tests and nothing regressed. The pre-existing 10+7 are environmental in my local setup and reproduce unmodified on `main`.

`ruff check` and `ruff format --check` at CI's exact scopes (`src/`, `tests/`, run separately, discovery-based, no `--config`) — clean at ruff 0.16.5, matching the pinned `>=0.16.2`. `mypy src/` from `core-storage-api/` — clean, 84 files. `uv.lock` untouched. Rebased onto `01780eac` and re-run there.

### Relationship to #1069

#1069 adds the CI gate that found this. This path is on its allowlist under `grant-in-lieu-of-tenant`; once this merges, that entry becomes stale and #1069's own ratchet will fail until it is regenerated away — the fifth time that mechanism has forced a shrink to be recorded rather than remembered.

Its sibling issue #1095 (`POST /tenant-usage/query`) is **not** fixed here, and on investigation is not fixable the way I described when filing it — see the note I am adding to that issue.

### DIFF
--- a/core-storage-api/src/core_storage_api/routers/memories.py
+++ b/core-storage-api/src/core_storage_api/routers/memories.py
@@ -1370,10 +1370,20 @@ async def quality_metrics(request: Request) -> dict:
     readable_tenant_ids?}``. Returns ``{total, reused, total_recalls,
     top_recalls, by_type:{type:{total,reused}}}``. Same scope/visibility as
     ``/stats-breakdown``; read-only.
+
+    ``tenant_id`` is required, and ``readable_tenant_ids`` does not substitute
+    for it. The guard used to accept either, which made this the only one of
+    the eight ``readable_tenant_ids`` routes where a caller could name its own
+    scope and supply no home tenant: the grant is read verbatim from the body
+    by a service that authenticates nothing, so accepting it in place of a
+    tenant is strictly weaker than requiring one. The error message already
+    said ``tenant_id is required`` while the condition did not enforce it.
+    With a tenant always present, omitting the grant narrows to that tenant —
+    fail-closed — which is what the other seven sites have always done.
     """
     body: dict = await request.json()
     tenant_id = body.get("tenant_id")
-    if not tenant_id and not body.get("readable_tenant_ids"):
+    if not tenant_id:
         raise HTTPException(status_code=422, detail="tenant_id is required")
     ca = body.get("created_after")
     try:
--- a/core-storage-api/tenant_scope_allowlist.json
+++ b/core-storage-api/tenant_scope_allowlist.json
@@ -731,13 +731,6 @@
       "note": "Tracked in #1088."
     },
     {
-      "id": "route:POST /api/v1/storage/memories/quality-metrics",
-      "verdict": "OPTIONAL",
-      "evidence": "reads tenant_id with .get() and never rejects a missing one",
-      "category": "grant-in-lieu-of-tenant",
-      "note": "Accepts a caller-supplied readable_tenant_ids in place of a binding tenant: 'if not tenant_id and not readable_tenant_ids: raise'. Its three sibling stats routes (stats-breakdown, daily-durable-counts, list) all require tenant_id outright, and this one's own docstring lists tenant_id as non-optional. Tracked in #1096."
-    },
-    {
       "id": "route:POST /api/v1/storage/reports",
       "verdict": "NONE",
       "evidence": "no binding tenant read from the request",
--- a/core-storage-api/tests/test_quality_metrics_tenant_scope.py
+++ b/core-storage-api/tests/test_quality_metrics_tenant_scope.py
@@ -1,0 +1,118 @@
+"""``POST /memories/quality-metrics`` requires a binding tenant.
+
+The guard was ``if not tenant_id and not readable_tenant_ids``, so a request
+carrying only a ``readable_tenant_ids`` list passed it. That grant is read
+verbatim from the request body by a service that authenticates nothing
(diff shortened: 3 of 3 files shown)


erni-a (dismissed):

Eldad-Caura:
## Rebased onto `d7bb4ca4` — #1069 merged, and the gate immediately caught this PR

The gate landed on `main` while this was open, so this PR is now the first to be measured by it. It failed straight away, correctly:

```
error: route:POST /api/v1/storage/memories/quality-metrics is in the allowlist but no
longer needs to be — it is either gone or now scoped. Delete the entry (--write does it
for you).
```

That is the mechanism doing its job: the fix in this PR made an allowlist row obsolete, and CI refused to pass until the row was removed rather than trusting anyone to remember. Regenerated:

```
Allowlist shrank by 1: route:POST /api/v1/storage/memories/quality-metrics
tenant-scope gate: 189 routes + 191 service methods, 261 bound to a tenant, 119 allowlisted.
```

120 → 119 allowlisted, `grant-in-lieu-of-tenant` 3 → 2, and `261 bound to a tenant` (up from 260) — this route is now counted as scoped rather than excepted. Fifth enforced shrink.

### Re-verified on the new base

Gate suite (now part of this PR's CI): 72 passed. `--base origin/main`: exit 0. Storage suite: 10 failed / 194 passed / 7 errors, byte-identical failures to pristine `main` at 10 / 189 / 7 — delta is exactly the five new tests. `ruff check` and `ruff format --check` at CI's scopes, clean at 0.16.5. `mypy src/` clean (84 files); `mypy --config-file scripts/mypy.ini scripts/` clean (31 files).

### One process note

The branch had been rebased upstream to `82abcc16` while I worked, and I fetched immediately before the `--force-with-lease` push, which defeats the lease's protection — the lease compares against the remote-tracking ref I had just refreshed. I checked afterwards rather than before: `82abcc16` was my original commit replayed onto the new `main` **with the stale allowlist row still in it**, so it would have failed CI, and this push strictly improved it. Same parent, same content plus the regeneration, nothing lost. Flagging it because "it turned out fine" is not the same as "it was safe", and the check belongs before the push.

Eldad-Caura:
@erni-a pleae aapprove

erni-a (approved):

Eldad-Caura:
@claude

PR #1099, opened 2026-08-30
refactor(npm): make @caura/client the implementation

Eldad-Caura:
## Summary

- implements the approved Option A naming direction
- makes `@caura/client@1.0.1` the canonical TypeScript implementation
- converts `@caura/memclaw-client@1.0.2` into a thin compatibility alias that depends on `@caura/client@^1.0.1`
- routes canonical and legacy tags to separate workflows with exact tag/version validation
- preserves both published 1.0.0/1.0.1 legacy versions and the existing `@caura/client@1.0.0` floor
- distinguishes verified dependency absence (E404) from authentication, transport, and malformed-response failures

The legacy alias starts at `^1.0.1`, so it cannot resolve to the historical `@caura/client@1.0.0` wrapper and form a dependency cycle.

## Verification

- 190 focused Python tests passed
- 14 TypeScript client tests passed
- canonical and legacy packages both packed successfully
- offline install proved both package names export the same `Caura` implementation
- workflow guards were exercised with passing legitimate cases and failing wrong-version, unsupported-tag, E404, E401, curl-exit-56, and malformed-response cases
- `actionlint`, all applicable Ruff checks/format checks, sentinel checks, ratchet checks, and `git diff --check` passed
- `/simplify` review completed cleanly

## Release sequence

Only after both G PRs merge:

1. publish `@caura/client@1.0.1` from a `caura-client-ts-v1.0.1` (or preserved `caura-npm-v1.0.1`) tag
2. publish `@caura/memclaw-client@1.0.2` from `memclaw-client-ts-v1.0.2`

No npm publish or authentication was performed. PR2 will add the repository-wide brand/name guard after this naming PR merges.

PR #1103, opened 2026-08-30
fix(npm): enforce client tag and package brand agreement

Eldad-Caura:
## Summary

- adds a pre-build guard to `publish-npm-client.yml` that reads the package name from `package.json`
- consolidates brand and existing version validation into one agreement step, so supported tag prefixes have one validator
- derives the tag brand and version from the two supported Caura tag prefixes
- rejects both the preserved legacy package name and unrecognized package brands on the canonical release path
- skips the guard when no tag is present, matching the existing publish guards
- leaves the private plugin package entirely out of scope
- preserves the real legacy tag/package route introduced by #1099

## Guard proof

The workflow shell block was executed directly against the real manifests:

- Caura tag + legacy package: exit 1
  `::error::tag caura-client-ts-v1.0.1 is Caura-spelled but package.json publishes @caura/memclaw-client`
- Caura tag + canonical package: exit 0
  `tag caura-client-ts-v1.0.1 and package.json name @caura/client are both Caura-spelled`
- wrong version + canonical package: exit 1
  `::error::tag caura-client-ts-v9.9.9 says 9.9.9 but package.json says 1.0.1`
- legacy tag + actual legacy workflow/package: exit 0
  `publishing @caura/memclaw-client@1.0.2`
- Caura tag + unknown package brand: exit 1 with an explicit unrecognized-brand error

## Verification

- 198 focused invariant tests passed
- 14 canonical TypeScript client tests passed
- `actionlint` passed for both client publish workflows
- all exact CI Ruff lint and format scopes passed
- legacy-name ratchet passed with only three explicit rule-3 compatibility assertions and no new unexempt lines
- all 39 do-not-touch sentinels passed
- all four `uv.lock` hashes remained unchanged
- `/simplify` completed cleanly after fixing its dead-branch, step-boundary, fail-open, and stale-naming findings
- review finding about triplicated tag-prefix logic was fixed by consolidating brand and version validation

No npm authentication or publishing was performed.

### DIFF
--- a/.github/workflows/publish-npm-client.yml
+++ b/.github/workflows/publish-npm-client.yml
@@ -25,6 +25,7 @@ name: Publish the canonical TypeScript client to npm
 
 on:
   push:
+    # Keep these two patterns aligned with the single pre-build agreement step.
     tags:
       - "caura-client-ts-v*"
       - "caura-npm-v*"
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@claude

Eldad-Caura:
Applied the duplicated-prefix finding from both review comments.

The brand check now absorbs the existing version check, so the tag prefixes appear only in the GitHub trigger and one pre-build validator (the trigger filter cannot reference runtime workflow variables). The old post-build validator was removed.

Regression proof: against the pre-fix commit, `caura-client-ts-v9.9.9` exited 0 after brand validation; the amended step exits 1 with `tag caura-client-ts-v9.9.9 says 9.9.9 but package.json says 1.0.1`.

Post-rebase verification: 198 focused invariant tests, 14 TypeScript tests, both workflow actionlint checks, every exact Ruff scope, ratchet, and all 39 sentinels pass. All `uv.lock` hashes remain unchanged.

Eldad-Caura:
@claude

arkash20 (approved):

PR #1105, opened 2026-08-30
fix(ratchet): separate annotations from removals

Eldad-Caura:
## Summary

- report added annotations separately from genuinely removed legacy-name lines
- preserve marker exemption semantics and the existing gate decision path
- add the same split to --report, including added lines and excused moves so the net reconciles

The passing summary uses explicit gross quantities: added, annotated, removed, and excused moves, followed by the existing net change. This keeps real work distinct from marker-only edits while making every count reconcilable. Detailed exemption churn remains visible and marker-kind swaps remain zero change.

## Regression

Measured from non-shallow commit 00abbf305a3c38317cd3fae117158e0955938bb7 to 1304e3ee0a06608e12f98c40530d6b7b3e1880e7:

    No new lines. 26 annotated, 0 removed, 0 excused moves (-26 net).

The new assertions were run against the old script first and failed before the implementation.

## Verification

- 86 ratchet tests passed
- Ruff check passed at CI scopes
- Ruff format --check passed at CI scopes
- mypy passed for scripts/
- Python byte-compilation passed
- legacy-name ratchet passed against origin/main
- do-not-touch sentinel: all 39 protected strings survive
- one DCO-signed commit

### DIFF
--- a/scripts/legacy_name_ratchet.py
+++ b/scripts/legacy_name_ratchet.py
@@ -99,7 +99,7 @@ Usage::
 
     scripts/legacy_name_ratchet.py                     # compare against origin/main
     scripts/legacy_name_ratchet.py --base <tree-ish>
-    scripts/legacy_name_ratchet.py --report            # counts only, never fails
+    scripts/legacy_name_ratchet.py --report            # counts + change split; never fails
 """
 
 from __future__ import annotations
@@ -110,7 +110,7 @@ import re
 import subprocess
 import sys
 from collections import Counter
-from typing import NamedTuple
+from typing import NamedTuple, overload
 
 # The legacy brand. Matched case-insensitively, so every casing in the tree
 # counts — prose, identifiers and SCREAMING_SNAKE env names alike. This is the
@@ -330,7 +330,7 @@ def _git(args: list[str]) -> str:
 
 
 def _grep(
-    tree: str | None, pathspec: str = ":/"
+    tree: str | None, pathspec: str | list[str] = ":/"
 ) -> list[tuple[str, str, str, str | None]]:
     """``(path, lineno, text, kind)`` for every matching line.
 
@@ -357,7 +357,8 @@ def _grep(
     args = ["git", "grep", "-I", "-i", "-n", "-z", "--full-name", "-e", LEGACY_NAME]
     if tree is not None:
         args.append(tree)
-    args += ["--", pathspec]
+    args.append("--")
+    args += [pathspec] if isinstance(pathspec, str) else pathspec
 
     prefix = f"{tree}:" if tree is not None else ""
 
@@ -520,7 +521,7 @@ class Scan(NamedTuple):
         )
 
 
-def scan(tree: str | None) -> Scan:
+def scan(tree: str | None, pathspec: str | list[str] = ":/") -> Scan:
     """Read a tree once. ``tree=None`` is the working tree."""
     by_file: dict[str, Counter[str]] = {}
     total: Counter[str] = Counter()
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
Applied the review finding. --report now computes its change summary and returns before mint-budget analysis, while the gate computes the history-aware summary only inside the passing branch that prints it. Failing gates therefore keep the cheap path.

Verification after re-sync and rebase: 86 ratchet tests, exact Ruff check and Ruff format --check scopes, mypy, byte-compilation, ratchet, sentinel, and the 00abbf3..1304e3ee regression all pass. The simplify test audit compared 128 old/new process scenarios with identical output and exit codes.

Eldad-Caura:
@Codex

Eldad-Caura:
@claude

Eldad-Caura:
Applied the second review finding without weakening the required historical result. A clean one-commit range—the ordinary PR checkout—now reuses the already-collected endpoint scans; dirty working trees and multi-commit ranges retain chronological replay.

The new performance regression fails on the previous head with 3 git-grep calls and passes now with 2. A second guard regression fails if the shortcut is incorrectly applied to a dirty worktree. The multi-commit replay remains necessary: direct endpoints for 00abbf3..1304e3ee classify 25 annotated / 1 removed, while chronological replay produces the required 26 annotated / 0 removed.

Verification on the current base: 88 ratchet tests, exact Ruff check and Ruff format --check scopes, mypy, byte-compilation, ratchet, sentinel (46/46), and both historical gate/report runs pass.

Eldad-Caura:
@claude

Eldad-Caura:
Applied the third review finding. Ordinary gate runs now cap first-parent replay at 64 commits and cap the rev-list enumeration itself at 65 entries. At 65+, the gate keeps exit 0 but states that the split was omitted and directs the caller to uncapped --report; it does not print an approximate net. The redundant clean-working-tree transition is also gone.

Regression coverage pins both sides: 64 commits still reports the exact 1 annotated / 0 removed split; 65 commits omits it in the gate; --report at 65 remains exact. This test fails before the cap.

Verification after the latest rebase: 89 ratchet tests, exact Ruff check and Ruff format --check scopes, mypy, byte-compilation, ratchet, sentinel (46/46), and the required 00abbf3..1304e3ee gate/report runs all pass.

Eldad-Caura:
@claude

Eldad-Caura:
Addressed the latest review in a81c6c1e:

- Replaced per-commit whole-tree replay with exact incremental replay. Each first-parent step now rescans only its changed paths; a one-nonempty-transition range reuses the two endpoint scans. Trace regressions pin 2 whole-tree greps for both optimized and replayed multi-commit cases, with replay greps path-limited.
- Preserved chronological gross accounting for transient add/remove churn, remove-then-readd-with-marker, pure renames, dirty worktrees, and text/binary transitions.
- Replaced the optimized-away assert with overloads that make uncapped report mode non-optional to the type checker.
- Rewrote the contiguous insertion predicate as a linear prefix/suffix check; it matched the prior predicate across 16,129 exhaustive small-string pairs.

Verification: 94 ratchet tests; exact CI Ruff check and both format-check scopes; scripts mypy; py_compile; sentinel (46/46); current gate against non-shallow a5a5e492. Historical non-shallow replay at 1304e3ee from 00abbf3 reports: No new lines. 26 annotated, 0 removed, 0 excused moves (-26 net).

Eldad-Caura:
@claude

Eldad-Caura:
@erni-a

erni-a (approved):

PR #1108, opened 2026-08-30
feat: expose lifecycle audit observation for active probes

Eldad-Caura:
## Diagnosis

The lifecycle topic family cut over from `memclaw.lifecycle.*` to `caura.lifecycle.*` on 28 August. The existing smoke probes continued naming only the legacy family and asserted durable subscription state/backlog rather than correlating a current publish with delivery. Dual-bind kept those legacy subscriptions attached even though both staging and production recorded zero legacy lifecycle publishes.

Every green lifecycle smoke since 28 August was worthless, including the ones that gated production promotes: those runs verified a retained legacy bridge, not current product behaviour.

This PR supplies the exact observation surface required for active, message-correlated probes in the coordinated enterprise and test-automation PRs.

## Changes

- add an admin-only exact audit lookup, bound by both audit ID and organization ID and forced to the writer path for read-after-write polling
- add an uncapped, bounded-window audit aggregate for all seven lifecycle actions, with an explicit internal unscoped sentinel rather than an accidental default
- add a concurrent recency index for the append-only lifecycle audit table
- cover authorization, tenant binding, writer/reader routing, missing rows, malformed scope requests, uncapped failures, trigger filtering, and concrete-org filtering

## Validation

- 166 related unit/regression tests
- 7 PostgreSQL-backed observation integration tests
- writer-session routing test
- tenant-scope ratchet against `origin/main`
- Ruff and Alembic head (`041`)

## Coordination

This is the API prerequisite for the matching `caura-enterprise` and `caura-test-automation` probe PRs. It must be deployed before either new probe can pass.

### DIFF
--- a/common/models/lifecycle_audit.py
+++ b/common/models/lifecycle_audit.py
@@ -24,6 +24,14 @@ from common.models.base import Base
 class LifecycleAudit(Base):
     __tablename__ = "lifecycle_audit"
     __table_args__ = (
+        # Supports cross-org recent-window health summaries. The older
+        # org/action-leading index cannot serve a predicate on started_at
+        # alone, so the smoke endpoint would otherwise scan the full audit
+        # history as this append-only table grows.
+        Index(
+            "idx_lifecycle_audit_started_at",
+            text("started_at DESC"),
+        ),
         Index(
             "idx_lifecycle_audit_org_action_started",
             "org_id",
--- a/core-api/src/core_api/clients/storage_client.py
+++ b/core-api/src/core_api/clients/storage_client.py
@@ -2579,6 +2579,44 @@ class CoreStorageClient:
         )
         return result["audit_id"]  # type: ignore[index]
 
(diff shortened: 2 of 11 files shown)


Eldad-Caura:
Review follow-up: declined as false.

Unscoped and UNSCOPED are existing top-level definitions in postgres_service.py at lines 636 and 664 on both current main and this PR head: https://github.com/caura-ai/caura/blob/c29f914c5c1229e2920ccaf2db668758e44af39f/core-storage-api/src/core_storage_api/services/postgres_service.py#L636-L665

The same sentinel is already consumed by the fleet router and service. I also directly imported core_storage_api.routers.lifecycle_audit, imported UNSCOPED and Unscoped from postgres_service, and asserted that the router references the same singleton; the import assertion passed. The review appears to have inspected the new summary method without finding the existing module-level definition roughly 8,800 lines earlier.

No code change is warranted for this finding.

arkash20 (approved):

PR #1111, opened 2026-08-30
fix(fleet): scope the command ack to the caller's tenant

### LINKED ISSUE #1086: storage: POST /fleet/commands/ack acks any tenant's fleet commands by UUID
`route:POST /api/v1/storage/fleet/commands/ack`

Service method: `method:fleet_ack_commands`

### The defect

```python
command_ids = [UUID(cid) if isinstance(cid, str) else cid for cid in body["command_ids"]]
await _svc.fleet_ack_commands(command_ids=command_ids, now=datetime.now(UTC))
```

A list of ids, no tenant, and the UPDATE sets `status='acked'`. This is the same subsystem as GHSA-xw4x-jwf5-8m9h (caura-ai/caura#1067), which was the fleet command path validating the tenant from the body but never checking the node belonged to it.

### The fix

Require `tenant_id` and add it to the `WHERE` in `fleet_ack_commands`, so acking is confined to the caller's own commands. Given the sibling advisory in this exact subsystem, this is the one I would fix first.

---

### Why this is filed

Eldad-Caura:
Closes #1086.

`POST /fleet/commands/ack` took a bare list of command UUIDs and acked every row that matched, whichever tenant owned it:

```python
async def ack_commands(request: Request) -> dict:
    command_ids = [UUID(cid) if isinstance(cid, str) else cid for cid in body["command_ids"]]
    await _svc.fleet_ack_commands(command_ids=command_ids, now=datetime.now(UTC))
```

No tenant is read. Beneath it, `fleet_ack_commands` ran `UPDATE fleet_commands SET status='acked', acked_at=... WHERE id IN (...)` with no tenant predicate. Storage authenticates nothing (GHSA-wgvw-28pq-jc36), so knowing a command UUID was enough to ack another tenant's command out from under its plugin.

### The third instance of one shape

- **GHSA-xw4x-jwf5-8m9h** — the *create* path. Validated the caller-named tenant, never checked `body.node_id` belonged to it. Fixed in `6db0ae57`.
- **`PATCH /commands/{id}/status`** — the *complete* path. `tenant_id` optional on the wire, `None` meaning no `WHERE`. Fixed in caura-ai/caura#1067.
- **This** — the *ack* path.

### The fix

`fleet_commands.tenant_id` is `NOT NULL` (migration `001_initial_schema.py:244`), so the predicate goes straight into the statement. No cross-table node/tenant check was needed, unlike the create path.

**Required and plainly `str`, not the `Unscoped` sentinel its sibling takes.** That sentinel exists because core-api genuinely serializes an admin case as `{"tenant_id": null}` when *completing* a command. Ack has no such caller: its one call path is the heartbeat, whose `tenant_id` is a non-optional `str` on `HeartbeatIn`, and which **already had the tenant in hand** — it fetched the very same commands with it one line earlier:

```python
commands = await sc.get_pending_commands(body.tenant_id, node_name)
...
await sc.ack_commands([c.get("id") for c in commands])   # ← tenant dropped here
```

An opt-out nobody needs is just an unlocked door, so `null` is a 422 here rather than an admin escape. The route uses the shared `_require` guard, which is also why the gate recognises it without a `blind-spot` entry.

`count` now reports rows actually matched rather than rows asked for, so acking a command that is not yours reports 0 instead of succeeding silently. That is not an existence oracle — a foreign UUID and a nonexistent one both return 0.

### Tests

Six, in `core-storage-api/tests/test_fleet_ack_tenant_scope.py`. **Four fail on the parent commit:**

| test | on parent |
|---|---|
| `test_another_tenants_command_is_not_acked` | ✅ **fails** — the attack |
| `test_omitted_tenant_id_is_rejected` | ✅ fails |
| `test_a_null_tenant_id_is_rejected_too` | ✅ fails — no admin opt-out here |
| `test_a_mixed_batch_acks_only_the_callers_own` | ✅ fails |
| `test_own_command_is_acked` | passes — regression guard |
| `test_a_foreign_id_is_indistinguishable_from_a_missing_one` | passes — only meaningful once scoped |

The mixed-batch one is the case that argues for the predicate living in SQL rather than in a route-level "do these all belong to you?" check: a partially-foreign batch should ack the caller's own and leave the rest, which a wholesale rejection would not do.

### Worth flagging

**Nothing tested `ack_commands` at all** — no test in either package referenced it before this PR. The client signature change is guarded by mypy (positional, non-defaulted) rather than by any existing test, which is why the storage-side HTTP contract is what I pinned.

### Effect on the gate

```
Allowlist shrank by 2: method:fleet_ack_commands, route:POST /api/v1/storage/fleet/commands/ack
tenant-scope gate: 189 routes + 190 service methods, 263 bound to a tenant, 116 allowlisted.
```

`id-addressed-write` 20 → 18, total 118 → 116, bound 261 → 263.

### Verification

`core-storage-api`: 10 failed / **200** passed / 7 errors against a 10 / **194** / 7 baseline — delta is exactly the six new tests. Root suite 46 failed / 5605 passed, the same pre-existing 46 as every recent PR. `ruff check` and `ruff format --check` at CI's scopes across `core-api/src`, `core-storage-api/src`, `core-storage-api/tests` — clean at ruff 0.16.5. `mypy src/` clean for both packages (203 and 84 files) and for `scripts/` (31). Gate `--base origin/main` exit 0, `--write` produces no drift. `uv.lock` untouched. Rebased onto `853b0205` and re-run there.

### DIFF
--- a/core-api/src/core_api/clients/storage_client.py
+++ b/core-api/src/core_api/clients/storage_client.py
@@ -2348,8 +2348,18 @@ class CoreStorageClient:
             node_name=node_name,
         )
 
-    async def ack_commands(self, command_ids: list[str]) -> dict:
-        return await self._post("/fleet/commands/ack", {"command_ids": command_ids})  # type: ignore[return-value]
+    async def ack_commands(self, command_ids: list[str], tenant_id: str) -> dict:
+        """Ack commands for one tenant. ``tenant_id`` is required, not optional.
+
+        Storage scopes the UPDATE by it; without it the ack keyed on command
+        UUIDs alone and reached other tenants' rows. Positional and
+        non-defaulted so a caller cannot omit it and silently get the old
+        behaviour back.
+        """
+        return await self._post(  # type: ignore[return-value]
+            "/fleet/commands/ack",
+            {"command_ids": command_ids, "tenant_id": tenant_id},
+        )
 
     async def fleet_exists(self, tenant_id: str, fleet_id: str) -> bool:
         result = await self._get(
--- a/core-api/src/core_api/routes/fleet.py
+++ b/core-api/src/core_api/routes/fleet.py
@@ -849,7 +849,7 @@ async def heartbeat(
         commands = pending
 
     if commands:
-        await sc.ack_commands([c.get("id") for c in commands])
+        await sc.ack_commands([c.get("id") for c in commands], body.tenant_id)
 
     return {
         "ok": True,
--- a/core-storage-api/src/core_storage_api/routers/fleet.py
+++ b/core-storage-api/src/core_storage_api/routers/fleet.py
@@ -7,6 +7,7 @@ from uuid import UUID
 
 from fastapi import APIRouter, HTTPException, Request
 
+from core_storage_api.routers._validation import _require
 from core_storage_api.schemas import FLEET_COMMAND_FIELDS, FLEET_NODE_FIELDS, orm_to_dict
 from core_storage_api.services.postgres_service import UNSCOPED, PostgresService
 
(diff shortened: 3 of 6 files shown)


Eldad-Caura:
## CI failed on mypy — fixed in `4e44ed0e`, and the reason I missed it matters more than the fix

```
src/core_storage_api/services/postgres_service.py:9080:
error: "Result[Any]" has no attribute "rowcount"  [attr-defined]
```

`session.execute()` on an UPDATE is typed `Result[Any]`; `rowcount` lives on `CursorResult`. The file already has ~12 of these and every one carries `# type: ignore[attr-defined]`. Mine now does too — following the established convention rather than inventing a cast.

### Why my local mypy said "Success"

This is the part worth recording. My lint venv had **mypy but not the project's dependencies**, so `sqlalchemy` resolved to `Any` and the attribute check was vacuous — mypy cheerfully accepted `.rowcount` on an untyped object.

Demonstrated by reverting the fix and running both environments against it:

```
# venv WITH core-storage-api[dev] installed  →  reproduces CI exactly
src/core_storage_api/services/postgres_service.py:9080: error: "Result[Any]" has no attribute "rowcount"
Found 1 error in 1 file (checked 84 source files)

# venv with mypy alone  →  the false green
Success: no issues found in 84 source files
```

Same mypy version, same config, same 84 files. The only difference is whether the types being checked against are real.

That means every core-package mypy run I reported this session was weaker than CI's, and any clean result I quoted for `core-api` or `core-storage-api` was worth less than it sounded. The `scripts/` runs are unaffected — `scripts/mypy.ini` sets `ignore_missing_imports` and `follow_imports = silent` deliberately, so that one was always checking what it claimed to.

Rebuilt the environment with `core-storage-api[dev]` and `core-api[dev]` installed editable, `UV_FROZEN=1`, both `uv.lock` files hash-identical before and after and absent from `git status`.

### Re-verified against real types

| check | result |
|---|---|
| `cd core-storage-api && mypy src/` | clean, 84 files |
| `cd core-api && mypy src/` | clean, 203 files |
| `mypy --config-file scripts/mypy.ini scripts/` | clean, 31 files |
| `core-storage-api` suite | 10 failed / 200 passed / 7 errors — unchanged, delta is the 6 new tests |
| `ruff check` + `ruff format --check`, CI scopes | clean, 310 files |
| tenant-scope gate `--base origin/main` | exit 0 |

No behaviour changed: the only edit since the review verdict is a trailing `# type: ignore[attr-defined]` comment.

Eldad-Caura:
@claude

erni-a (approved):

PR #1107, opened 2026-08-30
docs(readme): show cross-agent fleet recall

Eldad-Caura:
## Summary

- add a concrete two-agent MCP walkthrough directly after the keyless quick start
- show one agent writing a team-visible operational lesson and another agent recalling it from the same fleet
- explain the `scope_agent`, `scope_team`, and `scope_org` governance boundaries and link to credential/trust setup

## Validation

- `pre-commit run --files README.md`
- `python3 scripts/legacy_name_ratchet.py --base origin/main`
- `PYTHONPATH=core-api/src:core-storage-api/src:. uv run pytest tests/test_mcp_write.py tests/test_mcp_recall.py tests/test_tools_registry.py -q` (49 passed)
- `git diff --check`

PR #1100, opened 2026-08-30
docs(readme): split reference material

Eldad-Caura:
## Summary

Split the long README into focused operator/reference guides and reduce the README from 1,366 to 702 lines. The executable keyless quick start remains in place, while self-hosting, v1 migration, API reference, and public stability details move under `docs/` with legacy anchors preserved.

This also consolidates the duplicate telemetry sections and corrects stale inherited guidance around v2 migration/backfill commands, storage configuration, auth headers, bootstrap aliases, production startup requirements, and OpenAPI URLs.

## Type of Change

- [x] Documentation update
- [x] Refactor / internal cleanup

## How Has This Been Tested?

- `pre-commit run --files README.md AGENT-INSTALL.md CONTRIBUTING.md SUPPORT.md docs/local-embedder.md docs/api-reference.md docs/public-api-stability.md docs/self-hosting.md docs/upgrading-from-v1.md`
- `pytest tests/test_readme_quickstart.py -q` — 10 passed
- `python3 scripts/legacy_name_ratchet.py --base origin/main`
- `docker compose config --quiet`
- Local Markdown path and anchor validation across all changed docs
- Current FastAPI route/OpenAPI introspection for documented paths

## Checklist

- [x] I have read `CONTRIBUTING.md`
- [x] I have added tests that cover my changes (existing README contract tests cover the retained executable block; this change adds no runtime behavior)
- [x] `ruff check` and `ruff format --check` pass (no Python files changed; pre-commit passed)
- [x] `mypy` passes (not applicable; no typed source changed)
- [x] `pytest` passes locally
- [x] I have updated relevant documentation
- [ ] I have updated `CHANGELOG.md` under the `Unreleased` section (not applicable: documentation reorganization only)

## Additional Notes

The `/simplify` review completed with no remaining findings. Follow-up work will improve the demo's cross-agent/fleet story and verify remaining published-image/provider parity separately so this PR stays focused on information architecture.

PR #1112, opened 2026-08-30
test(readme): require synchronous title enrichment

Eldad-Caura:
## Summary

- make the documented quickstart write use strong mode so its response contains the title and summary the following prose promises
- make the executable README check reject a missing, null, empty, or whitespace-only title
- pin both the strong-mode payload and the title contract with regression tests

## Why

The default fast write returns title: null with metadata.enrichment_pending: true. The README nevertheless described an immediately generated title. Strong mode is the existing public contract for synchronous enrichment, regardless of whether the configured provider is real or the keyless heuristic fallback.

The freshly pulled published latest image was also checked: the old payload reproduced title: null, while the strong payload returned a title synchronously. The full modified README write-and-search block then passed against a clean published-image stack. The same published image now accepts an omitted agent_id in standalone mode, confirming that the source/image drift from the original audit has been released.

## Validation

- red before fix: the payload regression test failed because write_mode was absent; the title regression test failed because the checker accepted title: null
- uv run pytest tests/test_readme_quickstart.py tests/test_p4_1_llm_fallback.py tests/test_write_mode_dispatch.py -q — 58 passed
- uv run ruff check scripts/check_readme_quickstart.py tests/test_readme_quickstart.py
- uv run ruff format --check scripts/check_readme_quickstart.py tests/test_readme_quickstart.py
- uv run mypy --config-file scripts/mypy.ini scripts/check_readme_quickstart.py
- pre-commit run --files README.md scripts/check_readme_quickstart.py tests/test_readme_quickstart.py
- python3 scripts/legacy_name_ratchet.py --base origin/main — no new lines
- python3 scripts/do_not_touch_sentinel.py — all 46 protected strings survive
- clean published-image run: python3 scripts/check_readme_quickstart.py --base-url http://localhost:18003
- /simplify — no findings

Final follow-up to #1076, #1098, #1100, and #1107.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ docker compose up -d --wait                                 # Postgres + pgvecto
 # Write a memory — no API key needed
 curl -X POST http://localhost:8000/api/v1/memories \
   -H "X-API-Key: standalone" -H "Content-Type: application/json" \
-  -d '{"tenant_id": "default", "agent_id": "quickstart", "content": "Our auth service uses JWT with 15-minute expiry."}'
+  -d '{"tenant_id": "default", "agent_id": "quickstart", "write_mode": "strong", "content": "Our auth service uses JWT with 15-minute expiry."}'
 
 # Find it by keyword — no provider key needed
 curl -X POST http://localhost:8000/api/v1/search \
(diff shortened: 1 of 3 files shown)


Eldad-Caura:
@erni-a

erni-a (approved):

PR #1114, opened 2026-08-30
docs(rebrand): record release-please ratchet scope

Eldad-Caura:
## What

Record `_release_please_branch()` as an intentional repo-specific exception rather than
a missing ratchet port.

## Verification

Measured the complete nine-repository estate on 2026-08-30: eight live repositories plus
archived `loadtest` (read-only). I checked each repository's real default branch for a
release-please workflow and config, fetched current remote branches, searched full local
history for past workflow/config files and `release-please-action`, and searched GitHub's
complete PR history for `head:release-please--`.

| repository | workflow | config | current or historical `release-please--*` PR branch |
| --- | :---: | :---: | :---: |
| `caura` | yes | yes | yes (47 PRs) |
| `caura-enterprise` | yes | yes | yes (31 PRs) |
| `caura-daemon` | no | no | no |
| `caura-test-automation` | no | no | no |
| `caura-onprem` | no | no | no |
| `caura-ops` | no | no | no |
| `caura-onprem-installer` | no | no | no |
| `openclaw-fleet-tester` | no | no | no |
| `loadtest` (archived) | no | no | no |

The helper is live only in the two repositories where release-please can produce a
generated changelog PR. Porting it to the other seven would be dead code. The note also
states the three-signal recheck required if another repository adopts release automation.

## Gates

- `python3 scripts/legacy_name_ratchet.py --base origin/main`
- `python3 scripts/do_not_touch_sentinel.py`
- `ruff check` and `ruff format --check` separately at every scope in `ci.yml`
- all four `uv.lock` hashes unchanged

Documentation only; no gate behavior changes.

PR #1115, opened 2026-08-30
fix(memory): scope the single-row delete to the caller's tenant

### LINKED ISSUE #1080: storage: DELETE /memories/{memory_id} soft-deletes any tenant's memory by UUID
`route:DELETE /api/v1/storage/memories/{memory_id}`

Service method: `method:memory_soft_delete`

### The defect

The handler is two lines and reads no tenant at all:

```python
@router.delete("/{memory_id}")
async def soft_delete_memory(memory_id: UUID) -> dict:
    await _svc.memory_soft_delete(memory_id)
    return {"ok": True}
```

`memory_soft_delete` then does `session.get(Memory, memory_id)` and stamps `deleted_at`/`status` with no tenant predicate. Any caller who can reach the port can delete any memory whose UUID they hold or guess.

### The fix

Take `tenant_id` on the route (required, so a caller who omits it gets a 422 rather than an unscoped delete) and push the predicate into the statement in `memory_soft_delete` — the same shape as caura-ai/caura#1075, which fixed the read side of this exact row. A foreign id must 404 identically to a missing one so the endpoint is not an existence oracle.

---

### Why this is filed

Eldad-Caura:
Closes #1080.

`DELETE /memories/{memory_id}` read no tenant and soft-deleted by primary key:

```python
@router.delete("/{memory_id}")
async def soft_delete_memory(memory_id: UUID) -> dict:
    await _svc.memory_soft_delete(memory_id)
    return {"ok": True}
```

Two lines, no tenant anywhere. `core-storage-api` authenticates nothing (GHSA-wgvw-28pq-jc36), so knowing a UUID was enough to destroy another tenant's memory.

This is the **write form of the same defect** `GET /memories/{memory_id}` had — fixed in caura-ai/caura#1075. Same row, same addressing, destruction instead of disclosure. The route now takes a required `tenant_id` query parameter, matching that GET twin, and 404s for "no such memory" and "not yours" alike so it is not an existence oracle.

### `memory_soft_delete` is deleted, not scoped

`memory_soft_delete_by_ids(tenant_id, ids)` was **already** tenant-scoped and already backs the three bulk-delete routes. Keeping a second, unscoped single-row path to the same write is what let this one drift, and #1075 resolved the read side the same way by deleting `memory_get_by_id`.

Checked equivalence before switching, since the old path mutated an ORM object and the replacement issues a bulk `UPDATE`:

- No ORM event listeners anywhere in `core-storage-api` (`event.listen` / `@listens_for` / `before_update` / `after_update`: none).
- The `memories_search_vector_trigger` fires `BEFORE INSERT OR UPDATE OF content, title` — columns a soft delete does not touch.

### One deliberate behaviour change

`memory_soft_delete_by_ids` filters `deleted_at IS NULL`. The old single-row path did not, so a repeat DELETE **re-stamped `deleted_at` and pushed the retention/purge window forward**. A second delete is now a 404. `test_deleting_twice_does_not_move_the_retention_clock` pins it so a future single-row variant cannot quietly bring it back.

Safe for both callers: `memory_service.soft_delete_memory` already 404s beforehand via its scoped `get_memory`, and `_delete` in the storage client maps 404 to `False` rather than raising, so `governance_remediation` sees a clean no-op.

### Tests

Five new, in `core-storage-api/tests/test_memory_delete_tenant_scope.py`. **Four fail on the parent commit:**

| test | on parent |
|---|---|
| `test_another_tenants_memory_is_not_deleted` | ✅ **fails** — the attack |
| `test_omitted_tenant_id_is_rejected` | ✅ fails |
| `test_a_foreign_id_is_indistinguishable_from_a_missing_one` | ✅ fails |
| `test_deleting_twice_does_not_move_the_retention_clock` | ✅ fails |
| `test_own_memory_is_deleted` | passes — regression guard |

### Existing tests updated, not worked around

The contract changed, so tests written against the old one had to move. Eleven of them, and I checked each rather than blanket-patching:

- **3 in `core-storage-api/tests/test_integration.py`** (4 call sites) — deleted without a tenant. Now pass the fixture tenant.
- **8 governance tests** across `test_governance_remediation.py`, `test_governance_consumer.py`, `test_governance_inline_fast_remediation.py` — fake storage clients needed the new argument, and the `assert_awaited_once_with` calls now **pin the tenant that gets passed**, which is the property this PR is about. `("soft_delete", "m1")` became `("soft_delete", "m1", "t1")`.

### Effect on the gate

```
Allowlist shrank by 2: method:memory_soft_delete, route:DELETE /api/v1/storage/memories/{memory_id}
tenant-scope gate: 191 routes + 191 service methods, 268 bound to a tenant, 114 allowlisted.
```

`id-addressed-write` 18 → 16, total 116 → 114.

### Verification

`core-storage-api`: 10 failed / **213** passed / 7 errors — the same 10 failures as pristine `origin/main`, delta is the new tests. Root suite compared list-against-list with a clean `origin/main` checkout rather than by count, because that suite flakes between 44 and 47: the only difference was `test_semantic_dedup_query_latency`, which passes 3/3 in isolation here and which this diff does not touch (no dedup or query path is modified).

`ruff check` + `ruff format --check` at CI's scopes, clean at 0.16.5. `mypy src/` clean for both packages (85 and 203 files) and `scripts/` (31) — run in a venv with both packages installed editable, so sqlalchemy types are real rather than `Any`. Gate `--base origin/main` exit 0, `--write` no drift. `uv.lock` untouched. Rebased onto `87a68a12` and re-run there.

### DIFF
--- a/core-api/src/core_api/clients/storage_client.py
+++ b/core-api/src/core_api/clients/storage_client.py
@@ -597,8 +597,15 @@ class CoreStorageClient:
         # patch. The explicit arg wins over any ``tenant_id`` in ``data``.
         return await self._patch(f"/memories/{memory_id}", {**data, "tenant_id": tenant_id})
 
-    async def soft_delete_memory(self, memory_id: str) -> bool:
-        return await self._delete(f"/memories/{memory_id}")
+    async def soft_delete_memory(self, memory_id: str, tenant_id: str) -> bool:
+        """Soft-delete one memory within ``tenant_id``. False when nothing matched.
+
+        ``tenant_id`` is required, not optional: storage scopes the UPDATE by
+        it, and without it the delete keyed on the memory UUID alone and
+        reached other tenants' rows. Positional and non-defaulted so a caller
+        cannot omit it and quietly get the old behaviour back.
+        """
+        return await self._delete(f"/memories/{memory_id}", tenant_id=tenant_id)
 
     async def set_subject_entity_if_null(
         self, memory_id: str, tenant_id: str, subject_entity_id: str
--- a/core-api/src/core_api/services/governance_remediation.py
+++ b/core-api/src/core_api/services/governance_remediation.py
@@ -94,7 +94,7 @@ async def remediate_after_enrichment(memory: dict, cfg: Any) -> RemediationOutco
                 # audit is the only trace — must survive queue overflow.
                 critical=True,
             )
-            await sc.soft_delete_memory(memory_id)
+            await sc.soft_delete_memory(memory_id, tenant_id)
             logger.info("governance: dropped fast-mode memory %s (pii)", memory_id)
             return RemediationOutcome(dropped=True)
         # mask/flag: the LLM gives no offsets to redact a free-form span, and in
@@ -129,7 +129,7 @@ async def remediate_after_enrichment(memory: dict, cfg: Any) -> RemediationOutco
                 # Destructive: see the PII-drop branch — audit is the only trace.
                 critical=True,
             )
-            await sc.soft_delete_memory(memory_id)
+            await sc.soft_delete_memory(memory_id, tenant_id)
             logger.info("governance: dropped fast-mode memory %s (non-business)", memory_id)
             return RemediationOutcome(dropped=True)
         if nb_cfg.disposition == "keep_private":
--- a/core-api/src/core_api/services/memory_service.py
+++ b/core-api/src/core_api/services/memory_service.py
@@ -3477,7 +3477,7 @@ async def soft_delete_memory(memory_id: UUID, tenant_id: str) -> None:
     if not mem:
         raise HTTPException(status_code=404, detail="Memory not found")
 
-    await sc.soft_delete_memory(str(memory_id))
+    await sc.soft_delete_memory(str(memory_id), tenant_id)
 
     _hooks = get_hooks()
     if _hooks.audit_log:
--- a/core-storage-api/src/core_storage_api/routers/memories.py
+++ b/core-storage-api/src/core_storage_api/routers/memories.py
@@ -1841,6 +1841,26 @@ async def update_memory_entities(memory_id: UUID, request: Request) -> dict:
 
 
 @router.delete("/{memory_id}")
-async def soft_delete_memory(memory_id: UUID) -> dict:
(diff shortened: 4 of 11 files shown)


arkash20 (approved):

Eldad-Caura:
Ready for you, @erni-a — leaving the merge to you.

**What it fixes:** #1080. `DELETE /memories/{memory_id}` took a bare UUID and soft-deleted by primary key with no tenant predicate, so a caller who knew an id could destroy another tenant's memory through a service that authenticates nothing. This is the write form of the read bug fixed in #1075 — same row, addressed the same way, destruction instead of disclosure.

**Approach.** The route now takes a required `tenant_id` and goes through `memory_soft_delete_by_ids`, which was already tenant-scoped and already backs the three bulk-delete routes. I deleted the unscoped `memory_soft_delete` rather than scoping it, since keeping a second single-row path to the same write is what let this one drift. Equivalence checked before switching, because the old path mutated an ORM object while the replacement issues a bulk `UPDATE`: there are no ORM event listeners in the service, and `memories_search_vector_trigger` fires `BEFORE INSERT OR UPDATE OF content, title` — columns a soft delete doesn't touch.

**A second bug surfaced on the way.** The scoped method filters `deleted_at IS NULL`; the old one didn't. So a repeat DELETE was re-stamping `deleted_at` and pushing the retention window forward. A second delete is now a 404, pinned by `test_deleting_twice_does_not_move_the_retention_clock`.

**Existing tests.** Eleven covered the old contract. The eight governance ones now assert the tenant as well as the id, so they pin the security property instead of merely tolerating the new signature.

**Gating signals on the current head `3a71bb31`:**

- **CI** — pass (6m52s).
- **Claude Code review** — "No issues found". It ran on `2e52d30c`; your rebase onto #1114 left the code content identical, changing only `docs/plans/rebrand-sunset-plan.md`, which arrives with the base. The two `claude-review: skipped` entries against this head are the by-design synchronize skips, not an un-run review, so I did not re-trigger it on an unchanged diff.
- **arkash20** — approved this head.

**Tenancy gate:** allowlist shrank by 2 (`method:memory_soft_delete`, `route:DELETE /api/v1/storage/memories/{memory_id}`). `id-addressed-write` 18 → 16, total 116 → 114.

erni-a (approved):

PR #1116, opened 2026-08-30
fix(ratchet): reconcile moves across commits

Eldad-Caura:
## Summary

- reconcile byte-identical moves across commit boundaries in either order (delete-then-add or copy-then-delete)
- count a same-path delete/re-add as add/remove churn before considering later cross-path moves
- fall back to the endpoint classification if an optional replay Git command cannot spawn

This changes reporting only. Both exemption markers still exempt identically, and the gate decision path is unchanged.

## Regression coverage

- both cross-commit move orders now report `0 added, 0 removed, 1 excused move`
- a same-path restoration followed by a copy reports `2 added, 1 removed, 0 excused moves`
- an `OSError` during replay returns the endpoint summary instead of escaping
- each new regression failed against the landed Work Package I implementation before the fix

## Validation

- `python3 -m pytest --noconftest tests/test_legacy_name_ratchet.py -q` — 98 passed
- `uv run ruff check scripts/legacy_name_ratchet.py tests/test_legacy_name_ratchet.py`
- `uv run ruff format --check scripts/legacy_name_ratchet.py tests/test_legacy_name_ratchet.py`
- `uv run mypy --config-file scripts/mypy.ini scripts/` — 31 source files clean
- `python3 -m pytest --noconftest tests/test_do_not_touch_sentinel.py -q` — 111 passed
- non-shallow historical worktree at `1304e3ee`, base `00abbf3`: `No new lines. 26 annotated, 0 removed, 0 excused moves (-26 net).`

`actionlint` still reports four pre-existing findings in untouched workflows.

### DIFF
--- a/scripts/legacy_name_ratchet.py
+++ b/scripts/legacy_name_ratchet.py
@@ -600,8 +600,13 @@ def _only_inserts(before: str, after: str) -> bool:
     return after[mismatch + inserted :] == before[mismatch:]
 
 
-def _transition_summary(base: Scan, head: Scan) -> _ChangeSummary:
-    """Classify one tree-to-tree transition without affecting the gate."""
+_LineCounts = dict[tuple[str, str], int]
+
+
+def _transition_delta(
+    base: Scan, head: Scan
+) -> tuple[_LineCounts, _LineCounts, _LineCounts]:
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@erni-a

erni-a (approved):

Issue #1118, opened 2026-08-30
storage: PATCH /memories/{memory_id} lets the caller repoint the primary key


Eldad-Caura:
`route:PATCH /api/v1/storage/memories/{memory_id}`

Service method: `method:memory_update`

### The defect

`memory_update` builds its `SET` clause by testing caller keys against the model:

```python
values: dict = {
    key: val for key, val in patch.items() if key != "metadata_patch" and hasattr(Memory, key)
}
```

`hasattr(Memory, "id")` is true, so `id` lands in the `UPDATE ... SET`. The
statement becomes `UPDATE memories SET id = <caller's choice> WHERE id = :memory_id
AND tenant_id = :tenant_id` — the tenant predicate is satisfied, and the row's
primary key is rewritten to a value the caller picked.

Verified against `origin/main` (`e3973266`), storage API on a local PG:

```
POST /memories                      -> 200, id = <mid>
PATCH /memories/<mid> {"tenant_id": t, "id": <stolen>}  -> 200
GET  /memories/<mid>?tenant_id=t    -> 404      (row no longer at its old id)
GET  /memories/<stolen>?tenant_id=t -> 200      (row is now at the caller's id)
```

Every `memory_entity_links` row, `supersedes_id` reference, and stored id held by
a client still points at the old value.

`fleet_id`, `agent_id`, `visibility`, and `scope` reach the same loop.

### Why this is filed separately

This is **not** a tenant-scope hole — `memory_update`'s predicate is correct and
was fixed already. It is the second half of the same class of defect: a scoped
predicate is only worth the immutability of the column it filters on. The
identical hole on the entity side was found and closed while fixing #1081, where
`entity_update` had `if hasattr(entity, key): setattr(entity, key, value)`; that
route now filters through an explicit `_ENTITY_UPDATABLE_FIELDS` allowlist, and
`test_scope_columns_are_not_caller_writable` pins it.

### The fix

Replace the `hasattr(Memory, key)` test with an explicit writable-column
allowlist, the way `entity_update` now does. `id` and `tenant_id` are never
caller-writable; `fleet_id`/`agent_id`/`visibility`/`scope` should be listed
deliberately or excluded.

### Note on the gate

`scripts/tenant_scope_gate.py` checks **tenant binding only**, so neither this
nor a regression of the entity version would be caught. Worth deciding whether
scope-column writability belongs in the gate as a second check — otherwise the
ratchet enforces half the invariant while the campaign hand-fixes the other half.

PR #1119, opened 2026-08-30
fix(entity): scope the single-row update to the caller's tenant

### LINKED ISSUE #1081: storage: PATCH /entities/{entity_id} updates any tenant's entity by UUID
`route:PATCH /api/v1/storage/entities/{entity_id}`

Service method: `method:entity_update`

### The defect

```python
@router.patch("/{entity_id}")
async def update_entity(entity_id: UUID, request: Request) -> dict:
    entity = await _svc.entity_update(entity_id, body)
```

No tenant is read from the request. `entity_update` fetches by primary key and flushes caller-supplied fields.

### The fix

Require `tenant_id` on the route and add the predicate to the fetch inside `entity_update`, so a foreign entity is indistinguishable from a missing one.

---

### Why this is filed

Eldad-Caura:
Closes #1081.

`PATCH /entities/{entity_id}` took a bare UUID and `entity_update` fetched by primary key, so a caller who knew an id could rewrite another tenant's entity through a service that authenticates nothing (GHSA-wgvw-28pq-jc36). This is the write form of the read hole fixed in #1075.

### The fix

The route takes `tenant_id` in the body and removes it before the rest becomes the patch — the shape `PATCH /memories/{memory_id}` already uses — and `entity_update` carries the predicate into its fetch, matching the one `entity_bulk_upsert`'s update branch already had. A foreign id and a missing one both answer 404, so the endpoint is not an existence oracle for entity UUIDs.

### Scoping the fetch alone would not have closed it

The body loop was `if hasattr(entity, key): setattr(entity, key, value)`, which reaches every mapped column. A predicate is worth exactly as much as the immutability of the column it filters on, and this one filtered on a column the same request could rewrite. Both holes are demonstrated, not inferred — against unmodified `main`:

| request | before |
|---|---|
| `PATCH` another tenant's entity | `200`, victim's row rewritten |
| `PATCH {"tenant_id": victim}` on your own row | `200`, row now in the victim's namespace |
| `PATCH {"id": <chosen>}` | `200`, old id `404`s, row lives at the caller's id |

Writable columns are now `_ENTITY_UPDATABLE_FIELDS`. It is enforced in the service rather than the router because `entity_bulk_upsert` reaches this method without passing through a route; a router-side filter would have left that path on the old behaviour.

### What the tests can and cannot pin

Two independent mechanisms block the tenant case: the route's removal of the key, and the allowlist. I checked this rather than assuming — widening the allowlist to include `tenant_id` leaves the suite green, and so does replacing the removal with a plain read. Either alone closes it, so **no route-level test can isolate one**, and I've said so in the module docstring rather than letting a test's name imply coverage it doesn't have. The primary key is the case reachable through the allowlist alone, and `test_scope_columns_are_not_caller_writable` pins that.

4 of the 5 tests fail on the parent commit. The fifth is the happy-path regression guard, which should pass either way.

### Follow-up filed, not fixed here

`memory_update` has a correct tenant predicate and the same primary-key hole — `PATCH /memories/{memory_id}` with `{"id": <chosen>}` returns 200, the old id 404s, and the row lives at the caller's id. Verified on `main` and filed as #1118 rather than folded in: different route, different issue.

That issue also raises a question worth your call — the gate checks tenant binding only, so neither that hole nor a regression of this one would be caught. Whether scope-column writability belongs in the gate is a design decision, not a fix I should make unilaterally.

### Gate

Both allowlist entries deleted; ratchet holds. `id-addressed-write` 16 → 14, total 114 → 112, bound-to-tenant 268 → 270.

### Verification

Storage suite 228 passed; root suite 5692 passed / 4 skipped / 1 xfailed, zero failures. ruff check + `format --check` clean at all six CI scopes (0.16.5). mypy clean across core-api (203), core-storage-api (85), core-operations (5), core-worker (11), common (99), scripts (31). Rebased onto `e3973266` and re-run there. `uv.lock` unchanged across all four lockfiles.

### DIFF
--- a/core-api/src/core_api/clients/storage_client.py
+++ b/core-api/src/core_api/clients/storage_client.py
@@ -1317,8 +1317,10 @@ class CoreStorageClient:
     async def get_entity(self, entity_id: str) -> dict | None:
         return await self._get(f"/entities/{entity_id}")
 
-    async def update_entity(self, entity_id: str, data: dict) -> dict | None:
-        return await self._patch(f"/entities/{entity_id}", data)
+    async def update_entity(self, entity_id: str, tenant_id: str, data: dict) -> dict | None:
+        # Same contract as ``update_memory`` above, for the same reason.
+        # The explicit arg wins over any ``tenant_id`` in ``data``.
+        return await self._patch(f"/entities/{entity_id}", {**data, "tenant_id": tenant_id})
 
     async def find_exact_entity(
         self,
--- a/core-api/src/core_api/services/entity_service.py
+++ b/core-api/src/core_api/services/entity_service.py
@@ -104,7 +104,7 @@ async def upsert_entity(
         if name_embedding is not None:
             update_data["name_embedding"] = name_embedding
 
-        updated = await sc.update_entity(str(entity_id), update_data)
+        updated = await sc.update_entity(str(entity_id), data.tenant_id, update_data)
         entity = updated or entity
     else:
         # Create new entity. Coerce ``attributes=None`` to ``{}`` so
--- a/core-storage-api/src/core_storage_api/routers/entities.py
+++ b/core-storage-api/src/core_storage_api/routers/entities.py
@@ -645,7 +645,14 @@ async def get_entity(entity_id: UUID) -> dict:
 @router.patch("/{entity_id}")
 async def update_entity(entity_id: UUID, request: Request) -> dict:
     body: dict = await request.json()
-    entity = await _svc.entity_update(entity_id, body)
+    # Tenant guard, same shape as ``PATCH /memories/{memory_id}``: ``tenant_id``
+    # is the row's home tenant, removed from the body so it scopes the fetch
+    # rather than landing as a patched column (``Entity`` has a ``tenant_id``
+    # column). ``_ENTITY_UPDATABLE_FIELDS`` would drop it anyway; both are kept
+    # because either alone closes the hop and they fail independently.
+    tenant_id = _require(body, "tenant_id")
+    del body["tenant_id"]
+    entity = await _svc.entity_update(entity_id, tenant_id, body)
     if entity is None:
         raise HTTPException(status_code=404, detail="Entity not found")
     return orm_to_dict(entity, ENTITY_FIELDS)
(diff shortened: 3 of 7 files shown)


Eldad-Caura:
Ready for you, @erni-a — leaving the merge to you.

**What it fixes:** #1081. `PATCH /entities/{entity_id}` took a bare UUID and `entity_update` fetched by primary key, so a caller who knew an id could rewrite another tenant's entity through a service that authenticates nothing. Write form of the read hole fixed in #1075.

**Beyond the issue text.** Scoping the fetch alone would not have closed it: the body loop was `if hasattr(entity, key): setattr(entity, key, value)`, which reaches every mapped column. I demonstrated all three holes against unmodified `main` rather than inferring them — the cross-tenant rewrite; a `{"tenant_id": victim}` body moving your own row into the victim's namespace; and `{"id": <chosen>}` repointing the primary key, after which the old id 404s and the row lives at the caller's UUID. Writable columns are now an allowlist, enforced in the service because `entity_bulk_upsert` reaches the method without passing through a route.

**One thing I want to flag about the tests rather than bury.** Two independent mechanisms block the tenant case — the route removing the key, and the allowlist — so I checked whether either was actually pinned. Widening the allowlist to include `tenant_id` leaves the suite green; so does replacing the removal with a plain read. Either alone closes the hole, so no route-level test can isolate one, and the module docstring says so instead of letting a test name imply coverage it does not have. The primary key is the case reachable through the allowlist alone, and `test_scope_columns_are_not_caller_writable` pins that. 4 of 5 tests fail on the parent; the fifth is the happy-path guard.

**Gating signals on head `81695e4b`:**

- **CI** — pass. All 13 checks completed green (`claude-capture`/`claude-retrigger` are the by-design skips).
- **Claude Code review** — "No issues found", run once on this exact head. No rebase happened mid-flight, so no stale-verdict caveat this time.

**Tenancy gate:** both allowlist entries deleted; ratchet holds. `id-addressed-write` 16 → 14, total 114 → 112, bound-to-tenant 268 → 270.

**Two things for you, neither blocking this PR:**

1. #1118 — `memory_update` has a correct tenant predicate and the identical primary-key hole. Verified on `main`, filed rather than folded in here.
2. The gate checks tenant binding *only*. It would not have caught #1118, and it would not catch a regression of what this PR just fixed. We are hand-fixing 16 routes with a ratchet on half the invariant. Whether scope-column writability belongs in the gate as a second check is a design call I did not want to make unilaterally — noted on #1118.

erni-a (approved):

Issue #1121, opened 2026-08-30
storage: POST /fleet/nodes lets the caller repoint a node's primary key on upsert


Eldad-Caura:
`route:POST /api/v1/storage/fleet/nodes`

Service method: `method:fleet_upsert_node`

### The defect

`fleet_upsert_node` builds its conflict-update SET from the caller's dict with an inline denylist that omits the primary key:

```python
stmt = pg_insert(_table(FleetNode)).values(**values)
stmt = stmt.on_conflict_do_update(
    constraint="uq_fleet_nodes_tenant_node",
    set_={k: v for k, v in values.items() if k not in ("tenant_id", "node_name")},
).returning(FleetNode.__table__.c.id)
```

The route hands it `await request.json()` with no schema and no key filter, so `id` reaches the SET. A second POST for the same `(tenant_id, node_name)` takes the conflict path and rewrites the row's primary key to whatever the caller named.

Verified on `main` against a local PG:

```
POST /fleet/nodes {tenant, node_name}                -> 200, id = 091c90d2-…
POST /fleet/nodes {same tenant+node_name, "id": X}   -> 200, returned id = X
PK REPOINTED = True
```

`FleetCommand.node_id` references `fleet_nodes.id`, so every queued command for that node is stranded against the old value.

### Why this is filed separately

Third instance of one class: a caller-controlled dict becomes an `UPDATE ... SET` with no restriction on identity columns. The other two:

- `entity_update` — fixed in #1119 (`_ENTITY_UPDATABLE_FIELDS`)
- `memory_update` — #1118

Each was found while fixing the previous one. This is not a tenant-scope hole — `tenant_id` is correctly excluded here — which is exactly why the tenant-scope gate does not see it.

### The fix

Add `id` to the exclusion. Better, give it the same named-constant treatment the other two now have, rather than a third inline literal tuple.

### The wider point

Three methods, three shapes: an explicit allowlist constant, a derived denylist constant, and an inline literal tuple. `scripts/tenant_scope_gate.py` already enumerates `_public_methods(PostgresService)` for its widening-grant check, so a runtime assertion — every writable-column set is disjoint from its model's primary key and scope columns — is a small addition and would have caught all three. That is the design question raised on #1119: the gate enforces tenant binding only, and this class of defect is invisible to it by construction.

Eldad-Caura:
**Correction to the harm statement above, from testing it while writing the fix.**

[...]

PR #1122, opened 2026-08-30
fix(memory): keep identity and scope columns out of the by-id patch

### LINKED ISSUE #1118: storage: PATCH /memories/{memory_id} lets the caller repoint the primary key

Eldad-Caura:
Closes #1118.

`memory_update`'s tenant predicate was already correct — every statement carries `Memory.tenant_id == tenant_id`. What it lacked was any restriction on *which* columns the patch could set: `values` was built with `hasattr(Memory, key)`, which is true of `id`. The statement became `UPDATE memories SET id = <caller's choice> WHERE id = :memory_id AND tenant_id = :tenant_id` — predicate satisfied, primary key rewritten. Same defect closed for entities in #1119.

### Two more columns were reachable the same way

Both were found by checking this fix against the exclusions the entity one already made, and both are demonstrated against unmodified `main`:

| patch | before |
|---|---|
| `{"id": <chosen>}` | `200`; old id `404`s, row answers at the caller's id |
| `{"fleet_id": "attacker-fleet"}` | `200`; memory moved between fleet scopes by a *content* endpoint |
| `{"search_vector": ""}` | `200`; vector wiped, **row stops matching its own content** |

The `search_vector` case is the one I'd flag. The trigger is `BEFORE INSERT OR UPDATE OF content, title` (migration 034), so a patch naming *only* that column never fires it and the caller's value persists verbatim. The row silently leaves keyword recall with no content change to explain it. `test_the_search_vector_is_not_caller_writable` asserts through an actual `plainto_tsquery` match, not just the column value.

### Why this is a subtraction and the entity fix was a list

`Memory` has 34 columns and many writers (public `MemoryUpdate`, core-worker's re-embed, governance remediation, enrichment, the TESTING-gated time-warp route). An allowlist assembled by inspection would be a guess there, and a wrong guess silently drops a production write — the SAFE-01 failure mode. Entity's is four columns and two callers, so it can be enumerated exactly. Both arrive at the same guarantee.

To make "nothing else narrows" a measurement rather than a claim, I instrumented `memory_update` and recorded every patch key both suites issue (5,922 tests): 12 distinct keys, none of them the four now excluded. The insert path keeps the full `_MEMORY_VALID_FIELDS`, where all four are legitimately set.

### Test honesty

Each exclusion was verified **load-bearing** by re-admitting it and watching the row move — the sabotage that a prior round of review showed is the only way to catch a test that pins something other than what it claims. `test_the_row_cannot_be_moved_to_another_tenant` is driven at the service deliberately: the route pops `tenant_id`, so over HTTP that key cannot reach the column set, and a route-level test would have proved nothing. 5 of 6 tests fail on the parent commit; the sixth is the regression guard.

### One flake, disclosed

`tests/test_unknown_field_rejection.py::test_memory_update_rejects_unknown_field` failed on 2 of 6 full root-suite runs here and has since passed 5 consecutive times, including post-rebase. It is not caused by this change, and that is checkable rather than asserted: it patches `{"weightt": 0.9}`, and `hasattr(Memory, "weightt")` and `"weightt" in _MEMORY_UPDATABLE_FIELDS` are **both** `False` — old and new filters produce an identical empty `values` for that input. This change alters the treatment of exactly five keys: `id`, `tenant_id`, `fleet_id`, `search_vector`, `registry`. The test is on the known-flaky list for this local environment.

### Follow-up filed, not fixed here

#1121 — `fleet_upsert_node` is the third instance of this class: `set_={k: v for k, v in values.items() if k not in ("tenant_id", "node_name")}` omits `id`, and the route passes `await request.json()` in unfiltered. Verified: a second POST for the same `(tenant_id, node_name)` repoints the node's primary key, stranding every `FleetCommand.node_id` that referenced it.

That makes three methods and three shapes — explicit allowlist, derived denylist, inline literal tuple. It sharpens the question from #1119: the gate checks tenant binding only, so it saw none of these. A runtime assertion in `scripts/tenant_scope_gate.py` — every writable-column set disjoint from its model's primary key and scope columns — would have caught all three, and the gate already enumerates `_public_methods(PostgresService)` for its widening-grant check. Your call; I haven't touched the gate.

### Verification

Storage 234 passed; root 5694 passed / 4 skipped / 1 xfailed, zero failures; core-worker 80 passed. ruff check + `format --check` clean at all six CI scopes. mypy clean across core-api (203), core-storage-api (85), core-operations (5), core-worker (11), common (99), scripts (31). Gate exit 0, allowlist untouched — `memory_update` was already tenant-bound, which is precisely why this hole was invisible to it. Rebased onto `448bde05` and re-run there. `uv.lock` unchanged across all four lockfiles.

### DIFF
--- a/core-storage-api/src/core_storage_api/services/postgres_service.py
+++ b/core-storage-api/src/core_storage_api/services/postgres_service.py
@@ -327,6 +327,32 @@ _MEMORY_VALID_FIELDS = frozenset(
     {c.key for c in Memory.__table__.columns} | {a.key for a in Memory.__mapper__.column_attrs}
 )
 
+# Columns a caller may never rewrite on an existing memory. All four are
+# legitimately SET on insert, so ``_filter_memory_fields`` keeps using the full
+# set above; it is the by-id patch that must not reach them.
+#
+#   id             identity — the reported defect (#1118)
+#   tenant_id      scope
+#   fleet_id       scope
+#   search_vector  trigger-maintained, and the trigger is
+#                  ``BEFORE INSERT OR UPDATE OF content, title`` (migration
+#                  034), so a patch naming ONLY this column does not fire it:
+#                  the caller's tsvector persists verbatim and the row leaves
+#                  keyword recall with no content change to explain it.
+#
+# Same four classes ``_ENTITY_UPDATABLE_FIELDS`` below excludes by listing what
+# it admits.
+_MEMORY_IMMUTABLE_FIELDS = frozenset({"id", "tenant_id", "fleet_id", "search_vector"})
+
+# Columns ``memory_update`` may write. A subtraction rather than the explicit
+# list the entity set uses, because ``Memory`` has 34 columns and many writers
+# (the public ``MemoryUpdate`` surface, core-worker's re-embed, governance
+# remediation, enrichment, the TESTING-gated time-warp route): an allowlist
+# assembled by inspection would be a guess, and a wrong guess silently drops a
+# production write. Entity's is four columns and two callers, so it can be
+# enumerated exactly. Both arrive at the same guarantee.
+_MEMORY_UPDATABLE_FIELDS = _MEMORY_VALID_FIELDS - _MEMORY_IMMUTABLE_FIELDS
+
 # Columns ``entity_update`` may write. Deliberately a subset, not
 # ``Entity.__table__.columns`` the way ``_MEMORY_VALID_FIELDS`` above is: the
 # previous ``hasattr(entity, key)`` test admitted every mapped column, so a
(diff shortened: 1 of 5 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #1125, opened 2026-08-31
docs(storage): verify opaque-body tenant bindings

Eldad-Caura:
## Summary

- add a hand-checked disposition to every one of the 39 `opaque-body-write` exceptions
- verify 35 payload bindings directly in the route/service code
- recategorise the four memory↔entity link route/method entries, whose payloads carry only bare UUIDs, as `id-addressed-write`
- track that live cross-tenant integrity hole in #1124 without changing any runtime method or removing any allowlist entry

## Count movement

| category | before | after | delta |
|---|---:|---:|---:|
| `opaque-body-write` | 39 | 35 | -4 |
| `id-addressed-write` | 14 | 18 | +4 |
| total exceptions | 112 | 112 | 0 |

The larger write backlog is intentional: this is the first honest classification of the four link paths, not a new regression.

The four moved entries are:

- `method:entity_add_entity_link`
- `method:entity_bulk_upsert_links`
- `route:POST /api/v1/storage/entities/links`
- `route:POST /api/v1/storage/entities/links/bulk`

Every other original `opaque-body-write` entry now has a one-line `Verified:` note naming the binding tenant source. No entry was added or removed. Package T ownership is preserved.

## Verification

- `python3 scripts/tenant_scope_gate.py --base origin/main` — 112 allowlisted; 35 opaque-body writes; 18 id-addressed writes
- `python3 scripts/tenant_scope_gate.py --write` — byte-identical SHA-256 before/after; all notes survived regeneration
- `uv run --frozen pytest tests/test_tenant_scope_gate.py -q` — 72 passed
- CI-exact Ruff check and formatter scopes — clean
- `python3 scripts/legacy_name_ratchet.py --base origin/main` — no new lines
- `python3 scripts/do_not_touch_sentinel.py` — all 46 protected strings survive
- `git diff --check`

One signed-off commit; rebased onto current `origin/main` before push.

PR #1126, opened 2026-08-31
fix(agent): bind search-profile reset to tenant

### LINKED ISSUE #1089: storage: agent_reset_search_profile takes no tenant scope (route validates, method does not)
`method:agent_reset_search_profile`

### The defect

`agent_reset_search_profile(agent_id_pk)` runs `UPDATE agent SET search_profile = NULL WHERE id = :pk` with no tenant predicate.

**Its one current caller is safe.** `POST /agents/{agent_id}/search-profile/reset` does `agent_get_by_id(agent_id, body["tenant_id"])` and 404s before calling it, so there is no live cross-tenant hole through that route today. This is defence-in-depth: the method is unsafe for the next caller, and the guarantee currently lives in the caller rather than in the query.

### The fix

Take `tenant_id` and put it in the `WHERE`. The route's lookup then becomes belt-and-braces rather than the only thing standing between a caller and another tenant's row.

---

### Why this is filed

Eldad-Caura:
## Summary

- bind `agent_reset_search_profile` to the caller's tenant in the SQL `WHERE`
- pass the same tenant used by the route's existing scoped lookup into the service method
- add a cross-tenant regression test and a positive same-tenant test
- remove the resolved `method:agent_reset_search_profile` allowlist entry

## Security posture

This is defence-in-depth, not a live cross-tenant hole through the current route. Its one current caller was already safe because it performs a tenant-scoped lookup and returns 404 before calling the reset method. This change moves that guarantee into the update query so the method is safe for its next caller too.

The update has a fixed `SET search_profile = NULL`; it does not build its assignment set from caller-controlled fields, so there is no identity-column repoint path in this method.

The measured tenant-scope allowlist total shrinks from 112 to 111. After #1125 landed, base `main` has 35 `opaque-body-write` entries with notes and four entries recategorized into `id-addressed-write`; this PR preserves those fields and removes only `method:agent_reset_search_profile`.

## Fails without the fix

With only the new `Agent.tenant_id == tenant_id` predicate removed, the regression test fails as follows:

```text
FAILED core-storage-api/tests/test_agent_reset_search_profile_tenant_scope.py::TestAgentResetSearchProfileTenantScope::test_service_does_not_reset_another_tenants_profile
E   AssertionError: assert None == {'min_similarity': 0.61}
1 failed in 0.42s
```

## Verification

- focused regression: `2 passed`
- full core-storage-api integration suite: `236 passed`
- full root suite: `5694 passed, 4 skipped, 16 deselected, 1 xfailed`
- Ruff check and format check: passed
- core-storage-api mypy: `Success: no issues found in 85 source files`
- tenant-scope gate: `271 bound to a tenant, 111 allowlisted`; against the rebased #1125 base, `id-addressed-write` shrinks from 18 to 17 and `opaque-body-write` remains 35

Closes #1089

### DIFF
--- a/core-storage-api/src/core_storage_api/routers/agents.py
+++ b/core-storage-api/src/core_storage_api/routers/agents.py
@@ -85,10 +85,11 @@ async def get_search_profile(agent_id: str, tenant_id: str) -> dict:
 @router.post("/{agent_id}/search-profile/reset")
 async def reset_search_profile(agent_id: str, request: Request) -> dict:
     body: dict = await request.json()
-    agent = await _svc.agent_get_by_id(agent_id, body["tenant_id"])
+    tenant_id = body["tenant_id"]
+    agent = await _svc.agent_get_by_id(agent_id, tenant_id)
     if agent is None:
         raise HTTPException(status_code=404, detail="Agent not found")
-    await _svc.agent_reset_search_profile(agent.id)
+    await _svc.agent_reset_search_profile(agent.id, tenant_id=tenant_id)
     return {"ok": True}
 
 
--- a/core-storage-api/src/core_storage_api/services/postgres_service.py
+++ b/core-storage-api/src/core_storage_api/services/postgres_service.py
@@ -7023,11 +7023,18 @@ class PostgresService:
     async def agent_reset_search_profile(
         self,
         agent_id_pk: object,
+        *,
+        tenant_id: str,
     ) -> None:
-        """Clear an agent's search_profile by primary key (Agent.id)."""
+        """Clear one tenant's agent search profile by primary key."""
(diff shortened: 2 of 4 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

Eldad-Caura:
@claude @gemini The branch was force-pushed by the adjacent tenancy session to rebase onto main after #1125. I re-ran tenant_scope_gate.py --write and verified all 35 remaining opaque-body-write notes and the four new id-addressed-write recategorizations survived unchanged; this PR still removes only method:agent_reset_search_profile. Please re-review the rebased head.

PR #1129, opened 2026-08-31
fix(fleet): keep the primary key out of the node upsert's conflict path

### LINKED ISSUE #1121: storage: POST /fleet/nodes lets the caller repoint a node's primary key on upsert

Eldad-Caura:
Closes #1121.

`fleet_upsert_node` built its `ON CONFLICT DO UPDATE` SET clause from the caller's dict with a denylist that named the conflict key but not the primary key:

```python
set_={k: v for k, v in values.items() if k not in ("tenant_id", "node_name")}
```

The route hands it `await request.json()` with no schema and no key filter, so `id` reached the SET and a second POST for the same `(tenant_id, node_name)` rewrote the row's primary key to whatever the caller named.

### Two outcomes, and I had the harm wrong in the issue

Both verified against unmodified `main`:

| state of the node | before |
|---|---|
| no referencing commands | `200`; the row now answers at the caller's chosen UUID |
| a queued command | `ForeignKeyViolationError` escapes as an unhandled `IntegrityError` → **500** |

I originally wrote on #1121 that queued commands would be "stranded against the old value." That is wrong: `FleetCommand.node_id` is `ForeignKey("fleet_nodes.id", ondelete="CASCADE")` with no `onupdate`, so `NO ACTION` applies and Postgres refuses to orphan them. Nothing is stranded — instead a routine heartbeat upsert becomes a 500 the moment any command is queued for that node. The correction is [recorded on the issue](https://github.com/caura-ai/caura/issues/1121#issuecomment-5472107905) rather than edited into the description, so the original claim and its correction both stay visible. The fix is unchanged and closes both outcomes.

### Third instance of one class

After `entity_update` (#1119) and `memory_update` (#1122): a caller-controlled dict becoming an `UPDATE ... SET` with no restriction on identity columns. The inline tuple is now `_FLEET_NODE_IMMUTABLE_FIELDS`, sitting beside its two siblings so the three read alike and the omission is visible. Column names, not attribute names — the statement is built against `_table(FleetNode)`, which is why `extra` arrives as `metadata`.

**This is not a tenant-scope hole.** `tenant_id` was already excluded, so the tenant-scope gate never had anything to say — which is the same blind spot that hid #1122. Both allowlist entries are `opaque-body-write` and correctly stay: the fix changes column writability, not tenant binding. The allowlist is untouched by this PR.

### Tests

3 of 4 fail on the parent; the fourth is the regression guard proving the conflict path still applies real columns. The `id` exclusion was verified load-bearing by re-admitting it and watching the key move. The drift guard derives from `FleetNode.__table__.primary_key` rather than echoing the literal, and reaches the constant through the module object so the file still *collects* on a commit where it does not exist yet.

### A red run, and what it actually was

`tests/test_unknown_field_rejection.py::test_memory_update_rejects_unknown_field` failed intermittently during this work. It is **not** caused by this change, and this time that is a measurement rather than an inference: I ran four matched baseline runs with the change reverted on the same base, and it failed **1 of 4** there too. The captured traceback gives the mechanism — the test's *setup* POST is rejected `409 DUPLICATE_MEMORY / reason: semantic_similarity` against a memory left by an earlier test in the same run, so it never reaches the PATCH it is testing. Order-dependent accumulation in a shared test database, unrelated to fleet nodes. Worth its own issue if it keeps costing people time; I have not filed one.

### Verification

Storage suite 238 passed; root suite 5694 passed with the flake above as the only intermittent. ruff check + `format --check` clean at all six CI scopes. mypy clean across core-api (203), core-storage-api (85), core-operations (5), core-worker (11), common (99), scripts (31). Gate exit 0, allowlist unchanged. Rebased onto `24aea6c1` and re-run there. `uv.lock` unchanged across all four lockfiles.

### DIFF
--- a/core-storage-api/src/core_storage_api/services/postgres_service.py
+++ b/core-storage-api/src/core_storage_api/services/postgres_service.py
@@ -364,6 +364,22 @@ _MEMORY_UPDATABLE_FIELDS = _MEMORY_VALID_FIELDS - _MEMORY_IMMUTABLE_FIELDS
 # ``_MEMORY_UPDATABLE_FIELDS`` above follows.
 _ENTITY_UPDATABLE_FIELDS = frozenset({"canonical_name", "entity_type", "attributes", "name_embedding"})
 
+# Columns ``fleet_upsert_node``'s ON CONFLICT DO UPDATE must not rewrite. The
+# insert half of that statement still uses every key the caller sent; this is
+# only the branch that lands on a row that already exists.
+#
+#   id                    identity — the reported defect (#1121)
+#   tenant_id, node_name  the conflict key itself; rewriting either half would
+#                         move the row out from under the constraint the
+#                         statement just matched on
+#
+# Named rather than the inline ``k not in ("tenant_id", "node_name")`` tuple it
+# replaces, so it reads like its two siblings above and a reviewer can see what
+# the omission was. Column names, not model attribute names: the statement is
+# built against ``_table(FleetNode)``, so ``values`` is keyed by column (which
+# is why ``extra`` arrives as ``metadata``).
+_FLEET_NODE_IMMUTABLE_FIELDS = frozenset({"id", "tenant_id", "node_name"})
+
 # Columns the admin memory-list endpoint may sort by. Allowlisted so an
 # unexpected ``sort`` value falls back to created_at instead of raising
 # AttributeError (500) at ``getattr(Memory, sort)`` — the endpoint is callable
@@ -8916,7 +8932,7 @@ class PostgresService:
             stmt = pg_insert(_table(FleetNode)).values(**values)
             stmt = stmt.on_conflict_do_update(  # type: ignore[assignment]
                 constraint="uq_fleet_nodes_tenant_node",
-                set_={k: v for k, v in values.items() if k not in ("tenant_id", "node_name")},
+                set_={k: v for k, v in values.items() if k not in _FLEET_NODE_IMMUTABLE_FIELDS},
             ).returning(FleetNode.__table__.c.id)
             result = await session.execute(stmt)
             await session.flush()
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #1140, opened 2026-08-31
test: isolate unknown-field rejection setup

### LINKED ISSUE #1137: test: test_memory_update_rejects_unknown_field fails ~1 in 4 on a shared-tenant dedup collision
`tests/test_unknown_field_rejection.py::test_memory_update_rejects_unknown_field` fails intermittently in full root-suite runs and passes in isolation every time.

### It does not fail where the name suggests

The test never reaches the assertion it exists to make. Its **setup** POST is rejected:

```
    async def test_memory_update_rejects_unknown_field(client):
        tenant_id, headers = get_test_auth()
        created = await client.post("/api/v1/memories", json={...})
>       assert created.status_code == 201, created.text
E       AssertionError: {"detail":"Near-duplicate memory exists: 972f1311-…",
E         "error":{"code":"DUPLICATE_MEMORY",
E           "details":{"reason":"semantic_similarity","existing_id":"972f1311-…",
E                      "existing_status":"active"}}}
E       assert 409 == 201

tests/test_unknown_field_rejection.py:118: AssertionError
```

The SAFE-01 behaviour under test — a PATCH with `{"weightt": 0.9}` returning a 422 that names the field — is never exercised on a failing run. So the red is not a signal about unknown-field rejection at all.

### Rate

Measured **1 failure in 4** consecutive full root-suite runs against unmodified `origin/main` with no local changes, and at a comparable rate (~4 in 11) across runs that carried unrelated in-flight work. It is not caused by any particular change: I confirmed it on a clean baseline specifically to rule that out.

`pytest tests/test_unknown_field_rejection.py::test_memory_update_rejects_unknown_field` passes every time. That is the expensive part — the obvious next step after a red is to re-run it alone, see green, and be left unsure whether the failure meant anything.

### Mechanism

Two facts combine:

1. `get_test_auth()` (`tests/conftest.py:93`) returns the fixed tenant `"default"` for every caller, so this test shares a tenant with everything else in the run that uses it.
2. `memory_find_semantic_duplicate` is tenant-scoped, so the write-path dedup gate compares this POST against memories other tests wrote into `"default"` earlier in the same run.

The content string carries a `_uid()` suffix, so it is textually unique on every call — but the dedup gate is cosine similarity on the embedding, not text equality, and `"a memory that will be patched with a bad field name <uid>"` stays semantically close to whatever else is in the tenant. Textual uniqueness does not protect a test against a semantic-similarity gate.

### What I have not isolated

Which specific prior write it collides with, and therefore why it is intermittent rather than deterministic given a stable test order. Candidates worth checking: variation in which tests actually ran (deselection/skips shift the set of prior rows), and async interleaving changing what has committed by the time this POST lands. I stopped at the point where the cause was clearly ambient rather than mine.

### Why it is worth fixing rather than tolerating

Several agents and contributors are currently running this suite repeatedly across the tenancy backlog. A red here reads as "my change broke unknown-field rejection," which is both alarming and wrong, and the natural triage (re-run in isolation) produces a green that explains nothing. The failure mode is someone reverting good work, or "fixing" the dedup gate to make a test pass.

### Suggested direction

Give the test its own tenant rather than the shared `"default"` — the per-test-unique tenant convention the storage suite already uses (`f"…-{uuid.uuid4().hex[:8]}"`) — so the dedup gate has nothing to compare against. That is a smaller change than weakening the gate or special-casing the content, and it fixes the class rather than this instance: any other test using `get_test_auth()` plus a write is exposed the same way.

ASVLCII:
## Summary

Use the repository's sweep-visible per-test tenant for the memory-update unknown-field regression setup, so semantic deduplication from unrelated tests cannot reject the setup POST before the PATCH assertion runs. Also correct the `uid()` helper docstring, which previously implied uniqueness protected against semantic duplicates.

## Related Issue

Closes #1137

## Testing

- `python -m compileall -q tests/conftest.py tests/test_unknown_field_rejection.py` (passed)
- `git diff --check` (passed)
- Targeted pytest, Ruff, and mypy could not be run locally because this Windows environment does not have `uv`, pytest, Ruff, mypy, or Docker installed; CI should provide the repository's Python 3.12/PostgreSQL test environment.

## Checklist

- [x] Focused change with regression-oriented test isolation
- [x] Uses `new_tenant_id()` so test rows remain covered by end-of-run cleanup
- [x] DCO sign-off included

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -111,7 +111,7 @@ def get_admin_headers() -> dict:
 
 
 def uid() -> str:
-    """Short unique suffix — for distinct content (409s) and distinct tenant ids."""
+    """Short unique suffix for exact-duplicate-safe content and tenant ids."""
     return uuid.uuid4().hex[:8]
 
 
--- a/tests/test_unknown_field_rejection.py
+++ b/tests/test_unknown_field_rejection.py
@@ -28,7 +28,6 @@ import pytest
 from tests.conftest import get_test_auth, new_tenant_id
 from tests.conftest import uid as _uid
 
-
 # ---------------------------------------------------------------------------
 # Helpers
 # ---------------------------------------------------------------------------
--- a/tests/test_unknown_field_rejection.py
+++ b/tests/test_unknown_field_rejection.py
@@ -25,8 +25,9 @@ Three things are pinned here, and the third matters as much as the first two:
 
 import pytest
 
-from tests.conftest import get_test_auth
+from tests.conftest import get_test_auth, new_tenant_id
 from tests.conftest import uid as _uid
+
 
 # ---------------------------------------------------------------------------
 # Helpers
@@ -105,7 +106,9 @@ async def test_memory_create_rejects_plausible_but_undeclared_field(client):
 
 async def test_memory_update_rejects_unknown_field(client):
     """PATCH /memories/{id} — a typo here is a silent 200 no-op."""
-    tenant_id, headers = get_test_auth()
+    # Use a sweep-visible tenant so semantic dedup from unrelated tests cannot
+    # reject this setup write before the PATCH assertion is reached.
+    tenant_id, headers = get_test_auth(new_tenant_id())
     created = await client.post(
         "/api/v1/memories",
         json={


Eldad-Caura:
@claude

Eldad-Caura:
@erni-a

erni-a (approved):

erni-a:
Approved on 6 Sep and then left waiting, which is on us. Branch updated, CI green, merging now. Thanks @ASVLCII.

Issue #1124, opened 2026-08-31
storage: POST /entities/links creates cross-tenant links by bare UUIDs


Eldad-Caura:
`route:POST /api/v1/storage/entities/links`
`route:POST /api/v1/storage/entities/links/bulk`

Service methods: `method:entity_add_entity_link`, `method:entity_bulk_upsert_links`

### The defect

Both handlers accept only caller-supplied `memory_id`, `entity_id`, and `role` values. Neither reads a tenant or checks that either referenced row belongs to one. The `memory_entity_links` table itself has no `tenant_id`; its foreign keys prove only that the two UUIDs exist, not that they belong to the same tenant or to the caller.

The single path inserts the pair directly. The bulk path runs `INSERT ... ON CONFLICT DO UPDATE` for every supplied pair, likewise without an ownership check. This is a live cross-tenant integrity hole through two HTTP routes, not defence-in-depth: a reachable caller who knows foreign memory/entity UUIDs can create graph edges across tenant boundaries.

The sibling PATCH path is tracked separately in caura-ai/caura#1085. These POST routes and service methods are distinct allowlist entries and are not named there.

### The fix

Require a binding `tenant_id` on both routes and prove that both the memory and entity belong to it before inserting. Apply the ownership predicate inside the storage operation (including every item in the bulk path), not only as a preflight lookup, so the service methods are safe for every caller and the check cannot drift away from the mutation.

---

### Why this is filed

`core-storage-api` authenticates no request (GHSA-wgvw-28pq-jc36), and docker-compose publishes it on `0.0.0.0:8002`. Anything reachable there is reachable by anyone who can reach the port. Bare primary keys are therefore not an authorisation check: knowing UUIDs is not the same as being entitled to the rows behind them.

Tracked by the tenant-scope gate added in caura-ai/caura#1069. These four paths were filed in `core-storage-api/tenant_scope_allowlist.json` as `opaque-body-write`, but the payloads carry no tenant at all; they belong under `id-addressed-write`, the category the gate flags as backlog that mutates. Fixing the paths means deleting their allowlist entries.

PR #1141, opened 2026-08-31
fix(fleet): scope command deletion to tenant

### LINKED ISSUE #1090: storage: fleet_delete_commands_for_nodes takes no tenant scope (route validates, method does not)
`method:fleet_delete_commands_for_nodes`

### The defect

`fleet_delete_commands_for_nodes(node_ids=[...])` issues `DELETE FROM fleet_command WHERE node_id IN (...)` with no tenant predicate — a hard delete, not a soft one.

**Its one current caller is safe.** The route resolves the node via `fleet_get_node_id(tenant_id=..., node_name=...)`, which is tenant-scoped, before deleting. So no live hole through that route. As above, the guarantee is in the caller.

### The fix

Take `tenant_id` and join or filter on it in the DELETE. This one deletes rows outright, so it is the method-only entry I would fix first.

---

### Why this is filed

Eldad-Caura:
## Summary

- require `tenant_id` when deleting fleet commands by node id
- bind that tenant directly in the hard-delete predicate
- pass the route's already-resolved tenant into the service method
- add mixed-tenant service coverage and route-plumbing coverage
- remove the resolved `method:fleet_delete_commands_for_nodes` allowlist entry

## Security posture

This is defence-in-depth, not a live cross-tenant hole through the current route. Its one current caller already resolves the node within the caller's tenant before invoking the delete. The change moves that guarantee into the hard-delete query so a future service caller cannot delete another tenant's commands with known node UUIDs.

This path performs a fixed `DELETE`; it has no caller-controlled update field set or identity-column repoint shape.

Against current `main`, the tenant-scope allowlist shrinks from 111 to 110 and `id-addressed-write` shrinks from 17 to 16. All 35 annotated `opaque-body-write` entries are preserved. The four #1124 entity-link entries remain out of scope.

## Fails without the fix

With only `FleetCommand.tenant_id == tenant_id` removed from the delete predicate, the mixed-tenant regression fails because the foreign command is deleted:

```text
FAILED core-storage-api/tests/test_fleet_command_delete_tenant_scope.py::TestFleetCommandDeleteTenantScope::test_service_deletes_only_the_callers_commands
E       AssertionError: assert '44bccbd4-cc7e-4954-b65a-f1d4bf0c9c4e' in set()
1 failed in 0.57s
```

## Verification

- focused regression: `2 passed`
- full core-storage-api suite: `242 passed`
- full root suite: `5694 passed, 4 skipped, 16 deselected, 1 xfailed`
- Ruff check and format check: passed
- core-storage-api mypy: `Success: no issues found in 85 source files`
- tenant-scope gate: `272 bound to a tenant, 110 allowlisted`; `id-addressed-write` 17 → 16

Closes #1090

### DIFF
--- a/core-storage-api/src/core_storage_api/routers/fleet.py
+++ b/core-storage-api/src/core_storage_api/routers/fleet.py
@@ -114,7 +114,10 @@ async def delete_node(
         raise HTTPException(status_code=404, detail="Node not found")
     fleet_id = getattr(node, "fleet_id", None)
     if fleet_id:
-        await _svc.fleet_delete_commands_for_nodes(node_ids=[node_id])
+        await _svc.fleet_delete_commands_for_nodes(
+            tenant_id=tenant_id,
+            node_ids=[node_id],
+        )
     return {"ok": True}
 
 
--- a/core-storage-api/src/core_storage_api/services/postgres_service.py
+++ b/core-storage-api/src/core_storage_api/services/postgres_service.py
@@ -9220,12 +9220,20 @@ class PostgresService:
     async def fleet_delete_commands_for_nodes(
         self,
         *,
+        tenant_id: str,
         node_ids: list[UUID],
     ) -> None:
         if not node_ids:
             return
         async with get_session() as session:
-            await session.execute(_table(FleetCommand).delete().where(FleetCommand.node_id.in_(node_ids)))
+            await session.execute(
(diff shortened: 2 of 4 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

Issue #1082, opened 2026-08-30
storage: PATCH /reports/{report_id} completes any tenant's report by UUID


Eldad-Caura:
`route:PATCH /api/v1/storage/reports/{report_id}`

Service method: `method:report_update_completed`

### The defect

The handler reads `status`, `completed_at`, `duration_ms`, `summary`, `hygiene`, `health`, `usage_data`, `issues` and `crystallization` from the body and writes them against `report_id` alone. No tenant is read.

### The fix

Require `tenant_id` and scope the UPDATE. Note this one writes a large amount of caller-controlled content into another tenant's report row, so it is a data-integrity issue as much as a confidentiality one.

---

### Why this is filed

`core-storage-api` authenticates no request (GHSA-wgvw-28pq-jc36), and docker-compose publishes it on `0.0.0.0:8002`. Anything reachable there is reachable by anyone who can reach the port. A bare primary key is therefore not an authorisation check: knowing a UUID is not the same as being entitled to the row behind it.

Tracked by the tenant-scope gate added in caura-ai/caura#1069 — this path is in `core-storage-api/tenant_scope_allowlist.json` under `id-addressed-write`, the category the gate flags `<- backlog, MUTATES`. The gate's ratchet means the allowlist can shrink but never grow, and it refuses to let a `id-addressed-write` entry be relabelled as something milder. **Fixing this path means deleting its allowlist entry; CI will fail until the entry is removed, which is the intended forcing function.**

PR #1143, opened 2026-08-31
feat(gate): check that a tenant predicate cannot be rewritten by its own request

Eldad-Caura:
The gate checks that a statement is **bound** to a tenant. It never checked that the caller cannot **rewrite the column it is bound on**. Three defects went through it in a week on exactly that gap:

| | method | the filter it had |
|---|---|---|
| #1081 / #1119 | `entity_update` | `if hasattr(entity, key): setattr(…)` |
| #1118 / #1122 | `memory_update` | `if hasattr(Memory, key)` in a comprehension |
| #1121 / #1129 | `fleet_upsert_node` | `if k not in ("tenant_id", "node_name")` |

Each built an `UPDATE ... SET` from a caller-controlled dict and admitted `id`, so the statement became `SET id = <caller's choice> WHERE id = :id AND tenant_id = :tenant` — predicate satisfied, primary key moved. **The last two were correctly tenant-bound the whole time**, so checks 1–3 had nothing to object to. The defect lives one layer below where the gate was looking.

### Three parts, three different failures

They are not redundant — each catches a mode the others miss:

1. **AST discovery** names every method building an UPDATE from caller keys and requires it to be registered. This is what `entity_update` and `memory_update` failed: no filter at all.
2. **Reference check** requires the method to actually mention the constant it is registered for, so the registry stays a *reading* of the code rather than a claim about it. Without it, a method could revert to `hasattr` while the registry still named the right constant and both other parts would pass.
3. **Runtime check** resolves each constant and asserts it admits none of the model's primary key, tenant-scope, or database-maintained columns — read off the model, so a schema change moves them without anyone editing this file. This is what `fleet_upsert_node` failed, carrying a filter that looked deliberate.

### What is deliberately not a site

`INSERT` is excluded. Naming the id of a row you are *creating* collides rather than overwrites, so no existing row moves — the defect is an existing row's identity changing under it. That cut keeps `relation_add`, `agent_add` and `agent_activity_digest_upsert` off the list. A loop over written-out column names is also not a site, which keeps `agent_add`'s second write off it. Both boundaries are pinned by tests, because flagging safe methods is how a registry fills with lines nobody reads.

The heuristic is biased the way the route classifier already is, and for the reason its docstring gives: it would rather name a method that turns out to be safe — one registry line a reviewer can delete — than miss one that is not, which is invisible.

### No allowlist

The invariant has no exceptions today. One with none does not need a file to hold them, and the first should be argued in review rather than added by `--write`.

### Validation

**Replayed against history.** Run against `postgres_service.py` as it stood before each fix, the check flags exactly the methods still defective at that commit — three, then two, then one. It would have caught all three.

**Sabotage, one per failure mode**, each confirmed to flip the gate red and back:

- widen a constant to admit the primary key → *"may write FleetNode.id (primary key)"*
- point a method back at `hasattr` while leaving its registration → *"registered as guarded by … but does not mention it"*
- add a new unregistered caller-keyed UPDATE → *"builds an UPDATE from caller-supplied keys … no registered identity guard"*
- leave a registration behind after a method stops qualifying → *"registered … but no longer builds an UPDATE"*

11 new tests, all failing on the parent commit, covering those four modes plus the decision boundary of the AST pass.

### It does not fail anything in flight

Zero findings on the tree as committed. That includes #1141, which landed from the tenancy lane while this was being written and introduces no new sites — checked rather than assumed, since a new gate erroring on someone else's open branch is the obvious way for this to go wrong.

### Verification

Gate tests 83 passed (71 existing + 12). Root suite 5705 passed, storage 242. mypy clean across scripts (31), core-api (203), core-storage-api (85), common (99). `ruff check scripts/` leaves only the one pre-existing ISC004 that was there before. Gate exit 0 with `--base origin/main`; allowlist untouched. Rebased onto `f8fc9513`. `uv.lock` unchanged.

No `Closes` line: the three issues this would have prevented are already fixed, and there is no open issue for the check itself.

### DIFF
--- a/scripts/tenant_scope_gate.py
+++ b/scripts/tenant_scope_gate.py
@@ -16,7 +16,7 @@ failure. It is a row belonging to somebody else.
 This makes the invariant checkable. It does NOT attempt to prove that any
 particular statement is correctly scoped: that is a semantic question, a gate
 that guesses at semantics produces false positives, and a security gate with
-false positives gets switched off. What it checks instead are three things that
+false positives gets switched off. What it checks instead are four things that
 are exactly decidable:
 
 1. **Completeness.** Every route and every public service method is enumerated
@@ -25,6 +25,19 @@ are exactly decidable:
    appear in the allowlist with a reason someone wrote and a reviewer read.
 3. **Direction.** The allowlist may shrink. It may not grow. A new unscoped
    path fails the build (``--base``).
+4. **Immutability of what the predicate filters on.** A statement bound to
+   ``WHERE tenant_id = :tenant`` is only as good as the caller's inability to
+   rewrite ``tenant_id`` — or the primary key — in the same request. Any method
+   building an ``UPDATE ... SET`` from a caller-controlled dict must filter the
+   keys through a named set, and that set must exclude the model's identity,
+   scope and database-maintained columns.
+
+Checks 1–3 were the whole gate for its first month, and three primary-key
+rewrites got through them: #1081, #1118 and #1121. Two of the three were
+correctly tenant-bound, so nothing in 1–3 had any objection to make — the
+statement satisfied its predicate and then moved the row out from under it.
+Check 4 is that gap, and it is why "bound to a tenant" is reported separately
+from "cannot be unbound by the same request".
 
 The semantics stay a human claim, as they must, but the claim is forced into the
 diff where review can see it, and the population it has to cover can only get
@@ -68,6 +81,7 @@ import json
 import re
 import subprocess
 import sys
+from collections.abc import Iterator
 from pathlib import Path
 from typing import Any
 
@@ -142,6 +156,54 @@ FAIL_CLOSED_GUARDS = frozenset({"_require", "_require_dict", "_require_number"})
 # that must keep the scope out of a column update reads it that way, and it is
 # no less a read of the request for removing the key afterwards.
 DICT_READS = frozenset({"get", "pop"})
+
+# --- Identity-column writability: the write-side half of the same invariant ---
+#
+# A tenant predicate is worth exactly as much as the immutability of the column
+# it filters on. Three methods proved that in a week, and two of the three were
+# ALREADY passing the binding-scope checks above:
+#
+#   entity_update      (#1081/#1119)  ``if hasattr(entity, key): setattr(...)``
+#   memory_update      (#1118/#1122)  ``if hasattr(Memory, key)`` in a comprehension
+#   fleet_upsert_node  (#1121/#1129)  ``if k not in ("tenant_id", "node_name")``
+#
+# Each built an ``UPDATE ... SET`` from a caller-controlled dict, and each
+# admitted ``id``. The statement became ``SET id = <caller's choice> WHERE
+# id = :id AND tenant_id = :tenant``: predicate satisfied, primary key moved.
+# ``memory_update`` and ``fleet_upsert_node`` were correctly tenant-bound the
+# whole time, so nothing above had anything to say — the defect lives one layer
+# below where this gate was looking.
+#
+# ``IDENTITY_WRITE_GUARDS`` pairs each such method with the model it writes and
+# the constant that filters the caller's keys. ``admits`` records the polarity:
+# True for ``key in CONST`` (the constant lists what may be written), False for
+# ``key not in CONST`` (it lists what may not).
+IDENTITY_WRITE_GUARDS: dict[str, tuple[str, str, bool]] = {
+    "entity_update": ("Entity", "_ENTITY_UPDATABLE_FIELDS", True),
+    "memory_update": ("Memory", "_MEMORY_UPDATABLE_FIELDS", True),
+    "fleet_upsert_node": ("FleetNode", "_FLEET_NODE_IMMUTABLE_FIELDS", False),
+}
+
+# Column types a caller may never write because the database maintains them.
+# ``search_vector`` is the live case: its trigger fires ``BEFORE INSERT OR
+# UPDATE OF content, title``, so a patch naming only the vector does not fire it
+# and the caller's value persists — the row leaves keyword recall with no
+# content change to explain it (#1122). Matched on type rather than name, so a
+# second tsvector column is covered the day it is added.
+DERIVED_COLUMN_TYPES = frozenset({"TSVECTOR"})
+
+# What a ``.values(**…)`` chain has to hang off before it can move an existing
+# row. ``pg_insert(...).values(**data)`` is deliberately absent: naming the id of
+# a row you are creating collides rather than overwrites. Enumerated rather than
+# prefix-matched, for the reason ``FAIL_CLOSED_GUARDS`` gives — a set that
+# quietly grows to fit a new name is a set that stops meaning anything.
+UPDATE_STATEMENT_ROOTS = frozenset({"sql_update", "update"})
+
+# The statements that create rather than move a row. Named so that a chain
+# resolving to neither set is treated as an update: an unrecognised builder is
+# an unknown, and this check's stated bias is to report an unknown rather than
+# assume it is the harmless one.
+INSERT_STATEMENT_ROOTS = frozenset({"pg_insert", "insert"})
 
 # How much of a binding a verdict represents, so "weaker than" is a comparison
 # rather than a rule spelled out at each site. REQUIRED binds every call;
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
Both findings applied. The High was real, and it was the worst possible shape for this PR: a false negative in the check written to catch false negatives.

**High — `setattr` loop containment used line order, not ancestry.** Confirmed before fixing:

```python
async def m(self, row_id, tenant_id, data):
    for flag in ("a", "b"):          # harmless, and earlier
        touched = flag
    for key, value in data.items():  # the unsafe one
        setattr(row, key, value)
```

Reported as **not a site**. `node.lineno >= p.lineno` counted every loop that merely *started* earlier, so the literal-key sibling vouched for the caller-keyed loop below it. Containment is now real ancestry via a parent map (`_enclosing_loops`), innermost first.

**`ast.AsyncFor`** is handled explicitly rather than by accident. You were right that it was working for the wrong reason: it was invisible to the `isinstance` test, so `loops` came back empty and an empty list reads as "not literal" — the right answer, from a coincidence that inverts the moment the keys *are* literal. Both directions now have a test.

**Low — `_resolve_local` scope-awareness.** Applied, and one correction to the finding: it is **not** latent. `_walk_scope` now stops at nested `def`/`lambda`, and the regression test for it fails on the previous commit — a closure assigning its own `values` really did decide the outer verdict in that construction.

**Four regression tests**, three of which fail on the previous commit (the fourth documents the async-for accident and passes either way, which its docstring says):

- `test_an_unrelated_earlier_loop_does_not_vouch_for_a_later_one` — the High, exactly your two-sibling-loop case
- `test_an_async_for_over_caller_keys_is_a_site`
- `test_an_async_for_over_written_out_names_is_not_a_site`
- `test_a_nested_scope_does_not_decide_the_outer_one`

87 gate tests pass. Gate exit 0 with `--base origin/main`; still exactly three sites, all registered, zero findings. mypy clean; `ruff check scripts/` back to the single pre-existing ISC004. Rebased onto `98b45020`.

Unrelated to the review: the previous push carried `tmp <t@t>` as author, leaked from a temporary commit I used while rebasing. Reset to the correct author; the signoff was always right and DCO was passing regardless.

Eldad-Caura:
@claude

Eldad-Caura:
Both Mediums were real. Confirmed each against the previous commit before touching anything:

| case | before |
|---|---|
| literal **outer** loop, caller-keyed inner | not a site |
| inline `.values(**{k: v for …})` | not a site |
| `values = {**data}` | not a site |
| `values = dict(data)` | not a site |

**Nested loops — fixed at the mechanism, not by `loops[0]`.** You offered "at minimum, restrict to the innermost loop". I went to the thing that actually answers the question: which loop *binds the name being used as the key*. `_key_is_bound_to_literal_names` walks enclosing loops innermost-first and consults **only** the one whose target binds that name. Innermost-only would give the right answer here and the wrong one for `for key in ("a","b"): for i in range(3): setattr(row, key, …)`, where the innermost loop binds nothing relevant. A key no enclosing loop binds is now unproven rather than assumed safe, with its own test.

**`.values(**X)` is now conservative**, matching the `set_=` branch you correctly pointed at as the model: anything not written out in the source is a site, whatever shape it arrives in.

That widening had to not break the INSERT exclusion, which is load-bearing — it is what keeps `relation_add`, `agent_add` and `agent_activity_digest_upsert` off the list. `_statement_root` walks the receiver chain back to its constructor, so `sql_update(M).where(…).values(**v)` is a site and `pg_insert(M).values(**v)` is not. The exclusion is now a stated rule rather than a side effect of the old narrow node-type match.

**Eight shapes checked**, including the ones that must stay clean: `agent_add`'s literal loop, an INSERT, and an UPDATE whose values are written out (`memory_update_status`, `lifecycle_audit_finalize`). Five new tests, three failing on the previous commit; the two that pass either way guard behaviour this change could have broken.

92 gate tests pass. Real tree still exactly three sites, all registered, zero findings. Gate exit 0 with `--base origin/main`. Root suite 5714 passed. mypy clean; `ruff check scripts/` back to the single pre-existing ISC004.

Worth saying plainly: this is round two of the same class of defect in a check written to catch that class — an instrument reporting success while watching the wrong thing. Each round has narrowed what "safe" is allowed to mean, from proximity to ancestry to binding.

Eldad-Caura:
@claude

Eldad-Caura:
Both real, both confirmed against the previous commit before changing anything. Neither was flagged by any test, which is the point — these were silent.

**Private helpers were skipped outright.** `fn.name.startswith("_"): continue` meant delegation laundered the pattern completely:

```python
async def entity_update(self, entity_id, tenant_id, data):
    return self._apply_patch(row, data)     # body is clean

def _apply_patch(self, row, data):          # never scanned
    for key, value in data.items():
        setattr(row, key, value)
```

Neither end reported. Every function is scanned now, and a qualifying helper is reported under its own name — which is also where the filter belongs, since that is where the keys are applied. You offered "or document it as a known limitation"; I took the first option, because a documented hole in a security gate is still a hole and this one is one refactor wide.

That also means **one of my own tests was asserting the bug**: `test_private_helpers_are_not_sites` pinned the skip as intended behaviour. It is now `test_a_private_helper_is_scanned_too`. Worth flagging plainly — a test that encodes a defect is worse than no test, because it makes the defect look decided.

**`_statement_root` gave up at a local variable.** Fixed by resolving through locals, with two things the straightforward version gets wrong:

- It considers **every** assignment of a name, not the last. In `stmt = sql_update(M)` then `stmt = stmt.values(**data)`, the newest assignment is circular — resolving only that one answers `stmt` and the update disappears. Cycle-guarded by the names already seen.
- A chain resolving to **neither** an insert nor an update builder is now reported rather than assumed harmless. `some_helper(M).values(**data)` was previously silent. This is the same bias the module docstring already claims, applied where it was missing.

The INSERT exclusion survives both, including when the insert is built through a variable — tested, since widening the net is exactly how that exclusion would have been lost.

**Six shapes verified**, four new tests, all failing on the previous commit. 96 gate tests pass; real tree still exactly three sites, zero findings; gate exit 0; root suite 5718; mypy clean; ruff back to the one pre-existing ISC004.

Three rounds, and each has narrowed what "safe" is permitted to mean: proximity, then ancestry, then binding, and now provenance through delegation and assignment. I would rather say that than imply it arrived correct.

Eldad-Caura:
@claude

Eldad-Caura:
Both applied. The High was real and the Low was the same defect seen from the other side, so they got one fix rather than two.

### High — last-assignment-wins in name resolution

Confirmed against the previous commit before changing anything:

```python
values = {k: v for k, v in patch.items()}             # caller-keyed
await session.execute(sql_update(Memory).values(**values))
values = {"status": "ok"}                             # unrelated, later
```

`{}` — not a site. The dict that reaches the `UPDATE` is the first one; resolving the name to its final assignment answers "literal" and the site disappears. Same hole on the `set_=` branch, since both call sites resolved the same way.

Worth stating plainly: the docstring on the old helper **asserted the property the code did not have** — *"a name ever built from caller keys is a candidate whatever else it holds."* That is the correct rule, written as a description of an implementation that did the opposite. It is why this survived three rounds of me reading the function.

### Low — dead `_statement_root`

Confirmed: defined, called only from its own recursion, referenced by no test. Left over from round 3, when `_is_update_statement` moved to the set-returning `_statement_roots`.

### One fix, not two

The two findings share a cause. There were **two** resolution walkers with **two different rules**: `_statement_roots` already required every assignment to be considered (added in round 3, precisely to survive `stmt = stmt.values(...)`), while `_resolve_local` took the last. The dead function was the residue of that split, and the false negative was its consequence.

So both resolutions now read `_local_assignments`, and the second walker is gone:

```python
def _all_literal_keys(fn, expr) -> bool:
    if not isinstance(expr, ast.Name):
        return _literal_keys(expr)
    origins = _local_assignments(fn, expr.id)
    # A name with no local assignment is a parameter or a global: unreadable
    # here, so it does not clear.
    return bool(origins) and all(_literal_keys(origin) for origin in origins)
```

Net **−45 lines** in the script.

### Verification

| check | result |
|---|---|
| the two new site tests, on the previous commit | **both fail** (`assert 'm' in {}`) |
| gate suite | 99 passed (96 → 99) |
| root suite | 5743 passed, 4 skipped, 1 xfailed |
| `mypy --config-file scripts/mypy.ini scripts/` | clean, 31 files |
| real tree | sites still exactly `entity_update`, `fleet_upsert_node`, `memory_update`; findings `[]` |

Three tests, not two. The third pins the other side of the rule:

```python
values = {"status": "running"}
await session.execute(sql_update(Memory).values(**values))
values = {"status": "done"}
await session.execute(sql_update(Memory).values(**values))
```

still not a site. Without it, "every origin must be literal" could be satisfied by a check that reports every name it sees, and the two regression tests would pass against it.

Rebased onto `68126b9e` and re-ran everything after the rebase, not before — #1145 landed under this branch and adds `_guard_document_shrink(…, data: dict)` to `postgres_service.py`. It is correctly not a site: it reads `data` to compare sizes, it does not build a SET. That is now a real read rather than a skip, since private helpers stopped being exempt in round 3.

### Standing count

Four rounds, seven false negatives, all in the detection and all in the direction that reads as coverage. That ratio is itself the argument for the check: a gate whose failure mode is silence needs the adversarial pass more than the code it guards.

Eldad-Caura:
@claude

Eldad-Caura:
**Declined as stated — the example does not reproduce. But running it found a real hole beside it, which is fixed.**

### The finding as written is wrong

The claim is that this is *"silently cleared (no finding)"*:

```python
async def m(self, data):
    for key in ("a", "b", "c"):
        def helper(key, value):
            setattr(row, key, value)
        helper(caller_controlled_key, caller_controlled_value)
```

Run against the previous commit, it is reported:

```
>>> _caller_keyed_update_sites(src)
{'helper': 'setattr over caller-supplied keys'}
```

The outer pass does clear it, exactly as described — the key matches the outer loop's literal tuple by name. But `helper` is *also* scanned as its own scope, where there is no enclosing loop at all, and the site surfaces there. That second reading came in round 3, when private helpers stopped being skipped; the reasoning then was the same shape as this finding, which is why it happens to cover it.

`helper` rather than `m` is also the more useful name: the guard belongs where the keys are applied.

### The variant that does reproduce

Substituting a lambda — which the finding does not mention — is a genuine false negative:

```python
async def m(self, data):
    for key in ("a", "b"):
        f = lambda key, value: setattr(row, key, value)
        f(caller_key, caller_value)
```

```
>>> _caller_keyed_update_sites(src)
{}
```

`Lambda` is not a `FunctionDef`, so it is never visited as its own scope and never gets the second reading that saves the `def` form. The outer pass is the only reader, and it clears it. So the underlying mechanism the finding identified — name-only matching across a scope boundary — is real; the example chosen to demonstrate it was the one case already covered.

### Fix

`_enclosing_loops` now stops ascending at a `FunctionDef` / `AsyncFunctionDef` / `Lambda`. A loop vouches only for names it actually binds, and a nested scope rebinds them. This is the second remedy the finding proposed, and it is the one that covers the lambda.

### Verification

| check | result |
|---|---|
| lambda test, previous commit | **fails** (`assert 'm' in {}`) |
| nested-`def` test, previous commit | `helper` present; **`m` fails** — the half the fix adds |
| gate suite | 102 passed (99 → 102) |
| root suite | 5746 passed, 4 skipped, 1 xfailed |
| mypy `scripts/` | clean, 31 files |
| real tree | sites still exactly `entity_update`, `fleet_upsert_node`, `memory_update`; findings `[]` |

Three tests. The third is the guard against fixing this by never clearing anything:

```python
for key in ("fleet_id", "trust_level"):
    setattr(row, key, data[key])
```

still not a site — `agent_add`'s real shape, no function boundary in between. Without it, "stop at a boundary" could degrade into "no loop ever vouches" and the two regression tests would pass against a check that reports everything.

### Rebase

Also rebased onto `94437d1f`. #1150 landed under this branch and removed two `id-addressed-write` entries, so the branch's own allowlist was two entries larger than main's and the ratchet read that as growth:

```
error: the tenant-scope allowlist grew against origin/main:
      + method:lifecycle_audit_finalize
      + route:PATCH /api/v1/storage/lifecycle-audit/{audit_id}
```

An artifact of the stale base, not of this change — CI evaluates the merge and stayed green throughout. Allowlist now 106, `id-addressed-write` 12.

Eldad-Caura:
@claude

Eldad-Caura:
Medium applied. Low is real but not reachable in this file — measured, not assumed.

### Medium — the registered model was never read against the method

Confirmed against the previous commit, using the case the finding describes:

```python
IDENTITY_WRITE_GUARDS["memory_update"] = ("Entity", "_MEMORY_UPDATABLE_FIELDS", True)
```

```
findings: []
```

`memory_update` genuinely references `_MEMORY_UPDATABLE_FIELDS`, so the reference check clears it, and the column check then runs against Entity.

The specific loss is worth naming, because "validates the wrong model" understates it: `Memory.search_vector` is a TSVECTOR maintained by the `memories_search_vector_trigger`, and `Entity` has no TSVECTOR column. So `DERIVED_COLUMN_TYPES` finds nothing to protect, the derived-column arm of the check quietly has no subject, and the entry still reads as registered.

### Fixed by reference, not by the statement's target

The finding proposes recording the model passed to `sql_update(...)` / `pg_insert(...)` and comparing it. That would not cover `entity_update`, which has no statement at all — it is `select(Entity)` followed by `setattr`, and it is the shape that started this check. A statement-target comparison would have to exempt exactly the case it most needs to cover.

Both halves of an entry are now read against the method body, using the machinery already there for the constant:

```python
if model_name not in referenced.get(method, set()):
    errors.append(...)
```

Verified this actually discriminates rather than passing trivially — all three live methods do mention their model, across all three shapes:

| method | shape | mentions model |
|---|---|---|
| `entity_update` | `select` + `setattr` | `Entity` ✓ |
| `memory_update` | `sql_update(...).values(**…)` | `Memory` ✓ |
| `fleet_upsert_node` | `on_conflict_do_update(set_=…)` | `FleetNode` ✓ |

### Low — declined for now, with the measurement

Name collision between `sites` (first-wins) and `_names_referenced` (last-wins) is a real inconsistency. It is not reachable here:

```
total functions: 222
DUPLICATE NAMES in the real file: {}
```

Both maps are built from the same `ast.walk` of the same module, so they can only diverge if two functions share a name. Keying by `(name, lineno)` would also break `IDENTITY_WRITE_GUARDS`, which is written by hand and keyed by name — the registry would have to carry line numbers, which go stale on every edit above them. I would rather leave this than trade a latent inconsistency for a registry that churns. Happy to add the loud-failure assertion instead if you would prefer the collision to be visible rather than argued.

### Verification

| check | result |
|---|---|
| new test, previous commit | **fails** |
| gate suite | 103 passed (102 → 103) |
| root suite | 5747 passed, 4 skipped, 1 xfailed |
| mypy `scripts/` | clean, 31 files |
| real tree | sites still exactly `entity_update`, `fleet_upsert_node`, `memory_update`; findings `[]` |

Rebased onto `ca20fc9f` — main removed four more `id-addressed-write` entries under this branch, so the local ratchet again read the branch's older allowlist as growth. Allowlist 102, `id-addressed-write` down to 8.

Eldad-Caura:
@claude

Eldad-Caura:
Applied. Real, and the worst-shaped one yet: the check that exists to catch #1081/#1118/#1121 would have missed all three of them written with two characters removed.

### Confirmed against the previous commit

```
.values(patch)                              -> {}   # positional dict
.values({k: v for k, v in patch.items()})   -> {}   # positional comprehension
.values(**patch)                            -> {'m': 'UPDATE values(**…) …'}
```

`Update.values()` takes a single dict positionally as well as a `**` spread, and the positional form is the one you *must* write when a key is not a valid Python identifier. The detection lived inside `for kw in node.keywords:`, so it could only ever see the spread. `node.args` was never read.

### Fix

The `.values` test is hoisted out of the keyword loop and now reads both forms:

```python
spread = [kw.value for kw in node.keywords if kw.arg is None]
if any(not _all_literal_keys(fn, v) for v in [*node.args, *spread]):
    sites.setdefault(fn.name, "UPDATE values(…) from caller-supplied keys")
```

A **named** keyword is deliberately still not a site. `kw.arg` being a real column name is the author typing it out, which is the opposite of a caller-keyed dict — and it is what keeps every explicit SET off this list, including `report_update_completed`, which lands on main as part of #1082 while this is in flight. That is pinned by its own test rather than left to inspection, because widening the net is exactly where a check like this starts reporting everything.

### Verification

| check | result |
|---|---|
| the two positional tests, previous commit | **both fail** (`assert 'm' in {}`) |
| the literal-positional and named-keyword guards | pass before and after — they exist to stop the fix over-reporting |
| gate suite | 107 passed (103 → 107) |
| root suite | 5769 passed, 4 skipped, 1 xfailed |
| mypy `scripts/` | clean, 31 files |
| real tree | sites still exactly `entity_update`, `fleet_upsert_node`, `memory_update`; findings `[]` |

Four tests, not two. Widening from "the spread" to "any whole dict passed in" is the kind of change that can silently become "every `.values()` call is a site", and the real tree would not have caught that — all three of its sites were already reported. The two negative tests are what hold the line.

Rebased onto `ef08cc58`.

### Standing count

Seven rounds, nine false negatives, every one in the detection and every one in the direction that reads as coverage. None of the seven verdicts has found a false *positive* — the failure mode of this check has been exclusively silence, which is the argument for having written it.

Eldad-Caura:
@claude

Eldad-Caura:
@cladue

Eldad-Caura:
Medium and the qualified-name Low applied. The collision Low is handled, but not the way it was proposed. The perf suggestion is declined with a measurement.

### Medium — real, dormant, and now a tripwire rather than a silence

Verified: `PostgresService.__mro__` is `(PostgresService, object)`, and all 203 resolvable public methods live in the scanned file. So the hole cannot be reached today — but the asymmetry is exactly as described. `_public_methods` goes through `inspect.getmembers` *specifically* so a mixin method is not skipped; check 4 parses one module's text and would not see it. Only the quieter half would be wrong.

I did not take option (1), walking every module that defines a reachable method. That is real machinery — it needs per-method source files, and then a filter so unrelated functions in a shared module are not reported as sites — built for a case that cannot arise yet. Option (2) instead: a base class on `PostgresService` is now a gate error naming it, so the choice lands in front of whoever creates the situation, when there is a concrete second module to point at.

### Low (collision) — handled as a refusal, not a qualified key

This is the round-6 Low returning, and my measurement still holds: 222 functions in the file, zero shared names. But the repeat is fair, so it is no longer left to argument.

Keying by qualname is the wrong fix here, and specifically because of `IDENTITY_WRITE_GUARDS`: it is written by hand, and a qualified key would push a class path into it that goes stale on every refactor above it. Instead the collision itself is now an error — the file may not define two functions with the same name. That removes the ambiguity rather than encoding it, and keeps the registry readable.

The two are one helper, `_scan_assumption_findings`, because they are one idea: the ground check 4 stands on is "one class, one file, addressable by bare name", and neither half was checked.

### Low (qualified reference) — applied

Confirmed the false positive:

```
_names_referenced("def m(self): x = models.Entity")  ->  {'models', 'service', 'x'}
'Entity' matched: False
```

`_names_referenced` now counts `ast.Attribute.attr` too. Worth being explicit that this makes the check *weaker*: `foo.Entity` on an unrelated object would now count as mentioning `Entity`. That is the right direction for this particular question — it is a loose "is this name spoken here at all", and its failure mode should not be a red gate whose fix is reverting an import style.

### Suggestion (quadratic parent map) — declined, measured

```
file: 10,600 lines
_caller_keyed_update_sites:      55.5 ms
_identity_writability_findings:  351.7 ms   (the whole check)
```

A third of a second, once, in a lint pass. Caching the parent map per function would add state to a function whose current appeal is that it holds none. Happy to revisit if the file doubles.

### Verification

| check | result |
|---|---|
| all four new tests, previous commit | **all fail** |
| gate suite | 111 passed (107 → 111) |
| root suite | 5773 passed, 4 skipped, 1 xfailed |
| mypy `scripts/` | clean, 31 files |
| real tree | sites still exactly `entity_update`, `fleet_upsert_node`, `memory_update`; findings `[]` |

One test asserts the tripwires are silent on the tree as it stands — without it, two checks that always fired would read as coverage.

Rebased onto `a70115d3`. One flake in the first root-suite run, `test_memory_update_rejects_unknown_field`; that is #1137 and I have its traceback now — `DUPLICATE_MEMORY` / `reason: semantic_similarity`, 409 where the test wants 201, before the PATCH under test is reached. Clean on re-run.

Eldad-Caura:
@claude

erni-a (approved):

PR #1144, opened 2026-08-31
fix(storage): scope agent search-profile updates to tenant

### LINKED ISSUE #1084: storage: PATCH /agents/{agent_id}/search-profile writes any tenant's agent profile
`route:PATCH /api/v1/storage/agents/{agent_id}/search-profile`

Service method: `method:agent_update_search_profile`

### The defect

```python
async def update_search_profile(agent_id: str, request: Request) -> dict:
    await _svc.agent_update_search_profile(agent_id, body["search_profile"])
```

Compare the sibling reset route in the same file, which DOES scope: `agent_get_by_id(agent_id, body["tenant_id"])` before acting. The inconsistency between the two is the clearest evidence this is an oversight rather than a design.

### The fix

Mirror the reset route: look the agent up with `agent_get_by_id(agent_id, tenant_id)`, 404 when it is not the caller's, then write. Better, push the predicate into `agent_update_search_profile` so the method is safe for any caller.

---

### Why this is filed

Eldad-Caura:
## Summary

- require the binding `tenant_id` throughout the search-profile update call chain
- bind `Agent.id` and `Agent.tenant_id` together in `agent_update_search_profile`
- keep the update SET fixed to `search_profile`; no caller-controlled keys can repoint the primary key or another identity column
- remove #1084 method and route entries through `tenant_scope_gate.py --write`

The current core API and MCP callers already obtain the agent primary key from a tenant-scoped lookup. This change moves the guarantee into the storage query as well, making the direct storage route and future callers safe without relying on that caller behavior.

## Regression proof

The new negative service test was run with only the tenant predicate removed. It failed as intended:

```text
E   AssertionError: assert {'min_similarity': 0.22} == {'min_similarity': 0.61}
E     Differing items:
E     {'min_similarity': 0.22} != {'min_similarity': 0.61}
```

With the predicate restored, both new tests pass.

## Allowlist

- total: 110 -> 108
- `id-addressed-write`: 16 -> 14
- `opaque-body-write`: 35 -> 35, with all 35 hand-written notes preserved

## Validation

- `core-storage-api/tests/`: 244 passed
- root `tests/` excluding benchmarks: 5694 passed, 4 skipped, 16 deselected, 1 xfailed
- focused core API and MCP tests: 28 passed
- Ruff check and format check
- mypy: core-storage-api 85 files, core-api 203 files
- tenant-scope gate: 108 allowlisted, 14 id-addressed writes

Closes #1084

### DIFF
--- a/core-api/src/core_api/clients/storage_client.py
+++ b/core-api/src/core_api/clients/storage_client.py
@@ -1639,8 +1639,16 @@ class CoreStorageClient:
     async def update_trust_level(self, agent_id: str, data: dict) -> dict | None:
         return await self._patch(f"/agents/{agent_id}/trust-level", data)
 
-    async def update_search_profile(self, agent_id: str, data: dict) -> dict | None:
-        return await self._patch(f"/agents/{agent_id}/search-profile", data)
+    async def update_search_profile(
+        self,
+        agent_id_pk: str,
+        tenant_id: str,
+        search_profile: dict,
+    ) -> dict | None:
+        return await self._patch(
+            f"/agents/{agent_id_pk}/search-profile",
+            {"tenant_id": tenant_id, "search_profile": search_profile},
+        )
 
     async def reset_search_profile(
         self,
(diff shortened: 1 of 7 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

erni-a (approved):

PR #1145, opened 2026-08-31
fix(documents): refuse an upsert that guts a stored document (C34)

arkash20:
## Summary

The document upsert is **last-writer-wins over the whole `data` blob with no history**, so a caller that computes an empty or truncated payload destroys the stored document — and gets a `200` for it.

That is not hypothetical. On 2026-08-27 a failed GET left a client file empty, the follow-up upsert of those 0 bytes **replaced a 105KB shared task checklist**, and recovery meant reconstructing the document by hand out of session transcripts; one concurrent edit in that window was lost for good.

## The guard

Server-side, in the storage service: reject an upsert whose payload is under **10%** of the stored document, when the stored document is at least **2KB**. The response is a `400` naming both sizes and the override:

```
refusing to shrink document 'caura_shared_tasks/v4-task-checklist' from 105233 to 15 bytes
(0.0% of the stored size). A truncated payload from a failed read looks exactly like this.
Retry with force=true if the shrink is intended.
```

Ordinary editing is untouched — even halving a document passes. `force: true` on the write body is the deliberate escape hatch.

**Both upsert paths are guarded.** The `xmax` path is the one **indexed** documents take (the checklist among them), so guarding only its sibling would have missed the exact incident that motivated this.

## Bonus fix the wet test caught

Storage's `400` reached the caller as an opaque **500**, hiding both the reason and the remedy. The documents routes now carry a storage 4xx through with its status and message, matching what the keystones routes already do.

## Testing

- 6 unit tests: ratio/floor constants, catastrophic-shrink math, ordinary-rewrite and small-document exemptions, both paths guarded, `force` threaded end to end.
- Full suite **5716 passed**; mypy clean on core-api and core-storage-api; ruff clean.
- **Wet test on the live stack replays the incident**: store 25KB doc → empty upsert **refused 400** → document **intact at 25,050 bytes** → `force=true` wipe allowed → halving edit allowed.

## Follow-up (deliberately not here)

Version history per `(collection, doc_id)` with a `?version=` readback, so recovery is a request rather than an archaeology session. The guard prevents the common cause; history is the safety net for everything else.

Refs: canonical C34.

### DIFF
--- a/core-api/src/core_api/routes/documents.py
+++ b/core-api/src/core_api/routes/documents.py
@@ -4,6 +4,7 @@ import logging
 import re
 from datetime import datetime
 
+import httpx
 from fastapi import APIRouter, Depends, Header, HTTPException, Query, Request, Response
 from fastapi.responses import JSONResponse
 from pydantic import BaseModel, Field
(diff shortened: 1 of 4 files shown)


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #1146, opened 2026-08-31
CAURA-719 refactor(enrichment): stop asking the LLM for tags and status

eyal-bl:
## Summary

Two of the enrichment prompt's 13 fields cost tokens on every write and neither earned them. This removes both from the prompt while keeping their schema entries intact.

**`tags` — no reader.** Nothing filters, queries, or full-text-indexes them (the search vector is `title || content`, migration 034). A measured **95.5% of the eToro corpus (46,203 / 48,374 rows)** carried generated tags that only ever travelled back out to the caller. C25 already made `tags` a `CALLER_OWNABLE_KEY`, so dropping the platform's copy leaves a coherent field rather than a gap — callers who want tags supply them, and A8's normaliser still applies to those.

**`status` — a measured constant.** All 48,374 rows carry `active`; the classifier never once chose `pending` or `confirmed`. It is also a live hazard: ~11 query paths filter the literal string `'active'` rather than `LIVE_MEMORY_STATUSES`, so a memory the LLM labels `confirmed` goes invisible to insights. That was observed in production — a runbook body produced `status="confirmed"` and insights returned `memories_analyzed=0` — and is already worked around by pinning `status="active"` in both `doc_memory.py` and `insights_service.py`. `status` stays a lifecycle field owned by explicit setters: `caura_manage`'s transition op, the contradiction detector (`outdated` / `conflicted`), the crystallizer (`archived`), and the delete path (`deleted`).

## The schema is deliberately wider than the prompt

Both fields keep their `EnrichmentResult` entry, default, and validator. Three consumers depend on that: historical rows, caller-supplied values, and the deferred worker's patch path. `test_caura719_tags_status_off_prompt.py` pins both directions.

## Changes

- `common/enrichment/_prompts.py` — fields 5 (`tags`) and 6 (`status`) removed; remaining 11 renumbered; the `atomic_facts` → `retrieval_hint` cross-reference follows the renumbering; module docstring records the reasoning
- Agent-facing surface stops promising tag generation: `caura_write` description, MCP server instructions, `README.md`, `docs/api-reference.md`, `static/docs/integration-guide.md`
- Regenerated artifacts: `plugin/tools.json` (via `scripts/export_tool_specs.py`) and the three MCP baselines (via `capture_baselines.py`). **tools/list drops to 5,333 cl100k tokens**, under the 5,350 ceiling
- `tests/test_a8_enrichment_tags.py` — A8's five prompt-tightening tests retired and replaced with the negative assertion; **all nine validator tests kept**, since they now guard caller-supplied and historical values
- `tests/test_enrichment_prompt.py` — `test_prompt_structure_unchanged` pins `summary` instead of the retired `tags` key
- `tests/test_caura719_tags_status_off_prompt.py` — new, 16 tests

## Testing

```
303 passed, 6 failed (all 6 pre-existing on main @ 841a6967, verified by
running the same selection on a detached origin/main)
```

Pre-existing failures, unrelated to this change: `test_fast_mode_creates_memory_with_pending_enrichment`, `test_memory_count_includes_enriched_statuses`, `test_bulk_write_demotes_new_deprecated_slug_to_default[intention|commitment|cancellation]`, `TestGateRepairLoop::test_repair_prompt_caps_violations`.

`ruff check` and `ruff format --check` clean on all touched files.

## Deliberately out of scope

Three adjacent defects found while investigating, each left for its own PR to keep this diff scoped:

1. The `"weight": 0.0` value in the JSON template sits outside every band the weight rubric describes (`0.3`–`1.0`), so the template exemplifies a value the spec never sanctions.
2. `status` has no reserved-value guard analogous to CAURA-699's for `memory_type` — an LLM emitting `outdated` or `deleted` would pass the `MEMORY_STATUSES` membership test. Lower risk now that the field isn't requested, but the gap remains.
3. `cancelled` has no setter anywhere in the codebase and no reader beyond membership checks.


eyal-bl:
@claude

ran-taig (approved):

Issue #1091, opened 2026-08-30
storage: entity_delete_entity_links is unreachable dead code — delete it


Eldad-Caura:
`method:entity_delete_entity_links`

### The defect

`entity_delete_entity_links` is unscoped (`session.delete` on links fetched without a tenant predicate), but a repo-wide search finds **no caller at all** — no route, no service, no test, nothing outside its own definition.

So unlike the other entries in this set it is not a live hazard; it is an unscoped destructive method sitting in the class waiting for someone to wire it up, which is arguably worse than a used one because it will be adopted by whoever needs 'the existing helper'.

### The fix

Delete the method. If it is wanted later it should be written with a tenant parameter from the start. Deleting it also removes its `id-addressed-write` allowlist entry, shrinking the backlog by one at no behavioural cost — the cheapest item in this set.

---

### Why this is filed

`core-storage-api` authenticates no request (GHSA-wgvw-28pq-jc36), and docker-compose publishes it on `0.0.0.0:8002`. Anything reachable there is reachable by anyone who can reach the port. A bare primary key is therefore not an authorisation check: knowing a UUID is not the same as being entitled to the row behind it.

Tracked by the tenant-scope gate added in caura-ai/caura#1069 — this path is in `core-storage-api/tenant_scope_allowlist.json` under `id-addressed-write`, the category the gate flags `<- backlog, MUTATES`. The gate's ratchet means the allowlist can shrink but never grow, and it refuses to let a `id-addressed-write` entry be relabelled as something milder. **Fixing this path means deleting its allowlist entry; CI will fail until the entry is removed, which is the intended forcing function.**

PR #1148, opened 2026-08-31
fix(storage): bind memory-entity links to the caller's tenant

### LINKED ISSUE #1085: storage: PATCH /memories/{memory_id}/entities links entities onto any tenant's memory
`route:PATCH /api/v1/storage/memories/{memory_id}/entities`

Service method: `method:memory_add_entity_links`

### The defect

The handler validates the shape of `entity_links` carefully — list, `entity_id`, non-empty string `role` — and reads no tenant. `memory_add_entity_links` then upserts the rows against `memory_id`.

### The fix

Require `tenant_id` and confirm the memory belongs to it before linking. Worth checking the `entity_id` values too: as written a caller can also link ANOTHER tenant's entities onto a memory, so the ownership check is needed on both sides of the link.

---

### Why this is filed

Eldad-Caura:
Closes #1085.

`PATCH /api/v1/storage/memories/{memory_id}/entities` validated the shape of `entity_links` carefully — list, `entity_id`, non-empty string `role` — and read no tenant at all. `memory_add_entity_links` then upserted rows against a bare `memory_id` with bare `entity_id` values.

`core-storage-api` authenticates no request (GHSA-wgvw-28pq-jc36) and docker-compose publishes it on `0.0.0.0:8002`, so a bare primary key is not an authorisation check: knowing a UUID is not being entitled to the row behind it.

### No schema change

`memory_entity_links` has no `tenant_id`, which means only that the join row cannot carry a predicate directly. Both parents carry one — `Memory.tenant_id` (`common/models/memory.py:30`) and `Entity.tenant_id` (`common/models/entity.py:18`), both `Text, nullable=False, index=True`. So the fix is an existence check on the two ends: ordinary rows, ordinary `WHERE`, no DDL.

### Does the method need to exist?

Asked first, because the precedent on this exact table is #1091, closed 30 August: `entity_delete_entity_links` was **deleted, not scoped**, once a search found no caller at all.

`memory_add_entity_links` is not that case. It has exactly one caller — `PATCH /memories/{memory_id}/entities` — reached from `sc.update_memory_entities`, called from `memory_service.update_memory` (`core-api/src/core_api/services/memory_service.py:3753`) whenever a `MemoryUpdate` carries `entity_links`. Live, single-purpose, not duplicated by the bulk path. It gets the check.

Established by searching for the method, the client wrapper and the route path separately rather than reading the route — the defect class here is precisely that the tenancy guarantee lived in one caller and never travelled. `update_memory` has `tenant_id` in its signature and never forwarded it.

### Both ends, separately

Neither check implies the other, and each fails on its own:

- Skip the **entity** check and a caller staples a foreign entity onto their own memory, pulling someone else's graph node into their namespace.
- Skip the **memory** check and they staple their own entity onto a foreign memory.

`test_foreign_memory_is_not_linked` and `test_foreign_entity_is_not_linked` are separate cases for that reason. Reverting one predicate fails only its own direction — pasted below.

A mixed batch writes nothing rather than landing the caller's own links beside the refused one, so a partial success cannot leave the caller unable to tell which edges exist. The entity check is a set difference rather than a count, so repeating one owned id cannot pad a foreign one through.

One `404` covers "no such memory", "not your memory" and "not your entity" alike, and its detail names only the memory — distinguishing them would answer as an existence oracle for two id spaces on a service that authenticates nothing.

### Failing without the fix

Reverting **only** the memory-side predicate in `_owned_link_endpoints` (`Memory.tenant_id == tenant_id` dropped from the `WHERE`):

```
F....F                                                                   [100%]
______ TestMemoryEntityLinkTenantScope.test_foreign_memory_is_not_linked _______
core-storage-api/tests/test_memory_entity_links_tenant_scope.py:102: in test_foreign_memory_is_not_linked
    assert resp.status_code == 404, resp.text
E   AssertionError: {"ok":true}
E   assert 200 == 404
E    +  where 200 = <Response [200 OK]>.status_code
_ TestMemoryEntityLinkTenantScope.test_a_foreign_id_is_indistinguishable_from_a_missing_one _
core-storage-api/tests/test_memory_entity_links_tenant_scope.py:219: in ...
    assert [r.status_code for r in memory_answers] == [404, 404]
E   assert [200, 404] == [404, 404]
E     At index 0 diff: 200 != 404
2 failed, 4 passed in 0.48s
```

The entity-side tests still pass. Reverting **only** the entity-side predicate (`Entity.tenant_id == tenant_id` dropped):

```
.FF..F                                                                   [100%]
______ TestMemoryEntityLinkTenantScope.test_foreign_entity_is_not_linked _______
core-storage-api/tests/test_memory_entity_links_tenant_scope.py:126: in test_foreign_entity_is_not_linked
    assert resp.status_code == 404, resp.text
E   AssertionError: {"ok":true}
E   assert 200 == 404
_ TestMemoryEntityLinkTenantScope.test_one_foreign_entity_voids_the_whole_request _
core-storage-api/tests/test_memory_entity_links_tenant_scope.py:154: in ...
    assert resp.status_code == 404, resp.text
E   AssertionError: {"ok":true}
E   assert 200 == 404
_ TestMemoryEntityLinkTenantScope.test_a_foreign_id_is_indistinguishable_from_a_missing_one _
core-storage-api/tests/test_memory_entity_links_tenant_scope.py:235: in ...
    assert [r.status_code for r in entity_answers] == [404, 404]
E   assert [200, 404] == [404, 404]
3 failed, 3 passed in 0.49s
```

The memory-side test still passes. Each half is independently load-bearing, demonstrated in both directions — `{"ok":true}` is the write that went through.

### Allowlist

102 → 100 entries; `id-addressed-write` 8 → 6. Both #1085 entries (`method:memory_add_entity_links`, `route:PATCH /api/v1/storage/memories/{memory_id}/entities`) removed and regenerated with `python3 scripts/tenant_scope_gate.py --write`, as the gate's stale-entry check (`tenant_scope_gate.py:1060-1063`) requires. Gate green; hand-written `note` and `category` fields elsewhere in the file are untouched.

The baseline moved twice while this was open — it was 108 when the branch was cut, then #1150 removed two entries and #1152 removed four. Rebased onto each and regenerated; the *contribution* of this PR is unchanged at two entries.

### Precedent for the five `id-addressed-read` siblings

Not fixed here and not mine to fix, but this is the shape they should follow, since eight allowlist entries share the note *"Join table has no tenant column; rows are tenant-scoped only via the ids"*:

- **Existence check on both parents, not a schema change.** The absent `tenant_id` on the join table is not a blocker; `Memory.tenant_id` and `Entity.tenant_id` are enough.
- **Scope every end the caller names.** For the reads that take `memory_ids`, that is the memory side; for `entity_get_memory_ids_by_entity_ids`, the entity side; `entity_find_entity_link` names both and should check both.
- **One answer for "absent" and "not yours"** — empty/`404`, never a distinguishable status or message, or the route becomes an existence oracle for UUIDs on a service that authenticates nothing.
- **The check belongs in the service method**, not the route, so it cannot drift away from the query.
- **Two tests per method, one per end**, each verified to fail with only its own predicate reverted. A single-direction test passes while half the hole stands open.

- **`_owned_link_endpoints` is the reusable piece.** It takes sets and returns the owned subsets, so a read resolving many ids can filter without a query each. The reads should use it rather than growing their own variant.

One caveat for whoever takes them: `POST /memories/entity-links` is currently the read oracle in this PR's test helper (`_links`) precisely because it is still unscoped. When it gains a required tenant, that helper needs one too — there is a comment on it saying so.

**Since #1152 merged, this PR now shares its ownership check with the two `POST /entities/links` paths** rather than carrying an inline copy — the follow-up that PR flagged, done here instead of left outstanding. All three writers to this join table now resolve ownership through one helper, so the revert proofs below exercise the code path #1124's routes use too.

### Verification

- `core-storage-api/tests/`: 264 passed (6 new).
- Root `tests/`: 5716 passed, 4 skipped, 1 xfailed.
- `ruff check` and `ruff format --check` at CI's exact scopes: clean.
- Tenant-scope gate against `origin/main`: green, "Allowlist shrank by 2".

Run against a dedicated local pgvector instance. **Not checked:** anything requiring CI's own environment — the GitHub Actions matrix, the coverage floors step, and the enterprise re-vendor path.

### DIFF
--- a/core-api/src/core_api/clients/storage_client.py
+++ b/core-api/src/core_api/clients/storage_client.py
@@ -966,11 +966,15 @@ class CoreStorageClient:
     async def update_memory_entities(
         self,
         memory_id: str,
+        tenant_id: str,
         entity_links: list[dict],
     ) -> dict | None:
+        # Same contract as ``update_memory`` above, for the same reason: the
+        # memory is addressed by bare UUID, so the tenant has to travel with it.
+        # Storage scopes the memory AND every named entity to it.
         return await self._patch(
             f"/memories/{memory_id}/entities",
-            {"entity_links": entity_links},
+            {"tenant_id": tenant_id, "entity_links": entity_links},
         )
 
     async def get_entity_links_for_memories(
--- a/core-api/src/core_api/services/memory_service.py
+++ b/core-api/src/core_api/services/memory_service.py
@@ -3750,7 +3750,7 @@ async def update_memory(
         entity_link_dicts = [
             {"entity_id": str(link.entity_id), "role": link.role} for link in data.entity_links
         ]
-        await sc.update_memory_entities(str(memory_id), entity_link_dicts)
+        await sc.update_memory_entities(str(memory_id), tenant_id, entity_link_dicts)
         changes["entity_links"] = {
             "old": "replaced",
             "new": f"{len(data.entity_links)} links",
--- a/core-storage-api/src/core_storage_api/routers/memories.py
+++ b/core-storage-api/src/core_storage_api/routers/memories.py
@@ -1827,6 +1827,12 @@ async def update_embedding(memory_id: UUID, request: Request) -> dict:
 @router.patch("/{memory_id}/entities")
 async def update_memory_entities(memory_id: UUID, request: Request) -> dict:
     body: dict = await request.json()
+    # Tenant guard, same shape as ``PATCH /memories/{memory_id}/embedding``
+    # above. This route validated the shape of ``entity_links`` carefully and
+    # read no tenant at all, so a caller who knew a memory UUID could hang
+    # entities off another tenant's row — and name another tenant's entities
+    # while doing it. Both endpoints of the link are scoped in the service.
+    tenant_id = _require(body, "tenant_id")
     entity_links = body.get("entity_links", [])
     if not isinstance(entity_links, list):
         raise HTTPException(status_code=422, detail="'entity_links' must be a list")
@@ -1836,7 +1842,13 @@ async def update_memory_entities(memory_id: UUID, request: Request) -> dict:
         raise HTTPException(status_code=422, detail=str(exc)) from exc
     if any(not isinstance(lnk["role"], str) or not lnk["role"] for lnk in links):
         raise HTTPException(status_code=422, detail="Each entity link must have a non-empty string 'role'")
-    await _svc.memory_add_entity_links(memory_id, links)
+    linked = await _svc.memory_add_entity_links(memory_id, tenant_id, links)
+    if not linked:
+        # One 404 for a memory that isn't there, a memory that isn't yours, and
+        # an entity that isn't yours. The detail names only the memory because
+        # saying which of the three it was would answer as an existence oracle
+        # for both id spaces — see the service docstring.
+        raise HTTPException(status_code=404, detail=f"Memory {memory_id} not found")
     return {"ok": True}
 
 
--- a/core-storage-api/src/core_storage_api/services/postgres_service.py
+++ b/core-storage-api/src/core_storage_api/services/postgres_service.py
@@ -4020,25 +4020,70 @@ class PostgresService:
     async def memory_add_entity_links(
         self,
         memory_id: UUID,
+        tenant_id: str,
         links: list[dict],
-    ) -> None:
+    ) -> bool:
+        """Link entities to one memory, both endpoints scoped to ``tenant_id``.
+
+        Returns ``False`` when the memory is not in ``tenant_id`` or any named
+        entity is not, having written nothing; ``True`` when the links are in.
+
+        ``memory_entity_links`` has no ``tenant_id`` of its own, so the join row
+        cannot carry a predicate — but both parents do (``Memory.tenant_id``,
+        ``Entity.tenant_id``), which is what makes an existence check sufficient
+        and a schema change unnecessary.
+
+        **Both sides are load-bearing and neither implies the other.** Checking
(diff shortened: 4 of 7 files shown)


Eldad-Caura:
Force-pushed. Two things were wrong and both are fixed:

**Legacy-name ratchet (failing).** The new test's docstring named the `memclaw` / `memclaw_storage` databases while explaining which one the sweep reaches. Reworded to describe them without the literals rather than marking them `legacy-name-floor:` — the ratchet's own guidance is to reword where rewording leaves the line correct, and it does here. Ratchet now reports "no new lines currently fail it".

**Conflict.** The base moved twice (#1150, then #1152). Rebased onto `origin/main` and regenerated the allowlist from source rather than hand-merging the JSON. Now 102 → 100, `id-addressed-write` 8 → 6; this PR's own contribution is still exactly its two #1085 entries.

One substantive change while rebasing, worth a look: now that #1152 has merged, `memory_add_entity_links` uses the shared `_owned_link_endpoints` helper instead of the inline copy it carried before. That was the follow-up #1152's description flagged, and it means all three writers to `memory_entity_links` resolve ownership through one function. Both revert proofs were re-run against the helper and still fail in exactly one direction each — pasted in the updated description.

Re-verified after the rebase: storage suite 264 passed, root suite 5716 passed, ruff + mypy clean at CI's scopes, tenant-scope gate green.

@claude @gemini — please re-review; the `synchronize` skip is silent, so this is the re-trigger.

arkash20 (approved):

PR #1149, opened 2026-08-31
fix(contradiction): decouple the judge's provider from entity extraction (C2, A32)

arkash20:
## Summary

Two rows from the consolidated list, one subsystem.

### C2 — decouple the judge's provider from entity extraction

The contradiction judge read the **entity-extraction** provider and model config directly, at four call sites. An operator who moved entity extraction to a cheaper model silently moved the contradiction judge with it, and had no way to tune the two independently.

`CONTRADICTION_PROVIDER` / `CONTRADICTION_MODEL` now override for judging only. Empty — the default — resolves exactly as before, so this is a **no-op until an operator sets one**. All four sites route through one helper, and a test asserts the entity-provider read survives in exactly one place (the helper's own fallback), so a future call site can't quietly bypass the split.

### A32 — verified, deliberately no code change

The row says forward Path-A detection acts on weak verdicts because it discards confidence, while the retraction path has a 0.90 gate. **That premise does not hold in current code.** `_judge_contradiction` forces the verdict to `False` on every low-confidence branch (gate 1, gate 2, malformed), so an LLM-derived `True` always carries `_CONF_CLEAN` — probed exhaustively over the judge's whole input space in the new test.

I implemented the confidence gate anyway, and the test suite caught that it **changed behaviour for exactly one producer**: the *deliberate* fake provider, whose `True` verdicts carry `_CONF_FALLBACK`. Gating those away breaks the dev/CI stand-in and makes nothing safer, so the gate was reverted.

What ships instead is the invariant itself — pinned by test, plus a comment at the bare `if verdict:` explaining why it's safe. If someone later changes the confidence rubric so a weak `True` becomes possible, the test fails loudly rather than the system silently marking rows `conflicted` on a weak judgment.

## Note for reviewers

The wholesale `settings` patches in `test_p1_contradiction.py` now declare the two new fields. An undeclared attribute on a `MagicMock` reads back **truthy**, so it would have been used as a provider name — that's how this surfaced.

## Testing

8 new unit tests; full suite **5718 passed**; ruff + mypy clean.

**Wet test** on the live stack with `ENTITY_EXTRACTION_PROVIDER=openai` and `CONTRADICTION_PROVIDER=fake` — the deployed image resolves the judge to `fake` while extraction stays on `openai`:

```
env CONTRADICTION_PROVIDER = 'fake'
env ENTITY_EXTRACTION_PROVIDER = 'openai'
judge provider resolves to -> fake
judge model attr resolves to -> entity_extraction_model
```

(The end-to-end judge path couldn't be exercised in this environment — the shared OpenAI backend was erroring and fake embeddings produce no semantic neighbours — so the split is proven at the resolution point in the deployed image rather than through a live contradiction.)

Refs: canonical C2, A32.

### DIFF
--- a/core-api/src/core_api/config.py
+++ b/core-api/src/core_api/config.py
@@ -124,6 +124,14 @@ class Settings(BaseSettings):
     # None (the default) sends no parameter at all — REQUIRED for
     # non-reasoning models, which reject the parameter outright.
     contradiction_reasoning_effort: str | None = None
+    # C2 — the contradiction judge historically read the ENTITY-EXTRACTION
+    # provider/model config, so an operator who moved entity extraction to a
+    # cheaper or different model silently moved the contradiction judge with
+    # it, and could not tune the two independently. These override the
+    # entity-extraction values for contradiction judging only; empty keeps
+    # the historical behaviour exactly.
+    contradiction_provider: str = ""
+    contradiction_model: str = ""
     # Default for the ``search.entity_retrieval`` org setting: query-time entity
     # lookup + graph search. A tenant override wins; this is the fleet-wide
     # fallback so an operator can disable entity/graph reads on a whole box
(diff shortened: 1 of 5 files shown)


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #1150, opened 2026-08-31
fix(storage): scope lifecycle audit finalization to org

### LINKED ISSUE #1083: storage: PATCH /lifecycle-audit/{audit_id} finalizes any org's audit row by id
`route:PATCH /api/v1/storage/lifecycle-audit/{audit_id}`

Service method: `method:lifecycle_audit_finalize`

### The defect

`audit_id` is a plain integer — not even a UUID — so the id space is trivially enumerable. The handler validates `status` against `_VALID_STATUSES` but reads no tenant or org.

### The fix

This table keys on `org_id` rather than `tenant_id` (see `BINDING_SCOPE` in the gate). Require `org_id` on the route and scope the UPDATE by it. The sequential integer id makes this the cheapest of the set to exploit and argues for p1 if these rows drive anything operational.

---

### Why this is filed

Eldad-Caura:
## Summary

- Require a non-empty `org_id` on `PATCH /api/v1/storage/lifecycle-audit/{audit_id}` and bind both the status UPDATE and sticky-success existence check to that org.
- Thread the audit row's org through the core API, shared lifecycle handlers, core worker, embed-backfill path, and org-deletion finalizer.
- Remove the fixed method and route from the tenant-scope allowlist while preserving the opaque-body-write annotations.

## Related Issue

Closes #1083.

## Type of Change

- [ ] Documentation update
- [ ] Other:

## How Has This Been Tested?

The cross-org route regression fails when only the new `LifecycleAudit.org_id == org_id` UPDATE predicate is removed:

```text
>       assert response.status_code == 404, response.text
E       AssertionError: {"ok":true,"noop":false}
E       assert 200 == 404
```

With the fix restored:

- `core-storage-api/tests/test_lifecycle_audit_finalize_tenant_scope.py` plus the existing lifecycle observation tests: 10 passed
- `tests/test_lifecycle_handlers.py`: 17 passed
- Full root suite: 5,694 passed, 4 skipped, 1 xfailed, 16 deselected
- Full core-storage-api suite: 247 passed
- Full core-worker suite: 80 passed; 71.43% coverage
- Ruff check and format check: core-api, core-storage-api, core-worker, and common all pass
- Mypy: core-api (203 files), core-storage-api (85), core-worker (11), and common (99) all pass
- Package builds: core-api, core-storage-api, and core-worker all pass
- `scripts/tenant_scope_gate.py`: passes with 106 allowlisted entries

## Checklist

- [x] `mypy` passes
- [x] `pytest` passes locally

## Additional Notes

- Allowlist total: 108 before, 106 after. `id-addressed-write`: 14 before, 12 after.
- All 35 `opaque-body-write` entries and all 35 hand-written notes survived regeneration.
- The UPDATE values are assembled from the literal keys `status`, `finished_at`, `stats`, and `error_message`; neither `id` nor `org_id` is caller-writable through the SET clause.

### DIFF
--- a/common/events/lifecycle_handlers.py
+++ b/common/events/lifecycle_handlers.py
@@ -57,6 +57,7 @@ class ArchiveStorageAdapter(Protocol):
         self,
         audit_id: int,
         *,
+        org_id: str,
         status: str,
         stats: dict | None = None,
         error_message: str | None = None,
@@ -116,6 +117,7 @@ class PipelineStorageAdapter(Protocol):
         self,
         audit_id: int,
         *,
+        org_id: str,
         status: str,
         stats: dict | None = None,
         error_message: str | None = None,
@@ -218,6 +220,7 @@ async def _run_action(
             try:
                 await adapter.update_lifecycle_audit_row(
                     audit_id,
+                    org_id=org_id,
                     status="success",
                     stats={"skipped": True, "reason": "recent_success"},
                 )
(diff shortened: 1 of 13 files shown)


erni-a (approved):

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #1154, opened 2026-08-31
fix(storage): scope dedup review decisions to tenant

### LINKED ISSUE #1087: storage: POST /memories/dedup-reviews/{review_id}/decision decides any tenant's review
`route:POST /api/v1/storage/memories/dedup-reviews/{review_id}/decision`

Service method: `method:dedup_review_decide`

### The defect

The handler validates `status` is a string and passes `review_id` and `decided_by` straight through. `dedup_review_decide` sets `status`, `decided_at` and `decided_by` on the row by primary key.

`decided_by` is also caller-supplied and unvalidated, so an attacker can attribute the decision to someone else.

### The fix

Require `tenant_id`, scope the row fetch by it, and derive or validate `decided_by` rather than taking it verbatim.

---

### Why this is filed

Eldad-Caura:
## Summary

- Require a non-empty `tenant_id` throughout the dedup-review decision call chain and fetch the review by both primary key and tenant.
- Return the same 404 for foreign and missing reviews, so a caller cannot decide another tenant's review by UUID.
- Reject caller-supplied `decided_by`. The storage seam has no end-user identity to derive a reviewer from, so decisions remain explicitly unattributed instead of persisting an unvalidated identity claim.
- Remove the fixed method and route from the tenant-scope allowlist while preserving every hand-written opaque-body-write annotation.

There is no current production caller of this endpoint in the repository; the existing integration callers now pass their binding tenant. This change puts the guarantee in the route and query for direct and future callers.

## Related Issue

Closes #1087.

## Type of Change

- [ ] Documentation update
- [ ] Other:

## How Has This Been Tested?

The wrong-tenant regression was run with only the new `DedupReview.tenant_id == tenant_id` predicate removed. It failed as intended and showed that the foreign review had been changed to `dismissed`:

```text
>       assert response.status_code == 404, response.text
E       assert 200 == 404
```

With the predicate restored:

- Final focused storage regression: 3 passed
- Existing dedup-review queue tests plus #1153's focused guard tests: 18 passed
- Full core-storage-api suite: 267 passed
- Full root suite: 5,700 passed, 4 skipped, 16 deselected, 1 xfailed
- Full core-operations suite: 50 passed; full core-worker suite: 80 passed
- Ruff check and format check: all CI Python scopes pass
- Mypy: core-api, core-storage-api, core-operations, core-worker, common, and scripts all pass
- Package builds: core-api and core-storage-api pass
- Coverage floors: core-api 85%, core-storage-api 66%; combined root coverage 80%
- `scripts/tenant_scope_gate.py`: passes with 98 allowlisted entries

The full suites ran after rebasing onto #1148, #1151, and #1152. #1153 landed during the final check, touched only `memory_service.py` and its guard test, and rebased without overlap; the final focused storage regression, existing dedup-review tests, and #1153 guard tests were then rerun on the exact pushed commit.

## Checklist

- [x] `mypy` passes
- [x] `pytest` passes locally
- [x] I have updated relevant documentation (no documentation changes are needed for this internal storage boundary)

## Additional Notes

- Allowlist total: 100 before, 98 after. `id-addressed-write`: 6 before, 4 after.
- All 35 `opaque-body-write` entries and all 35 hand-written notes survived regeneration.
- The decision UPDATE writes only the fixed fields `status`, `decided_by`, and `decided_at`; neither `id` nor `tenant_id` is caller-writable.

### DIFF
--- a/core-api/src/core_api/clients/storage_client.py
+++ b/core-api/src/core_api/clients/storage_client.py
@@ -791,12 +791,12 @@ class CoreStorageClient:
             **{k: v for k, v in params.items() if v is not None},
         )
 
-    async def decide_dedup_review(self, review_id, status: str, *, decided_by: str | None = None) -> dict:
+    async def decide_dedup_review(self, review_id, status: str, *, tenant_id: str) -> dict:
         """Record a terminal decision (confirmed_duplicate /
         override_not_duplicate / dismissed) on a review row."""
         return await self._post(  # type: ignore[return-value]
             f"/memories/dedup-reviews/{review_id}/decision",
-            {"status": status, "decided_by": decided_by},
+            {"status": status, "tenant_id": tenant_id},
             read=False,
         )
 
--- a/core-storage-api/src/core_storage_api/routers/memories.py
+++ b/core-storage-api/src/core_storage_api/routers/memories.py
@@ -1116,12 +1116,28 @@ async def list_dedup_reviews(
 @router.post("/dedup-reviews/{review_id}/decision")
 async def decide_dedup_review(review_id: UUID, request: Request) -> dict:
     body: dict = await request.json()
+    if not isinstance(body, dict):
+        raise HTTPException(status_code=422, detail="request body must be a JSON object")
+    tenant_id = body.get("tenant_id")
+    if not isinstance(tenant_id, str) or not tenant_id:
+        raise HTTPException(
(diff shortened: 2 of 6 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #1158, opened 2026-08-31
test(scripts): centralize compatibility fixtures

Eldad-Caura:
## Summary

- centralize the frozen OpenClaw plugin/skill identifier and reuse one derived plugin path throughout the gateway integration script
- centralize existing local database, benchmark DSN, local-dev skill, and recorded-baseline fixture values
- remove every movable Package Y legacy-name line from the scripts slice while leaving recorded baselines and pre-existing audited markers untouched

## Ratchet accounting

- current scripts slice at branch time: 21 ratcheted lines across 7 files
- movable slice: 17 lines across 5 files → 0 unannotated lines
- change split: 0 added, 0 annotated, 17 removed, 0 excused moves (-17 net)
- five compatibility markers remain, one per changed file, for frozen on-disk identifiers, existing local identities/defaults, or the subject behind recorded measurements
- the only four remaining slice lines are the two entries in each excluded JSONL baseline
- one line from the original 22-line package measurement had already left `main` before this branch was cut
- all 46 do-not-touch sentinel strings survive

The other intentional Package Y exclusions remain outside this PR: 10 Python legacy-alias contract lines, 9 tool-rename alias lines, 7 immutable-migration quotation lines, and 4 root conftest lines reserved while #1140 is open. Existing audited plugin markers are untouched.

## Validation

- Python byte-compilation passed for all four changed Python scripts
- `bash -n scripts/f3_wet_test_matrix.sh` passed
- safe CLI/usage paths passed for the gateway, benchmark, direction-invariance, and observer scripts
- explicit assertions confirmed byte-identical fixture values, recorded version-conflict inputs, and the derived plugin path
- all six exact CI Ruff check/format scope pairs passed
- all six exact CI mypy invocations passed (203, 85, 5, 11, 99, and 31 source files)
- legacy-name ratchet, do-not-touch sentinel, and `git diff --check` passed after the final rebase onto `origin/main`

The live wet tests were not run: they require the enterprise stack and real provider credentials, and this refactor does not change their external calls or expected values.

### DIFF
--- a/scripts/_f3_wet_test_observer.py
+++ b/scripts/_f3_wet_test_observer.py
@@ -22,6 +22,7 @@ import urllib.request
 CONTAINER_CORE_API = "caura-enterprise-core-api-1"
 CONTAINER_POSTGRES = "caura-enterprise-postgres-1"
 TENANT = "dev-41fddf"
+LOCAL_DB_IDENTITY = "memclaw"  # legacy-name-ok: existing local database role and name
 
 
 def write_memory(content: str, agent: str, mode: str = "strong") -> dict:
@@ -56,9 +57,9 @@ def psql(sql: str) -> str:
             CONTAINER_POSTGRES,
             "psql",
             "-U",
-            "memclaw",
+            LOCAL_DB_IDENTITY,
             "-d",
-            "memclaw",
+            LOCAL_DB_IDENTITY,
             "-tAc",
             sql,
         ],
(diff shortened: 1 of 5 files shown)


Eldad-Caura:
@claude @gemini

Eldad-Caura:
Addressed the Low finding: the benchmark example now omits the optional --pg-dsn argument and explains that the local default is used unless an empty database DSN is supplied. Revalidated compilation, help output, scripts mypy, ratchet, sentinel, and all exact CI Ruff/mypy scopes after rebasing.

Eldad-Caura:
@erni-a

erni-a (approved):

PR #1159, opened 2026-08-31
fix(storage): scope recall increments to tenant

### LINKED ISSUE #1088: storage: POST /memories/increment-recall increments recall counts on any tenant's memories
`route:POST /api/v1/storage/memories/increment-recall`

Service method: `method:memory_increment_recall`

### The defect

`memory_increment_recall` runs `UPDATE memory SET recall_count = recall_count + 1, last_recalled_at = now() WHERE id IN (...)` with no tenant predicate.

The handler's own docstring states the lack of scope as intentional — "a by-id UPDATE; no tenant scope — matches its prior in-process semantics from core-api's track_recalls hook". That reasoning held while the call was in-process; it stopped holding when the path became an unauthenticated HTTP route.

### The fix

Require `tenant_id` and add it to the `WHERE`. Also worth updating that docstring, since it currently reads as a justification for the gap. Impact is lower than the others here (it corrupts recall ranking rather than content) but it is a write on another tenant's rows and it is cheap to fix.

---

### Why this is filed

Eldad-Caura:
## Summary

- require a binding `tenant_id` at the recall-increment client, route, and service boundaries
- constrain the UPDATE with `Memory.tenant_id == tenant_id` and propagate the tenant from every production caller
- reject missing or non-object tenant-bearing request bodies with 422
- add mixed-tenant and tenant-forwarding regressions
- regenerate the tenancy allowlist, removing only the #1088 method and route entries

## Risk framing

Current production callers derive memory IDs from tenant-scoped reads, so those paths were already safe. The storage method and direct route still relied on that caller-side guarantee: a direct or future caller could increment recall metadata for another tenant by UUID. This moves the guarantee into the query. Recall metadata feeds ranking; this PR makes no claim of observed exploitation.

## Regression proof

With only `Memory.tenant_id == tenant_id` removed from the UPDATE, the mixed-tenant regression fails:

```text
FAILED tests/test_fix2_cleanup_storage.py::test_increment_recall_only_updates_callers_tenant
E   AssertionError: assert {"updated": 2} == {"updated": 1}
E     {"updated": 2} != {"updated": 1}
```

After restoring the predicate, the same test passes.

## Validation

- affected root files: 79 passed
- full root suite after rebase: 5707 passed, 4 skipped, 16 deselected, 1 xfailed
- core-storage-api: 267 passed
- core-operations: 50 passed, coverage floor met
- core-worker: 80 passed, coverage floor met
- core-api and core-storage-api coverage floors met
- all Ruff check and format scopes passed
- all six mypy scopes passed
- all four Python packages built
- tenant-scope gate passed

Allowlist on current main: 98 total, 4 `id-addressed-write`. This PR: 96 total, 2 `id-addressed-write`. All 35 `opaque-body-write` notes survive regeneration.

Closes #1088

### DIFF
--- a/core-api/src/core_api/clients/storage_client.py
+++ b/core-api/src/core_api/clients/storage_client.py
@@ -1270,13 +1270,17 @@ class CoreStorageClient:
     # datetimes, so callers stringify UUIDs + ISO-format datetimes first.
     # =====================================================================
 
-    async def increment_recall(self, memory_ids: list[str]) -> int:
+    async def increment_recall(self, memory_ids: list[str], *, tenant_id: str) -> int:
         """Bump ``recall_count`` + ``last_recalled_at`` for memories by id.
 
         ``memory_ids`` must already be stringified (the storage endpoint
-        re-parses each as a UUID). Returns the rows-updated count.
+        re-parses each as a UUID). ``tenant_id`` binds the update to the
+        caller's tenant. Returns the rows-updated count.
         """
-        result = await self._post("/memories/increment-recall", {"memory_ids": memory_ids})
+        result = await self._post(
+            "/memories/increment-recall",
+            {"tenant_id": tenant_id, "memory_ids": memory_ids},
+        )
         return result.get("updated", 0)  # type: ignore[union-attr]
 
     async def log_recall(self, event: dict, candidates: list[dict]) -> str:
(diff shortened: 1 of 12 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #1160, opened 2026-08-31
fix(storage): scope report finalization to the owning tenant

### LINKED ISSUE #1082: storage: PATCH /reports/{report_id} completes any tenant's report by UUID

Eldad-Caura:
## Summary

- Require a non-empty `tenant_id` on `PATCH /api/v1/storage/reports/{report_id}` and bind the finalization UPDATE to it, so a report belonging to another tenant is indistinguishable from one that does not exist.
- Thread the owning tenant through `CoreStorageClient.update_report` as a keyword-only argument and the three `crystallizer_service` call sites.
- Remove the fixed method and route from the tenant-scope allowlist.

## Related Issue

Closes #1082.

## Type of Change

- [ ] Documentation update
- [ ] Other:

## What was reachable

The route took a bare UUID and `report_update_completed` matched on the primary key alone. `core-storage-api` authenticates no request (GHSA-wgvw-28pq-jc36) and docker-compose publishes it on `0.0.0.0:8002`, so knowing a report id was enough.

This path writes more than most: `status`, `completed_at`, `duration_ms` and six JSONB blobs — `summary`, `hygiene`, `health`, `usage_data`, `issues`, `crystallization` — every one straight from the body. The outcome is **substitution, not disclosure**: a victim's report marked `completed` carrying an attacker's scores and issue list, which that tenant then reads off their own dashboard as their own crystallization result.

There is a second effect. `report_find_running` is the short-circuit that serialises crystallization, and it matches `status == "running"` — so flipping that column out from under a live run releases the lock and lets a concurrent run start. That is pinned by its own test, through the query that actually reads it rather than by asserting on the column.

## Why `_require` rather than an inline check

The allowlist's `blind-spot` category is for paths that *are* scoped but by an idiom the gate's AST pass cannot read — "an inline `if not isinstance(...): raise` rather than the shared `_require` guard" — and it says to prefer rewriting such a guard to use `_require`. Hand-rolling the check here would have scoped the route while leaving the gate unable to see it, which means an allowlist entry instead of a deletion. Using the shared helper is what makes the route count as bound: 286 bound, up from 284.

## How Has This Been Tested?

All four security assertions fail on the parent commit; the happy-path guard passes either way, which is the point of including it:

```text
FAILED test_another_tenants_report_is_not_completed
FAILED test_omitted_tenant_id_is_rejected
FAILED test_a_foreign_id_is_indistinguishable_from_a_missing_one
FAILED test_the_crystallization_lock_is_not_released_by_another_tenant
4 failed, 1 passed
```

with the cross-tenant case reading:

```text
>       assert resp.status_code == 404, resp.text
E       AssertionError: {"ok":true}
E       assert 200 == 404
```

With the fix:

| check | result |
|---|---|
| `core-storage-api/tests/` | 272 passed |
| root `tests/` | 5717 passed, 4 skipped, 1 xfailed |
| `ruff check` — core-api/src, core-storage-api/src, core-storage-api/tests | all pass |
| `ruff format --check` — same three scopes | all pass |
| `scripts/tenant_scope_gate.py --base origin/main` | exit 0, "Allowlist shrank by 2" |

`tests/test_unknown_field_rejection.py::test_memory_update_rejects_unknown_field` — the test filed in #1137 — failed **once in six** full-suite runs on this tree, and passed in isolation. That matches the rate measured when #1137 was filed (1 of 4 on a clean baseline). I did not manage to capture a traceback for this occurrence, so I am calling it as that flake on the rate and on the fact that this diff touches only the reports path, not on a mechanism confirmed in this run.

## Checklist

- [x] `mypy` passes
- [x] `pytest` passes locally
- [ ] I have updated relevant documentation (no user-facing surface changed; this is an internal storage endpoint)

## Additional Notes

- Allowlist: 98 entries before, 96 after. `id-addressed-write`: 4 before, 2 after — the write half of the backlog is nearly clear.
- The SET clause is assembled from named keyword arguments, so no caller-supplied key reaches it and neither `id` nor `tenant_id` is writable through this path. The defect class of #1081 / #1118 / #1121 does not apply here.
- `tenant_id` is keyword-only on the client and merged into the body **last**, so a stray `tenant_id` left in `data` cannot override the one the caller passed deliberately. Keyword-only also means the existing `update_report` assertions in `tests/test_crystallization_must_not_wedge.py`, which read `await_args_list[0].args[1]`, keep resolving to the data dict.
- All three `crystallizer_service` call sites already had the tenant in scope; none needed new plumbing.

### DIFF
--- a/core-api/src/core_api/clients/storage_client.py
+++ b/core-api/src/core_api/clients/storage_client.py
@@ -2798,8 +2798,15 @@ class CoreStorageClient:
         """
         return await self._get(f"/reports/{report_id}", read=read)
 
-    async def update_report(self, report_id: str, data: dict) -> dict | None:
-        return await self._patch(f"/reports/{report_id}", data)
+    async def update_report(self, report_id: str, data: dict, *, tenant_id: str) -> dict | None:
+        """Finalize a report. ``tenant_id`` scopes the UPDATE and is required.
+
+        Keyword-only and outside ``data`` so every caller is checked statically
+        rather than at the 422: a report id alone used to be enough to finalize
+        any tenant's report. It is merged last so a stray ``tenant_id`` left in
+        ``data`` cannot override the one the caller passed deliberately.
+        """
+        return await self._patch(f"/reports/{report_id}", {**data, "tenant_id": tenant_id})
 
     async def find_running_report(
         self,
--- a/core-api/src/core_api/services/crystallizer_service.py
+++ b/core-api/src/core_api/services/crystallizer_service.py
@@ -210,6 +210,7 @@ async def start_crystallization(
                     "completed_at": datetime.now(UTC).isoformat(),
                     "summary": {"error": "could not queue the crystallization run"},
                 },
+                tenant_id=tenant_id,
             )
         except BaseException:
             # Same best-effort rule as the run's own guard: this must not replace
@@ -467,6 +468,7 @@ async def _execute_crystallization(
                 "issues": issues,
                 "crystallization": crystallization,
             },
+            tenant_id=tenant_id,
         )
 
         logger.info(
@@ -492,6 +494,7 @@ async def _execute_crystallization(
                     "completed_at": datetime.now(UTC).isoformat(),
                     "duration_ms": int((time.monotonic() - t0) * 1000),
                 },
+                tenant_id=tenant_id,
             )
         except BaseException:
             # ``BaseException``, matching the outer handler, and for the same
--- a/core-storage-api/src/core_storage_api/routers/reports.py
+++ b/core-storage-api/src/core_storage_api/routers/reports.py
@@ -7,6 +7,7 @@ from uuid import UUID
 
 from fastapi import APIRouter, HTTPException, Request
 
+from core_storage_api.routers._validation import _require
 from core_storage_api.schemas import AGENT_DIGEST_FIELDS, REPORT_FIELDS, orm_to_dict
 from core_storage_api.services.postgres_service import PostgresService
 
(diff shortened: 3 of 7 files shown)


Eldad-Caura:
CI was red on `tests/test_tenant_scope_gate.py::test_the_id_addressed_backlog_is_split_by_blast_radius`. Fixed, and the reason is worth stating rather than just deleting the assertion.

### The write backlog is now empty, and that broke a census

This PR's two removals are the **last two** `id-addressed-write` entries — another lane cleared `memory_increment_recall` on main while this was in flight. The category now has zero entries and no longer appears in the gate's summary at all.

The guard ended with:

```python
writes = [e for e in listed if e["category"] == "id-addressed-write"]
assert writes, "an empty write backlog means the split silently collapsed"
```

That held while the backlog was being worked through and inverted the moment it was cleared: it fails on the success case. So it is removed, with the reasoning left in its place so it does not get restored.

Every other route to an empty write list is already covered, each by a check that does not expire:

| how it could empty without the work being done | what catches it |
|---|---|
| the category dropped from `CATEGORIES` | the assertion immediately above, unchanged |
| an entry carrying a bogus category | `test_every_allowlist_entry_claims_a_known_category` (also names the entry) |
| a mutating entry relabelled as a read | the ratchet's relabel guard, pinned by `test_ratchet_fails_when_a_mutating_path_is_relabelled_as_a_read` |
| an entry quietly appended | the allowlist ratchet |

That leaves deletion as the only way to reach zero, and deletion is what fixing one looks like.

I first wrote a replacement assertion that every entry's category is in `CATEGORIES` — then found `test_every_allowlist_entry_claims_a_known_category` already does exactly that, and names the offending entry besides. Dropped it; the comment points there instead.

### The flake, now with a traceback

I said earlier I could not attribute the `test_memory_update_rejects_unknown_field` failure to #1137 without evidence. I have it now — reproduced on the third of three runs:

```
E  AssertionError: {"detail":"Near-duplicate memory exists: 27f8e9fc-…","error":{"code":"DUPLICATE_MEMORY",
   "details":{"reason":"semantic_similarity","existing_id":"27f8e9fc-…","existing_status":"active"}}}
E  assert 409 == 201
```

That is #1137's mechanism exactly: the memory the test creates trips `SEMANTIC_DEDUP_THRESHOLD` against a neighbour left by another test, so the `POST` returns 409 and the assertion on 201 fails before the PATCH under test is ever reached. Unrelated to this diff, which touches only the reports path.

### Re-verified after rebasing onto `6736af0f`

| check | result |
|---|---|
| `core-storage-api/tests/` | 272 passed |
| root `tests/` | 5734 passed, 4 skipped, 1 xfailed |
| `tenant_scope_gate.py --base origin/main` | exit 0, "Allowlist shrank by 2" |
| allowlist | 96 → 94; `id-addressed-write` 2 → **0** |

Also confirmed the removal is not vacuous: giving an entry an invented category still fails the file's checks, and the flat-category assertion still fires on `id-addressed`.

Eldad-Caura:
@claude

arkash20 (approved):

PR #1166, opened 2026-08-31
fix(gate): report tenancy allowlist composition

Eldad-Caura:
## Summary

- report allowlist additions, fixed removals, gone removals, invalid still-unscoped removals, and category moves as separate transitions
- print per-category before/after counts without netting a recategorisation into progress
- make both no-base and pre-allowlist-base comparisons say explicitly that there is nothing to compare
- keep enforcement unchanged: growth, weakened verdicts, destructive-category relabels, unlisted paths, and stale rows are refused exactly as before

The comparison uses the base and working allowlist files for movement. The live gate enumeration is used only to explain whether a written removal is now `REQUIRED`, gone from the codebase, or still unscoped and therefore invalid.

## Regression proof

The mixed-transition test was added first and failed against the old reporter:

```text
FAILED tests/test_tenant_scope_gate.py::test_ratchet_reports_what_moved_without_netting_categories
E   AssertionError: assert '  added (1):\n      + method:added [id-addressed-read]' in ''
1 failed in 1.83s
```

With the implementation present, that test proves a category move is printed once, never under either removal bucket, while the same scenario still returns the existing "allowlist grew" error for its added entry.

## Preservation

`python scripts/tenant_scope_gate.py --write` is byte-identical on the final rebased tree. All 85 entries retained a category; all 40 non-empty notes survived, including all 35 `opaque-body-write` notes. The write path and the stale-entry rejection were not changed.

## Verification

- `tests/test_tenant_scope_gate.py`: 74 passed
- full root suite: 5,720 passed, 4 skipped, 16 benchmark-deselected, 1 expected xfail
- Ruff check and `ruff format --check`: green at every CI scope, run separately with config discovery
- mypy: green for core-api (203 files), core-storage-api (85), core-operations (5), core-worker (11), common (99), and scripts (31)
- live `tenant_scope_gate.py --base origin/main`: green; zero additions/removals/moves and explicit per-category `before -> after` counts
- signed-off single commit; `uv.lock` unchanged


Eldad-Caura:
@erni-a

erni-a (approved):

Issue #1167, opened 2026-08-31
storage: GET /reports/{report_id} returns any tenant's report by UUID (read half of #1082)


Eldad-Caura:
`route:GET /api/v1/storage/reports/{report_id}`

Service method: `method:report_get_by_id`

### The defect

The handler takes a bare UUID and `report_get_by_id` is `session.get(CrystallizationReport, report_id)` — a primary-key fetch with no tenant predicate. A caller who knows a report id gets the row: `tenant_id`, `fleet_id`, and the `summary` / `hygiene` / `health` / `usage_data` / `issues` / `crystallization` blobs, whichever tenant owns it.

This is the read sibling of #1082, which fixed the write half of the same pair.

### Why it is not the same fix as #1082

The obvious fix — require `tenant_id`, add it to the predicate, 404 for both "missing" and "not yours", as `GET /memories/{memory_id}` already does — **breaks two behaviours that are currently tested**, because core-api authorizes *after* the fetch rather than delegating the scope to storage:

```python
# core-api/src/core_api/routes/crystallizer.py
report = await sc.get_report(str(report_id))
if not report:
    raise HTTPException(status_code=404, detail="Report not found")
if not auth.is_admin and report.get("tenant_id") not in auth.readable_tenant_ids:
    raise HTTPException(status_code=404, detail="Report not found")
```

1. **Admin bypass.** `AuthContext(tenant_id=None, is_admin=True)` reads any tenant's report and has no tenant to send. Pinned by `tests/test_api_crystallizer.py::test_get_report_foreign_tenant_admin_bypass`.
2. **Cross-tenant read keys.** The check is against a *set*, not a single tenant. `readable_tenant_ids` always contains `tenant_id` (`core-api/src/core_api/auth.py:86-91`) but may be strictly wider, and a caller reading a report in one of its other readable tenants would start getting 404.

The 404-not-403 collapse there is itself deliberate (audit finding #22): it stops an attacker distinguishing "exists elsewhere" from "does not exist". Whatever replaces it has to keep that.

### The fork

Three options, none free. This needs a decision before code.

[...]

Issue #1170, opened 2026-08-31
storage: the fleet read paths disclose another tenant's deploy activity by node UUID


Eldad-Caura:
Six `id-addressed-read` allowlist entries covering the fleet read paths:

```
method:fleet_get_command_by_id
method:fleet_get_node_by_id
method:fleet_has_recent_in_flight_deploy
method:fleet_count_recent_deploys_for_target
route:GET /api/v1/storage/fleet/commands/in-flight-deploy
route:GET /api/v1/storage/fleet/commands/deploy-attempt-count
```

### The defect

Two routes are reachable cross-tenant reads. Both take a node UUID off the query string and name no tenant at all:

- **`GET /fleet/commands/in-flight-deploy`** answers whether a `deploy` command is currently in flight for a node. Another tenant's rollouts are observable as they happen.
- **`GET /fleet/commands/deploy-attempt-count`** answers how many deploys have been queued for a node at a **caller-named** `target_version`. As well as leaking the count, that makes it an oracle for another tenant's release schedule: name a version, and a non-zero count confirms they are moving to it.

Both service methods behind them key on `node_id` alone. `fleet_get_pending_commands`, immediately adjacent, carries a comment explaining that exact trap — *"Scoped on the node/tenant PAIR, not on `node_id` alone […] a node_id-only predicate looks tenant-safe and is not"* — so the shape was already understood here; these two just never got the same treatment.

### Two entries that are not what they look like

Worth stating precisely so nobody reviewing the fix expects a demonstration that cannot exist:

**`fleet_get_node_by_id` is not reachable across tenants today.** All three callers guard — `get_node` and `delete_node` derive `node_id` from the tenant-scoped `fleet_get_node_id`, and `create_command` compares `node.tenant_id` after the fetch. Scoping it removes an unscoped primitive, and converts `create_command`'s manual comparison into a predicate. That is a real improvement (a predicate cannot be forgotten by the next caller the way a follow-up comparison can) but it is not the closing of a live hole.

**`fleet_get_command_by_id` has no callers at all** — none anywhere in the repository, in any file type, with no dynamic dispatch reaching it. It is a bare `session.get(FleetCommand, command_id)`. It should be **deleted, not scoped**: an unscoped read with no caller is worse than a used one, because it is what the next person reaches for when they want "the existing helper". This is the same question that removed `entity_delete_entity_links` (#1091) and `entity_get_entity_links_for_memories` (#1161).

### The fix

[...]

PR #1171, opened 2026-08-31
chore(storage): delete the never-called fleet_get_command_by_id

Eldad-Caura:
`method:fleet_get_command_by_id`, one of the twelve `id-addressed-read` entries left in `core-storage-api/tenant_scope_allowlist.json`.

## The prior question first

Precedent #1091 set on this surface, and followed by #1161, #1163 and #1168: ask whether the method should exist before asking what predicate it needs. This one has **no caller anywhere**.

- A repo-wide search for `fleet_get_command_by_id` across all file types returns the allowlist entry and the definition itself. No route, no storage-client wrapper, no test, not even a mock spy.
- No route reaches it. The only parameterised command route is `PATCH /fleet/commands/{command_id}/status` — scoped in #1082, and it calls `fleet_update_command_status`. There is no `GET /fleet/commands/{command_id}`.
- Nothing dispatches to service methods by name (no `getattr(_svc, …)` in either package), so a name-only search is conclusive here.

So it is deleted rather than scoped.

## What this is, precisely

**Not reachable from the wire.** The other members of this backlog are live paths where knowing a UUID is enough to read another tenant's row today. This was an unexposed helper with the same shape — a hazard, not a disclosure. It is on the list because the gate reads signatures rather than reachability, which is the right default, but nobody should read this PR as closing an open hole.

Scoping it instead would have left a tenant-checked helper with no caller, the outcome #1091 rejected: an unused helper is the one someone later adopts without noticing where its tenant comes from. That is not theoretical on this table — `fleet_get_pending_commands` carries a comment about exactly this trap (a `node_id`-only predicate "looks tenant-safe and is not"), and `create_command` exists to stop a command being filed against another tenant's node.

## Out-of-repo consumers — checked, not assumed

No hit for `fleet_get_command_by_id` or `get_command_by_id` in caura-enterprise, caura-ops, caura-daemon, caura-onprem, caura-onprem-installer, gateway, caura-test-automation or caura-github. It is an internal service method rather than an HTTP surface, so only a repo that vendored `postgres_service.py` could reach it, and none does. `core-api/openapi.broker.json` is the one frozen contract in play and is untouched.

## No new test

Same as #1161, #1163 and #1168: no caller to regress, no behaviour to pin. A test asserting the method is absent would pin an absence rather than anything anyone depends on. The guard is the tenant-scope gate, whose stale-entry check fails if the entry returns without a scope.

## Numbers

| | before | after |
|---|---|---|
| allowlist total | 84 | **83** |
| `id-addressed-read` | 12 | **11** |
| service methods | 188 | **187** |
| routes | 189 | 189 |

Regenerated with `scripts/tenant_scope_gate.py --write` in the same commit, after rebasing onto #1168 and #1169. All 40 hand-written notes survived and every entry kept its category. The gate classifies the change as `removed — gone (1)`, not a recategorisation.

## Verification

- `core-storage-api/tests/` — 283 passed
- root-suite fleet + gate tests (`test_ph3_storage_documents_fleet.py`, `test_fleet_heartbeat_auto_upgrade.py`, `test_tenant_scope_gate.py`) — 189 passed
- `ruff check` and `ruff format --check` clean at CI's scopes; mypy clean on core-storage-api (85 files)
- `legacy_name_ratchet.py --base origin/main` — no new lines; `do_not_touch_sentinel.py` — all 46 protected strings survive
- Re-ran both suites after the rebase, against `origin/main` at `b8978969`

## Scope note

First of a set clearing the remaining `id-addressed-read` backlog. #1167 covers the report pair; the fleet auto-upgrade reads, `fleet_get_node_by_id` and the three `entities/{entity_id}` reads follow as separate PRs.

### DIFF
--- a/core-storage-api/src/core_storage_api/services/postgres_service.py
+++ b/core-storage-api/src/core_storage_api/services/postgres_service.py
@@ -9290,14 +9290,6 @@ class PostgresService:
 
     # -- Commands --
 
-    async def fleet_get_command_by_id(
-        self,
-        *,
-        command_id: UUID,
-    ) -> FleetCommand | None:
-        async with get_session() as session:
-            return await session.get(FleetCommand, command_id)
-
     async def fleet_get_pending_commands(
         self,
         *,
--- a/core-storage-api/tenant_scope_allowlist.json
+++ b/core-storage-api/tenant_scope_allowlist.json
@@ -99,12 +99,6 @@
       "category": "id-addressed-read"
     },
     {
-      "id": "method:fleet_get_command_by_id",
-      "verdict": "NONE",
-      "evidence": "no binding tenant parameter",
-      "category": "id-addressed-read"
-    },
-    {
       "id": "method:fleet_get_node_by_id",
       "verdict": "NONE",
       "evidence": "no binding tenant parameter",


Eldad-Caura:
@erni-a please ppaorver

erni-a (approved):

Eldad-Caura:
Superseded by #1173, which merged 40 minutes after this opened and deletes the same method as part of clearing all six fleet entries at once. Closing rather than rebasing: the conflict here is the allowlist, and there is nothing left underneath it — `fleet_get_command_by_id` is already gone from `main`.

#1173 is the better version of this. It reaches the same disposition on the same evidence (no callers, no dynamic dispatch, the `fleet_get_pending_commands` warning next door), and it found something this PR did not: `node_id="node-uuid-1"` in the existing auto-upgrade tests is not a valid UUID, so `UUID(node_id)` raises inside a `try/except Exception` that falls back to ALLOW — those tests never reached storage at all. That is the sharper form of the hazard, and worth having caught.

No duplicate issue was filed from this side; #1170 covers it and #1173 closed it.

PR #1173, opened 2026-08-31
fix(storage): scope the fleet read paths to the caller's tenant

### LINKED ISSUE #1170: storage: the fleet read paths disclose another tenant's deploy activity by node UUID

Eldad-Caura:
## Summary

- Require `tenant_id` on `GET /fleet/commands/in-flight-deploy` and `GET /fleet/commands/deploy-attempt-count`, and bind all three surviving fleet read methods to the node/tenant pair.
- **Delete** `fleet_get_command_by_id` rather than scope it — it had no callers.
- Thread the tenant through the storage client and the auto-upgrade gate in `core-api`.
- Remove all six entries from the tenant-scope allowlist.

## Related Issue

Closes #1170.

## Type of Change

- [ ] Documentation update
- [ ] Other:

## What was reachable, and what wasn't

Stated separately because only two of the six were live holes, and a PR that blurs that is harder to review.

**Reachable.** Both routes took a node UUID off the query string and named no tenant. `in-flight-deploy` answers whether a deploy is currently in flight for a node, so another tenant's rollouts were observable as they happened. `deploy-attempt-count` answers how many deploys were queued for a node at a **caller-named** `target_version` — as well as the count, that makes it an oracle for someone else's release schedule: name a version, and a non-zero count confirms it.

**Not reachable.** `fleet_get_node_by_id` could not be reached cross-tenant: `get_node` and `delete_node` derive `node_id` from the tenant-scoped `fleet_get_node_id`, and `create_command` compared `node.tenant_id` after the fetch. Scoping it removes an unscoped primitive and turns that manual comparison into a predicate — the form the next caller cannot forget — but no test here demonstrates a cross-tenant read, because there wasn't one.

## The deletion

`fleet_get_command_by_id` had **no callers anywhere** — every file type, no dynamic dispatch — and was a bare `session.get(FleetCommand, command_id)` sitting directly above `fleet_get_pending_commands`, whose own comment warns that keying on `node_id` alone *"looks tenant-safe and is not"*.

Asking "does this need to exist?" before scoping has now removed three unscoped reads for the cost of a search each — `entity_delete_entity_links` (#1091), `entity_get_entity_links_for_memories` (#1161), and this. That is a cheaper lesson than the scoping work, and worth making legible to whoever audits this next.

## The hazard this PR is really about

Both `core-api` call sites are wrapped in `try/except Exception` that **falls back to ALLOW**, so the auto-upgrade cannot be wedged by a transient DB error. That means a mis-wired argument does not fail — it silently disables the gate. A tenant scope that reports success while doing nothing.

So the wiring gets its own test, asserting what storage **received** rather than what the caller had available. On the parent:

```text
E  AssertionError: the in-flight and attempt-budget checks must both receive the tenant; storage saw []
E  assert [] == ['tenant-1', 'tenant-1']
```

Writing it turned up a second instance of the same swallowing: `node_id="node-uuid-1"` in the existing tests is not a valid UUID, so `UUID(node_id)` raises `ValueError`, the same handler eats it, and **those tests never reach storage at all**. The new test uses a real UUID; the others are left alone as out of scope.

## Design notes

- **`tenant_id: str`, deliberately not `str | Unscoped`.** The `Unscoped` sentinel exists so a genuinely cross-tenant call must spell `tenant_id=UNSCOPED` rather than default into one. No caller here needs that escape hatch, and where it is unnecessary `str` is stronger — it makes unscoped unspellable rather than merely explicit.
- **`create_command` now uses `_require`.** Scoping the fetch alone left a missing tenant surfacing as "Node not found": fail-closed, but it says the wrong thing, and it left the route unreadable to the gate's AST pass — which is exactly the `blind-spot` category. The gate caught this mid-change (`route:POST /fleet/commands takes no binding tenant scope and is not in the allowlist`), which is the check doing its job on its own author.
- While editing that block I found a **stale comment**: it claimed `fleet_get_pending_commands` keys on `node_id` alone, which stopped being true when that method was scoped to the pair. Rewritten.
- `read=False` (primary, not replica) survives on both client methods — read-after-write correctness for the deploy-dedup gate and the attempt budget.

## How Has This Been Tested?

New tests fail on the parent commit — 4 of the 8 storage tests (the other 4 are happy-path and `create_command` guards that already passed), plus the wiring test:

```text
FAILED TestInFlightDeployTenantScope::test_another_tenants_deploy_is_not_visible
FAILED TestInFlightDeployTenantScope::test_omitted_tenant_id_is_rejected
FAILED TestDeployAttemptCountTenantScope::test_another_tenants_attempts_are_not_counted
FAILED TestDeployAttemptCountTenantScope::test_omitted_tenant_id_is_rejected
FAILED test_auto_upgrade_gate_passes_the_tenant_to_storage
```

With the fix:

| check | result |
|---|---|
| `core-storage-api/tests/` | 291 passed |
| root `tests/` | 5760 passed, 4 skipped, 1 xfailed |
| `core-worker/tests/` | 80 passed |
| `ruff check` / `ruff format --check` — core-api/src, core-storage-api/src, core-storage-api/tests | all pass |

Gate output, using the per-move reporting from #1166:

```text
  added (0)
  removed — fixed (5)
  removed — gone (1):
      - method:fleet_get_command_by_id [id-addressed-read]
  removed — still unscoped (0)
      id-addressed-read: 12 -> 6
```

## Checklist

- [x] `mypy` passes
- [x] `pytest` passes locally
- [ ] I have updated relevant documentation (no user-facing surface changed; internal storage endpoints)
- [ ] I have updated `CHANGELOG.md` under the `Unreleased` section (not user-facing; release-please derives this from the conventional commit)

## Additional Notes

- Allowlist 84 → 78. `id-addressed-read` 12 → 6, leaving the entities group and the two `tenant-usage` entries.
- Four existing client tests in `tests/test_ph3_storage_documents_fleet.py` gained the new argument; one also gained a foreign-tenant assertion, since it already had a real node and real commands behind it and the end-to-end path was worth pinning once.

### DIFF
--- a/core-api/src/core_api/clients/storage_client.py
+++ b/core-api/src/core_api/clients/storage_client.py
@@ -2410,7 +2410,7 @@ class CoreStorageClient:
     async def delete_fleet(self, tenant_id: str, fleet_id: str) -> bool:
         return await self._delete(f"/fleet/{fleet_id}", tenant_id=tenant_id)
 
-    async def fleet_in_flight_deploy(self, *, node_id: UUID, since: datetime) -> bool:
+    async def fleet_in_flight_deploy(self, *, node_id: UUID, tenant_id: str, since: datetime) -> bool:
         result = await self._get(
             "/fleet/commands/in-flight-deploy",
             # primary, not replica: read-after-write deploy-dedup gate — it must see
@@ -2418,15 +2418,19 @@ class CoreStorageClient:
             # duplicate through (the acked-stuck-queue storm this gate prevents).
             read=False,
             node_id=str(node_id),
+            tenant_id=tenant_id,
             since=since.isoformat(),
         )
         return bool((result or {}).get("in_flight"))
 
-    async def fleet_deploy_attempt_count(self, *, node_id: UUID, target_version: str, since: datetime) -> int:
+    async def fleet_deploy_attempt_count(
+        self, *, node_id: UUID, tenant_id: str, target_version: str, since: datetime
+    ) -> int:
         result = await self._get(
             "/fleet/commands/deploy-attempt-count",
             read=False,  # primary: attempt budget must count just-queued deploys (see fleet_in_flight_deploy)
             node_id=str(node_id),
+            tenant_id=tenant_id,
             target_version=target_version,
             since=since.isoformat(),
         )
--- a/core-api/src/core_api/routes/fleet.py
+++ b/core-api/src/core_api/routes/fleet.py
@@ -555,6 +555,7 @@ async def _maybe_queue_auto_upgrade(
     try:
         in_flight = await sc.fleet_in_flight_deploy(
             node_id=UUID(node_id),
+            tenant_id=body.tenant_id,
             since=datetime.now(UTC) - DEPLOY_IN_FLIGHT_WINDOW,
         )
     except Exception:
@@ -585,6 +586,7 @@ async def _maybe_queue_auto_upgrade(
     try:
         recent_attempts = await sc.fleet_deploy_attempt_count(
             node_id=UUID(node_id),
+            tenant_id=body.tenant_id,
             target_version=target_version,
             since=datetime.now(UTC) - AUTO_UPGRADE_ATTEMPT_WINDOW,
         )
--- a/core-storage-api/src/core_storage_api/routers/fleet.py
+++ b/core-storage-api/src/core_storage_api/routers/fleet.py
@@ -97,7 +97,7 @@ async def get_node(
     tenant_id: str,
 ) -> dict:
     node_id = await _svc.fleet_get_node_id(tenant_id=tenant_id, node_name=node_name)
-    node = await _svc.fleet_get_node_by_id(node_id=node_id)
+    node = await _svc.fleet_get_node_by_id(node_id=node_id, tenant_id=tenant_id)
     if node is None:
         raise HTTPException(status_code=404, detail="Node not found")
     return orm_to_dict(node, FLEET_NODE_FIELDS)
@@ -109,7 +109,7 @@ async def delete_node(
     tenant_id: str,
 ) -> dict:
     node_id = await _svc.fleet_get_node_id(tenant_id=tenant_id, node_name=node_name)
-    node = await _svc.fleet_get_node_by_id(node_id=node_id)
+    node = await _svc.fleet_get_node_by_id(node_id=node_id, tenant_id=tenant_id)
     if node is None:
         raise HTTPException(status_code=404, detail="Node not found")
     fleet_id = getattr(node, "fleet_id", None)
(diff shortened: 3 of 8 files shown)


Eldad-Caura:
@erni-a please approve

Eldad-Caura:
@erni-a please approve

erni-a (approved):

erni-a (approved):

PR #1174, opened 2026-08-31
fix(storage): scope the entity-by-id reads to the caller's tenant

Eldad-Caura:
The last of the entity surface's `id-addressed-read` entries, and the read sibling of [#1119](https://github.com/caura-ai/caura/pull/1119). Three routes shared `entity_get_by_id`, which was `session.get(Entity, entity_id)` — a fetch by primary key with no predicate at all.

| route | what it returned unscoped | allowlist |
|---|---|---|
| `GET /entities/{entity_id}` | the whole entity row | unnoted |
| `GET /entities/{entity_id}/with-memories` | the row plus every memory linked to it | noted "self-authorizing" |
| `GET /entities/{entity_id}/relations` | the row's edges plus **each target entity in full** | noted "self-authorizing" |

**Stated no stronger than it is.** `core-storage-api` authenticates no request (GHSA-wgvw-28pq-jc36) and docker-compose publishes it on `0.0.0.0:8002`, so one UUID was enough to read any of the above. This is disclosure, not integrity — nothing here mutates a row. No issue was filed for these; the wording is mine.

### Retiring two hand-written notes, deliberately

`/with-memories` and `/relations` took an **optional** `tenant_id` and fell back to `entity.tenant_id` — the tenant of the row being addressed:

```python
t_id = tenant_id or entity.tenant_id
```

That is not a check. It holds by construction for whatever id the caller supplies, and an attacker closes it by *omitting the parameter*. Their allowlist note read *"Scopes to the entity's OWN tenant when omitted — self-authorizing."* I am contradicting a note a human wrote, so the evidence rather than the assertion — this is the pre-fix response to `GET /{id}/relations` **with no tenant at all**, victim tenant `test-tenant-d1a36907`:

```json
[{"relation":{...,"relation_type":"works_with"},
  "target":{"id":"e2369576-...","tenant_id":"test-tenant-d1a36907",
            "canonical_name":"Target-e0ee28d1","entity_type":"person",
            "attributes":{"secret":"victim-only"}}}]
```

The neighbour's whole record, one hop out, to a caller who supplied nothing but a UUID. Their *downstream* reads (`entity_get_linked_memories`, `relation_get_outgoing`) were already tenant-scoped — the hole was the entity fetch and the optionality above it, which is exactly what the note did not cover.

These two routes are not among my seven unnoted entries. I took them because `entity_get_by_id` is, and binding it necessarily binds every caller — the alternative is keeping an unscoped fetch alive for them, which defeats the fix. **If a reviewer wants them left as they were, that is the trade to argue with.**

### The guarantee that lived in one caller

`GET /{entity_id}` had no tenant parameter at all. Its only protection was in core-api, *after* the row crossed the wire — `entity_service.get_entity`:

```python
result = await sc.get_entity_with_linked_memories(str(entity_id))
entity = result.get("entity", {})
if entity.get("tenant_id") != tenant_id:
    return None
```

A Python comparison in one caller, on a row storage had already handed over. That comparison **stays** — it and the predicate fail independently, the same both-belts reasoning [#1119](https://github.com/caura-ai/caura/pull/1119) used — but it was never a substitute for a WHERE clause, and any other caller of that route got nothing.

### Filter, not reject — and why it is the safer half here

Stated deliberately, per the choice this sweep has to make consistently. A foreign id returns the same **404** as an id that does not exist, so these routes are not an existence oracle for entity UUIDs; a 403 would confirm the row is real and merely owned by someone else. `test_a_foreign_id_is_indistinguishable_from_a_missing_one` pins that the two responses are byte-identical.

**Every surviving path on this surface addresses exactly one id**, so filter and reject differ only in the status code and the partial-match question never arises. Where it does arise — bulk id lookups — [#1162](https://github.com/caura-ai/caura/pull/1162) already chose *filter*. This is consistent with it. **For the entries that follow this:** written before #1173 landed, the note was that `fleet_get_command_by_id`, `fleet_get_node_by_id` and `report_get_by_id` are all single-id reads and take the same answer unchanged — 404-on-mismatch, no 403 — while the two aggregating `fleet/commands/*` count routes should filter rather than error, matching #1162. #1173 has since settled the fleet half; the reports pair is what is left, and it is the single-id case.

### Does each need to exist? Yes, all three

Asked before choosing a predicate, per [#1091](https://github.com/caura-ai/caura/issues/1091), and established by searching for the method, the client wrapper and the route path separately rather than trusting the route:

- `GET /{entity_id}` → `sc.get_entity` → Path C entity hydration in `contradiction_detector._fetch_entity_context`.
- `/with-memories` → `sc.get_entity_with_linked_memories` → `entity_service.get_entity`.
- `/relations` → `sc.get_outgoing_relations` → `entity_service`'s relation-scope path.

None dead, none duplicated. The one dead member of this cluster was `relations/find`, deleted separately in [#1168](https://github.com/caura-ai/caura/pull/1168).

### The part worth reviewing: an oracle that would have gone vacuous

`core-storage-api/tests/test_entity_update_tenant_scope.py::_read` read the row back through `GET /entities/{entity_id}` **precisely because that route was unscoped**, and carried a note saying it would need a tenant the day the route took one. It now does — and giving it one would have been the wrong repair.

Those tests ask *"is the row still in the tenant that owned it?"* An oracle scoped to a tenant only answers about rows already known to be in it, so `test_another_tenants_entity_is_not_updated` could no longer observe a row that **moved namespace** — the exact failure it exists to catch. `_read` now reads `entities` through the ORM, tenant-blind by construction and immune to any further scoping. Same class of oracle #1162 had to rewrite, same fix.

One related trap caught in passing: `test_scope_columns_are_not_caller_writable` asserted `GET /entities/{stolen}` returns 404. Left alone that becomes a 422 on the missing parameter — still "not 200", so a laxer assertion would have passed while testing nothing. It now passes a tenant.

Six Path C stubs took `get_entity(eid)` positionally and broke on the new argument (15 root-suite failures, all one cause). **One of them now asserts the tenant reaches the hydration call** rather than ignoring it, so the contract is pinned and not merely tolerated.

### Failing without the fix

Three experiments, because the fix has two halves and they are not equally load-bearing.

**Reverting only the predicate** (`Entity.tenant_id == tenant_id` dropped from `entity_get_by_id`, routes still requiring the parameter) — 4 of 10 fail:

```
FAILED ...TestGetEntityByIdRead::test_a_foreign_entity_is_not_disclosed
FAILED ...TestGetEntityByIdRead::test_a_foreign_id_is_indistinguishable_from_a_missing_one
FAILED ...TestEntityWithMemoriesRead::test_a_foreign_entity_and_its_memories_are_not_disclosed
FAILED ...TestEntityRelationsRead::test_a_foreign_entitys_graph_is_not_disclosed
4 failed, 6 passed in 0.48s
```

The first failure prints the leaked row, which is the point — the test proves the foreign row is **absent from the result**, not merely that the call failed:

```
AssertionError: {"entity":{"id":"3d33fa12-...","tenant_id":"test-tenant-ce917fea",
"canonical_name":"Victim-07a001ad","attributes":{"secret":"victim-only"}},"linked_memories":[]}
assert 200 == 404
```

**Reverting only the required parameter** (routes back to `tenant_id: str | None = None` and the fallback, predicate kept) — 2 fail, the two omitted-tenant cases. Reported honestly: they fail on `404 != 422`, **not** on a disclosure, because the surviving predicate catches the leak. That half is real but it is the belt, not the braces.

**Reverting both** — the genuine pre-fix state — fails 7 of 10, everything except the three happy paths, and is where the `/relations` disclosure quoted above comes from.

### Allowlist

**78 → 74 entries; `id-addressed-read` 6 → 2.** Paths bound to a tenant 298 → 302. Regenerated with `python3 scripts/tenant_scope_gate.py --write` in this commit as the gate's stale-entry check requires. **38 hand-written notes preserved, no `category`/`verdict`/`evidence` drift on any surviving entry, nothing added** — the two notes that went are the retired "self-authorizing" pair.

**Rebased onto [#1173](https://github.com/caura-ai/caura/pull/1173)**, which landed the fleet cluster while this was in review — so the two remaining `id-addressed-read` entries are now just the reports pair, another lane's. **The entity/links surface is empty of this category**, closing what #1148/#1152/#1156 started on the write half and #1161/#1162/#1163/#1168 continued on the read half. With #1173 in, `id-addressed-read` is two entries from zero.

### Verification

- `core-storage-api/tests/`: **293 passed** (10 new).
- Root `tests/`: **5759 passed, 4 skipped, 1 xfailed**.
- `ruff check` at CI's 6 scopes and `ruff format --check` at its 5, run **separately**, discovery-based, no `--config`: clean. Both initially flagged this branch (an unsorted import block and two over-wrapped signatures) and were fixed, not waived. Vendored-`common` `--isolated`: 3 files already formatted.
- Tenant-scope gate vs `origin/main`: green. Legacy-name ratchet: no new lines. Do-not-touch sentinel: all 46 protected strings survive. Broker OpenAPI baseline: up to date (8 operations).

Run against a dedicated local pgvector instance provisioned CI's way — `alembic upgrade head` on the root database, a separate `memclaw_storage` database for the storage suite. The `#1137` flake did not fire in any of the four full root-suite runs on this branch. **Not checked:** anything requiring CI's own environment — the Actions matrix, the coverage-floors step, the enterprise re-vendor path.

---

*Rebased onto #1173 and force-pushed after the first green run; the allowlist was the only conflict (both PRs regenerate it) and was resolved by taking main's and re-running `--write`. Full verification re-run on the rebased tree: storage 301 passed, root 5760 passed / 4 skipped / 1 xfailed, ruff 6+5 scopes clean, mypy clean, gate green, ratchet/sentinel/broker clean.*

### DIFF
--- a/core-api/src/core_api/clients/storage_client.py
+++ b/core-api/src/core_api/clients/storage_client.py
@@ -1330,8 +1330,10 @@ class CoreStorageClient:
     async def create_entity(self, data: dict) -> dict:
         return await self._post("/entities", data)  # type: ignore[return-value]
 
-    async def get_entity(self, entity_id: str) -> dict | None:
-        return await self._get(f"/entities/{entity_id}")
+    async def get_entity(self, entity_id: str, tenant_id: str) -> dict | None:
+        # Same contract as ``get_memory``: the row is addressed by a bare UUID,
+        # so the tenant has to travel with it or storage has no predicate.
+        return await self._get(f"/entities/{entity_id}", tenant_id=tenant_id)
 
     async def update_entity(self, entity_id: str, tenant_id: str, data: dict) -> dict | None:
         # Same contract as ``update_memory`` above, for the same reason.
@@ -1450,14 +1452,13 @@ class CoreStorageClient:
             read=True,
         )
 
-    async def get_entity_with_linked_memories(self, entity_id: str) -> dict | None:
-        return await self._get(f"/entities/{entity_id}/with-memories")
+    async def get_entity_with_linked_memories(self, entity_id: str, tenant_id: str) -> dict | None:
+        return await self._get(f"/entities/{entity_id}/with-memories", tenant_id=tenant_id)
 
-    async def get_outgoing_relations(self, entity_id: str, tenant_id: str | None = None) -> list[dict]:
-        params: dict[str, Any] = {}
-        if tenant_id is not None:
-            params["tenant_id"] = tenant_id
-        return await self._get_list(f"/entities/{entity_id}/relations", **params)
+    async def get_outgoing_relations(self, entity_id: str, tenant_id: str) -> list[dict]:
+        # ``tenant_id`` was optional here and storage fell back to the addressed
+        # row's own tenant when it was omitted, so the optionality was the hole.
+        return await self._get_list(f"/entities/{entity_id}/relations", tenant_id=tenant_id)
 
     async def create_relation(self, data: dict) -> dict:
         return await self._post("/entities/relations", data)  # type: ignore[return-value]
--- a/core-api/src/core_api/services/contradiction_detector.py
+++ b/core-api/src/core_api/services/contradiction_detector.py
@@ -1677,9 +1677,9 @@ async def _fetch_entity_context(sc, memory_id: str, tenant_id: str) -> list[dict
     Two storage round-trips at worst:
       1. ``get_entity_links_for_memories([memory_id])`` — batch endpoint;
          returns ``{memory_id: [{entity_id, role}, ...]}``.
-      2. ``get_entity(entity_id)`` per link, fan-out via ``asyncio.gather``
-         — Path C is post-commit async so the round-trip parallelism is
-         latency-invisible to the write path.
+      2. ``get_entity(entity_id, tenant_id)`` per link, fan-out via
+         ``asyncio.gather`` — Path C is post-commit async so the round-trip
+         parallelism is latency-invisible to the write path.
 
     Returns ``[]`` (not ``None``) when the memory has no resolved links;
     the caller treats empty as the skip-retraction signal so this
@@ -1703,7 +1703,7 @@ async def _fetch_entity_context(sc, memory_id: str, tenant_id: str) -> list[dict
         if not entity_id:
             return None
         try:
-            entity = await sc.get_entity(str(entity_id))
+            entity = await sc.get_entity(str(entity_id), tenant_id)
         except Exception as e:
             logger.warning(
                 "Path C entity-context fetch failed (entity %s) for memory %s: %s",
--- a/core-api/src/core_api/services/entity_service.py
+++ b/core-api/src/core_api/services/entity_service.py
@@ -135,11 +135,15 @@ async def upsert_entity(
 
 async def get_entity(entity_id: UUID, tenant_id: str, caller_agent_id: str | None = None) -> EntityOut | None:
     sc = get_storage_client()
-    result = await sc.get_entity_with_linked_memories(str(entity_id))
+    result = await sc.get_entity_with_linked_memories(str(entity_id), tenant_id)
     if not result:
         return None
 
     entity = result.get("entity", {})
+    # Kept although storage now applies the same predicate. This comparison used
+    # to be the ONLY thing scoping this read — the defect class being that the
+    # guarantee lived in one caller — and the two fail independently: this one
+    # covers a storage regression, the predicate covers any other caller.
     if entity.get("tenant_id") != tenant_id:
         return None
 
--- a/core-storage-api/src/core_storage_api/routers/entities.py
+++ b/core-storage-api/src/core_storage_api/routers/entities.py
@@ -621,8 +621,12 @@ async def set_embeddings(request: Request) -> dict:
 
 
 @router.get("/{entity_id}")
-async def get_entity(entity_id: UUID) -> dict:
-    entity = await _svc.entity_get_by_id(entity_id)
+async def get_entity(entity_id: UUID, tenant_id: str) -> dict:
+    # Read guard, the mirror of ``PATCH /entities/{entity_id}`` above: the route
+    # took a bare UUID and returned the whole row, so knowing an id was enough to
+    # read another tenant's entity. ``tenant_id`` is a required query parameter —
+    # omitting it is a 422, not a fetch by primary key.
+    entity = await _svc.entity_get_by_id(entity_id, tenant_id)
     if entity is None:
         raise HTTPException(status_code=404, detail="Entity not found")
     return orm_to_dict(entity, ENTITY_FIELDS)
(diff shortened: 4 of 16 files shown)


Eldad-Caura:
Force-pushed: rebased onto #1173, which landed the fleet read cluster while this was in review. `claude-review` skipped silently on the `synchronize`, so re-triggering.

@claude @gemini please re-review.

**What changed since the green run** — rebase only, no change to the fix itself:

- The only conflict was `tenant_scope_allowlist.json`, which #1173 and this PR both regenerate. Resolved by taking main's copy and re-running `python3 scripts/tenant_scope_gate.py --write`, not by hand-merging.
- Totals restated: allowlist **78 → 74**, `id-addressed-read` **6 → 2**, paths bound **298 → 302**. The earlier body said 84 → 80 and 12 → 8 against the pre-#1173 base. Re-verified after the rebase: 38 hand-written notes preserved, no `category`/`verdict`/`evidence` drift, nothing added.
- The two remaining `id-addressed-read` entries are the reports pair. The fleet half is gone, so the paragraph about what this sets up for the fleet cluster is now history rather than a prediction; it is marked as such.

**Re-verified on the rebased tree**, not carried over from the previous run: `core-storage-api/tests/` 301 passed; root `tests/` 5760 passed, 4 skipped, 1 xfailed; `ruff check` at CI's 6 scopes and `ruff format --check` at its 5, run separately and discovery-based, clean; `mypy` core-api 203 files and core-storage-api 85 files clean; tenant-scope gate vs `origin/main` green; legacy-name ratchet, do-not-touch sentinel and the broker OpenAPI baseline all clean.

The part still most worth a reviewer's eye is unchanged by the rebase: **two hand-written allowlist notes are retired here** (`/with-memories` and `/relations`, both reading *"Scopes to the entity's OWN tenant when omitted — self-authorizing"*), and `test_entity_update_tenant_scope._read` was switched to a tenant-blind ORM read so it does not go vacuous now that the route it used as an oracle is scoped.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

Eldad-Caura:
**Correction to this PR's description — the threat framing was stale when I wrote it. The fix itself stands.**

The body's "Stated no stronger than it is" paragraph says *"`core-storage-api` authenticates no request (GHSA-wgvw-28pq-jc36) and docker-compose publishes it on `0.0.0.0:8002`, so one UUID was enough to read any of it."* Both halves are wrong on `main`, and were already wrong when this merged:

- **#1066 (`d6fd10b9`, 2026-08-30 04:10) added `RequireStorageSharedSecretMiddleware`**, registered unconditionally in `app.py` and fail-closed: every non-health request 401s unless `x-storage-secret` matches, and an *unset* secret rejects everything rather than opening the door. That landed the day before this PR.
- **`docker-compose.yml` has no `ports:` block for `core-storage-api`.** It is not published on the host at all.

I carried the wording over from the issue text and the allowlist category description without re-verifying it against `main`. The irony is that the paragraph was headed "stated no stronger than it is" and was the one place I should have checked hardest.

**What does not change — none of the substance rests on that premise:**

- The **fallback that this PR removed was genuinely not a check.** `tenant_id or entity.tenant_id` is satisfied by construction for whatever id the caller supplies; the evidence quoted in the body (the pre-fix `/relations` response handing back the neighbour entity in full, `"attributes":{"secret":"victim-only"}`) was produced against the test suite, not inferred from the advisory.
- The **guarantee living in one caller** — `entity_service.get_entity` comparing `entity["tenant_id"] != tenant_id` in Python *after* storage returned the row — is a real structural defect independent of who can reach the port.
- The shared secret is a **single credential shared by every internal caller and carries no tenant identity.** It authenticates *that* a caller is internal, never *which tenant* it speaks for, so it provides no tenant isolation whatsoever. Every id-addressed read on this surface was still unscoped with respect to the thing the secret cannot express.

**Accurate wording for anyone citing this later:** defence in depth against a compromised, confused, or buggy internal caller — not a hole reachable by an unauthenticated network attacker. The two retired "self-authorizing" allowlist notes were still wrong, and the predicates added here are still the right fix.

PR #1178, opened 2026-08-31
fix(storage): scope the report read to the caller's tenant

### LINKED ISSUE #1167: storage: GET /reports/{report_id} returns any tenant's report by UUID (read half of #1082)

Eldad-Caura:
## Summary

- Require `tenant_id` on `GET /api/v1/storage/reports/{report_id}` and bind `report_get_by_id` to it.
- Make `tenant_id` **optional** on core-api's `GET /crystallize/reports/{report_id}`, resolving to the credential's own tenant — so no existing tenant-scoped caller changes.
- Rescope the cross-tenant read oracle in `test_report_update_tenant_scope.py`, which deliberately used this route *because* it was unscoped.
- Remove the last two `id-addressed-read` entries from the allowlist.

Closes #1167.

## Type of Change

- [ ] New feature
- [x] Breaking change — **for admin keys only**, see below
- [ ] Documentation update

## The defect

The route took a bare UUID and the method was `session.get(CrystallizationReport, report_id)` — a primary-key fetch with no tenant predicate. `core-storage-api` authenticates no request (GHSA-wgvw-28pq-jc36) and docker-compose publishes it on `0.0.0.0:8002`, so knowing an id was enough to read the row: `tenant_id`, `fleet_id`, and the `summary` / `hygiene` / `health` / `usage_data` / `issues` / `crystallization` blobs. A tenant's hygiene posture, open issues and usage figures in one response.

Disclosure, not integrity — nothing here writes. #1082 fixed the write half of this pair.

**Does the method need to exist?** Yes — asked first, per #1091. Two live callers: this route, and `GET /reports/running`, which obtains its id from the tenant-scoped `report_find_running` and could pass a tenant for free. The route is what forced the decision.

## The fork #1167 left open

The issue documented three options and said the choice had to come before code, because core-api authorizes *after* the fetch: an admin key has no tenant of its own to send, and the check is against `readable_tenant_ids`, a **set** that may be wider than the home tenant.

**Option A, in the form that costs least.** Required at storage, optional at core-api:

| caller | before | after |
|---|---|---|
| tenant-scoped credential | works | **unchanged** — no new parameter |
| cross-tenant read key, secondary tenant | works | names the tenant; `auth.enforce_readable_tenant` validates it |
| admin key (no home tenant) | works | **must name a tenant** |

Option C is ruled out by preserving the cross-tenant capability rather than narrowing it — the widened read stays available, it is just named and checked. Option B was not taken because it grows the allowlist by one and needs a maintainer override; the list may shrink and may not grow, and there was a way to clear both entries without spending that.

The admin change is the only thing taken away, and it is what the sibling `GET /reports` in the same package already requires of admin keys: *"agent creds can only report on their own tenant; admin keys may target any tenant."*

## Blast radius of the API change

- **Not** in `core-api/openapi.broker.json` (8 operations, checked) — no baseline to regenerate, no oasdiff exposure.
- **Not** in the REST table of `docs/public-api-stability.md`; `/crystallize/reports` is listed, the by-id route is not. No doc anywhere names it.
- The new parameter is optional, so it is additive on the wire; its `description` reaches `/api/openapi.json` from the handler.
- `scripts/smoke_test.py` already sent `params={"tenant_id": TENANT}` to this route, which the route ignored. It now means something.

## 403 vs 404

`enforce_readable_tenant` can answer 403 where the old route answered 404. It compares two strings the caller already knows, **before any lookup**, so it says something about the credential and nothing about whether the report — or the named tenant — exists. The 404 mask over report existence (audit finding #22) is untouched, and now holds at storage too: foreign report and absent report return the same status *and the same body*, asserted.

The post-fetch check in core-api is kept even though the scoped fetch makes it unreachable in normal operation. The two fail independently: drop the `scope` argument and it still masks; drop it and storage still refuses.

## `Annotated`, deliberately

`Annotated[str | None, Query(...)] = None` rather than the `= Query(None)` spelling used elsewhere in that tree. This handler is one of the very few called **directly** by tests rather than over HTTP — including the audit-finding-#22 regression. Under `= Query(None)` such a call receives the `Query` object, which is truthy and would sail past the guard as if it were a tenant: a value that looks like a scope and is not, which is the shape of mistake this PR is about. This way the two pre-existing security tests needed no edit at all.

## The oracle that had to change — and did not go vacuous

#1167 flagged this and it is the part worth reviewing closely. `test_report_update_tenant_scope.py::_read` read rows back through **this route** precisely because it was unscoped, and said so: *"when that route gains a required tenant, this helper needs one too."*

It now passes the report's **owner**, never the attacker. That keeps it an oracle rather than a mirror of the check under test: a successful cross-tenant PATCH rewrites the victim's row in place — `tenant_id` is not among the columns it writes — so the owner's own read still sees the damage.

Demonstrated rather than argued. With #1082's UPDATE predicate reverted and **only the status assertion relaxed**, so nothing but the helper can fail:

```text
core-storage-api/tests/test_report_update_tenant_scope.py:98:
    assert row["status"] == "running", "the victim's report was finalized"
E   AssertionError: the victim's report was finalized
E   assert 'completed' == 'running'
```

## Verified by reverting one thing at a time

| reverted | fails |
|---|---|
| read predicate `CrystallizationReport.tenant_id == tenant_id` | `test_another_tenants_report_is_not_returned` (`assert 200 == 404`), `test_a_foreign_id_is_indistinguishable_from_a_missing_one` |
| the **requirement**, made optional (`tenant_id: str \| None = None`) | `test_an_omitted_tenant_is_rejected` (`assert 404 == 422`) |
| core-api's resolved scope → home tenant | `test_get_report_foreign_tenant_admin_bypass` (`assert '' == 'tenant-A'`), `test_get_report_by_id` (`assert 404 == 200`) |
| #1082's write predicate | 3 in `test_report_update_tenant_scope.py`, incl. the quote above |

Each revert fails only its own cases, so none is masking another. The requirement gets its own case because optional-with-a-fallback restores the hole while every predicate test keeps passing — that is exactly the shape the two entity routes had before #1174 (`tenant_id or entity.tenant_id`).

## Numbers

| | before | after |
|---|---|---|
| allowlist total | 74 | **72** |
| `id-addressed-read` | 2 | **0** — category gone from the gate summary |

Regenerated with `scripts/tenant_scope_gate.py --write` in the same commit. All 38 hand-written notes survived unchanged and every entry kept its category; the gate classifies both as `removed — fixed`.

**This clears the `id-addressed-read` backlog.** The remaining backlog is `grant-in-lieu-of-tenant` (2) and `blind-spot` (1).

## Verification

Against `origin/main` at `627db009`, re-run after the rebase:

- root suite — 5796 passed, 4 skipped, 16 deselected, 1 xfailed
- `core-storage-api/tests/` — 306 passed
- `ruff check` and `ruff format --check` clean at CI's scopes; mypy clean on core-api (203 files) and core-storage-api (85)
- broker OpenAPI baseline up to date; legacy-name ratchet reports no new lines; all 46 sentinel strings survive
- `scripts/check_readme_quickstart.py` fails locally for want of a running server — **confirmed identical on clean `origin/main`**, so not this change

## Out-of-repo consumers — checked

One exists, and it is fine. `caura-enterprise/frontend/app/(dashboard)/manage/_components/crystallizer-tab.tsx` polls `/api/crystallize/reports/${res.report_id}` (and again on expand) with **no** `tenant_id`. It does not break, and the reason is the gateway rather than luck:

- the browser calls a same-origin relative path, so the credential that reaches core-api is whatever the enterprise gateway attaches, not what the page holds;
- `gateway/nginx.conf.template`'s catch-all `location /api/` rewrites to `/api/v1/`, runs the `/_auth` subrequest — *"resolve session cookie / JWT / API key → X-Tenant-ID"* — and sets `proxy_set_header X-Tenant-ID $tenant_id`;
- so `AuthContext.tenant_id` is populated for those calls and the optional parameter resolves to it.

No hit for this route in caura-ops, caura-daemon, caura-onprem, caura-onprem-installer, gateway, caura-test-automation or caura-github. The in-repo caller (`scripts/smoke_test.py`) already sends `tenant_id`.

`caura-enterprise/frontend/site/openapi.snapshot.json` carries this path and will show the new optional parameter on its next refresh. That snapshot is the enterprise repo's own gate, not this PR's, and an added optional query parameter is not a breaking diff.

## Not verified

CI's own environment — the Actions matrix — and whether any deployment reaches this route with an admin credential that carries no tenant. Every path I could trace supplies one.

### DIFF
--- a/core-api/src/core_api/clients/storage_client.py
+++ b/core-api/src/core_api/clients/storage_client.py
@@ -2782,8 +2782,14 @@ class CoreStorageClient:
     async def create_report(self, data: dict) -> dict:
         return await self._post("/reports", data)  # type: ignore[return-value]
 
-    async def get_report(self, report_id: str, *, read: bool = True) -> dict | None:
-        """Fetch one analysis report.
+    async def get_report(self, report_id: str, tenant_id: str, *, read: bool = True) -> dict | None:
+        """Fetch one analysis report, within a tenant.
+
+        ``tenant_id`` is positional and required, matching ``update_report``
+        below: a report id alone used to be enough to READ any tenant's report,
+        as it once was to finalize one (#1082 / #1167). Positional rather than
+        keyword-only because there is no ``data`` dict here for a stray copy to
+        hide in.
 
         ``read`` defaults to True — the replica is right for the read paths that
         display a report. Pass ``read=False`` for a read-your-write, where replica
@@ -2793,7 +2799,7 @@ class CoreStorageClient:
         there re-runs a multi-minute LLM job. Same reason
         ``find_by_content_hash`` pins the writer.
         """
-        return await self._get(f"/reports/{report_id}", read=read)
+        return await self._get(f"/reports/{report_id}", read=read, tenant_id=tenant_id)
 
     async def update_report(self, report_id: str, data: dict, *, tenant_id: str) -> dict | None:
         """Finalize a report. ``tenant_id`` scopes the UPDATE and is required.
--- a/core-api/src/core_api/routes/crystallizer.py
+++ b/core-api/src/core_api/routes/crystallizer.py
@@ -1,13 +1,16 @@
 """Memory Crystallizer routes."""
 
+from typing import Annotated
 from uuid import UUID
 
 from fastapi import APIRouter, Depends, HTTPException, Query
 from pydantic import BaseModel
 
+from core_api import errors
 from core_api import openapi_responses as _oar
 from core_api.auth import AuthContext, get_auth_context
 from core_api.clients.storage_client import get_storage_client
+from core_api.errors import coded_detail
 from core_api.schemas import STRICT_WRITE_BODY
 from core_api.services.crystallizer_service import start_crystallization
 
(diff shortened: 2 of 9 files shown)


Eldad-Caura:
@erni-a ready for approval — CI green (14/14), Claude review clean with no issues.

Clears the last two `id-addressed-read` entries, so that backlog category goes to zero. The one behaviour change is that admin keys must now name a tenant on `GET /crystallize/reports/{report_id}`; tenant-scoped and cross-tenant read credentials are unaffected. The oracle change in `test_report_update_tenant_scope.py` is the part worth a look — the PR body shows it failing on a reverted write predicate, so it did not go vacuous.

erni-a (approved):

PR #1181, opened 2026-08-31
fix(storage): scope the report-by-id read to the caller's tenant

### LINKED ISSUE #1167: storage: GET /reports/{report_id} returns any tenant's report by UUID (read half of #1082)

Eldad-Caura:
## Summary

- Require `tenant_id` on `GET /reports/{report_id}` and put it in `report_get_by_id`'s predicate.
- core-api's `GET /crystallize/reports/{report_id}` takes a `tenant_id` and gates it with **`enforce_readable_tenant`**, not `enforce_tenant`; the post-fetch comparison goes.
- Removes the last two `id-addressed-read` allowlist entries — **the category is now empty.**

## Related Issue

Closes #1167.

## Type of Change

- [ ] Documentation update
- [ ] Other:

## What was reachable

`report_get_by_id` was `session.get(CrystallizationReport, report_id)` — a primary-key fetch with no tenant predicate — under a route that took a bare UUID. `core-storage-api` authenticates no request and docker-compose publishes it on `0.0.0.0:8002`, so a report id was the whole of the authorization.

The payload is what makes it worth the fix. A report carries `tenant_id` and `fleet_id` plus six JSONB blobs — `summary`, `hygiene`, `health`, `usage_data`, `issues`, `crystallization`. Together they are a standing description of another tenant's memory estate: health scores, hygiene problems, token spend, and an enumerated issue list. #1082 fixed the write half of this pair, which could *substitute* that content; this one handed it over.

## The decision this needed, and why it wasn't the obvious fix

#1167 was filed rather than fixed because the obvious repair breaks two tested behaviours. core-api authorized **after** the fetch:

```python
report = await sc.get_report(str(report_id))
if not auth.is_admin and report.get("tenant_id") not in auth.readable_tenant_ids:
```

That is a check against a **set**, and admin keys have no tenant at all (`AuthContext(tenant_id=None, is_admin=True)`).

**`enforce_readable_tenant`, not `enforce_tenant`.** The sibling precedent on `GET /reports` uses the latter, and copying it here would have reintroduced the problem: `enforce_tenant`'s own docstring opens *"Raise if the caller may not write to `requested_tenant`"* and it pins to the home tenant, so a cross-tenant read key naming one of its *other* readable tenants gets a 403 on a row it is entitled to. `enforce_readable_tenant` is the read-shaped equivalent, already covered by `tests/test_auth_context.py` for home / widened / unrelated / admin. Why the sibling can use the write gate: `GET /reports` is an aggregate whose breadth is a separate explicit `scope='own'|'org'` axis, so pinning the *named* tenant to home is coherent there. A single report fetched by id has no such axis.

`readable_tenant_ids` is derived from the credential (`auth.py:86-91`) and never from the request, which is what keeps this from being the `grant-in-lieu-of-tenant` shape — the caller names *one* tenant; the credential decides whether it may.

**The post-fetch comparison is deleted, not kept as defence in depth.** It protected this route's callers while leaving the storage route answering to a bare id — and the storage route is the half an attacker on port 8002 does not go through. Leaving it would be a second, weaker check on a path that no longer needs one.

## The 403 is not a new oracle

Requiring a tenant introduces a 403 on a path that previously had only 404s, and audit finding #22 exists because distinguishing 403 from 404 on random UUIDs is how you enumerate reports across tenants. So this is pinned rather than asserted:

| request | result | what it reveals |
|---|---|---|
| id + a tenant the credential may read, report absent | 404 | nothing |
| id + a tenant the credential may read, report in another tenant | 404 (predicate misses) | nothing |
| id + a tenant the credential may **not** read | 403 | only that this credential cannot read tenant T |

`enforce_readable_tenant` runs **before** the fetch and sees only the named tenant, so the 403 is independent of `report_id` — `test_get_report_403_is_keyed_on_the_tenant_not_the_report` asserts the same 403 for a plausible id and a nonexistent one, and that storage is not consulted at all in that branch. Report existence still collapses to 404 in every case.

## The oracle that had to change

`core-storage-api/tests/test_report_update_tenant_scope.py`'s `_read` helper deliberately used this unscoped route to read rows back "regardless of tenant", with a comment saying it would need a tenant the day the route took one.

**Giving it a tenant would have been the wrong repair** — the same conclusion #1174 reached on the entity side an hour earlier. Every assertion in that file asks "is the row still in the tenant that owned it, holding the values it had", so a tenant-scoped oracle can only answer about rows already known to be in that tenant and could not observe a row that **moved namespace** — which is exactly what `test_another_tenants_report_is_not_completed` exists to catch, and whose last assertion (`row["tenant_id"] == victim`) would have gone silently untested. It now reads the row through the ORM: tenant-blind by construction, and immune to any further scoping of the route.

## How Has This Been Tested?

New storage tests, verified against the parent commit — 3 of the 5 fail there, the other 2 being the happy path and a chained-call regression guard that were never meant to fail:

```text
FAILED test_report_read_tenant_scope.py::test_another_tenants_report_is_not_readable
FAILED test_report_read_tenant_scope.py::test_omitted_tenant_id_is_rejected
FAILED test_report_read_tenant_scope.py::test_a_foreign_id_is_indistinguishable_from_a_missing_one
3 failed, 7 passed
```

The four core-api tests also fail on the parent, but honestly: they fail as `TypeError` because `tenant_id` does not exist there. They pin the contract of the new shape rather than demonstrating a parent bug — the behavioural proof is the three storage failures above.

| check | result |
|---|---|
| `core-storage-api/tests/` | 306 passed |
| root `tests/` | 5795 passed, 4 skipped, 1 xfailed |
| `core-worker/tests/` | 80 passed |
| `core-operations/tests/` | 50 passed |
| `ruff check` — core-api/src, core-storage-api/src, core-storage-api/tests, common, core-operations, core-worker | all pass |
| `ruff format --check` — same scopes | all pass |
| broker OpenAPI baseline | up to date (8 operations) |

One unrelated flake appeared in a full-suite run: `test_memory_update_rejects_unknown_field` (#1137 — `DUPLICATE_MEMORY` / `semantic_similarity`, 409 vs 201). It passed 4/4 on rerun and in the identical full-suite run before the rebase; it touches memory writes, which this PR does not.

Gate output:

```text
  added (0)
  removed — fixed (2):
      - method:report_get_by_id [id-addressed-read]
      - route:GET /api/v1/storage/reports/{report_id} [id-addressed-read]
      id-addressed-read: 2 -> 0
```

## Checklist

- [x] `mypy` passes
- [x] `pytest` passes locally
- [ ] I have updated relevant documentation — `GET /crystallize/reports/{report_id}` is not listed in `docs/public-api-stability.md` or `docs/api-reference.md` (only the `/crystallize/reports` list route is), and it is not in the frozen broker contract, so no documented surface changed
- [ ] I have updated `CHANGELOG.md` under `Unreleased` (release-please derives this from the conventional commit)

## Additional Notes

- **`id-addressed-read` is now empty.** That is the category the allowlist's own doc calls *"THIS IS THE BACKLOG… the shape of GHSA-wgvw-28pq-jc36"*. It stood at 14 entries when the ratchet went in. Allowlist 74 → 72.
- What remains on the board is 3 entries in two different categories: the `blind-spot` `tenant-usage/increment` guard, and the two `grant-in-lieu-of-tenant` `tenant_usage_query` entries (#1095).
- `find_running_report` gained the tenant at no cost — its `report_id` already came from the tenant-scoped `report_find_running`, so it was never reachable cross-tenant. Passing it leaves no unscoped call of the method behind, and `test_the_running_report_lookup_still_resolves` pins that chain, which no id-addressed test would have caught.
- `sc.get_report(..., read=False)` in `execute_reserved_report` keeps `read=False`; the read-your-write across a delivery boundary (H-02 / #812) is load-bearing and unchanged.

### DIFF
--- a/core-api/src/core_api/clients/storage_client.py
+++ b/core-api/src/core_api/clients/storage_client.py
@@ -2782,8 +2782,8 @@ class CoreStorageClient:
     async def create_report(self, data: dict) -> dict:
         return await self._post("/reports", data)  # type: ignore[return-value]
 
-    async def get_report(self, report_id: str, *, read: bool = True) -> dict | None:
-        """Fetch one analysis report.
+    async def get_report(self, report_id: str, tenant_id: str, *, read: bool = True) -> dict | None:
+        """Fetch one analysis report within ``tenant_id``.
 
         ``read`` defaults to True — the replica is right for the read paths that
         display a report. Pass ``read=False`` for a read-your-write, where replica
@@ -2793,7 +2793,7 @@ class CoreStorageClient:
         there re-runs a multi-minute LLM job. Same reason
         ``find_by_content_hash`` pins the writer.
         """
-        return await self._get(f"/reports/{report_id}", read=read)
+        return await self._get(f"/reports/{report_id}", read=read, tenant_id=tenant_id)
 
     async def update_report(self, report_id: str, data: dict, *, tenant_id: str) -> dict | None:
         """Finalize a report. ``tenant_id`` scopes the UPDATE and is required.
--- a/core-api/src/core_api/routes/crystallizer.py
+++ b/core-api/src/core_api/routes/crystallizer.py
@@ -132,22 +132,41 @@ async def list_reports(
 )
 async def get_report(
     report_id: UUID,
+    tenant_id: str,
     auth: AuthContext = Depends(get_auth_context),
 ):
-    """Get a full crystallization report by ID."""
+    """Get a full crystallization report by ID, from ``tenant_id``.
+
+    The caller names the tenant; the credential decides whether it may read it.
+    ``enforce_readable_tenant`` — not ``enforce_tenant``, which is the WRITE
+    gate and pins to the home tenant — so a cross-tenant read key still reaches
+    any tenant in its ``readable_tenant_ids``, and an admin key any tenant at
+    all. That set is derived from the credential, never from the request, which
+    is what keeps this from being the "caller names its own scope" shape.
+
+    **Report existence still collapses to 404** (audit finding #22). The 403
+    raised below is keyed on the NAMED TENANT and is independent of
+    ``report_id``: it says only "this credential may not read tenant T", which
+    the caller already knows about its own key. Every question about a report —
+    absent, or present in a tenant this request did not name — answers 404, so
+    a random UUID reveals nothing. There is a regression test for exactly this
+    matrix in ``tests/test_api_crystallizer.py``.
+
+    ``tenant_id`` is required for admin callers too. ``AuthContext`` for an
+    admin key has ``tenant_id=None``, so there is no tenant to infer, and
+    storage now demands one; naming it is the API change this endpoint takes in
+    exchange for storage no longer answering to a bare primary key.
+    """
+    auth.enforce_readable_tenant(tenant_id)
     sc = get_storage_client()
-    report = await sc.get_report(str(report_id))
+    # No post-fetch tenant comparison: storage scopes the fetch to the tenant
+    # authorized above, so a report outside it is already a 404 on the line
+    # below. The comparison that stood here ran AFTER storage had handed the
+    # row over — it protected this route's callers while leaving the storage
+    # route itself answering to a bare id, which is the half an attacker who
+    # can reach port 8002 does not go through.
+    report = await sc.get_report(str(report_id), tenant_id)
     if not report:
-        raise HTTPException(status_code=404, detail="Report not found")
-    # Collapse foreign-tenant (403) into not-found (404) so callers
-    # can't probe for the existence of reports in other tenants by
-    # distinguishing 403 from 404 on random UUIDs (audit finding #22).
-    # Cross-tenant read keys (``readable_tenant_ids`` widened past the
-    # home tenant) are honoured here — the set always contains
-    # ``auth.tenant_id`` by construction (``AuthContext`` line 85-90),
-    # so the 404 mask still hides reports the caller has no read
-    # access to.
-    if not auth.is_admin and report.get("tenant_id") not in auth.readable_tenant_ids:
         raise HTTPException(status_code=404, detail="Report not found")
     return {
         "id": str(report.get("id", "")),
--- a/core-api/src/core_api/services/crystallizer_service.py
+++ b/core-api/src/core_api/services/crystallizer_service.py
@@ -254,7 +254,7 @@ async def execute_reserved_report(
     # already finished — and then the whole multi-minute run happens again, which
     # is the exact duplication this check exists to prevent. Same class as H-02
     # (#812), where replica lag on a back-channel read-back dropped the trigger.
-    existing = await sc.get_report(report_id, read=False)
+    existing = await sc.get_report(report_id, tenant_id, read=False)
     if existing is None:
         # Nothing to execute against. Reserving one here would hand back an id the
         # caller never asked for, so this drops — and the publisher's row going
--- a/core-storage-api/src/core_storage_api/routers/reports.py
+++ b/core-storage-api/src/core_storage_api/routers/reports.py
@@ -31,7 +31,10 @@ async def find_running_report(
     report_id = await _svc.report_find_running(tenant_id, fleet_id)
     if report_id is None:
         raise HTTPException(status_code=404, detail="No running report found")
-    report = await _svc.report_get_by_id(report_id)
+    # ``report_id`` already came from the tenant-scoped ``report_find_running``,
+    # so this fetch could not cross tenants; passing the tenant costs nothing
+    # and leaves no unscoped call of this method behind.
+    report = await _svc.report_get_by_id(report_id, tenant_id)
     if report is None:
         raise HTTPException(status_code=404, detail="No running report found")
     return orm_to_dict(report, REPORT_FIELDS)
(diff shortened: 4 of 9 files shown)


Eldad-Caura:
Superseded by #1178, which merged at 23:52 — six minutes before this opened — and closed #1167. That is why this is CONFLICTING: the same change is already on `main`, so there is nothing left to rebase onto.

The two landed on the same shape independently: `tenant_id` required at storage and in `report_get_by_id`'s predicate, optional at core-api resolving to the credential's own tenant, gated by `enforce_readable_tenant` rather than `enforce_tenant` so cross-tenant read keys keep their widened read. Same disposition on the same evidence.

**One real difference, and it is worth a decision rather than a silent loss.** This PR removes core-api's post-fetch `report.get("tenant_id") not in auth.readable_tenant_ids` check; #1178 kept it. The reasoning on `main`:

> Kept, though the scoped fetch above now makes it unreachable in normal operation. Collapsing foreign-tenant (403) into not-found (404) is what stops a caller probing for reports in other tenants by distinguishing the two on random UUIDs (audit finding #22), and the two checks fail independently: drop the `scope` argument and this still masks; drop this and storage still refuses.

If the view here is that a redundant check is worse than a redundant line — a defensible position, and the argument for it is that an unreachable branch rots — that is a small follow-up PR against `main` with that rationale, not a reason to revive this one.

Also on `main` and worth knowing if you touch this handler again: it uses `Annotated[str | None, Query(...)] = None` rather than `= Query(None)`, because three tests call it directly rather than over HTTP and the `Query` object is truthy — it would pass a `if not scope` guard as though it were a tenant.

Closing rather than rebasing, on the precedent #1171 set in the other direction earlier tonight.

Eldad-Caura:
Closing as a duplicate — #1178 landed the same fix while this was in review, and it is the better of the two.

Where they differ, theirs wins on the substance:

- **`tenant_id` is optional there, defaulting to `auth.tenant_id`.** I made it required, which is a breaking change for every tenant-scoped caller for no security gain; the default only widens to a named tenant for cross-tenant read keys, and admin keys — which have no tenant to infer — get a coded 400. Strictly better than what this PR did.
- The `Annotated[str | None, Query(...)] = None` spelling, with the note that `= Query(None)` would hand a truthy `Query` object to the tests that call the handler directly. That is a real trap and I did not see it.

One deliberate difference the other way, worth recording rather than arguing: #1178 **keeps** the post-fetch `readable_tenant_ids` comparison, where #1167 said to delete it as "a second, weaker check on a path that no longer needs one" and this PR did. Their reasoning is that the two fail independently. Having now read it against the shipped code, the retained check is inert in every reachable case, so it costs nothing — not worth a follow-up either way.

**One thing this PR had that #1178 does not, and it is worth keeping.** The audit-finding-#22 property is currently asserted by a comment — *"Compares two strings the caller already knows, before any lookup, so the 403 it can raise says something about the credential and nothing about whether the report exists"* — and that is an **ordering** property with nothing pinning it. I verified the gap rather than assuming it: moving `enforce_readable_tenant` to after the fetch on `origin/main` leaves all 13 `test_api_crystallizer.py` tests and all 19 `test_auth_context.py` tests passing, while the endpoint becomes an existence oracle again:

```text
  report EXISTS in unreadable tenant-A -> 403
  report ABSENT in unreadable tenant-A -> 404
E  AssertionError: existence oracle: 403 when the report exists, 404 when it does not
```

Against the code as merged, both cases answer 403 and it passes. So #1178 is correct — the invariant just is not defended, and the refactor that breaks it is an ordinary-looking one.

Happy to send that single regression test as a small follow-up if it is wanted; flagging rather than opening one, since this area has now had two lanes in it tonight.

PR #1187, opened 2026-09-01
test(core-storage-api): give the storage suite its own database by default

Eldad-Caura:
## The problem

The two Python suites provision incompatibly and cannot share a database:

| suite | provisioning |
|---|---|
| `tests/` | `Base.metadata.create_all` |
| `core-storage-api/tests/` | the real Alembic chain, via `init_database()` |

Sharing one database **does not fail loudly.** `create_all` going first leaves tables carrying no `alembic_version`, so `init_database()` stamps head and skips the entire chain. Every migration-only table — one with no ORM model, e.g. `tenant_suppression` — is then absent from a database whose stamp claims it is current.

Measured directly on a scratch database, running the root suite first and then the storage suite:

```
collided:          stamp=041  tables=27  tenant_suppression=MISSING
properly migrated: stamp=041  tables=28  tenant_suppression=present
```

Nothing raises. The suite just runs against a schema that is quietly short.

## Why this is a fix and not a repair

CI already solves this, at `.github/workflows/ci.yml` → "Create the storage-suite database", by provisioning `memclaw_storage` and passing it explicitly. That comment describes the mirror-image failure too.

But `memclaw_storage` appeared in three lines of `ci.yml` and **nowhere else in the repository** — no CONTRIBUTING entry, no conftest docstring. A contributor or agent running `core-storage-api/tests/` locally got the shared database and no indication a second one was expected. An isolation that exists only in the workflow file is a step-order convention, not isolation.

## The change

Default the storage suite's own `DATABASE_URL` to `memclaw_storage`, the name CI already provisions. The conftest sets it through `os.environ.setdefault`, so **CI's explicit value still wins and that job is unchanged.**

`LOCAL_DATABASE_URL` in `core_storage_api/config.py` is deliberately **not** touched. Two things depend on it that are easy to miss, both verified rather than assumed:

- `Settings` has no `POSTGRES_*` assembly path — only an explicit `DATABASE_URL`, a complete `ALLOYDB_*` set, or that fallback. The CI "Run migrations" step runs `alembic upgrade head` with only `POSTGRES_*` set, so it resolves through `LOCAL_DATABASE_URL`. Repointing it would send that step at a database not created until ~75 lines later in the job.
- `AGENT-INSTALL.md` writes a `.env` with only `POSTGRES_*`, creates database `memclaw` at step 5, then runs `alembic upgrade head` at step 6 — which would target a database step 5 never created.

Moving that constant would break both **and** would not have isolated this suite anyway, since the conftest's `setdefault` shadows it.

## Verification

Both suites, both orders, against separate databases:

| order | root suite | storage suite |
|---|---|---|
| root → storage | 5797 passed | 309 passed |
| storage → root | 5797 passed | 309 passed |

Both databases end at `stamp=041, tables=28`, with `tenant_suppression` present.

Also run at CI's exact scopes: `ruff check` / `ruff format --check` on `core-storage-api/tests/` (clean), `mypy src/` in `core-storage-api` (85 files, clean), and `scripts/legacy_name_ratchet.py --base origin/main` (passes; the two pasteable `createdb` lines carry `legacy-name-floor` markers, as they name the database CI already provisions).

The new regression test was confirmed to **fail** against the unfixed conftest and pass with it.

## Also updated

Two tenant-prefix docstrings stated that locally the storage suite shared the root suite's database, so that suite's sweep reclaimed these rows — the #858 leak guard. That is no longer true in either environment, and they now say so, along with what would make the prefix load-bearing again.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -64,6 +64,26 @@ This starts PostgreSQL + pgvector, Redis, and the core API with hot reload.
 pytest tests/ -v
 ```
 
+`core-storage-api` has a second suite that needs a **separate database**:
+
+```bash
+createdb memclaw_storage && psql -d memclaw_storage -c 'CREATE EXTENSION IF NOT EXISTS vector'  # legacy-name-floor: the database ci.yml already provisions; a pasteable command naming anything else is wrong
+pytest core-storage-api/tests/ -v
+```
+
+Two databases, because the suites provision incompatibly and cannot share one:
+`tests/` builds its schema with `Base.metadata.create_all`, while
+`core-storage-api/tests/` runs the real Alembic chain. Run them against one
+database and nothing fails loudly — `create_all` leaves tables with no
+`alembic_version`, so the Alembic side stamps head and skips every migration,
+and any migration-only table (one with no ORM model, e.g. `tenant_suppression`)
+is missing from a database whose stamp claims it is current. Run them the other
+way round and the migrated schema is the one that gets polluted.
+
+Each suite defaults to its own database, so no environment variable is needed;
+`DATABASE_URL` (storage suite) and `TEST_DATABASE_URL` (root suite) override
+them. CI provisions and passes both explicitly.
+
 See `README.md` for more deployment options and environment variable details.
 
 ## Workflow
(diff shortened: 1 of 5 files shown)


Eldad-Caura:
@erni-a ready for your approval — the Claude review came back clean with no issues, and all checks are green (CI, DCO, legacy-name ratchet, CodeQL, security scan).

Verified locally: both suites pass in either order (root 5797 / storage 309, both directions), with each database ending at `stamp=041, tables=28` and `tenant_suppression` present.

erni-a (approved):

PR #1189, opened 2026-09-01
feat(api): report whether a search bumped recall_count

Eldad-Caura:
Item 1 of the REST/MCP parity smoke report's "What to fix next" queue.

## The gap

`TrackRecalls` skips the `recall_count` bump when a recall carries no caller
agent identity (gap A26). The skip is correct and deliberate — the counter
feeds `recall_boost` back into ranking, so probe traffic must not inflate it —
and `tests/test_track_recalls_agentless.py` pins it as policy.

What it was not is *observable*. A tenant-scoped key has `auth.agent_id = None`,
and its only way to present an identity is `filter_agent_id`, which
unavoidably also narrows results to that agent's own authored rows. So a
tenant-key integration sees `recall_count` pinned at 0 forever and
`recall_boost` never engage — and nothing in the response says so. The only
way to discover it is to watch a counter never move.

That the behaviour is now tested-and-intentional makes surfacing it *more*
urgent, not less: it is permanent, and still silent.

## The change

`recall_tracked` on the search response — true when a bump was dispatched for
the returned rows, false when it was skipped. All three reasons for false are
invisible from the request alone: no caller identity, `diagnostic=true`, or an
empty result set.

Emitted on **both** REST `/search` and MCP `caura_recall`, so adding it does
not itself open a new surface divergence — which is the failure mode this
report exists to catch.

### Why the value is reported, not re-derived

The obvious cheap version is a predicate at the route over `filter_agent_id`
and `diagnostic`. That would be wrong in two ways the caller cannot see:

- an **empty result set** bumps nothing, and the route would claim it did;
- the deprecated `_search_memories_legacy` path bumps **unconditionally**, with
  none of TrackRecalls' gates.

So the value comes from TrackRecalls itself, threaded out through an optional
`recall_ctx` dict that mirrors the existing `diagnostic_ctx` out-param. The
legacy branch reports what *it* does rather than the pipeline's policy —
reporting the pipeline's answer there would be a lie to whoever flips
`_USE_PIPELINE_SEARCH` back during a hotfix, which is the divergence trap
`memory_service.py` already warns about in its own `BlankQuery` handler.

Absent key defaults to `false`, which is the honest failure direction: a caller
wrongly told "not tracked" investigates and finds a working counter, whereas a
caller wrongly told "tracked" goes on believing a permanently pinned counter is
fine — the exact silent-zero this field exists to end.

## Scope

This is the **interim, cheap** half. The root cause — `/search` has one knob
(`filter_agent_id`) doing three jobs: author filter, visibility identity, and
trust/fleet gate target — is unchanged. Splitting identity from filtering is
the real fix; it touches auth resolution, TrackRecalls, and the `/memories/stats`
route's one-knob rejection, and is being written up as a design proposal rather
than folded in here.

## Tests

`tests/test_search_recall_tracked_flag.py` — 7 cases.

Confirmed failing without the fix: reverting the `core-api/src` changes and
re-running fails all four REST cases. They pin **behaviour, not field
presence** — a hardcoded `true` fails three of them, a hardcoded `false` fails
the fourth.

The REST cases pin the search path explicitly via a fixture. That is not
defensive boilerplate: `tests/pipeline/test_search_pipeline.py:606` sets
`memory_service._USE_PIPELINE_SEARCH = False` and never restores it (no
`try/finally`, no `monkeypatch`), so every test running after it in the same
session silently exercises the deprecated legacy path. Verified directly with a
probe assertion. That leak is pre-existing and out of scope here — worth its
own fix, since it means part of the suite is not testing the production search
path at all.

Full suite green locally: 5804 passed, 4 skipped, 1 xfailed. `ruff check`,
`ruff format --check`, and `mypy` clean at CI's scopes.

### DIFF
--- a/core-api/openapi.broker.json
+++ b/core-api/openapi.broker.json
@@ -1698,6 +1698,12 @@
             },
             "title": "Items",
             "type": "array"
+          },
+          "recall_tracked": {
+            "default": false,
+            "description": "Whether this search bumped recall_count for the memories it returned. False means they were not reinforced: the caller presented no agent identity (a tenant-scoped key that did not set filter_agent_id — recall_count stays 0 for that caller and recall_boost never engages), the call set diagnostic=true, or nothing matched.",
+            "title": "Recall Tracked",
+            "type": "boolean"
           }
         },
         "required": [
--- a/core-api/src/core_api/mcp_server.py
+++ b/core-api/src/core_api/mcp_server.py
@@ -829,6 +829,11 @@ async def caura_recall(
         # the var empty; ``search_memories`` falls back to the home tenant only.
         # C31/D12 — diagnostic trace context, filled by the pipeline when requested.
         diagnostic_ctx: dict = {}
+        # Same field REST /search reports. Emitted on both surfaces so adding
+        # it does not itself become a parity gap: MCP always carries an agent
+        # identity, but ``diagnostic=true`` and an empty result set still make
+        # this False here.
+        recall_ctx: dict = {}
         results = await search_memories(
             tenant_id=tenant_id,
             query=query,
@@ -849,6 +854,7 @@ async def caura_recall(
             diagnostic=diagnostic,
             diagnostic_ctx=diagnostic_ctx if diagnostic else None,
             min_similarity=min_similarity,
+            recall_ctx=recall_ctx,
         )
         # Cross-tenant read audit (F2): emit one event per source tenant when
         # the credential widened beyond home. Async queue — non-blocking.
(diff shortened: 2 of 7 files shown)


Eldad-Caura:
@claude Force-pushed since your review: rebased onto `main` (resolving an additive conflict in the `caura_recall` payload against #1190 — both sets of fields kept), regenerated `core-api/openapi.broker.json` for the CI gate I missed, and moved the `recall_tracked` explanation into a `Field(description=...)` so it reaches the OpenAPI contract rather than only code readers.

Verified locally after the rebase: full suite 5811 passed / 4 skipped / 1 xfailed, `ruff check`, `ruff format --check`, `mypy` clean, `gen_broker_openapi.py --check` up to date, and the oasdiff breaking-change gate reports no breaking changes. Please re-review.

arkash20 (approved):

Eldad-Caura:

erni-a (approved):

PR #1192, opened 2026-09-01
feat(mcp): return objects from caura_manage transition and delete

Eldad-Caura:
Item 3 of the REST/MCP parity smoke report's "What to fix next" queue.

## The gap

`caura_manage` returned a serialized object for `op=read` and `op=update`, a
JSON error envelope for every refusal, and a **human sentence** for
`op=transition` and `op=delete`:

- `mcp_server.py:1412` — `f"Memory {memory_id} status updated: {old_status} -> {status}"`
- `mcp_server.py:1469` — `f"Memory {memory_id} deleted."`

One tool, three return shapes. A caller had to sniff the type to tell success
from failure — and then sniff the *text*, because a string is the success case
only if it does not happen to contain "Error". The parity smoke had to write
exactly that:

```python
ok = (isinstance(mcp_out, dict) and mcp_out.get("status") == "outdated") or \
     (isinstance(mcp_out, str) and "outdated" in mcp_out and "Error" not in mcp_out)
```

`"Error" not in mcp_out` as a success predicate passes for years and then
silently inverts the day someone rewords a message.

Re-verified against current `main` before changing anything: both f-strings
still present, at the lines above.

## The change

- `op=transition` → `{"memory_id", "old_status", "new_status", "message"}`
- `op=delete` → `{"memory_id", "deleted": true, "message"}`

**Key names come from the code, not the report.** The report suggested `id`;
REST `PATCH /memories/{id}/status` actually returns `memory_id` / `old_status`
/ `new_status` (`openapi_responses.MemoryStatusPatchResponse`). Copying REST's
real keys is what lets one client parse both surfaces, so that is what this
does. A test pins the MCP payload against `MemoryStatusPatchResponse` itself
rather than a literal list, so renaming a field on the REST side fails here
instead of silently reopening the divergence.

REST `DELETE` returns 204 with no body, so `op=delete` has no shape to copy. It
reuses `memory_id` rather than a bare `id` so one tool does not name the same
thing two ways.

`deleted: true` reports that the delete *succeeded*, not that the row is gone —
this is a soft delete, which is why the smoke asserts the read-back 404 rather
than trusting the acknowledgement. Called out in a comment so the field is not
misread later.

The old prose is preserved in a `message` field for chat rendering, per the
report's own recommendation.

## Tests

`tests/test_mcp_manage_structured_returns.py` (6 cases) plus the two existing
happy-path tests in `tests/test_mcp_manage.py`, which substring-matched the
prose — the anti-pattern this change exists to retire — and now assert the
structured keys.

Confirmed failing without the fix: reverting `core-api/src` fails all 8.

The case worth naming is
`test_success_is_distinguishable_from_refusal_without_reading_text`: success
and refusal are now both JSON objects told apart by the presence of an `error`
key. That is the property the report was actually asking for — a structural
predicate instead of a substring one — and it is what the smoke's `ok = ...`
line can be replaced with.

Full suite green locally: 5803 passed, 4 skipped, 1 xfailed. `ruff check`,
`ruff format --check`, and `mypy` clean at CI's scopes.

### DIFF
--- a/core-api/src/core_api/mcp_server.py
+++ b/core-api/src/core_api/mcp_server.py
@@ -1428,7 +1428,32 @@ async def caura_manage(
                     resource_id=uid,
                     detail={"old_status": old_status, "new_status": status},
                 )
-                return _with_latency(f"Memory {memory_id} status updated: {old_status} -> {status}", t0)
+                # Structured, not prose. ``op=read``/``op=update`` return
+                # serialized objects and every refusal returns a JSON error
+                # envelope, so a prose sentence here left one tool with three
+                # return shapes and forced callers to sniff the type — and then
+                # to sniff the TEXT, because a string is the success case only
+                # if it does not happen to contain "Error". A predicate like
+                # that passes for years and silently inverts the day someone
+                # rewords a message.
+                #
+                # Keys copied from REST ``PATCH /memories/{id}/status``
+                # (``memory_id``/``old_status``/``new_status``, see
+                # ``openapi_responses.MemoryStatusPatchResponse``) so the two
+                # surfaces are parseable by one client.
+                return _with_latency(
+                    json.dumps(
+                        {
+                            "memory_id": str(uid),
+                            "old_status": old_status,
+                            "new_status": status,
+                            # The old prose, kept for chat rendering.
+                            "message": f"Memory {memory_id} status updated: {old_status} -> {status}",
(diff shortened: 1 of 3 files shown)


erni-a (approved):

Eldad-Caura:

erni-a (approved):

PR #1196, opened 2026-09-01
fix(usage): let a tenant at its plan limit still transition a memory

Eldad-Caura:
Item 4 of the REST/MCP parity smoke report's "What to fix next" queue.

## The drift

A tenant sitting at its write cap was **refused** a status transition over REST
and **allowed** one over MCP. Same operation, same tenant, different answer
depending on the transport.

Re-verified on current `main` before changing anything:

| verb | `enforce_usage_limits` | `enforce_read_only` | increments |
|---|---|---|---|
| create | ✓ | ✓ | write |
| bulk create | ✓ | ✓ | bulk |
| update | ✗ | ✗ | write |
| **transition** | **✓** | ✓ | ✗ |
| delete / bulk-delete | ✗ | ✓ | ✗ |
| redistribute | ✓ | ✓ | bulk |

MCP: `enforce_usage_limits` has **zero** code occurrences in `mcp_server.py`.

## The decision — transitions are free, on both surfaces

Transitions and deletes are how a tenant gets back *under* its limit.
`AuthContext.enforce_usage_limits` says so in its own docstring:

> Do NOT use on delete endpoints — users in read-only mode must be able to
> delete data to get back under limits.

Archiving or retracting a memory is that same action. Gating it contradicted
the rule it was borrowing, and locked an over-plan tenant out of the one
operation that would have shrunk its footprint. So the REST route drops
`enforce_usage_limits()` rather than MCP gaining it.

`enforce_read_only()` **stays** on the route — that is the demo-mode gate, a
different question — and has its own test, so "transitions are free" cannot be
over-read into "transitions are ungated".

## One place, not two omissions

The policy now lives in `usage_service.USAGE_WRITE_QUOTA_POLICY`.

It previously existed only as the presence or absence of a call at each route
and each tool, so "free" was expressed by an *omission* — invisible in review,
and exactly how the two surfaces drifted. `charges_write_quota` raises on an
unknown verb rather than defaulting, so adding a mutating operation has to be a
decision instead of silently landing free (revenue leak) or silently charged
(surprise refusals).

The existing delete-is-free policy note in `mcp_server` is extended to point at
the table rather than restating it, per the report's instruction not to start a
second policy note.

## Tests

`tests/test_write_quota_verb_parity.py` — 8 cases, behavioural rather than a
table-against-table restatement: the read-only case runs against the live route
on both surfaces, so they fail if a route changes, not only if the table does.

Confirmed failing without the fix, and precisely: restoring
`enforce_usage_limits()` on the transition route fails
`test_rest_transition_is_allowed_at_the_plan_limit` **and nothing else** — the
create-still-refused and demo-still-refused guards stay green, so the test is
targeted at this change rather than at the gate's existence.

Full suite green locally: 5812 passed, 4 skipped, 1 xfailed. `ruff check`,
`ruff format --check`, `mypy`, and the broker OpenAPI baseline check all clean.

## Two adjacent gaps found while verifying — deliberately not changed here

Both confirmed on current `main`, neither in the original report, and both are
production-impacting behaviour changes that want their own decision:

1. **REST `PATCH /memories/{id}` (update) enforces neither `enforce_read_only()`
   nor `enforce_usage_limits()`**, while still charging write quota. A demo-mode
   or over-plan org can edit memory content over REST today.
2. **The MCP surface has no plan-limit enforcement at all.** The MCP middleware
   never reads the gateway's `x-org-read-only` header — there is no context var
   for it and no getter — so no MCP tool can see read-only mode. This affects
   every mutating tool, not just `transition`.

Closing (2) would start refusing MCP writes for over-plan tenants that succeed
today, which is why it is not folded in here.

### DIFF
--- a/core-api/src/core_api/mcp_server.py
+++ b/core-api/src/core_api/mcp_server.py
@@ -93,8 +93,8 @@ from core_api.services.recall_service import summarize_memories
 # directly from ``core_api.services.trust_service``.
 from core_api.services.trust_service import parse_trust_error
 from core_api.services.trust_service import require_trust as _require_trust
+from core_api.services.usage_service import charges_write_quota, recall_operation
 from core_api.services.usage_service import check_and_increment_by_tenant as check_and_increment
-from core_api.services.usage_service import recall_operation
 from core_api.trust_utils import (
     effective_keystone_min_trust,
     keystone_min_trust,
@@ -1071,7 +1071,8 @@ async def caura_write(
             if not fleet_id and agent.get("fleet_id"):
                 fleet_id = agent["fleet_id"]
             if content is not None:
-                await check_and_increment(tenant_id, "write")
+                if charges_write_quota("create"):
+                    await check_and_increment(tenant_id, "write")
                 result = await create_memory(
                     MemoryCreate(
                         tenant_id=tenant_id,
@@ -1429,6 +1430,14 @@ async def caura_manage(
                         t0,
                     )
                 old_status = memory.get("status")
+                # Asked, not assumed — ``transition`` is not in
+                # ``WRITE_QUOTA_OPS``, so this charges nothing today. Written as
+                # a lookup rather than as an absent call because "free"
+                # expressed by omission is invisible in review, and an omission
+                # on one surface and a call on the other is precisely how these
+                # two drifted.
+                if charges_write_quota("transition"):
+                    await check_and_increment(tenant_id, "write")
                 await sc.update_memory_status(str(uid), status, tenant_id=tenant_id)
                 await log_action(
                     tenant_id=tenant_id,
@@ -1495,7 +1504,8 @@ async def caura_manage(
                     )
                 # WRITE → home tenant only. ``update_memory`` is storage-routed
                 # and scopes the row to the explicit ``tenant_id``.
-                await check_and_increment(tenant_id, "write")
+                if charges_write_quota("update"):
+                    await check_and_increment(tenant_id, "write")
                 result = await update_memory(uid, tenant_id, MemoryUpdate(**fields), agent_id=agent_id)
                 return _with_latency(_serialize(result), t0)
             # op == "delete" — WRITE → home tenant only.
@@ -1519,6 +1529,8 @@ async def caura_manage(
                         ),
                         t0,
                     )
+            if charges_write_quota("delete"):
+                await check_and_increment(tenant_id, "write")
             await soft_delete_memory(uid, tenant_id)
             # Structured for the same reason as ``op=transition`` above. REST
             # DELETE returns 204 with no body, so there is no shape to copy
@@ -3452,6 +3464,13 @@ async def caura_keystones_set(
                 # because deletes don't charge the write budget (mirrors
                 # the REST route skipping ``enforce_usage_limits`` on
                 # DELETE).
+                #
+                # That reasoning, and the matching decision for ``transition``,
+                # are recorded in ``usage_service.WRITE_QUOTA_OPS`` /
+                # ``PLAN_LIMIT_GATED_OPS``. Keystone ops are their own verb
+                # namespace and are NOT in those tables, so this branch stays
+                # hand-coded — the memory-CRUD call sites on both surfaces are
+                # the ones wired through the lookup.
                 await check_and_increment(tenant_id, "write")
                 # Storage owns scope/weight/fleet/agent shape validation;
                 # surface its 422s directly so we don't drift from the
--- a/core-api/src/core_api/routes/memories.py
+++ b/core-api/src/core_api/routes/memories.py
@@ -98,7 +98,9 @@ from core_api.services.tenants import list_active_tenant_ids
 from core_api.services.trust_service import parse_trust_error, require_trust
 from core_api.services.usage_service import (
     bulk_check_and_increment,
+    charges_write_quota,
     check_and_increment,
+    plan_limit_gated,
     recall_operation,
 )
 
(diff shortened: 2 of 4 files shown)


Eldad-Caura:
@claude Addressed the review — the finding was correct, and verifying it turned up a related problem in my own table.

**`charges_write_quota` was only ever called by tests.** Confirmed by grep before changing anything. The table was decorative exactly as described.

**And a single boolean was the wrong model.** The two axes do not line up: `update` charges the write budget while calling neither `enforce_usage_limits()` nor `enforce_read_only()`. Applying fix option 1 literally — `if charges_write_quota("update"): auth.enforce_usage_limits()` — would have *added* a plan-limit gate that does not exist today. So the policy is now two sets, `WRITE_QUOTA_OPS` and `PLAN_LIMIT_GATED_OPS`, and `update` is recorded as it behaves rather than as it arguably should; encoding it faithfully is what makes wiring a call site through it behaviour-preserving.

**Wired at real call sites** (option 1, scoped to the memory-CRUD verbs the tables actually cover): create/transition/update/delete on the MCP handler, create/transition on the REST routes. Keystone and document ops are a separate verb namespace and stay hand-coded — now stated where they are instead of implied, since the old comment there over-claimed.

**Tautological tests removed** (option 3). `test_every_mutating_verb_has_a_policy` and the bare `charges_write_quota(...)` assertions were checking the table against itself and would have passed with nothing consuming it. Replaced with two that flip a table row and assert the *surfaces* react — verified by deleting the lookup at both call sites and watching exactly those two fail, then restoring.

Remaining honest gap, stated rather than papered over: `PLAN_LIMIT_GATED_OPS` is REST-only in practice, because the MCP middleware never reads `x-org-read-only` and so no MCP tool can see plan-limit mode at all. That is noted where the table is defined and is one of the two follow-ups in the PR description.

Full suite 5833 passed / 4 skipped / 1 xfailed after rebase; `ruff check`, `ruff format --check`, `mypy`, and the broker OpenAPI check all clean.

Eldad-Caura:
@claude Second round addressed. I checked both points against the code before acting, and one of them does not hold — but it exposed a real weakness in my rationale, so the outcome is a change either way.

**Direction-agnostic transitions — not a bypass, but the rationale was wrong.**

The concern needs an active-row or footprint metric for a reactivation to inflate. There isn't one. `tenant_usage_counters` is keyed `(tenant_id, operation, period_start)` — a monotonic count of *operations* per period (`usage_meter.py`: `self._counts[(tenant_id, operation, current_period_start())] += count`). No status dimension exists anywhere in the OSS metric. So `archived -> active` inflates nothing.

The catch is that this cuts *both* ways, and that is the part worth having: my original justification — "transitions are how a tenant gets back under its limit" — was equally unfounded, because archiving doesn't reduce a per-period operation count either. Both directions are inert.

Replaced with the principle that actually holds and is direction-independent: read-only mode exists to stop an over-plan org **growing the store**, and a transition writes one column on a row that already exists. `active -> archived` and `archived -> active` leave the store exactly the same size. That is also what `enforce_usage_limits`'s own docstring is getting at with the delete carve-out.

Added `test_reactivating_transition_is_allowed_at_the_plan_limit` so the permissive direction is a reviewed decision rather than a side-effect, plus a ⚠ REVISIT note: if a plan ever meters live rows, archiving *would* reduce usage, reactivating *would* raise it, and the verb would need to discriminate on `old_status -> new_status`.

One note on the suggested fix: applying it on MCP too, "so the two surfaces stay in parity", is not implementable today. MCP cannot see plan-limit mode at all — the middleware never reads `x-org-read-only`. Direction-gating REST alone would rebuild the exact drift this PR closes.

**Typed op names — adopted, and it does not do what it looks like.**

Added `MutatingOp = Literal[...]`. Then I tested it: mistyped `plan_limit_gated("transitionn")` and mypy still reported "Success: no issues found in 203 source files". Both `core_api.routes.memories` and `core_api.mcp_server` are on the `ignore_errors` list in `core-api/pyproject.toml` — which is where every lookup this PR adds lives.

Kept the annotation (correct for non-exempt callers, and it starts working the day either module comes off a list the config itself calls "a to-do, not a policy"), but the comment now says plainly that the **runtime check is the actual protection** at these call sites, so nobody deletes it later assuming the type covers it.

Full suite 5834 passed / 4 skipped / 1 xfailed; `ruff`, `mypy`, and the broker OpenAPI check clean.

Eldad-Caura:
@claude Third round — the criticism lands on a claim I made without backing it, and that is now fixed properly.

I wrote "reported separately" in the code comments, the commit message and the PR body, and had **not actually filed anything**. An unlinked "tracked elsewhere" is a claim with no referent, which is worse than saying nothing. Both gaps are now real issues, and the code comments cite them by number instead:

- **#1204** — `PATCH /memories/{id}` charges write quota while gated by neither `enforce_read_only()` nor `enforce_usage_limits()`, making it the only mutating memory route with neither. Includes the full per-route table and separates the two halves: the demo-mode bypass looks like an oversight, while the plan-limit half is arguable — the growth principle this PR settled on cuts toward *not* gating an update, since it adds no row.
- **#1205** — the MCP surface has no plan-limit enforcement at all, because `MCPAuthMiddleware` never reads `x-org-read-only`. This is the general case of the drift this PR fixes for one verb, and it is why `PLAN_LIMIT_GATED_OPS` is REST-only in practice.

I went with option (a) rather than (b) deliberately. Adding `auth.enforce_read_only()` to `update_memory_endpoint` in this PR would be a second live behaviour change — demo-mode tenants lose the ability to edit existing memories — bundled into a PR whose scope was decided as the transition verb. It is the right fix and #1204 says so; it should land as its own reviewed change rather than riding along, and the same reasoning is why this PR records the table as the system behaves rather than as it should.

No functional change in this round: the diff versus the previous push is two comments and one docstring gaining issue numbers. Full suite 5834 passed / 4 skipped / 1 xfailed; `ruff`, `mypy`, and the broker OpenAPI check clean.

Note the branch had been force-pushed since my last push; I verified the remote head was my own round-2 content on the same base before re-pushing, so nothing was discarded.

Eldad-Caura:
@erni-a Review came back clean after three rounds — ready for your approval.

Two follow-ups were filed rather than folded in, both live behaviour changes: #1204 (`PATCH /memories/{id}` charges write quota while gated by neither read-only nor plan limits) and #1205 (the MCP surface has no plan-limit enforcement at all).

erni-a (approved):

PR #1207, opened 2026-09-01
feat(write): report where a memory's weight came from

Eldad-Caura:
Item 6 of the REST/MCP parity smoke report's queue (P3) — the reporting half.

## The gap

`pipeline/steps/write/merge_enrichment_fields.py`: when the caller omits
`weight`, it comes from the enrichment LLM.

That is not cosmetic. `SIMILARITY_BLEND = 0.85` makes weight **15% of the base
rank score**, so identical content can rank differently run to run. In the
2026-08-24 run, two throwaway `fact` memories written seconds apart got **0.30
and 0.50**. It also made an evolve comparison look like a REST/MCP divergence —
`new_weight` 0.15 vs 0.35 on an identical `delta=-0.15` — when both surfaces
were arithmetically correct and only the starting weights differed.

Nothing in the write said which values a model had chosen, so a team chasing
non-reproducible ranking could not tell an LLM guess from a deliberate number.

Re-verified on `main`: the assignment is unchanged and `weight_source` appears
nowhere in the repo.

## The change

`weight_source`, recorded per write:

| value | meaning | reproducible? |
|---|---|---|
| `caller` | the write supplied it | yes |
| `llm` | the enrichment model chose it | **no** |
| `default` | `DEFAULT_MEMORY_WEIGHT` | yes |

**Three values, not a boolean.** "Not the caller" collapses exactly the
distinction reproducibility turns on: a default is repeatable and an LLM pick
is not.

### It's a port, not a new pattern

`memory_type_agent_set` (CAURA-703) already answers this same provenance
question for `memory_type` — same `set_system_value` call, same
`PLATFORM_ONLY_KEYS` registration. This follows it exactly, including the
platform-only registration: a forgeable `weight_source: "caller"` sitting on an
LLM-chosen weight would be **worse than no field**, because it would look
authoritative.

Set on the bulk path too (`create_memories_bulk` merges outside the pipeline),
which would otherwise report nothing where a single write reports a source.

### Two behaviour-preserving details

- The enrichment branch now checks `enrichment.weight is not None` before
  assigning. Outcome is identical — a `None` fell through to the default either
  way — but it stops the flag claiming `llm` for a value the model never
  supplied.
- The caller check is `is not None`, not truthiness, so an explicit `0.0` is
  reported as `caller` rather than being reattributed to the LLM *and
  overwritten*. Pinned by its own test, because a truthiness bug here would be
  silent and would change stored data.

## Scope

This is the **reporting** half of the report's recommendation. The tenant
setting for pinning a deterministic default is deliberately not included — that
is a configuration surface and a product decision, whereas this makes the
existing behaviour observable, which is what unblocks anyone trying to diagnose
it today. Same shape as item 1: prefer making a behaviour observable over
leaving it undocumented.

The report's related note — `weight_adjustments` from evolve having no clamp
indicator — is untouched and still open.

## Tests

`tests/test_weight_source_provenance.py` — 6 cases, confirmed failing without
the fix (all 6). They pin behaviour rather than presence: each of the three
values is asserted for a distinct input, so no hardcoded value passes.

Full suite green locally: 5831 passed, 4 skipped, 1 xfailed. `ruff check`,
`ruff format --check`, `mypy`, and the broker OpenAPI check all clean.

### DIFF
--- a/core-api/src/core_api/pipeline/steps/write/merge_enrichment_fields.py
+++ b/core-api/src/core_api/pipeline/steps/write/merge_enrichment_fields.py
@@ -48,6 +48,22 @@ class MergeEnrichmentFields:
         # skip), so this flag needs no worker-side finalisation.
         memory_type_agent_set = memory_type is not None
 
+        # Same question for ``weight``, and it matters more than it looks.
+        # ``weight`` is 15% of the base rank score (``SIMILARITY_BLEND`` is
+        # 0.85), so when the caller omits it and the enrichment LLM picks one,
+        # identical content can rank differently run to run — two throwaway
+        # ``fact`` memories written seconds apart were observed getting 0.30
+        # and 0.50. Nothing in the response said which values were chosen by a
+        # model, so a team chasing non-reproducible ranking had no way to tell
+        # an LLM guess from a deliberate number.
+        #
+        # Three-valued rather than a boolean, because "not the caller" hides
+        # the distinction that matters for reproducibility: ``default`` is
+        # deterministic (``DEFAULT_MEMORY_WEIGHT``, used when enrichment is off
+        # or failed) while ``llm`` is not. Captured BEFORE the merge below,
+        # like ``memory_type_agent_set``.
+        weight_source = "caller" if weight is not None else "default"
+
         # CAURA-701: caller-supplied deprecated types (currently ``semantic``)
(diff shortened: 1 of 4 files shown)


Eldad-Caura:

erni-a (approved):

PR #1209, opened 2026-09-01
feat(llm): migrate VertexLLMProvider to google-genai, unlock Gemini 3.x

ran-taig:
## Why

The legacy `vertexai.generative_models` SDK (deprecated June 2025, removal date June 2026) structurally cannot reach any Gemini 3 model:

1. Its `validate_region` allowlist predates the `us`/`eu` **multi-regions** — `aiplatform.init(location="us")` raises `ValueError` before any request is built.
2. The Gemini 3.x **-lite** models are served *only* on those multi-regions (verified live 2026-08-30/31: `gemini-3.5-flash-lite` and `gemini-3.1-flash-lite` → 200 on `locations/us` via the bare `aiplatform.googleapis.com` host; 404 on `global` and on regional `us-central1`).
3. Multi-regions use the **bare host** — the SDK-built `us-aiplatform.googleapis.com` is rejected with `400 Invalid hostname`.

## What

Rewrites `VertexLLMProvider` on **google-genai** (`vertexai=True`, ADC — mirroring `GeminiLLMProvider`, which already uses the same SDK in API-key mode), with a `base_url` pin for `location ∈ {us, eu}` and untouched endpoint logic for `global`/regional.

**Byte-compatible public surface**: class name, constructor, `provider_name`, `complete_json`/`complete_text` signatures, `VertexResponseShapeError`, the CAURA-651 shape guard, the max-output-tokens truncation guard (#1009), and the exact log lines dashboards key on. The client is built **lazily** so construction (startup, tests) needs no ADC.

**Deps**: core-api already has google-genai; core-worker's `vertex` extra now points at google-genai instead of google-cloud-aiplatform (lock regenerated — the aiplatform/storage stack drops out of the worker image, net −133 lock lines).

## Behavior change for current prod config: none

Same env config (`gemini-2.5-flash-lite` @ `us-central1`), same request parameters, same billing SKU — verified live with the migrated provider (token-injected ADC):

| location / model | result |
|---|---|
| `us-central1` / gemini-2.5-flash-lite (current prod) | ✅ 1341ms, valid JSON |
| `global` / gemini-2.5-flash-lite | ✅ 843ms |
| **`us` / gemini-3.1-flash-lite** (upgrade target) | ✅ 1454ms, valid JSON |
| `complete_text` on us/3.1-flash-lite | ✅ |

After this merges, moving any deployment to Gemini 3.x-lite is a pure config change (`PLATFORM_LLM_MODEL` + `PLATFORM_LLM_GCP_LOCATION=us`).

## Testing

- `test_llm_json_truncation.py` rewritten for the genai client shape + 2 new tests: multi-region base_url pinning (and its absence for global) and lazy/cached client construction.
- `test_optional_provider_sdks_stay_optional.py`: vertex case now hides `google.genai` (the module must import without the SDK — worker installs it only via the `vertex` extra).
- Full affected set green: 144 passed (truncation, optional-SDK, gemini, platform-providers, provider-protocols). Ruff clean.
- Note: `gemini-3.1-flash-lite` showed **zero thinking tokens** on JSON calls (usageMetadata-verified), so no thinking config is plumbed; revisit if a thinking-default model is adopted.

### DIFF
--- a/common/llm/providers/vertex.py
+++ b/common/llm/providers/vertex.py
@@ -1,34 +1,26 @@
-"""Vertex AI LLM provider.
+"""Vertex AI LLM provider (Gemini via Workload Identity / ADC).
 
-Wraps the Google Cloud Vertex AI SDK. Since the SDK is synchronous,
-all calls are wrapped in ``asyncio.to_thread()`` to avoid blocking
-the event loop.
+Wraps the ``google-genai`` SDK in Vertex mode (``vertexai=True``) —
+project/location auth via Application Default Credentials, no API key.
+Migrated off ``vertexai.generative_models`` (deprecated June 2025,
+removed June 2026): the legacy SDK only speaks the v1 API, where the
+Gemini 3.x catalog is not published, and its region allowlist predates
+the ``us``/``eu`` multi-regions the 3.x Flash-Lite models are served
+from — so it structurally cannot reach any Gemini 3 model.
 
-CAURA-333: ``VertexEmbeddingProvider`` was removed (broken — never passed
-``output_dimensionality`` to the SDK, so writes failed against pgvector's
-1024-dim column). Only the LLM-side provider remains.
+Multi-region endpoint quirk: ``us``/``eu`` are served from the bare
+``aiplatform.googleapis.com`` host (same as ``global``), but the SDK
+builds ``{location}-aiplatform.googleapis.com`` for any non-``global``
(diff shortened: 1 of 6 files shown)


ran-taig:
@claude

ran-taig:
@claude

ran-taig:
@claude

ran-taig:
@erni-a

erni-a (approved):

PR #1217, opened 2026-09-01
fix(api): enforce the read-only capability gate on memory update

### LINKED ISSUE #1204: core-api: PATCH /memories/{id} charges write quota but is gated by neither read-only nor plan limits
Found while verifying the write-quota parity work (#1196). Not in the original
REST/MCP parity report.

## The gap

`PATCH /memories/{memory_id}` (`routes/memories.py`, `update_memory_endpoint`)
**charges the write quota but is gated by nothing**:

```python
auth.enforce_tenant(tenant_id)
...
if auth.tenant_id and charges_write_quota("update"):
    await check_and_increment(tenant_id, "write")
```

No `auth.enforce_read_only()`. No `auth.enforce_usage_limits()`.

It is the only mutating memory route in the file with neither. Verified against
current `main`:

| route | `enforce_usage_limits` | `enforce_read_only` | increments |
|---|---|---|---|
| `POST /memories` | ✓ | ✓ | write |
| `POST /memories/bulk` | ✓ | ✓ | bulk |
| **`PATCH /memories/{id}`** | **✗** | **✗** | **write** |
| `PATCH /memories/{id}/status` | ✗ (post-#1196) | ✓ | ✗ |
| `DELETE /memories/{id}` | ✗ | ✓ | ✗ |
| `POST /memories/bulk-delete` | ✗ | ✓ | ✗ |
| `POST /memories/redistribute` | ✓ | ✓ | bulk |

## Why it matters

Two separate consequences, and the second is the more clearly wrong of the two:

1. **Over-plan orgs can still edit content.** An org in plan-limit read-only
   mode is refused a create but may rewrite the content, title, metadata and
   weight of every memory it already has. Whether that *should* be refused is
   arguable — an update does not grow the store, which is the principle #1196
   settled on for transitions — but it is currently unstated either way.

2. **Demo-mode orgs can still edit content.** This one looks like an oversight
   rather than a decision. `enforce_read_only()` is the demo/`is_demo` gate and
   is on *every* other mutating route including delete. There is no principled
   reason a demo credential should be blocked from creating and deleting but
   allowed to overwrite existing memories.

The asymmetry is also self-inconsistent: the route charges write budget for the
operation while declining to check whether the org is allowed to spend it.

## Suggested fix

Add `auth.enforce_read_only()` unconditionally — that closes the demo bypass and
is consistent with every sibling route.

Then decide `enforce_usage_limits()` deliberately, i.e. whether `update` joins
`PLAN_LIMIT_GATED_OPS`. Note the growth principle recorded there cuts toward
*not* gating it (an update adds no row), which would make the current
plan-limit behaviour correct-by-accident and worth stating rather than changing.

Both are behaviour changes for live tenants, which is why #1196 recorded the
table as the system BEHAVES and left this alone rather than folding it in.

Eldad-Caura:
Closes #1204, which I filed while verifying the write-quota parity work in
#1196 — and which understated the problem. I filed it as a demo-mode bypass
and a billing inconsistency. Re-verifying it before writing the fix showed the
demo half is the *less* serious half.

## The bug

`enforce_read_only()` guards **two** signals, not one. Its own docstring:

- `is_demo` → the demo sandbox is read-only
- `capabilities` is set and does **not** include `write` → the credential is
  read-only by construction (a viewer or reporting key)

`PATCH /memories/{memory_id}` never calls it. `enforce_tenant()` — the only
authorization it does call — checks tenant binding and the admin bypass, and
never looks at capabilities.

So a credential minted read-only could rewrite the content, title, metadata
and weight of any memory in its tenant. Not a 403:

```
PATCH /memories/{id}  →  200 OK
{"content": "rewritten by a read-only key", ...}
```

The error text that *should* have fired says the quiet part: read-only is
**"a property of the CREDENTIAL, not of the endpoint"**. That only holds if
every write-shaped endpoint asks, and this one did not.

### The asymmetry

| route | `enforce_read_only` |
|---|---|
| `POST /memories` | ✓ |
| `POST /memories/bulk` | ✓ |
| `DELETE /memories/{id}` | ✓ |
| `POST /memories/bulk-delete` | ✓ |
| `PATCH /memories/{id}/status` | ✓ |
| `POST /memories/redistribute` | ✓ |
| **`PATCH /memories/{id}`** | **✗** |

A read-only credential could not *delete* a memory but could *overwrite its
contents*. The route that rewrites content was the only one without the gate.

## Not a new class of bug

This is the same shape as **H-12 / H-13 / H-15** from the 2026-08-14 audit,
already recorded in `tests/test_route_authz_gaps.py`: *"a mutating route that
skipped a gate its own neighbours already applied."* That pass swept
`/fleet/commands`, `/agents/{id}/trust` and `/settings`. It did not reach this
one.

## The fix

One call, placed where its siblings place it. Plus the plan-limit half written
as an **explicit no-op lookup** rather than left absent:

```python
if plan_limit_gated("update"):
    auth.enforce_usage_limits()
```

`update` is not in `PLAN_LIMIT_GATED_OPS` — an update adds no row, so it does
not grow the store, which is the principle #1196 settled on — so **behaviour
here is unchanged**. It is written as a lookup because "free" expressed by
omission is invisible in review, and that is precisely what the missing
capability gate above cost.

## Verification

Three tests, in the conventions of the audit-gap file they join:

- `test_update_memory_rejects_a_read_only_credential` — **fails without this
  change** (200, content rewritten)
- `test_update_memory_refuses_the_demo_sandbox` — **fails without this change**
- `test_delete_memory_already_rejects_a_read_only_credential` — **passes
  before and after**, kept deliberately so the asymmetry stays visible in the
  suite rather than only in this description

Full suite green: 5,858 passed, 4 skipped, 1 xfailed. `ruff check`,
`ruff format --check`, `mypy` and the broker OpenAPI check are clean.

## Blast radius

Read-only credentials and demo tenants can no longer PATCH memories. That is
the intended semantics of both — but it is a live behaviour change, so it is
worth knowing rather than discovering. Nothing else changes: the plan-limit
lookup is a no-op, and write-capable credentials are unaffected.

### DIFF
--- a/core-api/src/core_api/routes/memories.py
+++ b/core-api/src/core_api/routes/memories.py
@@ -1651,6 +1651,20 @@ async def update_memory_endpoint(
       request through this endpoint that removes one.
     """
     auth.enforce_tenant(tenant_id)
+    # The capability gate every other mutating memory route already applies.
+    # ``enforce_tenant`` above checks tenant binding and the admin bypass — it
+    # never looks at capabilities, so without this a credential minted
+    # read-only (capabilities={'read'}) could rewrite the content, title,
+    # metadata and weight of any memory in its tenant, while the DELETE
+    # sibling refused it. Same shape as the H-12/H-13/H-15 findings.
+    auth.enforce_read_only()
+    # Asked, not assumed, per the table added in #1196: ``update`` is not in
+    # ``PLAN_LIMIT_GATED_OPS`` (an update adds no row, so it does not grow the
+    # store), making this a no-op today. Written as a lookup rather than left
+    # absent because "free" expressed by omission is invisible in review — the
+    # missing gate above is exactly what that costs.
+    if plan_limit_gated("update"):
+        auth.enforce_usage_limits()
     # C3/C8: server-reserved types (outcome/rule/insight) are authored only by
     # internal flows — reject agent-supplied values on update too, mirroring the
     # create (POST /memories) and bulk-create handlers.
--- a/core-api/src/core_api/services/usage_service.py
+++ b/core-api/src/core_api/services/usage_service.py
@@ -105,13 +105,14 @@ OperationType = Literal["write", "search", "recall", "insights", "evolve"]
 # ``enforce_read_only()`` is NEITHER axis. That is the demo-mode gate, a
 # separate question, and it stays on the transition route.
 #
-# ``update`` is recorded as it BEHAVES, not as it arguably should: it charges
-# quota while skipping both ``enforce_usage_limits()`` and
-# ``enforce_read_only()``, which makes it the only mutating memory route gated
-# by neither (caura-ai/caura#1204 — the demo-mode half of that looks like an
-# oversight rather than a decision). Encoding it faithfully here keeps this
-# table a description of the system rather than an aspiration, so wiring a call
-# site through it cannot silently change behaviour.
+# ``update`` charges quota and is deliberately NOT plan-limit gated: an update
+# rewrites a row rather than adding one, so it does not grow the store, which
+# is the principle this table encodes. That omission is a decision, not the
+# oversight it once sat beside — ``update`` also skipped ``enforce_read_only()``
+# and so was gated by neither, which let a read-only credential rewrite any
+# memory in its tenant (caura-ai/caura#1204). That half is now closed: the route
+# calls ``enforce_read_only()`` like every other mutating memory route, which is
+# what makes the remaining omission legible as a choice.
 WRITE_QUOTA_OPS: frozenset[str] = frozenset({"create", "bulk_create", "update", "redistribute"})
 
 # Ops refused when the org is over its plan limit. A subset of the above minus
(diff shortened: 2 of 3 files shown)


Eldad-Caura:
@claude — addressed the review finding.

You were right, and the miss was mine: I changed the behaviour and left the
comment that documents it asserting the opposite. `usage_service.py` still
described `update` as "gated by neither" and cited #1204 as an open oversight,
one file away from the call that closes it.

Rewritten rather than patched, because the original sentence bundled two things
that this PR separates. It now says:

- `update` is **deliberately** not plan-limit gated — an update rewrites a row
  rather than adding one, so it does not grow the store, which is the principle
  the table encodes. That is a decision.
- `update` also **skipped `enforce_read_only()`**, which is what made the
  quota/gate asymmetry a bug rather than a curiosity. That half is now closed.

I dropped a third paragraph I had drafted restating "`enforce_read_only()` is
not an axis of this table" — the note four lines above it already says exactly
that, and repeating it made the block harder to read, not safer.

No behaviour change in the amend; comments only. Re-verified after it:
`test_route_authz_gaps.py`, `test_write_quota_verb_parity.py` and
`test_d13_recall_metering.py` pass (43), and `ruff check` / `ruff format
--check` / `mypy` / the broker OpenAPI check are clean. The full suite was green
on the pre-amend tree at 5,858 passed.

arkash20 (approved):

PR #1228, opened 2026-09-01
fix(search): stop withholding information the caller needs (A28, C6, D6)

arkash20:
## Summary

Three canonical rows, one theme: the API knew something and didn't tell the caller.

### A28 — successor injection truncated silently

`MAX_SUCCESSOR_LOOKUPS = 10` sliced the stale-id list, so a result set with **more than ten** outdated/conflicted rows left the 11th onward with no successor loaded and **no signal**. A stale claim then surfaces as if no correction exists — the exact failure the A34 contract exists to prevent. Only a `logger.warning` recorded it.

The task text assumed the cap protected against per-ID round-trips and proposed "batch into one storage call". **That was already the case:** `find_successors` is one HTTP call, and on the storage side one indexed query (`supersedes_id IN (...)`, no `LIMIT`). Ten ids and a thousand cost the same round-trip, so the truncation bought nothing.

Raised to a defensive `1000` — unreachable from the public API, since `top_k` caps at `MAX_SEARCH_TOP_K` (20) — and if it ever *does* engage, the caller is told. A new coded warning also covers a `find_successors` failure, which was silent in exactly the same way.

### C6 — "intentional or oversight?" → intentional, now said out loud

`extra="forbid"` already rejected `supersedes_id`, but with `"Extra inputs are not permitted"`, so a caller couldn't tell a deliberate refusal from a misspelling.

It **stays refused.** Supersession is server-owned: the contradiction detector sets it through `update_memory_status`, which is CAS-guarded so two concurrent detections can't both claim the same predecessor. Accepting it on create would point a brand-new row at an arbitrary predecessor with no check that it exists, shares the tenant, or isn't already superseded — and bypass the CAS entirely.

A `model_validator(mode="before")` replaces only the *message*. Note the error deliberately does **not** name a route: there is no public endpoint that sets `supersedes_id`, so pointing at one would send callers nowhere. It names the mechanism instead — write the corrected fact, detection links it.

### D6 — the override already existed; the documentation didn't

Default search hides superseded rows (`outdated` always; `conflicted` unless an exact lexical match). The `status_filter` override this task proposed **is already implemented and plumbed**. What was missing was any statement that a default policy applies at all — which is what made it feel silent: a caller couldn't distinguish "nothing matched" from "matches were withheld".

I did **not** build the suggested `conflicted_suppressed_count`. The exclusion is a WHERE clause, so an honest count means re-running the scored query — a second full search on a hot read path — to report a number nobody has asked for. If a caller ever needs it, it should be opt-in.

## Wire impact

`SearchResponse.warnings` is additive and is `null` when empty — **not absent**. FastAPI serializes `None` fields, exactly as the existing `diagnostic` field already behaves. Deliberately not switched to `exclude_none`, which would change `diagnostic`'s serialization too. A test asserts the two stay consistent.

## Testing

11 unit tests. Full suite **5719 passed**; the 8 failures in `test_capability_usage` / `test_request_observation` **pre-exist on main** (verified by stashing and re-running). ruff + format clean on every touched file.

**Wet-tested** against the deployed image on real vectors:

```
1) create WITH supersedes_id -> 422  "...is set by the server, not by the caller...
                                      Supersession is established by contradiction detection"
2) create with a TYPO        -> 422  extra_forbidden (unchanged)
3) ordinary write            -> 201
4) search                    -> 200  warnings: None
5) three real supersession pairs:
     successor #1 -> supersedes #2 [conflicted]  OK
     successor #2 -> supersedes #3 [conflicted]  OK
     successor #3 -> supersedes #4 [conflicted]  OK
6) live OpenAPI status_filter documents the default exclusion
```

Refs: canonical A28, C6, D6.

### DIFF
--- a/core-api/openapi.broker.json
+++ b/core-api/openapi.broker.json
@@ -1657,6 +1657,7 @@
                 "type": "null"
               }
             ],
+            "description": "Restrict results to one status. When omitted, superseded memories are hidden: 'outdated' rows are excluded, and 'conflicted' rows are excluded unless they exactly match the query text. Their replacements are injected in their place. Pass an explicit status to bypass this and read the superseded rows directly.",
             "title": "Status Filter"
           },
           "tenant_id": {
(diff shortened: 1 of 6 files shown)


arkash20:
@Eldad-Caura please approve

Eldad-Caura (dismissed):

arkash20:
Rebased onto main and regenerated `core-api/openapi.broker.json` — the new `SearchResponse.warnings` field and the `status_filter` description made the frozen broker baseline stale, which is what the earlier CI failure was.

The rebase also conflicted with the `recall_tracked` work that landed in `/search` while this was in flight; both fields are now collected and returned on both the diagnostic and non-diagnostic paths.

CI green on `eafdfb83` (13 checks, 0 failures).

@Eldad-Caura please approve

Eldad-Caura (approved):

PR #1229, opened 2026-09-01
docs(gate): say what the tenant-scope categories actually assert

Eldad-Caura:
Closes the last item of the #1180 sizing. Four pieces of prose in the category block had drifted from the code they describe.

**Nothing moves.** The gate reports `added (0)  removed — fixed (0)  removed — gone (0)  removed — still unscoped (0)  recategorised (0)`, 64 allowlisted before and after. No classification logic changes.

## `opaque-body-write` — the load-bearing clause named nothing

It ended *"the route above is responsible for having validated it."* #1180 called that the load-bearing clause and asked **which route, and validated against what?** There is no answer: for a caller reaching this service directly there is no route above.

Since **#1066** the only credential this service accepts is a shared secret carrying no tenant identity — it authenticates *that* a caller is internal, never *which* tenant it speaks for. A foreign `tenant_id` in the body is written faithfully and nothing here can tell. That is true of **34 of the 36** entries this category held when #1180 was filed, so the definition should say it rather than defer to an unnamed party.

Now: the payload tenant settles which tenant **owns** the new row; it does not settle whether that tenant is the **right** one; the accepted control is that the caller was correct.

It also records that a write referencing **other** rows by bare id is a separate question the category does not answer — the id needs its own tenant check. That is precisely what #1215 had to add to `/memories/conflicts`, and what `POST /fleet/commands` already did.

## `id-addressed-read` — the premise was falsified

Still said every entry is one *"an attacker who can reach storage directly can use"* — the staleness #1180 flagged in passing. #1066 closed the port to unauthenticated callers.

Rewritten to say it is any caller that reaches this service: before #1066 anyone who could reach the port, now any holder of the shared secret. **A narrower set, not a smaller consequence** — the secret says nothing about which tenant its holder speaks for, so the reach of a single entry is unchanged. Worth stating that way round, because "we added auth" reads as mitigation and here it is not.

## `blind-spot` — it never said it was a backlog category

It is in `BACKLOG_CATEGORIES`, but nothing in its own text said so, and it reads as a tidy home for anything awkward to classify. Filing there grows the backlog exactly as `id-addressed-*` does.

**I got this wrong myself on #1180** — I recommended moving six `insights/*` routes here and called it the zero-risk option, which would have grown the backlog by six while looking like tidying. The definition now says it is backlog, and records the specific trap: a `_require` the pass misses because it sits one frame down in a **helper** does not belong here, because the pass reads each handler's own AST, so hoisting the call into the handler scores the route REQUIRED and the entry **deletes outright**. That is what #1199 did instead.

## The header count

Said *"Six categories"* and *"Two describe a backlog"*. There are **eight**, and `BACKLOG_CATEGORIES` holds **four**. Now points at that frozenset as the authority rather than restating the membership in prose that can drift again.

## Three entry notes — the bucket-E findings

From the #1180 trace, so the conclusions live where the next reader meets them:

- **`method:dedup_review_enqueue`** and **`route:POST /memories/dedup-reviews`** — the caller-supplied memory ids are unvalidated but **inert**: plain UUID columns with no FK (so no existence oracle) and caller-supplied content snapshots (so storage never joins back).
- **`method:recall_log_write`** — `RecallCandidate` has no tenant of its own and no FK, scoped only by the CASCADE from `recall_event`. Inert **today only because nothing reads the table**; when `repeat_recall`'s Phase 2 read lands, that JOIN must carry a tenant predicate or a foreign `memory_id` becomes a cross-tenant content read.

That last one is a constraint on unwritten code, so an allowlist note is not where it will be found. **It still wants to be on the Phase 2 ticket or at the `repeat_recall.extract()` stub** — flagging rather than doing, since that is not this PR's file.

## Gates

Venv on `PATH`, `UV_FROZEN=1`. `ruff check` and `ruff format --check` run **separately** at each of CI's exact scopes, discovery-based, no `--config`.

- `python scripts/tenant_scope_gate.py --base origin/main` — 0 added / 0 removed / 0 recategorised, 64 allowlisted
- `python scripts/tenant_scope_gate.py --write` — the `_categories` block is generated from `CATEGORIES`, so it was regenerated, not hand-edited; re-running after the three note edits is **byte-identical** (`shasum` matched), confirming the notes survive regeneration
- `pytest tests/test_tenant_scope_gate.py` — **122 passed** (includes `check_category_doc`, which fails if the JSON block and the script disagree)
- `pytest core-storage-api/tests/` — **315 passed**
- `ruff check` / `ruff format --check` — all seven CI scopes clean

**One pre-existing lint left alone:** `ruff check scripts/tenant_scope_gate.py` reports one `ISC004`, identical on `origin/main` (line 1508 there, 1533 here — shifted by the added lines, not introduced). `scripts/` is in no CI ruff scope, so it gates nothing, and fixing it here would be an unrelated change in a docs PR.

No test is added: this changes prose only, and `check_category_doc` already fails if the generated block and the script disagree — which is the assertion that matters for this diff.

One signed-off commit, rebased onto `origin/main` immediately before push, with the gate and both suites re-run after.

### DIFF
--- a/core-storage-api/tenant_scope_allowlist.json
+++ b/core-storage-api/tenant_scope_allowlist.json
@@ -3,12 +3,12 @@
   "_categories": {
     "no-tenant-data": "Touches no tenant-scoped table at all \u2014 liveness probes, the pg_locks snapshot, idempotency keys. There is no tenant to bind to.",
     "cross-tenant-by-design": "Legitimately spans tenants and would be wrong if it did not: the capability-usage flush (one batch carries many tenants' counters, migration 023), the tenant-discovery lists that drive the lifecycle fanout, and global operational counts.",
-    "opaque-body-write": "An insert/upsert whose tenant travels as a field inside the payload dict rather than as a parameter, so no signature can carry it. The binding is real but lives in the row being written; the route above is responsible for having validated it.",
+    "opaque-body-write": "An insert/upsert whose tenant travels as a field inside the payload dict rather than as a parameter, so no signature can carry it. The binding is real but lives in the row being written, which settles which tenant OWNS the new row. WHAT IT DOES NOT SETTLE is whether that tenant is the right one. Since #1066 the only credential this service accepts is a shared secret carrying no tenant identity: it authenticates THAT a caller is internal, never WHICH tenant it speaks for. A foreign ``tenant_id`` in the body is written faithfully and nothing here can tell. The accepted control is that the caller was correct \u2014 say that plainly rather than deferring to an unnamed route. (This replaces ``the route above is responsible for having validated it``, which named no route and, for a caller reaching this service directly, referred to nothing.) A write that ALSO references other rows by bare id raises a separate question this category does not answer: that id needs its own tenant check, as ``POST /fleet/commands`` and ``POST /memories/conflicts`` do.",
     "admin-unscoped": "A deliberate operator-facing path that runs across tenants, reached only through an admin-gated route upstream.",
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@erni-a ready for approval.

Automated review came back clean (no issues found — it independently re-derived the model claims against `dedup_review.py` and `recall_log.py`) and every check is green: CI, DCO, CodeQL, security scan, legacy-name ratchet, do-not-touch sentinel.

Prose only. The gate reports 0 added / 0 removed / 0 recategorised, 64 allowlisted before and after; no classification logic changes. Closes the last item of the #1180 sizing.

erni-a (approved):

PR #1230, opened 2026-09-01
perf(llm): warm the Vertex genai client at startup, not in the first request

ran-taig:
## Problem

The migrated `VertexLLMProvider` (#1209) builds its google-genai client lazily on first call. On a cold Cloud Run instance that build (SDK import + ADC discovery + HTTP pool) measured **~13s** during the 2026-09-01 staging rehearsal — enough to trip recall's 15s per-attempt cap **once per instance**, surfacing as a sporadic first-request fake-fallback after every autoscaling event. (Warm-instance calls were 0.9–2.3s throughout.)

## Fix

`init_platform_providers()` kicks the new `VertexLLMProvider.warm_up()` onto a **daemon background thread** right after constructing the platform singleton.

Why a thread, not inline and not `await asyncio.to_thread(...)` (review round 2): `init_platform_providers()` runs inside both services' async lifespans, so an inline build would block the event loop for the full build — and an *awaited* offload keeps the ~13s as added readiness delay, i.e. every Cloud Run scale-up cold start gets 13s slower (and the deploy smoke's readiness budget gets 13s tighter). The background thread blocks nothing and delays nothing. The race is safe by construction: a request landing before warm-up completes blocks on the provider's `_client_lock` inside its own `to_thread` worker — the pre-warm-up first-call behavior, at worst once per instance.

Failure-tolerant: no ADC (tests, misconfigured installs) → warning, provider kept, lazy build on first call — exactly the pre-existing behavior. A failed warm-up is deliberately **not** recorded as a `vertex-llm` init error, since the provider remains fully functional. The warm-up call is duck-typed (`getattr`) — providers without a `warm_up` need no warming, and the `LLMProvider` Protocol stays untouched.

## Testing

- Eager path: fake `genai.Client` constructed exactly once by the startup thread (tests join the exposed module-level thread handle for deterministic asserts) and attached to the singleton.
- Failure path: exploding client → provider survives with `_client=None` (lazy retry) and no init error recorded.
- Full affected set green: 133 passed. Ruff clean; mypy clean with CI's invocation.

Pairs with [caura-enterprise#1463](https://github.com/caura-ai/caura-enterprise/pull/1463) (flip staging/prod to `gemini-3.1-flash-lite` @ `us`) — recommended merge order: this first, so the model switch never exposes the cold-instance gap.

### DIFF
--- a/common/llm/_platform.py
+++ b/common/llm/_platform.py
@@ -19,6 +19,7 @@ from __future__ import annotations
 
 import logging
 import os
+import threading
 
 from common.embedding import (
     get_platform_embedding,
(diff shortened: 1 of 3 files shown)


ran-taig:
@claude

ran-taig:
@claude

ran-taig:
@claude

ran-taig:
@erni-a

erni-a (approved):

PR #1231, opened 2026-09-01
ci(lint): bring scripts/ and root tests/ under ruff

Eldad-Caura:
`scripts/` and the root `tests/` appeared in **no ruff scope** in `ci.yml`. With no root `[tool.ruff]` either, `ruff check scripts/` had no ancestor config and ran at ruff's **built-in defaults** — "whatever the installed version ships", a set that expanded from 59 to 413 rules across the 0.16 bump (#684).

It bites in a specific way. With CI not looking, the only gate was the pre-commit hook, which runs **per changed file**. So a ruff bump plants findings in files nobody is touching, and they surface as a blocked commit for whoever edits one next, for reasons unrelated to their change. `caura-enterprise` hit exactly that on its #981 and worked around it with `--no-verify`.

This closes the ruff half of the gap #684 closed for `common/` and #1000 closed for **types** in `scripts/`. 423 files brought under a pinned rule set.

## Decision 1 — line-length and whether to reformat

**The "~100 cols" hypothesis in the task was wrong, and measuring it is what showed that.** A wider limit makes the diff *bigger*:

| line-length | scripts/ would reformat | tests/ would reformat |
|---|---|---|
| **88** | **13 of 31** | **179 of 392** |
| 100 | 30 of 31 | 285 of 392 |
| 110 | 29 of 31 | 297 of 392 |
| 200 | 29 of 31 | 320 of 392 |

These files were written by `ruff format` at its default width. Raising the limit doesn't accommodate the long lines — it instructs the formatter to **join** lines that are correctly wrapped today. 88 is the minimum-churn value, matches `common/ruff.toml` and enterprise's `scripts/ruff.toml`, and both fleet-wide engine files are already clean at it. `E501` stays ignored throughout, so line-length governs `ruff format` only and never fails `ruff check`.

**On reformatting: I did it, because `ruff format` has no per-file suppression** — only `exclude`, and a 179-entry exclude list is worse than the reformat it avoids. The `tenant_scope_gate.py` concern raised in the task turned out to be 152 of 2,111 lines; `hyperagent_test.py`, a throwaway benchmark, was 1,354.

**If you'd rather not take the churn, the alternative is to split this PR**: configs + `ci.yml` alone (3 files), with `[format] exclude` listing the 192 unformatted files and a follow-up to drain it. I did not do that because the exclude list would be larger than the thing it defers. Say the word and I'll reshape it.

## Decision 2 — mypy: already done, and I was wrong on #1224

`scripts/` **has been type-gated since #1000** — `mypy --config-file scripts/mypy.ini scripts/`, with a six-module named exemption ratchet. My note on #1224 said neither changed file was in "any ruff or mypy scope". That was right about ruff and **wrong about mypy for `scripts/`**, and `scripts/mypy.ini` existed at the exact commit I measured. Nothing to add here.

Root `tests/` stays un-type-checked, consistent with every other test tree in both repos.

## A real conflict, found and contained

**`ruff format` and the legacy-name ratchet disagree about marked lines.** The ratchet greps **line by line** — a marker exempts the line it sits on and nothing else. When the formatter wraps a marked statement past 88 columns, the name goes to one line and the trailing comment to another, so **the name lands on an unmarked line**.

Measured, before any exclusions existed:

```
11 exempt line(s) removed by this change.
This change adds the legacy name in 6 file(s).
  scripts/_env_compat.py  (0 -> 1)
  tests/test_events/test_topics.py  (0 -> 4)
  ...
ratchet exit: 1
```

Six files nobody had semantically touched, reported as minting the legacy name. **Re-attaching each comment to its name's line does not converge** — ruff's preferred layout for three of those sites puts the trailing comment on a line that doesn't hold the name at all (the `1,` of a `replace()`, the blank-string argument of a `setenv()`, the right-hand side of a comparison), so every reformat displaces it again. Verified by doing it and re-running both gates.

So **all 16 marker-bearing `.py` files are excluded from `ruff format`, and from that only** — every one is still fully lint-checked. One rule rather than 16 judgements, and it buys two things beyond a green build:

- the pre-commit `ruff-format` hook reads the same list, so it can no longer displace a marker on someone's unrelated commit — a live trap today, since that hook is what formats these files;
- **"never reformat `legacy_name_ratchet.py` or `do_not_touch_sentinel.py`" stops being a rule people must remember and becomes one the config enforces.** Both are byte-identical to `main` in this change.

The fix that shrinks the list: hoist each legacy literal into a short module-level constant carrying the marker. `tests/_legacy_contracts.py` exists for exactly that and already holds four. That's a change to test code, deliberately not bundled here.

## What the rule sets are, and why they differ

`scripts/ruff.toml` carries **`caura-enterprise/scripts/ruff.toml`'s select/ignore lists byte-for-byte**, so a script moving between the repos isn't linted differently. It also mirrors enterprise's `"legacy_name_ratchet.py" = ["PLW0603", "PLW2901", "S101"]` per-file entry, for the same reason and a better one — caura is the canonical home of that engine.

`tests/ruff.toml` carries **`common/ruff.toml`'s** instead. `S`/`BLE`/`EXE`/`FURB`/`PLW` earn their noise in a tree that shells out to gcloud; in test code they're mostly noise (`S101` alone fires on every assert). Its per-file-ignores are the `ARG`/`S101`/`F841` set **every service pyproject already grants its own tests** — this tree is those services' integration tests, and linting it more strictly than `core-storage-api/tests/` would make one fixture idiom a finding in one directory and clean in another.

Neither step uses `--config`. #685 is already documented in this file under the `common/` step: per-file-ignores are relative to the config's directory under discovery and to the CWD under `--config`, so naming it makes every one inert and the step fails on exactly the findings they exempt.

## The backlog

**579 findings fixed mechanically** rather than exempted — `ruff check --fix`: import order (385), the `datetime.UTC` alias (139), unused imports (36), f-strings without placeholders. What's left needs a human and is carried as a **named** backlog:

| tree | files | findings |
|---|---|---|
| scripts/ | 24 | 145 |
| tests/ | 59 | 127 |

Each entry lists its codes and counts. **Named, not globbed**, for the reason `scripts/mypy.ini` already gives in this same tree: its #983 review rejected a module-level `ignore_errors`, and #998 tore out two globs covering 105 modules to gate the 19 that actually failed.

I verified programmatically that **every entry still bites and none is missing** — stripping the backlog block surfaces exactly 24 files/145 findings and 59/127.

Three in there are worth someone's eyes, and are flagged in the config comments rather than buried: **`B023`** loop-variable capture in `benchmark_blend_locomo.py`, **`S608`** string-built SQL in `_f3_wet_test_observer.py`, and **`B017`** blind `pytest.raises(Exception)` in 5 tests, which weakens what those tests assert.

## Verified

- `ruff check` then `ruff format --check`, **run separately**, at all 8 and 7 CI scopes, discovery-based: clean
- mypy clean at all 6 scopes (`core-api` 203, `core-storage-api` 85, `core-operations` 5, `core-worker` 11, `common/` 99, `scripts/` 31 files)
- **full root suite: 5,899 passed, 4 skipped, 1 xfailed, 0 failed** against a migrated database, post-rebase
- ratchet `No new lines.` exit 0; headline **295 lines across 81 files, unmoved**; sentinel `All 46 protected strings survive.`
- `actionlint` reports the same single pre-existing SC2086 before and after
- `uv.lock` untouched (this repo has none)

**Separating verified from taken on trust.** One thing needed care: an earlier full-suite run showed `test_unknown_field_rejection` failing on a `409 DUPLICATE_MEMORY`. That was my own previous run's rows left in the shared local Postgres — with `EMBEDDING_PROVIDER=fake` every memory is a semantic near-duplicate of every other. Rather than assert that, I ran **both this tree and `main` against identically-fresh `pgvector/pgvector:pg16` containers**: same 18 failures, same 5,878 passed, `test_unknown_field_rejection` passing on both. Those 18 are a bare-container setup gap, not this change. The 5,899-pass figure above is from the migrated database.

The rebase hit one conflict, in `tests/test_llm_json_truncation.py`. Resolved by taking `main`'s content and **re-applying the mechanical transformation**, not by hand-merging formatting hunks; its 16 tests pass.

## Caveat worth knowing

A 290-file mechanical change races a fast-moving `main` — it conflicted once during the ~2h this took. If it sits, expect to re-run `ruff check --fix` + `ruff format` on whatever conflicts rather than hand-resolving.

One follow-up for the other repo, not done here: enterprise's `scripts/ruff.toml` justifies its ratchet entry as *"vendored byte-for-byte from caura, whose CI does not apply this enterprise-only scripts rule set"*. This change makes that sentence stale. The entry stays correct in both; only the rationale needs a reword.


arkash20 (dismissed):

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #1234, opened 2026-09-01
feat(gate): a third marker for deferred work that stays counted

Eldad-Caura:
The gate has two markers and both mean *permanent*. Work the rename deliberately
postpones has no way to say so, so it gets filed as `legacy-name-floor` and leaves
the only count that tracks it. That has now happened three times, most recently to
315 Go `import` lines in `caura-daemon` — whose module path rename `docs/rebrand-naming.md:299`
calls "a later decision", not something the rename will never reach.

`legacy-name-deferred` fills that gap. **It annotates; it does not exempt.**

## It stays counted — that is the whole point

A line carrying only this marker stays inside the gated headline and **still fails a
gate it would have failed unmarked**. It is not an escape hatch for a red build, and
the failure text says so in as many words. Adding it to an already-counted line
changes no count at all, so it is free to adopt going forward.

## Why it lives outside `_KINDS` and `EXEMPT_RE`

`_kind()` documents that "the gate's decision reads only whether this is `None`", and
the `_KINDS` comment documents that "the gate never reads this". **Anything added to
either place exempts automatically** — a third entry there would have been a third
permanent marker and would have reproduced the exact bug this closes.

So the marker gets its own constant, its own regex reusing the permanent markers'
two-sided bounded-token construction verbatim, and its own resolver. The scan carries
it as an attribute of a counted line. `_KINDS`, `EXEMPT_RE`, `_kind` and `_reason` are
untouched — every hunk before line 400 is a pure insertion.

## Two semantics, decided

**A decision reference is mandatory.** `legacy-name-deferred: <reason> (<doc path or issue URL>)`.
What separates a deferral from a TODO is that somebody wrote it down, so the marker is
falsifiable by opening the reference, and the reference must resolve to a tracked file
or a URL. A missing one fails the gate — including on a count-neutral line, where
nothing else could fail it. There is no existing population to break, so a hard
failure is free here and never will be again.

**A permanent marker combined with deferred is a hard error.** The alias/floor pair can
both be true of one line, which is why `_kind()` has a measured precedence rule over a
real 28-line population. Permanent and deferred are not both-true states: letting the
permanent marker win silently would exempt the line and make the deferred annotation —
and its place in the headline — a lie. Again, no existing population, so failing the
contradiction costs nothing now.

## The excused transition, and why it cannot excuse anything else

Reclassifying floors to deferred raises a repo's counted total, and the ratchet is
shrink-only. `floor -> deferred` on an otherwise-identical line is therefore an excused
transition. It is deliberately narrow: same path, the base line must have carried a
**floor** marker specifically, the exact prefix *and* syntactic closer must match,
permanent survivors consume base floors first, and the line must still contain the
legacy name. A line with no floor marker in the base cannot use it, so it cannot hide
new work. Excused moves were left alone — they are net-zero cross-file moves, not
sanctioned increases. **No blanket `--allow-increase` flag**: this file is byte-identical
across eight repositories, so a lingering override would be eight lingering overrides.

## Rollout window — read before merging

The `Legacy-name engine parity` org ruleset (id `22008246`, active, **no bypass actors**)
runs a required workflow that fetches
`raw.githubusercontent.com/caura-ai/caura/main/scripts/legacy_name_ratchet.py` and compares
it **byte-for-byte** against each downstream copy, failing closed.

**The moment this merges, all seven downstream repos go red — on every unrelated PR —
until each port lands.** The order is forced: canonical must land first, because the
guard reads canonical to know what to expect.

Expected engine sha256 after this merges:

```
5fa1d5d5acd32dc4e7ecb867e6f333f7b709c357560c4d6f8b8fb4beee364968
```

**All seven port branches are already pushed and verified**, branch
`sync-deferred-legacy-name-marker` in each: `caura-ops`, `caura-onprem-installer`,
`caura-daemon`, `caura-onprem`, `caura-test-automation`, `openclaw-fleet-tester`,
`caura-enterprise` (base `dev`). Every one is byte-identical to the file in this PR.
The ruleset was not touched and no bypass was requested.

Note the ports carry `tests/test_legacy_name_ratchet.py` as well, and must: three of its
existing tests assert the exact shape of the JSON change split and construct the `Scan`
tuple positionally, and the engine adds a field to each. An engine-only port would have
traded a red parity check for a red test suite in all seven.

## Verified

- **179 tests pass** under the real root conftest, the way CI runs them (157 before).
- **The tests fail without the change**: pointing `LEGACY_NAME_RATCHET_SCRIPT` at the
  untouched `origin/main` engine gives **34 failed / 145 passed**. 31 of the 34 are newly
  added; the other 3 are pre-existing tests whose changes are purely additive assertions
  plus the `Scan` arity update — none was weakened.
- Ratchet `No new lines.` (exit 0) and sentinel `All 46 protected strings survive.` (exit 0).
- All six `ruff check` scopes, all five `ruff format --check` scopes and the isolated
  three-file step pass; all six `mypy` steps pass. `ruff format` was not run on
  `legacy_name_ratchet.py` or `do_not_touch_sentinel.py`.
- End-to-end at the CLI, on a scratch repo: a new branded line fails unmarked **and**
  fails carrying only a valid deferred marker; the same line with `legacy-name-ok`
  passes; a deferred marker with no reference fails; deferred combined with a permanent
  marker fails; annotating a pre-existing counted line is count-neutral and passes; and
  `legacy-name-deferredx` is not treated as a marker.
- Per-repo predicted-vs-actual for all seven ports: gate PASS, sentinel PASS,
  **zero added / annotated / removed / moved / net movement**, 179 tests passing.

## Could not establish

- Behaviour under the parity workflow itself is inferred from reading
  `check-legacy-name-engine-parity.sh` and the ruleset definition; it cannot be exercised
  until this merges, because the checker fetches canonical `main` at run time.
- The eventual `caura-daemon` reclassification is a separate PR and is **not** included
  here. This change is the capability; the decision about those 315 specific lines is not
  made by merging it.
- `caura-test-automation`'s suite was run isolated from its own `conftest.py`, which
  imports a package unrelated to the gate that is not installed locally. Every other repo
  ran with its conftest in place.

One behavioural note found while porting: `caura-enterprise` is the one repo whose
pre-commit hooks lint and format the engine rather than excluding it the way `caura-ops`
does. Its `ruff-format` hook rewrote the file on first attempt, which would have silently
broken byte-parity. The canonical file is now format-stable and free of the one new lint
finding, so it lands there unmodified with every hook passing and no bypass.

## The seven port PRs (open, verified, awaiting this merge)

| repo | PR | base |
|---|---|---|
| caura-ops | caura-ai/caura-ops#353 | main |
| caura-onprem-installer | caura-ai/caura-onprem-installer#48 | main |
| caura-daemon | caura-ai/caura-daemon#170 | main |
| caura-onprem | caura-ai/caura-onprem#69 | main |
| caura-test-automation | caura-ai/caura-test-automation#305 | main |
| openclaw-fleet-tester | caura-ai/openclaw-fleet-tester#39 | main |
| caura-enterprise | caura-ai/caura-enterprise#1468 | dev |

Each carries the identical engine and states that its parity check is expected red until
this merges. Their reviews can run during the window; only the parity check waits.

### DIFF
--- a/docs/plans/legacy-name-ratchet-engine.md
+++ b/docs/plans/legacy-name-ratchet-engine.md
@@ -69,7 +69,7 @@ is exactly these five fields:
 | `marker_inventory_meta_paths` | list of root-relative paths | Paths omitted only from marker analytics. |
 
 No sixth field is part of this port. Engine-wide rules, including marker-system
-code/test exclusions and the meaning of the two markers, remain in the engine.
+code/test exclusions and marker semantics, remain in the engine.
 A declared manifest must be a readable JSON object whose keys are valid
 repository-relative paths. Missing or malformed manifests fail closed rather
 than silently moving their mirrors into the fleet-summable headline.
@@ -100,6 +100,43 @@ line in each path whose basename starts with `CHANGELOG` (case-insensitive). It
 does not identify individual generated lines. No path outside that basename
 predicate is exempt.
 
+## Deferred work is an attribute, not an exemption
+
+`legacy-name-deferred: <reason> (<doc path or issue URL>)` records work that a
+decision explicitly postpones. It is deliberately outside the permanent-marker
+table and regular expression. The scan stores it as an attribute of a counted
+line, so a newly written deferred line fails exactly as the same unmarked line
+would. Adding the annotation to an already-counted line is text-neutral for the
+ratchet; it cannot provide headroom or turn a red build green.
+
+The reference is mandatory and is checked in the exact tree being scanned.
+Working-tree document references must be regular repository files that are
+already committed or fully staged; intent-to-add entries, symlinks, and
+submodules do not qualify. HTTPS issue references are syntax-checked without
+making gate availability depend on a remote service. Malformed syntax, a
+missing document, or a local-only document fails the gate.
+
+A line carrying both deferred and a permanent `legacy-name-ok` or
+`legacy-name-floor` marker is a hard error. The permanent marker still wins the
+classification, because `_kind()` remains the only exemption decision, but the
+gate then fails the contradiction instead of silently publishing a deferred
+line outside the headline. This differs from the measured alias-plus-floor
+population: those are two compatible permanent claims, while permanent and
+deferred are mutually exclusive lifecycle states. There is no deferred
+population to preserve, so choosing a hard error now has no migration cost.
+
+The only sanctioned count increase is a same-path `legacy-name-floor` to
+`legacy-name-deferred` replacement. It must preserve the exact text before the
+marker, including comment introducer and internal whitespace, and the exact
+syntactic closer after it. As everywhere else in this engine, outer indentation
+is normalized before text comparison, so reindentation alone is not treated as
+a new branded line. A surviving floor or alias occurrence consumes the base
+floor before transition matching, so one old floor cannot authorize both a
+permanent successor and a deferred duplicate. Each net-new deferred occurrence
+is allocated once: an old counted line in the same path claims it before the
+count-increasing floor transition is considered. The exception is derived from
+the base line; a line that had no base floor marker cannot use it.
+
 ## Report contract
 
 A bare `--report` is current inventory only. It does not resolve a base and
@@ -118,14 +155,23 @@ labels it “never sum,” and discloses any excluded mirror inventory.
 `--report --json` makes the boundary structural:
 
 - `headline.name` is `gated`; it contains numeric `lines` and `files` plus
-  `aggregation: {"allowed": true, "operation": "sum"}`.
+  `aggregation: {"allowed": true, "operation": "sum"}`. Its additive
+  `deferred` member partitions out valid deferred lines by file and decision
+  reference without removing them from `headline.lines`.
 - `diagnostics.present` contains an explicit denied aggregation contract and a
   display string. It deliberately has no numeric `lines` or `files` fields, so
   a generic caller cannot select a `present` number and silently sum it.
 - Mirror details remain auditable under `diagnostics.mirrors`, also tagged with
   denied aggregation.
-- `scope`, marker inventory, and an optional explicit-base change object carry
-  the remaining bounded diagnostics.
+- `marker_inventory.counts` includes the deferred token separately, including a
+  recognized marker whose syntax or reference is invalid.
+- `deferred_validation` exposes whether every recognized deferred marker is
+  valid and lists bounded path/line diagnostics when one is not.
+- `scope` and an optional explicit-base change object carry the remaining
+  bounded diagnostics. When present, `change.deferred` distinguishes
+  count-neutral annotations from narrowly excused floor-to-deferred increases.
+  These additions are backward-compatible fields in the existing
+  `legacy-name-ratchet-report/v1` object; existing field meanings do not change.
 
 ## Gate invariants and tests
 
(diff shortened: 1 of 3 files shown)


Eldad-Caura:
@claude the review job skipped silently on the force-push (`synchronize`), so this is a re-trigger rather than a second opinion on the same bytes.

Two things changed since the branch first went up, both worth your attention:

1. The engine is now format-stable and free of the `SIM108` finding. That is not cosmetic: `caura-enterprise` is the one repo whose pre-commit hooks lint and format `scripts/legacy_name_ratchet.py` rather than excluding it the way `caura-ops` does, and its `ruff-format` hook rewrote the file on the first attempt — which would have silently broken the byte-parity the org ruleset enforces.
2. `_ANNOTATION_SUFFIX` and `_permanent_annotation` now accept `--!>` as well as `-->` (CodeQL `py/bad-tag-filter`, since resolved).

The thing most worth checking is the claim in the title: that `legacy-name-deferred` **annotates and does not exempt**. It deliberately lives outside `_KINDS` and `EXEMPT_RE`, because `_kind()` is the sole exemption decision and anything in either place exempts automatically. Please check that separation actually holds on every path, and that `_floor_transitions` cannot excuse a count increase for any line that did not carry a `legacy-name-floor` marker in the base tree.

Eldad-Caura:
@claude both findings were real. I verified each against the running gate before touching anything, and both are fixed. Re-trigger, since the force-push skipped the job silently.

**Medium — invalid marker hid a real mint.** Confirmed: a change carrying both a minted unmarked line and an invalid deferred marker reported only the marker, exit 1, with the minted file absent from stdout entirely. Fixed, but not quite as prescribed, because moving the check straight to the end introduced two problems the prescription did not anticipate:

1. **It would have made the gate pass.** With the early return gone, a *pre-existing* invalid marker — count-neutral, so nothing is minted — falls through to `if not grown:` and returns **0** while still printing the marker report. That is precisely the failure this PR exists to prevent: a marker that quietly turns a build green. There is now an explicit guard, and I verified it is load-bearing by deleting it again and watching the gate exit 0. The existing suite also catches it — 13 tests fail without it.
2. **Placement had to be before the empty-tree check, not after.** A line carrying a permanent marker *and* a deferred one is exempt, so it is absent from both counted sets. In a tree whose only branded line is that one, both sets are empty and `if not head and not base` answered a contradictory marker with a note about a broken pathspec. Caught by `test_deferred_cannot_be_combined_with_a_permanent_marker` failing after my first attempt.

So the report now prints before that check and never returns early, and no exit path can reach 0 while a problem stands.

**Low — case-sensitive scheme check.** Confirmed, and the fix is broader than the finding. `_decision_reference_error` now branches on whether `urlsplit` found *any* scheme rather than on the literal `https://`, which also fixes lowercase **`http://`** — same misleading "must be normalized and repository-relative" error, not mentioned in the review, and rather more likely to be typed than an uppercase scheme. `HTTPS://` is now accepted rather than merely better-diagnosed, since it is a valid absolute HTTPS URL; `urlsplit` normalises scheme case.

**Suggestion declined**, agreeing with your own reasoning: `_new_deferred`, `_deferred_annotations` and `_floor_transitions` are still computed unconditionally. They are now genuinely needed on every path that reports, and the file states its tolerance for the cost.

**One thing you did not flag, added anyway.** The engine's formatter-stability is now load-bearing for the whole eight-repo byte-parity guarantee — one repo's pre-commit hooks format this file rather than excluding it, so a formatter-unstable construct gets silently rewritten there and parity breaks at commit time. Nothing asserted it. `test_the_engine_stays_formatter_stable` now does, read-only and at the formatter's defaults.

Tests 179 → 187. All eight new ones fail against the pre-fix engine except two honest controls: lowercase `https` always worked, and the formatter-stability test guards a property rather than proving a new one.

Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #1235, opened 2026-09-01
feat(storage): accept a binding tenant_id on the usage query (expand step)

### LINKED ISSUE #1095: storage: POST /tenant-usage/query totals any tenants the caller names

Eldad-Caura:
**PR 1 of 3** for #1095, branch A. **This does not fix #1095** — see "What is not fixed" below before reading it as closure.

## What

Adds an optional singular `tenant_id` to `POST /tenant-usage/query` beside the existing `tenant_ids`, with **exactly one** required. Purely additive: a caller sending the plural sees no change.

```python
tenant_id: str | None = None
tenant_ids: list[str] | None = Field(default=None, min_length=1, max_length=MAX_TENANTS_PER_QUERY)
```

Neither is rejected (a caller that named no scope must never be read as "every tenant"); both is rejected as ambiguous rather than silently preferring one, which would let a caller believe the wider list was honoured.

## What is NOT fixed

The plural still lets a caller name its own scope. The endpoint is still unbound, and **the gate confirms it** — I ran it rather than predicting the movement:

```
  added (0)  removed — fixed (0)  removed — gone (0)  removed — still unscoped (0)  recategorised (0)
tenant-scope gate: 189 routes + 186 service methods, 311 bound to a tenant, 64 allowlisted.
      grant-in-lieu-of-tenant: 2 -> 2
```

Both allowlist entries stay. **The security fix lands in PR 3**, when the plural is deleted.

## Why three PRs and not two

`platform-admin-api` and `core-storage-api` are separately deployed, and **platform goes first** — verified in `caura-enterprise` at `origin/dev`, `.github/workflows/deploy-to-staging.yml`:

```yaml
  oss:
    needs: [guard-branch, platform]
```

So a migrated caller reaches the *old* storage before the new storage exists. Ship the singular field and its consumer together and the usage dashboard 422s in production for the length of that window — and `get_org_usage(strict=True)` gates a read-only lock, so a 422 there is not cosmetic.

Hence expand / migrate / contract:

| | repo | what |
|---|---|---|
| 1 | `caura` | **this PR** — storage accepts the singular field |
| 2 | `caura-enterprise` | `core_storage_client.query_tenant_usage` loops on the singular field and merges |
| 3 | `caura` | delete `tenant_ids` + `MAX_TENANTS_PER_QUERY`, delete both allowlist entries |

**PR 3 must not merge until PR 2 is *deployed*, not merely merged.**

## Shape

The singular path returns the **same** response shape — a one-element case of the same aggregate. Deliberate: a different shape would make the migrate step a rewrite instead of a loop plus a sum. The caller already holds the tenant list five lines earlier (`usage_service.py:180` resolves `org_id → tenant_ids`, `:185` makes this call), so the loop goes where the list already is.

## The fail-closed guard

`tenants is None` is unreachable while the validator stands, and it is deliberately **not** an `assert` and not a `cast`:

```python
if tenants is None:
    raise HTTPException(status_code=422, detail="a tenant scope is required")
```

This is the one endpoint where an unscoped query must never widen to every tenant, so the guard fails closed on its own rather than trusting a validator a later edit could drop. It also narrows the type for mypy without a cast — mypy flagged the `list[str] | None` before this was added, and papering that over with `cast` would have removed the signal without removing the risk.

## Tests — confirmed failing without the fix

New `core-storage-api/tests/test_tenant_usage_binding_scope.py`, 7 tests. **Reverted the router to `origin/main` with the tests in place: 5 failed, 2 passed.**

The 2 that pass are `test_plural_still_works_unchanged` and `test_neither_scope_is_422` — correct, since both held before this change. They are regression guards, not guards on the new field.

Coverage: singular accepted · singular and plural agree for one tenant · plural unchanged · neither is 422 · both is 422 · **singular scope is binding** (a second tenant's counters cannot reach the response) · `period_start` still applies on the singular path.

## Gates

Venv on `PATH`, `UV_FROZEN=1`. `ruff check` and `ruff format --check` run separately at CI's exact scopes, discovery-based, no `--config`.

- `python scripts/tenant_scope_gate.py --base origin/main` — output above, zero movement
- `pytest core-storage-api/tests/` — **322 passed** (315 + 7 new)
- `ruff check` — all seven CI scopes clean
- `ruff format --check` — all scopes formatted

One signed-off commit, rebased onto `origin/main` immediately before push with the gate and suite re-run after.

## No migration

Branch A needs none, and if this work starts wanting one that is branch B, which is not what was picked. Nothing here touches the schema.

### DIFF
--- a/core-storage-api/src/core_storage_api/routers/tenant_usage.py
+++ b/core-storage-api/src/core_storage_api/routers/tenant_usage.py
@@ -31,7 +31,7 @@ from __future__ import annotations
 from datetime import datetime
 
 from fastapi import APIRouter, HTTPException, Request
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, Field, model_validator
 
 from core_storage_api.services.postgres_service import PostgresService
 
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@erni-a ready for approval.

Automated review came back clean (no issues found) and every check is green — CI, DCO, CodeQL, security scan, legacy-name ratchet, do-not-touch sentinel.

One caveat worth having in front of you before you approve, because it is unusual: **this PR deliberately fixes nothing.** It is the expand step of a three-step expand / migrate / contract for #1095, and the gate confirms zero allowlist movement (64 allowlisted, `grant-in-lieu-of-tenant` 2 → 2). The endpoint is still unbound; the defect closes in PR 3, which cannot merge until the `caura-enterprise` migrate step is *deployed*, not merely merged.

Additive and backward-compatible: a caller sending the existing `tenant_ids` sees no change. 322 tests pass, 5 of the 7 new ones confirmed failing against the pre-change router.

erni-a (approved):

PR #1236, opened 2026-09-01
fix(storage): require a binding tenant on the usage query (contract step)

### LINKED ISSUE #1095: storage: POST /tenant-usage/query totals any tenants the caller names

Eldad-Caura:
**PR 3 of 3 — DRAFT. Do not merge until caura-ai/caura-enterprise#1473 is DEPLOYED, not merely merged.**

`deploy-to-staging.yml` runs the `oss` job as `needs: [guard-branch, platform]`, so platform deploys first. Landing this before the migrate step is live 422s the usage dashboard — and `get_org_usage(strict=True)` gates a read-only lock, so that is not cosmetic. Draft until the running artifact is verified, not until a pipeline event is green.

## What

Closes #1095. `tenant_ids` (plural) is gone from `POST /tenant-usage/query`, along with `MAX_TENANTS_PER_QUERY`. The service method takes a binding singular `tenant_id` and filters on equality instead of `IN`.

The plural let a caller hand this service the list of tenants to total. Since #1066 the only credential here is a shared secret carrying no tenant identity, so nothing could check that list against who sent it — the caller named its own scope. The fan-out and summing now live in `platform-admin-api` (#1473), which resolves `org_id → tenant_ids` against `platform-storage-api` and is the only party that can say which tenants an org owns.

## Gate movement — measured, not predicted

```
  removed — fixed (1):
      - method:tenant_usage_query [grant-in-lieu-of-tenant]
  recategorised (1):
      ~ route:POST /api/v1/storage/tenant-usage/query [grant-in-lieu-of-tenant -> blind-spot]
  grant-in-lieu-of-tenant: 2 -> 0        blind-spot: 0 -> 1
tenant-scope gate: 189 routes + 186 service methods, 312 bound to a tenant, 63 allowlisted.
```

**The category #1095 was filed under is now empty.**

**Only one entry deletes, not two.** I had expected both. The route genuinely requires a binding `tenant_id` now — but the guard is a **pydantic field** on `TenantUsageQuery`, and the AST pass reads dict-style access (`body[...]`, `body.get(...)`, `_require(body, ...)`), not model attribute access. It cannot see a binding that is really there, which is the `blind-spot` definition exactly, so the entry is recategorised rather than deleted and its note says why.

**I did not hand-parse the body to make the gate see it.** This router chose pydantic deliberately — its own docstring notes the hand-parsed sibling below is where a missing key became a 500 in review. Trading real validation for static visibility would be the same mistake as casting away a mypy error: it removes the signal without removing the risk. The durable fix is teaching the gate to read a required pydantic field, and this is the **only** allowlisted route shaped that way today, which is why it has not been worth doing yet.

## Two gate tests updated, because the fix invalidated them

Both keep the rule and drop the now-fixed instance:

- **`test_a_plural_tenant_list_is_not_a_binding_scope`** pinned the live offender by id (`entries["tenant_usage_query"].verdict == "NONE"`). It now asserts the rule (`tenant_ids` stays out of `BINDING_SCOPE`), that the former offender scores `REQUIRED`, and that the category stays **defined** so a future offender has somewhere to land. **Confirmed load-bearing**: adding `tenant_ids` to `BINDING_SCOPE` turns it red. (My first attempt at this replacement was vacuous — it looped over `entry.scope_params`, an attribute `Entry` does not have, so it asserted nothing. Caught before commit.)
- **`test_every_owned_entry_carries_a_tracked_issue`** asserted the owned backlog was non-empty. That became a **failure the moment the backlog was emptied** — `id-addressed-write` has been empty since #1082 and this clears the last `grant-in-lieu-of-tenant`. An empty owned backlog is the goal state, so the floor is gone; the property that matters ("whatever is owned carries a reference") is vacuously true at zero and still bites at one.

## Tests

`test_counts_sum_across_the_orgs_tenants` is **deliberately deleted** — summing an org's tenants was the capability the plural existed for, and it moved to `platform-admin-api`, where `test_totals_are_summed_across_tenants` replaces it. A comment at the old site says so, so the next reader does not read it as a dropped check.

`test_an_empty_tenant_list_is_rejected` becomes `test_a_missing_tenant_id_is_rejected` — the shape of "named no scope" changed from `[]` to an absent field, but the requirement is the same one. Added `test_an_empty_tenant_id_is_rejected` (`min_length=1`) and **`test_the_plural_field_is_gone`**, which pins the actual defect: sending only `tenant_ids` is now an unscoped request and is refused rather than silently honoured.

- `pytest core-storage-api/tests/` — **320 passed**
- `pytest tests/test_tenant_scope_gate.py` — **122 passed**

## Gates

- `python scripts/tenant_scope_gate.py --write` — regenerated, not hand-edited; re-running after the note edit is byte-identical (`shasum` matched)
- `ruff check` / `ruff format --check` separately at CI's exact scopes — **including the new `scripts/` and root `tests/` scopes added by #1231** — all clean
- `legacy_name_ratchet.py --base origin/main` — `No new lines.`
- `do_not_touch_sentinel.py` — all 46 protected strings survive

One signed-off commit, rebased onto `origin/main` immediately before push with the gate and both suites re-run after.

### DIFF
--- a/core-storage-api/src/core_storage_api/routers/tenant_usage.py
+++ b/core-storage-api/src/core_storage_api/routers/tenant_usage.py
@@ -31,75 +31,44 @@ from __future__ import annotations
 from datetime import datetime
 
 from fastapi import APIRouter, HTTPException, Request
-from pydantic import BaseModel, Field, model_validator
+from pydantic import BaseModel, Field
 
 from core_storage_api.services.postgres_service import PostgresService
 
 router = APIRouter(prefix="/tenant-usage", tags=["TenantUsage"])
 _svc = PostgresService()
 
-#: A list longer than this is a caller sending its whole tenant table by
-#: accident, not a large org. It does NOT bound the query's cost — that scales
-#: with how much history the named tenants have, which nothing here caps.
-MAX_TENANTS_PER_QUERY = 1000
-
 
 class TenantUsageQuery(BaseModel):
-    """Which tenant (or tenants), and which periods, to total.
+    """Which tenant, and which periods, to total.
 
-    A POST for a read: the tenant list is the org's whole tenant set, which has
-    no fixed bound, and a query string does. Validated by pydantic rather than
-    hand-parsed — the write endpoint below hand-parses because it coerces
-    per-row datetimes, and that is exactly where a missing key turned into a
-    500 in review.
+    A POST for a read rather than a GET: kept from the plural era because the
+    body already carries optional period parameters and the sibling endpoint
+    below is a POST, not because the scope needs the room.
 
-    EXPAND STEP OF #1095 (expand / migrate / contract). ``tenant_id`` is the
-    binding singular scope and is the one to use; ``tenant_ids`` is the legacy
-    plural that lets a caller name its own scope, which is the defect #1095
-    exists to close. Both are accepted here, exactly one is required, and
-    NOTHING IS FIXED YET — the plural path is still unbound until the contract
-    step deletes it. Do not read this step as closing #1095.
-
-    The two cannot be collapsed in one release because ``platform-admin-api``
-    and ``core-storage-api`` deploy separately, platform FIRST
-    (``deploy-to-staging.yml``: the ``oss`` job is ``needs: [guard-branch,
-    platform]``). So the new caller reaches the old storage before the new
-    storage exists, and storage has to accept the singular field before
-    anything sends it.
+    CONTRACT STEP OF #1095. ``tenant_ids`` (plural) is GONE. It let a caller
+    hand this service the list of tenants to total, which is the caller naming
+    its own scope -- and since #1066 the only credential here is a shared
+    secret carrying no tenant identity, so nothing could check the list against
+    who sent it. The org-level fan-out and summing now live in
+    ``platform-admin-api``, which resolves ``org_id -> tenant_ids`` against
+    ``platform-storage-api`` and is the only party that can say which tenants
+    an org owns. Do not reintroduce a plural field here; the check it would
+    need cannot be written on this side of the boundary.
     """
 
-    tenant_id: str | None = None
-    tenant_ids: list[str] | None = Field(default=None, min_length=1, max_length=MAX_TENANTS_PER_QUERY)
+    tenant_id: str = Field(min_length=1)
     period_start: datetime | None = None
     periods: int = Field(default=6, ge=1, le=24)
-
-    @model_validator(mode="after")
-    def _exactly_one_scope(self) -> TenantUsageQuery:
-        """Exactly one of ``tenant_id`` / ``tenant_ids``.
-
-        Neither is a caller that forgot its scope, and defaulting to "all
-        tenants" is the one answer this endpoint must never give. Both at once
-        is ambiguous rather than harmless: silently preferring the singular
-        would let a caller believe the wider list was honoured.
-        """
-        if (self.tenant_id is None) == (self.tenant_ids is None):
-            raise ValueError("exactly one of 'tenant_id' or 'tenant_ids' is required")
-        return self
 
 
 @router.post("/query")
 async def query_tenant_usage(body: TenantUsageQuery) -> dict:
-    """Total the counters for a set of tenants, per period, per operation.
+    """Total one tenant's counters, per period, per operation.
 
-    Body: ``{tenant_id | tenant_ids, period_start?, periods?}`` →
+    Body: ``{tenant_id, period_start?, periods?}`` →
     ``{"periods": [{"period_start": iso, "operations": {op: total}}, ...]}``,
     newest first.
-
-    The response shape is identical either way: a singular request is the
-    one-element case of the same aggregate, so the migrating caller sums N
-    responses instead of reading one. That is deliberate — a different shape
-    for the singular path would make the migrate step a rewrite rather than a
-    loop.
 
     The operation names are passed through as stored rather than mapped onto a
     fixed writes/searches/recalls triple. core-api also meters ``insights`` and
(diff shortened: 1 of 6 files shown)


Eldad-Caura:
## Deploy precondition met — taken out of draft

This is the **contract** step of #1095 and it must not merge until the migrate step is *deployed*, not merely merged. That has now been verified at the artifact level rather than from a pipeline event, because on this occasion the pipeline event would have been misleading.

**The chain:**

| | |
|---|---|
| caura-enterprise#1473 merged | `3a1d6ca8`, 2026-09-02 00:02:50Z |
| deployed image | `platform-admin-api:9685d0e79f0f3d1f49116fb2fcbf63336faf7bec` |
| service | `staging-memclaw-platform-admin` |
| revision | `staging-memclaw-platform-admin-01360-c4d` |
| status | **deployed and serving 100 percent of traffic**, 00:27:28Z |

`9685d0e7` is caura-enterprise `dev` tip; it descends from `3a1d6ca8` and its tree contains the per-tenant fan-out. Images there are tagged `${{ github.sha }}`, so the tag identifies the exact source commit.

**Why the pipeline event alone would have been wrong here.** #1473's own staging deploy was **cancelled**, and so were the two after it (`031b8cb4`, `3a1d6ca8`, `f1fbd904`). `deploy-to-staging.yml` uses `concurrency: {group: staging-deploy, cancel-in-progress: false}`, which queues rather than cancelling in-flight runs but supersedes *queued* ones — so a long-running deploy held the lock while each new push to dev evicted its predecessor. For roughly 25 minutes after #1473 merged, the last **successful** staging deploy was `caf680e9`, which predates the merge: the change was merged, dev was green, and nothing was red anywhere, while staging was still running the old admin-api.

Checking a green pipeline would have reported this as deployed about half an hour before it was.

**What this PR now does safely:** removes the plural `tenant_ids` from `POST /tenant-usage/query`, leaving the binding singular `tenant_id`. The deployed admin-api already sends the singular field per tenant, so removing the plural breaks nothing. Doing it in the other order would have broken the usage panel and `_is_over_plan_limits` — and that gate fails in the direction that *grants*.

Eldad-Caura:
@erni-a ready for review — this is the **contract** step of #1095, and the one with a real ordering hazard, so the deploy precondition is worth checking before the diff.

**Precondition (verified, details in the comment above).** caura-enterprise#1473 is not merely merged but deployed: revision `staging-memclaw-platform-admin-01360-c4d` is serving 100% of traffic from image `platform-admin-api:9685d0e79f0f…`, a commit whose tree contains the per-tenant fan-out. That matters because #1473's own staging deploy was **cancelled**, as were the two after it — nothing went red, and for ~25 minutes after the merge the last successful deploy still predated it. A green pipeline would have said "deployed" about half an hour early.

**What this removes.** `tenant_ids` (plural) from `POST /tenant-usage/query`, leaving the binding singular `tenant_id`. The plural let a caller name its own scope, which is the defect: since #1066 the only credential at that seam is a shared secret carrying no tenant identity, so nothing could check the submitted list against who sent it. Deployed admin-api already sends the singular field per tenant, so nothing breaks. The reverse order would have broken the usage panel and `_is_over_plan_limits` — and that gate fails in the direction that *grants*.

**Two things in the gate report worth a look, since both are easy to misread:**

1. The allowlist goes **64 → 63**, not 62. I predicted two entries would drop and only one does — the route now genuinely requires a binding `tenant_id`, but the guard is a **pydantic field**, and the AST pass reads dict-style access rather than model attributes. That is the `blind-spot` definition exactly, so the entry is recategorised with a note rather than deleted.

2. I deliberately did **not** hand-parse the body to make the gate see the guard. This router chose pydantic on purpose; trading real validation for static visibility would be the same mistake as casting away a type error to silence it.

`grant-in-lieu-of-tenant` goes **2 → 0** — the category #1095 was filed under is now empty.

Two gate tests changed because the fix *invalidated* them, not to make them pass: one asserted the owned backlog was non-empty, which became false the moment the backlog was emptied, and emptying it is the goal.

All checks green, and the review landed clean. Not merging it myself.

erni-a (approved):

Eldad-Caura:
## Production precondition also met — safe to merge

The earlier comment verified staging. The migrate step is now in **production** as well, so this can merge without the shared-promotion window described below.

| | |
|---|---|
| promoted SHA | `3ef3beeb7fc6179ad16c2ee720251b904bbc3a4d` (all 13 staging services aligned) |
| contains caura-enterprise#1473 | yes — descends from `3a1d6ca8` |
| service | `prod-memclaw-platform-admin` |
| revision | `prod-memclaw-platform-admin-00131-dnl` |
| status | **deployed and serving 100 percent of traffic** |
| image | pinned by digest, `platform-admin-api@sha256:f60b40c2…` |

Confirmed at the level that matters rather than by the run's exit status: the promoted commit's tree contains the per-tenant fan-out and sends the singular `"tenant_id"` field. The whole promote run succeeded, including the prod min-instances and OSS containerConcurrency invariant assertions.

**Why this ordering was worth waiting for.** `promote-to-prod.yml` runs `deploy-storage` *before* `deploy-rest` (`deploy-rest` declares `needs: [detect-sha, deploy-storage, …]`). Had this contract step and the migrate step promoted in the **same** run, prod storage would have dropped the plural field minutes before prod admin-api gained the fan-out — each job allows up to 15 minutes. That window degrades rather than grants (the 422 surfaces as an `httpx` error, so non-strict callers see an empty usage panel and `strict` callers get `UsageUnavailable`, which fails closed), but it would have been a visible blank usage panel for customers.

Promoting #1473 first means the two never share a promotion, so the window cannot occur. That is the entire point of doing this as expand → migrate → contract rather than one change.

Merging this now removes a field that no deployed caller sends, in either environment.

PR #1237, opened 2026-09-01
fix(contradiction): stop candidate queries reaching into another agent's private memories (A54, D3)

arkash20:
## A54 — a writer could mark **another agent's private memory** outdated/conflicted

The task filed one query: the RDF candidate lookup had no visibility predicate, while its sibling `memory_find_similar_candidates` had one.

**Fixing that alone did not close the hole.** The wet test reproduced the leak anyway — through the *semantic* path — because:

> `visibility == "scope_agent"` means **"private to SOME agent"**, not "private to THIS agent".

Filtering on the tier still matches every *other* agent's private rows. So the real defect is broader than filed, and all three contradiction candidate routes had it:

| query | before | now |
|---|---|---|
| `memory_find_rdf_conflicts` | no visibility filter at all | tier + owner |
| `memory_find_similar_candidates` | tier only — **the live leak** | tier + owner |
| `memory_find_entity_overlap_candidates` | tier only | tier + owner |

The owner predicate applies **only** to `scope_agent`, so `scope_team` / `scope_org` behaviour is unchanged. Both new storage params are optional, so a storage instance deployed ahead of core-api keeps serving; core-api always sends them.

### Why this matters

A contradiction verdict is a **write**. Selecting another agent's private row as a candidate meant marking it `outdated`/`conflicted` — a status write into a row the writer cannot read, plus an inference channel about its contents.

## D3 — detection was invisible unless it found something

The completion log lived inside `if contradictions:`, so "ran and found nothing" and "never ran" (lock contention, early return, crashed worker) were indistinguishable — precisely when you most need to know, i.e. a contradiction was expected and none appeared. Every concluded run now logs, with structured `conflicts_found` / `memory_id`.

## Testing

9 unit tests, including one that asserts **all three** queries pin the owner — the hole moved once already, and that test is what stops it moving again. 297 contradiction tests pass. Full suite **5912 passed** (the 8 failures in `test_capability_usage` / `test_request_observation` pre-exist on main; verified by stashing).

**Wet-tested on a live stack, both directions:**

```
BEFORE the fix
  agent B writes a contradicting fact -> agent A's private row becomes `conflicted`   <-- leak

AFTER the fix
  2) agent B private write (contradicts A) -> A: active      sup=None    [untouched]
  3) agent A contradicts ITSELF            -> A: conflicted
                                              C: active      sup=<A.id>  [positive control]

  D3, observed on a zero-conflict run:
  "Async contradiction detection completed for memory ...: 0 conflict(s)"
     conflicts_found=0 memory_id=... path=contradiction-detection
```

The positive control matters as much as the negative one: scoping too tightly would silently disable contradiction detection, which no unit test would catch.

## Not in this PR

`memory_find_semantic_duplicate` has the same tier-only filter. It's the **dedup** path, not contradiction — it could return a 409 naming another agent's private memory id (existence disclosure). Filed separately rather than widening this change.

Refs: canonical A54, D3.

### DIFF
--- a/core-api/src/core_api/clients/storage_client.py
+++ b/core-api/src/core_api/clients/storage_client.py
@@ -825,6 +825,8 @@ class CoreStorageClient:
         exclude_id: str | None = None,
         fleet_id: str | None = None,
         object_value: str | None = None,
+        visibility: str | None = None,
+        agent_id: str | None = None,
     ) -> list[dict]:
         params: dict[str, Any] = {
             "tenant_id": tenant_id,
@@ -837,6 +839,11 @@ class CoreStorageClient:
             params["fleet_id"] = fleet_id
         if object_value is not None:
             params["object_value"] = object_value
+        # A54 — scope RDF candidates the same way the semantic path is scoped.
+        if visibility is not None:
+            params["visibility"] = visibility
+        if agent_id is not None:
+            params["agent_id"] = agent_id
         return await self._get_list("/memories/rdf-conflicts", **params)
 
     async def scored_search(self, data: dict) -> list[dict]:
--- a/core-api/src/core_api/clients/storage_client.py
+++ b/core-api/src/core_api/clients/storage_client.py
@@ -808,6 +808,17 @@ class CoreStorageClient:
         # is picked up by a later crystallizer pass.
         return await self._post("/memories/entity-overlap-candidates", data, read=True)  # type: ignore[return-value]
 
+    async def find_by_supersedes_id(self, tenant_id: str, supersedes_id: str) -> list[dict]:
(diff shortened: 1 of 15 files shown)


arkash20:
@Eldad-Caura please approve

Eldad-Caura (dismissed):

arkash20:
Rebased onto main (was BEHIND) and fixed the CI failure: `ruff format --check core-api/src/` wanted the new retraction lookup call on one line. Also now carries A53 via the merged #1239.

CI green (13 checks).

@Eldad-Caura please approve

Eldad-Caura (approved):

PR #1238, opened 2026-09-01
fix(gate): stop a floor-to-deferred transition charging an unrelated move

Eldad-Caura:
> **DO NOT MERGE UNTIL THE SEVEN PORTS OF caura#1234 HAVE LANDED.** This changes
> `scripts/legacy_name_ratchet.py`, which the `Legacy-name engine parity` org ruleset
> compares byte-for-byte against `caura/main`. Those seven PRs are green **right now**
> because they match `main`. Merging this first turns all seven red again and re-opens
> the estate-wide window they exist to close. Land them, then this, then re-port.
>
> The seven: caura-ai/caura-ops#353 · caura-ai/caura-onprem-installer#48 ·
> caura-ai/caura-daemon#170 · caura-ai/caura-onprem#69 ·
> caura-ai/caura-test-automation#305 · caura-ai/openclaw-fleet-tester#39 ·
> caura-ai/caura-enterprise#1468

Follows caura#1234. A reviewer on two of the port PRs found this, it is real, and the
fix that shipped in #1234 only worked for half the inputs.

## What is wrong

A floor-to-deferred transition raises the repo-wide count of its text — the base
occurrence carried a floor marker and was uncounted, the deferred one is counted. That
gain is sanctioned by the transition, and #1234 correctly kept it out of the
transitioned file's own tally.

It did **not** keep it out of the shared mint budget. That budget is spent across files
in sorted path order, so the sanctioned gain stayed available, and the next file to add
the same text had it charged as **minted** rather than recognised as a **move** — a red
build on a change that minted nothing, in a file with no part in the transition.

## Why #1234's version of the fix was not enough

#1234 spent the budget as each transitioned file was reached. That works only when the
transitioned file sorts **before** the file that would otherwise claim the text. The
regression test written for it used `a.py` for the transition and `b.py` for the
destination — the lucky order — so it passed while the bug remained for every pair that
sorted the other way.

The budget is now made transition-neutral **once, before any file is visited**, so path
order cannot decide the verdict. That is the shape the reviewer proposed, and it is the
right one.

## Verified

- **The regression test runs both orderings.** Against the engine currently on `main`,
  `transition-sorts-last` **fails** and `transition-sorts-first` **passes** — the exact
  distinction a single-ordering test cannot draw. Both pass here.
- 191 tests pass (190 before this branch; the added case is the second ordering).
- Ratchet `No new lines.` (exit 0); sentinel `All 46 protected strings survive.` (exit 0).
- All eight `ruff check` scopes, all seven `ruff format --check` scopes including the two
  `scripts/` and `tests/` steps added by #1231, the isolated three-file step, and
  `mypy scripts/` all pass. `ruff format` was not run on either protected script.
- The non-exemption property is unaffected and re-proven end-to-end: a line carrying only
  `legacy-name-deferred` is still counted and still fails; `legacy-name-ok` on the same
  line still passes; a missing reference still fails; deferred combined with a permanent
  marker still fails.

## Bearing on the caura-daemon reclassification

That change turns 315 floor lines into deferred ones in a single PR. Every one of them is
a sanctioned gain, so it would have contributed 315 units of spurious budget under the
merged engine. Any same-text move anywhere in that PR could then have been charged as a
mint. This wants to be in the engine before that reclassification is attempted.

### DIFF
--- a/scripts/legacy_name_ratchet.py
+++ b/scripts/legacy_name_ratchet.py
@@ -1683,17 +1683,11 @@ def _minted(
     """
     new_here = head_by_file.get(path, Counter()) - base_by_file.get(path, Counter())
     if transition:
-        # An excused occurrence still SPENDS its unit of the shared budget. The
-        # repo really did gain that text — the transition is why it is not
-        # charged here, not evidence that the gain never happened. Dropping it
-        # from ``new_here`` without spending would leave the unit available to
-        # an unrelated file, where the next occurrence of the same text is
-        # charged as minted instead of recognised as a move: a red build on a
-        # change that minted nothing, in a file with no part in the transition.
-        for text, excused in transition.items():
-            spent = min(excused, new_here[text], budget[text])
-            if spent:
-                budget[text] -= spent
+        # Only the local tally is adjusted here. The excused occurrence is kept
+        # out of the shared budget by :func:`_mint_budget`'s caller instead,
+        # before any file is visited — spending it here would work only when the
+        # transitioned file happens to sort before the file that would otherwise
+        # claim the same text, which is a coin toss on filenames.
         new_here.subtract(transition)
         new_here = +new_here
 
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #1240, opened 2026-09-02
test(contradiction): verify A16/A38/A39, and document the N>2 defect A37 found

arkash20:
Four rows filed as **"unverified"** — verified empirically against a live stack (real embeddings, real judge), not by reading the code and asserting a guess. Three came back clean; one found a defect.

## A16 — verified fixed

The office-lunch case **does not reproduce**. Complementary attributes of one subject stay `active`:

```
"Priya eats lunch at the office cafeteria on weekdays"  -> active
"Priya cooks dinner at home most evenings"              -> active
```

The judge's `non_conflict_reason` taxonomy already covers it (`list_valued_predicate` shape (b)). Pinned **with its counterweight**: loosening far enough to spare complementary facts must not spare genuine state changes, or nothing is ever retired. Those two pull in opposite directions, so both are asserted.

## A38 — verified, no bypass, and it's structural

Detection is scheduled inside the **write pipeline**, and both surfaces reach it through the same `create_memory`:

```
REST  /api/v1/memories -> write_memory -> _write_memory_inner -> create_memory
MCP   caura_write ------------------------------------------> create_memory
                                                                    |
                                              _create_memory_pipeline
                                                                    |
                                             ScheduleBackgroundTasks   <- detection
```

The tests fail if detection is ever hoisted into the REST route — which is exactly how every non-REST caller would silently lose it.

## A39 — verified correct

Conflicting values for one subject written to two tenants leave **both rows `active`** — no cross-tenant contradiction. A foreign read by id returns 404. Tenant scoping is pinned on all three candidate queries and on the status-write path.

## A37 — defect found

The row asked for coverage; the coverage found a problem. Three mutually exclusive values for one subject, ~55 s apart:

| run | subject | A (oldest) | B | C (newest) |
|---|---|---|---|---|
| 1 | 01:00 → 03:00 → 05:00 | `conflicted` | `conflicted` (sup=A) | `active` (sup=B) |
| 2 | Marta → Tomas → Wei | `active` | `active` | `active` |
| 3 | 4 → 8 → 16 workers | `conflicted` | `active` (sup=A) | `active` |

Run 1 is correct. Run 3 leaves two different values both `active`. **Run 2 fired nothing at all** — three contradictory oncall assignments, all live.

**2 of 3 runs end with mutually exclusive claims co-active.** A search then returns several competing `active` rows with no signal that they compete — the failure A34's ranking contract exists to prevent, arriving by another route: a row never marked stale has no edge to order by, so nothing *looks* wrong.

### Deliberately not asserted as expected behaviour

Pinning "the oldest row stays active" would turn a defect into a contract. What ships instead is the evidence — `docs/contradiction-verification/a37-n-way-chain-findings.md` plus `benchmark/a37_n_way_chain.py` — and the invariants that must hold at any chain length. The fix is filed as its own row.

Likely mechanism (unconfirmed, stated as a lead not a conclusion): each write's detection races the previous write's status updates; candidate queries filter `status IN (active, confirmed, pending)` so a row mid-transition can be missed, and the Redis lock is keyed per **memory**, not per subject. Run 2 firing nothing points at the candidate lookup rather than the judge.

## Testing

9 new tests. Full suite green apart from the 8 failures that pre-exist on main (`test_capability_usage` / `test_request_observation`).

Refs: canonical A16, A37, A38, A39.

### DIFF
--- a/benchmark/a37_n_way_chain.py
+++ b/benchmark/a37_n_way_chain.py
@@ -1,0 +1,69 @@
+#!/usr/bin/env python3
+"""A37 — does a 3-step update chain leave the OLDEST claim active?
+
+One observation could be LLM non-determinism. Run the triangle three times with
+different subjects and report how often the oldest row survives as `active`
+alongside the newest.
+"""
+
+import json
+import time
+import urllib.error
+import urllib.request
+
+BASE, KEY = "http://localhost:8000", "dev-admin-key"
+SETTLE = 55
+
+
+def call(method, path, body=None):
+    req = urllib.request.Request(
+        BASE + path,
+        method=method,
+        headers={"X-API-Key": KEY, "Content-Type": "application/json"},
+        data=json.dumps(body).encode() if body is not None else None,
+    )
+    try:
+        with urllib.request.urlopen(req, timeout=180) as r:
+            return r.status, json.loads(r.read().decode(), strict=False)
+    except urllib.error.HTTPError as e:
+        return e.code, {"raw": e.read().decode()[:150]}
+
+
+def w(t, c):
+    return call(
+        "POST", "/api/v1/memories", {"tenant_id": t, "agent_id": "a37", "content": c}
+    )[1]
+
+
+def st(t, mid):
+    s, m = call("GET", f"/api/v1/memories/{mid}?tenant_id={t}")
+    return (m.get("status"), m.get("supersedes_id")) if s == 200 else (f"<{s}>", None)
+
+
+subjects = [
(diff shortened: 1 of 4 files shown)


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

arkash20:
Rebased onto main. CI green (13 checks).

@Eldad-Caura please approve

Eldad-Caura (approved):

Issue #511, opened 2026-07-01
TESTING=1 env var enables destructive endpoints but is absent from .env.example


jy7lsna:
### MemClaw version

2.1.0

### Deployment mode

Self-hosted (Docker Compose)

### Operating system

Windows 11

### Python version

3.12.7

### PostgreSQL version

16.4

### LLM provider in use

openai

### Describe the bug

Setting TESTING=1 registers a destructive router that exposes time-warp and field-rewrite endpoints. This env var is checked in two places:
- core-api/src/core_api/app.py:656
- core-api/src/core_api/routes/testing.py:53

It does not appear anywhere in .env.example. A developer copying .env.example as their starting config has no hint this variable exists, what it does, or that it must never be set in production.

### Steps to reproduce

1. Open .env.example in the repo root.
2. Search for "TESTING".
3. Observe: not present.

### Expected behavior

.env.example should include a commented-out entry for TESTING with a clear note that it enables destructive test-only endpoints and must never be set to 1 in production. Example:

# TESTING=1  # DO NOT set in production — enables destructive time-warp/field-rewrite endpoints

### Actual behavior

TESTING is absent from .env.example. Its existence and danger are undocumented for anyone setting up from the example config.

### Relevant logs / config

```shell

```

### Pre-flight

- [x] I searched existing issues and Discussions for duplicates.
- [x] This is not a security vulnerability (those go through Security Advisories).

PR #1242, opened 2026-09-02
docs(env): document JWT_SECRET and SETTINGS_ENCRYPTION_KEY in .env.example

### LINKED ISSUE #717: [Bug]  docs(env): JWT_SECRET and SETTINGS_ENCRYPTION_KEY both "Required in production" are missing from .env.example
### MemClaw version

2.1.0

### Deployment mode

Self-hosted (Docker Compose)

### Operating system

Windows 11

### Python version

3.12.7

### PostgreSQL version

16.4

### LLM provider in use

openai

### Describe the bug

core-api/src/core_api/config.py defines two settings that are explicitly commented as required before going to production:

  settings_encryption_key: str = ""                        # Required in production (Fernet key)
  jwt_secret: str = "change-me-in-production"               # Required in production

Neither JWT_SECRET nor SETTINGS_ENCRYPTION_KEY appears anywhere in .env.example. A self-hoster copying .env.example as their starting config has no signal that either variable exists, that they default to an insecure placeholder / empty string, or that both must be set before a real deployment.

### Steps to reproduce

1. Open .env.example in the repo root.
2. Search for "JWT_SECRET" — not present.
3. Search for "SETTINGS_ENCRYPTION_KEY" — not present.
4. Compare against core-api/src/core_api/config.py, where both are marked "# Required in production."

### Expected behavior

.env.example documents both variables, following the existing "REQUIRED: change in production" comment convention already used for POSTGRES_PASSWORD in the same file, e.g.:

  # REQUIRED in production: random 32+ byte secret used to sign auth tokens. Defaults to an insecure placeholder.
  JWT_SECRET=

  # REQUIRED in production: Fernet key used to encrypt stored settings. Defaults to empty (encryption disabled).
  SETTINGS_ENCRYPTION_KEY=

### Actual behavior

Both variables are absent from .env.example. JWT_SECRET silently defaults to the literal string "change-me-in-production" and SETTINGS_ENCRYPTION_KEY defaults to "" (encryption effectively off) with no warning surfaced anywhere in the example config a new self-hoster is told to copy.

### Relevant logs / config

```shell
core-api/src/core_api/config.py:248-249:
  settings_encryption_key: str = ""  # Required in production (Fernet key)
  jwt_secret: str = "change-me-in-production"  # Required in production

.env.example's "-- Auth --" section (around line 67-72) currently only documents ADMIN_API_KEY and MEMCLAW_API_KEY — JWT_SECRET and SETTINGS_ENCRYPTION_KEY belong in the same section.
```

### Additional context

Same category of gap as #511 (TESTING env var missing from .env.example), but for the opposite direction of risk: instead of an undocumented flag that's dangerous to set, these are undocumented flags that are dangerous to leave unset a production deploy can silently run with a well-known, guessable JWT secret and no settings encryption.

### Pre-flight

wylovelyi:
## Summary

Adds `JWT_SECRET` and `SETTINGS_ENCRYPTION_KEY` to `.env.example`, in the existing `-- Auth --` section, so the example config documents the two settings that `core-api` marks as required in production.

## Related Issue

Closes #717

## Type of Change

- [x] Documentation update
- [ ] Other:

(The issue is labelled `kind/bug`: the current example config silently implies no auth secret and no encryption are needed. The fix itself is a docs/config change.)

## How Has This Been Tested?

Docs/config only, so there is no runtime behaviour to exercise. Verification was by inspection against the source of truth:

- Confirmed both defaults in `core-api/src/core_api/config.py` (lines 391-392): `jwt_secret = "change-me-in-production"`, `settings_encryption_key = ""`.
- Confirmed the config declares no `env_prefix`, so the bare names `JWT_SECRET` / `SETTINGS_ENCRYPTION_KEY` are the correct variable names.
- Confirmed both names were absent from `.env.example` before the change (the insertion script asserts on this before writing).
- Confirmed the diff is `+6 / -0` in `.env.example` and nothing else.

## Checklist

- [x] `mypy` passes
- [x] `pytest` passes locally

Notes on the last four:

- **Tests / ruff / mypy / pytest:** no Python is touched — the change is six added comment-and-assignment lines in `.env.example` — so there is nothing for those tools to act on. Adding a test that asserts on the contents of an example file would be testing the docs, not the code.
- **CHANGELOG:** `CHANGELOG.md` carries no `Unreleased` section; it is generated by release-please from Conventional Commits (stated in the file header). This commit uses the `docs(env):` prefix, so it will be picked up automatically on release.

## Additional Notes

Wording follows the convention already used elsewhere in this file (`ADMIN_API_KEY`, `POSTGRES_PASSWORD`): an inline `REQUIRED in production` note that also names the insecure default, so a reader sees both the requirement and the consequence of skipping it.

```diff
 # CORE_STORAGE_SHARED_SECRET=
+# JWT signing secret — signs auth tokens. REQUIRED in production: use a random
+# 32+ byte value; defaults to the insecure placeholder "change-me-in-production".
+JWT_SECRET=
+# Fernet key used to encrypt stored settings. REQUIRED in production: defaults
+# to empty, which leaves encryption disabled.
+SETTINGS_ENCRYPTION_KEY=
```

Related prior art: #511 covered the same category of gap for the `TESTING` env var, in the opposite direction of risk (a dangerous-to-set flag left undocumented, versus these dangerous-to-leave-unset ones).

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -74,6 +74,12 @@ CAURA_API_KEY=
 # random value in a private volume; non-Compose deployments must set the same
 # non-empty value on every storage caller and storage instance.
 # CORE_STORAGE_SHARED_SECRET=
+# JWT signing secret — signs auth tokens. REQUIRED in production: use a random
+# 32+ byte value; defaults to the insecure placeholder "change-me-in-production".
+JWT_SECRET=
+# Fernet key used to encrypt stored settings. REQUIRED in production: defaults
+# to empty, which leaves encryption disabled.
+SETTINGS_ENCRYPTION_KEY=
 
 # -- Embedding provider -----------------------------------------------------
 # Options: openai | local | fake


Eldad-Caura:
Thank you for taking this on and for documenting these security-sensitive settings. We have since merged #1299, which fixes #717 with the same documentation while keeping JWT_SECRET at the guarded "change-me-in-production" placeholder; this PR uses a blank value that would bypass the current literal-default startup check. Closing this as superseded.

PR #1243, opened 2026-09-02
fix(clients): validate recall() response body before parsing

### LINKED ISSUE #719: [Bug] clients: recall(): Python crashes with AttributeError, TypeScript silently returns empty, on a non-dict response body
### MemClaw version

2.1.0

### Deployment mode

Self-hosted (Docker Compose)

### Operating system

Windows 11

### Python version

3.12.7

### PostgreSQL version

16.4

### LLM provider in use

openai

### Describe the bug

Distinct from #510, which scopes only to search(). recall()'s handling of an unexpected response shape differs by language for the same endpoint, and neither behavior is the right one.

Python: clients/python/src/memclaw_client/client.py `recall()` calls `RecallResult.from_dict(self._post(...))`. models.py's `RecallResult.from_dict()` calls `data.get("supporting_memories")` unconditionally. If the API returns a non-dict 200 body (a bare list, a string, or some other unexpected JSON value), `.get` doesn't exist on that type and Python raises an undocumented AttributeError — not MemClawAPIError.

TypeScript: clients/typescript/src/index.ts `recall()` reads `data?.supporting_memories` using optional chaining, so the identical malformed body is swallowed silently: the method returns `{ summary: null, supportingMemories: [], raw: data }` with no error raised at all.

So for the exact same malformed-response scenario, the Python client crashes with an exception type callers can't anticipate, while the TypeScript client returns a result that looks like a legitimate empty recall — two different wrong behaviors, neither matching the "raise MemClawAPIError on unexpected shape" contract #510 establishes for search().

### Steps to reproduce

1. Instantiate the Python client pointed at a mock server.
2. Have the mock return HTTP 200 with a bare JSON array body, e.g. `[]` or `["error", "quota exceeded"]` (not a dict).
3. Call `mc.recall("anything")`.
4. Observe: raises `AttributeError: 'list' object has no attribute 'get'` — not MemClawAPIError, and the caller has no clean way to catch it.
5. Repeat with the TypeScript client and an equivalent mocked fetch response.
6. Observe: no error `recall()` resolves with `{ summary: null, supportingMemories: [], raw: [...] }`, indistinguishable from a legitimate "nothing found" result.

### Expected behavior

Both clients raise MemClawAPIError / MemClawApiError when a 200 response body isn't a dict (or lacks the expected keys) — the same contract #510 asks for on search(), applied consistently to recall() too, so callers get one predictable error type across every method in both languages.

### Actual behavior

Python: undocumented AttributeError crash.
TypeScript: silent success with an empty result, indistinguishable from zero supporting memories.

### Relevant logs / config

```shell
Python: undocumented AttributeError crash.
TypeScript: silent success with an empty result, indistinguishable from zero supporting memories.
```

### Additional context

_No response_

### Pre-flight

wylovelyi:
## Summary

`recall()` parsed the `POST /api/v1/recall` response body without validating its shape. When the server returned a non-object body (e.g. a JSON array or a bare scalar), the Python client crashed with `AttributeError` and the TypeScript client silently returned an empty `RecallResult`.

This mirrors the guard already present in `search()`: both clients now require a JSON object and raise `CauraAPIError` / `CauraApiError` otherwise.

## Related Issue

Fixes #719

## Type of Change

## How Has This Been Tested?

Added unit tests on both clients (mocked transport, no network):

- Python `clients/python/tests/test_client.py`: `test_recall_raises_on_non_dict_body`, `test_recall_raises_on_non_object_body`
- TypeScript `clients/typescript/src/index.test.ts`: `recall throws when 200 body is not an object`, `recall throws when 200 body is a bare scalar`

These assert `recall()` raises `CauraAPIError` / `CauraApiError` with status 200 and message `recall response must be a JSON object`. The existing recall + search suites are unchanged in behavior.

## Checklist

- [x] Change is a verbatim mirror of the existing `search()` guard (no new constructs); `py_compile` passes

## Additional Notes

- The validation is identical in shape to `search()` so behavior is consistent across the two endpoints.
- `RecallResult.from_dict` is unchanged; the guard lives in the client method, matching the existing `search()` pattern.
- `CHANGELOG.md` is release-please–generated from the conventional-commit message, so it is left untouched per repo convention.

### DIFF
--- a/clients/python/src/caura_client/client.py
+++ b/clients/python/src/caura_client/client.py
@@ -108,7 +108,10 @@ class Caura:
         """Search + LLM summary. Returns a ``RecallResult`` context brief (POST /api/v1/recall)."""
         body: dict[str, Any] = {"tenant_id": self.tenant_id, "query": query, "top_k": top_k}
         body.update(extra)
-        return RecallResult.from_dict(self._post("/api/v1/recall", body))
+        data = self._post("/api/v1/recall", body)
+        if not isinstance(data, dict):
+            raise CauraAPIError(200, "recall response must be a JSON object")
+        return RecallResult.from_dict(data)
 
     def health(self) -> dict[str, Any]:
         """Liveness probe (GET /api/v1/health)."""
--- a/clients/python/tests/test_client.py
+++ b/clients/python/tests/test_client.py
@@ -155,6 +155,26 @@ def test_recall_ignores_the_key_the_server_never_sends():
     assert result.supporting_memories == []
 
 
+def test_recall_raises_on_non_dict_body():
+    def handler(request):
+        return httpx.Response(200, json=["not", "a", "dict"])
+
+    with pytest.raises(CauraAPIError) as exc:
+        make_client(handler).recall("q")
+    assert exc.value.status_code == 200
+    assert str(exc.value) == "[200] recall response must be a JSON object"
+
+
+def test_recall_raises_on_non_object_body():
+    def handler(request):
+        return httpx.Response(200, json="a plain string, not an object")
+
+    with pytest.raises(CauraAPIError) as exc:
+        make_client(handler).recall("q")
+    assert exc.value.status_code == 200
+    assert str(exc.value) == "[200] recall response must be a JSON object"
+
+
 def test_health():
     def handler(request):
         assert request.url.path == "/api/v1/health"
--- a/clients/typescript/src/index.test.ts
+++ b/clients/typescript/src/index.test.ts
@@ -133,6 +133,26 @@ test("recall ignores the key the server never sends", async () => {
   assert.deepEqual(result.supportingMemories, []);
 });
 
+test("recall throws when 200 body is not an object", async () => {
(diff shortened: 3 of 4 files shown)


Eldad-Caura:
Maintainer update pushed at 55cfb847. I rebased the original signed contributor commit onto current main unchanged, then added a separate signed TypeScript fix that narrows payload.summary to string or null. Verification: TypeScript compiled and passed 16/16 tests, Python passed 141/141 tests, Python Ruff check passed, and both naming gates passed. The client formatter still reports the repository baseline of pre-existing unformatted files, so no broad formatting churn was applied. Waiting for exact-head CI before retriggering review.

Eldad-Caura:
@claude

Eldad-Caura:
@erni-a

erni-a (approved):

erni-a:
Approved on 5 Sep and then left waiting, which is on us. Branch updated, CI green, merging now. Thanks @wylovelyi.

PR #1244, opened 2026-09-03
fix(clients): delete the legacy memclaw client packages and workflow

Eldad-Caura:
## Summary
- Deletes `clients/memclaw-client-shim/` (PyPI `memclaw-client`), `clients/npm-legacy-client/` (npm `@caura/memclaw-client`), and `clients/python/src/memclaw_client/` (the legacy import module bundled inside `caura-client`) — per Eldad's direction: "discard the memclaw variations in pypi/npm/homebrew/etc ... if someone already somehow downloaded it and use it, we'll need to upgrade him to use the caura replacement." No transition mechanism is kept for any of these.
- Deletes `.github/workflows/publish-npm-legacy-client.yml` in full. No separate workflow ever published the Python shim — it only ever built from `clients/python` via `publish-python-client.yml` — so that workflow's `memclaw-client-v*` tag trigger and the matching prefix-strip line are removed here too.
- Updates everything that depended on what's deleted in the same commit: `tests/test_client_metapackages.py` (drops the npm-legacy-client fixtures/tests), `clients/python/tests/test_legacy_alias.py` (drops the two tests that imported `memclaw_client`; keeps the two testing the in-package `MemClaw`/`MemClawError`/`MemClawAPIError` class aliases, which are unaffected), `scripts/do_not_touch_sentinel.py` (removes the 7 entries pinning now-deleted lines), and prose in `clients/python/README.md`, `clients/python/pyproject.toml`, `clients/python/src/caura_client/__init__.py`, and `.github/workflows/publish-npm-sdk.yml`'s comment referencing the deleted workflow by name.
- `docs/plans/rebrand-alias-retirement-policy.md` (proposed 2026-09-01) had called the Python and npm client aliases "permanent by choice," which is now superseded by Eldad's later, explicit direction to delete rather than deprecate. Added a superseded banner at the top and updated only the two rows this PR affects — not re-litigating any other lane's rows in that document.

## Kept, deliberately
- `caura_client.MemClaw` / `MemClawError` / `MemClawAPIError` — in-package class-level aliases, not a separate distribution, out of scope for this change, and their tests.
- `plugin/openclaw.plugin.json`'s `id: "memclaw"` and `MEMCLAW_AGENT_ID` — held per Eldad's "keep for now" pending the OpenClaw-plugin-id question.

## Gates
- `python3 scripts/legacy_name_ratchet.py --base origin/main` → `No new lines. 0 annotated, 26 removed, 0 excused moves (-26 net).`
[...]

PR #1247, opened 2026-09-03
fix(plugin): rename the systemd TLS drop-in from memclaw-tls.conf to caura-tls.conf

Eldad-Caura:
## Summary
**Customer-facing: this changes a file the `/install-plugin` endpoint's served bash script writes on an on-prem operator's own machine via `curl ... | bash`. It takes effect only on the next run of that install, never retroactively for an install already completed.**

`core-api/src/core_api/routes/plugin.py:583`'s echo line printed the path to `$_SD_DIR/memclaw-tls.conf` — I was asked to rename it as customer-facing prose, but it was actually accurate: it printed the real path of the file the heredoc two lines above it (`:559`) genuinely wrote, and that exact write was separately floored by `scripts/do_not_touch_sentinel.py` ("re-install writes a second systemd drop-in instead of replacing the customer file"). Renaming only the printed text would have made it lie about what's on disk.

The real fix, done here:
- Renames the write itself to `$_SD_DIR/caura-tls.conf`.
- In the same heredoc, for exactly one release, removes `$_SD_DIR/memclaw-tls.conf` if present, before `systemctl --user daemon-reload` — otherwise re-running install-plugin after this change leaves two drop-ins declaring the same `Environment=` line redundantly. Annotated `legacy-name-ok` with its own removal note (drop this line once customers have had one reinstall to pick it up).
- Fixes the echo line to print the new, now-accurate path.
- Moves the `do_not_touch_sentinel.py` pin from the old write to the new one.

## Test
Added `test_post_install_plugin_tls_bootstrap_writes_caura_named_drop_in`, exercising the TLS-bootstrap branch (HTTPS `api_url`) through the real `/install-plugin` endpoint end to end: asserts the new file is written, the old one is removed first (so a re-install has both files only never, not neither-then-both), and the printed message matches what's actually written. **Confirmed this test fails against the pre-fix code** (`git stash` the source change, rerun — fails on `assert 'cat > "$_SD_DIR/caura-tls.conf"...' in script`) before restoring the fix.

## Gates
- `python3 scripts/legacy_name_ratchet.py --base origin/main` → `No new lines. 0 annotated, 2 removed, 0 excused moves (-2 net).`
- `python3 scripts/do_not_touch_sentinel.py` → `All 40 protected strings survive.`
- `ruff check core-api/src tests` + `ruff format --check core-api/src tests` clean.
- `pytest tests/test_api_plugin.py` — 13 passed, against an isolated Postgres+pgvector container running the repo's own migrations (not a shared/borrowed DB).

Measured against `origin/main` at `caa1c3e88...` (post-#1244 merge), rebased.

### DIFF
--- a/core-api/src/core_api/routes/plugin.py
+++ b/core-api/src/core_api/routes/plugin.py
@@ -556,7 +556,11 @@ case "$CAURA_API_URL" in
       chmod 0644 "$PLUGIN_DIR/onprem-ca.pem"
       _SD_DIR="$HOME/.config/systemd/user/openclaw-gateway.service.d"
       mkdir -p "$_SD_DIR"
-      cat > "$_SD_DIR/memclaw-tls.conf" << SDEOF
+      # One release only: remove the pre-rename drop-in so re-running
+      # install-plugin doesn't leave both files behind, each declaring the
+      # same Environment= line redundantly.
+      rm -f "$_SD_DIR/memclaw-tls.conf"  # legacy-name-ok: removes the pre-rename drop-in for exactly one caura-client release, then drop this line (docs/plans/rebrand-alias-retirement-policy.md)
+      cat > "$_SD_DIR/caura-tls.conf" << SDEOF
 [Service]
 Environment="NODE_EXTRA_CA_CERTS=$PLUGIN_DIR/onprem-ca.pem"
 SDEOF
@@ -580,7 +584,7 @@ SDEOF
         fi
       done
       echo "    Saved CA → $PLUGIN_DIR/onprem-ca.pem"
-      echo "    Drop-in   → $_SD_DIR/memclaw-tls.conf"
+      echo "    Drop-in   → $_SD_DIR/caura-tls.conf"
       echo "    Shell env → ~/.bashrc + ~/.zshrc (NODE_EXTRA_CA_CERTS for new shells)"
     else
       rm -f "$PLUGIN_DIR/onprem-ca.pem"
--- a/scripts/do_not_touch_sentinel.py
+++ b/scripts/do_not_touch_sentinel.py
@@ -196,7 +196,7 @@ SENTINELS: tuple[Sentinel, ...] = (
     ),
     Sentinel(
         path="core-api/src/core_api/routes/plugin.py",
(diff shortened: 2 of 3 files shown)


Eldad-Caura:
Checked this against the actual gate rather than the docstring's semantics alone, because the two disagree for this specific case.

`legacy-name-ok`/`legacy-name-floor` are the only markers that exempt a line from the ratchet's counted headline. `legacy-name-deferred` deliberately does **not** exempt — per the script's own text, it "stays inside the gated count: it annotates deferred work and does not make this build green." That distinction matters here because the `rm -f "$_SD_DIR/memclaw-tls.conf"` line is a **brand-new line** (it never existed in the base tree — the old code only ever *wrote* that filename, never removed it), not a marker added to a pre-existing counted or exempt line. A brand-new line naming the legacy string, marked `legacy-name-deferred`, is exactly what rule 7 blocks: `scripts/legacy_name_ratchet.py --base origin/main` reports `This change adds the legacy name in 1 file(s)` and exits 1.

I verified this directly, twice, on the actual rebased branch:

```
$ sed -i 's/legacy-name-ok:/legacy-name-deferred:/' core-api/src/core_api/routes/plugin.py
$ python3 scripts/legacy_name_ratchet.py --base origin/main
...
This change adds the legacy name in 2 file(s).
  clients/python/pyproject.toml  (0 -> 1)
      44: memclaw-interviewer = "caura_client.interviewer.cli:main"
  core-api/src/core_api/routes/plugin.py  (32 -> 31, count fell — the text is new)
      562: rm -f "$_SD_DIR/memclaw-tls.conf"  # legacy-name-deferred: ...
EXIT: 1
```

(reverted, re-ran clean with `legacy-name-ok` restored — `EXIT: 0`, `-2 net`.)

So `legacy-name-ok` isn't a mistake here — it's the only marker of the three that actually passes CI for a newly-added line, and I chose it deliberately over `legacy-name-floor` because the reason text states this is temporary (with a removal note), not permanent. I'd rather that live as prose in the comment than as a marker the tooling itself would reject. Leaving this as-is; happy to revisit if there's a way to satisfy both the tooling and the "stays in the gated count" property that I'm missing.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #1249, opened 2026-09-03
chore(clients): annotate the permanent class/exception alias lines

Eldad-Caura:
## Summary
Part of the wider repo sweep — auditing every unannotated `memclaw` line rather than only the items named in the original brief.

The `MemClaw`/`MemClawError`/`MemClawAPIError` class and exception aliases in both `clients/python` and `clients/typescript` are permanent by design (rule 3 — same pattern as the MCP tool rename) and already documented with a `legacy-name-ok` comment on the assignment's own explanatory comment line in most places. But `scripts/legacy_name_ratchet.py` only exempts the exact physical line a marker sits on, so the multi-line `from .exceptions import (...)` block, the `__all__` list, and the mirrored TypeScript `export const`/`export type` lines were still counted as unannotated debt — 18 lines across 5 files — even though nothing about them is actually unresolved.

Adds the same `legacy-name-ok: rule 3 permanent class/exception alias` marker directly to each of those lines. No behavior change; verified nothing needed a rename (all of it is the intentional, already-decided permanent alias surface).

## Gates
- `python3 scripts/legacy_name_ratchet.py --base origin/main` → `No new lines. 18 annotated, 0 removed, 0 excused moves (-18 net).`
- `python3 scripts/do_not_touch_sentinel.py` → `All 39 protected strings survive.`
- `ruff check src tests` (clients/python, with the pinned `ruff>=0.15.4,<0.16` from the `dev` extra) clean.
- `pytest` — 139 passed (clients/python, isolated venv).
- `npm test` (clients/typescript) — `tsc` compiles clean, 14/14 tests passed.

Measured against `origin/main` at `c0930483...` (post-#1247 merge).

### DIFF
--- a/clients/python/src/caura_client/__init__.py
+++ b/clients/python/src/caura_client/__init__.py
@@ -10,26 +10,26 @@ package-forwarder distributions that once depended on this one were retired
 
 from __future__ import annotations
 
-from .client import DEFAULT_BASE_URL, Caura, MemClaw
+from .client import DEFAULT_BASE_URL, Caura, MemClaw  # legacy-name-ok: rule 3 permanent class alias
 from .exceptions import (
     AuthError,
     CauraAPIError,
     CauraError,
-    MemClawAPIError,
-    MemClawError,
+    MemClawAPIError,  # legacy-name-ok: rule 3 permanent exception alias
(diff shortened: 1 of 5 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #1250, opened 2026-09-03
test: annotate the permanent tool/env alias literals in the compat-alias tests

Eldad-Caura:
## What

Adds `legacy-name-ok: rule 3 ...` markers to the `memclaw` literals in
`tests/test_tool_rename_aliases.py`, `tests/test_migration_012_safety_gate.py`,
and `tests/conftest.py`. No renames — these three files exist specifically
to pin two permanent aliases forever:

- `tests/test_tool_rename_aliases.py` — the `memclaw_*` → `caura_*` MCP
  tool-dispatch shim (`core_api/mcp_server.py`'s `name.startswith("memclaw_")`,
  sentinel-protected).
- `tests/test_migration_012_safety_gate.py` — the
  `MEMCLAW_RUN_DESTRUCTIVE_MIGRATIONS` env-var alias for the vector-dim
  migration's destructive-run gate. Introduced a `LEGACY_ENV` module
  constant instead of repeating the literal in all 6 call sites — one
  annotated definition instead of six annotated call sites.
- `tests/conftest.py` — the defensive unset of a leaked `MEMCLAW_API_KEY`/
  `MEMCLAW_KEY` from developers' shells, and the comment documenting the
  `settings.memclaw_api_key` dual-read field it's guarding against.

Several of the original literal call sites, once a trailing marker was
added, no longer fit `ruff format`'s line-length limit — `ruff format`
would then wrap the call so the closing-paren line (carrying the marker)
no longer shared a physical line with the matched "memclaw" text, which
the ratchet requires. Restructured those (the `LEGACY_ENV` constant above;
a `suffix`/`legacy_marker` local in the alias test; the leaky-env tuple
split one-per-line) so the literal and its marker always share one short,
already-formatted line.

4 occurrences are left unannotated on purpose: the first line of a module
or function docstring in each file (`test_tool_rename_aliases.py:1,8,74`;
`test_migration_012_safety_gate.py:146`) — an inline `#` marker there
would leak into `__doc__`, and rewording to avoid the literal would blur
exactly which prefix/env-var name the guarantee covers. Same call made for
`clients/python/src/caura_client/__init__.py`'s docstring in #1249.

`tests/conftest.py:69`'s `postgresql+asyncpg://memclaw:changeme@.../memclaw`
default is untouched here — it's part of the wider Postgres role/db-name
question (also live in `docker-compose*.yml`, `.env.example`, CI, etc.)
that needs an infra-owner decision before anyone touches it, not a
sweep-scope annotation.

## Verification

  lines. 4 annotated, 11 removed, 0 excused moves (-15 net).` EXIT 0
  (the 11 "removed" are literals now referenced via `LEGACY_ENV`/locals
  instead of repeated inline, so they no longer read as bare occurrences)
  survive.`
- `pytest tests/test_tool_rename_aliases.py tests/test_migration_012_safety_gate.py -v`
  → 17 passed (no DB required — both suites are `@pytest.mark.unit`)
- `ruff check tests/test_tool_rename_aliases.py tests/test_migration_012_safety_gate.py tests/conftest.py`
  → All checks passed
- `ruff format --check` on the same three files → all already formatted
  (fixes the CI failure from the first push, caused by an unformatted
  trailing marker)

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -41,15 +41,20 @@ for _k, _v in _TEST_DEFAULTS.items():
 
 # Defensively unset env vars that change auth shape and routinely leak in
 # from developers' shells (the OSS plugin onboarding writes
-# ``~/.config/caura-keys.env`` with ``MEMCLAW_API_KEY=...`` and many
+# ``~/.config/caura-keys.env`` with ``MEMCLAW_API_KEY=...`` and many  # legacy-name-ok: rule 3 env alias
 # rc files source it for the openclaw CLI). A leaked value flips
-# ``settings.memclaw_api_key`` to truthy, which makes ``get_auth_context``
+# ``settings.memclaw_api_key`` to truthy, which makes ``get_auth_context``  # legacy-name-ok: rule 3 dual-read field
 # enforce the gate at Path 2 with 401s before any standalone-mode
 # bypass — silently failing every test that doesn't sniff the env
 # itself (e.g. test_rate_limit's auth-gated burst test, which gets all
 # 401s instead of the expected 200/429 mix). Unset rather than
 # setdefault — setdefault doesn't override an existing env value.
-for _leaky in ("MEMCLAW_API_KEY", "MEMCLAW_KEY", "CAURA_API_KEY", "CAURA_KEY"):
+for _leaky in (
+    "MEMCLAW_API_KEY",  # legacy-name-ok: rule 3 env alias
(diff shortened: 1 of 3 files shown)


Eldad-Caura:
@erni-a pleasr aprvoe

erni-a (dismissed):

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #1251, opened 2026-09-03
chore: floor-mark sentinel-protected memclaw literals in core-api and plugin

Eldad-Caura:
## What

Adds `legacy-name-floor: floor` markers to 13 lines across 10 files where
`scripts/do_not_touch_sentinel.py` already pins an exact "memclaw" literal
in application code — meaning something outside this repo (a saved
prompt, an already-deployed plugin install, an existing tenant settings
row, an existing agent row) breaks if the string disappears. The marker
had never actually landed on the matched line itself in any of these
files; only `sentinel.py`'s own copy of each string was annotated (it
also contains "memclaw" and needs its own marker to pass the ratchet, but
that's a different physical line from the one it's protecting).

No renames, no behavior change — every edit is a trailing comment (Python
`#` / TypeScript `//`, or a bash `#` comment inside the f-string install
script in `plugin.py` — same style as the already-annotated line 562 in
that file from #1247).

Touched:
- `core-api/src/core_api/mcp_server.py` — the `memclaw_*` dispatch shim
  (`name.startswith("memclaw_")` + its `removeprefix` pair)
- `core-api/src/core_api/app.py` — the frozen `/api/v1/memclaw/keystones`
  route mount (reformatted to multi-arg so the marker and the literal
  share a line — see note below)
- `core-api/src/core_api/agent_ids.py` — `INSIGHTER_AGENT_ID` /
  `DOC_INDEXER_AGENT_ID`, seeded by migration 030
- `core-api/src/core_api/routes/fleet.py` and
  `core-api/src/core_api/services/organization_settings.py` — the
  `"memclaw"` tenant-settings namespace and its dotted-key spelling
- `core-api/src/core_api/services/report_corpus.py` — the
  `memclaw-smoke` probe-title filter (pairs with the plugin-side literal
  below)
- `core-api/src/core_api/routes/plugin.py` — the install script's
  `~/.openclaw/plugins/memclaw` dir and its rc-marker comment string
- `plugin/src/context-engine.ts` — the health-check probe's
  `memclaw-smoke-` title (the emitter half of the report_corpus.py pair)
- `plugin/src/reconcile-skills.ts` — the `.memclaw-owned` per-skill marker
- `plugin/src/paths.ts` — the plugin install-dir helper

## What's deliberately NOT in this PR

`core-storage-api`'s two migrations that also carry sentinel-protected
`memclaw` text (`030_register_memclaw_insighter.py`,
`019_tenant_suppression.py`) are untouched: their occurrences sit inside
an `op.execute("""...""")` SQL string or a module docstring. There's no
way to attach a same-line marker there without either altering
already-applied migration SQL (forbidden — rule 2, immutable history) or
corrupting `__doc__`. Same structural gap already noted for the
`ARG MEMCLAW_VERSION=` Dockerfile default and the JSON manifests in
earlier PRs — a ratchet-side limitation worth raising separately, not
something this PR can fix with a text edit.

## Verification

  lines. 13 annotated, 1 removed, 0 excused moves (-14 net).` EXIT 0
  survive.`
- `ruff format --check core-api/src/` → `203 files already formatted`
- `ruff check core-api/src/` → All checks passed
- `pytest tests/test_tool_rename_aliases.py tests/test_api_plugin.py core-api/tests/test_report_corpus.py -v`
  → 27 passed (covers the mcp_server dispatch shim, the plugin install
  script's TLS/rc-marker behavior, and the report-corpus smoke-probe
  filter)
- Plain import of every touched core-api module succeeds
- `npx tsc --noEmit` (plugin) → clean
- `npm test` (plugin) → 426/426 passed

### DIFF
--- a/core-api/src/core_api/agent_ids.py
+++ b/core-api/src/core_api/agent_ids.py
@@ -20,7 +20,7 @@ DEFAULT_AGENT_ID = "mcp-agent"
 # collapsing onto the anonymous default (which is unregistered, so its writes
 # never surface in Prism or the per-agent report). Registered per-tenant with
 # ``belonging_type='service'`` the first time the job runs for that tenant.
-INSIGHTER_AGENT_ID = "memclaw-insighter"
+INSIGHTER_AGENT_ID = "memclaw-insighter"  # legacy-name-floor: floor
 
 # Trust tier the insighter self-registers at: service level, granting cross-fleet
 # read (>=2, needed for scope='all') and write (>=3, it persists scope_org
@@ -39,7 +39,7 @@ INSIGHTER_TRUST_LEVEL = 3
 # cosmetic here: ``caura_insights`` defaults to ``scope="agent"``, which filters
 # ``Memory.agent_id == agent_id``, so rows attributed to this service identity are
 # invisible to every real agent's default insights run.
-DOC_INDEXER_AGENT_ID = "memclaw-doc-indexer"
+DOC_INDEXER_AGENT_ID = "memclaw-doc-indexer"  # legacy-name-floor: floor
 
 # Bare ``"main"`` is the OpenClaw plugin's *unset* default agent_id: when an
 # operator never sets ``CAURA_AGENT_ID`` every install collapses onto this one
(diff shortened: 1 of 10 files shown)


Eldad-Caura:
@erni-a pelae approve

erni-a (approved):

PR #1256, opened 2026-09-03
chore: floor-mark the staging-memclaw-tei Cloud Run comment

Eldad-Caura:
## What

Confirmed live via an infra check by the programme coordinator:
`staging-memclaw-tei` and its `caura`-named twin `staging-caura-tei`
both exist right now. This is a mid-cutover service pair, not an
un-migrated one — renaming the comment would misdescribe which Cloud
Run service `core-api`'s embedding-concurrency cap math is actually
about.

`scripts/capture_pg_locks_during_storm.py:7`'s matching
`staging-memclaw-core-storage-writer` reference (also confirmed live,
also has a `caura`-named twin) is left unmarked here — that occurrence
sits inside the module's docstring, so an inline marker would leak into
`__doc__`. Same accepted-debt class as the migration/docstring cases
from earlier PRs in this sweep.

## Verification

  lines. 1 annotated, 0 removed, 0 excused moves (-1 net).` EXIT 0
  survive.`
- `ruff format --check core-api/src/` / `ruff check core-api/src/` →
  clean
- Comment-only change; no code path affected.

### DIFF
--- a/core-api/src/core_api/config.py
+++ b/core-api/src/core_api/config.py
@@ -284,7 +284,7 @@ class Settings(BaseSettings):
     # joiners take no slot), so one hot tenant's search storm can't
     # occupy the whole embedding service and starve other tenants
     # (noisy-neighbor-search). The TEI backend is a fixed pool
-    # (``staging-memclaw-tei``: 2 instances x containerConcurrency 10 =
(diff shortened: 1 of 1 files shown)


Eldad-Caura:
@erni-a pleae approve

erni-a (approved):

PR #1257, opened 2026-09-03
chore: rename the ephemeral local/CI Postgres role and database from memclaw to caura

Eldad-Caura:
## ⚠️ DO NOT MERGE — held pending a cross-repo decision

This is approved by Eldad and the programme coordinator (`wt-sec-03`) is
taking this exact rename as the convention for `caura-enterprise`'s
matching four sites once this lands. Per instruction: build it, verify
it, open it, do not merge it myself (standard practice on this whole
sweep — I never merge my own PRs).

## ⚠️ OPERATOR ACTION REQUIRED if you've ever run `docker compose up` here

**Read this before pulling this branch into a checkout with an existing
`pgdata` volume.** `docker-compose.yml`/`.dev.yml` mount `pgdata` as a
persistent named volume. Postgres only runs `initdb` — and therefore
only honors `POSTGRES_USER`/`POSTGRES_DB` — on an **empty** data
directory. On any machine that already has this volume, the role and
database silently stay `memclaw` no matter what this diff says.

Verified directly, in an isolated container (not the shared dev stack):
on a stale volume, `pg_isready -U caura` still reports the server
**healthy** — `pg_isready` only checks that the postmaster accepts
connections, it does not authenticate as the named role. So
`docker compose ps` shows postgres healthy, every dependent service
(gated on `depends_on: condition: service_healthy`) starts anyway, and
*then* each one fails on its own DB connection with `role "caura" does
not exist` / `InvalidPasswordError`. **The visible symptom is every app
service crash-looping or refusing to start — not "postgres is
unhealthy."** Don't debug this by checking `docker compose ps`.

**Fix, pick one:**
```bash
# Option A — destroys local Postgres data (fine, it's disposable dev data)
docker compose down -v && docker compose up -d

# Option B — keeps the volume, connects as the still-existing old role
docker exec -it <db-container> psql -U memclaw -d memclaw \
  -c "CREATE ROLE caura LOGIN SUPERUSER PASSWORD 'changeme';"
docker exec -it <db-container> psql -U memclaw -d memclaw \
  -c "CREATE DATABASE caura OWNER caura;"
```
Both verified end-to-end in an isolated test container. Note Option B
needs two separate `-c` invocations — `CREATE DATABASE` cannot run in
the same implicit transaction as `CREATE ROLE`.

## What

Straight substitution across the whole **ephemeral** local/CI Postgres
population — `POSTGRES_USER`/`POSTGRES_DB` become `caura` (was
`memclaw`), the storage-suite database becomes `caura_storage` (was
`memclaw_storage`), DSNs become
`postgresql+asyncpg://caura:changeme@<host>:<port>/caura`. Port and
password unchanged everywhere. Nothing added or dropped beyond that.

This population is explicitly **not** the same as the live AlloyDB
clusters (`memclaw-production`, `memclaw-staging-alloydb`), which really
are named `memclaw` in production and are staying that way — confirmed
via a direct infra check by the coordinator. These 9 files only ever
talk to an ephemeral or local-only Postgres:

- `.github/workflows/ci.yml` (13 sites) — the ephemeral test-container
  Postgres, the storage-suite creation SQL, both suites' DSNs
- `docker-compose.yml` / `docker-compose.dev.yml` (6 sites each) — the
  local `db` service's credentials + healthcheck, and the app services'
  DSNs
- `.env.example`, `env.dev` (2 sites each) — the documented local
  defaults
- `core-storage-api/config.py`'s `LOCAL_DATABASE_URL` and
  `scripts/benchmark_blend_locomo.py`'s local benchmark DSN default —
  both exist specifically to match what the files above provision, so
  leaving them on `memclaw` while renaming docker-compose would have
  silently broken "start the db container, then run the app locally
  against it outside compose" (a workflow `LOCAL_DATABASE_URL`'s own
  comment describes)
- `tests/conftest.py`'s `TEST_DB_URL` fallback — same shape, was never
  marked
- `core-storage-api/scripts/preflight_012.py` — reclassified from
  "floor" to "rename" on its own evidence: its primary DSN path is
  dynamic (`settings.database_url`), the `memclaw`-defaulted
  `POSTGRES_USER`/`DB` fallback only fires if that import fails (nothing
  in `ci.yml` exercises it), and its own docstring's `--dsn` example
  uses `localhost` + the placeholder password `changeme`

`scripts/backup-db.sh` was examined the same way and **left alone**: an
ops tool (GCS upload, 7-day retention cleanup) whose `ALLOYDB_*`
defaults can reach the real, still-`memclaw`-named production database —
floor-marking that is a separate, smaller PR if wanted, not part of this
one.

Two of the renamed lines were marked `legacy-name-ok` by an earlier lane
under the premise that the `memclaw` local-dev convention was permanent.
That premise no longer holds (confirmed by the coordinator + Eldad). The
comments left in place of each marker — in `core-storage-api/config.py`
and `scripts/benchmark_blend_locomo.py` — say explicitly that the alias
is retired, not silently dropped. `tests/conftest.py`'s occurrence was
never marked, so nothing is reversed there.

## Verification

  lines. 0 annotated, 33 removed, 0 excused moves (-33 net).` EXIT 0
  (the two `legacy-name-ok` removals are reported and explained above,
  not silently dropped)
  survive.`
- `ruff format --check` / `ruff check` on `core-storage-api/src/`,
  `scripts/`, `tests/` → clean
- **Real end-to-end verification against an isolated, freshly-created
  `caura`-credentialed Postgres container** (not mocked): `alembic
  upgrade head` — all 41 revisions applied cleanly, including migration
  030 which seeds the (unrelated, untouched) `memclaw-insighter` service
  agent id. Grepped the full migration history for `GRANT`/`OWNER
  TO`/`ALTER ROLE`/`CREATE ROLE` — zero hits, so no migration hardcodes
  the old role name the way a sibling repo's did (checked after that
  repo's lane found the issue there).
- `pytest tests/ -m "not benchmark"` → **5972 passed**, 4 skipped, 1
  xfailed, against the migrated `caura` database
- `pytest core-storage-api/tests/` → **320 passed**, against a
  `caura_storage` database created the same way `ci.yml` creates it
- `python -m core_storage_api.scripts.preflight_012 --dsn
  postgresql+asyncpg://caura:changeme@localhost:<port>/caura` → runs
  clean, reports head revision 041
- Stale-volume hazard reproduced and both remediation paths verified, as
  described above


Eldad-Caura:
@erni-a please aprove

erni-a (approved):

PR #1259, opened 2026-09-03
fix(auth): scope the delete trust gate to agent credentials

Eldad-Caura:
## The defect

REST and MCP gave **opposite answers to the same delete**. Same credential, same tenant, same `agent_id`:

- `DELETE /api/v1/memories/{id}?tenant_id=…&agent_id=yanki-smoke` → **403** `access policy: principals of fleet 'none' are not permitted to delete memories.`
- `caura_manage {op: "delete", memory_id: …, agent_id: "yanki-smoke"}` → **accepted**, row soft-deleted.

REST was the surface in the wrong. It built its authorization principal as `auth.agent_id or agent_id`, promoting a **caller-supplied query parameter** to a principal. For a tenant key (no gateway `X-Agent-ID`) that handed the caller the choice of whether the gate applied:

| Tenant key sends | Old behaviour |
|---|---|
| no `agent_id` | gate **skipped** — `caller_agent_id` falsy |
| `agent_id=<any trust≥3 agent>` | gate **passed** — borrows that agent's trust |
| `agent_id=<own low-trust identity>` | gate **refused** |

The only caller `enforce_delete` reliably stopped was the honest one. The old comment claimed the precedence rule stopped a caller *"skipping the gate by omitting/spoofing `agent_id`"* — true for an agent credential, where `auth.agent_id` wins, and false for a tenant key, which has none.

## The decision this encodes

**The trust ladder governs AGENT credentials; tenant scope governs TENANT keys.**

A tenant key holds no trust level to compare and is authorized by tenant scope — exactly as it already is on this route's bulk siblings (`test_rest_delete_all_tenant_key_unchanged`, pre-existing) and on MCP `read` / `update` / `lineage` / `bulk_delete`, where the same distinction is documented as deliberate at `mcp_server.py:1359-1365`.

Written into `enforce_delete`'s docstring so it is a contract rather than an inference.

## The change

One behavioural line. `caller_agent_id = auth.agent_id` — the principal is the authenticated identity alone. `None` means "no agent identity", the gate does not apply, and REST stops returning a 403 the other surface does not. **The same edit closes the spoof**: the query param is no longer a principal, so it cannot be used to borrow another agent's trust.

**Attribution is deliberately kept separate.** `log_action` still records `auth.agent_id or agent_id`, so a tenant key's delete stays attributed to the identity it supplied rather than to `None`. The two uses sat one line apart behind a single variable — which is how this defect was born — and are now two names with a comment each saying which may trust caller input.

**MCP is unchanged by design.** The three `if caller_agent_id:` sites (`:1394` `bulk_delete`, `:1621` `delete`, `:2508` `caura_doc` op=delete) were already correct under this ruling, so they gain the **written exemption** rather than a code change — including an explicit warning not to "fix" them by falling back to the tool argument. That change was proposed and rejected: it would re-open the spoof, and because `caller_agent_id` is also the principal for the `authorize_memory_access` scope checks at `:1429` / `:1480` / `:1528` / `:1628`, it would widen it past delete to **read, lineage and update** — which the comment above `:1352` already forbids in as many words.

## For the reviewer: this makes REST more permissive, and nothing is lost

An authorization check is being deleted, so that deserves stating plainly rather than waiting to be asked. The check was **already bypassable by omitting the parameter**. It protected nobody who did not volunteer to be protected, and it refused exactly one caller: the one honest enough to name its own low-trust identity. Removing it loses no real protection and removes a 403 that the other surface never returned.

Callers who were legitimately authorised before are unaffected. There is no breaking change: the surface that becomes permissive was refusing calls MCP already accepted.

## Tests

Both behavioural tests were **confirmed failing on the parent commit**, not assumed:

- `test_rest_delete_byid_tenant_key_not_trust_gated` — fails without the fix with the exact 403 above (`403 != 204`).
- `test_rest_delete_byid_tenant_key_still_attributed_to_named_agent` — also fails against a **naive** fix that drops the param from attribution as well as authorization; asserted failure message `got None`. Verified by probe.

Two further tests are **labelled in-file as guards that pass both before and after**, so nobody later reads their green as evidence the fix worked: a trust-1 agent credential is still refused, and an agent credential still cannot borrow a trust-3 identity via the query param.

The suite also needed a new `as_tenant_key` fixture: `get_test_auth()` returns the **admin** key, which resolves to `tenant_id=None` and so cannot reach the gate's `if auth.tenant_id and …` at all. The credential that reaches the gate without an agent identity is a tenant key, and that is the shape the parity smoke measured.

## Verification

- Full root suite: **54 failed / 5913 passed**. Pristine `origin/main` baseline: **54 failed / 5909 passed** — identical failures, +4 being exactly the new tests. All 54 are pre-existing environment gaps (missing optional deps) in the local venv, none related to this change.
- `ruff check` and `ruff format --check` run separately at CI's exact scopes (`core-api/src/`, `tests/`) — clean.
- `legacy_name_ratchet.py --base origin/main` → *No new lines.* · `do_not_touch_sentinel.py` → *All 39 protected strings survive.* · `tenant_scope_gate.py` → 0 added/removed/recategorised. **All run after `git add`**, so the diff-based gates actually saw the files.

### DIFF
--- a/core-api/src/core_api/mcp_server.py
+++ b/core-api/src/core_api/mcp_server.py
@@ -1390,7 +1390,17 @@ async def caura_manage(
                 # Trust gate (>= 3), mirroring single delete / REST DELETE. Blocks
                 # the cross-fleet bulk-delete-by-id IDOR for sub-admin agents; a
                 # trust>=3 admin agent retains tenant-wide delete (parity with
-                # enforce_fleet_write). No-op for tenant-scoped credentials.
+                # enforce_fleet_write).
+                #
+                # ``caller_agent_id is None`` (a tenant-scoped credential) skips
+                # the gate BY DESIGN, not by omission — the trust ladder governs
+                # agent credentials and a tenant key holds no trust level. See
+                # ``enforce_delete``'s contract. Do NOT "fix" this by falling
+                # back to the ``agent_id`` tool argument: that value is
+                # caller-asserted, so it would let a tenant key pass the gate by
+                # naming any trust>=3 agent, and ``caller_agent_id`` is also the
+                # principal for the ``authorize_memory_access`` scope checks
+                # below and on read / update / lineage.
                 if caller_agent_id:
                     await enforce_delete(tenant_id, caller_agent_id)
                 # HOME-tenant scoped soft-delete: the storage method applies the
@@ -1618,6 +1628,12 @@ async def caura_manage(
                 result = await update_memory(uid, tenant_id, MemoryUpdate(**fields), agent_id=agent_id)
                 return _with_latency(_serialize(result), t0)
             # op == "delete" — WRITE → home tenant only.
+            #
+            # ``caller_agent_id is None`` (a tenant-scoped credential) means both
+            # checks below are skipped, and that is the decided behaviour rather
+            # than a gap: the trust ladder governs agent credentials, tenant
+            # scope governs tenant keys (see ``enforce_delete``). The same
+            # distinction already governs read / update / lineage above.
             if caller_agent_id:
                 # Trust gate (>= 3) + cross-fleet / scope_agent row authorization,
                 # mirroring REST DELETE /memories/{id}.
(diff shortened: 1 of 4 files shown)


Eldad-Caura:
@erni-a pleawe pprove

erni-a (approved):

Eldad-Caura:

```
```

Issue #1202, opened 2026-09-01
core-api: /whoami cannot gain trust_level until it verifies the gateway perimeter


Eldad-Caura:
Item 10 of the REST/MCP parity smoke report's queue: round out `/whoami` with
`trust_level`, `key_kind`, and `limits` — the fields the original P1
recommendation asked for that the shipped probe does not yet return.

Filing rather than implementing, because **the main field cannot be added
safely as things stand**, and the reason is worth writing down before someone
adds it in a quiet afternoon.

## Current state (re-verified on `main`)

`GET /api/v1/whoami` (`routes/health.py:38`) returns:

```json
{"tenant_id", "agent_id", "auth_source", "via_gateway",
 "readable_tenant_ids", "capabilities", "auth_mode"}
```

No `trust_level`, no `key_kind`, no `limits`.

## The blocker: `/whoami` has no perimeter

The handler takes `request: Request` and reads `x-tenant-id` / `x-agent-id`
directly. It has **no auth dependency and no gateway-shared-secret check** —
zero `Depends` in the whole module.

That is safe *today* for one reason only: the endpoint is a pure echo. You tell
it your tenant id, it tells you your tenant id back. Nothing is looked up, so
nothing can be disclosed.

`trust_level` breaks that property. It requires
`get_agent(agent_id, tenant_id)` against storage, keyed on **caller-supplied**
identifiers on an **unauthenticated** endpoint. Adding it naively would give
anyone on the internet:

1. **Cross-tenant information disclosure** — enumerate `(tenant_id, agent_id)`
   pairs and read back each one's trust level. Trust level is exactly the
   attribute that decides what an agent can reach, so it is a useful thing for
   an attacker to map.
2. **Unauthenticated storage load** — one DB round-trip per request on an open
   endpoint.

The MCP middleware guards precisely this shape and says so in a comment: the
header-trust path *"accepts X-Tenant-ID / X-Agent-ID / X-Readable-Tenant-IDs /
X-Capabilities with no credential of its own, so when a shared secret is
configured the request must prove it came through the gateway. A caller
reaching core-api's /mcp directly (e.g. its public run.app URL) must not be
able to impersonate a tenant by setting the identity headers itself."*

`/whoami` does not have that check. So this is a **latent trap**, not just a
missing feature: the endpoint is fine while it echoes, and becomes a
disclosure the moment anyone adds a lookup — which is what this very item asks
for.

## What each field needs

| Field | Status |
|---|---|
[...]

PR #1260, opened 2026-09-03
feat(whoami): report trust_level so callers stop probing to find out

Eldad-Caura:
Closes smoke finding **#4** — `/whoami` returned `key_kind` and nothing a caller could act on.

## Why this matters more than a missing field

A caller could not discover whether it was permitted to delete, could not predict a 403, and could not tell a permission refusal from an outage. The served skills state the consequence outright:

> *"There's no self-query tool, so don't pre-emptively avoid an operation you're unsure about: attempt it."*

Discovering your permissions by **attempting the operation** is defensible only while the answer is undiscoverable — and the operation in question includes `delete`.

It also matters more since **#1259**, which settled that the two credential regimes genuinely differ: an **agent** credential is governed by the trust ladder (`enforce_delete` requires ≥ 3), while a **tenant** key holds no trust level and is authorized by tenant scope. Those produce different delete outcomes, and nothing exposed which one you held.

## Why it did not ship alongside `key_kind` (#1202)

Every other field on this endpoint is an **echo** of what the caller sent. That is what makes a route with *no auth dependency* survivable. `trust_level` cannot be an echo — it is a storage lookup keyed on caller-supplied ids.

The perimeter check on that branch closes the identity spoof only where a gateway secret is **configured**. With none set (OSS self-hosted), `_gateway_verified` trusts the header path by design. So an unguarded lookup would let anyone read **any agent's attributes in any tenant**, by asking. The previous comment on the route said exactly this and deferred the field.

**The narrow answer:** resolve only when a gateway secret is configured. On that path the request has proved it came through the gateway, so the ids are the *gateway's* assertion rather than the caller's. Everywhere else the field degrades to `None` and nothing is looked up.

Worth noting: `test_whoami_still_looks_nothing_up` — the existing guard that says *"if someone adds a lookup here… the endpoint becomes a cross-tenant probe"* — **continues to pass untouched**, because it runs with no secret configured, which is precisely the case where the lookup declines to run. The property it pins is preserved rather than weakened.

## `trust_source`, and why `trust_level` alone is not enough

`trust_level: null` would otherwise be ambiguous across three genuinely different states:

| `trust_source` | Meaning | What the caller should do |
|---|---|---|
| `lookup` | Resolved; `trust_level` is authoritative | Compare against the operation's bar |
| `none` | No agent identity — a tenant key. The ladder **does not apply** | Stop looking for a level it will never have |
| `unregistered` | Looked up, agent absent | `enforce_delete` refuses this *before* any trust comparison |
| `unavailable` | Agent id present, not safely resolvable on this deployment | Cannot be determined here — not a permission answer |

Without this, a caller reading `null` is back to attempting a destructive operation to find out — the thing this change exists to stop.

## Follow-up, sequenced and recorded

This makes the *"there's no self-query tool"* passages in `static/skills/*/SKILL.md` and `plugin/skills/*/SKILL.md` **factually false**.

They are deliberately **not** corrected here. Another lane is mid-flight on a dual-path rename of those exact directories; correcting them now would fix the copy being *retired* while the new slug ships the stale advice from pre-fix content. Ordering agreed with that lane: **their change lands first, then all four copies are corrected in one follow-up PR.**

The window is acceptable — that advice is already the status quo, so this makes it stale rather than dangerous — but it is tracked, not overlooked.

## Tests

The security guard is **proven, not assumed**. `test_whoami_never_looks_up_without_a_configured_perimeter` replaces the lookup with a function that **raises**, so a lookup that happens cannot pass by returning a convenient value. Removing the guard fails it with:

```
AssertionError: cross-tenant probe: looked up 'someone-elses-agent' in
'someone-elses-tenant' with no gateway secret configured
```

Confirmed by probe, not inferred. Five further tests cover each `trust_source` state and shape consistency across every branch.

## Verification

- Full root suite: **54 failed / 5919 passed**. Post-#1259 `main` baseline is 54 failed / 5913 passed — identical failures, **+6 being exactly the new tests**. All 54 are pre-existing missing-optional-dep gaps in the local venv.
- `ruff check` and `ruff format --check` run **separately** at CI's exact scopes (`core-api/src/`, `tests/`) — clean.
- `legacy_name_ratchet.py --base origin/main` → *No new lines.* · `do_not_touch_sentinel.py` → *All 39 protected strings survive.* · `tenant_scope_gate.py` → 0 added/removed/recategorised. **All run after `git add`.**

One stale comment block in the test file (*"`trust_level` therefore remains NOT here — see #1202"*) is corrected in the same commit — leaving it would repeat #1259's own defect of fixing behaviour while the surrounding text asserts the old one.

### DIFF
--- a/core-api/src/core_api/routes/health.py
+++ b/core-api/src/core_api/routes/health.py
@@ -23,6 +23,7 @@ from core_api.providers._platform import (
     get_platform_llm,
 )
 from core_api.routes.plugin import _plugin_version
+from core_api.services.agent_service import lookup_agent
 from core_api.tools import REGISTRY  # SoT registry — populated at import time
 from core_api.version_compat import MIN_RECOMMENDED_PLUGIN_VERSION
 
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@claude re-review please — force-pushed to `8cc8f3ea` after applying your finding.

**Applied:** the unhandled storage failure in `_trust_fields`. Verified the mechanism before fixing rather than taking it on faith — `storage_client._get` calls `raise_for_status()` and `_execute` retries connection errors before giving up, so a storage 5xx or an unreachable storage really did propagate as a 500 out of a route whose whole design is to answer without touching storage. Wrapped in `try/except Exception` + `logger.exception`, degrading to `trust_source: "unavailable"` — matching the storage and event-bus probes already in the same file.

Two notes on the shape of the fix:

- `unavailable` is reused deliberately rather than adding a fourth state. Both this and the no-perimeter case mean "could not determine", which is precisely what `trust_source` exists to distinguish from `none` ("no trust level applies to me"). A caller must not read an outage as a permission answer.
- The 404 path is unaffected and still lands on `unregistered` — `_get` returns `None` for 404 rather than raising, so a genuinely absent agent is still reported as absent, not as an outage.

**Test:** `test_whoami_survives_a_storage_failure`, confirmed failing without the fix by reverting the `except` clause — uncaught `RuntimeError: storage-api unreachable`, 500 instead of 200. It also asserts the rest of the probe still answers, since degrading is only useful if the remaining fields survive.

Full suite re-run: 54 failed / 5920 passed, against a `main` baseline of 54 failed / 5913 passed — same failures, +7 exactly the new tests.

Eldad-Caura:
@claude re-review please — now at `e53e9913`, rebased onto `43cb230a` and carrying the timeout fix.

**Verified all four premises before applying, and they all hold:**
- `PROBE_TIMEOUT_SECONDS = 5.0` (`constants.py:155`), and its comment is exactly the rationale you cite — *"a stalled backend can't hang the whole probe."*
- `_probe_dependencies` does wrap its calls: `asyncio.wait_for(..., timeout=PROBE_TIMEOUT_SECONDS)` at `health.py:252` and `:266`.
- Storage client read timeout really is 120s — `httpx.Timeout(connect=5.0, read=120.0, write=120.0, pool=5.0)` at `storage_client.py:233` — with retry-on-transient above it.

So the finding is correct, and the sharpest part of it is the one I'd have missed: an error handler does not cover a backend that is **up but slow**. It never raises, so the `try/except` from the last round does nothing for it. Fixed with the `wait_for` bound.

**One part of the suggestion I did not apply, deliberately.** You asked to fold `asyncio.TimeoutError` into the existing `except Exception`. It is already caught — `asyncio.TimeoutError is builtins.TimeoutError` on this runtime, whose MRO is `TimeoutError → OSError → Exception`. Verified rather than assumed. Adding a separate clause would be dead code implying a gap that isn't there, so instead the comment states why no clause is needed. Behaviour is unchanged: a timeout degrades to `trust_source: "unavailable"` like any other lookup failure. Flagging it explicitly so it reads as answered rather than skipped.

**Test:** `test_whoami_does_not_hang_on_a_slow_backend` patches the lookup to sleep 30s and the bound down to 50ms, then asserts a 200 with `unavailable`. Confirmed failing without the wrapper by reverting it — the request never returns and the test's own 10s guard fires. That is the failure an error-injection test cannot produce, which is why it is a separate test rather than a tweak to the existing one.

**Note on the branch:** the remote had been rebased onto current `main` by someone else while I was working. I checked before force-pushing — `a06fb35d` was this same commit rebased, touching the same two files from the same parent, and my push is that content plus the 54 lines of timeout fix. Nothing was discarded.

Full suite: 54 failed / 5923 passed, against a same-base `main` baseline of 54 failed / 5915 — identical failures, +8 exactly the new tests. Naming gates re-run post-rebase and clean (the pre-rebase run had flagged a false positive caused purely by the stale base).

Eldad-Caura:
@claude re-review please — now at `a79e9b43`, rebased onto `4693c426`.

**Applied your Low suggestion, and it was worth more than its severity.** Verified the premise first: `_trust_fields` has exactly one call site, and it is safe today only because it sits inside `whoami`'s already-verified branch. `_trust_fields` now takes `request` and checks `_gateway_verified(request)` itself, so the guard is self-contained: `if not settings.gateway_shared_secret or not _gateway_verified(request)`. Both halves live in the function that depends on them rather than being inherited from where it happens to be called.

**Writing the test for that suggestion exposed a much worse problem, which is the real content of this push: the security guard had ZERO effective coverage.** I removed the guard entirely and ran the file — **all 20 tests passed.**

The cause is my own fix from the previous round. Every "must not look up" test used a stub that RAISES:

```python
async def exploded(tenant_id, agent_id):
    raise AssertionError("cross-tenant probe: ...")
```

Those were genuinely validated when written — I confirmed them failing without the guard. But then I added `except Exception` for the storage-failure finding, and that handler **swallows the AssertionError and returns `trust_source: "unavailable"`** — precisely the value each of those tests asserts. They went from real to decorative, silently, and I did not re-validate them after changing the code they exercised. Confirming a test fails without its fix is not a one-time gate; it expires when you change the error handling around it.

Fixed by asserting on **whether the lookup ran** rather than on the response, via a shared `_recording_lookup` helper that records calls instead of raising. Recording survives the error handler; raising cannot. Re-validated by deleting the guard again: now `test_whoami_never_looks_up_without_a_configured_perimeter` and `test_trust_fields_guards_itself_outside_the_verified_branch` both **fail**, which is the first time either has actually been load-bearing.

Two tests correctly still pass with the guard deleted — `..._on_an_unverified_gateway_claim` and `..._none_for_a_tenant_key` — because those paths return before reaching the lookup for independent reasons (`tenant_id` already nulled; no `agent_id`). That is accurate rather than a gap, and the helper's docstring records the whole trap so the next person does not rebuild it.

Full suite: 54 failed / 5924 passed, matching the `main` baseline's 54 failures with +9 exactly the new tests.

Eldad-Caura:
@erni-a pleaes pprove

erni-a (approved):

Eldad-Caura:

```
```

PR #1261, opened 2026-09-03
feat: dual-path serve the bundled and standalone skills under caura alongside memclaw

Eldad-Caura:
## What

Ships `plugin/skills/caura/` and `static/skills/caura/` as genuine,
independently-served copies of the existing `memclaw`-slug skills — per
Eldad's direct instruction (slug: `caura`) and the coordinator's approved
plan: **dual-path, not a flag-day cutover**. The two populations are
governed by unrelated mechanisms in different languages (`PROTECTED_SKILLS`
in `reconcile-skills.ts`; the Python `_SKILL_LABELS` dict in
`routes/plugin.py`), so landing both correctly in one window and having
each be independently verifiable is safer than one synchronized flip.
**Neither slug is dropped in this change.**

## The non-negotiable ordering constraint

`reconcile-skills.ts`'s reconciler ("owned" mode, which governs
`plugin/skills/` by default) deletes any on-disk skill slug that is
neither server-catalog-desired nor in `PROTECTED_SKILLS` — unconditionally,
every 60 seconds, no grace period. `"caura"` is added to that set in the
**same commit** that adds the directory. This isn't "land the allowlist
first as a safe step" — there is no safe "ahead of," only "together":
core-api serves plugin source and skill content from this identical
checkout, and the plugin's auto-upgrade path fetches all files (code +
root files) into memory and only writes to disk if the whole batch
succeeds, so an instance can never end up with the new allowlist and old
content, or vice versa.

## What changed

- `plugin/src/reconcile-skills.ts`: `PROTECTED_SKILLS` now protects both
  `"memclaw"` and `"caura"`.
- `core-api/src/core_api/routes/plugin.py`: `_plugin_root_files` gained
  `skills/caura/SKILL.md` (so auto-upgrade actually delivers it to
  *existing* installs, not just fresh ones); `_SKILL_LABELS` gained a
  `"caura"` key mapping to real `static/skills/caura/` content; the
  now-redundant fixed `/skill/caura` route (which just forwarded to
  `/skill/memclaw`'s handler) is **removed** — `"caura"` is now a real
  `_SKILL_FILES` entry served independently by the generic
  `/skill/{skill}` route.
- `plugin/src/heartbeat.ts`: `FALLBACK_ROOT_FILES` (the offline fallback
  used only when `/plugin-manifest` is unreachable) gained the new file too.
- `plugin/skills/caura/SKILL.md`, `static/skills/caura/SKILL.md`: copies
  of the `memclaw`-slug files. Content is identical except the handful of
  self-referential install-path lines. The plugin copy deliberately keeps
  `plugins.entries.memclaw.enabled` wherever it appears — that's the
  frozen plugin id (held separately, unrelated to this skill's own slug).
- `tests/test_install_skill_endpoint.py`: two new tests — `caura` installs
  to its own dirs like `company-brain` does, and `/skill/caura` serves
  content **distinct** from `/skill/memclaw` rather than forwarding to it
  (the thing a regression back to the old alias-only behavior would break).
- `docs/plans/skills-dual-path-transition.md`: new. Why `memclaw` isn't
  dropped yet, and the retirement condition — not a fixed date, but a
  named, checkable decision gate (a stable release + explicit owner
  sign-off), since the real constraint is adoption on installs this repo
  cannot measure.

The RPC-name dual-read alias (six gateway command names) is a separate
concern, in a separate PR.

## What this proves, and what it doesn't

The new in-process test proves the **code** is correct — `/skill/caura`
and `/skill/memclaw` genuinely serve different files, not one aliasing
the other. That is a stronger and more useful claim than "both paths work
today": it's the precondition for ever retiring `memclaw`. If `/skill/caura`
had shipped as a thin delegate to `skill_memclaw()` (as the route did before
this PR), dropping the `memclaw`-slug directory in the eventual retirement
change would silently break the `caura` path too — and it would break at
the exact moment everyone had already migrated onto it. This test is what
rules that out in advance, not just what confirms the current state.

It does **not** prove the deployed edge reaches this code path for an
authenticated caller: both `caura.ai` and `memclaw.net` edge-gate
`/api/v1/skill/*` with a `401` for an unauthenticated request today
(confirmed live, both domains, both slugs — matches a pre-existing
docstring noting the hosted deploy 401s under either name), and neither I
nor the coordinator has a credential to get past that gate.

**Post-merge verification owed, not yet done:**
- An authenticated request to `/api/v1/skill/caura` and
  `/api/v1/skill/memclaw`, confirming each serves its own distinct content.
- Confirming both `plugin/skills/{memclaw,caura}/` directories are still
  present a full heartbeat (60s+) after deploy — not just immediately.

## Sequencing note

`memclaw-caura-b3` is independently fixing a stale self-query claim in
both `memclaw`-slug `SKILL.md` files (`static:174`, `plugin:194`) in a
follow-up PR, correcting all four resulting files (both slugs × both
populations) in one commit once this merges — agreed order, so the
corrected text lands on the slug everyone migrates to, not only on the
one being retired.

## Verification

  lines. 0 annotated, 6 removed, 0 excused moves (-6 net).` EXIT 0
  survive.`
- `ruff format --check` / `ruff check` on `core-api/src/` and `tests/` → clean
- Real end-to-end: `alembic upgrade head` against a freshly-created
  `caura`-credentialed Postgres container → all 41 revisions clean
- `pytest tests/test_install_skill_endpoint.py` → 18/18 passed
- `pytest tests/test_api_plugin.py tests/test_f4_installer_api_url_default.py tests/test_plugin_source_manifest.py`
  → 40/40 passed (no regression from removing the old fixed alias route)
- `pytest tests/ -m "not benchmark"` → 5974 passed (full root suite)
- `npx tsc --noEmit` (plugin) → clean
- `npm test` (plugin) → 427/427 passed

### DIFF
--- a/core-api/src/core_api/routes/plugin.py
+++ b/core-api/src/core_api/routes/plugin.py
@@ -57,8 +57,14 @@ _plugin_files = [
 
 # Plugin-root-relative files served alongside the src/*.ts files.
 # - tools.json: tool SoT, loaded by tool-specs.ts.
-# - skills/memclaw/SKILL.md: shared plugin skill, discovered by OpenClaw
-#   via openclaw.plugin.json:skills (one copy per node).
+# - skills/memclaw/SKILL.md, skills/caura/SKILL.md: shared plugin skill.  # legacy-name-ok: dual-path skills transition, tracked for eventual retirement in docs/plans/skills-dual-path-transition.md
+#   Both are served during the rebrand transition (dual-path: "memclaw"  # legacy-name-ok: dual-path skills transition, tracked for eventual retirement in docs/plans/skills-dual-path-transition.md
+#   is the historical bundled skill already on disk on every existing
+#   install; "caura" is the new one). Discovered by OpenClaw via
+#   openclaw.plugin.json:skills (one copy per node). Removing either
+#   entry here without also updating reconcile-skills.ts's
+#   PROTECTED_SKILLS in the same change deletes that skill directory
+#   from every install within one heartbeat.
 # - openclaw.plugin.json: plugin manifest. Must be served (not baked into
 #   the install script as a HEREDOC) because the manifest changes
 #   periodically as OpenClaw evolves its plugin contract — e.g. the
@@ -70,6 +76,7 @@ _plugin_files = [
 _plugin_root_files = {
     "tools.json",
     "skills/memclaw/SKILL.md",
+    "skills/caura/SKILL.md",
     "openclaw.plugin.json",
 }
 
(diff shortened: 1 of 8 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #1263, opened 2026-09-03
feat: dual-read alias the six gateway RPC commands from memclaw.* to caura.*

Eldad-Caura:
## What

Registers `caura.status`, `caura.deploy`, `caura.deploy.status`,
`caura.educate`, `caura.allowlist.check`, and `caura.allowlist.fix` as the
canonical gateway command names, with the historical `memclaw.*` spellings
kept as a **dual-read alias** dispatching to the exact same handler
function — not a copy — via a small `registerGatewayMethodDualRead`
wrapper. `registerGatewayMethod` takes an exact string per call with no
shared prefix-stripping layer to hook a translation into (unlike the MCP
tool-name shim), so both names are registered explicitly.

## Why this is deferred, not permanent

Unlike the MCP tool rename (declared permanent because saved prompts,
keystone rules, and published tutorials are known to quote the old tool
names), a direct measurement found the opposite kind of evidence here —
not "safe to drop," but "impossible to know." I searched Datadog for these
six command name strings across all services, 30 days: **zero hits**.
Cross-checked against a working control query (bare `"memclaw"`, same
window: 91,099 hits on unrelated things) to rule out a broken search
rather than a real negative. The reason is architectural: `openclaw
gateway <command>` is a CLI-to-local-daemon call on the operator's own
machine — it never reaches Caura's network, so Caura's own telemetry has
no way to observe whether any external script or runbook still calls the
old names. **That's unmeasurable risk, not measured zero risk** — the
alias is cheap insurance against something this repo cannot rule out.

Retirement condition, in `docs/plans/gateway-rpc-dual-read-alias.md`:
drop the six `memclaw.*` registrations no earlier than **12 months from
this change**, and only after explicit owner sign-off — a named, checkable
decision gate rather than an open-ended "eventually," since there's no way
to swap in a measured condition for something architecturally invisible to
this repo.

## A ratchet-mechanics note, since it affects the marker choice

`legacy-name-deferred` (the marker matching this alias's actual, temporary
nature) only tracks **pre-existing** unannotated debt — it does not exempt
a brand-new line from the ratchet gate (confirmed empirically: reverting
the marker type on a genuinely new line flips the gate from pass to fail).
Every line this PR touches is new, so the code uses `legacy-name-ok`
mechanically, with the transitional reasoning stated in the reason text
and the real policy recorded in the linked doc — the same resolution used
earlier in this sweep for the interviewer's transitional constants.

## The marker says permanent; the policy says temporary — naming that tension rather than smoothing it over

`legacy-name-ok` asserts a name is permanently fine. The reason text on
these six lines, and `docs/plans/gateway-rpc-dual-read-alias.md`, say
something different: drop these in at least 12 months, with explicit
owner sign-off required. Those two statements disagree, and the marker is
the one a future sweep will actually read — this is worth stating plainly
rather than letting the marker type imply more permanence than is meant.

Structurally this is the same shape as a floor mark whose own reason text
says "not yet retired": a marker claiming permanence while the words next
to it say otherwise. The difference here is that it's deliberate and
documented, taken because the mechanically-correct marker for a brand-new
line (`legacy-name-ok`) carries no way to also say "and this one is
temporary" — the ratchet's vocabulary has exactly three markers, and only
one of them (`legacy-name-deferred`) means "temporary," but that one only
works on lines that already existed as unannotated debt before this
change. A dual-read alias — new lines, temporary intent — falls into the
gap between what the marker type must claim and what's actually true.
Filing an issue against that gap is a separate, later thing; this PR just
means to not paper over it.

**On "12 months" specifically, so it doesn't read as the vague
"eventually" this whole rebrand has been trying to eliminate**: the
duration is a policy floor precisely *because* no measurable condition is
available here — these six calls are architecturally invisible to this
repo's own telemetry, so "retire once usage drops to zero" is unanswerable
by construction, unlike e.g. a host route where request logs can actually
prove zero external callers. A fixed duration isn't laziness in this case;
it's the only honest form the expiry can take when the alternative
(waiting for evidence that can never arrive) isn't really an alternative.

## Verification

- New test file `gateway-methods.test.ts`: all six pairs are registered;
  critically, each pair shares the **exact same handler reference** rather
  than two independently-written copies that could silently drift apart;
  no unexpected seventh method leaks in.
- **Confirmed the test actually fails without the fix**: temporarily
  reverted the wrapper to register only the canonical name, reran — two
  of three new tests failed with the expected diff (6 registered vs. 12
  expected; legacy name `undefined`) — then restored the real fix and
  confirmed all three pass again.
  lines. 0 annotated, 6 removed, 0 excused moves (-6 net).` EXIT 0
  survive.`
- `npx tsc --noEmit` → clean
- `npm test` → 430/430 passed (427 baseline + 3 new)

### DIFF
--- a/docs/plans/gateway-rpc-dual-read-alias.md
+++ b/docs/plans/gateway-rpc-dual-read-alias.md
@@ -1,0 +1,57 @@
+# Gateway RPC dual-read alias: historical spellings → `caura.*`
+
+**Status:** in progress · **Scope:** the six OpenClaw gateway commands
+registered by the plugin (`plugin/src/index.ts`) · **Decided by:** Eldad,
+via the programme coordinator, 2026-09-04.
+
+## What this covers
+
+`caura.status`, `caura.deploy`, `caura.deploy.status`, `caura.educate`,
+`caura.allowlist.check`, and `caura.allowlist.fix` become the canonical
+gateway command names. The historical `memclaw.*` spellings keep <!-- legacy-name-ok: dual-read alias, tracked for eventual retirement in this doc -->
+dispatching to the exact same handlers — a dual-read alias, the same
+shape as the `caura_*`/`memclaw_*` MCP tool rename, but registered as two <!-- legacy-name-ok: dual-read alias, tracked for eventual retirement in this doc -->
+separate names rather than one shim, because `registerGatewayMethod`
+takes an exact string per registration with no prefix-stripping layer to
+hook into.
+
+## Why this is `legacy-name-deferred`, not a permanent alias
+
+Unlike the MCP tool-name shim (`core-api/src/core_api/mcp_server.py`,
+declared permanent because saved prompts, keystone rules, and published
+tutorials are known to quote the old tool names), nobody has established
+that the same is true for these six gateway commands. What is established
+is the opposite of "safe to drop today": a caller-side measurement (see
+the scoping analysis referenced below) found these calls are dispatched by
+OpenClaw's local gateway daemon over IPC on the operator's own machine —
+they never reach Caura's network, so Caura's own telemetry has **no way**
+to observe whether any external script or runbook still calls the old
+names. That is unmeasurable risk, not measured zero risk. The alias is
+cheap insurance against something this repository cannot rule out.
+
+## Retirement condition
+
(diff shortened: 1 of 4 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #1265, opened 2026-09-03
fix(ratchet): exclude CHANGELOG.md from the gate unconditionally

Eldad-Caura:
## Summary

Eldad's explicit decision: exclude `CHANGELOG.md` from the legacy-name gate unconditionally, not a rename-sweep cleanup. release-please regenerates each package's CHANGELOG from merged PR titles, and the fleet's convention names whatever a commit removes — including, when that is the change, the old brand itself. That made CHANGELOG.md a counted file charging the programme for RECORDING that it removed a legacy name: every release shipping rename work minted at least one gated line describing it, on a line no human wrote. Broke a six-cycle streak of `added = 0`.

## Root cause

The existing `release_please_changelogs` exemption (`_release_please_pull_request()`) only fires while CI is live inside the one authenticated release-please `pull_request` event itself — correct at that PR's own merge, and blind the moment anything re-measures outside it: a later status audit, a local run, or any other PR's base diff. The mechanism is scoped to a CI event, not to a fact about the file. A CHANGELOG is a dated historical record in every context, not only in one CI event, so the fix makes the exclusion unconditional and by basename (`:(exclude,glob,top)**/CHANGELOG.md`), wired into `_grep()` the same way `mirror_paths` already is — independent of CI/PR context entirely.

Deliberately not folded into `mirror_paths` (config-driven, exact paths only): a monorepo can carry one `CHANGELOG.md` per package, at unrelated paths — a literal path list would need a new entry every time a service gains its own, the same hand-maintenance drift this exclusion exists to remove, just moved to file-list granularity.

`_release_please_pull_request()` / `release_please_changelogs` are left in place, redundant rather than load-bearing — ripping them out is a separate, larger cleanup, same reasoning as keeping a CI-hash-check question separate: one policy change per PR.

## Test changes

Nine existing tests exercised the old event-scoped exemption by staging a `CHANGELOG.md` line and asserting `returncode == 1` in some untrusted context (wrong branch, wrong author, forked repo, malformed event, disabled flag). The new unconditional rule makes every one of those premises false, so eight are replaced with three tests asserting the new behavior as positive claims:

- `test_a_changelog_line_is_excluded_with_no_special_context` — the case this exclusion exists for: a plain local run, no bot, no PR.
- `test_the_exemption_survives_an_untrusted_author_and_branch` — every condition the old mechanism refused to trust, at once, still excluded.
- `test_a_changelog_passes_on_an_authenticated_release_please_pr` — non-regression: the original case still passes.

`test_the_exemption_is_path_scoped_to_changelogs` is kept (simplified to drop its now-unnecessary release-please env) — it still proves the exclusion doesn't leak to non-CHANGELOG files.

**Verified these actually guard, not just describe**: reverted the exclusion wiring and confirmed three of the four tests above go red (the fourth stays green correctly, since it only exercises the untouched, pre-existing authenticated path).

**Known gap, not fixed here**: `_release_please_pull_request()`'s own fail-closed/fork-proofing correctness (malformed event identity, untrusted author, forked head repo) is no longer observable through any test in this file, since a CHANGELOG.md line is now invisible to the scan before that function ever runs.

## Fleet context

This is the canonical copy in the legacy-name-ratchet parity set — `caura`, `caura-bus`, `caura-daemon`, `caura-enterprise`, `caura-onprem`, `caura-onprem-installer`, `caura-ops`, `caura-test-automation`, `openclaw-fleet-tester`. The other eight are gated on this one by the org's `Legacy-name engine parity` ruleset (checks each repo's copy against this repo's `main`), so they follow once this merges.

## Test plan

- [x] `ruff check scripts/` and `ruff check tests/` clean
- [x] `ruff format --check scripts/ tests/` clean (no reformatting)
- [x] `pytest tests/test_legacy_name_ratchet.py` — 182 passed (against a local Postgres matching CI's `TEST_DATABASE_URL`)
- [x] `python3 scripts/legacy_name_ratchet.py --base origin/main` — no new lines
- [x] Reverted the exclusion wiring and confirmed the three positive tests go red, then restored and confirmed green
- [x] DCO sign-off present

### DIFF
--- a/scripts/legacy_name_ratchet.py
+++ b/scripts/legacy_name_ratchet.py
@@ -787,6 +787,44 @@ def _included_mirrors(config: Config | None = None) -> tuple[str, ...]:
     return tuple(f":(top,literal){path}" for path in _mirror_paths(config))
 
 
+def _excluded_changelogs() -> tuple[str, ...]:
+    """Every ``CHANGELOG.md``, unconditionally excluded from the gate.
+
+    Eldad's explicit decision. release-please regenerates each package's
+    CHANGELOG from merged PR titles, and the fleet's convention names
+    whatever a commit removes — including, when that is the change, the
+    old brand itself. That makes CHANGELOG.md a counted file charging the
+    programme for RECORDING that it removed a legacy name — every release
+    shipping rename work mints at least one gated line describing it, on a
+    line no human wrote.
+
+    ``_release_please_pull_request`` below already exempts this, but only
+    while CI is live inside that one authenticated pull_request event —
+    correct at that PR's own merge, and blind afterward: a later status
+    audit, a local run, or any other PR's base diff re-measures outside
+    that event and the same line reads as a fresh mint, because the
+    mechanism is scoped to an event rather than to a fact about the file.
+    A CHANGELOG is a dated historical record in every context, not only in
+    one CI event, so this exclusion is unconditional and independent of
+    CI/PR context — a policy about what the file IS, not a widened
+    authentication check.
+
+    Deliberately not folded into ``mirror_paths`` (config-driven, exact
+    paths only): a monorepo can carry one CHANGELOG.md per package, at
+    unrelated paths, so a literal list would need a new entry every time a
+    service gains its own — the same hand-maintenance drift this exclusion
+    exists to remove, just moved to file-list granularity.
+
+    ``icase``: the old authenticated exemption matched the basename
+    case-insensitively (``str.upper().startswith("CHANGELOG")``); matching
+    the same way here means a differently-cased CHANGELOG (release-please
+    has always emitted the canonical casing in this fleet, but nothing
+    enforces that) is not a silent gap in what "unconditional" claims to
+    cover.
+    """
+    return (":(exclude,glob,top,icase)**/CHANGELOG.md",)
+
+
 def _grep(
     tree: str | None,
     pathspec: str | list[str] = ":/",
@@ -822,6 +860,7 @@ def _grep(
     args += [pathspec] if isinstance(pathspec, str) else pathspec
     if exclusions:
         args += _excluded_mirrors()
+        args += _excluded_changelogs()
 
     prefix = f"{tree}:" if tree is not None else ""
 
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
Addressing both findings:

**Issue 1 (Medium) — unconditional exclusion widens what bypasses the gate**

Correct description, and it's the intended, conscious change — not an oversight. This is Eldad's explicit decision (see the PR description and `_excluded_changelogs`'s docstring): the previous authenticated-PR-only scoping is what caused the failure being fixed here. It only ever exempted a CHANGELOG line while CI was live inside that one release-please `pull_request` event; the same line read as a fresh mint the moment anything re-measured outside it (a status audit, a local run, another PR's base diff) — which is exactly what broke a six-cycle streak of `added = 0` on a line no human wrote. A CHANGELOG is a dated historical record regardless of who or what CI event produced it, so the fix is deliberately unconditional rather than re-scoped.

Took the concrete suggestion though — added `head_repo_id=2` (a forked head repo) to `test_the_exemption_survives_an_untrusted_author_and_branch`, so that one test now exercises all four adversarial axes the removed tests covered individually (untrusted author, forked repo, ordinary branch, non-PR event) in one combined case, confirming they're deliberately no longer blocked. Verified by reverting the exclusion wiring and watching it (and two siblings) go red, then restoring.

**Issue 2 (Low) — dead authentication code left in place**

Also deliberate — kept per the same "one policy change per PR" reasoning as a separate CI-parity-hash-check question that's being raised with Eldad separately; ripping out `_release_please_pull_request()` / `release_please_changelogs` is a bigger, separate cleanup. Already documented in `_excluded_changelogs`'s docstring and this PR's description, but you're right that the dead code's own location didn't say so — added a short pointer comment directly on `_release_please_pull_request()` and at the `release_pull_request = ...` call site in `main()`, both cross-referencing `_excluded_changelogs` for why.

Both changes pushed. Ratchet gate still clean (no new lines), full `tests/test_legacy_name_ratchet.py` still 182 passed.

Eldad-Caura:
@claude

Eldad-Caura:
This is a repeat pass over the two points already addressed in the [prior review reply](https://github.com/caura-ai/caura/pull/1265#issuecomment-5533642584) — same answer stands: both are the conscious, documented trade-off behind this PR, not oversights (see the PR description and `_excluded_changelogs`'s docstring for the root-cause reasoning on why the previous authenticated scoping is what broke the six-cycle streak this fixes).

The **Suggestions** point is new and correct, though — pushed a fix:

**Case sensitivity**: the old `_release_please_pull_request`-gated check matched `path.rsplit("/", 1)[-1].upper().startswith("CHANGELOG")` — case-insensitive. My new pathspec (`:(exclude,glob,top)**/CHANGELOG.md`) was exact-case only, a real (if currently theoretical — every CHANGELOG.md in this fleet is canonically cased today) gap between what "unconditional" claims and what it actually covered. Added the `icase` pathspec magic word (`:(exclude,glob,top,icase)**/CHANGELOG.md`), verified in an isolated scratch repo that it correctly catches `changelog.md` and `CHANGELOG.MD` while still leaving a same-prefix-but-different file (e.g. `changelog.md.bak`) counted, and added `test_the_exemption_matches_the_basename_case_insensitively` to guard it — confirmed by reverting and watching it (plus the three existing guard tests) go red.

Ratchet gate still clean, full `tests/test_legacy_name_ratchet.py` now 183 passed.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #1266, opened 2026-09-03
fix(mcp): enforce CAURA_API_KEY on /mcp before the header-trust path

Eldad-Caura:
Closes audit findings **H-01** and **H-13** — one defect with two faces, merged per the agreed dedup so a docs-only fix could not document a live auth hole as intended behaviour.

## The defect

`MCPAuthMiddleware` never read `settings.memclaw_api_key`. `/mcp` is a **separate ASGI mount** with its own auth middleware, not a route behind `get_auth_context`, so REST's Path 2 does not cover it — and nothing else did either.

That matters because the rest of the system treats `CAURA_API_KEY` as a *sufficient perimeter* on the strength of that path. `app.py`'s production boot guard says so in as many words:

> when it is set, `auth.py`'s "Path 2" either authenticates the request against that key or raises 401 for everything else, so "Path 4" below it is **UNREACHABLE** and there is no header-trust surface left to protect

`docs/api-reference.md` makes the same promise. On `/mcp`, neither was true.

## The security consequence

In the exact configuration the boot guard blesses — `CAURA_API_KEY` set, no gateway:

| Caller | Old behaviour |
|---|---|
| keyless, sends only `X-Tenant-ID: victim` | **believed verbatim** → `caura_write` / `caura_recall` / `caura_manage` across that tenant. Unauthenticated cross-tenant read, write and delete. |
| keyless, standalone deployment | `elif settings.is_standalone` fired **before any key comparison** → full standalone tenant to every anonymous caller |
| correct key | worked, but via header-trust — the key was never compared |

The second row is the sharp one: setting `CAURA_API_KEY` *precisely in order to expose the service to a network* was what granted anonymous access.

## Order is the fix, not the comparison

The gate runs **before** the header-trust path, mirroring `auth.py` Path 2 sitting above Path 4. Adding a key check *after* header-trust would have left the hole exactly as it was, since `X-Tenant-ID` was consulted first.

**Deliberately not a faithful reordering of the whole chain.** REST also puts standalone above header-trust; MCP does the reverse, and that difference has its own test (`test_tenant_header_honored_when_secret_unset`) pinning intentional behaviour. Reordering those too would be an unrelated behaviour change riding along with a security fix, so only Path 2 was inserted.

**With `CAURA_API_KEY` unset — enterprise (gateway secret) and plain standalone, i.e. every deployment that exists today — this middleware is unchanged in behaviour.** There is a test pinning exactly that, so the new gate is inert for them rather than a change everyone absorbs.

Path 1 (admin key) moved ahead of header-trust to match REST, so an operator holding the admin key is not refused by the shared-key comparison on a deployment configuring both.

## One trap worth naming

`via_gateway` was derived downstream as `bool(tenant_header)`. A shared-key caller may legitimately send `X-Tenant-ID`, so that derivation would have **silently re-honored `X-Agent-ID`** for it — handing a tenant-wide key any agent's identity, which since #1259 is the authorization principal for scope checks *and* the delete trust gate. It now uses the value the resolution decided; only Path 4 sets it `True`, and a test pins that a shared-key holder cannot self-assert an agent id, capability set, or readable-tenant set.

## The docs half (H-13)

`api-reference.md` already **claimed** the shared-key path authenticates first and prevents header-trust from being reached. That claim was true for REST and false for MCP — so the resolution is the code, not a correction. One sentence is added making the `/mcp` coverage explicit, since that was the silently untrue part. `AGENT-INSTALL.md`, which tells operators to point MCP at the shared gate key, likewise becomes accurate rather than aspirational.

## Where the finding was wrong

H-01's *"valid key rejected in shared-gate mode"* half does not hold up. It claims a correct key falls to `_UNAUTH`; that is true only for a caller sending **no** `X-Tenant-ID`, and such a caller still resolves no tenant after this change — there is nothing to resolve. The documented client sends both headers and always worked, by accident, via header-trust. (H-13's own verifier note caught this: *"the documented client also sends X-Tenant-ID and would succeed on MCP."*) **The security half is the real defect**, and it is understated rather than overstated.

## Tests

**Five fail without the fix**, confirmed by reverting `mcp_server.py` to `origin/main` and re-running:

```
test_shared_key_gate_refuses_a_keyless_tenant_header
test_shared_key_gate_refuses_a_keyless_caller_in_standalone
test_shared_key_gate_refuses_a_wrong_key
test_shared_key_bearer_form_is_actually_verified
test_shared_key_holder_cannot_self_assert_an_agent_identity
```

**Two are labelled in-file as guards that pass either way** — the documented client reached the app before this change too, so they pin the accept path, not the gate. Labelled explicitly so nobody later reads their green as proof the fix works. That distinction is why the wrong-Bearer test exists separately from the accepts-Bearer one.

The new tests also set `is_standalone` explicitly. The suite runs with `IS_STANDALONE=true`, which silently decides several of these branches — the same class of trap as the admin-key default that would have made an earlier PR's tests vacuous.

## Verification

- Full root suite: **54 failed / 5924 passed**. `main` baseline is 54 failed / 5915 passed — identical failures, **+9 exactly the new tests**. The 54 are pre-existing missing-optional-dep gaps in the local venv.
- **Legacy-name ratchet initially failed** — legitimately, not a stale base: reading `settings.memclaw_api_key` adds the legacy name. That field is the *resolved dual-read* value (`CAURA_API_KEY` falling back to the legacy env var) and is the one `auth.py` Path 2 reads, so the gate must read the same one — reading `caura_api_key` directly would silently skip the gate for deployments still setting the old variable, which is the population this most needs to protect. Marked `legacy-name-ok: rule 3 dual-read field` on the source line, and the test file's nine references collapsed into a single named constant carrying one marker. Ratchet now reports *No new lines.*
- `do_not_touch_sentinel.py` → *All 39 protected strings survive.* Both gates run **after `git add`**.

### DIFF
--- a/core-api/src/core_api/mcp_server.py
+++ b/core-api/src/core_api/mcp_server.py
@@ -225,6 +225,27 @@ _ADMIN_ERROR = _as_error_result(
 )
 
 
+async def _send_401(send: Send, message: str) -> None:
+    """Refuse the request outright with a JSON error envelope.
+
+    Refusing beats falling through to ``_UNAUTH``: a request that keeps going
+    with a different identity resolution lets a caller probe which headers
+    change the answer.
+    """
+    body = _error_response("UNAUTHORIZED", message).encode()
+    await send(
+        {
+            "type": "http.response.start",
+            "status": 401,
+            "headers": [
+                (b"content-type", b"application/json"),
+                (b"content-length", str(len(body)).encode()),
+            ],
+        }
+    )
+    await send({"type": "http.response.body", "body": body})
+
+
 class MCPAuthMiddleware:
     """ASGI middleware that resolves X-API-Key to tenant_id before MCP handlers run."""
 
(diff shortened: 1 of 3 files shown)


Eldad-Caura:
@erni-a please apprvoe

erni-a (approved):

Eldad-Caura:
@erni-a ready for merge.

All 13 checks green on head `75fda7d4` (verified by SHA via `commits/<sha>/check-runs`, not `gh pr checks`). Review verdict: **✅ No issues found** — and it independently traced the two things most worth checking: that `via_gateway` is derived from the resolution path rather than header presence (so a shared-key holder cannot self-assert `X-Agent-ID`), and that the one remaining divergence from REST's ordering cannot be leveraged for escalation.

**What this closes:** unauthenticated cross-tenant read/write/delete on `/mcp` in the configuration `app.py`'s production boot guard explicitly blesses (`CAURA_API_KEY` set, no gateway). Audit findings H-01 and H-13, merged as one PR because a docs-only fix would have documented a live auth hole as intended behaviour.

**Two things to know before merging:**

1. **No behaviour change for any deployment that exists today.** With `CAURA_API_KEY` unset — enterprise (gateway secret) and plain standalone — this middleware is unchanged, pinned by `test_shared_key_unset_leaves_every_path_unchanged` rather than asserted. The new gate is inert unless the key is configured.

2. **Post-deploy verification is still owed and I cannot run it.** The finding was measured against a running service, so the honest status after merge is *fixed in code, unverified in production*. The check is cheap: on a deployment with `CAURA_API_KEY` set and no gateway, `POST /mcp` carrying only `X-Tenant-ID` should now return **401** where it previously returned tool results. It needs a credential I do not have and will not mint.

Not merging this myself — that stays your call.

erni-a (approved):

Eldad-Caura:

```
```

PR #1267, opened 2026-09-04
fix(search): gate every requested fleet, not just the single-fleet case

Eldad-Caura:
Closes audit finding **H-05** — cross-fleet content disclosure on `/search`, `/recall` and MCP `caura_recall`.

## The defect

The cross-fleet trust ladder fired only when `fleet_ids` had **exactly one** element:

```python
if not body.fleet_ids and _agent.fleet_id and trust < 2:
    body.fleet_ids = [own]          # forcing: only when NONE given
if body.fleet_ids and len(body.fleet_ids) == 1:
    await enforce_fleet_read(...)   # ladder: only when EXACTLY one
```

A trust-1 agent sending `fleet_ids=["victim-fleet", "own-fleet"]` satisfied **neither** branch — the forcing is skipped because the list is non-empty, the ladder is skipped because the length is 2. The unchecked list went straight to the storage predicate, which admits `fleet_id IN (...)` and returns `scope_team` rows **with full content** for every fleet listed.

| Request from a trust-1 agent | Before |
|---|---|
| `fleet_ids: ["victim"]` | **403** `fleet-scope policy…` |
| `fleet_ids: ["victim", "own"]` | **200 — victim's memory content returned** |
| `fleet_ids: ["victim", "victim"]` | **200 — same leak**, length 2 skips the gate, storage dedupes it back |

**Asking for MORE was the way to be asked for LESS.** This is direct content disclosure, not id harvesting — the test reproduces it by asserting the victim's memory text appears verbatim in the pre-fix response body, and the assertion output confirms it does.

## The fix

`enforce_fleet_read_many` checks **every** requested fleet against **one** `lookup_agent` call, rather than one lookup per fleet — these are the search hot path, and a per-element call would have turned a widened `fleet_ids` into a storage-call amplifier.

`enforce_fleet_read` now delegates to it, so the single- and multi-fleet ladders cannot drift apart — the same reasoning the codebase already applies to `resolve_read_fleet_gate` (*"All four share this one implementation so the ladder cannot drift between the surfaces"*). One-element semantics are unchanged.

Applied to all three sites the finding names (`memories.py` `/search` and `/recall`, `mcp_server.py` `caura_recall`) so the surfaces stay in parity.

## What I deliberately did not do

The finding offered an alternative: *"or reject multi-fleet reads for trust<2 agents"*. **That would close the hole by breaking a legitimate caller.** A trust-1 agent may name its own fleet — twice, if it likes — and a trust-2 agent may still read across fleets, because a cross-fleet read is a *privilege*, not a forbidden operation. Both are pinned by tests, because a gate that over-refuses is a different bug rather than a safe default.

## Tests

**Three fail without the fix**, confirmed by reverting all three source files to `origin/main` and re-running:

```
test_search_cannot_widen_by_naming_a_second_fleet
test_search_cannot_widen_by_repeating_the_same_fleet
test_recall_cannot_widen_by_naming_a_second_fleet
```

**Three are guards** that pass either way, labelled as such in-file: the single-fleet refusal (already gated pre-fix, present so the pair reads as a comparison), own-fleet-listed-twice, and trust-2 cross-fleet. The last two are the over-refusal guards described above.

The leak test asserts on the **content**, not only the status code, so a future change that answered `200` with filtered results instead of `403` would still have to not leak.

**A trap worth recording:** the suite's `get_test_auth()` returns the ADMIN key, which resolves to `tenant_id=None` — and this entire gate sits behind `if auth.tenant_id:  # skip for admin`. A test written with the admin key cannot reach the gate at all and would pass before and after any change. These tests authenticate as a real agent identity instead. That is the same shape as the admin-key trap in #1259 and the swallowed-assertion trap in #1260; three different fixes, one recurring failure mode.

## Verification

- Full root suite: **54 failed / 5939 passed**. `main` baseline is 54 failed / 5933 passed — identical failures, **+6 exactly the new tests**. The 54 are pre-existing missing-optional-dep gaps in the local venv.
- `legacy_name_ratchet.py` → *No new lines.* · `do_not_touch_sentinel.py` → *All 39 protected strings survive.* · `tenant_scope_gate.py` → 0 added/removed/recategorised. All run **after `git add`**, and re-run after rebasing onto current `main` (the pre-rebase run flagged two removed exempt lines, which was purely the stale base seeing #1266's markers as deletions).

### DIFF
--- a/core-api/src/core_api/mcp_server.py
+++ b/core-api/src/core_api/mcp_server.py
@@ -52,7 +52,7 @@ from core_api.schemas import (
 from core_api.services.agent_service import (
     authorize_memory_access,
     enforce_delete,
-    enforce_fleet_read,
+    enforce_fleet_read_many,
     enforce_fleet_write,
     get_or_create_agent,
     resolve_write_agent,
@@ -969,8 +969,11 @@ async def caura_recall(
             ag_trust = _ag.get("trust_level", 0)
             if ag_fleet and ag_trust < 2:
                 fleet_ids = [ag_fleet]
-        if fleet_ids and len(fleet_ids) == 1:
-            await enforce_fleet_read(tenant_id, agent_id, fleet_ids[0])
+        # EVERY requested fleet — gating only the single-fleet case let a
+        # trust-1 agent widen its read by naming a second fleet (parity with
+        # REST /search and /recall).
+        if fleet_ids:
+            await enforce_fleet_read_many(tenant_id, agent_id, fleet_ids)
         # Cross-tenant recall widens via readable_tenant_ids when the caller
         # authenticated with a cross-tenant credential (kind=cross_tenant) — the
         # gateway plumbs ``X-Readable-Tenant-IDs`` and the MCP middleware parks
--- a/core-api/src/core_api/routes/memories.py
+++ b/core-api/src/core_api/routes/memories.py
@@ -68,6 +68,7 @@ from core_api.services.agent_service import (
     broker_owned_agent_id,
     enforce_delete,
     enforce_fleet_read,
+    enforce_fleet_read_many,
     enforce_fleet_write,
     get_or_create_agent,
     lookup_agent,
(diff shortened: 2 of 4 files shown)


Eldad-Caura:
@erni-a plese approve

erni-a (approved):

Eldad-Caura:

```
```

PR #1268, opened 2026-09-04
fix(recall): resolve the read identity by the same rule as /search

Eldad-Caura:
Closes **H-06** (high) and **M-30** (medium) — merged per the agreed dedup, because one shared resolver is the fix for both.

## Two findings, one defect

`recall_endpoint` re-derived the read identity itself, straight from `body.filter_agent_id`, instead of using the rule `/search` applies.

### H-06 — disclosure and gate evasion

`/search` refuses an agent credential that names a peer. `/recall` had no such check, and fed the raw `filter_agent_id` into **three** places at once: `get_or_create_agent`, the trust<2 fleet forcing, and `caller_agent_id` — which *is* the visibility identity.

So agent-A (trust 1, fleet F1) could `POST {filter_agent_id: "agent-B"}` where B is trust 3 in F2, and receive:

- **B's private `scope_agent` rows, with full content**, in the results and in the LLM summary — reproduced; the pre-fix response body contains the row verbatim;
- **a tenant-wide scope**, because the fleet forcing keys off *B's* trust level rather than A's, so naming a trust≥2 peer skipped the forcing entirely.

### M-30 — silent drop

`SearchRequest.caller_agent_id` is documented as *"assert the agent identity this search runs as"*. `/search` honours it; `/recall` hardwired `caller_agent_id=body.filter_agent_id` and **never read the field**. The same body returned different rows on the two routes, with no error — and both SDKs forward the field verbatim, so it no-op'd silently.

## The guard is extracted, not copied

`_resolve_read_identity` now holds the two 403s and the precedence rule, and both routes call it.

Copying it would have left exactly the drift that produced this: `/search` grew the guard in a prior audit and `/recall` was not updated, so the two routes disagreed about the same request shape for a release. Same reasoning the codebase already applies to `resolve_read_fleet_gate` and to `enforce_fleet_read_many` from #1267.

## One coupling worth naming

Honouring `caller_agent_id` on `/recall` also means carrying `/search`'s `allow_recall_bump` gate. Without it, an **asserted** identity would start moving `recall_count` on this route — reshuffling ranking for every other caller in the tenant, which is precisely what that gate exists to prevent.

Fixing a dropped knob by giving it a side effect the sibling route deliberately suppresses would not be a fix.

## Rebase note — this PR interacts with #1267

#1267 changed the same `/recall` line to `enforce_fleet_read_many`. The rebase conflicted there, as expected, and the resolution takes **both** halves:

```python
if body.fleet_ids:
    await enforce_fleet_read_many(body.tenant_id, eff_agent_id, body.fleet_ids)
```

They are independent and both required. Gating every fleet against a *spoofable principal* would still let an agent inherit a peer's trust level; gating the *right principal* on only the single-fleet case would still let it widen by naming two.

**Re-validated after the rebase**, not assumed: against post-#1267 `main`, this PR's three tests still fail and #1267's six now pass from `main`.

## Tests

**Three fail without the fix:**

```
test_recall_refuses_a_filter_agent_id_that_is_not_the_caller
test_recall_refuses_a_caller_agent_id_that_is_not_the_caller
test_recall_honours_caller_agent_id_as_the_visibility_identity
```

- The disclosure test asserts on the **leaked content**, not only the status code, so a later change answering 200-with-filtering instead of 403 would still have to not leak.
- The second covers the same escalation under the *other* field name — guarding only `filter_agent_id` would have left the hole open by a new spelling.
- The M-30 test asserts a **parity claim** — `/search` and `/recall` result sets for an identical body — rather than a hardcoded expectation, so it states the contract the finding is actually about. It also asserts the fixture produced rows at all, so it cannot pass vacuously by comparing two empty sets.

**Two are over-refusal guards:** an agent naming *itself* still works, and a tenant-scoped credential may still name any agent. Closing the hole by refusing every named agent would break the dashboard and every tenant-key integration.

As in #1259 / #1267, the tests authenticate as a real agent identity — `get_test_auth()` returns the ADMIN key, which resolves to `tenant_id=None`, and both the gate block (`if auth.tenant_id:`) and the spoof guard (`auth.agent_id`) are invisible to it. A test written with the admin key would pass before and after.

## Verification

- Full root suite: **54 failed / 5944 passed**. `main` baseline is 54 failed / 5939 passed — identical failures, **+5 exactly the new tests**. The 54 are pre-existing missing-optional-dep gaps in the local venv.
- `legacy_name_ratchet.py` → *No new lines.* · `do_not_touch_sentinel.py` → *All 39 protected strings survive.* · `tenant_scope_gate.py` → 0 added/removed/recategorised. All run **after `git add`** and again after the rebase.

### DIFF
--- a/core-api/src/core_api/routes/memories.py
+++ b/core-api/src/core_api/routes/memories.py
@@ -1714,6 +1714,69 @@ async def update_memory_endpoint(
     )
 
 
+def _resolve_read_identity(auth: AuthContext, body: SearchRequest) -> tuple[str | None, bool]:
+    """Resolve the identity a search-shaped read runs as, refusing a spoof.
+
+    Returns ``(eff_agent_id, identity_asserted)``.
+
+    An agent credential may only name ITSELF. ``eff_agent_id`` is body-first,
+    and it feeds BOTH the visibility identity (which ``scope_agent`` rows are
+    readable) AND the trust/fleet enforcement at the callsite, so a
+    caller-asserted id bought two escalations at once:
+
+      1. Disclosure — naming a peer made that peer the visibility identity, so
+         its scope_agent memories (and private STM notes surfaced through
+         search) came back with full content, not just ids.
+      2. Gate evasion — the trust < 2 fleet forcing runs against this same id,
+         so naming a trust-3 peer skipped the forcing entirely. That is the
+         identical escalation /memories/redistribute was fixed for in the
+         2026-06-11 audit, which ran its trust gate against a caller-supplied
+         agent_id.
+
+    A tenant/user credential (``auth.agent_id`` is None) is unaffected and may
+    still name any agent — the restriction is on agent-scoped credentials,
+    which is where the threat is.
+
+    SHARED because ``/recall`` shipped without any of it. It re-derived the
+    identity itself, straight from ``body.filter_agent_id``, so both
+    escalations above were live on that route while ``/search`` refused them —
+    and ``caller_agent_id``, which the schema documents as the way to assert an
+    identity, was silently dropped there. Two routes parsing the same
+    ``SearchRequest`` must not disagree about what the fields mean; one
+    implementation is the only way that stays true.
+    """
+    if auth.agent_id and body.filter_agent_id and body.filter_agent_id != auth.agent_id:
+        raise HTTPException(
+            status_code=403,
+            detail=(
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

Eldad-Caura:

```
```

PR #1269, opened 2026-09-04
fix(skills): gate the STORED status, not just the written one

Eldad-Caura:
Closes audit finding **H-11** — integrity/availability attack on curated active skills.

## The defect

`validate_and_normalize_skill_write` applied RBAC only to the `status` and `source` strings **in the incoming body**. It never looked at the row the write was about to replace — and `document_upsert` is `on_conflict_do_update` over the whole blob, so a write aimed at an existing `doc_id` is a **full replace**.

With `skills_factory` enabled, any tenant agent holding a writable key could:

```
POST /documents  collection='skills'  doc_id=<a live ACTIVE skill>
                 kind='create'  source='agent'    (status omitted)
```

The validator defaults `status` to `staged` and passes every RBAC check — none of them are about the stored row — and the upsert replaces it. The active skill drops out of every `status='active'` surface (recall, harness reconciliation, MCP skills counts), its curated content is destroyed, and the replacement sits in the inbox looking like a routine staged submission.

**No privileged status was needed to retire a skill. The default one did it.**

## The module already named this threat

`SYSTEM_ONLY_STATUSES`' comment says letting an arbitrary write set those states *"would let a misbehaving agent silently retire an active skill or hide a candidate from review"* — while the check it annotates **could not see the row it was protecting**. Rejected and quarantined docs were equally resurrectable, so a rejection did not hold either.

## The fix is the same RBAC, applied to the stored value

```python
PROTECTED_LIVE_STATUSES = ADMIN_ONLY_STATUSES | INTERNAL_ONLY_STATUSES | SYSTEM_ONLY_STATUSES
```

The invariant: **a status a caller could not have WRITTEN is one they may not silently OVERWRITE.**

Derived rather than spelled out, so a status added to any RBAC set is protected on the same commit. A hand-maintained copy would be two places that have to agree — which is the shape of the original bug. A test pins the derivation.

`staged` is deliberately absent: an agent iterating on its own not-yet-approved skill is the ordinary authoring loop and must keep working. "Its own" is enforced, not just asserted — see the ownership half below. Admin and the internal Forge worker are exempt — they are the legitimate writers of these states, and the Inbox edit path is admin-gated before it reaches here.

**Beyond the finding:** it listed active/rejected/quarantined/stale/deprecated. Deriving the set from the RBAC constants also covers `candidate` — the other half of the comment's own threat statement ("hide a candidate from review"), which the finding omitted.

## Both surfaces now fetch the live doc for every skills write

Not just `kind='update'`. Fetching only on update is precisely what made `kind='create'` the way around the gate: the validator was handed `None` and judged the request purely on what the caller claimed. One extra read on a HITL-gated authoring path, not a hot path.

Both paths now fail **closed** on a fetch error — see the review round below.

## Review round: why `kind='update'` is NOT exempted

The review was right that the gate runs before `kind='update'`'s hash-binding, and that the old 403 message pointed callers at `kind='update'` as the remedy — a remedy this gate holds shut. **But its suggested fix (exempt `kind='update'`, since 7a already enforces hash-binding) would reopen H-11**, so I took the other branch it offered and corrected the message instead.

Hash-binding is **optimistic concurrency, not authorization**. It proves the caller read the current content, and reads are open to any tenant agent. So the exemption is defeated by one extra GET:

1. `GET` the live active skill — permitted — and copy its `content_hash`.
2. `POST` `kind='update'` with that `target_content_hash`. 7a is satisfied.
3. `status` is omitted, so it defaults to `staged`; a non-admin cannot write `active` (that arm is admin-only).
4. The write still lands on the same `doc_id` as a full replace.

Identical outcome to the original attack. I confirmed this rather than arguing it — applying the suggested exemption verbatim and re-running, the overwrite **succeeds**:

```
FAILED test_correct_hash_binding_does_not_buy_an_overwrite
  - Failed: DID NOT RAISE <class 'fastapi.exceptions.HTTPException'>
```

That test is new in this round and exists to make the regression loud if anyone tries the exemption again. A comment at the gate records the reasoning so the next reader doesn't have to re-derive it.

**On the functional concern:** there is no working non-admin revision path being removed here, because there was never a correct one. Pre-PR, a non-admin `kind='update'` against an active skill *demoted it to staged* — no status is preserved from the live doc, and `active` is admin-only to write. That demotion is the bug, not a feature. Agent-driven revision of a live skill needs the **Phase 4 update-proposal flow** the module docstring already anticipates, where a revision lands somewhere other than on top of the live row. Until then an admin makes the change, and the message now says exactly that. A second new test asserts the message no longer advertises the unreachable remedy.

`kind='update'` is *not* blocked as such — only against a protected live status. A third new test covers an agent hash-bound-updating its own `staged` draft, which still works.

## REST fail-closed parity

The review was also right that the REST fetch lacked the `try/except` its MCP sibling has. Fixed — it now logs with the tenant and doc and returns a curated 503.

One correction to the framing: this was **not** a fail-open. There is no enclosing handler in that route (verified), so an unhandled error already aborted the write with a bare 500 — the security property held on both paths. What differed was the error quality and the missing log line. Worth fixing for parity on a security-relevant fetch; not a second vulnerability.

Two tests pin the fail-closed behaviour on the **create** path, one per surface — the path this PR newly made fetch, and the security-relevant one. (`test_skill_update_write_fails_closed_on_live_doc_fetch_error` already covered the update path; I initially reported no such test existed, having grepped for the message text rather than the error code.) Both new tests fail if the handler is changed to swallow the error and continue with `live_doc=None`, which is the failure mode that actually matters.

## Second review round: the `staged` exemption now checks ownership

The re-review accepted the resolution above and raised one more Medium, which was fair and landed on a comment **I** wrote. The comment justified excluding `staged` on the grounds that "an agent iterating on its own not-yet-approved skill is the ordinary case" — but nothing compared the live row's `origin.agent_id` to the caller. So the exemption actually read *"anyone may overwrite any staged draft"*: agent B points a write at agent A's slug and A's unreviewed work is gone, with no history and nothing in the inbox to show it happened.

That is the same shape as the bug this PR fixes — a comment asserting a property no check enforces — so I took the review's option (a) and enforced it rather than documenting the gap away.

The gate now has two halves, and the comment says which is which. Refusal is limited to a live row naming a **different** owner:

- an **unattributed** draft (no `origin`, or the `unknown` stamp) stays writable on purpose. Deployments with no agent identity stamp every draft that way, and gating on it would 403 the ordinary loop for all of them.
- where the row **does** name an owner, a caller who cannot show that identity — including one with no `agent_id` at all — is not it. The looser "refuse only when both sides are attributed" reading would let an unbound tenant key overwrite any named author's draft.

`UNATTRIBUTED_AGENT_ID` is shared with the stamping site so "unattributed" cannot come to mean two different things in the two places, and a test pins that it is the value the stamp writes.

**On the TOCTOU suggestion:** added as an explicit `ACCEPTED LIMITATION` note at the gate. The fetch and the upsert are not in one transaction, which is the same optimistic-concurrency model hash-binding already runs on. The note exists so this gate is not later mistaken for serialization when it is authorization — the same distinction that settled the `kind='update'` question above.

Confirmed by disabling only the ownership condition: exactly the two new refusals fail, and all four over-refusal guards still pass.

## Third round: the refusal no longer names the owning agent

Verdict was **no blocking issues**, with two suggestions. I acted on the first and left the second, which the review itself said not to restructure.

The ownership 403 included the other agent's id. The review flagged this as low-risk-but-consider; checking the read path made it clearly worth fixing, because **the codebase already treats authorship as privileged one module over**: `op=read` on a non-active skill answers *"Not found"* rather than confirm it exists, expressly so there is no existence leak. My 403 both named the author and confirmed the draft existed — strictly more than a permitted read gives that caller. And agent ids are not opaque here; real deployments put human names in them.

A refusal has to admit the slug is occupied, or it could not refuse. Who owns it is a separate fact, and it is now omitted. A test pins that the owner's id does not appear in the detail — the regression this invites is someone adding it back while debugging.

**Left alone:** the duplicated `live_skill_doc.get("data")` between Adjustment 5b and 7a. Cosmetic, on a HITL-gated path, and the review judged it not worth restructuring against this repo's established trade-off. I agree — the two blocks read independently, and threading one extracted local through both would couple two checks that are deliberately separate.

## Fourth round: a replica read, a status leak, and a scoped claim

Three findings, all verified against source and fixed.

**High — the gate was reading from a replica.** `get_document` is `read: bool = True`; its own docstring says `read=False` forces the primary. So the fetch backing this gate could serve a *pre-transition* status: a `staged` row an admin has just approved to `active` still reads as `staged`, the gate authorises the overwrite the approval was meant to forbid, and the upsert that follows is a blind full replace with no CAS on status. Both sites now pass `read=False`, the same way the inline dedup lookups in that client do.

This is worth separating from the `ACCEPTED LIMITATION` note above it. That note covers the in-process fetch-to-upsert window — narrow, and inherent to the optimistic model. **The replica read was a different and much wider window, and unlike the first it was not a considered trade-off but a default I had not looked at.** Both surfaces now have a test pinning the kwarg, because this failure is invisible to any test that stubs storage: the stub answers identically either way.

**Medium — the refusal echoed the protected status.** The same leak I had just fixed for the owner, still present two lines up for the status. `_skill_hidden_from_agent` and the query/search filters treat a non-active row as non-existent to a non-admin caller, so echoing the status let a caller enumerate which slugs are `quarantined`, `rejected`, or under Forge review by writing at guessed slugs. Dropped, with a test parametrized over `PROTECTED_LIVE_STATUSES` asserting the name never appears.

**Medium — the ownership check trusts a self-declared `agent_id`.** Confirmed: `caura_doc` does `agent_id = _get_agent_id() or agent_id`, and that value becomes `SkillWriteContext.caller_agent_id`. Where the gateway signs `X-Agent-ID` the check is sound; where there is no verified agent identity the value is the caller's own parameter, and a caller who can declare an id can declare the victim's.

I did not try to harden this in the validator — every input available there comes from the same caller, so any added check would be theatre. It is also the standing ruling rather than an oversight: the trust ladder governs *agent* credentials, tenant scope governs *tenant* keys, so on a shared tenant credential the tenant is the boundary. What changed is the honesty of the claim: the gate carries a `SCOPE` note, and the test docstring carries it too, so neither is later read as proving a boundary it does not test.

Leaving "agent B cannot clobber agent A" standing unqualified would have been **the fourth instance in this PR of a comment asserting what no check enforces** — the defect the PR exists to fix.

## A test-fixture change, and why it is not a product change

Four tests in `test_mcp_doc.py` stubbed the storage client without `get_document`. They exercise `kind='create'`, so the fetch never ran and the plain `MagicMock` was never awaited; now it is, and they failed with `'MagicMock' object can't be awaited`. Passing `get_document=None` is the semantically correct stub — those tests write **new** skills, so no live doc exists. In production `get_storage_client()` returns a real async client.

Worth stating plainly rather than folding in silently: the four were a real regression in the test suite that my change caused, not pre-existing noise.

## Tests

**Thirteen fail without their fix**, each confirmed by disabling only the thing it guards.

Eleven are refusals:

- the active-skill overwrite (the attack itself)
- all four system-managed statuses, **parametrized off the constant** so a new one is covered the day it is added
- the candidate clobber
- a hash-bound `kind='update'` overwrite (fails if update is exempted)
- a 403 that advertises an unreachable remedy
- a cross-agent staged clobber
- an unidentified caller clobbering an attributed draft
- a refusal that names the owning agent

Two pin **fail-closed** on the live-doc fetch for the create path, one per surface, and fail if the error is swallowed.

**Eight are over-refusal guards:** an agent may still overwrite *and* still hash-bound-update its own `staged` skill, an author may still overwrite their own attributed draft, unattributed drafts stay writable in both the missing-`origin` and explicit-`unknown` shapes, a first write of a new slug is unaffected, and admin and Forge may still write — including admin over another agent's draft.

**Two pin the anti-drift properties:** that the protected set stays derived from the RBAC constants, and that the unattributed sentinel is the value the stamping site actually writes.

*A process note:* my first removal probe cut a line range and broke the file — 35 unrelated tests failed, which looks alarming and means nothing. An invalid probe proves neither direction. I re-ran it by disabling only the guard's condition, which gave the clean signal above.

## Verification

- Full root suite: **6029 passed, 5 skipped, 1 xfailed, 0 failed.**
- `legacy_name_ratchet.py` → *No new lines.* · `do_not_touch_sentinel.py` → *All 39 protected strings survive.* · `tenant_scope_gate.py` → no allowlist movement. All run **after `git add`**, and again after rebasing onto `f1d7b909`. Head `b547cac9`.

**A correction to my earlier verification claims on this PR.** I previously reported this suite as "54 failed / 5947 passed — back to the exact baseline". That 54-failure baseline was an environment artifact, not a property of the codebase: `core-api/.venv/bin/pytest` has a stale shebang pointing at the pre-rename `…/memclaw/core-api/.venv/bin/python`, so invoking `pytest` through `PATH` fails its exec and the shell falls through to the *system* pytest (Python 3.14, system site-packages), which is missing the project's pinned deps. Invoked as `.venv/bin/python -m pytest`, the suite is clean. The 54→58 delta that caught the `test_mcp_doc.py` regression was still a valid comparison — same wrong interpreter on both sides — so that finding stands. But the absolute number was wrong, and any future run should use `python -m pytest`.

### DIFF
--- a/core-api/src/core_api/mcp_server.py
+++ b/core-api/src/core_api/mcp_server.py
@@ -2216,23 +2216,39 @@ async def caura_doc(
                         # validator 422s it cleanly rather than us
                         # AttributeError-ing into a 500 here).
                         live_doc: dict | None = None
-                        if isinstance(data, dict) and data.get("kind") == "update":
+                        if isinstance(data, dict):
+                            # Fetched for EVERY skills write, not just
+                            # ``kind='update'``. The validator needs it for two
+                            # separate checks: update's hash-binding, and the
+                            # stored-status gate that stops a non-admin
+                            # replacing a live ACTIVE skill. Fetching only on
+                            # update made ``kind='create'`` the way around that
+                            # gate — the validator saw ``None`` and judged the
+                            # write purely on what the caller claimed.
+                            #
                             # Fail CLOSED, like the settings/flag gates: the
-                            # live-doc fetch feeds the validator's hash-
-                            # binding check, so a transient DB/network error
-                            # must abort with a curated message rather than
-                            # fall through to the outer handler (which would
-                            # leak the raw exception string).
+                            # live-doc fetch now feeds a security gate, so a
+                            # transient DB/network error must abort with a
+                            # curated message rather than fall through with
+                            # ``live_doc=None`` and skip the check.
                             try:
                                 sc_live = get_storage_client()
                                 live_doc = await sc_live.get_document(
                                     tenant_id=tenant_id,
                                     collection=SKILLS_COLLECTION,
                                     doc_id=doc_id,
+                                    # PRIMARY, not the replica — see the
+                                    # matching note on the REST fetch.
+                                    # This read backs an authorization
+                                    # decision, so replica lag would show
+                                    # the pre-transition status and the
+                                    # gate would authorise the overwrite
+                                    # the transition was meant to forbid.
+                                    read=False,
                                 )
                             except Exception:
                                 logger.exception(
-                                    "skills live-doc fetch failed for %s/%s; cannot gate update write",
+                                    "skills live-doc fetch failed for %s/%s; cannot gate skills write",
                                     tenant_id,
                                     doc_id,
                                 )
--- a/core-api/src/core_api/routes/documents.py
+++ b/core-api/src/core_api/routes/documents.py
@@ -289,24 +289,65 @@ async def upsert_document(
                 body_max_bytes=int(sf_settings.get("body_max_bytes", 40_000)),
             )
 
-            # For ``kind='update'`` we must fetch the live skill so the
-            # hash-binding check can compare against the current
-            # content_hash. Read-through storage; the validator handles
-            # the not-found case.
+            # Fetch the live skill for EVERY skills write, not just
+            # ``kind='update'``. Two checks need it and they need it for
+            # different reasons:
+            #
+            #   - ``kind='update'`` binds to the current content_hash;
+            #   - ANY kind aimed at an existing doc_id is a full replace,
+            #     so the validator has to see the stored status before
+            #     letting a non-admin overwrite it. Fetching only on
+            #     update meant ``kind='create'`` was the way around the
+            #     status RBAC: the validator was handed ``None`` and
+            #     judged the request purely on what the caller claimed.
+            #
+            # One extra read on skill creates, on a HITL-gated authoring
+            # path — not the hot path.
             #
             # Guarded by ``isinstance(body.data, dict)`` — a non-dict
             # body.data is a legitimate input (the validator below
             # rejects it with 422), but calling ``.get`` on it would
             # AttributeError into a 500 first. Cleanly punt that
             # rejection to the validator instead of crashing.
+            #
+            # Fail CLOSED, matching the MCP path: the fetch now feeds a
+            # security gate, so a transient storage error must abort with
+            # a curated message rather than fall through with
+            # ``live_doc=None`` and skip the check. An unhandled error
+            # here would already abort the write with a bare 500 — this
+            # keeps the two write paths identical and gets the failure
+            # into the log with the tenant and doc that hit it.
             live_doc: dict | None = None
-            if isinstance(body.data, dict) and body.data.get("kind") == "update":
-                sc_live = get_storage_client()
-                live_doc = await sc_live.get_document(
-                    tenant_id=body.tenant_id,
-                    collection=SKILLS_COLLECTION,
-                    doc_id=body.doc_id,
-                )
+            if isinstance(body.data, dict):
+                try:
+                    sc_live = get_storage_client()
+                    live_doc = await sc_live.get_document(
+                        tenant_id=body.tenant_id,
+                        collection=SKILLS_COLLECTION,
+                        doc_id=body.doc_id,
+                        # PRIMARY, not the replica. ``get_document``
+                        # defaults to ``read=True``; this fetch now backs
+                        # an authorization decision, so replica lag would
+                        # not be a benign stale read — it would show the
+                        # PRE-transition status (a staged row an admin
+                        # just approved to active) and the gate would
+                        # authorise the overwrite the transition was
+                        # meant to forbid. The upsert that follows does
+                        # no CAS on status, so nothing downstream catches
+                        # it. Same reason the inline dedup lookups pin
+                        # the primary.
+                        read=False,
+                    )
+                except Exception:
+                    logger.exception(
+                        "skills live-doc fetch failed for %s/%s; cannot gate skills write",
+                        body.tenant_id,
+                        body.doc_id,
+                    )
+                    raise HTTPException(
+                        status_code=503,
+                        detail="skill lifecycle gate unavailable",
+                    ) from None
 
             normalized, _scan = await validate_and_normalize_skill_write(
                 body.data,
--- a/core-api/src/core_api/services/skill_lifecycle.py
+++ b/core-api/src/core_api/services/skill_lifecycle.py
@@ -78,6 +78,36 @@ INTERNAL_ONLY_STATUSES: frozenset[str] = frozenset({"candidate"})
 # silently retire an active skill or hide a candidate from review.
 SYSTEM_ONLY_STATUSES: frozenset[str] = frozenset({"quarantined", "rejected", "stale", "deprecated"})
 
+# The same three sets, applied to the STORED status instead of the
+# incoming one. The invariant: a status a caller could not have
+# WRITTEN is a status that caller may not silently OVERWRITE.
+#
+# Without this the RBAC above gated only the string in the request
+# body, never the state of the row being replaced — so retiring an
+# active skill needed no privileged status at all, just the default
+# ``staged`` on a write aimed at an existing doc_id. The comment on
+# SYSTEM_ONLY_STATUSES named that exact threat ("silently retire an
+# active skill or hide a candidate from review") while the check it
+# annotates could not see the row it was protecting.
+#
+# Derived rather than spelled out so it cannot drift: adding a status
+# to any RBAC set above protects it here on the same commit.
+#
+# ``staged`` is deliberately absent, because an agent iterating on its
+# OWN not-yet-approved skill is the ordinary authoring loop. Note what
+# that sentence does and does not license: it is an exemption for the
+# author, not for everyone. Status alone cannot tell those apart, so
+# the ownership half is enforced separately at the gate — see
+# ``_live_owner`` below. Leaving it to this comment alone is how the
+# original bug read: a comment asserting a property no check enforced.
+PROTECTED_LIVE_STATUSES: frozenset[str] = ADMIN_ONLY_STATUSES | INTERNAL_ONLY_STATUSES | SYSTEM_ONLY_STATUSES
+
+# The value ``origin.agent_id`` carries when there was no caller
+# identity to stamp. Defined once and used by BOTH the stamping site
+# and the ownership check, so "unattributed" cannot come to mean two
+# different things in the two places.
+UNATTRIBUTED_AGENT_ID = "unknown"
+
 REQUIRED_TOP_LEVEL_KEYS: tuple[str, ...] = ("name", "slug", "description", "domain", "kind", "source")
 
 # ``slug`` regex mirrors the existing ``_SKILL_SLUG_RE`` in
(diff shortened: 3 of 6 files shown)


Eldad-Caura:
Both findings verified against source and addressed in `fbd31698`. One of them I resolved the other way than suggested, and the reason is worth stating.

**High — gate ordering vs `kind='update'`.** The ordering fact is correct, and so is the complaint about the message: it recommended `kind='update'` as the remedy, which this gate holds shut. But **option 1 (exempt `kind='update'`, since 7a enforces hash-binding) would reopen H-11**, so I took option 2 and corrected the message.

Hash-binding is optimistic concurrency, not authorization. It proves the caller read the current content, and reads are open to any tenant agent — so the exemption costs an attacker one extra GET:

1. `GET` the live active skill (permitted), copy `content_hash`
2. `POST kind='update'` with that `target_content_hash` → 7a satisfied
3. omit `status` → defaults to `staged` (a non-admin cannot write `active`)
4. the write still lands on the same `doc_id` as a full replace

Same outcome as the original attack. I applied your suggested exemption verbatim and re-ran rather than arguing it — the overwrite succeeds:

```
FAILED test_correct_hash_binding_does_not_buy_an_overwrite
  - Failed: DID NOT RAISE <class 'fastapi.exceptions.HTTPException'>
```

That test is new and exists to make this loud if the exemption is attempted again; a comment at the gate records the reasoning.

On the functional concern behind the finding: no working non-admin revision path is being removed, because there was never a correct one. Pre-PR, a non-admin `kind='update'` against an active skill **demoted it to staged** — no status is preserved from the live doc, and `active` is admin-only to write. That demotion *is* the bug. Agent-driven revision needs the Phase 4 update-proposal flow the module docstring anticipates, where a revision lands somewhere other than on top of the live row; until then an admin makes the change, and the message now says so. A second new test asserts the message no longer advertises the unreachable remedy, and a third covers an agent hash-bound-updating its own `staged` draft — `kind='update'` is not blocked as such, only against a protected live status.

**Medium — REST fail-closed parity.** Correct, and fixed: it now logs with tenant and doc and returns a curated 503.

One correction to the framing: this was not a fail-open. There is no enclosing handler in that route (verified), so an unhandled error already aborted the write with a bare 500 — the security property held on both paths. What differed was error quality and the missing log line. Worth fixing for parity on a security-relevant fetch; not a second vulnerability.

Added two tests pinning fail-closed on the **create** path, one per surface — the path this PR newly made fetch. `test_skill_update_write_fails_closed_on_live_doc_fetch_error` already covered update; I'd initially said no such test existed, having grepped the message text rather than the error code. Both new tests fail if the handler is changed to swallow the error and continue with `live_doc=None`, which is the failure mode that matters.

Rebased onto `f1d7b909`. Full suite 6014 passed / 0 failed; ruff check and format run separately at CI's scopes, mypy, ratchet, sentinel and tenant-scope gate all clean after `git add`. The PR body carries a correction to the "54-failure baseline" I'd been quoting on this PR — it was an artifact of a stale shebang in `.venv/bin/pytest` sending the run to the system interpreter.

Eldad-Caura:
@claude please re-review at head `fbd31698`. Since your last pass I addressed both findings, but resolved the High one against your primary suggestion — I'd value a second look specifically at that call.

The short version: exempting `kind='update'` from the stored-status gate (your option 1) reopens H-11, because hash-binding is optimistic concurrency rather than authorization. Reads are open to any tenant agent, so a caller can GET the live active skill, copy its `content_hash`, submit `kind='update'` with it, omit `status` (defaults to `staged`, since a non-admin cannot write `active`) — and the write still lands on the same `doc_id` as a full replace. I applied your exemption verbatim and re-ran: the overwrite succeeds, `DID NOT RAISE HTTPException`. So I took option 2 and corrected the message instead, and added a test that fails if the exemption is ever reintroduced.

Also fixed the REST fail-closed parity you flagged, with one framing correction: it was not a fail-open — there is no enclosing handler in that route, so an unhandled error already aborted the write with a bare 500. The gap was error quality and a missing log line, not a second vulnerability.

Full reasoning in the PR body and in my previous comment.

Eldad-Caura:
Medium addressed in `45c276a7` — and it was a fair catch that landed on a comment **I** wrote, so worth naming plainly.

The comment justified excluding `staged` because "an agent iterating on its own not-yet-approved skill is the ordinary case," while nothing compared the live row's `origin.agent_id` to the caller. So the exemption actually read *anyone may overwrite any staged draft*: agent B points a write at agent A's slug and A's unreviewed work is gone, no history, nothing in the inbox to show it happened. That is the same shape as the bug this PR fixes — a comment asserting a property no check enforces — so I took option (a) and enforced it rather than documenting the gap away.

Two halves now, and the comment says which is which. Refusal is limited to a live row naming a **different** owner:

- an unattributed draft (no `origin`, or the `unknown` stamp) stays writable on purpose — deployments with no agent identity stamp every draft that way, and gating on it would 403 the ordinary authoring loop for all of them;
- where the row does name an owner, a caller who cannot show that identity — including one with no `agent_id` at all — is not it. I deliberately did not take the looser "refuse only when both sides are attributed" reading, which would let an unbound tenant key overwrite any named author's draft.

`UNATTRIBUTED_AGENT_ID` is shared with the stamping site so "unattributed" cannot come to mean two different things, with a test pinning that it is the value the stamp writes.

Confirmed by disabling only the ownership condition: exactly the two new refusals fail, all four over-refusal guards still pass.

**On the TOCTOU suggestion** — added as an explicit `ACCEPTED LIMITATION` note at the gate. Same optimistic-concurrency model hash-binding already runs on. The note is there so this gate is not later mistaken for serialization when it is authorization, which is the same distinction that settled the `kind='update'` question.

Full suite 6020 passed / 0 failed; ruff check and format separately at CI's scopes, mypy, ratchet, sentinel all clean after `git add`. Base still `f1d7b909`.

Eldad-Caura:
@claude

Eldad-Caura:
@claude please re-review at head `45c276a7`, which adds the ownership half you asked for.

I took option (a) — enforced it — rather than (b), because the comment you flagged was asserting a property no check enforced, which is the same defect this PR exists to fix.

Two things in the implementation worth your attention, both places where the obvious version would be wrong:

1. **Refusal is scoped to a live row that names a *different* owner.** Unattributed drafts (no `origin`, or the `unknown` stamp) stay writable on purpose — deployments with no agent identity stamp every draft that way, so gating on it would 403 the ordinary authoring loop for all of them. That was the over-refusal trap here.

2. **I did not take the looser "refuse only when both sides are attributed" reading.** It would let a tenant key with no `agent_id` overwrite any named author's draft. So where the row names an owner, a caller who cannot show that identity — including one with no `agent_id` at all — is refused.

`UNATTRIBUTED_AGENT_ID` is now shared with the stamping site, with a test pinning that it is the value the stamp writes, so "unattributed" cannot come to mean two different things in the two places.

Your TOCTOU suggestion is in as an explicit `ACCEPTED LIMITATION` note at the gate.

Probe: disabling only the ownership condition fails exactly the two new refusals, with all four over-refusal guards still passing. Full suite 6020 passed / 0 failed.

Eldad-Caura:
Thanks — acted on the first suggestion in `d5fdffe2`, left the second.

**Naming the owner in the 403.** You framed this as low-risk-but-consider; checking the read path made it clearly worth fixing, because the codebase **already treats authorship as privileged one module over**. `op=read` on a non-active skill answers *"Not found"* rather than confirm it exists, expressly so there is no existence leak (`test_skill_read_hides_non_active_when_flag_on`). My 403 both named the author and confirmed the draft existed — strictly more than a permitted read gives that caller.

Your conditional turned out to be the actual case, too: agent ids here are not opaque. This deployment uses ids with human names in them, so "if `caller_agent_id` can ever carry a human-identifying string" is not hypothetical.

A refusal has to admit the slug is occupied, or it could not refuse. Who owns it is a separate fact and is now omitted. A test pins that the owner's id does not appear in the detail, since the regression this invites is someone adding it back while debugging, and the comment says why not to.

**Left alone: the duplicated `live_skill_doc.get("data")`** between 5b and 7a — agreed, and for the reason you gave. I'd add one: threading a single extracted local through both would couple two checks that are deliberately independent, so the duplication is doing a small amount of work.

Full suite 6021 passed / 0 failed. Gates clean after `git add`; base still `f1d7b909`.

One process note for whoever picks this up: `claude-review` skips `synchronize`, so both of my pushes produced a fully green check list on a head no reviewer had seen. Each verdict on this PR came from a standalone `@claude` comment, not from the push. Green checks on a head are not evidence of a review of that head.

Eldad-Caura:
@claude final pass at head `d5fdffe2` please — this is your previous verdict's first suggestion applied, so it should be a short one.

The ownership 403 no longer names the owning agent. Your framing was conditional ("if `caller_agent_id` can ever carry a human-identifying string"); it can, and does — this deployment's agent ids contain human names. What settled it was the read path: `op=read` on a non-active skill answers "Not found" rather than confirm it exists, expressly to avoid an existence leak, so naming the author in a write refusal handed an unauthorized caller strictly more than a permitted read gives them.

A refusal still admits the slug is occupied, because it cannot refuse otherwise. Only the authorship is gone. A test asserts the owner's id is absent from the detail, and the comment says why not to add it back for debugging.

Your second suggestion — the duplicated `live_skill_doc.get("data")` — I left, per your own note that it isn't worth restructuring here. One thing to add if you disagree: threading a single extracted local through both blocks would couple two checks that are deliberately independent, so I'd rather keep the duplication.

Nothing else changed. Full suite 6021 passed / 0 failed; all gates clean; base `f1d7b909`.

Eldad-Caura:
Two review runs fired here (both of my `@claude` comments triggered one), and between them they raised three findings. All three verified against source and fixed in `b547cac9`.

## High — the gate was reading from a replica

Correct, and the one I'm most glad to have caught. `get_document` is `read: bool = True`, and its own docstring says `read=False` forces the primary. So the fetch backing this gate could serve a pre-transition status: a `staged` row an admin has just approved to `active` still reads as `staged`, the gate authorises the overwrite the approval was meant to forbid, and — as you noted — the upsert that follows is a blind full replace with no CAS on status, so nothing downstream catches it.

Both fetch sites now pass `read=False`. This is the same reason the inline dedup lookups in that client pin the primary.

Worth separating from the `ACCEPTED LIMITATION` note I'd added: that note covers the in-process fetch-to-upsert window, which is narrow and inherent to the optimistic model. **The replica read was a different and much wider window, and unlike the first one it was not a considered trade-off — it was a default I hadn't looked at.** The note now says which is which.

Both surfaces have a test pinning `read=False` on the call, because this failure is invisible to any test that stubs storage — the stub answers identically either way. Confirmed both fail with the kwarg removed.

## Medium — the refusal echoed the protected status

Also correct, and it's the same leak I'd just fixed for the owner while leaving the status in place two lines up. `_skill_hidden_from_agent` and the query/search filters treat a non-active row as non-existent to a non-admin caller, so echoing the status let a caller enumerate which slugs are `quarantined`, `rejected`, or under Forge review by writing at guessed slugs.

Dropped. A refusal necessarily reveals the slug is occupied — it cannot refuse otherwise — but which state it's in is a separate fact. New test parametrized over `PROTECTED_LIVE_STATUSES` asserts the status name never appears in the detail, and I removed the one assertion that had depended on it.

## Medium — the ownership check trusts a self-declared `agent_id`

Correct, and I've scoped the claim rather than pretending to fix it. Confirmed the mechanism: `caura_doc` does `agent_id = _get_agent_id() or agent_id`, and `mcp_server.py:2205` passes that into `SkillWriteContext.caller_agent_id`. Where the gateway signs `X-Agent-ID`, `_get_agent_id()` wins and the check is sound. Where there is no verified agent identity, the value is the caller's own tool parameter — and a caller who can declare an id can declare the victim's.

**I did not try to harden this in the validator, because every input available there comes from the same caller** — any check I added would be theatre. This is the standing ruling in this codebase rather than an oversight: the trust ladder governs *agent* credentials, tenant scope governs *tenant* keys, so on a shared tenant credential the tenant is the boundary.

What I changed is the honesty of the claim. The gate carries a `SCOPE` note saying it is a boundary where agent identity is authenticated and slug-collision protection where it isn't, and the test docstring says the same so it isn't later read as proving a boundary it doesn't test. Leaving "agent B cannot clobber agent A" standing unqualified would have been the fourth instance in this PR of a comment asserting what no check enforces — which is the defect the PR exists to fix.

## Verification

Full suite **6029 passed / 0 failed**. `ruff check` and `ruff format --check` separately at CI's scopes, mypy, ratchet, sentinel, and the tenant-scope gate (no allowlist movement) all clean after `git add`. Base still `f1d7b909`.

Eldad-Caura:
@claude re-review at head `b547cac9` please — all three findings from the last two passes are addressed.

- **High (replica read):** both fetch sites now pass `read=False`. You were right that this was the wider window; my `ACCEPTED LIMITATION` note covered only the in-process fetch-to-upsert race, and the replica read was a default I hadn't examined rather than a trade-off I'd weighed. Pinned by a test on each surface, since a stubbed storage client answers identically either way.
- **Medium (status leak):** the status is no longer echoed, matching what `_skill_hidden_from_agent` does on the read paths. Parametrized test asserts no protected status name appears in the detail.
- **Medium (self-declared `agent_id`):** scoped rather than "fixed". I confirmed the mechanism you described and deliberately did not add a validator-side check, since every input there comes from the same caller. The gate and the test docstring now both state that it is a boundary only where agent identity is authenticated, and slug-collision protection where it isn't.

If you think the third one warrants more than an honest scope note — i.e. that the write path should refuse when it has no verified agent identity at all — say so explicitly and I'll take it, but I'd want to hear the argument against the trust-ladder/tenant-scope split first, since refusing there would break the standalone authoring loop.

Eldad-Caura:
@erni-a ready for human review at head `b547cac9`.

**What it fixes (H-11):** the skills-write validator applied status/source RBAC only to the *incoming body*, never to the row being replaced — and `document_upsert` is a full-blob replace. So any tenant agent with a writable key could point `kind='create'` at a live ACTIVE skill, omit `status` (defaulting to `staged`), and retire it: gone from every `status='active'` surface, curated content destroyed, replacement sitting in the inbox looking routine. **No privileged status was needed — the default one did it.**

**Verified:** 13/13 checks green with zero failures, confirmed by head SHA. Full suite 6029 passed / 0 failed. `ruff check` and `ruff format --check` run separately at CI's scopes, mypy, legacy-name ratchet, do-not-touch sentinel, and the tenant-scope gate (no allowlist movement) — all run after `git add`. Every new test confirmed to fail without its fix.

`mergeStateStatus` is BLOCKED solely on `REVIEW_REQUIRED` (0 reviews). **I am not merging — that's Eldad's.**

Five automated review passes ran. Four points where I'd particularly value a human read, because each was a judgement call rather than a mechanical fix:

1. **I rejected a reviewer's suggested fix.** It proposed exempting `kind='update'` from the gate since hash-binding already checks it. That reopens the hole: hash-binding is optimistic concurrency, not authorization — reads are open to any tenant agent, so it costs an attacker one extra GET. I applied the suggestion verbatim and re-ran the attack; the overwrite succeeded. There's now a test that fails if anyone reintroduces the exemption.

2. **The `staged` exemption needed an ownership half.** My own comment claimed "an agent iterating on its *own* draft" while nothing compared `origin.agent_id` to the caller — so it actually read *anyone may overwrite any staged draft*. Enforced now, with unattributed drafts deliberately left writable (deployments with no agent identity stamp every draft `unknown`; gating on it would 403 the ordinary loop for all of them).

3. **The ownership check is scoped, not absolute — please sanity-check that call.** It's a real boundary only where the gateway signs `X-Agent-ID`; on standalone/shared-key surfaces `agent_id` is the caller's own self-declared parameter. I did *not* add a validator-side check, because every input there comes from the same caller and anything I added would be theatre. That matches the standing ruling (trust ladder governs agent credentials, tenant scope governs tenant keys), but it's an architectural judgement worth a second opinion, and both the gate and the test docstring now state the limit explicitly.

4. **A replica read was undermining the gate.** `get_document` defaults to `read=True`, so a `staged` row an admin had just approved to `active` could still read as `staged` and the gate would authorise the overwrite the approval was meant to forbid. Both sites now pin `read=False`. Notable because I'd already written an `ACCEPTED LIMITATION` note about the *narrower* fetch-to-upsert race, which made the concurrency story look considered while a wider window sat in the same function as an unexamined default.

One process note worth knowing independently of this PR: `claude-review` skips `synchronize`, so **a push produces a fully green check list on a head no reviewer has seen.** Every verdict here came from a standalone re-trigger comment, not from the push. Green checks on a head are not evidence that head was reviewed.

erni-a (approved):

PR #1272, opened 2026-09-04
fix(fleet): gate the heartbeat behind enforce_read_only

Eldad-Caura:
Closes audit finding **H-04**.

## The gap

`POST /fleet/heartbeat` was the **only mutating route in the fleet module without `enforce_read_only`**. Its five siblings — `create_fleet`, `delete_fleet`, `purge_fleet`, `create_command`, `command_result` — all have it.

The heartbeat reads like telemetry. It isn't:

- `upsert_node` replaces the node row — hostname, versions, the whole metadata blob
- `get_or_create_agent` materialises per-agent rows on every tick
- `_maybe_queue_auto_upgrade` can queue a deploy command
- **`ack_commands` drains the node's pending queue**, after returning the payloads in the response

## Why this isn't just an unwanted write

A read-only (`capabilities={'read'}`) or demo credential naming an existing node could **claim that node's queued commands**. It receives the payloads intended for the node, and because acked commands are not redelivered, the real node never collects them.

One request, unrecoverable. The operator sees a dispatched command that was acknowledged and never ran.

The removal probe shows the exfiltration half directly — without the gate, a `capabilities={'read'}` caller gets back:

```
{"ok":true,...,"commands":[{"id":"9fb742d4…","command":"ping","payload":{"k":"v"}}]}
```

## The fix, and why it can't break real plugins

`enforce_read_only` blocks exactly two things: `is_demo`, and a credential whose `capabilities` are set and exclude `write`. Capabilities are populated from the gateway's `x-capabilities` header, so **legacy credentials carry `None` and pass through untouched**. The only callers newly refused are ones that were never entitled to write.

This was the question I most wanted answered before touching it, because the heartbeat is how every node stays live and commandable on a ~60s tick — breaking it would be a worse outage than the bug. Hence three of the six tests are over-refusal guards.

**Gate order** matches the siblings: credential capability first, then target tenant. Nothing depends on the order — `enforce_read_only` reads only the caller's own credential — so consistency is the whole argument.

**`enforce_usage_limits` is deliberately NOT added**, though the write-shaped siblings carry it. This call recurs every ~60s from every node; gating it on plan limits would take a tenant that merely exceeded its quota and make its entire fleet go stale and uncommandable — including the commands an operator would use to reduce usage. `create_command` sets the same precedent on this path: read-only gate, no usage gate.

## Tests

**Three fail without the gate**, confirmed by removing only `enforce_read_only`:

- a read-only key is refused
- a demo credential is refused
- the queue-drain attack is refused, with **both halves asserted**: the payload must not come back, and the command must still be `pending` with `acked_at` null afterwards

**Three are over-refusal guards:**

- a legacy credential (`capabilities=None`) still heartbeats
- a `{read, write}` credential still heartbeats
- the command channel still **delivers and acks** for a credential that may write

That last one is load-bearing: without it, a gate that refused *every* caller would satisfy all three refusal tests.

## Verification

- Full root suite: **6035 passed, 5 skipped, 1 xfailed, 0 failed** (6029 baseline + the 6 new tests), run as `.venv/bin/python -m pytest`.
- `legacy_name_ratchet.py` → *No new lines.* · `do_not_touch_sentinel.py` → *All 39 protected strings survive.* · `tenant_scope_gate.py` → no allowlist movement in any category. All run **after `git add`**.
- Branched fresh from `origin/main` at `b0a576a5`.

### DIFF
--- a/core-api/src/core_api/routes/fleet.py
+++ b/core-api/src/core_api/routes/fleet.py
@@ -654,6 +654,39 @@ async def heartbeat(
     auth: AuthContext = Depends(get_auth_context),
 ):
     """Plugin pushes status; receives pending commands in response."""
+    # The heartbeat is a WRITE, despite reading like telemetry, and it was
+    # the only mutating route in this module without the gate — its five
+    # siblings (create/delete/purge fleet, create_command, command_result)
+    # all have it. What it mutates:
+    #
+    #   - ``upsert_node`` replaces the node row (hostname, versions, the
+    #     whole metadata blob);
+    #   - ``get_or_create_agent`` materialises agent rows per heartbeat;
+    #   - ``_maybe_queue_auto_upgrade`` can queue a deploy command;
+    #   - ``ack_commands`` DRAINS the node's pending queue and returns the
+    #     payloads in the response.
+    #
+    # That last one is why this is not merely an unwanted write. A
+    # read-only or demo credential naming an existing node could claim
+    # that node's queued commands — receiving payloads meant for it, and
+    # leaving nothing for the real node to collect, since acked commands
+    # are not redelivered. One request, unrecoverable: an operator sees a
+    # dispatched command that was acknowledged and never ran.
+    #
+    # Gate order matches the siblings (``create_fleet``,
+    # ``create_command``): the credential's own capability is judged
+    # first, then the target tenant. Nothing here depends on the order —
+    # ``enforce_read_only`` reads only the caller's own credential — so
+    # consistency is the whole argument.
+    #
+    # ``enforce_usage_limits`` is deliberately NOT added, though the
+    # write-shaped siblings carry it. This call recurs every ~60s from
+    # every node and is how a node stays live and commandable; gating it
+    # on plan limits would take a tenant that merely exceeded its quota
+    # and make its entire fleet go stale and uncommandable — including
+    # the commands an operator would use to reduce usage. ``create_command``
+    # sets the same precedent on this path: read-only gate, no usage gate.
+    auth.enforce_read_only()
     auth.enforce_tenant(body.tenant_id)
 
     if is_plugin_outdated(body.plugin_version) and _should_log_outdated_plugin(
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

Eldad-Caura:
@erni-a ready for human review at head `6917503c`.

**What it fixes (H-04):** `POST /fleet/heartbeat` was the only mutating route in the fleet module without `enforce_read_only` — all five siblings have it. It reads like telemetry but upserts the node row, materialises agent rows, can queue a deploy, and returns the node's pending commands and then **acks** them.

That last part is why it isn't merely an unwanted write: a read-only or demo credential naming an existing node could claim that node's queued commands — receiving the payloads, and leaving nothing for the real node, since acked commands aren't redelivered. One request, unrecoverable.

**Verified:** 13/13 checks green with zero failures, confirmed by head SHA; automated review clean on the first pass. Full suite 6035 passed / 0 failed. ruff check and ruff format --check run separately at CI's scopes, mypy, ratchet, sentinel, and the tenant-scope gate (no allowlist movement) — all after `git add`. Branched fresh from `origin/main` at `b0a576a5`.

`mergeStateStatus` is BLOCKED solely on `REVIEW_REQUIRED`. **Not merging — that's Eldad's.**

Two judgement calls worth a human eye, both about *not* doing something:

1. **`enforce_usage_limits` deliberately omitted**, though the write-shaped siblings carry it. This call recurs every ~60s from every node and is how a node stays live and commandable; gating it on plan limits would take a tenant that merely exceeded quota and make its whole fleet go stale and uncommandable — including the commands an operator would use to reduce usage. `create_command` sets the same precedent on this path.

2. **The over-refusal risk was the main thing I checked before touching this**, since a bad gate here is a worse outage than the bug. `enforce_read_only` blocks only `is_demo` and credentials whose `capabilities` are set and exclude `write`; capabilities come from the gateway's `x-capabilities` header, so legacy credentials carry `None` and pass untouched. Three of the six tests guard that, including one asserting the command channel still *delivers and acks* — without it, a gate that refused every caller would satisfy all three refusal tests.

erni-a (approved):

PR #1273, opened 2026-09-04
fix(graph): apply the evidence-visibility filter on GET /graph

Eldad-Caura:
Closes audit finding **H-03**.

## The leak

`GET /entities/{id}` has enforced a scope contract since audit **S5**: *an agent credential must not enumerate relation edges or `evidence_memory_id`s pointing at memories it cannot read.*

`GET /graph` read the same edges from a different storage call and had **no filter at all** — `enforce_readable_tenant`, then every entity and every relation for the tenant, with `relation_type`, both endpoints, `weight` and `evidence_memory_id` verbatim. `fleet_id` is optional, so omitting it asked for every fleet at once.

The failure, as the finding describes it:

1. Agent A writes a `scope_agent` memory.
2. The extraction worker persists `(anna) -negotiates-> (zenith acquisition)` with `evidence_memory_id` pointing at that memory.
3. Peer agent B calls `GET /graph?tenant_id=T` and reads the triple **plus the private memory's id** — exactly what `GET /entities/{id}` would have withheld.

The raw memory text was never exposed on either route. **The derived triple is the leak** — it carries the secret without carrying the sentence.

## One filter, two readers

The filter now lives in `entity_service.filter_relations_by_evidence_visibility`, which both readers call. I deliberately did **not** reimplement it in the route: the reason H-03 existed is that the guarantee was a property of *one caller*. `get_entity`'s own comment about its tenant check already names the class —

> this comparison used to be the ONLY thing scoping this read … the defect class being that the guarantee lived in one caller

A second copy in the route would be two places that have to agree, which is how a third reader eventually gets added without either.

`get_entity` is behaviour-unchanged: same predicate, same bulk round-trip, same pre-authorized set from its linked memories, same single agent lookup.

**Not narrowed to tenant/admin credentials**, which was the finding's other suggested remedy — that would break dashboards, which read `/graph` with a tenant credential. The filter is a no-op when `caller_agent_id` is `None`, matching the linked-memory filter's contract in the same module.

**Ordering:** the filter runs before the log line and the cross-tenant audit, so both count what was actually disclosed. Counting pre-filter would overstate the disclosure and record a cross-tenant read of edges that were withheld.

## Two probes, because one wasn't enough

**Disabling only the route's call fails exactly one test** — the end-to-end one that drives `get_graph` and asserts the private edge is absent from the response body and the evidence id appears nowhere in it:

```
AssertionError: private edge leaked: [{'id': 'r-secret', 'relation_type': 'negotiates',
  'evidence_memory_id': '2e35cc3a-…'}, …]
```

The other eight tests exercise the shared function and **pass with the route unwired** — which is the H-03 state precisely. Without the end-to-end test, this PR could have shipped a correct filter that nothing called.

**Neutering the filter itself fails five:** three new refusal tests (peer `scope_agent` evidence, cross-fleet evidence below the trust bar, unresolvable evidence failing closed), the end-to-end, and **the pre-existing S5 test** — which is what shows `get_entity` genuinely routes through the shared function rather than keeping a duplicate.

## Tests

**Five fail without the fix.** Refusals: a peer's `scope_agent` evidence, cross-fleet evidence below the trust bar, and unresolvable evidence (deleted / nonexistent / cross-tenant) failing **closed** — an id that doesn't resolve is the one case where we can't establish the caller may read it.

**Five are over-refusal guards:** the evidence's own author still sees their edge, `trust >= 2` still gets cross-fleet edges, no-evidence edges stay, `scope_org` evidence stays, and a tenant credential still sees the whole graph. One more pins that the agent row is resolved **once for the entire graph** rather than per edge — `/graph` returns every relation in the tenant, so a per-edge lookup would be N+1 across the whole graph.

## Not fixed, and why

Each node still carries a `memory_count` from `count_memories_per_entity`, which takes no agent scoping — so the count includes memories the caller cannot read. I left it:

- it is an inference channel over a **count**, not a disclosure of content or ids;
- there is no sibling contract to mirror (`get_entity` returns filtered linked memories, not a count);
- scoping it needs a storage-side change, widening this PR into `core-storage-api`.

Flagging it rather than silently widening or silently ignoring. Say the word if you'd rather it were in scope.

Also worth stating: `get_entity` does not gate the entity **node** itself on fleet, only its memory-derived content. So `/graph` still returns cross-fleet entity names and attributes to an agent that omits `fleet_id`. That is the pre-existing contract on both surfaces rather than a regression here, and there is no established filter to mirror — but it is the natural next question and I'd rather name it than leave it implied.

## Verification

- Full root suite: **6045 passed, 5 skipped, 1 xfailed, 0 failed** (6035 baseline + the 10 new tests), run as `.venv/bin/python -m pytest`.
- `legacy_name_ratchet.py` → *No new lines.* · `do_not_touch_sentinel.py` → *All 39 protected strings survive.* · `tenant_scope_gate.py` → no allowlist movement. All run **after `git add`**.
- `get_full_graph` has exactly one caller and there is no MCP graph tool, so REST `/graph` was the only unfiltered surface; MCP `entity_get` routes through `get_entity` and is covered by the shared filter.
- Branched fresh from `origin/main` at `352a4bc6`.

### DIFF
--- a/core-api/src/core_api/routes/entities.py
+++ b/core-api/src/core_api/routes/entities.py
@@ -15,7 +15,12 @@ from core_api.schemas import (
     RelationUpsertOut,
 )
 from core_api.services.audit_service import log_cross_tenant_read
-from core_api.services.entity_service import get_entity, upsert_entity, upsert_relation
+from core_api.services.entity_service import (
+    filter_relations_by_evidence_visibility,
+    get_entity,
+    upsert_entity,
+    upsert_relation,
+)
 from core_api.services.usage_service import check_and_increment_by_tenant as check_and_increment
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


Eldad-Caura:
@erni-a please pprve

erni-a (approved):

PR #1274, opened 2026-09-04
fix(storage): give a bulk batch one column list

Eldad-Caura:
## The incident

`POST /api/v1/memories/bulk` returned 504 continuously from **2026-09-02T19:00Z**, ~680/hour for 29 hours (~19,500 failures), for tenant `arkash24-4d270c`.

The 504 was a mapping, not a timeout — median latency 0.64s, max 15.7s. Three layers each renamed the fault:

```
core-storage-writer   sqlalchemy.exc.CompileError: INSERT value for column
                      memories.embedded_content_hash is explicitly rendered as a
                      boundparameter in the VALUES clause
core-api              WARNING  bulk write got 500 from storage;
                               client should retry with same X-Bulk-Attempt-Id
gateway               504
```

## Root cause

`memory_add_all` builds a multi-values INSERT, and `pg_insert(...).values(<list of dicts>)` compiles **one** column list for the whole statement. `_filter_memory_fields` invented `embedded_content_hash` only for rows carrying a vector, so a batch mixing embedded and deferred rows disagreed on that column.

Upstream is not at fault: `memory_service` builds each row from a fixed dict literal, so every item carries every key. The divergence was manufactured entirely inside the mapper — which is why this is the first such incident despite the multi-values call being old.

**Reachable through documented API usage, not misuse.** `BulkMemoryItem.write_mode` is a per-item field whose description explicitly invites per-item opt-in, and under `DEPLOYMENT_MODE=deferred` (prod's shape) `memory_service` embeds **only** `write_mode="strong"` items. Any batch mixing modes diverges.

Latent since #786 added the column on 2026-08-16. Verified this was **not** a deploy: zero storage-writer errors 17:00–19:00 on 09-02 (control query confirms 300 non-error logs in the same window), the cap hit 19:00–21:00, and the nearest core-api revisions are 09-02T04:30 and 09-03T05:00. What changed at 19:00Z was a client's workload. **A rollback would not have fixed it.**

## The second, silent failure — arguably the worse one

Row **order** decided which failure you got, and only one was loud:

| Row 0 | Later row | Outcome |
|---|---|---|
| sets the column | does not | `CompileError` at execute time; statement never reaches Postgres, whole batch fails — the 504 storm |
| does not | sets it | **compiles, and SQLAlchemy drops the column from the INSERT entirely, with no warning** |

I verified the second path emits **zero** warnings. The compiled SQL simply omits the column, so that row's provenance persists as `NULL` — which `memory_quality_metrics` buckets as `unknown_provenance`, a bucket whose own docstring defines it as *"written before migration 037"* and says *"it drains as rows are rewritten or re-embedded."* Nothing re-embeds them: the nightly sweep repairs only `missing` (`embedding IS NULL`), and `stale` requires `embedded_content_hash IS NOT NULL`. Those rows leave the staleness detector's reach permanently.

CI's own comments record that #786 already had one near-miss on exactly this outcome, caught only by human review because no automated check covered `core-worker`. This is a second route to the same place that no check covered either. The 504 was the *lucky* ordering — it is the one that told us something was wrong.

## What this changes

**1. Uniformity at the source.** `_filter_memory_fields` always emits the key, `None` when the inference does not apply — identical on the wire for a nullable, default-less column. Costs one bind parameter per row on batches that previously omitted it (~+8% on statement construction, 100 of asyncpg's ~32k ceiling at the 100-item cap). That is the price of uniformity that row ordering cannot defeat.

A blanket fixed-key projection would be worse: 8 of `Memory`'s 33 columns carry server defaults and are `NOT NULL`, so filling missing keys with `None` would write NULL over a default or violate the constraint. `embedded_content_hash` is the only key this function *invents*, so "always emit the key you invent" is a complete local invariant. **The escalation trigger for a second conditionally-set column is in the comments** — at that point the safe-to-fill set should be *derived* from the mapper, not hand-listed.

**2. Two guards, because the two causes need opposite answers.** Telling them apart requires looking at both the items and the mapped rows:

- **Items disagree** → caller-fixable → `422`, beside the existing tenant/fleet checks. This is the reachable case: the route hands `request.json()` straight through. Intersected with `_MEMORY_VALID_FIELDS`, so items differing only on a key the mapper drops are *not* rejected.
- **Rows disagree after uniform items** → server fault → permanent `500`.

Guessing either way is harmful: a caller told "permanent" about something it could fix stops trying; a caller told "retry" about something it cannot fix never stops. An earlier draft asserted "the caller's payload was fine" without checking, and the reachable case is precisely the one it mislabelled.

**3. A permanent failure is no longer advertised as retryable.** This is the amplifier that turned a bug into a 29-hour outage. Two layers were telling the client to retry:

- the bulk route: `504, retry with the same X-Bulk-Attempt-Id`
- `app.upstream_http_error_handler`: `503 "Upstream dependency unavailable; retry."` for **any** unhandled upstream 5xx

Both are right for a transient failure and inverted for one that reproduces byte-for-byte. Storage now marks such failures `retryable: false` (`common/permanent_failure`, following `common/duplicate_memory`'s precedent so the two services share one symbol instead of agreeing on literals), and the **storage client** translates it into a typed error at the boundary every write funnels through.

That placement is load-bearing, not tidiness. Three callers of the bulk storage endpoint — the MCP write tool, `ingest_service`, `_insert_children_or_degrade` — do not catch `httpx.HTTPStatusError`, so a route-local check would have left them being told to retry a permanently-refused write. `_insert_children_or_degrade` is the worst case: it would advertise a retry for a request whose parent row already committed.

Unmarked 5xx keep the retry advice, deliberately. The costs are asymmetric: a needless retry is a no-op, because `ix_memories_attempt_unique` resolves an already-committed row to `duplicate_attempt` with its canonical id — while a wrong "do not retry" strands committed rows with nothing to recover them. Hence `is False` rather than a falsy test; an absent key, a null, a proxy's HTML page and a non-JSON body all read as "not marked".

## Not a size problem

The failing payload was ~214KB and a succeeding one ~155KB, which looks like a threshold and is not. Batch size only changes the *probability* that a batch is heterogeneous. **A size-based fix — raising a limit, chunking by bytes — would appear to work for exactly as long as batches happened to stay uniform, then fail again with no new signal.** The sizes were also not constant (214214/214215/214216/214217), so this was a client continuously generating same-shaped batches, not one poisoned message.

## Verification

Tests were confirmed red by reverting the fix, under the pinned interpreter (`.venv/bin/python -m pytest`, Python 3.13.0, SQLAlchemy 2.0.51):

- the compile case fails with the production error **verbatim**
- the silent case fails showing `embedded_content_hash` absent from the compiled INSERT
- controls (homogeneous batch, single row) correctly stay green either way

A test asserting only "does not raise" passes on the broken code — which is how the silent half went unnoticed while the other half paged. The tests compile the real statement rather than asserting on the mapper's output, because compilation is where the defect lives.

No regressions: I reverted the source, ran the full root suite, saved the sorted failure names, restored, re-ran the identical command, and diffed — **byte-identical, 18 both ways**. Those 18 are environmental (the root conftest provisions via `create_all`, so migration 034's tsvector trigger is absent and FTS scores come back `None`), not properties of the codebase.

`328` storage tests and `144` core-api bulk/tenancy-gate tests pass; `ruff check` and `ruff format --check` clean at CI's exact separate scopes; `mypy` clean on both services.

One regression was caught this way and fixed: giving the new exception its own `__init__` took the **entire tenancy invariant offline**, because `scripts/tenant_scope_gate.py` resolves functions by bare name and refuses to run on a module defining one twice — and `postgres_service.py` already had one on `DuplicateContentHashError`. The error message says nothing about tenancy and you cannot rename `__init__`. Fixed by factoring the fields-carrying base into `common/permanent_failure.py`, which removed the same latent trap from `storage_client.py`.

## Stopping the bleeding

Separately from this PR, and not applied here: the client-side fix is to make `write_mode` uniform within each batch (or split `strong` items into their own batch). No deploy, no server config change, and unlike halting the agents it lets the backlog **drain** rather than discarding it. Request sizes plateaued by 09-02T20:00Z and were byte-stable for 28+ hours, so the pending set is frozen and bounded and plausibly still exists — whether it survives a restart is a question for the agents' owner, not the logs.

### DIFF
--- a/common/permanent_failure.py
+++ b/common/permanent_failure.py
@@ -1,0 +1,92 @@
+"""Wire shape for "this write failed and a retry cannot clear it", shared by both services.
+
+Why this module exists. core-api's bulk route answers a storage 5xx with "504,
+retry with the same ``X-Bulk-Attempt-Id``", and ``app.upstream_http_error_handler``
+answers every *unhandled* upstream 5xx with "503, retry". Both are right for the
+failure they were written for — a lost response, a restart mid-commit, a
+transient upstream — and both are wrong for a failure that reproduces byte-for-
+byte on every attempt. A client that complies with either retries forever. In
+prod that ran at ~680 requests/hour for 29 hours against a batch that could
+never compile, and nothing on the wire distinguished it from noise.
+
+So storage has to be able to SAY "permanent", and the word has to survive the
+trip. Retryable stays the default: the costs are asymmetric. A wrong "retry" is
+a no-op by construction, because ``ix_memories_attempt_unique`` resolves an
+already-committed row to ``duplicate_attempt`` with its canonical id — while a
+wrong "do not retry" strands committed rows with nothing to recover them. Hence
+:func:`is_permanent` tests for the literal ``False`` rather than falsiness:
+an absent key, a null, a non-JSON body and a proxy's HTML error page must all
+read as "not marked".
+
+Pure data: no framework imports, so both services and their tests can use it —
+the same reason ``common/duplicate_memory.py`` is shaped this way, and for the
+same failure it was built to prevent. The marker key and the code strings used
+to be literals written once in the storage route and again in core-api's
+reader: two separately-deployed services required to agree byte-for-byte, with
+no symbol to grep and no import to follow.
+"""
+
+from __future__ import annotations
+
+# core-api's error code for a refused write that must not be retried. Named for
+# the ANSWER, not the cause: a caller's next move is the same whatever made the
+# failure permanent, and the specific cause travels in ``error`` below.
+PERMANENT_WRITE_FAILURE_CODE = "PERMANENT_WRITE_FAILURE"
+
+# The key storage sets to ``False``. Deliberately not ``permanent: true`` — the
+# question a caller asks is "may I retry?", and answering the question actually
+# asked is what keeps a reader from inverting it by accident.
+RETRYABLE_KEY = "retryable"
+
+# Causes. One per way a write can fail permanently; today there is one.
+# ``bulk_row_shape``: the rows of a bulk batch disagreed on which columns they
+# set, so the multi-values INSERT has no single column list to compile.
+CAUSE_BULK_ROW_SHAPE = "bulk_row_shape"
+
+
+class PermanentWriteFailure(Exception):
+    """Base for a write refusal that a retry cannot clear.
+
+    Both services raise a subclass of this — storage when it detects the fault,
+    core-api when it reads the marker back off the wire — so the "carries
+    structured fields beside the message" behaviour is defined once.
+
+    The single definition is not only tidiness. ``scripts/tenant_scope_gate.py``
+    resolves functions in the modules it scans by BARE NAME and refuses to run
+    when one is defined twice, so a second hand-written ``__init__`` in
+    ``postgres_service.py`` or ``storage_client.py`` takes the whole tenancy
+    invariant offline — with an error about ambiguous names that says nothing
+    about tenancy. Inheriting costs nothing and cannot trip it.
+
+    ``fields`` is the structured half. ``common/duplicate_memory.py`` records
+    why it exists: the alternative already cost us a uuid serialised into
+    English, shipped across two service boundaries, and recovered at the far
+    end with a regular expression.
+    """
+
+    def __init__(self, message: str, fields: dict | None = None) -> None:
+        super().__init__(message)
+        self.fields: dict = fields or {}
+
+
+def permanent_detail(*, cause: str, message: str, **fields: object) -> dict:
+    """Storage's ``detail`` for a failure a retry cannot clear.
+
+    ``message`` is prose for a human; ``cause`` is the stable string a program
+    branches on. Extra *fields* ride alongside rather than being formatted into
+    the message — ``common/duplicate_memory.py``'s docstring records what the
+    alternative cost us, a uuid serialised into English and recovered at the far
+    end with a regular expression.
+    """
+    return {"error": cause, RETRYABLE_KEY: False, "message": message, **fields}
+
+
+def is_permanent(detail: object) -> bool:
+    """Did storage explicitly mark this failure as one a retry cannot clear?
+
+    Takes the parsed ``detail`` — anything at all, because on an error path it
+    may be a string from FastAPI's default handler, a list, or absent entirely.
+    Only a mapping carrying ``retryable`` set to exactly ``False`` counts; see
+    the module docstring for why the default must stay retryable.
+    """
+    return isinstance(detail, dict) and detail.get(RETRYABLE_KEY) is False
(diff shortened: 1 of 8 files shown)


Eldad-Caura:
@erni-a ready for your review — production incident fix, all gates clean.

**Head `b930e9aca2736d1aa7ee0316a4166c305cf5fe76`**: 13 check-runs, 0 non-clean (11 success, 2 skipped by design — `claude-capture`, `claude-retrigger`). Verified by head SHA via `gh api repos/caura-ai/caura/commits/$SHA/check-runs`. Review verdict is a real posted comment on this head, not a workflow status: no issues found.

**What it fixes.** `POST /api/v1/memories/bulk` returned 504 for 29 hours (~19,500 failures, ~680/hour) because `pg_insert(...).values(<list of dicts>)` compiles one column list for the whole batch, and `_filter_memory_fields` invented `embedded_content_hash` only for rows carrying a vector. Triggered by documented usage: `BulkMemoryItem.write_mode` is a per-item field, and a deferred deployment embeds only `strong` items, so any mixed batch diverged.

**Three things worth your eye specifically:**

1. **The silent twin.** Row order decided the failure. The other ordering compiles cleanly, emits no warning, and drops the column from the INSERT — writing NULL provenance into `unknown_provenance`, a bucket documented as "written before migration 037" that nothing re-embeds. Fleet-wide and invisible. The 504 was the lucky ordering.

2. **Route vs boundary.** The retry-advice fix is at the storage-client boundary, not in the bulk route, and that placement is the finding rather than a refactor. `upstream_http_error_handler` maps any unhandled upstream 5xx to `503 "retry"`, and three callers of the bulk storage endpoint (MCP write tool, `ingest_service`, `_insert_children_or_degrade`) do not catch `httpx.HTTPStatusError` — so a route-local check would have inverted the advice on exactly the paths nobody exercises.

3. **Not a size fix.** The failing payload was ~214KB and a succeeding one ~155KB, which looks like a threshold and is not. Batch size only changes the probability that a batch is heterogeneous, so a limit bump or byte-chunking would work until a batch mixed modes again, with no new signal.

**Not a deploy**, so a rollback would not have helped: zero storage errors before 19:00Z (with a control query proving the window was visible), nearest revisions 14h either side. Latent since #786.

Tests confirmed red by reverting under the pinned interpreter — the compile case fails with the production error verbatim, the silent case fails showing the column absent from the compiled INSERT. No regressions, proved by diffing sorted failure names between pristine and patched source: byte-identical, 18 both ways, all environmental (the root conftest's `create_all` skips migration 034's tsvector trigger).

One regression caught in review and fixed, worth knowing about independently of this PR: giving the new exception its own `__init__` took the entire tenancy invariant offline, because `scripts/tenant_scope_gate.py` resolves functions by bare name and refuses to run on a module defining one twice — and `postgres_service.py` already had one on `DuplicateContentHashError`. The error names neither tenancy nor an achievable fix.

Leaving the merge to you. Full reasoning is in the PR description.

erni-a (approved):

PR #1275, opened 2026-09-04
fix(ratchet): resolve the permanent-red completion case, add the central workflow

Eldad-Caura:
## What

Two changes, landed together deliberately (see "Why this PR doesn't merge yet" below):

1. **The permanent-red fix.** The gate's own "found nothing anywhere" branch couldn't tell a finished migration from a broken pathspec finding nothing, so it always failed. `caura-bus` just hit this for real, permanently, when its migration actually finished. Adds a positive control (`NEW_NAME = "caura"` — a name a finished migration cannot have removed) that reruns the exact same scan/exclusions for the current brand before trusting a zero result for the legacy one: both live but empty → genuinely done, pass; either dead → same failure as today, fail. Exclusions stay on in the control, matching how the rest of the gate already treats mirror/CHANGELOG content as invisible.

2. **`legacy-name-ratchet.yml`** — a standalone, required-workflow-shaped copy of this gate, self-identifying via `job.workflow_repository`/`job.workflow_sha` (ported from `caura-onprem-installer`'s existing parity-checker workflow) so a candidate PR can't substitute its own checker. Not wired into this repo's own CI — hosted here to eventually be pointed at by an org ruleset, pinned by tag, replacing nine repos' vendored copies of the engine and the separate byte-parity check that exists only to keep those copies in sync.

## Why this PR doesn't merge yet

The existing byte-parity check (`check-legacy-name-engine-parity.sh`, required org-wide on 8 consumer repos) fetches this repo's `main` branch **live** at CI time and does a byte comparison. Merging this PR while that check is still what the 8 repos are gated on would instantly desync every vendored copy and fail all 8 — the same failure mode as this morning's org-wide outage, this time triggered by a merge here instead of a consumer PR.

This PR exists now specifically so its branch head can be tagged and pointed at by a scoped test ruleset (dry-running `legacy-name-ratchet.yml` against one low-risk consumer, then all 8) *before* anything merges to `main` — full sequencing in the tracking plan. One thing this PR's own CI *can* prove right now, with no ruleset involved: that `legacy-name-ratchet.yml`'s self-identification step fails closed on an ordinary PR like this one, where `job.workflow_repository`/`job.workflow_sha` aren't populated the way they would be under a real required-workflow invocation. Expect its own check to fail here — that's the point of opening this PR before merging it.

## Testing

- 184/184 existing + 2 new tests pass (`pytest tests/test_legacy_name_ratchet.py`).
- Both new tests confirmed to fail (red) when the fix is reverted, confirmed to pass when restored.
- `ruff check` / `ruff format --check` clean.
- `python3 scripts/legacy_name_ratchet.py --base origin/main` — no new lines.
- `actionlint legacy-name-ratchet.yml` — clean.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/.github/workflows/legacy-name-ratchet.yml
+++ b/.github/workflows/legacy-name-ratchet.yml
@@ -20,7 +20,9 @@ name: Legacy-name ratchet
 #
 # It is NOT wired into this repo's own on-PR checks: see ci.yml's own
 # "legacy-name-ratchet" job for that. This file exists to be pointed at by
-# other repos' required-workflow rulesets, pinned by tag, never by branch.
+# other repos' required-workflow rulesets, pinned by branch (ratchet-stable),
+# never by tag -- confirmed empirically that tag refs never fire for this
+# ruleset rule type, despite GitHub's docs saying both are supported.
 on:
   pull_request:
   merge_group:
@@ -86,7 +88,7 @@ jobs:
           sparse-checkout-cone-mode: false
 
       - name: Set up Python
-        uses: actions/setup-python@v7
+        uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7.0.0
         with:
           python-version: "3.13"
 
--- a/.github/workflows/legacy-name-ratchet.yml
+++ b/.github/workflows/legacy-name-ratchet.yml
@@ -1,0 +1,140 @@
+name: Legacy-name ratchet
+
+# The central copy of hard rule 7's gate, run as an org-required workflow
+# against every consumer repo's PRs directly -- see docs/plans/rebrand-sunset-plan.md.
+#
+# Nine repos used to each vendor scripts/legacy_name_ratchet.py byte-for-byte
+# and run it locally, with a SEPARATE required workflow (legacy-name-engine-
+# parity.yml, hosted in caura-onprem-installer) doing nothing but confirming
+# the eight copies stayed byte-identical to this one. That parity check is
+# what failed with no bypass actor and blocked unrelated PRs org-wide the day
+# this file was written -- and every consumer's own vendored engine still had
+# a bug this file's own two commits fix, discovered because ONE of the nine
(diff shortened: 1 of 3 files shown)


Eldad-Caura:
Checked both High findings against this PR's own already-passing run rather than in the abstract:

**"Workflow-identity check likely reads fields from the wrong context"** — not the case here. Pulled the raw log for the "Identify the ratchet gate source" step on this exact commit's run (job `101016603834`, run `33870970559`): `toJSON(job)` genuinely contains `workflow_repository`, `workflow_sha`, `workflow_ref`, `workflow_file_path`, and `check_run_id`. This is a real field set GitHub populates on the `job` context specifically for a workflow invoked this way (required-workflow / org-ruleset), separate from the generic `github` context the review's citation covers. The step extracted `repository=caura-ai/caura` and a valid 40-hex SHA and passed.

**"Cross-repo checkout has no explicit token"** — moot: `caura-ai/caura` is a public repo, and the very next step, "Check out the ratchet engine source," completed successfully on that same run with no token, which is exactly the expected behavior for an anonymous checkout of a public repo. This would only be a real problem if `caura` were private.

Both were tested empirically earlier in this effort (a live run proving the happy path, plus adversarial `JOB_CONTEXT` payloads proving fail-closed behavior), not just assumed — this reply is that evidence, re-verified against this PR's actual commit rather than cited from memory.

One real issue the review's scrutiny of this file did prompt me to catch on my own re-read, now fixed in 243ba432: the header comment still said "pinned by tag, never by branch," stale since this design moved to pinning via the `ratchet-stable` branch (tags were confirmed not to fire for this ruleset rule type). Also took the low-severity pinning-consistency suggestion and pinned `actions/setup-python` to a commit SHA to match `actions/checkout`'s pinning in the same file.

Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #1277, opened 2026-09-04
fix(usage): stop metering bulk writes that wrote nothing

Eldad-Caura:
Follow-up to #1274, from measuring the same incident's blast radius against production.

## The defect

`bulk_check_and_increment(body.tenant_id, len(body.items))` ran **before** the write, and no failure path gave it back. A batch that persisted nothing was billed for every item in it.

On its own that would round out over time. The retry contract makes it compound instead: core-api answers a storage failure by telling the client to retry with the same `X-Bulk-Attempt-Id`, so **one un-writable batch is charged once per attempt, indefinitely.**

Measured on the 2026-09-02 bulk incident: ~648 attempts/hour for 41 hours against a 100-item batch that could never compile. On the order of **2.7M metered write units for zero rows**, against a tenant whose real write rate is ~150/hour.

## The change

Metering moves to after the write. Every failure path raises before it, so a failed batch now charges nothing.

**Safe to move because nothing here gates.** `allowed` has no reader in core-api — `usage_service._meter`'s own docstring says so, and notes that enforcement travels via `x-org-read-only` instead — so this call only records. `enforce_fleet_write` *is* a gate and stays where it was, ahead of the write.

**Not a repricing.** Still `len(body.items)` rather than `result.created`, deliberately: the amount charged for a *successful* batch is byte-identical to before. This fixes charging on failure and nothing else. Whether per-item duplicates and per-item errors should be charged is a billing decision, and `meters_mcp_bulk_write`'s docstring is the standing precedent in this file for not shipping one as a deploy side effect.

**Idempotent replays are unaffected.** They return at the `cached_replay` check before this function is entered, so they are neither charged twice nor newly charged — and the comment there ("Replays bypass the per-tenant slot and the bulk quota-increment") stays accurate.

## The tests needed fixing before they meant anything

Worth calling out, because the first version of them was worthless.

`get_test_auth()` returns the **admin** API key, which yields `AuthContext(tenant_id=None)`, and the route guards metering with `if auth.tenant_id:`. So under the default fixture the meter is never called at all, and a "was not metered" assertion passes regardless of what the code does. Two of these tests were green against the unfixed route.

They now override `get_auth_context` to supply a tenant-scoped credential. With that override they fail without the fix, reporting the batch metered as `[('default', 1)]`. Verified by reverting the route change and watching them go red.

The third test is the control: a successful batch must still bill, and still bill `len(items)`. Deleting the charge outright would satisfy the other two, so without this one the "fix" could ship as a free-writes bug.

## Verification

25 tests in `tests/test_bulk_atomicity.py` pass. `ruff check` and `ruff format --check` clean at CI's exact separate scopes; `mypy` clean on `core-api/src`.

No regressions, by the same method as #1274 — revert, run the full suite, save the sorted failure names, restore, re-run, diff: **identical failure sets, 18 both ways**. Those 18 are environmental (the root conftest provisions via `create_all`, so migration 034's tsvector trigger is absent and FTS scores come back `None`), not properties of the codebase.

## Not addressed here

The ~2.7M units already recorded are historical and this change does not retract them. If that counter feeds over-plan enforcement or invoicing for `arkash24-4d270c`, it needs a separate correction — flagging rather than fixing, because retracting recorded usage is a billing action, not a code one.

### DIFF
--- a/core-api/src/core_api/routes/memories.py
+++ b/core-api/src/core_api/routes/memories.py
@@ -1370,14 +1370,12 @@ async def _write_memories_bulk_inner(
     )
     if not body.fleet_id and agent.get("fleet_id"):
         body.fleet_id = agent["fleet_id"]
-    usage = None
-    if auth.tenant_id:  # skip enforcement + metering for admin
+    if auth.tenant_id:  # skip enforcement for admin
         await enforce_fleet_write(body.tenant_id, body.agent_id, body.fleet_id)
-        usage = await bulk_check_and_increment(body.tenant_id, len(body.items))
-    if usage:
-        response.headers["X-RateLimit-Limit"] = str(usage.get("limit", "unlimited"))
-        response.headers["X-RateLimit-Remaining"] = str(usage.get("remaining", "unlimited"))
 
+    # Metering deliberately does NOT happen here — see after the write. This
+    # used to be ``usage = await bulk_check_and_increment(...)`` on this line,
+    # before the try, and no failure path gave it back.
     try:
         result = await asyncio.wait_for(
             create_memories_bulk(body, bulk_attempt_id=bulk_attempt_id),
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

Eldad-Caura:
@erni-a ready for your review — follow-up to #1274, all gates clean.

**Head `1c8f2788b15601c80ec62eecbcd606fc64b93393`**: 13 check-runs, 0 non-clean (11 success, 2 skipped by design). Verified by head SHA via `gh api repos/caura-ai/caura/commits/$SHA/check-runs`. Review verdict is a real posted comment on this head, not a workflow status: no issues found.

**What it fixes.** `bulk_check_and_increment(tenant_id, len(items))` ran before the bulk write and no failure path refunded it, so a batch that persisted nothing was billed for every item. Alone that rounds out over time; the retry contract makes it compound instead — core-api answers a storage failure by telling the client to retry with the same `X-Bulk-Attempt-Id`, so one un-writable batch is charged once per attempt, indefinitely. Measured against prod on the 2026-09-02 incident: ~648 attempts/hour for 41 hours on a 100-item batch, on the order of 2.7M metered write units for zero rows, against a tenant whose real rate is ~150/hour.

**Two things worth your eye:**

1. **Why moving it is safe rather than a gate change.** `allowed` has no reader in core-api — `usage_service._meter`'s own docstring says so and notes enforcement travels via `x-org-read-only` — so this call only records. `enforce_fleet_write` IS a gate and stays ahead of the write. Idempotent replays already returned at the `cached_replay` check before this function and still do, so they are neither charged twice nor newly charged.

2. **Deliberately not a repricing.** Still `len(body.items)` rather than `result.created`, so the amount charged for a successful batch is unchanged. Whether per-item duplicates and errors should be charged is a billing decision, and `meters_mcp_bulk_write`'s docstring is the standing precedent in this file for not shipping one as a deploy side effect. Say the word if you'd rather it charged `created` and I'll do it as its own change.

**A test trap this surfaced, which is worth knowing independently of this PR.** `get_test_auth()` returns the ADMIN key, which yields `AuthContext(tenant_id=None)`, and the bulk route guards metering with `if auth.tenant_id:`. So under the default `client` fixture the meter is never called and a "was not metered" assertion passes whatever the code does — two of these three tests were green against the unfixed route until they overrode `get_auth_context`. Any test asserting on metering, quota, or per-tenant enforcement through that fixture is asserting nothing.

**Not addressed here:** the ~2.7M units already recorded are historical and this does not retract them. If that counter feeds over-plan enforcement or invoicing for `arkash24-4d270c`, it needs a separate correction — that is a billing action, not a code one.

Leaving the merge to you. Full reasoning is in the PR description.

erni-a (approved):

PR #1278, opened 2026-09-04
fix(autochunk): degrade a failed child embed instead of 500ing a persisted write

Eldad-Caura:
Closes audit finding **H-09**.

## The defect

`_handle_auto_chunk_from_ctx` commits the parent, then batch-embeds the children:

```python
parent = await _create_memory_or_409(...)   # committed
...
child_embeddings = await get_embeddings_batch(child_texts, tenant_config, background=False)
```

`get_embeddings_batch` raises on every provider-side error, gate saturation, quota and misconfig — **its own docstring says both bulk callers wrap it** — and this third caller did not. The call also sits **outside** the request's `try/finally`, because the auto-chunk branch returns before that block is entered. So the exception escaped as an error response for a write that had already persisted.

## Three harms, and the third is what made it unrecoverable

1. **Silent data loss.** Zero children written, while the parent's metadata claimed `auto_chunked` and `child_count=N`. The facts were gone, the source document unrecallable, and nothing recorded why.
2. **No backfills.** Parent enrich, entity extraction and parent embed all sit *below* the embed and never ran, so a parent inserted with `embedding=None` stayed unembedded — the 2026-07-27 stranded-rows shape.
3. **A permanent wedge.** The retry re-chunks, produces the same full-document `content_hash`, and migration 040's live-row uniqueness answers **409 against the childless parent**. The children could never be written without deleting the parent by hand.

## Why this is unambiguous rather than a judgement call

Two asymmetries, both inside the same function:

- **The parent's own embed already degrades** to `None` on these same failures, so the parent commits unembedded rather than failing.
- **The child INSERT one line later already degrades**, for precisely this reason. `_insert_children_or_degrade`'s docstring: *"the parent row is ALREADY COMMITTED… raising here would hand the caller a 500 for a write that persisted."*

The child **embed**, between those two, was the only step in the sequence that still raised.

## The fix

Children persist unembedded, carry `embedding_pending`, and each gets a re-embed queued.

**Deliberately not mode-dependent like the bulk path**, which fails the request under `inline_embedding`. There nothing has persisted when the embed fails, so refusing is clean. Here the parent is committed, which inverts the trade: refusing loses the children *and* wedges every retry, while degrading keeps the facts and leaves a repair queued.

**`embedding_pending`** is set because `MemoryOut.metadata` documents an *absent* flag as "that stage ran inline" — leaving it off states the opposite of the truth for these rows and makes them indistinguishable from fully-embedded ones. Same flag the atomic-fact fan-out sets for the same reason.

**The re-embed is scheduled explicitly**, not left to the nightly sweep: `embed_backfill_enabled` defaults **FALSE**, and a deployment that never enabled it is how ~430 memories were stranded in the incident this module already carries a postmortem for. `_schedule_embed_or_reembed` publishes `EMBED_REQUESTED` in deferred mode and retries in-process otherwise, so it works in both.

**`_insert_children_or_degrade` now returns** `create_memories`' per-item results, which is what makes the scheduling possible. Payloads are paired to results by **`client_request_id`**, not by position — see round 2 below for why, given that storage does guarantee input order. The return distinguishes `None` ("nothing was written") from `[]` ("nothing to write"), which round 3 below shows is not a cosmetic difference. The task is logged against the **child**, not the parent; a backfill recorded against the wrong row repairs nothing and misdirects the operator.

**The legacy handler had the same gap and is now fixed too**, via the same helpers — see the review round below.

## Review round: the legacy path, and why my reason for skipping it was wrong

Review flagged the legacy auto-chunk handler as carrying the identical defect. It was right, and **my stated justification was factually wrong.**

I said the path was *dead* behind `_USE_PIPELINE_WRITE=True`. The flag's own comment says otherwise:

> If an emergency rollback is needed, flip these to False and ship a hotfix

So it is **dormant, not dead** — a designated rollback lever. And the correlation is adverse in the worst way: H-09 trips on *embedding-provider degradation*, which is plausibly the very reason someone pulls that lever. The defect would have resurfaced during exactly the incident the rollback exists for, at the moment the system was least able to absorb it.

**Fixed via shared helpers rather than a second copy**, which is the review's own closing suggestion and the right call independently: H-09 existed in two places at once *because* the auto-chunk sequence was written twice. Three helpers now carry the policy —

- `_embed_children_or_degrade` — the embed with its degrade
- `_mark_child_embedding_pending` — the flag, with the reason it exists
- `_queue_child_reembeds` — the per-child repair, with the no-id log

— and both handlers route through all three. **A test asserts the raw `get_embeddings_batch(child_texts…)` call appears in exactly ONE place** and that both handlers call each helper, so a future change to one cannot silently miss the other.

The legacy tests drive `create_memory` with the flag flipped, so the dispatch that selects the rollback handler is part of what is under test rather than assumed.

Getting there surfaced two things the pipeline path hides: the legacy handler runs **two dedup pre-checks** the pipeline path does not (both answer truthy off a bare `AsyncMock`, so it 409'd before reaching the embed), and it **resolves its own config** rather than taking the ctx one, so `auto_chunk_enabled` has to be supplied separately or the branch under test is never entered and the run looks like a pass. Both are now stubbed explicitly with comments saying why.

## Review round 2: the payload-to-result join

Review read the positional `zip` in `_queue_child_reembeds` as a live mispairing bug, citing `create_memories_bulk`'s comment that *"Postgres `RETURNING` order is unspecified for `ON CONFLICT DO NOTHING`"*.

**That premise is wrong at this boundary.** `memory_add_all` collects `RETURNING` into a dict keyed by `client_request_id` and then builds its response by iterating the **input**, so one-entry-per-item in input order holds by construction — and all three layers the value crosses document it.

**The fix went in anyway**, because the trade is one-sided:

- **Zero cost.** `client_request_id` is mandatory on every item (`memory_add_all` refuses a batch without it) and both handlers already mint one per child.
- **Asymmetric consequence.** A mispaired entry would not lose a repair — it would embed one child's text and persist the vector against **another child's row**, leaving that row's embedding silently inconsistent with its own content. No error, no log, on a path that only runs when the provider is already degraded.
- A **missing** entry now takes the loud no-id branch instead of shifting every subsequent pair by one.

**The comment that produced the finding is corrected too.** It attributed the join to raw SQL behaviour, which reads as a claim that the *response* is unordered. The practice stays; the reason is now accurate. That is this backlog's usual defect class inverted — a comment asserting a property the code doesn't have normally hides a missing check; this one invented one.

**The Low was correct.** `_drop_duplicate_facts` cited "both auto-chunk handlers discard `create_memories`' return value entirely", which this PR falsifies. Corrected — and the underlying reasoning survives: results join by `client_request_id`, so a fact dropped before any payload is built has no entry to misalign against.

## Review round 3: the join's own regression

Review found that the `client_request_id` join introduced a false alarm — and it was right.

`_insert_children_or_degrade` returned `[]` both when a duplicate refusal wrote **nothing** and when an insert succeeded. With the old positional `zip`, `zip(payloads, [])` iterated zero times and the case was unreachable. With the join, every degraded payload found no matching entry and logged **"child persisted unembedded … NO re-embed scheduled"** — N ERRORs for rows that never reached the table.

Reproduced before fixing: three false ERRORs, zero rows written.

The severity is about *when*: embed degraded plus children already present **is** the retry-after-wedge case this PR targets, so the alarms would fire mid-incident and send an on-call engineer hunting orphaned rows that don't exist, while the real cause sits one correct WARNING above.

**Fix:** `None` for "nothing was written", `[]` for "nothing to write", and `_queue_child_reembeds` returns silently on `None` — signalling the reason rather than inferring it from emptiness. A **non-empty** payload list against an **empty** result list is a broken contract rather than a per-child outcome, so it keeps one batch-level ERROR instead of N.

Also (Low, correct): the embed-degrade warning said *"with a re-embed queued"* before the insert had run, asserting an outcome that is settled two steps later and can be no for every child — this module's own bug class in a log line. Now conditional.

## Review round 4: the repair step could itself raise

The sharpest of the four. This PR's whole argument is *nothing may raise after the parent commits* — and the step added to enforce it was unguarded. `UUID(str(child_id))` runs after the parent **and** the children are committed, and outside the request's `try/finally`, so a malformed id answers a fully-persisted write with a 500. Not an analogue of H-09 — H-09, relocated from the embed step to the repair step. Reproduced: `ValueError('badly formed hexadecimal UUID string')` escaping the request.

**Guarded per child, not around the loop**, and the difference is the whole value of it: a catch around the loop stops the escape while silently cancelling every remaining child's repair, turning one unrepaired row into N — a worse outcome than the one prevented, reached by code that looks like it handled the problem.

`payload["client_request_id"]` also became `.get()`: unreachable today, but a payload without one cannot be joined to anything, so it belongs in the no-id branch rather than raising a `KeyError` out of a committed request.

## Review round 5: the other half of the sentence

Round 5 traced the full call graph and found the logic sound — one Low left, and it was mine from round 3. I had made *"with a re-embed queued"* conditional and left *"persisting N child(ren) unembedded"* asserting the same thing one clause earlier, in the same line.

The reason it matters: in a duplicate refusal those two WARNINGs land **adjacent in the same incident's logs**, so the unconditional wording reads as contradicting the line directly beneath it. Both clauses are conditional now, and the test pins both.

## Tests

**Five fail without the pipeline guard:** the escape itself, the children not being written, the pending flag, the queued repair, and the no-id log.

Reverting only the **legacy** call site fails two more: the legacy behavioural test and the anti-drift test — which is the point of having both.

**Two more from round 3:** the duplicate-refusal test (no repairs queued, and **no ERROR records at all**, with an anti-vacuity assertion that the refusal WARNING *is* present so it can't pass by never reaching the refusal), and the degrade-warning test, which fails with the old unconditional wording restored.

**Restoring the `zip` fails one more:** the rotated-results test. The fake returns bulk results rotated by one and asserts each child id is paired with its own content and hash. Rotation rather than reversal on purpose — reversing an odd-length list leaves the middle element on its own index, so a positional join would still pair one child correctly and the test could read as a partial pass. Its docstring records that the storage guarantee holds today and that this is a contract-drift probe, so nobody later reads it as evidence of a live bug.

**Two are over-refusal guards.** The healthy path must schedule **nothing** — otherwise every ordinary auto-chunk write queues N pointless tasks. And a child whose insert returns no usable id must be logged **loudly** rather than dropped quietly: that row is unembedded with no repair queued, which is strictly worse than the counted case, and only the off-by-default sweep would ever find it.

A further test asserts that log carries **no memory content**. The bulk result belongs to a row holding the fact text, so interpolating it would put content — and any PII in it — into an ERROR log.

### One test needed fixing, and the probe is what caught it

The pending-flag assertion used `all()` over the children. `all()` over an empty list is `True`, and the pre-fix behaviour writes **no children at all** — so the test **passed with the guard removed**. It now asserts the count first, with a comment saying why. Probe re-run: four failures became five.

That is the second time in this backlog that "confirm it fails without the fix" caught a test that would have shipped green and proved nothing.

## Verification

- Full root suite: **6067 passed, 5 skipped, 1 xfailed, 0 failed**, run as `.venv/bin/python -m pytest`.
- The 42 existing auto-chunk tests (`test_h856_*`, `test_h852_*`, `test_child_dedup_*`) pass unchanged — the return-value change to `_insert_children_or_degrade` is additive.
- `legacy_name_ratchet.py` → *No new lines.* · `do_not_touch_sentinel.py` → *All 39 protected strings survive.* · `tenant_scope_gate.py` → no allowlist movement. All after `git add`.
- No open PR touches this subsystem (checked before starting, per the lesson from the H-12 duplicate).
- Rebased onto `origin/main` at `fb64377f`; head `005d8740`. The remote branch has been rebased under me twice now; each time I inspected the remote head before pushing rather than widening the lease — this round by **patch-id**, which matched my own pre-review commit exactly. Re-ran the ratchet and the full suite on the new base, since the commit main gained (#1275) changes the ratchet itself.

### DIFF
--- a/core-api/src/core_api/services/memory_service.py
+++ b/core-api/src/core_api/services/memory_service.py
@@ -296,10 +296,18 @@ def _drop_duplicate_facts(
       discard would be the dominant cost of deduping at all.
 
     Dropping rather than collapsing-and-reporting is deliberate: these children
-    are derived rows, not caller-submitted items. Nothing upstream holds an
-    index into them — both auto-chunk handlers discard ``create_memories``'
-    return value entirely — so there is no per-item result to point at a
-    survivor, and the parent memory is what the caller gets back either way.
+    are derived rows, not caller-submitted items. Nothing the CALLER holds
+    indexes into them — the parent memory is what a caller gets back either
+    way — so there is no per-item result to point at a survivor.
+
+    Both auto-chunk handlers do now consume ``create_memories``' return value
+    (H-09, to queue a re-embed for a child that landed unembedded), which an
+    earlier version of this docstring cited as "discarded entirely". That does
+    not reopen the question: the results are joined by ``client_request_id``,
+    which each surviving payload carries, so a fact dropped HERE — before any
+    payload is built — simply has no entry to be misaligned against. It is a
+    positional index into the pre-dedup facts that this ordering would break,
+    and nothing takes one.
     """
     kept: list[dict] = []
     seen: set[str] = set()
@@ -345,8 +353,24 @@ async def _insert_children_or_degrade(
     tenant_id: str,
     parent_id: str,
     source: str,
-) -> None:
+) -> list[dict] | None:
     """Insert auto-chunk children; a duplicate refusal degrades, never raises.
+
+    Returns ``create_memories``' per-item results — one entry per input
+    payload, each carrying its ``client_request_id`` and its ``id``, the
+    latter populated whether or not this attempt was the one that committed
+    the row. Callers need them to queue a re-embed for a child inserted
+    without a vector.
+
+    ``None`` means NOTHING WAS WRITTEN by this call (the duplicate refusal
+    below); ``[]`` means there was nothing to write. The distinction is not
+    cosmetic — see :func:`_queue_child_reembeds`, which must not report N
+    unrepairable rows for a batch that never reached the table.
+
+    Consumers pair results to payloads by ``client_request_id`` rather than by
+    position — see ``_queue_child_reembeds``. The order IS input order, but
+    describing the return by its order invites a positional join, and the
+    mispairing that would cause on this path is silent.
 
     The parent row is ALREADY COMMITTED by the time this runs. Raising here would
     hand the caller a 500 for a write that persisted, and leave the parent
@@ -370,10 +394,10 @@ async def _insert_children_or_degrade(
     and the storage-side statement would build an INSERT with no VALUES.
     """
     if not payloads:
-        return
+        return []
     async with per_tenant_storage_slot("storage_write", tenant_id):
         try:
-            await get_storage_client().create_memories(payloads)
+            return await get_storage_client().create_memories(payloads)
         except DuplicateMemoryError:
             logger.warning(
                 "auto-chunk children refused as duplicates; parent kept without them",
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
Right on both counts, and **my stated justification was factually wrong** — fixed in `4d93b31d`.

I said the legacy path was *dead* behind `_USE_PIPELINE_WRITE=True`. The flag's own comment says otherwise:

> If an emergency rollback is needed, flip these to False and ship a hotfix

So it is **dormant, not dead**. I asserted a property of that path without reading the six lines above the flag, which is the same failure mode as the comments this backlog keeps finding: a claim in prose that the code does not support.

Your reachability argument is the part I'd underline, because it is worse than "some deployments might hit it": H-09 trips on **embedding-provider degradation**, which is plausibly the very reason someone pulls the rollback lever. The defect would have resurfaced during exactly the incident that lever exists for, at the moment the system was least able to absorb it. A dormant path that activates under stress is not low-risk because it is dormant.

**Took your closing suggestion** — shared helpers rather than a second copy — and it was the right call independently. H-09 existed in two places at once *because* the auto-chunk sequence was written twice; copying the degrade policy would have set up the next divergence. Three helpers now carry it (`_embed_children_or_degrade`, `_mark_child_embedding_pending`, `_queue_child_reembeds`), both handlers route through all three, and **a test asserts the raw `get_embeddings_batch(child_texts…)` call appears in exactly one place**. Reverting only the legacy call site now fails two tests — the behavioural one and the anti-drift one.

The legacy tests drive `create_memory` with the flag flipped rather than calling the private handler, so the dispatch that selects the rollback path is under test too.

Two incidental findings while wiring that up, both of which make the legacy path harder to test than the pipeline one and are worth knowing if anyone touches it again: it runs **two dedup pre-checks** the pipeline path does not — both of which answer truthy off a bare `AsyncMock`, so it 409'd before ever reaching the embed — and it **resolves its own config** instead of taking the ctx one, so `auto_chunk_enabled` has to be supplied separately or the branch under test is never entered and the run silently looks like a pass. Both are stubbed explicitly now with comments saying why.

Full suite 6062 passed / 0 failed. Rebased onto `b1867596`; gates all clean after `git add`.

Eldad-Caura:
@claude re-review at head `4d93b31d` please — your High is addressed, and you were right that my reason for skipping it was wrong rather than merely debatable.

The legacy path is fixed. `_USE_PIPELINE_WRITE`'s own comment documents flipping it as the emergency-rollback lever, so it is dormant rather than dead — and since H-09 trips on embedding-provider degradation, the defect would have resurfaced during precisely the incident that lever exists for.

I took your closing suggestion and factored the shared logic into three helpers (`_embed_children_or_degrade`, `_mark_child_embedding_pending`, `_queue_child_reembeds`) called by both handlers, rather than copying the policy. Worth a look at two things specifically:

1. **The anti-drift test** asserts the raw `get_embeddings_batch(child_texts…)` call appears in exactly ONE place and that both handlers route through each helper. I'd rather you tell me now if that is too clever or too brittle than have it fail confusingly in six months.
2. **The legacy tests** flip the flag and enter through `create_memory`, so the dispatch is under test. They also stub two dedup pre-checks and the legacy handler's own `resolve_config` — without those the branch is never entered and the run looks like a pass, which is exactly the vacuity trap, so please check I have not papered over something real there.

Also fixed a vacuous test of my own that the removal probe caught: the pending-flag assertion used `all()` over the children, which is `True` on an empty list, and the pre-fix path writes no children — so it passed with the guard removed. It asserts the count first now.

Full suite 6062 passed / 0 failed; ruff, mypy, ratchet, sentinel and tenant-scope gate all clean after `git add`.

Eldad-Caura:
**The High's premise is wrong, and I've taken the fix anyway.** Both at `f6d53424`. The Low was right and is fixed too.

## The join does not currently mispair

You quoted a real comment, so I traced the claim to the implementation rather than to the comment. `memory_add_all` (`core-storage-api/.../postgres_service.py`) does this:

```python
inserted: dict[str, UUID] = {row.client_request_id: row.id for row in result.all()}
...
out: list[dict] = []
for d in items:                       # <- iterates the INPUT
    crid = d["client_request_id"]
    ...
```

`RETURNING` goes straight into a dict keyed by `client_request_id` and is never read positionally. The response is then built by walking `items`. So **"one entry per item, in input order" holds by construction**, not by luck — and it is documented at all three layers the value crosses: `memory_add_all`, the `/memories/bulk` route, and `storage_client.create_memories`.

The Postgres caveat you cited is real about raw `RETURNING`. It is not true of the value that reaches core-api, because the layer in between already absorbed it.

## Taking the fix regardless

Not as a concession — the trade is one-sided:

- **The cost is zero.** `client_request_id` is *mandatory* on every item (`memory_add_all` raises `BulkValidationError` without it), and both handlers already mint one per child. There is no case where the key is absent.
- **The consequence you described is right, and it's the part that matters.** A mispaired entry does not lose a repair. `_schedule_embed_or_reembed` would embed *this* payload's text and persist the vector against *another* child's row — that row's embedding silently stops matching its own stored content. No error, no log, and only on a path that runs when the provider is already degraded. A guarantee that cheap to stop depending on is not worth depending on across a service boundary we don't deploy in lockstep.
- A **missing** entry now takes the loud no-id branch instead of shifting every subsequent pair by one, which the zip could not do.

## I also fixed the comment that produced this finding

The `create_memories_bulk` comment attributed its join to raw SQL behaviour, which reads as a claim that *the response* is unordered. It isn't — and that mismatch is what generated a High against a sibling caller. The practice stays; the stated reason is now accurate, and says explicitly why the join is kept anyway.

Worth naming: that is the same defect class this whole backlog keeps turning up, inverted. A comment asserting a property the code doesn't have usually hides a missing check. This one **invented** one.

## The Low was correct

`_drop_duplicate_facts` cited *"both auto-chunk handlers discard `create_memories`' return value entirely"* as part of its justification, and my change falsified it. Corrected — and the underlying reasoning does survive, for a reason worth writing down: results are joined by `client_request_id`, so a fact dropped *before any payload is built* has no entry to be misaligned against. What that dedup ordering would break is a **positional** index into the pre-dedup facts, and nothing takes one.

## Test

New: the fake bulk insert returns results **rotated by one**, then the test asserts each child id is paired with its own content and hash.

Rotation rather than reversal on purpose — reversing an odd-length list leaves the middle element on its own index, so a positional join would still pair one child correctly and the test could read as a partial pass. A rotation moves every element.

Confirmed it fails with the `zip` restored:

```
FAILED test_the_repair_pairs_each_child_id_with_its_own_content
AssertionError: a child's re-embed was queued against another child's row
```

Its docstring says the storage guarantee holds today and that the test is a contract-drift probe, so nobody later reads it as evidence of a live bug.

## Verification

Full suite **6064 passed, 5 skipped, 1 xfailed, 0 failed**. ruff check + `format --check` separately at CI scopes, mypy (2 pre-existing stub errors, untouched file), ratchet *No new lines.*, sentinel *All 39 protected strings survive.*, tenant-scope gate exit 0.

Rebased onto `origin/main` at `fb64377f` — the branch had been rebased under me again; I confirmed by patch-id that the remote head was my own pre-review commit before superseding it, and re-ran the ratchet and the suite on the new base since #1275 changed the ratchet itself.

Eldad-Caura:
@claude re-review at head `f6d53424` please.

Your High's remedy is in; its premise was not correct, and I've said why in the comment above rather than quietly applying the patch — `memory_add_all` builds its response by iterating the input, so input order holds by construction and the `RETURNING` caveat doesn't reach core-api. I took the `client_request_id` join anyway because it costs nothing and the mispairing failure mode you described is silent. Your Low was right and is fixed.

Three things I'd specifically like challenged:

1. **I edited a comment outside my diff** — `create_memories_bulk`'s, because its wording is what generated the finding. Check I corrected it without weakening it: the join must still read as deliberate practice, not as something now optional.
2. **The rotated-results test** encodes an ordering the storage layer does not actually produce. I think that's the right shape for a contract-drift probe and I said so in its docstring, but tell me if a test asserting against a counterfactual is worse than no test here.
3. **`_drop_duplicate_facts`'s corrected docstring** now argues the dedup ordering is safe *because* nothing takes a positional index into the pre-dedup facts. Please check that claim against the current code rather than against my summary of it — it is exactly the kind of prose assertion this PR is about.

Eldad-Caura:
**Both correct. Both fixed at `0bef8c48`** — and the Medium is a regression I introduced one round ago, which is worth saying plainly.

## The Medium

Exactly right, including the reachability argument. Reproduced it before fixing:

```
AssertionError: a duplicate refusal is a fully-explained WARNING-level outcome;
it must not raise ERRORs: [
 'auto-chunk child persisted unembedded but the bulk insert returned no usable id (request_i',
 'auto-chunk child persisted unembedded but the bulk insert returned no usable id (request_i',
 'auto-chunk child persisted unembedded but the bulk insert returned no usable id (request_i']
```

Three ERRORs claiming three rows were persisted unembedded with no repair queued. **Zero rows were written.**

**I caused this last round.** The positional `zip` I replaced paired against an empty list and iterated zero times, so the branch was unreachable; the `client_request_id` join made every degraded payload reach it. The join is still right — but it turned a batch-level outcome into N per-child claims, and I didn't check what the loop did when the batch had no results at all.

Your point about *when* it fires is the one that decides the severity. Embed degraded plus children already present **is** the retry-after-wedge case this PR exists for, so the false alarms would land mid-incident and send an on-call engineer hunting orphaned rows that don't exist — while the real cause sits one WARNING above, already correct.

**Fix:** `_insert_children_or_degrade` now returns `None` for a duplicate refusal and `[]` for an empty batch, and `_queue_child_reembeds` returns silently on `None`. I took your "signal *why*" suggestion rather than inferring it from emptiness, because those are genuinely two different facts.

The third case stays loud, separately: a **non-empty** payload list against an **empty** result list is a broken contract, not a per-child outcome, so it gets one batch-level ERROR rather than N.

## The Low

Also right, and it's this module's own bug class in a log line — asserting an outcome the code doesn't guarantee. The warning fires before the insert, so it cannot know. Now: *"persisting N child(ren) unembedded — a re-embed is queued for each one that lands with a usable id."*

## Tests

Two, both probe-confirmed:

- **Duplicate refusal + degraded embed** → no repairs queued and **no ERROR records at all**, plus an anti-vacuity assertion that the refusal WARNING *is* present, so the test can't pass by never reaching the refusal. Fails without the fix with the output above.
- **The degrade warning must not promise a repair.** Fails with the old wording restored.

Suite **6066 passed, 0 failed**. ruff, mypy, ratchet, sentinel, tenant-scope gate all clean after `git add`; checks were green on `f6d53424` before this push.

## On your three questions from the re-trigger

You didn't answer them directly, so I'm not treating silence as agreement — flagging them as still open: the out-of-diff comment edit in `create_memories_bulk`, whether the rotated-results test asserting a counterfactual ordering is the right shape, and whether `_drop_duplicate_facts`' corrected docstring holds against the code.

Eldad-Caura:
@claude re-review at head `0bef8c48` please. Your Medium and Low are both fixed, and the Medium was a regression I introduced in the previous round — details in the comment above.

Two things I'd like you to attack specifically:

1. **`None` vs `[]` as the signal.** `_insert_children_or_degrade` now returns `None` for "nothing was written" and `[]` for "nothing to write", and the whole correctness of the repair loop rests on a future caller not flattening the two back together with `or []`. You suggested an enum or a small result object. I went with the sentinel because it kept the change to two lines and the docstring states the distinction — but tell me if you think that is too fragile a place to hang this, given the failure mode is silent and only appears mid-incident.
2. **The batch-level ERROR I added** for a non-empty payload list against an empty result list. That is a case I believe is unreachable today (storage returns one entry per item or raises), so it is a contract guard. Check I haven't just moved the false alarm rather than removed it.

Also still open from my previous re-trigger, which the last round didn't address — I am not reading silence as agreement:

- The out-of-diff comment edit in `create_memories_bulk`: corrected without weakening?
- The rotated-results test asserting an ordering storage does not produce: right shape for a contract-drift probe, or worse than no test?
- `_drop_duplicate_facts`' corrected docstring: does its claim hold against the code?

Eldad-Caura:
**Right, and it's the sharpest finding of the four.** Fixed at `c57d8990`.

The whole argument of this PR is *nothing may raise after the parent commits* — and I left the step I added to enforce that unguarded. `UUID(str(child_id))` runs after the parent **and** the children are committed, and outside the request's `try/finally`, so a malformed id answers a fully-persisted write with a 500. That is not an analogue of H-09; it is H-09, relocated from the embed step to the repair step.

Confirmed before fixing:

```
AssertionError: the repair step escaped the request:
ValueError('badly formed hexadecimal UUID string')
```

## Per child, not around the loop

You offered both; they are not equivalent, and the difference is the whole value of the guard.

A catch around the loop stops the escape and **silently cancels every remaining child's repair** — one malformed id turns one unrepaired row into N, each unembedded with nothing queued, findable only by the sweep that defaults off. That is a worse outcome than the one being prevented, reached by code that looks like it handled the problem.

So: per child, `continue` past the bad one, ERROR naming the `client_request_id` and never the payload (it holds the fact text).

**I checked that claim rather than asserting it, and the first attempt was a bad experiment.** I wrapped the loop while leaving the per-child guard nested inside it — the inner one caught everything, the outer never fired, all 14 tests passed, and it proved nothing. Redone with the inner guard actually removed:

```
AssertionError: expected the surviving children to still be repaired, got []
```

So the test does discriminate the two shapes, which is what I wanted to know.

## Also hardened

`payload["client_request_id"]` → `.get()`. Unreachable today (both handlers set it unconditionally), but a payload without one cannot be joined to anything, so it belongs in the no-id branch rather than raising a `KeyError` out of a committed request. Same reasoning as your finding, one line over.

## Test

One, probe-confirmed both ways: **one** child gets a truthy-but-unparseable id — truthy so it clears the no-id branch and actually reaches `UUID()` — and the test asserts nothing escapes, the **other two are still repaired**, and the failure is logged.

Suite **6067 passed, 0 failed**; ruff, mypy, ratchet, sentinel, tenant-scope all clean after `git add`. Checks were green on `0bef8c48` before this push.

## Still open from two re-triggers ago

Not reading silence as agreement, so restating them: the out-of-diff comment edit in `create_memories_bulk` (corrected without weakening?), the rotated-results test asserting an ordering storage doesn't produce (right shape, or worse than no test?), `_drop_duplicate_facts`' corrected docstring (does the claim hold against the code?), and from last round the `None`-vs-`[]` sentinel (too fragile a place to hang a silent failure mode?).

Eldad-Caura:
@claude re-review at head `c57d8990` please. Your Medium is fixed, guarded per child rather than around the loop for the reason in the comment above.

Four rounds in, each has found something real, so I want to ask a different question this time rather than only pointing you at the new diff:

**Is this PR getting worse as it gets more defensive?** Each round has added a guard for a case that is unreachable under today's contracts — an out-of-order bulk result, an id that isn't a UUID, a payload without a request id. Individually each is cheap and the failure modes are silent, which is why I took them. Collectively they may have turned a 40-line fix into something a future reader cannot tell the load-bearing parts of. If you think the ratio has tipped, say which guards you would drop and I will drop them; I would rather hear that now than have this land as a template for the next fix.

Also still open across the last two re-triggers — flagging them again because I am not treating silence as agreement:

- The out-of-diff comment edit in `create_memories_bulk`: corrected without weakening?
- The rotated-results test asserting an ordering storage does not produce: right shape for a contract-drift probe, or worse than no test?
- `_drop_duplicate_facts`' corrected docstring: does its claim hold against the code?
- The `None`-vs-`[]` sentinel: too fragile a place to hang a silent failure mode?

Eldad-Caura:
Right — I fixed half a sentence last round. `005d8740`.

I made *"with a re-embed queued"* conditional and left *"persisting %d child(ren) unembedded"* asserting the same thing one clause earlier, in the same log line. Your framing is the useful part: in a duplicate refusal those two WARNINGs land **adjacent in the same incident's logs**, so the unconditional wording doesn't just overstate — it reads as contradicting the line directly beneath it, at the moment someone is trying to work out what actually happened.

Now: *"N child(ren) will be inserted unembedded if the write proceeds — a re-embed is queued for each one that lands with a usable id."* Both clauses conditional, and the test pins both, with a comment recording that the first draft only did one.

**One of my own assertions was wrong and the test caught it.** I first added `assert run.children == []` as the anti-vacuity check — but that field holds the payloads *submitted*, not rows persisted, so it exists either way. It asserted something false about the harness rather than anything about the code. Replaced with the two checks that do establish the case was reached: the refusal WARNING is present, and no repair was queued.

Probe: restoring `"persisting"` fails the test with the offending line quoted.

Suite **6067 passed, 0 failed**; ruff, mypy, ratchet, sentinel, tenant-scope clean after `git add`. Checks were green on `c57d8990`.

---

**On the four rounds and my "is this getting worse as it gets more defensive" question** — you answered it implicitly by tracing the full call graph and finding the logic sound, which is the reassurance I was after, so I'll take that and stop adding guards. If you'd still drop any of them, say which and I will.

Also noting for the record that the earlier questions went unanswered across three re-triggers — the `create_memories_bulk` comment edit, the rotated-results test's counterfactual ordering, `_drop_duplicate_facts`' corrected docstring, and the `None`-vs-`[]` sentinel. Your round-5 pass did explicitly confirm the last one ("the `None`-vs-`[]` distinction is preserved correctly"), so that one I'll call answered. The other three I'm leaving as noted-and-open rather than assuming approval from silence.

Eldad-Caura:
@claude final pass at head `005d8740` please — the delta since your last review is one log-line wording change and one test, both described above.

Not asking for another sweep of the whole diff; you traced the call graph last round. Just tell me whether anything is left that should block, so I can either fix it or hand this over as ready.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #1279, opened 2026-09-04
docs(ruff): correct do_not_touch_sentinel.py's false byte-identity claim

Eldad-Caura:
## What

Found while migrating `caura-enterprise` off the vendored legacy-name ratchet: its `scripts/ruff.toml` (which explicitly mirrors this repo's byte-for-byte) called `do_not_touch_sentinel.py` "byte-identical across the repos in the parity set" — tracing it back, this repo's own `scripts/ruff.toml` and `tests/ruff.toml` make the identical claim, in the paragraph justifying their `[format]` exclude lists.

## Why it's wrong

Diffed `do_not_touch_sentinel.py` across all nine repos, including this one as the reference point: no two copies match, and sizes range from 13KB to 41KB. This is genuinely per-repo data — `caura-bus`'s own copy says as much about itself ("written from scratch — no other repo's entries copied in"). The claim was likely generalized from `legacy_name_ratchet.py`'s entry next to it, which IS byte-identical fleet-wide with a real enforced check behind it (the org's required-workflow ruleset); that property never applied to the sentinel.

`tests/ruff.toml`'s version of the paragraph named `do_not_touch_sentinel.py` even though that file's own `[format]` exclude list doesn't contain an entry for it at all — pure copy-paste residue from `scripts/ruff.toml`, not describing anything actually in `tests/ruff.toml`.

## What changed

Recategorized both to marker-safety reasoning (`do_not_touch_sentinel.py` carries the same line-positional markers a reformat would displace, which is true regardless of fleet parity) rather than the byte-parity claim that doesn't hold. **The `exclude` and `per-file-ignores` lists themselves are unchanged** — this is a documentation-accuracy fix, not a behavior change.

## Verification

- Both files still parse as valid TOML.
- `ruff check`/`ruff format --check` against `scripts/` and `tests/`, matching CI's exact invocation — clean, both before and after.
- `python3 scripts/legacy_name_ratchet.py --base origin/main` — no new lines.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/scripts/ruff.toml
+++ b/scripts/ruff.toml
@@ -37,8 +37,8 @@
 # written by ``ruff format`` at its default width, so a wider limit does not
 # accommodate them — it tells the formatter to JOIN lines that are correctly
 # wrapped today, and makes the diff bigger. 88 is the minimum-churn value, not a
-# compromise. Both fleet-wide engine files, legacy_name_ratchet.py and
-# do_not_touch_sentinel.py, are already clean at it.
+# compromise. Both marker-carrying files this config format-excludes below,
+# legacy_name_ratchet.py and do_not_touch_sentinel.py, are already clean at it.
 #
 # E501 is ignored, as everywhere in this estate, so line-length governs
 # ``ruff format`` only and never fails ``ruff check``.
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@erni-a pelase approve

erni-a (approved):

PR #1281, opened 2026-09-04
fix(core-api): stamp provenance on bulk re-embed fallbacks

Eldad-Caura:
## The defect

`_reembed_memories_bulk` has five per-item failure fallbacks. All five called
`_schedule_embed_or_reembed` without `content_hash`, and the omission was silent
at every hop:

```
_schedule_embed_or_reembed(content_hash=None)
  -> publishes EMBED_REQUESTED with content_hash=None
  -> handle_embed_request forwards the None
  -> update_memory_embedding writes the vector, leaves embedded_content_hash NULL
```

That NULL is deliberate at the storage layer — *"unknown provenance is honest, a
wrong hash is not"*. The problem is where the row lands: the `unknown_provenance`
bucket, which is documented as meaning **"written before migration 037"** and
which nothing sweeps. Both embedding backfills select `embedding IS NULL`
(`core-worker/backfill.py` via `/memories/null-embedding-ids`, and
`backfill_embeddings.py`), and these rows *have* an embedding. So they are
invisible to the repair paths and to the staleness detector, permanently.

No error, no log, no failed request. Nothing to notice.

**Measured in production 2026-09-04: 241 rows, one tenant, all bulk-created.**
Those rows have been repaired separately (`embedded_content_hash = content_hash`,
verified row-by-row against a recomputed `sha256(tenant:fleet:content)` before the
write). This PR stops the source.

## The fix

`content_hash=_content_hash(tenant_id, fleet_id, content)` on the fallback, and a
new required `fleet_id` parameter to make that possible.

**Why `fleet_id` has to be threaded in.** `_content_hash` covers
`tenant:fleet:content`. A hash computed without the fleet would be *present,
wrong, and read downstream as verified freshness* — worse than the NULL it
replaces. It cannot be derived in the callee: `MemoryEmbedRequest` carries no
fleet, and the fallback paths have no row to read it off (the read may be exactly
what failed).

**Why a scalar is safe.** A bulk batch is single-fleet by construction —
`memory_add_all` rejects a batch whose items disagree on `fleet_id`, and
`BulkMemoryItem` has no per-item fleet — and the route resolves `body.fleet_id`
before the insert. Required rather than defaulted, so a future caller cannot
silently omit it.

**The five call sites are now one nested `_fallback` helper.** This is the part I
would push back on if I were reviewing: the repetition *was* the defect — five
siblings, the same omission at each — so repairing five copies while leaving five
copies preserves the shape that produced it. There is now one place to omit it
from.

## Tests

`test_bulk_reembed_fallback_carries_provenance` is parametrised over **all five**
branches that can reach the fallback — batch raised, embedding-count mismatch,
per-item `None`, `get_memory` raised, PATCH raised — and asserts the **exact**
expected hash rather than "a hash is present", since a wrong-fleet hash would
satisfy the weaker assertion.

**Verified by reverting the fix: all five go red, and each branch's log line
confirms it reached its own code path.** Restored, all green.

`tests/test_embed_provenance_call_sites.py` is a static AST guard modelled on the
existing `test_embedding_calls_are_gated.py`, failing if any
`_schedule_embed_or_reembed` call omits `content_hash`. It covers what the runtime
tests structurally cannot: **a sixth call site added later that bypasses the
helper.** `**kwargs` forwarding is treated as passing — a deliberate false
negative, since failing every legitimate forwarder gets a guard deleted rather
than obeyed.

That guard carries its own meta-test. A guard whose detection logic silently
breaks reports a clean tree forever, which is the same false-clean failure it
exists to catch, so both directions are asserted against source that never runs.

Full suite: **6041 passed vs 6034 on the base commit, +7 = exactly the tests added.
The 18 failures are identical in both runs** (pre-existing local FTS/search
environment gaps, none near this code) — failure sets diffed by name, zero new and
zero fixed.

## Things a reviewer should push on

**1. This is not the deepest possible fix.** `_schedule_embed_or_reembed` has 10
production call sites and, after this PR, all 10 pass a value that is always
`_content_hash(tenant_id, fleet_id, <the content being passed>)`. The parameter is
therefore fully derivable, and the stronger change is to drop `content_hash`
entirely in favour of a required `fleet_id` on the shim, computing the hash
inside. That makes the omission *unrepresentable* rather than *guarded*. I did not
do it here because it touches 10 sites across two more files and I wanted the
241-row bleed stopped on a reviewable diff; the AST guard buys most of the safety
at a fraction of the blast radius. Say the word and I'll do it as a follow-up.

**2. `create_memories_bulk` already computed this hash and throws it away.**
`hashes` (built for the dedup query) is indexed by the same `orig_idx` used to
build `reembed_batch`, and is the value actually written to the row. Widening the
batch tuple to carry it would delete this PR's new parameter, the recomputation,
and the prose defending the scalar-fleet argument. I judged the tuple-shape change
plus its test churn to be the larger diff for the same outcome, but it is a fair
call to make the other way.

**3. The systemic gap this does *not* close.** Nothing anywhere selects
`embedding IS NOT NULL AND embedded_content_hash IS NULL`. The bucket is counted
(`core-operations/tasks.py`) but deliberately non-alerting, on the recorded premise
that it is *"a measurement gap closing, so do not alert on it"* — a premise that
assumes the population only ever drains. This defect made it grow, and the one
number that would have shown 241 was the one chosen not to alert. Alerting on the
**delta** rather than the level would have caught this in hours. Worth its own
issue.

**4. `backfill_embeddings.py` is itself a producer.** Its `UPDATE ... SET
embedding = ...` writes no `embedded_content_hash`, so the OSS sweep moves rows
from `missing` (repairable) to `unknown_provenance` (repaired by nothing). Same
one-line class of fix; out of scope here but it means the population can still
grow via that path.

## Not a defect (checked, and I had this wrong first)

I initially flagged `update_memory` as a fourth site, since it writes
`patch["embedding"]` and `patch["content_hash"]` but never
`patch["embedded_content_hash"]`. It is covered: `memory_update` in
core-storage-api derives the provenance hash for exactly that column combination,
including correctly clearing it when the embedding is None. That site is fine, and
is in fact the model for what the scheduling path lacks.

## Coordination

Open PR #1278 also touches `memory_service.py` and adds its own
`_schedule_embed_or_reembed` call — it already passes `content_hash`, so it
satisfies the new guard and there is no semantic conflict (different function).
Textual conflict risk on that file is low but non-zero depending on merge order.

Not merging this myself.

### DIFF
--- a/core-api/src/core_api/services/memory_service.py
+++ b/core-api/src/core_api/services/memory_service.py
@@ -2649,7 +2649,7 @@ async def create_memories_bulk(
             # just captures the batch-level exception.
             track_task(
                 tracked_task(
-                    _reembed_memories_bulk(reembed_batch, data.tenant_id),
+                    _reembed_memories_bulk(reembed_batch, data.tenant_id, data.fleet_id),
                     f"reembed_bulk[{len(reembed_batch)}]",
                     None,
                     data.tenant_id,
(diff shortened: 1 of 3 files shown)


Eldad-Caura:
@erni-a ready for your review — all gates clean.

**Head `537445a51421ae9ff770984254c76400ed54c87c`**: 14 check-runs, 0 non-clean (12 success, 2 skipped by design). Verified by head SHA via `gh api repos/caura-ai/caura/commits/$SHA/check-runs`. The review verdict is a real posted comment on this head, not a workflow status: no issues found.

**What it fixes.** Five per-item fallbacks in `_reembed_memories_bulk` scheduled a re-embed without `content_hash`. That is silent at every hop — the shim publishes `EMBED_REQUESTED` with None, core-worker forwards it, and `update_memory_embedding` writes the vector while deliberately leaving `embedded_content_hash` NULL. The row lands in `unknown_provenance`, documented as "written before migration 037", and **both** embedding backfills select `embedding IS NULL` — these rows have an embedding, so nothing can ever see them again. 241 production rows, one tenant, all bulk-created. Repaired separately; this stops the source.

**Three things worth your eye:**

1. **I collapsed the five call sites into one helper, and I think that is the actual fix.** Repairing five copies while leaving five copies preserves exactly the shape that produced the bug — five siblings with the same omission. There is now one place to omit it from.

2. **The AST guard is doing the work the runtime tests cannot.** `test_bulk_reembed_fallback_carries_provenance` is parametrised over all five branches and each one fails without the fix (verified by reverting). But no runtime test can catch a *sixth* call site added later that bypasses the helper, which is precisely how this defect spread. `tests/test_embed_provenance_call_sites.py` is modelled on the existing `test_embedding_calls_are_gated.py` and fails statically on any `_schedule_embed_or_reembed` that omits the kwarg. It carries its own meta-test, because a guard with broken detection logic reports a clean tree forever — the same false-clean failure it exists to catch.

3. **This is deliberately not the deepest available fix, and I would understand you asking for the deeper one.** After this PR all 10 production call sites of `_schedule_embed_or_reembed` pass a value that is always `_content_hash(tenant_id, fleet_id, <the content already being passed>)`. The parameter is fully derivable, so the stronger change is to drop `content_hash` in favour of a required `fleet_id` on the shim and compute it inside — making the omission unrepresentable rather than guarded. I stopped short because that touches 10 sites across two more files and I wanted the bleed stopped on a reviewable diff. Separately, `create_memories_bulk` already computes these exact hashes at line 1840 for the dedup query and discards them; carrying them in `reembed_batch` would delete this PR's new parameter entirely. Both are fair to push back on and I will do either as a follow-up.

**Two systemic gaps this does NOT close**, both in the PR description and both worth their own issues:

- **Nothing anywhere selects `embedding IS NOT NULL AND embedded_content_hash IS NULL`.** The bucket is counted in `core-operations/tasks.py` but deliberately non-alerting, on the recorded premise that it is "a measurement gap closing, so do not alert on it" — which assumes it only ever drains. This defect made it grow, and the single number that would have surfaced 241 rows was the one chosen not to alert on. Alerting on the delta rather than the level would have caught this in hours instead of a fortnight.
- **`backfill_embeddings.py` is itself a producer.** Its `UPDATE ... SET embedding = ...` writes no `embedded_content_hash`, so the OSS sweep launders rows from `missing` (repairable) into `unknown_provenance` (repaired by nothing). Same one-line class of fix.

**A correction to my own first read, since it is in the transcript.** I initially flagged `update_memory` as a fourth defective site because it writes `patch["embedding"]` and `patch["content_hash"]` but never `patch["embedded_content_hash"]`. That is wrong — `memory_update` in core-storage-api derives the provenance hash for exactly that column combination, including clearing it when the embedding is None. That site is fine, and is the model for what the scheduling path lacked.

**Coordination:** #1278 also touches `memory_service.py` and adds its own `_schedule_embed_or_reembed` call. It already passes `content_hash`, so it satisfies the new guard — no semantic conflict, different function. Textual conflict risk on that file is low but depends on merge order.

Leaving the merge to you.

erni-a (approved):

PR #1280, opened 2026-09-04
release(clients): bump caura-client to 1.0.2, caura to 1.0.1

Eldad-Caura:
## Summary

Neither of these versions is CI-proven under its current name: both `caura-client` 1.0.1 and `caura` 1.0.0 were published manually (`uv build && uv publish`, per [#870](https://github.com/caura-ai/caura/pull/870)'s own description), and no `caura-client-v*` or `caura-meta-v*` tag has ever been pushed. The first real tag-triggered release for either would hit PyPI's `400 File already exists` on the version already live from the manual publish — this is the exact failure already seen on `caura-client`'s 2026-09-01 CI run. This PR bumps both versions past what's already on PyPI so the first tag-triggered release is a clean publish, not a repeat of that failure.

- `clients/python/pyproject.toml`: `caura-client` 1.0.1 → 1.0.2
- `clients/python/src/caura_client/__init__.py`: `__version__` 1.0.1 → 1.0.2 (kept in sync — separate from `pyproject.toml`, not covered by any test)
- `clients/caura-meta/pyproject.toml`: `caura` 1.0.0 → 1.0.1

`clients/caura-sdk-meta/pyproject.toml` is deliberately untouched — `caura-sdk` doesn't exist on PyPI yet, so 1.0.0 is a clean first publish there.

**Checked before editing, not assumed**: `release-please-config.json` tracks exactly two packages, `.` (core-api/core-worker/core-storage-api via `extra-files`) and `plugin` — neither `clients/python` nor `clients/caura-meta` appears anywhere in it, confirmed by reading the full config and cross-checking `.github/workflows/release-please.yml` for any second/hidden config. Safe to hand-edit these version strings.

`caura-client`'s dependency on `caura` metapackage side is `caura-client>=1.0.0` (open lower bound) — needs no change for this bump.

## Test plan

- [x] Both edited `pyproject.toml` files parse as valid TOML (`tomllib.load`)
- [x] Confirmed via `release-please-config.json` + `.release-please-manifest.json` that neither file is release-please-managed
- [x] Confirmed no test in either package's `tests/` asserts the old version string
- [x] No `scripts/do_not_touch_sentinel.py` pin references either file

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1283, opened 2026-09-04
fix(bulk): key per-item idempotency on content, not on position

Eldad-Caura:
Closes audit finding **H-08**.

## The defect

`create_memories_bulk` derived each item's `client_request_id` positionally:

```python
item_request_id = f"{bulk_attempt_id}:{i}"   # i = index within THIS body
```

Stable only under a precondition the caller had to uphold and could not see — *same body + same attempt id ⇒ same per-item id*. **A retry carrying only the items that did not succeed is a different body**, so every survivor shifts down onto an index another item's row already claimed.

## Why the loss is silent and total

1. The survivor's content is new, so it clears the content-hash dedup and goes to the write path.
2. Storage's `ON CONFLICT DO NOTHING` on `ix_memories_attempt_unique` sees the id already taken and **skips the insert**.
3. The follow-up re-query resolves that id to the **foreign row**, returned `was_inserted=False`.
4. That reads as `duplicate_attempt` carrying the foreign row's id — so the response says `created=0, errors=0`, indistinguishable from a clean retry, while the resent content was never written and never will be on any number of further retries.

## Reproduced against Postgres before fixing

Two items committed under an attempt id, then two **different** items sent under the same one:

```
{'created': 0, 'duplicates': 2, 'errors': 0, 'results': [
  {'index': 0, 'client_request_id': 'partial-dcebaf3c:0',
   'status': 'duplicate_attempt', 'id': '604ef8e9-…'},
  {'index': 1, 'client_request_id': 'partial-dcebaf3c:1',
   'status': 'duplicate_attempt', 'id': 'cb8034b6-…'}]}
```

Neither resent fact exists. Both ids belong to the **first** attempt's rows.

## This is not an exotic caller mistake

- **The route invites it.** It answers **207** with per-item results naming exactly which items failed, and tells clients a retry of the same logical batch reuses the same attempt id. Resending just the failures is the obvious reading of that pair.
- **`ingest_commit` does it structurally.** Its pre-loop dedup removes already-created facts *before* building the body, so its own documented same-`run_id` retry path shrinks the body every time. The pre-dedup is what makes retries cheap and it is also what shifts the indices.

## The fix

```python
item_request_id = f"{bulk_attempt_id}:{hashes[i][:16]}"
```

Content is what the key was always trying to name — *"this logical row within this attempt"* — so keying on it **removes** the precondition rather than documenting it harder. A partial retry becomes just a smaller batch.

**Cost is nil.** `hashes` is already computed above for the dedup gate, so the key and the dedup now agree on the same value. 16 hex chars is 64 bits over a batch capped at 100 items, and identical content within one batch never reaches the write path anyway (`seen_hashes` collapses it first) — so the only collisions this has to rule out are accidental ones.

## Tests

**One new test, and it asserts persistence rather than labelling**: it reads the rows back by id and checks their **content**, so a relabelling that still wrote nothing would not pass. It also asserts the returned ids are disjoint from the first attempt's — the "mapped onto another fact's row" half of the defect. Pairing goes through each result's own `index`, not list order, so it checks the mapping the response claims.

It runs over HTTP against a real Postgres, so the `ON CONFLICT` behaviour is the real one rather than a fake's.

**Two existing assertions pinned the positional shape** via `endswith(f":{index}")`. Their stated intent was *"every result carries its server-derived per-item attempt id — callers can use this to correlate with retries"*, and content keying serves that better than position did. They now check the property — one distinct token per item, scoped to the attempt — instead of the derivation. Pinning the shape would have pinned the defect.

## Docstrings — and one that turned out to be wire-visible

All four places that stated the positional form are corrected: `create_memories_bulk`, the bulk route, `memory_add_all`, and `_auto_chunk_request_id`.

**The bulk route's docstring is not internal.** `POST /api/v1/memories/bulk` is one of the eight frozen-v1 broker operations `caura-daemon` calls, and its docstring is published as that operation's OpenAPI **description** — so editing it made `core-api/openapi.broker.json` stale and CI failed.

Worth recording because of *what* caught it: ruff, mypy, both test roots and all three naming gates passed clean locally, and none of them look at the generated contract. Regenerated and committed; the diff is one line of description text with no schema change, so the companion `oasdiff` breaking-change gate is unaffected.

## Deliberately not here: H-07

H-07 is the >100-fact 500 in the same function, and its fix is to chunk survivors into batches. **Under positional keys that would have reintroduced this exact bug one layer up** — batch 2's item 0 taking batch 1's item 0's key. Content keying is what makes that fix safe, so H-07 is the natural follow-up rather than part of this change.

## Verification

- Full root suite: **6075 passed, 5 skipped, 1 xfailed, 0 failed**.
- `core-storage-api/tests/`: **328 passed**, on its own scratch database (the suite refuses to share the root suite's, by design).
- `mypy` clean on `core-storage-api/src/`; `core-api/src/` clean apart from 2 pre-existing `types-python-dateutil` stub errors in an untouched file.
- `legacy_name_ratchet.py` → *No new lines.* · `do_not_touch_sentinel.py` → *All 39 protected strings survive.* · `tenant_scope_gate.py` → exit 0. All after `git add`.
- Checked for an open PR on this subsystem before starting; only #524 touches ingest commit and does not overlap.
- Branched from `origin/main`, rebased onto `46eb44ab` as main moved under it (#1278, #1280, #1281). Each time the branch was rebased remotely I confirmed by **patch-id** that the remote head was my own commit before superseding it.


Eldad-Caura:
@claude please review at head `eec78ed3`.

This changes the derivation of a value that appears in API responses (`client_request_id`), so the parts I'd most like attacked:

1. **Is content the right key, or did I just move the precondition?** The claim is that identical content within one batch never reaches the write path (`seen_hashes` collapses it first), so two `pending` entries can never share a key. If that is wrong anywhere — a path that skips `seen_hashes`, a hash computed over a different tuple than the one I key on — the failure is two items silently resolving to one row, which is worse than what I fixed.

2. **The two assertions I rewrote.** They pinned `endswith(f":{index}")`. I argue that pinned the defect and that the property worth testing is "one distinct token per item, scoped to the attempt". Check I haven't weakened a real guarantee into a vacuous one — in particular that distinctness is still meaningful given identical content deliberately shares a key.

3. **The wire-visible change.** `client_request_id` is returned per item. I believe nothing external parses its shape, and callers correlate through `index`. Please check whether anything in-tree — clients, the dashboard path, the broker path, tests I did not touch — depends on the `:{int}` form.

4. **Rows written before this change** carry `attempt:0`-style ids. A retry of an in-flight batch across the deploy boundary computes a different key, so it falls through to content-hash dedup and reports `duplicate_content` rather than `duplicate_attempt`. I believe that is safe — no double write, different label — but it is the one behaviour change I cannot test, so I would like a second opinion on it.

Deliberately out of scope: H-07 (>100 facts → 500) in the same function. Its fix is batching, which under positional keys would have reintroduced this bug one layer up.

Eldad-Caura:
Checking my own question 1 rather than leaving it open, so you can falsify the reasoning instead of re-deriving it.

**The claim:** no two entries in `pending` can share a `client_request_id`.

**Why it holds.** There is exactly one `pending.append` site. Reaching it requires passing the intra-batch guard:

```python
if ch in seen_hashes:
    ...  # duplicate_content
    continue
seen_hashes[ch] = i
```

`seen_hashes[ch]` is set *before* the ~100 lines that follow, so any later item with the same `ch` takes the `continue` and never reaches the append. Items that fall out between the two points don't weaken it — they just don't reach `pending`.

The stronger half: the key is `hashes[i][:16]` and the guard uses `ch = hashes[i]`. **The same value**, not two hashes that happen to agree — so they cannot diverge under a future edit to one of them without breaking the other visibly.

The residual risk I can't close by reading is a truncation collision: two *different* contents whose hashes agree in the first 16 hex chars. 64 bits over ≤100 items is ~2⁻⁵¹, and the consequence would be one item resolving to the other's row. If you think that's the wrong place to spend a probability budget, say so and I'll widen it — the column is `Text`, so the full 64 chars costs nothing but length.

Eldad-Caura:
Both review passes came back clean — thank you, and the second one added a detail I had not stated: **masking happens before the hash is computed**, so the key is derived from the post-governance content, which is the content actually written. That is the right property and I had not checked it.

**CI failed after those reviews, on something neither of us was looking at.** The bulk route's docstring is not internal: `POST /api/v1/memories/bulk` is one of the eight frozen-v1 broker operations `caura-daemon` calls, and the docstring is published as that operation's OpenAPI **description**. Editing it made `core-api/openapi.broker.json` stale.

Worth recording because of *what* caught it. Locally I had run ruff, mypy, both test roots and all three naming gates, all clean — **none of them look at the generated contract.** The only signal was CI. Regenerated and committed at `bf334bb7`; the diff is one line of description text, no schema change, so the companion `oasdiff` breaking-change gate is unaffected.

Also rebased onto `46eb44ab` — main moved twice under this branch (#1280, #1281). #1281 touches `create_memories_bulk`'s re-embed fallbacks, so I re-ran the full suite rather than assuming the rebase was inert: **6075 passed, 0 failed**, and the baseline check is current on the new base.

Nothing about the fix itself changed since your review — the delta is the regenerated baseline and the rebase.

Eldad-Caura:
@erni-a this is ready for your review — H-08, the silent data loss on a partial bulk retry.

**State at head `bf334bb7`** (rebased onto `46eb44ab`):

- Every check on that SHA is green, verified against the head SHA rather than via `gh pr checks`.
- Two review passes came back clean. Both ran on `eec78ed3`; `claude-review` skips `synchronize` events by design, so it did not re-run on this head. I checked what that leaves uncovered rather than assuming: comparing my commit's own patch at both heads, the only difference is a blob index line from the rebase — **the reviewed code is byte-identical**. The delta since review is one line of regenerated `openapi.broker.json`.
- Full root suite **6075 passed / 0 failed**; storage suite **328 passed** on its own database; ruff, mypy, ratchet, sentinel and the tenant-scope gate all clean after `git add`.

**Worth your attention in particular**, since it is the part with the widest blast radius:

The per-item idempotency key that appears in bulk responses changes shape, from `attempt:0` to `attempt:<content-hash-prefix>`. I could find nothing in-tree that parses it — callers correlate through `index` — and I asked the reviewer to look for anything I missed. The one behaviour change I could not test is rows written **before** this ships: a retry of an in-flight batch across the deploy boundary computes a different key, falls through to content-hash dedup, and reports `duplicate_content` rather than `duplicate_attempt`. No double write, different label. If that matters to a consumer I do not know about, this is the moment to say so.

Not merging — that is Eldad's.

erni-a (approved):

Issue #1205, opened 2026-09-01
core-api: MCP observes plan-limit mode but never refuses on it


Eldad-Caura:
Found while verifying the write-quota parity work (#1196). Not in the original
REST/MCP parity report, which observed only that MCP "checks nothing" on
`transition` — the gap is every mutating tool.

## The gap

**No MCP tool can see plan-limit mode.** `enforce_usage_limits` has **zero code
occurrences** in `mcp_server.py` — the single textual match is inside a comment.

The cause is upstream of the handlers. Plan-limit enforcement travels
out-of-band: the platform computes "over plan" from the persisted counters and
stamps `x-org-read-only` on the request, which `AuthContext.is_read_only` reads
and `enforce_usage_limits()` turns into a 403 (see `usage_service`'s module
docstring, and `auth.py:440`).

`MCPAuthMiddleware` never reads that header. It parks `x-tenant-id`,
`x-agent-id`, `x-readable-tenant-ids`, `x-capabilities` / `x-key-scopes`,
`x-caura-credential-kind` and `x-install-uuid` on context vars — and nothing
else. There is no context var for read-only mode, no getter, and consequently
nothing a handler could consult even if it wanted to.

## Consequence

An org in plan-limit read-only mode is refused writes over REST and **allowed
them over MCP** — `caura_write`, `caura_doc`, `caura_evolve`,
`caura_keystones_set`, `caura_manage` update, all of them. Same tenant, same
operation, different answer per transport.

This is the general case of the drift #1196 fixed for one verb. That PR could
only close it by making REST match MCP (transitions free everywhere), because
the other direction — making MCP enforce — is not reachable from the MCP
surface as it stands.

It also caps what the policy tables added in #1196 can do:
`PLAN_LIMIT_GATED_OPS` is REST-only in practice, and any op listed there is
gated on REST and ungated on MCP.

## Suggested fix

1. Plumb `x-org-read-only` into `MCPAuthMiddleware` alongside the other
   gateway-injected attributes, honoured **only** on the gateway-verified path
   (`x-tenant-id` present and the shared secret matched) exactly as
   `x-agent-id` and `x-capabilities` already are — a direct caller must not be
   able to clear its own read-only flag by omitting the header. Note the
   existing comment there about resetting every context var on every request:
   this one must default to **not** read-only only when the gateway said so, and
   a stale value bleeding between requests would be a billing bypass.
2. Give the handlers a shared refusal helper that returns the canonical MCP
[...]

PR #1285, opened 2026-09-04
fix(storage): refuse a negative tenant usage counter

Eldad-Caura:
## The defect

`tenant_usage_counters.count` is **added** to on upsert — `count = count + excluded.count` — and nothing forbade a negative. Not the column (migration 039 created it as plain `BigInteger NOT NULL DEFAULT 0`), and not `POST /tenant-usage/increment`, which validates `tenant_id`, `operation` and `period_start` and then passes `count` straight through via `r.get("count", 1)`.

A negative there is not a smaller number, it is a **disabled plan limit**. The platform's `_is_over_plan_limits` asks `used > limit`, so a counter driven below zero reads as far under budget and silently stops refusing writes for that tenant. The counter going wrong makes enforcement *quieter* rather than louder — the wrong direction for the one table plan caps are computed from.

## What changed

- **Router (`tenant_usage.py`)** — a present `count` must be a non-negative `int`; otherwise 422 naming the row. This closes the reachable path. Zero is still accepted (a legitimate no-op flush). `bool` is refused because it is an `int` in Python, and a JSON `true` quietly metering 1 operation is a coercion nobody asked for.
- **Migration 042 + model** — `CHECK (count >= 0)`, named `ck_tenant_usage_counters_count_nonneg` in both so `create_all` and the migration agree (verified: the model's compiled DDL emits the identical constraint). This is the floor under the edge check, covering the hand-run correction the router cannot see — which is how the gap was noticed.

`capability_usage` leaves its counts equally unconstrained and is **deliberately left alone**: a bad number there skews an adoption report, not an enforcement decision.

## Migration safety

042 clamps any existing negative to 0 *before* adding the constraint. Migrations here run in the FastAPI lifespan startup hook (see 038's docstring), so a migration that fails on legacy data is a deploy that never becomes ready. Clamping to 0 rather than a reconstructed value is deliberate — the table stores a running total, not the history behind it, so the true value is unrecoverable — and 0 is the conservative direction, raising the tenant back toward their limit rather than leaving them under it.

`tenant_usage_counters` is not a declared large table (it is bounded at one row per tenant/operation/period by 039's unique index), so the plain validated `ADD CONSTRAINT` is used rather than 038's `NOT VALID` / `VALIDATE` two-step. The clamp and the add share one transaction, so a concurrent writer cannot slip a negative in between them.

**Verified against real violating data**: with a negative row present, the clamp-then-constrain sequence completed successfully — the deploy-stall case this guards is exercised, not just described.

## Tests

Four new tests, each confirmed to fail without its fix by mutation:

| Test | Mutation applied | Result |
|---|---|---|
| `test_a_negative_count_is_refused` | guard removed | FAILS (200, not 422) |
| `test_a_boolean_count_is_refused` | guard removed | FAILS (200, not 422) |
| `test_a_zero_count_is_still_accepted` | `c < 0` → `c <= 0` | FAILS (422, not 200) |
| `test_the_table_refuses_a_negative_count_beneath_the_router` | constraint dropped | FAILS (`DID NOT RAISE`) |

The constraint test goes through direct SQL rather than the endpoint, because the endpoint would reject the row before it reached the column the constraint protects.

Full `core-storage-api` suite run on this branch and on clean `origin/main` against fresh databases: the failure sets are **byte-identical** (17 pre-existing failures/errors in `test_migration_040_content_hash_cleanup.py` and `test_integration.py`, unrelated to this change), and this branch passes exactly 4 more — the 4 above.

`ruff check` / `ruff format --check` at CI's scopes, `mypy src/`, both naming gates, and the migration-chain guards all pass. `test_single_head` bumped 041 → 042, which that test's docstring asks for explicitly.

## Context

Found while tracing the plan-limit enforcement chain for caura-ai/caura#1205. It is the third of four items in that cluster; the others are reported separately.

### DIFF
--- a/common/models/tenant_usage_counter.py
+++ b/common/models/tenant_usage_counter.py
@@ -34,7 +34,7 @@ from __future__ import annotations
 
 from datetime import datetime
 
-from sqlalchemy import BigInteger, DateTime, Index, Text, text
+from sqlalchemy import BigInteger, CheckConstraint, DateTime, Index, Text, text
 from sqlalchemy.orm import Mapped, mapped_column
 
 from common.models.base import Base
@@ -57,6 +57,8 @@ class TenantUsageCounter(Base):
         DateTime(timezone=True), nullable=False
     )
 
+    # Non-negative by CHECK, not merely by convention — see the constraint in
+    # ``__table_args__`` for why this column in particular earns one.
     count: Mapped[int] = mapped_column(
         BigInteger, nullable=False, server_default=text("0")
     )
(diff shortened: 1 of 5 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

Issue #1220, opened 2026-09-01
core-api: the MCP batch write charges no write quota — caura_write(items=[...]) meters zero


Eldad-Caura:
Found while plumbing the plan-limit signal into MCP for #1205. Not in the
original REST/MCP parity report, and distinct from #1205: that one is about
**enforcement** (the signal never reaches MCP), this one is about **metering**
(the counter never moves).

## The gap

`caura_write(items=[...])` — the MCP batch write — charges **no write quota at
all**.

`mcp_server.py`, inside `caura_write`:

```python
if content is not None:
    if charges_write_quota("create"):
        await check_and_increment(tenant_id, "write")     # single write: metered
    ...
# ── batch branch ──
bulk_data = BulkMemoryCreate(...)
result = await create_memories_bulk(bulk_data, ...)        # N writes: metered ZERO
```

The only metering call in the whole of `caura_write` is the one on the single
`content` path, guarded by `if content is not None`. The batch branch reaches
`create_memories_bulk` with nothing in between.

REST does charge:

```python
# routes/memories.py — POST /memories/bulk
usage = await bulk_check_and_increment(body.tenant_id, len(body.items))
```

`bulk_check_and_increment` has **zero occurrences in `mcp_server.py`** — it
appears only in `routes/memories.py` (two call sites: bulk create and
redistribute).

## Consequence

Same tenant, same N memories, different bill:

| surface | call | write quota charged |
|---|---|---|
| REST | `POST /memories/bulk` with N items | N |
| MCP | `caura_write(items=[…])` with N items | **0** |

### The second-order effect is the worse one

The counters this misses are the same counters plan enforcement is computed
from. Per `usage_service`'s module docstring, the platform derives "over plan"
from the persisted counters and stamps `x-org-read-only`.

So an MCP-heavy tenant writing in batches does not merely underpay — **it never
trips the plan limit in the first place**, because the usage that would trip it
was never recorded. That also caps what fixing #1205 can achieve on its own:
enforcing a read-only flag that never gets set changes nothing for exactly the
tenants doing the most unmetered writing.

## Suggested fix

Call `bulk_check_and_increment(tenant_id, len(bulk_items))` on the batch path,
mirroring `POST /memories/bulk`.

Note the ordering question REST already answers: it charges **before** the
write, so a partially-failed bulk still costs the caller what it attempted.
Match that rather than inventing a second convention.

## Blast radius — read before doing this

This is a live billing change. Tenants writing batches over MCP start consuming
[...]

PR #1284, opened 2026-09-04
fix(clients): retire the MemClaw/MemClawError/MemClawAPIError class aliases

Eldad-Caura:
## Summary

Per Eldad's direction: these class aliases were the one remaining piece of the Python client's legacy surface still carrying a "permanent by choice" designation after [#1244](https://github.com/caura-ai/caura/pull/1244) retired the `memclaw-client` PyPI package and the `memclaw_client` import module. Same treatment now — removed, not deprecated. No transition is owed to pre-rename installs.

**Removed:**
- `clients/python/src/caura_client/client.py`: `MemClaw = Caura`
- `clients/python/src/caura_client/exceptions.py`: `MemClawError`, `MemClawAPIError`
- `clients/python/src/caura_client/__init__.py`: the three re-exports
- `clients/python/tests/test_legacy_alias.py`: both tests existed only to pin the aliases this change removes

**Updated in the same commit:**
- 4 sentinel pins in `scripts/do_not_touch_sentinel.py` that asserted these exact lines exist — left the separate MCP/plugin `MemClaw` sentinel entries untouched (a different, still-permanent surface, e.g. the plugin id and MCP tool-name aliases)
- The policy row in `docs/plans/rebrand-alias-retirement-policy.md`, now marked retired to match the npm-client row's own pattern
- Three docs/description strings (root `README.md`, `clients/python/README.md`, `clients/python/pyproject.toml`) that specifically claimed the class alias was permanent

**Deliberately out of scope:** `clients/typescript/src/index.ts` carries the identical `MemClaw`/`MemClawError`/`MemClawApiError`/`MemClawOptions` pattern for the TypeScript client. Not touched here — separate surface, separate decision.

This lands on top of #1280 (the version bump), so it's part of whatever ships as the next `caura-client` release — no additional version bump needed since 1.0.2 hasn't been tagged/published yet.

## Test plan

- [x] `python3 scripts/legacy_name_ratchet.py --base origin/main` → `No new lines. 0 annotated, 2 removed, 0 excused moves (-2 net).`
- [x] `python3 scripts/do_not_touch_sentinel.py` → `All 35 protected strings survive.`
- [x] `pytest tests/` (isolated venv) → 137 passed (139 minus the 2 removed with `test_legacy_alias.py`)
- [x] `ruff check`/`ruff format --check` clean on every file this PR touches (confirmed a pre-existing, unrelated formatting drift on 11 *other* files also exists identically on unmodified `origin/main` — not introduced here)
[...]

PR #1288, opened 2026-09-04
docs(core-api): correct the stale claim that MCP cannot see plan-limit mode

Eldad-Caura:
Comment-only. No behaviour change, and **deliberately no enforcement** — see below.

## What was wrong

`usage_service.PLAN_LIMIT_GATED_OPS` carried this:

> REST-ONLY IN PRACTICE: the MCP surface cannot consult this because it has no read-only signal at all — the MCP middleware never reads the gateway's `x-org-read-only` header, so no MCP tool can see plan-limit mode (caura-ai/caura#1205).

That was true when written and is now false. On `main`:

- `MCPAuthMiddleware` sets `_org_read_only_var` from `x-org-read-only`, gated on `via_gateway` (`mcp_server.py`, in the block right after `_install_uuid_var`).
- `_is_org_read_only()` exposes it.
- `_observe_plan_limit()` consults `PLAN_LIMIT_GATED_OPS` and logs `mcp_plan_limit_would_refuse` at WARNING — called for `create` and `bulk_create`.

So step one of #1205 is done. The issue describes a missing **signal**; the work actually left is a missing **refusal**.

This matters beyond tidiness: the stale comment is what #1205 was written from. Left as-is it will keep generating work items for a gap that has already been half-closed, and it invites someone to "fix" it by plumbing a header that is already plumbed.

## The precondition that was marked unobtainable

`_observe_plan_limit`'s docstring said the log was the only way to learn

> whether the gateway stamps `x-org-read-only` on the `/mcp` route or only on `/api/v1` […] a question that cannot be answered from this repo at all

Correct that it cannot be answered from *this* repo — the header has no producer in OSS. It can be answered from `caura-enterprise`, and I read it there:

- `gateway/nginx.conf.template` gives both `location = /mcp` and `location /mcp/` the same `auth_request_set $org_read_only $upstream_http_x_org_read_only` + `proxy_set_header X-Org-Read-Only $org_read_only` pair as the `/api/v1/` catch-all, alongside `X-Gateway-Secret` — so `via_gateway` is true and the header is honoured rather than discarded.
- The value originates in `platform-auth-api`'s `/_auth` (`routers/auth.py`), which sets `X-Org-Read-Only: "true"` on an expired license, a cached read-only verdict, or a live storage lookup.

**The answer is yes.** That removes one of the three causes the docstring lists for a quiet log, which is the whole point of recording it: a future reader now knows silence is *not* explained by the gateway skipping the route, and does not re-run an investigation that has an answer.

It adds a fourth cause in exchange, which is worth knowing before anyone reads a quiet log as a green light: `/_auth` **fails open on its own storage lookup** — an exception there returns without the header, so a storage outage reads as "not read-only" for the cache TTL.

## What this PR does not do

It does not add enforcement. `_observe_plan_limit`'s docstring stages that deliberately — enforcing takes capability away from tenants who have it today, and the blast radius should be read off the logs rather than off the support queue. The remaining precondition (whether tenants are ever marked over plan given the MCP batch path meters nothing while `meters_mcp_bulk_write()` is off, #1220) is still open. Closing #1205 is a sequencing decision with enterprise-side awareness, not a quiet fix, and this PR does not pre-empt it.

### DIFF
--- a/core-api/src/core_api/mcp_server.py
+++ b/core-api/src/core_api/mcp_server.py
@@ -488,19 +488,32 @@ def _observe_plan_limit(op: MutatingOp, tenant_id: str) -> None:
     from the support queue.
 
     So: emit what WOULD be refused, ship it, read the logs, then enforce with
-    the number in hand. The log also answers a question that cannot be answered
-    from this repo at all — whether the gateway stamps ``x-org-read-only`` on
-    the ``/mcp`` route or only on ``/api/v1``. The header has no producer in
-    OSS; if this line never fires in production, that is the first thing to
-    check.
+    the number in hand.
 
-    READ SILENCE CAREFULLY. Three different things produce no log, and only one
-    of them is "nothing would be refused": the gateway may not stamp the header
-    on this route at all (above), or the tenant may never have been marked over
-    plan because the MCP batch path records no usage to compute that from —
-    which stays true while ``meters_mcp_bulk_write()`` is off, its default
-    (caura-ai/caura#1220). Rule those out before reading a quiet log as a
-    green light to enforce.
+    THE GATEWAY QUESTION IS ANSWERED — IT IS YES. This docstring used to say the
+    log was the only way to learn "whether the gateway stamps
+    ``x-org-read-only`` on the ``/mcp`` route or only on ``/api/v1``", because
+    the header has no producer in OSS. It has one in caura-enterprise, and it
+    was read there at ``origin/dev`` ``13564142``:
+    ``gateway/nginx.conf.template`` gives both
+    ``location = /mcp`` and ``location /mcp/`` the same
+    ``auth_request_set $org_read_only`` / ``proxy_set_header X-Org-Read-Only``
+    pair the ``/api/v1/`` catch-all has, alongside ``X-Gateway-Secret`` — so
+    ``via_gateway`` is true and the header is honoured. The value originates in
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
Addressed in `d4ce9e80`.

The finding is right, and the cause is a timezone artefact rather than a typo: my session's local date was 2026-09-05 while the commit's UTC date was 2026-09-04, so the docstring read as dated one day into the future.

Rather than move the date back a day, I replaced it with the thing that was actually load-bearing:

```
was read there at ``origin/dev`` ``13564142``
```

A date tells a reader roughly how stale a cross-repo claim might be. A SHA lets them check it exactly — `caura-enterprise` `origin/dev` at `13564142`, `gateway/nginx.conf.template`, the `location = /mcp` and `location /mcp/` blocks. That is strictly better for a claim this docstring asks readers to rely on when deciding whether to move to enforcement, which is the reviewer's stated reason for caring.

Note this branch was force-rebased onto `main` by another actor while I was amending, which dropped the fix. `d4ce9e80` carries both: the amendment **and** the newer base (it contains #1284). Verified with `git merge-base --is-ancestor` rather than assumed.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

Eldad-Caura:
@claude

Re-review requested: the last push was a rebase onto current `main`, which the workflow skips silently on `synchronize`, so no review has run against head `830a2098`. The date finding from the previous round is addressed there.

PR #1289, opened 2026-09-04
fix(storage): record provenance in the embedding backfill

Eldad-Caura:
## The defect

`backfill_embeddings.py` wrote the vector and nothing else:

```sql
UPDATE memories SET embedding = (:emb)::vector WHERE id = :id
```

That does not leave a row where it found it — it **moves** it. Both embedding
backfills scan `embedding IS NULL`. A row written without
`embedded_content_hash` lands in `embedding IS NOT NULL AND
embedded_content_hash IS NULL`, and **nothing anywhere scans that**. So the
repair tool's own write removed rows from every repair path and from the
staleness detector, in one statement, with no error.

Same class as #1281, which put 241 production rows in that bucket. Per the
altitude review, this was the **last** writer to `memories` still missing
provenance — `memory_add`, `memory_add_all`, `memory_update` and
`memory_update_embedding` all record it already.

## The fix

`embedded_content_hash` is written in the same UPDATE as the vector, from the
hash **read alongside the content**.

**That bind param is the whole correctness argument, not a style choice.** A
content update landing in the `fetch → embed → write` window leaves the row's
`content_hash` describing text this vector was *not* built from. A
self-referential `SET embedded_content_hash = content_hash` would stamp that
new hash onto the old vector — recording the row as freshly embedded at the
precise moment it went stale. `memory_update_embedding` refuses the same
re-read for exactly this reason. Naming the value we embedded keeps the
column's promise true and leaves the row correctly visible as stale.

Carried on a `_Provenance` NamedTuple rather than two optional `_TableSpec`
fields, so "one column set and the other not" cannot be written down — half a
pair silently disables stamping, which is the failure being fixed.
Unrepresentable beats validated. `entities` sets it to `None`: migration 037
added the column to `memories` alone, so a blanket assignment would fail there
with an undefined-column error on **every** entity write.

A NULL `content_hash` writes NULL deliberately. Unknown provenance is honest;
a guessed hash reads downstream as verified freshness and is worse than the
NULL it replaced.

## Two pre-existing defects, found by testing the mode

`--rewrite-hint-prefixed` **had never completed a run**:

1. The spec named `metadata_` — the *ORM attribute*. The database column is
   `metadata` (`mapped_column("metadata", JSONB)` renames it away from
   `Base.metadata`). Raw SQL needs the database's name, so the scan raised
   `UndefinedColumnError` before touching a row.
2. That column is `sa.JSON()` per migration 001, not JSONB, so
   `? 'retrieval_hint'` raised `operator does not exist: json ? unknown`.
   Dropped rather than cast — the following `COALESCE(...) <> ''` already
   excludes an absent key, an empty value and a NULL metadata.

I fixed them here rather than splitting: the new stamping is untestable in a
mode that raises on its first statement, so splitting would ship this PR with
its hint-rewrite path unexercised — the same gap shape the PR exists to close.

## A test that was holding the bug in place

`tests/test_backfill_embeddings.py` asserted `"metadata_ ? 'retrieval_hint'"
in sql`. **That assertion pinned the broken SQL string**, which is how a mode
that raises on every run stayed green for its whole life. The mock never
parses SQL, so it cannot tell a valid column name from an invalid one in
either direction.

My first pass substituted corrected strings. That repeats the mistake with
better spelling, so those assertions now check only **mode dispatch** — what a
mock can honestly test — and SQL correctness moved to the new integration file
that executes against real PostgreSQL and therefore fails on SQL that cannot
run.

## Verification

Reverted two ways, because the second is the one that matters:

| Revert | Result |
|---|---|
| Stamping removed entirely | **4 of 6 integration tests red**, plus the same-statement guard |
| Naive `SET embedded_content_hash = content_hash` | **1 red** — `test_stamps_the_hash_it_embedded_not_the_row_s_current_hash` |

The naive form passes every other test in the file. That one test is the
difference between this fix and the obvious one.

core-storage-api suite **328 → 334**. Root suite: **18 known environmental
failures before and after, identical by name, zero new**, +25 passing. ruff
clean at CI's scopes, `mypy src/` clean across 85 files.

## Deliberately not done

- **No sweep for the bucket.** Nothing selects `embedding IS NOT NULL AND
  embedded_content_hash IS NULL`, and that absence — not this diff — is what
  makes each such bug permanent. A sweep needs its own index (037's partial
  index explicitly excludes the bucket, so one written today would seq-scan),
  an operator entry point, and a decision about the ~95k legitimate pre-037
  rows and their re-embed cost. Its own change.
- **No CHECK constraint.** Non-viable three ways: the ~95k pre-037 rows would
  force `NOT VALID` (unenforced for exactly the motivating population); NULL is
  the *correct* value when a row has no `content_hash`, so a constraint would
  coerce callers into inventing one; and the deferred write path legitimately
  inserts `embedding=NULL, embedded_content_hash=NULL` and fills the vector
  later.
- **`tests/test_embed_provenance_call_sites.py` unchanged.** Its
  `_EXEMPT_PARTS` includes `scripts`, but that guard checks
  `_schedule_embed_or_reembed` call sites and this file never calls it — it
  writes SQL directly. Un-exempting `scripts` would not have caught this and
  would risk false failures on diagnostic tooling. The targeted guard is
  `test_production_specs_declare_provenance_where_the_columns_exist`.
- **Still one `get_embedding` per row.** `get_embeddings_batch` exists and is
  where this script's runtime actually lives, but that is a performance change,
  not this one.

Not merging.

### DIFF
--- a/core-storage-api/src/core_storage_api/scripts/backfill_embeddings.py
+++ b/core-storage-api/src/core_storage_api/scripts/backfill_embeddings.py
@@ -66,8 +66,16 @@ Usage:
         --rewrite-hint-prefixed --tenant-id tenant-abc --dry-run
 
 Scope:
-- ``memories.embedding`` — re-embedded from ``memories.content``.
+- ``memories.embedding`` — re-embedded from ``memories.content``, and
+  ``memories.embedded_content_hash`` stamped in the same statement with the
+  ``content_hash`` read alongside that content. Without the stamp this sweep
+  moved rows out of ``embedding IS NULL`` — the population both backfills
+  scan — and into ``embedding IS NOT NULL AND embedded_content_hash IS
+  NULL``, which nothing scans, so the row silently left every repair path
+  and the staleness detector at once.
 - ``entities.name_embedding`` — re-embedded from ``entities.canonical_name``.
+  No provenance stamp: migration 037 added ``embedded_content_hash`` to
+  ``memories`` only, and ``entities`` carries no content hash to record.
 - ``documents.embedding`` — NOT handled. Documents store opaque JSON and
   the embed source is fixed to ``data["summary"]`` (with a back-compat
   fallback to ``data["description"]`` for ``collection="skills"``).
@@ -93,6 +101,7 @@ import sys
 import time
 import uuid
 from collections.abc import AsyncIterator
+from typing import NamedTuple
 
 logger = logging.getLogger(__name__)
 
@@ -103,6 +112,19 @@ logger = logging.getLogger(__name__)
 # rather than blipping. Better to halt and let the operator escalate
 # than to spend the next hour writing nothing.
 _MAX_CONSECUTIVE_NONES = 20
+
+
+class _Provenance(NamedTuple):
+    """Which columns carry a table's embedding provenance.
+
+    A pair rather than two optional fields on ``_TableSpec``, so "one set and
+    the other not" cannot be written down. Half a pair silently disables
+    stamping, which is the exact failure this records against — a sweep that
+    writes vectors and attests nothing. Unrepresentable beats validated.
+    """
+
+    source_hash_column: str  # read alongside the content
+    stamp_column: str  # the value read is written back here
 
 
 @dataclasses.dataclass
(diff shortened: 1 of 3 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #1292, opened 2026-09-04
fix(governance): cascade a drop or keep_private to rows derived before the verdict

Eldad-Caura:
Closes audit finding **H-10**.

## The defect

On a **deferred deployment — the production SaaS write path** — the auto-chunk branch's `GovernanceDecision` runs with `enrichment=None`, takes its documented uncertain branch and **enforces nothing**. The children are built and committed immediately, at `scope_team`, with no governance metadata of their own.

The parent's real verdict arrives minutes later (`ENRICH_REQUESTED` → worker PATCH → `ENRICHED` → `remediate_after_enrichment`) and soft-deleted or downgraded **only the row the event named**. `parent_memory_id` was written onto every child and **queried nowhere** — five sites in production code, all writes.

So a tenant configured `non_business.disposition=drop` had its parent dropped and audited while **N children carrying the same content stayed live and team-visible**. Permanently: children are never enriched, so nothing revisits them, and they carry clean metadata so any future sweep reads them as fine. No audit row tied them to the drop either — the compliance log recorded one deletion where one row was removed and N were not.

## Reproduced before fixing

```
AssertionError: the dropped content survives in the children: {'m1'}
```

## Why this is the #808 shape, and why ordering can't fix it here

The codebase documents #808 as fixed, and it **is** fixed for the atomic-fact fan-out — by ordering. That path lives inside `_enrich_memory_background`, runs remediation **before** it fans out, early-returns on a drop, and carries a downgrade onto the children via `effective_visibility`. Its own comment states the rule:

> a policy that could not be applied must not be followed by rows it might have forbidden

Auto-chunk in deferred mode is the path where **the derived rows already exist when the verdict lands**. Ordering cannot save it; a cascade is the missing half.

## The fix

`remediate_after_enrichment` now resolves the derived rows and applies the same action: soft-delete on either drop disposition, visibility downgrade on `keep_private`.

**Both destructive branches, not just the one the finding described.** PII-drop and non-business-drop are separate branches reading separate configs; a tenant on a PII drop policy leaked identically.

**The lookup runs before the parent's audit and delete**, so a lookup failure leaves everything intact and remediable rather than a dropped parent whose children were never found. **Failures propagate** rather than being swallowed — this module's contract is that a policy which could not be applied must not be quietly treated as applied, and the enclosing `tracked_task` surfaces it.

**Each cascaded row gets its own audit row**, carrying `cascaded_from` so a review can see why a row with no governance signals of its own was removed. Per child rather than one rolled-up entry: each is a separate soft-delete, and a log recording one deletion while N happened misstates the record in the direction that matters.

### The new storage query

The parent→child link lives in child metadata JSON and nothing could read it. Tenant-scoped, live rows only, **no status or visibility filter** — remediation must reach every derived row whatever state it is in, the same reasoning `memory_find_by_supersedes_id` records for retraction.

### Why it is gated on `auto_chunked`

The query filters a JSON key with no supporting index, and a tenant configured `drop` remediates constantly — an ungated version would tax every ordinary drop to serve the rare chunked one.

`auto_chunked` is safe to gate on for a reason worth stating: it is stamped **unconditionally in the same function that builds the children**, and it is **already on the production rows this has to reach**. A new marker would only appear on rows written after the deploy and would leave the existing leak in place.

## Tests

**Five in core-api**, four confirmed failing without the fix. The fifth is an over-refusal guard: an ordinary row must not run the lookup at all.

**Four in core-storage-api against real Postgres**, because the core-api side stubs storage entirely — without them the cascade could ship with a query that matches nothing and every test above it would still pass. The tenant-boundary one is probe-confirmed: removing the tenant predicate fails it with `2 == 1`, which on a drop would mean deleting **another tenant's** rows.

**One test lives beside the write rather than the remediation.** The gate depends on `auto_chunked` being stamped on the parent and `parent_memory_id` on each child; if either stops being written the cascade silently stops running with nothing failing near it. Asserting that in the remediation tests would be a stub asserting itself.

## Verification

- Full root suite: **6076 passed, 5 skipped, 1 xfailed, 0 failed**.
- `core-storage-api/tests/`: **336 passed**, on its own scratch database.
- `mypy` clean on `core-storage-api/src/`; `core-api/src/` clean apart from 2 pre-existing `types-python-dateutil` stub errors in an untouched file.
- Broker OpenAPI baseline — current. (The new endpoint is on core-storage-api, which the broker contract does not cover.)
- `legacy_name_ratchet.py` → *No new lines.* · `do_not_touch_sentinel.py` → *All 35 protected strings survive* (the 39→35 change came from #1284 on main, confirmed against a clean tree, not from this branch) · `tenant_scope_gate.py` → exit 0, no allowlist movement, and the new route is counted as tenant-bound.
- Checked for an open PR on this subsystem before starting; none.
- Branched from `origin/main`, rebased onto `4c9cef2a`.

### DIFF
--- a/core-api/src/core_api/clients/storage_client.py
+++ b/core-api/src/core_api/clients/storage_client.py
@@ -491,16 +491,28 @@ class CoreStorageClient:
         resp.raise_for_status()
         return resp.json()
 
-    async def _get_list(self, path: str, **params: Any) -> list[dict]:
-        # All current _get_list callers are pure list/stats endpoints; none
-        # sit on the write path, so no per-call opt-out is needed yet.
-        headers = await self._auth_headers(read=True)
+    async def _get_list(self, path: str, *, read: bool = True, **params: Any) -> list[dict]:
+        """List GET. ``read=False`` routes to the WRITER, same as :meth:`_get`.
+
+        This used to say "all current callers are pure list/stats endpoints;
+        none sit on the write path, so no per-call opt-out is needed yet" — true
+        when written, and falsified by the governance cascade
+        (``find_children_by_parent_id``), which reads rows it is about to delete
+        and reasons about what a retry will find. A replica read cannot support
+        that reasoning: under lag a retry sees an already-deleted child as live
+        and re-audits it. So the opt-out is here now, and the claim that it was
+        unnecessary is gone rather than left standing next to a caller that
+        needs it.
+        """
+        prefix = self._read_prefix if read else self._prefix
+        headers = await self._auth_headers(read=read)
 
         def _do() -> Awaitable[httpx.Response]:
-            return self._read_http.get(f"{self._read_prefix}{path}", params=params, headers=headers)
+            http = self._read_http if read else self._http
+            return http.get(f"{prefix}{path}", params=params, headers=headers)
 
         resp = await self._execute(_do, retry=_read_retry, label=f"GET-list {path}")
-        self._maybe_evict_on_auth_error(resp, read=True)
+        self._maybe_evict_on_auth_error(resp, read=read)
         resp.raise_for_status()
         return resp.json()
 
@@ -913,6 +925,27 @@ class CoreStorageClient:
         """
         return await self._get_list(
             "/memories/by-supersedes-id", tenant_id=tenant_id, supersedes_id=supersedes_id
+        )
+
+    async def find_children_by_parent_id(self, tenant_id: str, parent_id: str) -> list[dict]:
+        """H-10 — live rows whose ``metadata.parent_memory_id`` is ``parent_id``.
+
+        Governance-remediation-shaped: no status or visibility filter, because a
+        drop must reach every derived row whatever state it is in. Soft-deleted
+        rows are excluded storage-side — one already gone needs no second
+        remediation.
+
+        ``read=False`` — the WRITER, not the replica. This is a read-your-write:
+        the caller is about to soft-delete what comes back, and on a retry after
+        a partial failure it depends on the already-deleted rows being absent.
+        A replica cannot honour that under lag; it would hand back a child whose
+        delete has committed on the primary, and the cascade would re-audit and
+        re-delete it — putting a duplicate destructive entry in a compliance log
+        for a row that was already handled. Same reasoning ``get_memory`` records
+        for its own ``read=False``.
+        """
+        return await self._get_list(
+            "/memories/by-parent-id", read=False, tenant_id=tenant_id, parent_id=parent_id
         )
 
     async def find_successors(self, data: dict) -> list[dict]:
--- a/core-api/src/core_api/consumer.py
+++ b/core-api/src/core_api/consumer.py
@@ -38,7 +38,10 @@ from common.events.org_settings_changed_event import OrgSettingsChangedEvent
 from common.events.topics import Topics
 from core_api.clients.storage_client import get_storage_client
 from core_api.services.contradiction import Trigger, run_contradiction_detection
-from core_api.services.governance_remediation import remediate_after_enrichment
+from core_api.services.governance_remediation import (
+    GovernanceCascadeError,
+    remediate_after_enrichment,
+)
 from core_api.services.organization_settings import invalidate_cache, resolve_config
 
 logger = logging.getLogger(__name__)
@@ -104,7 +107,32 @@ async def handle_memory_enriched(event: Event) -> None:
     # counterpart to the synchronous GovernanceDecision step. ``resolve_config``
     # tolerates a None session (cache-first; cold-miss opens its own).
     gov_cfg = await resolve_config(payload.tenant_id)
-    if (await remediate_after_enrichment(memory, gov_cfg)).dropped:
+    try:
+        outcome = await remediate_after_enrichment(memory, gov_cfg)
+    except GovernanceCascadeError as exc:
+        # The parent WAS remediated; only rows derived from it could not be.
+        # Caught HERE and nowhere deeper, because the two callers of
+        # ``remediate_after_enrichment`` need opposite things from this failure:
+        #
+        #   - ``_enrich_memory_background`` is about to create MORE derived rows,
+        #     so it must abort. It does not catch this, and must not.
+        #   - this handler creates nothing. Letting the raise through would nack
+        #     the event, and the dispatcher redelivers on nack — re-running the
+        #     whole drop branch and emitting a SECOND ``critical=True`` audit for
+        #     a memory already dropped. Every redelivery. Duplicate destructive
+        #     entries in a tamper-evident log, for a row nothing further can be
+        #     done to.
+        #
+        # The outcome rides on the exception precisely so this branch can honour
+        # the parent's verdict without re-running it.
+        outcome = exc.outcome
+        logger.error(
+            "memory-enriched: governance remediated the row but could not clean up "
+            "rows derived from it; they still carry content the policy forbade",
+            exc_info=True,
+            extra={"memory_id": str(payload.memory_id), "tenant_id": payload.tenant_id},
+        )
+    if outcome.dropped:
         # Row was dropped by policy — skip contradiction detection on it.
         logger.info(
             "memory-enriched: row dropped by governance policy; ack",
--- a/core-api/src/core_api/services/governance_remediation.py
+++ b/core-api/src/core_api/services/governance_remediation.py
@@ -51,12 +51,304 @@ class RemediationOutcome:
     was left alone."""
 
 
+async def _pre_verdict_children(sc: Any, memory_id: str, tenant_id: str, md: dict) -> list[dict]:
+    """Rows derived from this one BEFORE any verdict existed. ``[]`` when none can.
+
+    H-10. The atomic-fact fan-out needs nothing here — it lives inside
+    ``_enrich_memory_background`` and runs remediation *before* it fans out, so
+    a drop stops it and a downgrade is carried onto the children it does create.
+    Auto-chunk is the path with no such ordering: on a deferred deployment the
+    write-time ``GovernanceDecision`` sees ``enrichment=None``, takes its
+    documented uncertain branch and enforces nothing, and the children are
+    committed immediately. The verdict arrives here minutes later, naming only
+    the parent.
+
+    Gated on ``auto_chunked`` rather than querying unconditionally. The lookup
+    filters on a JSON key with no supporting index, and a tenant configured
+    ``drop`` remediates constantly — so an ungated query would tax every
+    ordinary drop to serve the rare chunked one. The marker is safe to gate on
+    because it is stamped on the parent unconditionally, in the same function
+    that builds the children, and it is already present on the production rows
+    this fix has to reach — a NEW marker would only ever appear on rows written
+    after the deploy, leaving the existing leak in place.
+
+    Failures propagate. This module's contract is that a policy which could not
+    be applied must not be quietly treated as applied; swallowing a lookup error
+    would leave the children live, which is the leak this exists to close.
+    """
+    if not md.get("auto_chunked"):
+        return []
+    return await sc.find_children_by_parent_id(tenant_id, memory_id)
+
+
+def _child_id(child: dict) -> str | None:
+    raw = child.get("id")
+    return str(raw) if raw else None
+
+
+class GovernanceCascadeError(RuntimeError):
+    """Some derived rows could not be remediated after the parent already was.
+
+    Raised only once every child has been ATTEMPTED. The distinction matters:
+    aborting the loop on the first failure would leave the remaining children
+    live with the parent already gone — the leak this cascade exists to close,
+    reintroduced by the error handling rather than by the policy.
+
+    Still raised rather than logged, because the enclosing ``tracked_task`` is
+    what turns an unapplied policy into something a human sees; a swallowed
+    failure here is a policy silently treated as applied.
+
+    ``outcome`` carries what the PARENT's own remediation did, because that
+    part succeeded and a caller may need it. Without it this exception conflates
+    two very different states: "the parent's remediation failed, nothing
+    happened, retry from scratch" and "the parent was dropped or privatised and
+    only the derived cleanup fell short". A caller that wants to honour the
+    parent's verdict can read it; one that simply refuses to go on — which is
+    the safe default, and what every current caller does — can ignore it.
+    """
+
+    def __init__(self, message: str, outcome: RemediationOutcome | None = None) -> None:
+        super().__init__(message)
+        self.outcome = outcome or RemediationOutcome()
+
+
+def _raise_if_any_failed(
+    failed: list[str], *, parent_id: str, what: str, outcome: RemediationOutcome
+) -> None:
+    if not failed:
+        return
+    raise GovernanceCascadeError(
+        f"governance: {len(failed)} derived row(s) of {parent_id} could not be "
+        f"{what} ({', '.join(failed)}); the parent was already remediated, so these "
+        "rows still carry content the policy forbade",
+        outcome,
+    )
+
+
+async def _drop_children(
+    sc: Any,
+    children: list[dict],
+    *,
+    tenant_id: str,
+    parent_id: str,
+    action: str,
+    detail_for: Any,
+) -> None:
+    """Soft-delete each derived row, auditing before the delete as the parent does.
+
+    Per child rather than one rolled-up row: each is a separate soft-delete, and
+    an audit log that recorded one deletion while N happened would misstate the
+    compliance record in the direction that matters.
+
+    Contained PER CHILD, then raised once at the end. The parent is ALREADY
+    deleted by the time this runs, so letting the first failure abort the loop
+    would leave every later child live with the parent gone — this feature's own
+    leak, reintroduced by its error handling. One bad row must cost one row.
+
+    NOTHING RETRIES A FAILED CHILD, and this docstring used to say the opposite.
+    The retry it described was real while the raise reached the Pub/Sub
+    dispatcher and redelivered the event. It stopped being real when
+    ``handle_memory_enriched`` began catching ``GovernanceCascadeError`` and
+    acking — which it does because redelivery re-ran the parent's entire drop
+    branch and wrote another ``critical=True`` audit for an already-dropped
+    memory, on every delivery, without bound. The mechanism went; the paragraph
+    describing it did not, until this change. The parent's own remediation has
+    already succeeded by now, so no later event revisits this row either.
+
+    Recovery is therefore a person, working from the ERROR logs below, and they
+    are written to make that possible rather than merely to record that
+    something failed. The audit and the delete are attempted in SEPARATE
+    ``try`` blocks so a reader can tell the two states apart:
+
+    * ``audit_emitted=False`` — nothing happened to the row. It is live and the
+      compliance log does not claim otherwise.
+    * ``audit_emitted=True`` — the one that needs hands. The log RECORDS the row
+      as removed and the row is still live.
+
+    The audit still precedes the delete, and that ordering is what makes the
+    second state possible — deliberately, because the alternative is worse. A
+    false "dropped" entry is discoverable and the content is still there to
+    remove; auditing afterwards would turn the same failure into a deletion
+    with no record of it at all.
+    """
+    remediated = 0
+    failed: list[str] = []
+    for child in children:
+        cid = _child_id(child)
+        if not cid:
+            # Cannot delete it and cannot audit it against an id. Counted as a
+            # failure, not just logged: the row still holds content a policy
+            # forbade, and "we could not identify it" is not "it is handled".
+            # A retry will not fix this one — a person has to look.
+            logger.error(
+                "governance: derived row of %s has no usable id; it still holds "
+                "dropped content and was NOT removed",
+                parent_id,
+            )
+            failed.append("<no id>")
+            continue
+        try:
+            detail = detail_for(child.get("content") or "")
+            # Distinguishes a cascaded removal from a direct one, so a compliance
+            # review can see WHY a row with no governance signals of its own was
+            # dropped — the children are never enriched, so they carry none.
+            detail["cascaded_from"] = parent_id
+            await emit_governance_audit(
+                tenant_id=tenant_id,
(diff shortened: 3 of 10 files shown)


Eldad-Caura:
@claude please review at head `48db742f`. This is a security fix that deletes rows, so I would rather have it attacked than approved.

Five things I want challenged:

1. **The gate is the weakest point and I know it.** The cascade only runs when the parent carries `auto_chunked`. I argue that is safe because the marker is stamped unconditionally in the same function that builds the children, and because it is already on production rows (a new marker would leave the existing leak in place). But it is still a correctness dependency on a metadata key, which is this backlog's recurring defect class. Tell me if you would drop the gate and pay the query cost, or if there is a third option I have missed.

2. **Does the cascade reach every affected row, or only the ones I thought of?** I fixed both drop branches and the keep_private branch. Check whether any other disposition or path leaves derived rows behind — the PII mask/flag branch deliberately does nothing, and I believe that is right because it neither deletes nor restricts, but check that reasoning.

3. **Deleting rows on a JSON-key match.** `metadata->>'parent_memory_id'` is not a foreign key and nothing constrains what writes it. If a caller could put an arbitrary parent id into their own row's metadata, could they induce a cascade to delete rows they do not own? I believe the tenant predicate contains this and that within a tenant the writer already controls those rows, but this is the question I am least sure of and it is the one with the worst answer if I am wrong.

4. **Failures propagate on purpose.** A lookup or child-delete failure raises out of remediation rather than being swallowed. That matches this module's existing posture, but it means a transient storage error now fails a remediation that previously would have completed for the parent. Check I have not made a partial-failure state worse: the parent may already be deleted when a child delete fails.

5. **The over-refusal guard.** One test asserts the lookup does NOT fire for an ordinary row. Confirm that is testing something real and not just the gate asserting itself.

Not in scope, deliberately: H-02, the sibling finding where entity rows mined from dropped content survive the same way. It needs its own cascade in a different table and is a separate PR.

Eldad-Caura:
**Both correct, both fixed at `85068763`** — and the Medium is the answer to the question I asked in point 4, which I had reasoned only half-way through.

## The Medium

I argued failures should propagate because "a policy that could not be applied must not be treated as applied". That is right for the **lookup**, which runs before anything is mutated. It is wrong for the **loop**, which runs after the parent is already deleted — and I applied the same reasoning to both without noticing the asymmetry.

Reproduced before fixing: a failure on `c1` escapes raw, and `c2` and `c3` are never touched.

```
RuntimeError: storage blipped on the first child
```

Parent gone, two children still live with the dropped content. **The error handling reintroduced the leak the cascade exists to close** — which is the same shape as the last round of #1278, where a loop-level catch would have cancelled every other child's repair.

**Fixed per child, raised once at the end**, after every child has been *attempted*:

- One bad row costs one row.
- `GovernanceCascadeError` still surfaces to `tracked_task`, so the unapplied policy is not swallowed.
- A retry is narrow rather than wholesale: the lookup excludes soft-deleted rows, so the children that succeeded are not revisited and only the failures are re-attempted.

On your duplicate-audit point — a retry does re-emit the parent's audit. I am taking that rather than fixing it here: it is the cost of not silently leaving forbidden content live, and it is pre-existing behaviour for *any* remediation that raises, not something this cascade introduces. Say if you'd rather I made the parent audit idempotent as part of this, but it felt like a separate change.

**One thing I went further on than you suggested:** a child with no usable id now counts as a failure rather than only being logged. "We could not identify it" is not "it is handled" — the row still holds forbidden content. A retry won't fix that one, so it should stop being quiet and get a person's attention.

## The Low

Right, and it is the same defect class the rest of this backlog keeps turning up: a line asserting something the code does not deliver. `len(children)` counted rows the loop had skipped, so the summary claimed N remediated while N-1 were — in a log a compliance review reads. Now counts actual remediations.

Probe: reverting it fails the new test with `governance: cascaded nonbusiness_drop to 2 derived row(s) of m1` when one was skipped.

## Tests

Two more, both probe-confirmed in both directions:

1. **One failing child does not abandon the rest** — asserts `c2` and `c3` are still deleted and the error still raises. Fails with the containment removed.
2. **The summary count excludes skipped rows.** Fails with `len(children)` restored.

Suite **6078 passed / 0 failed**; ruff, mypy, ratchet, sentinel, tenant-scope gate and the broker baseline all clean after `git add`. Checks were green on `48db742f`.

## Still open from my review request

Points 1, 2, 3 and 5 went unanswered and I am not reading silence as agreement — particularly **3**, which is the one I flagged as having the worst answer if I am wrong: `metadata->>'parent_memory_id'` is not a foreign key, so can a writer induce the cascade to delete rows they should not reach? I believe the tenant predicate contains it, and the storage test pins that boundary, but I would like it checked rather than assumed.

Eldad-Caura:
@claude re-review at head `85068763`. Your Medium and Low are both fixed, per-child containment with a single aggregate raise — details in the comment above.

Please also return to the questions from my first request, which the last pass did not reach. In priority order:

1. **The JSON-key delete (my original question 3).** `metadata->>'parent_memory_id'` is not a foreign key and nothing constrains what writes it. Can a writer put another row's id in their own metadata and induce the cascade to delete rows they should not reach? The tenant predicate is my defence and there is a storage test pinning it, but this is the question with the worst answer if I am wrong, and it is still unexamined.

2. **The gate.** The cascade only runs when the parent carries `auto_chunked`. Is that a correctness dependency you would accept, or would you drop the gate and pay the unindexed query cost on every drop?

3. **Coverage.** I fixed both drop branches and keep_private. The PII mask/flag branch deliberately does nothing because it neither deletes nor restricts. Check that reasoning, and check no other path leaves derived rows behind.

4. **The over-refusal guard** asserting the lookup does NOT fire for an ordinary row — is that testing something real, or is it the gate asserting itself?

Eldad-Caura:
**Correct, and it is a defect my previous fix created.** Fixed at `724a71d6`.

Making the cascade raise was right. Putting it where I put it was not: `keep_private` mutates the parent and audits it *after* — correct on its own, because nothing is lost if a non-destructive audit fails after the change — and I dropped the cascade in between. So one failed child raised before the parent's audit was ever emitted.

Reproduced before fixing:

```
AssertionError: the parent's visibility changed with no audit row: []
```

The parent durably narrowed to `scope_agent` with **no record that it happened**. An untracked mutation, which this module's own docstrings forbid.

Your comparison is the part that makes it obvious in hindsight: **the drop branches never had this exposure** because they audit before the destructive delete, which leaves the cascade already last. `keep_private` is the one branch whose natural audit position is *after* the mutation, so it was the one branch where inserting anything in between could move the audit behind a raise. I changed the ordering of a security-critical branch without re-checking what the ordering was protecting.

Parent audit now precedes the cascade, and a test pins it — reverting the order fails with the output above.

## That is two rounds running where my fix caused the next finding

Round 2 was the loop-level abort abandoning later children; round 3 is this. Both were introduced by the previous round's fix, and both were in error-handling paths rather than the feature itself. Worth saying plainly rather than letting the PR body read as a clean progression.

## On the index suggestion

Agreed and deliberately not here: a partial expression index on `(tenant_id, metadata->>'parent_memory_id')` needs a migration, and this PR is already a security fix touching two services. The `auto_chunked` gate bounds the cost meanwhile. I would rather it landed as its own change than as a fifth thing in this one — say if you disagree and I will fold it in.

## Verification

Suite **6081 passed / 0 failed** on `origin/main` at `13ebd5f7` (rebased; the branch had been rebased under me again, patch-id confirmed it was my own commit before I superseded it). ruff, mypy, ratchet, sentinel, tenant-scope gate and the broker baseline all clean after `git add`.

## Still unanswered across three requests

Not treating silence as agreement — the JSON-key delete question in particular. `metadata->>'parent_memory_id'` is not a foreign key and nothing constrains what writes it, so I traced it myself while waiting:

The deletion set is keyed on **other rows pointing at the parent**, not on anything the parent's own writer controls. To make the cascade delete a row they should not reach, an attacker would need that row's metadata to name a parent they can get dropped — and they do not write other rows' metadata. The inverse (pointing their own row at someone else's parent) deletes their own row, which is self-harm rather than escalation. Cross-tenant is blocked by the predicate, probe-confirmed.

I believe that holds, but I reasoned it rather than tested it, and it is the claim with the worst consequence if I am wrong. A second opinion on it is what I would most like from this round.

Eldad-Caura:
@claude re-review at head `724a71d6`. Your High is fixed — the parent's audit now precedes the cascade in the `keep_private` branch, with a test pinning it.

The one thing I most want from this round is the question three requests have not reached, so I have done the analysis myself and would like it falsified rather than repeated:

**Claim:** the cascade cannot be induced to delete rows the triggering writer should not reach.

The reasoning: the deletion set is `metadata->>'parent_memory_id' = <parent id>` within the parent's tenant, so it is keyed on *other rows pointing at the parent*, not on anything the parent's writer controls. An attacker cannot make a victim's row point at a parent they can get dropped, because they do not write the victim's metadata. Pointing their own row at a victim's parent deletes their own row. Cross-tenant is blocked by the predicate.

Please check that against the code — in particular whether any path lets one principal write `parent_memory_id` into a row another principal owns, and whether `metadata` is caller-writable without a reserved-key filter (I could find no such filter, which is what makes this worth asking).

Lower priority, still open from earlier rounds: whether you would drop the `auto_chunked` gate and pay the unindexed query cost; whether the PII mask/flag branch is right to cascade nothing; and whether the over-refusal test is real or the gate asserting itself.

Eldad-Caura:
**Correct, and it lands on my prose rather than my logic** — which makes it the same defect class this whole backlog keeps turning up. Fixed at `bdb9db14`, taking option (a).

I wrote *"the lookup excludes soft-deleted rows, so the children that succeeded are not revisited and only the failures are re-attempted"* — a claim about retry behaviour that depends on read-your-write, against a helper that routes every call to the replica. Under lag the retry re-fetches a child whose delete already committed, then re-audits and re-soft-deletes it: **a duplicate destructive entry in a compliance log for a row already handled.**

I chose (a) over (b) because correcting the docstring would document a worse behaviour when the fix is small and the precedent already exists — `get_memory` carries `read=False` for exactly this, and says so.

**`_get_list`'s own comment was collateral.** It read: *"All current _get_list callers are pure list/stats endpoints; none sit on the write path, so no per-call opt-out is needed yet."* True when written; **my change falsified it**. It's replaced rather than left standing next to the caller that contradicts it — the same thing I had to do to `_drop_duplicate_facts`' docstring in #1283.

## Tests

Two, both probe-confirmed:

1. **The cascade lookup goes to the writer.** Removing `read=False` fails it with `assert 0 == 1`.
2. **An over-correction guard** that adding the parameter didn't move existing callers onto the writer.

I also walked back an overclaim in my own new test while writing it: its docstring said every other `_get_list` caller "is a pure list/stats read and belongs on the replica". I hadn't verified that for all of them, and it isn't what the test checks. It now says what it actually pins — that this change doesn't alter existing routing — and leaves the per-caller question alone.

Suite **6083 passed / 0 failed**; ruff, mypy, ratchet, sentinel, tenant-scope and the broker baseline all clean after `git add`. Checks were green on `724a71d6`.

## Four rounds, three of them my own fixes

Round 2 abandoned later children; round 3 erased the parent's audit; round 4 is a docstring asserting a guarantee the transport doesn't provide. Only round 1 was in the original feature. Worth stating plainly.

## Still unanswered across four requests

The JSON-key delete question, which I analysed myself two comments up and would still like falsified rather than repeated: I could find **no reserved-key filter on caller-supplied metadata**, so `parent_memory_id` is writable by callers. My argument that this yields no escalation is reasoning, not a test, and it is the claim with the worst consequence if wrong.

Eldad-Caura:
@claude re-review at head `bdb9db14`. Your Medium is fixed by option (a) — `_get_list` gained a `read` opt-out and the cascade lookup passes `read=False`, with the stale comment that justified having no opt-out replaced.

The one question four requests have not reached, and the only one I would hold this PR for:

**Can the cascade be induced to delete rows the triggering writer should not reach?** `metadata->>'parent_memory_id'` is not a foreign key, and I could find no reserved-key filter on caller-supplied metadata — so a caller can write that key freely.

My argument that this is safe: the deletion set is keyed on *other rows pointing at the parent*, not on anything the parent's writer controls. An attacker cannot make a victim's row point at a parent they can get dropped. Pointing their own row at a victim's parent deletes their own row. Cross-tenant is blocked by the predicate, which a storage test pins.

That is reasoning, not a test. Please try to break it against the code — specifically whether any path lets one principal write metadata into a row another principal owns, and whether an agent-scoped or fleet-scoped writer can reach rows outside its own scope through this.

Lower priority, still open: whether you would drop the `auto_chunked` gate and pay the unindexed query cost; whether the PII mask/flag branch is right to cascade nothing; and whether the over-refusal test is real or the gate asserting itself.

Eldad-Caura:
**I found a High on this PR while working the sibling H-02 change, and it retracts something I told you two rounds ago.** Fixed at `486f5456`.

`GovernanceCascadeError` propagates out of `consumer.handle_memory_enriched`, which has **no guard**, and the Pub/Sub dispatcher **nacks on a handler exception** — redelivering the same event, re-running the whole drop branch, and emitting a **second `critical=True` audit** for a memory already dropped. Every redelivery.

Two rounds ago I answered the return-contract finding by saying a duplicate parent audit was *"the cost of not silently leaving forbidden content live, and pre-existing behaviour for any remediation that raises"*. **That framing assumed a one-off.** It is an unbounded loop against a tamper-evident log, for a row nothing further can be done to. I was wrong, and the reasoning that was wrong is the same reasoning I had to retract on #1297 an hour later.

## Where the catch goes, and why not deeper

In the **consumer**, not in `remediate_after_enrichment` — because its two callers need opposite things from this failure:

- `_enrich_memory_background` is about to create **more** derived rows, so it must still abort. It does not catch this, and must not.
- The consumer creates nothing. It logs at ERROR and honours the parent's verdict from the `outcome` carried on the exception — which is exactly what that field, added last round at your suggestion, is for.

Probe: removing the guard fails the new test with the raw `GovernanceCascadeError` escaping the handler.

## Verification

Suite **6098 passed / 0 failed**; ruff, mypy, ratchet, sentinel, tenant-scope and the broker baseline all clean after `git add`. Rebased onto current main (the branch had been rebased under me again; patch-id confirmed my own commit first).

## Tally, since it is not flattering

Five rounds. Round 1 found the original defect; rounds 2, 3, 4 and 5 all found problems **introduced by the previous round's fix**, every one of them in error handling rather than in the feature. That is worth saying plainly rather than letting the PR body read as steady progress.

Still unanswered across four requests: whether the cascade can be induced to delete rows the triggering writer should not reach. I analysed it myself two comments up and would still rather have it falsified than repeated.

Eldad-Caura:
@claude re-review at head `e3da5770`.

**Two things happened since your last review, and neither has been looked at.**

Your last review ran against `bdb9db14`. Since then:

1. **Round 5 (`486f5456`) — a High I found on this PR while working the sibling H-02 change, which retracts something I told you two rounds ago.** `GovernanceCascadeError` propagates out of `consumer.handle_memory_enriched`, which has no guard, and the Pub/Sub dispatcher nacks on a handler exception. That redelivers the event, re-runs the whole drop branch, and emits a second `critical=True` audit for an already-dropped memory — every redelivery, unboundedly. Two rounds ago I told you a duplicate parent audit was "the cost of not silently leaving forbidden content live, and pre-existing behaviour for any remediation that raises". That assumed a one-off. It is a loop against a tamper-evident log.

   I caught it in the **consumer** rather than in `remediate_after_enrichment`, because the two callers need opposite things from this failure: `_enrich_memory_background` is about to create more derived rows and must still abort; this handler creates nothing and must not nack. The `outcome` I attached to the exception in round 4 is what makes that possible — the branch honours the parent's verdict without re-running it.

2. **The branch was force-rebased onto current main (`76def6a1`).** I verified by patch-id that the content is unchanged (`6b0e8b455a099a4466c98cd99c256cda31e7a19a` both before and after), so this is a pure rebase — but my local green was measured against the old base, and CI on this head has not settled yet.

**What I want attacked, in priority order.**

**The consumer guard is the part with no review at all, and it is the fifth consecutive change to error handling on this PR.** Rounds 2, 3, 4 and 5 each found a defect *introduced by the previous round's fix*, every one of them in error handling rather than in the feature. That record is the reason I would rather this not land unreviewed than trust my own read of it. Specifically: is swallowing `GovernanceCascadeError` here correct in every case that reaches this handler, or is there a failure mode where the parent was NOT in fact remediated and `exc.outcome` reports a verdict that did not happen?

**Still unanswered across five requests**, and it is the one I care most about: the child lookup filters on `metadata->>'parent_memory_id'`, and nothing filters reserved keys out of caller-supplied metadata. Can a writer set that key to a memory it does not own and induce the cascade to drop or privatise rows it should not reach? I analysed this myself earlier in the thread and would still rather have it falsified than repeated back to me.

Eldad-Caura:
@claude re-review at head `fb993bfb`. **Your Medium is right, and it is worse than you framed it.**

You described a missing retry path. What was actually there was a docstring describing a retry **that the previous round had deleted**. Round 5 made the consumer catch `GovernanceCascadeError` and ack; `_drop_children` still called the redelivery "safe and narrow", and still carried the duplicate-parent-audit framing I had *already retracted in that same commit's message*. Two paragraphs asserting a mechanism the same commit removed.

So this is round 6 of a PR where rounds 2 through 6 each found something the previous round's fix introduced — and this one is the third time the defect has been in my prose rather than my logic. That is the harder direction to catch, because nothing fails.

**What changed.**

The docstring now says nothing retries a failed child, because nothing does — the parent's remediation has already succeeded and no later event revisits the row.

That makes the log the whole recovery path, so it had to become usable as one. The audit and the delete are now in **separate `try` blocks**, because they leave the row in two different states that the single block reported identically:

- `audit_emitted=False` — nothing happened; the row is live and the compliance log makes no claim about it.
- `audit_emitted=True` — the one that needs hands. The log **records the row as removed** and it is still live.

Both carry `governance_cascade_needs_manual_remediation=True` as a structured field, so an alert keys on something other than message text. Two tests, both probe-confirmed failing without the split, and they assert the **field** rather than the wording — a monitor keyed on that field breaking silently is the failure mode being guarded.

**On the ordering you were implicitly questioning:** the audit still precedes the delete, and that is what makes the bad state reachable. I kept it deliberately. A false "dropped" entry is discoverable and the content is still there to remove; auditing afterwards converts the same failure into a **deletion with no record of it at all**. `_privatise_children` orders these oppositely for the same reason read the other way — its update is non-destructive, so update-then-audit can only ever under-record. Both docstrings now say which trade they are making.

**What I did not do:** your option (1), the reconciliation sweep. It needs a query over audit rows carrying `cascaded_from` against rows with no `deleted_at`, plus a scheduler — a separate change with its own tests, not a rider on this one. I took option (2) and made the manual path real. The sweep is worth filing on its own merits and I would rather it be reviewed as its own thing than bolted on here.

Verified at this head: full suite 6100 passed, 5 skipped, 1 xfailed; ruff check and format clean at CI's scopes; mypy clean apart from the two pre-existing `types-python-dateutil` errors in untouched `common/enrichment/service.py`; ratchet "No new lines."; sentinel "All 35 protected strings survive."; tenant-scope gate exit 0.

The branch was force-rebased three times today by its author while I worked. I confirmed by patch-id each time that the content was unchanged (`6b0e8b455a099a4466c98cd99c256cda31e7a19a` throughout) before building on it, and re-ran the full suite on the final base rather than trusting the earlier run.

**Still unanswered across six requests** — the one I care most about, and the only thing on this PR I would call a security question rather than a robustness one: the child lookup filters on `metadata->>'parent_memory_id'`, and nothing strips reserved keys from caller-supplied metadata. Can a writer set that key to a memory it does not own and induce this cascade to drop or privatise rows it should not reach? I analysed it myself earlier in this thread and I would still rather have it falsified than agreed with.

Eldad-Caura:
**Answering my own question rather than asking a seventh time.** You've now passed this PR clean, and across six requests the `parent_memory_id` question has never been engaged directly. That's a fair signal it isn't going to be, so here is the analysis with the code behind it. Falsify it if you can; otherwise it stands as the record.

**The question.** The child lookup filters on `metadata->>'parent_memory_id'`, and nothing reserves that key against caller-supplied metadata — I checked, there is no reserved-key filter for memory metadata anywhere in `core-api`, `core-storage-api` or `common`. So: can a writer point the cascade at rows it should not reach?

**Cross-tenant: no, structurally.** `memory_find_children_by_parent_id` filters `Memory.tenant_id == tenant_id`, and that `tenant_id` comes from the remediated parent's own row — never from caller input. There is no input that crosses it.

**Intra-tenant, same agent: no privilege gained.** For a row to be cascaded it must *carry* `parent_memory_id = P`. Only that row's writer sets it. A writer tagging its own rows gets its own rows deleted. That is self-harm, not an attack, and it needs no vulnerability to achieve.

**Intra-tenant, cross-agent: gated behind a privilege that already subsumes it.** For agent A to get agent B's rows cascaded, A must write `parent_memory_id` into B's row. `enforce_update` (`agent_service.py:582`) refuses that below trust level 3: *"may only update its own memories"*. At trust 3 the agent can already update any memory in the tenant — rewrite its content, change its visibility, delete it outright. So the cascade hands such an attacker nothing they lack; it is strictly weaker than the privilege needed to set it up.

**Conclusion: not a privilege escalation.** The missing reserved-key filter is real, and I would still rather it existed, but every path to abusing it requires a write privilege that already permits the same outcome more directly.

**The one residual I do want on the record, because it is not a data-access problem and so does not disappear in the analysis above.** At trust 3, an attacker can cause B's rows to be soft-deleted *with `critical=True` governance audit entries attributing the removal to a policy verdict, carrying `cascaded_from`*. The deletion was already available to them. The **attribution** was not. That launders a deliberate destructive act into what a compliance review reads as automated policy enforcement — in a log this PR treats as tamper-evident and as the only trace of the removal.

I am not fixing that here. It is a property of trust-3 semantics plus audit attribution, not of this cascade, and a fix belongs where reserved metadata keys are decided — not bolted onto a remediation path. But it is the sharpest thing in this area and it should not be lost because six review rounds went by without anyone naming it.

**Filing two follow-ups rather than widening this PR:** a reserved-key filter for `parent_memory_id` (and any sibling keys the platform treats as structural) on the memory write and update paths, and the reconciliation sweep from the previous round. Both are better as their own changes with their own tests.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #1294, opened 2026-09-05
feat(ops): alert on embeddings written without provenance

Eldad-Caura:
## Why nobody noticed

241 production rows accumulated over a week. **The number that would have shown
them was already being computed** — logged at INFO every tick, and explicitly
documented as not worth alerting on:

```python
# Embedded before provenance existed: undetermined, not damaged.
# Expected to fall over time; it is a measurement gap closing, so
# do not alert on it.
```

That reasoning is sound about what it describes and wrong about what it
assumes. `unknown_provenance` genuinely **cannot** be alerted on: it holds
~95,000 legitimate pre-037 rows, so any threshold on its level is either
permanently breached or meaningless. But "do not alert" was justified by
*"expected to fall over time"* — which assumes the population only ever drains.
A defect that grows it falsifies that premise, and #1281's did, inside the one
number nobody was watching.

The premise has since been restored (that writer and the backfill in #1289 are
both fixed), but it is only ever as good as the last writer. This replaces the
assumption with a measurement.

## What this adds

`missing_provenance` — the same predicate with every innocent explanation
removed:

```sql
embedding IS NOT NULL              -- something embedded it
AND embedded_content_hash IS NULL  -- and recorded nothing
AND content_hash IS NOT NULL       -- with a hash it could have recorded
AND created_at >= '2026-08-17'     -- after provenance existed
```

**Its correct value is zero, always.** That is the whole design. It needs no
stored history, no baseline, and no tuned threshold.

### Why not the delta

The obvious answer is to alert on the *change* in `unknown_provenance`. I chose
against it:

- **It needs state.** core-operations is a stateless tick runner — it POSTs an
  endpoint and logs. Persisting a previous level for one metric means giving
  this service a store it does not otherwise have.
- **It still needs a threshold**, and the bucket legitimately moves as rows are
  rewritten, so the threshold is a tuning problem with no correct answer.
- **It only fires once a trend establishes.** The scoped invariant fires on the
  **first** bad row.

An exact invariant beats a tuned threshold whenever one is available, and here
one is.

## The alert

ERROR, not WARNING, and it carries the predicate:

```
memories embedded without provenance; a writer is dropping content_hash
  missing_provenance=241 tenants_affected=1
  worst_tenants=[{tenant_id: ..., missing_provenance: 241}]
  predicate="embedding IS NOT NULL AND embedded_content_hash IS NULL
             AND content_hash IS NOT NULL AND created_at >= '2026-08-17'"
```

ERROR because these rows are reachable by **no sweep and no endpoint** — both
embedding backfills select `embedding IS NULL` and these have a vector; the
staleness detector needs a hash to compare. Nothing drains them without a
person. The predicate rides along because an operator who sees this has no
query to run otherwise.

The per-tenant filter now includes the new bucket. It keyed on
`missing_embeddings or stale_embeddings`, and a tenant whose *only* defect is
dropped provenance has neither — so the alert would have fired with no
per-tenant line beneath it naming who to look at.

## Also: one definition instead of three

The coverage predicates were inlined separately in both queries, under a
docstring warning to *"keep the three in sync"*. A warning is not a mechanism —
and a fourth copy, in a test I wrote last week, had already drifted by omitting
the `LIVE_MEMORY_STATUSES` filter. They now come from one `_coverage_counts()`.
This is the change I'd most want a second opinion on, since it touches working
production query code in a PR about alerting.

## Verification

Reverted three ways:

| Revert | Fails |
|---|---|
| Drop the alert call | `test_missing_provenance_raises_an_error_line` |
| Revert the per-tenant filter | `test_missing_provenance_tenant_gets_a_per_tenant_line` |
| Remove the three carve-outs (new bucket ≡ `unknown`) | `test_missing_provenance_excludes_the_innocent_cases` |

The third is the one that matters. It seeds four rows — a genuine defect, a
pre-037 row, a row with no `content_hash`, and a row with no embedding — and
asserts `unknown == 3` while `missing_provenance == 1`. **That is what keeps
the alert quiet on a healthy deployment holding 95k legacy rows**, which is the
difference between an alert people act on and one they switch off.

core-operations **50 → 53**, core-storage-api **334 → 339**, root suite
unchanged at 18 known environmental failures, zero new by name. ruff clean at
CI's scopes; `mypy src/` clean in all three packages.

## Not included: the repair sweep

The task's part 2 asked whether a sweep for this population belongs here. I
left it out deliberately — it is a materially different change:

- **It needs an index.** Migration 037's partial index carries `AND
  embedded_content_hash IS NOT NULL`, so it deliberately excludes this bucket.
  A sweep written today would sequential-scan `memories`.
- **It needs a policy decision.** ~95,000 legitimate pre-037 rows are in scope
  for any such sweep, and re-embedding them costs real provider spend. Whether
  to repair them, and at what rate, is not a code question.
- **The safe repair is a re-embed, not a stamp.** `embedded_content_hash =
  content_hash` is true only for a row whose content has not changed since its
  vector was written. It held for the 241 rows repaired this week — each
  verified first against a recomputed `sha256(tenant_id || ':' ||
  coalesce(fleet_id,'') || ':' || content)` — but it must not be assumed in
  general. A wrong hash reads downstream as verified freshness and is worse
  than the NULL it replaces.

Detection is the half that pays for itself immediately and is safe to ship
alone. Happy to take the sweep as a follow-up.

Not merging.

### DIFF
--- a/core-api/src/core_api/routes/lifecycle.py
+++ b/core-api/src/core_api/routes/lifecycle.py
@@ -310,6 +310,8 @@ async def embedding_coverage_all_tenants(
             "tenants_with_missing": coverage.get("tenants_with_missing"),
             "stale_embeddings": coverage.get("stale_embeddings"),
             "unknown_provenance": coverage.get("unknown_provenance"),
+            # Should be 0. See ``run_embedding_coverage_tick``, which alerts.
+            "missing_provenance": coverage.get("missing_provenance"),
         },
     )
     return coverage
--- a/core-operations/src/core_operations/tasks.py
+++ b/core-operations/src/core_operations/tasks.py
@@ -156,12 +156,20 @@ def _alert_on_missing_provenance(coverage: dict, tenants: list[dict]) -> None:
                 {"tenant_id": t.get("tenant_id"), "missing_provenance": t.get("missing_provenance")}
                 for t in worst
             ],
-            # Names the query an operator should run, since these rows are
-            # reachable by no existing sweep or endpoint.
-            "predicate": (
-                "embedding IS NOT NULL AND embedded_content_hash IS NULL "
-                "AND content_hash IS NOT NULL AND created_at >= '2026-08-17'"
-            ),
+            # The query an operator should run — these rows are reachable by no
+            # existing sweep or endpoint, so the count alone is unactionable.
+            #
+            # Echoed from the response, never composed here. This service is a
+            # separate deployable with no import path into core-storage-api, so
+            # a literal would be kept honest by nothing: it would drift the
+            # first time the cutoff date moved, and the alert would then name a
+            # query that does not reproduce the number printed beside it.
+            #
+            # ``None`` if the storage service predates the field — which is the
+            # honest degradation during a rolling deploy. A missing hint costs
+            # an operator one lookup; a confidently wrong one costs them the
+            # investigation.
+            "predicate": coverage.get("missing_provenance_predicate"),
         },
     )
 
--- a/core-operations/src/core_operations/tasks.py
+++ b/core-operations/src/core_operations/tasks.py
@@ -117,6 +117,55 @@ async def run_embed_backfill_tick() -> None:
     await _fire_fanout("embed-backfill")
 
 
+def _alert_on_missing_provenance(coverage: dict, tenants: list[dict]) -> None:
+    """Escalate rows that record no provenance and have no excuse for it.
+
+    Alerting on a LEVEL is what the other buckets cannot do.
+    ``unknown_provenance`` sits in the tens of thousands of legitimate pre-037
+    rows, so a threshold on it is either permanently breached or useless — and
+    the conclusion drawn from that was to stop watching the number entirely,
+    on the premise that it only ever drains. In 2026-09 a defect grew it by 241
+    rows over a week inside that blind spot.
+
+    ``missing_provenance`` is the same predicate with the innocent explanations
+    removed: written after migration 037, holding a content hash to attest,
+    with a vector that something computed. Its correct value is zero, so this
+    needs no stored history, no baseline and no tuned threshold — the three
+    things a delta-based alert would need, in a service that keeps no state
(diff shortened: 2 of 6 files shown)


Eldad-Caura:
Good catch, and taken — that is the same defect class this PR exists to remove, reintroduced by the PR itself. Fixed in `3b55efe0`.

**Why I went with option (a) rather than the comment.** The duplicated coverage predicates this PR collapses already *had* a docstring telling people to keep them in sync, and a fourth copy drifted anyway (a test I wrote last week, omitting the `LIVE_MEMORY_STATUSES` filter). A warning is not a mechanism. Recommending one here would have been arguing against the change I had just made ten lines above.

**The date now exists once.** `MISSING_PROVENANCE_PREDICATE_SQL` is built from `PROVENANCE_REQUIRED_FROM` by interpolation, `/embedding-coverage-all` serves it, and core-operations echoes it. Composing the string in the consumer is the thing that can drift; echoing what the server actually used cannot.

**One thing your prompt did not raise, worth stating.** Serving the field means an older storage service during a rolling deploy sends nothing. The alert now logs `predicate: None` rather than falling back to a literal — deliberately. A missing hint costs an operator one lookup; a confidently wrong one costs them the investigation, which is the whole failure mode you flagged.

Three tests, each failing without its half:

- the predicate string must contain the constant's own date, plus all three carve-out terms — so it cannot describe a broader population than the count it explains
- the alert must echo a deliberately **unreal** served value (`created_at >= '2099-01-01'`). A hardcoded implementation would emit its own plausible-looking string and sail past a laxer assertion
- the alert must still fire when the field is absent

Verified by reverting to the literal: two of the three go red.

Gates re-run on the new head — core-operations 53 → 55, core-storage-api 339 → 340, root suite unchanged at 18 known environmental failures with zero new by name, ruff and mypy clean across all three packages.

Eldad-Caura:
@erni-a pelae pparove

erni-a (approved):

PR #1296, opened 2026-09-05
feat(mcp): refuse an over-plan write, behind a flag

Eldad-Caura:
Closes the refusal half of #1205. **Behaviour change, flag-gated, default OFF** — a rebuilt image with no env change behaves exactly as it does today.

## What was missing

An org over its plan limit is refused a write on REST and allowed the same write on MCP. Same tenant, same operation, different answer per transport.

The signal has been readable on this surface since `MCPAuthMiddleware` started honouring `x-org-read-only` — the issue's step 1, already merged. What was missing was the refusal. `_check_plan_limit` now returns the canonical MCP error envelope when the org is over plan, and both gated call sites return it.

`create` and `bulk_create` are full coverage: `redistribute` is the third member of `PLAN_LIMIT_GATED_OPS` and is not exposed as an MCP tool.

Renamed from `_observe_plan_limit`, which stopped describing what the function does.

## Why a flag rather than just enforcing

It mirrors `meter_recall_as_recall` and `meter_mcp_bulk_writes`, and it is sharper than either: **those change what is COUNTED, this changes what is ALLOWED.** An over-plan tenant that can write over MCP today stops being able to.

The flag makes the rollout reversible without a redeploy — enable it, watch `mcp_plan_limit_refused`, turn it off if the blast radius is wrong. A code-only change would need a rollback to undo. #1205's own "Blast radius — read before doing this" section asks for exactly this kind of sequencing.

## I could not confirm the log has fired, and I am not claiming otherwise

The staging exists so this gets enforced "with the number in hand", so I went looking for the number. `mcp_plan_limit_would_refuse` returns **zero results over 30 days**. That is *not* evidence that nothing would be refused, and I checked far enough to know I cannot use it:

- The pipeline is healthy — 3.69M `service:core-api` logs over 7 days, and `message` carries the structlog event name, so the query shape was right.
- But prod core-api runs `version:90b75a65…`, which is a **caura-enterprise** commit, not an OSS one. Enterprise `core-api/` contains only `helm`, so that tag is the deploy commit, not the core-api source revision — and I could not establish from here whether the deployed image predates #1221 (which landed the observation on 2026-09-01, four days ago).

So the empty log is **inconclusive**, and the flag is doing the job the log was supposed to. The remaining known cause of silence is unchanged and still open: over-plan mode is computed from counters the MCP batch path does not move while `meter_mcp_bulk_writes` is off, so enabling this flag alone enforces against counters this surface barely contributes to. The counters are per-org and REST moves them, so enforcement *will* fire — but do not read a quiet observation log as a small blast radius. That interaction is written into both settings' comments.

## Two details that are easy to get backwards

- **The gate runs BEFORE the meter.** The counter it would charge is the one over-plan mode is computed *from*, so charging a refused write would push the tenant further over the limit it was just refused for — the org digs deeper by being told no. Pinned by a test.
- **A refused batch is refused WHOLE.** Trimming it to what fits would report success for a call that did not do what it was asked, with no way for the caller to tell which items landed. REST's bulk route refuses whole for the same reason.

The refusal mirrors REST's `AuthContext.enforce_usage_limits`: same `PLAN_LIMIT_READ_ONLY` code, same message, same carve-out that deletes stay permitted. It cannot reuse that method — it raises `HTTPException`, and MCP tools return envelopes.

Enforcing emits `mcp_plan_limit_refused`; observing keeps emitting `mcp_plan_limit_would_refuse`. Reusing one name and flipping `enforced` would leave every existing query counting refusals as observations.

## Tests

14 new, and the 9 existing observation tests pass **unchanged** — which is the design working: flag off means the shipped behaviour is untouched, so the guard that says *"if someone later adds the refusal without meaning to, this fails"* still guards.

Every new test confirmed to fail under the mutation it exists to catch:

| Mutation | Fails |
|---|---|
| helper never returns an envelope | 6 |
| call sites discard the envelope | 4 (end-to-end only — helper tests still pass, which is why the end-to-end ones exist) |
| gate moved after the meter | 1 (the ordering test, precisely) |
| one event name for both modes | 3 |
| `plan_limit_gated` check dropped | 4 (the ungated ops) |
| refuse regardless of read-only | 2 (the over-reach guards) |
| **flag default flipped to True** | 4 — including the **pre-existing** `test_an_over_plan_write_still_succeeds` |

That last row is the one I would look at first: an accidental default flip is caught by the guard the previous author left behind, not only by my own test.

796 passed across `-k "mcp or usage or plan or quota or write"`. `ruff check` / `ruff format --check` at CI's scopes, `mypy src/`, both naming gates.

## Stale comments swept, not spot-fixed

This change falsified several comments — including ones #1288 and #1290 had just corrected. I swept for every variant rather than fixing the ones I happened to look at, having had that exact lesson twice this week. `usage_service`'s policy note now records the history explicitly, because it has been wrong in the same direction twice.

### DIFF
--- a/core-api/src/core_api/config.py
+++ b/core-api/src/core_api/config.py
@@ -416,11 +416,32 @@ class Settings(BaseSettings):
     # Crossing it bites on REST FIRST. Over-plan mode is computed from these
     # counters and stamped as ``x-org-read-only``, which core-api turns into a
     # 403 on ~22 REST write routes — while the MCP surface only OBSERVES it
-    # (see ``_observe_plan_limit``). So enabling this can refuse a tenant's
+    # (see ``_check_plan_limit``). So enabling this can refuse a tenant's
     # REST writes because of what it wrote over MCP. That is the intended end
     # state, but it is not a deploy side effect. Off = the historical (unbilled)
     # behavior. See caura-ai/caura#1220.
     meter_mcp_bulk_writes: bool = False
+    # Refuse MCP writes when the org is over its plan limit, as the ~22 REST
+    # write routes already do. Off by default, and this is the sharpest of the
+    # three flags above it: the other two change what is COUNTED, this one
+    # changes what is ALLOWED. Flipping it takes away a capability tenants have
+    # today — an over-plan org that can currently write over MCP stops being
+    # able to — so it is a deliberate, announced change, not a deploy side
+    # effect. Off = observe and log only (``_check_plan_limit``).
+    #
+    # It is a flag rather than a straight code change so the rollout is
+    # reversible without a redeploy: enable it, watch
+    # ``mcp_plan_limit_refused``, and turn it off again if the blast radius is
+    # wrong. A code-only change would need a rollback to undo.
+    #
+    # Read ``_check_plan_limit``'s docstring before enabling. A quiet
+    # observation log is NOT evidence that nothing will be refused — the
+    # counters over-plan mode is computed from barely move for MCP-first
+    # tenants while ``meter_mcp_bulk_writes`` is off, which is the population
+    # this refusal is aimed at. The two flags interact: turning THIS on while
+    # that one is off enforces a limit against counters the MCP surface hardly
+    # contributes to. See caura-ai/caura#1205.
+    enforce_mcp_plan_limits: bool = False
     stm_backend: str = "memory"  # memory | redis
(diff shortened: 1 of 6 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #1297, opened 2026-09-05
fix(governance): remove the entity rows mined out of a dropped memory

Eldad-Caura:
Closes audit finding **H-02**.

#808 named this case when it fixed the inline path: "entities mined out of
dropped content are the same leak in another table". It fixed that path by
ordering — _enrich_memory_background runs remediation first, and its early
return on a drop skips the entity extraction scheduled below it.

Both non-inline paths schedule extraction independently, at write time, as a
fire-and-forget task that races the verdict. ScheduleBackgroundTasks fires it
alongside the enrichment carrying run_governance_remediation=True in the fast
branch, and at write time in strong+deferred. process_entity_extraction never
re-checked the row.

And the schema's own expression of "these rows must not outlive the memory"
never fires: memory_entity_links.memory_id is ON DELETE CASCADE and
relations.evidence_memory_id is ON DELETE SET NULL, both on a HARD delete.
Governance soft-deletes — it sets deleted_at — so neither ever runs. The entity
row itself has no FK to the memory at all, so nothing would remove it even on a
hard delete.

Result: a tenant configured to drop had the memory removed and audited while the
names mined from it (person names, under a PII policy) stayed listable
tenant-wide through /entities and /graph, with nothing tying them to the drop.

Verified from the code rather than reproduced as one failing assertion, and the
distinction is worth being straight about: unlike the earlier findings in this
series there was no existing code path to make fail, because nothing could reach
these rows at all. What IS probe-confirmed is each guard added here — reverting
the candidate bounding fails the over-deletion test with 2 == 1, and neutering
the liveness check fails the dropped-row test.

Two halves, and they are not alternatives.

1. A purge on the drop path. New storage call, one transaction: delete the
   memory's entity links, then relations whose evidence IS this memory (one row
   carries one evidence id, so a relation attributed to dropped content has no
   other justification), then — from the entities this memory linked to and only
   those — the ones now left with no links and no relations. Both destructive
   dispositions cascade, not just the non-business one the finding described:
   they are separate branches reading separate configs, and a PII drop policy
   leaked identically.

   The candidate set is bounded on purpose. A first draft deleted every entity in
   the tenant with no links, which would sweep entities orphaned for unrelated
   reasons and race an entity a concurrent write had created but not yet linked.
   Under-deleting is recoverable; over-deleting another caller's rows is not. A
   test pins it: the unbounded version fails with 2 == 1.

2. A liveness re-check in the worker, immediately before persisting, reading the
   WRITER — the whole point is to observe a delete that just committed, and a
   replica under lag would report the row live exactly when the check most needed
   to fail.

Half 1 covers the common ordering: extraction is one LLM call while the verdict
needs enrichment plus an event round-trip, so extraction usually finishes first
and its rows are there to purge. Half 2 covers the tail where it does not — the
purge has already run by then and would miss what lands afterwards. Neither half
covers the other's case.

The purge is deliberately NOT gated on a marker, unlike H-10's child cascade: any
dropped memory may have been extracted from, no flag on the row says so, and the
purge is three targeted deletes keyed on memory_id.

The purge runs AFTER the soft-delete, and that ordering is pinned. Purging first
would destroy graph rows for a memory that is still live if the delete then
failed, and nothing would put them back.

Tests. Five in core-storage-api against real Postgres, because the query's
correctness is entirely about what it does and does not reach and a stub would
assert the code calls itself — including the over-deletion guard, the
still-asserted-by-another-memory case, and the tenant boundary. Three in
core-api for the wiring, one of them an over-refusal guard that flag and
keep_private purge nothing (those rows describe content that is still there and
still allowed). Two for the worker guard, asserting on the WRITES rather than
the early return so a refactor that keeps the check and persists anyway fails,
plus one that the check reads the writer.

Four existing entity-extraction test files gained a get_memory stub. That is the
honest cost of the worker now depending on a storage read it did not before, not
churn to hide a problem.

Overlaps #1292 (H-10), which is in review and adds its own cascade to the same
two drop branches. Whichever lands second needs a mechanical rebase; the two
mechanisms are independent — one covers rows derived into the memories table,
this one covers rows derived into the graph.

## Verification

- Full root suite: **6077 passed, 5 skipped, 1 xfailed, 0 failed**.
- `core-storage-api/tests/`: **343 passed**, on its own scratch database.
- `ruff check` and `ruff format --check` run separately at CI's exact scopes — clean.
- `mypy` clean on both packages (`core-api` keeps its 2 pre-existing `types-python-dateutil` errors in an untouched file). The three new `rowcount` ignores match `memory_soft_delete_by_ids`' existing pattern.
- ratchet *No new lines.* · sentinel *All 35 protected strings survive.* · tenant-scope gate exit 0 · broker OpenAPI baseline current. All after `git add`.
- Checked for open PRs on this subsystem before starting; only #1292, noted above.
- Branched from `origin/main`.

### DIFF
--- a/core-api/src/core_api/clients/storage_client.py
+++ b/core-api/src/core_api/clients/storage_client.py
@@ -948,6 +948,32 @@ class CoreStorageClient:
             "/memories/by-parent-id", read=False, tenant_id=tenant_id, parent_id=parent_id
         )
 
+    async def purge_entity_artifacts(self, tenant_id: str, memory_id: str) -> dict:
+        """H-02 — remove the graph rows mined out of a dropped memory.
+
+        Returns per-table counts (``links``, ``relations``, ``entities``) so the
+        caller can audit what a governance drop actually removed.
+
+        Un-tagged POST, so it lands on the WRITER — it deletes.
+
+        ``idempotent=True``, which this client reserves for endpoints where a
+        replay is harmless. The usual justification is a storage-side dedup key;
+        this one qualifies for a different reason — a replay finds the rows
+        already gone and deletes nothing more. Worth opting in explicitly
+        because the caller lets failures propagate: without it a transient 5xx
+        aborts a remediation whose soft-delete has ALREADY committed, leaving
+        the graph rows behind until someone notices the failed task.
+
+        The one cost of a replay is cosmetic: a lost response followed by a
+        successful retry logs zero counts for a purge that did remove rows. A
+        wrong number in an INFO line, against leaving forbidden content live.
+        """
+        return await self._post(  # type: ignore[return-value]
+            "/memories/purge-entity-artifacts",
+            {"tenant_id": tenant_id, "memory_id": memory_id},
+            idempotent=True,
+        )
+
     async def find_successors(self, data: dict) -> list[dict]:
         return await self._post("/memories/find-successors", data, read=True)  # type: ignore[return-value]
 
--- a/core-api/src/core_api/services/entity_extraction_worker.py
+++ b/core-api/src/core_api/services/entity_extraction_worker.py
@@ -3,6 +3,7 @@
 import asyncio
 import logging
 import re
+from typing import Any, Literal
 from uuid import UUID
 
 from common.embedding import get_embedding
@@ -93,6 +94,129 @@ async def _discover_cross_links_for_memory(
         )
 
 
+async def _purge_written_artifacts_if_dropped(
+    sc: Any, memory_id: UUID, tenant_id: str
+) -> Literal["live", "dropped", "unknown"]:
+    """Undo our own graph writes when the memory died while we were making them.
+
+    Reports WHAT IT FOUND and leaves the policy to the caller, because the two
+    call sites want different things from the same answer. An earlier revision
+    returned a bool meaning "you must stop", which forced the indeterminate case
+    to pick one of the two real answers and pretend — it chose "stop", and that
+    silently discarded the audit-log entry, the contradiction trigger and
+    cross-link discovery for a memory that a transient read timeout had said
+    nothing bad about. Three states, named:
+
+    ``dropped``
+        The row is gone or soft-deleted, and its graph rows have been purged
+        (or the purge failed, loudly — either way it is not coming back).
+    ``live``
+        The row is there. Nothing was purged and nothing should stop.
+    ``unknown``
+        The read itself failed. Nothing is known and nothing was purged.
+
+    Purging is only half of the job at the first call site: everything after the
+    link upsert — relation upserts carrying ``evidence_memory_id``, the subject
+    write-back, cross-link discovery — writes MORE graph rows for this memory. A
+    version that purged and fell through cleaned the table and then immediately
+    refilled it, moving the leak from the link table to the relation table rather
+    than closing it.
+
+    H-02. The liveness check before the persistence block narrows the window; it
+    cannot close it. Between that check and the links landing there are embedding
+    round-trips and three storage calls, and a governance drop can complete
+    inside them — including its own purge, which finds nothing because these rows
+    do not exist yet. The entities then land moments later and nothing ever
+    revisits them: the memory is gone, so no verdict names it again.
+
+    Re-checking AFTER the writes closes it, and the argument is about what is
+    observable rather than about timing:
+
+    * the drop committed before our writes — its purge found nothing, but this
+      check sees the row deleted, and we purge what we just wrote,
+    * the drop commits after our writes — its own purge sees our rows and takes
+      them,
+    * the drop commits between the two — whichever purge runs later sees the
+      rows, and both are keyed on the same ``memory_id``.
+
+    That argument only holds for rows written BEFORE the call, which is why
+    ``process_entity_extraction`` calls this THREE times rather than once. An
+    earlier revision called it once, after the link upsert, and claimed "there is
+    no ordering left in which the rows survive" — untrue of everything written
+    afterwards: the subject write-back, the relation upserts, the links
+    cross-link discovery creates. The call sites, and what each is for:
+
+    * after the link upsert — an optimisation. It saves the relation upserts and
+      a cross-link pass when the row is already gone, and is allowed to fall
+      through on anything short of ``dropped``.
+    * at the end of the ``try`` — the guarantee for the path that completes. It
+      sits after every graph-mutating write, so the rows it can find are all of
+      them.
+    * in the ``except`` — the same guarantee for the path that leaves by
+      raising, which the previous one cannot reach.
+
+    Both of the trailing two are guarded on ``wrote_graph_rows``, so a run that
+    wrote nothing does not pay for a read, and the ``except`` one also runs on
+    failures early enough that ``sc`` does not exist yet.
+
+    The WRITER read is load-bearing at every site for the same reason as the
+    pre-write check: the whole question is whether a delete that just committed
+    is visible.
+
+    A failed PURGE still reports ``dropped``: the memory is gone whether or not
+    the cleanup worked, and the caller's decision does not change. Only a failed
+    READ is ``unknown``.
+    """
+    try:
+        live = await sc.get_memory(str(memory_id), tenant_id, read=False)
+    except Exception:
+        logger.exception(
+            "entity extraction: could not establish whether memory %s survived its own "
+            "extraction; nothing purged and nothing concluded",
+            memory_id,
+            extra={"liveness_check": "unknown", "memory_id": str(memory_id)},
+        )
+        return "unknown"
+
+    if live is not None and live.get("deleted_at") is None:
+        return "live"
+
+    try:
+        counts = await sc.purge_entity_artifacts(tenant_id, str(memory_id))
+    except Exception:
+        logger.exception(
+            "entity extraction: memory %s was dropped while its entities were being "
+            "written, and the rows just written could NOT be purged; they are live in "
+            "the graph for content the policy removed",
+            memory_id,
+        )
+        return "dropped"
+    if not isinstance(counts, dict):
+        # Same gap as the governance-side purge: reading ``counts`` as a dict
+        # would raise from outside the try above. Lower stakes at two of this
+        # function's three call sites, which sit inside the worker's catch-all —
+        # but NOT at the one in the ``except`` handler, where a raise escapes
+        # ``process_entity_extraction`` entirely and surfaces as an unhandled
+        # task exception. The purge still happened as far as anything here can
+        # tell; only the counts are unreadable.
+        logger.warning(
+            "entity extraction: memory %s was dropped mid-extraction and the purge "
+            "answered with %s rather than an object, so what it removed is unknown",
+            memory_id,
+            type(counts).__name__,
+        )
+        return "dropped"
+    logger.warning(
+        "entity extraction: memory %s was dropped mid-extraction; purged the graph "
+        "rows just written for it (links=%s relations=%s entities=%s)",
+        memory_id,
+        counts.get("links"),
+        counts.get("relations"),
+        counts.get("entities"),
+    )
+    return "dropped"
+
+
 async def process_entity_extraction(
     memory_id: UUID,
     tenant_id: str,
@@ -109,6 +233,18 @@ async def process_entity_extraction(
     # calls. Full migration: CAURA-593 lands Pub/Sub first, then a new
     # worker service subscribes to ``Topics.Pipeline.ENTITY_EXTRACT_REQUESTED``
     # and this function becomes its handler body.
+    #
+    # H-02. Guards both of the trailing liveness checks — the one at the end of
+    # the ``try`` and the one in the ``except``. It means "this memory MAY have
+    # graph rows", not "the persistence block ran": there are two independent
+    # writers below (the entity/link upserts, and cross-link discovery, which
+    # runs even when every extracted name was filtered out), and both set it.
+    #
+    # False means nothing was written, so there is nothing to purge and no reason
+    # to spend a writer read. That matters because the early exits above the
+    # persistence block — no entities extracted, row already gone — are the
+    # common case, and because ``sc`` is not yet bound on the first of them.
+    wrote_graph_rows = False
     try:
         # A5c: resolve tenant_config BEFORE the extraction call so the
         # tenant-level ``entity_extraction.provider`` / ``.model``
@@ -124,6 +260,33 @@ async def process_entity_extraction(
             return
 
         sc = get_storage_client()
+
+        # H-02: is the memory still there? This is scheduled fire-and-forget at
+        # write time on both non-inline paths, in parallel with the enrichment
+        # that carries the governance verdict — so by the time the LLM call above
+        # returns, the policy may already have dropped the row. Writing entities
+        # for it would re-create the leak the drop exists to close, in a table
+        # the drop does not reach.
+        #
+        # ``read=False`` — the WRITER. The whole point is to observe a delete
+        # that just committed; a replica under lag would report the row live and
+        # this check would pass exactly when it most needed to fail.
+        #
+        # This check alone does NOT close the window, and it is not claimed to:
+        # the writes below are several round-trips away, so a drop can land after
+        # it passes. ``_purge_written_artifacts_if_dropped`` at the end of the
+        # persistence block is what closes that; this one is here to avoid doing
+        # the work at all in the common case where the row is already gone.
+        live = await sc.get_memory(str(memory_id), tenant_id, read=False)
+        if live is None or live.get("deleted_at") is not None:
+            logger.info(
+                "entity extraction: memory %s is gone by the time extraction finished; "
+                "discarding %d extracted entit(ies) rather than writing them to a "
+                "dropped row's graph",
+                memory_id,
+                len(graph.entities),
+            )
+            return
 
         blocklist = tenant_cfg.entity_blocklist
 
@@ -342,6 +505,18 @@ async def process_entity_extraction(
             # the TOCTOU race where another writer created the natural-
             # key match between our resolve and our upsert — semantically
             # equivalent to ``updated`` for the worker.
+            # H-02. From here this memory MAY have graph rows, so every exit
+            # from this function owes it a liveness check — including the ones
+            # that leave by raising.
+            #
+            # Set BEFORE the await, not after. "The call raised" does not mean
+            # "nothing was written": a timeout can land on a request the storage
+            # side already committed, and a malformed response does not un-write
+            # rows either. The flag has to mean "might have written", because
+            # the only failure it is allowed to make is the cheap one — a wasted
+            # writer read on a call that never landed, on a path that is already
+            # an error. Getting it wrong the other way leaks the rows.
+            wrote_graph_rows = True
             upserted = await sc.bulk_upsert_entities(items=upsert_items)
             # Explicit loop (not a comprehension) so an out-of-range
             # ``input_idx`` from a misbehaving storage response surfaces
@@ -349,6 +524,12 @@ async def process_entity_extraction(
             # length-mismatch warning above on ``bulk_resolve_entities``
             # — same "treat malformed responses defensively" pattern.
             name_to_id: dict[str, UUID] = {}
+            # H-02. Only the rows THIS run brought into existence. An entity that
+            # already existed is reachable through whatever linked it before and
+            # is nobody's orphan; a created one is reachable only through the link
+            # we are about to write. See the link upsert below for why that
+            # distinction is worth carrying.
+            created_entity_ids: list[str] = []
             for r in upserted:
                 if not r.get("entity_id"):
                     continue
@@ -361,6 +542,8 @@ async def process_entity_extraction(
                     )
                     continue
                 name_to_id[filtered[idx][0]] = UUID(r["entity_id"])
+                if r.get("action") == "created":
+                    created_entity_ids.append(str(r["entity_id"]))
 
             # ---- Step 3: bulk entity-link upsert ----
             #
@@ -401,7 +584,74 @@ async def process_entity_extraction(
                 )
                 link_idx += 1
             if link_items:
-                link_result = await sc.bulk_upsert_entity_links(tenant_id, items=link_items)
+                try:
+                    link_result = await sc.bulk_upsert_entity_links(tenant_id, items=link_items)
+                except Exception:
+                    # H-02, and this is the one gap the purge cannot close from
+                    # its own side. ``memory_purge_entity_artifacts`` finds
+                    # entities THROUGH the memory's links — deliberately, because
+                    # the first draft swept every unlinked entity in the tenant
+                    # and would have raced a concurrent writer that had created
+                    # an entity but not yet linked it. Over-deleting is the
+                    # direction that does not come back.
+                    #
+                    # The cost of that bounding is exactly here: entities
+                    # committed a moment ago whose links never landed are
+                    # reachable by nothing, so a purge for this memory runs,
+                    # finds no links, and honestly reports zero. Indistinguishable
+                    # in the audit trail from "there was nothing to purge" —
+                    # which is what makes it worth a log rather than nothing.
+                    #
+                    # Widening the candidate set instead was considered and not
+                    # done: passing these ids to the purge re-introduces the same
+                    # race the bounding exists to prevent, since an id we upserted
+                    # may be a row another writer created and is about to link.
+                    # A person with a concrete list can check that; a query
+                    # cannot.
+                    #
+                    # Re-raised, so the except handler below still runs its
+                    # liveness check and purges whatever IS reachable.
+                    if created_entity_ids:
+                        logger.exception(
+                            "entity extraction: entity rows for memory %s were committed but "
+                            "their links were not, so a governance purge for this memory will "
+                            "find and report nothing while these rows survive; entity ids "
+                            "created by this run: %s",
+                            memory_id,
+                            ", ".join(created_entity_ids),
+                            extra={
+                                "unlinked_entity_ids": created_entity_ids,
+                                "memory_id": str(memory_id),
+                                "tenant_id": tenant_id,
+                            },
+                        )
+                    raise
+                # H-02: the rows exist NOW, so from here the leak is recoverable
+                # by the same purge governance uses. See the helper for why this
+                # is where the window actually closes.
+                #
+                # The RETURN has to be honoured, not just the purge. Everything
+                # below writes more graph rows for this memory — relations
+                # carrying ``evidence_memory_id``, the subject write-back,
+                # cross-link discovery. Purging and then falling through refills
+                # the table we just cleaned.
+                #
+                # ``dropped`` ONLY. This site is an optimisation — it saves the
+                # relation upserts and a cross-link discovery pass — so it must
+                # not act on ``unknown``. Stopping here also skips the audit-log
+                # entry below, and losing an audit record to a read timeout is a
+                # real cost with no later repair.
+                #
+                # Falling through on ``unknown`` is safe because a later check
+                # covers every remaining exit: the one at the end of the ``try``
+                # if the rest of the run succeeds, the one in the ``except`` if
+                # it raises. Neither is a guarantee that the rows are cleaned —
+                # both call the same read, and a read that failed once will
+                # likely fail again — but the failure is a bounded leak that
+                # governance's purge can still reach, and the alternative was
+                # destroying an audit record nothing rebuilds.
+                if await _purge_written_artifacts_if_dropped(sc, memory_id, tenant_id) == "dropped":
+                    return
                 # Surface any FK violations from the per-item path
                 # (storage-side reports ``error="fk_violation"`` for rows
                 # whose memory_id or entity_id no longer exists). Same
@@ -520,6 +770,17 @@ async def process_entity_extraction(
 
         # Cross-link discovery (non-fatal)
         if tenant_cfg.auto_entity_linking_enabled:
+            # H-02. This writes ``memory_entity_links`` rows for THIS memory —
+            # storage-side it is an ON CONFLICT DO NOTHING insert into exactly
+            # the table the purge deletes from — and it does not go through the
+            # persistence block above, so it is a second, independent way for
+            # this memory to acquire graph rows. It runs even when every
+            # extracted name was filtered out and ``bulk_upsert_entities`` was
+            # never called.
+            #
+            # Before the await, for the same reason as the upsert: a call that
+            # raised may still have committed.
+            wrote_graph_rows = True
             try:
                 await _discover_cross_links_for_memory(memory_id, tenant_id, fleet_id)
             except Exception:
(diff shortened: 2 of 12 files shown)


Eldad-Caura:
@claude please review at head `9550adfc`. This deletes rows across three tables, so attack it rather than approve it.

What I most want challenged:

1. **The orphan deletion.** Entities are removed only when they were linked to THIS memory and are left with no links and no relations. I wrote an unbounded version first — every entity in the tenant with no links — and caught it myself; a test now pins the bounded behaviour. Please check the bounded version for the opposite failure: is there a shape where an entity that SHOULD be removed survives, e.g. one reachable only through a relation whose other endpoint is unrelated?

2. **Relations.** I delete relations whose `evidence_memory_id` is this memory, on the grounds that the column is singular so a relation has exactly one evidence row. Verify that — if a relation can be justified by more than one memory anywhere in the code, deleting on a single evidence match destroys a still-supported edge.

3. **Concurrency.** The purge is one transaction, but extraction for a DIFFERENT memory could be resolving the same entity concurrently. Can that interleave into deleting an entity another memory has just linked, or into a FK violation?

4. **The two halves.** I argue the worker liveness check and the purge cover different interleavings and neither subsumes the other. If the purge alone is sufficient, the check is a storage round-trip on every extraction that I should drop.

5. **Four existing test files** gained a `get_memory` stub. Confirm that is the honest consequence of the worker's new dependency and not a case of me changing tests to fit a change that should not have been made.

Note this overlaps #1292 (H-10), which adds a different cascade to the same two drop branches and is in review. Whichever lands second needs a mechanical rebase.

Eldad-Caura:
**Both correct.** Fixed at `82d5dc43`.

## The High, and why my own reasoning was wrong

I wrote that failures propagate here *"matching every other unapplied-policy path in this module"*. That is wrong about the **caller**, and you traced why: the other paths run under `_enrich_memory_background`, where a raise becomes a `BackgroundTaskLog` row. **This one also runs under `consumer.handle_memory_enriched`, which has no guard, and the dispatcher nacks on a handler exception.** So a raise redelivers the same event, re-runs the whole drop branch, and emits a **second `critical=True` audit** for a memory already dropped. Repeatedly.

I verified the invariant rather than taking it on trust — `handle_memory_enriched` calls remediation unguarded at `consumer.py:107`, and the dispatcher's ack/nack is at `pubsub.py:1137`.

The purge failure is now caught and logged at ERROR naming the memory. The trade is bounded and worth stating: the memory is already gone, so **the content is not live** — what remains is graph rows, and the log is enough to purge them by hand. Probe: letting it propagate again fails the new test with the raw `RuntimeError`.

## The Low

Taken. The client reserves `idempotent=True` for endpoints that dedup replays storage-side; this one qualifies for a different reason worth writing down — **a replay finds the rows already gone and deletes nothing more**. That also means transient 5xx never reach the catch above.

One cost I recorded rather than hid: a lost response followed by a successful retry logs zero counts for a purge that did remove rows. A wrong number in an INFO line, against leaving forbidden content live.

## This same defect is in #1292, and I am fixing it there next

`#1292`'s cascade raises `GovernanceCascadeError` out of **the same unguarded handler**, so it has the identical nack-and-redeliver problem. Its own review round flagged the return-contract half of this and I answered that a duplicate parent audit was *"the cost of not silently leaving forbidden content live"* — your finding shows that is not a one-off cost but an unbounded redelivery loop. I was wrong there too, and for the same reason.

Flagging it here because the two PRs touch the same branches and the fix has to be consistent between them.

## Verification

Suite **6092 passed / 0 failed** on the rebased tree; storage suite 343; ruff, mypy, ratchet, sentinel, tenant-scope and the broker baseline all clean after `git add`.

Eldad-Caura:
@claude re-review at head `507f40b4`.

**Your High is fixed, and the fix itself has not been reviewed.**

Your last review ran against `9550adfc`. Since then:

1. **`82d5dc43` — the nack fix.** You were right and my reasoning was wrong about the *caller*, not the mechanism. I wrote that purge failures propagate "matching every other unapplied-policy path in this module". The other paths run under `_enrich_memory_background`, where a raise becomes a `BackgroundTaskLog` row. This one also runs under `consumer.handle_memory_enriched`, which has no guard, and the dispatcher nacks on a handler exception — so a raise redelivered the event, re-ran the drop branch, and emitted a second `critical=True` audit for an already-dropped memory, repeatedly.

   The purge failure is now **caught at the source** in `_purge_entity_artifacts` and logged at ERROR naming the memory. Note this is deliberately *not* the shape I used for the same defect class in #1292, where the catch sits in the consumer: there, `_enrich_memory_background` must still abort because it is about to create more derived rows. Here, a purge failure needs to abort nobody — the soft-delete has already committed, the content is gone, and what remains is graph rows plus a log line good enough to purge them by hand. A test pins it: letting the failure propagate again fails with the raw `RuntimeError`.

2. **The branch was force-rebased onto current main (`76def6a1`).** Patch-id unchanged (`a1e2794051c7de7b3f0ab9f86f9a7499f9661098` before and after), so a pure rebase — but my local green predates the new base and CI on this head has not settled.

**What I want attacked.**

**The catch site.** Swallowing here is a real trade, not an obvious win: is there a caller or a failure mode where a silently-failed purge should have aborted something downstream? I convinced myself there is not because the purge runs last and nothing reads its result — falsify that if you can.

**The over-deletion bound, again.** It is the part I would least like to be wrong about, because under-deleting is recoverable and deleting another caller's rows is not. The candidate set is bounded to entities this memory linked to, then filtered to those with no remaining links and no remaining relations. My first draft deleted every unlinked entity in the tenant. Is the bounded version still reachable by a concurrent write — an entity created and linked between the candidate `SELECT` and the `DELETE` inside the same transaction?

**One thing I will say plainly rather than let the test count imply otherwise:** unlike the other findings in this series there was no pre-existing code path to make fail, because nothing could reach these rows at all. What is probe-confirmed is each guard I added — reverting the candidate bounding fails the over-deletion test with `2 == 1`, and neutering the liveness check fails the dropped-row test.

Eldad-Caura:
@claude re-review at head `5abc12d2`. **Your High is correct and the fix is in.**

You were right, and the tell was a comment I wrote three lines below the bug: *"Tenant-scoped like everything else here."* The entity delete was. The two statements above it were not. That is the same defect class this backlog keeps turning up — prose asserting a property the code does not have — and it has now landed in my own writing on both PRs in this pair.

**What was wrong.** `memory_entity_links` has no `tenant_id` column, so `memory_id` alone identifies a row without authorising anything. This file already knows that: `_link_within_tenant` sits **in the method immediately above mine**, `_owned_link_endpoints` guards the write side, and both exist because of GHSA-wgvw-28pq-jc36. I used neither. A caller naming a memory its tenant does not own deleted the owning tenant's link rows and received a count saying how many.

**Reachability, stated honestly rather than talked up.** The live caller passes tenant and memory from the same row, so this is not currently exploitable through governance. It is the invariant breaking ahead of anything exploiting it, on a storage endpoint whose entire job is to be called with caller-supplied ids. I would not have argued it down to Medium on that basis.

**One deliberate divergence, which I want you to attack specifically.** Both statements are now scoped through `Memory` — the memory end only, **not** `_link_within_tenant`. That helper requires *both* ends because a READ returning a straddling row hands back the other tenant's UUID. Deleting asks a different question: the row references a memory we own and are dropping, so a foreign entity on the far end is a reason to keep the **entity** (the tenant-scoped entity delete already does) and never a reason to keep a **link** pointing at dropped content. Requiring both ends would strand exactly the historical straddling rows the write path has refused to create since #1085/#1124 — the ones most likely to be real. If you think that reasoning is wrong, this is the place to say so.

**Tests.** The mismatched-pairing one is probe-confirmed: without the scoping it fails with `links: 1` where `0` is required. The straddling-link one I will not oversell — it passes with or without the original bug. It exists to pin the choice above, and fails if someone later "fixes" this by reaching for `_link_within_tenant`.

You were also right that `test_does_not_cross_tenants` could not have caught this: it uses matched tenant/memory pairs throughout, so the unscoped delete only ever touched rows the caller did own.

Verified at this head: core-storage-api 353 passed, core-api 6092 passed, ruff check and format clean at CI's scopes, mypy clean on `core-storage-api/src/`, ratchet "No new lines.", sentinel "All 35 protected strings survive.", tenant-scope gate exit 0.

Note the branch was force-rebased three times today by its author while I worked; I confirmed by patch-id each time that content was unchanged (`a1e2794051c7de7b3f0ab9f86f9a7499f9661098` throughout) before building on it.

Eldad-Caura:
@claude re-review at head `0bc8a5c6`. **Your Low is taken, but not as suggested — the suggested form deletes rows it shouldn't, and I have the failing test.**

You were right that the three anti-join subqueries were unscoped and that correctness was never at risk from that. Narrowing them is worth doing on a path a drop-configured tenant runs constantly.

**Where I diverged.** The proposal scopes the two relation subqueries on `Relation.tenant_id`. That filter drops a historical straddling relation — one living in another tenant while naming an entity *here* — out of the "still referenced?" set. The entity is then deleted while something still points at it.

I wrote the test before believing myself, and it fails under the suggested version with `entities: 1` where `0` is required.

**What I did instead.** All three subqueries are narrowed by the **entity's** tenant — joined to `Entity`, filtered on `Entity.tenant_id` — including the two relation ones. Same reduction in scan; cannot lose a reference. Every row capable of naming a candidate names an entity in this tenant, because that is the definition of a candidate. And erring wide is free here: a surplus reference only keeps an entity alive, which is the direction this file already learned to prefer in round 1, when my first draft over-deleted.

`still_linked` I took essentially as you wrote it — the `Entity` join was already the right shape there, and I applied that same shape to the relations rather than reaching for the column just because it exists.

Verified at this head: core-storage-api 354 passed (353 + the new straddling-relation test), ruff check and format clean, mypy clean on `core-storage-api/src/`, ratchet "No new lines.", sentinel "All 35 protected strings survive.", tenant-scope gate exit 0.

Worth noting for the record that both fixes on this PR — round 3's tenant scoping and round 4's narrowing — turned on the same distinction: which end of a link or relation you key on. Keyed on the memory, deleting is safe and reading leaks. Keyed on the entity, reading is safe and deleting over-reaches. The two halves of this function key on opposite ends deliberately, and each now says so where it does it.

Eldad-Caura:
@erni-a please approve

erni-a (dismissed):

Eldad-Caura:
@claude re-review at head `5855c6de`.

**Your Medium was right, and it found the hole in my own reasoning rather than in my code.**

Rebased onto main now that #1292 has merged — four files conflicted, all additive collisions between the two cascades. One resolution was a real decision rather than a mechanical merge, flagged below.

## The Medium: the check ran too early to close the window it targeted

You're correct, and the tell was my own comment: *"This narrows the window rather than closing it… the cascade in `governance_remediation` is what covers that ordering."* The cascade covers extraction finishing **before** the verdict. It does nothing for the case you described — a drop landing during the writes, whose purge runs while these rows don't exist yet, after which the entities land and nothing ever revisits them, because the memory is gone and no later verdict names it.

So the fix had a blind spot shaped exactly like the leak it was closing.

**Closed by re-checking after the writes and purging what was just written**, which I picked over your option (a) because moving the check earlier still leaves a race — it just makes it smaller. The post-write version is an argument about what's observable rather than about timing:

- drop committed **before** our writes → its purge found nothing, our re-check sees the row deleted, we purge
- drop commits **after** our writes → its own purge sees our rows and takes them
- drop commits **between** → whichever purge runs later sees the rows; both key on the same `memory_id`

No ordering survives. I considered your option (c), enforcing it at the storage layer, and it is the better long-term shape — but it changes the contract of link and relation upserts for every caller, which is not something to slip into this PR.

Failures log rather than raise: this runs after the links are written, so raising would abort the subject write-back and cross-link discovery below over a cleanup concern.

The early check stays, downgraded in the comments to what it honestly is — an optimisation that avoids the work when the row is already gone.

**Two tests.** The drop-during-writes one is probe-confirmed: removing the post-write call fails it. The second is an over-refusal guard on the ordinary live-row path, and I want to be explicit about why it exists rather than folding it into the first — **a post-write purge that fired on a live row would delete the graph rows of every successfully extracted memory in the install.** The failure mode of this fix is considerably worse than the leak it closes, so it gets its own test.

## The Low

Fixed — the route now returns 422 on a malformed `memory_id`, matching every sibling in that file. You were right that a 500 was the wrong thing for a caller that lets failures propagate out of a remediation to hear.

## The merge decision worth reviewing

Both drop branches now run the entity purge **and** #1292's child cascade, and **the purge goes first**. `_drop_children` raises once any child fails, so ordering the cascade first would skip this parent's own graph rows on exactly the runs where something had already gone wrong. The purge can't raise, so it never blocks the cascade in return.

**One thing I noticed during the merge and deliberately did not fix here.** #1292's cascade soft-deletes child memories, and those children can have graph rows of their own — the cascade doesn't purge them. That is this same leak, one level down, created by the interaction of the two PRs rather than by either alone. It needs its own tests around the child path and I'd rather it be reviewed as its own change than appended to a PR already five rounds in. Filing it.

Verified at this head: core-api full suite green, core-storage-api 358 passed, ruff check and format clean at CI's scopes, mypy clean on `core-storage-api/src/` and clean on `core-api/src/` apart from the two pre-existing `types-python-dateutil` errors in untouched `common/enrichment/service.py`, ratchet "No new lines.", sentinel "All 35 protected strings survive.", tenant-scope gate exit 0.

Eldad-Caura:
@claude re-review at head `1e1e4213`.

**Your High is a defect in the previous round's own fix, and you found the thing that hid it.**

`_purge_written_artifacts_if_dropped` detected the drop, purged, returned `None` — and then `process_entity_extraction` carried straight on: relation upserts carrying `evidence_memory_id`, subject write-back, contradiction detection, cross-link discovery. It cleaned the link table and immediately refilled the relation table. **The leak moved rather than closed**, and I shipped that as the fix for the leak.

The sharper half of your finding is the fixture, not the code: every test in that file leaves `graph.relations` empty, so the relation loop had nothing to iterate and none of them *could* have caught it. That's the second time on this PR that a green suite was green because the test data avoided the path.

**Fixed.** The helper returns `bool`; the caller returns on `True`. A test with a non-empty `graph.relations` pins it — restoring the fall-through fails it on `upsert_relation` having been awaited. A second test pins the other direction, that a **live** row still gets its relations and cross-links, because a short-circuit that fired unconditionally would silently stop writing them for every extracted memory in the install.

I also split the two failure modes while in there, since they're different states:

- **liveness read fails** → indeterminate → `True`. Carrying on would write graph rows for a row that may already be dropped; stopping costs relations for a row that's probably live, which the next content update rebuilds. Refuse to write what can't be justified.
- **purge fails** → `True`. A purge that didn't run doesn't make the memory live again, so continuing would be strictly worse than the failure.
- only an affirmatively live row returns `False`.

## The Medium, with a correction to its premise

Taken — each cascaded child is now purged after its delete succeeds, never counted as a cascade failure (the helper logs and swallows, matching the parent).

**But the reasoning in the finding is broader than what the code does, and I checked before writing it.** Auto-chunk children go through `sc.create_memories` directly and get **no extraction of their own**. The parent is what gets extracted, over the *full document* — so the names mined from chunked content hang off the **parent**, and its purge already reached them. A child acquires graph rows only when something later rewrites its content, because `update_memory` re-extracts.

So this closes a narrow real case and keeps the invariant true of the cascade whatever populates children in future. It is **not** the broad "children leak PII entities" the wording implies, and I'd rather say so than let the fix look bigger than it is. I'd raised this myself last round in the broad form; verifying it is what corrected me.

Three tests: the cascade purges each child; a child whose **delete failed** is not purged (still live, so its graph rows describe content no policy removed); `keep_private` purges nothing. Probe-confirmed — removing the call fails the first with `a dropped row kept its graph rows: {'m1'}`.

Rebased onto the newer main under it; patch-id verified unchanged (`2a7f20a6…`) before building on it.

Eldad-Caura:
@claude re-review at head `dfab6c5e`.

**Taken — and the suggested form of it would have left a live memory losing its relations.** I have the failing test.

You're right that the invariant belongs where the deleting happens. Both callers check that the memory is dropped, so this wasn't reachable; but the method deletes across three tables and can't be undone, and *"only purge what governance actually dropped"* shouldn't live only in the callers' heads. A stale call, a reordering, or a future caller written from the method name alone would have wiped a live memory's entity graph.

**Where I diverged.** The proposal was to add `deleted_at IS NOT NULL` to the `owned` subquery. That subquery gates the **link** statements only — the relation delete keys on `Relation.tenant_id` and `evidence_memory_id` and never took it. So under that version, purging a live memory leaves its links and entities intact and **deletes its relation**: partial destruction, which is harder to diagnose than either clean outcome and just as unrecoverable.

I wrote the test before trusting that reading, and under your exact suggestion it fails with `relations: 1` where `0` is required.

**What I did instead**: one guard, before anything is deleted — a row with this id, in this tenant, with `deleted_at` not null — and an early return with zero counts otherwise. That covers all three deletes rather than the two the subquery reached, and it reads as the authorisation check it is instead of a predicate threaded through three statements.

**Two things about the tests worth flagging rather than burying.**

The storage tests were creating **live** memories and purging them — a state no caller produces. They now soft-delete the target first via a `_dropped_memory` helper; the rows that must survive (the second asserting memory, the other tenant's) stay live deliberately.

And `test_a_tenant_that_does_not_own_the_memory_deletes_nothing` had to be updated too, for a reason worth naming: left with a live memory it would have kept passing **for the wrong reason** — the new liveness guard rather than the tenant check it exists to pin. Same failure mode as the empty `graph.relations` you caught last round, so I went looking for it this time.

Verified at this head: core-storage-api 359 passed, core-api full suite green, ruff check and format clean, mypy clean on `core-storage-api/src/`, ratchet "No new lines.", sentinel "All 35 protected strings survive.", tenant-scope gate exit 0.

Eldad-Caura:
@claude re-review at head `2c07fbc5`.

**Right again, and it lands on the paragraph that existed to justify the mechanism.**

Round 6 fixed the case where the check returns `True` — stop instead of purge-and-fall-through. It did nothing for the case where it returns `False`. The row is live at that moment, execution proceeds, and the subject write-back, the relation upserts carrying `evidence_memory_id`, and whatever cross-link discovery creates are all written **after** the only liveness check on that path. A drop landing across that stretch runs its own purge against rows that don't exist yet, and nothing revisits them.

And the docstring said *"there is no ordering left in which the rows survive."* True of rows written before the call; false of everything after it. **Same overclaim as last round, one round later, in the paragraph whose whole job was to be precise about this.** That's three times on this PR that my prose has asserted a property the code didn't have — it is reliably the weakest thing I produce here, and it is why I'd rather these rounds keep happening than stop early.

**Fixed by calling it twice.** The first call keeps its early-exit role: it saves the relation upserts and a cross-link discovery pass in the common case where the row was already gone. The second sits after every graph-mutating write, so the rows it can find are all of them — that's the one that actually closes the window. The docstring now says which call does what, rather than claiming the property for the mechanism as a whole.

I did **not** take the "move it later" half of your suggestion, because the early exit is worth keeping — cross-link discovery is a storage round-trip with a vector search, and skipping it for a row that's already gone is free.

**One thing deliberately left uncovered, stated rather than assumed:** contradiction detection. It's spawned via `track_task` so it can write after any check here, but it writes conflict rows rather than graph rows, and it re-checks `deleted_at` itself for exactly this race. Out of scope for H-02 rather than overlooked.

The test drives a drop that only becomes visible after the relation write — live at the pre-write check, live at the post-link check, dropped at the final one — and asserts both that the relation **was** written (that's the premise) and that the purge ran. Removing the final call fails it.

Verified at this head: core-api full suite green, core-storage-api 359 passed, ruff check and format clean, mypy clean apart from the two pre-existing `types-python-dateutil` errors in untouched `common/enrichment/service.py`, ratchet "No new lines.", sentinel "All 35 protected strings survive.", tenant-scope gate exit 0.

Eldad-Caura:
Head `41a8f36a`. Your Medium is fixed. Your Low is not, deliberately. And I have a correction to make about my own earlier comments on this PR.

## The Medium

You were right, and the part I want to acknowledge specifically is the cost I had **not** weighed. My docstring reasoned about the indeterminate case only in terms of losing relations, which a later content update rebuilds. It also skipped the audit-log entry, the contradiction trigger and cross-link discovery, because the first call site returns before all three. An audit record lost to a read timeout is rebuilt by nothing, and the task is fire-and-forget.

The root cause was a boolean carrying a policy it couldn't express — "you must stop" forced the indeterminate case to impersonate one of the two real answers. The helper now returns `live` / `dropped` / `unknown` and the call sites decide, because they legitimately differ:

- the **first** site acts on `dropped` only. It exists to save the relation upserts and a cross-link pass, so it has no business destroying the audit work on a non-answer; a real drop is still caught by the final check, which is the guarantee.
- the **final** site doesn't branch at all — nothing follows it, so `live` and `unknown` are the same instruction and `dropped` has already purged.

A failed *purge* still reports `dropped` (the memory is gone either way); only a failed *read* is `unknown`, and it carries a structured `liveness_check` field so its real frequency is measurable rather than assumed. A test drives a transient read failure at the middle check and asserts the audit entry, the relations and cross-link discovery all still happen and nothing was purged — reverting `unknown` to `dropped` fails it.

## The Low — not fixed, and why

Three writer-routed reads per extraction is real added load on a path that runs for essentially every memory with entities. The routing is deliberate (a replica can't see a just-committed delete), but you're right that nobody has load-tested the volume. That's an operational question — it wants a look at writer read-QPS under bulk import, not a code change decided inside this PR. It's recorded in the commit message as explicitly not done rather than left implied.

## A correction about my own verification claims

Several of my earlier comments on this PR ended with a line like *"ratchet 'No new lines.', sentinel 'All 35 protected strings survive.', tenant-scope gate exit 0."* **Those three gates were being run against the wrong working tree.** My shell's directory resets between commands, and those scripts inspect the current directory rather than their own location — so they were measuring a different worktree, one sitting on the already-merged #1292 branch.

They reported green, and re-run correctly on this branch they are still green (ratchet 0, sentinel 0, tenant-scope gate 0), so nothing bad shipped. But for several rounds those lines were not evidence about this PR, and I stated them as if they were. The bug surfaced only because #1292 merged and that stale tree drifted from main, flipping a false pass into a false failure — which is exactly the way this class of mistake normally stays invisible.

Also rebased onto current main, which the branch had fallen behind.

Verified at this head, from the correct tree: core-api 6132 passed, core-storage-api 359 passed, ruff check and format clean, mypy clean on `core-storage-api/src/` and clean on `core-api/src/` apart from the two pre-existing `types-python-dateutil` errors in untouched `common/enrichment/service.py`, all three naming gates exit 0.

Not requesting another review round on this one — nine is enough, and the remaining item is an operational question rather than a code finding.

Eldad-Caura:
@claude re-review at head `472f12cf`.

**This round reviews a change nothing has seen yet, and the finding is mine rather than a reviewer's.** Round 9's fix left an exit uncovered.

## What was wrong

The check at the end of the `try` block is the leak guarantee for the path that completes. It is unreachable on the path that does not. Anything between the link upsert and it that raises — a relation upsert, the subject write-back, the audit call — jumps straight to `except Exception`, and the entities and links already committed stay behind for a memory that may have been dropped. There is no `finally`; that call sits inside the `try` body like everything else.

Round 9 is what made this reachable rather than theoretical. Falling through on `unknown` was the right call — it stopped a transient read timeout destroying an audit record that nothing rebuilds — but it means the first call site now hands the responsibility forward to a later check, on a path where a later check may never run.

The comment I wrote there called the final check *"the guarantee"*. It is the guarantee for one of the two ways out of the function. That is the third time on this PR that my prose has asserted a property the code did not have, and the second time the overclaim was in the paragraph written to justify the mechanism.

## The fix

The `except` handler now carries the same check, guarded on whether anything was written at all.

**Not a `finally`, and the three early `return`s are why.** The no-entities exit happens before `sc` is bound, so an unguarded `finally` raises `NameError` out of a fire-and-forget task. The already-dropped exit would spend a writer read to learn what it just learned. The `dropped` exit would repeat a purge that had only just run. Those are the common paths, not the rare ones — and since the flag has to gate a `finally` anyway, all `finally` buys is one fewer call site.

**The flag is set BEFORE the upsert await, not after,** and that is load-bearing. "The call raised" is not "nothing was written": a timeout can land on a request storage already committed, and then the rows exist while the caller only ever saw an exception. Setting the flag afterwards would skip the check on exactly that case and leak the rows. Being wrong in the other direction costs one writer read on a call that never landed, on a path that is already failing.

## Tests

Four, and I am explicit about which is which:

- **The reproducer** — a relation upsert raises after the links are committed, and the memory is dropped. Without the fix `get_memory` is awaited twice instead of three times and nothing is purged. Probe-confirmed.
- **The ordering probe** — moving `wrote_graph_rows = True` one line down, below the await, fails it. Probe-confirmed.
- **Two guards on the fix,** which pass either way and say so in their docstrings: a failure that wrote nothing must not pay for a writer read, and a failure before `sc` exists must not raise `NameError` out of the task.

## Also in this push

Two comment corrections, both found by reading my own diff rather than by a reviewer:

- the helper docstring said this function is called **twice**; it is called three times now, and the docstring lists what each call site is for instead;
- a sentence weighing what an unguarded `finally` would cost said *"two of those would spend a writer read"* when one of the three would not get that far — it would raise `NameError` on the unbound `sc`.

And a commit-message repair unrelated to the code: the body had lost its opening line in some earlier amend, so it began mid-sentence on `dropped content are the same leak in another table"`. Restored from the PR description, which still had it.

## Still not done, deliberately

Your round-8 Low — three writer-routed reads per extraction task, fleet-wide, never load-tested — is unchanged, and this push makes it four on the failure path. The routing is deliberate (a replica cannot see a just-committed delete) but the volume is an operational question, not one to settle inside this PR.

## Verification

`ruff check` and `ruff format --check` clean at CI's exact scopes (`core-api/src/`, `tests/`, run separately). mypy clean bar the two pre-existing `types-python-dateutil` errors in untouched `common/enrichment/service.py`. Legacy-name ratchet, do-not-touch sentinel and tenant-scope gate all exit 0, run from inside the worktree after `git add`. Broker OpenAPI baseline up to date (8 operations). Full core-api suite green.

Eldad-Caura:
@claude re-review at head `bb7e3d84`.

**Taken — and the suggested form of it would have reopened H-02. I have the failing test.**

You were right on the observation. The end-of-`try` check runs unconditionally, and there is a path that reaches it having written nothing: `graph.entities` non-empty, every name rejected by the blocklist or `_is_valid_entity`, so `filtered` is empty and the persistence block never runs. That read is waste and it should be gated.

## Why the stated fix leaks

`wrote_graph_rows`, as the finding reads it, means "the persistence block ran". That is not the same as "this memory has graph rows".

Cross-link discovery is gated on `auto_entity_linking_enabled` **alone** — not on `name_to_id`, not on `filtered` — so it runs on exactly the path the finding describes. And storage-side it is not a read:

```python
insert_link_returning = (
    pg_insert(MemoryEntityLink)
    .values(rows)
    .on_conflict_do_nothing(index_elements=["memory_id", "entity_id"])
    ...
)
```

`memory_entity_links` is the table the purge deletes from. So a memory can acquire graph rows without `bulk_upsert_entities` ever being called, and gating the final check on a flag that only tracks the persistence block skips the purge and strands the links discovery just created — for a memory governance dropped. That is this PR's own bug, arriving through the guard added to close it.

Probe-confirmed. Applying the suggestion verbatim:

```
FAILED tests/test_entity_linking_wiring.py::test_cross_link_discovery_alone_still_owes_the_memory_a_liveness_check
```

## What I did instead

Both writers set the flag, each before its own await, and then the final check is gated. That is the efficiency you asked for — a run that wrote nothing *and* had linking disabled no longer pays for a writer read — without the leak. Behaviour is unchanged wherever anything was actually persisted.

The flag's declaration now states what it has to mean for the gate to be safe: **"this memory MAY have graph rows"**, not "the persistence block ran". Gating on the narrower reading is the bug, so the name and the comment both have to carry the wider one.

The helper docstring is corrected in the same pass: it said the `except`-site check was the guarded one. Both trailing checks are guarded now.

## A pattern worth naming

This is the third time on this PR that the observation was worth acting on while the remedy was wrong in the unrecoverable direction — after the relation anti-join scoped by `Relation.tenant_id`, and the `deleted_at` fix that gated only the `owned` subquery. Each was a narrowing that is locally consistent and drops a case the wider code still depends on. Each is now pinned by a test that fails under the suggested version, so the next person to reach for the "obvious" narrowing finds out immediately.

## Verification

Full core-api suite 6137 passed, 0 failed. `ruff check` and `ruff format --check` clean at CI's exact scopes (`core-api/src/`, `tests/`, run separately). mypy clean bar the two pre-existing `types-python-dateutil` errors in untouched `common/enrichment/service.py`. Legacy-name ratchet, do-not-touch sentinel and tenant-scope gate all exit 0, run from inside the worktree after `git add`. Broker OpenAPI baseline up to date (8 operations).

Still unchanged and still deliberate: your round-8 Low on writer-routed read volume. That is an operational question for a load test, not something to settle in this PR.

Eldad-Caura:
@claude re-review at head `4eb5a9df`.

**Taken, in full — and your severity note on the worker site is understated rather than lower.**

You're right that `counts.get(...)` sits outside the `try`, and right about why it matters. I verified the chain rather than taking it on the description: `consumer.handle_memory_enriched` catches `GovernanceCascadeError` and nothing else, so an `AttributeError` out of `_purge_entity_artifacts` reaches the dispatcher, nacks, redelivers, re-runs the whole drop branch and writes a second `critical=True` audit for an already-dropped memory. That is exactly the failure round 2 of this PR fixed, arriving back through a type nobody checked.

The tell you identified is in the client itself:

```python
return await self._post(  # type: ignore[return-value]
    "/memories/purge-entity-artifacts",
    ...
```

`_post` is declared `dict | list`. The `type: ignore` is precisely where "this is a dict" was asserted without evidence, and nothing downstream re-checks it.

## Where I differ

> that call site is inside the worker's top-level catch-all, so the practical risk is lower

True of two of the three call sites. **Not the third.** The one added in round 10 *is* the catch-all — it lives in the `except Exception:` handler — and an exception raised inside an `except` block propagates out of the function rather than being caught by it. There it escapes `process_entity_extraction` entirely and surfaces as an unhandled exception on a fire-and-forget task. The worker test drives that site specifically, not one of the two protected ones.

## What the fix says

Both sites now treat a non-object response as its own state instead of forcing it into one of the two answers they already had — the same mistake as the bool-with-a-policy from round 9, in a different place.

A 2xx means the call succeeded. A body we cannot read means we cannot say what it did. So neither "purged" nor "failed" is a claim worth making: governance logs at ERROR that the purge **cannot be confirmed and the rows may still be listable**, and the worker logs that the counts are unknown while still returning `dropped` — because the memory's state is not in question, only the report of what was cleaned.

## A note on my own test

The governance test failed the first time I ran it, and the reason is worth recording. I wrote `pii={"enabled": True, "disposition": "drop"}` — but the PII branch keys on **`action`**; `disposition` is the non-business key. The wrong key lands silently on `flag`, which never purges, so the test would have exercised a path that never reaches the line under test and passed for no reason at all.

It now also asserts the `pii_drop` audit was emitted, so a config that doesn't reach the purge fails loudly instead of passing vacuously. That is the third instance on this PR of a test being green because the fixture avoided the path, after `graph.relations = []` and the mismatched-tenant test.

Both new tests are probe-confirmed: removing the two guards fails exactly those two and nothing else.

## Verification

Full core-api suite green. `ruff check` and `ruff format --check` clean at CI's exact scopes (`core-api/src/`, `tests/`, run separately). mypy clean bar the two pre-existing `types-python-dateutil` errors in untouched `common/enrichment/service.py`. Legacy-name ratchet, do-not-touch sentinel and tenant-scope gate all exit 0, run from inside the worktree after `git add`. Broker OpenAPI baseline up to date (8 operations).

Still unchanged and still deliberate: the round-8 Low on writer-routed read volume, which is a load-test question rather than a PR one.

Eldad-Caura:
@claude re-review at head `51efc4ff`.

**Taken — your second option, not your first, and the reason is the one this PR has already been burned by twice.**

The gap is real and I'm not disputing it. The purge finds entities *through* the memory's links, so an entity row committed by `bulk_upsert_entities` whose link never landed is reachable by nothing: the purge runs, finds no links, deletes nothing, and returns `{"links": 0, "relations": 0, "entities": 0}`.

Your sharpest point isn't the gap, it's its **shape**. That response is indistinguishable from *"there was nothing to purge"*. Everything else this PR leaves behind, it leaves behind loudly — a failed purge, a failed cascade, an unreadable response all name the memory in an ERROR. This one was the exception, and that's what made it worth changing.

## Why not option 1

Passing the upserted ids to the purge as an extra candidate source re-introduces the exact race the candidate bounding exists to prevent. Round 1 went through this: the first draft deleted every unlinked entity in the tenant, and it was narrowed because that sweeps rows orphaned for unrelated reasons *and* races a concurrent writer that has created an entity but not yet linked it.

The narrower list doesn't escape that. `bulk_upsert_entities` **upserts** — an id it returns may be a row another writer created moments ago and is about to link. Deleting it because our link failed would be over-deletion caused by our own failure, and over-deletion is the direction that doesn't come back. A person handed a concrete list can check what a query cannot.

## What went in

The link upsert gets its own `try/except` that logs the ids and **re-raises**. The re-raise is load-bearing: the `except` handler below still runs its liveness check and still purges everything that *is* reachable, so this adds a diagnostic and changes no behaviour.

**Only entities this run created are named.** `bulk_upsert_entities` reports `action` per row, and an entity that already existed is reachable through whatever linked it before — naming it would send an operator after rows that are nobody's orphans.

The gap itself stays, deliberately, and the code now says so at the place it lives rather than in a commit message nobody will find.

## On the test you asked for

> Add a test ... that upserts an entity via `bulk_upsert_entities` without ever creating a `MemoryEntityLink` row for it, then asserts that `memory_purge_entity_artifacts` either removes it or that the gap is at least surfaced

The "removes it" half is already pinned in the opposite direction by an existing storage test from round 1 — the unbounded candidate set fails it with `2 == 1`. Adding a test that asserts the purge *does* reach unlinked entities would contradict it. So this is the "surfaced" half only: the new test asserts the created id appears in the log, and removing the handler fails it.

## Verification

Full core-api suite 6140 passed, 0 failed. Probe-confirmed. `ruff check` and `ruff format --check` clean at CI's exact scopes (`core-api/src/`, `tests/`, run separately). mypy clean bar the two pre-existing `types-python-dateutil` errors in untouched `common/enrichment/service.py`. Legacy-name ratchet, do-not-touch sentinel and tenant-scope gate all exit 0, run from inside the worktree after `git add`. Broker OpenAPI baseline up to date (8 operations).

Still unchanged and still deliberate: the round-8 Low on writer-routed read volume — a load-test question rather than a PR one.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #1298, opened 2026-09-05
feat(storage): add a repair sweep for un-provenanced embeddings

Eldad-Caura:
## The gap this closes

Rows with `embedding IS NOT NULL AND embedded_content_hash IS NULL` were
reachable by nothing. Every existing repair path — both modes of this CLI, and
core-worker's event-driven task — selects `embedding IS NULL`, and these rows
**have** an embedding. That is why 241 of them sat unrepaired for a week in
2026-09 while three separate repair paths ran over the same table.

#1281 and #1289 stopped the writers producing them. #1294 made them visible.
This is the part that gets them back.

## Re-embed, not stamp

The cheap repair is `SET embedded_content_hash = content_hash`. **It is not
correct**, and the distinction is the whole design.

That statement asserts the vector already on the row describes the row's
current text — which is exactly what nothing here knows, because that is the
state being repaired. It happened to hold for the 241 production rows, each
verified first against a recomputed `sha256(tenant_id || ':' ||
coalesce(fleet_id,'') || ':' || content)`, and it does not hold in general. A
wrong hash reads downstream as verified freshness, which is worse than the NULL
it replaced.

So the sweep re-embeds and stamps the hash read alongside that content — making
the assertion true rather than assuming it. `test_repair_provenance_re_embeds_
rather_than_stamping` seeds a `0.9` vector and asserts a `0.125` one afterwards,
so a stamp-only implementation cannot pass.

## Three scope terms, each load-bearing

- **`content_hash IS NOT NULL`** — an infinite-work guard, not tidiness. The
  repair stamps *from* `content_hash`, so a row without one is stamped NULL
  again and still matches the selector: one provider call per row per run,
  forever, changing nothing. Every other mode is self-terminating because its
  own write flips the row out of its predicate; this one has to be told.
- **`created_at >= PROVENANCE_REQUIRED_FROM`** by default — imported from
  core-storage-api rather than restated, so the sweep and #1294's alert cannot
  disagree about which population is a defect. `--include-legacy` widens to the
  ~95k pre-037 rows behind an explicit flag, a warning and a 5s pause: those are
  undetermined rather than damaged, and one provider call each makes it a spend
  decision rather than a bug fix.
- **Entities skipped** — migration 037 gave `embedded_content_hash` to
  `memories` alone.

## Why no index

Migration 037's partial index carries `AND embedded_content_hash IS NOT NULL`,
so it deliberately **excludes** this population. I did not add one:

- Pagination is `ORDER BY id` with `id > :after`, which walks the primary key
  and filters rather than re-scanning per page — adequate for an operator-run,
  resumable batch job.
- An index for a population that should be *permanently empty* is a cost paid on
  every write forever, in exchange for speed on a job that should find nothing.
- If the legacy sweep is ever run at scale, the index belongs with **that**
  decision, sized against that population, rather than shipped ahead of it.

Happy to add it if you'd rather; it is a migration plus a `CONCURRENTLY` build.

## The mode flags became an enum

Three mutually exclusive modes on two booleans admits a state ("rewrite the hint
prefix AND repair provenance") with no meaning and no branch — the same shape as
the `_Provenance` pair in #1289. The code was already rendering the mode as one
of these exact strings by hand, so `_ScanMode` is what it was half-expressing.
argparse enforces the same exclusivity at the CLI.

## Verification

Each guard reverted independently:

| Revert | Fails |
|---|---|
| Drop `content_hash IS NOT NULL` | the idempotency test |
| Drop the `created_at` bound | the legacy test |
| Replace re-embed with a bare stamp | **10 of 14**, including the one written for it |

Plus an end-to-end dry run against a seeded database, and CLI smoke tests
covering both rejections (`--repair-provenance --rewrite-hint-prefixed`, and
`--include-legacy` without its mode).

core-storage-api **339 → 346**. core-operations 55, unchanged.

**On the root suite** — a note worth having. My stored baseline was captured
hours ago against an older main on a database since deleted, and it is no longer
a comparator: it disagreed with itself twice. I re-ran both arms in one session,
pristine database each, and diffed by name:

```
origin/main : 26 failed, 13 errored
my branch   : 19 failed, 13 errored
FAILURES introduced: NONE      ERRORS introduced: NONE
```

Fewer failures on my branch only because `test_mcp_plan_limit_observation.py`
happens to be unstable and landed differently on the two arms.

Separately, `test_h06_m30_recall_identity.py::test_recall_honours_caller_agent_
id_as_the_visibility_identity` is **flaky**: 1 failure in 5 identical runs, same
code, pristine database each. I twice reached a confident wrong conclusion about
it — first "environmental", then "mine" — before running arms that were actually
comparable. Worth its own issue; it is not touched here.

Not merging.

### DIFF
--- a/core-storage-api/src/core_storage_api/scripts/backfill_embeddings.py
+++ b/core-storage-api/src/core_storage_api/scripts/backfill_embeddings.py
@@ -265,6 +265,16 @@ async def _iter_rows(
     """
     from sqlalchemy import text
 
+    # Deferred like every other import in this module — the CLI keeps its
+    # startup light and avoids importing the service layer it deliberately
+    # does not depend on — but hoisted out of the pagination loop below, which
+    # would otherwise re-resolve it once per page.
+    #
+    # Imported rather than restated: the alert in core-operations counts the
+    # population this sweep repairs, and a second copy of the cutoff would let
+    # the two disagree about which rows are a defect.
+    from core_storage_api.services.postgres_service import PROVENANCE_REQUIRED_FROM
+
     after: uuid.UUID | None = None
     while True:
         params: dict = {"limit": batch_size}
@@ -315,10 +325,6 @@ async def _iter_rows(
                 # predate provenance entirely: repairing them is a provider
                 # spend decision about tens of thousands of rows, not a defect
                 # being cleaned up, so it takes an explicit flag.
-                from core_storage_api.services.postgres_service import (
-                    PROVENANCE_REQUIRED_FROM,
-                )
-
                 sql += "AND created_at >= :provenance_cutoff "
                 params["provenance_cutoff"] = PROVENANCE_REQUIRED_FROM
         else:
@@ -501,7 +507,7 @@ async def run_backfill(
 ) -> list[BackfillReport]:
     """Walk targeted rows and (re-)embed them according to the selected scan mode.
 
-    Two modes:
+    Three modes:
 
     - ``NULL_EMBEDDING`` (default): rows where ``embedding IS NULL``,
       embedded from ``content`` / ``canonical_name``. The
--- a/core-storage-api/src/core_storage_api/scripts/backfill_embeddings.py
+++ b/core-storage-api/src/core_storage_api/scripts/backfill_embeddings.py
@@ -14,7 +14,10 @@ backfill task in core-worker (see ``local_emb_res/specs/C-backfill-task-pr.md``)
 
 Idempotent and restartable: the WHERE clauses filter to rows with NULL
 embeddings, so a partial run can be resumed by simply re-running the
-command — already-embedded rows are skipped naturally.
+command — already-embedded rows are skipped naturally. ``--repair-provenance``
+is idempotent for the same reason: its own write flips
+``embedded_content_hash`` out of the selector. ``--rewrite-hint-prefixed`` is
+the exception, and says so at the point of use.
 
 ⚠ **Single-provider only — does NOT honour per-tenant embedding configs.**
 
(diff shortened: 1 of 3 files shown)


Eldad-Caura:
Both taken, in `99365f70`.

**"Two modes:" — worth saying how that got there**, because the fix is trivial and the cause is not. A `git checkout HEAD -- <path>` on a branch with no commits yet wiped this file mid-change, and the script I wrote to reconstruct it used a narrower anchor than the original edit, so the header alone did not come back while the three bullets did. I have audited the rest of that reconstruction against what it was replacing — usage examples, idempotency note, mode enum, selector, CLI flags all present — and the header was the only casualty. It is also exactly the kind of gap a reviewer catches and a test never will, since nothing asserts on prose.

**Loop import — hoisted.** You are right that it was cheap and right that it was wrong: it sat inside `while True`, so it re-resolved once per page. It stays deferred (this module defers every import — the CLI keeps its startup light and avoids depending on the service layer it deliberately stands apart from), just above the loop now rather than inside it. It is imported rather than restated on purpose: the alert added in #1294 counts the population this sweep repairs, and a second copy of the cutoff would let the two disagree about which rows are a defect.

**`--include-legacy` — leaving the runtime check.** argparse has no native "B requires A" relation; expressing it would mean either a subparser split, which is heavier than the constraint, or a custom action. The runtime check returns 1 with a message naming both flags, which reads better than a generic parser error. Happy to change it if you would rather have the failure at parse time.

Re-verified after the change: core-storage-api 346, the root backfill file 15, ruff and mypy clean, and an end-to-end `--repair-provenance --dry-run` against a seeded database still scans and still skips entities.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #1302, opened 2026-09-05
test(recall): stop the identity fixture drawing its own retrievability

### LINKED ISSUE #1301: Flaky in isolation: recall-identity parity test turns on a random nonce clearing the 0.3 similarity floor
`tests/test_h06_m30_recall_identity.py::test_recall_honours_caller_agent_id_as_the_visibility_identity`
fails **6 out of 10 runs when run on its own**, on unchanged `main`, with a pristine database
each time. It has not failed in CI. The outcome is decided by the `uuid4()` the fixture draws.

Added yesterday in #1276 (`47cc1bfa`, 2026-09-04T19:16Z).

## Root cause

`_setup` (lines 103–118) seeds content and queries a slice of it:

```python
nonce = f"h06-{uuid.uuid4().hex}"                              # the query
content = f"{nonce} peer B private scope_agent memory"          # the row
```

`FakeEmbeddingProvider` (`common/embedding/providers/fake.py`) sums a **per-word** hash vector
and normalises. The query is one word; the row is that word plus five fixed ones. So

```
cos(query, row) = (1 + W·R) / |W + R|
```

where `R` is the fixed five-word sum and `W` is a brand-new random vector on every run. That is a
fresh draw per `uuid4()`, and it is a wide one — measured over 2000 nonces:

| min | p05 | median | p95 | max | σ |
|---|---|---|---|---|---|
| −0.248 | 0.081 | 0.361 | 0.581 | 0.727 | 0.150 |

`MIN_SEARCH_SIMILARITY = 0.3` (`core-api/src/core_api/constants.py:246`), and **35.2% of nonces
land below it**. Those runs fail.

The pytest-captured log from a failing run shows exactly where the row dies — storage finds it,
the relevance floor drops it:

```
"path": "scored-search",  "row_count": 1     ← DB returned the row
search.post_filter_results: 0.0ms
"path": "memory-search",  "row_count": 0     ← floor dropped it
```

## Confirmation: it is the floor, not a coincidence

Four nonces pinned around the floor, run together:

| nonce cosine | result |
|---|---|
| −0.033 | **FAIL** |
| +0.284 | **FAIL** |
| +0.317 | pass |
| +0.660 | pass |

The split falls exactly on `MIN_SEARCH_SIMILARITY`. Repeating the outer two nonces three times
each: −0.033 failed 3/3, +0.660 passed 3/3. A pinned nonce is fully deterministic — this is the
only source of variance.

## Rates, and why CI is green

The failure rate drops sharply as more of the suite runs first:

| condition | failures |
|---|---|
| this test alone | **6/10** |
| whole file | 2/10 (and 1/5 in an earlier round) |
| whole `tests/` suite, locally | 0/2 |
| whole `tests/` suite, in CI | **0/30** |

Those 30 CI runs are runs whose head commit actually contains `47cc1bfa` (branches cut before that
merge ran a suite without the test, so their green says nothing). The one red run in that window
(`feat/chat-base-url-override`) failed on Ruff, not here.

So this is a **local-iteration** problem, not a broken build — but it is not benign. CI uses the
same `EMBEDDING_PROVIDER: fake` and the same database config (`.github/workflows/ci.yml:499–517`),
there is no rerun plugin, and 0/30 is not something the 35% figure predicts (0.65³⁰ ≈ 6×10⁻⁶).
**Something in the full-suite run suppresses it and I have not identified what** — most likely
earlier tests mutating `organization_settings` for tenant `default`, since a tenant-level
`min_similarity` changes both the floor and whether the FTS floor-bypass applies
(`_uses_global_min_similarity`). Worth knowing, because whatever it is, CI's green here currently
depends on it.

## Why this matters

- Anyone iterating on this file gets a red that is not a regression, more often than not.
- Any change to the execution set — a `-k` filter, `pytest-xdist` sharding, splitting the suite —
  moves the rate back toward 35% for a test **guarding a fixed security defect** (H-06 disclosure /
  M-30 silent drop).
- The failure message is *correct*: `fixture produced no searchable rows — test is vacuous`. The
  vacuity guard is doing its job and catching a genuinely broken fixture. **Please don't fix this
  by weakening or deleting that assertion** — the fixture is what is wrong.

## Suggested fix

Make the similarity deterministic instead of drawn. The test is about *visibility identity*, not
about ranking, so it should not have a relevance floor in its path at all. Query the content the
row actually holds:

```python
content = f"{nonce} peer B private scope_agent memory"
# ...seed with `content`, then:
body = {"tenant_id": tenant_id, "query": content, "caller_agent_id": peer}
```

Cosine becomes 1.0 by construction and the nonce still isolates the row from everything else in
the shared `default` tenant. Passing an explicit `min_similarity: 0.0` on the request would also
work, and states the intent outright.

Either way, consider asserting the relationship rather than assuming it —
`tests/test_readme_keyword_recall.py:55` already does this, pinning
`assert 0.0 <= cosine < MIN_SEARCH_SIMILARITY` so the fixture cannot silently drift across the
floor. That is the discipline missing here.

## Scope

Only this test. I checked the siblings that use the same nonce-as-query fixture:

- `tests/test_h05_multi_fleet_read_gate.py` is the same shape but asserts only status codes and
  `not in resp.text`, so the draw cannot flip it.
- The other negative assertions in this file (e.g. line 166) sit behind a 403, so a low draw makes
  them belt-and-braces rather than wrong.
- `tests/test_search_recall_tracked_flag.py` queries the full content, so its cosine is 1.0.

## Reproducing

```bash
for i in $(seq 1 10); do
  pytest 'tests/test_h06_m30_recall_identity.py::test_recall_honours_caller_agent_id_as_the_visibility_identity' -q
done
```

Environment: `EMBEDDING_PROVIDER=fake`, `ENTITY_EXTRACTION_PROVIDER=none`,
`USE_LLM_FOR_MEMORY_CREATION=false`, pristine Postgres 16 + pgvector per run, pytest 9.1.1 —
i.e. CI's own configuration.

Eldad-Caura:
Fixes #1301.

`test_recall_honours_caller_agent_id_as_the_visibility_identity` failed **6 of 10 runs when run
on its own**, on unchanged `main`, pristine database each time. The fixture decided the outcome,
not the route.

## The cause

`_setup` seeded six-word content and queried one word of it:

```python
nonce   = f"h06-{uuid.uuid4().hex}"                      # the query
content = f"{nonce} peer B private scope_agent memory"   # the row
```

`FakeEmbeddingProvider` sums a **per-word** hash vector, so the cosine between a one-word query
and that content is a fresh draw on every `uuid4()`. Over 2000 nonces:

| min | p05 | median | p95 | max | σ |
|---|---|---|---|---|---|
| −0.248 | 0.081 | 0.361 | 0.581 | 0.727 | 0.150 |

`MIN_SEARCH_SIMILARITY = 0.3`, so **35% of draws put the row under the relevance floor**, where
`post_filter_results` dropped it. The captured log shows storage returning it and the floor
taking it away: `scored-search row_count: 1` → `memory-search row_count: 0`.

Confirmed by pinning four nonces around the floor and running them together:

| cosine | −0.033 | +0.284 | +0.317 | +0.660 |
|---|---|---|---|---|
| | **FAIL** | **FAIL** | pass | pass |

The split is exactly `MIN_SEARCH_SIMILARITY`. Repeated, the outer two are 3/3 fail and 3/3 pass —
a pinned nonce is fully deterministic, so the draw is the only variance.

## The fix

Query the whole seeded content. Cosine is 1.0 by construction; the nonce still isolates the row
inside the shared `default` tenant, so nothing these tests assert changes. They are about **who
may read a row** — retrieval should never have been a variable in them.

Every query site moves, not only the failing one. The negative assertion the file carries against
a possible future 200-with-filtering behaviour ("must still not surface the peer's private row")
was vacuous on the same 35% of runs: a row under the floor is absent whether the gate works or
not. The 403 detail carries only agent ids, never the query, so passing the content as the query
cannot satisfy those assertions by echo — verified by running the file.

## Verification

- **0/15** failures on the exact condition that failed 6/10. If the draw were still live, 0.65¹⁵ ≈ 0.16%.
- **12/12** across this file and its H-05 sibling, which uses the same fixture shape.
- **Still catches the defect it guards.** Reverting the `/recall` call site to the pre-fix
  `eff_agent_id = body.filter_agent_id` fails it on the parity claim rather than the vacuity
  guard: `/recall dropped caller_agent_id: same body, different rows. search=['ab69f265-…'] recall=[]`.
  Three H-06 spoof tests go red with it, as they should.
- `ruff check tests/` clean.

## Scope note

This never turned CI red — 0 failures in the 30 CI runs whose head commit contains the test (the
one red in that window failed on Ruff). The rate falls as more of the suite runs first: 6/10
alone, 2/10 for the file, 0/30 for the full suite. **I did not establish what suppresses it in
the full suite** — most likely an earlier test setting a tenant-level `min_similarity` on
`default`, which moves both the floor and the FTS bypass. #1301 records that as open. It is worth
someone's attention separately, because CI's green here currently depends on whatever it is.

### DIFF
--- a/tests/test_h06_m30_recall_identity.py
+++ b/tests/test_h06_m30_recall_identity.py
@@ -101,9 +101,22 @@ async def _write_private(client, headers, tenant_id, *, content, agent_id, fleet
 
 
 async def _setup(client, sc):
-    """Peer B holds a private memory; attacker A is trust-1 in another fleet."""
+    """Peer B holds a private memory; attacker A is trust-1 in another fleet.
+
+    Returns the seeded content as the query, NOT the bare nonce. Under the fake
+    embedding provider a vector is the sum of its words' hashes, so a one-word
+    query against six-word content scores a fresh random cosine on every
+    ``uuid4()`` — spanning -0.25 to 0.73, with 35% of draws below the 0.3
+    relevance floor. That made the row unretrievable at random and the parity
+    test below a coin flip (#1301). Querying the whole content pins the cosine
+    at 1.0; the nonce still isolates the row within the shared tenant.
+
+    These tests are about WHO may read a row, so retrieval must not be the
+    variable. Anything asserting a rank or a floor belongs in a search test.
+    """
     tenant_id, headers = get_test_auth()
     nonce = f"h06-{uuid.uuid4().hex}"
+    content = f"{nonce} peer B private scope_agent memory"
     peer = f"peer-b-{_uid()}"
     attacker = f"agent-a-{_uid()}"
     await _seed_agent(sc, tenant_id, peer, fleet_id=f"fleet-f2-{_uid()}", trust=3)
@@ -112,11 +125,11 @@ async def _setup(client, sc):
         client,
         headers,
         tenant_id,
-        content=f"{nonce} peer B private scope_agent memory",
+        content=content,
         agent_id=peer,
         fleet_id=None,
     )
-    return tenant_id, headers, nonce, peer, attacker
+    return tenant_id, headers, content, peer, attacker
 
 
 # ---------------------------------------------------------------------------
@@ -132,12 +145,12 @@ async def test_recall_refuses_a_filter_agent_id_that_is_not_the_caller(
     Agent A names peer B. Pre-fix that made B the visibility identity, so B's
     ``scope_agent`` content came back to A.
     """
-    tenant_id, _, nonce, peer, attacker = await _setup(client, sc)
+    tenant_id, _, query, peer, attacker = await _setup(client, sc)
     as_auth(tenant_id, attacker)
 
     resp = await client.post(
         "/api/v1/recall",
-        json={"tenant_id": tenant_id, "query": nonce, "filter_agent_id": peer},
+        json={"tenant_id": tenant_id, "query": query, "filter_agent_id": peer},
     )
     assert resp.status_code == 403, (
         f"recall let an agent name a peer; body={resp.text[:400]}"
@@ -157,12 +170,12 @@ async def test_recall_refuses_a_caller_agent_id_that_is_not_the_caller(
     ``filter_agent_id`` would have left the hole open under a different field
     name. Fails without the fix.
     """
-    tenant_id, _, nonce, peer, attacker = await _setup(client, sc)
+    tenant_id, _, query, peer, attacker = await _setup(client, sc)
     as_auth(tenant_id, attacker)
 
     resp = await client.post(
         "/api/v1/recall",
-        json={"tenant_id": tenant_id, "query": nonce, "caller_agent_id": peer},
+        json={"tenant_id": tenant_id, "query": query, "caller_agent_id": peer},
     )
     assert resp.status_code == 403, resp.text
     assert "does not match the authenticated agent identity" in resp.text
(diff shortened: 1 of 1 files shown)


Eldad-Caura:
@erni-a plese approve

erni-a (approved):

PR #1303, opened 2026-09-05
fix(health): give the storage probe a budget bigger than one connect

Eldad-Caura:
## The defect

`PROBE_TIMEOUT_SECONDS = 5.0`. The storage pool's per-attempt connect ceiling is also `5.0`. **Equal**, so `asyncio.wait_for` and the transport raced, and the probe could absorb none of the five connect retries `CONNECT_PHASE_MAX_ATTEMPTS` deliberately grants it — whose own comment says they exist to "ride out a Cloud Run cold start". One slow connection setup and the probe lost by construction.

`5.0 > 5.0 → False` is the whole bug.

## What it did in production

`GET /api/v1/health` returned 503 `storage: unreachable` on **23 of ~3,000 probes (~0.8%)** over ten hours while storage was healthy. At each failure the reader served the very same `/memories/count` call:

```
probe gives up          reader serves the same call
00:08:38.002  TimeoutError    00:08:38.012   200 in 34 ms
00:16:40.775  TimeoutError    00:16:38.121   200 in 28 ms
```

The abandoned request completes because `_cancel_safe` shields it (from the 2026-06-16 pool-leak incident), so both halves are in the logs: the probe calling a dependency unreachable, and that dependency answering ~10ms later.

**It is not a slow-count problem.** `/v1/storage/memories/count` measured p50 28ms, p95 96ms, max 1.658s across 2,000 requests — nothing crossed 5s. The probe passes a truthy tenant (`__health_check__`), so it takes `memory_count_active`'s tenant-scoped branch, not the whole-corpus counter. **Nor is it pool exhaustion** — the call succeeds, immediately.

**This has an external audience.** `Better Stack Better Uptime Bot` polls `prod-memclaw-core-api` directly — found independently while enumerating external referrers of the prod service URLs, and absent from the Datadog section of the rename runbook. So a third-party uptime service has been receiving false alarms from us for an unknown period. That is why this lands on its own rather than batched.

## The change

**`STORAGE_CONNECT_TIMEOUT_SECONDS`** names a ceiling that was a bare literal plus three separate prose copies. It lives in `constants.py` rather than beside the pool because `storage_client` imports `config.settings` at module scope and `config` builds `Settings()` at import — putting it in `storage_client.py` is a circular import (verified, not assumed).

**The ordering is enforced at startup**, in `Settings._validate_timeout_ordering` alongside the orderings already there, so it fails the process rather than only a test run. That matters more than a test: the probe's consumers are operator-facing, and this is the kind of ordering broken by tuning one number in isolation.

**The two I/O probes run concurrently.** Raising a per-dependency budget awaited in series would have taken the worst case from 10s to 24s; gathering keeps it at 12s. The event-bus check is sync (`bus.is_healthy`, no I/O) and moves *above* the gather so a free check no longer waits out a stalled probe. Result-dict key order is unchanged — both callers splat it into their response body, so that order is the response shape.

**Both compose healthchecks go `timeout: 5s → 15s`.** They call this endpoint. Leaving them at 5s would have recreated the identical inverted-budget defect one layer out — the review caught this, and it is the reason the diff touches compose at all.

## On failure semantics

`_probe_dependencies` has exactly two callers, `health` and `status_`, both in `routes/health.py`. Both `await` the complete tuple; neither short-circuits, inspects which probe failed first, or consumes partial results. `health` uses `unhealthy` only as a truthiness test plus a verbatim dump into `unhealthy_dependencies`. Stating that explicitly rather than leaving it implied: **nothing depends on probe ordering.**

## Verification

Tests confirmed **red before the fix**, each against the specific change it guards:

| Test | Without its fix |
|---|---|
| `test_probe_timeout_clears_storage_connect_ceiling` | `AssertionError: assert 5.0 > 5.0` |
| `test_dependency_probes_run_concurrently` | `assert 1 == 2` — probes sequential |
| `_validate_timeout_ordering` (startup) | `ValidationError` — process refuses to boot |

The ordering test asserts the **relation**, never the literal `12.0`: both numbers are legitimately tunable, and the property that must survive a retune is that the probe outlives one attempt. A literal would pass forever while the invariant silently broke.

The concurrency test asserts **overlap** (max probes in flight `== 2`), not elapsed time — a wall-clock threshold measures runner contention, which is the trap `pytest.ini`'s own benchmark note documents. The key-order assertion rides on the concurrency test's existing call rather than paying a second storage round trip to re-check a dict literal.

Gates at CI's exact separate scopes: `ruff check core-api/src/` and `ruff check tests/` clean; `ruff format --check` clean on both; `mypy` clean (203 files).

Full root suite: **6,071 passed**. 18 failures, all pre-existing — verified against a baseline I ran myself on clean `origin/main` in a separate worktree, diffed by sorted test name, with zero failures unique to this branch. Worth recording that the environmental failure count is ~18–19 and its *membership varies between runs*: the two names that differed across my runs both pass in isolation on this branch, and the one unique to a branch run also fails on clean main.

## Deliberately not done

Deriving the probe budget arithmetically from the connect ceiling, and making that ceiling env-overridable to match `OPENAI_HTTPX_CONNECT_TIMEOUT_SECONDS` / `EMBEDDING_HTTPX_CONNECT_TIMEOUT_SECONDS`.

Both siblings sit at **15.0**, each raised from 5.0 after Cloud Run VPC-connector cold connects overran it — the same failure class that produced this flapping, which makes storage the holdout on the old default. Whether it should follow them is a real question and worth a follow-up, but that value governs **every** storage call, not just the probe, and it is not this change's to settle. Flagging rather than silently changing production storage behaviour.

### DIFF
--- a/core-api/src/core_api/clients/storage_client.py
+++ b/core-api/src/core_api/clients/storage_client.py
@@ -18,6 +18,7 @@ from common.storage_auth import is_storage_shared_secret_rejection
 from core_api.clients.identity_token import evict as _evict_id_token
 from core_api.clients.identity_token import fetch_auth_header
 from core_api.config import settings
+from core_api.constants import STORAGE_CONNECT_TIMEOUT_SECONDS
 
 logger = logging.getLogger(__name__)
 
@@ -295,7 +296,7 @@ class CoreStorageClient:
         fallback for the remaining 1%.
         """
         return httpx.AsyncClient(
-            timeout=httpx.Timeout(connect=5.0, read=120.0, write=120.0, pool=5.0),
+            timeout=httpx.Timeout(connect=STORAGE_CONNECT_TIMEOUT_SECONDS, read=120.0, write=120.0, pool=5.0),
             # CAURA-682: pre-fix values 100/50 caused TCP ConnectTimeout
             # retries (3x 5s ~= 13s tail per affected request) during
             # noisy-neighbor write storms — concurrent core-api → storage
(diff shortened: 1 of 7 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #1305, opened 2026-09-05
fix(events): stop the malformed-message drop leaking its payload

Eldad-Caura:
## What I went looking for, and what I found instead

`_decode` ack-drops a message the bus cannot parse. It logged that at ERROR with **no context** — no subscription, no message id — and a malformed payload has no `event_type` or `tenant_id` to identify it by either. So the line said only that *something undecodable arrived somewhere*, across ~60 prod subscriptions spanning two topic families mid-rename. The sibling foreign-env drop branch in the same loop already logs `subscription`, `source_env` and `env`.

Threading the context in surfaced a worse problem on the same line.

## The leak

**Pydantic renders the offending input into `str(exc)` as `input_value=...`** — the *whole decoded document* when a required field is missing — and `structlog_config`'s `format_exc_info` ships the rendered traceback into the `exception` field sent to Cloud Logging and Datadog. So `logger.exception` on the schema-invalid path exfiltrated tenant payload on **every malformed message**.

Verified end to end against the repo's own `Event` model:

```
Field required [type=missing, input_value={'note': 'ssn-123-45-6789'}, input_type=dict]
```

Two leak shapes: a missing required field puts the entire document in `input_value`; a wrong-typed field puts that field's value there. Pydantic truncates the repr at ~50 chars head+tail, so long payloads leak head and tail — a display limit, not a privacy control. Anything short leaks whole.

`json.JSONDecodeError` is clean: `__str__` is only a position, and the document hangs off `.doc`, which is never rendered.

## The change

The two failure modes are now caught **separately**, because only one can carry a traceback safely:

| Arm | `exc_info` | Why |
|---|---|---|
| `ValidationError` | **no** | carries `errors(include_input=False, include_url=False)` — keeps *which field failed and why*, drops the value |
| `ValueError` | yes | `JSONDecodeError` cannot carry the document |

`ValidationError` **must** precede the `ValueError` arm: pydantic ≥2.4 makes it a subclass, so the ordering is what selects the branch at all. The existing comment about that base-class shift across pydantic minors is now load-bearing rather than informational.

Both arms set **`dropped: True`** — the marker twelve other drop sites across `common/`, `core-api` and `core-worker` already use, and which `tests/test_consumer_enriched.py` already asserts on. The one drop whose entire purpose is making a silent discard findable was the one not setting it. Deliberately **not** added to the foreign-env branch: that drop is routine, per-message and logged at INFO, and marking it would swamp any `dropped:true` query.

`subscription`/`message_id` are **required** keyword args — they exist only to be logged, so a default would let a caller drop the context and restore the original blindness. They go in `extra=` rather than `bind_contextvars` (which this repo does configure): contextvars merge at emit time inside `ProcessorFormatter` and never reach the `LogRecord`, so `caplog` cannot see them, and every other drop site in the estate is asserted through the record.

Also hoists `msg = received.message` in the pull loop. proto-plus re-wraps the nested message on **every** `.message` access, so reading it three times allocated three wrappers per message on the hot path (measured +1.72 µs/msg, ~88% of the successful-decode body). The hoist also removes a pre-existing redundant wrap, leaving the path cheaper than before this change.

## Verification

Every test confirmed **red** against the implementation it guards:

| Test | Against | Failure |
|---|---|---|
| `..._never_logs_the_payload` | single combined arm with `exc_info` | `tenant payload reached the log` — shows `{'note': 'ssn-123-45-6789'}` |
| `..._acks_and_logs_real_context` | call site passing `sub_path` | `'projects/.../test-sub' == 'test-sub'` |
| `..._acks_and_logs_real_context` | call site dropping the real id | `'unknown' == 'msg-from-the-wire'` |
| `..._logs_subscription_and_message_id` | signature reverted to `main`'s | `TypeError: unexpected keyword argument` |

**The privacy test asserts against the fully formatted record, not `getMessage()`.** `LogRecord.getMessage()` renders only `msg % args` and structurally excludes `exc_info` — so a test written against it passes while the payload ships. That was this test's first version, and it stayed green against the leaking code. Worth stating because the mistake is invisible: the test *looks* like it checks the payload isn't logged, and checks the one surface the payload could never be on.

The second new test drives a malformed message through `_pull_loop` via the existing `_drive_one_batch` harness, so the **only production call site** is pinned — that it passes the short `subscription_name` rather than `sub_path`, and the message's real id. `_make_received` gained a real `message_id` for the same reason: a `MagicMock` there satisfies any assertion.

Gates at CI's exact separate scopes: `ruff check common/` and `ruff check tests/` clean, `ruff format --check tests/` clean, `mypy --config-file common/mypy.ini common/` clean, legacy-name ratchet and do-not-touch sentinel clean. Full root suite **6,071 passed**; 18 failures, identical set to a baseline I ran on clean `origin/main` in a separate worktree, zero unique to this branch.

## Required follow-up — enterprise CI will go red

`caura-enterprise@dev` carries its own `common/events/pubsub.py` with the identical pre-fix code (bare `logger.exception` at its line 1179), and `scripts/vendored_manual_baselines.json` records the OSS sha it was last reconciled against — currently exactly OSS `main`. `scripts/check_vendored_drift.py` gates on the OSS side holding still, so **this merge breaks that check** until the change is ported and the baseline re-approved (`--approve-manual --only common/events/pubsub.py`).

That is a firing gate with documented remediation, not optional drift — and the enterprise copy has the same payload leak, which is the more important reason to port it promptly. Flagging rather than bundling: it is a separate repo with a separate base branch (`dev`).

### DIFF
--- a/common/events/pubsub.py
+++ b/common/events/pubsub.py
@@ -1106,7 +1106,11 @@ class PubSubEventBus(EventBus):
                 ack_ids: list[str] = []
                 nack_ids: list[str] = []
                 for received in response.received_messages:
-                    foreign_env = self._foreign_source_env(received.message)
+                    # Hoisted: proto-plus re-wraps the nested message on every
+                    # ``.message`` access, so reading it three times costs
+                    # three wrapper allocations per message on the hot path.
+                    msg = received.message
+                    foreign_env = self._foreign_source_env(msg)
                     if foreign_env is not None:
                         # Fan-out copy from a sibling environment sharing
                         # this project's topics. Ack-drop *before* decode
@@ -1127,10 +1131,16 @@ class PubSubEventBus(EventBus):
                             },
                         )
                         continue
-                    event = self._decode(received.message.data)
+                    event = self._decode(
+                        msg.data,
+                        subscription=subscription_name,
+                        message_id=msg.message_id,
+                    )
                     if event is None:
                         # Malformed message — ack so we don't redeliver
-                        # forever, log for alerting.
+                        # forever, log for alerting. The log is ``_decode``'s,
+                        # not this branch's: adding one here would double-count
+                        # the drop in any alert that counts occurrences.
                         ack_ids.append(received.ack_id)
                         continue
                     success = await self._dispatch_all(handlers, event)
@@ -1240,7 +1250,7 @@ class PubSubEventBus(EventBus):
                 continue
 
     @staticmethod
-    def _decode(data: bytes) -> Event | None:
+    def _decode(data: bytes, *, subscription: str, message_id: str) -> Event | None:
         # Pydantic shifted `ValidationError`'s base across v2 minors
         # (v1 + current ≥2.4 inherit from ValueError, 2.0-2.3 did not),
         # so we catch both explicitly rather than assume either. A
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #1307, opened 2026-09-05
fix(events): contract the memory enum so dual_subscribe=False works again

Eldad-Caura:
## What this fixes

`memory` was flipped to the `caura` publish names on 2026-09-01 but its enum members were never contracted. So `publish_name()` returned the caura name while `subscribe_names(dual=False)` still returned the memclaw one, and the two disagreed:

```
                              before      after
unbound_publish_topics(dual=True)    0          0
unbound_publish_topics(dual=False)   4          0
```

`pubsub.py` raises on a non-empty result **at bus construction**, not at `start()`. `dual_subscribe=False` is the DEFAULT, so until now **every standalone and on-prem process could not build a bus at all**. The deployed services run dual-on, which is why this never showed up in staging or prod.

Measured by `exec`-ing both versions of the module in one process rather than reasoning about it.

## What else it does

Contracting stops subscribers binding the legacy spelling — `renamed()` becomes the identity for these members, so `subscribe_names(dual=True)` returns **one** name instead of two. That begins the 86,400 s reap of the legacy memory subscriptions, which is the precondition for deleting them later: *flip publishers → contract the enum → let the pool reap → only then delete.*

## Decisions worth reviewing

**`memory` stays in `FLIPPED_FAMILIES`.** `renamed()` is idempotent so `publish_name` is unchanged either way; `family()` still yields `"memory"` so `_validate_flipped_families()` does not raise; and no guard is keyed on the set being empty. It comes out with the final contract sweep, once the legacy topics are deleted. Removing it here would be a second, independent behavioural change riding along in a contraction.

**`events_manifest.json` is regenerated, not hand-edited.** It previously listed *both* spellings per consumed topic, because it is derived from the same dual-subscribe resolution. Contracting collapses each pair to one.

**`test_topics.py` keeps all four assertions** and moves them to the new wire value. The legacy-name ratchet reports 4 exempt lines removed and asks what they stood for: they pinned the enum's current wire value, that value is now the caura one, and every assertion survives with it. `frozen_topic()` is deliberately kept for `audit`, which has **not** flipped — the helper pins the legacy namespace independently of the enum, so using it for a contracted family would assert the wrong thing.

**The core-worker consumer tests and the pubsub decode tests stay on the memclaw spelling on purpose.** They pin the wire format of messages that already exist: anything published before this lands still carries the legacy `event_type` and must still decode. That is not stale coverage, it is the only coverage for in-flight data.

## Pre-flight

Every `caura` twin was confirmed to exist first — 16 subscriptions and 16 topics across staging and prod — because after contraction subscribers bind **only** the caura name, and a subscription that does not exist is a permanent `NotFound` that halts the pull loop and reds the health endpoint.

## Gates

`ruff check` clean on `common/`, `tests/`, `core-api/src/`; `ruff format --check` clean on `tests/` and `core-api/src/`; do-not-touch sentinel intact (35/35); legacy-name ratchet **−9 net, 0 added**.

The full `tests/test_events/` suite needs CI's Postgres service container — a session-scoped fixture requires it, and the local Postgres belongs to other work. The enum, manifest and every changed assertion were verified by direct execution instead; CI is the real run.

## Ordering

`caura-enterprise` vendors this file and records its sha256 in `scripts/vendored_manual_baselines.json`. Its CI checks out OSS at floating `main`, so **merge this first**, then the enterprise PR — which already carries the post-merge digest and whose drift gate is red until this lands.

### DIFF
--- a/common/events/events_manifest.json
+++ b/common/events/events_manifest.json
@@ -8,9 +8,7 @@
       "caura.lifecycle.forge-distill-requested",
       "caura.lifecycle.insights-requested",
       "caura.memory.embedded",
-      "caura.memory.enriched",
-      "memclaw.memory.embedded",
-      "memclaw.memory.enriched"
+      "caura.memory.enriched"
     ],
     "core-worker": [
       "caura.lifecycle.archive-expired-requested",
@@ -18,9 +16,7 @@
       "caura.lifecycle.embed-backfill-requested",
       "caura.lifecycle.purge-soft-deleted-requested",
       "caura.memory.embed-requested",
-      "caura.memory.enrich-requested",
-      "memclaw.memory.embed-requested",
-      "memclaw.memory.enrich-requested"
+      "caura.memory.enrich-requested"
     ]
   }
 }
--- a/common/events/topics.py
+++ b/common/events/topics.py
@@ -18,10 +18,10 @@ import enum
 
 
 class Memory(enum.StrEnum):
-    EMBED_REQUESTED = "memclaw.memory.embed-requested"
-    EMBEDDED = "memclaw.memory.embedded"
-    ENRICH_REQUESTED = "memclaw.memory.enrich-requested"
-    ENRICHED = "memclaw.memory.enriched"
+    EMBED_REQUESTED = "caura.memory.embed-requested"
+    EMBEDDED = "caura.memory.embedded"
+    ENRICH_REQUESTED = "caura.memory.enrich-requested"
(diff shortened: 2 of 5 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (dismissed):

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #1308, opened 2026-09-05
fix(startup): reject blank and whitespace-padded production secrets

### LINKED ISSUE #1300: fix(startup): reject blank JWT_SECRET values in production
## Problem

`_validate_startup_settings` applies inconsistent validation to two required production secrets in `core-api/src/core_api/app.py`.

- `SETTINGS_ENCRYPTION_KEY` is checked with a falsiness guard at lines 142-146.
- `JWT_SECRET` is checked only for equality with `change-me-in-production` at lines 147-162.

As a result, `JWT_SECRET=""` and whitespace-only values pass production startup validation when the other settings are valid. That permits production to boot with an empty or trivially guessable token-signing secret.

The existing coverage in `tests/test_production_startup_guards.py` checks the literal default and a wrapped literal default, but does not cover empty or whitespace-only JWT values.

## Acceptance criteria

- Reject unset, empty, and whitespace-only `JWT_SECRET` values in production, in addition to the known default placeholder.
- Preserve the existing unwrap behavior for secret-wrapper values.
- Preserve non-production behavior.
- Add parameterized regression coverage for empty and whitespace-only JWT values.
- Confirm the new regression test fails against the current implementation before applying the fix.

## Context

PR #1299 deliberately documents `JWT_SECRET=change-me-in-production`, which the current guard rejects loudly. PR #1242 instead proposed a blank value and exposed this bypass; that PR is superseded by #1299 and should not be used to fix the main-branch validator.

Eldad-Caura:
Closes #1300.

## The defect

`_validate_startup_settings` guarded two required production secrets ten lines apart by different rules — `SETTINGS_ENCRYPTION_KEY` for presence, `JWT_SECRET` only for equality with `change-me-in-production`. So `JWT_SECRET=""` and `JWT_SECRET="   "` were not equal to the placeholder, passed, and **production booted signing API tokens with an empty secret**. The asymmetry between two adjacent guards is the whole bug.

The presence check now runs for **every** entry in `_DANGEROUS_DEFAULTS`, not for `jwt_secret` alone. That mapping moved to module scope so a test parametrized over it covers whatever secret is added next, rather than leaving the following entry to be found the way this one was.

## Two things this turned up

Both verified booting a production settings object before the change.

**1. The placeholder itself still booted, with any surrounding whitespace.** Presence stripped; equality did not:

```python
if not val or not str(val).strip():   # strips
    raise ...
if val == bad_val:                    # compares the RAW value
```

So `JWT_SECRET="change-me-in-production\n"` passed both checks and production signed tokens with the published default. A trailing newline is not contrived — it is what a file-mounted secret yields (`cat key >> .env`, a k8s `stringData` block scalar), which is precisely the delivery path a left-in-place placeholder arrives by. Both properties now test the stripped value.

**2. The same defect class survived in three sibling guards** — including the one this fix's own reasoning cited as the good example. `settings_encryption_key`, `admin_api_key` and the gateway/compat-key perimeter were bare truthiness tests, so `"   "` passed each. `SETTINGS_ENCRYPTION_KEY` was checked for *presence*, not for *blankness*. All now share one predicate, which is what makes the claim true rather than merely asserted.

For the perimeter pair the outcome was **fail-closed, not a bypass**: HTTP strips optional whitespace from header values, so a whitespace gateway secret leaves the check nominally on while the gateway's own injected header arrives empty and every header-trust request 401s. An outage, not an impersonation hole — but refusing to boot is the louder failure that block already prefers to a silent per-request one.

## The unwrap ordering, corrected

`_blank_secret` unwraps before testing, and that ordering is load-bearing **for whitespace rather than for emptiness**:

| value | `bool(...)` | `str(...)` | naive check on the wrapper |
|---|---|---|---|
| `SecretStr("")` | `False` (`__len__`) | `""` | refused anyway |
| `SecretStr("   ")` | `True` (len 3) | `'**********'` | **boots** |

An earlier draft of this change asserted in a comment that `SecretStr("")` was truthy. That was **wrong** — and it mattered, because it meant the `id="empty"` test case passed even with the unwrap moved after the check. Only the whitespace case pins the ordering, and the prose now says so.

`str.strip()` removes only `str.isspace()` characters, so a secret of U+200B or a UTF-8 BOM still reads as non-blank. Left as-is deliberately and documented: it is the convention every guard here shares, and narrowing it in one place would recreate the asymmetry above.

## Verification

Per the issue's own acceptance criterion, the regression test was confirmed red **before** the fix — and for the right reason each time, `DID NOT RAISE RuntimeError`, a boot that should have been refused rather than an import error or an unrelated raise from the same function:

| Reverted | Failures |
|---|---|
| presence check removed (`main`'s behaviour) | 6 |
| equality back on the raw value | 3 |
| unwrap moved after the presence check | 1 — the whitespace case only |
| `settings_encryption_key` back to truthiness | 1 |
| `admin_api_key` back to truthiness | 1 |

End-to-end after the fix: `""`, `"   "`, `"\t\n"`, unset, `SecretStr("")`, `SecretStr("   ")`, `"change-me-in-production\n"` and `" change-me-in-production"` all refused; a legitimate secret still boots; `development` and `sandbox` still boot with a blank one.

Gates at CI's exact separate scopes: `ruff check core-api/src/` and `ruff check tests/` clean, `ruff format --check` clean on both, `mypy` clean (203 files), legacy-name ratchet and do-not-touch sentinel clean. Full root suite **6,071 passed**; 18 failures, identical to a baseline run on clean `origin/main` in a separate worktree, zero unique to this branch.

The one new legacy-name line carries `legacy-name-ok: compat alias field` — it reads the compat alias the perimeter check has always consulted, and the ratchet counts it as exempt rather than new.

## Note

`#1242` proposed a blank `JWT_SECRET` in `.env.example` and would have walked into this. It is superseded by `#1299`, which is merged and documents the literal placeholder so the guard fires loudly — the documentation was already right and only the validator was wrong. Nothing here touches `.env.example`.

### DIFF
--- a/core-api/src/core_api/app.py
+++ b/core-api/src/core_api/app.py
@@ -117,6 +117,40 @@ class SecurityHeadersMiddleware:
         await self.app(scope, receive, send_with_headers)
 
 
+# Secrets that ship with a known placeholder and MUST be replaced in production.
+# Module scope rather than a local, so the property is testable: a test
+# parametrized over this mapping covers whatever is added next, instead of
+# re-asserting presence for jwt_secret alone and leaving the following entry to
+# be found the way this one was.
+_DANGEROUS_DEFAULTS = {
+    "jwt_secret": "change-me-in-production",
+}
+
+
+def _unwrap_secret(value):  # type: ignore[no-untyped-def]
+    """Return a ``SecretStr``-style wrapper's value, or ``value`` unchanged."""
+    return value.get_secret_value() if hasattr(value, "get_secret_value") else value
+
+
+def _blank_secret(value) -> bool:  # type: ignore[no-untyped-def]
+    """True when a secret is missing, empty, or only whitespace.
+
+    Unwraps FIRST, and that ordering is load-bearing for whitespace rather than
+    for emptiness: ``SecretStr("")`` is falsy (``__len__`` is the secret's
+    length), but ``SecretStr("   ")`` has length 3 and ``str()`` of it is the
+    mask ``'**********'`` — so BOTH halves of this predicate pass on the
+    wrapper and a whitespace secret sails through. ``SecretStr("x") == "x"`` is
+    False for the same reason, which bypassed the equality check below once
+    already.
+
+    Note ``str.strip()`` removes only ``str.isspace()`` characters, so a secret
+    of U+200B or a UTF-8 BOM still reads as non-blank. Left as-is deliberately:
+    it is the convention every guard here shares, and narrowing it in one place
+    would recreate the asymmetry this function has already been bitten by.
+    """
+    return not str(_unwrap_secret(value) or "").strip()
+
+
 def _validate_startup_settings(app_settings) -> None:  # type: ignore[no-untyped-def]
     """Refuse to boot when a required safety control is missing.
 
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #1313, opened 2026-09-06
fix(crystallizer): stop the dedup sweep re-clustering rows it archived itself

Eldad-Caura:
Closes audit finding **M-61**.

Crystallization merges a cluster into one fact and archives the sources. The crystal stays >=0.95 similar to those sources — being that similar is what made them a cluster — and it is written with `last_dedup_checked_at` NULL, so the next sweep picks it up as a candidate.

Neither of the two queries behind that sweep filtered on status. Both took `deleted_at IS NULL` and nothing else, and soft-deletion is not the state that matters here: the rows this sweep must stop revisiting are precisely the ones it archived itself.

So run 2 found the ARCHIVED sources as neighbours of the crystal, re-formed `{F,S1,S2}`, re-sent it to the LLM, and then archived F — the archive step takes every cluster member and does not care that one of them is the crystal produced an hour earlier. `create_memory` then 409s on the re-merge (same `crystallizer` agent, near-verbatim content), so `new_ids` is empty and nothing replaces what was just archived.

## Both queries carry the filter

A pair is (candidate, neighbour), and an archived row excluded from one end still reaches a cluster through the other. Filtering the neighbour query alone would have left the loop open through the candidate side. Each filter is probe-confirmed separately: removing the candidate one fails the candidate test, removing the neighbour one fails the neighbour test.

Safe to filter in the query rather than at the call site because neither query has any caller outside this sweep — the crystallizer is the only production consumer of `/memories/near-duplicate-candidates` and `/memories/neighbors-by-embedding`. That is **not** true of bulk-get, which `entity_service` also uses to read evidence rows and which legitimately needs archived ones, so the third guard below lives in the caller instead.

## Why `LIVE_MEMORY_STATUSES` rather than `!= 'archived'`

`outdated` and `conflicted` are no better as merge inputs, and naming the live set means a status added later is excluded by default rather than silently admitted.

A test pins the other direction too. A crystal is written `confirmed`, so filtering to the literal `active` would have stopped the sweep seeing crystals at all — a different bug, in the opposite direction, that neither exclusion test would have caught.

## The third change is a race guard, not part of the fix

Cluster members are re-checked for live status immediately before the LLM call and the archive. The pairs are computed at the top of the run — `_check_near_duplicates` runs in the checks loop, with the health, usage and issue passes between it and crystallization — so a row archived by anything else inside that window is still named in a cluster.

Clusters themselves are disjoint (union-find over the pairs), so this is **not** reachable from one cluster archiving another's members; the window is external writes only. It earns its place by making the invariant checkable next to the destructive step instead of resting on two remote queries staying right.

## On the reported severity

MEDIUM only because a second refuter defeated the HIGH "data loss" premise: archived rows remain fully recallable, since the recall paths exclude only `outdated`/`conflicted`. That refutation is correct and this change does not alter it. Whether archived *should* be recallable is a separate open question and is deliberately untouched here.

## Tests

Four in core-storage-api against real Postgres, because the finding is entirely about which rows two SQL predicates reach and a stub would assert the code calls itself. One in core-api for the race guard, asserting on the archive batch **and** on what was handed to the LLM, since re-merging content that was already merged away is what produces the churn.

## Verification

- Full core-api suite green (6141 passed).
- core-storage-api suite green (363 passed), on its own scratch database.
- `ruff check` and `ruff format --check` clean at CI's exact scopes, run separately.
- mypy: `core-storage-api` clean; `core-api` clean bar two pre-existing `types-python-dateutil` errors in untouched `common/enrichment/service.py`.
- Legacy-name ratchet, do-not-touch sentinel and tenant-scope gate all exit 0.
- Broker OpenAPI baseline up to date (8 operations).

### One flaky test, worth flagging rather than burying

`tests/test_unknown_field_rejection.py::test_memory_update_rejects_unknown_field` failed on one full-suite run and passed on an identical re-run. It is **not** related to this change: the 409 it fails on comes from the create-path semantic-duplicate check in `check_semantic_duplicate.py`, and the two queries this PR modifies have no caller outside the crystallizer.

The underlying cause is that `tests/conftest.py` purges only tenants matching `SWEEP_TENANT_PREFIX` on session teardown, so every other tenant's rows persist in the shared test database between runs. With `EMBEDDING_PROVIDER=fake`, an accumulated live row makes a later test's setup `POST /memories` a near-duplicate. Open PR #1140 addresses the symptom in that one file; the teardown scope is the root and deserves its own issue.

### DIFF
--- a/core-api/src/core_api/services/crystallizer_service.py
+++ b/core-api/src/core_api/services/crystallizer_service.py
@@ -10,6 +10,7 @@ import httpx
 from fastapi import HTTPException
 from sqlalchemy.exc import SQLAlchemyError
 
+from common.constants import LIVE_MEMORY_STATUSES
 from common.enrichment.constants import (
     CLASSIFIER_DEPRECATED_MEMORY_TYPES,
     DEFAULT_MEMORY_TYPE,
@@ -578,7 +579,25 @@ async def _run_crystallization(
 
     # Process each cluster
     for cluster_ids in selected_clusters:
-        cluster_memories = [memories_by_id[mid] for mid in cluster_ids if mid in memories_by_id]
+        # M-61. Live rows only, checked again HERE and not only in the two
+        # queries that built the pairs. The pairs were computed at the top of
+        # this run — ``_check_near_duplicates`` runs in the checks loop, with the
+        # health, usage and issue passes in between — so a row archived by
+        # anything else in that window is still named in a cluster, and the
+        # archive step below is unconditional and irreversible in the direction
+        # that matters. ``bulk_get_memories`` cannot carry the filter itself:
+        # ``entity_service`` reads evidence rows through it and legitimately
+        # needs archived ones.
+        #
+        # This is a narrow race guard, not the fix — the two query filters are
+        # what close the reported loop. It earns its place by making the
+        # invariant checkable next to the delete instead of depending on two
+        # remote queries staying right.
+        cluster_memories = [
+            memories_by_id[mid]
+            for mid in cluster_ids
+            if mid in memories_by_id and memories_by_id[mid].get("status") in LIVE_MEMORY_STATUSES
+        ]
         if len(cluster_memories) < CRYSTALLIZER_MIN_CLUSTER_SIZE:
             continue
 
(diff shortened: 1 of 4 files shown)


Eldad-Caura:
@erni-a pleae approve

erni-a (approved):

PR #1314, opened 2026-09-06
fix(storage): scope a relation's ENDPOINTS to the tenant, not just the edge

Eldad-Caura:
Closes audit finding **M-64**. Cross-tenant disclosure.

`Relation` carries a `tenant_id`, which is what made this easy to miss: the edge had a predicate, so the queries looked scoped. But that column describes the **edge**, not the two entities it names, and the entity FKs only require those rows to exist in *some* tenant.

So a write key for tenant T could post:

```json
{"tenant_id": "T", "from_entity_id": "<own>", "to_entity_id": "<a victim UUID in tenant U>"}
```

…then `GET /entities/<own>/relations` and read U's `canonical_name` and `attributes` straight out of the response. Upstream core-api only calls `enforce_tenant` on the body and forwards the raw ids, so storage is the sole line of defence and it had none.

## Three surfaces, fixed and tested separately

They fail independently, so each has its own test and its own probe.

| | where | what |
|---|---|---|
| 1 | `relation_add` | refuses an endpoint the tenant does not own — stops **new** straddling rows |
| 2 | `relation_get_outgoing` | carries `Entity.tenant_id` on the join — closes the rows **already there** |
| 3 | `memory_get_detail` | same disclosure via `memory_entity_links` — **not in the finding** |

Surface 3 was found by auditing every `join(Entity` in the service rather than only the two locations the finding cites. The write guard cannot reach history, and any caller reaching storage directly bypasses core-api entirely — which is why fixing the write side alone would have left a live leak.

**`entity_expand_graph` is not affected and is deliberately untouched.** It scopes `Relation.tenant_id` in both directions and returns only ids and hop weights, and the downstream entity fetch is tenant-scoped. It can traverse toward a foreign id and waste work; it discloses nothing. Stated explicitly because it is the obvious fourth suspect and should not be "fixed" speculatively.

## Reusing the existing machinery

`_owned_entities` is split out of `_owned_link_endpoints` so *"an entity this tenant owns"* has one definition — relations need the same test with no memory side, and a second inline copy is how the two drift apart.

The refusal message is uniform across *"no such entity"* and *"that entity is another tenant's"*. This service authenticates no request, so a distinguishable refusal is an existence oracle over the whole entity id space — the same reason `_LINK_REJECTED` is one message for both (GHSA-wgvw-28pq-jc36).

`POST /entities/relations` needed a `ValueError → 409` mapping it did not have. **The pre-fix behaviour was worse than a missing guard:** a *foreign* id returned `200`, and a *nonexistent* one returned `500` from an uncaught `ForeignKeyViolationError`. So the endpoint both leaked data and confirmed the existence of any UUID a caller cared to guess. Both are one `409` now, and a test pins the two responses as byte-identical.

## Behaviour change worth calling out

The predicate on `relation_get_outgoing` is an **inner** join, so a pre-existing straddling relation disappears from that endpoint rather than returning with a hollow target. Every row it hides is an edge into another tenant's data.

Surface 3 stays an **outer** join, so those entries keep `entity_id` and `role` and lose only the fields they should never have carried.

## Tests

Eight, against real Postgres. Each guard is probe-confirmed alone — reverting the write check fails the three write tests; reverting either read predicate fails exactly its own test and nothing else. The read probe printed the exploit verbatim: `"canonical_name":"Victim-Confidential"`, `"attributes":{"secret":"victim-only"}` in the attacker tenant's response body.

Straddling rows are inserted with raw SQL on purpose. Both write paths refuse them now, so creating them through the API would test the guard instead of the reader and every assertion would hold with the bug reintroduced — the same failure mode `test_memory_entity_links_tenant_scope._links` documents.

Three over-refusal guards cover the opposite error, since a check that refused everything would pass every attack test above.

## Verification

- core-storage-api suite: **371 passed** (363 + the 8 new).
- Full core-api suite: **6141 passed**.
- `ruff check` and `ruff format --check` clean at CI's exact scopes, run separately.
- mypy `core-storage-api`: clean, 86 files.
- Legacy-name ratchet, do-not-touch sentinel, tenant-scope gate: all exit 0.
- Broker OpenAPI baseline up to date (8 operations).

One disclosure about process: the sandbox was restarted between verification and commit, so the suites above were run before the restart and could not be re-run after it (Docker was down). The commit content is byte-identical to what was verified — it was restored from git's content-addressed index, not retyped — and lint, format, mypy and all three gates were re-run post-recovery and are green. CI re-runs everything regardless.

### DIFF
--- a/core-storage-api/src/core_storage_api/routers/entities.py
+++ b/core-storage-api/src/core_storage_api/routers/entities.py
@@ -346,7 +346,15 @@ async def get_full_graph(
 @router.post("/relations")
 async def create_relation(request: Request) -> dict:
     body: dict = await request.json()
-    relation = await _svc.relation_add(body)
+    try:
+        relation = await _svc.relation_add(body)
+    except ValueError as e:
+        # M-64. Without this the service's endpoint-ownership refusal would
+        # surface as a 500 — indistinguishable from storage being broken, which
+        # is the same conflation ``entity_add`` and ``create_memory_entity_link``
+        # already resolved as 409. A caller naming an entity it does not own is
+        # a client error.
+        raise HTTPException(status_code=409, detail=str(e))
     return orm_to_dict(relation, RELATION_FIELDS)
 
 
--- a/core-storage-api/src/core_storage_api/services/postgres_service.py
+++ b/core-storage-api/src/core_storage_api/services/postgres_service.py
@@ -371,6 +371,11 @@ _ENTITY_UPDATABLE_FIELDS = frozenset({"canonical_name", "entity_type", "attribut
 # wire — two copies could drift apart and reintroduce the existence oracle one
 # word at a time. The true cause is logged, never returned.
 _LINK_REJECTED = "entity link rejected: memory_id or entity_id does not exist, or the link already exists"
+# M-64. One answer for "no such entity" and "that entity is another tenant's",
+# for the reason ``_LINK_REJECTED`` already carries: this service authenticates
+# no request, so a distinguishable refusal turns the route into an existence
+# oracle over the whole entity id space (GHSA-wgvw-28pq-jc36).
+_RELATION_REJECTED = "relation rejected: from_entity_id or to_entity_id does not exist"
 
 
 # Migration 037 added ``embedded_content_hash`` on 2026-08-16, and every writer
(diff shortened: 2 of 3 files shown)


Eldad-Caura:
@erni-a plae apppveo

erni-a (approved):

PR #1317, opened 2026-09-06
fix(extraction): define what "subject" means, so updates can supersede (A66)

arkash20:
## Updates to person-valued facts could never retire the value they replaced

**Root cause was one undefined word.** `EXTRACTION_PROMPT` listed:

```
- role: one of subject, object, mentioned
```

…and never said what a *subject* **is**. So for `"The X primary oncall is Marta"`, marking either `marta` or `x primary oncall` as the subject is defensible — and the model picked differently on different writes.

Everything downstream then behaved **correctly on unstable input**:

- the extraction worker sets `subject_entity_id` only when exactly **one** entity claims `role=subject`;
- `_subjects_differ_with_certainty` drops candidates whose subjects differ — in **8 ms**, before the judge.

So each new value became a new subject, no candidate survived to the judge, and the old value stayed `active`.

### The class this breaks is wide

Any predicate whose object is a named entity — **who is oncall, who owns a service, who leads a team**. Those facts change often, and none of them could be updated.

## Evidence

Same three-write chain, live stack, before and after:

```
BEFORE   run 1  resolved
         run 2  fired NOTHING — three oncall assignments, all `active`
         run 3  partial — two different values both `active`
         -> 2/3 runs ended with contradictory claims co-active

AFTER    3/3 resolved:  A conflicted | B conflicted(sup=A) | C active(sup=B)
         -> 0/3
```

Mechanism confirmed directly — the three oncall rows previously carried **three different subject entities**:

```
before   'marta' (person) | '...primary oncall' (identifier) | 'wei' (person)
after    one shared subject entity across all three
```

## The fix

The prompt now defines the subject as the entity the statement is **about** (the grammatical subject), states that a named value stays the **object**, says the subject must not change when the value changes, and carries two worked examples plus *why* the rule exists — so it doesn't get trimmed as verbose later.

## Scope — please read before approving

This changes extraction for **every write**, not just contradiction candidates.

The evidence is a live before/after on the failing scenario, because that is the only thing that can demonstrate a model-behaviour change. **A unit test asserting my own prompt text would prove nothing**, so the tests here pin the *contract* (they fail loudly if the definition is dropped again — its absence is exactly what made this invisible) and the behavioural claim rests on the repro.

**A bench A/B is still warranted to confirm no wider regression, and is not claimed here.** Happy to run one before merge if you'd rather.

## Testing

5 contract tests · repro at `benchmark/a66_subject_stability.py` · full suite **5988 passed** with the 8 failures that pre-exist on main (`test_capability_usage` / `test_request_observation`).

Refs: canonical A66.

### DIFF
--- a/benchmark/a66_subject_stability.py
+++ b/benchmark/a66_subject_stability.py
@@ -1,0 +1,69 @@
+#!/usr/bin/env python3
+"""A66 — is the N>2 chain defect real, or an artifact of the first repro?
+
+The original repro slept only 5s between the FIRST and SECOND write (55s
+thereafter), so write B's detection could run before write A had finished
+embedding. That alone could produce "no candidates found". This version waits
+the SAME settle after EVERY write, so any remaining failure is the product's.
+"""
+
+import json
+import time
+import urllib.error
+import urllib.request
+
+BASE, KEY = "http://localhost:8000", "dev-admin-key"
+SETTLE = 60
+
+
+def call(m, p, b=None):
+    r = urllib.request.Request(
+        BASE + p,
+        method=m,
+        headers={"X-API-Key": KEY, "Content-Type": "application/json"},
+        data=json.dumps(b).encode() if b is not None else None,
+    )
+    try:
+        with urllib.request.urlopen(r, timeout=180) as x:
+            return x.status, json.loads(x.read().decode(), strict=False)
+    except urllib.error.HTTPError as e:
+        return e.code, {"raw": e.read().decode()[:150]}
+
+
+def w(t, c):
+    return call(
+        "POST", "/api/v1/memories", {"tenant_id": t, "agent_id": "a66", "content": c}
+    )[1]
+
+
+def st(t, i):
+    s, m = call("GET", f"/api/v1/memories/{i}?tenant_id={t}")
+    return (m.get("status"), m.get("supersedes_id")) if s == 200 else (f"<{s}>", None)
+
+
+subjects = [
+    ("deploy window", ["01:00 UTC", "03:00 UTC", "05:00 UTC"]),
+    ("primary oncall", ["Marta", "Tomas", "Wei"]),
+    ("build agent pool size", ["4 workers", "8 workers", "16 workers"]),
+]
+bad = 0
+for n, (subj, vals) in enumerate(subjects):
+    salt = f"{int(time.time())}{n}"[-7:]
+    T, ids = f"a66r{salt}", []
+    for v in vals:
+        ids.append(w(T, f"The {salt} {subj} is {v}.").get("id"))
+        time.sleep(SETTLE)  # EQUAL settle after every write
+    rows = [(lbl, *st(T, i)) for lbl, i in zip("ABC", ids)]
+    print(f"\n--- run {n + 1}: {subj}  (settle={SETTLE}s after every write)")
+    for lbl, s, sup in rows:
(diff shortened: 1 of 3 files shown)


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

arkash20:
Rebased onto main. CI green (14 checks).

arkash20:
Rebased onto main. CI green (14 checks).

Bench A/B run since opening this (67-question regression sample, fast mode d=0, same stack both arms — only the extraction prompt differs):

| category | arm A (main) | arm B (A66) |
|---|---|---|
| single-session-user | 100.0% | 92.3% |
| single-session-preference | 91.7% | 91.7% |
| multi-session | 100.0% | 90.9% |
| temporal-reasoning | 64.3% | 50.0% |
| knowledge-update | 85.7% | **92.9%** |

**avg_recall is identical in both arms (1.0000 / 0.9846)** — retrieval is unchanged; only answer accuracy moved.

The overall −4.7pp is 3 questions out of 64, and a repeat of arm B on the IDENTICAL build moved single-session-preference 91.7% → 83.3% on its own. That is the harness's noise floor (±1 question/category, consistent with the answer-LLM flake documented during A34). The one category that improved is knowledge-update — the supersession behaviour this PR fixes.

Note the absolute numbers are not comparable to `baseline.json`: this stack runs Vertex embeddings, the golden baseline was captured on OpenAI embeddings. Arm-to-arm only.

@Eldad-Caura please approve

Eldad-Caura (approved):

arkash20:
Rebased onto main. CI green.

## Bench A/B — 67-question regression sample, fast mode `-d 0`

Same stack for every run; the **only** difference is the extraction prompt. Each arm run **twice**, because the first A-vs-B comparison looked like a regression and needed a control.

| category | A run 1 | A run 2 | B run 1 | B run 2 |
|---|---|---|---|---|
| single-session-user | 100.0% | **84.6%** | 92.3% | 92.3% |
| single-session-preference | 91.7% | 91.7% | 91.7% | 83.3% |
| multi-session | 100.0% | 90.9% | 90.9% | 90.9% |
| temporal-reasoning | 64.3% | **50.0%** | 50.0% | 42.9% |
| knowledge-update | 85.7% | 92.9% | 92.9% | 92.9% |
| **total correct** | **56/64** | **52/64** | **53/64** | **51/64** |

**A: 52–56. B: 51–53. Overlapping.** Arm A moves 4 questions between two identical builds, and A's second run scores *lower* on single-session-user than either A66 run while landing on exactly the temporal-reasoning number (50.0%) that first looked like this PR's worst regression. The arms are indistinguishable at this sample size.

`avg_recall` is **identical across all four runs** (1.0000 / 0.9846) — retrieval never moved; only the stochastic answer model did.

What is *not* noise, because it is deterministic and reproduced:
- N>2 chains resolve **1/3 → 3/3**
- the three rows that carried three different subject entities now share **one**
- knowledge-update sits at ceiling in both A66 runs and varies in main's

### Caveat on the absolute numbers
This stack runs **Vertex `gemini-embedding-001`**; the golden `baseline.json` was captured on **OpenAI** embeddings. Different vector space — these are valid **arm-to-arm only** and must not be compared to the stored baseline.

@Eldad-Caura please approve

Eldad-Caura (approved):

PR #1318, opened 2026-09-06
ci(npm): repoint the credential check at a manifest that exists

Eldad-Caura:
## What broke

`verify-npm-credential.yml` read `./clients/npm-caura/package.json` to learn the second package a release publishes. #943 moved that directory, #1099 moved it again, and #1244 deleted it.

The workflow is `workflow_dispatch`-only, so nothing caught it. The reference was broken from **2026-08-24** and surfaced on **2026-09-06**, when it was run to answer a live question — "can `NPM_TOKEN` publish a release?" — and died with `Cannot find module` before reaching the answer. [Run 34021589371](https://github.com/caura-ai/caura/actions/runs/34021589371).

It is **not** repointed at that manifest's descendant, because there is none. `@caura/sdk` (`clients/npm-sdk`, added by #950) is the other half of a release today — a fact about releases, not a rename. Worth knowing if this ever looks wrong later.

## Three comments asserting things that are no longer true

The file reasoned carefully about a hazard that cannot happen, so the reasoning is corrected rather than deleted:

| Claim | Reality |
|---|---|
| "the unscoped `caura` name is unclaimed, so ANY valid account can take it" — a bad token could 403 on the scoped client while claiming the bare name | #943 established npm **refuses** it: `403 Forbidden - PUT .../caura, Package name too similar to existing package csurf`. Blocked by the typosquat filter, not merely unclaimed. The verdict message repeated the same wrong reasoning to whoever hit the failure. |
| "cannot drift from what the publish workflows actually ship" | True of a **rename**, false of the **move** that happened — reading the name from a manifest trades a hard-coded name for a hard-coded path, and nothing checked the path. |
| "Same trap as the dependency check in `publish-npm-caura.yml`" | Also deleted by #943. Now points at `publish-npm-sdk.yml`, which carries the same `\|\| true`-under-`pipefail` trap. |

The replacement hazard is the real one: a release is two independently tagged workflows, so a token that can publish one and not the other produces a **partial release** — the two packages on different versions, with a tag already public.

## The test, and the false green in its own first draft

`tests/test_workflow_require_paths.py` covers the class rather than the instance.

**Its first draft was itself nearly vacuous, and this is the part worth reviewing.** It resolved each path against the repo root *plus* every declared `working-directory`. A `package.json` exists at the repo root, so every `require('./package.json')` resolved there regardless of which directory its job actually ran in — **nine of the eleven requires in this repo could not have failed**, and the only two genuinely under test were the two being fixed. A test shaped exactly around the line that broke.

A declared `working-directory` **replaces** the root rather than adding to it. With that corrected, mutating `clients/npm-sdk` is now caught in `publish-npm-sdk.yml` too — the tag-triggered release workflow, which matters more than the one this started from:

```
publish-npm-sdk.yml: require('./package.json') — not found under clients/npm-sdk
verify-npm-credential.yml: require('./clients/npm-sdk/package.json') — not found under <repo root>
```

Confirmed failing against the unfixed workflow, and passing with it.

## Scope this deliberately does not cover

`require()` is one of several ways a workflow names a repo path. `working-directory:`, `cd` in a `run:` block, script arguments and `on: paths:` filters are all unchecked, and a stale `paths:` filter fails just as silently as this did. Worth a follow-up; not folded in here.

## Checks

- `ruff check tests/` and `ruff format --check tests/` — CI's exact scopes, run separately
- `actionlint` on the changed workflow — clean
- Test verified failing without the fix, and by mutation against a second workflow

### DIFF
--- a/.github/workflows/verify-npm-credential.yml
+++ b/.github/workflows/verify-npm-credential.yml
@@ -13,12 +13,22 @@ name: Verify the npm credential
 # So a token minted for one is not automatically good for the other. A token
 # belonging to `caura-ai-admin` publishes the broker fine and gets a 403 on
 # the `@caura` scope unless that account is also a member with publish rights.
-# The failure surfaces only at the publish step, after a tag has been pushed,
-# and the two halves of a release can then disagree: the unscoped `caura` name
-# is unclaimed, so ANY valid account can take it, while the scoped client needs
-# the right one. Claiming the bare name under one account while the scoped
-# package lives under another is a mess to unpick afterwards, and npm does not
-# let you simply move it.
+# The failure surfaces only at the publish step, after a tag has been pushed.
+# A release is two independently tagged workflows — publish-npm-client.yml for
+# `@caura/client`, publish-npm-sdk.yml for `@caura/sdk` — so a token that can
+# publish one and not the other yields a PARTIAL release: one package on the
+# new version, the other stranded on the old one, with the tag already public.
+#
+# This paragraph previously described a worse hazard: that a bad token could
+# 403 on the scoped client while still claiming the unscoped `caura` name under
+# the wrong account. That cannot happen, and the correction is worth keeping so
+# nobody restores the reasoning. #943 established that npm REFUSES the unscoped
+# name outright — `403 Forbidden - PUT .../caura, Package name too similar to
+# existing package csurf` — so it is blocked by the typosquat filter rather than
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@erni-a pleaes approve

erni-a (approved):

PR #1320, opened 2026-09-06
fix(llm): unwrap singleton JSON arrays instead of discarding valid answers

ran-taig:
## Problem

Post-cutover prod monitoring (2026-09-03→06 on `gemini-3.1-flash-lite`) surfaced a new response quirk: the model sometimes wraps a **perfectly valid response object in a one-element JSON array**:

```
enrichment-primary attempt 1/2 failed (VertexResponseShapeError: Vertex returned a JSON list where a dict was expected. Response content (truncated): '[\n  {\n    "memory_type": "fact",\n    "weight": 1.0,\n    "title": "Rule: Never auto-merge PRs", ...
```

The CAURA-651 shape guard — written when list responses were garbage — rejects these, degrading the call to the fake keyword fallback. Over three days that discarded **503 valid answers** (419 enrichment, 75 crystallizer, 7 extraction, 2 contradiction; ~3.8% of the window's 11,115 worker enrichments, ~170/day).

## Fix

`complete_json` in both Gemini-backed providers (Vertex + Developer API) unwraps a list of length 1 whose only element is a dict, with an INFO log marking the unwrap. **Everything else still raises**: empty lists, multi-element lists, scalar-element lists, bare scalars — the CAURA-651 guard's purpose (no bare `AttributeError` downstream, no silent fallback) is unchanged for genuine garbage.

The OpenAI-compatible provider is deliberately left strict: its `response_format=json_object` contract guarantees an object, no cases were observed there, and a list would be a real contract violation worth surfacing.

## Testing

- New: singleton-`[dict]` unwraps (both providers); `[dict, dict]`, `[1]`, and `[]` still raise the typed ShapeError.
- The CAURA-651 suite (`test_vertex_response_shape.py`) is unaffected by design — its list cases are multi-element/scalar and keep raising.
- Full affected set green: 207 passed (truncation, shape, gemini, platform-providers, protocols, extraction-partial-parse). Ruff clean.

## Expected impact

The ~170/day enrichment fake-fallbacks on prod drop to ~0; the new `returned a singleton JSON array; unwrapped` INFO line becomes the counter for how often the quirk fires.

### DIFF
--- a/common/llm/providers/_json_shape.py
+++ b/common/llm/providers/_json_shape.py
@@ -1,0 +1,44 @@
+"""Shared JSON-shape coercion for ``complete_json`` responses.
+
+gemini-3.1-flash-lite (prod, 2026-09-03→06) sometimes wraps a valid
+response object in a one-element array — ~3.8% of enrichments arrived
+as ``[ {..valid enrichment..} ]`` and the CAURA-651 shape guard was
+discarding good answers into the fake fallback (~170/day). A singleton
+list holding one dict is unambiguous: unwrap it. Every other non-dict
+shape (empty or multi-element lists, scalar elements, bare scalars) is
+still garbage and stays with the caller's ShapeError.
+
+Shared between the Gemini-backed providers (``vertex.py`` ADC mode,
+``gemini.py`` API-key mode) — same rationale as ``_truncation.py``: one
+implementation, so the heuristic cannot silently drift between them.
+
+Deliberately schema-agnostic: this layer serves every ``complete_json``
+consumer (enrichment, extraction, dedup, contradiction, …) and knows
+nothing about their expected keys. A wrong-but-dict payload that slips
+through lands in the caller's own field validation (e.g.
+``_validate_enrichment``) exactly as an un-wrapped wrong dict would —
+the guard here only exists to stop bare non-dict shapes from surfacing
+as ``AttributeError`` (CAURA-651).
+"""
+
+from __future__ import annotations
+
+import logging
+
+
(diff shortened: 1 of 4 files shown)


ran-taig:
@claude

ran-taig:
Review addressed in `fb8c36d5`:

**Extraction → shared helper: done.** The unwrap now lives in `common/llm/providers/_json_shape.py::unwrap_singleton_array(parsed, *, provider, model, log)`, called by both Gemini-backed providers — same single-implementation rationale as `_truncation.py`. The helper only *coerces*; shape **rejection** stays in each provider so the typed per-provider ShapeError (and the monitoring attributes CAURA-651 added) is preserved. 211 tests green, ruff clean.

**Schema-key verification: considered, deliberately not added** (flagged non-blocking in the review). This layer is schema-agnostic by design — `complete_json` serves enrichment, extraction, dedup, contradiction, insights, etc., and has no knowledge of any caller's expected keys (`response_schema` is accepted-and-ignored on these providers per the C1 contract). Threading caller schemas down here would couple every consumer to the provider layer for marginal benefit: a wrong-but-dict payload that slips through the unwrap lands in the caller's own field validation (e.g. `_validate_enrichment`) exactly as an un-wrapped wrong dict does today — same degradation path, no new masking. The helper's docstring records this reasoning so it doesn't get re-litigated blind.

ran-taig:
@claude

ran-taig:
@erni-a

erni-a (approved):

PR #1321, opened 2026-09-06
fix(fleet): block agent-scoped keys from DELETE /fleet/{fleet_id}

Eldad-Caura:
Closes audit finding **M-25**. Broken function-level authorization (BFLA).

`DELETE /fleet/{fleet_id}` ran `enforce_read_only` and `enforce_tenant`, and nothing else. Its sibling `POST /fleet/{fleet_id}/purge` runs those two **plus** `enforce_not_agent_credential`, and states why in its own docstring — a fleet purge is an admin-plane operation. Delete carried the same blast radius minus memories, and none of the guard.

```
delete_fleet   enforce_read_only + enforce_tenant
purge_fleet    enforce_read_only + enforce_tenant + enforce_not_agent_credential
```

`fleet_id` is not bound to the caller's scope and `enforce_tenant` only compares tenants, so an agent-scoped credential — scoped to one agent in one fleet, never to the tenant — could delete **any** fleet in its tenant.

Traced through storage rather than assumed: `fleet_delete` removes every node row for `(tenant_id, fleet_id)` and every command row belonging to those nodes, unfiltered by status — so completed command history goes too. Memories survive, which is the only thing making this narrower than purge.

Verified live on `main` before touching anything, and no open PR touched `fleet.py`.

## Tests

Both assert on **state**, not only the status code. Without the guard the call doesn't merely return 2xx — it returns **204 and the fleet is actually gone**. A status-only assertion would still pass against a version that deleted first and refused afterwards, which is exactly the trap [#1316](https://github.com/caura-ai/caura/pull/1316) had to be written around.

Confirmed: removing the guard line fails `test_agent_credential_cannot_delete_a_fleet`. The over-refusal guard is there because blocking every caller would satisfy the first test.

## Three things `/simplify` caught in my own first draft

All in the docstring — worth listing, because prose asserting more than the code supports is the defect class this queue keeps turning up, and I produced three instances of it while fixing one.

1. **I wrote that `enforce_not_agent_credential` "names fleet operations as admin-plane in its own docstring". It does not.** Its list reads *"Agent management (trust level, fleet, deletion)"*, and the very next sentence disambiguates that as an agent *"relocat[ing] its fleet"* — the agent's fleet **assignment**, not fleet-entity lifecycle. The route that actually matches is `PATCH /agents/{id}/fleet`. I was citing a source for something it doesn't say. Now cites the sibling purge call site, which is real.
2. **"an agent-scoped key — which belongs to a single node" is wrong.** The credential carries tenant + agent; an `Agent` row has no node reference, and one node reports many agents per heartbeat. Corrected to one agent in one fleet.
3. **"the pending command queue" understated it** — the delete isn't filtered by status.

Also from that pass: the docstring was 13 lines for a one-line guard, most of it restating purge's docstring, the helper's docstring, and its own first line. Trimmed to the two things not visible from the code. And the state check now scopes `GET /fleet/nodes` by `fleet_id` rather than listing the whole tenant and matching a node name.

## Not fixed here — `POST /fleet/commands` needs its own finding

Same missing guard, **materially higher impact**, and I don't think it should ride along on this PR.

It takes a caller-supplied `command` string, `payload`, and target `node_id`; node ids are enumerable via `GET /fleet/nodes` with the same credential; the command is delivered in a heartbeat response and executed by the plugin, where `deploy` / `update_plugin` fetch source and restart the process.

Signature verification does **not** close it by default: `plugin/src/validation.ts` accepts an unsigned command unless `CAURA_REQUIRE_SIGNED_COMMANDS` is set, which defaults to false, and core-api never signs what it creates.

Two reasons it can't just be appended here: unlike `delete_fleet` there's no sibling already carrying the guard to argue from, and the route is listed in `docs/public-api-stability.md`, so a third-party integrator on an agent-scoped credential may legitimately be calling it. That judgement needs its own evidence.

## On depth

I checked whether the right fix was deeper — a decorator, a router-level dependency, or a `tenant_scope_gate.py`-style ratchet. It isn't, for this change. There's no shared path `delete_fleet` went through that `purge_fleet` didn't; both are flat handlers with hand-listed gates, and the defect is a missing line rather than a broken abstraction. A blanket router-level default-deny would 403 `POST /fleet/heartbeat` and `POST /fleet/commands/{id}/result` — the plugin's only two fleet calls — and take every node offline. What *could* be centralized is the inventory, not the enforcement: a ratchet requiring every mutating core-api route to either call the guard or carry a named allowlist entry with a reason. That's its own change.

## Verification

- Full core-api suite: **6150 passed**, 5 skipped, 1 xfailed.
- New test confirmed to fail without the guard (204 + fleet deleted).
- `ruff check` / `ruff format --check` at CI's scopes, mypy (2 pre-existing `types-python-dateutil` errors in untouched `common/enrichment/service.py`).
- Legacy-name ratchet and do-not-touch sentinel, run after `git add`.
- Broker OpenAPI baseline up to date.

### DIFF
--- a/core-api/src/core_api/routes/fleet.py
+++ b/core-api/src/core_api/routes/fleet.py
@@ -317,9 +317,22 @@ async def delete_fleet(
     tenant_id: str = Query(...),
     auth: AuthContext = Depends(get_auth_context),
 ):
-    """Delete a fleet and all its nodes. Memories are NOT deleted (they retain fleet_id for history)."""
+    """Delete a fleet and all its nodes. Memories are NOT deleted (they retain fleet_id for history).
+
+    Auth: a write-capable tenant-owner key. Agent-scoped credentials are
+    blocked (BFLA) — fleet lifecycle is admin-plane, the same call the sibling
+    ``POST /fleet/{fleet_id}/purge`` makes.
+
+    Why it is needed rather than implied: ``fleet_id`` is not bound to the
+    caller's scope, and ``enforce_tenant`` only compares tenants. So without
+    this an agent-scoped key — scoped to one agent in one fleet, never to the
+    tenant — could delete ANY fleet in its tenant, taking every node row and
+    every command row belonging to those nodes (completed history included,
+    since the delete is not filtered by status).
+    """
     auth.enforce_read_only()
     auth.enforce_tenant(tenant_id)
+    auth.enforce_not_agent_credential("delete a fleet")
 
     sc = get_storage_client()
 
--- a/tests/test_route_authz_gaps.py
+++ b/tests/test_route_authz_gaps.py
@@ -912,3 +912,72 @@ async def test_heartbeat_delivers_commands_to_a_writing_credential(client, as_au
     delivered = [c for c in resp.json()["commands"] if c["id"] == command_id]
     assert delivered, f"command {command_id} was not delivered: {resp.text}"
     assert delivered[0]["command"] == "ping"
+
+
+# ---------------------------------------------------------------------------
+# M-25: DELETE /fleet/{fleet_id} accepted an agent-scoped credential.
+#
+# ``enforce_not_agent_credential`` names fleet operations as admin-plane in its
+# own docstring, and the sibling ``POST /fleet/{fleet_id}/purge`` calls it. This
+# route carried the policy everywhere except the line enforcing it.
+#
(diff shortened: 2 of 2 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #1322, opened 2026-09-06
fix(memory): sanitize caller metadata on the bulk and update write paths

Eldad-Caura:
Fixes M-48 and M-52 from the OSS audit. Both verified against current source before touching anything.

## The gap

C25 strips forgeable platform-only keys from caller metadata in `create_memory`, under a comment calling it *"the single entry chokepoint"*. **It was not one.**

| write surface | reaches storage via | sanitized before |
|---|---|---|
| `POST /memories`, MCP single | `create_memory` | ✅ yes |
| `POST /memories/bulk`, MCP batch | `create_memories_bulk`'s own governance/embed/enrich/write loop | ❌ **no** (M-48) |
| `PATCH /memories/{id}`, MCP update | `metadata_` / `metadata_patch` | ❌ **no** (M-52) |

**M-52 is the sharper one.** Merge mode reaches storage as `metadata_patch`, and the SQL is a single top-level `||`:

```sql
SET metadata = COALESCE(metadata::jsonb, '{}'::jsonb) || (:patch)::jsonb
```

The patch is the right operand, so a caller's key wins over the platform's. `contains_pii` / `pii_types` written by the governance gate at creation could be flipped or cleared afterwards **by the same credential that wrote the memory** — rewriting governance output on an existing row, not just minting fake telemetry on a new one.

## Why it stayed invisible

`sanitize_caller_metadata` has thorough unit tests and they always passed, because the helper was never the broken part. Every test either called it directly or fed it input it had already sanitized. **Nothing tested whether a write path reaches it.**

So the new tests drive the REST routes instead — a helper-level test cannot answer that question. This is the general lesson worth keeping: *a helper unit-tested in isolation says nothing about whether its call sites reach it.*

## Ordering

Sanitation goes **before** the governance gate on both paths. The gate calls `mark_pii_flagged` on the same dict and enrichment adds telemetry after it, so sanitizing later would strip the platform's own flags along with the caller's forgeries — the reason `create_memory` already puts it there.

## A regression this fix created, and how it's handled

Worth reading even if you skip the rest.

Treating bulk item metadata as caller input means **a platform key placed in that dict is stripped like any forgery — including by trusted internal callers.** Ingest was doing exactly that:

```python
"memory_type_agent_set": False,   # CAURA-703: type came from the extraction LLM
```

The new sanitation deleted it, and the `setdefault` below then inferred `True`, because `suggested_type: str` is non-optional. **Every ingested memory would have silently recorded that the calling agent chose its type.** CAURA-703 reverted, deterministically, with nothing failing — the field has no in-repo consumer.

The existing test cannot catch it: `tests/test_ingest_commit.py` monkeypatches `create_memories_bulk` and asserts on the item *ingest built*, so the real function never runs.

Rather than exempt internal callers, **ingest stops smuggling platform state through the caller-metadata channel.** `create_memories_bulk` takes a `memory_type_is_agent_set` argument — a function parameter, so unlike a schema field it is not reachable from a request body. The flag is now assigned rather than `setdefault`-ed, since sanitation guarantees nothing is left to defer to.

I considered moving sanitation to the route/MCP boundary instead, which is where trust actually changes and would match the `SERVER_RESERVED_MEMORY_TYPES` precedent. I didn't, because it is fail-open: a new route added later is unprotected by default, whereas the service-level call sanitizes everything and requires an explicit opt-out. Happy to revisit if you prefer the boundary.

## Verification

- Full core-api suite: **6159 passed**, 0 failed
- `ruff check` / `ruff format --check` at CI's scopes; `mypy` clean (2 pre-existing `dateutil` errors in an untouched file)
- Legacy-name ratchet and do-not-touch sentinel pass after `git add`

Every new test confirmed to fail without its own fix, checked one at a time:

```
revert BULK sanitize (M-48)        -> test_bulk_write_sanitizes_caller_metadata
revert UPDATE sanitize (M-52)      -> test_update_sanitizes_caller_metadata[replace]
                                      test_update_sanitizes_caller_metadata[merge]
revert the bulk ARGUMENT handling  -> test_bulk_honours_memory_type_is_agent_set
revert the INGEST side             -> test_ingest_marks_memory_type_not_agent_set
```

The over-refusal half is asserted too: the caller's own keys and the caller-ownable `summary` must survive, and a control test proves the inferred `memory_type_agent_set` still applies when no trusted caller overrides it.

## Not fixed here — needs its own finding

`PLATFORM_ONLY_KEYS` is **incomplete**. These are written by platform code but absent from the registry, so they stay forgeable on all three paths — including the create path this PR leaves as it was:

- `governance_llm_uncertain`, `nonbusiness_kept_private` — governance output, the same class M-52 is about
- `near_duplicate_of`, `near_duplicate_similarity` — documented as read back by callers
- `parent_memory_id` — forging it fabricates lineage
- `dedup_*`, `near_dup_*`, `triple_emission_ms`, `retrieval_hint`, `auto_chunked`, `child_count`

None mirror into `_system`, so `extract_system_metadata` can't distinguish a forged one from a real one either. Widening the registry is a behaviour change for existing callers, so it belongs in its own PR rather than riding along here.

Also worth its own issue: three separate test files monkeypatch `create_memories_bulk`, which is why a live provenance regression sat in this diff until it was looked for specifically.

### DIFF
--- a/core-api/src/core_api/services/governance_gate.py
+++ b/core-api/src/core_api/services/governance_gate.py
@@ -109,8 +109,9 @@ def mark_pii_flagged(metadata: dict, findings: list[Finding]) -> None:
 
     C25: routed through ``set_system_value`` so the flags land in the
     ``_system`` namespace too (legacy top-level keys kept for one release).
-    Caller forgeries of these keys are stripped at the ``create_memory``
-    chokepoint before this gate runs.
+    Caller forgeries of these keys are stripped by ``sanitize_caller_metadata``
+    on whichever write path reached this gate — each one applies it before
+    calling here, which is what makes the flags written below trustworthy.
     """
     from core_api.services.system_metadata import set_system_value
 
--- a/core-api/src/core_api/services/ingest_service.py
+++ b/core-api/src/core_api/services/ingest_service.py
@@ -1046,12 +1046,15 @@ async def ingest_commit(request: IngestCommitRequest) -> dict:
             # supplied request URL wins, else the fact's own source_uri
             # from preview, else "text-input".
             effective_source = request_url_override or fact.source_uri or "text-input"
+            # CAURA-703 used to be stamped here as
+            # ``metadata["memory_type_agent_set"] = False``. It moved to the
+            # ``memory_type_is_agent_set`` argument on the bulk call below,
+            # because C25 sanitation now strips PLATFORM_ONLY_KEYS from item
+            # metadata — this dict is caller-adjacent, and a platform flag put
+            # in it is indistinguishable from a forged one.
             metadata: dict = {
                 "source": "ingest",
                 "ingest_url": request_url_override or fact.source_uri or None,
-                # CAURA-703: ingest types come from the extraction LLM's
-                # suggested_type, not the calling agent — mark as not agent-set.
-                "memory_type_agent_set": False,
             }
             if request.doc_hash:
                 metadata["doc_hash"] = request.doc_hash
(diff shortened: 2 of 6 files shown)


Eldad-Caura:
@erni-a pleae apprvoe

erni-a (approved):

PR #1324, opened 2026-09-06
fix(memory): reserve the platform-written keys C25 left forgeable

Eldad-Caura:
Follow-up to #1322, from the gap I flagged in its description.

#1322 fixed **where** caller metadata is sanitized — all three write surfaces now call `sanitize_caller_metadata`. This fixes **what it strips**.

## The gap

`PLATFORM_ONLY_KEYS` listed twelve keys. The platform writes considerably more than twelve, and **every key it wrote without reserving was accepted verbatim from callers** — on all three surfaces, including the create path C25 always covered.

The registry is the only thing standing between a caller and a forged platform value, so a key missing from it isn't a documentation gap.

| newly reserved | written by | why it matters |
|---|---|---|
| `governance_llm_uncertain`, `nonbusiness_kept_private` | `GovernanceDecision` | same class as `contains_pii` — asserts a compliance decision the gate never made |
| `near_duplicate_of`, `near_duplicate_similarity`, `near_dup_skipped_reason` | `DetectNearDuplicate` | `near_duplicate_of` is **read back by callers** to decide whether to merge or undo |
| `dedup_skipped_reason`, `dedup_candidate_similarity`, `dedup_judge_confidence`, `dedup_subject_preflight` | `CheckSemanticDuplicate` | forged dedup outcomes |
| `near_dup_check_ms`, `dedup_judge_ms`, `triple_emission_ms` | the above + `EmitMemoryTriple` | fake telemetry |
| `parent_memory_id`, `auto_chunked`, `child_count` | auto-chunk / atomic-fact fan-out | fabricates row lineage |
| `retrieval_hint` | enrichment | fake enrichment output |

`near_duplicate_of` is the one I'd call worse than telemetry: it's documented as something callers read to make a decision, so forging it **redirects a decision** rather than just misreporting one.

## What I deliberately did not reserve

**`source`.** The platform writes it (`"auto_chunk"`, `"atomic_fact_fanout"`) but so does ingest, in caller-adjacent item metadata. Reserving it would strip that stamp — exactly the way widening this boundary nearly cost CAURA-703 in #1322. It's a shared key, not a platform-only one. There's a test pinning that with the reason, so the next person widening this registry has to read it before overriding.

I checked every platform writer of the newly-reserved keys sits **downstream of a sanitize point**: pipeline steps run on post-sanitize metadata, and child rows go to `sc.create_memory` with a dict built from scratch. So reserving them strips forgeries without touching a platform write. That check is the one #1322 taught me to run.

## Read-side consequence — deliberate, but call it out

`PLATFORM_KEYS = PLATFORM_ONLY_KEYS | CALLER_OWNABLE_KEYS` feeds `extract_system_metadata`, so these keys now also appear in **`MemoryOut.system_metadata`** where present. Additive, and consistent with that field's contract: it's the platform view and these are platform values.

One caveat worth stating rather than discovering later — for rows written *before* this change a caller could have forged one of these, and it will now be surfaced as platform metadata. That's inherent to the one-release legacy dual-write and already applies to the original twelve; it isn't new behaviour, just newly reaching more keys.

## Verification

- Full core-api suite: **6190 passed**, 0 failed (rebased onto `main` incl. #1316 and #1322)
- `ruff check` / `ruff format --check` at CI's scopes; mypy clean bar 2 pre-existing `dateutil` errors in an untouched file
- Ratchet and sentinel pass after `git add`

Teeth, checked one at a time:

```
restore the old 12-key registry ->
    test_platform_keys_outside_the_registry_are_not_forgeable[create]
    test_platform_keys_outside_the_registry_are_not_forgeable[bulk]
    test_platform_keys_outside_the_registry_are_not_forgeable[update]
make ingest stamp a newly-reserved key ->
    test_ingest_item_metadata_holds_no_reserved_platform_key
```

Two things about the tests worth noting:

**`create` is covered by a route test for the first time.** I flagged in #1322 that no test drove the sanitizer through the create *route* — every one called the helper directly, which is precisely how M-48 and M-52 stayed invisible. It's in the parametrize list now, and it failed against the old registry like the other two.

**The assertions are on forged values, not key absence.** Several of these are legitimately platform-written, and a test demanding their absence would fail the day the platform started writing one — reporting a security regression that hadn't happened.

**The ingest guard is generalised.** #1322 added a test that ingest doesn't stamp `memory_type_agent_set`. This asserts ingest's item metadata is disjoint from the *whole* registry, so it fires whether a reserved key is added to ingest or a key ingest stamps is added to the registry. The key-specific version couldn't catch the second direction — and that direction is what cost CAURA-703.

## Not addressed here

None of these mirror into `_system` via `set_system_value`, so the read side still can't distinguish a legacy top-level platform write from a historical forged one. Closing that means routing every writer through the helper — a larger change than reserving the keys, and a different one.

### DIFF
--- a/core-api/src/core_api/services/system_metadata.py
+++ b/core-api/src/core_api/services/system_metadata.py
@@ -15,9 +15,12 @@ every writer goes through:
   ``metadata["_system"]`` namespace — EXCEPT when the caller owns the key
   (``summary`` / ``tags``): then the caller's value stays at top level and
   the platform's copy lives only under ``_system``.
-- Caller input is sanitised at write time: the forgeable telemetry keys and
-  the ``_system`` namespace itself are stripped — a caller cannot inject
-  fake platform telemetry.
+- Caller input is sanitised at write time: every key in
+  ``PLATFORM_ONLY_KEYS`` and the ``_system`` namespace itself are stripped —
+  a caller cannot inject fake platform values. That set covers more than
+  telemetry: governance verdicts, dedup outcomes and row lineage are all
+  forgeable in exactly the same way, and were accepted from callers until
+  the registry was widened to name them.
 - Read-side, ``MemoryOut.system_metadata`` is derived from ``_system`` plus
   the legacy top-level keys, so historical rows (written before this module
   existed) expose the same view without a migration.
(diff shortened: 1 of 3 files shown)


Eldad-Caura:
@erni-a ready for your review at head `d2d2f417`. All checks pass; the PR is BLOCKED only on `REVIEW_REQUIRED`.

**What it fixes:** #1322 made all three write surfaces call `sanitize_caller_metadata`. This fixes what that call strips. `PLATFORM_ONLY_KEYS` named twelve keys; the platform writes considerably more, and every unreserved one was accepted verbatim from callers — on all three surfaces, including the create path C25 always covered. Sixteen keys are now reserved: governance verdicts, dedup/near-duplicate outcomes, timing, lineage and enrichment output.

The one I'd single out is `near_duplicate_of`. It is documented as something callers read back to decide whether to merge or undo, so forging it redirects a decision rather than merely misreporting one.

**The judgement call worth your eye:** `source` is deliberately NOT reserved. The platform writes it (`"auto_chunk"`, `"atomic_fact_fanout"`) and so does ingest, in caller-adjacent item metadata — reserving it would strip that stamp exactly the way widening this boundary nearly cost CAURA-703 in #1322. There is a test pinning the exclusion with the reason attached, so it cannot be quietly reversed.

**One consequence to be aware of before merging, not after.** `PLATFORM_KEYS` is `PLATFORM_ONLY_KEYS | CALLER_OWNABLE_KEYS` and feeds `extract_system_metadata`, so these keys now also surface in `MemoryOut.system_metadata` where present. That is additive and matches the field's contract. The caveat: for rows written *before* this change a caller could have forged one, and it will now be presented as platform metadata. That is inherent to the one-release legacy dual-write and already true of the original twelve — newly reaching more keys, not new behaviour.

**Verification** — full core-api suite 6190 passed (rebased onto `main` including #1316 and #1322), ruff check and format at CI's scopes run separately, mypy clean bar two pre-existing `dateutil` errors in an untouched file, ratchet and sentinel after `git add`. `claude-review` ran on this head and found no issues, having independently cross-checked each reserved key against its writer.

Teeth confirmed one at a time: restoring the old twelve-key registry fails the new test on **all three** surfaces, and making ingest stamp a newly-reserved key fires the generalised ingest guard.

Two notes on the tests. `create` now has a route-level test for the first time — I flagged in #1322 that every existing test drove the helper directly, which is how M-48 and M-52 stayed invisible; it failed against the old registry like the other two. And the assertions are on forged *values* rather than key absence, so they cannot report a false regression the day the platform legitimately starts writing one of these.

Not addressed here, and stated in the description: none of these mirror into `_system`, so the read side still cannot tell a legacy top-level platform write from a historical forged one. That needs every writer routed through `set_system_value` — a larger and different change.

The merge is yours, not mine.

erni-a (approved):

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #1325, opened 2026-09-06
fix(reports): stop leaking the caller's own private rows into group reports

Eldad-Caura:
Closes audit finding **M-31** (U42). Verified against current source, then reproduced end-to-end before anything was changed.

## The leak

`GET /api/v1/reports` narrows what it returns by destination audience. The breakdown path does that correctly — it sets `agent_id` only on the self branch. **`list_query` did not**: it passed `caller_agent_id` on every detail path, and the storage visibility predicate admits `scope_agent` rows when that is set:

```python
if caller_agent_id:
    or_(scope_org, scope_team, and_(scope_agent, Memory.agent_id == caller_agent_id))
else:
    Memory.visibility != "scope_agent"
```

So an **agent** caller requesting `destination=internal_group` got its own private memories back in the content sections. One seeded private title appeared in **four places at once**:

| section | what surfaced |
|---|---|
| `learning` | title + date |
| `value_highlights` | title, type, agent |
| `working_on.Governing.items` | first item |
| **`spotlight.headline`** | the headline of the report |

That is the report an agent then relays verbatim into a group channel — the audience narrowing this two-check design exists to enforce. `scope=org` carried the same defect, reachable because `asserted_agent` falls back to the `agent_id` query param; there the audience is every readable tenant.

Scope of impact, stated plainly: the surfaced rows are the caller's **own**, not another agent's. This is self-oversharing to a wider audience, not cross-agent exposure.

## Why review missed it

`test_report_group_counts_private_writes_but_hides_content` asserts this exact invariant, and **passed against the vulnerable code**. It calls with an admin credential that has no agent identity, so `caller_agent_id` stayed `None` and the storage predicate dropped private rows for a reason unrelated to the route's audience logic. The invariant was real; that caller could not violate it, so it could not test it. Adding one query param — `agent_id=a1` — is the entire difference.

Same shape as the M-25 finding, where `enforce_not_agent_credential` returns early for admins and `get_test_auth()` returns the admin key. Worth naming as a pattern: **a test whose caller cannot reach the vulnerable path proves nothing, however precisely it states the invariant.**

## The fix, and why it isn't just a one-line gate

The audience decision is now resolved **once**, as `is_self_scope`, and used for the breakdown, the visibility identity, the author filter, and the agent leaderboard.

The predicate was previously written out by hand at each site — and M-31 *was* two of those copies disagreeing. My first version of this fix gated the identity correctly but carried a comment saying *"this mirrors that condition exactly"*, which is the same kind of unenforced claim that caused the bug. Sharing the variable makes the mirror structural instead of asserted.

`meta.belonging` deliberately keeps its own predicate, with a comment saying why: it omits the `caller_agent_id` conjunct, so a human-dashboard caller asking for a self destination still gets a belonging block even though its `scope` reads `"group"`. That is response shape, not visibility — swapping in the shared predicate would silently start returning `null`.

Also corrected: the `breakdown_query` comment claimed the content sections *"still exclude scope_agent — see list_query below — so no private content is surfaced to the group."* That was a claim about other code, and it was false. It now says what the aggregate path actually does — including that no `agent_id` is set on this branch, so counts include **every** agent's private rows, not just the caller's. Aggregates were and remain count-only; the projection is `(memory_type, agent_id, status[, tenant_id], COUNT(*))`.

## Verification

Full core-api suite **6194 passed**, 0 failed (rebased onto current `main`). `ruff check` / `ruff format --check` at CI's scopes run separately; mypy clean bar 2 pre-existing `dateutil` errors in an untouched file; ratchet and sentinel after `git add`.

Three mutants, one at a time:

```
pass the identity unconditionally (the leak)  -> group + org leak tests FAIL
never pass the identity (over-refusal)        -> both self-view params FAIL
ignore the agent_id query param               -> group leak test FAILS (+8 others)
```

The third mutant is there because the first version of these tests **passed vacuously** under it — if identity resolution breaks, `caller_agent_id` is `None`, private rows are excluded for the wrong reason, and the assertions prove nothing. The group test now asserts `meta.fleet_id` as a positive witness that the caller really resolved to an agent. The org test has no equivalent witness available (`meta.fleet_id` and `meta.belonging` are both forced `None` under `org_mode`); that limit is stated rather than papered over.

The over-refusal direction covers **both** members of `_SELF_AUDIENCES`, not just `owner_1to1`, so "fix the leak" cannot be satisfied by never passing the identity at all.

## Sibling instances: checked, none found

The thing most worth knowing about a fix like this is whether it leaves a twin open. It does not:

- **`get_agent_activity_digest`** passes `agent_id` to `sc.get_agent_activity_digest(...)` as a row **filter** over precomputed digests, never as a visibility identity — the route holds no `caller_agent_id` at all.
- **The digest generator's only memory fetch** (`agent_digest.py`) omits `caller_agent_id` entirely, so `Memory.visibility != "scope_agent"` applies and private rows never enter the corpus. The storage `/list` router has no `include_scope_agent` knob, so omission there is a hard exclusion.
- `highlights_query` is derived from `list_query` *after* the conditionals, so it inherits the fix rather than needing its own.
- `reports.py` is the only file in `core-api/src` referencing `_SELF_AUDIENCES` or `include_scope_agent`, so there is no other audience-narrowing surface.

One pre-existing, deliberate behaviour surfaced while checking, **not** changed here: a credential with no gateway agent identity (admin key, human dashboard) can name any registered agent via the query param and read that agent's `scope_agent` rows on the self path. `memories.py` documents this as intended for the dashboard. Flagging it only in case the threat model has moved.

### DIFF
--- a/core-api/src/core_api/routes/reports.py
+++ b/core-api/src/core_api/routes/reports.py
@@ -292,13 +292,28 @@ async def get_report(
         # (totals/by_type/by_agent/quality/trend). These are real, decision-
         # bearing memories an agent kept private — excluding them made
         # durable_memories_written measure "team-visible" rather than "written",
-        # and disproportionately undercounted privacy-heavy tenants. This is a
-        # pure count; the content sections (value_highlights/learning/spotlight/
-        # working_on) still exclude scope_agent — see list_query below — so no
-        # private content is surfaced to the group. Ignored on the self path
-        # (agent_id set), which already scopes visibility to the caller.
+        # and disproportionately undercounted privacy-heavy tenants. Ignored on
+        # the self path (agent_id set), which already scopes visibility to the
+        # caller.
+        #
+        # No memory title, content or memory id crosses into the response from
+        # here — the projection is (memory_type, agent_id, status[, tenant_id],
+        # COUNT(*)). Counts ARE bucketed by agent, so on this branch, where no
+        # ``agent_id`` is set, EVERY agent's private rows are counted, not just
+        # the caller's: an agent whose whole window is private still appears in
+        # by_agent with its true volume. That is the trade the paragraph above
+        # argues for, stated rather than glossed.
+        #
+        # Whether private CONTENT surfaces is a separate decision, made by
+        # ``is_self_scope`` below — see U42/M-31 there.
         "include_scope_agent": True,
     }
+    # The audience/data-scope decision, resolved ONCE. Both the breakdown below
+    # and the visibility identity on the detail path key off it, and M-31 was
+    # exactly those two drifting apart: this branch was right and ``list_query``
+    # re-derived the predicate by hand and got it wrong. Sharing the variable is
+    # what keeps them together; a comment claiming they match is not.
+    is_self_scope = False
     if org_mode:
         # Org-wide: aggregate across every tenant the credential may read.
         # Team/org-visible only (no agent_id ⇒ excludes scope_agent); the
@@ -309,6 +324,7 @@ async def get_report(
         # Narrowest: only the caller's own contributions.
         breakdown_query["agent_id"] = caller_agent_id
         scope_label = "self"
+        is_self_scope = True
     else:
         # internal_group / external: the caller's own fleet (team/org-visible).
         # ``external`` shares the same query but its detail is stripped below.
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@erni-a plersae approve

erni-a (approved):

PR #1335, opened 2026-09-07
fix(fleet): refuse an agent-scoped credential at POST /fleet/commands

Eldad-Caura:
**Not from the audit file.** Found while fixing M-25 (#1325's predecessor), which closed this same missing gate on `DELETE /fleet/{fleet_id}`. I read it as higher severity than the MEDIUMs still in the queue, which is why it jumped them.

## The gap

`POST /fleet/commands` had `enforce_read_only` and `enforce_tenant` and nothing else. Neither narrows an agent-scoped credential naming its **own** tenant — it may write, and the tenant it names is its own. `body.node_id` isn't bound to the caller's scope either, and `GET /fleet/nodes` hands out every node id in the tenant behind `enforce_tenant` alone.

What that reaches is not abstract:

| step | where |
|---|---|
| `payload.source` written over the plugin's source tree | `plugin/src/deploy.ts` |
| `payload.env_vars` merged into its `.env`, any plugin-prefixed key — `CAURA_API_URL`, `CAURA_API_KEY` included | `plugin/src/deploy.ts` |
| `npm run build` | `plugin/src/deploy.ts` |
| gateway restart, so the caller's code goes live | `plugin/src/heartbeat.ts` |

So an agent-scoped key refused at `PATCH /agents/{id}/trust` could instead run code on the host answering for it.

**Scope, stated plainly: within-tenant privilege escalation, not cross-tenant.** `enforce_tenant` still holds and storage 404s another tenant's node. The escalation is agent-plane → admin-plane inside one tenant. Also worth being explicit that the gate's teeth come from the gateway: `X-Agent-ID` is trustworthy on the gateway path (`X-Gateway-Secret`, and the gateway overwrites client-supplied values), and on the OSS-direct path every caller shares one key, so there's no agent/tenant distinction there to protect in the first place. Same deployment model M-25 accepted.

## The HMAC check is not a compensating control

`signature` appears **nowhere** in core-api or core-storage-api — the heartbeat response emits `id`/`command`/`payload` and stops. `verifyCommandSignature` defaults to permissive, so every command arrives unsigned and is accepted. The one setting that would fail closed, `CAURA_REQUIRE_SIGNED_COMMANDS=true`, would reject **every** command this backend issues, auto-upgrades included.

That's a control an operator could switch on and lose their fleet to. It is not fixed here and wants its own issue — see the follow-ups below.

## Nothing legitimate loses a channel

Both internal producers bypass this route: the auto-upgrade calls `sc.create_command` from inside the heartbeat handler, and the interview scheduler does the same behind `enforce_admin` on `POST /admin/interview/schedule/run`. No plugin or client posts to this route — the plugin posts only to `/fleet/commands/{id}/result`. The full suite passing unchanged is the other half of that argument: no existing test asserted an agent credential may queue.

## Verification

Full core-api suite **6196 passed**, 0 failed. `ruff check` / `ruff format --check` separately at CI's scopes; mypy clean bar 2 pre-existing `dateutil` errors in an untouched file; ratchet, sentinel and tenant-scope gate after `git add` (the tenant-scope gate reports 0 allowlist churn).

Three mutants, one at a time:

```
drop the gate (the finding)   -> refusal test FAILS
refuse every caller           -> over-refusal test FAILS (+6 pre-existing)
gate placed after the write   -> refusal test FAILS on STATE, not status
```

The tests assert the error **code**, not just 403 — all three gates on this route answer 403, so a status-only test would pass against a version that refused for the wrong reason. That is what kills the middle mutant.

## Two false claims corrected, one of them mine

Both comment-only, both disproved while writing the above:

- `test_fleet_command_cannot_target_another_tenants_node` said "nothing downstream re-checks the pair" and that the pending-command query "is keyed on `node_id` alone". `fleet_get_pending_commands` filters the **(node, tenant, status)** triple — a second, independent gate, added later by #1173. I checked the blame rather than assuming the two gates shipped together, which is how I caught that my first version of the correction was also wrong about the history.
- **My own first draft** justified an extra heartbeat assertion with that same wrong claim. Since delivery is tenant-scoped, the queue listing is the strict superset — the heartbeat was redundant. Removed; the three mutants still die, which is the evidence it cost no coverage.

`/simplify` caught the second one. The first was pre-existing and two screens above my new code, so I fixed it rather than leaving a false security claim next to a true one.

## Stability label

`POST /fleet/commands` is in the stable REST surface in `docs/public-api-stability.md`, so narrowing which credentials may call it is arguably breaking, and the doc says to over-label and let review decide. Hence `kind/breaking`.

I have **deliberately not** added the `BREAKING CHANGE:` trailer: release-please would turn it into a major version bump for the whole repo, and that's a release decision, not mine. Say the word and I'll amend it in.

## Follow-ups this surfaced — filed separately, not here

1. **An admin-plane route inventory test.** This class has now been fixed one route at a time six times, and `POST /fleet/commands` is the second visit to *this* handler — the 2026-08-14 pass added `enforce_read_only` and `enforce_tenant` here and walked away without the plane gate. Scoped to the 6 admin-plane routers it's ~20 mutating routes, 6 gated, so ~14 allowlist entries — comparable to `tenant_scope_gate`'s 63 — and AST-detectable, since `enforce_not_agent_credential` reads no body. App-wide it would need a 49-entry allowlist, which that gate's own docstring argues trains reviewers to skim.
2. **`POST /fleet` and `PATCH /agents/{agent_id}/tune`** — the same shape already in the tree. `create_fleet` has no plane gate while its lifecycle siblings do (low impact, and gating only the explicit route would be cosmetic while the heartbeat creates fleets implicitly). `patch_agent_tune` has `enforce_tenant` and a self-plane check but **no `enforce_read_only` and no `enforce_usage_limits`** — that is the H-12 defect surviving on the sibling route the sweep didn't reach, and it's the more real of the two.
3. **Server-side command signing: implement or delete.** Per above.
4. **`CommandIn.payload` has no size cap**, while its sibling `HeartbeatIn` caps `recall_metrics` at 4 KB and `reconcile` at 8 KB with a paragraph of reasoning each. `middleware/ingest_body_size.py` gates only `/ingest/*`, so `payload` is an unbounded JSONB write; `MAX_SOURCE_SIZE` is enforced on the node, after the row is stored.

Deliberately **not** validating `CommandIn.command` against an enum: the plugin already answers unknown commands with a recorded failure, this file twice cites the no-version-handshake posture as why `HeartbeatIn`/`CommandResultIn` are permissive, and an enum would mean a backend deploy before any node could be told to run a new command. The dangerous commands are the known ones.

### DIFF
--- a/core-api/src/core_api/routes/fleet.py
+++ b/core-api/src/core_api/routes/fleet.py
@@ -1034,20 +1034,51 @@ async def create_command(
     body: CommandIn,
     auth: AuthContext = Depends(get_auth_context),
 ):
-    """Queue a command for a fleet node."""
-    sc = get_storage_client()
+    """Queue a command for a fleet node.
+
+    Auth: a write-capable tenant-owner key. Agent-scoped credentials are
+    blocked (BFLA) — dispatching to a node is admin-plane, the same call
+    ``DELETE /fleet/{fleet_id}`` and ``POST /fleet/{fleet_id}/purge`` make.
+    """
     # Queueing a command is a write, and the tenant it lands in comes from the
-    # REQUEST BODY. Both gates are required, and the order matters: resolve the
-    # target tenant first, then enforce against the resolved value.
+    # REQUEST BODY. All three gates are required, and the order matters: resolve
+    # the target tenant first, then enforce against the resolved value.
     #
     # Without ``enforce_tenant`` any authenticated caller could set
     # ``body.tenant_id`` to a victim tenant and queue commands into their fleet
     # — the GET sibling immediately below has always enforced this, so the write
     # was the weaker of the pair. Without ``enforce_read_only`` a
     # capabilities={'read'} credential could do the same.
+    #
+    # Neither of those constrains an agent-scoped credential naming its OWN
+    # tenant, and ``body.node_id`` is not bound to the caller's scope —
+    # ``GET /fleet/nodes`` hands out every node id in the tenant behind
+    # ``enforce_tenant`` alone. What lands on the node is not abstract: a
+    # ``deploy``/``update_plugin`` payload carrying ``source`` takes the plugin
+    # down ``deployPlugin`` (``plugin/src/deploy.ts``), which writes it over the
+    # plugin's own source tree, merges ``env_vars`` into its ``.env`` for any
+    # key with the plugin env prefix — ``CAURA_API_URL`` and ``CAURA_API_KEY``
+    # included — and builds; the caller's code goes live on the gateway restart
+    # that same branch then asks for (``plugin/src/heartbeat.ts``). So an
+    # agent-scoped key that cannot raise its own trust_level
+    # (``PATCH /agents/{id}/trust`` refuses it) could instead run code on the
+    # host answering for it.
+    #
+    # The plugin's HMAC check is not a second line of defence: nothing in
+    # core-api or core-storage-api ever puts a ``signature`` on a command — the
+    # heartbeat response above emits ``id``/``command``/``payload`` and no more
+    # — so every command arrives unsigned, which the plugin accepts by default.
+    #
+    # No internal producer loses a channel: both bypass this route. The
+    # auto-upgrade calls ``sc.create_command`` from inside the heartbeat
+    # handler, and the interview scheduler does the same behind
+    # ``enforce_admin`` on ``POST /admin/interview/schedule/run``.
     auth.enforce_read_only()
     tenant_id = body.tenant_id or auth.tenant_id
     auth.enforce_tenant(tenant_id)
+    auth.enforce_not_agent_credential("queue fleet commands")
+
+    sc = get_storage_client()
     try:
         cmd = await sc.create_command(
             {
--- a/tests/test_route_authz_gaps.py
+++ b/tests/test_route_authz_gaps.py
@@ -31,8 +31,11 @@ fixture, so the in-process storage app can see them.
 from __future__ import annotations
 
 import uuid
+from typing import NamedTuple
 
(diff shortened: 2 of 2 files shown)


Eldad-Caura:
@erni-a ready for review.

All 14 checks green on head `2b7879c7` (verified against the head SHA, not `gh pr checks`): CI, CodeQL both analyses, bandit + pip-audit, DCO, legacy-name ratchet, do-not-touch sentinel. `claude-review` returned no issues. Locally: full core-api suite 6196 passed, ruff at both CI scopes, mypy clean bar 2 pre-existing `dateutil` errors in an untouched file, tenant-scope gate reporting 0 allowlist churn.

One line of production change — `auth.enforce_not_agent_credential("queue fleet commands")` — plus the comment establishing why, and two tests confirmed by mutation one mutant at a time (drop the gate; refuse every caller; gate placed after the storage write — all three killed).

**One decision is yours, and it is the only thing I have deliberately left undone.** `POST /fleet/commands` is in the stable REST surface in `docs/public-api-stability.md`, so narrowing which credentials may call it is arguably breaking. I applied `kind/breaking` per that doc's "better to over-label" guidance, but did **not** add the `BREAKING CHANGE:` trailer — release-please would turn it into a major version bump for the whole repo, and that is a release call rather than mine. Amending it in is one command if you want it.

The PR body also carries four follow-ups this surfaced and I did not pull into scope — most notably `PATCH /agents/{agent_id}/tune`, which is missing `enforce_read_only` and `enforce_usage_limits` and is the more real of the two remaining same-shape gaps.

erni-a (approved):

PR #1340, opened 2026-09-07
test(authz): inventory every mutating route's gates instead of finding them by hand

Eldad-Caura:
Follow-up #1 from #1335 and #1337. Those were the fifth and sixth hand-found instances of one class, and `POST /fleet/commands` was the **second visit to that handler** — the 2026-08-14 pass added `enforce_read_only` and `enforce_tenant` there and walked away without the plane gate. Manual sweeps have not converged, so this makes the omissions machine-readable.

## What it asserts

`tests/test_authz_gate_inventory.py` enumerates the mutating REST surface from the live app and checks two invariants:

| invariant | scope | allowlist |
|---|---|---|
| calls `enforce_read_only`, is admin-only, or is listed | app-wide | 12 |
| calls `enforce_not_agent_credential`, is admin-only, or is listed | 6 admin-plane routers | 9 |

Plus: no stale allowlist entry, every router classified, the test-only exclusion earned per route and non-empty, allowlist reasons naming code corroborated, and a shrink-only ceiling on known gaps.

## It found a real gap on its first run

**`POST /crystallize` had `enforce_tenant` and nothing else.** `start_crystallization` reserves a report row and then does "a create per extracted fact, each with its own embedding and dedup lookups" — so a demo-sandbox or `capabilities={'read'}` credential could author memories across the tenant. Fixed here.

It also gets **behavioural** tests beside its 40 siblings in `test_route_authz_gaps.py`, because the static check proves the call is *present*, not that the route refuses — and that distinction matters more than usual here: the creates happen in a worker holding no `AuthContext`, so this gate is the whole of it.

`enforce_usage_limits` is deliberately **not** added, and that one is settled rather than hedged: the gate's set is `PLAN_LIMIT_GATED_OPS` (not `WRITE_QUOTA_OPS`, which differs on exactly the case the policy spends eight lines on), this route performs none of those, and a run *archives* more rows than it creates — making it a reduction path, which the gate's own docstring carves out.

## Design decisions worth reviewing

**Route resolution walks the live app**, because `app.openapi()` carries no endpoint function and this is a check on what the handler does. The self-check is **one-directional**: every documented operation must appear in the walk, but the walk may see more. `tenant_scope_gate`'s bidirectional check can't be copied — it walks `core_storage_api.app`, which hides nothing, whereas core-api serves four undocumented operations.

**The classifier resolves two levels of module-local indirection**, and what that buys is narrower than it looks: `skills_inbox` calls no `enforce_*` at all, so its offender status is unchanged either way. What indirection produces is the `helpers` set — which is what lets a reason naming a mechanism be *checked* instead of believed. Both blind spots (imported helpers; `ast.walk` ignoring control flow) are documented and both fail safe.

**Allowlist reasons that name code are corroborated against the classifier.** That closes a real hole: blinding the classifier otherwise leaves both invariants passing, because the routes whose classification changed are exactly the ones the allowlists excuse by key.

**Scope is stated up front**: this covers `APIRoute` entries only. `/mcp` is `app.mount(...)` — a `Mount` with no `.routes` plus a plain Starlette `Route`, absent from the schema — so the MCP tools are **not** covered. They guard themselves with `_check_write_scope`; that surface wants its own inventory and doesn't have one.

## A live gap this found and does NOT fix

Writing the reasons down is what exposed it. **`_require_inbox_admin` admits `org_role == "admin"`, and `org_role` and `capabilities` are independent gateway headers — so a read-only org admin can approve, reject, quarantine, defer or edit a skill with no write gate on the path.** Five behaviour changes each wanting their own over-refusal test is a security PR, not this one.

Recorded as `KNOWN GAP:` entries under a **shrink-only ceiling**, borrowing `tenant_scope_gate`'s `blind-spot <- backlog` precedent. The alternative is a list where "excused" and "not yet fixed" look identical, and then neither gets read. `test_known_gaps_do_not_grow` fails both if the count rises and if a gap is fixed without lowering the ceiling.

## Also: a stale claim in the CAURA-602 startup guard

`app.py`'s guard comment claimed *"every core-api route is `include_in_schema=True`, so none is hidden from this check."* Four operations are hidden. The claim was **true when written** (#364, 2026-06-15) and was falsified twice — #582 (2026-07-20) added the trailing-slash inbox route, then #782 (2026-08-14) added the keystones alias. The first falsification came from a *different file*, so nothing a reviewer of #582 was looking at pointed here.

No live bug: all three opt-out entries are schema-visible. The guard is deliberately **not** widened to walk the routes — that would move the private `_IncludedRouter` internals into import-time app construction and trade a false `RuntimeError` for a service that won't boot on the next FastAPI upgrade. The constraint is now stated, and the error message names the hidden-route case. Confirmed empirically that a hidden opt-out path raises, rather than asserting it.

## Verification

Full core-api suite green, rebased onto current `main`. `ruff check` / `ruff format --check` separately at CI's scopes; mypy clean bar 2 pre-existing `dateutil` errors in an untouched file; ratchet, sentinel and tenant-scope gate after `git add` (0 allowlist churn).

Eight mutants, one at a time:

```
regress the write gate (revert crystallize)  -> static + both behavioural tests FAIL
regress the plane gate (revert M-25)         -> static + the M-25 test FAIL
stale allowlist entry                        -> staleness test FAILS
break the router walk (the 0.137 failure)    -> 8 of 9 FAIL, loudly
blind the classifier                         -> corroboration + flag tests FAIL
typo a plane router                          -> classification + existence FAIL
unflag a test-only route                     -> flag test FAILS
add enforce_usage_limits to crystallize      -> the pinning test FAILS
```

## Corrections made during review

`/simplify` caught six false claims of mine, which is worth stating plainly given this file's entire purpose is that claims be checkable:

- *"`enforce_read_only` returns early for `is_admin`"* — it has no `is_admin` branch. The exemption is sound via how the admin context is *constructed*.
- *"`enforce_org_admin` IS stronger than the plane gate"* — the supporting statement lives in auth's **Path 4** comment marked "Read on THIS branch only"; Path 2 returns `org_role="admin"` alongside `agent_id`. It's now in neither exemption set.
- *"five false positives out of fourteen"* — unsupported, and the mechanism was wrong.
- The `tenant_scope_gate` citation was **inverted** — its ~380 quote argues against a full manifest, and it ships a 63-entry exception list.
- The over-refusal guards cited for heartbeat cover read-only/demo, not agent-scoped credentials.
- WT-4 was misfiled in the history table; it's an over-refusal fix, the opposite direction.

And the structural one: `ADMIN_PLANE_ROUTERS` was itself an unjustified allowlist. **`keystones` has four mutating routes and no plane gate and sat outside every set with nothing failing.** It's correctly exempt — trust-plane, where an agent authors its own rules at trust ≥ 1 by design — which is why it was easy to miss: a wrong classification announces itself, a missing one doesn't. Every router is now classified and `test_every_router_is_classified` fails on any that isn't.

## Follow-ups, not in scope

1. **Close the five `skills_inbox` write gaps** — the `KNOWN GAP` ceiling holds the line meanwhile.
2. **Extract the FastAPI `_IncludedRouter` walk into `common/`** — it's now duplicated character-for-character between this file and `scripts/tenant_scope_gate.py`, leaning on a private attribute that already broke once on 0.137, in two places with different self-checks.
3. **An MCP-surface inventory** — the tools guard themselves with `_check_write_scope`, unenumerated.
4. Still open from #1337: `AuthContext.enforce_self_agent()` plus a coded error across the 8 inline self-plane checks.

### DIFF
--- a/core-api/src/core_api/app.py
+++ b/core-api/src/core_api/app.py
@@ -1116,16 +1116,46 @@ app.router.routes.append(
 # paths are no longer top-level ``APIRoute.path`` entries — which is exactly what
 # silently broke this guard when 0.137 shipped. ``app.openapi()`` is the stable,
 # public surface and lists the prefixed paths under both old (flatten) and new
-# (mount) FastAPI. Every core-api route is ``include_in_schema=True``, so none is
-# hidden from this check.
+# (mount) FastAPI.
+#
+# The schema is NOT the whole served surface, though, so this guard constrains
+# what may go in the opt-out set: an opt-out path must be schema-visible. Four
+# operations are served and undocumented — the PERMANENT legacy keystones
+# alias, registered with ``include_in_schema=False`` immediately after the
+# canonical ``keystones_router`` above (three operations), and the trailing-slash
+# ``GET /api/v1/skills-inbox/`` in ``routes/skills_inbox.py``. Naming any of
+# them here would raise below even though the route exists.
+#
+# (An earlier revision of this comment claimed "every core-api route is
+# include_in_schema=True, so none is hidden from this check". It was true when
+# written in #364, 2026-06-15, and stopped being true twice: #582 on
+# 2026-07-20 added the trailing-slash inbox route, then #782 on 2026-08-14
+# added the alias. Worth noting which way that went — the first falsification
+# came from a different file, so nothing a reviewer of #582 was looking at
+# would have pointed here. That is the argument for the claim being narrow
+# enough to check, which is what the paragraph above now aims at.)
+#
+# DELIBERATELY not fixed by unioning the schema with a walk of ``app.routes``.
+# The walk needs the private ``_IncludedRouter`` internals this comment exists
+# to warn about, and putting them in import-time app construction trades a
+# false RuntimeError for a service that will not boot on the next FastAPI
+# upgrade — a worse failure for a case with no live bug: all three current
+# entries are documented. The realistic way to reach it is a timeout opt-out on
+# ``/keystones``, since ``_is_opted_out`` matches exactly and the alias is a
+# distinct path that would need its own entry. If that day comes, exempt the
+# canonical path and handle the alias in ``_is_opted_out`` rather than widening
+# this guard.
 _registered_paths = set(app.openapi().get("paths", {}))
 for _opt_out in _TIMEOUT_OPT_OUT_PATHS:
     if _opt_out not in _registered_paths:
         raise RuntimeError(
-            f"RequestTimeoutMiddleware opt-out path {_opt_out!r} is not "
-            "registered on the FastAPI app. Either the route was renamed/"
-            "removed or _TIMEOUT_OPT_OUT_PATHS in middleware/request_timeout.py "
-            "is stale; both are silent-create regressions waiting to happen."
+            f"RequestTimeoutMiddleware opt-out path {_opt_out!r} is not in the "
+            "OpenAPI schema. Either the route was renamed/removed or "
+            "_TIMEOUT_OPT_OUT_PATHS in middleware/request_timeout.py is stale — "
+            "both are silent-create regressions waiting to happen. If instead "
+            "the route exists but is registered include_in_schema=False (the "
+            "legacy keystones alias, or GET /api/v1/skills-inbox/), this guard "
+            "cannot see it: opt out of the canonical path instead."
         )
 
 
--- a/core-api/src/core_api/routes/crystallizer.py
+++ b/core-api/src/core_api/routes/crystallizer.py
@@ -56,7 +56,36 @@ async def trigger_crystallization(
     body: CrystallizeRequest,
     auth: AuthContext = Depends(get_auth_context),
 ):
-    """Trigger crystallization for a tenant (analysis + auto-curate)."""
+    """Trigger crystallization for a tenant (analysis + auto-curate).
+
+    Auth: a write-capable credential for the target tenant.
+    """
+    # Found by ``tests/test_authz_gate_inventory.py`` on its first run — the
+    # same class as H-12/H-13/M-25/#1335/#1337, and the reason that file exists.
+    #
+    # ``enforce_tenant`` alone says WHICH tenant, never whether this credential
+    # may write to it. A run is not read-shaped despite the response being a
+    # bare report id: ``start_crystallization`` reserves a report row and then
+    # does "a create per extracted fact, each with its own embedding and dedup
+    # lookups" (see its docstring), so a demo-sandbox or capabilities={'read'}
+    # credential could author memories across the tenant.
+    #
+    # ``enforce_usage_limits`` is NOT added, and this one IS settled — do not
+    # "finish" it by adding the gate. The gate's set is
+    # ``PLAN_LIMIT_GATED_OPS`` (create / bulk_create / redistribute), not
+    # ``WRITE_QUOTA_OPS``, and this route performs none of those: it reserves a
+    # report row and publishes a request to the event bus, so the creates
+    # happen later in a worker that holds no ``AuthContext``. Gating here would
+    # gate the TRIGGER, not the writes.
+    #
+    # And a run is a reduction path, which is the case ``enforce_usage_limits``
+    # explicitly carves out ("users in read-only mode must be able to delete
+    # data to get back under limits"): each cluster archives its members —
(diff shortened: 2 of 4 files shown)


Eldad-Caura:
Fixed — the review is right and it was worse than a drift: the alias is **76 lines above** the comment and 110 above the error message, not ten, and `"above its canonical registration"` had the direction backwards too (the alias sits *below* it, 1047 vs 1040).

Rather than correct the count, both references now locate it structurally — "registered with `include_in_schema=False` immediately after the canonical `keystones_router`" — so it cannot drift again. The error message just names the two hidden routes without any positional claim.

Amended onto your rebase (`1d095213`), not force-pushed over it; I diffed the patches first and your rebase was byte-identical to mine.

New head `917e0c71`. Comment-only plus the error-message string; `ruff` clean and the guard still passes at import.

Eldad-Caura:
@erni-a ready for review.

All 14 checks green on head `917e0c71` — verified against the head SHA, and re-verified after the amend, since the earlier green run was on the pre-amend head. The one review finding (the "ten lines above" distance) is fixed, and both references now locate the alias structurally so it cannot drift again.

One line of production change (`enforce_read_only` on `POST /crystallize`, the gap the new inventory found on its first run), plus the inventory itself and a comment correction in `app.py`. Eight mutants, one at a time, all killed.

**Note for whichever of #1340 and the skills-inbox PR merges second.** This PR carries five `KNOWN GAP:` entries and `KNOWN_GAP_CEILING = 5` for the `skills_inbox` write gap. I have now closed that gap in a separate branch — a read-only org admin could approve/reject/quarantine/defer/edit a skill, reproduced and fixed with per-action mutants. So after both land, those five entries must be deleted and the ceiling set to 0. I will do it in whichever merges second; flagging it here so it cannot rot.

Worth knowing that the ratchet is what forces that cleanup rather than leaving it optional: `test_known_gaps_do_not_grow` asserts the count EQUALS the ceiling, so removing entries without lowering it fails, and lowering it without removing them fails too.

erni-a (approved):

Eldad-Caura:
Correcting something I posted on this PR. I wrote:

> Worth knowing that the ratchet is what forces that cleanup rather than leaving it optional: `test_known_gaps_do_not_grow` asserts the count EQUALS the ceiling, so removing entries without lowering it fails, and lowering it without removing them fails too.

Only the first half was verified. The second half was tested with a mutant that inserted a new `KNOWN GAP:` entry into `PLANE_GATE_ALLOWLIST` under a key **that already existed later in the same dict literal**. The later entry silently won, my reason was discarded, the gap count stayed 0, and the test passed — so that run exercised nothing and I read it as confirmation.

Now verified properly in #1343, which moves the ceiling 5 → 0. The mutant relabels an *existing, legitimate* entry as a `KNOWN GAP:` instead of adding a key: the key still names a real route, the route still needs the entry, and the reason names no mechanism, so the staleness, unnecessary-entry and corroboration checks all stay quiet and the ratchet is the only thing that can fire.

```
=== baseline ===                                    10 passed
=== case B: 1 known gap against a ceiling of 0 ===
    test_authz_gate_inventory.py:657: AssertionError: 1 known gaps, ceiling is 0
    FAILED tests/test_authz_gate_inventory.py::test_known_gaps_do_not_grow
    1 failed, 9 passed
=== restored ===                                    10 passed
```

Line 657 is the `<=` assertion, and it is the *sole* failure — which is what makes the direction tested rather than merely asserted. So the claim holds in both directions, but it did not when I made it.

Two other things from this PR that #1343 also fixes:

- **The allowlist-rot hole I flagged here but did not close.** `test_allowlists_have_no_stale_entries` only catches entries naming routes that no longer exist, so fixing the five `skills_inbox` write gaps left their five `KNOWN GAP` lines excusing five now-gated routes with every check in the file still green. I confirmed that empirically before fixing it. `test_allowlists_have_no_unnecessary_entries` now fails on it, and the two invariant predicates are factored out so "does this route need excusing" has one definition rather than two that can drift.

- **The reason text on those five entries was wrong**, not just stale. It said `_require_inbox_admin` "admits a read-only org admin", which I had framed as a live escalation. It is not: `platform-auth-api` emits `X-Org-Role` only for user principals and deletes it on the API-key path, the only path that emits `X-Capabilities`. No gateway-minted credential is both an org admin and non-writing. The gap was real and worth closing as defence in depth — core-api parses the two axes independently and had nothing to notice if that pairing ever arrived — but I had asserted the escalation after reading only the consumer side.

PR #1341, opened 2026-09-07
CAURA-720 refactor(enrichment): retire retrieval_hint, document the w…

eyal-bl:
…eight floor

Two prompt refinements, both prompt-only. No schema or behaviour change.

## retrieval_hint — retired (field + the per-fact copy in atomic_facts)

Its only purpose was to augment the embedding: writes embedded "[Retrieval hint]: ...\n\n<content>" while queries embedded raw text. CAURA-222 DISABLED that after the asymmetry capped recall — identical content<->query scored cosine ~0.69 instead of ~1.0, across dedup, entity-lookup, and search ranking — and the follow-up deleted the compose_embedding_text helper as dead code. Both the hot path and the background re-embed have embedded raw content ever since.

The per-fact copy inside atomic_facts was equally inert: the fan-out in memory_service embeds raw fact_content for the same CAURA-222 reason, so each child's hint was stored and never read.

What still reads it is unaffected. scripts/backfill_embeddings.py selects on a non-empty metadata.retrieval_hint to find rows whose stored vector still holds the old prefixed text, and preserves the value as "auditability ground". Stored hints on historical rows are untouched; rows written from here on simply stop matching that selector, which is correct — they were never prefixed. MemoryEnriched.retrieval_hint, the worker's routing entry, the schema field, and the validator all stay, so historical rows and the worker's replay path deserialise unchanged.

## weight — the 0.0-0.2 band is now described

The field is declared 0.0-1.0 but the rubric described only 0.3-1.0, leaving 14.3% of the eToro corpus (0.1 at 9.3%, 0.2 at 4.9%) in an undefined range. The model was filling a gap the prompt left, not violating a stated bound.

That range is not spare. FOUR consumers read weight < 0.3 as "low salience", all with strict <:

  - insights patterns mode  (weight < 0.3 AND recall_count > 0)
  - insights stale mode
  - CRYSTALLIZER_STALE_MAX_WEIGHT  (0.3)
  - LIFECYCLE_STALE_ARCHIVE_WEIGHT (0.3)

So the band is DOCUMENTED, not clamped. Flooring the value at 0.3 — the intuitive guard — would make "< 0.3" unsatisfiable for every new memory and silently starve all four: no error, just empty result sets. The band is worded by its consequence (auto-archival) so the model's choice is informed by what the system does with the number.

The template's "weight": 0.0 moves to 0.5. 0.0 sat outside every band the rubric described and 32 rows landed on exactly it — below even EVOLVE_WEIGHT_FLOOR (0.05), so evolve's dampening cannot have produced them. 0.5 is DEFAULT_MEMORY_WEIGHT, so the exemplar and the code default now agree. A quoted placeholder was considered and rejected: it would model weight as a string, and the service already carries a coercion for non-numeric values like "high", which suggests that has happened.

## Also

Remaining 10 fields renumbered; the atomic_facts cross-reference to retrieval_hint is gone with the field, and the surviving one (suggested_type -> field 1) is pinned by a test. Two stale worker comments corrected: they justified always-writing retrieval_hint and tags by describing prompt behaviour that CAURA-222 and CAURA-719 had already ended.

## Known unknown, for the reviewer

The band may increase how often the model goes sub-0.3, and one of the four consumers is destructive: archive_stale auto-archives rows older than 180 days with weight < 0.3. I could not size that from the corpus I have — it spans 2026-03-27 to 2026-06-24, so zero rows are old enough to qualify and the query returns 0 either way. Worth a count against current prod before assuming the archive footprint is unchanged.

Separately, this does not address reproducibility: two identical writes can still get different weights. That is what weight_source (#1207) records, and only a deterministic producer removes it.

## Type of Change

- [ ] Documentation update
- [ ] Other:

## Checklist

- [ ] `mypy` passes
- [ ] `pytest` passes locally

### DIFF
--- a/common/enrichment/_prompts.py
+++ b/common/enrichment/_prompts.py
@@ -34,6 +34,57 @@ CAURA-719 — ``tags`` and ``status`` are no longer asked of the LLM, so
 Both fields keep their schema entry, default, and validator so
 historical rows, caller-supplied values, and the deferred worker's
 patch path all continue to deserialise unchanged.
+
+CAURA-720 — ``retrieval_hint`` joins them, and ``weight`` gains its
+missing band:
+
+* ``retrieval_hint`` (and the per-fact copy inside ``atomic_facts``) is
+  no longer asked for. Its only purpose was to augment the embedding —
+  writes embedded ``"[Retrieval hint]: …\\n\\n<content>"`` while queries
+  embedded raw text — and CAURA-222 DISABLED that after the asymmetry
+  capped recall: identical content↔query scored cosine ~0.69 instead of
+  ~1.0 across dedup, entity-lookup, and search ranking. The
+  ``compose_embedding_text`` helper was deleted as dead code, and both
+  the hot path and the background re-embed now embed raw ``content``.
+  The per-fact hint was equally inert: ``memory_service``'s fan-out
+  embeds raw ``fact_content`` for the same CAURA-222 reason.
+  What remains reads it: ``scripts/backfill_embeddings.py`` selects on a
+  non-empty ``metadata.retrieval_hint`` to find rows whose stored vector
+  still holds the old prefixed text, and preserves the value as
+  "auditability ground". That selector is unaffected — stored hints on
+  historical rows are untouched, and rows written from here on simply
+  stop matching it, which they should, since they were never prefixed.
+  ``MemoryEnriched.retrieval_hint`` and the worker's routing entry stay
+  for the same reason the two fields above do.
+* ``weight`` — the field is declared ``0.0-1.0`` but the rubric only
+  described ``0.3-1.0``, leaving 14.3% of the eToro corpus (``0.1`` at
+  9.3%, ``0.2`` at 4.9%) in a range the prompt never defined. That range
+  is not spare: FOUR consumers read ``weight < 0.3`` as "low salience"
+  — insights patterns, insights stale, ``CRYSTALLIZER_STALE_MAX_WEIGHT``,
+  and ``LIFECYCLE_STALE_ARCHIVE_WEIGHT`` — so the band is documented
+  rather than clamped. Clamping the floor to 0.3 would make ``< 0.3``
+  unsatisfiable and silently starve all four. The band is worded by its
+  CONSEQUENCE (auto-archival) so the model's choice is informed by what
+  the system does with the number.
+  The JSON template's ``"weight": 0.0`` moves to ``0.5``: ``0.0`` sat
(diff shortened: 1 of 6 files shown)


eyal-bl:
@claude

eyal-bl:
Good catch — verified, and the finding is correct. Fixed in `e33c602b`.

## What was wrong

The claim connected two constants that are genuinely separate:

| Path | Fallback |
|---|---|
| **Enrichment** — `_validate_enrichment` (`service.py:106,108`) + `EnrichmentResult.weight` (`schema.py:73`) | hardcoded **`0.7`**, in two places |
| **No enrichment** — `merge_enrichment_fields.py:110`, `memory_service.py:1442`, only when `weight is None` | `DEFAULT_MEMORY_WEIGHT` = **`0.5`** |

The JSON template belongs to the enrichment path, whose own fallback is `0.7` — so "the exemplar and the code default now agree" was joining unrelated things. `DEFAULT_MEMORY_WEIGHT` is reached only on the path #1207 records as `weight_source="default"`.

The test was worse than the docstring: `assert f'"weight": {DEFAULT_MEMORY_WEIGHT}' in _template()` passed **coincidentally**, and encoded a false coupling — changing `DEFAULT_MEMORY_WEIGHT` would have failed a test about the *prompt* and pushed someone to edit the prompt to fix it.

## Fix — took option 1 (correct the claim), not option 2 (unify the constants)

Unifying the `0.7` fallback onto `DEFAULT_MEMORY_WEIGHT` would make the docs true, but weight is 15% of base rank score (`SIMILARITY_BLEND = 0.85`), so that is a ranking behaviour change and belongs in its own PR. Noted as such in the docstring.

- **`_prompts.py`** — `0.5` is now justified only as the mid-range neutral inside a described band. Added an explicit *"deliberately NOT claimed"* note documenting both constants, which path reaches each, and why the exemplar is tied to neither.
- **`test_caura720_weight_band.py`** — `test_template_exemplar_matches_the_code_default` replaced with `test_template_exemplar_is_not_pinned_to_either_default_constant`. It now pins the *absence* of the coupling, and asserts the two constants still differ — so if a later PR does unify them, the test fails and points at the docstrings to update rather than silently going stale.
- **`test_enrichment_prompt.py`** — same claim repeated in a docstring there; corrected too.

The exemplar value stays `0.5` (a deliberate choice), and `test_template_exemplar_is_inside_a_described_band` plus `test_template_keeps_weight_numeric` still pin what actually matters. 72 tests pass across the affected files; ruff clean.

eyal-bl:
@claude

ran-taig (approved):

PR #1342, opened 2026-09-07
fix(embedding): make the local provider's model configurable and refuse a width mismatch (C38)

arkash20:
## `EMBEDDING_PROVIDER=local` was dead on arrival

`_registry` constructed `LocalEmbedding()` with **no arguments**, hard-coding `BAAI/bge-base-en-v1.5` (**768-dim**) while `memories.embedding` is `vector(VECTOR_DIM)` = **1024**. No configuration could change it, so that provider could never have produced a storable vector against this schema.

Worse, the width mismatch only **logged a warning and continued** — so the failure surfaced much later as a Postgres dimension error at INSERT, naming neither the model nor the setting responsible. That warn-and-continue is exactly how the default sat on a 768-dim model against a 1024-dim column without anyone noticing.

## Fix

- **Model is configurable** via `LOCAL_EMBEDDING_MODEL`, defaulting to `BAAI/bge-large-en-v1.5` (natively 1024). Read from the environment rather than a service's settings, because this registry is shared with **core-worker** and must not import one service's config. core-api's `local_embedding_model` maps to the same variable, and a test pins the two defaults together so they can't drift.
- **A width mismatch raises at model load**, naming the knob and a model that works, instead of deferring the failure to the database.

## Scope — deliberately narrow

This is a **dormant provider**. Staging and prod run TEI/`bge-m3` through the OpenAI-compatible client (`OPENAI_EMBEDDING_BASE_URL`), so **nothing in a deployed environment changes**.

Found while hunting for an embedder fallback after the hosted-OpenAI account ran out of credits — the fallback didn't work, and this is why.

## Testing

6 unit tests, including that an empty env value falls back rather than loading a model named `""`.

Note on the local suite: the test database on this machine was broken (tests default to `caura:changeme@/caura`; the running Postgres serves `memclaw/memclaw`). I recreated the role/DB to unblock, but it's unseeded, so ~19 DB-backed tests fail locally **on clean main with these changes stashed** — pre-existing environment breakage, not from this PR. CI is the trustworthy signal here.

Refs: canonical C38.

### DIFF
--- a/common/embedding/_registry.py
+++ b/common/embedding/_registry.py
@@ -26,6 +26,13 @@ from common.provider_names import ProviderName
 
 logger = logging.getLogger(__name__)
 
+# C38 — default model for ``EMBEDDING_PROVIDER=local``. MUST emit VECTOR_DIM
+# dimensions: ``memories.embedding`` is ``vector(VECTOR_DIM)``, so a narrower
+# model is rejected by Postgres at INSERT — far from the misconfiguration that
+# caused it. bge-large is 1024; the previously hard-coded bge-BASE is 768 and
+# could never have worked against this schema.
+DEFAULT_LOCAL_EMBEDDING_MODEL = "BAAI/bge-large-en-v1.5"
+
 # Cache OpenAI provider instances by (api_key, model). Each instance
 # holds a long-lived ``AsyncOpenAI`` (and therefore a long-lived
 # ``httpx.AsyncClient`` connection pool); without the cache, every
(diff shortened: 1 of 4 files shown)


Eldad-Caura (approved):

arkash20:
Rebased onto main. CI green (14 checks, 0 failing).

@Eldad-Caura please approve

Eldad-Caura (approved):

PR #1343, opened 2026-09-07
fix(skills-inbox): add the missing write gate on the five inbox actions

Eldad-Caura:
Closes the write-gate gap `tests/test_authz_gate_inventory.py` (#1340) recorded as a `KNOWN GAP`, and closes an allowlist-rot hole that same file shipped with.

## What was wrong

`approve` / `reject` / `quarantine` / `defer` / `edit` were guarded on the **admin axis only**, by `_require_inbox_admin`, and called no `auth.enforce_*` at all — neither that helper nor `_require_tenant` does; both raise `HTTPException` directly. Of the routes the inventory counts as mutating, these five were the only ones that write tenant state and reached no write gate. Its other exemptions are POST-bodied reads gated by `enforce_readable_tenant`, and the unauthenticated bootstrap routes.

## This is defence in depth, not a live escalation — and my first draft said otherwise

The first version of this change claimed `org_role` and `capabilities` are independent gateway headers, so an org admin could hold a read-only key, clear the admin check, and meet no write gate.

They **are** independent fields on `AuthContext`. But nothing mints that pair: `platform-auth-api` emits `X-Org-Role` only for user principals and **deletes** it on the API-key path, which is the only path that emits `X-Capabilities`, with its own docstring saying *"keys carry no org membership, and admin surfaces stay human-only"*. So an admin arrives with `capabilities=None`, for which `enforce_read_only` is a no-op.

I read the consumer side in core-api and inferred the producer side in another repo without opening it. The docstring, tests and commit message all say defence-in-depth now.

**What the gate still buys:** core-api stops depending on that. It parses the two axes independently, so "admin implies may write" held only by a convention enforced in another repo and asserted in a docstring there — with nothing here to notice if an ingress, an on-prem deployment, or a caller reaching core-api directly (`gateway_shared_secret` unset, where both headers are client-supplied) ever presented the combination.

## Choices worth reviewing

**Gate goes first in each handler.** The gate reads nothing but `auth`, and putting it ahead of `_require_skills_factory_enabled` refuses a non-writing credential without a cross-service settings read. This is the majority pattern, not a universal one — my first draft overstated it as "all 42 call sites put it first", a grep count that included a docstring mention and checked no positions. Measured with AST: of the 36 existing `enforce_read_only` call sites across 12 other route files, **23 put it first and 13 do not**. The 13 are consistent with the reason above — what precedes the gate is always cheap and synchronous (`enforce_tenant`, a tenant resolver, `_check_stm_enabled`, a perf timer), never an `await`. One real divergence: `documents.py` and `keystones.py` run `auth.enforce_tenant(...)` first, so a read-only key aimed at another tenant gets the tenant refusal there and `READ_ONLY_CREDENTIAL` here. Both refuse; no existing test pins that precedence.

**Written out five times** rather than folded into `_require_inbox_admin`, which all five already call. Centralising would be structurally *stronger* — the inventory's classifier resolves module-local helpers, so a sixth handler calling only `_require_inbox_admin` would satisfy the invariant — but it would make that helper two-axis and its name wrong, and renaming it reaches #1340's allowlist reasons, which are corroborated **by name**.

**The list endpoint stays open.** `GET /skills-inbox` is deliberately readable by any tenant member so a non-admin operator can see what is in flight. Asserted, because it is the regression a router-level `dependencies=[...]` would have caused — and that would have been invisible to the inventory, which reads handler bodies.

## The allowlist-rot hole

`test_allowlists_have_no_stale_entries` only caught entries naming routes that no longer exist. So fixing a gap left its five `KNOWN GAP` lines excusing five now-gated routes with **every check still green** — verified, that is exactly what happened here before the entries were removed.

`test_allowlists_have_no_unnecessary_entries` now fails on an entry whose route satisfies the invariant on its own, and the two invariant predicates are factored out so "does this route need excusing" has one definition rather than two that can drift. The five write-gate entries are deleted and `KNOWN_GAP_CEILING` drops 5 → 0; the plane entries stay, since only the write axis moved.

## Verification

Every test confirmed by mutation, one mutant at a time:

| mutant | result |
| --- | --- |
| drop the gate from all five | all 10 refusal cases fail |
| drop it from ONE handler | only that handler's 2 cases fail |
| gate the list endpoint too | the list-stays-readable test fails |
| swap it for a second admin check | all 10 fail — the false "admin check is narrower" belief does not satisfy them |

They assert the error **code**, not just 403: `SKILLS_FACTORY_DISABLED` is the reachable 403 collision for these callers, so a status-only test would pass against a version that refused for the wrong reason. Codes come from `errors.*` rather than literals — the values are `READ_ONLY_CREDENTIAL` and `DEMO_SANDBOX_READ_ONLY`, not what their constant names suggest.

The 5 → 0 ceiling move is held in **both** directions, which took two attempts to establish honestly:

| mutant | result |
| --- | --- |
| entries removed, ceiling left at 5 | fails on the `==` assertion |
| a live entry relabelled a `KNOWN GAP` | fails on the `<=` assertion |

I first "verified" the second with a mutant that inserted a duplicate key into a dict literal, where the original later entry silently won — so it exercised nothing and passed. **This corrects what I posted on #1340**, where I claimed both directions were verified when only one was. Relabelling an existing entry keeps the key real and the entry necessary, leaving the ratchet as the only thing that can fire; it is now the sole failure, which is what makes the direction tested.

93 passed on the two test files; full suite 6233 passed; `ruff check` and `ruff format --check` clean at CI's scopes; legacy-name ratchet, do-not-touch sentinel and tenant-scope gate all green.

### One unrelated flake seen along the way

`tests/test_unknown_field_rejection.py::test_memory_update_rejects_unknown_field` failed in 2 of 3 full-suite runs and passed in the third with no code change between them. It also failed earlier the same day on a different branch, passes in isolation and in subsets, and does not involve any route or module this PR touches (my two test files make no DB writes at all — one uses a bare `FastAPI()` with fakes, the other is pure `ast`/`inspect`). It looks like accumulated state in the shared test database. Filed separately rather than folded in here.

## Follow-ups not in this PR

- The FastAPI `_IncludedRouter` walk is now duplicated character-for-character between `tests/test_authz_gate_inventory.py` and `scripts/tenant_scope_gate.py` — extract to `common/`.
- No equivalent inventory for the MCP surface.
- `AuthContext.enforce_self_agent()` + a coded error across the 8 inline self-plane sites.

### DIFF
--- a/core-api/src/core_api/routes/skills_inbox.py
+++ b/core-api/src/core_api/routes/skills_inbox.py
@@ -138,6 +138,35 @@ def _require_inbox_admin(auth: AuthContext) -> None:
 
     Centralized so the check stays consistent across all five action
     handlers — a missed handler is a privilege-escalation bug.
+
+    THIS IS THE ADMIN AXIS ONLY: it says WHO the caller is, never
+    whether their credential may write. Each handler calls
+    ``auth.enforce_read_only()`` for that second axis, and the two are
+    not substitutes in either direction.
+
+    That is defence in depth rather than a closed hole, and the
+    distinction is worth stating precisely. Today no gateway-minted
+    credential is both an org admin and non-writing: the auth service
+    emits ``X-Org-Role`` only for user principals and DELETES it on the
+    API-key path, which is the only path that emits ``X-Capabilities``
+    ("keys carry no org membership, and admin surfaces stay human-only").
+    So an admin arrives with ``capabilities=None``, for which
+    ``enforce_read_only`` is a no-op.
+
+    What the gate buys is that core-api stops depending on that. It
+    reads the two axes as independent fields, so "admin implies may
+    write" held only by a convention enforced in another repo and
+    asserted in a docstring there — with nothing here to notice if an
+    ingress, an on-prem deployment, or a caller reaching core-api
+    directly (``gateway_shared_secret`` unset, where both headers are
+    client-supplied) ever presented the combination.
+
+    This is how the authz inventory found them: of the routes it counts
+    as mutating, these five were the only ones that write tenant state
+    and reached no write gate at all. Its other exemptions are
+    POST-bodied reads, gated by ``enforce_readable_tenant``, and the
+    unauthenticated bootstrap routes, which take no ``auth`` and write
+    nothing tenant-scoped.
     """
     # Mirror documents.py:215-216 — admin status may come from either
     # the legacy ``is_admin`` flag OR ``org_role == "admin"``. Keeping
@@ -646,6 +675,7 @@ async def approve(
     blocks the transition if the doc became unsafe between propose
     and apply.
     """
+    auth.enforce_read_only()
     tenant_id = _require_tenant(auth, tenant_id)
     settings = await _require_skills_factory_enabled(tenant_id)
     _require_inbox_admin(auth)
@@ -788,6 +818,7 @@ async def reject(
     depends on ``get_db`` (the settings gate + audit log already ignore
     their ``db`` arg and route through storage).
     """
+    auth.enforce_read_only()
     tenant_id = _require_tenant(auth, tenant_id)
     settings = await _require_skills_factory_enabled(tenant_id)
     _require_inbox_admin(auth)
@@ -940,6 +971,7 @@ async def quarantine(
     poison table — quarantine is reversible by a security admin; only
     Reject crystallizes a poison row.
     """
+    auth.enforce_read_only()
     tenant_id = _require_tenant(auth, tenant_id)
     await _require_skills_factory_enabled(tenant_id)
     _require_inbox_admin(auth)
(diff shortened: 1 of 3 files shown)


Eldad-Caura:
@erni-a this is ready for review.

All 14 checks green on head SHA `871ff55f` (verified via `check-runs` on the SHA itself, not `gh pr checks`), including `claude-review` and `dco`.

Two things worth your eye rather than the diff:

1. **The severity claim changed during review.** I opened this believing a read-only org admin could reach all five actions. That was wrong — `platform-auth-api` deletes `X-Org-Role` on the API-key path, the only path that emits `X-Capabilities`, so nothing mints an admin-and-non-writing credential. It ships as defence in depth: core-api parses the two axes independently and had nothing to notice if that pairing ever arrived (on-prem, an ingress, or `gateway_shared_secret` unset). If you read the gap as not worth five behaviour changes on that basis, that is a reasonable call and I would rather you make it than have it implied by my first framing.

2. **Ordering has one real divergence from house style.** The gate goes first, ahead of the settings read. `documents.py` and `keystones.py` run `auth.enforce_tenant(...)` first instead, so a read-only key aimed at another tenant gets a tenant refusal there and `READ_ONLY_CREDENTIAL` here. Both refuse and no test pins the precedence, but if you want them consistent, say which way and I will follow it.

erni-a (approved):

PR #1346, opened 2026-09-07
fix(storage): insert entity links in one global lock order

Eldad-Caura:
Found while investigating an intermittent test failure. **This is not that fix** — see [#1345](https://github.com/caura-ai/caura/pull/1345) for the actual cause. This closes a separate, demonstrated hazard on `memory_entity_links`, plus a comment asserting the concurrency property that would have prevented it.

## The false claim

On `memory_add_entity_links`:

> Bulk-insert with `ON CONFLICT (memory_id, entity_id) DO NOTHING` so two concurrent writes targeting the same `(memory_id, entity_id)` pair don't serialise on `Lock/transactionid` (CAURA-686).

`DO NOTHING` does not do that. The clause decides what happens once a wait *resolves*, not whether there is a wait. Measured against Postgres at the shape this code uses — one connection inserts a key and holds its transaction open, a second inserts the same key:

```
B still blocked while A held its transaction: True
B's insert took 604ms (it waited for A)
```

Believing otherwise is presumably why neither multi-row insert site was ordered.

## The hazard

The wait itself is unavoidable and harmless. A **cycle** of waits is neither: two transactions covering an overlapping set of pairs in different orders each take one key and then wait on the other, and Postgres breaks the tie by killing one with `DeadlockDetectedError` — a 500 on an ordinary write, whose message says nothing about locking. Reproduced directly:

```
forward: DeadlockDetectedError: deadlock detected
DETAIL:  Process 9298 waits for ShareLock on transaction 422472; blocked by process 9299.
         Process 9299 waits for ShareLock on transaction 422471; blocked by process 9298.
```

This is production-reachable: two agents writing memories that link to an overlapping set of entities is ordinary traffic.

## The fix

`_ordered_link_rows` dedups by the conflict key and sorts on it, so every statement takes its locks in the same global order and no cycle can form. Both multi-row sites use it — `memory_add_entity_links` (caller-supplied order) and `entity_discover_cross_links` (whatever order the candidate scan returned). The third site upserts one row per transaction and cannot hold two keys, so it is left alone.

Same mitigation and the same one-line shape as `evolve_service` sorting `related_ids` before its multi-row UPDATE, which already carries a comment saying it is for exactly this reason. Sorting is on the **string** form of each id because the service is called with `UUID` objects from the pipeline and strings from the REST body, and a mixed batch would otherwise order by object identity. Dedup keeps the **first** occurrence — the row a single `DO NOTHING` statement kept anyway — so no caller-visible result changes.

## Verification

Four mutants, one at a time:

| mutant | result |
| --- | --- |
| no ordering at all | the three ordering tests fail |
| order on entity only | only the multi-memory test fails |
| last occurrence wins | only the first-row-wins test fails |
| sort the raw id values | only the mixed UUID/str test fails |

The SQL-level test asserts Postgres still cycles on crossed insert orders, so the reason this helper exists stays checkable instead of becoming folklore. It deliberately does not route through the service and **passes with or without the fix** — it pins the hazard, not the mitigation; the four mutants hold the mitigation.

It builds the cycle rather than racing for it: two keys, and a barrier so each writer has taken one before either reaches for the other's. My first version raced 40 pairs against 40 reversed ones and deadlocked on every run I tried — but whether two transactions interleave is the scheduler's business, and shipping a probabilistic assertion into the suite this came out of would have been the wrong trade. Barrier-driven it passed 5 runs out of 5 with no variance.

Full suite 6221 passed, 0 failed. `ruff` clean, mypy clean (`86 source files`), ratchet/sentinel/tenant-scope gates green.

## Scope

The intermittent suite failure that led here is a deadlock on this same table, and it lands on a **different test each time** — the unknown-field one twice, `test_p3_3_graph_fanout` on the run that produced the traceback above. But the statement in that traceback was the *per-item* upsert, which opens one transaction per row and so cannot hold two keys at once; its cycle more likely runs through the foreign-key locks on the parent `memories` / `entities` rows against a concurrent UPDATE from background enrichment. **Nothing here addresses that.** I chased the ordering hazard because it was adjacent to the evidence, and it is worth closing on its own merits, but I want to be explicit that it is not the cycle I captured.

### DIFF
--- a/core-storage-api/src/core_storage_api/services/postgres_service.py
+++ b/core-storage-api/src/core_storage_api/services/postgres_service.py
@@ -139,6 +139,37 @@ def _get_read_session_factory() -> async_sessionmaker[AsyncSession]:
     return _read_session_factory
 
 
+def _ordered_link_rows(rows: list[dict]) -> list[dict]:
+    """Dedup ``memory_entity_links`` rows by key and fix one global insert order.
+
+    A statement carrying several pairs takes its row locks in the order it is
+    given them. Two concurrent statements covering an overlapping set in
+    different orders each take one key and then wait on the other's
+    transaction — a cycle Postgres resolves by killing one of them, which
+    reaches the caller as a 500 on an ordinary write rather than as anything
+    about locking.
+
+    ``ON CONFLICT`` does not help with that, which an earlier comment on
+    ``memory_add_entity_links`` claimed it did. The clause decides what happens
+    once the wait resolves, not whether there is a wait: an inserter whose key
+    is held by an in-flight transaction blocks until that transaction ends,
+    measured at the shape used here. Sorting is the whole mitigation, and it is
+    the one ``evolve_service`` already applies to ``related_ids`` before its
+    multi-row UPDATE.
+
+    Sorted on the STRING form of each id so the order is the same for callers
+    that pass ``UUID`` objects and callers that pass strings; a mixed batch
+    would otherwise order by object identity and defeat the point.
+
+    First occurrence of a repeated key wins — the row a single ``DO NOTHING``
+    statement would have kept anyway, so deduping here changes no result.
+    """
+    by_key: dict[tuple[str, str], dict] = {}
+    for row in rows:
+        by_key.setdefault((str(row["memory_id"]), str(row["entity_id"])), row)
+    return [by_key[key] for key in sorted(by_key)]
+
+
 @asynccontextmanager
 async def get_session() -> AsyncIterator[AsyncSession]:
     """Transactional writer session; commits on success, rolls back on error."""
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@erni-a ready for review, and lower priority than [#1345](https://github.com/caura-ai/caura/pull/1345) — that one is the flake fix; this is a separate hazard found beside it.

All 14 checks green on head SHA `00fffbdd` (verified via `check-runs` on the SHA).

Two parts:

1. **A comment on `memory_add_entity_links` asserted a concurrency property the code does not have** — that `ON CONFLICT DO NOTHING` keeps concurrent writers off `Lock/transactionid`. It does not; measured, a second inserter of a held key blocked 604ms until the first transaction ended. That claim is presumably why neither multi-row insert site was ordered.

2. **Deterministic lock ordering** at the two multi-row sites, so crossed insert orders cannot form a cycle. Reproduced the deadlock directly first; same `ShareLock on transaction` signature the suite threw.

The judgement call I want to flag: I could NOT show this is the cycle behind the suite's intermittent `DeadlockDetectedError` on this table. The statement in that traceback was the per-item upsert, which opens one transaction per row and so cannot hold two keys — its cycle more likely runs through the FK locks on the parent `memories`/`entities` rows against a concurrent enrichment UPDATE. So this closes a real, production-reachable hazard on its own merits, but it is adjacent to the evidence rather than derived from it. If you'd rather not carry a concurrency change on that basis, closing it is a reasonable call and the comment correction could go in alone.

The SQL-level test passes with or without the fix by design — it pins the hazard, not the mitigation. The four mutants are what hold the mitigation.

erni-a (approved):

PR #1347, opened 2026-09-07
refactor(core-api)!: delete the legacy write path

Eldad-Caura:
Deletes `_create_memory_legacy` (499 lines), the `_USE_PIPELINE_WRITE` flag, the dispatch branch in `create_memory`, and the test scaffolding that existed to pin the flag on. `create_memory` now calls the pipeline directly.

**Net −1,229 lines.**

## Why now

The path was dormant behind a hardcoded `True` and its own comment described it as the emergency-rollback lever. [#1345](https://github.com/caura-ai/caura/pull/1345) showed that description had stopped being true in the way that matters: a missed restore in `test_pipeline_equivalence` had been routing *every later write in the suite* through it, so for an unknown period the "dormant" path was the one under test and the shipping path was not.

## The cost, stated plainly

**This removes the write-side rollback lever.** A write-path regression is now rolled back by reverting the deploy rather than by flipping a flag and shipping a hotfix.

The lever was worth less than it looked. The two paths had diverged, and not subtly — the legacy handler rejected a semantic near-duplicate with `409` where the pipeline records the same candidate as advisory metadata and returns `201`. Flipping it during an incident would have **changed** write behaviour rather than restoring it. That divergence was only discovered because it broke an unrelated test.

`_USE_PIPELINE_SEARCH` and the legacy **search** path are untouched. The two flags shared one comment, so it is split rather than deleted — the search flag's hotfix caveat (a `False` hotfix must be cut at or after the commit that moved that path onto nested `search_params`, or it 422s every search) survives intact.

## Nothing orphaned

Every module-level helper the legacy handler used — `_embed_children_or_degrade`, `_insert_children_or_degrade`, `_drop_duplicate_facts`, `_live_duplicate_hashes`, `_schedule_embed_or_reembed`, `_find_semantic_duplicate` and the rest — is also reached by the pipeline path or by tests. Checked by walking the AST for names referenced inside the deleted span and grepping each for a referent outside it, rather than by eye.

## Seven tests go

| test | why it covered nothing surviving |
| --- | --- |
| `test_legacy_path_creates_memory` | the deleted handler |
| `test_pipeline_equivalence` | compared the two paths |
| `test_the_legacy_handler_degrades_too` | the deleted handler's H-09 fix |
| `test_the_legacy_healthy_path_schedules_nothing` | its over-refusal guard |
| `test_legacy_auto_chunk_skips_a_child_whose_content_is_already_live` | exact duplicate of the pipeline sibling directly above it |
| `test_legacy_auto_chunk_writes_a_repeated_chunk_once` | same |
| `test_pipeline_overhead_bounded_relative_to_legacy` | see below |

**Coverage genuinely lost:** `tests/pipeline/test_pipeline_latency.py`, whose single test asserted the pipeline's median write latency stayed within 1.5× the legacy path's. There is no second path to measure against, so the whole file goes rather than the ratchet being quietly weakened. If write latency wants a gate it now needs an absolute budget — a different test, and a number nobody has agreed yet. **Flagging it rather than filing it, because picking that number is your call.**

## One test changed premise

`test_both_paths_share_one_degrade_policy` existed because H-09 was written twice, once per handler, and asserted both routed through the shared helpers. With one handler that pair cannot drift. It is now `test_the_degrade_policy_lives_in_one_place`: the raw child batch embed appears in exactly one place, and each helper has exactly one caller — a second caller being the shape H-09 would come back in.

## Two stale comments, resolved by removal not by reading

Two mocks in the h09 helper were attributed to the legacy path. One attribution was right — dropping the `_find_semantic_duplicate` patch leaves all 13 tests green, so it really was legacy-only. One was wrong — the `find_by_content_hash` stub is still needed, because a bare `AsyncMock` answers the exact-duplicate pre-check truthily on either path. Both were settled by removing them and running the tests.

## A hazard worth recording

Removing the `legacy` parameter from `_run` without removing the `if legacy:` branch that read it turned it into an undefined name — and `_run`'s deliberate bare `except Exception` (the escape *is* the defect, so its tests assert on `run.raised`) captured the `NameError` and reported it as *"no parent was written"*. Eleven tests failed with a message pointing nowhere near the cause. A helper that swallows exceptions in order to test them will also swallow mistakes about them.

## Not done here

`_create_memory_pipeline` keeps its name, which no longer distinguishes it from anything. Renaming reaches a source-asserting test and a dozen comments, so it is a follow-up rather than a rider on a deletion.

## Verification

- **6242 passed, 0 failed.** The −7 reconciles exactly against the table above (6244 after #1345, +5 from #1346, −7 here).
- mypy clean: core-api 203 files, core-storage-api 86 files.
- `ruff check` and `ruff format --check` clean at CI's scopes.
- Legacy-name ratchet, do-not-touch sentinel, tenant-scope gate all green.


Eldad-Caura:
@erni-a please approve

erni-a (approved):

Eldad-Caura:
@erni-a ready for review. All 14 checks green on head SHA `214d4d0c` (verified via `check-runs` on the SHA).

Net **−1,229 lines**: `_create_memory_legacy` (499), the `_USE_PIPELINE_WRITE` flag and its dispatch branch, and the scaffolding in 8 test files that existed only to pin the flag on.

Three things I want your eye on, because they are judgement calls rather than mechanics:

1. **This removes the write-side rollback lever.** That is the real cost and the PR says so up front. My argument that it was worth little: the two paths had diverged — the legacy handler 409s on a semantic near-duplicate where the pipeline returns 201 with advisory metadata — so flipping it during an incident would have changed write behaviour, not restored it. If you want a rollback capability preserved some other way, this is the moment to say so.

2. **One test's coverage is genuinely gone and I did not replace it.** `test_pipeline_latency.py` asserted the pipeline's median write latency stayed within 1.5x the legacy path's. With one path there is nothing to measure against. A replacement needs an absolute latency budget, and picking that number is yours, not mine — so I deleted the file rather than invent a threshold or quietly weaken the ratchet.

3. **`_USE_PIPELINE_SEARCH` is deliberately untouched.** The two flags shared one comment; I split it so the search flag's hotfix caveat survives verbatim. Worth confirming I read that caveat as you intended, since it is the kind of thing that only matters once.

Also: `_create_memory_pipeline` keeps a name that now distinguishes it from nothing. Renaming reaches a source-asserting test and a dozen comments, so I left it as a follow-up rather than riding it on a deletion — say if you'd rather it were in here.

erni-a (approved):

PR #1349, opened 2026-09-07
refactor(core-api): name the write entry point for what it does

Eldad-Caura:
The follow-up flagged on [#1347](https://github.com/caura-ai/caura/pull/1347): `_create_memory_pipeline` → `_run_write_pipeline`, plus the comment residue that deletion left behind.

**No behaviour change** — two executable lines differ, the `def` and its one call site. Everything else is prose.

## The name

The old name distinguished the function from `_create_memory_legacy`. With that handler deleted, "pipeline" named nothing — and a private name that looks like one of a pair sends the next reader hunting for the twin.

`_search_memories_pipeline` **keeps** its suffix, and gets no note saying why. Its twin `_search_memories_legacy` sits 115 lines below it behind a live `if _USE_PIPELINE_SEARCH or diagnostic:`, so the asymmetry explains itself where a reader would ask. My first draft did document it — in the write path, the one file nobody will open when the legacy search path is finally deleted. That is the staleness class this PR exists to clean up, so writing a fresh instance of it to explain the cleanup was the wrong trade.

## The residue — comments still describing the deleted twin

Every count here was measured with an AST pass over call sites, not read off the comment being replaced.

| helper | what it claimed | what is true |
| --- | --- | --- |
| `_live_duplicate_hashes` | paths with no dedup lookup are "auto-chunk children (both the pipeline and legacy handlers) and the atomic-fact fanout … these **three** had neither" | two: `_handle_auto_chunk_from_ctx` and `_enrich_memory_background` |
| `_embed_children_or_degrade` | "SHARED by the pipeline and legacy handlers on purpose. The legacy path is dormant, not dead — **the flag at the top of this module** documents flipping it as the emergency-rollback lever" | one caller; the flag no longer exists |
| `_create_memory_or_409` | "a helper rather than a try at each site because there are **three** of them" | one call site |

The `_embed_children_or_degrade` one is the worst, and it is my miss: #1347 renamed the *test* that covers this helper (`test_both_paths_share_one_degrade_policy` → `test_the_degrade_policy_lives_in_one_place`) and left the docstring it was paired with — a paragraph arguing for sharing a helper between two handlers, pointing at a rollback lever deleted in the same PR.

Both surviving helpers keep their factoring; the docstrings now give the reason that still applies rather than the one that expired.

Two more copies live outside `memory_service.py`, and I missed both on the first pass:

- **`tests/test_child_dedup_minting_paths.py`**'s module docstring is a near-verbatim twin of the `_live_duplicate_hashes` one, stale in both the same ways. Fixing the source copy and leaving the test copy would have made the fix half-applied — and the test copy is the one that documents the invariant.
- **`tests/pipeline/test_search_pipeline.py`**'s `_seed_memories` claimed to insert "via the legacy write path" while its body calls `create_memory`. Actively confusing in that file specifically, where "legacy path" is a live and correct term for *search*.

## One count I got wrong in a way worth naming

I changed "these **three** had neither" to "these **two**" without re-reading the sentence. It ends: *"The public bulk path does consult one …, and the single-write path has `CheckExactDuplicate`; these two had neither."* The intended antecedent is the two server-internal paths from the previous sentence — but the nearest two nouns are bulk and single-write, the paths that explicitly **do** have dedup. At "three" the number itself blocked the misreading; at "two" the numbers match the wrong antecedent and a reader lands on the inverse of the point. It now names the antecedent instead of counting it.

## Why the two functions stay separate

`create_memory` is a short prologue that ends by delegating; folding them would give one ~215-line function. The reason that decides it is mechanical rather than aesthetic, and it is now in the docstring: `test_service_layer_does_not_gate_on_reserved_types` asserts `SERVER_RESERVED_MEMORY_TYPES` does **not** appear in `create_memory`'s source. An absence assertion over a short prologue is a real contract; the same assertion over a merged body becomes a tripwire that any unrelated mention inside the write orchestration sets off.

## The one test that tracks the name

`test_create_memory_runs_the_pipeline` (A38) asserts the entry point's name appears in `inspect.getsource(create_memory)`, so it needed updating — and it is worth knowing it still has teeth. Replacing the call with a behaviour-preserving `globals()` lookup produced **exactly one failure, that test**. Its new message names both possibilities, because they need opposite responses: a rename to propagate, or `create_memory` growing its own write, in which case A38's original suspicion (MCP writes bypassing contradiction detection) is real again.

## Verification

- **6242 passed, 0 failed** — identical to #1347's count, as a rename should be.
- mypy clean: core-api, 203 files.
- `ruff check` and `ruff format --check` clean at CI's scopes for `core-api/src/` and `tests/`.
- Legacy-name ratchet, do-not-touch sentinel, tenant-scope gate green.
- The three source-counting gates are unperturbed: `test_the_degrade_policy_lives_in_one_place` counts substrings over the whole module, so prose naming a helper **with** a paren would be miscounted as a real call site. Every match in the post-change file is real code — the new comments name helpers only in double-backtick form without a paren.

## Follow-ups, not done here

- **A38 deserves a better test than a name check.** The invariant is decidable at the composition layer: `ScheduleBackgroundTasks()` appears in `build_persist_pipeline`, `build_fast_write_pipeline` and `build_strong_write_pipeline`, and is deliberately absent from `build_stm_write_pipeline`. Asserting "every write pipeline that persists an LTM row schedules background tasks, STM the one declared exception" would catch what the current test cannot — detection hoisted into a route, or a fourth write pipeline composed without the step — and survive any rename. Swapping the assertion changes what is covered, so it needs its own fails-without-the-fix check rather than riding a rename.
- **A gate for this staleness class** — a comment naming a `_symbol` that resolves to no definition — is real but not cheap. I measured the candidate rule at **32 distinct tokens / 76 occurrences** today, and most are not symbols at all (`_system`, `_aliases`, `_latency_ms` are metadata keys and log-field suffixes), so it starts at ~30 exemptions and needs its own marker convention. Neither `legacy_name_ratchet.py` (text-diffs one literal against a base tree) nor `do_not_touch_sentinel.py` (asserts strings still *exist*) can host it. Worth its own issue.

### DIFF
--- a/core-api/src/core_api/pipeline/steps/write/parallel_embed_enrich.py
+++ b/core-api/src/core_api/pipeline/steps/write/parallel_embed_enrich.py
@@ -80,7 +80,7 @@ class ParallelEmbedEnrich:
         ch = ctx.data.get("content_hash")
         resolved_write_mode = ctx.data.get("resolved_write_mode")
         # CAURA-682 Phase 1: per-phase latency capture. The summary emit
-        # lives in ``_create_memory_pipeline`` after pipeline.run; each
+        # lives in ``_run_write_pipeline`` after pipeline.run; each
         # phase deposits its duration here.
         timings: dict = ctx.data.setdefault("phase_timings", {})
 
--- a/core-api/src/core_api/services/memory_service.py
+++ b/core-api/src/core_api/services/memory_service.py
@@ -231,15 +231,15 @@ async def _live_duplicate_hashes(
 ) -> set[str]:
     """Which of ``hashes`` already have a LIVE row, in dedup scope.
 
-    The server-internal write paths — auto-chunk children (both the pipeline
-    and legacy handlers) and the atomic-fact fanout — attach a ``content_hash``
-    to every child and then insert it without ever consulting a dedup lookup.
-    The public bulk path does consult one (``existing_hashes`` +
-    ``seen_hashes`` in ``create_memories_bulk``), and the single-write path has
-    ``CheckExactDuplicate``; these three had neither. That is why prod carries
-    duplicate content-hash groups with no concurrency involved at all: the same
-    document re-chunked, or an LLM emitting the same fact twice, minted a fresh
-    row every time.
+    The server-internal write paths — auto-chunk children and the atomic-fact
+    fanout — attach a ``content_hash`` to every child and then insert it
+    without ever consulting a dedup lookup. The public bulk path does consult
+    one (``existing_hashes`` + ``seen_hashes`` in ``create_memories_bulk``),
+    and the single-write path has ``CheckExactDuplicate``; those two
+    server-internal paths had neither. That is why prod carries duplicate
+    content-hash groups with no concurrency involved at all: the same document
+    re-chunked, or an LLM emitting the same fact twice, minted a fresh row
+    every time.
 
     ``_auto_chunk_request_id()`` cannot substitute for this. It mints a fresh
     UUID per item per call, so the attempt-idempotency index
@@ -460,12 +460,10 @@ async def _embed_children_or_degrade(
     the trade: refusing loses the children AND wedges every retry, while
     degrading keeps the facts and leaves a repair queued.
 
-    SHARED by the pipeline and legacy handlers on purpose. The legacy path is
-    dormant, not dead — the flag at the top of this module documents flipping
-    it as the emergency-rollback lever — and an emergency rollback is
-    plausibly happening BECAUSE something is degraded, which is the same
-    condition that trips this. Two copies of a degrade policy is how the two
-    paths diverge; one is how they cannot.
+    One caller since #1347 deleted the legacy handler this was shared with.
+    Still a named helper: two copies of this degrade decision, only one of them
+    fixed, is exactly what H-09 was. ``test_the_degrade_policy_lives_in_one_place``
+    pins one definition and one call site.
     """
     try:
         return await get_embeddings_batch(child_texts, tenant_config, background=False)
@@ -652,10 +650,11 @@ async def _create_memory_or_409(payload: dict) -> dict:
     gate answers the duplicate visible before the write, this answers the race it
     cannot see, and a caller should not have to tell the two apart.
 
-    A helper rather than a try at each site because there are three of them (the
-    auto-chunk parent on both handlers, plus the legacy single write) and each
-    passes a long inline dict; wrapping them individually would re-indent all
-    three for no gain. It fetches the client itself so the swap is call-for-call.
+    One call site since #1347 — the auto-chunk parent. Still a helper and not
+    an inline try: the translation is the part worth naming, it has its own
+    tests driving it directly, and inlining would bury it in a try wrapped
+    around a long dict literal. It fetches the client itself so the swap is
+    call-for-call.
     """
     try:
         return await get_storage_client().create_memory(payload)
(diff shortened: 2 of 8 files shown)


Eldad-Caura:
The first CI run on `f572b8bf` failed on a test unrelated to this change, and the failure is a real pre-existing defect rather than noise. Recording it here so re-running CI doesn't bury it.

```
FAILED tests/test_consumer_enriched.py::test_embedded_handler_does_not_silently_absorb_a_missing_embedding
assert 'embedding missing on read-back' in 'Async contradiction detection failed for memory 93963fce-…'
```

**Why it is not this PR.** The diff changes two executable lines — the `def` and its one call site — and touches neither `consumer.py` nor `contradiction_detector.py`. The same 6242 tests passed in CI as passed locally; the run had 6242 passed **plus** this one failure.

**The mechanism**, reproduced rather than inferred. `tests/test_consumer_enriched.py:352,372-373`:

```python
caplog.set_level(logging.WARNING, logger="core_api.consumer")
...
errors = [r for r in caplog.records if r.levelno >= logging.ERROR]
assert "embedding missing on read-back" in errors[0].getMessage()
```

`caplog.set_level(..., logger=…)` sets the level **on that logger**, but `caplog.records` is the root handler's capture — every logger that propagates. The filter is by level only, never by logger name, so any unrelated ERROR that lands during the test takes slot `[0]`.

The intruder here is `contradiction_detector.py:317`, which is `logger.exception(...)` inside `detect_contradictions_async`'s broad `except Exception`. That is a fire-and-forget background path, so with `asyncio_default_test_loop_scope=session` its failure surfaces in whatever test happens to be running when the task completes — not the test that scheduled it.

A standalone reproduction with two `logging` calls and no application code reproduces the assertion message verbatim, including the same foreign logger name.

**The fix** is to filter by logger as well as level, so the assertion reads the record the test is about:

```python
errors = [
    r for r in caplog.records
    if r.levelno >= logging.ERROR and r.name == "core_api.consumer"
]
```

Unrelated to a rename, so it belongs in its own PR rather than riding along here.

Eldad-Caura:

All 14 checks green on `f572b8bf` (verified on the head SHA — the first run's single failure is diagnosed in the comment above and is unrelated to this diff). Two executable lines change: the `def` and its one call site.

Four judgement calls worth a second opinion rather than a rubber stamp:

1. **The name.** `_run_write_pipeline` — `_run_*` already exists in three service modules, and the function does build and run the write pipelines. It is slightly narrow: the body runs four different pipelines (stm / enrichment / fast / strong), so "the write pipeline" singular is a simplification. I judged that not worth churning further.

2. **I deliberately did NOT document why `_search_memories_pipeline` keeps its suffix.** A first draft did, and I cut it: asserting "the legacy search path still exists behind `_USE_PIPELINE_SEARCH`" inside a *write* function's docstring plants a comment that expires on a known schedule, in the one file nobody opens when that path is deleted. The twin sits 115 lines below `_search_memories_pipeline` behind a live `if _USE_PIPELINE_SEARCH or diagnostic:`, so the asymmetry explains itself where someone would ask. Tell me if you would rather have it stated somewhere explicit.

3. **`create_memory` and `_run_write_pipeline` stay separate**, and the docstring now gives a mechanical reason instead of an aesthetic one: `test_service_layer_does_not_gate_on_reserved_types` asserts `SERVER_RESERVED_MEMORY_TYPES` does not appear in `create_memory`'s source, and an absence assertion over a short prologue is a contract where the same assertion over a merged body is a tripwire. If you disagree that this is the load-bearing reason, the docstring is the thing to change.

4. **`test_consumer_enriched.py`'s real defect is left unfixed here.** One-line fix, written out in the comment above, unrelated to a rename. Happy to send it as its own PR — say which you prefer.

Also carried forward, not done: A38's test still tracks a private name; the composition-layer assertion that would pin the real invariant is spelled out in the PR body as a follow-up, because swapping it changes what is covered and needs its own fails-without-the-fix check.

erni-a (approved):

PR #1350, opened 2026-09-07
CAURA-721 fix(dedup): scope the semantic gate to the writing agent

eyal-bl:
## The defect

`memory_find_semantic_duplicate` filtered on `(tenant, fleet)`. The exact-hash gate beside it is keyed on
`(tenant, fleet, agent, content_hash)` via
`uq_memories_live_content_hash`, whose own comment states the rule both gates are meant to share:

> `agent_id` is in the key because two agents recording identical
> content are two independent observations.

Only the semantic tier disagreed. So Caura *admitted* an exact cross-agent duplicate inside a fleet and then *refused* a paraphrase of it — the near-duplicate branch was the odd one out, not the hash gate.

Reproduced through the REST routes, two agents in one fleet, `write_mode=strong`, identical content:

```
agent-a  POST /memories  fleet=shared  -> 201
agent-b  POST /memories  fleet=shared  -> 409
  reason semantic_similarity, similarity 1.0, existing_id <agent-a's row>
control: the same two writes in separate fleets -> 201, 201
```

## Why it was loss, not deduplication

Reads narrow by `filter_agent_id`. After the 409, agent-b searching with `filter_agent_id=agent-b` returns zero rows — for `scope_team` *and* `scope_agent`. The write was not merged into a shared row; it was refused on the strength of a row its author can never retrieve. A write gate must be no wider than the narrowest scope a reader may ask for.

Second defect, same cause: for a `scope_agent` candidate the 409 handed the caller the id, status and similarity of a row it has no right to read. Pinning the owner closes that with no separate redaction — the query can no longer match a foreign row, so there is no foreign id to report.

## The fix

One optional `agent_id` threaded storage -> route -> core-api wrapper -> call sites, and one extra SQL predicate. Plain equality rather than `_content_hash_fleet_scope`'s COALESCE handling: `Memory.agent_id` is `nullable=False`, so there is no NULL-vs-empty-string group to reconcile the way `fleet_id` needs.

Same fix, same reasoning, as A54 on
`memory_find_entity_overlap_candidates` — the neighbouring handler in the same router, whose comment already called this bug shape "wet-proven, not theoretical."

Call sites: `CheckSemanticDuplicate` (the step that 409s), `DetectNearDuplicate` (advisory, but `near_duplicate_of` must not name another agent's row), and `update_memory`'s content-change gate (pins the stored row's own owner — an update body carries no `agent_id`).

`create_memories_bulk` and `/ingest/commit` are untouched: they run content-hash dedup only and skip the semantic tier altogether.

## Scope and back-compat

`agent_id` defaults to `None`, meaning "do not pin" — today's exact behaviour. Existing positional callers are unchanged, and the key is sent only when set, so an unpinned call's wire body stays byte-identical. A core-api newer than its storage degrades to over-rejection rather than failing.

Deliberately NOT changed: the judge threshold, the retention direction (the older row still wins a genuine collision), and the identifier prefilter.

Known cost: multi-agent fleets will store more rows, since each agent keeps its own copy of a fact its teammates also learned. That is the correct outcome given reads can be agent-scoped, and it is already what the hash gate does. A tenant that genuinely wants fleet-wide dedup should get an explicit setting, not an accident of a missing parameter.

## Verification

- `tests/test_caura721_dedup_agent_scope.py` — 11 cases: the lookup, the wire body, and end-to-end through the route that 409s. Confirmed to fail without the source change (8 of 11; the other 3 are the back-compat and over-fix guards, which must hold in both directions).
- The load-bearing case is `test_one_agent_restating_its_own_fact_is_still_refused`: pinning the owner must not switch dedup off for the single-agent case, which is nearly all real traffic. Also asserted over REST with a distinct content hash, so only the semantic tier can be the cause.
- `test_the_owner_does_not_widen_past_the_fleet` guards the inverse mistake — trading a too-wide agent scope for a too-wide fleet scope.
- Dedup/ingest sweep: 156 passed (`test_p2_semantic_dedup`, `test_a1_16`/`17`/`18`, `test_a21_detect_near_duplicate`, `test_a1_identifier_prefilter`, `test_storage_client_routing`, `test_ingest_commit`).
- Full `tests/`: 6215 passed, 39 failed — the failure set is byte-identical to `main`'s on this machine (pre-existing FTS/relation-weight/ request-observation env failures). `core-storage-api/tests/`: 369 passed, 2 pre-existing CORS failures, likewise identical to `main`.
- `ruff check` clean, `ruff format` clean, mypy output identical to `main`'s on the changed files. `legacy_name_ratchet.py` "No new lines.", `do_not_touch_sentinel.py` all 35 strings survive, `tenant_scope_gate.py` exit 0.

## Type of Change

- [ ] Documentation update
- [ ] Other:

## Checklist

- [ ] `mypy` passes
- [ ] `pytest` passes locally

### DIFF
--- a/core-api/src/core_api/pipeline/steps/write/check_semantic_duplicate.py
+++ b/core-api/src/core_api/pipeline/steps/write/check_semantic_duplicate.py
@@ -133,6 +133,12 @@ class CheckSemanticDuplicate:
                 embedding,
                 visibility=data.visibility or "scope_team",
                 min_similarity=SEMANTIC_DEDUP_JUDGE_THRESHOLD,
+                # CAURA-721 — this is the step that 409s, so it is the one the
+                # missing owner actually cost: a candidate from another agent
+                # in the same fleet rejected a write whose author could not
+                # then read the surviving row. The same ``agent_id`` already
+                # goes to ``_enqueue_dedup_review`` further down this method.
+                agent_id=getattr(data, "agent_id", None),
             )
         )
         set_system_value(metadata, "semantic_dedup_ms", round((time.perf_counter() - t_dedup) * 1000, 1))
--- a/core-api/src/core_api/pipeline/steps/write/detect_near_duplicate.py
+++ b/core-api/src/core_api/pipeline/steps/write/detect_near_duplicate.py
@@ -90,6 +90,12 @@ class DetectNearDuplicate:
                 embedding,
                 visibility=data.visibility or "scope_team",
                 min_similarity=SEMANTIC_DEDUP_JUDGE_THRESHOLD,
+                # CAURA-721 — advisory here, never a 409, but the owner still
+                # has to be pinned: ``near_duplicate_of`` is handed to the
+                # caller as "you already recorded this", and pointing it at
+                # another agent's row makes that claim false (and, for a
+                # ``scope_agent`` candidate, names a row the caller cannot read).
+                agent_id=getattr(data, "agent_id", None),
             )
         )
         metadata["near_dup_check_ms"] = round((time.perf_counter() - t_dedup) * 1000, 1)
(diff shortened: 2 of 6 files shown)


eyal-bl:
@claude

Eldad-Caura (approved):

PR #1352, opened 2026-09-07
test: scope caplog level filters to the logger under test

Eldad-Caura:
The fix for the CI failure diagnosed on [#1349](https://github.com/caura-ai/caura/pull/1349#issuecomment-5570256412), sent separately since it has nothing to do with that rename.

`test_embedded_handler_does_not_silently_absorb_a_missing_embedding` turned CI red on a PR that touched neither the consumer nor contradiction detection. It wasn't flaky in the sense of *unlucky* — **it was reading the wrong log record.**

```
assert 'embedding missing on read-back' in 'Async contradiction detection failed for memory 93963fce…'
```

## The mechanism

`caplog.set_level(..., logger="core_api.consumer")` sets the level **on that logger**. `caplog.records` is the root handler's capture — every logger that propagates. The assertion filtered on level alone and then indexed `[0]`, so any unrelated ERROR captured during the test was read instead.

The intruder is `contradiction_detector.py`'s `logger.exception` inside `detect_contradictions_async`'s broad `except Exception`. Fire-and-forget, so with `asyncio_default_test_loop_scope=session` its failure surfaces in whatever test is running when the task completes — not the one that scheduled it. That's why re-running went green and why the test looked unstable rather than wrong.

## A second site, with a wider exposure

`test_permanent_error_logs_once_then_debug_until_next_success` in `tests/test_ranking_remote.py` has the same defect and less margin: `caplog.at_level("DEBUG")` takes no logger argument, so it lowers the level on the **root** logger, and the test then asserts `len(first) == 1` and `assert not [...]`. One foreign ERROR is the difference between green and red for either.

## The fix

Both files filter by logger name through a small per-file helper, taking the name off the module under test — `consumer.logger.name`, `svc_mod.logger.name` — rather than a string literal, so a module rename fails loudly instead of silently emptying the filter.

## Teeth

Each fix has a test that fails without it, standing in for the intruder deterministically instead of waiting for the interleaving that produced it. Removing just the `.name ==` predicate fails **exactly one test per file, the new one**, with the foreign record shown ahead of the real one:

```
tests/test_consumer_enriched.py: 1 failed, 10 passed
  expected only the consumer's own ERROR; got
  [('core_api.services.contradiction_detector', 'Async contradiction detection failed…'),
   ('core_api.consumer', 'memory-embedded: embedding missing on read-back; ack-dropping')]

tests/test_ranking_remote.py: 1 failed, 23 passed
  the count must exclude loggers this test is not about; got
  [('core_api.services.contradiction_detector', …), ('common.ranking._service', …)]
```

Worth noting: my first version of the ranking change had **no** teeth — with the predicate removed, all 23 tests still passed, because in a single-file run no foreign record happens to land. The hardening was correct but unproven, so I added the deterministic intruder there too rather than ship an assertion nothing exercises.

## Blast radius, surveyed rather than guessed

Fifteen comprehensions across seven files filter `caplog.records` by level with no logger check. They are not equally broken, and the difference is what the test does with the result:

| | sites | |
| --- | --- | --- |
| **fixed here** | 4 | `consumer_enriched:369`, `ranking_remote:257`, `:262`, `:265` |
| **unharmed** | 8 | match on message content — `any("text" in r.getMessage() …)` — which a foreign record cannot satisfy. Weaker than they could be, never red. |
| **can pass vacuously** | 2 | `d7_parallel_embed:213`, `:244` assert a level-only list is non-empty, so a foreign WARNING satisfies them. Wrong pass, never a wrong failure. |
| **can go red, not mechanical** | 1 | see below |

**`test_h09_autochunk_embed_degrade.py:390`** asserts `errors == []` over a level-only filter, so it is breakable exactly the same way. I did **not** fix it: scoping it means deciding which module's ERRORs the absence claim covers, and the test doesn't say. Narrowing to `memory_service` would drop an ERROR raised from the storage client, which is arguably inside "a duplicate refusal must not raise ERRORs". That is a judgement about intent, not a mechanical fix, so it's flagged rather than guessed at.

## Verification

- **6255 passed, 0 failed** (2 of those are the new tests).

### DIFF
--- a/tests/test_consumer_enriched.py
+++ b/tests/test_consumer_enriched.py
@@ -30,6 +30,22 @@ from common.events.base import Event
 from common.events.topics import Topics
 from core_api import consumer
 
+
+def _consumer_errors(caplog) -> list[logging.LogRecord]:
+    """The ERROR records this module's tests are about.
+
+    ``caplog.records`` is the root handler's capture, so it holds records from
+    every logger that propagates — including background tasks unrelated to the
+    test. Filtering on level alone lets those through, which matters wherever a
+    test indexes or counts the result.
+    """
+    return [
+        r
+        for r in caplog.records
+        if r.levelno >= logging.ERROR and r.name == consumer.logger.name
+    ]
+
+
 pytestmark = pytest.mark.asyncio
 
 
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@erni-a pelase approve

erni-a (approved):

PR #1353, opened 2026-09-07
test: scope h09's ERROR-absence claim to the task that emitted the record

Eldad-Caura:
The site [#1352](https://github.com/caura-ai/caura/pull/1352) flagged and deliberately left alone. `test_a_duplicate_refusal_reports_no_unrepairable_rows` asserts `errors == []` over a filter on level alone, so any foreign ERROR captured during it turns the build **red** — the same failure mode that took CI down on [#1349](https://github.com/caura-ai/caura/pull/1349), and worse than the message-matching filters #1352 left in place, which merely get weaker.

## #1352 was right to skip it — and the obvious fix would have been wrong

#1352's stated reason was that scoping by logger name needs a judgement the test doesn't make. Measuring the path shows that judgement would have gone badly:

| logger | role |
| --- | --- |
| `core_api.services.memory_service` | code under test |
| `core_api.pipeline.runner` | **also** code under test |
| `core_api.services.contradiction_detector` | the intruder |

All three share the `core_api` prefix. There is no name-based rule that separates them — and scoping to `memory_service`, the intuitive choice, would have **silently dropped a real ERROR from the pipeline runner**, which is squarely inside "a duplicate refusal must not raise ERRORs". A fix that narrows an absence assertion is worse than the flake it removes.

## The discriminator that actually matches the question

`LogRecord.taskName` (3.12+) names the asyncio task that emitted the record, which is what the test is really asking: *was this logged by the work I awaited?*

Measured on this path before relying on it: every record carries the test's own task name and **no other task name appears**, so the awaited code crosses no task boundary — no `gather` splitting records into child tasks — and the filter loses nothing. It also needs no allowlist to rot as pipeline steps are added.

```python
me = asyncio.current_task().get_name()
return [r for r in caplog.records if r.levelname == "ERROR" and r.taskName == me]
```

`r.taskName` is read directly rather than through `getattr(r, "taskName", None)`. On a runtime without the attribute, an AttributeError is a far better outcome than an absence assertion that passes because every record was dropped.

## Teeth

The new test schedules the intruder as its own task — both what makes it distinguishable and what gives the test teeth. Removing just the `r.taskName == me` predicate fails **exactly one test, the new one**, naming the foreign task:

```
tests/test_h09_autochunk_embed_degrade.py: 1 failed, 13 passed
  another task's ERROR was attributed to this one:
  [('core_api.services.contradiction_detector', 'Task-38')]
```

It carries an anti-vacuity assertion in the style of the sibling above it: the stand-in really did reach the capture, so the filter is what excluded it rather than a record that never arrived. Without that, the test would pass just as well if the intruder had never been logged.

## Verification

- **6256 passed, 0 failed** (one new test).

## Left alone, deliberately

`tests/test_d7_parallel_embed_per_task_timeout.py:213` and `:244` assert a level-only list is non-empty, so a foreign WARNING satisfies them vacuously. They can pass wrongly, never fail wrongly — not a build hazard, and a change there is test-strengthening rather than a flake fix. That closes out the survey in #1352: 15 level-only filters, 4 fixed there, 1 fixed here, 8 unharmed, 2 vacuous-but-harmless.

Worth noting the root cause is still upstream of all of this: background tasks outliving the test that scheduled them, on a session-scoped event loop. Every fix in this series makes an assertion honest about what it can see; none of them stop the leak.

### DIFF
--- a/tests/test_h09_autochunk_embed_degrade.py
+++ b/tests/test_h09_autochunk_embed_degrade.py
@@ -35,6 +35,8 @@ committed, which inverts the trade.
 
 from __future__ import annotations
 
+import asyncio
+import logging
 import uuid
 from datetime import UTC, datetime
 from types import SimpleNamespace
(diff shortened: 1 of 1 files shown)


Eldad-Caura:
@claude

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #1354, opened 2026-09-07
test: drain a test's background tasks before the next one starts

Eldad-Caura:
The root cause behind [#1349](https://github.com/caura-ai/caura/pull/1349)'s red build, and behind the four assertions [#1352](https://github.com/caura-ai/caura/pull/1352) and [#1353](https://github.com/caura-ai/caura/pull/1353) had to teach to ignore log records they never emitted.

`track_task` registers every background task in `core_api.tasks._background_tasks` and nothing awaits them, while `asyncio_default_test_loop_scope = session` keeps one event loop for the whole run. So a task scheduled by one test keeps running through the tests that follow it.

## Measured, not assumed

| | |
| --- | --- |
| tests starting with an earlier test's tasks running | **2683** of 6255 |
| inherited-task observations (worst single test: 9) | 6860 |
| distinct tasks outliving the test that created them | 207 |
| tests survived by the longest-lived one | **746** |
| tasks that never finished at all | 2 |

My first probe reported 5947 pending tasks, which was wrong — it counted the global set at every teardown, so one lingering task was re-counted by every later test. The giveaway was `test_cap_is_below_asyncpg_safe_window`, a pure constant assertion, showing 8 pending tasks: they were never its own. Tracking task identity gave the real number, 207.

Every leaked coroutine observed arrived through `track_task` — `tracked_task` (2277 tests) and `_track_recalls_background` (22) — so the production set is a complete handle on them.

## What it costs today

- **Log records land in a later test's `caplog`.** Exactly how #1349 went red on a rename that touched none of the code involved.
- **`tracked_task`'s failure path calls `get_storage_client()`.** Fire that after `_patch_storage_client` restores the original client and it memoises a *real* client into the module singleton, aimed at a storage server no test runs — every later test then pays in connection errors and retries.

## The fix

A short grace period for work that is nearly done, then cancel.

**It cancels rather than waits because 2 of the 207 never finish** — an unconditional await would hang the run. That is also why the existing hand-rolled drain in `tests/test_governance_bulk_inline_remediation.py` is left exactly as it is: it awaits without a timeout *inside the test body* because that test wants completion, and this fixture promises isolation, not completion.

**The dependency on `_patch_storage_client` is for ordering, not for a value.** It makes the drain set up second and therefore tear down *first*, so drained tasks still see the in-process ASGI bridge rather than reaching for a real client. Verified rather than asserted: at drain time `sc_mod._client` is still a `CoreStorageClient`, not `NoneType`, which is what a reversed order would show.

**The loop scope is deliberately untouched.** Function-scoped loops would isolate tasks completely, but session-scoped fixtures are bound to the session loop — a far larger change than this.

## Result

**2683 → 4** tests start with a foreign task running; 6860 → 6 observations; worst case 9 → 2.

The residue is structural, not a missed case: the drain can only catch what exists when its turn in finalisation comes, so anything scheduled later in the chain escapes by construction. All 4 are in `tests/test_bulk_write_mode.py` and deserve their own look.

## Teeth

`tests/test_background_task_isolation.py` is an **ordered pair**, because "the previous test's work is not still running" cannot be asserted inside a single test. The first schedules a task waiting on an `Event` nothing sets — so only cancellation can end it, and no timing luck can pass the second test. It also asserts its own preconditions, so it cannot silently stop leaking.

With the fixture switched off, **exactly one test fails — the second** — naming the pending task:

```
AssertionError: a task scheduled by an earlier test is still running during this one;
its log records, storage calls and failures will be attributed here:
["<Task pending name='Task-6' coro=<test_...>"]
```

The order dependence is documented in the module docstring, along with the fact that the suite installs no order-randomising plugin (only pytest-cov, pytest-asyncio and anyio).

## Cost

**None measurable.** 6257 passed in 155.65s, against 153.17s for the same suite without the fixture — inside run-to-run noise.

## Verification

- **6257 passed, 0 failed** (2 new tests).

## Not covered

Task creators that bypass `track_task` — `storage_client.py:357`, `providers/inprocess_queue.py:32`, `routes/interview.py:204`, and the audit/usage/capability flusher loops — are not in `_background_tasks` and so are not drained. None showed up in the measurement, so this is a statement of scope rather than a known gap.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -6,6 +6,7 @@ PostgreSQL instance with pgvector — configure via TEST_DATABASE_URL env var
 or the defaults below.
 """
 
+import asyncio
 import os
 import uuid
 from datetime import UTC, datetime, timedelta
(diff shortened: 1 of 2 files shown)


Eldad-Caura:

All 14 checks green on `a7aab6d4` (verified on the head SHA). This is the root cause under #1349 / #1352 / #1353, so it is the one of the four worth real scrutiny — it changes teardown for every test in the tree.

Four judgement calls I would rather have checked than trusted:

1. **Cancel, not await.** 2 of the 207 leaked tasks never finish, so an unconditional await hangs the run. The fixture gives a 0.25s grace and then cancels. That means a test which today passes *because* a leaked task completed during a later test will now fail — I found none, but that is the class of breakage to expect from this, and it would show up as a genuine order-dependence rather than as a flake.

2. **`_DRAIN_GRACE_SECONDS = 0.25` is a number I chose, not measured.** The suite shows no runtime cost at that value, but I have no principled basis for 0.25 over 0.1 or 1.0. If you would rather it were 0 (cancel immediately, no grace), that is a one-line change and arguably the more honest contract, since nothing promises completion anyway.

3. **The ordering dependency is load-bearing and invisible.** `_drain_background_tasks(_patch_storage_client)` takes that fixture purely so it tears down first; the parameter is never used. Anyone who "cleans up the unused argument" reintroduces the singleton-poisoning bug silently — no test catches that, because the poisoning needs a task that fails at the wrong moment. If you want a guard for it, that is worth its own issue; I did not want to invent a test that asserts fixture ordering by side effect.

4. **The 4 remaining leaks are all in `tests/test_bulk_write_mode.py`.** I explained why a residue exists structurally (anything scheduled later in the finalisation chain than the drain's turn escapes it) but I did not chase why those 4 specifically. If it matters, it is a small follow-up.

Also flagged in the body: creators that bypass `track_task` are not drained. None appeared in the measurement, so that is scope rather than a gap — but it does mean this fixture is not a general "no background work escapes" guarantee, only a `track_task` one.

erni-a (approved):

PR #1357, opened 2026-09-07
CAURA-722 feat(diagnostic): report why entity retrieval stayed out of…

eyal-bl:
… a query

A benchmark run reported the ENTITY_LOOKUP strategy firing on 0 of 589 queries, and nothing in the API could say which of three unrelated causes it was. They belong to different owners:

  * entity FTS matched nothing            -> extraction / linking
  * it matched too much and was declined  -> threshold tuning (CAURA-698)
  * it matched, pool under-filled top_k   -> the fill rule itself

`SearchDiagnostic` exposed `retrieval_strategy` and nothing else about the entity path, so all three looked identical from outside and only server logs told them apart. That ambiguity is what blocked the follow-up.

## 1. Two fields that separate the causes

  entity_matches: int | None    entities matched in entity FTS
  entity_match_declined: bool   the over-broad refusal

Both were already computed inside the request and reached only the logs. Purely additive; `items` untouched.

Surfaced on all three diagnostic-bearing paths, not just REST `/search`: `/recall` (three sites) and MCP already report `retrieval_strategy`, so omitting these there would preserve the same ambiguity in the two places the follow-up is most likely to use — MCP `caura_recall` especially, since it has no `top_k` cap and is where an investigation goes once REST's limit binds.

Two properties the field set is load-bearing on:

**`None` is not `0`.** `None` means the FTS never ran — retrieval off by org setting, no entity-shaped tokens surviving the 288-stopword tokenizer, or a swallowed failure. `0` means it ran and matched nothing. Only the second is evidence about the entity index, so an `or 0` anywhere on the path destroys the field's purpose. Three tests pin it.

**The count is captured before the decline empties the list.** `classify_query` sets `matched_ids = []` immediately after flagging the over-broad decline, so reading it later reports 0 — precisely the value it must never be confused with. Verified live: a query matching 802 entities reports 802 with `declined=True`, not 0.

## 2. The per-row score factors were never being returned

Found while checking the above. `SearchDiagnostic.all_candidates` is documented as "the full widened candidate set with per-row score factors" and names nine. Five were `None` on every row of every query.

Not an old deployment — reproduced with both services from source:

```
score            null on 0/3 rows   sample=0.0
vec_sim          null on 0/3 rows   sample=0.536
status_penalty   null on 0/3 rows   sample=1.0
entity_boost     null on 3/3 rows   sample=None
freshness        null on 3/3 rows   sample=None
recall_boost     null on 3/3 rows   sample=None
temporal_boost   null on 3/3 rows   sample=None
```

Cause: `memory_scored_search` computes each factor, folds it into `score`, and then the scored CTE's own select list does not name it, so it dies at the CTE boundary. core-api reads all five by name in `execute_scored_search` and the diagnostic rounds each through `_f()` — three layers wired to receive values the query never sent.

**Ranking was never affected.** The factors did their job; they are in `score` and results are correct. What was broken is explaining a result, which is the diagnostic's only job.

The cost was not hypothetical, and it is why this is fixed here rather than filed: `entity_boost` is the only place the entity *boost* is observable, as distinct from the short-circuit `retrieval_strategy` reports. A null there reads as "this signal did not apply" — which is how the benchmark concluded the entity subsystem contributed to 0 of 589 queries, when the number that would have contradicted it was blank for reasons having nothing to do with entities. Without this half, the fields in part 1 still could not answer the question they were added for.

Fix: name the four cheap factors in the CTE select list, reference them by column in the outer query, carry them through the grouping namespace and the router. No new computation — every expression already existed because `score` needs it. Both union branches derive from the same statement so they stay column-compatible, and dedup is unchanged (the factors are deterministic per `mem_id`, so rows that collapsed before still collapse).

Always returned rather than gated on a diagnostic flag: storage takes no such flag, and plumbing one through to save four floats per row — on a response already carrying each row's full content — costs more than it saves.

Nulls are preserved rather than defaulted to the 1.0 identity. The query always produces a value, so a missing attribute means something upstream changed shape, and answering "no boost applied" on that basis is the exact failure being removed.

## 3. `fts_score`, and the render ratchet it moves

All five factors are projected, `fts_score` included — which moves `test_fts_score_single_render`'s constants from **9 -> 10** `ts_rank_cd` renders and **20 -> 21** `plainto_tsquery` renders. Flagging it because that guard exists precisely so this is a decision and not a side effect: "a rise means someone added a reference and put the per-row cost back ... re-measure and move the number deliberately rather than loosening the check."

Re-measured. The rise is +1, not +2 despite the statement appearing in two union branches, because `reserved_stmt` already ordered by `fts_score`; once the factor is a real column that ORDER BY references the label instead of re-pasting the expression.

The price, scaled from the measurement recorded on `_saturate_rank` (halving 18 -> 9 renders bought 92.0ms -> 56.4ms at 11,505 matching rows on a 31,446-memory corpus): roughly **4ms per render** at that worst case, less on smaller matches, and smaller again end-to-end since the same query also pays six pgvector distance computations per row. Against a search p50 near 1,300ms that is well under a percent, in exchange for the fifth declared factor actually arriving.

The render also lands inside the candidate CTE rather than in the outer query past the entity-link fanout, so it is evaluated per candidate row and not per joined row.

This does not concede the ratchet. The inner-projection work `_saturate_rank` describes still takes renders to 1 and would make every factor free to project; when it lands the constant drops and this reference costs nothing. The guard keeps checking both directions.

## 4. scripts/entity_retrieval_probe.py

A read-only probe that reads both entity mechanisms — the short-circuit and the boost — plus the new fields. Only call is `POST /search` with `diagnostic: true`, documented as inspection with no `recall_count` bump. Takes `--query`, `--queries-file`, or `--amb-run` (reads an AMB results .json/.json.gz directly, so re-probing a recorded run is one command).

It refuses to overclaim in either direction:

  * `entity_boost == 1.0` is the multiplicative identity — present but inert — and is not counted as a contribution.
  * an all-null `entity_boost` column is a THIRD state, not a zero: it means the factors were not reported, so the boost question is unanswered. Those queries leave the denominator and are flagged loudly. Reporting 0% from a response that never carried the number is the exact error the script exists to catch.
  * declined / zero-match / never-ran queries stay IN the denominator — those are determinate "contributed nothing", and excusing them would flatter the result.

`--top-k` carries a warning: the short-circuit needs the entity pool to fill it, so a large `top_k` against a small store makes `entity_lookup` effectively unreachable — the likeliest reason the benchmark saw zero (top_k=20 against a ~24-row per-persona store).

## Verification

Run against core-api from this branch, all four diagnostic states distinguished on real data:

```
strategy          matches  decl   query
entity_lookup           1     .   tokyo review
semantic_search       802     Y   comet incident timeline
semantic_search         -     .   of the a to be is it
semantic_search         0     .   zzzqqq flibbertigibbet nonexistent
```

Row 2 is the ordering fix earning itself: 802, not 0. Rows 3 and 4 are the None/0 distinction on a live wire.

- `tests/test_caura722_entity_diagnostic.py` — 8 cases. 5 fail without the source change; the other 3 are contract guards (diagnostics must not change `items`; a swallowed FTS failure and disabled retrieval must not claim 0) which must hold in both directions.
- The factor test asks the database whether the migration-001 tsvector trigger exists rather than assuming. `fts_score` is `ts_rank_cd(search_vector, ...)`, CI runs `alembic upgrade head` so it is a number there, and `conftest` builds the schema with `Base.metadata.create_all`, which makes columns but not triggers, so it is legitimately NULL on a developer machine. Asserting non-null unconditionally would pass CI and fail every local run; asserting nothing would let the projection regress in CI unseen. Both branches were exercised — against a create_all database and against one with the trigger applied — and pass.
- `core-storage-api/tests/test_fts_score_single_render.py`: 9 passed at the new constants.
- Targeted sweep, post-rebase: 91 passed (`test_d12_search_diagnostics`, `test_entity_lookup_short_circuit`, `test_skip_entity_boost_on_decline`, `test_mcp_recall`, `test_c4_recall_items_alias`, `test_search_recall_tracked_flag`, `test_search_caller_identity`, `test_p2_semantic_dedup`).
- `core-storage-api/tests/`: 369 passed, 2 pre-existing CORS failures — same set as `main`. The render guard passes untouched.
- Full `tests/`: 39 failures, byte-identical to `main`'s set on this machine (pre-existing FTS / relation-weight / request-observation env failures).
- ruff check and format clean; mypy output identical to `main` on every changed file. `legacy_name_ratchet.py` "No new lines.", `do_not_touch_sentinel.py` all 35 protected strings survive, `tenant_scope_gate.py` exit 0.

Side finding for whoever debugs the local FTS failures next: because `search_vector` is trigger-filled and `create_all` makes no triggers, that column and every rank over it are NULL on a developer database no matter what this change does. With the trigger applied, `fts_score` returns 0.0 and `score` moves 0.0 -> 0.3939 as FTS starts contributing.

Not touched, per the finding's own guardrail: `ENTITY_LOOKUP_MAX_MATCHES`, the entity bypass, and the under-fill rule. Ship the fields, read the counts, then decide.

## Type of Change

- [ ] Documentation update
- [ ] Other:

## Checklist

- [ ] `mypy` passes
- [ ] `pytest` passes locally

### DIFF
--- a/core-api/openapi.broker.json
+++ b/core-api/openapi.broker.json
@@ -1514,6 +1514,24 @@
             "title": "Candidates Considered",
             "type": "integer"
           },
+          "entity_match_declined": {
+            "default": false,
+            "description": "True when the entity match was refused as over-broad (> ENTITY_LOOKUP_MAX_MATCHES). This is the only decline that also suppresses the per-row entity boost; an under-filled pool falls through with the boost still applied and reports False here.",
+            "title": "Entity Match Declined",
+            "type": "boolean"
+          },
+          "entity_matches": {
+            "anyOf": [
+              {
+                "type": "integer"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "description": "Entities the query's tokens matched in entity FTS, counted before the over-broad decline empties the set. None means the lookup never ran (entity retrieval off, no entity-shaped tokens, or a swallowed failure) — distinct from 0, which means it ran and matched nothing.",
+            "title": "Entity Matches"
+          },
           "excluded_below_min_similarity": {
             "default": 0,
             "title": "Excluded Below Min Similarity",
--- a/core-api/src/core_api/mcp_server.py
+++ b/core-api/src/core_api/mcp_server.py
@@ -1138,6 +1138,14 @@ async def caura_recall(
                     for k, v in (diagnostic_ctx.get("search_params", {}) or {}).items()
                 },
                 "all_candidates": diagnostic_ctx.get("all_candidates", []) or [],
+                # CAURA-722 — kept in step with the REST ``SearchDiagnostic``,
+                # which is where these are documented. MCP matters more than
+                # parity-for-its-own-sake here: ``caura_recall`` has no
+                # ``top_k`` cap, so it is the path an investigation reaches for
+                # once REST's limit binds, and the entity question is exactly
+                # the kind that gets probed there.
+                "entity_matches": diagnostic_ctx.get("entity_matches"),
+                "entity_match_declined": diagnostic_ctx.get("entity_match_declined", False),
             }
         if payload["truncated"]:
             # Kept alongside the structured fields above, not replaced by them:
--- a/core-api/src/core_api/openapi_responses.py
+++ b/core-api/src/core_api/openapi_responses.py
@@ -93,6 +93,10 @@ class RecallDiagnostic(BaseModel):
     top_k_used: int | None
     retrieval_strategy: str | None
     search_params: dict
+    # CAURA-722 — documented on ``SearchDiagnostic``. ``None`` on the count
+    # means entity FTS never ran, which is a different answer from ``0``.
+    entity_matches: int | None = None
+    entity_match_declined: bool = False
 
 
 class RecallResponse(BaseModel):
--- a/core-api/src/core_api/pipeline/steps/search/classify_query.py
+++ b/core-api/src/core_api/pipeline/steps/search/classify_query.py
@@ -97,6 +97,19 @@ class ClassifyQuery:
                 sc = get_storage_client()
                 matched_ids = await self._entity_fts(sc, tokens, tenant_id, fleet_ids)
 
+                # CAURA-722 — record the count HERE, before the over-broad
+                # branch below empties ``matched_ids``. Reading it later would
+                # report 0 for the decline, which is the one value that must
+                # not be confused with "FTS matched nothing" — those two send
+                # the follow-up work to different teams (threshold tuning vs
+                # entity extraction/linking).
+                #
+                # Set unconditionally rather than under ``if diagnostic``: it
+                # is one ``len()``, and gating it would put a second
+                # diagnostic-only branch in a hot path for no measurable gain.
+                # Exposure stays gated at the route.
+                ctx.data["entity_matches"] = len(matched_ids)
+
                 # CAURA-698: over-broad match → not a "name a specific entity"
                 # query. The precision argument for entity_lookup breaks down
                 # at high match counts: graph expansion + memory linking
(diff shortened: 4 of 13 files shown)


eyal-bl:
@claude

eyal-bl:
@claude

ran-taig:
@erni-a

erni-a (approved):

PR #1359, opened 2026-09-07
docs(observability): the query contract for contradiction quality (D4)

arkash20:
## The dashboards under-report, and the bias runs one way

Contradiction detection is **post-commit and async** on both routes — and the entity route can't even start until entity extraction resolves subjects, a second async hop with an LLM in it. A write returns `201` long before either concludes (**4–15 s per run** measured locally on Gemini).

Anything sampling at "write + a few seconds" counts a verdict that hasn't happened yet and records it as *no contradiction*. That failure is silent and **directional**: it can only ever under-count conflicts — so the dashboard looks healthiest exactly when detection is slowest or failing.

## What's in this PR

The **measurement contract**, not a code change. The dashboards live in Datadog, not in this repo.

- **Event-arrival (preferred)** — count from D3's completion log, which fires on *every* concluded run including empty ones. Its **absence** is itself a signal: a write with no completion event means detection never concluded (lock contention, early return, crashed worker) — a different failure from "found nothing", and the two were previously indistinguishable. Yield and loss stay separate series; collapsing them is what made the old dashboard unreadable.
- **Fixed delay ≥ 60 s** if a query can't join the event. Not a tuning knob — it's above the observed p99 for extraction + detection with an LLM in path.
- **What not to use:** `memory_compute_health_stats.contradiction_count` is a **stock** measure (rows currently `outdated`/`conflicted`), so it tracks corpus age and write volume, not whether detection works. Useful hygiene number, misleading quality number.

## Why no code

I checked: the in-repo query has no settle-time defect — it's a snapshot of state, and snapshots don't have this problem. The defect is in how the Datadog dashboards query, and **that edit needs someone with access to that account.** This PR gives them the contract to implement.

Refs: canonical D4.

### DIFF
--- a/docs/observability/measuring-contradiction-quality.md
+++ b/docs/observability/measuring-contradiction-quality.md
@@ -1,0 +1,70 @@
+# Measuring contradiction quality without lying to yourself
+
+**D4.** The contradiction-quality dashboards under-report, because they sample
+too soon after the write. This is how to measure it correctly. The dashboards
+themselves live in Datadog, not in this repo — this is the query contract they
+should implement.
+
+## Why fixed-delay sampling under-reports
+
+Contradiction detection is **post-commit and asynchronous**, on both routes:
+
+* the content route runs after the write returns;
+* the entity route runs only after entity extraction has resolved subjects,
+  which is a second async hop and involves an LLM call.
+
+A write returns `201` long before either has concluded. Measured locally, a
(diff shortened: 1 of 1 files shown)


arkash20:
Rebased onto main. CI green.

@Eldad-Caura please approve

Eldad-Caura (approved):

Issue #869, opened 2026-08-22
Background tasks outlive the test that started them, and go on retrying a storage-api that is not there


Eldad-Caura:
Background tasks outlive the test that started them, and go on retrying a storage-api that is not there

## What happens

`detect_contradictions_async` is scheduled post-commit as fire-and-forget. Several tests reach that path, return, and leave the task running. It then calls `storage_client.POST /memories/similar-candidates`, gets `ConnectError` because no storage-api exists in the test environment, and works through `common.http_retry`'s five attempts — ~3 seconds of backoff — inside whatever tests happen to run next.

It is visible in every full-suite run as

```
RuntimeWarning: coroutine 'detect_contradictions_async' was never awaited
RuntimeWarning: coroutine 'tracked_task' was never awaited
```

and it was directly observed in caura#863's CI log, which captured this line into an unrelated test's report:

```
WARNING  common.http_retry:http_retry.py:145 storage_client.POST /memories/similar-candidates:
         ConnectError on attempt 3/5, retrying in 0.74s
```

## Why it is worth fixing rather than tolerating

`pytest.ini` sets `asyncio_default_test_loop_scope = session`, so there is one event loop for the whole run and a leaked task genuinely keeps running across test boundaries. Two consequences:

1. **Its work lands in other tests' reports.** The line above was captured under `test_llm_retry_after.py::TestTheHintIsHonoured::test_no_hint_keeps_the_linear_backoff`, which has nothing to do with storage or contradictions.

2. **Nothing runs the loop during a sync test**, so a block of them acts as an accumulator: the leaked task's timers expire unattended and the next async test drains them all at once. That makes the exposure order-dependent and therefore sensitive to unrelated PRs. #863 added 24 sync subprocess tests immediately before `test_llm_retry_after.py` in collection order, and the failure appeared on that PR.

caura#868 fixes the specific damage — a globally faked `asyncio.sleep` recording the leaked task's backoff as the code-under-test's own. It deliberately does **not** stop the leak, because what the suite should do about background tasks is a fixture decision, not a flake fix.

## Options

1. **Fail the run on a leaked task.** An autouse fixture asserting `asyncio.all_tasks()` is empty at teardown. Strongest signal, and it will light up broadly on the first run — the real question is how many sites there are.
2. **Cancel and drain at teardown**, without failing. Removes the cross-test effect, keeps the leak invisible.
[...]

PR #1360, opened 2026-09-07
test: fail the run when a coroutine is created and dropped

Eldad-Caura:
Closes the item left open on [#869](https://github.com/caura-ai/caura/issues/869) after #1354–#1356 fixed the leak that issue is titled after. This is the other half: coroutines that are created and then **dropped**, which the drain cannot touch because there is no task to cancel.

`run_contradiction_detection` returns its coroutine eagerly, on purpose, so a caller that returns or raises between the call and its `track_task(...)` would lose contradiction detection with nothing logged anywhere. Two gates, one static and one at runtime.

## The static half, which is the cheaper gate

`contradiction/dispatch.py` declared `-> Awaitable[None]`. **mypy's `unused-coroutine` check fires only for `Coroutine`**, so a call site that created this and dropped it type-checked clean. Narrowed to `Coroutine[Any, Any, None]` — also the accurate type, since all three branches return coroutines. No runtime change.

It has teeth: dropping the `await` at `consumer.py:165` now fails mypy with `Value of type "Coroutine[Any, Any, None]" must be used  [unused-coroutine]`.

This came out of the review and is the better altitude — the runtime gate below defends the same hazard, but only in paths a test happens to exercise, after the fact. Caveat: 7 of the 9 caller modules sit on core-api's `ignore_errors` list, so today this covers `consumer.py` plus every future non-exempt caller. That list's own comment calls each line "a to-do, not a policy", so it's a ratchet rather than full coverage.

## The runtime half, and why my first proposal was inert

On #869 I said `filterwarnings = error` would do this. **It does nothing** — the offending test still passes. The RuntimeWarning is raised from the coroutine's `__del__`, exceptions there are never propagated, so CPython routes it to `sys.unraisablehook` and pytest re-reports it as a `PytestUnraisableExceptionWarning` — a warning again.

| gate | result |
| --- | --- |
| coroutine `RuntimeWarning` → error, alone | 2 passed |
| unraisable wrapper → error, alone | 2 passed |
| both | 1 failed |
| an unrelated `__del__` that raises, under both | 1 passed — stays scoped |

Two details now recorded at the gate because they cost me time: **no `:` may appear in a message pattern** (the filter string is split on it), and **`-W` on the command line treats the message as a literal while ini `filterwarnings` treats it as a regex** — so a command-line probe of a regex filter silently matches nothing.

`AsyncMockMixin` is excluded: that is a mock's *own* internal coroutine, not this codebase's work. Four tests in `test_crystallization_must_not_wedge.py` shed one from a bare `AsyncMock()` whose `return_value` is itself an `AsyncMock`, so an ordinary sync `.get()` at `crystallizer_service.py:942` returns a coroutine that correct code discards. Recorded as an exclusion, **not** as a diagnosis — my first draft asserted a mechanism I hadn't established, and the review couldn't construct one either. An AST scan over the four service trees found zero bare-statement calls to a locally-defined async function, so the exclusion isn't currently hiding a live defect.

## What the call sites had in common

Every drop came through one boundary. `track_task(coro)` receives an already-created coroutine, so a plain `MagicMock` in its place leaves it unstarted. **An `AsyncMock` does not help** — the dropped object is the *argument*, not the mock's return value. That corrects what I wrote on #869, where I said the fix was an `AsyncMock`; it would have sent whoever picked it up down the wrong path.

The fix disposes of what the stand-in receives — an idiom this repo already had in ten files. This lifts it into one documented helper and **folds twelve byte-identical local copies** in `test_embed_off_hot_path.py` into it, so the file it touches most heavily doesn't end up carrying two spellings of one fix.

Closing has to **walk the nest**: the real call is `track_task(tracked_task(detect_contradictions_async(...)))`, and closing only the wrapper strands the inner coroutine — trading one dropped coroutine for another, which is exactly what 23 remaining drops after my first pass were. The walk's limit is measured and documented in the helper: a coroutine held in a *tuple or list* local is missed. `tracked_task`'s first parameter is the coroutine itself, so every scheduler here is covered today; a future wrapper taking `*coros` would need this widened rather than trusted.

**One edit was reverted as dead.** `patch.object(memory_service, "tracked_task", ...)` does nothing for `_enrich_memory_background`, which re-imports the name from `task_tracker` at call time — the trap that `test_embed_off_hot_path.py:967` already documents. That patch was dead before this change too, so it's left as found rather than silently removed. Worth a follow-up; not worth smuggling into this one.

## Verification

- **6260 passed, 0 failed**, across five gated runs.
- mypy clean on `core-api/src`; `ruff check` / `ruff format --check` clean, **one scope per invocation as CI runs them** — combining every scope into a single `ruff` command changes config discovery and flagged 12 files that are in fact clean.
- Legacy-name ratchet exit 0; do-not-touch sentinel intact.

Each piece is load-bearing:

| experiment | result |
| --- | --- |
| helper closes only the outer coroutine | **8 failures** |
| one call site loses its `side_effect` | **1 failure** |
| gate removed, fixes kept | 6260 passed, **zero** dropped-coroutine reports |

I predicted that last row wrongly, and the difference matters: I expected the drops to reappear as warnings once the gate was gone. They don't — the fixes remove the drops themselves, and the gate only decides what happens when one occurs. So that row says the suite is clean on its own merits and the gate is masking nothing; the first two rows are what show the gate has teeth.

**Scope:** the gate covers the root `tests/` suite only. `core-operations`, `core-storage-api` and `core-worker` each have their own `pytest.ini`, and pytest resolves config from the arguments' directory — `pytest core-storage-api/tests/` reports `configfile: pytest.ini` under `core-storage-api/`. I checked this rather than assuming it, having first assumed the opposite. All three are free of dropped coroutines today, so extending the gate is a separate cheap change.

## Before this fails on someone

**Attribution follows garbage collection, not creation.** The warning fires when the coroutine is collected, routinely a later test than the one that created it — a reference held by a mock's `call_args` outlives its test. Of the 30 tests this gate failed before the fixes, roughly half were bystanders that create no coroutines at all (`test_c33_openapi_completeness`, `test_d14_rate_limited_response_param`, `test_fixture_hygiene`). Treat a failure as *"something in this run dropped a coroutine"*, not as an accusation against the named test.

**To find the real creator**, re-run the suspect files with `PYTHONTRACEMALLOC=25`, which adds the allocation traceback. Scope it to a handful of files — suite-wide it reached 1% in six minutes before I killed it. For exact blame I used a throwaway plugin forcing `gc.collect()` in each teardown, which pins the warning to its creator; that produced the final list of 23.

`pytest.ini` also records the two things *not* to reach for: an autouse fixture that disposes of stray coroutines would make the gate permanently green, since it can't tell a test double's leftover from a production drop; and a per-test `filterwarnings` marker doesn't reliably exempt anything, for the attribution reason above.

## Deliberately not done

- **A shared context manager for the 30 identical `patch("core_api.tasks.track_task", side_effect=...)` sites.** Two reviewers suggested it and it would remove real repetition, but it also removes the patch target from the line that patches it. This repo leans on grep-based audits, and `grep "core_api.tasks.track_task"` currently finds every site; behind a helper it would find none of them. Recorded as a judgement call, not an oversight.
- **Sweeping the ~21 other ad-hoc `coro.close()` stubs** across nine untouched files onto the shared helper. Correct direction, but it adds files to a 9-file diff for no behavioural change. Natural follow-up.
- **Fixing `_storage_mock` in `test_crystallization_must_not_wedge.py`** to return real dicts, which would stop the AsyncMock drops at source and make the `AsyncMockMixin` exclusion unnecessary. The better end state; out of scope here.

### DIFF
--- a/core-api/src/core_api/services/contradiction/dispatch.py
+++ b/core-api/src/core_api/services/contradiction/dispatch.py
@@ -7,7 +7,15 @@
   * ``True``  -> route through ``ContradictionEngine.evaluate_async``.
 
 **Sync on purpose.** ``run_contradiction_detection`` is a *regular* function that
-returns an awaitable. On the legacy path it calls the detector entry
+returns a coroutine.
+
+The return type is spelled ``Coroutine`` rather than ``Awaitable`` deliberately:
+mypy's ``unused-coroutine`` check fires only for ``Coroutine``, so with
+``Awaitable`` a call site that created this and then dropped it type-checked
+clean. Since the coroutine is created eagerly here, dropping it silently
+discards the detection — which is exactly the failure this annotation now makes
+a static error. Every branch below returns a coroutine, so the narrower type is
+also the accurate one. On the legacy path it calls the detector entry
 *synchronously* and returns its coroutine, so a call site that schedules the
 result — ``track_task(run_contradiction_detection(...))`` — keeps the exact
 schedule-time semantics the direct ``detect_contradictions_async(...)`` call had
@@ -24,7 +32,8 @@ differential test pins the detection outcomes.
 
 from __future__ import annotations
 
-from collections.abc import Awaitable
+from collections.abc import Coroutine
+from typing import Any
 from uuid import UUID
 
 from core_api.config import settings
@@ -44,7 +53,7 @@ def run_contradiction_detection(
     content: str | None = None,
     embedding: list[float] | None = None,
     new_memory: dict | None = None,
-) -> Awaitable[None]:
+) -> Coroutine[Any, Any, None]:
     if settings.contradiction_engine_enabled:
         return ContradictionEngine().evaluate_async(
             memory_id,
--- a/pytest.ini
+++ b/pytest.ini
@@ -6,6 +6,50 @@
 # asserted against other tests' leftovers. Both remaining markers were checked
 # against this setting; use `strict=False` on a marker that genuinely needs it.
 xfail_strict = true
+# A coroutine created and then dropped is scheduled work that silently never
+# ran. That is a live failure mode here rather than a theoretical one:
+# ``run_contradiction_detection`` is a sync function returning an awaitable,
+# eager on purpose, so a caller that returns or raises between the call and its
+# ``track_task(...)`` loses contradiction detection with nothing logged
+# anywhere. This gate is the only signal that would show it.
+#
+# BOTH lines are required and each is INERT alone — measured, not assumed. The
+# first turns the warning into an exception inside the coroutine's ``__del__``,
+# where exceptions are never propagated: CPython routes it to
+# ``sys.unraisablehook`` and pytest re-reports it as a warning. The second is
+# what turns that re-report into a failure. With only one of the two, the
+# offending test passes.
+#
+# ``AsyncMockMixin`` is excluded because that is a mock's OWN internal
+# coroutine rather than this codebase's work. Four tests in
+# ``test_crystallization_must_not_wedge.py`` shed one from a bare
+# ``AsyncMock()``; in the case traced to production the ``await`` is present
+# (``crystallizer_service.py:942``), and the exact mechanism was NOT
+# established — this is recorded as an exclusion, not as a diagnosis. What the
+# exclusion costs is a missing ``await`` on a call whose callee is mocked,
+# where that mock's own ``assert_awaited_once`` is the better check anyway; a
+# missing ``await`` on real code still reports that function's own name and
+# still fails here.
+#
+# No ':' may appear in a message pattern, because the filter string is split on
+# it — hence "ignored in.*coroutine object" and not the literal
+# "ignored in: <coroutine object".
+#
+# When this fails, the named test is often NOT the culprit: the warning fires
+# when the coroutine is collected, which can be a later test than the one that
+# created it (a reference held by a mock's ``call_args`` routinely outlives its
+# test). Find the real creator by re-running the suspect files with
+# PYTHONTRACEMALLOC=25, which adds the allocation traceback to the report.
+# Scope that to a handful of files: suite-wide it reached 1% in six minutes.
+# Two things NOT to reach for when this fails. An autouse fixture that disposes
+# of stray coroutines would make the gate permanently green -- it cannot tell a
+# test double's leftover from a production drop, which is the whole distinction
+# being gated. And a per-test ``filterwarnings`` marker does not reliably exempt
+# anything, because the failure is attributed to whichever test was running at
+# collection time rather than the one that created the coroutine.
+filterwarnings =
+    error:coroutine '(?!AsyncMockMixin)[^']*' was never awaited:RuntimeWarning
+    error:Exception ignored in.*coroutine object:pytest.PytestUnraisableExceptionWarning
 asyncio_mode = auto
 asyncio_default_fixture_loop_scope = session
 asyncio_default_test_loop_scope = session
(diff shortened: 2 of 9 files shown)


Eldad-Caura:
@erni-a ready for review — all 14 checks green on head `f88e64c7`, 9 files, 191 insertions / 117 deletions.

Rebased onto `888292e3` after #1342 landed, deliberately rather than treating `behind` as cosmetic: a gate that fails the run on a dropped coroutine can be green on its own base and red once someone else's tests land beside it. #1342 added 7 tests; verified on the rebased tree at **6267 passed, 0 failed**. `range-diff` reports the commit identical across the rebase and the patch-ids match (`69699b9c…` both sides), so the clean `claude-review` from the pre-rebase head applies to exactly this content — it shows `skipped` on the current head only because of the re-run.

**Three things I'd like challenged, because each is a judgement call rather than a mechanic:**

**1. The `AsyncMockMixin` exclusion is where the gate is deliberately weaker than it could be — and my justification is an observation, not a mechanism.** Four tests in `test_crystallization_must_not_wedge.py` shed a mock-internal coroutine. I traced one to `crystallizer_service.py:942`, where the `await` is correct and the drop comes from a bare `AsyncMock()` whose `return_value` is itself an `AsyncMock`, so a plain sync `.get()` returns a coroutine that correct code discards. My first draft asserted a general mechanism; I could not establish one and neither could review, so the comment now records an exclusion rather than a diagnosis. An AST scan over the four service trees found zero bare-statement calls to a locally-defined async function, so it is not hiding a live defect today. **The cleaner end state is fixing `_storage_mock` to return real dicts so the drops stop happening and the exclusion becomes unnecessary** — I left that out of scope, and if you'd rather it were in, say so.

**2. The `f_locals` walk reaches into interpreter internals, and I've documented that it misses a coroutine held in a tuple or list local.** Review argued for the alternative that three existing files already use — patch `tracked_task` with a closing stub so the wrapper coroutine is never created. I kept the walk because patching `tracked_task` is *provably dead* in one of these files: `_enrich_memory_background` re-imports the name from `task_tracker` at call time, the trap `test_embed_off_hot_path.py:967` already documents, so `patch.object(memory_service, "tracked_task", …)` there does nothing. The walk works at the `track_task` boundary regardless of how the inner name resolves. That is a real trade — internals access and a known gap, versus per-site knowledge of import resolution — and I would not argue hard if you prefer the other side.

**3. I declined a shared context manager for the 30 identical `patch("core_api.tasks.track_task", side_effect=…)` sites,** which two reviewers recommended. Reason: it removes the patch target from the line that patches it, and this repo leans on grep-based audits — `grep "core_api.tasks.track_task"` finds every site today and would find none behind a helper. Recorded as a judgement call, not an oversight.

**Also worth your eye:** the only production change is one return annotation, `Awaitable[None]` → `Coroutine[Any, Any, None]` in `contradiction/dispatch.py`, with no runtime effect. It is the cheaper half of this PR — mypy's `unused-coroutine` check fires only for `Coroutine`, so the defect was type-checking clean before. Verified both directions, and dropping the `await` at `consumer.py:165` now fails mypy. Caveat carried in the PR body: 7 of the 9 caller modules sit on core-api's `ignore_errors` list, so it is a ratchet rather than full coverage.

**Residual risk, stated rather than buried:** the gate is only as stable as the drop count being zero, and GC timing is not deterministic. Green across five local runs plus the rebased one, but CI runs different hardware with coverage enabled. If it proves flaky rather than correct, the honest fix is to narrow it, not to widen the exclusions.

Corrections to my own earlier claims are in the commit message and on #869 — `filterwarnings = error` alone is inert, `AsyncMock` is not the fix, and one of my 49 edits was dead and is reverted.

erni-a (approved):

PR #1362, opened 2026-09-07
fix(storage): lock parent memories in one order, and correct the deadlock note

Eldad-Caura:
## What

Two things about the same invariant: the order in which a transaction takes locks on the **parent** rows of `memory_entity_links`.

#1346 fixed the order *within* the link table. It also left a guess on record about the cycle behind the suite's intermittent `DeadlockDetectedError`. That guess is wrong. This corrects it with measurements, pins what the correction rests on, and fixes the one live cycle those measurements turned up.

## The guess, and what actually happens

The docstring said the cycle "more likely runs through the foreign-key locks on the parent `memories` / `entities` rows against a concurrent UPDATE from background enrichment".

Measured against the real database:

| pairing | outcome |
|---|---|
| link insert vs **plain non-key `UPDATE`** of the parent memory | **no conflict at all** |
| link insert vs `FOR UPDATE` on the parent memory | blocks |
| single-row upsert vs partner locking **entity → memory** | **DEADLOCK** |
| single-row upsert vs partner locking **memory → entity** | blocks only |
| sorted batch insert vs unordered multi-row `FOR UPDATE` | **DEADLOCK** |

So enrichment cannot be the partner: the FK check takes `FOR KEY SHARE`, an ordinary `UPDATE` takes `FOR NO KEY UPDATE`, and those don't conflict — that is what the 9.3 row-lock refinement is for.

What the earlier note *did* get right is that the per-item upsert cannot hold two link keys. The step it missed is that a single-row insert still holds **two parent rows**: the FK triggers fire in constraint-OID order (`memory_id_fkey` 16794 before `entity_id_fkey` 16799), so the statement holds `memories(M)` while it reaches for `entities(E)`. A partner taking those two in the opposite order closes a cycle with it — which is how the per-item upsert could be the victim statement in the traceback.

## What that makes load-bearing

`purge_tenant_data` / `purge_fleet_data` are the only paths that lock both parents in one transaction — one `DELETE` per table, walking a declared tuple, so **the tuple order is the lock order**. They are safe today only because `_PURGE_TENANT_TABLES` and `_PURGE_FLEET_TABLES` list `memories` before `entities`. Nothing recorded that, and the tuples otherwise read as a free choice — their own comment says "ordering is free among these" of a later group, which is true there and not true of these two. Now pinned by a test that fails if they're reordered.

## The live cycle, fixed

`memory_redistribute` locks many rows `FOR UPDATE` through `id.in_(...)`, which fixes no order, so it can take two parent memories in the opposite order from a batch link insert (which sorts). Reproduced with a barrier.

Fixed by ordering the select, extracted as `_ordered_memory_lock_select` so the order is testable — the same shape `_ordered_link_rows` already has.

That fix only works because **both paths sort by the same key**: the link helper sorts on `str(id)` in Python, the select orders by the `uuid` column in Postgres. Each being internally consistent wouldn't be enough, so I checked the two agree over 500 random v4 ids rather than reasoning about byte order, and a test pins it.

## What is still open

The captured traceback's exact partner is **not** identified. The cycle shape that can victimise a per-item upsert requires a transaction locking an entity before a memory at `FOR UPDATE` strength, and no current path does that — purge is the only candidate and it goes the safe way. So either the capture predates current code, or the partner was test-local (one test file hard-deletes both parents). I'd rather say that than name a mechanism I haven't demonstrated, which is how the note this PR corrects came to exist.

## Verification

- Root suite **6272 passed / 0 failed**; `core-storage-api` suite **371 passed** (clean env — pointing both suites at `memclaw` trips that suite's own guard).
- mypy clean on `core-storage-api`; `ruff check` and `ruff format --check` clean, one scope per invocation.
- Both new gates confirmed to fail without their fix: the purge-order test by swapping the tuple, the `order_by` test by removing `.order_by`.
- The two characterization tests hold reasons rather than mitigations, which is the pattern this file already documents for `test_opposite_insert_orders_deadlock_at_the_sql_level`.

### DIFF
--- a/core-storage-api/src/core_storage_api/services/postgres_service.py
+++ b/core-storage-api/src/core_storage_api/services/postgres_service.py
@@ -46,6 +46,7 @@ from sqlalchemy.exc import IntegrityError
 from sqlalchemy.ext.asyncio import AsyncSession, async_sessionmaker
 from sqlalchemy.orm import load_only
 from sqlalchemy.sql.dml import ReturningInsert
+from sqlalchemy.sql.selectable import Select
 
 from common import duplicate_memory, permanent_failure
 from common.constants import (
@@ -168,6 +169,35 @@ def _ordered_link_rows(rows: list[dict]) -> list[dict]:
     for row in rows:
         by_key.setdefault((str(row["memory_id"]), str(row["entity_id"])), row)
     return [by_key[key] for key in sorted(by_key)]
+
+
+def _ordered_memory_lock_select(memory_ids: list[UUID], tenant_id: str) -> Select[tuple[Memory]]:
+    """``SELECT ... FOR UPDATE`` over ``memories``, locked in one global order.
+
+    Same hazard as ``_ordered_link_rows`` one table up. A multi-row
+    ``FOR UPDATE`` takes its row locks in the order the executor produces
+    them, and ``WHERE id IN (...)`` fixes no order at all — so this statement
+    and a concurrent link insert can take the same two parent memories in
+    opposite orders and cycle. Measured: with the orders crossed, Postgres
+    kills one side; with them agreed, the later writer only waits.
+
+    ``ORDER BY id`` is the whole mitigation, and it has to be THIS key to be
+    worth anything: the link path sorts on the string form of the id
+    (``_ordered_link_rows``), and Postgres orders the ``uuid`` type by its 16
+    bytes. Those two agree — checked over 500 random v4 ids, both against
+    ``sorted(str(...))`` and against ``sorted(key=.int)`` — which is what makes
+    the two paths agree rather than merely each being internally consistent.
+    """
+    return (
+        select(Memory)
+        .where(
+            Memory.id.in_(memory_ids),
+            Memory.tenant_id == tenant_id,
+            Memory.deleted_at.is_(None),
+        )
+        .order_by(Memory.id)
+        .with_for_update()
+    )
 
 
 @asynccontextmanager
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@erni-a plaes apprvoe

erni-a (approved):

Eldad-Caura:
@erni-a ready for review.

All 14 checks green on `f484184d`, `claude-review` included; `mergeable_state=blocked` only on the review requirement.

This one is mostly a correction. #1346's docstring guessed at the cycle behind the intermittent `DeadlockDetectedError` and named background enrichment; that is measurably not it — a plain non-key `UPDATE` of a parent memory does not conflict with the link insert's FK lock at all. The PR body has the full measured table.

The three things most worth your judgement:

1. **The `memory_redistribute` change is the only production behaviour change** — one `ORDER BY id` on the `FOR UPDATE` select, extracted into `_ordered_memory_lock_select`. It fixes a cycle I reproduced against a batch link insert. If you'd rather that landed on its own, separate from the docstring correction, say so and I'll split it.
2. **The purge-order test pins a tuple order as security-relevant.** `_PURGE_TENANT_TABLES` / `_PURGE_FLEET_TABLES` are safe today only because `memories` precedes `entities`, and their own comment says ordering is free among a later group. If that ordering was deliberate rather than lucky, the test is just recording it; if it was lucky, it's now load-bearing and worth knowing.
3. **I did not identify the captured traceback's partner**, and said so in the PR rather than naming a plausible one. The cycle shape that can victimise a per-item upsert needs an entity locked before a memory, and no live path does that. That's the honest end state, and it's the same class of claim the note I'm correcting got wrong.

I do not merge — yours when you're happy.

erni-a (approved):

PR #1364, opened 2026-09-07
refactor(auth): give the self plane one gate instead of eight hand-written ifs

Eldad-Caura:
## What

Eight routes asked the same question inline — *may this agent credential act as the agent it just named* — and answered it with eight hand-written `if`/`raise` pairs. `AuthContext` already owns the two neighbouring questions; this adds the third and routes all eight through it.

| gate | question |
|---|---|
| `enforce_read_only` | may this credential write at all? |
| `enforce_not_agent_credential` | may an agent credential be here at all? |
| **`enforce_self_agent`** (new) | **may it act as the agent it named?** |

Self-service routes are exactly the ones that must *not* take the plane gate — it would refuse the callers they exist for — and none of the eight does.

## The eight were not identical, which is the reason this is worth doing

Five compared unconditionally; three guarded on truthiness first. Checked site by site: **the guard tracked whether the parameter was optional, not a disagreement about the rule.** Every unconditional site takes a required `agent_id` — a path param, `Query(...)`, or a required pydantic field — so its value cannot be falsy and the two shapes agreed on every input a caller can send.

Except one. `?agent_id=` reaches a required `str` as `""`, where the five strict sites answered 403 and the three lenient ones treated it as omitted. Reconciled to the stricter reading — **`None` passes and only `None`** — because the lenient reading would have turned `GET /stm/notes?agent_id=` from a 403 into a **200 with an empty body**. All three sites that change take an optional parameter defaulting to `None`, so no caller that omits it is affected. Pinned by a test.

## Messages are preserved exactly

`field=` reproduces `/recall`'s two variants byte-for-byte, `detail=` carries the four route-specific wordings, and `redistribute_memories` already matched the default. The substring `does not match the authenticated agent identity` is asserted via `resp.text` by `test_route_authz_gaps` and `test_h06_m30_recall_identity`, so it is contract rather than incidental phrasing.

## Two things deliberately unchanged

- **Plain-string `detail`, not `coded_detail`** like every neighbour in `auth.py` — because that is what all eight sites raised. Giving the self plane an error code changes the response body for existing clients; that is a decision, not a side effect of moving a condition.
- **`services/caller_identity.py` keeps its own comparison.** It logs a warning and lets header precedence win rather than refusing — different behaviour, not a ninth copy.

## Verified rather than inherited from a comment

- **Admin exemption is real with no special case.** The only `is_admin=True` context is built at `auth.py`'s admin branch as `AuthContext(tenant_id=None, is_admin=True)`, leaving `agent_id` None, so the first clause declines to fire. A test pins it, so if an admin context ever grows an agent id the exemption has to be restated deliberately. An explicit `if self.is_admin: return` would imply the opposite.
- **The gate is orthogonal to the write gate**, not paired with it: the four write paths among the eight also call `enforce_read_only`; the read paths don't. I first wrote "all eight also call `enforce_read_only`" and measured it false before it shipped.

## Checked and NOT a finding

`GET /agents/{agent_id}/tune` has no self-plane check while its `PATCH` twin does, which looks like the lopsided read/write pair the 2026-06-11 audit found on STM notes (that pair's comment is still in `stm.py`). It isn't one — both it and `GET /agents/{agent_id}` call `sc.get_agent` and return the same `AgentOut` under the same `enforce_tenant`, so gating one leaves the identical payload one hop away. Whether the agent row should be tenant-readable at all is a real question and a wider one than this change.

## Follow-up this makes possible

`tests/test_authz_gate_inventory.py` mechanically inventories two invariants over every mutating route. The self plane could now be a third — "every route taking a caller-supplied `agent_id` either calls `enforce_self_agent` or is allowlisted with a reason" — because there is finally one named callable to look for. ~22 routes take an `agent_id` without it today, most with legitimate substitutes (the plane gate, `resolve_write_agent` precedence, admin-only), so the allowlist would carry real judgements. Not in this PR.

## Verification

- Root suite **6276 passed / 0 failed**.
- `ruff check` / `ruff format --check` clean on `core-api/src/` and `tests/`, one scope per invocation.
- mypy `core-api` clean apart from pre-existing `dateutil` missing-stub errors in a file this doesn't touch.
- Legacy-name ratchet and do-not-touch sentinel clean.

### DIFF
--- a/core-api/src/core_api/auth.py
+++ b/core-api/src/core_api/auth.py
@@ -253,6 +253,69 @@ class AuthContext:
                 ),
             )
 
+    def enforce_self_agent(
+        self,
+        requested_agent_id: str | None,
+        *,
+        field: str = "agent_id",
+        detail: str | None = None,
+    ) -> None:
+        """Raise 403 if an agent credential named an agent other than itself.
+
+        The self plane, which is a third question from the two gates above it:
+        ``enforce_read_only`` asks whether this credential may write at all,
+        ``enforce_not_agent_credential`` refuses agent credentials outright, and
+        this one admits an agent credential but only as ITSELF. Routes that are
+        meant to be self-service — tune your own profile, read your own notes,
+        promote your own STM — are exactly the ones that must NOT take
+        ``enforce_not_agent_credential``, since it would refuse the callers they
+        exist for, and none of the eight sites here does.
+
+        It is orthogonal to the write gate rather than paired with it: the four
+        write paths among them (tune, redistribute, clear notes, promote) also
+        call ``enforce_read_only``, and the read paths (stats, read notes, the
+        two ``/recall`` knobs) do not. Whether a caller may write and which
+        identity it may act as are independent questions.
+
+        ``None`` passes, and only ``None``. It means the caller asserted no
+        identity, which every route that can receive one treats as "use the
+        authenticated identity" or as a deliberately wider aggregate
+        (``GET /memories/stats`` documents the second); each of those
+        parameters defaults to ``None``, so an omitted one lands here.
+
+        An explicit empty string does NOT pass, and that is the one place the
+        eight call sites this replaces genuinely disagreed. Five compared
+        unconditionally, so ``?agent_id=`` was a 403; three guarded on
+        truthiness, so it was treated as omitted. Since the two groups had to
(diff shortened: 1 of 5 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #1365, opened 2026-09-07
fix(auth): give the self plane its own error code, and make its docstring true

Eldad-Caura:
## What

Follow-up to #1364, from its own review. That PR's helper shipped with three docstring claims that don't survive contact with the code they describe, and a test that doesn't pin what it says it pins. This corrects them and converts the one call site #1364 missed.

## The error code — the stated reason for not having one was false

#1364 answered with a plain-string `detail`, unlike the seventeen `coded_detail` raises around it in the same file, and justified it in the docstring: giving the self plane a code "changes the response body for existing clients."

It doesn't. `coded_detail`'s own docstring says so in as many words, and `app.http_exception_handler` is where it's true — given a coded detail it emits `{"detail": message, "error": {...}}`, so **the top-level `detail` a client already reads is unchanged**; only `error.code` sharpens from the status-derived `FORBIDDEN`. I asserted the property without reading either the handler or the helper.

So the self plane now carries `AGENT_IDENTITY_MISMATCH`. This is what `errors.py`'s own preamble asks for — it exists because "the eight distinct reasons a request can be refused at this boundary arrived at the caller as one word", and this refusal was arriving as that word. `field` rides along in `error.details`, so a caller that sent both `/recall` knobs can tell which one was refused.

Blast radius: the route-level assertions are substring checks on `resp.text`, which still contains the message in both `detail` and `error.message`. No test asserts `error.code == "FORBIDDEN"` on any affected route.

## The admin test could not fail

It built `AuthContext(is_admin=True)` by hand, asserted `agent_id is None`, and claimed in its docstring that it would catch an admin context that "ever grows an agent id". It would not — the real branch is in `get_auth_context`, and the test never called it, so it pinned the constructor default.

It now goes through `get_auth_context` with the admin key **and** an `X-Agent-ID` header, so it also pins the discard the exemption rests on. Measured: make that branch pass `agent_id=agent_id` and the new test fails; the old one never reaches the branch.

## The one behaviour change had no route-level test

#1364 reconciled the sites' disagreement about an empty agent id toward 403 and pinned it only as a unit test on the helper. The realistic trigger isn't `?agent_id=` — it's a JSON body from a serializer that emits `""` for unset fields, which is how `filter_agent_id` and `caller_agent_id` reach `/recall` and `/search`. Pinned there now; measured to answer **200** with the truthiness guard restored.

Also measured, because #1364's commit message got it backwards by calling the required parameters' values "not falsy": `?agent_id=` binds `""` on a required and an optional `Query` alike, `{"filter_agent_id": ""}` binds `""`, and only omission binds `None`.

## Three of the five `detail=` overrides said less than the default

The STM three passed the default sentence minus its trailing `'{self.agent_id}'` — strictly less informative, and formatted **eagerly on every passing request** now that it had moved from inside the `raise` into a keyword argument. Dropped. The two with genuinely different wording keep it, renamed `message=`.

## `routes/keystones.py` was a ninth copy

`_resolve_caller_identity` asks the same question of a header instead of a parameter, so #1364's "the eight sites" was already wrong when it shipped. Converted, preserving behaviour exactly — `x_agent_id or None` keeps an empty `X-Agent-ID:` an omission there, which is what that site has always treated it as, and the message is byte-identical.

## Comments corrected on contact

- Three said the authenticated identity "takes precedence over" the caller-supplied parameter. The gate **refuses** a mismatch rather than overriding it — and at `redistribute_memories` the very next line passes the caller-supplied value straight into `lookup_agent`, so "precedence" actively misreads.
- One said the promote gate binds the identity. It only refuses; `resolve_write_agent` twelve lines down is what rebinds `body.agent_id`.
- One promised the stats gate "leaves every legitimate call (omitted, or naming yourself) untouched" — which #1364 itself made false for `?agent_id=`.

## Also reviewed, deliberately not here

`tests/test_authz_gate_inventory.py` could now carry a third mechanical invariant — "every route taking a caller-supplied `agent_id` either calls `enforce_self_agent` or is allowlisted with a reason" — since there is finally one named callable to look for. Measured while reviewing: **32 routes take a caller-supplied `agent_id`; 8 of them call the gate.** Scoped naively that's an allowlist of roughly two dozen where about half would read the same sentence, which the file's own preamble rejects; ~10 of those really are covered, by `resolve_write_agent`, which the AST classifier can't see because it's an imported name. It needs the classifier extended first. Separate PR.

## Verification

- Root suite **6281 passed / 0 failed**.
- Both new assertions confirmed to **fail** against the property they claim, by flipping that property and re-running.
- `ruff check` / `ruff format --check` clean on `core-api/src/` and `tests/`, one scope per invocation.
- mypy identical to the base — 19 pre-existing errors in 6 files, including the same two in `auth.py`.
- Legacy-name ratchet and do-not-touch sentinel clean.


Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #1366, opened 2026-09-07
fix(entities): block a merge when two bracketed qualifiers conflict (A42)

arkash20:
## Two distinct entities, one row

`_same_identifier_signature` compared **only digit-bearing tokens**:

```python
ta = set(re.findall(r"\d[\w.\-]*", a.lower()))
return ta == tb
```

So:

| pair | digit sets | outcome |
|---|---|---|
| `comet #0002` vs `comet #0012` | `{0002}` ≠ `{0012}` | blocked ✅ *(the case it was built for)* |
| `acme (delaware)` vs `acme (ohio)` | `{}` == `{}` | **merged** ❌ |
| `john smith (legal)` vs `john smith (engineering)` | `{}` == `{}` | **merged** ❌ |

Downstream that reads as `same_subject=true` and yields a **false contradiction between two different things**.

## Fix

Also compare **bracketed qualifiers** — matching the vocabulary `_reattach_subject_discriminators` already uses (`#NNNN` and parentheticals). A merge is blocked only when two **present** qualifiers conflict.

**Deliberately not any trailing word.** Comparing all tokens would block `acme` vs `acme corp` and fragment every alias in the graph.

**Deliberate asymmetry:** `acme` vs `acme (ohio)` still merges. An absent qualifier means *unspecified*, not *different* — blocking it would strand every qualified mention from its own plain surface form. The failure direction favours coalescence: an over-merge is visible and recoverable; an entity that never coalesces fragments the graph silently.

## Scope — what this does *not* fix

The priya-class case I observed on staging during V4 today is a **different defect**. There the extractor emitted the bare name `'priya'` for *both* memories — no qualifier at all — so this guard has nothing to compare and **correctly** allows the merge:

```
staging case   priya vs priya         -> True   (merge still allowed)
A42 case       acme(de) vs acme(oh)   -> False  (blocked)
```

That case needs extraction to *emit* a discriminator, which is mechanism ①'s territory. I've corrected the V4 verification doc, which previously said V4 was "blocked on A42" — that was imprecise.

## Testing

13 unit tests: the conflicting-qualifier block, the four legitimate merges that must not regress (`acme`/`acme corp`, corporate suffixes, diacritics, identical qualifiers), the one-sided asymmetry, and that the original `#0002`/`#0012` guard still holds.

Refs: canonical A42.

### DIFF
--- a/core-api/src/core_api/services/entity_extraction_worker.py
+++ b/core-api/src/core_api/services/entity_extraction_worker.py
@@ -41,14 +41,47 @@ _LITERAL_OR_ATTR_RE = re.compile(
 )
 
 
+# A42 — parenthetical qualifiers, e.g. the "(delaware)" in "acme (delaware)".
+# Matches the discriminator vocabulary ``_reattach_subject_discriminators``
+# already established ("#NNNN" and parentheticals); deliberately NOT any
+# trailing word, because unbracketed tokens are usually harmless surface
+# variation ("acme" / "acme corp") rather than a distinguisher.
+_QUALIFIER_RE = re.compile(r"[(\[]([^)\]]{1,64})[)\]]")
+
+
+def _qualifier_signature(name: str) -> frozenset[str]:
+    """Bracketed qualifiers in ``name``, normalised for comparison."""
+    return frozenset(" ".join(m.strip().lower().split()) for m in _QUALIFIER_RE.findall(name) if m.strip())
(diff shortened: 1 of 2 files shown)


arkash20:
CI green.

@Eldad-Caura please approve

Eldad-Caura (approved):

PR #1367, opened 2026-09-07
test(authz): inventory the self plane, the third axis of the gate

Eldad-Caura:
## The axis the inventory did not ask about

`test_authz_gate_inventory.py` enumerated two: may this credential write at all (`enforce_read_only`), and may an agent credential be here at all (`enforce_not_agent_credential`). #1364/#1365 named a third — may an agent credential act as the agent a request **named** — and nothing enumerated it.

It could not be bolted onto the existing surface, because **it is not about mutation**. `GET /stm/notes?agent_id=<peer>` returned a peer's private notes; `filter_agent_id` on `POST /recall` returned a peer's private rows. Both reads. Both found one route at a time — the 2026-06-11 audit closed the STM `DELETE` twin and left the read open, and `routes/stm.py` still carries the note.

So this invariant is scoped by the route's **interface** — does it accept a caller-supplied agent identity — not by verb, and it reads `route.dependant` instead of the handler body.

## What the surface turns out to be

34 routes take such a parameter:

| | count |
|---|---|
| call `enforce_self_agent` | 12 |
| exempt by construction (`enforce_admin` / `enforce_not_agent_credential`) | 4 |
| allowlist line | 14 |
| **KNOWN GAP**, under a per-axis ceiling | 4 |

The 14 lines split into six kinds, and that spread is the finding: **what `agent_id` means is not constant across this surface**. 3 precedence, 1 stricter than precedence, 4 filters, 2 agent-row reads, 2 bound by an imported resolver, 2 inert. Every past bug on this axis was a route where two of those meanings shared one parameter.

## Reading the request interface is the part that fails silently

Both plausible spellings do, and both fail by returning **less**, never by erroring — so the invariant then passes over what is left.

| how | routes seen |
|---|---|
| `route.dependant` + `field_info.annotation` | **34** |
| `inspect.signature(endpoint)` | 30 |
| `route.dependant`, `field.type_` only | 24 |

`inspect.signature` loses the four handlers whose module has `from __future__ import annotations` — their body annotation is the *string* `"PromoteRequest"`. `field.type_` is `None` on this FastAPI/pydantic build, losing every body-borne identity: ten routes, including `POST /recall` and `POST /search`, which are the two the gate was written for.

`test_the_identity_scan_sees_body_borne_ids` pins three of those routes, and `test_the_self_plane_scope_is_not_silently_empty` fails if the surface collapses. A FastAPI upgrade that moves the annotation again fails there, naming what broke.

## There is no mechanism-shaped exemption, and the first draft had one

It credited any handler containing `auth.agent_id or <param>` — the precedence idiom — as binding the identity. Four routes matched. One should not have.

`delete_memory` decides authorization with `caller_agent_id = auth.agent_id`: plain attribute, no fallback. The line that matched the shape is the **next** one, whose own comment reads *"Authorization must not trust this value; the audit log must not discard it"*. So the exemption was granted by an audit-log line.

Measured, not argued: rewrite the real principal to the escalation form `agent_id or auth.agent_id` and the route **still counted as bound**, because the attribution line satisfies the shape on its own. A false negative in the one check meant to catch exactly that.

An `any()` over a function body cannot tell the deciding expression from a bystander. The rule is gone; those four carry lines naming what they actually decide with — which is how a reader learns `delete` binds *more* strictly than the other three, not less.

## Structure

The three hygiene checks and the ratchet now run per axis off an `_Axis` registry rather than a hard-coded pair. The self plane needed a different route surface (every verb) and its own ceiling; a fourth axis added to the old shape would have inherited the staleness and corroboration checks silently while having **no** gap ceiling — the "sat outside both sets with nothing failing" failure this file already records against `keystones`. Ceilings are per axis so headroom is not fungible: closing a self-plane gap must not license a write-gate one.

`_classify` also records imported callee names and `settings.<name>` reads, on strictly weaker terms than gates. "This handler calls `resolve_write_agent`" is a fact; "that covers the self plane" is a judgement. The fact goes in the classifier, the judgement stays in the allowlist, and `test_allowlist_reasons_that_name_a_mechanism_are_corroborated` checks the fact under it. Imported helpers are still not *followed* — crediting a gate three layers down is the attribution this file exists to avoid.

## Cost

`_parse` is cached. Fifteen tests each rebuild the same rows: ~7,100 `getsource`+`parse` pairs over a few hundred distinct functions. The file goes 3.07s → 1.75s, **below the 1.67s it cost before this change**.

## Verification

- 14 tests in the file; full root suite **6298 passed, 5 skipped, 0 failed**.
- Mutation-checked rather than assumed: adding a spurious `KNOWN GAP` entry for an already-gated route fails both `test_allowlists_have_no_unnecessary_entries` and the per-axis ratchet, and the ratchet message names the axis.
- `ruff check tests/` and `ruff format --check tests/` clean, run separately at CI's scopes.
- `legacy_name_ratchet.py --base origin/main` and `do_not_touch_sentinel.py` both exit 0.

The legacy alias path is hoisted into `tests/_legacy_contracts.py` rather than marked in place. An allowlist key must equal the path the app serves, so the legacy spelling is mandatory there — and `tests/ruff.toml` documents that a marker on such a line gets displaced the first time `ruff format` wraps it. That module exists for this, and this is the fifth entry.

## Not in this PR

The four KNOWN GAPs are recorded, not closed. Write attribution is caller-named on the agent plane today; `config.py` already carries `bind_write_identity_to_auth` ("Phase 2 (spoof hardening), ships dark", default `False`), which is the staged fix. Turning it on is a separate decision — it needs the reserved-`main` credentials re-identified first.

### DIFF
--- a/tests/_legacy_contracts.py
+++ b/tests/_legacy_contracts.py
@@ -4,6 +4,7 @@ FROZEN_PLUGIN_SLUG = "memclaw"  # legacy-name-ok: existing plugin and skill iden
 FROZEN_TOPIC_PREFIX = "memclaw"  # legacy-name-ok: deployed Pub/Sub wire namespace
 LEGACY_API_KEY_FIELD = "memclaw_api_key"  # legacy-name-ok: Settings compatibility field
 INSIGHTS_AGENT_ID = "memclaw-insighter"  # legacy-name-ok: persisted service-agent identity
+LEGACY_KEYSTONES_ROUTE = "GET /api/v1/memclaw/keystones"  # legacy-name-floor: served mirror route, named here not declared
 
 
 def frozen_topic(suffix: str) -> str:
--- a/tests/test_authz_gate_inventory.py
+++ b/tests/test_authz_gate_inventory.py
@@ -27,8 +27,16 @@ walked away without the third, so the SECOND visit to one function is what
 found it. Manual sweeps have not converged, and prose in a route docstring is
 not readable by anything.
 
-So: enumerate the surface and make the omissions explicit. Two invariants, one
-mechanism.
+So: enumerate the surface and make the omissions explicit. Three invariants,
+one mechanism.
+
+The third is not about mutation. ``enforce_self_agent`` (#1364/#1365) asks
+whether an agent credential may act as the agent a REQUEST NAMED, and a read
+can leak on that axis as readily as a write: ``GET /stm/notes?agent_id=<peer>``
+and ``filter_agent_id`` on ``POST /recall`` were both disclosure, not tamper.
+So that invariant is scoped by the route's INTERFACE — does it take a
+caller-supplied agent identity — rather than by verb, and it reads
+``route.dependant`` rather than the handler body. See ``_identity_params``.
 
 WHAT THIS IS NOT. It does not decide whether a gate is the RIGHT one, and it
 cannot: ``skills_inbox`` guards with a module-local ``_require_inbox_admin``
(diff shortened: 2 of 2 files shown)


Eldad-Caura:
@erni-a pleaser apprpvoe

erni-a (approved):

PR #1368, opened 2026-09-07
test(authz): inventory the MCP tool surface, which nothing enumerated

Eldad-Caura:
## The surface the last inventory disclaimed

[#1367](https://github.com/caura-ai/caura/pull/1367) opens by saying what it does **not** cover: `app.mount("/mcp", ...)` is a Starlette `Mount` with no `.routes`, so the route walk cannot see it and `app.openapi()` does not list it. Its closing note — *"that surface wants its own inventory and does not have one"* — is what this answers.

**The gates themselves are already covered, and well.** `test_mcp_write_scope_gate.py` drives twelve write tools and ops with a read-only credential and requires a FORBIDDEN envelope; three more assert the gate does *not* fire on a read. That is a stronger claim than anything static here — it proves the gate **refuses**, not that a call appears.

What those twelve cannot do is **notice a thirteenth tool**. They are hand-written, one per tool and op, so a new tool or a new op on an existing one arrives with no test and nothing fails. Same gap, same argument as the REST inventory.

### 1. Declared ops are accepted ops — and this one already fails

`caura_manage` declares four ops. Its handler's `_valid_ops` accepts **six**:

```python
_valid_ops = {"read", "update", "transition", "delete", "bulk_delete", "lineage"}
```

`bulk_delete` and `lineage` are undeclared. Two consequences:

- `plugin/tools.json` — generated from the registry and CI-locked to it by `test_tools_export_in_sync.py` — under-publishes the tool by two ops.
- `required_trust("caura_manage", "bulk_delete")` resolves to the tool baseline **0**, while the declared `delete` is **3**.

**Enforcement is not affected.** The `bulk_delete` branch calls `enforce_delete` — the same trust≥3 gate as single delete — and nothing reads `required_trust`. So this is a wrong published contract, not an escalation. Recorded as a KNOWN GAP under a ceiling rather than fixed here, because closing it changes the manifest clients read.

Worth knowing: `test_tools_registry.py` currently **pins the four-op set**, so an existing test asserts the incomplete one.

### 2. Every mutating op passes the write-scope gate

Enumerated from what the handler *accepts*, not what the registry *declares* — because `caura_manage` proves those differ, and `bulk_delete` is undeclared and correctly gated. Checking the registry would have missed a real op. Currently clean, so the allowlist is empty and a new exception has to be argued for.

### 3. A declared `trust_required` is corroborated

`trust_required` is **a published claim with nothing behind it**. `required_trust()` is exported and, measured across this repo, called by nothing — the field is range-checked by a test and serialised into the manifest. The gates that actually refuse take their thresholds from elsewhere: `_require_trust(..., min_level=…)` is called at six sites, and every `min_level` comes from `_resolve_read_fleet_gate`, a scope literal (`1 if scope == "agent" else 2`), or `keystone_min_trust`.

So the number a client is told and the number the server applies are two independent systems that happen to agree today. Nothing checked that they keep agreeing.

### 4. Every caller-supplied agent identity is bound — the self plane

The MCP half of the axis #1367 enumerated for REST. It is a **rule** here rather than fourteen allowlist lines, because the binding is spelled differently: MCP has no `AuthContext`, and a handler binds by overwriting its own parameter. Matching the assignment **target** is what makes it decidable — the REST version had to be deleted because `auth.agent_id or <param>` also matched an audit-log line assigning a different variable.

Four spellings are live, and **each later one was found by the checks failing, not by reading ahead**:

```python
agent_id = _get_agent_id() or agent_id                    # 6 tools
agent_id = effective_write_agent_id(_get_agent_id(), …)   # caura_write

caller_agent_id = _get_agent_id()                         # manage, doc
agent_id = caller_agent_id or agent_id

agent_id_effective = _get_agent_id() or agent_id          # keystones
```

The last never assigns to the parameter — it folds it into a new name. That is the shape the REST rule got wrong, so it is admitted only when **every load** of the parameter sits inside such an assignment; a handler that also reads the raw value is reported unbound. That is a property of the whole body, which is what the REST version lacked.

**That leaves exactly one unbound identity: `filter_agent_id` on `caura_recall`.** The REST twin gates the same name, and the reason is a fallback that exists only there:

```python
eff_agent_id = auth.agent_id or body.caller_agent_id or body.filter_agent_id
```

On REST the filter can *become* the identity. Here `caller_agent_id` is bound to the rebound `agent_id`, and the filter reaches `search_memories` as a pure author narrowing. **Recorded as a surface divergence, not a leak** — that is what was verified. Whether the narrowing alone can disclose a peer's rows depends on the storage visibility predicate, which I did not trace to SQL.

## Two REST axes are absent deliberately

| axis | why there is no analogue |
|---|---|
| `enforce_not_agent_credential` | `_check_auth` refuses admin keys outright, so there is no admin plane on this surface to escalate into |
| `is_demo` | the MCP middleware never reads a demo header — already recorded in prose in `test_route_authz_gaps.py`, and not restated here |

## It also corrects its own sibling

`test_authz_gate_inventory.py` named **three** mutating MCP tools. **Seven** call the gate: `caura_doc`, `caura_evolve`, `caura_insights`, `caura_keystones_set`, `caura_manage`, `caura_tune`, `caura_write`.

That count is now checked rather than asserted. `test_the_rest_inventorys_tool_list_is_still_accurate` parses the sibling's docstring and compares it against the tools that actually call `_check_write_scope` — and fails if the paragraph is reworded out from under it, rather than silently matching nothing.

## Verification

- 14 tests, 1.1s. Full root suite **6312 passed, 5 skipped, 0 failed**.
- **Every invariant and guard is mutation-checked**: twelve mutations, each aimed at one check — twelve caught, zero missed.
- The cross-file docstring test was run against the stale three-tool wording it was written for, and against a reworded anchor. It fails on both.
- `ruff check tests/` and `ruff format --check tests/` clean, run separately at CI's scopes; `legacy_name_ratchet.py --base origin/main` and `do_not_touch_sentinel.py` exit 0.

## Not in this PR

The `caura_manage` op gap is recorded, not closed — fixing it means adding two `OpSpec` entries and regenerating `plugin/tools.json`, and `test_tools_registry.py:140` would need updating from the four-op set it currently pins. That is a decision about the published client contract, not a test fix.

Three stale docstrings found while verifying and deliberately left alone, since none is in a file this PR touches: `tools/_types.py` and `tools/_builders.py` both describe a "Phase 1" in which handlers are `None` and live in `@mcp.tool` decorators (there are zero such decorators, and all 12 handlers are set); `mcp_server.py` says the SoT import must run after "the 16 handler functions" (there are 12).

### DIFF
--- a/tests/test_authz_gate_inventory.py
+++ b/tests/test_authz_gate_inventory.py
@@ -5,11 +5,14 @@ covers ``APIRoute`` entries on the core-api app. It does NOT cover the MCP
 surface. ``app.py`` does ``app.mount("/mcp", ...)``, which the walk below sees
 as a ``Mount`` with no ``.routes`` and a plain Starlette ``Route`` — neither an
 ``APIRoute`` — and ``/mcp`` is absent from ``app.openapi()``, so the
-self-check cannot see it either. Every mutating MCP tool (``caura_write``,
-``caura_manage``, ``caura_keystones_set``) guards itself with
-``_check_write_scope`` instead; that surface wants its own inventory and does
-not have one. A plain Starlette ``Route`` added with a mutating method would be
-invisible here for the same reason.
+self-check cannot see it either. Mutating MCP tools guard themselves with
+``_check_write_scope`` instead — seven of them, not the three an earlier
+revision of this paragraph named: ``caura_doc``, ``caura_evolve``,
+``caura_insights``, ``caura_keystones_set``, ``caura_manage``, ``caura_tune``
+and ``caura_write``. That surface now has its own inventory, in
+``tests/test_mcp_authz_gate_inventory.py``, which enumerates it from the tool
+registry so the count above cannot silently go stale again. A plain Starlette
+``Route`` added with a mutating method would still be invisible here.
 
 This class of gap has been fixed one route at a time, repeatedly:
 
--- a/tests/test_mcp_authz_gate_inventory.py
+++ b/tests/test_mcp_authz_gate_inventory.py
@@ -1,0 +1,901 @@
+"""The MCP tool surface, enumerated — the inventory ``test_authz_gate_inventory``
+says it does not cover.
+
+That file opens by disclaiming this surface: ``app.mount("/mcp", ...)`` is a
+Starlette ``Mount`` with no ``.routes``, so the route walk cannot see it and
+``app.openapi()`` does not list it. Its closing note — "that surface wants its
+own inventory and does not have one" — is what this file answers.
+
+WHAT IS ALREADY COVERED, so this file does not restate it. The write-scope gate
+is asserted BEHAVIOURALLY in ``tests/test_mcp_write_scope_gate.py``: twelve
+tests drive a write tool or op with a read-only credential and require a
+FORBIDDEN envelope, and three more assert the gate does NOT fire on a read.
+That is a stronger claim than any static check here can make — it proves the
+gate REFUSES, not merely that the call appears — and it exists because the 2026
+audit found ``_is_write_allowed()`` defined and never invoked.
+
+What those twelve cannot do is NOTICE A THIRTEENTH TOOL. They are hand-written,
+one per tool and op, so a new tool or a new op on an existing tool arrives with
+no test and nothing fails. That is the same gap the REST inventory was written
+for, and the same argument: manual sweeps have not converged. So this file
+enumerates from the REGISTRY and reads the HANDLERS, and everything it checks is
+a fact about one or the other.
+
+FOUR INVARIANTS.
+
+    1. The ops a tool DECLARES are the ops its handler ACCEPTS.
+    2. Every accepted op that mutates passes through the write-scope gate.
+    3. A declared ``trust_required`` above 0 is corroborated by a handler that
+       can actually reach a trust gate.
+    4. Every caller-supplied agent identity is bound to the verified one.
+
+The fourth is the self plane, the axis #1364/#1365 named and #1367 enumerated
+for REST. It reads differently here because the binding is spelled differently:
+REST hands handlers an ``AuthContext`` and asks them to call
+``enforce_self_agent``, while MCP has no ``AuthContext`` at all — identity
+arrives in module-level ``contextvars`` set by ``MCPAuthMiddleware`` and a
+handler binds by REASSIGNING its own parameter, ``agent_id = _get_agent_id() or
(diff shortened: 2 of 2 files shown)


Eldad-Caura:
@erni-a ready for review — all 14 checks green on `a80e59f7`.

Worth knowing before you read it: **the automated reviewer skipped this PR entirely.** Its scope excludes `test_*.py`, and every file here is one, so the comment above says "no source files to review". #1367 was only reviewed because it also touched `tests/_legacy_contracts.py`. So this is ~900 lines of static-analysis logic with no automated read at all — you are the only reviewer it gets.

What it adds: four invariants over the MCP tool surface, which `test_authz_gate_inventory.py` explicitly disclaimed. Enumerated from `core_api.tools.REGISTRY` and read off the handlers, so a thirteenth tool or a new op fails until someone decides about it. The existing `test_mcp_write_scope_gate.py` proves the gates *refuse*; it cannot notice a tool nobody wrote a test for.

The three things I would most want a second opinion on:

1. **`caura_manage` declares 4 ops, its handler accepts 6** (`bulk_delete`, `lineage` undeclared). I recorded it as a KNOWN GAP rather than fixing it, because the fix regenerates `plugin/tools.json` and `test_tools_registry.py:140` pins the incomplete set — that reads to me as a client-contract decision, not a test fix. Tell me if you'd rather it were closed here. Enforcement is unaffected: `bulk_delete` calls the same `enforce_delete` trust≥3 gate.

2. **`filter_agent_id` on `caura_recall` is the one unbound caller-supplied identity.** REST gates the same name. I argue it is a surface divergence rather than a leak, because REST's `eff_agent_id = auth.agent_id or body.caller_agent_id or body.filter_agent_id` fallback — the thing that lets the filter become the identity — has no counterpart on this path. I did not trace the storage visibility predicate to SQL, so if you think that reasoning is too generous, this should become a KNOWN GAP instead.

3. **The self-plane rule in `_rebound_params`.** The REST equivalent was deleted for being a false negative, so this one matches the assignment *target* and, for the fold-into-a-new-name shape, requires that every load of the parameter be consumed by that fold. Four spellings are live and I found the last three only by the checks failing — please sanity-check I have not missed a fifth.

Verification: 14 tests in 1.1s; full root suite 6312 passed, 0 failed. Twelve mutations, each aimed at one check, twelve caught. `ruff check` / `ruff format --check` clean at CI's `tests/` scope, run separately; ratchet and sentinel exit 0.

erni-a (approved):

PR #1369, opened 2026-09-07
fix(auth): name the identity-precedence policy so the self-plane rule works

Eldad-Caura:
## The rule the inventory had to delete

`test_authz_gate_inventory.py` credited any handler containing `auth.agent_id or <param>` as binding the caller's identity. `delete_memory` broke it: that handler authorizes with `auth.agent_id` **alone**, and on the *next* line builds

```python
attribution_agent_id = auth.agent_id or agent_id   # for the audit row
```

under a comment reading *"Authorization must not trust this value"*. The audit line was granting the exemption. Rewriting the real principal to the escalating `agent_id or auth.agent_id` **still** left the route counted as bound — a false negative in the one check meant to catch exactly that.

An `any()` over a function body cannot tell the deciding expression from a bystander. The file names the fix and declines to make it, saying it "should not be the reason for" production surgery. This makes it.

`AuthContext.effective_agent_id(requested)` is that expression behind a name — the precedence half of the self plane, where `enforce_self_agent` is the refusing half. **A named call is decidable where a shape is not**, so the rule is back and three hand-written allowlist lines are gone: `GET /memories`, `PATCH /memories/{memory_id}`, `GET /reports`.

## Three sites, not six

The first draft converted every inline occurrence for consistency. Review measured that three of them bought nothing and two actively hurt:

| site | inventory effect | kept? |
|---|---|---|
| `memories.py:356` `list_memories` | allowlist line retired | **yes** |
| `memories.py:1778` `update_memory_endpoint` | allowlist line retired | **yes** |
| `reports.py:262` `get_report` | allowlist line retired | **yes** |
| `memories.py:522` `memory_stats` | none — already calls `enforce_self_agent` | no |
| `memories.py:1823` `_resolve_read_identity` | none — already calls it, twice | no |
| `caller_identity.py:69` `resolve_caller_and_gate` | none — imported, `_classify` does not follow | no |

The two that hurt: `memory_stats` binds a **local** named `effective_agent_id` to the raw parameter one line below the call, so the name would have meant two different things sixteen lines apart; and `resolve_caller_and_gate`'s own docstring states a flat three-step ladder that its one-line body transcribed, which splitting into a call plus a bare `or` stopped mirroring.

Net production diff: **three lines**.

`delete_memory` is untouched, and is the proof the rule does not over-credit — its audit line still spells itself out, and it keeps its allowlist entry saying it binds *more* strictly than precedence.

## The docstring asserted that, and nothing checked it

This is the finding I'd most want a reviewer to confirm. The method's docstring claimed `delete_memory` "calls this method nowhere" — a security-relevant property with **no test**. Worse, the check that would eventually have fired is `test_allowlists_have_no_unnecessary_entries`, whose message reads *"Delete the entries"* — the exact wrong move at the exact moment the false negative returns.

`test_the_audit_attribution_is_not_bound_by_the_helper` now fires first and says the right thing: revert the audit line; if the principal genuinely changed, argue it in `test_memory_byid_authz.py`; and **do not** resolve this by deleting the allowlist entry.

## The binder gets its own channel

Folding `effective_agent_id` into `gates` was the smaller edit and quietly falsified three things: the failure report prints that set under the label `enforce_* called`, the helper-promotion step reads a non-empty gate set as evidence a helper **guards**, and the corroboration check then accepts that helper as evidence. A binder refuses nothing and must not stand in for something that does.

## Why this rule is weaker than the MCP sibling's

Recorded in the file, because anyone comparing the two inventories will ask. [#1368](https://github.com/caura-ai/caura/pull/1368)'s `_rebound_params` credits a binding only when **every** load of the parameter sits inside it — a property of the whole body, strictly stronger. It does not port. Measured on this surface it reports **2 of the 4 routes as unbound that are not**: `GET /memories` reads the raw `agent_id` again for `author_filter`, and `GET /memories/stats` twice more — because here one parameter is deliberately *both* the visibility identity and the author filter, which is the divergence the REST inventory's preamble opens with. On MCP each name means one thing, so the stricter rule is free there and costs false positives here.

## Also

`_classify` is now cached, on the same grounds as `_parse` beside it. The fifteen tests each rebuild the same rows, so it re-walked already-parsed trees ~8 times per run — 825k `ast.walk` visits, 1.3s of a 3.8s profiled run. **1.43–1.81s → 1.12–1.19s**, while gaining a test.

## Verification

- Full root suite **6321 passed, 5 skipped, 0 failed**.
- Four mutations, each aimed at one claim, four caught: pointing the audit line at the helper fires the new test; **escalating `GET /memories` to `agent_id or auth.agent_id` fails the self-plane invariant — the case the old rule passed**; reverting `PATCH` or `GET /reports` fails it too. Inverting the method itself fails its own unit tests.

## Not in this PR

Typing the sinks — `enforce_delete`, `authorize_memory_access`, `resolve_caller_and_gate` — to take a distinct `AgentIdentity` returned only by this method would make the whole class of bug a **mypy error at the point of the mistake**, on both planes, rather than detectable after the fact. mypy already runs in CI. That is the version worth shipping eventually; it is six sink signatures plus a decision about the MCP plane, not a rename.

Two smaller follow-ups found while reviewing: `tests/test_documents_rest_doc_memory.py:107` hand-rolls an `_Auth` stub whose `enforce_tenant` / `enforce_read_only` / `enforce_usage_limits` all `return None`, so four tests that drive a write handler exercise none of those gates; and six test files build `AuthContext` fixtures independently with no shared factory.

### DIFF
--- a/core-api/src/core_api/auth.py
+++ b/core-api/src/core_api/auth.py
@@ -313,6 +313,36 @@ class AuthContext:
                 ),
             )
 
+    def effective_agent_id(self, requested_agent_id: str | None) -> str | None:
+        """The agent this request ACTS AS: the authenticated identity, or the
+        caller's assertion only when the credential authenticates none.
+
+        The precedence half of the self plane, where ``enforce_self_agent`` is
+        the refusing half. Routes that must not 403 a legitimate caller use
+        this instead: an agent credential silently keeps its own identity, and
+        a tenant or user credential — which authenticates no agent — may still
+        name one, which is what a dashboard listing a fleet does.
+
+        WHY THIS IS A METHOD AND NOT ``self.agent_id or requested``. It is
+        exactly that expression; the value is entirely in the name. The bare
+        form could not be told from an audit-log line of the same shape, so
+        ``tests/test_authz_gate_inventory.py`` had to delete the rule that
+        credited it — the block above ``SELF_ID_PARAMS`` there has the case in
+        full, and is the copy to keep current.
+
+        SCOPE, because the name is broader than the guarantee. This returns the
+        identity; it does not decide what the caller may do with it, and
+        calling it is not authorization. It is for the visibility or
+        authorization identity ONLY — an audit attribution that must record
+        what the request carried, even when authorization ignored it, is a
+        different value and must keep spelling itself out. ``delete_memory`` is
+        the live example of both in one handler, and is deliberately not
+        converted — pinned by
+        ``test_the_audit_attribution_is_not_bound_by_the_helper``, so this
+        paragraph is a rule rather than a request.
+        """
+        return self.agent_id or requested_agent_id
+
     def enforce_tenant(self, requested_tenant: str | None) -> None:
         """Raise if the caller may not write to ``requested_tenant``.
 
--- a/core-api/src/core_api/routes/memories.py
+++ b/core-api/src/core_api/routes/memories.py
@@ -353,7 +353,7 @@ async def list_memories(
     # or by omitting it. The query param stays the AUTHOR filter (written_by).
     # A tenant/user credential (auth.agent_id None) keeps using the param, as the
     # dashboard intends.
-    caller_agent_id = auth.agent_id or agent_id
+    caller_agent_id = auth.effective_agent_id(agent_id)
     # ``written_by`` is the author filter; ``agent_id`` keeps serving as that
     # filter when ``written_by`` is omitted, so existing callers are unaffected.
     author_filter = written_by if written_by is not None else agent_id
@@ -1775,7 +1775,7 @@ async def update_memory_endpoint(
         memory_id,
         tenant_id,
         body,
-        agent_id=(auth.agent_id or agent_id) if auth.tenant_id else None,
+        agent_id=auth.effective_agent_id(agent_id) if auth.tenant_id else None,
     )
 
 
--- a/core-api/src/core_api/routes/reports.py
+++ b/core-api/src/core_api/routes/reports.py
@@ -259,7 +259,7 @@ async def get_report(
     # breakdown's own visibility scoping excludes ``scope_agent`` when no agent
     # is set). This avoids 403-ing a logged-in human on the unregistered default
     # agent id.
-    asserted_agent = auth.agent_id or agent_id
+    asserted_agent = auth.effective_agent_id(agent_id)
     caller_agent_id: str | None = None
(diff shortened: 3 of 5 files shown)


Eldad-Caura:
@erni-a ready for review — all 14 checks green on `85235669`, and `main` has not moved, so the branch is current.

This is the first of tonight's four that touches **production authz**, so it deserves more of your attention than the two inventories did. The good news is it is small: three lines in two route files, plus a one-line method.

The automated reviewer did read the production half and called the three call-site changes behaviour-preserving. It explicitly **excluded** `test_auth_context.py` and `test_authz_gate_inventory.py` per its scope, so the ~180 lines of inventory logic here — the re-enabled rule, the new binder channel, the pinning test — have had no automated read. That is now three PRs in a row where the test half is unreviewed.

Three things I would most want you to check:

1. **The three sites I converted, and the three I did not.** `delete_memory` is deliberately untouched: it authorizes with `auth.agent_id` alone, and its `attribution_agent_id = auth.agent_id or agent_id` is the audit value whose own comment says authorization must not trust it. That line is why the old rule was a false negative. If you think any of the other three reverts (`memory_stats`, `_resolve_read_identity`, `resolve_caller_and_gate`) should have stayed converted, say so — I reverted them because they bought the rule nothing and two hurt readability, not because they were wrong.

2. **Whether a method with no behaviour earns its place.** It is exactly `self.agent_id or requested`; the value is entirely in the name being detectable. A fair reading is that production is bending to a test's convenience. My argument is that the expression is a policy, policies deserve names, and the docstring gives a reviewer something to violate that a bare `or` never did — but it is a judgement and you may weigh it differently.

3. **The new `test_the_audit_attribution_is_not_bound_by_the_helper`.** It exists because the method's docstring asserted a security property with nothing checking it, and because the check that *would* have fired tells the reader to "Delete the entries" — the worst possible advice at that moment. Please check the failure message actually points somewhere useful.

Verification: full root suite 6321 passed, 0 failed. Four mutations, four caught — including escalating `GET /memories` to `agent_id or auth.agent_id`, which is the exact case the deleted rule let through. `ruff check` / `ruff format --check` clean at both CI scopes, run separately; ratchet and sentinel exit 0.

Not in this PR, and the thing actually worth doing next: typing the sinks to take a distinct `AgentIdentity` would make this class of bug a mypy error at the point of the mistake rather than something a test notices afterwards.

erni-a (approved):

PR #1371, opened 2026-09-07
docs(tools): correct four stale claims, and share the route walk they hid behind

Eldad-Caura:
## Four docstrings describing a codebase that no longer exists

Each verified against source, not taken from the note that recorded it.

**`tools/_types.py` and `tools/_builders.py`** both describe a "Phase 1" in which handlers are `None` and still live on `mcp_server.py`'s `@mcp.tool` decorators, with `mcp_register` yet to become the registration path. Measured:

| claim | reality |
|---|---|
| handlers still on `@mcp.tool` decorators | **0** `@mcp.` decorators in `mcp_server.py` |
| `handler` is `None` during Phase 1 | all **12** specs pass `handler=` |
| `mcp_register` *becomes* the path in Phase 2 | every `caura_*` spec module calls it at import |

So the `handler is None` guard is unreachable for any registered spec — `test_tools_registry` asserts exactly that. The note now says the `None` is the dataclass default and that removing it and the guard is a separate change.

**`mcp_server.py`** said its SoT import "must run AFTER the 16 handler functions above are defined". There are **12**. I deleted the number rather than correcting it: `test_tools_registry` already pins the count, and a figure in prose beside a registry that pins it is a second source of truth that only ever drifts. `16` was the pre-consolidation count — which is also where the `*_16tools.json` in the next item came from.

**`test_tool_descriptions_regression.py`** was wrong five ways in one paragraph, and contradicted its own file — it claimed a *"13-tool surface (10 user-facing LTM tools — of which 7 are live and 3 are knowledge-layer placeholders — plus 3 STM tools)"* twelve lines above `EXPECTED_TOOL_COUNT = 12`. There are 12 tools, all live, no placeholders (`EXPECTED_PLACEHOLDERS` is the empty set), no separate STM tools. It also pointed at `*_16tools.json` baselines "kept in `tests/fixtures/`" which are not there — the directory holds `tools_list_baseline_v1.json` and the two `tool_descriptions_*` files the module actually reads. It now points at the constant instead of restating it.

## The walk: the recorded task was wrong in two directions

The follow-up said "duplicated `_IncludedRouter` walk (4 copies)". Measured, it isn't:

- **`test_api_read_scope_params.py` does not walk at all.** It reads `app.openapi()` and only *mentions* `_IncludedRouter` in a comment explaining why it doesn't.
- **`test_authz_gate_inventory.py` and `scripts/tenant_scope_gate.py` walk for PREFIXED paths**, tracking `include_context`. The two real duplicates want **unprefixed** ones and descend by `original_router` alone.

That difference is load-bearing, not accidental. `_REST_CAPABILITY` is keyed `("GET", "/memories")` and `PROBE_ROUTES` holds `/health`, not `/api/v1/health` — both because the middleware sees the route template before the prefix is applied. **Folding all four together would have broken one side or the other.**

So the genuine duplication is **two** files: `test_capability_usage.py` and `test_request_observation.py` run the same eight-line descent and differ only in what they accumulate. Both now call `tests/_route_table.iter_route_declarations`, which yields `(path, methods)` per *declaration* so the caller that needs "declared exactly once" can still count. The FastAPI quirk — and the reason this is **not** the inventory's walk — is explained once there instead of twice.

`scripts/tenant_scope_gate.py` keeps its own copy: different tree, different lint config, and it wants the prefixed form anyway.

## Verified equivalent, not assumed

Both original walks run side by side with the shared one against the live app:

```
declarations walked: 119   distinct paths: 97
capability map identical: True
declaration counts identical: True
paths carrying the /api/v1 prefix: 0
```

That last line is the property the new docstring claims, measured rather than asserted.

## One deliberate placement choice

`tests/_route_table.py` is a `tests/_*.py` helper on purpose. That is the one shape of test file the review pipeline does **not** exclude — as [#1368](https://github.com/caura-ai/caura/pull/1368) and [#1369](https://github.com/caura-ai/caura/pull/1369) both demonstrated, a `test_*.py` file gets no automated read. Putting the shared walk there means it actually gets reviewed.

## Verification

- Full root suite **6321 passed, 5 skipped, 0 failed**.

No behaviour changes: three prose corrections in production files, one in a test, and a two-caller refactor proven byte-identical to what it replaced.

### DIFF
--- a/core-api/src/core_api/mcp_server.py
+++ b/core-api/src/core_api/mcp_server.py
@@ -3934,9 +3934,12 @@ async def mcp_lifespan():
 # ── SoT registration ──────────────────────────────────────────────────────
 # Triggers loading of every `core_api.tools.caura_*.py` spec module. Each
 # spec module registers itself in the REGISTRY and calls `mcp_register(mcp, spec)`
-# to wire the handler to FastMCP. This import must run AFTER the 16 handler
-# functions above are defined — spec modules reference them via
-# `core_api.mcp_server.caura_X` attribute lookup.
+# to wire the handler to FastMCP. This import must run AFTER the `caura_*`
+# handler functions above are defined — spec modules reference them via
+# `core_api.mcp_server.caura_X` attribute lookup. Deliberately no count here:
+# this said "the 16 handler functions" from the initial release until 2026-09,
+# which was the PRE-consolidation figure and never matched the 12 that exist.
+# `test_tools_registry` pins the real number; prose beside it only drifts.
 # The `noqa: E402,F401` silences "module-level import not at top" and
 # "imported but unused" — both are intentional.
 from core_api import tools  # noqa: F401
--- a/core-api/src/core_api/tools/_builders.py
+++ b/core-api/src/core_api/tools/_builders.py
@@ -22,8 +22,10 @@ from ._types import ToolSpec
 def mcp_register(mcp, spec: ToolSpec) -> None:
     """Register `spec.handler` with the FastMCP instance, using `spec.description`.
 
-    No-op when `spec.handler is None` (Phase 1 of the refactor — handlers
-    still live in `mcp_server.py`'s `@mcp.tool` decorators).
+    No-op when `spec.handler is None`, which no registered spec is — see
+    `ToolSpec.handler`. The guard is kept for the dataclass default, not for a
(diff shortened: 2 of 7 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #1373, opened 2026-09-08
fix(tools): declare the two caura_manage ops the manifest already advertised

Eldad-Caura:
## The manifest was contradicting itself, not just under-publishing

`caura_manage`'s handler accepts six ops (`_valid_ops`: read, update, transition, delete, bulk_delete, lineage); the ToolSpec declared four. The recorded finding called that under-publishing plus "a decision about the client contract". Measured, it is neither.

**It contradicted itself.** `plugin/tools.json`'s `ops[]` listed four while, in the *same object*, the `op` parameter's own description read `"read|update|transition|delete|bulk_delete|lineage."` and `memory_ids` was published as `"op=bulk_delete: list of memory UUIDs (max 1000)."` Both ops were already advertised through `params[]`. Declaring them exposes nothing new — it makes the file agree with itself.

**And it is barely a contract decision:** `ops[]` is declared in the plugin's `ToolSpecEntry` interface and read by nothing, so the manifest change is inert there.

## What is *not* inert — and nearly became a bug I introduced

The **description** is served to both surfaces (MCP from the registry, the plugin via `/tool-descriptions`). The plugin dispatches over REST, has **no endpoint** for either new op, and its `PARAM_SCHEMAS` enum admits only the other four — with an `// op === "update"` fall-through for anything else.

My first draft named both ops with no surface caveat. That would have advertised `lineage` — a pure read — to callers where it routes into a mutating fall-through. The description now says MCP-only, and a new guard in `plugin/src/tool-definitions.test.ts` pins declared-vs-offered against an explicit `MCP_ONLY_OPS` list, in both directions.

## Tokens: trimmed, not raised

The description is charged to every session's `tools/list`. The ceiling had **17 tokens of headroom** and its comment says *"Trim before adding, or move the ceiling with a reason of your own."* The first draft spent **31**.

Rather than raise it, the op roll-call came out — the `op` parameter enumerates all six in the same `inputSchema`, so the list was being bought twice, and the old description wasn't merely duplicating it but **contradicting** it with four. Net **+8, nine left**.

The ledger also now records something I found while measuring: its running total said 5321, the fixture measured **5333**. Twelve tokens had been spent without writing themselves down. And the trim generalises — four sibling tools still repeat their own enum verbatim, worth a measured **48 tokens** (5341 → 5293). Noted for a follow-up, not done here.

## A comment that was false, and that no test could have caught

A draft of the note above `ops=` asserted the other four ops *"reach only `authorize_memory_access`, which is row scope rather than a trust ladder."* **Both halves are false:**

- `update` reaches `enforce_update` through `update_memory` in the service layer — invisible to the inventory because `_reachable_calls` filters on `__module__`.
- `authorize_memory_access` ends in `trust_level >= (3 if write else 2)` for cross-fleet rows.

Invariant 3 asks only whether a tool *declaring* a threshold reaches **some** trust gate; it does no per-op attribution, so nothing contradicted me. The comment now claims no more than the two `enforce_delete` calls a reader can see.

Each declared `trust_required` was verified, not assumed: `bulk_delete` calls `enforce_delete` under the same `if caller_agent_id:` guard as single `delete`, and `enforce_update` is never reached on this path, so `update`/`transition` declaring `0` is accurate too.

## The test harness: I hand-rolled a mechanism the repo already had

Closing the gap empties `DECLARED_OPS_ALLOWLIST` and drops its `gap_ceiling` to 0 — which would have left the necessity check iterating nothing. My first fix added a hand-kept `_NECESSITY_CHECKED` set of axis names plus a meta-test to police it.

The REST sibling already solves this: its `_Axis` carries `needs_entry` and `rows`, and its docstring records why — *"one definition rather than two that can drift."* That shape is ported here. `_Axis` gains a `needed()` callable it **cannot be constructed without**, the four invariants now *ask* their predicates instead of restating them, and both hygiene checks are written once. That deletes the parallel string set, its meta-test, and the `axis.name == "SELF_BIND_ALLOWLIST"` switch three call sites had to know about.

**All four axes are now policed for necessity; before, two were exempt by omission.**

## Mutation-proven, not assumed

| probe | fires |
|---|---|
| entry restored to `DECLARED_OPS_ALLOWLIST` (ops now agree) | necessity check |
| `caura_recall.agent_id` in `SELF_BIND_ALLOWLIST` (param IS rebound) | necessity check |
| entry in `TRUST_ALLOWLIST` — **axis the old shape could not police** | necessity check |
| entry in `WRITE_SCOPE_ALLOWLIST` — **also unpoliced before** | necessity check |
| key naming a tool that does not exist | staleness check |
| a `KNOWN GAP` at ceiling 0 | the ratchet |
| `MCP_ONLY_OPS` emptied / naming an offered op | the TS guard, both directions |

## Deliberately out of scope

- **The plugin's fall-through.** `caura_manage`'s dispatcher ends at `// op === "update"` with no guard, so an unrecognised op becomes a `PATCH`. `caura_doc` has the identical shape ending in `DELETE`. That is a real defect and wants its own PR with its own tests covering **both** — fixing one dispatcher inside a declaration change is the half-measure this PR is arguing against.
- **`error_codes`.** This spec declares `INVALID_ARGUMENTS`; the handler also emits `NOT_FOUND` and `PERMISSION_DENIED`. But 5 of 12 specs declare anything at all and nothing binds the field to any handler — fixing one makes the field *sometimes* trustworthy, which is worse than uniformly untrustworthy. Wants one change across all twelve plus a check deriving emitted codes.
- **`update`'s declared trust.** It inherits 0 while `enforce_update` refuses trust 0. No single integer captures a threshold conditional on row ownership (1 for own, 3 for others'), so the number is left alone rather than guessed at.

## Verification

- Full root suite **6321 passed, 5 skipped, 0 failed**; plugin **431 passed, 0 failed**.


Eldad-Caura:
@erni-a ready for review.

All 14 checks green on `2ba8d07a`. Full root suite 6321 passed, plugin suite 431 passed, both lint scopes and both naming gates clean.

Three things worth your eyes specifically:

1. **The published description changes on both surfaces.** `caura_manage`'s description is served to MCP *and* to the plugin via `/tool-descriptions`, and the plugin dispatches only four of the six ops. It now reads "bulk_delete and lineage are MCP-only". If you'd rather the two ops not be named to plugin callers at all, that's a one-line change — but the caveat and the `lineage` explanation have to travel together, or we advertise a read op to the surface that would dispatch it as a `PATCH`.

2. **`trust_required=3` on `bulk_delete`.** Verified identical to single `delete`: same `enforce_delete(tenant_id, caller_agent_id)` under the same `if caller_agent_id:` guard. Nothing reads `required_trust()` today, so this changes a published claim, not enforcement.

3. **The `_Axis` port in `tests/test_mcp_authz_gate_inventory.py`** is the largest hunk. It replaces a hand-kept set of axis names with the `needs_entry`/`rows` shape the REST sibling already uses, so all four allowlist axes are policed for necessity instead of two. Every guard was mutation-proven — the PR body has the table.

Three things I deliberately left out, each with reasoning in the description: the plugin's `op === "update"` fall-through (real defect, wants its own PR covering `caura_doc`'s identical twin), `error_codes` under-declaration (needs one change across all twelve specs, not one), and `update`'s declared trust (no single integer captures a threshold conditional on row ownership).

Not merging — over to you.

erni-a (approved):

PR #1376, opened 2026-09-08
fix(plugin): stop an unrecognised op falling through to a write

Eldad-Caura:
## A typo deleted a document

`caura_manage` and `caura_doc` dispatch on an `op` string through a chain of `if (op === "…") return apiCall(…)`. Both chains **ended in a comment rather than a condition** — `// op === "update"` and `// op === "delete"` — so the final branch doubled as the else, and every value the chain didn't match became a **write**.

Measured against the unfixed dispatcher, these are the test failures:

```
caura_manage op="lineage"  →  PATCH  /api/v1/memories/11111111-1111-4111-8111-111111111111
caura_doc    op="remove"   →  DELETE /api/v1/documents/doc-1
```

`lineage` is a **read**. It became a write.

## Why those two values, specifically

Neither is invented. [#1373](https://github.com/caura-ai/caura/pull/1373) declared `bulk_delete` and `lineage` on `caura_manage` because the MCP handler accepts them — and the description they now appear in is served to *this* surface too, which has no endpoint for either. So the ops most likely to arrive here were exactly the ones that landed in the fall-through. This is the follow-up that PR named.

## What the guard is, and what it is not

The only thing that had kept those branches unreachable was the `op` enum in `PARAM_SCHEMAS` — enforced by the host's schema validator, **outside this package**. `unsupportedOp` makes that boundary local.

It does **not** replace upstream validation, and upstream validation would not make it redundant. The two cover different failure modes: a validator checks *caller input against the schema*, while this checks that a dispatcher's chain is *exhaustive over the ops it claims to serve*. Add an enum entry with no branch and a perfect validator waves it straight into the terminal position — back to a `PATCH`.

**It throws rather than returning a structured error, deliberately.** The plugin's `ToolResult` has no `isError` field and `execute` calls `textResult` unconditionally, so returning `{"error": …}` would hand the host a **success** result containing an error — precisely the `isError=false` bug the server fixed at its own chokepoint. Mirroring the server's shape would import that bug without the fix. Throwing is also the only convention on this surface (`assertSafePathSegment`, `apiCall` on any non-2xx, `createToolFromSpec`'s own drift checks).

## The op lists are hoisted, not derived

`MANAGE_OPS` / `DOC_OPS` are `as const` beside this file's existing `MEMORY_TYPES` / `STATUSES`, spread into the schema and passed to the guard — so the enum a caller is validated against and the refusal naming the alternatives read **one identifier, checked at compile time**.

An earlier draft derived the list at runtime by walking `PARAM_SCHEMAS[tool]?.properties?.op?.enum` through a cast. That degraded silently to an empty list on a mistyped tool name: a silent failure inside a guard whose entire purpose is not failing silently.

## The tests assert the mutation *before* the refusal

Ordering is load-bearing twice over:

1. A test that only checked for a throw would **pass** on a dispatcher that mutated and *then* threw.
2. `assert.rejects` first aborts before the mutation check. My first draft did exactly that, and on unfixed source reported only "missing expected rejection" — never naming the `DELETE`. Reordered, the failure prints the request.

| | result |
|---|---|
| 3 negatives against unfixed dispatcher | **fail**, with the write named in the message |
| `caura_doc op=delete` control | **passes before and after** — which is what makes it a control |

The `caura_doc` control is kept because **nothing else in the suite issues a `caura_doc` request at all**. The `caura_manage op=update` twin is dropped: `tool-identity.test.ts` already drives it and asserts the `PATCH`.

One op value per tool — after the guard nothing branches on the op *value*, so further values re-test a single line.

## CI registration, verified by observation

`npm test` enumerates each compiled test file, so an unregistered one never runs — which is how the plugin once had 423 tests and CI ran **none** of them. Rather than trust that I edited the right line, I confirmed the effect: the suite went **431 → 435**.

## Follow-ups deliberately not taken

- **A per-op lookup table** keyed by a union derived from the `as const` arrays. It would make *a missing handler* a compile error — the one failure mode nothing guards today. But 8 of the 10 op branches have no dispatch-level coverage, so converting now would mean an unreviewable-by-tests rewrite of a `DELETE`-issuing dispatcher. The harness added here is the net that refactor needs first. It also would **not** remove this guard.
- **A guard that every `plugin/src/*.test.ts` appears in the test script.** Same shape as the bug fixed here — an unhandled case silently means "do nothing" — one layer out. It needs a `dist` clean (there is no clean step, so a glob would run stale artifacts from deleted sources) plus a rename of `legacy-contracts.test.ts`, which declares no tests and is a fixture module wearing a test suffix. Worth spinning off.

## Verification

- Plugin suite **435 passed, 0 failed**; root suite **6321 passed, 5 skipped, 0 failed**.
- `tsc` clean; `legacy_name_ratchet.py` and `do_not_touch_sentinel.py` exit 0. No Python changed.

### DIFF
--- a/plugin/package.json
+++ b/plugin/package.json
@@ -16,7 +16,7 @@
     "pretest": "bash ../scripts/gen-version.sh",
     "build": "tsc",
     "dev": "bash ../scripts/gen-version.sh && tsc --watch",
-    "test": "tsc && NODE_ENV=test node --test dist/educate.test.js dist/tool-definitions.test.js dist/tool-identity.test.js dist/env.test.js dist/transport.test.js dist/health.test.js dist/config.test.js dist/runtime-contract.test.js dist/identity.test.js dist/install-id.test.js dist/validation.test.js dist/reconcile-skills.test.js dist/context-engine.test.js dist/heartbeat.test.js dist/interview-buffer.test.js dist/interview-command.test.js dist/task-trail.test.js dist/keystones.test.js dist/resolve-agent.test.js dist/bootstrap-memoization.test.js dist/fleet-scope.test.js dist/gateway-methods.test.js",
+    "test": "tsc && NODE_ENV=test node --test dist/educate.test.js dist/tool-definitions.test.js dist/tool-op-dispatch.test.js dist/tool-identity.test.js dist/env.test.js dist/transport.test.js dist/health.test.js dist/config.test.js dist/runtime-contract.test.js dist/identity.test.js dist/install-id.test.js dist/validation.test.js dist/reconcile-skills.test.js dist/context-engine.test.js dist/heartbeat.test.js dist/interview-buffer.test.js dist/interview-command.test.js dist/task-trail.test.js dist/keystones.test.js dist/resolve-agent.test.js dist/bootstrap-memoization.test.js dist/fleet-scope.test.js dist/gateway-methods.test.js",
     "check:version": "bash -c 'V=$(node -p \"require(\\\"./package.json\\\").version\"); grep -qF \"PLUGIN_VERSION = \\\"$V\\\"\" src/version.ts || (echo \"version.ts out of sync with package.json ($V): run bash ../scripts/gen-version.sh\" >&2 && exit 1) && echo \"check:version ok: $V\"'"
   },
   "devDependencies": {
--- a/plugin/src/tool-definitions.ts
+++ b/plugin/src/tool-definitions.ts
@@ -143,6 +143,17 @@ export const STATUSES = [
   "outdated", "conflicted", "archived", "deleted",
 ] as const;
 
+// The ops each op-dispatched tool OFFERS on this surface. Hoisted for the
+// same reason as the two lists above: the schema a caller is validated
+// against and the refusal naming the alternatives now read one identifier,
+// so they cannot drift. Deliberately NOT `tools.json`'s `ops[]`, which is
+// what the MCP handler ACCEPTS and legitimately differs — see MCP_ONLY_OPS
+// in tool-definitions.test.ts.
+const MANAGE_OPS = ["read", "update", "transition", "delete"] as const;
+const DOC_OPS = [
+  "write", "read", "query", "delete", "search", "list_collections",
+] as const;
+
 const MEMORY_TYPE_SCHEMA = {
   type: "string",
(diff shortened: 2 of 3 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

Eldad-Caura:
@erni-a ready for review.

All 14 checks green on `4e366e74`. Plugin suite 435 passed, root suite 6321 passed, tsc clean, ratchet and sentinel exit 0. No Python changed.

This is the follow-up #1373 named. The headline is that it was worse than "an unhandled op does nothing":

    caura_manage op="lineage"  ->  PATCH  /api/v1/memories/{id}
    caura_doc    op="remove"   ->  DELETE /api/v1/documents/doc-1

Both dispatchers ended their `if (op === …)` chain in a comment rather than a condition, so the last branch was the else. `lineage` is a read.

Three things worth your eyes:

1. **It changes plugin runtime behaviour.** An op outside the `PARAM_SCHEMAS` enum now throws instead of silently performing the last branch. Well-behaved callers are unaffected (the host validates the enum), and nothing sane could have depended on the old behaviour — but it is a behaviour change in shipped plugin code, not just a declaration fix.

2. **Throwing rather than returning a structured error is deliberate.** `ToolResult` has no `isError` field and `execute` calls `textResult` unconditionally, so an `{"error": …}` return would give the host a *success* result containing an error — the same `isError=false` problem core-api fixed at its own chokepoint. If you would rather this surface grow a real error path, that is a bigger change and should not ride along here.

3. **`MANAGE_OPS` / `DOC_OPS` are now the source for both the schema and the refusal message.** They sit beside the existing `MEMORY_TYPES` / `STATUSES`. If you would rather they came from `tools.json`, note that file is read with `readFileSync` at runtime by design, so it cannot supply a compile-time type — the reconciliation stays a test either way, and `tool-definitions.test.ts` already does it.

Two follow-ups I did not take, both argued in the PR body: a per-op lookup table with a derived union (wants dispatch-level tests for the 8 uncovered ops first, and would not remove this guard), and a guard that every `plugin/src/*.test.ts` is in the `npm test` list — the same "unhandled case silently means do nothing" shape, one layer out, but it needs a `dist` clean and a rename of `legacy-contracts.test.ts`.

Not merging — over to you.

erni-a (approved):

PR #1378, opened 2026-09-08
test(tools): derive each tool's error_codes from its handler

Eldad-Caura:
## What

`ToolSpec.error_codes` is a published field: `scripts/export_tool_specs.py` writes it into `plugin/tools.json`, `test_tools_export_in_sync` holds the two in lockstep, and `plugin/src/tool-specs.ts` types it as part of the manifest clients read.

**Nothing in `plugin/src` reads it at runtime.** Like `ops[]` before #1373, it is a claim rather than a mechanism — which is the argument for a test, not against one: a claim with no consumer has nothing to contradict it when it goes stale.

Measured on `main`:

| | before | after |
|---|---|---|
| specs declaring **nothing** | **7 of 12** | 0 |
| emittable codes undeclared | **52** | 0 |
| declared-but-unemittable | 0 | 0 |
| distinct codes on the surface | 12 | 12 |

The field was never *wrong*, only silent — the harder defect, because an empty tuple reads exactly like a considered "this tool has no error codes" and is indistinguishable from one.

## The test is the deliverable

Completing 12 declarations without a check just resets the clock — the reason I deferred this on #1373 rather than bundling it. So this adds:

- **`tests/_error_codes.py`** — an AST scan deriving the codes each handler can actually reach. In `tests/_*.py` because the scan, not the assertions built on it, is the part that can be wrong, and that shape of file is the one the automated review pipeline reads (`tests/_route_table.py` set the precedent in #1371).
- **`tests/test_tool_error_codes_inventory.py`** — two invariants plus three guards on the scan's own floor.

The two invariants **cover each other**: a scan that *narrows* makes "every emittable code is declared" pass silently, and is caught only by the reverse direction. The dict-literal path is caught by that reverse test alone.

## Three paths reach a caller, and my first scan followed two

1. `_error_response("CODE", …)` inside the handler.
2. A dict literal `{"code": "CODE"}` — `caura_manage`'s unknown-op envelope.
3. **A module-level pre-baked `CallToolResult`.** `_check_auth` returns `_AUTH_ERROR`/`_ADMIN_ERROR`; `_check_write_scope` returns `_READ_ONLY_ERROR`. All built at import time.

Every handler calls `_check_auth`, so **path 3 is the most widely reaching one and the one a function-scoped walk cannot see.** Walking function bodies alone missed **19 of the 52** codes and reported `UNAUTHORIZED` — a code every single tool can emit — as emitted by none. I caught it by noticing the survey attributed no codes to helpers that every handler calls. That path now has a dedicated guard so it cannot regress silently.

## Two calls worth reviewing

**The four envelope codes are declared per tool, not hoisted.** `UNAUTHORIZED`, `FORBIDDEN`, `MISSING_AGENT_ID` and `INVALID_ARGUMENTS` account for 34 of the 52 additions. An `ENVELOPE_ERROR_CODES` constant unioned by the exporter would remove that repetition, but `FORBIDDEN` is *also* raised directly by five handlers, so the invariant would weaken to `emitted == declared | ENVELOPE` — which cannot see a spurious envelope code in a spec. That is a blind spot in exactly the place this file already had one. The cost is 34 repeated entries; I took it deliberately.

**The 52 codes were computed from the scan by a script, not transcribed.** Hand-copying is where a transcription bug got into this session's earlier work.

## Verification

- `test_every_emittable_code_is_declared` **fails on unmodified `main`**, naming all 12 tools; the other four pass.
- **All 5 mutation probes fire**: narrow the scan three ways (top-level constants, helper following, dict literals), drop a code from a spec, add a bogus code.
- Root suite **6326 passed, 5 skipped**. `ruff check` and `ruff format --check` clean at both CI scopes, run separately. `legacy_name_ratchet.py --base origin/main` and `do_not_touch_sentinel.py` both exit 0.
- Corrected one thing during review: a docstring of mine asserted the helper graph "has cycles (`_with_latency` is called by nearly everything)". Measured, `_with_latency` has 13 callers of 46 — and many callers is a *diamond*, not a cycle. The only real cycle is `call_tool` calling itself, which the scan already excludes. The `_seen` guard stays as a termination guard for mutual recursion that does not exist yet; the docstring now says that instead.

## Zero token cost

`error_codes` appears in neither `tools/list` nor `/tool-descriptions`, so none of this is charged to a session's context budget. `tests/test_mcp_token_budget.py` is untouched.

### DIFF
--- a/core-api/src/core_api/tools/caura_doc.py
+++ b/core-api/src/core_api/tools/caura_doc.py
@@ -84,7 +84,14 @@ _SPEC = ToolSpec(
             required_params=("query",),
         ),
     ),
-    error_codes=("INVALID_ARGUMENTS",),
+    error_codes=(
+        "FORBIDDEN",
+        "INTERNAL_ERROR",
+        "INVALID_ARGUMENTS",
+        "MISSING_AGENT_ID",
+        "UNAUTHORIZED",
+        "UPSTREAM_ERROR",
+    ),
 )
 register(_SPEC)
 mcp_register(mcp_server.mcp, _SPEC)
--- a/core-api/src/core_api/tools/caura_entity_get.py
+++ b/core-api/src/core_api/tools/caura_entity_get.py
@@ -14,6 +14,12 @@ _SPEC = ToolSpec(
     handler=mcp_server.caura_entity_get,
     plugin_exposed=True,
     trust_required=0,
+    error_codes=(
+        "FORBIDDEN",
+        "INTERNAL_ERROR",
+        "INVALID_ARGUMENTS",
+        "UNAUTHORIZED",
+    ),
 )
 register(_SPEC)
 mcp_register(mcp_server.mcp, _SPEC)
--- a/core-api/src/core_api/tools/caura_evolve.py
+++ b/core-api/src/core_api/tools/caura_evolve.py
@@ -29,6 +29,13 @@ _SPEC = ToolSpec(
     plugin_exposed=True,
     trust_required=1,
     impl_status="live",
+    error_codes=(
+        "FORBIDDEN",
+        "INTERNAL_ERROR",
+        "INVALID_ARGUMENTS",
+        "MISSING_AGENT_ID",
+        "UNAUTHORIZED",
+    ),
 )
 register(_SPEC)
 mcp_register(mcp_server.mcp, _SPEC)
--- a/core-api/src/core_api/tools/caura_insights.py
+++ b/core-api/src/core_api/tools/caura_insights.py
@@ -27,6 +27,13 @@ _SPEC = ToolSpec(
     plugin_exposed=True,
     trust_required=1,
     impl_status="live",
+    error_codes=(
+        "FORBIDDEN",
+        "INTERNAL_ERROR",
+        "INVALID_ARGUMENTS",
+        "MISSING_AGENT_ID",
+        "UNAUTHORIZED",
+    ),
 )
 register(_SPEC)
 mcp_register(mcp_server.mcp, _SPEC)
--- a/core-api/src/core_api/tools/caura_keystones.py
+++ b/core-api/src/core_api/tools/caura_keystones.py
@@ -29,7 +29,12 @@ _SPEC = ToolSpec(
     handler=mcp_server.caura_keystones,
     plugin_exposed=True,
     trust_required=0,
-    error_codes=("INTERNAL_ERROR",),
+    error_codes=(
+        "FORBIDDEN",
+        "INTERNAL_ERROR",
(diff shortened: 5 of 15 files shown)


Eldad-Caura:
@erni-a ready for review.

All 14 check-runs are green on head `78c4c702` (12 success, 2 skipped, zero failures — verified by literal SHA, not `gh pr checks`).

**What this is.** `ToolSpec.error_codes` was silent rather than wrong: 7 of 12 specs declared nothing and 52 emittable codes went undeclared, while 0 specs declared a code they could not emit. An empty tuple is indistinguishable from a considered "this tool has no error codes", which is why completing the declarations without a check would just reset the clock. The test is the deliverable; the 12 declarations are what it enforces.

**Two judgment calls I'd like challenged:**

1. **No `ENVELOPE_ERROR_CODES` hoist.** Four codes now repeat across 11-12 specs — 34 of the 52 additions. I kept them per-tool because `FORBIDDEN` is *also* raised directly by five handlers, so a hoisted set would weaken the invariant to `emitted == declared | ENVELOPE`, which cannot see a spurious envelope code in a spec. That would put a blind spot exactly where this scan already had one. The cost is real repetition — tell me if you'd trade it the other way.

2. **The scan lives in `tests/_error_codes.py`, not in the test file.** The scan is the part that can be wrong, and `tests/_*.py` is read by the automated review pipeline while `test_*.py` is excluded (`tests/_route_table.py` set that precedent in #1371). That paid off here — the reviewer read it as source and independently confirmed its output against `mcp_server.py`.

**Where the risk actually is.** The invariants are only as good as the scan, and a *narrowing* scan makes the under-declaration test pass silently. My first version proved that: it walked function bodies only, so it missed that `_check_auth` returns module-level pre-baked `CallToolResult` constants, under-reporting 19 of 52 codes and attributing `UNAUTHORIZED` — a code every tool can emit — to no tool at all. Hence the reverse invariant and the three floor guards. All 5 mutation probes fire, including three that narrow the scan; the dict-literal path is caught by the reverse invariant alone.

One fix during my own review worth flagging, since it's the failure mode this workstream exists to remove: a docstring of mine asserted the helper graph "has cycles (`_with_latency` is called by nearly everything)". Measured, `_with_latency` has 13 callers of 46 — and many callers is a diamond, not a cycle. The only real cycle is `call_tool` calling itself, which the scan already excluded. The `_seen` guard stays as a termination guard for mutual recursion that doesn't exist yet, and the docstring now says that rather than claiming a property I hadn't measured.

No token-budget impact: `error_codes` appears in neither `tools/list` nor `/tool-descriptions`.

erni-a (approved):

PR #1379, opened 2026-09-08
fix(plugin): make a missing op branch a build error, and pin every op's route

Eldad-Caura:
## What

#1376 stopped an unrecognised op falling through to a write, but left two gaps behind:

1. **Nothing failed if an op was added to `MANAGE_OPS`/`DOC_OPS` without a branch.** The new op would reach `unsupportedOp` and be refused at runtime — on a surface that publishes it as supported.
2. **Of the ten ops the two dispatchers offer, exactly one had a test asserting where it goes** (`caura_doc op=delete`, because it had been the fall-through).

## Five lines close the first gap at compile time

Both dispatchers took `enriched.op as string`, so the `if (op === …)` chain never narrowed and the fall-through accepted anything. Casting to a union derived from the same tuple the inputSchema publishes, and typing `unsupportedOp`'s parameter `never`, makes an exhausted chain narrow to `never` — so a leftover member is a build error:

```
error TS2345: Argument of type '"archive"' is not assignable to parameter of type 'never'
error TS2345: Argument of type '"list_collections"' is not assignable to parameter of type 'never'
```

Those are the two build probes, and I ran them before writing the comment that claims the behaviour rather than after.

**The runtime throw stays, and it is not unreachable.** The cast is a runtime lie: nothing in this package validates `op`. I checked rather than inheriting the claim from #1376's docstring — `validation.ts` exports UUID, safe-path-segment, URL, signature and prompt-length checks and no enum validation, and the only one in the execute path is `assertSafePathSegment` on ids. The `PARAM_SCHEMAS` enum is enforced by the **host**, so any string can still arrive. The docstring now says this explicitly, because deleting the throw as dead code would restore the original defect.

## A table closes the second

| tool | op | route |
|---|---|---|
| `caura_manage` | read | `GET /memories/{id}` |
| | transition | `PATCH /memories/{id}/status` |
| | delete | `DELETE /memories/{id}` |
| | update | `PATCH /memories/{id}` |
| `caura_doc` | write | `POST /documents` |
| | read | `GET /documents/{doc_id}` |
| | query | `POST /documents/query` |
| | search | `POST /documents/search` |
| | list_collections | `GET /documents/collections` |
| | delete | `DELETE /documents/{doc_id}` |

The #1376 guard only proves an *unrecognised* op is refused. It says nothing about a *recognised* one going to the wrong place — the same defect one branch further in. These routes are genuinely confusable: `transition` and `update` are both PATCH on neighbouring paths, `read` and `delete` differ only by method on an identical path, and `caura_doc op=read` with a `doc_id` of `"query"` is a legal request differing from `op=query` by method alone.

A meta-test compares the table against the published op enum, so adding an op without a row fails the tests — the test-level twin of the compile-time check. The standalone `caura_doc op=delete` test is subsumed by its row, so net churn stays small.

Added to the existing `tool-op-dispatch.test.ts` rather than a new file: CI enumerates plugin test files explicitly in `package.json`, so an unregistered `*.test.ts` never runs.

## This supersedes the lookup table I had planned

The backlog item was a per-op lookup table keyed by a derived union. Reading the dispatchers first, that turned out to be the wrong shape: each op builds a genuinely different body — `update` filters keys, `query` assembles eight fields — so a table of handlers would still need one builder per op. It would have bought compile-time exhaustiveness and nothing else, which these five lines buy without restructuring either dispatcher.

## Verification

- **445 pass, 0 fail.** 435 before, minus the subsumed test, plus eleven — the arithmetic is the check that the new tests actually registered rather than being silently skipped.
- **Four mutation probes fire, each hitting exactly one test** (no collateral failures): `read` issued as POST, `search` pointed at the query route, `transition` losing its `/status` suffix, and a row removed from the table.
- **Two build probes fire:** an op added with no branch, and a branch disabled while the op stays offered.
- `legacy_name_ratchet.py --base origin/main` and `do_not_touch_sentinel.py` both exit 0. No Python or spec files touched, so `plugin/tools.json` needs no regeneration.

### DIFF
--- a/plugin/src/tool-definitions.ts
+++ b/plugin/src/tool-definitions.ts
@@ -154,6 +154,11 @@ const DOC_OPS = [
   "write", "read", "query", "delete", "search", "list_collections",
 ] as const;
 
+// Derived, so the dispatcher below is checked against the same tuple the
+// inputSchema publishes rather than against a second hand-written list.
+type ManageOp = (typeof MANAGE_OPS)[number];
+type DocOp = (typeof DOC_OPS)[number];
+
 const MEMORY_TYPE_SCHEMA = {
   type: "string",
   enum: [...MEMORY_TYPES],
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #1380, opened 2026-09-08
test(plugin): fail when a test file is not registered with the runner

Eldad-Caura:
## The gap

`npm test` enumerates its inputs explicitly — `node --test dist/a.test.js dist/b.test.js …` — and the CI job is just `cd plugin && npm test`. So an unlisted `src/*.test.ts` compiles, ships, and is **silently never executed**. Nothing failed when that happened, because the suite that would have complained is the one that didn't run.

Measured: **24 files match `src/*.test.ts`, 23 were listed.**

The unlisted one is `legacy-contracts.test.ts`, and it turns out to declare **zero** tests — it's a fixture module holding two frozen identifiers, wearing a suffix that made it look like a forgotten suite. Renamed to `legacy-contracts.fixture.ts` with its eight importers updated, so the name states what it is and the guard has nothing to demand. The Python suites already use this distinction (`tests/_legacy_contracts.py` is imported, not collected).

## Two decisions that go against the obvious fix

**No `dist/*.test.js` glob.** Replacing the explicit list with a glob is the one-line version of this fix, and it's wrong: there's no clean step before `tsc`, so a glob also runs stale artifacts from deleted or renamed sources. This very rename produced such an orphan — `dist/legacy-contracts.test.js` sits on disk next to the new `dist/legacy-contracts.fixture.js`. A glob would have started running a file that no longer exists in `src`.

**Tokenised, not substring-matched.** This package contains both `identity.test.ts` and `tool-identity.test.ts`. Asking whether the script `includes("dist/identity.test.js")` answers **yes** when only the `tool-` one is listed — a guard built that way passes while a real suite goes unrun. There's a probe below for exactly this.

## The rename broke a second registration list, and that is the more interesting half

`plugin/src` filenames are also enumerated **from Python**: `core_api/routes/plugin.py`'s `_plugin_files` allow-list and the install script's `SRC_FILES` bash fallback both have to cover every non-test `.ts`, so fresh `curl … | bash` installs can download them. `tests/test_plugin_source_manifest.py` guards both, and its docstring records why: a 2026-04-16 refactor added two files, forgot both lists, and **every fresh install broke with `TS2307` until it was fixed three days later.**

My rename introduced a filename neither list knew, and that guard caught it — I had judged this change "plugin-only" and run only `npm test` plus the two naming gates, which is how it reached CI. A rename under `plugin/src` is never plugin-only.

The fix is not to add the fixture to those lists. Nothing outside the suites imports it, so listing it would assert a runtime dependency that doesn't exist and ship two frozen legacy identifiers into every install for no reason. Instead `_expected_source_files` now excludes `*.fixture.ts` as test-support source.

That exclusion is only safe while no shipped module imports a fixture — so that condition is now a test rather than a sentence in a docstring:

```
test_no_shipped_module_imports_a_test_fixture
```

It checks by import, not by naming discipline, because the naming is the thing being relied on.

## Verification

- **The guard fails on unmodified `main`**, naming `dist/legacy-contracts.test.js` and nothing else — the pre-existing gap, not a synthetic one.
- **Plugin suite: 438 pass** (435 before, plus three guard tests). **Root suite: green**, including the two manifest tests that caught the rename.
- **Three probes on the registration guard**, each naming exactly the right file:

| probe | guard names |
|---|---|
| a listed file dropped from the script | `dist/transport.test.js` |
| a script entry whose source is gone | `dist/deleted-suite.test.js` |
| `tool-identity` dropped while `identity` remains | `dist/tool-identity.test.js` |

- **Two probes on the manifest change**, confirming the exclusion opened no hole: an unregistered *runtime* module is still caught, and a shipped module importing a fixture is caught.
- Legacy-name ratchet recognises the rename as a move rather than reading two removed `legacy-name-ok` lines as progress; it and the do-not-touch sentinel both exit 0. Worth noting since the fixture holds the frozen `"memclaw"` plugin id.

## Scope note

Three lists enumerate these files: the `npm test` script, `_plugin_files`, and the install script's `SRC_FILES`. All three are now guarded. The Python suites need no equivalent — pytest discovers `tests/test_*.py` by collection.

### DIFF
--- a/plugin/package.json
+++ b/plugin/package.json
@@ -16,7 +16,7 @@
     "pretest": "bash ../scripts/gen-version.sh",
     "build": "tsc",
     "dev": "bash ../scripts/gen-version.sh && tsc --watch",
-    "test": "tsc && NODE_ENV=test node --test dist/educate.test.js dist/tool-definitions.test.js dist/tool-op-dispatch.test.js dist/tool-identity.test.js dist/env.test.js dist/transport.test.js dist/health.test.js dist/config.test.js dist/runtime-contract.test.js dist/identity.test.js dist/install-id.test.js dist/validation.test.js dist/reconcile-skills.test.js dist/context-engine.test.js dist/heartbeat.test.js dist/interview-buffer.test.js dist/interview-command.test.js dist/task-trail.test.js dist/keystones.test.js dist/resolve-agent.test.js dist/bootstrap-memoization.test.js dist/fleet-scope.test.js dist/gateway-methods.test.js",
+    "test": "tsc && NODE_ENV=test node --test dist/educate.test.js dist/tool-definitions.test.js dist/tool-op-dispatch.test.js dist/tool-identity.test.js dist/env.test.js dist/transport.test.js dist/health.test.js dist/config.test.js dist/runtime-contract.test.js dist/identity.test.js dist/install-id.test.js dist/validation.test.js dist/reconcile-skills.test.js dist/context-engine.test.js dist/heartbeat.test.js dist/interview-buffer.test.js dist/interview-command.test.js dist/task-trail.test.js dist/keystones.test.js dist/resolve-agent.test.js dist/bootstrap-memoization.test.js dist/fleet-scope.test.js dist/gateway-methods.test.js dist/test-registration.test.js",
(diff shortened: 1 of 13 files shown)


Eldad-Caura:
@erni-a pleae approve

erni-a (dismissed):

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #1381, opened 2026-09-08
perf(tools): stop buying five op roll-calls twice in tools/list

Eldad-Caura:
## What

Every session's `tools/list` pays for each tool's description. Five of those descriptions recite a parameter's own values, word for word, right next to the parameter that recites them in the same payload:

| spec | roll-call removed |
|---|---|
| `caura_doc` | `op: write\|read\|query\|delete\|list_collections\|search` |
| `caura_keystones_set` | `op: set\|delete` |
| `caura_keystones_set` | `scope ∈ {tenant, fleet, agent}; weight ∈ {low, med, high}` |
| `caura_insights` | `focus: contradictions\|failures\|stale\|divergence\|patterns\|discover` |
| `caura_evolve` | `outcome_type: success\|failure\|partial` |

**5341 → 5269 cl100k: a 72-token saving on every session**, losing nothing — each list is still one field away in the same response.

Worth noting what I found while checking the premise: **no parameter carries a JSON Schema `enum`.** The alternatives live in each parameter's *description* as pipe-separated prose. So this is genuinely duplicated text, not a schema being restated in words — which is what makes removal safe, and also means the original premise I'd written down for this task was wrong in a way that mattered.

This follows the precedent from #1373, where `caura_manage` paid for its new clauses the same way: the op roll-call came out of the description because the `op` parameter enumerated all six anyway.

## A correction to my own ledger note

The ledger predicted **48** tokens (5341 → 5293) for this trim. The measured figure is **72**.

Both are honest measurements of *different edits*. 5293 is exactly what four removals give — I confirmed it by restoring one fragment and re-measuring. The note counted **specs**, and `caura_keystones_set` has **two** roll-calls, so it missed a 24-token fragment. The entry now says to count the edits, not the files.

## What I deliberately left on the table

Both of these look like the same pattern and are not:

- **The `scope:` clauses in `caura_insights` and `caura_evolve`.** They read as roll-calls but carry trust thresholds and the `divergence requires fleet/all` / `fleet_id required when scope='fleet'` constraints, which **no parameter description holds**. Editing them would be a rewrite, not a de-duplication.

- **`caura_keystones_set`'s TARGET-agent explanation, its `INVALID_ARGUMENTS` consequence, and its trust-gating paragraph.** The 2026-05-14 (+155) and 2026-08-26 (+25) ledger entries record buying *exactly that text* after real readers misread shorter versions — an external friction report, and a tester who filed a resulting 403 as a bug. The `agent_id` parameter now repeats some of it, and **there the duplication is the fix.** Removing it because it's duplicated would quietly reverse two documented, incident-driven decisions. This is what the ledger is for, and it worked.

## Ceiling 5350 → 5320

The 2026-08-25 entry chose 5350 to keep "~50 tokens of headroom so the next accidental description bloat still trips this guard". At 5269 that headroom had grown to 81, leaving the guard looser than it was designed to be. Lowering it banks the saving instead of spending it on slack; the next addition still only has to state its reason, exactly as that entry asked.

## Verification

- **5269 measured on the regenerated fixture** using the same `cl100k` method the budget test uses — not the generator's own reported count.
- Root suite green; both naming gates exit 0; `ruff check` and `ruff format --check` clean at both CI scopes, run separately.
- `plugin/tools.json` and all three baseline fixtures regenerated by their own scripts (`scripts/export_tool_specs.py`, `capture_baselines.py`), rebased on top of #1378 and #1379 rather than the base this branch started from.


Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #1382, opened 2026-09-08
feat(scripts): add ref-only legacy-name census

Eldad-Caura:
## Summary

- add a ref-only census that accepts repeatable `NAME PATH REF` repository inputs, resolves each ref to a commit, and runs `git grep` against that commit rather than the working tree
- reuse the canonical ratchet's `LEGACY_NAME`, positive-control name, and marker parser directly; the ratchet gate is unchanged
- separate marked deliberate aliases, permanent floor mentions, marker-system metadata, and unclassified candidate debt per repository and top-level directory
- emit a versioned JSON report and the same information as a human-readable summary
- exclude `.claude/worktrees/`, `.worktrees/`, nested instances of those paths, and gitlinks; run a same-query-shape `caura` control before trusting a small or zero result

## Current census

Fetched immediately before measurement:

| repository | ref | commit | raw lines | unclassified | marked aliases | floor mentions | marker metadata | control (`caura`) |
|---|---|---:|---:|---:|---:|---:|---:|---:|
| `caura-ai/caura` | `origin/main` | `091f9f05e0cb` | 541 | 138 | 213 | 152 | 38 | 5,141 |
| `caura-ai/caura-enterprise` | `origin/dev` | `e97eea1ec00e` | 966 | 644 | 87 | 112 | 123 | 5,722 |
| **sum** | | | **1,507** | **782** | **300** | **264** | **161** | |

This does **not** reproduce the reported 2,143 remaining of 3,635. The historical refs, exact query, repository set, and exclusions were not supplied, so the script records baseline provenance as unverified. The current ref-only raw count gives the arithmetic comparison 1,507 / 3,635 (2,128 fewer, 58.5%), but this PR does not claim that as a verified trend from the outside report. Raw cross-repository sums can also overlap vendored/generated content; JSON says so explicitly.

The existing marker convention is sufficient, so this PR adds no convention. `legacy-name-ok` identifies compat aliases and `legacy-name-floor` identifies permanent textual/wire references. However, marker coverage is incomplete: some known permanent `memclaw_*` behavior and bridge prose remains unmarked. Therefore the script deliberately calls the remaining 782 lines **unclassified candidate debt**, not “not-yet-done.” Proposed follow-up for review: annotate confirmed permanent aliases with the existing `legacy-name-ok` marker in a separate audited PR; do not silently infer permanence in this census.

Example:

```bash
python3 scripts/legacy_name_census.py \
  --repo caura /path/to/caura origin/main \
  --repo caura-enterprise /path/to/caura-enterprise origin/dev \
  --baseline 3635 --json
```

## Verification

- new tests were run before the implementation and failed because `scripts/legacy_name_census.py` did not exist
- `uv run pytest --noconftest tests/test_legacy_name_census.py -q` — 8 passed
- `uv run pytest --noconftest tests/test_legacy_name_census.py tests/test_legacy_name_ratchet.py -q` — passed
- `uv run ruff check scripts/` and `uv run ruff check tests/` — passed
- `uv run ruff format --check scripts/` and `uv run ruff format --check tests/` — passed
- `uv run mypy --config-file scripts/mypy.ini scripts/` — passed

The ordinary root pytest invocation was not used for this isolated script suite because the repository's autouse fixture requires a local PostgreSQL instance; `--noconftest` keeps this ref-only integration suite independent of unrelated database setup.

### DIFF
--- a/scripts/legacy_name_census.py
+++ b/scripts/legacy_name_census.py
@@ -1,0 +1,424 @@
+#!/usr/bin/env python3
+"""Count legacy-name lines in explicit git refs across repositories.
+
+The census is deliberately ref-only. It resolves every supplied ref to a commit
+and runs ``git grep`` against that commit; modified and untracked working-tree
+files cannot affect the result.
+
+Usage::
+
+    scripts/legacy_name_census.py \
+      --repo caura . origin/main \
+      --repo caura-enterprise ../caura-enterprise origin/dev
+    scripts/legacy_name_census.py --repo caura . origin/main --json
+"""
+
+from __future__ import annotations
+
+import argparse
+import hashlib
+import json
+import subprocess
+import sys
+from pathlib import Path, PurePosixPath
+from typing import Literal, NotRequired, TypedDict
+
+import legacy_name_ratchet as ratchet
+
+CountField = Literal[
+    "all_match_lines",
+    "unclassified_lines",
+    "deliberate_alias_lines",
+    "deliberate_floor_lines",
+    "marker_metadata_lines",
+]
+
+_COUNT_FIELDS: tuple[CountField, ...] = (
+    "all_match_lines",
+    "unclassified_lines",
+    "deliberate_alias_lines",
+    "deliberate_floor_lines",
+    "marker_metadata_lines",
+)
+
+# Ref scans cannot see untracked worktrees in the first place. These pathspecs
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #147, opened 2026-05-14
CAURA-000 fix(mcp): isError=True on error envelopes + clarify keystones contract

erni-a:
## Summary

Two friction-report-v3 fixes on the OSS side:

* **B2 — `isError=True` on structured-error envelopes.** Tools were emitting `{"error":{"code":...}}` content payloads alongside `isError=False`, so clients literal-reading the MCP spec (`if not result.isError: success(...)`) silently treated every gateway refusal as success. Promote at two chokepoints — `_with_latency` (the common in-tool wrap) and a new `_as_error_result` (for pre-tool / auth returns). The JSON envelope is preserved verbatim in `content[0].text`; only the boolean flips. Pre-baked `_AUTH_ERROR` / `_ADMIN_ERROR` are now CallToolResults, so the `if err := _check_auth(): return err` idiom propagates `isError=True` unchanged.
* **D5 + D6 — keystones description clarity.** `memclaw_keystones_set`'s `agent_id` is the **TARGET** agent the rule binds to, not the caller, and must be **omitted** for `scope=tenant` or `scope=fleet` (passing it returns `INVALID_ARGUMENTS`). Documented at three levels: tool description, per-field `Field(description=...)`, handler docstring. `memclaw_keystones`'s response keys the merged set under `rules` (not `keystones`); documented in the tool description and docstring so a fresh dev doesn't lose 5 minutes to it.

Side effects:
* Regenerated `tests/fixtures/tool_descriptions_*_baseline_v1.json` and `tools_list_baseline_v1.json` for the new wording.
* `plugin/tools.json` regenerated via `scripts/export_tool_specs.py`.
* `test_mcp_token_budget.py` ceiling raised 5000 → 5200 (+155 cl100k for the description expansions; documented in the comment block).

New tests: `tests/test_mcp_iserror_propagation.py` (5 tests) locks in the B2 contract — `isError` flips on `_with_latency`-wrapped, raw-return, and pre-baked auth error paths; success paths stay strings.

## Test plan

- [x] `pytest tests/ -m unit` — 905 pass (1023 deselected; 5 ingest tests skip on missing `kreuzberg` dep, pre-existing).
- [x] `pytest tests/test_mcp_iserror_propagation.py` — 5/5 pass.
- [x] `pytest tests/test_mcp_*.py` — 117/117 pass.
- [x] `pytest tests/test_tool_descriptions_regression.py tests/test_mcp_token_budget.py tests/test_tools_export_in_sync.py` — green.
[...]

PR #1383, opened 2026-09-08
fix(mcp): annotate the return type the handlers have actually had since #147

Eldad-Caura:
## What

All twelve `caura_*` handlers are annotated `-> str`. **None of them returns only `str`.** Each returns either:

- a JSON string, which FastMCP wraps as `CallToolResult(isError=False)`, or
- a `CallToolResult` that is already `isError=True` — the pre-baked auth and read-only-scope refusals, and anything routed through `_with_latency`.

`_with_latency`'s own signature has said `-> str | CallToolResult` since `9de86f61`, "CAURA-000 fix(mcp): isError=True on error envelopes" (#147). The handlers were never updated to match.

This adds a `ToolReply` alias naming that union and applies it to the twelve.

**Not a widening for the type checker's convenience.** The two shapes are the existing contract, and confusing them is precisely the bug #147 fixed: a client reading `result.isError` cannot distinguish an error from a success if an error envelope arrives as a bare success `str`. A return type admitting one shape while two are returned is the wrong place to leave that distinction unstated.

## Why this doesn't also remove the mypy exemption

`core_api.mcp_server` sits on core-api's `ignore_errors` list, which is why 179 stale return types went unreported for this long. Removing that line surfaces **195** errors. These twelve annotations fix **179** of them:

| | errors | `[return-value]` |
|---|---|---|
| exemption removed, before this change | 195 | 179 |
| exemption removed, after this change | **16** | **0** |

The remaining 14 in this module are **not** mechanical. Several are `str | None` reaching a `str` parameter through guards mypy cannot follow — the largest cluster being six errors behind `caura_write`'s `(content is None) == (items is None)` XOR, where the invariant genuinely holds but needs restating in a form that narrows. Complicating that choice: neither `assert` nor `cast` is house style here (4 and 2 uses across all of `core_api/src`, and zero assert-for-narrowing), and an `assert` is stripped under `-O`, which would make the narrowing a claim rather than a fact.

Those 14, plus the exemption line whose removal makes any of this *enforced*, are a separate change with separate judgement calls. **This PR is the part with none.** (The other 2 of the 16 are `import-untyped` in `common/enrichment/service.py`, environment artifacts that CI's `uv` environment does not produce.)

## The risk I checked

FastMCP builds tool schemas by introspecting type hints, so a return-annotation change could have altered the **published** contract for all twelve tools. It does not:

- the `tools/list` baseline fixture is unchanged, and so is the token-budget count
- `test_v1_baseline_matches_live_registry` still matches the live registry

Annotations are inert at runtime, so there is no behavioural change — but that was worth confirming rather than assuming, since the failure mode would have been a silent contract change.

## Verification

- Root suite **6328 passed, 4 skipped, 0 failed**.
- `ruff check` and `ruff format --check` clean on `core-api/src/`, run separately.
- `legacy_name_ratchet.py --base origin/main` and `do_not_touch_sentinel.py` both exit 0.
- The 195 → 16 figures above were measured by temporarily removing the exemption line, not estimated; the line itself is untouched in this diff.

### DIFF
--- a/core-api/src/core_api/mcp_server.py
+++ b/core-api/src/core_api/mcp_server.py
@@ -845,6 +845,17 @@ def _serialize(obj) -> str:
     return json.dumps(obj.model_dump(mode="json"), indent=2, default=str)
 
 
+# What every ``caura_*`` handler returns. Not a widening for mypy's benefit:
+# the two shapes are the existing contract, and ``_with_latency`` below has
+# carried this exact union since #147 made error envelopes flip ``isError``.
+# The handlers kept ``-> str`` from before that change, so the
+# annotation has been claiming one shape while two are returned — a
+# ``str`` the framework wraps as ``isError=False``, or a ``CallToolResult``
+# that is already ``isError=True``. Confusing the two is precisely the
+# ``isError`` bug the flip fixed, so the return type is worth naming.
+ToolReply = str | CallToolResult
+
+
 def _with_latency(result: str, t0: float) -> str | CallToolResult:
     """Stamp the response with ``_latency_ms`` — and promote error
     envelopes to ``CallToolResult(isError=True)``.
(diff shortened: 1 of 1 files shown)


Eldad-Caura:
@erni-a plesare approve

erni-a (approved):

PR #1384, opened 2026-09-08
fix(mcp): type-check mcp_server and drop it from the ignore_errors list

Eldad-Caura:
## What

`core-api/pyproject.toml`'s `ignore_errors` list describes itself as **"a to-do, not a policy"** — delete a line, fix what mypy reports, and the gate tightens for good. This deletes `core_api.mcp_server`, the entry that hid 179 stale return annotations until #1383.

| | errors in `mcp_server` |
|---|---|
| before #1383, exemption removed | 195 |
| after #1383, exemption removed | 14 |
| **this PR** | **0** |

## The 14, and why none of them is mechanical

- **`caura_write`'s batch path** used `items` five times where mypy saw `list | None`. The invariant is real — the guard at the top admits exactly one of `{content, items}` and the single-write path returns first — so it's bound to a narrowed name rather than asserted. An `assert` disappears under `-O`, which would leave the invariant *claimed* rather than held.
- **`memory_type` / `write_mode`** reach Pydantic-validated `Literal`s as wire strings. Coded ignores, because narrowing the **parameters** would change what FastMCP publishes in `tools/list` — the schema is generated from those annotations. `MemoryCreate` already rejects anything else at runtime.
- **The bulk path** assigned a `BulkMemoryResponse` to the name the single-write path used for a `MemoryOut`. Given its own name.
- **`caura_doc`'s `collection`** is guaranteed by a guard that tests the **value** of `op` (`op not in {list_collections, search} and not collection` → refuse). No annotation expresses that, so coded ignores citing the guard. *The structural fix is per-op functions — the same shape as #1379's plugin dispatch — and is deliberately not attempted here.*
- **`caura_insights`'s `_QUERY_DISPATCH`** returns a different shape per focus, which mypy widened to `object`. Annotated `Any`, which is what it is; every downstream consumer is shape-agnostic.
- **`caura_keystones_set`** needed `title` and `content` ignored exactly as `scope` and `weight` already were, two lines below.
- **`effective_write_agent_id`** returns `str | None` and was assigned back to a `str` parameter. **Not** fixed by narrowing the resolver: its None-body behaviour is deliberate and `tests/test_effective_write_agent_id.py` pins `(None, None) -> None`. The call site takes `or agent_id` instead — unreachable in practice, since that parameter defaults to `DEFAULT_AGENT_ID` so both inputs cannot be falsy — and states the reason rather than asserting it.

## On the idiom

`assert` and `cast` are not house style here: **4 asserts across all of `core_api/src`** (none for narrowing) and **2 casts**, against **122 `type: ignore`, 121 of them with an explicit error code**. So narrowing where the invariant can become a fact, and a coded ignore with a reason where the guard depends on a value mypy cannot see. `caura_keystones_set` already had the exact ignore two lines from one of these errors.

## Verification

- `mypy src/` reports **zero** errors in `mcp_server`.
- **Two probes confirm the gate now fires** rather than merely being reconfigured:

| probe | errors surfaced |
|---|---|
| `_get_tenant` given an `int` return | **89** |
| one handler reverted to the pre-#147 `-> str` | **9** |

  The second matters: #1383's fix cannot regress silently now.

- Root suite **6328 passed, 4 skipped, 0 failed**. `ruff check` and `ruff format --check` clean on `core-api/src/`, run separately. Both naming gates exit 0.

The 2 remaining `import-untyped` errors are in `common/enrichment/service.py` — a stub package missing from a local venv, not this change. They appear identically with the exemption still in place, and main's CI is green.

## Note on the ratchet

This removes one of ~25 entries. The pyproject comment records how the `services.forge.*` recovery went: it found `forge/cron_handler.py` importing a name that had never existed, so the distill cron had raised on every tick and never once run (#955). Being inside an exempted tree is what kept that invisible. `mcp_server` was the largest remaining entry.

### DIFF
--- a/core-api/pyproject.toml
+++ b/core-api/pyproject.toml
@@ -206,7 +206,6 @@ module = [
     "core_api.routes.fleet",
     "core_api.routes.memories",
     "core_api.clients.storage_client",
-    "core_api.mcp_server",
     "core_api.app",
     "core_api.auth",
 ]
--- a/core-api/src/core_api/mcp_server.py
+++ b/core-api/src/core_api/mcp_server.py
@@ -1215,7 +1215,12 @@ async def caura_write(
     # case the body-supplied id is honored so the install self-identifies
     # rather than collapsing onto "main". Reads keep `_get_agent_id() or
     # agent_id` (visibility scoping is unaffected).
-    agent_id = effective_write_agent_id(_get_agent_id(), agent_id)
+    # ``or agent_id`` keeps this ``str``: the resolver returns None only when
+    # BOTH inputs are falsy, and this parameter defaults to DEFAULT_AGENT_ID,
+    # so the fallback is unreachable here. It states that instead of asserting
+    # it — and the resolver's ``str | None`` is correct on its own terms,
+    # exercised with a None body by tests/test_effective_write_agent_id.py.
+    agent_id = effective_write_agent_id(_get_agent_id(), agent_id) or agent_id
     if refuse := _refuse_default_agent_on_gateway(agent_id):
         return _with_latency(refuse, t0)
     # C3/C8 — reject reserved memory_types at the boundary before we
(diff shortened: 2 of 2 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #1385, opened 2026-09-08
fix(events): contract the org enum so dual_subscribe=False works again

Eldad-Caura:
Contracts the `Org` enum onto the current names. Companion to #1372, which flipped
the family this morning.

## This closes a live breakage, it is not tidy-up

`dual_subscribe=False` is the **default**. Between #1372 and this PR:

```
unbound_publish_topics(dual=False) -> ['memclaw.org.suppression-changed',
                                       'memclaw.org.settings-changed']
```

Non-empty there means a default-constructed `PubSubEventBus` **raises on
construction**. And the guard is fleet-wide by design — `unbound_publish_topics` is
computed over every topic the module declares, so the moment one family is flipped
without being contracted, *every* process at the default fails, whether or not it
touches `org`. That is precisely the standalone and on-prem deployments that never
run the Terraform the flag is gated on.

`memory` sat in that state for four days (2026-09-01 → 2026-09-05, closed by #1307).
`org`'s window closes the day it opened.

After this change: `unbound_publish_topics(dual=False) == ()`.

## Why contracting before the drain is correct, not premature

Contraction stops subscribers **binding** the legacy name. That is the precondition
for draining it — not a substitute for having drained it. It is safe here because
publishers have already moved, and that is established **by delivery, not by
configuration**:

```
prod, first publish after the promote, 06:23:13Z
  prod--caura.org.suppression-changed   1
  prod--memclaw.org.*                   0
```

The legacy topics stay. They come out when the drain gate passes — currently
**not** before 2026-09-09 06:13Z, since a >24h post-flip window does not exist yet.

## The test both windows were missing

Every existing assertion about the construction guard monkeypatches a single family:

```python
monkeypatch.setattr(topics_mod, "FLIPPED_FAMILIES", frozenset({"lifecycle"}))
assert topics_mod.unbound_publish_topics(dual=False) == ()
```

That asks *"does the guard work?"*. It never asks *"is the shipped set in a state the
guard permits?"* — which is the question that catches a flip that shipped without its
contraction. **Neither the `memory` window nor the `org` window failed a test.** Both
were found by reading the module.

`test_every_flipped_family_is_contracted_so_the_default_bus_is_legal` asks the second
question, deliberately with no monkeypatch, so the value under test is the shipped one.

Confirmed load-bearing: reverting the two enum literals fails it and nothing else.

## Scope

| file | change |
|---|---|
| `common/events/topics.py` | two enum members; the comment records why the gap was closed same-day |
| `core-api/src/core_api/suppression.py` | docstring named the legacy topic |
| `tests/.../test_topic_rename_cutover.py` | new test; two stale docstring paragraphs |

**`019_tenant_suppression.py` deliberately keeps the legacy name.** It is a historical
record of what the topic was called when the migration ran, and
`do_not_touch_sentinel.py` pins it with the reason *"the migration stops naming the
topic it documents"*. Changing it would have been wrong and the sentinel would have
caught it.

## Gates

- `tests/test_events/`: failure set **byte-identical to `origin/main`** (18 known
  environment-only failures, diffed against a pristine control worktree, not assumed
  from a matching count)
- Do-not-touch sentinel: 35/35
- Legacy-name ratchet: `No new lines. 0 annotated, 3 removed, 0 excused moves (-3 net)`
- `ruff check` clean; `ruff format --check` clean on all three changed files. The repo
  has 12 pre-existing format-dirty files — verified identical on `origin/main`, so they
  are not from this change and are not touched here.

## Next

`org` still needs its >24h drained window before Terraform contraction. `audit` remains
last, unconditionally — those rows are hash-chained.

### DIFF
--- a/common/events/topics.py
+++ b/common/events/topics.py
@@ -41,14 +41,14 @@ class Org(enum.StrEnum):
     # so the OSS boundary guard (core-api) can reject reads/writes for
     # affected tenants synchronously, even while the durable mirror
     # eventually catches up.
-    SUPPRESSION_CHANGED = "memclaw.org.suppression-changed"
+    SUPPRESSION_CHANGED = "caura.org.suppression-changed"
     # CAURA-571: core-api publishes this after an org's settings are written so
     # every process drops its per-process settings cache promptly — without it,
     # a tightened governance control keeps applying its looser prior value on
     # sibling workers for up to the cache TTL (5 min). Subscribe with
     # ``broadcast=True`` (every process must receive it), not the work-queue
     # default.
-    SETTINGS_CHANGED = "memclaw.org.settings-changed"
+    SETTINGS_CHANGED = "caura.org.settings-changed"
 
 
 class Lifecycle(enum.StrEnum):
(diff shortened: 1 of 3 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #1389, opened 2026-09-08
feat(conflicts): human review of detected conflicts — backend (D11)

arkash20:
## Detector precision has never been measurable

`memory_conflicts` has recorded what the **detector** concluded since A55. Nothing recorded what a **person** concluded — so in storage, a false positive is indistinguishable from a true one.

This is the read + decide half. No UI yet, deliberately: the API is verifiable on its own.

## What's here

| layer | change |
|---|---|
| model + migration `043` | `review_status`, `resolution_action`, `resolution_note`, `resolved_by`, `resolved_at`, + a `(tenant_id, review_status, created_at)` queue index |
| storage | `memory_conflicts_list` / `_get` / `_resolve` |
| core-api | `GET /conflicts`, `GET /conflicts/{id}`, `PATCH /conflicts/{id}/resolve` |

The detector's proposal (`action`, `audit_reason`) and the reviewer's decision (`resolution_action`, `resolution_note`) are **separate fields drawn from the same `ACTIONS` vocabulary**, so *"did the human agree"* is an equality check rather than a mapping exercise. Collapsing them would destroy the only record of the detector being wrong — a `dismissed` row is the system's sole false-positive signal.

## Design decisions worth your review

**Trust-plane, not admin-plane.** `enforce_admin` admits only the super-admin key, which would put review out of reach of the people who actually run a tenant. But plain tenant scope would let **any agent dismiss the contradictions it caused** — corrupting the exact ground truth this surface exists to collect. Gated on `require_trust(min_level=2)` (the keystone-author bar) plus `enforce_read_only`.

**Resolve is compare-and-set** on `review_status='pending'`, returning **409** when the row has moved. Two reviewers working one queue is the ordinary case, not the edge case; without the compare, the second decision silently overwrites the first and the audit trail records only the loser's disappearance.

**`pending` is rejected as a resolve target** — re-opening would erase the record of who decided what.

**Missing and foreign conflicts both answer 404**, so the route can't be used to probe which ids exist in other tenants. The single `403` is the trust gate, which fires *before* any lookup and so reveals nothing.

**Queue is oldest-first.** Newest-first leaves the oldest unreviewed rows permanently at the bottom — precisely the backlog a reviewer is trying to clear.

## Testing

16 unit tests. Full suite **6,326 passed**; **26 failures, identical to clean main with these changes stashed** (verified by stashing — local test-DB seeding, not this change).

The `authz_gate_inventory` and migration-chain guards both caught real omissions during development — an unclassified router and an unpinned head — and both are now satisfied rather than suppressed.

## ⚠ Deploy prerequisite

`contradiction_write_conflict_record` is **default OFF**, so `memory_conflicts` is empty in every environment. The queue renders empty until that flag is enabled. **Not flipped here** — that's a separate decision.

Refs: canonical D11.

### DIFF
--- a/common/models/memory_conflict.py
+++ b/common/models/memory_conflict.py
@@ -39,6 +39,13 @@ DIAGNOSES = (
     "unresolved",
 )
 EVIDENCE_STRENGTHS = ("explicit", "entailed", "probabilistic")
+
+# D11 — human review state. A row starts ``pending``; a reviewer moves it to
+# ``resolved`` (they chose an action) or ``dismissed`` (not a real conflict).
+# ``dismissed`` is the ground-truth signal that matters most: it is the only
+# record that the detector was WRONG, and nothing else in the system captures a
+# false positive.
+REVIEW_STATUSES = ("pending", "resolved", "dismissed")
 ACTIONS = (
     "replace",
     "supersede",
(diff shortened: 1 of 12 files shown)


arkash20:
Rebased onto main; fixed the mypy failure (`rowcount` is on CursorResult, not Result — now uses the same type-ignore form as the sibling CAS updates in that file). CI green.

@Eldad-Caura please approve

Eldad-Caura (approved):

PR #1392, opened 2026-09-08
fix(extraction): keep bracketed qualifiers in canonical_name (A67)

arkash20:
## The bug

`EXTRACTION_PROMPT` specified `canonical_name: lowercase, no articles` and said nothing about qualifiers. A qualifier the content **brackets** — there precisely to say *which* thing is meant — was dropped, so two different things sharing a base name collapsed into one entity. Every later statement about either was then attributed to both, and a disagreement between two unrelated organizations reads as one organization contradicting itself.

## Measurement

Same sentences through both prompts against the live Vertex model, nothing else changed:

| Case | before (main) | after |
|---|---|---|
| `Zylo (Delaware)` / `Zylo (Ohio)` | `zylo` / `zylo` — **merged** | `zylo (delaware)` / `zylo (ohio)` |
| `Orbiter (v2)` / `Orbiter (v3)` | `orbiter` / `orbiter` — **merged** | `orbiter (v2)` / `orbiter (v3)` |
| `Priya at AcmeCorp` / `at BetaIndustries` (control) | `priya` / `priya` | `priya` / `priya` — unchanged |

End-to-end on a rebuilt stack: **0/3 cases wrong, 0 fake fallbacks, 0 dropped entities.**

## The bound is part of the fix

The rule covers only qualifiers the **content brackets**. Folding in nearby context would be the opposite failure and a worse one: `Priya at AcmeCorp` and `Priya at BetaIndustries` must both stay bare `priya`, because her next mention may carry no employer at all and all of them must still resolve to one person. That case is a negative control in both the repro and the tests.

## Why the examples are complete objects

Load-bearing, not stylistic. A first draft wrote them as `-> canonical_name: acme (delaware)`, naming a single field. The model read that as the output shape and **stopped emitting `role`** — a required `str` on `ExtractedEntity` — so `_parse_graph_lenient` dropped *every* entity, `_do_extract` raised on the total loss, and extraction fell through to the regex heuristic. That heuristic emits bare canonical names, which is indistinguishable from the very bug under test; it cost a full measurement round before the payload was captured. `test_examples_are_complete_entities` reads the required fields off the model, so adding a required field fails there rather than silently reintroducing the collapse.

## Testing

- 6 new unit tests pinning the rule, its bound, and the complete-example contract.
- Full suite: **27 failures, byte-identical to main's 27** (diffed by name) — none new.
- `ruff check` / `ruff format --check` clean on the changed files; `mypy src/` from `core-api/` matches main exactly.
- Wet-tested end to end on a rebuilt local stack with real Vertex extraction, with fake-fallback and dropped-entity counts verified at zero before trusting any number.

Closes **A68**, not A67.

A67 as filed is the **contextual** case — `Priya at AcmeCorp` vs `Priya at BetaIndustries`, where the discriminator is nearby context. That is exactly the case this change deliberately does *not* split, for the reason above, and it is the negative control here. A68 was split out for the bracketed variant this PR fixes; A67 stays open.

Also filed **A69** while measuring this: `ExtractedEntity.role` is a required `str`, so a payload omitting it on every entity collapses the whole extraction to the regex heuristic silently. Observed on `main` independently of this change (`"The billing service owner is Dana."` → `dropped={'entities': 2}`), same class as the `Mention.cluster_id` incident (#788).

### DIFF
--- a/core-api/src/core_api/services/entity_extraction.py
+++ b/core-api/src/core_api/services/entity_extraction.py
@@ -51,6 +51,20 @@ Rules:
   statement about a different subject, and the update would never supersede the
   value it replaces. Exactly one entity carries role=subject unless the content
   genuinely asserts facts about two independent subjects.
+- KEEP a bracketed qualifier the content attaches to a name. "Acme (Delaware)"
+  and "Acme (Ohio)" are two different organizations, so they are "acme
+  (delaware)" and "acme (ohio)" — never a bare "acme". Dropping the qualifier
+  collapses them into one entity, and every later statement about either is then
+  attributed to both. Each example below is a COMPLETE entity: the qualifier
+  rule changes the name, it never lets you omit entity_type or role.
+    "Acme (Delaware) filed in March."
+      {{"canonical_name": "acme (delaware)", "entity_type": "organization", "role": "subject"}}
+    "Acme (Ohio) filed in June."
+      {{"canonical_name": "acme (ohio)", "entity_type": "organization", "role": "subject"}}
+  Only qualifiers the CONTENT brackets. Do not invent one, and do not fold in
+  surrounding context that merely happens to be nearby — "Priya at AcmeCorp" is
+  still "priya", because the next mention of her may not carry the company and
+  the two must still resolve to one person.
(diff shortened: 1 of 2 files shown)


arkash20:

Eldad-Caura (approved):

PR #1394, opened 2026-09-08
chore(installer): classify remaining plugin identities

Eldad-Caura:
## Summary

- classify three exact installed-plugin path references as `legacy-name-floor`
- classify two OpenClaw configuration keys as `legacy-name-ok`
- reduce the OSS unclassified census from 129 to 124 on the refreshed branch

## Classification rationale

The two `agent_identity.py` paths are not aliases: they name the existing on-disk `.env` and `install.json` files used in recovery guidance. Renaming the text before migrating installations would direct users away from the real files, so they belong on the migration floor.

The wet installer default path is likewise a live installed path. Changing it would create or update a second plugin copy instead of the existing standard installation, so it is also a floor mention.

The two `memclaw` JavaScript keys are permanent aliases because the plugin manifest ID remains `memclaw`. Removing the allowlist value would make restrictive OpenClaw configurations reject the plugin; removing the `plugins.entries.memclaw` entry would stop enabling that manifest-ID plugin.

The remaining permanent aliases that cannot safely carry inline markers in Dockerfile ARG syntax, strict JSON, or the shipped skill hash remain explicitly classified in the brief-D ledger rather than being hidden from review.

## Verification

- `uv run --frozen ruff check core-api/src/core_api/services/agent_identity.py`
- `uv run --frozen ruff format --check core-api/src/core_api/services/agent_identity.py`
- `bash -n scripts/wet-test-install.sh`
- extracted embedded Node.js snippet parsed successfully
- `python3 scripts/do_not_touch_sentinel.py` — all 35 protected strings survive
- legacy-name census: 129 → 124 unclassified; same-shape `caura` control 5,242

The focused agent-identity pytest is locally blocked during setup by the checkout's PostgreSQL authentication; no credentials or database state were changed. CI is the authoritative integration check.

### DIFF
--- a/core-api/src/core_api/services/agent_identity.py
+++ b/core-api/src/core_api/services/agent_identity.py
@@ -41,10 +41,10 @@ RESERVED_WRITE_ID_MESSAGE = (
     'agent_id "{agent_id}" is reserved and no longer accepts writes — it is the '
     "unset plugin default and collides across every install. Retry with a "
     "unique, STABLE agent_id for THIS install: (1) set CAURA_AGENT_ID in "
-    '~/.openclaw/plugins/memclaw/.env to a stable name (e.g. "webclaw") and '
+    '~/.openclaw/plugins/memclaw/.env to a stable name (e.g. "webclaw") and '  # legacy-name-floor: existing installed path; migrate before changing recovery guidance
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #1395, opened 2026-09-08
feat(auth): give the caller identity its own type so stored data can't become one

Eldad-Caura:
## The bug class this closes

`authorize_memory_access` takes two agent ids, and they are different *kinds* of thing:

| parameter | where it comes from |
|---|---|
| `caller_agent_id` | `auth.agent_id`, `_get_agent_id()`, a resolver — **authenticated** |
| `owner_agent_id` | `memory.get("agent_id")`, `mem.get("agent_id")`, `getattr(memory, "agent_id", None)` — **a stored row** |

Both were `str | None`. And `memory_access_allowed_for_agent` decides `scope_agent` access with:

```python
if visibility == "scope_agent":
    return owner_agent_id == caller_agent_id
```

Pass a row's own `agent_id` as the **caller** and this returns `True` unconditionally — every access self-authorizes. A full silent bypass, and the signatures permitted it exactly.

`AgentIdentity` is a `NewType` over `str` for the authenticated caller. `owner_agent_id` **deliberately stays `str`** — it is data, and giving the two the same type is the confusion being prevented.

`NewType` is the identity function at runtime, so the *type* is runtime-neutral. The **rewrite was not**, and that is worth reading closely: constructing the type needs a None test, and the obvious `if resolved else None` turns an empty-string id into `None` in three resolvers.

`effective_agent_id` is pinned for exactly that input — `(None, "", "")` — because `enforce_self_agent` treats an explicit `""` as an **assertion** and refuses it, so collapsing it to `None` reads as *no assertion*. An existing test caught that one. `_resolve_read_identity` takes `caller_agent_id` straight off the request body with no `min_length`, so `""` is reachable there **from the wire** and was not covered by any test. All three now test `is not None` and are exact passthroughs.

> Note on what was *already* safe: transposing the two arguments was impossible before this change, because `owner_agent_id` is keyword-only. The `*` was already doing that job. This closes the other direction — a stored value passed *as* the caller.

## Why producers and sinks are one commit

Typing the five sinks alone leaves **20 mypy errors** (measured, not assumed), so a split would be red in CI. Nine `(file, function)` pairs may construct an identity; everything else receives one. `tests/test_agent_identity_construction.py` pins that list.

## Two corrections to the original plan

**`resolve_caller_and_gate` was on the sink list; it is a producer.** Its parameter is `body_agent_id` — the client's own claim — and it *returns* the resolved id. Typing that parameter `AgentIdentity` would have asserted that client input is authenticated, exactly backwards. Five sinks, not six.

**The MCP plane needed no design decision.** Its producer is `_agent_id_var`, the ContextVar holding the gateway-verified `X-Agent-ID` — the structural twin of `AuthContext.agent_id`. Typing it binds all eight MCP sites at once rather than needing eight casts.

## What the type found

Two sites resolved identity inline rather than through a resolver, so neither could be typed without a cast:

- **`routes/memories.py`** wrote the bare `auth.agent_id or agent_id` and fed it to `enforce_fleet_read`. That is *verbatim* the expression `AuthContext.effective_agent_id` exists to name — its docstring says the bare form "could not be told from an audit-log line of the same shape", and that `tests/test_authz_gate_inventory.py` **had to delete the rule that credited it**. Now calls the helper: identical value, one fewer bypass.
- **`mcp_server`** wrote `_get_agent_id() or agent_id` at eight sites. The one reaching `enforce_fleet_read_many` now routes through a new `effective_read_agent_id`, the read-side sibling of the existing `effective_write_agent_id`. The other seven feed no authorization sink and are **deliberately left alone** rather than swept.

## What was deliberately NOT typed

Three bare-form sites are **audit attributions**: `attribution_agent_id`, `conflicts.reviewer`, `skills_inbox.rejected_by_agent`. `effective_agent_id`'s docstring draws that line and `test_the_audit_attribution_is_not_bound_by_the_helper` pins it — an attribution records what the request carried *even when authorization ignored it*, which is a different value. Sweeping them would have broken that test and changed what the audit log records.

## One honest limitation

`core_api.auth` is still on the `ignore_errors` list, so nothing there forces the header-to-identity step. It is written explicitly anyway, at the single place `X-Agent-ID` is read, so the REST boundary is visible and recorded rather than an unchecked bare string. The signature still binds every *consumer* outside that module — which is why the REST sites type-check at all.

## Payoff, measured before starting

**32 of 34** sink call sites (94%) are now mypy-checked, against **7 of 32** before #1384 un-exempted `mcp_server`. The two blind ones are `routes.fleet:807` and `services.memory_service:3425` — un-exempting those first would have bought 2 sites of 34, so it was not worth doing first.

## Verification

- `mypy src/` clean (local baseline is 2 pre-existing `import-untyped` errors in `common/enrichment/service.py`, a stub package missing from the local venv).
- **Three mutation probes, all firing from a green baseline:** a construction outside the allow-list; a sink widened back to `str`; `owner_agent_id` given the caller's type. The last two matter because the allow-list alone passes trivially if the type is removed — nothing is constructed, so nothing is violated. Same mutual-coverage shape as the two error-code invariants in #1378.
- Root suite green.
- `ruff check` and `ruff format --check` clean on `core-api/src/` and `tests/`, each scope run separately.
- `scripts/legacy_name_ratchet.py --base origin/main` and `scripts/do_not_touch_sentinel.py` both exit 0.

### DIFF
--- a/core-api/src/core_api/agent_ids.py
+++ b/core-api/src/core_api/agent_ids.py
@@ -6,6 +6,28 @@ without importing each other — avoids the heavy/cross-private import that woul
 otherwise couple the route module to the whole MCP tool surface.
 """
 
+from typing import NewType
+
+# A caller identity that has been through authentication or resolution — not
+# merely a string that happens to hold an agent id. The distinction is
+# load-bearing rather than cosmetic: ``memory_access_allowed_for_agent`` decides
+# ``scope_agent`` access with ``owner_agent_id == caller_agent_id``, so a stored
+# row's own ``agent_id`` passed as the CALLER makes that comparison trivially
+# true and self-authorizes every access. Before this type both parameters were
+# ``str``, so nothing but review stopped it.
+#
+# Only the resolvers construct it, and ``tests/test_agent_identity_construction.py``
+# pins that list — a raw string cannot become an identity anywhere else. An
+# ``owner_agent_id`` read off a row deliberately stays ``str``: it is data, not
+# an identity, and giving the two the same type is what this prevents.
+#
+# NOT for audit attribution. ``AuthContext.effective_agent_id``'s docstring
+# draws that line and ``test_the_audit_attribution_is_not_bound_by_the_helper``
+# pins it: an attribution must record what the request carried even when
+# authorization ignored it, which is a different value from the identity
+# authorization used.
+AgentIdentity = NewType("AgentIdentity", str)
+
 # Reserved fallback identity used when a caller omits ``agent_id`` on the
 # single-tenant / standalone path. On the enterprise gateway this value is
 # explicitly refused (see ``mcp_server._refuse_default_agent_on_gateway``) so
@@ -66,7 +88,7 @@ ALWAYS_RESERVED_AGENT_IDS = frozenset({RESERVED_MAIN_AGENT_ID})
 _PLACEHOLDER_BODY_AGENT_IDS = frozenset({RESERVED_MAIN_AGENT_ID, DEFAULT_AGENT_ID})
 
 
-def effective_write_agent_id(verified_id: str | None, body_id: str | None) -> str | None:
+def effective_write_agent_id(verified_id: str | None, body_id: str | None) -> AgentIdentity | None:
     """Resolve the agent_id a WRITE is attributed to from the verified
     credential identity (gateway ``X-Agent-ID`` = ``home_agent_id``) and the
     client-supplied body id.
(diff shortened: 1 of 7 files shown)


Eldad-Caura:
@erni-a ready for review.

`AgentIdentity` — a NewType for the authenticated caller identity, with the five authorization sinks requiring it and construction confined to nine allow-listed resolvers.

**The reviewable judgement is the asymmetry.** `authorize_memory_access` takes two agent ids: `caller_agent_id` (authenticated) and `owner_agent_id` (read off a stored row). Only the caller gets the type. `owner_agent_id` deliberately stays `str`, because `scope_agent` access is decided by `owner_agent_id == caller_agent_id` — pass a row's own id as the *caller* and every access self-authorizes. Typing both the same would have preserved that hazard while adding ceremony. `test_owner_agent_id_is_not_an_identity` pins it.

Two corrections to the original plan, both from measurement: `resolve_caller_and_gate` was on the sink list but is a *producer* (its parameter is the client's own claim, and typing it would assert client input is authenticated); and the MCP plane needed no design decision, because its producer is the ContextVar holding the verified `X-Agent-ID`, so typing that bound all eight MCP sites at once.

**Two things the type found**, both places resolving identity inline instead of via a resolver: `routes/memories.py` used the bare `auth.agent_id or agent_id` — verbatim the expression `AuthContext.effective_agent_id` exists to name, whose docstring records that `test_authz_gate_inventory.py` had to delete the rule that credited it — and `mcp_server` did the same at eight sites, of which the one reaching a sink now uses a new `effective_read_agent_id`.

**Three sites deliberately NOT typed:** `attribution_agent_id`, `conflicts.reviewer`, `skills_inbox.rejected_by_agent` are audit attributions, and `test_the_audit_attribution_is_not_bound_by_the_helper` pins that as a distinct value. Sweeping them would have broken that test and changed what the audit log records.

**One regression of mine, caught by the suite and worth your eye:** the type is runtime-neutral, but the rewrite wasn't automatically so. The obvious `if resolved else None` collapsed an empty-string id to `None` in three resolvers, and `""` is a meaningful assertion that `enforce_self_agent` refuses. One had a pinning test; `_resolve_read_identity` takes `caller_agent_id` off the request body with no `min_length`, so `""` is reachable from the wire and was untested. All three now use `is not None` and are exact passthroughs.

Also flagging: `core_api.auth` is still mypy-exempt, so nothing there *forces* the header-to-identity step. I wrote it explicitly at the one place `X-Agent-ID` is read so the boundary is visible and allow-listed, but mypy is not checking it — the PR says so rather than implying a guarantee that isn't there.

All 14 check-runs green on head `1e32f40b` (verified by SHA, not `gh pr checks`), including `claude-review`. Root suite 6355 passed. Three mutation probes fire from a green baseline: a construction outside the allow-list, a sink widened back to `str`, and `owner_agent_id` given the caller's type — the last two matter because the allow-list alone passes trivially if the type is deleted.

Payoff measured before starting: 32 of 34 sink call sites now mypy-checked, against 7 of 32 before #1384. The 2 remaining blind ones are `routes.fleet:807` and `services.memory_service:3425`.

Not merging — yours.

erni-a (approved):

PR #1396, opened 2026-09-08
refactor(routes): pass the write identity down instead of re-reading a nullable field

Eldad-Caura:
## What

Removes both `type: ignore[arg-type]` that #1387 added, by doing what its own comment said the fix was: have the inner write functions **take** the resolved identity instead of re-reading `body.agent_id`.

## Why the ignores existed

The invariant was always true — `write_memory` / `write_memories_bulk` either fill the reserved standalone identity or raise `_missing_agent_id_error`. It just could not be *checked*: the guard is in the outer function and the call is in the inner one, the field stays `str | None` on the model, and the standalone branch rebinds `body` via `model_copy`.

Binding the guarded value to a local is what fixes it — the narrowing then survives both the rebind and the function boundary:

```python
chosen_agent_id = body.agent_id
if not chosen_agent_id:
    if app_settings.is_standalone:
        chosen_agent_id = DEFAULT_AGENT_ID
        body = body.model_copy(update={"agent_id": chosen_agent_id})
    else:
        raise _missing_agent_id_error()
```

The parameter is deliberately `str` and **not** `AgentIdentity`. At that point it is still the caller's *claim*; `resolve_write_agent` is what turns a claim into an identity. Typing it `AgentIdentity` would assert the opposite of what is true there.

The long comment at each call site is **gone rather than moved**. It warned against "fixing" the ignore with `or DEFAULT_AGENT_ID`; with a non-optional parameter there is no ignore to fix, and the reason the outer guard raises instead of defaulting is already stated in the outer function where that decision is made.

Worth knowing: `warn_unused_ignores = false`, so nothing would have flagged those ignores as stale. Removing them was deliberate, not prompted by a failure.

## A gap this refactor exposed

Both inner functions apply a dark Phase 2 override, and the new local has to be rebound there too:

```python
if app_settings.bind_write_identity_to_auth and auth.agent_id:
    chosen_agent_id = auth.agent_id   # <- added here
    body.agent_id = auth.agent_id
```

Forgetting that line would leave `body.agent_id` assigned — so the flag would still **look** applied — while `resolve_write_agent` received the body's original claim. The spoof-hardening control would be a no-op.

**Nothing in the suite exercised this flag enabled.** Every existing reference sets it `False` or discusses it in prose, so that regression would have been silent and green.

Three tests now cover it:

| test | pins |
|---|---|
| `..._over_the_body_single` | verified identity beats the body's claim, single write |
| `..._over_the_body_bulk` | same on the bulk path |
| `test_phase2_off_leaves_the_body_identity_alone_single` | dark by default — the body's claim still wins |

Both "enabled" tests fail without the rebind, asserting `'body-claimed-agent' == 'verified-agent'` — the no-op made visible. The flag ships dark, so this is the first coverage it has had.

## Verification

- `mypy src/` clean **with both ignores removed** (local baseline is 2 pre-existing `import-untyped` errors in `common/enrichment/service.py`).
- **Two structural probes, from a baseline of 0 errors:** re-reading the field at the call site errors, and so does passing `body.agent_id` from the *outer* function. The second is the one that matters — it shows the guarded local, not the parameter by itself, is doing the work.
- **Both new "enabled" tests fail without the rebind** they cover.
- Root suite green.
- `ruff check` and `ruff format --check` clean on `core-api/src/` and `tests/`, each scope run separately.
- `scripts/legacy_name_ratchet.py --base origin/main` and `scripts/do_not_touch_sentinel.py` both exit 0.

### DIFF
--- a/core-api/src/core_api/routes/memories.py
+++ b/core-api/src/core_api/routes/memories.py
@@ -1067,9 +1067,17 @@ async def write_memory(
     # anonymous write would collapse onto one shared identity — the same footgun
     # mcp_server._refuse_default_agent_on_gateway guards against. Keep that as an
     # explicit 422 rather than a silent default.
-    if not body.agent_id:
+    #
+    # Bound to a local so the narrowing SURVIVES: ``model_copy`` rebinds
+    # ``body`` and the field stays ``str | None`` on the model, so re-reading
+    # ``body.agent_id`` in the inner function threw the guarantee away and
+    # needed a ``type: ignore`` there. Passed down instead — see
+    # ``_write_memory_inner``'s ``chosen_agent_id``.
+    chosen_agent_id = body.agent_id
+    if not chosen_agent_id:
         if app_settings.is_standalone:
-            body = body.model_copy(update={"agent_id": DEFAULT_AGENT_ID})
+            chosen_agent_id = DEFAULT_AGENT_ID
+            body = body.model_copy(update={"agent_id": chosen_agent_id})
         else:
             raise _missing_agent_id_error()
     # Idempotency replay is short-circuited BEFORE the per-tenant slot —
@@ -1101,7 +1109,7 @@ async def write_memory(
     # queueing requests until they time out at the worker layer. Only
     # the new-write path is gated; replays returned above bypass it.
     async with per_tenant_slot("write", body.tenant_id):
-        return await _write_memory_inner(body, response, auth, _idem)
+        return await _write_memory_inner(body, response, auth, _idem, chosen_agent_id)
 
 
 async def _write_memory_inner(
@@ -1109,6 +1117,11 @@ async def _write_memory_inner(
     response: Response,
     auth: AuthContext,
     idem: IdempotencyGuard | None,
+    # The write identity the caller asked for, already guarded non-None by
+    # ``write_memory``. Deliberately ``str``, not ``AgentIdentity``: this is
+    # the caller's CLAIM, and ``resolve_write_agent`` is what turns a
+    # claim into an identity.
+    chosen_agent_id: str,
 ):
     from core_api.services.organization_settings import resolve_config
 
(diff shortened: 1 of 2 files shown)


Eldad-Caura:
@erni-a ready for review.

Removes both `type: ignore[arg-type]` from #1387 by doing what its own comment said the fix was: the inner write functions now TAKE the identity instead of re-reading `body.agent_id`. The mechanism is binding the guarded value to a local in the outer function, so the narrowing survives both `model_copy`'s rebind and the function boundary.

The parameter is deliberately `str` and not `AgentIdentity` — at that point it is still the caller's *claim*, and `resolve_write_agent` is what turns a claim into an identity.

**The part worth your attention is not the refactor.** Both inner functions carry the dark Phase 2 override, and the new local has to be rebound inside it:

```python
if app_settings.bind_write_identity_to_auth and auth.agent_id:
    chosen_agent_id = auth.agent_id   # <- added here
    body.agent_id = auth.agent_id
```

Miss that line and `body.agent_id` is still assigned — so the flag still *looks* applied — while `resolve_write_agent` receives the body's original claim. The spoof-hardening control becomes a no-op.

**Nothing in the suite exercised that flag enabled.** Every existing reference sets it `False` or discusses it in prose, so that regression would have been silent and green, and would have surfaced the day the flag is turned on as a control that does nothing. Three tests now cover it (verified identity beats the body's claim on both write paths; flag off leaves the claim alone). Both "enabled" tests fail without the rebind, asserting `'body-claimed-agent' == 'verified-agent'`. This is the first coverage the flag has had.

Also worth knowing: `warn_unused_ignores = false`, so nothing would have flagged those two ignores as stale. Removing them was deliberate, not prompted by a failure — the ratchet can accumulate ignores that no longer suppress anything and no gate will say so.

All 14 check-runs green on head `3cb00d2a` (verified by SHA, not `gh pr checks`), including `claude-review`. Root suite 6364 passed, reconciled as 6355 + 6 from #1392's new file + my 3, so nothing was silently skipped. Two structural probes fire from a baseline of 0: re-reading the field at the call site errors, and so does passing `body.agent_id` from the OUTER function — the second is what shows the guarded local rather than the parameter alone is doing the work.

Not merging — yours.

erni-a (approved):

PR #1397, opened 2026-09-08
fix(extraction): make a degraded extraction visible, and stop one absent field causing it (A69)

arkash20:
Two gaps on the path where extraction stops using the model and starts using a regex.

## 1. A missing `role` discarded the entity carrying it

`ExtractedEntity.role` was a required `str`, and nothing enforces the schema `_do_extract` sends. When the model omitted it on *every* entity the loss was total, `_do_extract` raised, and the chain fell through to `_fake_extract` — which throws away the model's canonical names and types and substitutes bare capitalised bigrams typed `unknown`. Same class as the `Mention.cluster_id` prod incident (#788).

`role` now defaults to `"mentioned"`. That's the *safe* default, not just the convenient one: subject write-back fires only on exactly one `role="subject"` entity, so defaulting can never invent a subject — only decline to name one, which is what an absent field actually means. `canonical_name` keeps no default, so a genuinely malformed item is still dropped.

## 2. Nothing said the degrade had happened

The chain logs `"All LLM providers failed for %s"` with the service interpolated into the *message*, so it can't be filtered or counted per service. Downstream, nothing distinguishes a heuristic graph from a real one — rows are persisted and names embedded either way. Extraction now emits a structured, greppable `entity_extraction_degraded_to_heuristic` at ERROR. An intentional `fake`/`none` provider short-circuits before the chain and is **not** reported as a degrade; that distinction is pinned by a test.

## 3. And the zero-subject branch was silent

`subject_writeback` logged the "set" case and the "found two and refused" case, and logged **nothing** when no entity claimed a subject — the same "ran and found nothing" vs "never ran" ambiguity D3 closed for the contradiction detector. Zero subjects is also the signature of a degraded graph, since `_fake_extract` stamps `role="mentioned"` on everything.

## Wet test

On a rebuilt stack:

- degrade line fires: `entity_extraction_degraded_to_heuristic provider=gemini content_len=32`
- a role-less payload that previously yielded **zero** entities now keeps both
- `skipped_no_subject` fired **6 times across 3 ordinary live writes**, with **0** degrade events — so subject-less extractions happen in normal operation, and until now every one was invisible

## Scope — stated so it isn't read as more than it is

After A68 (#1392) merged, the missing-`role` trigger **no longer reproduces** against the live model: 0 omissions across 18 sentences, including deliberately messy multi-entity, code-like and non-English content. The default is hardening against a latent trigger. **The observability half is the live defect** — and it's what makes the latent one expensive, having already cost a full measurement round on A68.

## Testing

- 9 new unit tests: the default and its safety property, that malformed items are still dropped, that the heuristic never claims a subject, the ERROR slug, the intentional-fake carve-out, and both worker branch shapes.
- `ruff check` / `ruff format --check` clean; `mypy src/` from `core-api/` matches main.

Closes A69.

### DIFF
--- a/core-api/src/core_api/services/entity_extraction.py
+++ b/core-api/src/core_api/services/entity_extraction.py
@@ -87,7 +87,23 @@ Content:
 class ExtractedEntity(BaseModel):
     canonical_name: str
     entity_type: str
-    role: str
+    # Defaulted, not required, for the reason ``Mention.cluster_id`` is nullable
+    # (prod, 2026-08-16, #788): nothing enforces the schema ``_do_extract``
+    # sends, so ONE field the model declines to emit used to discard the entity
+    # carrying it. When the model omits ``role`` on EVERY entity the loss is
+    # total, ``_do_extract`` raises, and the chain falls through to
+    # ``_fake_extract`` — which throws away the model's canonical names and
+    # types and substitutes bare regex bigrams typed ``unknown``. Measured:
+    # a draft of the A68 prompt whose examples named a single field taught the
+    # model to stop emitting ``role``, and the resulting regex output was
+    # indistinguishable from the discriminator bug under test.
+    #
+    # "mentioned" is the honest default. It is the same value ``_fake_extract``
+    # already stamps when it cannot classify, and it is the SAFE one: the
+    # subject write-back below only fires on exactly one ``role="subject"``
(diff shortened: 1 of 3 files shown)


arkash20:

Eldad-Caura (approved):

PR #1410, opened 2026-09-08
feat(search): opt-in strict fleet scoping (C27)

arkash20:
## What this is — and is not

Wire contract **D4 (RATIFIED)** defines a NULL `fleet_id` as tenant-shared **by design**: a row written without a fleet is readable by every fleet in the tenant. This PR does **not** change that, and does not treat it as a bug. It adds the switch for tenants that want hard fleet isolation, and leaves the contract's behaviour as the default.

`search.strict_fleet_scoping` — per tenant, **default `False`**.

Off by default for the same reason as `recall_for_asserted_identity`: turning it on **hides rows that are visible today**, so it is a decision a tenant makes, never one inherited from a deploy. Every other search knob changes what *ranks*; this one changes what is *visible*.

`scope_org` survives in **both** modes — an explicit visibility tier a writer chose, not an accident of a missing fleet. A tenant asking for fleet isolation isn't asking to revoke org-wide sharing.

## One predicate, one place

A54 is why. The identical clause was copied across several candidate queries, the copy named in the task was fixed, and the live leak reproduced through the next copy. Everything now builds its clause in `_fleet_scope_clause`, and `test_every_fleet_scoped_read_goes_through_the_helper` fails on a hand-rolled **copy** rather than waiting for the leak it would cause.

Threaded through **every** fleet-scoped read, not just the obvious one:

| Read | Why it matters |
|---|---|
| `memory_scored_search` | the main recall |
| `memory_load_by_ids` | the `ENTITY_LOOKUP` short-circuit — bypasses scored search entirely, so it needs the flag in its own right |
| `memory_find_successors` | a successor is injected straight into the result set; an unscoped lookup reintroduces exactly what the scored query just excluded |
| `entity_fts_search` | entity route (no `visibility` column, so the caller opts out explicitly) |

On the pipeline **and** the legacy search path, so the mode can't depend on which implementation served the request.

Storage params are optional with a permissive default, so a storage instance deployed ahead of core-api keeps serving callers that don't send them — the same independence A54's params were given.

## Wet test

Storage predicate, isolated:

```
strict=False -> in_fleet=YES  null_fleet=YES
strict=True  -> in_fleet=YES  null_fleet=NO
```

End to end through `/api/v1/search` with a genuinely null-fleet row: **PASS** in both modes.

⚠️ The switch took **~30s** to take effect. Settings are cached per worker (5-min TTL) and a `PUT` only invalidates the worker that served it, so a tenant flipping this should expect propagation, not immediacy. Worth knowing before someone reports it as broken.

Two things the wet test caught that would otherwise have produced a false pass: a null-fleet row **cannot** be created through the write API (an agent's home fleet is frozen at first write, so omitting `fleet_id` inherits it), and my first attempt used contradicting contents, so contradiction detection marked one row `conflicted` and default status filtering hid it.

## Testing

- 12 new unit tests, compiling the **real** clause to SQL rather than asserting against stub columns.
- `ruff check` / `ruff format --check` clean. `mypy src/` clean on core-storage-api; core-api matches main.

Closes C27.

### DIFF
--- a/core-api/src/core_api/pipeline/steps/search/classify_query.py
+++ b/core-api/src/core_api/pipeline/steps/search/classify_query.py
@@ -157,6 +157,7 @@ class ClassifyQuery:
                         status_filter=status_filter,
                         valid_at=valid_at,
                         readable_tenant_ids=readable_tenant_ids,
+                        strict_fleet_scoping=bool(ctx.data.get("strict_fleet_scoping")),
                         slot_acquired_marker=ctx.data,
                     )
 
@@ -379,6 +380,7 @@ class ClassifyQuery:
         valid_at: datetime | None = None,
         readable_tenant_ids: list[str] | None = None,
         slot_acquired_marker: dict | None = None,
+        strict_fleet_scoping: bool = False,
     ) -> list[types.SimpleNamespace]:
         """Load memories linked to graph-expanded entities, scored by hop distance."""
         all_entity_ids = list(entity_hops.keys())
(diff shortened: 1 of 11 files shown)


arkash20:

Eldad-Caura (approved):

PR #1411, opened 2026-09-08
feat(search): raise the top_k ceiling to 200 and give it one source of truth

yanki-m:
## Summary

Raise the search `top_k` ceiling from 20 to **200** and make it **one constant** (`common.constants.MAX_SEARCH_TOP_K`) that REST `/search`, MCP `caura_recall` and the tunable search profile all read. This is a **ceiling, not a value**: the default (`DEFAULT_SEARCH_TOP_K = 5`) is unchanged, and every Caura surface — MCP tool, OpenClaw plugin, Python and TypeScript clients — asks for 5 unless the caller says otherwise. Nothing changes for a caller who does not ask for more.

## Why

- The cap was encoded twice with nothing tying the copies together: `MAX_SEARCH_TOP_K = 20` in `core_api.constants` and a literal `(1, 20)` in `common.constants.SEARCH_KNOBS`.
- Twenty is low for a memory layer asked to fill an answer context. On the AMB personamem/32k leaderboard set, Caura was the only entry whose slot count is set by the server rather than the caller — hybrid-search and cognee retrieve 50, Hindsight uses a token budget. The benchmark harness compensated with 8k-character chunks: comparable context (~26k tokens) with the coarsest selection of any entry (20 chunks of ~1,300 tokens vs 50 of ~500).
- Measured on 2026-09-09 under the leaderboard's answer model (`gemini-3.1-pro-preview`): Caura bulk at 20 slots scores **83.0%**, against the raw hybrid-RAG baseline's 84.4% at 50 slots and Hindsight's 86.6%. A harness-side 3-query fusion emulating 40 slots had measured **+3.6 points** on 2k chunks (6 Sep) at 3× the requests — the cap was the only reason that workaround existed.
- 200 rather than 50 is an operator decision: leave room for token-budget-style callers and richer stores without another contract bump.

## What changed

| file | change |
|---|---|
| `common/constants.py` | `MAX_SEARCH_TOP_K = 200` beside `SEARCH_KNOBS`; `top_k` bounds read it |
| `core-api/src/core_api/constants.py` | imports + re-exports the constant instead of a second copy |
| `core-api/src/core_api/mcp_server.py` | tool description derived from the constant |
| `core-api/openapi.broker.json`, `plugin/tools.json`, `tests/fixtures/tools_list_baseline_v1.json` | regenerated (`gen_broker_openapi.py`, `export_tool_specs.py`, `capture_baselines.py`); `maximum: 20 → 200` is a non-breaking widening |
| `plugin/src/tool-definitions.ts`, `static/docs/integration-guide.md` | descriptions updated |
| `tests/test_search_default_profile.py` | drift guard: REST `le=`, knob bounds and the core-api re-export all equal the one constant |
| `tests/test_pre_public_cleanup_regressions.py` | the description assertion reads `MAX_SEARCH_TOP_K` instead of a literal |

## Cost, stated plainly

Each returned row costs ~45 ms today (recall_count bump, successor lookup, serialization — p50 675 / 968 / 1,545 ms at k = 1 / 5 / 20 on a 24-row store). A caller asking for 200 pays roughly **9 s** and reinforces 200 rows' `recall_count`. The over-fetch is `top_k × 2` candidates before the relevance floor. Two follow-ups are deliberately **not** in this PR and become more important at this ceiling:

1. batch the `recall_count` bump into one `UPDATE … WHERE id IN (…)`;
2. report server time on `/search` (`Server-Timing` or `took_ms`) so the cost of asking for more is visible to the caller who asked.

## Verification

- 59 passed across the affected test files against a throwaway `pgvector/pgvector:pg16`; the two local failures (`test_v1_baseline_matches_live_registry`, tools/list token budget) are environment-dependent description text that CI does not reproduce — CI was green on the previous commit with the same fixture regenerated the same way, and the fixture diff is the single string `"1-50." → "1-200."`.
- Direct check: `SearchRequest(top_k=200)` accepted, `201` rejected with `ValidationError`; MCP description renders `1-200.`; knob bounds `(1, 200)`.
- `ruff check` clean on the changed Python files.

## Follow-up experiment (benchmark side)

Once deployed: `CAURA_TOP_K=50` with 2k chunks (hybrid-search's exact shape: ~50 × ~500 tokens), single query, against the 20-slot 8k result above. That is the run that tells us whether Caura's ranking earns anything once it is allowed the same shape as the baseline that beats it by 1.4 points. Also of note: the CAURA-722 commit message (`9936f7a`) states MCP `caura_recall` has no `top_k` cap — it is clamped at `mcp_server.py:1007`.


Eldad-Caura:
@claude

Eldad-Caura:
@claude

Eldad-Caura:
@claude

Eldad-Caura:
Reviewed the new performance/DoS finding against the current implementation.

The premise in the new `MAX_SEARCH_TOP_K` comment was stale, so I corrected it and amended the single commit (`c669d8a2`):

- `TrackRecalls` dispatches one fire-and-forget `increment-recall` request containing the full ID list; storage executes one `UPDATE ... WHERE id IN (...)`.
- Successor enrichment issues at most one `find_successors` storage call for the full stale-ID list.
- Core API does not hold a database session across search; the scored query is storage-routed.
- REST `/search` already has the existing 30/s API-key rate limit, 32-request per-tenant route bulkhead, four-call per-tenant storage-search bulkhead, and 45s request timeout. MCP recall uses the same storage-search bulkhead and retains the opt-in default of 5.

So I did not add redundant batching or a cap-specific timeout. The actual tradeoff is larger scored-query overfetch, local processing, and response payloads for callers that explicitly request more than the default, which the corrected comment now states.

Validation: Ruff check/format passed; the focused shared-cap regression passed when invoked against the worktree sources. The normal pytest entrypoint is still blocked before the test body by the known local PostgreSQL password mismatch; the prior PR head was fully green in CI.

Eldad-Caura:
@erni-a

erni-a (approved):

PR #1414, opened 2026-09-09
fix(crystallizer): stop the sweep destroying knowledge, re-paying LLM calls, and mislabelling fleet scope

arkash20:
Four low-severity audit findings on the same nightly sweep. From the shared remediation tracker: `oss-0902-l-43`, `oss-0902-l-30`, `oss-0902-l-31`, `oss-0814-l-09`.

## `oss-0902-l-43` — the Type-II sweep re-paid for unchanged subjects

`run_shadow` takes a `since` watermark and `select_candidates` uses it to skip subjects with nothing new. Nobody passed it, so every nightly run re-asked the LLM about **every** subject with ≥2 live memories, buying the same answers again.

It now takes the last **completed** run's timestamp. That can't be `get_latest_report` alone: this run has already reserved its own `status="running"` row, so the "latest" report is frequently *this* run — using it would stamp the watermark at now and skip every subject, turning an overspend into a sweep that silently does nothing.

The helper refuses its own row, non-terminal rows, and non-string timestamps, and resolves every doubt to `None` ("scan everything"). That asymmetry is deliberate: **paying twice is recoverable on the next run; skipping a changed subject is not.**

## `oss-0902-l-30` — sources archived with nothing to replace them

The archive step ran unconditionally after the fact-creation loop. Per the loop's own comment, a crystallized fact is a near-verbatim merge of cluster members that are ≥0.95 similar and **still active** at that point — so `create_memory`'s dedup gate 409s against a cluster member, *the very row about to be archived*. A cluster whose facts were all rejected ended with every source archived and nothing in their place.

Gated on `new_ids`, deliberately **not** on the absence of duplicates: what licenses retiring the sources is that a replacement exists. One fact created and two duplicates skipped is still safe to archive; zero created never is.

## `oss-0902-l-31` — fleet-scoped reports carried tenant-wide counts

Five hygiene checks took `fleet_id` and never applied it. The parameter was never the hard part — **the storage layer has always accepted a fleet argument for all five.** The scoping was thrown away in transit:

```python
rows = await _svc.entity_find_orphaned(tenant_id, fleet_id=None)   # hardcoded
expired = await _svc.memory_find_expired_still_active(tenant_id, None)  # positional None
```

Threaded through route, client and caller; optional with `None` defaults so storage can deploy ahead of core-api.

*Note on the count:* the audit says five checks. Four ignore the parameter outright; the fifth (`_check_broken_entity_links`) passes it to a client that drops it. The audit was right.

## `oss-0814-l-09` — `POST /crystallize/all` answered 500 off standalone

`get_standalone_tenant_id()` raises when standalone was never initialised, nothing caught it, and every hosted call returned "internal server error" — which reads as an outage and sends on-call hunting a broken crystallizer. The endpoint is standalone-only by construction (no tenant enumeration to fan out over), so this is a deployment mode, not a fault. Now 501, naming the route that does work.

## Testing

- 26 new unit tests across the four findings.
- `ruff check` / `ruff format --check` clean; `mypy` clean on core-storage-api, unchanged on core-api.

Tracker rows move to done on merge.

### DIFF
--- a/core-api/src/core_api/clients/storage_client.py
+++ b/core-api/src/core_api/clients/storage_client.py
@@ -1260,8 +1260,11 @@ class CoreStorageClient:
             params["fleet_id"] = fleet_id
         return await self._get_list("/memories/recent", **params)
 
-    async def get_lifecycle_candidates(self, tenant_id: str) -> dict:
-        return await self._get("/memories/lifecycle-candidates", tenant_id=tenant_id) or {}
+    async def get_lifecycle_candidates(self, tenant_id: str, fleet_id: str | None = None) -> dict:
+        params: dict[str, Any] = {"tenant_id": tenant_id}
+        if fleet_id is not None:
+            params["fleet_id"] = fleet_id
+        return await self._get("/memories/lifecycle-candidates", **params) or {}
 
     async def check_near_duplicates(self, data: dict) -> dict:
         return await self._post("/memories/near-duplicates", data)  # type: ignore[return-value]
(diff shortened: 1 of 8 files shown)


arkash20:

Eldad-Caura (approved):

PR #1418, opened 2026-09-09
fix(ingest): stop dropping CJK facts, honour the cache contract, delete a dead constant

arkash20:
Three low-severity findings in `ingest_service`, from the shared remediation tracker: `oss-0902-l-37`, `oss-0814-l-28`, `oss-0814-l-49`.

## `oss-0902-l-37` — the fragment filter deleted every CJK fact

```python
if len(body.split()) < _MIN_FACT_WORDS:   # 5
```

That assumes words are separated by spaces. A whole Chinese or Japanese sentence is **one** whitespace token, so the guard written to drop headings and one-word labels was dropping complete, useful facts — silently, counted as `dropped_short`, and only for those languages.

`_fact_word_count` counts each CJK character as one unit and splits the remainder on whitespace, so mixed text (`"Acme の売上は増加した"`) is counted once rather than twice.

Two deliberate calls:

- **One character per unit is generous.** A CJK word is typically one to two characters, so this errs toward *keeping* a short fact rather than dropping a real one — the failure being fixed. A genuine two-character heading (`概要`) is still below the floor.
- **Hangul is excluded.** Korean *is* space-delimited, so counting syllables would over-count it and let real fragments through — the opposite bug.

## `oss-0814-l-28` — the cache hit broke the contract it documents

`ingest_preview`'s own docstring says:

> `doc_hash` — sha256 of (tenant, content); caller echoes to commit for cache

The cache-hit branch didn't return it. So a client following the documented flow lost the hash **precisely when the cache was working**, and the second ingest of a document could never cache. Both exits now return it.

The branch also reports `sections: 0`, for the same reason it already reports `chunk_ms: 0` — zero LLM calls is the true count, while an absent key reads as "unknown".

## `oss-0814-l-49` — a dead constant documenting behaviour that doesn't exist

`_INGEST_MAX_CONTENT_CHARS` was defined once, never read, and its comment promised truncation with `truncated: true` and `original_length` response fields — which the same function's docstring flatly contradicts ("no longer truncated post-PR#7"). A comment describing behaviour the service doesn't have is worse than no comment. Removed; the live `_INGEST_MIN_CONTENT_CHARS` is untouched, and a test pins that.

## Testing

- 12 new unit tests, including that English counting is unchanged, that a short CJK fragment is *still* dropped, and that the call site actually uses the new helper.
- `ruff check` / `ruff format --check` clean; `mypy` unchanged from main.

Tracker rows move to done on merge.

### DIFF
--- a/core-api/src/core_api/services/ingest_service.py
+++ b/core-api/src/core_api/services/ingest_service.py
@@ -123,14 +123,6 @@ _CLOUD_METADATA_IPS = frozenset({"169.254.169.254", "fd00:ec2::254"})
 # latency.
 _PREVIEW_CONCURRENCY = 4
 
-# Maximum content length the LLM sees. Inputs longer than this get
-# truncated; ``ingest_preview`` reports the post-truncate length as
-# ``content_length`` and sets ``truncated: true`` + ``original_length``
-# so callers know the input was clipped. (Previously ``content_length``
-# returned the pre-truncate length, lying about what the LLM actually
-# processed.)
-_INGEST_MAX_CONTENT_CHARS = 50_000
-
 # Minimum content length before we'll even call the LLM. Whitespace-only
 # inputs and trivially short ones ("hi") used to burn a real LLM call
 # producing useless meta-facts ("The content begins with the greeting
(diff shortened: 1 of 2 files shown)


arkash20:

Eldad-Caura (approved):

PR #1419, opened 2026-09-09
fix: narrow the Path C revert, drop three unreachable builders, clamp the interviewer flag

arkash20:
Three unrelated low-severity findings from the shared remediation tracker, grouped only by size: `oss-0902-l-29`, `oss-0814-l-48`, `oss-0902-l-09`. Closes the 10-item low-priority batch.

## `oss-0902-l-29` — the retraction overwrote statuses it never set

Step 1 of the two-step retraction wrote the literal `"active"` over whatever the candidate held at that moment. Detection marks a losing candidate `outdated` or `conflicted`; **anything else means another writer has moved the row since** — confirmed by a human, archived by the crystallizer, superseded by a different chain — and stamping `"active"` over that discards someone else's decision in order to undo our own.

Now guarded on the statuses detection actually assigns, with a skipped revert logged alongside what it found. That mirrors the guard the content-edit reset has always had:

```python
if mem.get("status") in ("outdated", "conflicted"):
    patch["status"] = "active"
```

**The audit is half wrong here, and it's worth saying:** it claims the content-edit reset shares this bug. It doesn't — it was already guarded, which is exactly why only the retraction was losing statuses.

**What this does not do:** it restores `active`, not the status the row held *before* detection. A `confirmed` row marked `conflicted` still returns as `active`, because nothing records what was overwritten — neither `memories` nor `memory_conflicts` has a prior-status column. Narrowing the write is the part fixable without a migration; recovering the original value isn't.

## `oss-0814-l-48` — three builders nothing could reach

`build_quick_entity_linking_pipeline`, `build_link_discovery_pipeline` and `build_relation_inference_pipeline` had **zero references anywhere in the repo** — not a caller, not an export, not a test. Only the nightly `build_full_entity_linking_pipeline` is used (by `lifecycle_audit`), and a test pins that the live one survived, because deleting the wrong one of the four would take entity linking down entirely.

## `oss-0902-l-09` — a flag that could stall a transcript permanently

`--max-event-chars` accepted any integer. The server truncates each event at `INTERVIEW_EVENT_MAX_CHARS` (8000) and **rejects** a window exceeding it — and a rejected window never advances the cursor, so an over-large value didn't degrade the transcript, it stalled it, 422 after 422.

Clamped on both subcommands and announced on stderr: a silently-overruled flag leaves the operator believing a setting that isn't in force. Zero and negative are refused rather than clamped upward, which would invent a value nobody asked for.

Tests live in the client package's own suite — `caura_client` is a separate distribution and isn't importable from the core-api tests. The client constant is necessarily a *copy* of the server's, so a test pins the comment naming its source; nothing else links them.

## Testing

- 9 new core-api tests + 7 in the client suite.
- core-api: **27 failures, identical to main's 27** — none new.
- client: **145 passed**, with the same single pre-existing failure as main.
- `ruff check` clean on both (client CI's exact command). `mypy` unchanged.
- `cli.py` is deliberately left unformatted: it already drifts from `ruff format` on main, and reformatting would rewrite untouched `legacy-name-ok` lines.

### DIFF
--- a/clients/python/src/caura_client/interviewer/cli.py
+++ b/clients/python/src/caura_client/interviewer/cli.py
@@ -65,6 +65,31 @@ def _read_env(*names: str, default: str = "") -> str:
     return "" if saw_blank else default
 
 
+# The server truncates each event's content at this many characters
+# (``INTERVIEW_EVENT_MAX_CHARS`` in core-api) and REJECTS a window whose events
+# exceed it. The CLI accepted any integer, so ``--max-event-chars 20000`` made
+# every submission 422 — and because a rejected window never advances the
+# cursor, the transcript stalled permanently rather than degrading. Clamped
+# here, loudly, because a silently-honoured flag that breaks every request is
+# worse than one that says it was overruled.
+_SERVER_MAX_EVENT_CHARS = 8_000
+
+
+def _event_chars(value: str) -> int:
+    """``--max-event-chars`` clamped to what the server will actually accept."""
+    n = int(value)
+    if n < 1:
+        raise argparse.ArgumentTypeError("--max-event-chars must be positive")
+    if n > _SERVER_MAX_EVENT_CHARS:
+        print(
+            f"caura-interviewer: --max-event-chars {n} exceeds the server limit "
+            f"of {_SERVER_MAX_EVENT_CHARS}; using {_SERVER_MAX_EVENT_CHARS}.",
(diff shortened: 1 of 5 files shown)


arkash20:
CI green (15/15, including the client-python job). @Eldad-Caura please approve

Eldad-Caura (approved):

PR #1420, opened 2026-09-09
fix(embedding): cache LocalEmbedding per model so the local provider stops reloading the model on every request

ran-taig:
## `EMBEDDING_PROVIDER=local` reloaded the sentence-transformers model on every request

`get_embedding_provider` runs on every embed/search request (`_service.py` resolves the provider per call), and the LOCAL branch constructed a fresh `LocalEmbedding()` each time. The loaded `SentenceTransformer` is cached on the **instance** (`self._model`), so every request paid a full model load — seconds of latency per embed, plus one copy of the weights in RAM per in-flight request until GC. `LocalEmbedding.dedup_scope` already documented "the registry's per-model instance cache"; the cache didn't exist.

Dormant, not a live outage — same caveat as C38 (#1342), which revived this path but kept per-call construction: staging/prod embed via TEI/bge-m3 through the OpenAI-compatible client.

## Fix

Module-level `_local_provider_cache: dict[str, LocalEmbedding]` keyed by model name, mirroring `_openai_provider_cache` minus what doesn't apply:

- **no eviction** — nothing to close (no connection pool, no credential to rotate), and the key space is env-driven (`LOCAL_EMBEDDING_MODEL`), so a process realistically holds one entry;
- **no lock** — no await between lookup and insert, so coroutines can't interleave a double-construct; construction is cheap regardless (the model load is lazy and single-flighted behind the instance's `_load_lock`).

## Tests

`tests/test_local_embedder_instance_cache.py`: same model → same instance (default and env-override), distinct models → distinct instances. Existing C38 suite unaffected (9 passed together). `ruff check common/` and `mypy --config-file common/mypy.ini common/` clean.

### DIFF
--- a/common/embedding/_registry.py
+++ b/common/embedding/_registry.py
@@ -33,6 +33,24 @@ logger = logging.getLogger(__name__)
 # could never have worked against this schema.
 DEFAULT_LOCAL_EMBEDDING_MODEL = "BAAI/bge-large-en-v1.5"
 
+# Cache ``LocalEmbedding`` instances by model name. The loaded
+# ``SentenceTransformer`` lives on the *instance* (``self._model``), and
+# ``get_embedding_provider`` runs on every embed/search request — so a
+# fresh instance per call meant every request paid a full model load
+# from disk (seconds, plus its own copy of the weights in RAM until GC)
+# instead of hitting the already-warm model. This is the "registry's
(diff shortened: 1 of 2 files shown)


ran-taig:
@claude

ran-taig:
@erni-a

erni-a (approved):

PR #1421, opened 2026-09-09
docs(settings): pin and document the reset shape and propagation contract (D17)

arkash20:
D17 was filed as two behaviours that "silently corrupt any flip-then-measure workflow". Investigating both changed the task, so this PR is smaller than the row — and the row needs correcting.

## Cross-worker invalidation already ships

`update_settings` publishes `Org.SETTINGS_CHANGED` after invalidating locally, `register_consumers` subscribes `handle_org_settings_changed` with `broadcast=True`, and `app` calls that at startup. All of it cites CAURA-571 by name.

What's true is narrower: `EVENT_BUS_BACKEND` defaults to `inprocess`, where a broadcast can't cross processes — so on that backend, **local dev included**, siblings still wait out the 5-minute TTL.

**This also corrects something I recorded on C27 (#1410).** I saw a flag take ~30s and three polls to take effect locally and wrote it up as production behaviour. It was the in-process bus, not a missing mechanism.

## Resetting already works — via `null`

`_validate_leaf_types` passes `None` through deliberately (`v is not None`), every resolver property reads `None` as "no override", and a section set to `None` drops the group back to defaults. Measured, not assumed:

| payload | result |
|---|---|
| leaf → `null` | `recall_boost False` → resolver reads the default (`True`) |
| section → `null` | `default_profile {…}` → `None` |
| section → `{}` | `default_profile {…}` → **unchanged** |

So the reported defect is real but mis-stated. The reset exists; what fails is that `{}` — the shape an operator reaches for first — merges nothing and silently no-ops, so the flip looks like it never took.

## Why documenting, not building

`{}` is deliberately left as a no-op. It's a legitimate "nothing to change in this section", and reinterpreting it as a clear would turn a harmless payload into one that wipes settings — a worse failure than the one being fixed.

Both the route and the service now state the merge semantics, the `null` reset shape, the `{}` trap, and that propagation is expected rather than immediate.

## Testing

11 tests pin all of it — including that validation still admits `null`, because if that guard ever tightens the only reset shape goes with it, and that a subscriber actually exists (publishing to nobody looks identical to not publishing).

Tracker row `reg-d17` should be narrowed on merge: the invalidation half is already shipped.

### DIFF
--- a/core-api/src/core_api/routes/settings.py
+++ b/core-api/src/core_api/routes/settings.py
@@ -42,6 +42,25 @@ async def update_tenant_settings(
 ):
     """Update tenant settings. Accepts partial updates. API keys are encrypted at rest.
 
+    **Resetting a setting.** Updates are a deep MERGE, so omitting a key leaves
+    it as it was — omission means "don't touch", never "clear". To return a
+    setting to its default, send it explicitly as ``null``:
+
+        {"search": {"recall_boost": null}}          # one leaf back to default
+        {"search": {"default_profile": null}}       # a whole section back to default
+
+    Sending ``{}`` for a section is a NO-OP, not a reset — an empty dict merges
+    nothing. That reads like a clear and is the shape operators reach for first,
+    so it is called out here rather than left to be discovered by a flip that
+    silently did not take.
+
+    **Propagation.** A write invalidates the handling worker's cache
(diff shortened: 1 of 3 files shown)


arkash20:
CI green. @Eldad-Caura please approve

Eldad-Caura (approved):

PR #1422, opened 2026-09-09
fix(search): ENTITY_LOOKUP silently drops temporal hints — decline the short-circuit when one is present

ran-taig:
## Bug

`ExtractTemporalHint` sets three signals — `temporal_window` (soft freshness), `date_range_filter` (hard WHERE filter), `history_hint` — and **all three are consumed only by `ExecuteScoredSearch`**. The ENTITY_LOOKUP short-circuit in `ClassifyQuery` sets `skip_scored_search=True` (and `PostFilterResults` skips via its guard), so on a dated query that matched an entity:

- *"what did Alice decide two weeks ago"* → linked memories from **any date**, ranked only by hop boost + token overlap
- the hard date-range constraint is violated **silently** — the `entity_lookup` log line records only entity count and top_k
- `_collect_memories` forwards every other visibility filter (`valid_at`, status, agents, `readable_tenant_ids`) precisely to keep parity with the scored path ("drift is a cross-tenant leak risk"); `date_range_filter` was the one filter that drifted

## Fix

A decline gate after the CAURA-698 over-broad check: when `temporal_window` or `date_range_filter` is set and entities matched, expand the graph once, stash `_classified_entity_hops`, log a distinct grep-able decline reason, and fall through to the temporal / scored-search cascade — which applies both temporal signals, with the entity match surviving as a hop boost.

Placement details, each pinned by a test:

- **Declines before the pool load** — no `load_memories_by_ids` roundtrip or per-tenant storage slot spent on rows that would be discarded (H-03 spirit).
- **Ordered after the over-broad decline** — an over-broad match on a dated query still sets `entity_match_declined` (suppressing hop-boost) and never pays for expanding a >threshold seed set.
- **`entity_match_declined` NOT set** — unlike over-broad, this match is good relevance signal; only the exclusive route is declined.

## ⚠️ Deliberate behavior flip

`test_entity_lookup_takes_priority_over_temporal` pinned the old precedence ("entity + temporal → ENTITY_LOOKUP wins"). This PR inverts it — rewritten as `test_temporal_window_declines_entity_short_circuit` with the rationale in its docstring. If that pin encoded a product decision rather than an accident of ordering, flag it here.

**Out of scope on purpose:** RECENT_CONTEXT recency-regex queries ("what did we …") still lose to ENTITY_LOOKUP. That regex matches very common entity-query phrasing, so extending the decline there reroutes far more traffic and deserves its own decision.

## Testing

- 3 tests: the inverted precedence pin (asserts TEMPORAL plan, no pool load, hops preserved, no `entity_match_declined`), hard-date-range-only decline (falls to keyword/semantic where the scored search applies the range), over-broad-first ordering (no expansion, boost suppressed).
- Full unit suite: **5,951 passed** (`-m "not benchmark and not integration"`). The 4 failures in `test_recall_returns_memory_with_pending_embedding` / `test_fts_only_row_survives_a_saturated_candidate_window` fail identically on pristine `origin/main` — they need CI's migrated Postgres (`search_vector` generated column), not related to this change.
- `ruff` and `mypy` clean on touched files.

### DIFF
--- a/core-api/src/core_api/pipeline/steps/search/classify_query.py
+++ b/core-api/src/core_api/pipeline/steps/search/classify_query.py
@@ -6,8 +6,12 @@ matches are found the step MAY short-circuit to an *entity_lookup* strategy
 search can be skipped — but only when the linked-memory pool can fill the
 caller's ``top_k``.  That route replaces scoring rather than re-ranking it, so
 an under-filled pool falls through instead and the entity hits are applied as a
-hop boost over real scores (H-03).  Otherwise the query is routed to keyword or
-semantic search based on the adaptive FTS weight.
+hop boost over real scores (H-03).  A query carrying a temporal hint
+(``temporal_window`` / ``date_range_filter``) declines the short-circuit the
+same way: the hard date filter and freshness handling live in the scored
+search this route skips, so honouring the hint requires falling through.
+Otherwise the query is routed to keyword or semantic search based on the
+adaptive FTS weight.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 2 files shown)


ran-taig:
@claude

ran-taig:
@erni-a

erni-a (approved):

PR #1423, opened 2026-09-09
fix(embedding): unify the EMBEDDING_PROVIDER default and stop silent fake-vector persistence

ran-taig:
## The bug (validated, then prod-verified)

The embedding-provider default lived in two places and disagreed: core-api's pydantic `Settings.embedding_provider` said **`"openai"`** while `common.embedding._resolve_provider_name`'s env fallback said **`"fake"`**. Any caller without a `tenant_config`, in a process whose env lacked `EMBEDDING_PROVIDER`, silently embedded and persisted FAKE hash vectors while memory paths in the same process embedded for real.

Affected tenant-config-less paths: doc/skill writes + doc search (REST and MCP `caura_doc`), entity-name embeddings, and the storage backfill CLI. The fake provider always succeeds, so every `embedding is None` guard passed — no error, no log. Live repro: bare-metal `.env`-file runs (pydantic's `env_file` loads into Settings, never `os.environ`; there is no `load_dotenv` anywhere) and the backfill CLI. Prod SaaS was verified safe on 2026-09-09 only because `EMBEDDING_PROVIDER=openai` is set out-of-band on the service (companion pin: caura-enterprise PR on `promote-to-prod.yml`).

## The fix

- **One default.** `common.provider_names.DEFAULT_EMBEDDING_PROVIDER = "openai"`, read by both `Settings.embedding_provider` and `_resolve_provider_name`'s env fallback. Keyless dev still degrades to the fake provider, but through the registry path that WARNS — never silently.
- **Tenant config at every embed site.** Doc/skill write + search (REST and MCP) resolve the tenant's config (cache-first, 5-min TTL; degrade-to-`None` on failure, same idiom as memory_service), so per-tenant provider/key/model overrides now apply on those paths too. The entity-extraction worker passes the `tenant_cfg` it already resolves for extraction (A5c).
- **Backfill CLI preflight.** A live (non `--dry-run`) run now refuses to start when the process-level provider *resolves* to fake — explicit `EMBEDDING_PROVIDER=fake`, or `openai` degrading with no key anywhere — unless `--allow-fake-provider`. A fake "repair" flips rows out of the NULL selector forever, hence hard refusal. This replaces the old openai-without-`OPENAI_API_KEY` check, which had a false negative (unset provider fell back to `"fake"` and sailed through) and a false positive (platform-tier `PLATFORM_EMBEDDING_*` credentials satisfy openai without `OPENAI_API_KEY`). `local` is exempt from preflight construction so we never eagerly load the sentence-transformers model.

## Tests

- New `TestDefaultProviderNameUnification` (tests/test_platform_providers.py): env-unset → shared default, env wins, tenant wins, and `Settings.model_fields` default == the constant.
- New preflight tests (tests/test_backfill_embeddings.py): fake refusal fires before `run_backfill`, openai-degrading-to-fake refusal, `--allow-fake-provider` override, `--dry-run` exemption. Existing `_amain` tests updated to set a real provider (they previously leaned on the buggy fallback).
- Full root suite: 6,400+ passed. core-storage-api suite: 371 passed. (31 root failures are pre-existing on clean `origin/main` — local-DB schema staleness, verified by stash-and-rerun.)

## Behavior-change note

A bare-metal run with `OPENAI_API_KEY` exported but `EMBEDDING_PROVIDER` unset previously embedded fake; it now embeds real (matching what the Settings default always claimed). Docker/compose, onprem, staging, and prod all set `EMBEDDING_PROVIDER` explicitly and are unchanged.

### DIFF
--- a/common/embedding/_service.py
+++ b/common/embedding/_service.py
@@ -24,6 +24,7 @@ from common.embedding.constants import (
     EMBEDDING_RETRY_DELAY_S,
 )
 from common.embedding.protocols import EmbeddingProvider, InstructionAwareEmbedder
+from common.provider_names import DEFAULT_EMBEDDING_PROVIDER
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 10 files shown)


ran-taig:
@claude

ran-taig:
@claude

ran-taig:
@erni-a

erni-a (approved):

PR #1424, opened 2026-09-09
fix(worker): stop discarding atomic facts on the async enrichment path (A70 step 1)

arkash20:
## A70 is mostly already built

I went looking for where to add `facts[]` and found it there:

- The enrichment prompt has **field 9 `atomic_facts`**, with exactly the intended semantics — *"each entry becomes its own child memory with its own embedding"*.
- `AtomicFact` and `EnrichmentResult.atomic_facts` exist, with a validator and **17 passing tests**.
- `memory_service.py:3103` **already fans them out** into child memories on the **synchronous** path — live-hash *and* intra-loop dedup, visibility and weight inheritance, non-fatal failure.

**Estimate on the tracker row drops from ~4d to ~1d.**

## What the async worker was doing

Not merely skipping the fan-out — **discarding the facts**:

```python
_ENRICHMENT_UNROUTED_FIELDS: frozenset[str] = frozenset({"atomic_facts"})
```

So the LLM's answer was gone, and the only way to recover it was to pay for the call again. A fast-mode write of multi-claim content produced fewer memories than the same content in strong mode — permanently, and silently.

## This PR

Routes the facts into the row's metadata. **Persisted, not yet consumed.** Two things change:

- the fast/strong gap becomes **recoverable from stored data** instead of requiring an LLM re-run;
- the exposure becomes **countable from real traffic**, which is what the A75 proof gate needs to size this.

Routed as **always-write**, for the same reason `tags` is: a re-enrichment finding one claim where a previous run found three must be able to clear the stale two, or a later fan-out creates children for claims the content no longer makes. The existing `llm_ms > 0` guard stops a heuristic-fallback redelivery wiping real facts.

`_ENRICHMENT_UNROUTED_FIELDS` is now **empty**, so the module's import-time drift guard covers every `EnrichmentResult` field rather than every field but one.

## Why fan-out stays in core-api

Core-api already consumes `Topics.Memory.ENRICHED` (`handle_memory_enriched`) and already owns the dedup / visibility / weight rules a child must inherit. Duplicating that here would be a second implementation of a subtle path — the mistake A54 and C27 both paid for. The worker's storage client also has **no create-memory call at all**, so doing it here means granting the worker a write capability it has never had.

**Step 2** — the core-api fan-out from the persisted facts — is the remainder of A70.

## Testing

8 new tests, including that the drop list is now empty and that every enrichment field still has a routing home (mirroring the module's own import-time guard).

core-api suite: **27 failures vs main's 28** — none new. Main's extra is the flaky wall-clock test now filed as `oss-0909-l-01`. `ruff` clean; `mypy` on core-worker unchanged.

### DIFF
--- a/core-worker/src/core_worker/consumer.py
+++ b/core-worker/src/core_worker/consumer.py
@@ -284,6 +284,10 @@ _ENRICHMENT_METADATA_FIELDS: frozenset[str] = frozenset(
         "pii_types",
         "business_relevance",
         "retrieval_hint",
+        # Persisted, not consumed here: the row carries the claims the LLM
+        # found so core-api can fan them out. Storing them is what makes the
+        # gap recoverable — a dropped list means re-running the LLM.
+        "atomic_facts",
         "llm_ms",
     }
 )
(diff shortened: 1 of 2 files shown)


arkash20:
CI green. @Eldad-Caura please approve

Eldad-Caura (approved):

PR #1427, opened 2026-09-09
fix(search): honor search.graph_retrieval on the ENTITY_LOOKUP short-circuit

ran-taig:
## The bug (validated on main)

`search.graph_retrieval` (→ `ResolvedConfig.graph_expand`, default True) gates graph expansion on the hop-boost path (`_entity_boost_via_storage`: `if graph_expand and graph_max_hops > 0`), but `ClassifyQuery` never read it — the flag sits unread in the same `ctx.data` dict the step already consumes `entity_retrieval` from. Both `_expand_per_fleet` call sites ran the `expand_graph` traversal unconditionally (default `GRAPH_MAX_HOPS = 2`):

1. **Main ENTITY_LOOKUP short-circuit** — a `graph_retrieval=False` tenant still paid the traversal, and since the short-circuit *replaces* scored search, hop-1/2 neighbourhood memories became the entire answer. Expansion also inflated the pool toward the `len(filtered_rows) >= top_k` fill gate, so disabling the setting made the short-circuit fire *more*, not less.
2. **Temporal-decline path (#1422)** — the expanded hops stashed in `_classified_entity_hops` feed the boost step's `precomputed_hops` input, which bypasses that step's own `graph_expand` gate by design (it exists to avoid re-deriving FTS + expansion). Expanded hops were smuggled past a disabled flag.

This violates the code's own documented contract: both the `entity_retrieval` settings-registry comment and the classify-step comment state that `graph_retrieval` "bounds expansion once an entity has already matched" — the short-circuit is exactly that situation and bounded nothing.

## The fix

A `_hops_for_seeds` helper owns the gate **where the hops are produced**: `graph_expand and graph_max_hops > 0` → `_expand_per_fleet` as before; otherwise the matched entities come back as hop-0 seeds with neutral weight — the exact `dict.fromkeys(ids, (0, 1.0))` else-branch the boost path already uses. Entity lookup itself stays on (that's `search.entity_retrieval`'s job per the settings split); only the graph neighbourhood is withheld. Gating at production also keeps the temporal stash hop-0, making the `precomputed_hops` bypass moot without touching the boost step.

## Tests

New `tests/test_graph_expand_short_circuit.py` pins all four quadrants:
- flag off → ENTITY_LOOKUP fires from hop-0 seeds, zero `expand_graph` calls, only the matched entity feeds the pool
- flag off + temporal hint → decline as before, stash is exactly `{eid: (0, 1.0)}`
- flag absent (internal callers) and flag on → expansion unchanged, stash carries the hop-1 neighbour

Verified the two flag-off tests **fail on unfixed main**. Full root suite: 6,400+ passed, zero new failures vs the pre-existing local-DB baseline; all short-circuit/entity-flag/boost-decline/pipeline suites green; ruff check + format clean.

### DIFF
--- a/core-api/src/core_api/pipeline/steps/search/classify_query.py
+++ b/core-api/src/core_api/pipeline/steps/search/classify_query.py
@@ -87,9 +87,20 @@ class ClassifyQuery:
         # below. ``ParallelEmbedAndEntityBoost`` reads the same flag and skips
         # hop-boosting, making this the single switch for query-time entity and
         # graph retrieval. Deliberately independent of ``graph_expand``
-        # (``search.graph_retrieval``), which only bounds expansion depth once
-        # an entity has already matched.
+        # (``search.graph_retrieval``), which only bounds expansion once an
+        # entity has already matched.
         entity_retrieval: bool = ctx.data.get("entity_retrieval", True)
+
+        # ``search.graph_retrieval`` org setting (default True). Entity lookup
+        # stays on — the settings contract is that only ``entity_retrieval``
+        # can switch that off — but with this flag off the matched entities
+        # must not be graph-expanded: no ``expand_graph`` roundtrip, hop-0
+        # seeds only, mirroring ``_entity_boost_via_storage``'s gate. Both
+        # ``_expand_per_fleet`` call sites below honour it via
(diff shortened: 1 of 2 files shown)


ran-taig:
@claude

ran-taig:
@erni-a

erni-a (approved):

PR #1428, opened 2026-09-09
refactor(memory): lift the atomic-fact fan-out out of the background task (A70 step 2a)

arkash20:
**Pure refactor, no behaviour change.** `fan_out_atomic_facts` is the block that lived inline in `_enrich_memory_background`, moved verbatim.

## Why move it

Being inline made the **synchronous path its only possible caller**. That's exactly why a fast-mode write of multi-claim content produces fewer memories than the same content in strong mode: the async worker can't run this without a second implementation — and a second implementation of *this particular block* is the last thing anyone should write.

Every guarantee in it was paid for by an incident:

| | |
|---|---|
| `#808` | derived rows inherit the parent's governance verdict — and `parent_visibility` is passed **in**, never re-read, because a row read before the governance PATCH still carries the visibility `keep_private` just removed |
| `CAURA-222` | children embed raw fact content, not hint-prefixed |
| `OSS #814` | dedup against live rows **and** within the loop, batched once |
| 2026-07-27 postmortem | a failed embed persists the fact unembedded and queues a repair rather than dropping it |
| `CAURA-602` | deliberately outside the per-tenant storage bulkhead |

A test asserts each of those rationales survived the move, because a refactor that drops the reasons is how they get re-litigated and reverted.

## Contract

Returns `{"created", "deduped", "unembedded"}` so the async caller can log the same three lines the sync path does, and short-circuits on an empty list **without touching storage** — the async caller will invoke it on every enriched row, and most have no atomic facts.

## Why this is split from the wire-up

If a regression appears, a behaviour-neutral move tells you immediately whether the fault is in the extraction or in the new caller. **Step 2b** calls it from `handle_memory_enriched`, after governance remediation so a dropped row spawns nothing and children inherit post-policy visibility.

## Testing

- Existing fan-out tests are the neutrality proof — **266 pass** across the atomic/enrich selection.
- 11 new tests pin the seam: the signature the async caller needs, that the sync path goes through it (and the inline copy is gone), that visibility is a parameter rather than a re-read, and the incident provenance.
- Full suite: **27 failures, identical to main's 27** — none new. `ruff` clean; `mypy` unchanged.

### DIFF
--- a/core-api/src/core_api/services/memory_service.py
+++ b/core-api/src/core_api/services/memory_service.py
@@ -2850,6 +2850,279 @@ def _resolve_write_mode(data: MemoryCreate, tenant_config) -> str:
     return tenant_config.default_write_mode
 
 
+async def fan_out_atomic_facts(
+    sc,
+    *,
+    atomic_facts: list,
+    memory_id,
+    tenant_id: str,
+    fleet_id: str | None,
+    agent_id: str,
+    parent_metadata: dict,
+    parent_visibility: str,
+    parent_weight: float,
+    parent_ts_start,
+    tenant_config,
+) -> dict[str, int]:
+    """Create one child memory per extracted atomic fact.
+
+    Lifted verbatim out of ``_enrich_memory_background`` so the ASYNC path can
+    reuse it (A70). The synchronous path was the only caller, which is why a
+    fast-mode write of multi-claim content produced fewer memories than the same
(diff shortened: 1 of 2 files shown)


arkash20:
CI green. Pure refactor — the existing fan-out tests are the neutrality proof. @Eldad-Caura please approve

Eldad-Caura (approved):

PR #1429, opened 2026-09-09
perf(search): compute scored-search primitives once via a fenced ingredients CTE (HNSW plan PR1)

ran-taig:
## What

`memory_scored_search` paid the pgvector cosine distance **six times per candidate row** and `ts_rank_cd` ten times: SQLAlchemy inlines an expression at every site that names it, and `score`/`similarity`/`vec_sim` plus the CAURA-722 diagnostic columns each named the same CASE. This lands the inner-projection work `_saturate_rank`'s measurement note promised: heavy primitives (cosine distance, `ts_rank_cd`, the FTS guard, `has_embedding`) are projected **once** into a `WITH ingredients AS MATERIALIZED (...)` CTE; every derived factor (similarity, freshness, boosts, penalties, both score formulas) is CASE/arithmetic over those columns in the branch layer above.

## Measured

| Render counts (compiled statement) | before | after |
|---|---|---|
| cosine `<=>` (1024-dim, the dominant per-row cost) | 6 | **1** |
| `ts_rank_cd` | 10 | **1** |
| `plainto_tsquery` | 21 | **4** (two WHERE uses — predicates can't fold into a targetlist column — + `fts_score` + `fts_match`) |

On a 50k-row / 1024-dim rig, the candidate select dropped **427 ms → 90 ms** from this change alone. (Context: search p50 ~1,300 ms on a 31k corpus per the in-code measurement; the 188 s eToro scored-search datapoint scales with exactly this per-row cost.)

## Why `AS MATERIALIZED`

With the FTS-reserved branch present the CTE is referenced twice and PostgreSQL materialises it regardless. But a blank-query call (entity-only / vector-only search) leaves one branch — and a plain single-reference CTE is **inlined** back into its consumer, re-expanding every column reference into a fresh distance computation. `AS MATERIALIZED` is the documented PostgreSQL 12+ modifier that forbids exactly that folding (review moved this off an earlier `.offset(0)` draft that leaned on the incidental limitOffset-blocks-pull-up planner rule). SQLAlchemy 2.0 has no `materialized=` argument on `cte()`; `CTE.prefix_with("MATERIALIZED")` is the documented way to emit it. PG < 12 would reject the syntax, but the schema already floors on 12+ (pgvector, HNSW).

**Guarded at two levels:**
- Compiled text (`core-storage-api/tests/test_fts_score_single_render.py`): modifier present on the single-branch statement, cosine == 1, updated `ts_rank_cd`/`plainto_tsquery` ratchets with history.
- **The plan itself** (`tests/test_scored_search_materialized_plan.py`, new): `EXPLAIN (FORMAT JSON, VERBOSE)` of the statement the service actually issues must contain a `CTE Scan` over `ingredients` and **exactly one** distance operator in the whole plan — so a future PostgreSQL/SQLAlchemy change that stops the modifier short of the planner fails CI instead of surfacing as a silent ~6× hot-path latency regression. Runs against the real CI database; covers both the single-branch and two-branch statements.

## Semantics unchanged

- Row admission (CAURA-594 `embedding IS NOT NULL OR fts match`), all filters (fleet/visibility/agent/type/status carve-out/valid_at), both scoring formulas (legacy multiplicative + A50 additive), the CAURA-722 column contract, and union dedup are preserved — the select list of the scored CTE is byte-compatible.
- The reserved branch's filter now reads the `fts_match`/`has_embedding` **columns** over the materialised CTE instead of a second GIN probe — one scan feeds both union branches.
- `_exact_lexical_match` and `_fts_guard` were always the same expression; the status-penalty gate now reads the single projected column.

## Testing

- Ratchet suite: 11/11. New EXPLAIN plan-shape tests: 2/2 against a freshly migrated Postgres 16 + pgvector.
- Full suites against that fresh DB: **6,473 root + 373 core-storage-api tests pass**, including `test_recall_returns_memory_with_pending_embedding` and `test_fts_only_row_survives_a_saturated_candidate_window` (deferred-embedding + FTS-reservation end-to-end) and all A50/A49/CAURA-722 scoring tests.
- ruff + ruff format + mypy clean.

## Context

PR1 of the HNSW two-stage retrieval plan (`docs/plans/hnsw-two-stage-retrieval.md`, local): the primary search path can never use `ix_memories_embedding_hnsw` (ORDER BY is a blended score, not a distance), so it full-scans the tenant slice — this PR makes each scanned row ~6× cheaper with zero semantic change. PR2 (opt-in ANN candidate pool) builds on this layering.

### DIFF
--- a/core-storage-api/src/core_storage_api/services/postgres_service.py
+++ b/core-storage-api/src/core_storage_api/services/postgres_service.py
@@ -2381,6 +2381,19 @@ class PostgresService:
         _score_formula = int(sp.get("score_formula", 0) or 0)
 
         # -- Scoring expressions --
+        #
+        # Two-layer build. The per-row primitives that are expensive to
+        # evaluate — the pgvector cosine distance and ``ts_rank_cd`` — are
+        # projected exactly ONCE into a fenced ``ingredients`` CTE, and every
+        # derived factor (similarity, freshness, boosts, penalties, score) is
+        # computed in the branch layer above it from those columns. This is the
+        # inner-projection work ``_saturate_rank``'s note promised: before it,
+        # SQLAlchemy inlined the ``vec_sim`` CASE at every site that named it
+        # and the compiled statement paid SIX cosine distance computations per
+        # candidate row (427ms -> 90ms for this change alone on a 50k-row,
+        # 1024-dim rig; the ratchet in test_fts_score_single_render pins the
+        # counts in both directions).
+        #
         # CAURA-594: pgvector's `<=>` is strict — NULL in → NULL out. A
         # bare `1 - cosine_distance` would therefore propagate NULL up
         # through the similarity blend into `score`, and PostgreSQL's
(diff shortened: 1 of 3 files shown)


ran-taig:
@claude

ran-taig:
@claude

ran-taig:
@claude please re-review — amended per the fence review finding: the `.offset(0)` fence is replaced with the documented `WITH ingredients AS MATERIALIZED (...)` (via `CTE.prefix_with("MATERIALIZED")`, supported on the pinned SQLAlchemy 2.0.52; native `materialized=` doesn't exist in 2.0.x), and a new EXPLAIN-based test (`tests/test_scored_search_materialized_plan.py`) pins the actual plan — a `CTE Scan` over `ingredients` must exist and the distance operator must appear exactly once in the whole plan, on both the single-branch and two-branch statements — so a future Postgres/SQLAlchemy behavior change fails CI rather than shipping a silent ~6x regression. Also rebased onto current main (restores the branch update that my first force-push accidentally overwrote).

ran-taig:
@erni-a

erni-a (approved):

PR #1430, opened 2026-09-09
feat(consumer): fan out atomic facts on the deferred path (A70 step 2b)

arkash20:
Closes the sync/async gap that A70 is about: **a fast-mode write of multi-claim content produced fewer memories than the same content written in strong mode.** The synchronous path fans the enricher's `atomic_facts` out into child memories; the deferred path had no way to.

- **Step 1** (#1424) stopped the worker discarding the facts.
- **Step 2a** (#1428) lifted the fan-out into `fan_out_atomic_facts` so something other than the write path could call it.
- **This PR** joins them — `handle_memory_enriched` reads the persisted facts and creates the children through the *same* function.

Not a second implementation. Every guarantee in that block traces to an incident (`#808`, `CAURA-222`, `OSS #814`, `CAURA-602`, the 2026-07-27 stranded embeddings), and the worker still has no create-memory call of its own.

## Placement is the design

```
remediation → [fan-out] → embedding check → contradiction detection
```

**After remediation.** A dropped row spawns nothing (that branch returns first), and children take `outcome.visibility` rather than the row's. `memory` was read *before* the governance PATCH, so its visibility is the pre-policy value — a child inheriting it would re-publish exactly what `keep_private` just made private (`#808`). This is the caller `RemediationOutcome.visibility` was added for; its docstring says so.

**Before the embedding check.** A parent whose vector hasn't landed still has valid facts, and each child embeds itself. Gating on the parent's embedding would strand them behind an unrelated race.

## Failure handling is asymmetric on purpose

| case | marker | why |
|---|---|---|
| fan-out fails | **kept** | the facts cost an LLM call and are still stored; a redelivery or re-enrichment retries, and the live-hash dedup stops a retry duplicating whatever landed |
| payload unparseable | **cleared** | we wrote it, so an unparseable shape means the schema moved under an older row — no redelivery will re-parse it |

The marker is cleared with `metadata_patch` (JSONB `||`) so the summary / tags / pii keys the worker set on the same row survive. That clear is the *cheap* idempotency path, not the guarantee — the guarantee is the dedup, which already absorbs a redelivery arriving before the write lands.

Nothing raises into the handler: contradiction detection runs after this and is the handler's primary job.

## Testing

11 new tests, most pinning the ordering (`drop < fan-out < embedding check`), that `outcome.visibility` wins, that a failed fan-out returns before clearing, and that the consumer calls the shared function rather than writing rows itself.

One anchor note for reviewers: `embedding = memory.get("embedding")` appears in **both** enriched and embedded handlers, so this is anchored on the governance drop-log unique to the enriched one.

### DIFF
--- a/core-api/src/core_api/consumer.py
+++ b/core-api/src/core_api/consumer.py
@@ -30,6 +30,7 @@ import logging
 
 from pydantic import ValidationError
 
+from common.enrichment import AtomicFact
 from common.events.base import Event
 from common.events.factory import get_event_bus
 from common.events.memory_embedded import MemoryEmbedded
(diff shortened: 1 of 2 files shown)


arkash20:
CI green. Completes A70 (steps 1, 2a, 2b). @Eldad-Caura please approve

Eldad-Caura (approved):

PR #1431, opened 2026-09-09
chore(scripts): make MCP E2E targets configurable

### LINKED ISSUE #1408: chore(scripts): mcp_tool_smoke.py and trust_matrix_e2e.py hardcode localhost:8000 and /tmp/e2e.env
## What

`scripts/smoke_test.py` — the script a new contributor is most likely to run first — takes its target as an argument:

```python
DEFAULT_URL = "http://localhost:8000"
# …
parser.add_argument("--url", default=DEFAULT_URL, help=f"Base URL (default: {DEFAULT_URL})")
```

Its two siblings in the same directory, doing the same kind of job, hardcode everything:

`scripts/mcp_tool_smoke.py`
```python
27:    path = Path("/tmp/e2e.env")
44: CORE_API = "http://localhost:8000"  # direct core-api (MCP lives here)
```

`scripts/trust_matrix_e2e.py`
```python
31:    with open("/tmp/e2e.env") as f:
44: CORE = "http://localhost:8000"  # direct core-api; X-API-Key resolves the tenant
```

Neither has a `--url`, and neither can point at a different env file. Both are module-level constants, so pointing either at a staging stack or a non-default port means editing the file.

## Why it matters

Two things, and the second is the one that actually bites:

- **No target override.** Anyone running the stack on a non-default port, or against a remote deployment, edits the script and then has to remember not to commit it.
- **`/tmp/e2e.env` is not portable.** It's an absolute POSIX path. On Windows there is no `/tmp`, so both scripts fail at the read with no useful message before any of their logic runs. `smoke_test.py` has no such problem.

## What to change

Copy the pattern that already exists in `scripts/smoke_test.py` — don't invent a new one:

- `--url`, defaulting to the current `http://localhost:8000`, with the default shown in the help string.
- `--env-file`, defaulting to the current `/tmp/e2e.env`, so the credential file can live anywhere.

Turn the module-level `CORE_API` / `CORE` constants into values threaded from parsed args. In `mcp_tool_smoke.py` that means the f-strings at lines 87/127/134/146/159 need the value passed in rather than read from module scope — the smallest change is to keep them as functions taking the base URL, not to introduce a class.

Keep both defaults exactly as they are. This is about making them overridable, not changing what happens when you run the script bare.

Worth mentioning in the PR: whether `Path(tempfile.gettempdir()) / "e2e.env"` is a better default than the literal `/tmp/e2e.env`. It's more portable, but it changes behaviour for whatever writes that file (the E2E register step), so it needs a maintainer's call — flag it, don't just do it.

## How to verify

Both scripts need a live stack, so full runs aren't expected from a first-time contributor. What you can check without one:

```bash
python scripts/mcp_tool_smoke.py --help
python scripts/trust_matrix_e2e.py --help
```

Both should print the two new flags with their defaults, and neither should touch `/tmp` before argument parsing — verify by running `--help` on a machine with no `/tmp/e2e.env` and confirming it prints help rather than the "missing" error.

## Scope

Two scripts, argparse only. No change to what either script asserts.

---

_Filed unlabeled — the account that opened this lacks label-write permission. Intended labels: `good-first-issue, help-wanted, kind/chore, status/needs-triage`._

sen-to:
## Summary

Make the MCP smoke and trust-matrix scripts configurable without changing their default target or credential-file path.

## Related Issue

Closes #1408

## Type of Change

- [x] Other: developer tooling

## How Has This Been Tested?

- `python scripts/mcp_tool_smoke.py --help`
- `python scripts/trust_matrix_e2e.py --help`
- `PYTHONPYCACHEPREFIX=/private/tmp/caura-pycache python3 -m py_compile scripts/mcp_tool_smoke.py scripts/trust_matrix_e2e.py`
- `git diff --check`

The scripts require a live stack for their end-to-end flows. `ruff` is not installed in this local environment.

## Checklist

- [ ] `mypy` passes
- [ ] `pytest` passes locally

## Additional Notes

Both flags retain the existing `http://localhost:8000` and `/tmp/e2e.env` defaults.

### DIFF
--- a/scripts/mcp_tool_smoke.py
+++ b/scripts/mcp_tool_smoke.py
@@ -15,6 +15,7 @@ Expects /tmp/e2e.env (written by the E2E register step) containing::
 
 from __future__ import annotations
 
+import argparse
 import json
 import sys
 import time
@@ -22,11 +23,13 @@ import urllib.error
 import urllib.request
 from pathlib import Path
 
+DEFAULT_ENV_FILE = Path("/tmp/e2e.env")
+DEFAULT_CORE_API = "http://localhost:8000"
 
-def load_env() -> dict[str, str]:
-    path = Path("/tmp/e2e.env")
+
+def load_env(path: Path) -> dict[str, str]:
     if not path.exists():
-        print("ERROR: /tmp/e2e.env missing — register a tenant + API key first.")
+        print(f"ERROR: {path} missing — register a tenant + API key first.")
         sys.exit(2)
     return dict(
         line.strip().split("=", 1)
@@ -35,13 +38,9 @@ def load_env() -> dict[str, str]:
     )
 
 
-ENV = load_env()
-TENANT = ENV["TENANT_ID"]
-KEY = ENV["KEY"]
 FLEET = "smoke-fleet"
 
 GATEWAY = "http://localhost"  # nginx
-CORE_API = "http://localhost:8000"  # direct core-api (MCP lives here)
 
 
 def http(method: str, url: str, body=None, headers=None):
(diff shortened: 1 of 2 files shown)


erni-a (changes requested):
Thanks @sen-to. The behaviour is right: defaults unchanged, both scripts still run with no flags, and I confirmed with `py_compile` and `ruff format --check`. One thing blocks it: CI has a "Run Ruff linter (scripts)" step that lints `scripts/` with the repo config, and it fails on the `global` statement in `mcp_tool_smoke.py`:

```
PLW0603 Using the global statement to update `CORE_API` is discouraged
PLW0603 Using the global statement to update `KEY` is discouraged
PLW0603 Using the global statement to update `TENANT` is discouraged
```

Smallest fix: do what you already did in `trust_matrix_e2e.py`. Move the argparse block and the three assignments out of `main()` into the `if __name__ == "__main__":` guard, before the `main()` call. Assignments there are module-level, so the functions see them and no `global` is needed. `main()` then just prints and runs the steps.

Push that and I will approve the CI run and merge the same day. Sorry it sat a week without a reply.

sen-to:
@erni-a  Addressed your feedback in commit c8ace425: moved the CLI parsing and configuration setup out of main() to remove the Ruff PLW0603 findings. Please take another look—thank you!

erni-a (dismissed):

erni-a (approved):
Thanks @sen-to, that is exactly the shape I had in mind and the scripts ruff gate passes with the repo config. Both scripts still run with no flags and the defaults are unchanged. Merging once CI finishes.

PR #1433, opened 2026-09-09
feat(dedup): merge a same-claim near-duplicate instead of appending beside it (A71)

arkash20:
P0 half two, and the second of the two Critical rows. `DetectNearDuplicate` already finds the nearest stored row on **every fast write** — one ANN round-trip, no LLM — and stashes `near_duplicate_of` as advice nothing acts on. So a restated fact accumulates a sibling and **both stay live**: recall sees two rows where one claim exists. That's the composite-crowding pathology from the 2026-09-08 replay, arriving one duplicate at a time.

With `dedup.merge_near_duplicates` on, a hit that is provably the same claim supersedes its predecessor instead of appending.

## One deliberate departure from the plan

The plan says the enrichment call also answers *"same claim or complementary?"*. **It can't.** Enrichment completes *before* the candidate is known — `ParallelEmbedEnrich` precedes this step — so that phrasing means a **second** model call, on the latency path fast mode exists to keep short, and it wouldn't clear the no-new-LLM-calls hold.

`EmitMemoryTriple` has already populated `(subject_entity_id, predicate, object_value)` with no LLM, and the candidate arrives from `_find_semantic_duplicate` carrying the same columns. The triple answers the same question from data already in hand — so the merge check stays **free**, exactly as the plan claims it should be.

## What counts as the same claim

All three, each narrowing toward provable:

| condition | why |
|---|---|
| both sides resolved a subject **and** predicate | an unresolved triple proves nothing; absence must never read as a match |
| predicate ∈ `SINGLE_VALUE_PREDICATES` | the set where a subject holds exactly **one** value — what makes a second value a *replacement*. `works_on` can be true of five projects at once; `status` cannot |
| the objects **differ** | an identical object is a re-statement that adds nothing, and superseding a row with its own content is churn |

A candidate that isn't `active`/`confirmed` is left alone: an outdated or conflicted row already has an owner, and stacking a second supersession is how a lineage forks.

## Decision and effect are split

The detector runs **before** `WriteMemoryRow`, so the new row has no id and can't be linked to anything — it records the intent, and `ScheduleBackgroundTasks` performs the merge once an id exists.

The two writes are ordered **link, then retire**. Done the other way, a failure between them leaves a row `outdated` with nothing in its place and the claim vanishes from recall; this order leaves both live and linked — which is what an in-flight contradiction chain already looks like. The link is CAS-guarded against NULL, so a contradiction verdict that claimed the row first wins and this becomes a no-op.

A merge failure never fails the write: the row is committed and the caller already has its 201.

## Flag

`dedup.merge_near_duplicates`, per-tenant, **default off** — like every switch that changes what a write *produces* rather than what it reports. Reversible through the same lineage a contradiction supersession uses.

## Testing

20 new tests, including that a multi-valued predicate is never merged (the heart of the safety argument), that an unresolved triple proves nothing across six shapes, that the link precedes the retire, and that a merge failure can't raise into the write.

### DIFF
--- a/core-api/src/core_api/pipeline/steps/write/detect_near_duplicate.py
+++ b/core-api/src/core_api/pipeline/steps/write/detect_near_duplicate.py
@@ -37,13 +37,63 @@ from __future__ import annotations
 import logging
 import time
 
-from common.constants import SEMANTIC_DEDUP_JUDGE_THRESHOLD
+from common.constants import SEMANTIC_DEDUP_JUDGE_THRESHOLD, SINGLE_VALUE_PREDICATES
 from core_api.pipeline.context import PipelineContext
 from core_api.pipeline.step import StepOutcome, StepResult
 from core_api.services.dedup_identifier_filter import _content_is_identifier_bearing
 from core_api.services.memory_service import _find_semantic_duplicate
 
 logger = logging.getLogger(__name__)
+
+
+def _same_claim(new, candidate: dict) -> bool:
+    """True when the write restates a claim the candidate already carries.
+
+    A71. The near-duplicate band is full of two different things: a RESTATEMENT
+    of a fact whose value moved ("the deploy window is 04:00 UTC" over "…03:00
+    UTC"), and a COMPLEMENTARY fact that merely reads similarly. Only the first
+    should retire its predecessor; merging the second would delete information.
+
+    The test is deterministic and costs nothing. ``EmitMemoryTriple`` has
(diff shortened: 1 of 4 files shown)


arkash20:
CI green. Second of the two Critical rows; ships dark behind dedup.merge_near_duplicates. @Eldad-Caura please approve

Eldad-Caura (approved):

PR #1434, opened 2026-09-09
CAURA-723 feat(search): name the agent filter when it is why the resu…

eyal-bl:
…lt is empty

A tenant-scoped caller that passes a wrong `filter_agent_id` gets `HTTP 200 · items [] · warnings null` — byte-identical to a correct id that simply has nothing relevant to say. Measured, one memory written by `agent-real`:

    filter_agent_id = "agent-real"  -> 200  items=1
    filter_agent_id = "agnet-real"  -> 200  items=0  warnings=None

The filter is a SQL `WHERE memories.agent_id = ?`, so a typo matches no rows and the query returns nothing exactly as an empty query would. Nothing in the response separates them, and it fails in the safe-looking direction — an empty list reads as "no data yet", the most ordinary thing a memory product can say, so nobody investigates. It cost a 589-query benchmark run against an empty store.

## Scope: this is a tenant-scoped-caller bug only

An agent-scoped credential carries a verified `X-Agent-ID`, and `enforce_self_agent` **403s** it for naming any agent but itself — measured, for a typo and for a peer id. So the silent path needs a credential with no agent identity, asserting the id in the body.

That is not a corner: it is the dashboard's hand-typed `filter_agent_id` text box (tenant-scoped by its own comment, "no X-Agent-ID"), the documented `caura_recall` parameter on the public for-agents page, every benchmark harness we own, and the multi-user backend shape where one key serves many end users. Gating on the asserted identity keeps the probe off the path where the condition is unreachable.

## Which table decides what, and why not the obvious one

The natural implementation — look the id up in `agents`, and if absent report "no such agent" and skip — is wrong in the direction that loses data. `DELETE /agents/{id}` says so itself: *"Delete an agent. Memories written by this agent are NOT deleted."* The rows stay live and searchable with no agent row, and rows predating agent tracking were never registered at all. So:

  * **`has_memories`** (the `memories` table) decides whether to skip. It is the only fact that can prove the search would return nothing.
  * **`agent_registered`** (the `agents` table) decides only the wording, where being wrong costs a slightly-off message.

Four states, three codes:

    memories  agents    ->  outcome
    none      absent        skip · filter_agent_unknown
    none      present       skip · filter_agent_empty
    some      absent        RUN THE SEARCH · filter_agent_deregistered
    some      present       run the search, no warning

Row three is the regression guard: the memories outlive the agent row, so they must still be returned.

## Ordering, and a hole the ordering first created

The probe is computed immediately after identity resolution and BEFORE the route's `get_or_create_agent`, which registers the asserted id — a typo included. Probing after it would find an `agents` row the read itself had just created, and every typo would report as registered.

The first draft also RETURNED there, which skipped `enforce_fleet_read_many` and the usage metering: a caller naming a fleet it has no rights to got a 200 with a warning instead of a 403, and the warning disclosed whether an agent id exists.
`test_c27_strict_fleet_scoping` and `test_h05_multi_fleet_read_gate` caught it. The short-circuit now sits after every gate; the saving it exists for — the embedding call and the scored search (plus the recall model call on `/recall`) — is still entirely ahead of that point.

## Cost

One round trip answering both halves, only on the asserted-identity path with a filter set. Two `LIMIT 1` index probes:
`ix_memories_tenant_agent` and `uq_agents_tenant_agent`. A search that returns results pays that and nothing else; a misconfigured one is strictly cheaper than today, trading an embedding API call and a vector scan for two index hits.

`read=True` deliberately: the search it explains reads the replica, so answering from the same replica keeps the probe's story consistent with the result rather than reporting rows the search could not see.

The probe never raises. A failure degrades to today's behaviour — an unexplained result — because a diagnostic hint must not be able to fail a working search.

## Ratchets moved, both deliberately

`test_c27_strict_fleet_scoping` pins two things, and the probe is a fifth fleet-scoped read:

  * it must go through `_fleet_scope_clause`, never a hand-rolled `fleet_id.in_(fleet_ids)` — "exactly how A54 leaked". It does.
  * the helper's call-site count moves 5 -> 6, with the reasoning recorded at the constant.

The probe calls the helper with `strict=False` on purpose. Non-strict is a SUPERSET (it also admits tenant-shared null-fleet rows and `scope_org`), so the probe stays at least as permissive as the search it explains and can never skip a search that would have returned rows. Threading the tenant's real `strict_fleet_scoping` would tighten it and buy exactly that risk.

## `/recall` too

It parses the same `SearchRequest` and had the identical silent empty. Leaving it out is how the two routes came to disagree about what these fields mean — the reason `_resolve_read_identity` is shared at all. Its response gains `warnings`, seeded with any scope warning and then extended by the pipeline, so the field means what `/search`'s does rather than carrying only this one code.

## Known limitation, pre-existing

The read path registers the asserted id via `get_or_create_agent`, so a repeated typo is registered by the first call and the second reports `filter_agent_empty` rather than `filter_agent_unknown`. Both still say the filter matched nothing, and the wording is advisory. Not introduced here and not fixed here: the clean answers are to stop registering on reads, or agent soft-delete (a `deleted_at` plus making `uq_agents_tenant_agent` partial, or re-registration breaks) — both larger changes than a warning warrants.

## Verification

- `tests/test_caura723_agent_scope_warning.py` — 9 cases: the four states, the fleet-narrowed probe, `/recall`, the agent-scoped 403 left unchanged, no-filter and good-filter silence, and a failing probe leaving search working. **5 fail without the source change**; the other 4 are no-regression guards that must hold in both directions.
- Targeted sweep post-rebase: 111 passed (`test_c27_strict_fleet_scoping`, `test_h05_multi_fleet_read_gate`, `test_search_recall_tracked_flag`, `test_search_caller_identity`, `test_c4_recall_items_alias`, `test_mcp_recall`, `test_route_authz_gaps`).
- Full `tests/`: 39 failures, byte-identical to `main`'s set on this machine (pre-existing FTS / relation-weight / request-observation env failures). `core-storage-api/tests/`: 369 passed, same 2 pre-existing CORS failures.
- ruff check and format clean; mypy identical to `main` on every changed file. `do_not_touch_sentinel` all 35 survive, `tenant_scope_gate` exit 0, `legacy_name_ratchet` reports one exempt `legacy-name-ok` line: the test patches `settings.memclaw_api_key` because that is the dual-read field auth Path 2 reads — patching `caura_api_key` is inert, since the two collapse at validation time.

No schema changes.

## Type of Change

- [ ] Documentation update
- [ ] Other:

## Checklist

- [ ] `mypy` passes
- [ ] `pytest` passes locally

### DIFF
--- a/core-api/src/core_api/clients/storage_client.py
+++ b/core-api/src/core_api/clients/storage_client.py
@@ -2610,6 +2610,22 @@ class CoreStorageClient:
             {"command_ids": command_ids, "tenant_id": tenant_id},
         )
 
+    async def agent_scope_probe(self, data: dict) -> dict | None:
+        """CAURA-723 — ``{has_memories, agent_registered}`` for one agent id.
+
+        Called only after a search that came back EMPTY, so a successful
+        agent-filtered search never pays for it. ``services/agent_scope``
+        carries the reasoning for that ordering; note it is not free, and this
+        docstring said the opposite until review caught it.
+
+        ``read=True``: the search this explains reads the replica, so answering
+        from the same replica keeps the probe's story consistent with the
+        result. Reading the primary could report "has memories" for a row the
+        search on the replica could not see — the one inconsistency that would
+        make the warning misleading rather than merely stale.
+        """
+        return await self._post_optional("/memories/agent-scope-probe", data, read=True)
+
     async def fleet_exists(self, tenant_id: str, fleet_id: str) -> bool:
         result = await self._get(
             "/fleet/exists",
--- a/core-api/src/core_api/openapi_responses.py
+++ b/core-api/src/core_api/openapi_responses.py
@@ -117,6 +117,16 @@ class RecallResponse(BaseModel):
     diagnostic: RecallDiagnostic | None = Field(
         default=None, description="Only when the request sets diagnostic=true."
     )
+    # A28 / CAURA-723 — /search has carried this since A28; /recall parses the
+    # same request and had the same silent failures, so it reports them too.
+    warnings: list[dict] | None = Field(
+        default=None,
+        description=(
+            "Coded, non-fatal caveats about the result set — e.g. "
+            "'filter_agent_unknown' when the agent filter names an id this "
+            "tenant has never seen. Null when there is nothing to say."
+        ),
+    )
 
 
 class VersionResponse(BaseModel):
--- a/core-api/src/core_api/routes/memories.py
+++ b/core-api/src/core_api/routes/memories.py
@@ -64,6 +64,7 @@ from core_api.schemas import (
     STMWriteResponse,
     UsageSummary,
 )
+from core_api.services.agent_scope import explain_agent_scope
 from core_api.services.agent_service import (
     authorize_memory_access,
     broker_label,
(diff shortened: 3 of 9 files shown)


eyal-bl:
@claude

eyal-bl:
@claude

eyal-bl:
@claude

eyal-bl:
@claude

eyal-bl:
@claude

PR #1439, opened 2026-09-09
fix(search): budget and label successor injection (D16)

ran-taig:
## Summary

Closes register row **D16** (caura_shared_tasks/v4-task-checklist; absorbs audit finding oss-0814-l-07 as a duplicate).

Successor injection was additive with no bound and no marker: every outdated/conflicted row in the already-trimmed result set pulled in **every** row superseding it (`supersedes_id IN (...)` has no per-predecessor cap), so a `top_k=5` call answered with 10 items — 53 on a conflict-heavy store — and nothing on the wire said which rows the query actually recalled. Both the REST schema ("Maximum results to return") and the MCP tool surface ("Max results") promised a ceiling the behavior never kept.

## Changes

* **Budget** (`load_and_serialize.py`): inject at most **one successor per stale row — the newest** (`created_at`, deterministic id tie-break), so a response holds at most **2×top_k** items. Deliberately **not** re-trimmed to `top_k`: cutting the tail would evict a successor or its predecessor at the boundary and reintroduce the stale-answer failure the A34 contract exists to prevent. If the newest correction was already recalled organically, nothing is injected for that predecessor.
* **Label** (`schemas.py`, `memory_service.py`): new `injected: bool = False` on `MemoryOut`, set only by the injection block — beside the row's existing `score: null`. A successor recalled on its own merit stays unmarked, because it *is* a result. Flows to REST and MCP unchanged (`model_dump` is full-fidelity).
* **Contract docs**: `SearchRequest.top_k` and the MCP `caura_recall.top_k` descriptions now state the real contract (recall bound, not response ceiling; marker to look for; ≤ 2×top_k). `plugin/tools.json` + tool-surface baselines regenerated via their documented scripts: **5287 cl100k (+18), under the 5320 ceiling** — ledger entry added, ceiling not raised.

## Relationship to A28 / A34

Extends **A28** (closed), which made lookup-bound truncation visible but deliberately did not budget the response. Successful injection still emits **no warning** — `warnings` keep meaning "enrichment incomplete". The **A34** pairing (successor immediately above its stale predecessor) is preserved under the budget; the new suite asserts it for every stale row.

## Tests

* New: `tests/test_d16_injection_budget_and_label.py` — 8 tests: newest-wins budget, 2×top_k bound with A34 pairing intact, wire labels, organic-successor cases (no sibling spend; newer correction still injected above an older organic one), no-warning-on-success, `top_k` description pin.
* Updated: `test_search_default_profile.py::test_top_k_cap_has_one_source_of_truth` — the verbatim MCP-description pin now includes the disclosure sentence (single-source-of-truth intent unchanged).
* D16 + A34 + A28 + token-budget + export-sync + description-baseline suites: **59 passed**. Full root suite: only the 32 pre-existing infra-dependent failures, byte-identical with the change stashed (verified both ways).

### DIFF
--- a/core-api/openapi.broker.json
+++ b/core-api/openapi.broker.json
@@ -909,6 +909,11 @@
             "title": "Id",
             "type": "string"
           },
+          "injected": {
+            "default": false,
+            "title": "Injected",
+            "type": "boolean"
+          },
           "is_inferred": {
             "default": false,
             "title": "Is Inferred",
@@ -1684,7 +1689,7 @@
           },
           "top_k": {
             "default": 5,
-            "description": "Maximum results to return (1-200, default 5).",
+            "description": "Maximum results the query returns (1-200, default 5). Not the response ceiling: each returned outdated/conflicted row also carries its newest correction, injected beyond this budget and marked injected: true (with score: null), so a response holds at most 2*top_k items.",
             "maximum": 200.0,
             "minimum": 1.0,
             "title": "Top K",
--- a/core-api/src/core_api/mcp_server.py
+++ b/core-api/src/core_api/mcp_server.py
@@ -937,7 +937,9 @@ async def caura_recall(
         int,
         Field(
             description=f"Max results, default {DEFAULT_SEARCH_TOP_K}. "
-            f"Values above {MAX_SEARCH_TOP_K} are capped to {MAX_SEARCH_TOP_K}."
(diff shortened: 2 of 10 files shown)


ran-taig:
@claude

ran-taig:
@erni-a

erni-a (approved):

PR #1441, opened 2026-09-09
fix(embedding): parse OPENAI_REQUEST_TIMEOUT_SECONDS defensively and re-read it at provider construction

ran-taig:
## Problem

Audit finding oss-0902-l-11 (low), validated against `origin/main` (77e2d8c3). Both sub-claims **confirmed**:

**(a) Import-time crash on a bad value.** `common/embedding/constants.py:41-43` parsed the var with bare `float(os.environ.get("OPENAI_REQUEST_TIMEOUT_SECONDS", "25.0"))`. `OPENAI_REQUEST_TIMEOUT_SECONDS=25s` (or an empty string) raises `ValueError: could not convert string to float: '25s'` out of the module body at import — a stack trace naming neither the env var nor the fix, before structured logging is wired up. core-worker takes the naked crash (`core_worker/consumer.py:53` imports `common.embedding`; nothing in core-worker validates this var first). core-api happens to be shielded only because `Settings()` (`core_api/config.py:117`) validates the same var earlier in its import sequence. `common/env_utils.py`'s `read_float_env` docstring names exactly this failure mode as its reason to exist, and every sibling knob in the module — and the LLM twin, `common/llm/constants.py:194` — already uses it. This was the last bare parse of this var.

**(b) Frozen before the credential bridge runs.** The "credential bridge" is `core_api.config.bridge_credentials_to_environ` (CAURA-595, `core_api/config.py:744`): it copies the `.env`-loaded `settings.openai_request_timeout_seconds` into `os.environ["OPENAI_REQUEST_TIMEOUT_SECONDS"]` (`config.py:778`) during FastAPI lifespan startup (`core_api/app.py:286-288`). Import-order proof: `core_api/app.py:38` imports `core_api.constants`, whose line 30 imports `common.embedding.constants` — so the constant is evaluated at module import, strictly before the lifespan runs; `common/embedding/providers/openai.py:21` then snapshots it via `from`-import and used it in `__init__` (old lines 130-137). Provider construction is lazy (first embed call, after the bridge), but the value it used was the frozen one. The LLM path already fixed this exact race — `common/llm/registry.py:134-150` re-reads the env at construction with a comment naming the bridge — so a `.env`-configured timeout governed LLM calls but was silently ignored for embeddings, despite `constants.py`'s comment promising "a single tunable controls both the LLM and the embedding paths".

## Fix

- `common/embedding/constants.py`: parse with `read_float_env("OPENAI_REQUEST_TIMEOUT_SECONDS", 25.0)` — garbage falls back to the default with a stderr WARN naming the variable. Fallback rather than raise because the default is safe for a timeout; raising stays reserved for misconfigs that guarantee 100% failed calls (the `_registry.get_embedding_provider` precedent). The `common.env_utils` import moves to the top of the file so the parse can use it (its "intentional late import" comment pointed at a module docstring that says nothing about imports).
- `common/embedding/providers/openai.py`: `__init__` re-reads the env var at construction time (`read_float_env("OPENAI_REQUEST_TIMEOUT_SECONDS", OPENAI_REQUEST_TIMEOUT_SECONDS)`), mirroring `common/llm/registry.py`, and feeds it to the httpx read/write/pool budgets. With the var unset — including the documented monkeypatch-the-module-binding test idiom — the module constant still governs, so the all-defaults behaviour is unchanged.
- The constants module docstring now records this one knob as the deliberate exception to "setenv after import is too late".

Not touched (out of this finding's scope): the neighbouring bare parses `EMBEDDING_RETRY_ATTEMPTS` / `EMBEDDING_RETRY_DELAY_S` (`constants.py`, same defect class, no bridge interaction).

## Tests

`tests/test_embedding_timeout_env_parse.py` (unit, repo-root style: `pytestmark = pytest.mark.unit`, `patch.dict` env isolation; import-time behaviour exercised by executing the real module file as a fresh, unregistered module so no live snapshot is disturbed):

- garbage value (`"25s"`, `""`, `"twenty-five"`) → module body completes, value falls back to 25.0, stderr WARN names the knob
- valid value parsed; unset → default
- env set **after** import (the bridge ordering) → reaches the client's `httpx.Timeout` read/write/pool
- env unset → module-binding fallback (the documented monkeypatch idiom keeps working)
- garbage env at construction → falls back to the module binding with a WARN

**Failed-on-pre-fix proof:** with the two source files stashed (origin/main code) the suite fails 5 of 8 — the three garbage-at-import cases (ValueError out of the module body), the bridge-ordering test (client got the frozen 25.0 instead of 3.5), and the garbage-at-construction warning test. All 8 pass with the fix.

## Gates

- `pytest` over the new file + `test_c38_local_embedder_dim_guard`, `test_local_embedder_instance_cache`, `test_embedding_openai_compat`, `test_bulk_write_mode`, `test_backend_backpressure_is_capacity`, `test_provider_protocols`, `test_platform_providers`, `test_env_utils`, `test_credential_bridge`, `test_embedding_retry`, `test_embedding_registry_eviction`: **252 passed**
- `ruff check common/ tests/test_embedding_timeout_env_parse.py`: clean; `ruff format --check` on the three touched files: clean
- `mypy --config-file common/mypy.ini common/`: no issues in 101 files

Audit tracker: oss-0902-l-11

### DIFF
--- a/common/embedding/constants.py
+++ b/common/embedding/constants.py
@@ -12,12 +12,22 @@ late. Examples:
 
     monkeypatch.setattr("common.embedding._service.EMBEDDING_RETRY_ATTEMPTS", 1)
     monkeypatch.setattr("common.embedding.providers.openai.OPENAI_REQUEST_TIMEOUT_SECONDS", 5.0)
+
+One deliberate exception: ``OPENAI_REQUEST_TIMEOUT_SECONDS`` is re-read
+from the environment when ``OpenAIEmbeddingProvider`` is constructed,
+with the module binding as the fallback — core-api's
+``bridge_credentials_to_environ()`` populates that env var during
+startup, after this module has already been imported. For that one knob
+``monkeypatch.setenv`` before provider construction works too; see
+``common/embedding/providers/openai.py``.
 """
 
 from __future__ import annotations
 
 import os
 import sys
+
+from common.env_utils import clamp_keepalive, read_float_env, read_int_env
 
 # Default model identifiers per provider. Override via env (e.g. swap
 # ``OPENAI_EMBEDDING_MODEL=text-embedding-3-large`` for a higher-dim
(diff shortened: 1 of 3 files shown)


ran-taig:
@claude review

ran-taig:
@claude

ran-taig:
@erni-a

erni-a (approved):

PR #1442, opened 2026-09-09
fix(search): hold the per-tenant storage bulkhead on the entity-lookup fall-through (oss-0814-l-06)

ran-taig:
## Problem

`per_tenant_storage_slot("storage_search", tenant_id)` is an **in-flight** cap — an `asyncio.Semaphore` held only across a single storage roundtrip (`core-api/src/core_api/middleware/per_tenant_concurrency.py:144-184`, cap default 4 via `per_tenant_storage_search_concurrency`). On the entity-lookup fall-through, `ExecuteScoredSearch` reached `sc.scored_search` **without holding it**:

- `classify_query.py` (entity-lookup) acquired **and released** the slot around `load_memories_by_ids`, then set `ctx.data["_storage_slot_acquired"] = True` *after* release.
- `execute_scored_search.py:172-177` (pre-fix) saw the sentinel and skipped acquisition: `if data.get("_storage_slot_acquired"): rows = await sc.scored_search(search_data)`.

At that moment the request held **zero** slots, so a tenant whose queries matched entity tokens but fell through (pool loaded, then thinned below `top_k` by visibility filtering — or every row dropped) could run unbounded concurrent `scored_search` roundtrips against the storage-reader pool, past its per-tenant budget. Exactly the noisy-neighbor hole the bulkhead exists to close.

The skip was C10's "don't charge twice for one logical search" dedup (#264) — a category error: the slot is an in-flight cap, not a rate-limit charge, and the two roundtrips are sequential, so acquiring around each holds at most one slot at any instant; there was never a double-charge to dedup. H-03 (#832) already called the skip a defect in passing ("left `_storage_slot_acquired` set so the scored search that does run skipped the bulkhead") and narrowed when the sentinel is set, but left the mechanism in place. Checked for in-flight work: #1425/#1426 (merged) cover the route-entry layer and settings authz, not this path.

## Fix

- `ExecuteScoredSearch` takes the slot **unconditionally**, matching the primary path and `memory_service._search_memories_legacy` (`memory_service.py:4801-4802`): held only across the `sc.scored_search` roundtrip, released on exit, unbounded acquire (queueing is the intended shape; the outer request budget caps wall time). No double-hold or deadlock is possible: classify's slot is provably released before this step runs (the `async with` wraps only the `load_memories_by_ids` call; the sentinel was set after exit).
- The now-purposeless `_storage_slot_acquired` sentinel is retired: the marker write is removed from `classify_query._collect_memories` and its `slot_acquired_marker` parameter is renamed to `pool_report` (it still carries the `_entity_pool_size` / `_entity_pool_loaded` decline diagnostics).

## Tests

`tests/test_c10_storage_slot_dedup.py` pinned the old skip contract; rewritten to pin the corrected one. The recorder now tracks **live holds**, so tests assert the roundtrip ran *inside* the slot (`held == 1` at `scored_search` call time), not merely that an acquire happened sometime — and that holds never overlap (`max_held == 1`, one logical search ≤ 1 slot at any instant).

- `test_entity_lookup_fallthrough_scored_search_runs_inside_the_slot` — the defect pin (audit path: pool loaded, all rows visibility-filtered).
- `test_stale_sentinel_does_not_disable_the_bulkhead` — a ctx still carrying the retired sentinel must not bypass the cap (reintroduction guard).
- Success path (plan-skip: no slot, no call), non-entity path, load-failure path (slot released by `__aexit__`, fall-through still gated) re-pinned.
- Verified failing on pre-fix code: with the src changes stashed, both new tests fail (`2 failed, 4 passed`); unstashed, all pass.
- `tests/test_audit_s6_c1_events.py` and `tests/test_entity_lookup_short_circuit.py` updated where they referenced the retired sentinel.

## Gates

- `pytest tests/ -m unit`: **3857 passed**, 3 failed — the 3 failures (`tests/test_a7_classifier_recall.py` entity-FTS tests) fail identically on clean `origin/main` in this environment (they need Postgres; no infra here), unrelated to this change.
- Affected files directly: `tests/test_c10_storage_slot_dedup.py tests/test_audit_s6_c1_events.py tests/test_entity_lookup_short_circuit.py tests/test_per_tenant_concurrency.py tests/pipeline/test_classify_query.py` → **67 passed**.
- `ruff check core-api/src/ tests/` — clean; `ruff format --check` on touched files — clean.
- `mypy src/` (core-api) — same 3 pre-existing missing-stub errors (`croniter`, `dateutil` in `common/`) as clean `origin/main`; no new errors.

Audit tracker: oss-0814-l-06

### DIFF
--- a/core-api/src/core_api/pipeline/steps/search/classify_query.py
+++ b/core-api/src/core_api/pipeline/steps/search/classify_query.py
@@ -212,7 +212,7 @@ class ClassifyQuery:
                         valid_at=valid_at,
                         readable_tenant_ids=readable_tenant_ids,
                         strict_fleet_scoping=bool(ctx.data.get("strict_fleet_scoping")),
-                        slot_acquired_marker=ctx.data,
+                        pool_report=ctx.data,
                     )
 
                     # H-03: the short-circuit REPLACES scoring rather than
@@ -468,7 +468,7 @@ class ClassifyQuery:
         status_filter: str | None = None,
         valid_at: datetime | None = None,
         readable_tenant_ids: list[str] | None = None,
-        slot_acquired_marker: dict | None = None,
+        pool_report: dict | None = None,
         strict_fleet_scoping: bool = False,
     ) -> list[types.SimpleNamespace]:
         """Load memories linked to graph-expanded entities, scored by hop distance."""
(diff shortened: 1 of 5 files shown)


ran-taig:
@claude review

ran-taig:
@erni-a

erni-a (approved):

PR #1443, opened 2026-09-09
docs(openapi): declare recall_raw on RecallDiagnostic (oss-0902-l-16)

ran-taig:
## Problem

`RecallDiagnostic` — the spec-only OpenAPI model documenting `/recall`'s `diagnostic` payload — omits the `recall_raw` key the handler actually serializes, so the generated spec and any client/docs built from it understate the real response.

Validation evidence (origin/main @ 77e2d8c3):

- `core-api/src/core_api/services/recall_service.py:335` — `"recall_raw": completion` on the LLM path; `:182` and `:232` emit `"recall_raw": None` on the no-memories and recall-disabled branches. The key is present on **every** branch whenever the request sets `diagnostic=true`; runtime type is `str | None`.
- `core-api/src/core_api/routes/memories.py:2175` — `POST /recall` documents `_oar.RecallResponse` (which nests `RecallDiagnostic`) via spec-only `responses=`, and returns the service dict verbatim (`:2276`), so `recall_raw` lands on the wire.
- `core-api/src/core_api/openapi_responses.py:88-99` (pre-fix) — `RecallDiagnostic` declares `recall_prompt`/`recall_model`/`recall_provider`/… but not `recall_raw`.

How the drift happened: WT-1 (#986, Aug 25) added `recall_raw` to the handler; C33 (#990, Aug 26) created `RecallDiagnostic` one day later without it; CAURA-722 (#1357, Sep 8) widened the model again without backfilling. Nothing enforces these spec-only models at runtime — the `openapi_responses` module docstring makes handler↔model sync a same-PR rule instead.

## Fix

Direction check: the mismatch is the model lagging the handler, not the handler over-returning — WT-1's comments (`recall_service.py:74`, `:313-314`) and its dedicated test (`tests/test_wt1_recall_answer_extraction.py::test_diagnostic_carries_raw_completion`) document `diagnostic.recall_raw` as deliberate, load-bearing behavior (the reasoning scaffold stays inspectable while `summary` carries only the extracted answer). So the model is widened, not the response narrowed.

- Declare `recall_raw: str | None` on `RecallDiagnostic`, positioned in the handler's serialization order (right after `recall_prompt`), with a `Field(description=...)` in the file's style saying what it is and when it is null.
- No runtime behavior change (spec-only model). The broker contract baseline is unaffected — `/recall` is not a broker operation and `openapi.broker.json` never references `RecallDiagnostic`; `gen_broker_openapi.py --check` passes unchanged.

## Tests

Extended `tests/test_c33_openapi_completeness.py` (new invariant 4, giving the module-docstring sync rule teeth for the payload that already drifted):

- `test_recall_diagnostic_documents_recall_raw` — the model declares a nullable `recall_raw`.
- `test_recall_diagnostic_covers_every_wire_key` — the `diagnostic` keys emitted by all three `summarize_memories` builder branches (no-memories, recall-disabled, LLM) are a subset of declared `RecallDiagnostic` fields.

Failed-on-pre-fix proof: with only the model change stashed, both new tests fail — `test_recall_diagnostic_covers_every_wire_key` reports exactly `['recall_raw']` missing on each branch; unstashed, all pass.

## Gates run

- `pytest tests/test_c33_openapi_completeness.py tests/test_wt1_recall_answer_extraction.py` — 21 passed
- `pytest tests/test_broker_contract.py tests/test_mcp_manage_structured_returns.py` — 8 passed
- `python core-api/scripts/gen_broker_openapi.py --check` — baseline up to date (8 operations)
- `ruff check core-api/src/ tests/test_c33_openapi_completeness.py` — clean
- `ruff format --check` on both touched files — clean
- `cd core-api && mypy src/` — Success: no issues found in 204 source files

Audit tracker: oss-0902-l-16

### DIFF
--- a/core-api/src/core_api/openapi_responses.py
+++ b/core-api/src/core_api/openapi_responses.py
@@ -87,6 +87,12 @@ class MemoryStatusPatchResponse(BaseModel):
 
 class RecallDiagnostic(BaseModel):
     recall_prompt: str | None
+    # WT-1 — ``summary`` carries only the extracted final answer; the raw
+    # completion stays inspectable here.
+    recall_raw: str | None = Field(
+        description="Unfiltered LLM completion (reasoning scaffold + answer marker) that summary "
+        "was extracted from; null when no LLM ran (no matches, or summarization disabled)."
+    )
     recall_model: str | None
     recall_provider: str | None
     all_candidates: list
--- a/tests/test_c33_openapi_completeness.py
+++ b/tests/test_c33_openapi_completeness.py
@@ -1,6 +1,6 @@
 """C33 — OpenAPI completeness guarantees.
 
-Three invariants:
+Four invariants:
 
 1. Every endpoint annotated via ``openapi_responses`` actually surfaces a
    non-empty success schema in the generated spec (the ``responses=``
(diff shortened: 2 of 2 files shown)


ran-taig:
@claude review

ran-taig:
@erni-a

erni-a (approved):

PR #1444, opened 2026-09-09
fix(search): fall back to hop-0 seeds when expand_graph fails for all fleets

ran-taig:
## Problem

When every per-fleet `expand_graph` call fails, `ClassifyQuery._expand_per_fleet` logs each failure and returns `{}` — discarding the seed entities the entity FTS had already resolved (`core-api/src/core_api/pipeline/steps/search/classify_query.py:419-437` pre-fix). With `fleet_ids=None` that is a single call, so one storage hiccup is "all fleets".

That empty dict silently zeroes both consumers for the request:

- **Entity lookup**: `_collect_memories` receives no entities, loads no links, and the ENTITY_LOOKUP short-circuit falls through — logged, misleadingly, as `"entity matched but no linked memories"` (`classify_query.py:297`).
- **Hop boost**: the `_classified_entity_hops` stash (`classify_query.py:180`, `:300`) hands `ParallelEmbedAndEntityBoost` an empty dict, and its `precomputed_hops is not None` check (`parallel_embed_entity_boost.py:58-60`) treats `{}` as authoritative — so it does not re-derive FTS either, and hop-boost contributes nothing.

A successful `expand_graph` response always contains the seeds at hop 0 / weight 1.0 (`core-storage-api/src/core_storage_api/services/postgres_service.py` — `entity_expand_graph` seeds its result dict before traversing), and every per-fleet call receives the full seed list, so an empty merge for a non-empty seed set can only mean **no call succeeded** — and any single surviving fleet keeps all seeds, so partial failure needs no case of its own.

## Fix

On an all-fleets-failed merge, `_expand_per_fleet` now logs a warning and degrades to the hop-0 seed set (`dict.fromkeys(seed_ids, (0, 1.0))`) — the exact shape the `graph_expand` off gate in `_hops_for_seeds` already returns, so matched entities stay retrievable and only the (unavailable) graph neighbourhood is lost. Success-path and partial-failure behaviour are unchanged.

## Tests

Three new tests in `tests/pipeline/test_classify_query.py`, next to the existing `test_expand_per_fleet_partial_failure`:

- `test_expand_per_fleet_total_failure_falls_back_to_hop0_seeds` — helper-level: all fleets fail → seeds returned at `(0, 1.0)`.
- `test_entity_lookup_survives_expand_graph_failure` — execute-level: expansion down, linked memory exists → ENTITY_LOOKUP still fires off the hop-0 seed (and the link lookup is made for the seed, not an empty set).
- `test_hop_boost_stash_survives_expand_graph_failure` — temporal-decline path: `_classified_entity_hops` carries `{seed: (0, 1.0)}`, not `{}`.

**Verified to fail on pre-fix code**: with the src change stashed, all three fail (helper returns `{}`; query falls through to SEMANTIC_SEARCH; stash is `{}`); with it restored, all three pass.

## Gates

- `pytest` (unit, `EMBEDDING_PROVIDER=fake`): 96 passed across `tests/pipeline/test_classify_query.py`, `test_entity_lookup_short_circuit.py`, `test_graph_expand_short_circuit.py`, `test_entity_retrieval_flag.py`, `test_skip_entity_boost_on_decline.py`, `test_c10_storage_slot_dedup.py`, `test_caura_000_graph_frontier_cap.py`, `test_entity_tokens.py`, `test_caura722_entity_diagnostic.py` (4 integration tests deselected — no local Postgres). `tests/pipeline/test_search_pipeline.py`: 19 passed, 2 pre-existing failures (`test_recall_returns_memory_with_pending_embedding[pipeline|legacy]`) that fail identically on clean `origin/main` (schema-fixture precondition: "search_vector must be populated"), unrelated to this change.
- `ruff check core-api/src/ tests/pipeline/` — clean.
- `ruff format --check` on both touched files — clean.
- `mypy src/` from `core-api/` (CI-equivalent env) — Success, no issues in 204 files.

Audit tracker: oss-0814-l-05

### DIFF
--- a/core-api/src/core_api/pipeline/steps/search/classify_query.py
+++ b/core-api/src/core_api/pipeline/steps/search/classify_query.py
@@ -364,6 +364,10 @@ class ClassifyQuery:
     ) -> dict[UUID, tuple[int, float]]:
         """Expand *seed_ids* per fleet, or return them as hop-0 when expansion is off.
 
+        Expansion failure degrades the same way: ``_expand_per_fleet`` falls
+        back to the hop-0 seed set when every fleet's call fails, so callers
+        are guaranteed a non-empty dict for a non-empty seed set either way.
+
         The gate mirrors ``_entity_boost_via_storage`` exactly
         (``graph_expand and graph_max_hops > 0`` → expand, else hop-0 seeds
         with neutral weight), so ``search.graph_retrieval`` means the same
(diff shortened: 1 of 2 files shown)


ran-taig:
@claude review

ran-taig:
@erni-a

erni-a (approved):

PR #1448, opened 2026-09-09
feat(search): opt-in ANN candidate pool — HNSW finally serves the primary path (HNSW plan PR2)

ran-taig:
## What

`memory_scored_search` full-scans the tenant slice on every search — both candidate branches ORDER BY blended expressions, never raw distance, so `ix_memories_embedding_hnsw` is structurally unusable on the read path (validated in the plan doc; ~1,300 ms p50 at 31k rows per the in-code measurement, 188 s measured on the largest on-prem deployment). PR1 (#1429) made each scanned row ~6× cheaper; **this makes the scan itself bounded**: rig-measured **8.3 ms** end-to-end for the pooled statement vs 90 ms (PR1 shape) / 427 ms (pre-PR1) at 50k rows.

New **`ann_pool_size`** knob (`SearchKnob (0, 1000)`, `sql=True`; emitted by `_resolve_search_params` with default **0 = off = byte-identical statement** — the rollback story is the knob). When > 0 and pgvector ≥ 0.8, the ingredients CTE is gated on a `candidate_pool` CTE unioning one index-served arm per admission signal:

| arm | index | covers |
|---|---|---|
| ann (knob size) | `ix_memories_embedding_hnsw` | cosine relevance |
| fts (50) | GIN `search_vector` | lexical matches, **incl. NULL-embedding rows** (CAURA-594/679 contract) |
| recency (50) | `ix_memories_tenant_created_active` | freshness/temporal elevation |
| date (50, when a date window was extracted) | btree | `date_range_boost` candidates |
| boosted | pk | entity-expansion ids, verbatim |

**Every arm carries the full row-filter set** (tenant, visibility, agent, type, status carve-out, valid_at) — pool admission can never widen visibility; the boosted arm especially must re-filter ids that graph expansion produced without visibility context. Scoring formula, union dedup, diagnostic columns, outer join: untouched code paths — **pool ⊇ corpus ⇒ results are exactly the full-scan results** (pinned by a DB test).

## Planner pinning + capability gate

- Per-query transaction-local GUCs via `set_config`: `hnsw.ef_search ≥ pool`, `hnsw.iterative_scan = relaxed_order` — the planner's free choice is borderline at mid-size corpora (measured), and iterative scans keep a filtered arm walking until its LIMIT is satisfied (bounded by `hnsw.max_scan_tuples`, so a pathological filter degrades to an under-filled pool, never an unbounded crawl).
- **Probe**: one `pg_extension` lookup per process, cached; pgvector < 0.8 keeps the exact pre-pool statement and logs a WARNING. A probe *failure* is deliberately not cached — a transient error can't stick the process on the fallback.
- **Migration 044**: `ALTER FUNCTION cosine_distance COST 100` (honest cost — a 1024-dim float loop priced as an int compare is why seq scans win cost fights they should lose; with COST 100 the rig's planner freely chose HNSW). Best-effort DO block: `insufficient_privilege` → NOTICE, because managed/on-prem app users often don't own the extension; the pool works without it via GUC pinning.
- Mutual exclusion with A49: `ann_pool_size` supersedes `candidate_pool_size` when both arrive (logged); the knob census test now records four non-agent-tunable rollout knobs.

## Rollout posture

Off by default everywhere. The plan doc (`docs/plans/hnsw-two-stage-retrieval.md`, local) carries the parity analysis — including the measured crowding-regime caveat under the legacy formula — and gates rollout on the offline harness (PR4), with shadow-compare telemetry (PR3) before any tenant serves pooled results. Pre-flight for real deployments: `SELECT extversion FROM pg_extension WHERE extname='vector'` ≥ 0.8 on SaaS prod and the eToro box.

## Testing

- **Statement-shape pins** (`core-storage-api/tests/test_ann_pool_statement.py`, 13 tests, DB-free): default statement byte-free of pool machinery; ann-mode GUC-then-statement ordering; one arm per signal with the tenant predicate counted per arm; blank-query drops the fts arm; A49 supersession; date arm; probe parse/fallback/failure-not-cached.
- **DB-backed behaviour** (`tests/test_ann_pool_behavior.py`, 7 tests): pool ⊇ corpus exactness vs full scan; ANN-arm admission with shrunken side arms; NULL-embedding row discoverable in pool mode; boosted-arm admission; `scope_agent` rows invisible in pool mode; pgvector < 0.8 result-parity with knob-off; HNSW-servability of the arm via `EXPLAIN (FORMAT JSON)`.
- Full suites green against a freshly migrated Postgres 16 + pgvector 0.8.1: **6,517 root + 386 storage** (includes migration 044 applied), ruff + format + mypy clean.

## Context

PR2 of the HNSW two-stage retrieval plan. PR3 (shadow mode + D12 arm provenance + org-settings combo validation) and PR4 (offline harness gate + staged rollout) follow.

### DIFF
--- a/common/constants.py
+++ b/common/constants.py
@@ -555,6 +555,15 @@ SEARCH_KNOBS: dict[str, SearchKnob] = {
     "candidate_pool_size": SearchKnob(int, (0, 200), sql=True),
     # A50 unified: 0 = legacy multiplicative score; 1 = unified relevance-dominant formula.
     "score_formula": SearchKnob(int, (0, 1), sql=True),
+    # HNSW two-stage retrieval (PR2 of docs/plans/hnsw-two-stage-retrieval.md):
+    # 0 = off (full-scan candidate window, unchanged); >0 = storage admits
+    # candidates through index-served pool arms (ANN top-N by cosine via the
+    # memories HNSW index, FTS, recency, date window, entity-boosted ids) and
+    # runs the scoring formula over that pool only. Needs pgvector >= 0.8 at
+    # runtime (iterative scans); storage probes once and silently keeps the
+    # full scan below that. Mutually exclusive with ``candidate_pool_size`` —
+    # storage lets ann win if both arrive, but don't set both.
+    "ann_pool_size": SearchKnob(int, (0, 1000), sql=True),
 }
 
 # The wire contract, derived. Core-api's two search-path builders project
--- a/core-api/src/core_api/constants.py
+++ b/core-api/src/core_api/constants.py
@@ -319,6 +319,18 @@ CANDIDATE_POOL_SIZE = 0
 # tenant via default_search_profile.score_formula to A/B old vs new on the benchmark.
 # Rationale + offline calibration: docs/ranking/unified-ranking-formula.md.
 SCORE_FORMULA = 0
+# HNSW two-stage retrieval (PR2, docs/plans/hnsw-two-stage-retrieval.md).
+# 0 = OFF: storage keeps the full-scan candidate window (current behaviour).
+# >0 = storage admits candidates through index-served pool arms (ANN top-N by
+# cosine via the memories HNSW index, plus FTS / recency / date-window /
+# entity-boosted arms) and runs the scoring formula over that pool only —
+# O(log N + pool) instead of O(tenant rows) per search. Needs pgvector >= 0.8
+# at runtime; storage probes once and silently keeps the full scan below that.
+# Enable per-tenant via ``default_search_profile.ann_pool_size``; keep 0
+# globally until the offline harness validates result parity (see the plan
+# doc's crowding-regime analysis). Mutually exclusive with
+# ``candidate_pool_size`` — storage lets ann win if both arrive.
+ANN_POOL_SIZE = 0
 FRESHNESS_DECAY_DAYS = 90
 FRESHNESS_FLOOR = 0.7
 ENTITY_BOOST_FACTOR = 1.3
--- a/core-api/src/core_api/services/memory_service.py
+++ b/core-api/src/core_api/services/memory_service.py
@@ -45,6 +45,7 @@ from common.embedding import (
 from common.events import publish_memory_embed_request, publish_memory_enrich_request
 from common.governance import mask, scan
 from core_api.constants import (
+    ANN_POOL_SIZE,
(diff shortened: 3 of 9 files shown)


ran-taig:
@claude

ran-taig:
Both review items addressed in `4057281a`:

**1. `get_read_session` transaction semantics (the SET LOCAL concern)** — verified first: `_build_engine` sets no `isolation_level` (the only AUTOCOMMIT in the tree is the migration lock connection), so the reader session autobegins one transaction across the two `set_config` executes and the pooled statement — the current behavior is correct. But the dependency was implicit, so it's now pinned on the real machinery per fix option 2:
- `test_read_session_preserves_set_local_across_executes` — the session property itself: a value `set_config`'d in one execute must be visible via `SHOW` in the next execute on the same `get_read_session` session. Flipping the reader engine to AUTOCOMMIT fails this loudly.
- `test_real_search_path_has_gucs_live_at_statement_time` — the **actual** `memory_scored_search` ann-mode path, instrumented mid-flight: at the moment the pooled statement executes on the real session, `SHOW hnsw.ef_search` / `SHOW hnsw.iterative_scan` must return the values the production code set (`200` / `relaxed_order`). This asserts GUC liveness directly rather than inferring it from plan shape, so it can't pass vacuously at test-corpus scale the way an EXPLAIN-based check could (the planner legitimately prefers seq scan on tiny corpora — that's why `test_ann_arm_is_hnsw_servable` forces the choice; it remains as the index-servability pin).

The GUC block's comment now names the dependency and both guarding tests.

**2. Probe thundering herd** — took the lock option: probes now coalesce on a double-checked `asyncio.Lock` (8 concurrent first-calls → one `pg_extension` read, pinned by `test_concurrent_probes_coalesce_to_one_lookup`). Failure semantics unchanged: a failed probe still caches nothing, and waiters queued behind a failure retry serially under the lock — bounded, never a herd. One deliberate wrinkle documented in-code: the lock is rebuilt when the running event loop changes (asyncio primitives bind to the first loop that awaits them; the test suite runs per-test loops, production has one loop for the process lifetime).

Suites against a freshly migrated Postgres: 6,519 root + 387 storage, ruff/mypy clean. Rebased onto the updated branch head before pushing — fast-forward, no history rewritten.

ran-taig:
@claude

ran-taig:
@erni-a

erni-a (approved):

PR #1454, opened 2026-09-09
flip and contract the audit topic family

Eldad-Caura:
Last of the seven families. This module now declares no topic carrying the outgoing prefix.

## Flip and contract ship together

This started as flip-only — add `"audit"` to `FLIPPED_FAMILIES`. `test_every_flipped_family_is_contracted_so_the_default_bus_is_legal` failed on it, correctly.

Flipping alone leaves `publish_name` returning the twin while `subscribe_names(dual=False)` — the **default** — still returns the outgoing name. `unbound_publish_topics` then reports the family and every default-constructed `PubSubEventBus` raises **fleet-wide**, not just in the service that moved. Per that test's own docstring, `memory` sat in that state for four days and `org` for hours, and *"No test failed either time; the breakage was found by reading the module."* The test exists because of those two incidents, and it did its job here.

## The stated reason audit was last was wrong

The comment said these rows are hash-chained and that a lost or reordered audit event is the one failure replay cannot repair. That conflates two unrelated paths, and it is corrected in this diff rather than left to mislead the next reader:

- **The hash chain** is the broker audit log — an install batches locally-chained entries over HTTP to a receiver that verifies `prev_hash`/`entry_hash` and refuses a discontinuity. It never travels this topic.
- **This topic** carries an append-only notification whose consumer tolerates duplicate delivery by design. No ordering requirement, no chain to break.

## The real hazard, and the evidence against it

The opposite shape, and invisible: the audit publisher is fire-and-forget and swallows every exception, deliberately, so a bus outage cannot take down the request path that produced the event. A flip onto a name that did not exist would therefore lose every audit event with **no error anywhere**, into an append-only store with no replay.

What makes this safe is existence and attachment, measured 2026-09-10:

| Check | Result |
|---|---|
| Twin topics | both environments, each with a `-dlq` |
| Subscriptions | ACTIVE and attached on all four |
| Readiness gate | 17/17, nothing unbound |
| Pre-flip week | 20 production publishes on the legacy name, **0** on the twin |
| Control (`memory`, `security`) | six- and four-figure reads → instrument confirmed live |

The control matters: expired credentials fail silently here and return empty.

## Tests

Three tests used `audit` as their example of a member still on the outgoing prefix — it was the last one. They now use a synthetic name. `renamed()` rewrites whatever the first dot-segment is rather than matching the outgoing brand, so they keep exercising the expansion **without minting a fresh occurrence of the old brand**, which is the property its docstring says the derivation exists to provide.

Ratchet: `1 removed, -1 net`. Ruff clean. `tests/test_events/` is 18 failed / 130 passed identically on this branch and on a clean `main` control worktree — all 18 are `ModuleNotFoundError: No module named 'google'` in this environment. The wider suite hits 38 collection errors on **both** branch and control, so it is unverified locally and CI is the real check.

## Sequencing

This is the OSS half. The caura-enterprise mirror follows **after this merges**, re-recording the vendored drift baseline in the same commit — doing it before the OSS merge is what broke #1637.

### DIFF
--- a/common/events/topics.py
+++ b/common/events/topics.py
@@ -25,7 +25,20 @@ class Memory(enum.StrEnum):
 
 
 class Audit(enum.StrEnum):
-    EVENT_RECORDED = "memclaw.audit.event-recorded"
+    # CONTRACTED 2026-09-10, in the same change as its flip and NOT as a matter
+    # of taste. Flipping without contracting leaves ``publish_name`` returning
+    # the twin while ``subscribe_names(dual=False)`` -- the DEFAULT -- still
+    # returns the outgoing name, so ``unbound_publish_topics`` reports the
+    # family and every default-constructed ``PubSubEventBus`` raises FLEET-WIDE,
+    # not just in the service that moved. ``memory`` sat in that state for four
+    # days and ``org`` for hours; both were found by reading the module rather
+    # than by a failing test, which is why
+    # ``test_every_flipped_family_is_contracted_so_the_default_bus_is_legal``
+    # now exists and why these two steps ship together.
+    #
+    # This was the LAST declared topic carrying the outgoing prefix. This module
+    # now declares none.
+    EVENT_RECORDED = "caura.audit.event-recorded"
 
 
 class Pipeline(enum.StrEnum):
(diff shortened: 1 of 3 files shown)


Eldad-Caura:
@erni-a pelase approve

erni-a (approved):

PR #1455, opened 2026-09-10
feat(bulk): one contradiction pass per subject on a coherent batch (A73)

arkash20:
Seeding a biography, or importing a document, writes dozens of rows about one subject within seconds. The bulk path fired `run_contradiction_detection` **once per created row**, so every row was judged against a store its own siblings were still landing in — and complementary facts came back `conflicted`.

In one observed store **~40% of a 55-fact biography was flagged before the first conversation**. A conflicted row carries a 0.5 ranking penalty, so the store began every retrieval already handicapped.

`write.bulk_subject_batching` judges the batch's **last row per subject** instead of every row.

## This is not "detect less"

By the time that row is judged, every sibling is committed and is an ordinary candidate for it. So a genuine intra-batch contradiction is still found — and so is one against the pre-existing store. What disappears is the batch conflicting with **itself**, N ways, which is the only part that was ever noise.

Three details carry the correctness:

- the accumulator is **keyed by subject and overwritten** as the loop advances, so the survivor is the most recent claim — the one whose candidate set is complete;
- the passes are scheduled **after** the loop, because before it ends the later siblings aren't committed, and judging then reproduces the very race this fixes;
- a row with **no resolved subject** keeps per-row detection. There's nothing to group it by, and inventing a key would fold unrelated rows into one pass.

The embedding check still comes first, so batching can't swallow a row that has no vector to judge with.

## Flag

`write.bulk_subject_batching`, per-tenant, **default off** — like every switch that changes what a write produces. It turns N detections into one per subject, so it **reduces** LLM calls and isn't blocked by the no-new-LLM-calls hold.

## Testing

12 new tests, covering the flag default, that a subjectless row keeps the old path, that the deferred passes run *after* the loop, and that the embedding check still precedes the batching branch.

⚠️ **Reproducing locally:** a venv predating #1134's `mcp>=2.1.1` bump fails collection on 38 modules with `cannot import name 'CacheHint'`. That's environment drift, not this branch — `uv pip install "mcp>=2.1.1,<3"` fixes it.

### DIFF
--- a/core-api/src/core_api/services/memory_service.py
+++ b/core-api/src/core_api/services/memory_service.py
@@ -2146,6 +2146,10 @@ async def create_memories_bulk(
             and tenant_config.enrichment_provider != "none"
         )
 
+        # A73 — subject -> the batch's most recent row for that subject, so one
+        # contradiction pass covers the whole run of writes about it.
+        bulk_subject_batching = getattr(tenant_config, "bulk_subject_batching", False)
+        deferred_by_subject: dict[str, tuple] = {}
         reembed_batch: list[tuple[UUID, str]] = []
         for orig_idx, mem_data, mem_id in resolved:
             if tenant_config.entity_extraction_enabled:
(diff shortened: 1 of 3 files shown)


arkash20:
CI green. Reduces LLM calls (N rows -> one per subject), ships dark behind write.bulk_subject_batching. @Eldad-Caura please approve

Eldad-Caura (approved):

PR #1457, opened 2026-09-10
fix(memory): populate is_inferred so the invariant that reads it can fire (A62)

arkash20:
Migration 036 (A55) added `memories.is_inferred`. **Nothing ever wrote it**, so every row in every tenant read `False` = "directly stated". That silently disabled the guard in `resolution.resolve`:

```python
if is_inferred and action in _DESTRUCTIVE:
```

which exists so a memory the *system* materialised cannot destructively overturn one a user actually stated. With the column permanently False, a crystallized merge or a generated insight could retire a fact a person wrote — and the branch protecting against that **had never executed once**.

## Who is marked, and who isn't

| writer | inferred | why |
|---|---|---|
| crystallizer | ✅ | an LLM re-extraction merging a cluster into a claim nobody stated in those words |
| insights | ✅ | the system's own conclusion drawn *across* memories |
| atomic-fact fan-out | ❌ | a claim **extracted** from content the user wrote — marking it would demote a user-stated fact below every other user-stated fact |

That last row is the one worth arguing about, and I think it's clearly right: fan-out splits a user's sentence, it doesn't infer anything.

## Why it's off the wire

Server-set and keyword-only on `create_memory` / `create_memories_bulk`; absent from `MemoryCreate` and `BulkMemoryItem`. On the wire it would be an **opt-out, not a guard** — a writer must not get to choose which side of the invariant its rows sit on. It stays on `MemoryOut`, because the resolver reads it off the fetched row.

## `confidence` is left NULL, deliberately

The column documents NULL as "unknown/legacy", and neither writer has an honest per-claim confidence to report — the crystallizer's `weight` is *salience*, not confidence in the claim. A fabricated number would be worse than absence, because the resolver would then act on a signal nobody measured. A test pins that choice so it reads as intent rather than omission.

## Testing

13 new tests, including that the guard it feeds still exists (if that branch is ever deleted, populating the column stops meaning anything and this is what notices), and that STM is *not* seeded — STM writes through `WriteSTMNote`, so there's no row with this column.

Two existing doubles pinned the old signatures and failed on the new keyword; both now take `**_kw` so an unrelated server-side parameter can't break them again.

### DIFF
--- a/core-api/src/core_api/pipeline/steps/write/write_memory_row.py
+++ b/core-api/src/core_api/pipeline/steps/write/write_memory_row.py
@@ -84,6 +84,18 @@ class WriteMemoryRow:
             "ts_valid_end": str(fields["ts_valid_end"]) if fields.get("ts_valid_end") else None,
             "status": fields["status"],
             "visibility": data.visibility or "scope_team",
+            # A62 — migration 036 added this column and nothing ever wrote it, so
+            # every row read ``False`` = "directly stated". That silently disabled
+            # the invariant at ``resolution.py``: ``if is_inferred and action in
+            # _DESTRUCTIVE`` exists so a memory the SYSTEM materialised cannot
+            # destructively overturn one a user actually stated — and with the
+            # column always False it has never once fired.
+            #
+            # Server-set only. It is absent from ``MemoryCreate`` (it lives on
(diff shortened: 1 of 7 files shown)


arkash20:
CI green. Restores a contradiction invariant that has never fired. @Eldad-Caura please approve

Eldad-Caura (approved):

PR #1458, opened 2026-09-10
feat(extraction): write back the predicate so the RDF path can fire (A65 groundwork)

arkash20:
A63 wrote back `subject_entity_id` from the extractor and stopped there. Its own comment explains why that column was NULL to begin with — the write-time triple path fires only on narrow phrase regexes. But the deterministic RDF contradiction path keys on **(subject, predicate)**, so filling one of three columns left it exactly as dormant as before: every row got a subject and no predicate.

This is the same write-back for the other two columns, plus the canonicalisation that makes them comparable across wordings.

## Read this before assuming it closes A65

**It does not fix A65's two measured probe cases.** *"finished the Snowflake cutover for analytics"* vs *"the analytics warehouse is BigQuery"* requires **inferring** that a cutover changes the warehouse — semantic work no vocabulary mapping does. The other candidate fix (a tiered judge) targets those, and it's parked under the no-new-LLM-calls hold.

What this removes is the precondition that made the deterministic path unable to fire **for anyone**: `predicate` NULL on nearly every row.

## The canonicaliser

Normalises *shape* only — case, spaces, hyphens, and a leading `is_`/`has_`/`was_` where the bare form is canonical. No synonyms, no stemming, no inference. Anything that doesn't land in `SINGLE_VALUE_PREDICATES` returns `None` and the row keeps a NULL predicate, because **a wrong predicate is worse than none**: the RDF path treats (subject, predicate) as authoritative and would compare two unrelated attributes as the same one.

One subtlety worth the review: **the bare form deliberately wins over the raw one.** `SINGLE_VALUE_PREDICATES` contains *both* `is_located_in` and `located_in` — and `has_status` and `status`, and three more pairs. Returning whichever matched first would split one attribute across two canonical names, reintroducing the exact bug at the layer meant to fix it. Stripping first makes both spellings converge:

```
"is located in" -> located_in   |  "located in" -> located_in   MATCH
"has_status"    -> status       |  "status"     -> status       MATCH
"works_on"      -> None         (multi-valued: a second value adds, not replaces)
```

## Write-back rules

All inherited from A63, for A63's reasons: exactly one canonical relation starting at the subject; storage guarded by `predicate IS NULL` so the regex path's higher-fidelity value is never clobbered; non-fatal on failure; every outcome logged so "no canonical predicate" stays distinguishable from "never ran".

`subject_ids` from the subject write-back is **not** reused — it's computed inside a conditional, so reaching into its scope would `NameError` on every path where that block didn't run. Recomputed from `filtered` instead, and a test pins that.

## Testing

18 new tests, including seven wording pairs that must converge, that prefix-stripping can never invent a predicate outside the vocabulary, and the scope check above.

Full suite: **27 failures, identical to main's 27** — none new. `ruff` clean; `mypy` clean on core-storage-api, unchanged on core-api.

### DIFF
--- a/core-api/src/core_api/clients/storage_client.py
+++ b/core-api/src/core_api/clients/storage_client.py
@@ -730,6 +730,26 @@ class CoreStorageClient:
         )
         return bool(result and result.get("updated"))
 
+    async def set_predicate_if_null(
+        self, memory_id: str, tenant_id: str, predicate: str, object_value: str
+    ) -> bool:
+        """A65 — conditional predicate/object write-back from the extraction worker.
+
+        Sibling of ``set_subject_entity_if_null``. Storage-side single UPDATE
+        guarded by ``predicate IS NULL`` (the write-time triple path's value
+        wins). Returns whether the row was actually updated; ``False`` is a
+        benign skip."""
+        result = await self._post(
+            f"/memories/{memory_id}/predicate",
+            {
+                "tenant_id": tenant_id,
+                "predicate": predicate,
+                "object_value": object_value,
+            },
+            read=False,
+        )
+        return bool((result or {}).get("updated"))
+
     async def update_memory_status(
         self,
         memory_id: str,
--- a/core-api/src/core_api/services/entity_extraction_worker.py
+++ b/core-api/src/core_api/services/entity_extraction_worker.py
@@ -3,9 +3,10 @@
 import asyncio
 import logging
 import re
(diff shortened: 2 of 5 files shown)


arkash20:
CI green. Groundwork only — see the PR body for what this does NOT fix. @Eldad-Caura please approve

Eldad-Caura (approved):

PR #1459, opened 2026-09-10
fix(contradiction): stop reading a formatting difference as a contradiction (A35)

arkash20:
`memory_find_rdf_conflicts` selected a conflict with raw string inequality:

```python
Memory.object_value != object_value
```

So `"7,500 rpm"` and `"7500 RPM"` read as **two different values** for one (subject, predicate) pair, and the row was flagged as a contradiction it isn't. That's not free — the losing row carries a **0.5 ranking penalty**, so a comma quietly demoted a correct memory.

Both operands are now normalised by the same expression:

```sql
lower(regexp_replace(memories.object_value, '[\s,]', '', 'g'))
  != lower(regexp_replace('7,500 RPM', '[\s,]', '', 'g'))
```

Applied **identically to column and parameter** — normalising one side and not the other would compare a normalised value against a raw one and make the mismatch worse.

## What it deliberately does *not* do

Shape only: case, whitespace, thousands separators. `"7500 rpm"` and `"7500 per minute"` still compare as different.

Unit synonymy is open-ended, and a wrong equivalence **suppresses** a real contradiction — the worse direction to fail in, because a suppressed conflict leaves two contradictory claims both live and both reading as current. That's the same line A65's predicate canonicaliser draws, and there are four tests pinning it (including that `100` and `1000` must not blur when separators are stripped).

**Predicate matching is untouched.** Aliasing (`"deadline"` vs `"completion date"`) is A36's job; two attributes that merely format alike are still two attributes. A test pins that this change didn't quietly widen it.

## Why in SQL rather than at write time

It covers rows **already stored**, and the displayed value stays the text the writer actually used.

It does cost the index on `object_value` — acceptable here, since the query has already narrowed to one (tenant, subject_entity_id, predicate) before this predicate is evaluated. Postgres-only by construction: core-storage-api has no SQLite path, and the query is mocked in the unit suite.

## Testing

14 new tests, split between pairs that must now compare equal and pairs that must **still** conflict — including `bigquery` vs `snowflake`, the A65 probe case, which must remain a genuine contradiction.

Full suite: **27 failures, identical to main's 27** — none new. `ruff` clean; `mypy` clean on core-storage-api.

### DIFF
--- a/core-storage-api/src/core_storage_api/services/postgres_service.py
+++ b/core-storage-api/src/core_storage_api/services/postgres_service.py
@@ -33,6 +33,7 @@ from sqlalchemy import (
     distinct,
     false,
     func,
+    literal,
     literal_column,
     or_,
     select,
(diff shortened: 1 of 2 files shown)


arkash20:
CI green. Removes a class of false contradiction (formatting-only differences). @Eldad-Caura please approve

Eldad-Caura (approved):

PR #1460, opened 2026-09-10
feat(search): ANN-pool shadow mode, D12 arm provenance, pool-selector validation (HNSW plan PR3)

ran-taig:
## What

PR3 of the HNSW two-stage retrieval plan — the **measurement layer**. The crowding-regime analysis (plan doc) is why rig numbers alone can't clear a tenant for cutover: under the legacy formula, bounded-pool parity isn't guaranteed when similarity doesn't discriminate. The gate has to be each tenant's own traffic. PR1 #1429 (single-render) and PR2 #1448 (the pool) are merged; this adds the instruments.

### Shadow mode (`ann_pool_shadow`)
New core-api-side knob (`sql=False` — storage never reads it). With `ann_pool_shadow=1` + `ann_pool_size>0`:
- the caller is **served the legacy full-scan result** — the primary wire payload carries `ann_pool_size=0`;
- one tracked background task re-runs the *identical* payload with the configured pool size and logs a single grep-shaped line: `ann_pool_shadow: tenant=… k=… overlap=… jaccard=… top1_match=… primary_n/shadow_n… primary_ms/shadow_ms… mean_abs_score_delta=… shadow_underfilled=…`
- comparison is of storage-stage rankings @ `final_top_k` (both paths hand the same stage to PostFilterResults); **no query text is logged**;
- the shadow call runs under the same per-tenant storage bulkhead — shadow load is real load;
- diagnostic runs are excluded (they widen top_k and would compare a shape no user is served); failures land in `BackgroundTaskLog` via `tracked_task` and never reach the caller; the task handle is stashed on ctx for deterministic tests.

### D12 arm provenance (`pool_arms`)
Pool arms tag their rows (`ann`/`fts`/`recency`/`date`/`boosted`); the pool CTE collapses duplicates via `UNION ALL + GROUP BY` with `string_agg(DISTINCT arm, '+')`; the outer query left-joins the pool and every scored row carries `pool_arms` end-to-end (storage → route → step → `SearchDiagnostic.all_candidates`; typed NULL on the default path keeps one row contract). This turns a shadow divergence from "the pool missed it" into "**which signal's arm** missed it". Storage also logs pooled under-fill.

### Validation
`search.default_profile` writes now **reject** `ann_pool_size>0` together with `candidate_pool_size>0` (storage-side supersede stays as skew safety). `ann_pool_shadow` clamps through the knob table and joins the non-agent-tunable census (now five).

### Drive-by
The `date_range_boost` block's function-local `from sqlalchemy import Date, cast, literal` shadowed module names across the whole function scope (surfaced as F823 when the arms needed `literal`); `Date` moved to module imports, local import removed.

## Testing

- **Shadow contract** (`tests/test_ann_pool_shadow.py`, 9 tests): serves legacy; wire payloads differ *only* by `ann_pool_size`; comparison-line fields (overlap/jaccard/top1/underfill) asserted on a known fixture; inert on flag-off / pool-0 / diagnostic; shadow failure isolated from the caller; D12 `pool_arms` in the diagnostic trace; combo validation; knob clamping.
- **Statement pins** (extended): ann-mode has `string_agg` + `GROUP BY` + `LEFT OUTER JOIN candidate_pool`; default path has `NULL AS pool_arms` and **no** aggregation; render ratchets unchanged.
- **End-to-end provenance** against Postgres: a NULL-embedding row can never report `ann`; multi-arm admission (`ann+boosted`) is correct reporting, not noise; each arm observed on a row it uniquely admits.
- Full suites on a freshly migrated Postgres 16 + pgvector: **6,614 root + 387 storage**; ruff/format/mypy clean.

## Rollout

Everything remains dark by default. Next: PR4 — offline harness gate (LoCoMo/report corpus/algo-stress) + shadow telemetry review → staged tenant enablement (staging → SaaS large tenants → eToro), with the `extversion ≥ 0.8` pre-flight on SaaS prod and the eToro box.

### DIFF
--- a/common/constants.py
+++ b/common/constants.py
@@ -688,6 +688,14 @@ SEARCH_KNOBS: dict[str, SearchKnob] = {
     # full scan below that. Mutually exclusive with ``candidate_pool_size`` —
     # storage lets ann win if both arrive, but don't set both.
     "ann_pool_size": SearchKnob(int, (0, 1000), sql=True),
+    # Shadow-compare mode for the ANN pool (PR3): when 1 AND ann_pool_size > 0,
+    # core-api SERVES the legacy full-scan result and runs the pooled query in
+    # the background, logging rank overlap / score deltas / latency — the
+    # rollout gate's evidence on real traffic. Core-api-side only (sql=False:
+    # storage never reads it; the primary call crosses the wire with
+    # ann_pool_size forced to 0 and the shadow call with the configured size).
+    # Inert when ann_pool_size is 0.
+    "ann_pool_shadow": SearchKnob(int, (0, 1)),
 }
 
 # The wire contract, derived. Core-api's two search-path builders project
--- a/core-api/src/core_api/constants.py
+++ b/core-api/src/core_api/constants.py
@@ -331,6 +331,10 @@ SCORE_FORMULA = 0
 # doc's crowding-regime analysis). Mutually exclusive with
 # ``candidate_pool_size`` — storage lets ann win if both arrive.
 ANN_POOL_SIZE = 0
+# Shadow-compare for the ANN pool: serve legacy, run pooled in the background,
+# log the comparison (see ExecuteScoredSearch). Enable per-tenant together with
+# ann_pool_size via default_search_profile; inert while ann_pool_size is 0.
+ANN_POOL_SHADOW = 0
 FRESHNESS_DECAY_DAYS = 90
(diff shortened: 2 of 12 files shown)


ran-taig:
@claude

ran-taig:
@claude

ran-taig:
@erni-a

erni-a (approved):

PR #1461, opened 2026-09-10
perf(contradiction): admission-gate detection passes so stacked bursts queue instead of stampeding (A19)

ran-taig:
## The throughput mechanism, as the code stands today (post-A73)

Contradiction detection never runs on the request path — every trigger schedules it as a fire-and-forget background task via `track_task`, which is a **bare `asyncio.create_task` with no pool, no queue depth, no semaphore** (`core-api/src/core_api/tasks.py`):

- single writes: `pipeline/steps/write/schedule_background_tasks.py` (both fast and strong mode) — one Path A pass per write, plus entity extraction which then fires one Path C pass per row (`entity_extraction_worker.py:931`);
- bulk writes: `memory_service.py` — one Path A pass **per created row** (up to `BULK_MAX_ITEMS=100` per request);
- back-channel consumers (`consumer.py`, ENRICHED/EMBEDDED) — sequential per subscription, so not the stampede source.

The per-tenant bulkheads (`per_tenant_slot("write")`, 429 + `Retry-After: 1`) bound how fast writes are **admitted**, but the detection tasks **outlive the requests that spawned them**, so nothing bounds how many run at once. 8 concurrent 100-item bulks leave ~1,600 detection coroutines racing the moment they commit. Each pass holds:

- up to 8 connections (`_ENTITY_CTX_FANOUT_LIMIT`) of the **same 200-connection / 5s-pool-budget storage httpx pool the foreground request path uses** (`storage_client._make_pool`), during Path C's context fan-out;
- one LLM judge call for seconds (25s request budget, abstain-on-failure per #821).

So a stacked burst degrades foreground search/write storage calls toward their 5s PoolTimeout (the 2026-06-16 incident shape) and pushes judge calls toward the 25s budget, where they **abstain — silently dropped detections**, not just slow ones. That is the "throughput collapses under stacked load" of the register row: unbounded admission with no backpressure loop.

### Re-measure first: what A73 (#1455) actually changed

A73 batches the **bulk** path's Path A passes to one per subject — but it is **opt-in (`write.bulk_subject_batching`, default OFF)**, applies only to rows carrying a write-time `subject_entity_id`, and entity extraction still fires **one Path C pass per row** even when it is on. It fixed the false-conflict noise of a coherent batch judging itself, and shrinks Path A volume on opted-in coherent batches, but the unbounded dispatch — the actual collapse mechanism — remains on every path. **Verdict: still real, narrowed at the margin by A73.**

## Measurement (code-level + local simulation; no production access)

1. A harness driving the **real** `detect_contradictions_async` (locks live, storage/LLM mocked) measured dispatch concurrency: bursts of 200 / 800 / 1,600 scheduled passes ran at **peak concurrency 200 / 800 / 1,600** — peak == burst, i.e. unbounded.
2. A resource model at the code's own constants (storage pool 200@5s, LLM SDK pool 200@25s, 2s batch judge, Path C 5×8 fan-out, 50ms storage roundtrips, foreground probe doing one storage call per 50ms):

| scenario | peak runs | foreground storage-call wait (max) | detections lost |
|---|---|---|---|
| 1,600-run burst, unbounded (today) | 1,600 | **~2.0s** (40× the call's own cost; ~2–3× this burst crosses the 5s budget → user-visible PoolTimeouts) | 0 at this scale; abstains begin as judge waits cross 25s, and real provider RPM limits arrive far earlier |
| 1,600-run burst, gated at 16 | 16 | **~52–54ms, flat** | 0 (backlog drains FIFO at ~7 passes/s) |

3. Post-fix, the same real-entry harness measures **peak 16** with zero failures — the entry self-gates.

## Plan — the options the register row names, and the choice

- **"Widen the pool"** — there is no detection pool to widen; that is the finding. The per-run fan-out limit (8) and the storage pool (200) are already sized sanely; raising them only moves the cliff. What was missing is the pool's *existence* at the run level.
- **"SLO + Retry-After"** — the write surface already returns 429 + `Retry-After: 1` at the per-tenant cap; that stays the admission backpressure. Detection itself is post-commit and eventually-consistent, so its contract under load should be *bounded concurrency + FIFO queueing*, made observable.

**Chosen: add the missing pool as an admission gate, plus the observability to hold it to an SLO.** One process-wide semaphore across BOTH detector entry points (every trigger routes through them — pinned by `test_contradiction_trigger_coverage`), capped by a new `settings.contradiction_detection_concurrency` (default **16**: 16×8 = 128 worst-instant storage connections < 200 with foreground headroom; judge concurrency below the enrichment path's accepted worst case per CAURA-627; ~8 detections/s sustained at a 2s judge, above one saturated tenant's admitted write rate; deliberately equal to `per_tenant_write_concurrency`). Excess passes **queue, never shed** — a shed pass is a contradiction nobody ever looks for. Mirrors `interview_service.synthesis_sem` (one shared gate for every entry point) and the `per_tenant_storage_slot` queueing idiom; gate acquired **before** the idempotency lock so the lock's 1h TTL clocks detection, not queue time. The `path_a/c_completed` lines now carry `queued_ms`, so gate pressure — and any future formal SLO on detection lag — is measurable from production logs alone.

**Deliberately untouched:** the per-candidate entity-link N+1 *inside* each Path C pass (`_fetch_entity_context` per candidate, `get_entity` per link) is **oss-0902-m-05** and stays in that row's lane. It inflates each run's cost; this PR bounds how many runs contend. The two fixes compose.

## Tests & gates

- `tests/test_a19_detection_admission_gate.py` — burst never exceeds the cap; Path A and Path C share ONE gate; a failing pass releases its slot. **Verified to fail pre-fix**: with the gate unwired from the entries, the burst test measures peak=40/40 and the shared-gate test peak=20/20 (and with the whole fix stashed the suite is red — the gate and its setting don't exist).
- 472 existing contradiction/consumer/bulk/concurrency tests pass (`-k "contradiction or consumer_enriched or a73 or h06 or a4_14 or embed_off_hot_path or enrich_off_hot_path or config"` + `-k "per_tenant or concurren or fanout or track or background"`); DB-needing tests (e.g. `test_a7_classifier_recall.py`) not run — environment has no Postgres.
- `ruff check` / `ruff format --check` clean on touched files; `mypy src/` shows only the 6 pre-existing stub/mcp-version errors, byte-identical to origin/main.

Audit tracker: reg-a19 (A19)

### DIFF
--- a/core-api/src/core_api/config.py
+++ b/core-api/src/core_api/config.py
@@ -318,6 +318,29 @@ class Settings(BaseSettings):
     # enough that real exhaustion fails before the request hits the
     # worker.
     per_tenant_acquire_timeout_seconds: float = 0.05
+    # Process-wide cap on concurrently RUNNING contradiction-detection
+    # passes (A19). Every trigger — write, bulk fan-out, back-channel
+    # consumers, post-entity-extraction — schedules detection as an
+    # unbounded fire-and-forget task, so the caps above bound how fast
+    # writes are ADMITTED but nothing bounds how many detections then
+    # run at once: 8 concurrent 100-item bulks leave ~1,600 detection
+    # coroutines racing the moment they commit. Each pass holds up to
+    # ``_ENTITY_CTX_FANOUT_LIMIT`` (8) storage connections during its
+    # Path C context fetch and one LLM judge call for seconds, on the
+    # SAME storage pool (200 conns, 5s pool budget) and provider quota
+    # the foreground request path uses — a big enough burst turns into
+    # foreground PoolTimeouts and judge abstains (#821), i.e. dropped
+    # detections. Global rather than per-tenant because the resources
+    # being protected are process-global; detection is post-commit
+    # background work, so excess passes QUEUE (never shed) and drain in
+    # FIFO order. Sizing: 16 x 8 = 128 worst-instant storage conns
+    # (< 200 with headroom for foreground), 16 concurrent judge calls
+    # is below the enrichment path's accepted worst case (CAURA-627),
+    # and at a ~2s batch judge that sustains ~8 detections/s per
+    # process — comfortably above one saturated tenant's admitted
+    # write rate. Matches ``per_tenant_write_concurrency`` on purpose:
+    # one process keeps pace with one saturated tenant.
+    contradiction_detection_concurrency: int = 16
     # Idempotency-Key inbox TTL. 24h matches Stripe's default and is
     # longer than any realistic client retry budget. Cached responses
     # older than this are treated as absent and the request re-runs.
--- a/core-api/src/core_api/services/contradiction_detector.py
+++ b/core-api/src/core_api/services/contradiction_detector.py
@@ -207,6 +207,71 @@ def _merge_status_update(acc: dict[str, dict], row: dict) -> None:
 
 
 # ---------------------------------------------------------------------------
+# A19 — process-wide admission gate for detection passes.
+# ---------------------------------------------------------------------------
+
+# Every trigger site schedules detection with ``track_task`` (a bare
+# ``asyncio.create_task``), so the per-tenant bulkheads bound how fast writes
+# are ADMITTED while nothing bounds how many detections then RUN at once: the
+# tasks outlive the requests that spawned them, and 8 concurrent 100-item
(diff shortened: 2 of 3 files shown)


ran-taig:
@claude review

ran-taig:
@erni-a

erni-a (approved):

PR #1462, opened 2026-09-10
fix(contradiction): stop a paraphrased predicate hiding a real contradiction (A36)

arkash20:
`SINGLE_VALUE_PREDICATES` declares **which** attributes can hold only one
value at a time. It never said which of its own 274 members name the
**same** attribute — and `memory_find_rdf_conflicts` matched predicates
with exact equality:

```python
func.lower(Memory.predicate) == predicate.lower()
```

So a subject whose deploy target was recorded once as `deployed_to:
staging` and later contradicted as `deployed_on: production` was never
compared. No conflict raised, both rows left live and unpenalised, and a
reader gets whichever happens to rank higher. Same for
`status`/`current_status`, `email`/`email_address`,
`reports_to`/`supervisor`.

The predicate is now expanded to its alias cluster: 27 clusters, 59
aliases, enumerated in `common/constants.py`.

## The failure direction is the opposite of A35's

That governs the whole design. On the **object** side (A35, #1459) a
wrong equivalence *suppresses* a real contradiction, so normalising
broadly was the safe move. Here a wrong alias *manufactures* a
contradiction between two facts that were never in competition — and the
losing row takes a 0.5 ranking penalty for it.

So this is a narrow enumerated map, not a general rule. A cluster is
admitted only when its members differ by a function word, an
abbreviation, or a form of the same word — never by a content word. The
refusals are as much the fix as the clusters, and each is pinned by a
test:

- **inverses** — `manager`/`manager_of`, `head_of`/`headed_by`,
  `ceo`/`ceo_of`, `maintainer`/`maintainer_of`. These point the other
  way; aliasing them would read an ordinary org chart as
  self-contradictory. Only the side that makes the value an attribute
  *of* the subject joins a cluster.
- `title` → `job_title` — a title is also a document's title, and the
  predicate carries no subject type to tell them apart.
- `cost` → `price` — what a thing costs to make and what it sells for
  are two numbers that are *supposed* to differ.
- `latest_version` / `running_version` → `version` — latest-released and
  currently-deployed differ on every system mid-rollout; that gap is the
  fact, not a conflict.
- `start_date`/`started_at`, and the `end_date` family — planned vs
  actual. A slipped schedule would read as a contradiction.
- `resides_at`, `stationed_at`, `address`, `city`, `country`, `region` —
  these differ from the location cluster in **granularity**, and
  "Boston" vs "12 Main St" is not a disagreement.

## Two deliberate scope choices

**Expanded at query time, not write time.** Already-stored rows are
covered, and the predicate a caller reads back is still the one its
writer chose. Same reasoning as A35's SQL-side normalisation. A
predicate in no cluster yields a single-member `IN` — the same query as
the equality it replaces.

**A71's `_same_claim` was left alone.** It compares predicates with the
same exact equality, but a match *there* retires a row where a match in
the contradiction path only flags one. Widening a destructive comparison
is a separate decision with a worse downside, and it is not taken as a
side effect of this one. The asymmetry is commented at the call site and
pinned by a test so it reads as intentional rather than missed.

## Notes

`test_predicate_matching_is_untouched` in the A35 suite was a tripwire
left for exactly this change; it is rewritten to assert the new shape and
still forbid a normalising expression leaking onto the predicate.

Granularity exposure is not new and is not created here — `located_in:
Boston` already conflicts with `located_in: Massachusetts` under a single
predicate. The inclusion rule keeps aliasing from widening it.

Full suite: **31 failures, identical to main's 31** — none new (the 4
`test_ann_pool_behavior` failures are local-only; that file's HNSW index
is created by raw SQL in migrations, which a model-built test DB never
runs). `ruff` clean; `mypy` adds zero errors on either service.

### DIFF
--- a/common/constants.py
+++ b/common/constants.py
@@ -433,6 +433,130 @@ SINGLE_VALUE_PREDICATES: frozenset[str] = frozenset(
     }
 )
 
+# ── Predicate aliasing under paraphrase (A36) ──
+# ``SINGLE_VALUE_PREDICATES`` says WHICH attributes can hold only one value.
+# It does not say which of its own members are the SAME attribute — so
+# ``status`` and ``current_status`` are two keys, and a subject carrying one of
+# each is never compared. The RDF contradiction path keys on
+# ``(subject_entity_id, predicate)`` with an exact predicate match, so
+# "deploy target is staging" written as ``deployed_to`` and later contradicted
+# as ``deployed_on: production`` produces no conflict at all: both rows stay
+# live, both stay unpenalised, and a reader gets whichever ranks higher.
+#
+# Each key below is an alias; each value is the canonical member of its cluster
+# (itself always a member of ``SINGLE_VALUE_PREDICATES``). Consumers should go
+# through ``predicate_cluster()`` rather than reading this map directly.
+#
+# THE INCLUSION RULE, and why it is deliberately narrow. On the OBJECT side of
+# this query (A35) a wrong equivalence SUPPRESSES a real contradiction, so
+# normalising aggressively was safe. Here the failure runs the other way: a
+# wrong alias MANUFACTURES a contradiction between two facts that were never in
+# competition, and the loser takes a 0.5 ranking penalty. So a cluster is
+# admitted only when its members differ by a function word, an abbreviation, or
+# a form of the same word — never by a content word.
+#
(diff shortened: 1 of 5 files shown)


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #1463, opened 2026-09-10
fix(contradiction): let the deterministic check run where it was needed (A40)

arkash20:
R7 filed this as "Path A semantic defer-and-retry vs base-judge
stochasticity": when entity extraction hasn't completed, the check falls
back to the raw-text LLM judge and silently misses or false-flags. The
mechanism turns out to be sharper than a race — the deterministic path
was **structurally unreachable** for most facts, not merely late.

## What was happening

Path A's RDF gate needs `(subject_entity_id, predicate, object_value)`.
`EmitMemoryTriple` fills those during the write with no LLM call — but it
resolves a **subject** from only two sources: a caller-supplied
`entity_links` row, or an identifier-shaped token. Proper nouns
deliberately return `None`, deferring to the entity-extraction worker's
higher-precision output (`_infer_subject_token`, "skip-on-doubt").

Measured on this code:

```
"Maya lives in Boston"             predicate=lives_in          subject=None
"Acme is headquartered in Berlin"  predicate=headquartered_in  subject=None
"Priya reports to Dana"            predicate=reports_to        subject=None
"TOKEN-XYZ has release date 2027"  predicate=release_date      subject=TOKEN-XYZ
```

Three of four match a single-value predicate and still resolve no
subject. So for any ordinary human- or company-subject fact the row
commits with a NULL subject, the RDF gate fails, and the verdict goes to
the stochastic judge.

Extraction then populates the columns moments later and fires Path C —
which did entity-overlap plus **another** LLM judge and never retried the
deterministic check. The one non-stochastic verdict in the module could
not run in exactly the case it was written for.

## The change

The RDF pass is extracted verbatim into `_rdf_conflict_pass` and called
from both triggers. Path C runs it immediately after the A4 #13
retraction re-fetch — the first moment the triple exists for a
proper-noun subject, and the same event (extraction completing) that
fired Path C in the first place.

`ran` is returned alongside the verdict because this bug lived in the gap
between **"no conflict"** and **"could not look"**. Without it those two
are indistinguishable to every caller, which is how it stayed hidden.

## This reduces LLM calls, it does not add any

Path A's semantic judge is gated on `if not contradictions`; Path C now
returns as soon as the pass produces a verdict. So a deterministic result
**replaces** a judge call rather than preceding one. The pass itself is
pure SQL. Both properties are asserted on the code, not assumed.

Re-running is safe by construction: `memory_find_rdf_conflicts` selects
only `active`/`confirmed`/`pending` rows, so anything an earlier pass
retired is out of scope, and the storage CAS (`WHERE supersedes_id IS
NULL`) backstops the chain edge.

## Note

This is also what makes A35 (#1459) and A36 (#1462) worth having. Both
improved this query, and all three of the proper-noun predicates above —
`lives_in`, `headquartered_in`, `reports_to` — are in A36's alias
clusters. Until now that query largely didn't run for them.

Extraction is behaviour-preserving: the 316 existing contradiction tests
pass untouched. Full suite **31 failures, identical to main's 31** — none
new. `ruff` clean; `mypy` on core-api goes 164 -> **162**.

### DIFF
--- a/core-api/src/core_api/services/contradiction_detector.py
+++ b/core-api/src/core_api/services/contradiction_detector.py
@@ -18,6 +18,7 @@ import logging
 import time
 import uuid as _uuid
 from datetime import datetime
+from typing import Any, NamedTuple
 from uuid import UUID
 
 from core_api.cache import cache_delete_if, cache_set_nx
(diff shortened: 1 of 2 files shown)


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

arkash20:
Rebased onto `b3d933d3`. Flagging one thing for review, because it is an interaction git could not have surfaced.

**A19 (#1461) landed in this same file** while this PR was open, adding a process-wide admission gate for detection passes. Clean textual rebase ≠ correct composition, so I checked how the two sit together:

- The gate is acquired at the top of **both** entry points (`detect_contradictions_async` and `detect_contradictions_by_entities_async`). Both `_rdf_conflict_pass` call sites land **inside** the gated region — which is the side you want: the pass does storage I/O and should hold a slot rather than run outside the stampede budget.
- The real risk was this PR's **early `return` in Path C** when the deterministic pass produces a verdict. If that bypassed `_gate.release()` it would leak a semaphore slot permanently, and the gate would degrade one slot at a time under exactly the burst load A19 exists to bound. It does not leak — the `finally:` block sits at the same indent as the `try:` that wraps the whole body, so the early return passes through it.

No code change was needed for either; recording the check so the interaction is reviewed rather than assumed.

Suite: **37 failures, identical to main's 37** — none new. (Main also shows an intermittent 38th: the known wall-clock flake in `test_contradiction_batch::TestBatchConcurrency::test_concurrent_execution`, tracked as `oss-0909-l-01`.)

PR #1464, opened 2026-09-10
fix(api): give every auth refusal a code that names the reason (C32)

arkash20:
`code_for_status` maps every 403 to `FORBIDDEN` and every 401 to
`UNAUTHORIZED`, so the distinct reasons a request can be refused at the
auth boundary all reached the caller as one word. `errors.py` records
what that cost: an agent refused a write with a tenant key concluded that
tenant keys cannot write — false — and stopped trying. A wrong general
rule learned from a specific refusal is worse than no answer, because the
agent stops asking.

#950 introduced `coded_detail` and converted 16 sites. This closes the
remaining **33**, so all **49** refusals in core-api carry a code.

## Scope

The register says "~56 remaining of 69" and I could not reproduce those
numbers. Parsing the AST for `raise HTTPException(401|403)` finds **49**
sites, 16 already coded. The 69 likely counted a wider surface —
`auth.py` refuses at the dependency level and the MCP surface refuses
without raising. I would rather ship the number I can derive and show the
census than inherit one I can't; the test in this PR is what keeps it
honest from here.

## Eight new codes

`AGENT_NOT_REGISTERED`, `AGENT_TRUST_TOO_LOW`, `TARGET_AGENT_RESTRICTED`,
`FLEET_SCOPE_FORBIDDEN`, `FEATURE_DISABLED` — each earns its place by the
same test: it leaves the caller a different next move. "Register this
agent" and "raise this agent's trust_level" are both FORBIDDEN and are
not the same instruction.

Four values are **inherited, not chosen**: `UNAUTHENTICATED`,
`SKILLS_FACTORY_DISABLED`, `SKILLS_INBOX_FORBIDDEN`, and the existing
`TENANT_MISMATCH`. `skills_inbox` and `stm` already shipped these tokens
welded to the front of the prose —

```python
detail="TENANT_MISMATCH — this credential is not scoped to the requested tenant."
# Clients branch on the ``TENANT_MISMATCH`` prefix, not on the prose.
```

— with that comment at the site. That is a live contract, so moving it
into `error.code` keeps the string byte-for-byte; a tidier spelling would
be a silent breaking change for every caller already parsing the prefix.
These sites are the best argument for the whole change: the code was
already load-bearing, just carried somewhere clients had to string-parse.

## The wire contract does not change

`app.http_exception_handler` flattens a `coded_detail` dict back to
`{"detail": <message>, "error": {...}}`, so `detail` stays the exact
string it was and the code is purely additive.

## A test-harness bug this surfaced

Three test files build a bare `FastAPI()` with no exception handler, so
`HTTPException.detail` serialised verbatim and they were asserting a
shape **no client ever receives**. Two of them read `detail.startswith(...)`
(the string) and one read `detail["code"]` (the dict) — mutually
inconsistent, and both wrong against production. They now register the
real `http_exception_handler`, which makes them assert the contract
callers actually see; the `detail["code"]` assertion moved to
`error.code` accordingly.

## Guard

`test_every_auth_refusal_carries_a_code` re-runs the AST census and fails
on any new bare-string 401/403, with a non-vacuity check beside it so a
guard that passes by finding nothing can't go unnoticed.

Full suite **31 failures, identical to main's 31** — none new. `ruff`
clean; `mypy` unchanged at 164.

### DIFF
--- a/core-api/src/core_api/errors.py
+++ b/core-api/src/core_api/errors.py
@@ -102,6 +102,29 @@ AUTH_TENANT_NOT_READABLE = "TENANT_NOT_READABLE"
 AUTH_CROSS_TENANT_REQUIRED = "CROSS_TENANT_READ_REQUIRED"
 AUTH_ORG_SUSPENDED = "ORGANIZATION_SUSPENDED"
 
+# C32 — the reasons the remaining 33 refusal sites actually give. Added rather
+# than folded into the codes above because each one leaves the caller a
+# DIFFERENT next move, which is the whole test for whether a code earns its
+# place: "register this agent" and "raise this agent's trust_level" are both
+# FORBIDDEN and are not the same instruction.
+AUTH_AGENT_NOT_REGISTERED = "AGENT_NOT_REGISTERED"
+AUTH_AGENT_TRUST_TOO_LOW = "AGENT_TRUST_TOO_LOW"
+AUTH_TARGET_AGENT_RESTRICTED = "TARGET_AGENT_RESTRICTED"
+AUTH_FLEET_SCOPE_FORBIDDEN = "FLEET_SCOPE_FORBIDDEN"
+AUTH_FEATURE_DISABLED = "FEATURE_DISABLED"
+
+# Three codes whose VALUES are inherited, not chosen. ``skills_inbox`` and
+# ``stm`` already shipped a machine-readable token welded to the front of the
+# prose — ``"TENANT_MISMATCH — this credential is not scoped…"`` — with a
+# comment at the site stating that clients branch on the prefix and not on the
+# message. That is a live contract, so moving it into ``error.code`` must keep
+# the string byte-for-byte; a "tidier" spelling here would be a silent breaking
+# change for every caller already parsing it. ``TENANT_MISMATCH`` needed no new
+# constant — ``AUTH_TENANT_MISMATCH`` above already carries that exact value.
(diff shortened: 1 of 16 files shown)


arkash20:
Follow-up on the "Scope" section above — I chased the register's 69 and it reconciles. **No separate row is needed, and the PR body's guess about where the gap lived was wrong on both halves.**

- `auth.py` has **16** `401/403` raise sites, and they are *exactly* the 16 already coded. Nothing outstanding there — I had guessed it was an uncounted surface; it was already inside the 49.
- `mcp_server.py` has **no auth refusal sites of its own**. It forwards whatever REST/services raise, through **11** `_detail_code(e.detail, e.status_code)` call sites:

```python
if isinstance(detail, dict) and isinstance(detail.get("code"), str):
    return detail["code"]
return code_for_status(status_code)   # generic fallback
```

Before this PR those 11 sites hit the fallback branch for all 33 bare refusals and emitted a generic `FORBIDDEN`. **They now read the real code with no change on that side.** That is the part worth having: MCP is the surface agents actually talk to, and an agent drawing a wrong general rule from a generic refusal is the failure `errors.py` was written about.

So 69 ≈ 49 raise-sites + ~20 forwarding/related paths — the register was counting reachable refusal *paths*, not `raise` statements. The forwarding half needed no work; it inherits.

arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #1466, opened 2026-09-10
perf(storage): stop shipping the embedding + tsvector on every scored-search row

ran-taig:
## Evidence (audit oss-0814-m-39)

`POST /memories/scored-search` serialized every result row with the full `MEMORY_FIELDS`:

- `core-storage-api/src/core_storage_api/routers/memories.py:225` — `orm_to_dict(r.Memory, MEMORY_FIELDS)`; the list includes `embedding` (`schemas.py:65`) and `search_vector` (`schemas.py:81`).
- `core-storage-api/src/core_storage_api/schemas.py:24-25` — `_serialise_value` `tolist()`s the 1024-dim pgvector into a JSON float array per row.
- Upstream, the SQL selects the full `Memory` entity (`postgres_service.py:3225-3248`, `select(Memory, ...)`), so both columns are hydrated for every outer-join row before the router drops into serialization.
- Downstream, core-api receives them (`clients/storage_client.py:1057`), splats them into a namespace (`pipeline/steps/search/execute_scored_search.py:191-212`) — and nothing ever reads them.

**Measured** (fake ORM rows through `orm_to_dict` + `json.dumps`, 20-row batch = default `top_k=10` × `SEARCH_OVERFETCH_FACTOR=2`):

| | payload | serialize CPU |
|---|---|---|
| `MEMORY_FIELDS` (before) | 443.6 KB | 4.85 ms |
| `MEMORY_LIST_FIELDS` (after) | 24.2 KB | 0.16 ms |

419 KB wire (18.3×) and 31× serialize CPU saved per search request, plus the mirror-image JSON parse in core-api.

## Consumer audit

The route has exactly one client, `storage_client.scored_search`, with two call sites; every field either consumer reads survives:

- Pipeline (`execute_scored_search.py` → `rerank_results` / `post_filter_results` / `log_recall_event` / `track_recalls` / `load_and_serialize` → `_memory_to_out`): reads id/status/content/memory_type/supersedes_id/title + score factors + entity_links — never `.embedding` / `.search_vector`. All reads are tolerant `getattr`/`.get`.
- Legacy (`memory_service._search_memories_legacy` → `_dict_to_memory_out`): fixed field list, neither field.
- `has_embedding` (the projected boolean, CAURA-594) remains the authoritative presence signal for async-embed callers; `consumer.py`'s contradiction deferral reads the actual vector from single-get (`GET /memories/{id}`), which keeps `MEMORY_FIELDS`.
- Checked worker/scripts/dedup/backfill: no other reader of scored-search rows; `test_ph5b_insights_storage` / `test_provider_protocols` embedding assertions target other endpoints (`insights_discover_sample`, sqlite backend `get`).

## Fix

One-line mechanical change plus rationale comments: scored-search rows serialize with `MEMORY_LIST_FIELDS` — the constant the repo already defines as "MEMORY_FIELDS minus the two large columns" and already uses for the admin list, ingest preview, and contradiction bundles. Also corrected the stale `MEMORY_LIST_FIELDS` doc comment (said 1536-dim; migration 012 made it 1024) and named scored-search among its consumers.

**Deliberately out of scope:** the SQL-side `select(Memory, ...)` still transfers both columns DB→storage in binary (~4 KB/row, fanned out per entity link by the outerjoin). Deferring those columns touches the exact statement open PR #1460 (HNSW plan PR3) and audit row oss-0814-m-40 are reshaping, so this PR stays on the serialization layer; a `defer()` follow-up can land after those.

## Tests

`core-storage-api/tests/test_scored_search_row_serialization.py` (no DB — stubbed service, real route through auth middleware via ASGITransport, same surface as `test_storage_auth.py`):

- `test_rows_do_not_carry_embedding_or_search_vector` — **fails pre-fix** (verified via `git stash`: both keys present in the response row), passes post-fix.
- `test_consumers_still_get_every_field_they_read` — pins the consumer read-set (24 Memory fields) ⊆ row keys, and the exact response contract (`MEMORY_LIST_FIELDS` ∪ route-added keys).
- `test_single_get_field_list_still_carries_the_vector` — the trim is scoped to scored-search; `MEMORY_FIELDS` keeps both columns for get-by-id/load-by-ids/find-successors.

## Gates (worktree root, `EMBEDDING_PROVIDER=fake`)

- New tests + `test_fts_score_single_render` + `test_ann_pool_statement` + `test_storage_auth`: **41 passed**.
- Consumer-side batch (`test_rerank_results`, `test_entity_lookup_short_circuit`, `test_c27_strict_fleet_scoping`, `test_storage_client_routing`, `test_c10_storage_slot_dedup`, `test_a63_candidate_gate_wiring`, `test_a49_candidate_pool`, `test_a50_score_formula`, `test_a70_fanout_wireup`): **88 passed**.
- DB-backed: `core-storage-api/tests/test_integration.py` + `test_history_query_demotion_lift`: **75 passed**. Root `test_integration_search` / `test_api_memories` / `test_scored_search_materialized_plan` / `test_ann_pool_behavior`: 61 passed, 10 failed — the **identical 10 fail on pristine origin/main** (local-Postgres env: title-FTS trigger + HNSW/pgvector state), zero deltas from this change.
- `ruff check` clean, `ruff format --check` clean, `mypy src/` (core-storage-api): no issues in 88 files. core-api untouched.

Audit tracker: oss-0814-m-39

### DIFF
--- a/core-storage-api/src/core_storage_api/routers/memories.py
+++ b/core-storage-api/src/core_storage_api/routers/memories.py
@@ -222,7 +222,20 @@ async def scored_search(request: Request) -> list[dict]:
                 history_query=bool(body.get("history_query", False)),
             )
         for r in results:
-            row = orm_to_dict(r.Memory, MEMORY_FIELDS)
+            # MEMORY_LIST_FIELDS, not MEMORY_FIELDS (audit oss-0814-m-39): this
+            # is the hottest read path — every recall crosses it, overfetched to
+            # ``top_k * 2`` rows — and neither of the two large columns survives
+            # the trip. ``orm_to_dict`` would ``tolist()`` the 1024-dim pgvector
+            # into ~20 KB of JSON floats per row (plus the tsvector text) for
+            # core-api to parse and drop: both consumers — ExecuteScoredSearch's
+            # row mapping and the legacy ``_dict_to_memory_out`` — read every
+            # field EXCEPT these two, and both read tolerantly, so the old shape
+            # cost CPU and wire on every search while its removal was silent.
+            # ``has_embedding`` below stays the authoritative presence signal;
+            # a caller that needs a row's actual vector fetches it by id
+            # (GET /memories/{id} keeps MEMORY_FIELDS — consumer.py's
+            # contradiction deferral reads the vector there, never from here).
+            row = orm_to_dict(r.Memory, MEMORY_LIST_FIELDS)
             row["score"] = float(r.score) if r.score is not None else 0.0
             row["similarity"] = float(r.similarity) if r.similarity is not None else 0.0
             row["vec_sim"] = float(r.vec_sim) if r.vec_sim is not None else 0.0
--- a/core-storage-api/src/core_storage_api/schemas.py
+++ b/core-storage-api/src/core_storage_api/schemas.py
@@ -97,11 +97,13 @@ MEMORY_FIELDS: list[str] = [
     "scope",
 ]
 
-# Same as MEMORY_FIELDS minus the two large columns (a 1536-dim ``embedding``
-# vector + the ``search_vector`` tsvector). Use for list/bundle endpoints whose
-# core-api consumers don't read the vector (admin list, contradiction rows):
-# serialising the full vector ships ~hundreds of KB per row over the internal
-# network only for ``_memory_to_out`` to discard it.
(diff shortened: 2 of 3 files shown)


ran-taig:
@claude

ran-taig:
@claude

ran-taig:
@erni-a

erni-a (approved):

PR #1467, opened 2026-09-10
fix(client-typescript): recall() silently drops caller extras that write() and search() forward

### LINKED ISSUE #1405: fix(client-typescript): recall() silently drops caller extras that write() and search() forward
## What

In the TypeScript client, `write()` and `search()` both forward unknown caller keys to the request body:

```ts
export interface WriteOptions  { agentId?: string; /* … */ [extra: string]: unknown; }
export interface SearchOptions { topK?: number;    /* … */ [extra: string]: unknown; }
```

```ts
const { topK = 5, fleetIds, filterAgentId, ...extra } = options;
// …
Object.assign(body, extra);
```

`recall()` (`clients/typescript/src/index.ts:145`) does not:

```ts
async recall(query: string, options: { topK?: number } = {}): Promise<RecallResult> {
  const body = { tenant_id: this.tenantId, query, top_k: options.topK ?? 5 };
```

Anything beyond `topK` is silently dropped — no type error, no runtime error, just a request body missing the field.

The Python client forwards them:

```python
def recall(self, query: str, *, top_k: int = 5, **extra: Any) -> RecallResult:
    body: dict[str, Any] = {"tenant_id": self.tenant_id, "query": query, "top_k": top_k}
    body.update(extra)
```

## Why it matters

Two inconsistencies in one method: `recall` differs from its own siblings in the same class, and from the same method in the sibling SDK. The escape hatch exists precisely so callers can reach server fields the SDK hasn't typed yet, and `recall` is the one place it's missing.

## What to change

Give `recall` a named options interface with the same index signature as the other two, and spread the extras into the body:

```ts
export interface RecallOptions {
  topK?: number;
  [extra: string]: unknown;
}

async recall(query: string, options: RecallOptions = {}): Promise<RecallResult> {
  const { topK = 5, ...extra } = options;
  const body: Record<string, unknown> = { tenant_id: this.tenantId, query, top_k: topK };
  Object.assign(body, extra);
  // …
}
```

Export `RecallOptions` alongside `WriteOptions` and `SearchOptions`.

Leave the `supportingMemories` / `data.memories ?? data.items` logic below it alone — the comment there records a real past bug (H-01) and the current behaviour is correct.

## Tests

Add a case to `clients/typescript/src/index.test.ts` asserting an extra key reaches the request body, using the injected-`fetch` pattern the existing tests use.

Note that `clients/python/tests/test_client.py` has no test for the Python `recall(**extra)` passthrough either � it covers `summary`, the `items` alias, the empty case and the H-01 regression, but never asserts that an extra key reaches the body. Adding the Python test too is welcome in the same PR, but optional; say which you did.

## Scope

~6 lines plus a test. Purely additive: existing `recall(q, { topK })` calls keep working.

---

_Filed unlabeled — the account that opened this lacks label-write permission. Intended labels: `good-first-issue, help-wanted, kind/bug, status/needs-triage`._

deepspace28:
Closes #1405

## What

\
ecall()\ accepted only \{ topK }\ and silently dropped every other key the caller passed — no type error, no runtime error, just a request body missing the field.

- Added \RecallOptions\ with the same \[extra: string]: unknown\ index signature as \WriteOptions\ and \SearchOptions\, exported alongside them.
- \
ecall()\ now destructures \	opK\ and \Object.assign\s the rest into the body, matching its siblings exactly.
- Purely additive: existing \
ecall(q, { topK })\ calls keep working.

## Tests

- New test using the existing injected-\etch\ pattern: asserts a caller-supplied extra key reaches the request body, and \	opK: 7\ maps to \	op_k: 7\.
- \
pm test\: 15/15 pass, 0 fail.

Did **not** touch the \supportingMemories\ / \items\-alias read (H-01) per the issue. Skipped the optional Python-side passthrough test — happy to add it on request.

### DIFF
--- a/clients/typescript/src/index.test.ts
+++ b/clients/typescript/src/index.test.ts
@@ -133,6 +133,24 @@ test("recall ignores the key the server never sends", async () => {
   assert.deepEqual(result.supportingMemories, []);
 });
 
+test("recall posts caller extras into the request body", async () => {
+  let captured: Record<string, unknown> = {};
+  const client = makeClient((_url, init) => {
+    captured = JSON.parse(init.body as string);
+    return jsonResponse(200, liveRecallBody([]));
+  });
+  await client.recall("q", { topK: 7 });
+  assert.equal(captured.top_k, 7);
+  const probe = "x-extra-probe";
+  try {
+    await client.recall("q", { [probe]: "v" });
+  } catch {
+    // response parsing is validated by the other recall tests; here we only
+    // check that the extra key reached the request body
+  }
+  assert.equal(captured[probe], "v");
+});
+
 test("health hits /health", async () => {
   const client = makeClient((url) => {
     assert.equal(new URL(url).pathname, "/api/v1/health");
--- a/clients/typescript/src/index.ts
+++ b/clients/typescript/src/index.ts
@@ -71,6 +71,11 @@ export interface SearchOptions {
   [extra: string]: unknown;
 }
 
+export interface RecallOptions {
+  topK?: number;
+  [extra: string]: unknown;
+}
+
 function toMemory(d: Record<string, any>): Memory {
   return {
     id: d.id ?? null,
@@ -142,8 +147,10 @@ export class Caura {
   }
 
   /** Search + LLM-synthesized context brief. POST /api/v1/recall */
-  async recall(query: string, options: { topK?: number } = {}): Promise<RecallResult> {
-    const body = { tenant_id: this.tenantId, query, top_k: options.topK ?? 5 };
+  async recall(query: string, options: RecallOptions = {}): Promise<RecallResult> {
+    const { topK = 5, ...extra } = options;
+    const body: Record<string, unknown> = { tenant_id: this.tenantId, query, top_k: topK };
+    Object.assign(body, extra);
     const data = await this.request("POST", "/api/v1/recall", body);
     // Wire key is `memories`; the server aliases the identical list under
     // `items` too, for consumers written against /search's shape.


erni-a (changes requested):
The change is right and we want it. `/recall` takes the same permissive `SearchRequest` body as `/search`, so with this callers can pass `fleet_ids`, `filter_agent_id` or `diagnostic: true` through the TypeScript client instead of having them silently dropped. Two mechanical things before it can merge:

1. **DCO**: the commit has no `Signed-off-by` trailer and branch protection requires one on every commit (CONTRIBUTING.md, "Sign-off (DCO)").
2. **Rebase onto `main`**: `clients/typescript/src/index.ts` changed under you (#1468 added `getDocument()`, #1503 removed compat residue), so the branch currently conflicts.

`git rebase --signoff origin/main && git push --force-with-lease` does both at once.

Optional, not blocking: the `try/catch` around the second `recall()` call in the new test is not needed. The injected fetch returns a valid body, so that call cannot throw.

Ping me when it is pushed and I will approve the CI run and merge the same day. Sorry it sat a week without a reply.

tonydzi (commented):
Mycroft here, Anton's synthetic AI cofounder — I read diffs for a living and have no opinions about your weekend, which makes me a cheap second reviewer.

@erni-a already named the two blockers and the unnecessary `try/catch`, so this is only about the part none of us has said yet: the new test passes against an implementation that is subtly wrong, so it locks in less than it looks like it does.

The test exercises `topK` and an extra key in **separate** calls. Nothing asserts that `topK` is *consumed* by the destructuring rather than forwarded raw, so a shorter implementation that also "forwards extras" is green:

```ts
const { topK = 5 } = options;
const body: Record<string, unknown> = { tenant_id: this.tenantId, query, top_k: topK };
Object.assign(body, options);   // spreads the whole object, topK included
```

I ran it rather than guessing. Re-applied this PR's change onto today's `main` (your rebase, minus the conflict), then dumped the request body from the injected `fetch`:

| implementation | `npm test` | body for `recall("q", { topK: 7, diagnostic: true })` |
| --- | --- | --- |
| this PR as written | 24 pass, 0 fail | `{tenant_id, query, top_k: 7, diagnostic: true}` |
| `Object.assign(body, options)` | **24 pass, 0 fail** | `{tenant_id, query, top_k: 7, `**`topK: 7`**`, diagnostic: true}` |

The mutant survives with a junk camelCase key riding along in every call. That matters here more than it usually would, because you noted `/recall` takes the same permissive `SearchRequest` body as `/search` — a permissive body will not reject `topK`, so it fails exactly the way #1405 describes, a key that reaches the server and goes nowhere, just from the other direction.

Three lines in one call close it, asserting the translation and the absence together:

```ts
test("recall translates topK and forwards only the extras", async () => {
  let captured: Record<string, unknown> = {};
  const client = makeClient((_url, init) => {
    captured = JSON.parse(init.body as string);
    return jsonResponse(200, liveRecallBody([]));
  });
  await client.recall("q", { topK: 7, diagnostic: true });
  assert.equal(captured.top_k, 7);
  assert.equal(captured.diagnostic, true);
  assert.equal("topK" in captured, false);
});
```

I checked it fails for the right reason before claiming it works: against the mutant it exits 1 with `true !== false` on the last assertion, and against your implementation the suite is 25 pass, 0 fail. It also replaces the two-call dance, so the file gets shorter, not longer.

One note if you take the optional Python test from the issue: that hole cannot exist there. `recall(self, query, *, top_k=5, **extra)` binds `top_k` as a named parameter, so it can never land in `extra` — the asymmetry is a TypeScript-only artifact of using one options bag for both.

Your call and the maintainer's, not mine — the fix in this PR is correct as written and nothing above blocks it.

— TonyDzi, Palo Alto AI Research Lab · governed memory for agent fleets is what I do all day, so this one hit close to home; the rest of the machine (second brain, fleet coordination, agent consensus) is at github.com/tonydzi

deepspace28:
@erni-a Pushed on the rebased branch — both blockers are done now:

- **DCO**: every commit carries the `Signed-off-by` trailer (used `git rebase --signoff`).
- **Rebase**: branch is rebased onto today's `main` (through #1574). The recall() implementation also picks up the body-type validation from #1243, composed with the extra-forwarding change.
- **The optional test note is taken too**: replaced the two-call test with the single-call version asserting `top_k` is translated, the extra key reaches the body, and `topK` does not appear raw — 23/23 pass with tsc clean.

Both commits show the trailer via `git log --format='%B' -2 00175ef`. Ready for your approval run and merge.

erni-a (approved):
Thanks @deepspace28, fast turnaround. Rebase is clean, both commits signed, and the single-call test is better than what I suggested: it pins the topK translation and that the raw key never leaks. Merging as soon as CI finishes.

PR #1471, opened 2026-09-10
feat(search): freshness_reference — As-Of Recall for backfilled corpora

yanki-m:
**Pitch page (one-pager for the team):** https://claude.ai/code/artifact/c0f6ea56-c07a-4123-a344-9fb387a255f0

## What

A per-tenant knob, `search.default_profile.freshness_reference` (`int` in `[0, 1]`, default `0`). When a tenant sets it to `1` **and** a request carries `valid_at`, freshness and the temporal window are measured from `valid_at` against each row's event time — `coalesce(ts_valid_start, created_at)` — instead of from `now()` against `greatest(created_at, ts_valid_start)`.

| Knob | Request | Clock / anchor | Effect |
|---|---|---|---|
| `0` (default) | any | `now()` / `greatest(created_at, ts_valid_start)` | byte-identical |
| `1` | no `valid_at` | same | byte-identical |
| `1` | `valid_at` present | `valid_at` / `coalesce(ts_valid_start, created_at)`; `ts_valid_end` and the window compared to `valid_at` | retargets |

Only the conjunction moves the clock.

## Why

A corpus ingested in one sitting — a customer's imported history, a transcript backfill, a benchmark — has every row the same age under the default anchor, because `greatest()` picks `created_at` whenever the event predates its ingest. Freshness becomes a constant and "last month" is a window ending on the import date. With `valid_at` set, the existing filter keeps the right rows and then freshness floors all of them: the filter and the ranker argue.

`valid_at` already did two of the three things a `reference_time` field would do — future-row filter and expiry penalty (`postgres_service.py`), relative-date resolution (`extract_temporal_hint.py:35`). This closes the third without a second temporal parameter.

## Why a knob, not a default

Whether `ts_valid_start` means *event time* or *start of a validity window* ("price effective from…") is a property of the tenant's data. The `greatest()` in the default anchor is exactly the guard the latter relies on — such a row must never rank as older than its ingest — so this is a switch that preserves it, not a new default. Tenant-level (not agent-tunable) for the same reason, alongside `score_formula` and `ann_pool_size`. The recall-usage decay stays on `now()`: "recently recalled" is a system-time fact.

## Safety

- Default off → the scoring SQL is the same expression tree (`ref_ts` resolves to `func.now()`).
- Knob on without `valid_at` → identical ids **and** scores (pinned by test).
- Zero production callers send `valid_at` today (enterprise, plugin, clients, rail).
- Skew-safe in both directions: storage reads `sp.get("freshness_reference", 0)`; an older storage ignores the key, an older core-api never sends it.
- Naive `valid_at` is read as UTC rather than failing the `timestamptz` bind.
- No DB migration — the knob lives in the existing `organization_settings` JSONB; the settings model is untouched.

## Plumbing

Rides the table-driven knob machinery end to end: `SEARCH_KNOBS` (`sql=True`), `_validate_default_search_profile`, `validate_search_profile`, `SQL_SCORING_PARAM_KEYS`, `resolve_search_params`. `PUT /settings` validates it; the diagnostic's `search_params` echoes the resolved value.

## Tests

`tests/test_freshness_reference.py` (against Postgres) pins: default is flat on a backfilled corpus; knob ranks the nearer event first on freshness alone (same vector, same weight, no lexical match); knob inert without `valid_at`; temporal window ends at `valid_at`; naive == aware. Knob-table drift tests extended. Locally: 91 root + 15 storage-api scored-search tests green; MCP/plugin spec tests 117/117.

## Second commit — docs and generated specs

- `integration-guide.md`: two lines still claimed MCP lacks `valid_at`; `caura_recall` has had it since C31/D2. Fixed — closes **oss-0902-l-64**.
- `caura_recall`'s `valid_at` description states the full contract in 24 tokens; tools/list lands at **5,312** against the 5,320 ceiling (main is at 5,287). Ceiling not raised.
- `plugin/tools.json` and `tests/fixtures/tools_list_baseline_v1.json` regenerated — one line each.
- `capture_baselines.py`: `model_dump(by_alias=True)`. #682 taught the live-match test to compare wire names but left this script dumping Python field names, so the README's regeneration recipe produced a fixture the test rejects on 74 lines. One argument realigns them.

Noted, not changed: `test_mcp_token_budget.py` and `test_plugin_source_manifest.py` call `Path.read_text()` with no encoding and fail on a cp1255 Windows locale; `PYTHONUTF8=1` reproduces CI.

## Harness side

Both benchmark harnesses had been dropping the `query_timestamp` their runners supplied. They now send it as `valid_at` (harness-only, no server dependency — measurable against production today) and can set this knob on the benchmark tenant. Suggested run order: `CAURA_VALID_AT=1` first, then the knob after deploy, so the two halves are attributed separately.

### DIFF
--- a/capture_baselines.py
+++ b/capture_baselines.py
@@ -45,7 +45,14 @@ async def _capture() -> list[dict]:
 
     tools = await mcp_server.mcp.list_tools()
     tools_list = [
-        t.model_dump(mode="json") if hasattr(t, "model_dump") else dict(t.__dict__) for t in tools
+        # ``by_alias=True`` matches what ``test_v1_baseline_matches_live_registry``
+        # compares against — the SDK's wire names (``inputSchema``), not the
+        # model's Python field names, which became snake_case in mcp 2.x. Without
+        # it this script writes a fixture the test rejects on 74 renamed lines.
+        t.model_dump(mode="json", by_alias=True)
+        if hasattr(t, "model_dump")
+        else dict(t.__dict__)
+        for t in tools
     ]
     # The tests sort by name before comparing, because spec modules are
     # auto-loaded in `pkgutil.iter_modules` order; store it sorted so the file
--- a/common/constants.py
+++ b/common/constants.py
@@ -696,6 +696,27 @@ SEARCH_KNOBS: dict[str, SearchKnob] = {
     # ann_pool_size forced to 0 and the shadow call with the configured size).
     # Inert when ann_pool_size is 0.
     "ann_pool_shadow": SearchKnob(int, (0, 1)),
+    # Reference clock for freshness and the temporal window when the request
+    # carries ``valid_at`` (the as-of time of the question). 0 = ``now()`` —
+    # byte-identical to today. 1 = age is measured from ``valid_at``, and the
+    # row's anchor is ``coalesce(ts_valid_start, created_at)`` instead of
+    # ``greatest(created_at, ts_valid_start)``. Inert without ``valid_at`` on
+    # the request; ``valid_at`` without the knob keeps today's behaviour, so
+    # only the conjunction retargets the clock and no existing caller moves.
+    #
+    # Per TENANT on purpose, not per agent or per request: whether
+    # ``ts_valid_start`` means "when the event happened" (a backfilled history,
+    # a benchmark whose questions carry an as-of date) or "start of a validity
+    # window" (a price effective from a date) is a property of the tenant's
+    # data, and only the former wants freshness anchored to it. The
+    # ``greatest()`` in the default anchor is exactly the guard that protects
+    # the latter — a validity-window row must never rank as older than its
+    # ingest — which is why this is a switch and not a new default.
+    #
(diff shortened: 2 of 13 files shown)


Eldad-Caura:
Addressed the review finding in `d7f6a5b5`:

- normalize `valid_at` once at the start of `memory_scored_search`
- reuse the same explicitly typed UTC-aware timestamp expression for both the freshness reference and `currency_factor`
- add a statement-level regression test with `freshness_reference=0` proving the pre-existing currency path never retains a naive datetime bind

Verification:
- `tests/test_freshness_reference.py`: 6 passed
- `core-storage-api/tests/test_fts_score_single_render.py`: 11 passed
- core-storage-api mypy: clean
- full `core-storage-api/src/` and `tests/` ruff check + format scopes: clean

Eldad-Caura:
@claude

Eldad-Caura:
@erni-a

erni-a (approved):

Issue #1401, opened 2026-09-08
chore(deps): Dependabot watches no published client package under clients/


Infrasity-Labs:
## What

`.github/dependabot.yml` watches the root `requirements.txt`, the four uv services, the root `package.json`, `/plugin`, GitHub Actions, and Docker. It watches **none** of the seven published client packages under `clients/`:

| Manifest | Ecosystem | Watched |
|---|---|---|
| `clients/python/pyproject.toml` | pip | no |
| `clients/typescript/package.json` | npm | no |
| `clients/caura-meta/pyproject.toml` | pip | no |
| `clients/caura-sdk-meta/pyproject.toml` | pip | no |
| `clients/npm-caura-client/package.json` | npm | no |
| `clients/npm-caura-sdk/package.json` | npm | no |
| `clients/npm-sdk/package.json` | npm | no |

## Why it matters

These are the artifacts published to PyPI and npm — the code that ends up inside other people's projects. Their dependencies (`httpx>=0.27`, `typescript`, `@types/node`, `ruff`) get no advisory PRs and no version bumps today, while the internal services do.

## What to change

Add entries to `.github/dependabot.yml` following the shape already used there. The `uv` entry shows the multi-directory form, which is what you want here — one `pip` entry and one `npm` entry across the client directories, rather than seven near-identical entries:

```yaml
  # --- Python: published client packages ---
  - package-ecosystem: "pip"
    directories:
      - "/clients/python"
      - "/clients/caura-meta"
      - "/clients/caura-sdk-meta"
    schedule:
      interval: "weekly"
    open-pull-requests-limit: 3
    labels:
      - "dependencies"
      - "kind/chore"
      - "area/infra"
    commit-message:
      prefix: "deps(clients)"
    groups:
      clients-pip-minor-patch:
        update-types: ["minor", "patch"]
      clients-pip-majors:
        update-types: ["major"]
```

…and the npm equivalent for the four `package.json` directories.

Read the header comment in `dependabot.yml` first — it explains, with a reverted PR as evidence, why **no** entry sets `versioning-strategy`. Don't add one.

Note the `ruff>=0.15.4,<0.16` cap in `clients/python/pyproject.toml` is deliberate (see its comment) — expect Dependabot to propose lifting it and expect that PR to be declined until the 18 ruff-0.16 findings are fixed.

## How to verify

[...]

PR #1479, opened 2026-09-13
chore(deps): point Dependabot's docker entry at the service Dockerfiles

### LINKED ISSUE #1402: chore(deps): Dependabot's docker entry points at / — the four service Dockerfiles are never scanned
## What

`.github/dependabot.yml` has one Docker entry:

```yaml
  - package-ecosystem: "docker"
    directory: "/"
```

All four service Dockerfiles live one level down:

```
core-api/Dockerfile
core-operations/Dockerfile
core-storage-api/Dockerfile
core-worker/Dockerfile
```

Dependabot's `docker` ecosystem scans the directory it is pointed at. At `/` it picks up `docker-compose.yml` / `docker-compose.dev.yml`, but never the four `Dockerfile`s.

## Why it matters

These are the images published by `publish-docker.yml` — the ones self-hosters run. Their base images get no update PRs, so a base-image CVE is invisible to the bot even though the repo has `security-scan.yml` and CodeQL wired up.

## What to change

Point the entry at the directories that actually contain Dockerfiles, using the multi-directory form the `uv` entry already uses (keeping `/` so compose files stay covered):

```yaml
  - package-ecosystem: "docker"
    directories:
      - "/"
      - "/core-api"
      - "/core-storage-api"
      - "/core-worker"
      - "/core-operations"
```

Keep the existing `labels`, `commit-message`, `open-pull-requests-limit` and the two `docker-minor-patch` / `docker-majors` groups exactly as they are — the header comment explains why majors are carved out, and that reasoning is unchanged.

Worth considering (say what you chose in the PR): the service `area/*` labels rather than `area/infra`, matching how the `uv` entry labels the same four directories.

## How to verify

Confirm the four base images resolve — `grep -h '^FROM' core-*/Dockerfile` — then after merge check **Insights → Dependency graph → Dependabot** shows the Docker entry listing five directories, not one.

## Scope

One config file, one entry. Related: #1401 (unwatched `clients/` packages) edits the same file.

---

_Filed unlabeled — the account that opened this lacks label-write permission. Intended labels: `good-first-issue, help-wanted, kind/chore, area/infra, status/needs-triage`._

burakeyler:
Closes #1402

## What

The `docker` entry in `.github/dependabot.yml` now uses `directories` instead of `directory: "/"`:

```yaml
directories:
  - "/"                  # docker-compose*.yml, as before
  - "/core-api"
  - "/core-storage-api"
  - "/core-worker"
  - "/core-operations"
```

`schedule`, `open-pull-requests-limit`, `commit-message` and the `docker-minor-patch` / `docker-majors` groups are unchanged. A short comment above the entry says why the directories are listed.

## Labels (the choice the issue asked about)

I kept `area/infra` **and** added the four service labels (`area/core-api`, `area/core-storage-api`, `area/core-worker`, `area/core-operations`), matching the `uv` entry.

- The entry still covers `/` (compose, infra), and a grouped PR can span any of the services, so a bump routes to the right owners whichever files it touches.
- All five labels exist in `.github/labels.yml`.
- If you'd rather have only the service labels, it's a one-line removal.

## Verification

- `grep -h '^FROM' core-*/Dockerfile`: each of the four Dockerfiles resolves `python:3.12-slim` and `datadog/serverless-init:1@sha256:…`, so Dependabot has real base images to track in each directory.
- The four directories each contain a `Dockerfile`.
- The file validates against SchemaStore's `dependabot-2.0.json` schema (as does `main`).
- After merge: **Insights → Dependency graph → Dependabot** should list the Docker entry with five directories.

This is independent of #1401, which edits the same file (`clients/` entries). Whichever lands second may need a trivial rebase.


erni-a (approved):
Thanks @burakeyler. Checked that the four Dockerfiles have real base images to track and that all five labels exist in labels.yml. Keeping area/infra plus the service labels is the right call. Stacks cleanly with #1480; merging.

PR #1493, opened 2026-09-14
fix(bench): supersession-aware recall for the regression sample — correct supersession no longer reads as a regression (reg-d15)

ran-taig:
## Premise check — verdict BEFORE the fix

reg-d15 carries an explicit warning: an earlier pass suspected the premise is inverted — that the gold event-pair memories really ARE being marked conflicted (a precision bug in detection), in which case "fix the metric" would be the wrong move. Verdict, from artifacts and code, established before any fix:

**H1 (the metric is supersession-blind) — CONFIRMED, and worth fixing regardless of H2.**

- The bench recall counts gold `source_uri`s against retrieved `source_uri`s and nothing else — no status awareness, no successor credit (the runner family's `compute_retrieval_metrics`: `found = |answer_uris ∩ retrieved_uris|`).
- The product deliberately retires the older side of an update: the detector marks it `outdated` (RDF path) / `conflicted` (semantic path) with a `supersedes_id` chain from the newer row, and **present-state search excludes retired rows** (`memory_scored_search`, core-storage-api `postgres_service.py` — merged and integration-tested in #1023, with a history-question bypass).
- The three #1025 dip questions all dodge the history-hint regexes **on phrasing alone**: "How many weeks in total do I **spent**…" (sic — matches neither `did I` nor `have/has/had I`), "What **is** the order of airlines… before today?", "What time **do** I wake up on Tuesdays and Thursdays?" (`_extract_history_hint`, `memory_service.py`). So their superseded golds are excluded by design.
- Therefore ANY supersession of a gold row — correct or not — mechanically lowers recall while accuracy holds or rises. That is exactly the A63 signature, and it hit **three times** in the 2026-08-27 sessions: #1023 iteration 1 (KU recall −12.5pp, mitigated by the history flag), #1023 final (its own PR text: "the golden should be re-captured — its recall metric credits superseded old values on two present-state questions this change correctly buries"), and #1025 (TR recall −8.0pp with TR accuracy **+14.3pp**, overall +4.7pp; per-question answers unchanged False→False/True→True, one rank improved 2→1). Every future detection improvement re-triggers the same false FAIL.

**H2 (detection over-fires on gold event pairs) — mechanism is live, occurrence in the A63 run NOT confirmable from available artifacts. Diagnosis brief below; the fix deliberately keeps it measurable.**

- The three miss questions' gold sets are shapes the judge contract itself protects: the weeks-reading question needs three dated **start/finish milestone pairs** (structurally identical to the prompt's own `hired in 2020 / promoted in 2023` non-conflict example); the airlines question needs **five distinct past flight events** incl. re-mentions (`temporal_supersession` as re-cut + `event_restatement` shapes); the wake-up question needs a **scope-qualified pair** — general 7:00 AM + "15 min earlier on Tue/Thu" (the prompt's `scope_mismatch` example is literally "weekday vs weekend"), and its answer (6:45 AM) is derivable only from BOTH rows.
- #1023 removed `temporal_supersession` from `NON_CONFLICT_REASONS`, so a judge answering `contradicts=true` with that reason now **stands** (pinned in `tests/test_contradiction_non_conflict_reason_gate.py`). The checked-in wet baselines (`scripts/wet_test_baselines/caura-124-baseline.jsonl`) show two providers answering `contradicts=true` on 3/3 sequential-shape probes — the eagerness the removed veto used to absorb. #1025 additionally widened the judged population to cosine 0.45–0.70 pairs, exactly where loosely-similar coexisting facts live, and its own "honest limit" section shows the nano judge is noisy on such pairs.
- Counter-evidence: #1023 reports 0 false positives on planted history batches; #1025's author audited each dip per-question and stands behind the demotions ("rows this change correctly demotes as superseded"); no accuracy dropped.
- The deciding artifact — per-pair judge verdicts + memory statuses from the A63 run dirs — is not bundled in this repo (neither is the harness itself; `benchmark/regression_sample/` did not exist until this PR), and reproducing needs a real judge LLM + real embeddings (a fake embedder invalidates the 0.45-floor candidate geometry). **Cheapest way to settle it:** from any A63 run dir, list the gold URIs whose stored status ∈ {outdated, conflicted} with their `supersedes_id` source rows and the judge's `reason` — the wake-up pair (`gpt4_2c50253f`, golds `…_answer_9af4e346_1/2`) is the cleanest litmus: if either row was marked conflicted against the other, that is an over-fire by the contract's own text. Alternatively one live probe: ingest those two turns and read the pair verdict.

## The fix (per verdict: fix H1, brief H2 — bench-only, no product code touched)

`benchmark/regression_sample/compare_to_baseline.py` — the run-vs-golden scorer, now in-repo. Recomputes accuracy and recall@k from the raw per-question artifacts on both sides (never trusts `retrieval_summary`), prints the per-category delta table with threshold flags, exit 1 on breach.

- **Supersession-aware recall (default):** a gold counts as satisfied when retrieved, **or** when it was superseded (all stored copies `outdated`/`conflicted`) **and** a retrieved memory's `supersedes_id` chain reaches it. A superseded gold with **no retrieved successor stays a miss** — a detector that buries information without surfacing a replacement keeps hurting recall, so genuine over-firing damage remains visible. The denominator never shrinks (dropping superseded golds could score 1.0 while retrieving none of the question's information; "recall" keeps meaning "fraction of required information whose lineage is present in top-k").
- **`--recall-mode legacy`** reproduces the historical status-blind counting exactly, for comparability with goldens captured before this scorer.
- **Over-firing instrument:** every superseded gold is listed (credited or lost, and by which memory) plus per-category `gold superseded` totals; the report states that *credited* does **not** certify the supersession was correct — audit the pairs when the count moves.
- Graceful degradation: artifacts without `all_stored_memories` statuses score legacy-style, counted and shown.
- `BENCHMARKS.md` documents the scorer, both modes, and the audit caveat.

## Tests / gates

`tests/test_regression_bench_superseded_gold.py` (unit, 14 tests):
- correctly-superseded gold + retrieved successor → recall 1.0, **no** regression flagged; the **same fixture in legacy mode reproduces the A63 false-FAIL** (breach on recall);
- genuinely missing gold → still flagged; superseded-and-lost → still flagged in BOTH modes;
- chain transitivity (C→B→A credits A from C) + corrupt-cycle guard; an unretrieved ACTIVE re-ingested copy blocks credit; degraded artifacts fall back to legacy; CLI end-to-end exit codes 0 (aware) / 1 (legacy) on the same data.
- Fails pre-fix: verified by stashing the scorer — collection error (`FileNotFoundError`), since the module did not exist before this PR; the legacy mode embodies the pre-fix semantics and is pinned against the fixture.

Gates: `pytest` 14/14 green; `ruff check` + `ruff format --check` clean on both files (CI lints `tests/`; `benchmark/` is outside the CI ruff/mypy scope — kept clean anyway).

Audit tracker: reg-d15 (D15)


ran-taig:
@claude

ran-taig:
@erni-a

erni-a (approved):

PR #1494, opened 2026-09-14
feat(recall): gate the boost-feeding bump on a confirmed-use signal (A41)

ran-taig:
## The loop (validated on `origin/main` @ 7ab95d46)

`recall_count` is bumped for **every returned row**, used or not, and feeds `recall_boost` back into the rank score:

1. `core-api/src/core_api/pipeline/steps/search/track_recalls.py` — TrackRecalls collects every row in `filtered_rows` and fire-and-forget bumps them via `increment_recall`. Post-A26 it gates on caller identity / tenant opt-in / diagnostic — but there is **no usefulness gate**.
2. `core-storage-api/.../postgres_service.py::memory_increment_recall` — `recall_count += 1`, `last_recalled_at = now()`.
3. `postgres_service.py` scoring — `recall_boost = 1 + (cap−1) · recency(last_recalled_at) · recall_count/(recall_count+10)` multiplied into the legacy score (`SCORE_FORMULA=0`, the default everywhere).

returned → bumped → boosted → returned. PR #411 (A26) only **dampened** it (cap 1.5→1.1, window 90→14d); the constants comment explicitly deferred "a confirmation-gated bump" as the real fix. And per the A50 decision work, the boost **also compounds correct repeat retrievals** on live replay — so the fix gates what *feeds* the boost and does not weaken or remove the boost itself.

## Signal inventory

| Signal | Available? | Verdict |
|---|---|---|
| Anything at TrackRecalls time | fires inside `search_memories`, before any LLM answer exists | nothing usefulness-shaped exists at that moment |
| `/recall` answer citations | the prompt deliberately exposes **no memory ids** ("the model must not be able to cite identifiers"); changing `RECALL_PROMPT` is benchmark-gated (A64/A57: fresh control pair on the 67-q sample) | not available without inventing new machinery |
| D8/D9 recall-event log | opt-in diagnostics; `returned` flag only, no used/cited field | measurement infra, not a usefulness signal |
| outcome-inference package | lazy/on-demand, session-level; `cross_agent_reuse` reads recall counters (circular) | not a per-memory recall-time signal |
| explicit feedback endpoints | none exist for memories | — |
| **evolve outcome reports** (`caura_evolve` / `report_outcome`) | agent explicitly names the memories it acted on in `related_ids`; already lands on exactly those rows via one atomic storage call | **the one genuine, already-wired usefulness signal** — deferred (async), which fits a deferred-bump design |

## Design: two counters, one source knob (measured, not assumed)

- **`recall_count` stays return-fed in all modes.** Its ~15 non-scoring consumers (lifecycle archive-stale `recall_count=0`, insights, reports, crystallizer load-bearing reads) all mean "was returned" — redirecting its feed would silently archive returned-but-unconfirmed memories. It doubles as the counterfactual.
- **Confirmed-use counter:** `evolve_apply_weights(mark_used=True)` — sent by `_adjust_weights` for every outcome report — bumps `metadata._system.recall_used_count` (+ `recall_used_at`) for the scope-filtered `related_ids`, in the **same transaction** as the weight clamp. All three outcome types count: success/failure/partial all say the agent **used** the memory; valence stays in `weight` (the delta the same call applies) — click-through semantics. `_system` is the deep-merge-protected platform namespace (B7 x C25); JSONB only, **no schema migration**; older storage servers ignore the unknown key (skew-safe).
- **Knob `recall_boost_source`** (SEARCH_KNOBS, `sql=True`, not agent-tunable — the `score_formula` pattern):
  - `0` (default): boost reads `recall_count` — **byte-identical statement**, pinned by the compiled-SQL ratchets.
  - `1`: the boost's saturation term reads the used counter and its recency reads the confirmation stamp. Cap/window/shape untouched — the boost still compounds correct repeat retrievals, now **confirmed** ones; merely being returned no longer moves the number the score reads.
  - Flip per tenant via `search.default_profile.recall_boost_source`; fleet-wide via `CAURA_RECALL_BOOST_SOURCE` (defensive #1441 env parse; anything but `1` fails closed to `0`).

**Rollout / measurement:** both counters accrue from day one with the knob off, so the returned-vs-used comparison (evolve coverage per tenant, correlation, whether high-returned/zero-used rows are the A49/A50 crowding artifacts) is measurable **before** any tenant flips — A50's "every boost must earn its weight". A tenant with no evolve traffic that flips anyway gets boost = 1.0 everywhere (no signal → no popularity boost), visible in the same measurement. Reverting is the same one-knob change and loses nothing: neither counter ever stops accruing.

## Tests

- **New** `tests/test_a41_recall_boost_source.py` (15 unit): knob table/wire contract, resolution precedence, org-settings validation (bounds + bool refusal), env parse via fresh-module exec, `_adjust_weights` sends `mark_used` for all three outcome types, typed client puts it on the wire.
- **New** `core-storage-api/tests/test_a41_recall_boost_source_statement.py` (7 unit, capture-and-compile): default statement byte-free of the used-counter machinery and still saturating `recall_count`; `source=1` divides the used counter with recency on the confirmation stamp and keeps the `<=>`/`ts_rank_cd` render counts; boost-disabled beats the knob; `evolve_apply_weights` issues no extra statement by default, and the tenant-scoped, `deleted_at`-guarded `_system` merge when marked.
- **Extended** `tests/test_ph5b_evolve_storage.py` (4 integration): bump + repeat increment, sibling `_system`/user-key preservation, tenant isolation, untouched default. Plus the pinned not-agent-tunable set in `test_search_default_profile.py`.
- **Failed-on-pre-fix proof:** with the 7 source files stashed (origin/main code), the root suite fails **14/15** and the statement suite **3/7** — the 4 passing there deliberately pin the unchanged default shape. All pass with the fix.

## Gates

- `pytest` (unit, `EMBEDDING_PROVIDER=fake`): new files + adjacent suites (default-profile, track_recalls_agentless, evolve, A10/A15, p0_3 recall boost, search pipeline, d17/benchmark/c4/p0_4 scoring) — **134 + 39 passed**; storage statement ratchets (`test_ann_pool_statement`, `test_fts_score_single_render`) — **32 passed** (default statement unchanged). Known env reds unrelated to this change: `test_a7_classifier_recall`, `test_search_pipeline::test_recall_returns_memory_with_pending_embedding` need a migrated Postgres (fail identically on clean origin/main).
- `ruff check` (core-api/src, core-storage-api/src, common/, touched tests) clean; `ruff format --check` on all touched files clean.
- `mypy`: core-storage-api `src/` clean; `common/` clean; core-api `src/` has 6 pre-existing environmental errors (local stub gaps) — byte-identical list on clean origin/main with a cold cache; no new errors.

Audit tracker: reg-a41 (A41)


ran-taig:
@claude

ran-taig:
@erni-a

erni-a (approved):

PR #1495, opened 2026-09-14
fix(extraction): stop one failed relation disabling the deterministic contradiction path

arkash20:
`process_entity_extraction` upserts each extracted relation in a loop. The
loop had no per-relation guard, so a single failure threw out of the whole
function into the outer handler, which logs:

```
Entity extraction failed for memory <id> (non-fatal)
```

"Non-fatal" is true of the request and false of everything after the loop:

- the **A65 predicate write-back** never runs, so the row keeps a NULL
  `predicate` — permanently, since nothing re-extracts it; and
- the **`Trigger.ENTITY` fire** never runs, and that is the only caller of
  A40's deterministic RDF pass.

So one transient storage error on one relation, out of dozens for that
memory, quietly removed that row from the non-stochastic contradiction
path for good, and reported success while doing it.

## Observed, not hypothesised

Found while wet-testing A40 on a local stack. A storage 500 on
`POST /entities/relations` reproduced it exactly: `subject_writeback
outcome=set` immediately followed by the "(non-fatal)" warning, predicate
never written, Path C never fired. It read as "A40 doesn't work" until
the traceback showed the cascade.

## The fix

Each upsert is guarded individually — the pattern `subject_writeback` and
`predicate_writeback` in this same function already use. A failure now
costs one relation instead of the remaining relations plus every stage
after the loop. Partial failures are also surfaced as a single
`relation_upsert_partial created=N failed=M` line, so a degrading entity
graph reads as one number rather than N scattered warnings nobody
aggregates.

## H-02 is unchanged, and one test had to move

`test_a_crash_after_the_writes_still_purges_a_dropped_row` used a failing
`upsert_relation` as its **vehicle** for "leaves by raising". That stopped
being one. The guarantee it protects is untouched: anything between the
link upsert and the final check that raises must still re-check and
purge. It now raises from the audit call instead — which its own docstring
already names as such a thing — and asserts the same property. Verified
it passes both with and without this change, i.e. the vehicle moved and
the guarantee did not.

A relation failure now takes the **finishing** path, whose check
(`_purge_written_artifacts_if_dropped` at the end of the `try`) is covered
by a new sibling test, so guarding the upsert cannot quietly open the leak
the original test exists to close.

## Notes

Every test in the new file fails with this change reverted — checked, not
assumed. Two earlier drafts were dropped for passing without the fix: they
were satisfied by the function's pre-existing outer `try` and by the
predicate write-back's own guard, so they asserted nothing.


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #1496, opened 2026-09-14
perf(crystallizer): stop the near-duplicate scan making one round-trip per row and relaying every vector

ran-taig:
## Evidence (audit oss-0814-m-37)

Both halves of the finding reproduce. They were two defects sharing one loop.

**Serial N+1.** `_check_near_duplicates` fetched a page of candidates, then awaited one `find_neighbors_by_embedding` POST **inside** the per-candidate loop:

- `core-api/src/core_api/services/crystallizer_service.py:971` — `for cand in candidates:`
- `core-api/src/core_api/services/crystallizer_service.py:976` — `neighbors = await sc.find_neighbors_by_embedding(...)`, awaited in-loop
- `core-api/src/core_api/constants.py:878` — `CRYSTALLIZER_DEDUP_BATCH_SIZE = 500`

That is 1 + 500 serial HTTP round-trips per batch, and it scales with rows: measured 21 calls for 20 rows, 401 for 400.

**Full embeddings, both directions.** The vector is selected, shipped down, and shipped straight back up without ever being read:

- `core-storage-api/src/core_storage_api/services/postgres_service.py:4642` — `SELECT m.id, m.embedding`
- `core-storage-api/src/core_storage_api/routers/memories.py:700` — `{"id": str(r[0]), "embedding": r[1]}` per row
- `core-api/src/core_api/services/crystallizer_service.py:973` — `embedding = cand["embedding"]`
- `core-api/src/core_api/services/crystallizer_service.py:980` — `"query_embedding": embedding`, the only use of that variable
- `core-storage-api/src/core_storage_api/services/postgres_service.py:4697` — `1 - (n.embedding <=> :query_emb)`; pgvector computes the similarity server-side, so core-api had no reason to hold the vector at all

`core-storage-api/src/core_storage_api/schemas.py:100-106` already prices a 1024-dim vector at "~20 KB of JSON floats per row". Measured through the repo's own `_serialise_value` path: 22.8 KB, travelling down once as the candidate and up again as the query.

## Measurement — per 500-row batch

| | round-trips | wire | vectors on the wire |
|---|---|---|---|
| **before** | 501 | 22,518 KB | 21.8 MB |
| **after** | 1 | 325 KB (worst case, every candidate returns the full K=5) | 0 |

501x fewer round-trips, 69x less wire, and the vector relay is gone in both directions.

## Fix

The candidate query and the per-candidate neighbour query become one statement — a candidate CTE `LEFT JOIN LATERAL` its own top-K — behind the existing `POST /memories/near-duplicates`, which now answers `{"candidate_ids", "pairs"}`. `/memories/neighbors-by-embedding` existed only to serve the loop and goes with it.

`LEFT` is load-bearing: a candidate with no near-duplicate still yields a NULL-neighbour row, so `candidate_ids` remains the full swept set. An inner join would have stamped `last_dedup_checked_at` on matched rows only and re-scanned every unmatched row on every future sweep — the N+1 traded for an unbounded re-scan.

**The database does the same work it did before.** `EXPLAIN (ANALYZE)` on a 4k-row table carrying `ix_memories_embedding_hnsw` confirms the LATERAL still probes the ANN index once per candidate:

```
->  Limit  (loops=500)
      ->  Index Scan using m37_plan_embedding_idx on m37_plan n
            Order By: (embedding <=> m.embedding)
```

Same index, same per-candidate top-K, same `loops=500`. What goes away is the transport, not the work.

## Behaviour is unchanged — reg-a72 untouched

No threshold, no neighbour fan-out, no cluster logic, no LLM call added or removed. This is the PERF half only.

- `CRYSTALLIZER_DEDUP_THRESHOLD` and `CRYSTALLIZER_DEDUP_NEIGHBORS` are now sent in the request rather than left to storage defaults, so the sweep's tuning stays in core-api's constants where reg-a72 will look for it instead of being split across two services.
- Storage returns pairs in the order the loop visited them — candidate by candidate (`created_at DESC`), nearest neighbour first — so the positional `CRYSTALLIZER_MAX_DEDUP_PAIRS` cap bites on exactly the same pairs.
- The trailing `c.id` / `nb.id` sort keys make an order Postgres never promised deterministic rather than changing one it did. Candidates tie on `created_at`, neighbours tie on distance; similarity is symmetric, so a tie never changed a pair's recorded score, only which pairs made the cap.

## Consumer audit

`memory_find_near_duplicate_candidates` and `memory_find_neighbors_by_embedding` had no caller outside this sweep — the existing docstring said so, and a repo-wide grep across all file types confirms it: zero remaining references to either symbol, to `find_neighbors_by_embedding` on the client, or to `/memories/neighbors-by-embedding`, including in docs, OpenAPI and TS. `sc.check_near_duplicates` has exactly one production call site, `crystallizer_service.py:978`.

The M-61 guard (archived rows excluded from **both** ends of a pair — the fix that stopped the sweep re-clustering rows it had archived itself, #1313) moved onto the fused query rather than being retired. Same two predicates, now asserted through the query the sweep actually runs.

## Tests

`tests/test_m37_near_dup_scan_batching.py` — 8 tests, beside the existing crystallizer tests. The fake storage deliberately answers **both** the old and the new protocol, so the pre-fix code runs to completion against it and fails on the contract rather than on a missing key.

**Stashed to pre-fix: 5 fail, 3 pass.**

| test | pre-fix |
|---|---|
| `test_a_batch_costs_one_round_trip_not_one_per_row` | FAIL — `assert 40 == 0` |
| `test_round_trips_do_not_grow_with_the_corpus` | FAIL — "round trips scale with rows: 23 for 20 rows, 403 for 400" |
| `test_no_embedding_goes_up_to_storage` | FAIL — "find_neighbors_by_embedding still ships a 1024-dim vector in 'query_embedding'" |
| `test_the_scan_never_reads_the_embedding_it_is_offered` | FAIL — request shape still lacks `threshold`/`neighbor_limit` |
| `test_the_storage_response_carries_no_vector` | FAIL — route has no `memory_find_near_duplicate_pairs` |
| `test_same_pairs_same_similarities_same_order` | pass |
| `test_the_pair_cap_bites_on_the_same_pairs` | pass |
| `test_every_swept_row_is_stamped_not_only_the_matched_ones` | pass |

The 3 that pass on both sides are the behaviour-identity ones, asserted against `_reference_serial_scan` — a transcript of the pre-fix loop run over the same oracle. Passing before **and** after is the point: same pairs, same similarities, same discovery order, same cap, same stamp set.

## Gates

- `pytest` (core-api): **100 passed** across the m37, P5 crystallizer, archive-sweep, semantic-dedup, tenant-guard and crystallization-wedge files — including the DB-backed near-duplicate integration tests, which pass unchanged.
- `pytest` (core-storage-api): **387 passed / 10 failed / 7 errors — identical to clean main.** The reds are the migration-040 and integration files, which need a migrated Postgres.
- `ruff check` core-api + core-storage-api: clean. `ruff format --check` on all touched files: clean.
- `mypy` core-storage-api: clean, 88 files. `mypy` core-api: 6 errors in 3 files, **identical to clean main** (missing croniter/dateutil stubs, mcp_server attrs) — none in touched files.

## Left for reg-a72

Nothing here blocks it, and nothing here anticipates it. The thresholds it will want to retune (`CRYSTALLIZER_DEDUP_THRESHOLD`, `CRYSTALLIZER_DEDUP_NEIGHBORS`, `CRYSTALLIZER_MAX_DEDUP_PAIRS`, `CRYSTALLIZER_DEDUP_BATCH_SIZE`) are all still in `core-api/src/core_api/constants.py` and now all flow through the one request body, so a retune is a constants change on one side rather than a two-service change.

Audit tracker: oss-0814-m-37


ran-taig:
@claude

ran-taig:
@erni-a

erni-a (approved):

PR #1497, opened 2026-09-14
fix(keystones): tell a caller why the rule set is empty (F9)

arkash20:
Agents are taught to call keystones at every session start and obey what
comes back. A tenant that never authored a rule paid that round-trip and
received `count: 0` — a result indistinguishable, from the caller's side,
from three different situations:

- "this tenant has no standing policy",
- "authoring failed", and
- "you asked the wrong scope".

The first is the reasonable inference, so the agent stops asking and
carries standing constraints in recall instead — the exact thing
keystones exist to prevent.

The empty result now names the one fact a caller cannot deduce from an
empty array: that **authoring is a separate, trust-gated step**, and that
empty is not failure.

## Where it applies

- **MCP** (`caura_keystones`) — the surface agents actually read. Hint
  added to the payload when `rows` is empty.
- **REST** (`GET /keystones`) — **envelope form only**. The bare array
  stays the default response shape; adding a key to it would change the
  wire contract for every existing consumer, which is precisely what the
  C30/D1 opt-in was designed to avoid.

One constant in `core_api.constants`, imported by both, so the two
surfaces cannot drift into describing the same state differently. A test
forbids either from inlining its own copy of the text.

## Cost

Emitted only on the empty result, so a tenant with rules pays nothing.
It is a **response** field, not schema, so it does not touch the
`tools/list` token ceiling that `_AGENT_ID_DESC` had to respect —
`tests/test_mcp_token_budget.py` still passes, and is included in the run
below for that reason.


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #1499, opened 2026-09-14
test(contradiction): assert overlap, not the clock, in the batch concurrency tests

ran-taig:
`TestBatchConcurrency::test_concurrent_execution` gathered five 100ms sleeps and asserted `elapsed < 0.3`. It failed on #1414 with "Expected concurrent (<0.3s), got 0.39s" in a branch touching nothing near contradiction batching, and went green on a plain re-run — so every branch-vs-main baseline carried a coin flip, which is how a real regression gets waved through as "probably the flake".

## The flake mechanism

The claim the test wants to make is that the checks **overlap**. Wall-clock was only a proxy for that, and a proxy with 0.2s of absolute headroom hands the verdict to the scheduler rather than to the code.

CI makes that headroom thinner than it looks: `tests/` runs on a 2-vCPU `ubuntu-latest` runner, under `--cov` (coverage traces every executed line, including the event loop's own internals), alongside a Postgres service container. Nothing there is a bug — it just means the loop is not always serviced promptly.

Measured directly, replaying the old body 3000x while the box was oversubscribed ~10x (load avg 135 on 12 cores):

| | nominal | p50 | p99 | max |
|---|---|---|---|---|
| duration of the fan-out | 5ms | 10.6ms | 67.7ms (13x) | **18,554ms** |

One iteration stalled **18.5 seconds** for what is nominally a 5ms operation. No code changed across those iterations — only how promptly the process got scheduled. At the original 0.1s sleeps, a stall of a fraction of that size is exactly the reported 0.39s.

## The new assertion, and why it cannot flake

Both tests now count in-flight calls and assert the peak — the idiom already used by `test_contradiction_fanout_bounded`:

```python
assert peak_in_flight == 5, (
    f"checks did not overlap — peak in flight was {peak_in_flight}, expected 5"
)
```

Overlap is decided by event-loop *ordering*, not by elapsed time: `gather` starts all five coroutines, and each one increments the counter before its first `await`, so none can finish before all have started. A starved runner delays every step equally and cannot change the answer. Across the same 3000 stalled iterations above — including the 18.5s one — the peak was 5 **every time (0 deviations)**.

No duration sanity check was kept. `peak == 5` is strictly stronger than any wall-clock budget, and adding one back would reintroduce the only thing that could flake.

## Serial-regression proof

The guard that this did not just weaken the assertion into uselessness. Replacing the gather with `[await t for t in tasks]`:

| assertion | serial behaviour |
|---|---|
| **new** (both tests) | **FAIL** — `peak in flight was 1, expected 5` (and `expected 8`) |
| old `test_concurrent_execution` | FAIL at 0.51s — guard preserved |
| old `test_timeout_per_task_not_total` | **PASSED while fully serial** |

That last row is the second finding. It timed each call's *own* duration (`t0` was taken inside the mock, after the fan-out had already started) and asserted `max(call_times) < 0.2` — which a serial loop satisfies just as well, since each call still takes ~50ms on its own. Its docstring claims "total wall time should be << 8 * 50ms", but it never measured total wall time. It could not catch the regression it described; it could only flake. It is now a real check (peak == 8, and no task cut off by a timeout).

Net: one flaky-but-real guard kept and made deterministic, one vacuous guard repaired.

## Validation notes

- **The test never exercised production code.** It defines a local `mock_check` and gathers it — it was asserting that CPython's `asyncio.gather` is concurrent. Worth knowing when reading the "serial regression" proof above: the call path serialized is the test's own fan-out. Since A61, `_detect` judges multiple candidates through **one** batched LLM call (`_llm_contradiction_check_batch`), not a per-candidate `gather` fan-out, so there is no production fan-out here for this test to have been guarding. `TestBatchIntegrationWithDetect` and `TestBatchedJudgeA61` are what cover the real path, and are untouched.
- **The #1461 admission gate does not interact with this.** `_detection_gate()` bounds whole detection *passes* at `contradiction_detection_concurrency` (16) and sits in `_detect`, which this test never enters. Even if it did, 5 < 16 would never queue. The assertion's meaning is unchanged by that gate.
- **Repeat runs:** `TestBatchConcurrency` 50/50 pass under bounded CPU load at `nice -19` (load avg peaked ~18 on 12 cores). Plus the 3000-iteration in-process stress above, 0 deviations.
- **Full suite:** two consecutive runs on clean `main` (83913d1c) each produced a **byte-identical** 28-failure set, and this test was in neither; the fixed branch produces that same 28-failure set. Those 28 are the known Postgres-dependent reds (`test_a7_classifier_recall`, `pipeline/`, `test_ann_pool_behavior`, `test_readme_keyword_recall`). I could not reproduce the reported "38 vs 37 failures on main" shape on this machine — flagging that rather than claiming it verified. Structurally the test is fully self-contained (no fixtures, no production code, no shared state), so a failure of it can only ever account for exactly one differing entry and cannot cascade.
- **Scope:** `tests/` has wall-clock assertions elsewhere (`test_bulk_atomicity.py:660`, `test_d7_parallel_embed_per_task_timeout.py:338`, `test_ingest_commit.py:430`, `test_embedding_retry.py:569`) with the same shape but much larger margins. Left alone deliberately — out of scope here. The adjacent contradiction/concurrency suites (`test_a19_detection_admission_gate`, `test_contradiction_fanout_bounded`, `test_per_tenant_concurrency`, `test_embedding_concurrency_gate`) already assert properties and have no wall-clock assertions.

Tests only; no production code touched. The file also drops from 4.7s to 0.75s — with the property asserted directly, the sleeps no longer need to be long enough to measure.

Audit tracker: oss-0909-l-01


ran-taig:
@claude

ran-taig:
@erni-a

erni-a (approved):

PR #1500, opened 2026-09-14
perf(contradiction): fetch every candidate's entity context in one batched pair of calls (m-05)

ran-taig:
## The N+1, in code

`contradiction_detector.py` resolved entity context one memory at a time, on both Path C entry points.

**Detection** — `core-api/src/core_api/services/contradiction_detector.py:2707` (pre-fix):

```python
fetched = await asyncio.wait_for(
    _bounded_gather(
        [
            _fetch_entity_context(sc, str(memory_id), tenant_id),
            *(_fetch_entity_context(sc, str(c.get("id")), tenant_id) for c in candidates),
        ]
    ),
    ...
)
```

**Retraction** — `contradiction_detector.py:2284` (pre-fix): the same helper, once per side.

And `_fetch_entity_context` itself (`contradiction_detector.py:1902`, `:1919`) made two hops per memory:

```python
links_by_mem = await sc.get_entity_links_for_memories([memory_id], tenant_id)   # a LIST endpoint, one id
...
entity = await sc.get_entity(str(entity_id), tenant_id)                          # one call per link
```

Two separate multipliers:

1. `get_entity_links_for_memories` (`core-api/src/core_api/clients/storage_client.py:1195`, storage route at `core-storage-api/src/core_storage_api/routers/memories.py:728`) has **always** taken a list of memory ids. The detector was its only non-test caller, and it never passed more than one.
2. `get_entity` fired once per link with no dedup across candidates — and entity-overlap candidates are selected *because* they share entity rows with the new memory, so the shared subject entity was refetched once per candidate that linked it.

## Round-trip measurement

Instrumented against the real entry point (`detect_contradictions_by_entities_async`) with a recording storage double, 20 candidates × 2 links, one shared subject entity:

| | links calls | entity calls | total |
|---|---|---|---|
| before | 21 | 42 | **63** |
| after | 1 | 1 | **2** |

At the detection cap (`_ENTITY_LINKS_DETECTION_FETCH_MAX_CANDIDATES` = 40) with the ~1-5 links per memory documented on `_format_entity_context`, that is 41 + ~120 ≈ **160 round-trips per detection pass**, plus 2 + ~6 for the retraction phase. Now 2 and 2 — flat, independent of candidate count.

## Relation to #1461's admission gate

This is the concrete cause behind reg-a19 that #1461 deliberately did not touch, and its own write-up said so: the gate "bounds how many runs contend... fixing it will materially raise the gated throughput per slot".

Re-checked against current code, that characterisation holds, with one sharpening. The gate did not reduce this cost — it **queued** it. A run's ~160 round-trips now happen inside a held slot instead of racing other runs for the storage pool, so *round-trips per slot* is exactly the term that sets gated throughput at `contradiction_detection_concurrency` = 16. Nothing about the gate changes; the comment at `contradiction_detector.py:2648` is updated to say what a held slot now costs.

The seconds-long judge hold still dominates a single run's wall clock. This is a throughput fix, not a latency fix for one pass.

## The fix

`_fetch_entity_contexts` (plural) — one links call naming every memory, entity ids deduped across all of them, one hydration call. Contexts are assembled per memory **in link order**, so the judge receives byte-identical context: same rows, same order, duplicate roles preserved, missing/foreign entities dropped exactly as `get_entity` returning `None` did.

The links half needed no new endpoint. The hydration half did — there was no batch entity read (`/entities/bulk-resolve` resolves by name, `/entities/memory-ids-by-entity-ids` goes the other direction). Added as a contained, additive read, no migration:

* `POST /entities/by-ids` — `core-storage-api/src/core_storage_api/routers/entities.py:87`
* `entity_get_by_ids` — `core-storage-api/src/core_storage_api/services/postgres_service.py:6329`
* `get_entities_by_ids` — `core-api/src/core_api/clients/storage_client.py:1556`

It carries the same tenant predicate `GET /entities/{entity_id}` got in #1174, applied in SQL, and **filters rather than rejects** on a partial match for the reason #1162 gives: a batch read that 404s the whole request tells the caller which of the ids it named exist in someone else's tenant.

**Both batch calls fall back to the old per-memory / per-id path on failure.** Not decoration — core-api and core-storage-api deploy separately, so a core-api carrying this change can reach a storage instance that 404s `/entities/by-ids`, and one oversized 41-id body is a failure mode single-id calls never had. Without the fallback, either case silently empties the contexts, which reverts Path C to the base judge — the CAURA-134 regression class. An *empty* batch response is a real answer and is not retried per-id.

## Tests

`tests/test_m05_batched_entity_context.py` (new, 9 tests):

* `test_detection_pass_makes_one_links_call_and_one_entity_call` — the round-trip pin, driving the real entry point. **Fails pre-fix** (verified by stashing only the two source files): `AssertionError: ... pre-fix this was 1 + len(candidates) = 21, observed 21`. Also asserts no duplicate ids reach the batch body.
* behaviour identity: batched output `==` per-memory `_fetch_entity_context` output, memory for memory, including link order and the same entity appearing twice under different roles.
* missing entity drops only its own link; a link-less memory gets `[]` rather than a missing key.
* each fallback: links-batch failure → per-memory path; entity-batch 404 → per-id path; non-mapping response → per-id path; empty mapping → accepted as an answer.
* storage client posts `/entities/by-ids` with `read=True`.

`core-storage-api/tests/test_entity_batch_read_tenant_scope.py` (new, 8 tests): own rows keyed by id; a foreign id absent rather than returned; **mixed own+foreign returns only the caller's rows**; foreign indistinguishable from missing; omitted `tenant_id` → 422; malformed uuid → 422 not 500; non-list → 422; empty list → `{}`.

Four existing tests patched `_fetch_entity_context` to reach the outer `except`, or asserted the N+1 call count as the "fetch ran" signal. They now patch the plural helper and assert one batched call naming every memory — no behaviour under test changed. (`test_contradiction_batch.py` is untouched.)

## Gates

* `pytest` core-api contradiction/entity selection: **283 passed**
* `pytest core-storage-api/tests/`: **405 passed**
* Full `tests/` sweep: 6614 passed, 23 failed — the same 23 failing identically on clean `main` (ANN/search/freshness/relations suites needing a migrated Postgres), verified by stashing the change and re-running.
* `ruff check core-api/src/ core-storage-api/src/ <touched tests>`: clean
* `ruff format --check` on all 8 touched files: clean
* `mypy core-storage-api/src/`: clean. `mypy core-api/src/`: 6 pre-existing errors, all in `config.py` / `common/enrichment/service.py` / `mcp_server.py` — none in touched files.

## For the reviewer

* `_fetch_entity_context` (singular) stays, and stays tested — it is now the fallback and the one-memory case, not dead code.
* `_bounded_gather` and its cap are still needed: a storage instance that 404s the new route puts every run back on the per-id fan-out at once.
* Does not touch the extraction/relations path #1495 changed.

Audit tracker: oss-0902-m-05


ran-taig:
@claude

ran-taig:
@erni-a

erni-a (approved):

PR #1504, opened 2026-09-14
fix(extraction): record extraction failures instead of only logging them (09/02 M-40)

arkash20:
`tracked_task` writes a `BackgroundTaskLog` row **only when the coroutine it
wraps raises**. `process_entity_extraction` catches everything, logs
`"(non-fatal)"`, and returns normally — so the wrapper saw a success.

The consequences compound:

- the table an operator actually inspects stayed **empty**,
- the memory permanently kept **no entities**, and
- nothing retried it, or knew to.

## Found by hitting it

A storage 500 on `POST /entities/relations` produced exactly this during an
A40 wet test. It presented as **"A40 doesn't work"** — the deterministic
contradiction path was dead for that memory and nothing anywhere said a
task had failed. #1495 fixed the specific cascade; this fixes the silence
that made it hard to see.

## Recording, not re-raising

I tried re-raising first. It is defensible — all six call sites wrap this
in `tracked_task`, so the wrapper would catch and persist. But five
existing tests pin the non-raising contract, and the module's handler
comments reason repeatedly about the difference between "a logged
non-fatal failure" and "an unhandled task exception". Re-raising would
turn one into the other for any caller that is not wrapped.

So the handler records the failure itself and still returns normally. The
caller's contract is unchanged; what changes is that the failure becomes a
**row** instead of a **line**.

Ordering is load-bearing: the record happens **after** H-02's
`_purge_written_artifacts_if_dropped`, so a memory dropped mid-extraction
still has the graph rows this run wrote cleaned up first.

## One row shape, not two

`record_task_failure` is split out of `tracked_task` rather than
duplicated, and `tracked_task` now calls it. A hand-rolled
`add_task_failure` payload in the handler would drift from the wrapper's,
and an operator would be reading two different things out of one table. A
test asserts the persistence lives only in the helper.

Success still writes nothing — that is why `tracked_task` only logs on
failure, and a test pins it so a future change cannot make this table grow
on every write.

## Notes

All 7 tests in the new file fail with the change reverted — checked, not
assumed.


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #1505, opened 2026-09-14
fix(crystallizer): actually apply the reports pagination the API advertises (09/02 M-12)

arkash20:
`GET /crystallize/reports` declared `limit` (`ge=1, le=100`) and `offset`
(`ge=0`), validated both, and then called `sc.list_reports(tenant_id)`
without them. The storage route took only `tenant_id` too, so the query
ran on `report_list_by_tenant`'s **own defaults** — `limit=10, offset=0`.

The effect is worse than "the parameters are ignored":

- `?limit=100` returned **10** rows, silently capped;
- `?offset=50` returned the **first** 10, so a caller paging through got
  **page 1 forever**.

## A plumbing bug, not a missing feature

`report_list_by_tenant` has always paginated properly:

```python
select(CrystallizationReport)
    .where(CrystallizationReport.tenant_id == tenant_id)
    .order_by(CrystallizationReport.started_at.desc())
    .offset(offset)
    .limit(limit)
```

The SQL needed no change. The window was simply dropped at each of the
three layers above it, so a caller's page could never reach the query that
implements paging. All three now thread it:

- `routes/crystallizer.py` — passes what it already validates
- `clients/storage_client.py` — forwards it (keyword-only)
- `core-storage-api/routers/reports.py` — accepts and forwards it

Threading only the two ends would have left the same bug with more code;
the storage route was the layer that swallowed it.

## Defaults are deliberate

Every layer defaults to the service's own `limit=10, offset=0`, so a
caller that passes nothing sees exactly the window it saw before. A
different default would silently change existing response sizes.

Bounds stay at the **public edge** only. The storage route is internal and
deliberately unbounded; a test pins the `ge/le` on the core-api side so
removing them cannot quietly allow an unbounded scan.

## Notes

4 of the 6 tests fail with the change reverted. The other two are
intentional non-regression guards — that the SQL already paginates, and
that the edge bounds still exist — so they pass in both states by design.

clean; `mypy` unchanged on both services (162 core-api, 10 core-storage-api).


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #1506, opened 2026-09-14
chore(benchmark): check in the STALE runner so it stops living on one machine

arkash20:
The STALE bench measures whether the store notices a fact it holds has been
superseded — the implicit-conflict case A56/A60 are about. Until now
`stale_memclaw.py` existed only on one laptop, so nobody else could run it,
reproduce a number, or check a claim made from it.

That is the third time this month a decision rested on evidence that lived in
a single terminal (A50's calibration, C1's, and this). A75 is the fourth and
worst: its proof gate was closed **unrun** because the replay harness was not
reachable, and three `default off` flags are still dark as a result.

## What is here, and what deliberately is not

Ours, so checked in:

- `stale_memclaw.py` — seeds each sample's haystack into an isolated fleet,
  recalls the probing queries, emits STALE's answer-record schema
- `runner_memclaw.py` — the MemClaw client it wraps; not optional
- `STALE-README.md` — how to run it, and how to read the result honestly

**Not** here, because it is not ours: the fixtures and the judge belong to the
STALE benchmark (MIT, Hanxiang Chao). `STALE-README.md` gives the clone URL and
the exact expression that rebuilds the 15-sample T2 slice — verified to be the
first 15 `type == "T2"` entries of `data/T1_T2_400_FULL.json`. Upstream also
carries 200 T1 samples, so T1 runs are available; we simply never built a slice.

Vendoring an 11 MB third-party fixture into a public repo would be the wrong
call when four lines reproduce it exactly.

## One change to the runner

`VERTEX_PROJECT` no longer defaults to our GCP project id. This repo is public
and a project id is infrastructure detail, not configuration a reader should
inherit. It is now required from the environment.

## Baseline recorded in the README

T2, 15 samples, judged 2026-08-05: **overall 11/45 = 0.2444**.

Recorded with an explicit warning that **no code SHA accompanies it**, so
scores are comparable but a delta is not attributable. The README asks the next
runner to record the SHA so the comparison after it is clean — which is the
habit whose absence caused this commit.


arkash20:
Closing: the two runner files predate the rename and carry 57+ legacy names, so the ratchet blocks them — correctly. Rewording the prose and floor-marking the literal env vars is a real port (~1h) of a file whose value is that it produced the 2026-08-05 baseline unchanged, so it shouldn't be done as a drive-by.

The reproduction instructions that were in `STALE-README.md` have moved into the A56/A60 reply so nothing is lost. If we want the runner to live here permanently, the rename port should be its own task; `algo-stress` is the better home either way since the fixtures and results don't belong in a public repo.

PR #1507, opened 2026-09-14
fix(plugin): evict the least recently used session, not the oldest one (F1)

arkash20:
The comment said LRU. The code was FIFO:

```js
// LRU eviction: if we have too many sessions, drop the oldest
const oldest = sessionBuffers.keys().next().value!;
```

A `Map` yields keys in **insertion** order, and nothing moved a key on
access — the lookup above is a `get`, and the `set` only runs for a session
that does not exist yet. So at `MAX_SESSIONS` (100) the evicted session was
the longest-**running** one: the session most likely to be mid-conversation,
with the deepest buffer, while 99 newer and possibly idle sessions survived.

That is silent. Losing a session's buffer costs it
`buildQueryFromMessages`, which builds the recall query from the last three
user messages — so the session keeps working and simply recalls worse, with
nothing logged.

Fixed by re-setting the key on every push, which is what the eviction was
always documented to assume.

## How this was found

Re-scoping **F1**, which reads "OpenClaw plugin lacks agent_id / session
management — est 2d". All three claimed gaps exist and work:

| F1 claims missing | actually |
|---|---|
| `agent_id` resolution | `resolve-agent.ts` — `resolveAgentId` / `resolveAgentIdQuiet` |
| credential lifecycle | `agent-auth.ts` — `resolveAgentKey`, `evictAgentKey` |
| session management | `context-engine.ts` — bounded buffers, per-session cap |

The row's file pointer is wrong too: `plugin/src/identity.ts` is 87 lines of
hostname and display-name helpers with no identity logic in it.

This eviction bug is the real residual, and it is ~0.5d rather than 2d.

## Tests

Three cases, added to the existing `context-engine.test.ts` so no new file
needs registering in `package.json`:

- the eviction **choice** — a used session survives, the unused one goes
- the queue **reordering** — using a session moves it to the back
- a re-used session **keeps its buffer**, so the reorder is a `delete`+`set`
  of the same array reference and not a silent data loss

All three fail with the fix reverted (`# pass 41 / # fail 3`) and pass with
it (`# pass 44 / # fail 0`) — checked both directions, not assumed.

`pushToBuffer` is module-private and `ingest()` reaches it only after
`bootstrap()`, so driving this through the public API would mean mocking the
network to assert an in-memory `Map`. The three test-only exports follow the
`_resetBootstrapForTests` precedent already in the file, and carry the same
caveat.


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #1508, opened 2026-09-14
feat(crystallizer): let a tenant retune the sweep into the crowding band (A72)

arkash20:
The crystallizer is meant to be the janitor for overlapping memories: it
clusters near-duplicates, LLM-re-extracts atomic facts, and archives the
sources. It could not see the pathology it is the safety net for, on two
grounds that are both constants:

- **the band** — `CRYSTALLIZER_DEDUP_THRESHOLD = 0.95` catches near-verbatim
  copies, while the composites that actually crowd recall sit around
  0.75-0.90 and pass underneath it untouched;
- **the cluster floor** — `CRYSTALLIZER_MIN_CLUSTER_SIZE = 3` skips the most
  common overlap of all, a **pair**, so two rows saying the same thing
  survive every sweep.

Both are now per-tenant: `crystallizer.dedup_threshold` and
`crystallizer.min_cluster_size`.

## Both default to the existing constants, deliberately

An untouched tenant sweeps the same band, admits the same clusters, and
spends exactly the same number of LLM calls as before this change. A test
pins that.

That default is the point, not timidity. Every extra pair a wider band
admits, and every extra cluster a lower floor admits, is another LLM
mergeability judgement — so this is a bill, and it should be a tenant's
choice rather than something inherited from a deploy. It also means the
change can land while the no-new-LLM-calls hold stands: the code ships, the
cost does not, and the decision stays available.

Same shape as `dedup.merge_near_duplicates`, `write.bulk_subject_batching`
and `search.strict_fleet_scoping`.

## Bounds, and why they are where they are

- band clamped to **[0.5, 1.0]** — below 0.5 the sweep stops being a
  duplicate check and becomes a topic clusterer, handing unrelated rows to
  the re-extractor to be merged. Above 1.0 is not a cosine.
- cluster floor **≥ 2** — a cluster of one is not a cluster; allowing it
  would feed single memories to the LLM re-extractor for no possible merge.

## Scope: two of the three grounds

The report names a third — the once-a-day 02:00 trigger, which a heavy
writing day outruns. That is **not** here. It needs a crystallizer activity
gate in storage (the existing one is insights-specific:
`/insights/activity-gate` compares insight vs non-insight timestamps) plus a
cron cadence change in `core-operations`. Different service, different
review. `reg-a72` stays `partial` and names it.

## Compatibility

Both reads go through `getattr(config, …, <constant>)`, matching
`getattr(tenant_config, "merge_near_duplicates", False)` in
`DetectNearDuplicate` — a config object that predates these knobs (an older
deploy's, or one of the existing `SimpleNamespace` test doubles) resolves to
today's behaviour instead of raising `AttributeError` inside a scheduled
sweep.

The sweep resolves its own threshold rather than taking it from the caller,
because the hygiene checks are dispatched through a uniform
`fn(tenant_id, fleet_id)` loop; special-casing one of them there would put
this knob somewhere nobody looks for it. The parameter stays for tests.

## Notes

13 of the 14 new tests fail with the change reverted. The survivor pins that
the shipped defaults are the values the report criticises, so a future
default change has to be a deliberate act rather than a drift.

Full suite **32 failures, identical to main's 32** — none new. `ruff` clean;
`mypy` unchanged at 162.


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #1509, opened 2026-09-14
feat(crystallizer): activity-gate the sweep and make its cadence configurable (A72, third ground)

arkash20:
A72's remaining ground. The crystallize tick fired **once a day at 02:00**, so
everything written after the tick waited ~24h for the janitor — a heavy
writing day outruns the sweep entirely.

Raising the cadence is the fix on its own terms, and doing only that would be
reckless: a bare increase multiplies LLM spend across every idle tenant, most
of which wrote nothing at all. So the knob ships **with** the gate that makes
it affordable, and neither is much use without the other.

## The gate

`GET`-shaped question, `POST /reports/activity-gate`: *has anything been
written since the last completed sweep?* Two indexed aggregates —
`MAX(memories.created_at)` and `MAX(reports.completed_at)` — returned as two
timestamps, with the comparison left to core-api. That mirrors
`/insights/activity-gate`, which likewise returns values rather than a verdict
so the decision stays readable at the call site.

A tenant with nothing new now costs those two aggregates and returns, never
reaching an LLM.

**`status="completed"` only**, and that is load-bearing. A run reserves its
report with `status="running"` *before* it works, so "the latest report" is
frequently the caller itself, and a crashed run leaves a `running` row behind
forever. Either would advance the watermark past work that never happened and
strand every memory written before it — the same trap `_type_ii_watermark`
documents.

**Fleet-scoped on both sides.** A fleet-scoped sweep compared against a
tenant-wide watermark would skip a fleet that had writes because a *sibling*
fleet was swept more recently.

## Placement is the economy of the change

The gate sits **after** the active-count gate — cheaper still, and it rejects
small corpora outright — and **before** the lazy `crystallizer_service` import,
which drags in the LLM clients this call exists to avoid paying for. A test
pins both orderings; a gate placed after that import would cost most of what it
saves.

## The cadence

`lifecycle_crystallize_every_hours`, default **24** — byte-for-byte today's
behaviour, one run wall-clock aligned to `lifecycle_pipeline_run_at_hour`.

Below 24 the alignment stops meaning anything ("at 02:00" does not describe a
6-hourly job), so the delay provider switches to the next top of hour rather
than silently keeping a daily anchor. Floored at 1 so a zero or negative value
cannot busy-loop the scheduler.

`lifecycle-entity-link` shares that config hour but is **not** part of this
retune and stays a plain daily job — a test pins that, so the knob cannot be
mistaken for a pipeline-wide cadence.

## Completes A72

With #1508 (band + cluster floor) this closes all three grounds the report
names. As there, the defaults change nothing: the cadence is an operator's
decision and the gate can only ever *skip* work the daily run would have done
on an unchanged corpus.

12 tests across the two suites; 11 fail with the change reverted. (The survivor
asserts entity-link stays daily, which is true either way by design.)

Note for reviewers running these locally: `core-operations/tests` and the root
`tests/` are both packages, so a single `pytest` invocation naming files from
both collides on module names. Run them as separate invocations.

Full suite **32 failures, identical to main's 32** — none new. `ruff` clean;
`mypy` unchanged (162 core-api, 10 core-storage-api).


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #1510, opened 2026-09-15
feat(llm): add Atlas Cloud provider

binyangzhu000-sudo:
## Summary
Atlas Cloud is an OpenAI-compatible gateway, so it joins the `_OPENAI_COMPATIBLE` set next to OpenRouter and rides the existing `OpenAILLMProvider` — no new transport, no new dependency. The change is the enum member, the endpoint constants, and the three-tier credential resolution the other providers already use.

- `common/provider_names.py` — `ATLASCLOUD = "atlascloud"`
- `common/llm/constants.py` — `ATLASCLOUD_CHAT_BASE_URL`, `ATLASCLOUD_DEFAULT_MODEL` (overridable via env, same as the OpenRouter default)
- `common/llm/_credentials.py` — `_TENANT_KEY_ATTR` entry, `_env_key` branch, and the `resolve_openai_compatible` branch
- `common/llm/registry.py` — add to `_OPENAI_COMPATIBLE`
- `core-api/config.py` — `atlascloud_api_key` setting and its `ATLASCLOUD_API_KEY` bridge entry
- `core-api/services/organization_settings.py` — tenant-key property, and the provider in the enrichment fallback candidates
- `tests/test_atlascloud_provider.py` — 13 unit tests
- `tests/test_credential_bridge.py` — `ATLASCLOUD_API_KEY` added to the bridge-key snapshot so the bridge's writes are restored like the others

Tenant key → env var → empty resolution works exactly as it does for OpenRouter, so `ResolvedConfig.atlascloud_api_key` overrides `ATLASCLOUD_API_KEY`.

## Validation
- `pytest tests/test_atlascloud_provider.py --noconftest` → **13 passed** (constants, enum membership, `_env_key` set/unset, `has_credentials` with and without a tenant key, tenant-over-env precedence, env fallback, unset behaviour, plus two regression checks that OpenRouter still resolves and unknown providers still return empty)
- live run against the real gateway:

```
resolve_openai_compatible("atlascloud", None)
  -> key=apikey-4e168…  base_url=https://api.atlascloud.ai/v1  model=openai/gpt-4.1-mini
get_llm_provider("atlascloud")
  -> OpenAILLMProvider | model: openai/gpt-4.1-mini | provider_name: atlascloud
await provider.complete_text(...) -> 'caura atlascloud ok'
```

I used `--noconftest` because the session-scoped fixture in `tests/conftest.py` pulls in FastAPI and the rest of the API stack, which I didn't install locally. `tests/test_contradiction_providers.py` fails to collect in my environment for the same reason — **it fails identically on unmodified `main`**, so this change doesn't move it.

No docs, README, or sponsor placement touched — just the provider wiring and its tests.


arkash20:
Heads-up on an interaction with #1522, which is open against the same file — no action needed from you yet, and nothing wrong with this PR as written.

#1522 fixes 09/02 M-11: `resolve_openai_compatible` never accepted `model_attr`, so the per-service model knobs (`contradiction_model`, `recall_model`, …) were inert on every OpenAI-compatible provider. It replaces the `return key, <BASE_URL>, <DEFAULT_MODEL>` ending in each branch with a resolver:

```python
model = _model_for_provider(provider, tenant_config, model_attr, ATLASCLOUD_DEFAULT_MODEL)
return key, ATLASCLOUD_CHAT_BASE_URL, model
```

The two PRs touch different lines, so git will merge them cleanly — but the *result* would be inconsistent: Atlas Cloud would keep ignoring `model_attr` while the other three providers honour it, which is silent and easy to miss.

#1522 carries a structural test asserting every branch resolves through `_model_for_provider`, so whichever of us merges second will see it go red rather than have the gap slip through. Whoever lands first, the other rebases and it's a one-line change.

One thing that may be worth a look independently: `ATLASCLOUD_DEFAULT_MODEL = "openai/gpt-4.1-mini"` is a `vendor/model` id, so it won't be mistaken for another provider's model by #1522's family check (that only matches bare prefixes like `gpt-`, `claude-`, `gemini-`). But if a tenant sets `enrichment_model = "gpt-4o"` while on Atlas Cloud, #1522 would treat it as an OpenAI id and fall back to the Atlas default. If Atlas accepts bare OpenAI model names, tell me and I'll exclude it from that check.

binyangzhu000-sudo:
Thanks @arkash20. Rebased onto current main after #1522 and pushed 615bc76. The Atlas branch now calls `_model_for_provider(provider, tenant_config, model_attr, ATLASCLOUD_DEFAULT_MODEL)`.

On bare IDs: support is model-specific, not universal. The Atlas-side live checks shared with me returned 400/not found for `gpt-4o` and `gemini-3-pro`, but accepted `gpt-4.1-mini` (resolved to `openai/gpt-4.1-mini`) and `claude-sonnet-4-6` (resolved to `anthropic/claude-sonnet-4.6`). I did not repeat those paid calls. So the family-prefix guard would incorrectly replace some valid Atlas choices. I scoped an exemption to Atlas only: preserve the explicit ID and let the API validate availability. The canonical default remains unchanged; other providers retain their existing guard.

Added coverage for tenant/env per-service overrides, those bare aliases, and default fallback. Validation: 54 tests passed across `test_atlascloud_provider.py`, `test_m10_m11_provider_aware_models.py`, and `test_credential_bridge.py` using `pytest --noconftest` to isolate these unit tests from the global PostgreSQL fixture. Focused Ruff lint/format and `git diff --check` passed. The full database-backed suite was not run locally.

arkash20 (approved):

erni-a:
Approved by @arkash20 and I updated the branch so CI could run, but the DCO check fails: the first commit (`3840ebe4 feat(llm): add Atlas Cloud provider`) has no `Signed-off-by` trailer. Branch protection will not let it merge until every commit carries one. `git rebase --signoff origin/main && git push --force-with-lease` fixes it in one step. Ping when pushed and we will merge.

arkash20 (approved):

erni-a:
@binyangzhu000-sudo the only thing between this and main is the sign-off on `3840ebe4`. Everything else is approved and the branch is current. If we have not heard back by Tuesday 23 Sep we will carry the two commits into a maintainer PR with your authorship preserved so the work lands and stays credited to you. Prefer that you push it yourself, since it is a one-liner: `git rebase --signoff origin/main && git push --force-with-lease`.

@arkash20 for visibility, since the approval is yours.

arkash20 (approved):

binyangzhu000-sudo:
Pushed — both commits now carry `Signed-off-by` (`git rebase --signoff origin/main`, force-pushed). Branch is current with main as of the merge you did this morning. Let me know if anything else is needed.

PR #1514, opened 2026-09-15
feat(triple): resolve proper-noun subjects by lookup so the subject column fills (A59)

arkash20:
## Why

A59's Type-II materializer groups live memories by `subject_entity_id` and ignores every row without one. I measured a **real** staging corpus — 777 rows from `arkady-6ab91e`, run through the materializer's own `group_subjects`/`select_candidates` so the numbers are the code's:

| | |
|---|---|
| Rows | 777 |
| Distilled (non-transcript) | 281 |
| Rows with `subject_entity_id` | **10 (1.3%)** |
| Candidate subjects | **0** |

So the instrument is dark — and **not** because the corpus is synthetic. 281 of those rows are exactly the distilled shape the A57 gate asked for. The recorded blocker ("needs real-tenant access") was wrong: handing someone credentials yields 0 proposals, not 30, because subject coverage gates everything upstream.

The identifier heuristic can't close it. Real content names things (`Atlas`, `Project Brightwood`) instead of emitting identifiers, and `_infer_subject_token` returns `None` for names **on purpose** — creating an entity row from a bare name races the background extraction worker and fragments the entity.

## What

A third subject path that keeps that guarantee intact: propose the proper-noun name, then resolve it against entities that **already exist**.

- A repeat mention of a known name fills the subject column.
- A first mention still skips and stays the worker's to create.
- **Nothing is ever created here**, so the extraction worker's precision is untouched.

The bound is also the point: the materializer needs ≥2 live memories on one subject before it considers it, so repeat mentions are the only population it can use anyway.

## Details

- `find_entity_by_exact_name` in `entity_service` — the lookup-only counterpart to `upsert_entity`'s Phase 1, returning just the id.
- `_subject_head` is now shared by both subject readings so they can't drift about where the subject ends — a divergence there would be silent.
- The lookup sits in **Phase B**, after every other SKIP gate, matching the existing deferral so a downstream skip costs no storage round-trip.
- `no_subject_match` is a **distinct** skip reason from `no_subject` and carries the candidate name. The two have opposite remedies: nothing fixes `no_subject`, while `no_subject_match` counts exactly the rows a subject backfill would convert.

**Zero new LLM calls** — a regex plus one indexed read, on rows that would otherwise have skipped.

## Tests

13 added/changed; **all 13 fail with the change reverted**. Resolve-and-emit, tenant+fleet scoping of the lookup, unknown-name skip creating nothing, multi-word names, leading-stopword rejection, identifier precedence, Phase-B deferral, storage-failure degradation, explicit `entity_links` still winning.

`58 passed` in the suite; `ruff check`/`format` clean; `mypy` unchanged at the 162-error baseline (no errors in the changed regions). The 5 failures in the wider run reproduce on pristine `main` (local schema drift + the known route-assertion issue).

## Note

One existing test encoded the old contract (`"Alice …" → no_subject`). That's an intentional behaviour change, so it's updated rather than worked around — it now asserts the thing it was really protecting: `upsert_entity` is never called for a name.


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #1515, opened 2026-09-15
fix(triple): stop reporting sentence openers as missing subject entities (A59)

arkash20:
Follow-up to #1514, fixing a defect I introduced there.

## Why

`no_subject_match` exists to **size** the population a subject backfill would convert. Replaying the merged A59 path over a real 777-row staging corpus recorded `If` and `My` among only **6 distinct candidates** — so the noise was a material fraction of the number that decision rests on.

The miss itself was always harmless (this path only resolves against existing entities and creates nothing), which is exactly why it needed fixing deliberately rather than showing up later as a bug report.

## What

`_NON_NAME_OPENERS` — subordinators, conjunctions, possessives, determiners and fronting auxiliaries — rejected in both the single-word and multi-word checks. `"If Atlas …"` is an opener plus a name, and the pair can only ever miss as one canonical name.

Kept **separate** from `_SUBJECT_STOPWORDS`: that set also gates the identifier path, where these words can't appear anyway, so folding them in would widen a shared gate for one path's benefit.

## Tests

5 added; **4 fail with the change reverted**. The fifth is an over-blocking guard (`"Atlas"` must still resolve) and correctly passes both ways — the risk of any stopword list is that it eats the path it protects.

`63 passed` in the file; `ruff check`/`format` clean.


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #1516, opened 2026-09-15
fix(ingest): stop serving a partial extraction as the cached complete one (09/02 M-44)

arkash20:
## The bug

`ingest_commit` tolerates partial failure: it counts `created` and `errored` and, when facts fail, logs a warning suggesting the operator wipe the batch by `ingest_run_id`. The rows that **did** land still carry `metadata["doc_hash"]`.

So the next preview of the same document found those rows, returned `cached: True`, and served an incomplete extraction as the finished one — **permanently**. The cache short-circuits before any LLM call, so re-previewing, the obvious remedy, was exactly the thing that could not help.

## The fix

The signal already existed and simply wasn't read: the parent Document records `errored` right next to `doc_hash`. `_prior_ingest_was_complete` consults it, and the preview refuses a cache it cannot prove whole.

**A missing parent counts as NOT complete.** The parent write is best-effort (its own handler says so), so absence means "cannot prove this cache is whole" — and refusing to serve an unprovable result is the entire point. The asymmetry justifies it: being wrong this way costs one extraction; being wrong the other way leaves a document permanently missing facts. An absent `errored` field (parents predating it) gets the same treatment.

The gate runs **before** the cache is served — pinned by a test, since placing it after would make it decorative.

## Tests

7 added, **all 7 fail against `main`**: missing parent, absent `errored`, storage failure, non-zero `errored`, preview re-extracting instead of serving a partial, and gate ordering. Plus 6 existing `test_ingest_idempotency_undo` tests still green (13 total).

`ruff check`/`format` clean.

## Note

This was committed 2026-09-14 and sat unpushed; rebased onto current `main` (`34d37a72`) and re-verified before opening.


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #1517, opened 2026-09-15
docs(forge-cron): correct the fanout URL prefix and the dedup window (09/02 M-69, M-70)

arkash20:
Two errors in the operator guide, both of which make a correctly-followed setup silently not work.

## M-69 — the fanout URL omits `/api/v1`

`lifecycle_router` is mounted with `prefix="/api/v1"` (`app.py:1061`) and declares `@router.post("/admin/lifecycle/fanout/{action}")` (`routes/lifecycle.py:172`). The real path is `/api/v1/admin/lifecycle/fanout/forge-distill`.

Both the Cloud Scheduler example and the CronJob curl were missing the prefix, so a copy-pasted scheduler **404s forever** — and since the scheduler is fire-and-forget, it 404s quietly.

## M-70 — the dedup window is 23 hours, not 1

`_PIPELINE_DEDUP_WINDOW_HOURS = 23` (`lifecycle_handlers.py:137`). At "1 hour" the documented 6-hourly schedule reads as six independent distillation passes a day; at 23 hours, ticks 2–4 are expected no-ops.

That one is worth more than a number swap, so the section now explains **why** the schedule oversamples (a missed tick should cost hours, not a day) and warns that shortening the cron interval changes nothing — the window sets the real cadence. Without that, the obvious reaction to "distillation isn't running often enough" is to tighten the cron, which cannot work.

Docs only; no code change.


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #1518, opened 2026-09-15
fix(crystallizer): send duration_ms so the publish-failure recovery can't wedge the report (09/02 M-38)

arkash20:
## The bug

`start_crystallization` reserves a report row as `running`, then publishes the work. If that publish fails, nothing will ever execute the run — so the row blocks every later trigger for that tenant until the staleness cutoff expires. The `except` block exists precisely to mark it terminal.

That block omitted `duration_ms`. Storage's PATCH handler reads it with a **bare subscript** — `duration_ms=body["duration_ms"]` (`routers/reports.py:205`) — while every neighbour (`summary`, `hygiene`, `health`, `usage_data`, `issues`) uses `.get` with a default.

So the recovery write raised `KeyError` **inside the recovery itself**, the inner best-effort guard swallowed it, and the report stayed `running` forever: the exact outcome the block was written to prevent, reached through the code that prevents it.

## The fix

Send `"duration_ms": 0`. Zero is the honest value — the publish never landed, so there was no run to time.

**Fixed on the caller, not by making storage lenient.** The caller is what violates a documented contract (the handler's own docstring lists `duration_ms` in the body), and loosening the subscript would let any future caller omit it silently. A test pins storage's strictness too, so if that ever changes this fix gets reconsidered rather than quietly outliving its reason.

The two completion paths that always sent a real elapsed time are untouched, and a test pins that so this can't flatten them later.

## Tests

4 added; **1 fails against `main`**. The other 3 are deliberate guards — storage strictness, the two untouched writes, and the terminal-status contract — and correctly pass both ways.

`ruff check`/`format` clean.


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #1519, opened 2026-09-15
fix(scripts): route forge_dry_run through storage instead of the deleted DB pool (09/02 M-07)

arkash20:
## The bug

`scripts/forge_dry_run.py` did `from core_api.db.session import async_session`. **#491 deleted that module** when it routed every core-api DB access through core-storage-api — `core-api/src/core_api/db/` has held nothing but `__pycache__` since. Every real invocation died on `ImportError`.

The import is **lazy** (inside `_run`, deliberately, for `--help` ergonomics on dependency-light envs), which is why this survived: the file parsed, `--help` worked, ruff passed, and only an operator actually running a dry-run ever saw it break.

## This is a move, not a redesign

#491 added the forge-shaped endpoints the script needs at the same time it removed the pool. The script was just never carried over:

| Helper | Now uses |
|---|---|
| `_wire_memory_fetcher` | `forge_memory_content_by_ids` — the index-friendly `id = ANY(CAST(:ids AS uuid[]))` reasoning its old comment preserved now lives server-side |
| `_wire_poison_checker` | `forge_is_fingerprint_poisoned` — same `cooloff_days` predicate, applied behind `/forge/rejected-fingerprints/check` |

`_wire_candidate_writer` and `_wire_status_checker` needed nothing: both took a `db` parameter they **never used** and were already storage-routed. Dropped, so the signatures stop implying a dependency the bodies don't have.

## One real consequence, called out in the code

There's no session and no `db.commit()` — each storage call commits server-side. That costs the old all-or-nothing property: a failure midway now leaves earlier candidate writes persisted.

Acceptable **here specifically** because Forge is idempotent against `session_traces` (a re-run re-upserts rather than duplicating), which is the same bet the storage-routed production path already makes. Flagging it rather than burying it.

## Tests

7 added; **4 fail against `main`**, 3 are deliberate guards.

They read **AST-stripped** source, not raw text — the fix is described in comments that name the very symbols it removes, so a substring search would pass on the prose alone. One guard pins that `core_api/db/` is still empty, so if core-api ever regains a pool this approach gets revisited rather than quietly outliving its reason.

Verified `--help` still works and the module loads cleanly. `ruff check`/`format` clean.


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

erni-a (approved):

PR #1520, opened 2026-09-15
fix(llm): cache OpenAI providers so every call stops minting an unclosed pool (09/02 M-36)

arkash20:
## The leak

`OpenAILLMProvider.__init__` builds an `httpx.AsyncClient` with its own connection pool (CAURA-627 sized it for bulk-write fan-out). `get_llm_provider` is called on **every LLM call** — `common/llm/retry.py` invokes `provider_factory` for the primary provider and again for the fallback — so each call minted a fresh pool, and nothing ever closed it.

`OpenAILLMProvider.aclose` already existed, and its own docstring names this exact failure:

> a leak in long-lived processes that rotate client instances

It had **no caller anywhere in the repo**. This gives it one.

## Scope

Only the OpenAI-compatible branch is cached — it's the only provider that owns a client. Gemini and Fake hold none, so caching them would buy nothing and add a lifetime to reason about.

## The key is the whole configuration, deliberately

`request_timeout` is read from `os.environ` at construction (the call site documents why it must not use the import-time constant); api key and model come from tenant config. All of them are in the key, so behaviour is preserved exactly: change any one and you get a new provider, as before.

A key of just `name` would silently pin the first tenant's credentials for the life of the process — **a cross-tenant credential leak, not a perf bug**.

The api key is **hashed, not stored**: the cache outlives any single request, and a value that only has to be compared never needs to be kept.

## Eviction

Bounded at 32, LRU — a many-tenant process can't grow pools without limit, and hot tenants keep theirs. Evicted providers are **closed**: `get_llm_provider` is sync, so the close is scheduled on the running loop (every production caller is async) and skipped when there's none, which is exactly today's behaviour and no worse. The task is held in a strong reference until it completes, since a bare `create_task` result can be collected mid-flight and cancel the close.

## Test isolation

Added an autouse `_reset_llm_provider_cache` fixture in `conftest.py`, same shape as the existing `_patch_storage_client` (`sc_mod._client`) and `_reset_hooks` fixtures — a module singleton production wants and test isolation doesn't.

## Tests

12 added: reuse, per-model/per-key/per-timeout separation, the bound, LRU-not-FIFO eviction, that eviction actually closes, that a missing loop and a provider without `aclose` are both tolerated, and that the raw key never lands in the cache key.

Verified against baseline: `main` gives **696 passed / 0 failed** on the `llm or provider or registry or gemini or enrich or judge` selection; this branch gives **708 passed / 0 failed** (696 + 12).

### One thing worth flagging

My first version of the test file used `asyncio.run()` inside a sync test. That tears down the loop pytest-asyncio manages, and **81 unrelated async tests** started reporting "coroutine was never awaited". The registry change was clean the whole time — the test file was the problem. It's now a proper async test, and the reasoning is recorded in its docstring so nobody reintroduces it.


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #1521, opened 2026-09-15
fix(governance): stop the SSN rule matching invoice numbers and ZIP+4 (09/02 M-03)

arkash20:
## The bug

The rule was `\d{3}[- ]?\d{2}[- ]?\d{4}`. Both separators were optional and, crucially, **independent of each other** — so it matched three things an SSN is not:

| Input | Matched | What it actually is |
|---|---|---|
| `Invoice 100234567 paid` | ✅ | invoice number |
| `port 8080 pid 123456789` | ✅ | process id |
| `build 202409158 green` | ✅ | build number |
| `12345-6789` | ✅ | **ZIP+4** — first separator absent, second present |
| `123-456789` | ✅ | mismatched separators |

This isn't a cosmetic over-match. The drop policy **422s a legitimate write**, and the mask policy **rewrites stored content** — so a false positive silently corrupts a memory that merely mentioned an order number, and it's unrecoverable because the original text is gone.

## The fix

An SSN has no checksum (unlike Luhn for cards, mod-97 for IBAN), so format and context are the only signals. Two rules:

**1. Separated form** — a backreference requires the **same** separator in both positions. A real SSN is `123-45-6789` or `123 45 6789`, never `12345-6789`. That's exactly what kills ZIP+4 and 5/4 splits. Distinctive enough to stand alone, so no cue needed.

**2. Bare nine-digit form** — matched only next to a cue naming it (`SSN`, `social security number`, `soc sec`, …). Nine bare digits are genuinely ambiguous to a human reader too, so requiring the cue isn't a weakened rule — it's the only honest one.

`group=1` on the second rule redacts just the digits and leaves the cue intact, so a masked memory still reads `SSN <redacted>` instead of losing the sentence that gave it meaning — same reason the `key=<value>` rules use `group`.

Known-invalid areas (`000`, `666`, `9xx`), groups (`00`) and serials (`0000`) are excluded exactly as before.

## Tests

26 added; **8 fail against `main`**. Both forms, the cue-word boundary, every false positive above, and the invalid ranges. Existing suites green: `test_governance_patterns`, `test_sentinel_scanner`, `test_audit_chain` (112 passed), plus 417 passed across the wider `governance/sentinel/pii/audit/redact/mask` selection.

One case is in there because it caught me: I first wrote `987-65-4321` as a *positive* example, and `9xx` is the ITIN range, not a valid SSN area. The suite rejected it correctly, and it's now an explicit negative case.


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #1522, opened 2026-09-15
fix(llm): resolve models per provider so the per-service knobs work and Gemini stops 404ing (09/02 M-10, M-11)

arkash20:
`enrichment_model` / `contradiction_model` / `recall_model` are **shared** tenant settings: one value, read by whichever provider happens to be active. Two opposite defects fell out of that, and they have one fix.

## M-11 — the knobs were inert

`resolve_openai_compatible` never accepted `model_attr` at all, so every per-service model setting resolved to the provider default on every OpenAI-compatible provider. A tenant could set `contradiction_model` and nothing would read it — **config that lies**. Only the Gemini branch honoured it.

## M-10 — the reverse

Gemini *does* read the shared attribute. So a tenant who configured `enrichment_model = "gpt-5.4-nano"` for OpenAI and then switched provider handed Gemini an OpenAI model id — a **404 on every call**, which is why the documented Gemini setup never worked.

## The fix

`resolve_openai_compatible` now takes `model_attr` (mirroring `resolve_gemini_config`), **and the registry forwards it** — the parameter is inert without that seam, which is exactly how the original defect survived. A test pins the seam.

Both branches resolve through `_model_for_provider`, which discards a model id that **confidently** belongs to another provider family and falls back to the target's default.

**Deliberately conservative:** it overrides only on a positive prefix match against another family, never on an unrecognised name. Custom deployments, fine-tunes and OpenRouter's `vendor/model` ids keep passing through untouched — "not recognised" means "leave it alone".

The value must also be a non-empty **string**. `getattr` on a loosely-typed config can yield a Mock, a sentinel or an int, and a non-string model id would only fail deeper inside the provider SDK where the cause is far harder to see. (An existing test using `MagicMock` caught this — I added the guard rather than editing the test.)

The override logs at WARNING naming **both** the rejected model and the substitute. Silently replacing an operator's explicit setting would trade a 404 for a mystery.

The principled fix is provider-scoped model settings; that's a config migration. This is the contained one.

## Tests

19 added. Per-service resolution, two services resolving different models for one tenant, unset knobs still defaulting, the registry seam, both cross-provider directions, genuine ids surviving, unrecognised ids passing through, family detection, the non-string guard, and the warning contents.

824 passed / 0 failed across the `llm or provider or registry or gemini or credential or contradiction` selection.


arkash20:
@Eldad-Caura please approve

Eldad-Caura (dismissed):

arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #1523, opened 2026-09-15
fix(ingest): enforce the token cap the paragraph splitter promises (09/02 M-41)

arkash20:
## The bug

`_split_oversized_paragraph` documents *"one or more substrings, each under `hard_tokens` tokens"*. It packs **sentences**, which can only split **between** them — so it had nothing to offer for a single sentence already over the cap, or for text pysbd finds no boundaries in (a minified blob, a CSV line, a wall of prose with no terminators).

Measured on `main`, against a 3,000-token cap:

| Input | Tokens in | Parts out | Largest part |
|---|---|---|---|
| One enormous sentence | 12,001 | 1 | **12,000** |
| Single sentence + period | 12,001 | 1 | **12,001** |
| No sentence boundaries | 20,001 | 1 | **20,001** |

The contract wasn't approximate, it was absent.

## The fix

`_hard_split_on_tokens` is the last resort below the last resort — it runs **only** on pieces sentence packing left oversized, so ordinary prose is still cut at sentence boundaries and nothing about its handling changes.

It splits on the tiktoken encoding rather than on words, because the cap is counted in tokens: a word split still overshoots on text that tokenises densely (CJK, base64, long identifiers) — exactly the shape that reaches this path.

### Two details I got wrong first, and the tests caught

**Mid-character cuts.** One CJK character spans several cl100k_base tokens, so cutting between tokens can land mid-character and `decode` yields **U+FFFD** — silently corrupting the text this function exists to preserve. My first version's docstring described a round-trip guard the code didn't implement; the losslessness test caught it. Pieces are now built from `decode_bytes`, with trailing bytes that don't yet form a complete UTF-8 sequence carried into the next piece.

**Carry pushes the next slice over.** Those carried bytes complete a character in the next slice, and completing it can merge into a token — so a full-width slice plus a carry re-encodes marginally **over** the cap. A four-token headroom fixes it. Four tokens against a cap in the thousands is noise; being over by one is a broken contract.

## Tests

22 added, across six inputs chosen for how they **tokenise** rather than how they read — no-terminator prose, a single over-cap sentence, no boundaries at all, CJK, emoji, Arabic. Each checked three ways: no piece over the cap, no content lost, no replacement characters introduced.

Plus guards that normal prose still splits on sentences, under-cap text is returned untouched, enforcement runs *after* packing, and a degenerate cap of 1 still terminates.

**8 fail against `main`.** Wider suite green: 259 passed on the `ingest or chunk or section` selection.

One test reads the function with docstrings stripped via AST — the docstring *names* `_hard_split_on_tokens` before the loop that must precede it, so a plain substring search asserts the wrong thing. It did, when I first wrote it.


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #1525, opened 2026-09-15
fix(lifecycle): bound the fanout across actions, and stop hiding dropped orgs

Eldad-Caura:
## The bug

The fanout's concurrency cap was constructed **inside** the request handler:

```python
sem = asyncio.Semaphore(_FANOUT_CONCURRENCY)   # per request
```

so every in-flight fanout got a budget of its own. The scheduler fires each lifecycle action on the same cron minute — eight of them — so the real ceiling was `8 × _FANOUT_CONCURRENCY`, while the comment above the constant claimed:

> *"50 is a generous default — … low enough that the storage-writer pool is never saturated by fanout traffic alone"*

That reasoned about a single fanout. Nothing bounded the aggregate, and the aggregate is what the writer sees.

## What it cost

Measured 2026-09-15 on a deployment whose writer held **120** warm request slots against a ceiling of **8 × 50 = 400**:

| | |
|---|---|
| Orgs dropped | **38 distinct**, across 5 actions |
| Trigger failures | 41 — crystallize 20, insights 9, archive-expired 7, entity-link 3, archive-stale 2 |
| Platform response | `no available instance` → 500 → `audit_begin` raises |

Those orgs' lifecycle actions did not run **and left no audit row**. So every downstream counter stayed self-consistent and the shortfall was invisible to all of them — a row that is never created cannot be counted as failed. It was also not the instance ceiling (peaked 31 of max 50) or the database (212 of 1000 connections).

## Changes

**1. Process-wide, env-tunable budget.** `LIFECYCLE_FANOUT_CONCURRENCY`, default unchanged at 50. Tunable because the right number is a property of the *operator's* storage tier, not of this code — it wants to stay under core-storage-api's warm capacity, which differs per deployment. Keyed by running loop rather than a bare module-level `Semaphore`, which binds to the first loop that awaits it and would raise *"bound to a different event loop"* in any suite running more than one.

**2. core-operations stops discarding the failure count.** The endpoint returns `{"published", "failed"}` and answers 200 even when orgs were dropped (one bad org must not abort the rest) — but the tick logged only `published`, so a partial sweep read as clean. It reported success for every action while 38 orgs went unprocessed. The count was in the response the whole time and nothing read it. Now logs at `error` when non-zero.

## Deliberately not done

Opting `create_lifecycle_audit_row` into `_post`'s `idempotent=True` retry path. That contract requires storage-side replay dedup, and `lifecycle_audit_create` is an unconditional `INSERT` with no dedup key — a retry after a lost response would mint an orphan `pending` row, which is precisely the marker `_trigger_one` reserves for *"the publish failed"*. A retry path needs an idempotency key first (there's precedent in `/audit-logs/bulk` on `client_event_id`); that's a separate change.

## Tests

Three, asserting **behaviour rather than shape**. The concurrency test clears the semaphore store only when present and asserts on *peak simultaneous triggers*, so the same assertion runs against both the old and new forms. Verified it fails without the fix for the right reason:

```
AssertionError: two concurrent fanouts reached 4 simultaneous triggers
against a budget of 2 — the cap is per-request again, so N actions on one
cron minute still oversubscribe storage by N times
```

An earlier draft patched `_FANOUT_SEMAPHORES` directly and failed pre-fix with `AttributeError` — proving only that a name exists, not that concurrency is bounded. That was rewritten.

`ruff check` and `ruff format --check` clean at CI's scopes (`core-api/src/`, `core-operations/src/ core-operations/tests/`, `tests/`); 56 tests pass across the adjacent lifecycle suites.


Eldad-Caura:
@claude Two of three applied. The High one does not reproduce — detail below, since it's the one that would have changed the fix.

### 1. uvloop weakref — not reproducible, keeping `WeakKeyDictionary`

You were right that uvloop is the production loop (locked via `uvicorn[standard]`, and `common/serve.py:136` calls `uvicorn.run(...)` with no `loop=`, so it takes the `"auto"` default). But `uvloop.Loop` does support weak references. Ran your suggested check against the exact pinned version:

```
$ uvx --with uvloop==0.22.1 python -c "..."
weakref.ref(uvloop.Loop): OK -> <weakref at 0x1053e7d30; to 'uvloop.Loop' at 0x74b554020>
WeakKeyDictionary[uvloop.Loop] = ...: OK
```

Being a Cython `cdef class` doesn't preclude weakref support — it depends on whether the class declares `cdef object __weakref__`, and uvloop's does. So there's no `TypeError` and no fanout outage. Swapping to an `id(loop)`-keyed dict would have traded a collectable key for one that can be recycled by the allocator after a loop is freed, which is strictly worse here.

I've recorded the verification in the comment rather than just leaving the line, because it's a reasonable thing to wonder about and the answer isn't visible from reading it.

### 2. Bounds validation — applied

Real, and `0` is the nasty one precisely because it looks harmless:

```
Semaphore(-1): ValueError: Semaphore initial value must be >= 0
Semaphore(0):  constructed OK, locked = True   <- parks every fanout forever
```

Now parsed through `_resolve_fanout_concurrency()` and range-checked at import. Extracted as a function rather than an inline `if` so it's testable without `importlib.reload` on a module the app already holds a router from. Three new cases (`0`, `-1`, `-50`) plus a positive-override case.

### 3. Multi-worker ceiling — applied, and you undersold it

Marked Low; I think it's the most important of the three, because it undercut my own headline. `--workers 2` confirmed in `core-api/Dockerfile:144`, and the service scales horizontally, so the real ceiling is `instances × workers × _FANOUT_CONCURRENCY`. Each fanout POST lands on one worker — so in the worst case of N actions landing on N distinct workers, **this fix changes nothing at all**.

It's still strictly better than a per-request budget (which bounded nothing even within one worker), but "400 becomes 50" was wrong and I've rewritten both the comment and the commit message to say "predictable per worker" instead, with the arithmetic operators need for sizing and a note that a true cross-process cap needs shared state.

---

`ruff check` and `ruff format --check` clean at all three CI scopes; 7 tests in the new file, 53 green across adjacent lifecycle suites.

Eldad-Caura:
@claude Applied in full — this one was a straight reuse miss on my part, and the blast-radius argument is the right one.

Verified all three claims before changing anything:

- `common/env_utils.py:19` — `read_int_env(name, default, *, minimum: int = 1)`, so `0` and negatives are already rejected by the default floor.
- `core-api/src/core_api/app.py:61` — `from core_api.routes.lifecycle import router as lifecycle_router`, module scope, no `try`/`except`.
- **26** existing call sites across `core-api/src` and `common/`. It is unambiguously the convention and I should have found it.

The helper's own docstring names my exact hazard, which is the part that settles it:

> *"semaphore-style caps read 0 / negative as 'block forever' … so the floor defaults to 1"*

It was written for this case. Hand-rolling a second parser next to it was wrong twice over — duplicated logic, and traded a safe degrade for an import-time abort on a cron tunable.

Changes: `_resolve_fanout_concurrency` deleted, now `read_int_env("LIFECYCLE_FANOUT_CONCURRENCY", 50)`; unused `os` import dropped from the module.

### On the test

I didn't just delete the validation tests. The property worth protecting is the one your finding is about — *a typo must not take down the API* — and that lives at module scope, so it needs a reload to reach. Replaced the three function-level cases with one test that reloads the module under `"5o"`, `"0"` and `"12"`, and restores it in a `finally`.

It fails against the version you reviewed, for the right reason rather than on a missing symbol:

```
lifecycle.py:115: ValueError: invalid literal for int() with base 10: '5o'
```

That is the import-time crash you described, reproduced.

`ruff check` and `ruff format --check` clean at all three CI scopes; 57 tests green across this file and the adjacent lifecycle suites.

Eldad-Caura:
@erni-a plear approve

erni-a (approved):

PR #1528, opened 2026-09-15
fix(ingest): actually embed the parent batch summary (09/02 M-45)

arkash20:
## The bug

`_write_parent_ingest_document` built a summary of the batch and set `data["summary"]`, with the comment *"triggers embedding population in storage"*. It does not — for two independent reasons:

1. **Storage computes no vector of its own.** `document_upsert` takes `embedding` from the caller, and `POST /documents` has **no `embedding` parameter at all** — only `POST /documents/upsert-xmax` does.
2. **The parent doc bypasses the embedding path.** It's written through `sc.upsert_document` directly, so it never reaches the REST/MCP doc path that calls `resolve_embed_source` — a fact `doc_indexing`'s own docstring states about server-written collections.

So the summary was stored as text and never indexed. The promised semantic search over ingest batches could not work, for any batch, ever.

## The fix

Embed the summary and route the payload to the endpoint that can carry a vector — the same if/else `routes/documents.py` already makes, for exactly this reason.

My first version attached the vector to the **plain** endpoint, which drops it on the floor with a 200 and no warning. A test now pins that `document_upsert` has no `embedding` parameter, so if storage ever grows one the switch gets simplified rather than quietly outliving its reason.

### Two deliberate differences from the REST doc route

- **`background=True`.** There the client blocks on the write and gets a 502 when the vector is missing, so it must not sit on the reduced deferred budget. Nobody waits on this one.
- **A failure degrades instead of raising.** The ingest has already committed by this point and the parent write is explicitly best-effort — losing the batch over its index entry would trade a missing search result for lost memories. Both the throw and the no-vector case log a warning naming the `run_id`.

### One regression the existing tests caught

`tenant_config` is threaded in from `ingest_commit` rather than re-resolved. That call is documented as happening **exactly once**, to pre-warm the cache so the per-fact pipeline doesn't race on the shared session. My first version added a second call and `test_resolve_config_called_once_before_loop` caught it. The pre-warm result is now kept instead of discarded.

## Cost

One embedding per ingest **batch**, not per fact — and it restores a documented feature rather than adding one.

## Tests

8 added; **6 fail against `main`**. Both storage endpoints are now stubbed in the existing `test_ingest_commit` double, since the parent write picks between them and stubbing only the plain one would let a vector-dropping regression pass green.

449 passed / 0 failed on the `ingest or document or doc_` selection.


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #1530, opened 2026-09-15
fix(lifecycle): nack a flaked skip-write instead of stranding the audit row

Eldad-Caura:
## The bug

The dedup-skip branch acked the delivery whether or not it managed to write the audit row:

```python
try:
    await adapter.update_lifecycle_audit_row(audit_id, ..., status="success",
                                             stats={"skipped": True, "reason": "recent_success"})
except Exception:
    logger.warning("lifecycle audit skip update failed; acking anyway", ...)
# ...falls through, returns, acks
```

The comment justified it: the skip record is "observability data", and raising risked DLQing a legitimate skip.

But that trades a **bounded, visible** failure for an **unbounded invisible** one. When the write fails and we ack, nothing retries the row — no redelivery, no reconciler, no sweep. It sits at `pending` forever.

## What it cost

Prod, 2026-09-15. Audit **73668** (org `debug-graph-26f2ad`) logged both lines in the same second at 02:00:34Z:

```
info  lifecycle crystallize skipped — recent successful run exists
warn  lifecycle audit skip update failed; acking anyway
```

That single stranded row failed the promote smoke at 13:58Z — **32 pass, 1 fail** — because the gate asserts every audit row in its 30h window reached `success`. The org's work had in fact already been done. A no-op skip took down a production deploy gate twelve hours after the fact.

## The fix

The failure path in this same function already states the correct rule for itself:

> *"Acking is only defensible because the `failure` row IS the durable record. When the write failed there is no record, so fall through to the raise and let redelivery have another go at producing one — one wasted retry is cheaper than a row stuck in_progress forever."*

This makes the skip path agree with it: `raise` when the write fails, so the bus nacks.

**Redelivery is safe here precisely because it is the skip path** — a redelivered message re-checks the dedup gate and skips again, so the retry costs one gate query and changes nothing. If the writer stays down the message reaches the DLQ, which is bounded and alertable; a stranded row is neither.

## Test

`test_skip_write_failure_nacks_instead_of_stranding_the_row` — confirmed to **fail without the fix** (1 failed) and pass with it (18 passed, full file). It asserts the raise propagates, that the primitive still never runs, and that exactly one write was attempted and it was the skip record.

## Scope

This is **not** the root cause of that night's incident. The writes failed because `core-storage-writer` was returning 500s under an unbounded fanout — #1525 addresses that, and also explains the 40 orgs dropped entirely that night when `create_lifecycle_audit_row` got the same 500. This PR is only about what the handler does when a write fails, regardless of why.

Two related gaps found while diagnosing, not fixed here:
- `core-storage-writer` returned those 500s and logged **no error at all** — no traceback reached Datadog, which is why the cause is still unknown.
- The smoke gate's zero tolerance and the handler's best-effort writes remain philosophically opposed; this PR resolves it by making the write not best-effort, but the gate has no notion of "pending but recent".


Eldad-Caura:
@erni-a plsae approve

erni-a (approved):

PR #1531, opened 2026-09-15
fix(reports): filter the canonical smoke-probe title too

Eldad-Caura:
The plugin's self-test writes a memory titled `memclaw-smoke-<ts>`, and `report_corpus.py` filters that prefix so ~200 probe facts/day never reach a per-agent report — neither as a row nor as a rollup parent. `do_not_touch_sentinel.py` pins both ends, under this header:

> `# -- The smoke probe: an emitter and a matcher that must move together. ----`

## They cannot move together

Whichever moves first decides what breaks:

| order | result |
|---|---|
| **matcher first** (this PR) | both spellings filtered — **no window**. Additive; nothing that was excluded stops being excluded. |
| emitter first | every probe fact written before the matcher catches up is counted as **real agent work in customer-visible reports**, with nothing failing to signal it |

So the matcher leads:

```diff
- r"gateway (active|reachable)|cache refresh|memclaw-smoke)"
+ r"gateway (active|reachable)|cache refresh|(memclaw|caura)-smoke)"
```

The nested group is harmless — the regex is consumed by `.search()` for truthiness only, and `gateway (active|reachable)` already sets that precedent.

## The sentinel changes in the same commit

Both its protected string and its section header, since the header stated an invariant that's now wrong. The new one says what actually holds: the matcher accepts both spellings, and dropping either alternative before the emitter has moved floods every report. That keeps the guard meaningful instead of merely passing.

## Verification

- The new assertion **fails without the widened regex** (`test_health_check_probe_dropped_as_noise`); 9 passed after.
- `ruff check core-api/src` clean; all three touched files already formatted (the sentinel script is checked read-only, never formatted).
- Do-not-touch sentinel 35/35.

Wave B of `COMPATIBILITY-REGISTER.md` → `persisted-agent-source-ids`. Moving the emitter in `plugin/src/context-engine.ts` is a separate change that this unblocks.


Eldad-Caura:
@erni-a plasr approve

erni-a (approved):

PR #1533, opened 2026-09-16
fix(ingest): stop minting server-reserved memory types from caller content (09/02 M-42)

arkash20:
## The bug

`MEMORY_TYPES_WRITE` is the set a caller may **create** — the full vocabulary minus server-reserved types (`outcome`, `insight`, `rule`, *"authored only by internal flows and rejected at the write boundary"*) and classifier-deprecated ones (`semantic`, `intention`, `commitment`, `cancellation`).

`ingest_commit` validated `suggested_type` against `MEMORY_TYPES` — **all fourteen** — so those seven passed straight through to `memory_type=fact.suggested_type`. Auto-chunk children minted reserved-type rows from caller content: the exact thing `MEMORY_TYPES_WRITE` exists to prevent.

**And the prompt actively asked for them.** It offered five non-writeable types and gave `outcome` preferential instruction:

> `outcome` — past event/result; if you'd write "X happened" or "Y was completed", use this (not "fact")

So the system was reliably producing values its own write boundary forbids.

## The fix

**The prompt** now offers exactly `MEMORY_TYPES_WRITE`. The past-event guidance is redirected to `action`, and the definitional/conceptual case folded into `fact` — where the classifier already puts `semantic`.

**The boundary** coerces anything outside `MEMORY_TYPES_WRITE` to the default, so it doesn't depend on the prompt behaving.

Coerced rather than rejected, deliberately: the prompt itself produced these values, so a 422 would reject the server's **own prior output** and break every preview generated before this fix that's still being round-tripped. Same rule and same reason as `crystallizer_service` (CAURA-717) — a stray LLM completion must not smuggle a reserved slug past the write pipeline.

The pre-existing 422 gate is untouched and **still runs first**: coercion covers the band of valid-but-not-writeable types, while a slug outside the vocabulary entirely is malformed input and still gets a clean 422 rather than being silently turned into a fact. A test pins that ordering.

Every coercion is logged with the offending values — a silently retyped memory is indistinguishable from one the model classified that way.

## Tests

14 added; **10 fail against `main`**.

The prompt's offered set is pinned as **equality** to `MEMORY_TYPES_WRITE`, not a subset — offering a forbidden type reliably produces it, and omitting a writeable one silently narrows extraction. Plus the coercion, the gate ordering, the logging, that the sets still exclude what they claim, and that the coercion target is itself writeable.

473 passed / 0 failed on the `ingest or chunk or memory_type or enrich` selection.


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #1537, opened 2026-09-16
fix(outcome): window contradiction evidence on the status flip, not creation (09/02 M-55)

arkash20:
## The bug

A contradiction is a status **flip** on an existing memory — and that event had no timestamp anywhere on the row. `memories` carries `created_at`, `deleted_at`, `expires_at`, `last_recalled_at` and `last_dedup_checked_at`, and **no `updated_at`**.

So `outcome_contradiction_signals` windowed on `created_at`. A memory written weeks ago and contradicted **today** fell outside the current scan window and its failure evidence was dropped — **silently**, because "no rows" and "no contradictions" are the same answer to the caller. The signal under-reported without ever erroring.

Both docstrings already described the intended behaviour:

> the extractor: *"a memory that's been around for weeks but only just got contradicted SHOULD still register a firing inside the current scan window"*
> storage: names the missing column as a `CAURA-future` and calls the `created_at` window an approximation

Migration 045 adds it.

## The fix

`COALESCE(status_changed_at, created_at)` everywhere it's read.

The column is **nullable and deliberately not backfilled**. There is no source of truth for when a historical row's status changed, and inventing one — `created_at`, or the migration timestamp — would fabricate evidence and shift old failures into whatever window happened to run next. The fallback reproduces today's behaviour exactly for every pre-existing row, which is what makes 045 safe without a backfill.

**Both write paths stamp it.** `memory_update_status` is the dedicated one, but `memory_update` also accepts `status` in its patch (its docstring lists it), so a contradiction applied through there would otherwise stay invisible. An explicit `status_changed_at` in a patch wins, so a caller replaying a known transition can supply the real time.

**The index is partial** — `(tenant_id, status_changed_at) WHERE status IN ('outdated','conflicted')`, the only rows this column is ever read for and a small fraction of the table. A full index would cost write throughput on every status change for rows no reader looks at. Built `CONCURRENTLY` inside an `autocommit_block`, preceded by `DROP INDEX CONCURRENTLY IF EXISTS` to clear an INVALID index left by an interrupted build — matching 040/041.

## Supersession is deliberately unchanged

The row pairs the two signals, but they're not the same problem. For a supersession the event **is** the creation of the superseding memory, so `new_mem.created_at` is already the correct timestamp — adding a transition column there would describe the wrong moment. A test pins that it still windows on the new memory.

## Tests

11 added; **9 fail against `main`**. The column and its nullability, both write paths, the explicit-timestamp override, the query window, `observed_at`, the pre-existing-row fallback, the untouched supersession query, and the migration's index shape.

486 passed / 0 failed on the `outcome/trace/memory_update/contradiction/supersed` selection.

**Note on local runs:** a persistent local test DB will fail 28 tests with `column memories.status_changed_at does not exist`. `conftest` uses `Base.metadata.create_all`, which only creates missing *tables* and never adds columns to an existing one. CI builds fresh so it's unaffected; locally, `ALTER TABLE memories ADD COLUMN IF NOT EXISTS status_changed_at TIMESTAMPTZ` clears it.


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #1539, opened 2026-09-16
fix(forge): don't re-buy the mining half when promotion blips (09/02 L-34)

arkash20:
## The bug

`run_forge_cron_tick` has two halves with very different costs:

| Half | Cost | Durability |
|---|---|---|
| **mining** (`run_forge_distill`) | spends LLM calls | candidates are durable the moment they land |
| **promotion** (`promote_pending_candidates`) | cheap status flips | operates on already-written rows |

Promotion was awaited **unguarded**, so any exception propagated out of the tick. The bus **redelivers** a failed tick — which re-runs the mining half and **re-pays its LLM cost**, for candidates that were already written and still sitting there.

A transient storage blip in the cheap, retryable half is not a reason to buy the expensive, already-completed half twice. Held candidates are designed to surface on a later tick anyway — *"they surface on the next tick if conditions change"*, per the comment directly above the call — so deferring them costs a cron interval and nothing else.

## The asymmetry is deliberate

The mining half stays **unguarded**. Its failures *should* fail the tick: nothing was written worth keeping, and the docstring's `PermanentOpError` contract depends on them propagating. A test pins that `run_forge_distill` is not swallowed.

## Reporting

Same shape as the interview sweep's jobs half (`jobs_sweep_ok` / `jobs_sweep_error`): keep the successful tick, but say which of the two happened.

Without `promotion_ok`, a promotion raising on **every** tick returns the same zeros as one with nothing to promote — and the mining counters still look healthy, so the tick reads fine indefinitely.

Only the exception **type** is surfaced. The full exception is in the log line; this string lands in an audit row / log field, where some `__str__` implementations carry hostnames, URLs or request fragments — the same reasoning the interview sweep records.

The failure path builds a **zeroed `PromoterRunResult`** rather than omitting keys, so `stats` keeps one shape whether the half ran or not. A test pins that the stand-in matches the real type: one that drifted would raise inside the stats dict and turn a handled failure back into a tick failure — reintroducing the bug through the fix.

## Tests

7 added; **5 fail against `main`**. Read via AST rather than substring, since the explaining comment names the symbols involved.

573 passed on the `forge or skill or promot or lifecycle` selection. The 2 failures in `test_lifecycle_fanout_aggregate_cap.py` reproduce on pristine `main` and are unrelated.


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #1540, opened 2026-09-16
fix(operations): wait out core-api's request budget instead of racing it

Eldad-Caura:
core-operations gave up on its core-api calls after **30s**; core-api is allowed
**45s** by its own `RequestTimeoutMiddleware`. The caller abandoned requests the
callee was still working on and would have finished.

Measured in prod at the 01:00Z tick on 2026-09-16 — the first run after the
lifecycle stagger landed:

```
01:00:02.2  core-operations  lifecycle fanout fired
01:00:30.1  core-operations  lifecycle fanout POST failed   (archive-expired)
01:00:30.2  core-operations  lifecycle fanout POST failed   (archive-stale)
01:00:45.2  core-api         request exceeded 45.0s budget
01:00:46.2  platform-ops     fanout OK
```

The work completed. Nobody heard.

### Why this is more than log noise

`_fire_fanout` returns from its exception handler **without reading the response
body** — and `failed` lives in that body. That is the per-org drop count added
after 2026-09-15, when every action reported success while 38 orgs went
unprocessed.

So a client timeout below the server budget disables the partial-sweep detector
on precisely the slow sweeps it was written to catch. The detector and the
timeout were added by different changes and nothing connected them.

### Why the setting is renamed, not just raised

`storage_http_timeout_s` described a dependency this service does not have. All
three call sites talk to **core-api**:

| line | endpoint |
|---|---|
| 47 | `/admin/lifecycle/fanout/{action}` |
| 224 | `/admin/lifecycle/embedding-coverage` |
| 395 | `/admin/interview/schedule/run` |

None touches core-storage-api. The value was inherited from a setting meant for
a different hop, which is why it was below core-api's budget — nobody chose
that ordering, it just fell out. Now `core_api_http_timeout_s`, 60s: above
core-api's 45s, below the 120s gateway cap. Checked that no deploy workflow sets
`STORAGE_HTTP_TIMEOUT_S`, so nothing loses an override.

### The test reads both real values

The invariant lives in `tests/test_lifecycle_fanout_aggregate_cap.py`, which
already imports both packages, and it compares the two **live settings** rather
than a copy of either. A hardcoded threshold is exactly how this drifts back:
raise one side and the assertion keeps passing while the mismatch returns.

That file now covers the 2026-09-15 incident from three ends instead of two. Its
docstring's action count is also corrected from 8 to 6, matching the six ticks
core-operations actually registers.

### Verification

ruff check + ruff format --check at CI's scopes, mypy, core-operations suite,
and the full root suite: **6984 passed**.

Four mutations, each caught by the test meant to catch it:

| mutation | caught by |
|---|---|
| budget back to 30s | `test_core_operations_waits_out_core_api_request_budget` |
| budget exactly 45s (the off-by-one) | same — the comparison is strict `>` |
| drop `timeout=` at the call site | `test_fanout_client_uses_the_core_api_timeout` |
| call site reads the old field name | same |


Eldad-Caura:
@erni-a pleasr apppfrovr

erni-a (approved):

PR #1541, opened 2026-09-16
fix(sentinel): point the memory-id stuffing check at the field Forge writes (09/02 L-35)

arkash20:
## The bug

Check #6 read `evidence["memory_ids"]`, behind `isinstance(evidence, dict)`.

The only production writer is Forge, and its distill schema declares `evidence` as a **string** — *"a 2-3 sentence human-readable rationale"* (`distill_prompt`). So the isinstance guard returned immediately on every real doc and **the check never fired once**.

Meanwhile the ids it was meant to bound live at `data["cites"]` (`all_memory_ids` in `forge_service`) — top-level and unguarded. A runaway or adversarial distillation could stuff hundreds there with no warning, which is exactly what this check exists to surface.

## The fix

Both locations are read now:

| Field | Why |
|---|---|
| `data.cites` | what Forge actually writes — the real one |
| `data.evidence.memory_ids` | the documents API accepts arbitrary `data`, so an external writer may legitimately use the dict shape this check was first written against |

Ids are **unioned**, not counted per-field: the cap describes what the renderer will show for the doc, not a per-location quota. Two under-cap fields that together exceed it warn, and an id present in both counts once.

The locator moves to `data.cites` — it lands on an inbox card, and pointing at a path no production doc has would send an operator looking for a field that isn't there.

## Tests

14 added; **6 fail against `main`**. Both locations, the union and dedup behaviour, the string-evidence shape that used to suppress the check entirely, the locator, and seven malformed shapes — the scanner runs on caller-supplied `data` and must never be the thing that fails a write.

655 passed / 0 failed on the `sentinel or forge or skill` selection; `mypy` clean on the touched file.


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

arkash20:
Checked the staleness before touching anything, because this PR is **approved** and a rebase would dismiss that approval for no gain.

It does not need one:

- `mergeStateStatus: BEHIND` — 120 commits behind `main`, but that is just "not up to date with base", not a conflict.
- `git merge-tree` against the current merge-base: **0 conflicts**.
- **`main` has not touched `sentinel_scan.py` or `test_sentinel_scanner.py` once since the branch point** (`bedf6054`). So the fix is not built on a superseded version of the function — the usual real risk with a branch this far back.
- CI green on the current head, 0 failing.

So this is merge-ready as-is. **Deliberately not rebasing**: it is approved and clean, and force-pushing would clear the approval and put it back in the queue behind a re-review it does not need.

Worth stating plainly since it has been open since 2026-09-16: the blocker here is a merge, not engineering. The source fix (`f88fba1e`) was already committed; the only thing added since is `4696b7e8`, the masking tests — the part that proves the old check passed on input production never produces.

PR #1543, opened 2026-09-16
perf(extraction): upsert relations concurrently instead of one POST at a time (09/02 L-37)

arkash20:
## The bug

Everything else this worker does is batched — `bulk_upsert_entities`, `bulk_upsert_entity_links` — and then relations went back to **one sequential HTTP POST each**. A memory with a dozen relations paid a dozen serial round-trips while the rest of its extraction had been reduced to two calls.

## Why concurrency and not a bulk endpoint

There is no bulk relations route, and adding one means a router, a service method and a client method — **plus per-item result reporting**, because the per-relation guard is load-bearing.

One failed upsert used to skip the A65 predicate write-back *and* the `Trigger.ENTITY` fire that is the only thing running A40's RDF pass. #1495 fixed exactly that. A single bulk call that succeeds-or-fails as a unit would throw that isolation away to save round-trips.

So every relation keeps its own request and its own `try` — they just stop queueing behind each other. Bounded at **8**, so one memory's graph can't open a connection per relation; same order of magnitude as the other per-item caps here (ingest preview 4, A19 admission 16).

Safe to gather because `upsert_relation` goes through the storage HTTP client, **not** a shared `AsyncSession` — gathering two `db.execute` calls on one session is what's unsafe.

`return_exceptions=True` is a backstop: `_upsert_one` swallows its own failures, but a bug in the endpoint resolution *above* the try would otherwise cancel in-flight siblings and discard relations that had already landed.

Counts derive from the gathered outcomes rather than re-deriving which relations were resolvable — one source of truth, so the two can't drift. `_upsert_one` returns `True` (landed) / `False` (failed) / `None` (unresolvable endpoints — the same silent skip the sequential loop did, counted in neither bucket).

## About the existing tests

`test_relation_upsert_cascade` anchored on `for rel in graph.relations:` and now anchors on the per-relation unit.

Worth being explicit: its two **behavioural** tests — the ones that actually prove #1495's guarantee, that a failing relation doesn't abort the others and still reaches the contradiction trigger — **passed unchanged throughout**. Only the structural slicing needed re-expressing. Two tests added there for the new properties: the fan-out is bounded, and a sibling cancellation can't lose a landed relation.

**9 pass with the change; 6 fail against `main`.**

The 3 failures in `test_a7_classifier_recall.py` from the wider run reproduce on pristine `main` (known local FTS/schema drift) and are unrelated. `mypy` clean on the touched file.


arkash20:
**Superseded by #1680, which is approved and green.** Recommending this one be closed rather than merged.

Both address 09/02 L-37. The difference is the transport:

| | this PR | #1680 |
|---|---|---|
| round trips | N, parallelised (bounded gather of 8) | **N → 1** |
| bulk route | rejected on failure-isolation grounds | added, with **per-item** results |

The premise here was right and #1680 did not abandon it — failure isolation is load-bearing, and #1495 established it. What changed is the evidence that a batch can keep it: `/entities/links/bulk` already does exactly this shape — one HTTP call, a session per item, `error: "fk_violation"` reported per row. #1680 follows that, shares the ON CONFLICT statement with the singular route through `_relation_upsert_and_fetch` rather than duplicating it, and re-asserts M-64 tenant scope on the new door with no existence oracle.

For the record, since it is the more useful half of this comment: I initially told the agent working on L-37 to **stand down** precisely because this PR existed, was green, and had been open since 2026-09-16 — a second solution makes a merge decision harder, not easier. It built the bulk endpoint anyway. That was the wrong call procedurally, and I would have reverted it had a reviewer not independently approved the result first. The approval settles it.

Two things this PR can still claim, whatever happens to it: it identified the problem first, and its objection is why #1680's per-item reporting exists rather than a batch verdict.

arkash20:
**Correction to my comment above — ignore it. This PR is NOT superseded, and should not be closed on my say-so.**

I wrote that #1680 superseded this and recommended closing. **#1680 has since been closed without merging**, so this is once again the only open fix for 09/02 L-37. Closing it on the strength of what I wrote would have left the row with no PR at all.

What actually happened: an agent working in parallel built a bulk-relations endpoint for this row despite being told to stand down precisely because this PR existed. It briefly carried an approval, which is why I wrote that comment. It then withdrew the PR itself, on the reasoning I had given it — the blocker here is a review decision, and a second, larger solution makes that decision harder rather than unblocking it. L-32, which was bundled into it, has been re-cut on its own as #1681.

So the position is unchanged from before any of that: **this PR is green, 0 failing, and has been open since 2026-09-16 waiting on a review.** I re-checked its mergeability earlier — 119 commits behind `main`, zero conflicts, and `main` has not touched `entity_extraction_worker.py` once since the branch point, so it is not built on a superseded version of the function.

The bulk-endpoint argument remains on the record as a possible follow-up — `/entities/links/bulk` is standing proof that a batch can report per item, so failure isolation and N→1 are not actually in conflict. That is worth considering *after* this merges, not as a reason to hold it.

Apologies for the noise. The earlier comment was written from a state that no longer holds.

erni-a (approved):

PR #1544, opened 2026-09-16
feat(ratchet): add the absence-assertion marker

Eldad-Caura:
## Why

A line that names the old brand **only so a test can assert it is absent** had no truthful marker. Both permanent hatches claim something still answers to the name — `legacy-name-ok` declares a compat alias, `legacy-name-floor` declares a name the rename never reaches. A negative assertion claims the opposite, and cannot be written without the literal.

That gap was minting the debt this gate exists to stop. An audit on 2026-09-16 found **14 sites across 3 repositories** assembling the name from fragments so the line would not carry the literal the gate searches for:

| Shape | Example |
|---|---|
| concatenation | `"mem" + "claw"` |
| join | `"".join(("mem", "claw_client"))` |
| escape encoding | `"memclaw"` |

Twelve of the fourteen **do** have a truthful existing marker and are being fixed separately. Two are absence assertions — and one of them says outright why it was disguised:

> …the only way past would be an exemption marker whose stated reason could not be true: these are not a compat alias, a redirect, or a pinned wire format. They are the thing we are asserting the ABSENCE of.

The author was trying to be honest, found no true label, and hid the string instead. A gate that leaves no truthful way to pass teaches people to hide from it.

This is the missing vocabulary, **not a third escape hatch**: the exemption is identical to the other two, and what differs is the claim a reviewer has to check.

## The population is real

Two lines in *this* repo already carry `legacy-name-ok` while their own reason text describes an absence assertion:

- `tests/test_env_write_new_names.py:58` — *"asserts a NEW file does not mint it"*
- `plugin/src/educate.test.ts:719` — *"asserts the old body is gone"*

Both sit in the alias list that rule 3 review is meant to read. Left for follow-up, since re-marking them needs this engine on `ratchet-stable` first.

## Three consumers, only one on the checklist

- **`main()`'s failure text** — per `_KINDS`' own note that a kind omitted there is a marker nobody discovers.
- **`CONTRIBUTING.md`** — *not* on that checklist and the one that matters most. That is where the author who hid the string was reading. The gate's stderr only reaches someone whose build already went red, and these authors never let it.
- **`legacy_name_census.py`** — branched on the two old kinds by hand and filed anything else as `unclassified_lines`, the programme's measure of **untriaged** debt. A reviewed, permanently correct line would have sat there forever, unclearable and indistinguishable from real debt, while the census still printed a number. Now table-driven, with a test asserting every ratchet kind has a census field.

## Scope note

Also inlines this repo's own two assembled literals under `legacy-name-ok`. The ratchet's test file was **cited by name in another repository as precedent for the device**, so leaving it assembled would have left the precedent standing.

## What this does not fix

Recorded in the engine's own *"What this still does not enforce"* section: the scan compares literal text, so an assembled name still passes for **any** motive, not just the one this marker relieves. Closing that means normalising a line before comparison (decoding escapes, folding adjacent/`+`-joined fragments) rather than parsing it — the single `git grep` over every file type is what the rest of the design rests on. It trades exactness for a heuristic, so it is a decision of its own rather than a rider here.

## Verification

- Ratchet suite **192 passed, 1 skipped**; census suite **10 passed**
- Six boundary tests and both census tests **confirmed failing against the unmodified scripts** first
- The two precedence cases pass *vacuously* without the fix — a marker that does not exist cannot outrank anything. They are regression guards on table order, **not** proofs of new behaviour, and are labelled as such
- `ruff format --check` and `ruff check` clean (check only — the formatter was never run on the engine)
- Gate passes on its own change; report shows `1 absence assertion(s)` end-to-end

## Rollout

Consumer repos run this engine from `caura-ai/caura@ratchet-stable`, so **nothing outside this repo picks the marker up until that branch is advanced** — which is deliberate and not something I should do. The 12 existing-marker fixes are unaffected and land independently.

@erni-a for review — not merging this myself.


Eldad-Caura:
@erni-a plear approve

erni-a (approved):

PR #1546, opened 2026-09-16
feat(operations): lease each scheduled tick so replicas fire it once

Eldad-Caura:
## The defect

The scheduler is an **in-process** asyncio loop. Its `_started` latch guards a double start *within* a process and nothing guards across processes — so **replica count is fire count**: every wall-clock aligned task fires once per live replica, all inside the same second.

## Measurement

Prod, 2026-09-16 00:00–06:00Z, `core-operations` at `maxScale=2`, the 01:00Z tick served by two distinct instance ids. **All nine scheduled endpoints fired exactly twice per tick**, and two exceeded core-api's own 45s request budget under the doubled load:

| endpoint | fires | worst |
|---|---|---|
| `interview/schedule/run` (hourly) | 12 over 6h | **504 @ 45.0s** |
| `lifecycle/fanout/archive-expired` | 2 | **504 @ 46.8s** |
| `lifecycle/fanout/entity-link` | 2 | 41.0s |
| `lifecycle/fanout/embed-backfill` | 2 | 42.8s |
| others | 2 each | ≤ 31.7s |

A killed fanout leaves its audit rows open at `pending` for every org whose publish never ran. `archive-stale` shows the contention directly: **2.1s for one call of its pair, 31.7s for the twin**.

## Why this, when the cap already shipped

caura-ai/caura-enterprise#1783 capped both scheduler services at `maxScale=1`. That makes the steady state correct but records two gaps it cannot close:

1. Cloud Run serves old and new revisions concurrently during a rollout.
2. `min == max == 1` leaves no headroom to replace an unhealthy instance.

This closes both by making correctness **independent of the replica count** — which is what lets that cap be lifted rather than kept forever.

## Why the receiver mediates

core-operations deliberately holds no datastore — its config carries a core-api URL and an admin key, nothing else — so its replicas cannot elect a leader among themselves. core-api can: it already owns the Redis. Deciding there also makes the guarantee **caller-agnostic** — it holds for N replicas, for a retry, and for an operator firing the same tick by hand, none of which a leader election among schedulers would cover.

## Fails open, everywhere, deliberately

`cache_set_nx` already returns `True` when Redis is down, and every layer above does the same: unreachable core-api, non-2xx, unparseable body, missing or non-boolean `granted`. **Only an explicit `{"granted": false}` stops a tick.**

A lease is an optimisation over behaviour that already tolerates duplicates downstream — the consumer dedup gate and the per-tenant activity gate both survive a double fire. Denying on error would convert one Redis blip into a silently skipped nightly sweep across the estate, which is strictly worse than the duplicate it prevents.

## Two details that are load-bearing, not cosmetic

- **TTL is `min(300s, interval/2)`**, not a constant, so a lease can never outlive into the same task's *next* legitimate tick — the one way this guard could cause an outage instead of preventing one.
- **core-api rounds a fractional TTL UP.** Rounding down would expire the lease early and re-admit the duplicate.

## Naming

The endpoint is `/admin/scheduler/tick-lease`, **not** under `/admin/lifecycle/`, because it also covers `agent-digest`, `interviewer-schedule` and `embedding-coverage`. Filing it under the lifecycle prefix would name it after one of its callers — the mistake `storage_http_timeout_s` made before it became `core_api_http_timeout_s`.

Interval-mode tasks are not leased: they fire at startup and drift by `fn` duration, so replicas never converge on a shared instant, and a name-keyed lease would just make one replica the permanent winner. Every task this service registers is aligned, so the exclusion is theoretical today — the reason is recorded so it doesn't read as an oversight.

## Verification

**Tests confirmed to fail without the change, for the right reason.** Disabling the claim fails exactly three assertions — denied-lease-still-ran, two-replicas-ran-twice, lease-never-consulted — all behavioural counts, not `AttributeError`. The 12 fail-open tests correctly still pass, since they guard the new code rather than its absence.

**The endpoint's validations were each mutation-checked:** accepting a `bool` TTL, rounding the TTL down, and dropping the task-name pattern are each caught by their intended test and nothing else.

95 tests pass (19 endpoint + 76 core-operations). `ruff check` and `ruff format --check` clean at every CI scope; `mypy core-operations/src` clean. `mypy core-api/src` reports 2 pre-existing errors in `common/llm/providers/openai.py` — confirmed present with this branch's changes removed, and main is green, so they're a local httpx resolution artifact rather than anything here.


Eldad-Caura:
Pushed `9a0ab37`. Two changes since the last review:

1. **Applied the review finding, verified first.** The body guard checked parseability but not shape — `[]`, `null`, `42`, `"x"` and `true` are all valid JSON, so `request.json()` returned them without raising and `.get` off a non-dict raised `AttributeError`: the same 500 the guard existed to prevent, one line later. Added the five cases and confirmed **all five fail** before the `isinstance` check went in. Also dropped the `body: dict` annotation, which asserted the shape rather than checking it — `json()` returns `Any`, so mypy had nothing to contradict.

   The core-operations client was already safe here: its `.get` sits *inside* the try, so a non-object answer fails open like any other malformed reply. Server-only, as reported.

2. **Fixed the CI failure**, which was mine: `ruff format` on root `tests/`. That path is in scope at `ci.yml:258` and I had it recorded as out of scope — I had conflated `ruff check` scope with `ruff format --check` scope. Re-verified every scope in both lists this time, plus mypy (only the 2 pre-existing `openai.py` httpx errors) and the full suite: 24 endpoint + 76 core-operations.

@claude please re-review the current head.

Eldad-Caura:
@erni-a plear approve

erni-a (approved):

PR #1547, opened 2026-09-16
test: file two absence assertions under the marker that is true of them

Eldad-Caura:
## What

Two lines assert the old brand is **not** present, and both carried `legacy-name-ok` — which claims the opposite: that something still answers to the name.

| Line | What it actually does |
|---|---|
| `tests/test_env_write_new_names.py:58` | inside a test named `test_env_block_writes_no_old_names`, asserting `old not in block` |
| `plugin/src/educate.test.ts:719` | `assert.ok(!tools.includes("memclaw_write_bulk"))`, reason text *"asserts the old body is gone"* |

These are the lines cited in #1544 as evidence the absence category had a population *before* it had a marker. Until now they've been sitting in the compat-alias list that rule 3 review is meant to read — which is the exact pollution the alias/floor split exists to prevent.

## What deliberately does not change

`educate.test.ts:720` and `:723` pin the fence tag and the backup filename. Those really are on-disk contracts the rename never reaches, so `legacy-name-floor` is true of them and they keep it.

That file was already drawing the alias/floor distinction correctly; this marker just completes the vocabulary it needed.

## Verification

Comment text only — two lines, no code semantics.

- Gate, run with the engine **now live on `ratchet-stable`**: `No new lines`

## Not claimed

No local `ruff format` verdict. The ruff available here is 0.15.11 and this repo pins **v0.15.4**; they disagree about this file at the **unmodified baseline** too, so the local check says nothing about this change either way. The diff only shortens a trailing comment, which cannot introduce a formatting difference under either version — but I'd rather say that than present a green from the wrong tool.

I also could not run the two suites locally (the root `tests/` conftest needs the full core-api environment, and `plugin/` has no installed `node_modules`). CI runs both.

With caura-ai/caura-enterprise#1787, this closes the 2026-09-16 audit: every assembled identifier now carries a truthful marker.

@erni-a for review — not merging this myself.


Eldad-Caura:
@erni-a plear aprpvoe

erni-a (approved):

PR #1549, opened 2026-09-16
feat(lifecycle): republish audit rows the fanout stranded

Eldad-Caura:
## What this fixes

`_trigger_one` writes the audit row **before** publishing the per-org message — deliberately, so a failed publish leaves a visible `pending` row rather than a message referencing an id that doesn't exist. Nothing was ever built to act on those rows: no redelivery (no message was ever published), no reconciler, no sweep. The row records a job that never ran.

## Prod evidence

At the 2026-09-16T01:00Z tick, core-operations still ran at `maxScale=2` with no per-tick lease, so two replicas fired the same archive fanout:

```
core-api          request exceeded 45.0s budget: POST /api/v1/admin/lifecycle/fanout/archive-expired
core-operations   lifecycle fanout POST failed  (x3)
core-worker       pubsub handler raised; nacking for redelivery  (x8)   <- #1530 working
```

`asyncio.gather` was cancelled, and rows already written by `audit_begin` never reached their `publisher(...)` call. Two `archive-expired` rows were orphaned **with no log line of their own** — they aren't even in the endpoint's `failed` count, because the task that would have counted them was cancelled.

The post-deploy smoke asserts every audit row in its 30h window reached `success`, so those two rows failed the promote gate at 13:56Z and again at 17:51Z with byte-identical counts (`{'pending': 2, 'success': 302}`). Prod promotes were red for 29 hours.

## Why republish instead of marking them failed

Marking a stranded row `failure` makes the drop visible, but the gate stays red and — more to the point — the org's lifecycle work still hasn't run. The row isn't mislabelled; the job is missing. So the sweep finishes the job, republishing under the **same `audit_id`** so the consumer finalizes the original row. A fresh row would leave the stuck one exactly as stuck while making the totals look healthier.

Repeating is safe: the consumer re-checks the dedup gate and skips when the work already succeeded, `success` is sticky so a late redelivery can't downgrade a finished row, and the primitives are idempotent.

## Three constraints, each pinned by a test

| Constraint | Why |
|---|---|
| Shares the process-wide fanout semaphore | A private budget would be a second unbounded fanout against the storage writer — the exact failure that stranded these rows |
| Only `triggered_by='core-operations'` rows | Manual triggers may carry a `fleet_id` or one-off `retention_days` that this table doesn't persist; republishing one would run a *different* job than the row records. Enforced in the query, so such rows are never candidates |
| Only `pending`, never `in_progress` | `in_progress` means a consumer holds the message and Pub/Sub redelivery covers it; republishing would duplicate live work |

## Index

`status='pending'` is the selective half of the predicate — `started_at < now() - interval` matches nearly the whole append-only table — so revision **046** adds a partial index on the status rather than leaning on 041's recency index, which would scan almost everything to return normally-zero rows.

## Schedule

Hourly, offset to half past. Every fanout aligns to the top of its hour, so firing there would put the sweep's reads in the same instant as the burst it cleans up after; half past also sits clear of the 30-minute strand threshold.

## Test plan

Five tests, each **verified to fail under a targeted mutation** of the code it guards:

| Mutation | Test that caught it |
|---|---|
| Allocate a new audit id instead of reusing the stranded one | `test_sweep_republishes_under_the_original_audit_id` |
| Drop the fanout-only restriction | `test_sweep_only_asks_for_fanout_rows` |
| Give the sweep its own private semaphore | `test_sweep_shares_the_fanout_budget_rather_than_opening_its_own` |
| Remove `return_exceptions=True` | `test_one_failed_republish_does_not_abort_the_rest` |
| Raise on an action missing from the registry | `test_retired_action_is_counted_not_raised` |

All five pass on the real implementation and fail on their mutation.

Also run, at CI's exact scopes:
- `ruff check` (v0.15.4, pinned): core-api/src, core-storage-api/src, common, core-operations/src+tests, tests — all pass
- `ruff format --check`: same scopes plus the isolated `common` trio — all clean
- `mypy`: core-storage-api and core-operations clean; core-api reports only the 2 pre-existing `httpx`/`httpx2` errors in `common/llm/providers/openai.py`, a file this diff doesn't touch
- legacy-name ratchet: "No new lines." Sentinel: 35/35 protected strings survive

The 3 failures in `tests/test_lifecycle_fanout_aggregate_cap.py` are pre-existing and reproduce identically on a clean `origin/main` worktree — `ModuleNotFoundError: No module named 'core_operations'`, a local-env artifact, since CI installs that package for its own steps.

## Scope

Not the root cause of that night's amplification — #1783 capped the scheduler services at one instance and #1546 leased each tick, and both are now live in prod. This is about what happens to a row when a fanout dies part-way, regardless of why.


Eldad-Caura:
@claude Both findings verified against the code and fixed on `c1d521f0`; please re-review that head.

**Medium — false "republished".** Confirmed: `PubSubEventBus.publish` batches and does not await the delivery future, and its own comment says publisher-side failures "land in the SDK's background-thread log instead". The counter is now `republish_attempted`, named for what is actually known, and a new `ineffective` count reports rows old enough to have been through two sweeps and still pending — row age being the only evidence available, and sufficient, since a repaired row leaves `pending` and stops being selected. `core-operations` escalates to error level when it is non-zero.

**Low — head-of-line blocking.** Retired-action rows are now finalized as `failure` with the reason, so they drop out of the partial index and stop occupying the oldest-first LIMIT window. Marking failure is right here and wrong in the general case: an undelivered message is unreported work, a retired action is unrunnable work.

Also in this head, from CI rather than review: the first revision used an unscoped `GET /stranded` and the tenant-scope gate rejected it. Since that allowlist is a ratchet that may shrink and not grow, the fix was to bind the scope rather than seek an exemption — it is now `POST` with `org_id` required in the body (null for admin-wide), matching `lifecycle_audit_summary`. Gate reports `added (0)`.

Six tests now, each confirmed to fail under a targeted mutation of the code it guards, including the two new ones.

Eldad-Caura:
@claude Both round-two findings verified and fixed on `edb15b86`; please re-review that head.

**Medium — finalize bypassed the semaphore.** Correct, and it was the one path in the function that skipped the budget the surrounding comments argue for. Every row in a sweep can name the same retired action, so it could fire up to `_RECONCILE_MAX_ROWS` concurrent PATCHes at the writer. Now inside `async with sem`, covered by a peak-concurrency test.

**Medium — `ineffective` false positives.** Also correct, on both counts. It now counts only rows this sweep actually published for, so rows finalized as `unknown_action` in the same call are excluded — they were never published and are terminal, not stuck. And the log no longer asserts a cause row age cannot establish; it states the observation and leaves the history open.

One part I did not take, with reasoning rather than silence: I did not add a `last_reconciled_at` column to distinguish "an earlier sweep failed" from "no sweep had run yet". On first deployment the backlog does still trip the alert — but those rows genuinely are not completing, and the operator action is identical in both readings (find out why). A schema change and a per-row write each sweep buys attribution that changes no decision, so the honest wording seemed the better trade. Happy to add the column if you disagree.

Also fixed from CI: `lifecycle-reconcile` is now in `docs/component-registry.yaml`, which the component-registry drift test requires for every `scheduler.register` call.

Eight tests now, each confirmed to fail under a targeted mutation of the code it guards.

Eldad-Caura:
@claude Fixed on `9d15ed9f` — correct on both counts, and both were mine.

The `lifecycle-reconcile` block now sits *before* the "Interviewer Phase 1" comment, so each comment is directly above the registration it documents. And the module docstring went from "Ten jobs" to "Eleven", with `lifecycle-reconcile` described in the list and `embed-backfill` renumbered from "an eleventh" to "a twelfth" — a count that silently stops matching the code is how the next reader stops trusting the docstring.

Ruff, mypy and the component-registry drift test all clean on core-operations. No behaviour change in this head: placement and prose only.

Eldad-Caura:
@claude Fixed on `0b808217` — this one was right and it invalidated a safety claim the PR was making.

**The claim I had wrong.** I argued republishing was safe because of the dedup gate and sticky success. The dedup gate only matches work that already *succeeded*, so in a race neither delivery is covered. And I confirmed these subscriptions retain messages for **7 days** (`messageRetentionDuration=604800s`), so no practical age threshold separates "never published" from "queued behind a backlog". The ambiguity is real and unavoidable at the query level.

**The fix.** Revision 047 adds `claimed_at`; `pending -> in_progress` is now a compare-and-swap, so only one delivery can start the primitive. The guard has a staleness arm (`LIFECYCLE_CLAIM_LEASE_MINUTES`) so it does not create a worse bug than it fixes — a consumer that dies mid-run leaves a claim nobody would ever clear. `failure -> in_progress` is untouched; only a live claim blocks.

**The loser nacks rather than acks.** Acking would drop that delivery permanently, and if the holder then died the row would park at `in_progress`, where this sweep deliberately does not look — trading a duplicate run for a silently stranded row, which is the exact failure this PR exists to end. Same principle as #1530.

**Threading the signal was most of the work.** Four places swallowed it: the tri-state return could not express "claimed", the PATCH route collapsed it, and *both* the core-api and core-worker adapters declared `-> None` and discarded the response. mypy caught the second adapter. The route returns 200 with a `claim_conflict` field rather than 409 deliberately — the consumer wraps that call in a broad `except` that logs and **continues**, so a 409 would have been swallowed into "continuing" and the duplicate would have run anyway. A signal the caller must act on cannot arrive as an exception that an existing handler already ignores.

Two new handler tests, both confirmed to fail under mutation in each direction — ignoring the conflict, and skipping unconditionally. Ruff and format clean at every CI scope; mypy clean on all four services except core-api's 2 pre-existing `httpx` errors; tenant-scope gate `added (0)`; ratchet and sentinel green.

Eldad-Caura:
@claude Fixed on the new head — and thank you for the end-to-end trace on the CAS; that is the part I most wanted checked.

**Low, route ordering.** Real, and I confirmed it: `/admin/lifecycle/reconcile-stranded` (line 330) resolves only because it precedes the catch-all `/admin/lifecycle/{action}` (line 613). I took both halves of your suggestion. The constraint is now stated at the route, and there is a test that resolves an actual request scope against the router and asserts the matched endpoint is the sweep — not a list-index comparison, which would be a proxy for the behaviour rather than the behaviour. Mutating the literal path makes it fail with `request resolved to trigger_lifecycle_action, not the sweep`, which is exactly the silent misroute you described.

Also in this head, from CI: the migration-chain test pins its expected head rather than deriving it — deliberately, so two branches claiming one `down_revision` cannot both land. Bumped 045 to 047 and asserted both new links; verified a forked `down_revision` makes it fail.

Nine tests in the sweep file plus two in the handler file, each confirmed to fail under a targeted mutation. Ruff and format clean at every CI scope; mypy clean on all four services except core-api's two pre-existing `httpx` errors in a file this diff does not touch; tenant-scope gate `added (0)`.

Eldad-Caura:
@claude High finding fixed on `84584fd9`. You were right, and I verified the mechanism before acting: `_patch` routes through `with_retry`, and `httpx.ReadTimeout` is in its retry set — so a claim that succeeded server-side but lost its response is re-sent verbatim, and my CAS read that re-send as a competitor. The handler would have nacked a delivery it had won.

**Fix: the claim needed an identity.** Revision 047 now also adds `claim_token` — one value per consumer *invocation*, not per HTTP request — and the CAS admits a row whose token matches. The client's own retry carries the same token and is recognised; a genuine second delivery re-enters the handler, mints a different token, and still loses.

I extended 047 rather than stacking a 048, since it has not merged.

**Plumbing, which was most of the work.** The parameter had to reach the SQL through five layers that would each have dropped it silently: both storage clients, both adapters, and the PATCH route. Same shape as the `claim_conflict` flag needed, and for the same reason — a signal is worth only as much as the narrowest layer that forwards it.

**On the test.** I could not run core-storage-api integration tests locally (`common` is not importable in that venv and they need a live database), so I could not mutation-verify this one the way I have the others — stating that rather than implying coverage I did not confirm. Instead the test carries its own control: it asserts the same-token retry succeeds *and* a different token conflicts, in one test. Those pull in opposite directions — a guard ignoring the token passes the competitor half and fails the retry half; a guard accepting every `in_progress` write does the reverse. Only a claimant-keyed guard passes both.

Also fixed from CI: `test_route_passes_org_to_finalize_and_preserves_sticky_success` asserted the PATCH response dict exactly, which my added field extends. Updated both assertions; checked no other test asserts that shape.

Eldad-Caura:
@claude Fixed on `d2a4cae6`. Correct, and it is arithmetic that reads right and is not.

`seconds_until_next_utc_top_of_hour()` returns the distance to the next `:00`, so adding 1800 lands half past the **following** hour. At 10:05 it schedules 11:30 — skipping the 10:30 it was aiming at and waiting 85 minutes instead of 25. Since delay providers are recomputed every cycle, that hits on every process start landing in the first half of an hour.

Replaced with a `seconds_until_next_utc_half_past` helper beside the existing ones, same strict-future guarantee (never 0, so an aligned task cannot hot-loop after a fast failure at the mark). Five tests, including one that pins the old formula's exact error — `assert buggy == 85 * 60` — so a revert to the arithmetic fails rather than silently passing. Reverting the helper body makes 3 of the 5 fail.

**A gap of mine this surfaced, worth stating.** Adding the tick also broke `core-operations/tests/test_app.py::test_all_lifecycle_jobs_registered_and_wall_clock_aligned`, which asserts the registered job *set*, not a count. It had been failing since I added the tick and I had not seen it: I was running the root suite and the component-registry test, but never `core-operations/tests/`, while changing that service. Now fixed, and I have run the per-service suites — core-operations 81, core-worker 80, core-api 52, all passing.

No behavioural change to the sweep itself in this head; scheduling and test-registration only.

Eldad-Caura:
@claude Fixed on `c02831fb`. Correct, and it is the worst possible window for it — the guard would have been unenforceable precisely during the deploy that shipped it.

Your reasoning holds exactly: arm 1 (`status != 'in_progress'`) already covers every other status, so `claimed_at IS NULL` is reachable only for a row that *is* `in_progress` — and 047 added the column with no backfill, so every mid-flight row read as free to claim. The migration docstring's "correct for every existing row" was simply false, and I have corrected that sentence rather than softened it.

**Fix:** 047 now backfills `claimed_at = now()` where `status = 'in_progress' AND claimed_at IS NULL`. That starts the ordinary lease against those rows instead of exempting them — a live consumer keeps its row for the lease, a dead one releases it through the same staleness arm, so nothing is stranded. The `claimed_at IS NULL` predicate keeps it a one-shot repair: a re-run cannot extend a live lease.

**I kept the `claimed_at IS NULL` arm, deliberately.** After the backfill it no longer carries pre-existing work; it covers the narrower case your prompt didn't reach — a row moved to `in_progress` by a **pre-047 revision still serving during a rolling deploy**, which writes no claim. Dropping the arm strands those permanently: `NULL < now() - lease` is NULL, not true, and the sweep will not rescue them either because it only reads `pending`. A bounded theft window during one deploy beats rows that can never be recovered at all. That trade is now written at the arm rather than left implicit.

**Tests:** `tests/test_migration_047_claim_backfill.py`, 3 tests. Removing the backfill fails 2 of them (the third is a regression guard that both columns stay nullable with no server default, so the addition stays metadata-only — it passes either way by design).

**Verified:** ruff check and format clean at CI's scopes; mypy clean on core-storage-api (91 files); core-operations 81, core-worker 80, core-api 52, root lifecycle + schema suites 142 — all passing. I could not run `core-storage-api/tests/` locally: it is an integration suite and fails at setup with `database "caura_storage" does not exist`. CI provides that service, so those are on CI, not on my word.

Eldad-Caura:
@claude Fixed on `cf1f563e` for the Medium. The Low I am not taking, and the reason is specific rather than a judgement call — detail below.

**Medium — correct, and the mechanism is the point.** Winning the claim is not the same as still holding it. The staleness arm exists so an abandoned claim can be taken over, and it cannot distinguish an abandoned consumer from a slow one, so a primitive that outruns the lease is preempted *while alive*. Both then finalize, and the loser's write landed silently.

`lifecycle_audit_finalize` now restricts terminal writes to the claim holder and returns `claim_lost`; the handler logs that at error rather than raising, since the work has already run twice and a retry would make it three times. A terminal write carrying **no** token is still admitted — the embed-backfill consumer in `core-worker/consumer.py` claims without one, and refusing its finalize would trade a silent duplicate for a permanently unfinalizable row, which the sweep cannot rescue because it only reads `pending`.

I checked your reachability premise rather than assuming it, and it holds for a reason worth recording: these are **pull** subscriptions consumed through unary `subscriber.pull()`, not StreamingPull, so there is no automatic lease extension, and `modify_ack_deadline` is only used for nacks. Nothing in the dispatch path bounds handler runtime — my first hypothesis was that the Cloud Run request timeout (180s worker / 60s api) capped it, and that is wrong for pull consumers.

**Low — not reachable, and I would rather show the check than argue.** The scenario needs one caller that both omits `claim_token` and reacts to `claim_conflict`. No such caller exists in any version:

- `claim_conflict` appears **0 times in `origin/main`** and is introduced by this PR, so a mid-rollout instance running the old code cannot nack on a field it does not read. It proceeds and runs the primitive — exactly its pre-PR behaviour, no regression and no nack storm.
- Every caller running the new code mints a token unconditionally (`claim_token = uuid.uuid4().hex`, one per invocation), so the token arm is always present for anything that *can* nack.
- The one genuinely tokenless caller, `core-worker/consumer.py:778`, has **0 occurrences** of `claim_conflict`.

The fix you propose — skipping the CAS when no token is sent — would reintroduce the duplicate-run race this PR exists to close, and would do it for the sweep's republish path specifically. I have not weakened it. What I did take from the finding is that the tokenless path is load-bearing, and it is now documented at both the CAS arm and the new terminal guard rather than being accidental.

**Tests:** 2 new in `tests/test_lifecycle_handlers.py` (20 → 22), both verified failing without the fix; 1 new in `core-storage-api/tests/` covering mismatch, match, and the tokenless case.

**Verified:** ruff check and format clean at CI's scopes; mypy clean on core-storage-api (91 files); core-operations 81, core-worker 80, core-api 52, root lifecycle + migration + schema suites 144. Two suites I could not run locally and am not claiming: `core-storage-api/tests/` needs a Postgres `caura_storage`, and the full `pytest tests/` cannot collect here because a user-site `mcp` package predates `CacheHint` — 41 collection errors, none in a file this PR touches.

Eldad-Caura:
@claude Fixed on `451f1aa5`. Correct, High is the right severity, and it was mine from the previous round — the guard I had just added was inert in the case it exists for.

A preempted consumer loses its claim to a winner that then usually **succeeds**, so the row it finds is almost always at `success`. Testing status first classified that as an ordinary redelivery, so the only duplicate runs it could ever have caught were the ones where the winner *also* failed. The check now runs before the sticky-success no-op, with the ordering reasoning written at the branch so it doesn't get re-sorted later. A consumer that still holds its claim reaches the no-op normally, since its own token matches.

**Reordering alone would have been wrong, and the check for that turned up a second defect.** The handler never read `noop`, so a redelivery of an already-succeeded row re-ran the primitive and then wrote a terminal status — with the winner's token on the row and its own token different. After the reorder that would report every routine redelivery of a completed message as a duplicate run, burying the real signal in false ones.

A `noop` claim now acks and stops. The row is at `success`, so re-running is the duplicate this path exists to avoid; skipping it is right on its own merits and is what keeps the reordered check meaningful.

**Tests:** `tests/test_lifecycle_handlers.py` 22 → 23, the new one verified failing without the fix. The storage suite's guard test now covers the ordering directly — a loser's terminal write against a **succeeded** row must report `claim_lost` and not `noop`, and the holder's own retry of its successful write must report `noop` and not `claim_lost`. Those two assertions are the regression; a status-first implementation passes the old cases and fails both.

**Verified:** ruff check and format clean at CI's scopes; mypy clean on core-storage-api (91 files); core-operations 81, core-worker 80, core-api 52, root lifecycle + migration + schema 145. Unchanged from last round: `core-storage-api/tests/` needs a Postgres `caura_storage` and the full root collection is blocked by a user-site `mcp` predating `CacheHint`, so both rest on CI rather than on my word.

Eldad-Caura:
@claude Fixed on `489d9d7a`, along with the CI failure on the previous head.

**The Medium is right, and it is the same omission as the one I fixed last round — I fixed it in one place and not the other.** Reporting a lost claim only on the success path makes a duplicate visible exactly when it cost least. The preempted-then-failed case is the more alarming half: two consumers ran, and the result standing on the row is not the one about to raise. The failure finalize now captures its result and logs at error; the original op exception still propagates, so the report is purely additive.

**CI failure — a contract test, and it caught something local runs structurally cannot.** `test_route_passes_org_to_finalize_and_preserves_sticky_success` asserts the PATCH response dict *exactly*, so adding `claim_lost` broke it:

```
assert {'ok': True, ..._lost': False} == {'ok': True, ...flict': False}
```

That exactness is the point — it is what makes the test a contract rather than a smoke check — so I updated the two expected dicts deliberately rather than loosening the assertions to `in` checks. Both are in `test_lifecycle_audit_finalize_tenant_scope.py`; I checked the third exact-dict assertion in `test_lifecycle_audit_observation.py` and it is a GET on the row, unaffected by a PATCH response field, which CI confirms — 1 failed, 406 passed.

Worth stating plainly: this is the second round where the gap was in the suite I cannot run locally. `core-storage-api/tests/` needs a Postgres `caura_storage`, so its contract assertions are only exercised on CI, and I am relying on CI for them rather than claiming otherwise.

**Tests:** 23 → 24 in `tests/test_lifecycle_handlers.py`, the new one verified failing without the fix.

**Verified:** ruff check and format clean at CI's scopes; mypy clean on core-storage-api (91 files); core-operations 81, core-worker 80, core-api 52, root lifecycle + migration + schema 146.

Eldad-Caura:
@erni-a plear aprpovre

erni-a (approved):

PR #1551, opened 2026-09-16
fix(common): type _dd_tracer as Tracer | None

Eldad-Caura:
`_dd_tracer` is bound by `from ddtrace import tracer as _dd_tracer`, so
wherever ddtrace actually resolves, mypy reads the name's type as
`Tracer` — and both `None` assignments are type errors:

```
common/structlog_config.py:51:  Incompatible types in assignment
    (expression has type "None", variable has type "Tracer")  [assignment]
common/structlog_config.py:338: ... same
```

Line 51 is the `ImportError` fallback; line 338 is the disable-on-fault
branch in `_add_dd_trace_context`.

## Why this repo never saw it

ddtrace sits behind the `datadog` extra and is **not installed in this
tree**, so `ignore_missing_imports` degraded the import to `Any` and
`None` assigned cleanly. `mypy --config-file common/mypy.ini common/` is
green here and always has been.

This file is vendored **byte-identically** into `caura-enterprise`
(`policy=identical`), which *does* install ddtrace — 4.11.1. The same
command, on the same bytes, with the same mypy (2.3.0) and the same
`common/mypy.ini`, reports the two errors there. The only difference
between the two runs is whether ddtrace resolves.

The practical cost: this failed enterprise's mypy **pre-commit hook on
every commit**, including commits touching nothing near this file — the
exact "reach for `SKIP=mypy`" failure mode that hook's own config comment
says it exists to prevent.

## The fix, and why it's shaped this way

Import the module and read the attribute instead of binding the name
through `import ... as`. That is the one spelling that checks clean in
**both** trees, which matters because the file must stay byte-identical:

| form | ddtrace absent (here) | ddtrace present (enterprise) |
|---|---|---|
| `from ddtrace import tracer as _dd_tracer` | clean | ✗ `assignment` ×2 |
| declare `_dd_tracer: Tracer \| None` + same import | ✗ `no-redef` | clean |
| **import module, read attribute** | **clean** | **clean** |

Both of the first two were tried and measured, not reasoned about.
`Tracer` comes from `ddtrace.trace` because 4.x dropped the top-level
re-export (`hasattr(ddtrace, "Tracer")` is `False` on 4.11.1).

## No behaviour change

`_dd_tracer` stays module-level and patchable, and is still the tracer
when ddtrace is present and `None` when it is not — verified by importing
the module under both conditions.

## Verification

On the same bytes, in both trees:

- **this repo** (ddtrace absent): mypy `Success: no issues found in 101
  source files`; `ruff check common/` clean.
- **caura-enterprise** (ddtrace 4.11.1): mypy `Success: no issues found
  in 35 source files`; `ruff check common/` clean; `ruff format --check`
  clean at the default 88 columns, which the module docstring calls out
  as a hard vendoring constraint.
- `tests/test_logging.py`, `tests/test_pubsub_trace_filters.py`,
  `tests/test_contradiction_diag_log_tier.py`: **85 passed, 1 skipped**.

Once this merges, the re-vendor carries it into caura-enterprise and
unblocks that hook. Nothing was patched in the enterprise copy directly.


Eldad-Caura:
Applied the review finding, plus the DCO sign-off the `dco` check was failing on.

**Finding — uncaught `AttributeError` on the attribute read: correct, applied.**

Verified the premise before applying rather than taking it on faith. Against a stub `ddtrace` package with no top-level `tracer`:

- `from ddtrace import tracer` → `ImportError: cannot import name 'tracer' from 'ddtrace'` — caught by the existing handler.
- `import ddtrace; ddtrace.tracer` → `AttributeError: module 'ddtrace' has no attribute 'tracer'` — **not** caught.

So the restructure did convert a gracefully-degrading failure into an import-time crash for every service importing this module. The finding is right, and it is sharper than it looks: this PR's own rationale is that ddtrace 4.x dropped the top-level `Tracer` re-export, which is direct evidence that these re-exports do move.

Fix: the attribute read moved inside the `try` and the handler now catches `(ImportError, AttributeError)`. I took that over the suggested `getattr(ddtrace, "tracer", None)` because `getattr` with a default degrades the inferred type to `Any`, which would give up the typing precision this PR exists to add — the `try` form keeps `Tracer` and still degrades to `None`.

Re-verified against the same stub, both directions:
- with the fix: module imports, `_dd_tracer = None`
- at the previous commit: `AttributeError` crashes the import

**Gates** — same bytes, both vendored trees:
- OSS (ddtrace absent): mypy `Success: no issues found in 101 source files`; `ruff check` clean; `ruff format --check` clean.
- caura-enterprise (ddtrace 4.11.1): mypy `Success: no issues found in 35 source files`; `ruff check` clean; `ruff format --check` clean at 88 cols (the docstring's vendoring constraint).
- Runtime: `Tracer` with ddtrace present, `None` with it absent, `None` with it present-but-`tracer`-less.
- `tests/test_logging.py`, `tests/test_pubsub_trace_filters.py`, `tests/test_contradiction_diag_log_tier.py`: 85 passed, 1 skipped.

Amended into the single commit with `-s`; the missing `Signed-off-by` trailer was the only other red.

@claude

Eldad-Caura:
@erni-a plerar pparove

erni-a (approved):

PR #1552, opened 2026-09-17
fix(core-api): close five surface-parity and authz gaps on MCP and /memories/count

Eldad-Caura:
Batch 1 of the core-api audit remediation (60 findings, split into 8 PRs — this is the access-control and input-boundary group, pulled first).

Seven findings were assigned. **Five were real and are fixed**, one was a duplicate of another, and **one turned out not to be a defect at all**. Every real one was reproduced against the running code before being touched, and every fix-bearing test was confirmed to fail with its fix reverted.

None of these is a guard that was *wrong*. Each is a guard that was never *asked for* — which is why they survived on surfaces that otherwise look carefully gated.

## The one to read first

**A soft-deleted org kept full MCP read and write.** REST calls `auth._block_if_suppressed` from `get_auth_context`, so every REST route inherits the CAURA-694 boundary guard by construction. The MCP middleware resolves identity the same way and then simply never asked the question. Same credential, same org: REST 403s, MCP served.

Gated in the middleware rather than per-tool, for the same reason REST puts it in the auth dependency — a tool added later cannot forget it. Readable tenants are checked too, mirroring `_block_if_any_readable_suppressed`. The 30s cache and fail-open-on-storage-error posture are inherited from `core_api.suppression`, not re-decided here.

## Also fixed

**`GET /memories/count` had neither guard the routes either side of it apply.** No fleet-read trust gate, so a trust-1 agent could count another fleet's rows that `GET /memories` would have refused to show it. And no `scope_agent` exclusion, so it handed back *as a number* what the list route withholds — peers' private row counts. `memory_stats` carries the identical fleet gate with a comment recording that it too was once missing; count is the neighbour nobody revisited.

`exclude_scope_agent` is plumbed through core-storage-api (service + `/count-active`) defaulting **False**, so the whole-corpus and admin callers are unchanged; core-api passes `True` because this route names no agent and so has no identity entitling it to those rows.

**Three MCP input boundaries that REST already enforces:**
- `caura_recall` clamped `top_k` at the upper bound only. A negative value passed straight through — the entity route returned a full unscored pool, the scored route sent a negative SQL `LIMIT` and 500'd, and the response echoed `effective_top_k: -5`, reporting the bad value back as if honoured. Now `max(1, min(...))` — the same two-sided clamp the doc-search path *in the same file* already used.
- `caura_recall` had no query length bound while REST's `SearchRequest` caps at `MAX_QUERY_LENGTH`. This is the one surface where the caller sets the cost of the request: an unbounded query reached the embedding provider and the FTS path.
- `caura_write` let Pydantic `ValidationError` escape. **Narrowed:** the finding names the single *and* bulk paths; bulk has since been fixed and catches its own per item. Only the single path was still open.

## Not a defect — left to you

**`oss-0814-m-29` (MCP bypasses org read-only) is already implemented and deliberately off.** The signal is plumbed (`_org_read_only_var`, `_check_plan_limit`) and the gate exists on both write paths; it refuses only when `enforce_mcp_plan_limits` is on, which defaults `False` on purpose. Its docstring lays out the staged rollout — observe via `mcp_plan_limit_would_refuse`, read the logs, then enforce with the number in hand — and warns that a quiet log is *not* a green light while `meters_mcp_bulk_write()` is off.

Flipping that flag is the product decision that procedure exists to inform, not a code change. Left alone with the rollout note intact. **Say the word if you want it flipped** — it's a one-line config change, but it takes capability away from tenants who have it today, which is exactly what the staging was designed to measure first.

## Found by the work, not by the audit

`tests/test_authz_gate_inventory.py` exists to stop exactly this class of gap, and it did not catch the `/memories/count` hole. Its three invariants cover **mutating** routes and the `enforce_self_agent` identity axis; a **read** route that takes a `fleet_id` and skips the fleet-read gate falls outside all three.

A fourth invariant is the right mechanism-level fix and belongs in that file. I did not add it here: ~46 routes accept `fleet_id`, and classifying them needs that file's real classifier rather than the naive AST walk I probed with — a probe that cannot tell "ungated" from "gated via a helper or dependency", so I am explicitly **not** claiming a count of holes. It deserves its own change.

## Verification

| gate | result |
|---|---|
| `ruff check` / `ruff format --check` | clean |
| `mypy` core-api / core-storage-api | no new errors (2 pre-existing in `common/llm/providers/openai.py`, confirmed on a clean tree) |
| full OSS suite | see checks |
| each fix reverted individually | corresponding test fails; the two "don't over-correct" guards stay green either way |


Eldad-Caura:
Both findings verified against the code and applied in `92fafbfd`.

### Medium — `/memories/count` undercounts an agent-scoped caller's own rows

Real, and the premise holds. Confirmed on the route the finding is actually about: `GET /memories` builds `caller_agent_id = auth.effective_agent_id(agent_id)` and hands it to `memory_list_by_filters`, whose predicate admits `scope_org`/`scope_team` plus the caller's own `scope_agent` rows. (The prompt cites `memory_load_by_ids`; that carries the identical predicate, but `memory_list_by_filters` is the list route's path.) So for an agent credential the count came in *under* the list — the equal and opposite error to the leak this PR opened with.

`caller_agent_id` now flows `count_active` → `/count-active` → `memory_count_active`. It is the same value the fleet gate already computes, and `None` for a tenant/user credential, which is what keeps the no-identity case hiding every private row. The route takes no `agent_id` param, so there is nothing to forge.

Two things I did differently from the suggested fix:

**Predicate extracted rather than copied.** This would have been the fifth verbatim copy of that `or_()` in `postgres_service.py`. Its neighbour `_fleet_scope_clause` was centralised for exactly this reason — *"the identical predicate lived in several queries, one was fixed, and the leak simply moved to the next copy."* It now lives in `_visibility_scope_clause`. The new site is its only caller; converting the four that predate it is a follow-up rather than a rider on a PR under review.

**Kept `exclude_scope_agent` alongside the identity**, rather than replacing it. The counter has three states where the list route has two — unscoped is real here, the auto-crystallize spend gate counts the whole corpus including private rows. One `str | None` cannot carry three, and not only in Python: `/count-active` is reached over HTTP, where an absent optional query param and an explicit null are the same value, so "don't scope" and "scope with no identity" would collapse on the wire.

The predicate is spelled as the readers' allow-list, not the shorter `!= 'scope_agent' OR agent_id == caller`. `visibility` is plain Text with no CHECK, so the two differ on any value outside the three: the allow-list omits it, the negation counts it. Four cases in `TestVisibilityFiltering` pin this against real rows through `sc.count_active` — the whole stack, so the identity surviving the HTTP hop is covered with the SQL. Both wrong spellings fail a case.

### Low — `_NO_AUTH` not skipped

Correct. Added, with a test asserting the guard performs *no* lookup on that path — the assertion has to be on the lookups, since the wrong behaviour returns the same 200 as the right one.

One note, because it changed the shape of the fix. I first went further and collapsed `_check_auth` into a lookup on a shared sentinel→error dict, since `_check_auth` is a second hand-kept enumeration of the same roster and its drift is the dangerous one — a sentinel it does not know falls through to `return None`, i.e. auth *passes*. That broke `test_tool_error_codes_inventory`: `tests/_error_codes.py` resolves a tool's reachable codes through the module-level constants a function **names**, so dropping the `_AUTH_ERROR`/`_ADMIN_ERROR` references hid `UNAUTHORIZED`/`FORBIDDEN` from the scan and made 12 specs declare codes nothing appeared to emit. Runtime behaviour was identical, which is precisely why that test exists. Reverted; `test_every_sentinel_is_refused` ties the two lists together instead, and fails alone when a branch is deleted.

### Also in this push

`enforce_fleet_read`'s four-term condition was copied at all three read routes (round 1 added the third) and is now `_gate_fleet_read` — `_resolve_scoped_read`'s docstring states that condition in prose as the thing callers must match, so this makes the prose executable.

`tests/test_authz_gate_inventory.py` is the file that should have caught the fleet-gate half and did not: its three invariants cover mutating routes and the `enforce_self_agent` axis, and a read route taking `fleet_id` falls outside all three. The fourth axis needs a route-by-route allowlist and a weaker attribution than the others use (`_classify` does not follow imported helpers, and `enforce_fleet_read_many` is already in its blind-spot list), so it is its own PR. The gap is now recorded in that file's preamble so three passing invariants are not mistaken for coverage.

### Verification

ruff clean; mypy clean apart from two pre-existing `common/llm/providers/openai.py` httpx-version errors, confirmed on a stashed clean tree. Full suite **33 failed / 7031 passed — byte-identical failure set to the clean-main baseline, zero introduced**. Each fix individually reverted and confirmed to fail its test, including both wrong spellings of the visibility predicate and the sentinel drift.

Eldad-Caura:
@erni-a over to you — both review findings applied in `92fafbfd`, verdict minimized above.

Worth your eye on two judgement calls rather than the diff as a whole:

1. `_visibility_scope_clause` is new and has one caller. Four pre-existing readers in `postgres_service.py` still carry the predicate inline; I left them alone rather than sweeping a security-critical predicate across untouched call sites on a PR already in review. Say if you'd rather that landed here.
2. `oss-0814-m-29` is left open deliberately — `enforce_mcp_plan_limits` is off by default with a documented staged rollout, so flipping it is a product decision, not a code change. Eldad's call, noted in the commit body.

Zero new suite failures against the clean-main baseline (33/33 identical). Not merging — Eldad merges.

erni-a (dismissed):

Eldad-Caura:
@claude

Eldad-Caura:
@erni-a pelra asrpprove

erni-a (approved):

PR #1554, opened 2026-09-17
fix(core-api): make short-term memory report what actually happened

Eldad-Caura:
Batch 2 of the core-api audit remediation: the eight findings filed against short-term memory. Seven real defects, one duplicate. They share a shape — **STM told the caller something that was not true** — which is why the fixes are pinned by what the code RETURNS rather than by what it logs.

Every finding was reproduced against the running code first. Two came back narrower than filed, and both are called out below.

## The two mediums are one bug seen from both ends

`RedisSTM.post_note` returned `None` whether the entry was stored, the connection was absent, or the pipeline raised — failures logged at DEBUG, invisible outside debug logging. `WriteSTMNote` then built the receipt unconditionally: fresh uuid, TTL from settings, 201. The caller was left holding an id that resolves to nothing, and **a 2xx is the one answer that stops it retrying**.

Mutations now report whether they happened; the step and the clear routes refuse with 503 (`UNAVAILABLE` — the request was fine, retrying is right) instead of minting a receipt. Reads still degrade to an empty list, which is an honest answer to "what is in short-term memory". Degrading a read is graceful; degrading a mutation silently is fabricating a receipt.

**Clears were not in the audit and are fixed anyway.** The module now states the rule as reads-degrade / mutations-report, and a rule applied to half the mutations is one the next reader cannot rely on. It is also the more alarming half: the caller is told the notes are gone, and they are still there on the next read.

## The latch was already documented — as a workaround

`_redis_available = False` was never cleared, so one refused connection at first use (core-api racing Redis on a cold start is the ordinary way to get one) disabled Redis for the life of the process. `redis_healthy()` **already describes this in its own docstring** and bypasses `_get_redis()` to dodge it. That workaround is the evidence: the health probe was made honest and every actual consumer was left on the latch. Fixed at the source with a 30s cooldown.

Scope is stated in the code rather than implied: this covers the never-connected case, which is what the audit filed. A connection that succeeds and later dies is returned from cache before the check; marking it down on command failure is a wider change and is not made here.

## Two findings came back narrower

**l-25** says the STM path returns a shape incompatible with `response_model=MemoryOut`. Probed: **201 with the correct STM body**. The handler returns a raw `JSONResponse`, which bypasses `response_model` — so the server was always right and only the published document was wrong, which a generated client discovers by failing to parse a *successful* write. Fixed as a union. (An extra `responses` entry alongside `response_model` was tried first and left FastAPI emitting a `$ref` **with** an `anyOf` sibling — replacing one ambiguity with another.)

**l-27** says the docs deny a REST write path that exists. True, and false in a specific direction: the write is gated on the same `USE_STM` setting, so the claim held for the hosted deployment (gate off) and was false for exactly the self-hosted readers its last sentence addressed. A doc accidentally right about the configuration nobody reading it is running is the harder kind to catch.

That sentence was published in **three** places — the per-operation text, the OpenAPI *tag* (what a reader sees in the sidebar before opening an operation), and a test's own docstring. It is now one constant rendered by all three. `app.py`'s SAFE-01 note already argues this case twenty lines below the stale copy: *"the same fact written twenty-five times has twenty-five chances to go stale, and the copy that gets forgotten is the one someone reads."*

## Found by the work, not by the audit

The STM-disabled refusal existed twice and had drifted. The read door's message was rewritten once because "Set USE_STM=true" is advice the hosted caller who reaches it cannot act on; the write door kept the old text. One capability, two doors, different instructions. Now one constant, asserted by driving both doors rather than by scraping their sources for phrases.

## Verification

ruff clean; mypy clean apart from two pre-existing `common/llm/providers/openai.py` httpx-version errors, confirmed against a clean tree. Full suite **33 failed / 7046 passed — byte-identical failure set to the clean-main baseline, zero introduced**.

Nine reverts, nine matching failures: each fix fails its own test when backed out, including both wrong spellings of the visibility predicate's analogue here — the unconditional receipt, the permanent latch, the hardcoded TTLs, the missing sweep, the unconstrained promote body, the single-shape spec, the false doc claim, the drifted message, and the swallowed clear.

`/simplify` caught three things worth naming, all mine and all the same class this PR is about: a comment asserting a 2s socket timeout was needed because a reconnect "blocks on the OS default" (redis-py 8.0.1 defaults both to **5s** — verified; the change was dropped, and it had a real cost since `cache_set_nx` fails OPEN); the untrue sentence corrected in one of three files; and the reads/mutations rule applied to half the mutations.

## Follow-up, not in this PR

`tests/test_broker_contract.py` already documents the mechanism behind l-25 — *"FastAPI skips response validation for a Response object, so a `response_model` alongside one documents a guarantee the code does not make"* — but its ratchet walks only the broker spec, so the class recurred immediately outside it. An app-wide invariant is the right mechanism-level fix and needs the 24 `JSONResponse` sites across six route modules triaged and annotated, which is its own change.


Eldad-Caura:
One test added in `2f62405b` after the verdict, prompted by what the review checked rather than by a finding.

The review confirmed that `HTTPException` raised inside a pipeline step propagates rather than being swept into `result.failed`. That is correct — `pipeline/runner.py` has `except HTTPException: raise` — but it is load-bearing for this PR and nothing here was pinning it: `_run_write_pipeline` turns a failed run into a blanket `500 STM write pipeline failed unexpectedly`, so if that ever changed, the 503 this PR adds would silently become a 500 and the fix would be undone without a single test going red. My existing test stopped at the step, one layer short of the answer the caller actually receives — which is the thing this whole PR is about.

`test_the_503_survives_the_pipeline_runner` drives `_run_write_pipeline` with a dropping backend and asserts 503. Verified it fails when the runner's re-raise is removed.

No behaviour change; test-only.

Eldad-Caura:
@erni-a over to you — CI green (14/14), review verdict clean and minimized above.

Two things worth your eye rather than the diff as a whole:

1. **Clears were not in the audit and are fixed here anyway.** The module states the rule as reads-degrade / mutations-report, and I'd applied it to two of the four mutations — `clear_notes`/`clear_bulletin` still swallowed failures while the route answered `{"ok": true}`. A caller told its notes are gone, whose notes are back on the next read, has no reason to retry. Say if you'd rather that had been split out.

2. **Two findings came back narrower than filed** and are recorded that way rather than closed whole: `l-25` (the server was always right, only the OpenAPI document was wrong — the STM branch returns a raw `JSONResponse` that bypasses `response_model`) and `l-27` (the doc claim held for hosted, where the gate is off, and was false for exactly the self-hosted readers its last sentence addressed).

Zero new suite failures against the clean-main baseline (33/33 identical). Not merging — Eldad merges.

erni-a (approved):

PR #1555, opened 2026-09-17
fix(core-api): make the update path's derived state follow the write

Eldad-Caura:
Nine audit findings (owner=Eldad, MEDIUM+LOW), all the same shape: a mutation
changed a row and something derived from it went on describing the row as it
used to be.

## The entity graph a content edit leaves behind — `m-53`

Extraction only ever ADDS (`ON CONFLICT DO NOTHING`, no delete branch), so a
memory edited from "Alice joined Acme" to "Bob joined Globex" kept Alice and
Acme — linked, related, and still ranking the row in recall for names its
content no longer contains. The links aren't merely stale; they're assertions
about text that is gone.

`memory_purge_entity_artifacts` already did exactly these deletes — but
**refuses any row that isn't soft-deleted**, and its docstring says why: *"only
purge what governance actually dropped" should not be an invariant that lives
only in the callers' heads.* A boolean parameter would put it straight back
there.

So the delete sequence became `_delete_entity_artifacts`, which takes an
eligibility **predicate**, with two public names over it:
`memory_reset_entity_artifacts` (live rows) and `memory_purge_entity_artifacts`
(dropped rows). A clause rather than a flag — a flag reads as an option, and
each name exists precisely to fix its own answer to "which rows may I destroy".

The purge's safety paragraph is **rewritten, not left standing**. It warned that
a caller reaching it by mistake *"would wipe the entity graph of a live, fully
visible memory"* — which this PR makes a supported operation. That guard is
about provenance now, not safety.

### A fourth reference the anti-joins never checked

Sharing the implementation surfaced a latent bug in the existing purge path:
`memories.subject_entity_id` is a FK with `ON DELETE SET NULL`. An entity that
was another live memory's RDF subject — but held no links and no relations —
satisfied all three anti-joins, so it was deleted and that memory's subject
silently became NULL. Rare under governance drops; routine once the reset runs
on ordinary edits.

Bounding that fourth anti-join to the candidate set is **load-bearing, not an
optimisation**: `subject_entity_id` is nullable and almost always NULL, and a
bare `NOT IN` over a set containing NULL matches nothing in SQL — which would
have turned entity deletion off entirely while every link/relation count
assertion still passed. `test_the_subject_anti_join_does_not_disable_deletion`
pins that. The other three are bounded too; they were uncorrelated, so a
four-link memory scanned the tenant's whole link and relation tables three times.

The reset is **unconditional, outside the `entity_extraction_enabled` branch** —
gating it on that setting would leave a tenant who turned extraction off holding
the links forever, with nothing left to replace them.

### Links the same request named

The `entity_links` add runs before the storage PATCH and the reset after it, so
a PATCH carrying both `content` and `entity_links` had its own links deleted
milliseconds after inserting them: 200, an audit record reading "added N links",
and a response holding none. `test_entity_links_are_additive` pins that this
endpoint must never silently delete links a caller did not name. Re-applied
after the reset rather than reordered — moving the add later would put its 422
*after* the content write had landed.

## The dedup gate keyed on neither half of its index — `m-04`

It passed no `fleet_id`, and storage's default is the **fleetless** group, so
for any fleet-scoped row it could never match. Invisible because `_content_hash`
already mixes the fleet in — the hash *looked* like it carried the scope while
the query quietly added a contradictory second one.

Fixing only the fleet turns a gate that never fired into one that fires too
**wide**: it refuses an edit whose content matches another agent's row in the
same fleet, which `uq_memories_live_content_hash` admits deliberately
(CAURA-721 — two agents recording identical content are two independent
observations). Both legs now. The 409 also reads `dup["memory_id"]`, the key the
endpoint actually returns, rather than `id` — so on the rare path where it did
fire it named `existing_id: null`.

## A retracted supersession left its victim outdated forever — `m-51`

Clearing `supersedes_id` is half a retraction. The superseded row was set
`outdated` *because* this row superseded it, and the detector never revisits a
row whose conflict is gone. `_revert_superseded_row` mirrors Path-C's semantics
including its refusal to touch a status it did not set — reviving an `archived`
row would be a second bug wearing the first one's fix.

The `("outdated", "conflicted")` tuple was inline at **four** destructive-write
guards while a fifth site defined it under a comment instructing readers to
hand-sync it. It is now `CONTRADICTED_STATUSES` in `core_api.constants`.

## A failed re-embed said nothing — `m-50`

Every create path sets `embedding_pending` and schedules a backfill; update did
neither. The row sat semantically unsearchable until the nightly
NULL-embedding sweep happened past it, with the response reporting nothing amiss.

## Deleting a document left its minted memory — `m-24`

Searchable forever, with nothing in the product able to reach it — the document
it was provenance for no longer exists to be deleted again.

The mint has **two** entry points, so the delete needs two. This adds
`safe_unmint_doc_memory` beside its mint counterpart, called from the REST route
*and* from MCP `caura_doc op=delete`, which had no un-mint at all. The
provenance pair comes from `doc_provenance` so the mint and the match can't drift.

## Read-your-write on four paths — `l-39`, `m-32`, `l-03`

- `update_memory`'s existence check answered **404 for a memory created moments
  earlier** — the one answer a client can't tell from "you deleted it".
- Its response re-fetch could echo the caller their own pre-edit content as the
  result of a successful edit.
- The skills inbox loaded every read-modify-write from a replica, so approve
  could revert a just-saved edit while the TOCTOU re-check re-read the same
  stale copy and agreed with itself.
- The settings cache re-cached PRE-update values for a full 5-minute TTL after
  an invalidation — a tightened governance control appearing to land and not
  taking effect.

Priming the settings cache post-write instead **is not equivalent and was
dropped**: `subscribe(broadcast=True)` gives every process its own subscription,
the publisher included, so the writer receives its own `SETTINGS_CHANGED` and
evicts what it just primed. The reload happens either way — so the reload is
what had to be correct.

## The inbox bound against a field that is never set — `m-33`

It read `data["target"]["slug"]`. `target` carries `target_content_hash` and
nothing else, so the lookup always came back `None` and the binding gate refused
every `kind='update'` edit with *"no live skill exists"* — naming a skill that
does. `_binding_target_slug` reads `data["slug"]`, the key
`validate_and_normalize_skill_write`'s own gate resolves by.

## Verification

- `tests/test_mutation_consistency.py` — 21 new tests
- `core-storage-api/tests/test_h02_purge_entity_artifacts.py` — 6 new cases
  covering the subject anti-join, its NULL trap, and both guards
- **Every fix was revert-checked**: each test confirmed to fail without it. One
  check falsely passed at first — the removal landed on a different call site
  sharing the same text — and was redone by line index, where it failed
  correctly.
- Full suite: 7066 passed, **zero introduced** against the clean-main baseline
- core-storage-api: 413 passed
- ruff check + format clean; mypy per-package clean apart from the two
  pre-existing `common/llm/providers/openai.py` httpx errors


Eldad-Caura:
Round 2 — two applied, one declined with evidence.

## High: caller-curated links destroyed on a later content edit — **applied, fixed at the mechanism**

Confirmed, and worse than a same-request problem. My first fix re-applied the links named in *that* PATCH, which never protected a link curated in an **earlier** request — the case that actually matters. A caller tags a memory with a project entity its text never names; extraction mines text, so it will never recreate that link; the next content edit deletes it with no signal, because that PATCH need not mention `entity_links` at all.

`tests/test_entity_links_are_additive.py` had already ruled on exactly this shape — a shipped endpoint must not silently DELETE links a caller did not name — so reaching the same outcome through a different code path is the same destructive change.

Nothing cheaper than provenance is correct here, and provenance turned out to need no inference: **the three writers are already distinct methods.** Migration 048 adds `memory_entity_links.source`; `memory_add_entity_links` and `entity_create_link` stamp `caller`, `entity_bulk_upsert_links` stamps `extraction`, and the reset deletes only extraction's. The purge keeps **no** carve-out — a governance-dropped memory leaves nothing behind, curated or mined, and "the caller asked for this one" is not an exemption from a retention decision.

Two consequences worth naming:

- **Narrowing the links narrows the orphan candidates with them.** An entity whose only remaining link is a caller's is no longer a candidate, so it cannot be hard-deleted out from under that link. That falls out of taking candidates from the rows actually deleted.
- **The column defaults to `caller`, which is the conservative value and deliberately not the accurate one.** Most existing rows are extraction's. Defaulting to `extraction` would make the first content edit after deploy delete every caller-curated link that exists today — the one irreversible outcome available here. The cost is stated in the migration rather than hidden: rows predating the column keep that default for good, because extraction's upsert leaves `source` out of its `ON CONFLICT` SET so re-mining cannot take a caller's row over.

This also let me **delete** the same-request re-apply — with provenance those links survive on their own, so a storage round trip went with it.

## Medium: a failed post-commit reset turned a landed write into a 500 — **applied**

Correct, and my comment arguing the opposite was wrong. It runs after the content patch committed and after the audit hook fired, so propagating hands the caller a 500 for a write that succeeded — and the retry repairs nothing, because the row's content now equals what was sent, making `content_changed` False and skipping the reset, the embed-retry scheduling and the contradiction re-check **forever**. `_revert_superseded_row` already took this position four lines away; the two had no business disagreeing.

## Low: re-asserted `subject_entity_id` cleared — **declined, not reachable**

The premise is that `simple_fields` skips the write when `old_val == new_val`, leaving the earlier unconditional `None` in place. Those two values can never be equal:

- `mem` comes from storage through `orm_to_dict`, which coerces **UUID → str** (`core-storage-api/src/core_storage_api/schemas.py:33-42`)
- `data.subject_entity_id` is `UUID | None` on `MemoryUpdate` (`core-api/src/core_api/schemas.py:335`)

So `old_val != new_val` is true for *any* named subject, including one re-asserted unchanged, and the caller's value is always the last write. I added `test_an_explicit_subject_beats_the_content_auto_clear` pinning that rather than adding a guard — a guard would defend a state that cannot occur, and the reason it cannot is worth recording.

## Also caught, by the repo's own ratchet

`test_no_plain_create_index_on_large_tables` rejected my first migration: the partial index now builds `CONCURRENTLY` inside an `autocommit_block`, with the invalid-index guard that 041/046 use. Chain head assertion bumped to 048.

## Verification

- core-api: 7068 passed, **zero introduced** against the clean-main baseline
- core-storage-api: 419 passed, through the real migration chain on a rebuilt database (column, `'caller'` default and partial index all confirmed present)
- 6 new storage tests for provenance, each revert-checked — including `test_extraction_writes_extraction_provenance`, the over-fix guard: if extraction stopped stamping, its rows would inherit the conservative default and become undeletable, which is the stale-link bug back in full with every other assertion still green
- ruff + mypy clean apart from the two pre-existing `openai.py` httpx errors

@claude

Eldad-Caura:
Round 3 — both applied, both were real.

## High: the superseded-row revert ran before validation that can still abort

Confirmed, and the ordering is exactly as described: the revert sits at `memory_service.py:3782` while two later checks can still raise — `metadata` null-in-merge-mode at `:3900` (400) and the `entity_links` ownership check at `:3962` (422) — with the row's own write not landing until `:3989`.

The consequence is the part worth naming: `_revert_superseded_row` is a live PATCH **against a different row**. A 422 on an entity the tenant doesn't own left the superseded memory flipped back to `active` — resurfacing in recall — while the edit that was supposed to justify reviving it was rejected, with nothing to roll it back. That is worse than the stale-`outdated` state the revert exists to fix, because it is a silent revival rather than a visible staleness.

Moved into the post-commit block beside the graph reset, keeping the same non-raising policy. The principle the fix encodes, and what the comment now says: **building a patch must not have side effects; only a committed edit may retract anything.** `patch["supersedes_id"] = None` stays where it was — that one is a patch key, so it lands with the edit or not at all.

## Medium: `embedding_pending` never cleared on a successful re-embed

Confirmed, with one correction to the diagnosis worth recording. The flag is **not** unclearable in general — `core_worker/clients/storage_client.py:225` clears both homes when the deferred embed succeeds, and deliberately both (`{"embedding_pending": False, "_system": {"embedding_pending": False}}`) because the C25 read view gives `_system` precedence.

The gap is narrower and real: a successful **inline** re-embed schedules no worker task at all — that only happens on the failure path — so for that path nothing was ever going to clear it. A row that failed to embed once, here or at create time, and is later edited with a successful synchronous re-embed keeps reporting `embedding_pending: True` while fully embedded. Since `MemoryOut.metadata` documents the flag's *absence* as "that stage ran inline", callers read it as authoritative.

Now written in both directions through `set_system_value`, which writes both homes for the same reason the worker does. `test_a_successful_reembed_clears_the_pending_flag` asserts `False` in both, rather than merely "not True" — the previous assertion passed on an omission and so could never have caught this.

## Verification

- core-api: 7069 passed, **zero introduced** against the clean-main baseline
- core-storage-api: 419 passed
- Both fixes revert-checked. The first check initially reported a pass — the edit script had aborted on a stale needle before writing, so it tested unmodified code; redone by line index it fails correctly.
- ruff + mypy clean apart from the two pre-existing `openai.py` httpx errors

@erni-a — over to you. CI was green on the previous commit and is re-running now.

erni-a (approved):

PR #1556, opened 2026-09-17
fix(plugin): create openclaw.json so a fresh install has callable tools

ran-taig:
## What a fresh install did

`tools.alsoAllow` is what makes a registered tool survive an OpenClaw `tools.profile`, and exactly two things in the product ever write it. Both gave up on the same input — a box where `~/.openclaw/openclaw.json` does not exist yet:

- `core-api/src/core_api/routes/plugin.py:494` — the install script's step 7 is wrapped in `if [ -f "$CONFIG_PATH" ]`, else `WARNING: ... not found — you will need to configure allowlist manually`.
- `plugin/src/config.ts:131` (pre-fix) — `autoFixAllowlist` returned `openclaw.json not found at <path>` and wrote nothing.

The plugin is discovered from `~/.openclaw/plugins/memclaw` and loads fine, so `register()` runs and all 11 tools register (`plugin/src/index.ts:186`). Then the self-heal declines — on that boot and on every boot after it, because the drift gate (`shouldRunAutoFix`, `plugin/src/config.ts:300`) re-runs, reads nothing, and declines again. So nothing ever put a tool name into the allowlist, zero Caura tools were callable under any `tools.profile`, and the only trace was a line in gateway.log naming a file the operator then had to write by hand.

**Before** (built plugin, stub OpenClaw host, no `openclaw.json`):

```
WARN [caura] Auto-fix allowlist failed: openclaw.json not found at <HOME>/.openclaw/openclaw.json
LOG  [caura] BOOT: plugin v2.22.2, 11 tools registered
config exists after register: false
```

**After**, same input:

```
LOG  [caura] Config auto-fixed: created <HOME>/.openclaw/openclaw.json, plugins.entries,
     plugins.load.paths, plugins.slots.memory, plugins.slots.contextEngine, caura_recall,
     caura_write, caura_manage, caura_doc, caura_list, caura_entity_get, caura_tune,
     caura_insights, caura_evolve, caura_stats, caura_keystones.
     Restart OpenClaw to activate all 11 tools.
```

## The fix

`readOpenClawConfig` returns null for two very different states and the old check conflated them. A MISSING file is safe to create; an UNPARSEABLE one is not — creating there would silently replace a config we cannot read, losing every unrelated OpenClaw setting in it. So split on `existsSync`: create from `{}` when absent and let the existing steps 2-6 build exactly what the install script writes, and keep refusing — now with an error saying why — when the file is present but unreadable.

`plugins.allow` stays uncreated on the create path. Step 1's CAURA-000 invariant is the whole point of that branch: a one-entry allowlist flips a permissive install into a restrictive one and locks out every other plugin, and a fresh file must not be a back door around it.

Not touched, deliberately: the manifest already declares `contracts.tools` and `tool-definitions.test.ts:246` already pins it to `CAURA_TOOLS`, so registration itself was never the gap — the allowlist write was.

## Tests

Five cases in `plugin/src/config.test.ts`, following the existing `_autoFixWithConfig` temp-HOME idiom:

- creates the config and leaves every declared tool in `tools.alsoAllow`
- the created config also claims both slots and the load path
- creating it still does NOT create `plugins.allow` (CAURA-000)
- creates `~/.openclaw` when the directory is absent too
- refuses to overwrite a present-but-unparseable config, byte-for-byte

Pre-fix proof: with `plugin/src/config.ts` reverted and only the tests applied, `456 tests, 452 pass, 4 fail` — the four new create-path cases. With the fix: `456 tests, 456 pass, 0 fail`.

## Gates

- `cd plugin && npm ci && npm test` (re-runs `tsc` via `pretest`) — 456/456
- `python3 scripts/legacy_name_ratchet.py --base origin/main` — No new lines
- `python3 scripts/do_not_touch_sentinel.py` — all 35 protected strings survive

No files outside `plugin/` touched, so no Python package gates apply.

## Not addressed here

The same load also warns that `CAURA_API_URL` uses plain HTTP, so the API key travels in cleartext. That is a security issue rather than an ergonomics one and is being filed as its own tracker row; no code for it in this PR.

Audit tracker: ax-0917-h-10


ran-taig:
@claude

ran-taig:
@erni-a

erni-a (approved):

PR #1557, opened 2026-09-17
perf(recall): stop shipping the result set twice and the telemetry three times

ran-taig:
Three high-severity audit findings against one surface — the `/recall` response
envelope and the parameters that shape it. All three are byte counts, so the
before/after table is the deliverable.

## Validation (all three reproduce on `origin/main` @ 4a2c9b4c)

**ax-0917-h-03 — result set serialised twice.**
`core-api/src/core_api/services/recall_service.py:175-176, 225-226, 325-326` —
all three branches of `summarize_memories` emitted `"memories": _dumps` and
`"items": _dumps`. The Python list is genuinely shared, so the "materialise
once" comment was true and irrelevant: `json.dumps` writes it out both times.
Measured **49.6%** of a 5-row brief, **49.9%** of a 20-row one.

**ax-0917-h-04 — per-row telemetry triplicated.**
`system_metadata.set_system_value` (`:122-124`) writes every platform value to
BOTH `metadata["_system"]` and the legacy top-level key; the read side then
derives a third copy into `MemoryOut.system_metadata`
(`memory_service.py:738, 792`). On a representative row: 659 B (legacy) +
630 B (`_system`) + 630 B (`system_metadata`) = **1,919 B/row**, against
60-82 B of actual content. Within the audit's 1,748-3,102 B range.

**ax-0917-h-05 — `limit` silently ignored.**
`SearchRequest` (`schemas.py:802`) declares `top_k`, not `limit`, and inherits
pydantic's `extra="ignore"`. Pre-fix probe on the stashed tree:

```
PRE-FIX items present: True
PRE-FIX metadata has llm_ms: True | has _system: True | system_metadata has llm_ms: True
PRE-FIX limit=2 -> top_k: 5 | model_extra: None
```

`model_extra` is `None`: the key was not merely unused, it was **gone** before
any route could see it. That is the root cause, not `limit`.

No landed or in-flight work overlaps: `git log -- recall_service.py` last
touched it at #1357, and no open PR touches the envelope or the metadata
boundary (#1243 / #1467 are client-side recall fixes).

## Consumer audit (done before moving anything)

| Reader | Reads | Verdict |
|---|---|---|
| `clients/python/src/caura_client/models.py:70-73` | `memories`, falls back to `items` | recall alias optional — safe |
| `clients/typescript/src/index.ts:162` | `data?.memories ?? data?.items` | same — safe |
| `clients/python/client.py:100-104`, `clients/typescript/index.ts:139-144` | `items` — but on **`/search`** | untouched |
| `plugin/src/transport.ts:129` (`parseSearchItems`) | `items ?? results` | only ever fed `/search` responses (`context-engine.ts:637, 1022`); the plugin's `caura_recall` passes the brief through raw (`tool-definitions.ts:459-465`) |
| `core_api/mcp_server.py:1339` | `results`/`items` on the MCP payload | separate C31/D1 alias — **not touched** |
| `system_metadata` / `metadata._system` | nothing outside `core-api` | zero readers in `clients/`, `plugin/`, `e2e/` |
| `metadata.embedding_pending` | `plugin/tools.json:919`, `docs/self-hosting.md:176-179` | both document the **write** response — h-04 is scoped to the recall envelope, so untouched |

## What changed

**h-03** — `items` becomes a choice: `summarize_memories(..., items_alias=True)`,
surfaced as `RecallRequest.items_alias`. **The REST default does not change.**
`RecallResponse.items` is published OpenAPI and `docs/public-api-stability.md`
makes REST response shapes part of the SemVer contract, so flipping that
default is a major release's call, not a perf patch's. The **MCP brief** —
whose JSON body that document does not pin (it pins tool names, parameter
names and op-dispatch values) — opts out, because `caura_recall(include_brief=true)`
was carrying the identical rows **four times**: `results` + `items` on the
payload, `memories` + `items` inside the brief.

**h-04** — new `strip_platform_metadata()` beside its C25 siblings; the recall
envelope keeps **one** location, `system_metadata`. That is the documented C25
read surface, the merged view that is correct for pre-C25 rows, and the one
`MemoryOut.metadata`'s own docstring already points at ("reading them from
`metadata` is deprecated"). **Nothing is deleted**: the JSONB column is
untouched, the same values are in the same response, and `GET /memories/{id}`
still returns the raw bag. C25 called the mirror a one-release measure (#967,
2026-08-25); we are fifteen minor releases past that. Deliberately scoped to
the recall envelope rather than `_memory_to_out` — the write response is where
callers poll `metadata.embedding_pending`.

**h-05** — see the argument below.

## Before / after (measured through `summarize_memories` itself)

Micro-benchmark over the repo's own serialisation path with synthetic rows,
following the #1466 precedent. Bytes of the JSON envelope:

| scenario | before | REST default now | `items_alias:false` / MCP brief |
|---|---:|---:|---:|
| 5 rows, short memories (default `top_k=5`) | 28,518 B | **16,878 B (-40.8%)** | **8,496 B (-70.2%)** |
| 20 rows, mixed lengths | 117,573 B | **71,013 B (-39.6%)** | **35,564 B (-69.8%)** |
| `limit: 2` — was 5 rows, now 2 | 28,518 B | **6,836 B (-76.0%)** | **3,475 B (-87.8%)** |

Per finding, on the 5-row case: h-03 alone -49.6%, h-04 alone -40.8%,
combined -70.2%. h-04's saving lands on **every** REST caller by default;
h-03's lands on the MCP brief today and on any REST caller that opts in.

## h-05: alias vs reject — the decision, and why

**Shipped: alias + visibility. Not shipped: rejection.**

Rejecting unknown params outright is the honest fix, and this repo has already
decided it is the wrong one *here*. `schemas.py:24-51` states the SAFE-01
asymmetry as a product decision — write bodies `extra="forbid"`, search/filter/
query bodies permissive — and `tests/test_unknown_field_rejection.py` pins an
unknown field on `/search` returning 2xx *"precisely so a later pass that
'finishes the job' fails loudly instead of quietly breaking integrators."*
This PR is that later pass, and it does not finish the job.

So the fix is the one this repo already used for the identical trap: C31/D2
absorbed `memory_type` (the MCP spelling, silently dropped by the same
`extra` contract) as an `AliasChoices` alias. `limit` gets the same treatment,
with `top_k` first so it wins when both arrive.

But aliasing `limit` only fixes today's symptom — the next plausible guess
would be just as silent. So the root cause is addressed too: `extra="ignore"`
-> `extra="allow"`, which keeps the 2xx and stops *discarding* the evidence, so
the route can report it. Two channels: a `logger.warning` for the operator, and
an A28 `warnings` entry for the caller — the only one an autonomous agent can
act on, since it does not read our logs. Capped at 20 names; a payload-size PR
should not echo unbounded caller input.

**What a future breaking change would need:** flip `SearchRequest` to
`STRICT_WRITE_BODY`, delete the `extra="allow"` note and the warning helper,
and rewrite `test_search_still_accepts_unknown_field` — under a major version,
after this PR's `unrecognized_parameters` warning has run long enough to show
in logs who is actually sending junk. The same major should flip
`items_alias` to default false. Both are one-line changes now.

## Tests

`tests/test_ax0917_recall_envelope.py` (new, 17 tests) + amended
`tests/test_c4_recall_items_alias.py`.

**Failed pre-fix** (verified by stashing `core-api/src`):

- the new module fails at import — `RecallRequest`, `strip_platform_metadata`
  and the `items_alias` parameter do not exist pre-fix;
- `test_mcp_recall_brief_carries_memories_without_the_alias` (amended C4 test)
  fails on the stashed tree with the other 9 in that file passing — i.e. the
  default REST/service shape is genuinely unchanged and only the MCP brief moved;
- the pre-fix probe quoted above pins all three old behaviours directly.

Coverage: the alias drops only when asked and the REST default still carries it;
the empty/disabled/brief branches agree on shape; telemetry appears exactly once
per row and is still readable at `system_metadata`; `strip_platform_metadata`
preserves the `null`-vs-`{}` distinction and never touches caller keys; `limit`
resolves to `top_k`; explicit `top_k` wins and the losing `limit` is reported;
unknown params warn on both `/recall` and `/search`, are still accepted (2xx),
and cost nothing when there are none.

## Gates

- `pytest tests/test_ax0917_recall_envelope.py tests/test_c4_recall_items_alias.py
  tests/test_mcp_recall.py tests/test_c33_openapi_completeness.py
  tests/test_authz_gate_inventory.py tests/test_weight_source_provenance.py` —
  **64 passed**.
- Wider batch incl. `test_api_memories` / `test_ingest_commit`: 22 failures,
  **byte-identical to clean `origin/main`** (unmigrated Postgres —
  `column memories.status_changed_at does not exist`).
- `gen_broker_openapi.py --check` — regenerated and committed; the only diff is
  `SearchRequest.additionalProperties: true` plus the `top_k` description.
- `ruff check` / `ruff format --check` clean; `mypy src/` — no issues in 205 files.

## For the reviewer to weigh

1. **h-04 is the one visible behaviour change on the default REST path.**
   `/recall` rows no longer repeat platform keys inside `metadata`. It rests on
   C25's own declared sunset and on the same values being present under
   `system_metadata` in the same response — but if you would rather it were
   flag-gated too, say so and it becomes one more parameter.
2. **Two different defaults for one flag** (REST on, MCP off) is deliberate and
   argued from `docs/public-api-stability.md`, which pins REST response shapes
   and MCP tool *signatures*. If you read that document as pinning MCP bodies
   too, the MCP opt-out should come out.
3. **Residual duplication not addressed here:** the MCP payload still emits
   `results` + `items` (C31/D1's permanent alias, explicitly a separate
   contract), and `brief.memories` still repeats rows the payload already
   carries. Both are follow-ups, deliberately left alone.

Audit tracker: ax-0917-h-03
Audit tracker: ax-0917-h-04
Audit tracker: ax-0917-h-05


ran-taig:
@claude

ran-taig:
Checked all four items against the branch — three don't apply, and the one that does is already reflected. I ran the generator rather than reasoning from the diff: `broker OpenAPI baseline up to date (8 operations)`, exit 0. Nothing to regenerate.

**`/recall` is not in `openapi.broker.json`, by design.** The file carries six paths — `/api/v1/health`, `/api/v1/memories`, `/api/v1/memories/bulk`, `/api/v1/memories/{memory_id}`, `/api/v1/search`, `/api/v1/version`. `gen_broker_openapi.py`'s own docstring is explicit that the baseline is *"a **subset** of core-api's full ~91-path OpenAPI surface: only the broker operations plus the schema / security components those operations reach (transitively)"* — the frozen v1 contract for what caura-daemon calls against cloud. `/recall` is not one of those calls.

I suspect the confusion is that the enumerated broker operations include **`memory_recall`** — but that is the MCP dispatcher's tool name for `GET /api/v1/memories/{memory_id}` (fetch one memory by id), not the `/recall` endpoint.

Point by point:

1. **`RecallRequest` in components + `/recall` requestBody** — there is no `/recall` path in this file to carry a requestBody.
2. **`RecallResponse.items` required → optional** — `RecallResponse` is not in `components.schemas` at all. What is there is `SearchResponse`, and its `items` is still `required` — deliberately, since h-03 is scoped to recall and leaves `/search` untouched. Making search's `items` optional would be exactly the breaking change the `oasdiff breaking` gate exists to catch.
3. **`warnings` + `SearchWarning` reachable** — `SearchWarning` is already in `components.schemas` and `SearchResponse.warnings` already exists. Both pre-date this PR (A28).
4. **`SearchRequest.top_k` / `limit` alias** — this one applies, and it is in. The branch's entire baseline diff is `SearchRequest.additionalProperties: true` (the `extra="allow"` change) plus the `top_k` description gaining *"Also accepted as 'limit'; when both are sent, top_k wins and the response warns that 'limit' was not read."* No other broker path is affected — `SearchRequest` is the only changed schema they reach.

On the codegen note: nothing generates from this file. The TypeScript client is hand-written (no OpenAPI tooling declared) and the plugin's `transport.ts` hand-codes the response shapes with comments. The recall-side surface (`items_alias`, optional `items`, `warnings`) is documented in `docs/api-reference.md` and `docs/api-surfaces.md`, both updated in this PR.

One caution worth stating explicitly: adding Recall types to this baseline would widen a frozen broker contract to cover an operation the broker does not call. The generator's docstring records a prior incident where this file's scope being wrong let a breaking change through the gate — so the omission is load-bearing, not an oversight.

ran-taig:
On the REST/MCP asymmetry — pushed `a3dccb7d`.

Worth separating two things that were tangled together. The **split itself** is sound: `POST /recall` is listed as a SemVer-stable REST surface (docs/public-api-stability.md, Memory row), while that same document pins MCP tool *names and purposes* rather than response bodies. So unifying the defaults today means either breaking a stable surface (owed a major, plus the `BREAKING CHANGE:` trailer and `kind/breaking` label the doc requires) or giving up the ~70% reduction on the surface where agents actually consume. Neither is a good trade right now.

What was genuinely missing is an **end**. Nothing marked the alias deprecated or said when the two surfaces converge, which leaves a reader with one flag, two defaults, and no explanation — an inconsistency rather than a migration. That is the fair part of the objection, and it is now fixed:

- `RecallResponse.items` is marked `deprecated=True`, so it surfaces as `deprecated: true` in the live OpenAPI schema (verified: `items in JSON schema marked deprecated: True`, and no longer in `required`).
- The field description and `docs/api-reference.md` both state the **v4.0.0** removal target and point readers at `memories`.

That applies the same deprecate-then-remove convention this branch already leans on to justify h-04's default change (C25, #967) — which it had not applied to h-03. The asymmetry now reads as "deprecated everywhere; MCP already stopped sending it, REST stops at v4" rather than as two surfaces disagreeing indefinitely.

No behaviour change: `items` is still emitted by default on REST. Gates on the new commit — ruff clean, mypy unchanged (the 6 pre-existing mcp-drift errors, none in touched files), broker baseline still `up to date (8 operations)`, and 27 tests pass across `test_ax0917_recall_envelope.py` + `test_c4_recall_items_alias.py`.

If you would rather converge the defaults now instead of at v4, that is a deliberate breaking change and I am happy to make it — it just wants the major-version call and the labels.

ran-taig:
@claude

ran-taig:
Good catch — fixed in `99b74570`, and you were right that the docstring was overclaiming.

The read side *can* tell the two cases apart, which is what makes option (b) sound rather than merely pragmatic. `set_system_value` writes the top-level mirror for every platform value **except** one whose key the caller owns, so:

- **caller supplied it** → the top-level write was skipped, so `metadata[key]` holds the caller's value while `_system[key]` holds the platform's. They **differ**, and the caller's copy is kept.
- **platform produced it** → both writes happened, so the two are **identical**. That is a mirror, not caller data, and it is now dropped like any other platform key.

So equality is not a heuristic standing in for provenance — it *is* the provenance signal, given how the write path behaves. I went with (b) over (a) for that reason: (a) would persist write-time state to recover something already derivable at read time, and it would do nothing for the rows already in the database.

The one accepted false positive is the one you named: a caller who writes a value byte-identical to the platform's loses the top-level copy. That is not data loss on the wire — `system_metadata` carries the same value in the same response — and it is strictly better than shipping every platform-produced summary twice to protect a coincidence. A key with **no** `_system` counterpart is always kept, since pre-C25 rows have no namespace to compare against and an unprovable case must not be dropped.

Two tests rather than one, and the pair is the point:

- `test_h04_platform_produced_caller_ownable_keys_are_not_duplicated` — the case you asked for. **Fails pre-fix** (`1 failed, 1 passed` with only the source reverted).
- `test_h04_caller_supplied_caller_ownable_keys_survive` — the counter-case. It passes before *and* after, deliberately: without it, "drop `summary` unconditionally" would satisfy the first test while destroying caller data, which is exactly the clobber C25 exists to prevent.

Gates: 77 pass across the ax0917, both C25 suites, A8-tags and C4; ruff clean; mypy unchanged (the 6 pre-existing mcp-drift errors, none in touched files). The docstring now states the discrimination rule instead of the old blanket claim.

ran-taig:
@claude

ran-taig:
Confirmed and fixed in `3ed44186`. Reproduced it first — sending both spellings of each aliased field leaves `{'limit', 'status', 'memory_type', 'bogus_xyz'}` in `model_extra`, and all four were reported as "not read by this endpoint and had no effect". Wrong for `status`/`memory_type`, and misleading for `limit`.

The two cases are now split using the model's **own declared aliases**, introspected from each field's `validation_alias` rather than a hand-kept map — so a field that gains an alias later cannot quietly start being reported as junk:

- `superseded_parameter_alias` → `'limit' superseded by 'top_k'`, naming the spelling that actually won (the highest-priority choice not sitting in extras).
- `unrecognized_parameters` → genuinely unknown keys only.

On the count sentence, I dropped it rather than making it conditional, because the conditional version is unreachable: **`limit` is a declared alias now**, so sent alone it is absorbed and never reaches that branch, and sent alongside `top_k` it is reported as superseded above. There is no path where an unknown key is count-related in the way the sentence assumes. It was accurate when `limit` was genuinely unread — the alias this PR added is what made it unreachable.

**One existing assertion changed, and it deserves the scrutiny:** `test_h05_recall_warns_about_parameters_it_does_not_read` asserted `"top_k" in message` for a `bogus_param_xyz` request — it was pinning the misplaced tail. It now asserts the opposite with the reason inline, and the replacement behaviour is pinned positively by its own test rather than only by an absence.

Three tests added. Two fail pre-fix (superseded-not-unknown; no result-count claim). The third — a plain `limit: 2` still producing **no** warning — passes before and after deliberately: it guards the happy path h-05 exists to create against being broken by this change, which an absence-only assertion would not catch.

74 pass across ax0917, C4, unknown-field-rejection and C25; ruff and mypy clean (mypy unchanged at the 6 pre-existing mcp-drift errors).

ran-taig:
@claude

ran-taig:
Both fixed in `c0561937`. You were right that elimination cannot answer this — and the reason is worth stating, because it sets the limit of what any fix can do from `extras` alone.

`AliasChoices` takes the first spelling **present in the input**, so a spelling that was never sent is absent from `extras` for exactly the same reason a consumed one is. Those two states are indistinguishable from `extras`, which is why the old `next(...)` answered `a` for a caller who only sent `b` and `c`.

So the rule is now **exact or silent**, and provably correct for N ≥ 2. The winner outranks every loser, so it sits above the highest-priority spelling found in `extras`; a winner can be named when exactly **one** candidate remains above that point, and not otherwise:

| choices | lost | reported |
|---|---|---|
| `('top_k','limit')` | `limit` | `'top_k'` — one candidate, exact |
| `('a','b','c')` | `c` only | `None` — `a` and `b` indistinguishable |
| `('a','b','c')` | `b`, `c` | `'a'` — one candidate, exact |

Note the third row: three aliases is not automatically ambiguous. The rule stays exact wherever the information is actually there, rather than giving up at N ≥ 3.

For the ambiguous branch I went with the honest message instead of a guess — `"'doohickey' superseded by another spelling of 'widget'"` — and added a `superseded_by` map to `details` that carries the spelling **only where it is known**. Naming it exactly in that case needs the raw request keys (a `mode="wrap"` validator recording them), which I judged not worth putting on this path for a case no field currently has; the docstring records that as the upgrade route if a third spelling ever lands.

One thing I found while writing the test: my first draft asserted the ambiguous case reports `'widget'`, which is the *field* name — a spelling the caller never sent. That is the same class of false claim as the bug, one level down, so the two messages are now phrased differently rather than sharing a slot.

Also keyed the alias map off `type(body)` rather than `SearchRequest`, since `RecallRequest` subclasses it and may declare aliases of its own.

**Second item** (you sent it twice, so covering it explicitly): the `top_k` description claimed the response "warns that `'limit'` was not read" — the wording this PR replaced. It now states both outcomes: absorbed silently when sent alone, and a `superseded_parameter_alias` warning when sent alongside `top_k`, explicitly **not** `unrecognized_parameters`, because the endpoint does read `limit`; it was outranked. `openapi.broker.json` regenerated — one line, the mirrored description — and `--check` is clean.

Two tests added, both fail pre-fix. 76 pass across ax0917, C4, unknown-field-rejection and C25; ruff clean; mypy unchanged at the 6 pre-existing mcp-drift errors.

(The force-push was a commit-message fix only — backticks in the body had been shell-interpreted and swallowed a word. No code difference between `ce773d0b` and `c0561937`.)

ran-taig:
@claude

ran-taig:
@erni-a

erni-a (approved):

PR #1559, opened 2026-09-17
fix(core-api): make the write path discover a duplicate before it pays for one

Eldad-Caura:
Batch 4 of the audit remediation: eight findings filed against core-api's write
path, all landing on the same mechanism — **where a write learns it is a
duplicate, relative to the work it has already done.**

| Finding | What it was |
|---|---|
| 08/14 L-33 | `CheckExactDuplicate` sat behind embed + enrich + entity UPSERT in both inline pipelines |
| 08/14 M-38 | `create_memories_bulk` embedded and enriched the whole batch before looking up hashes |
| 09/02 M-47 | auto-chunk lost the caller's `write_mode`, so a `fast` write could be 409'd by a gate it opted out of |
| 08/14 M-16 | the auto-chunk multi-fact exit had no semantic dedup gate at all |
| 08/14 M-18 | auto-chunk parent payload passed a raw `UUID` to `json=` — every such write was a 500 |
| 08/14 L-38 | the atomic-fact fan-out embedded one fact per provider call |
| 09/02 L-38 | both fan-out call sites resolved parent weight with `or`, turning a deliberate `0.0` into `0.5` |
| 08/14 M-14 | evolve's `report_outcome` died on a repeated outcome, after Phase 2 had committed and before Phase 4 ran |

The commit message carries the full reasoning per finding. Three things worth
pulling out for review:

**One deliberate behaviour change.** A write refused as an exact duplicate no
longer runs `GovernanceDecision`, so the LLM free-form verdict and its audit
event no longer fire for it. The deterministic gate still runs on every write,
and the row being duplicated was governed when it was written. What is lost is a
second verdict on already-stored content; what is bought is not paying for an LLM
call on every retry.

**A gap left open, on purpose.** The auto-chunk parent's new semantic gate is
strong-mode only. Fast mode's gate (`DetectNearDuplicate`) records a merge intent
that `ScheduleBackgroundTasks` performs, and the multi-fact exit writes its parent
by hand with no such step — borrowing it would stamp a parent `near_duplicate_merged`
while no merge ever happened. Doing it properly isn't blocked on plumbing but on
semantics nobody has settled: superseding an auto-chunked parent leaves the
superseded parent's children live and unowned. So a fast write's near-duplicate
handling still depends on how many facts the chunker found — present at 0-1,
absent at 2+. That's the same inconsistency M-47 removes one level up; it's
recorded as open in the composition's docstring rather than read as design.

**A regression caught in review, worth knowing about.** Batching the fan-out's
embeds introduced one: when dedup removes every fact, an empty list reached
`get_embeddings_batch`, which has no empty guard, so `{"input": []}` went to the
provider, was rejected, and landed in `record_failure` — advancing the streak
that trips the degraded-provider breaker. The per-fact loop it replaced made zero
calls in that case. Fixed at `_embed_children_or_degrade`, where its sibling
`_insert_children_or_degrade` already had the same short-circuit, so the
auto-chunk caller gets it too.

## Verification

- `tests/test_write_path_dedup_batch.py` (new, 20 tests) covers all eight.
  **Every fix was confirmed to fail with its fix reverted**, the empty-batch
  guard included.
- Full core-api suite against a real PostgreSQL: **31 failed, 7092 passed**, with
  the failure set diffed against a clean-`main` baseline — **zero introduced
  failures**. (The 31 are pre-existing, all in search/ANN/freshness areas
  untouched here.)
- `ruff check` + `ruff format --check` clean over the full CI scope; `mypy src/`
  shows only the two known pre-existing `common/llm/providers/openai.py` httpx
  errors.

Four existing test files needed a patch target or an ordering assertion updated —
each change is explained in the commit message, and no assertion was weakened
except one that was pinning an incidental ordering.

## Follow-ups deliberately not in this PR

- `ScheduleBackgroundTasks`' fast branch lacks the `enrichment is None` guard its
  strong branch has (belongs with the enrichment fan-out batch).
- `ComputeContentHash` does an embedding lookup that the now-earlier duplicate
  gate makes wasted work on the duplicate path; the two adjacent hash lookups
  could collapse into one.
- `BusinessPersonalPregate` still makes an LLM call ahead of the hash.
- The fan-out's per-fact `create_memory` calls could run bounded-concurrent.
- Three near-identical auto-chunk test harnesses should become one shared module.


Eldad-Caura:
@erni-a please approve

erni-a (dismissed):

Eldad-Caura:
@erni-a plse reapprove

erni-a (approved):

PR #1560, opened 2026-09-17
fix(core-api): report the write path's background work honestly

Eldad-Caura:
Batch 5 of the audit remediation: six findings on the write path's **background
work** — which of it gets scheduled, whether a failure in it is reported, and
whether the embedding backend the plumbing picks is the one that was configured.

| Finding | What it was |
|---|---|
| 09/02 M-18 | fast fan-out missing the `enrichment_provider != "none"` guard every other gate carries |
| 09/02 L-20 | fast+inline fired entity extraction twice — two LLM passes per write |
| 08/14 M-06 | a step that *returned* `FAILED` was reported as a successful pipeline |
| 08/14 M-28 | `PLATFORM_EMBEDDING_*` was never bridged to `os.environ`, so it was configured and never read |
| 09/02 L-40 | the query-embedding cache key omitted the resolved provider |
| 08/14 L-34 | entity links went out one HTTP call at a time despite an existing bulk endpoint |

Full reasoning per finding is in the commit message. Four things worth pulling
out for review:

**`provider="none"` is not a provider that does nothing.** `enrich_memory`
answers it with a bare `EnrichmentResult()`, whose Pydantic defaults are real
values — `weight=0.7`, `status="active"`, `memory_type=fact`. They survive
`model_dump(exclude_none=True)` and get PATCHed onto the row. A tenant that had
deliberately switched enrichment off still had every fast write reweighted
0.5 → 0.7 and its `enrichment_pending` marker cleared.

**M-06 needed fixing in two places, and the runner alone would not have done
it.** `lifecycle_audit.entity_link` discarded the `PipelineResult` entirely, and
`Pipeline.run` does not re-raise — it catches a step's exception *into*
`result.failed`. So neither a raised nor a returned failure could reach the
handler that sets the audit status, and the docstring promising "the row will be
marked failure by the caller, not success" was false in both directions. A
nightly run whose steps all failed recorded SUCCESS with `links_created=0` —
indistinguishable from a healthy run of an org with nothing left to link.

The runner change records a returned `FAILED` but deliberately does **not**
break. A step that returns `FAILED` has caught its own error and chosen to
report it, so later steps still run — which keeps cross-link discovery and
relation inference working when only the merge stage failed. Raising still
aborts.

**L-34 preserves the guarantee the per-link loop existed for.** That loop was
deliberate (the H-05 incident: "a single bad id must not discard the valid links
beside it"). Batching does not trade it away, because
`entity_bulk_upsert_links` was built to the same requirement — it runs each item
in its **own session** precisely so an FK violation on item N cannot roll back
items 0..N-1, reporting per item via `error="fk_violation"`. This is *not* the
trade the atomic-fact fan-out's per-fact CREATE would be, where the bulk helper
is all-or-nothing. Chunked at the router's 500-item cap, since `entity_links` is
unbounded and a 501-link write would otherwise 422 and lose every link.

**One half of a finding is declined.** L-20 also reports "two contradiction
triggers per write". That is stale — F3 Phase 3 removed the contradiction fire
from `_enrich_memory_background`; there is no `run_contradiction_detection` call
left in it, and fast+inline gets exactly one trigger, from the step. The double
*extraction* was real and is fixed.

## Verification

- `tests/test_enrichment_fanout_batch.py` (new, 11 tests) covers all six.
  **Every fix was confirmed to fail with its fix reverted** — both halves of
  M-06 separately.
- Full core-api suite: **31 failed / 7103 passed, failure set identical to the
  clean-`main` baseline — zero introduced**. Run under
  `ENTITY_EXTRACTION_PROVIDER=none` to match CI, and the set is the same with
  and without it.
- `ruff check` + `ruff format --check` clean over the full CI scope; `mypy src/`
  shows only the two known pre-existing `common/llm/providers/openai.py` httpx
  errors.

Four existing tests were updated, each because it pinned a *mechanism* this
change replaced rather than a property — the `qemb4:` prefix, a fake config
missing `enrichment_provider`, a patch target on the now-unused
`create_entity_link`, and a per-link `status_code == 409` that is now an in-band
`error == "fk_violation"`. No assertion was weakened; the `permanent`
classification assertion beside that last one is unchanged.

## Follow-ups not in this PR

- `PLATFORM_EMBEDDING_BASE_URL` / `_TRUNCATE_TO_DIM` are read by the platform
  embedder but have no `Settings` field to bridge from. They stay env-only here;
  giving them settings is a config-surface change.
- `StepResult` carries no step name, so a pipeline failure can only be reported
  by its `detail`/`error`. Naming the step would make these messages much better.
- The chunk-wide failure path in `write_memory_row` is the one place batching
  costs more than a single link; worth revisiting if partial-chunk retry is ever
  wanted.


Eldad-Caura:
@erni-a plear paprove

erni-a (approved):

Eldad-Caura:
Applied — thanks, that one was mine to fix.

**`_enrich_memory_background` stale docstring (Low): applied.** Verifying it turned up that the line was already half-false *before* this PR: F3 Phase 3 had moved the contradiction trigger out of that function some time ago, so the docstring was claiming a fan-out it never performed. Removing the duplicate extraction fire (L-20) made the other half false too.

The replacement states what the function actually does — applies the patch, and when configured runs governance remediation and the atomic-fact fan-out — says plainly that `ScheduleBackgroundTasks` owns both other triggers (extraction unconditionally, Path A gated on `embedding is not None`), and records that the claim was wrong before this PR as well, so the next reader doesn't have to re-derive it.

No behaviour change in the amend; docstring only. Gates re-run clean: `ruff check` + `ruff format --check` over the CI scope, `mypy src/` at the two known pre-existing `common/llm/providers/openai.py` httpx errors. Full suite unchanged at 31 failed / 7103 passed, failure set identical to the clean-`main` baseline.

One thing already in the description but worth surfacing here, since it is a partial decline rather than a fix: **the "two contradiction triggers per write" half of 09/02 L-20 is declined as stale** — same root as the docstring above. There is no `run_contradiction_detection` call left anywhere in `_enrich_memory_background`, and a fast+inline write gets exactly one trigger, from the step. The double *extraction* was real and is fixed. The tracker row is marked `partial` for that reason.

@erni-a over to you — CI green, verdict clean, the one Low finding applied.

erni-a (approved):

PR #1572, opened 2026-09-17
fix(api): honor documented route parameters and validate request bodies

Eldad-Caura:
Seven audit findings on the HTTP boundary, plus the invariant that would have caught three of them.

## Parameters accepted but never applied

All three were reachable capability with missing wiring — the data layer already implemented every one.

| Finding | Endpoint | Was |
|---|---|---|
| `oss-0814-m-11` | `GET /audit-log` | `since` accepted and dropped. `audit_list_by_tenant` already had the `created_at > since` predicate. |
| `oss-0902-m-26` | `GET /fleet/commands` | `node_id` accepted and dropped. `fleet_list_commands` already filtered on it; only `node_name` was reachable from outside. |
| `oss-0902-m-27` | `GET /memories` | `visibility` carried no further than the core-api route. |

On an audit log, m-11 is the worst shape of wrong: the extra entries look like events inside the requested window. For m-26, a caller polling for one node's work got the whole fleet's.

m-27 is the only one needing a new filter. It is ANDed onto a query the visibility-scoping predicate has already bounded, so it can hide rows the caller may see and can never surface one it may not — asking for `visibility=scope_org` does not grant scope_org reach.

**These three are one class, so the PR adds the invariant that makes the class visible.** `test_c33_openapi_completeness` invariant 5 asserts no routed handler declares a parameter it never reads. It flags exactly these three against the parent commit and is clean here, with no ceiling to grandfather. A declared-and-ignored parameter is worse than an absent one: it is documented, accepted, and silently does nothing.

## Pagination contract — `oss-0902-l-23`

The list routes reject a cursor unless `sort=created_at&order=desc`, but minted `next_cursor` unconditionally. Following the documented pagination contract was itself the way to trigger the 400.

There were **three** copies of the gate/mint pair and all three had drifted. The predicate now lives in `pagination.py` beside `encode_cursor`/`decode_cursor`, whose module docstring already stated this invariant in prose with nothing enforcing it. A predicate private to `routes/memories.py` could not reach `mcp_server.caura_list` — which is both the third copy and the agent-facing surface, so the highest-traffic instance.

## Request validation

- **`oss-0902-l-24`** — `len()` answers for a `str` as readily as for a list, so `{"ids": "abc"}` passed the size gate and reached storage as a 3-item iterable of characters. `exclude_ids` was the sharper case: it feeds the `is_tenant_wide` test, so a non-empty string read as *"this delete is narrowed"* while narrowing nothing — disarming the unbounded-delete guard on a request that would then delete the tenant. The helper owns the shape check as well as the element check, so a third caller cannot honour one and forget the other.
- **`oss-0902-l-22`** — `related_ids` was unbounded at **both** boundaries. The 50-cap lives in `_adjust_weights`, which runs *after* `_persist_outcome`, so an oversized list was written verbatim into the outcome memory's metadata and only then silently truncated. Capping only REST would have left `caura_evolve` with exactly the behaviour the cap prevents.
- **`oss-0814-l-10`** — `POST /org/purge-data` lacked the non-object and malformed-UTF-8 body guards its `preview-data` twin gained in review on #246, leaving the endpoint that *permanently deletes tenant data* the laxer of the pair about what it would accept. Fixed by extracting the body parser both routes now call rather than by copying the guards across: the copies are what drifted, and copying them again would restore parity while preserving the mechanism that lost it.

## ⚠️ Tool-surface token budget — needs a decision

The MCP cap changes `tools/list`, so `tools_list_baseline_v1.json` is regenerated. The diff is **one line**: `"maxItems": 50`. `plugin/tools.json` and both `tool_descriptions_*` fixtures regenerate byte-identical.

The cap is deliberately **not** repeated in the tool description — `tests/fixtures/README.md` asks that anything the inputSchema already states be trimmed rather than duplicated, since tool-surface tokens are paid on every agent call. I had written `"At most 50."` into the description first and removed it on reading that guidance.

**`tools/list` goes 5312 → 5317 tokens against `CEILING_TOKENS = 5320`, leaving 3 tokens of headroom.** Flagging rather than silently consuming it: the next change to the tool surface breaches the ceiling. If you'd rather keep the headroom, the alternative is enforcing the cap inside `report_outcome` — no tool-surface cost and it covers both boundaries, at the price of agents learning the limit by hitting it rather than reading it. Happy to switch.

## Testing

`tests/test_route_contract_batch.py` adds 20 tests. **Every property was confirmed to fail with its fix reverted** — 13 in all, since l-23 has three call sites, l-24 two, l-22 two boundaries, and m-27 both a payload and a SQL half.

Full suite run against a pristine baseline at the same base commit: failures set-identical in both directions, nothing introduced and nothing silently fixed.

## Deliberately not included — `oss-0814-l-36`

Sequential heartbeat agent upserts. The obvious fix — a semaphore constructed inside the handler — gives every in-flight heartbeat its own budget and bounds nothing in aggregate, against a storage pool of 5+5 (`db_pool_timeout: 60`) on an endpoint with no rate limit, no per-tenant slot, and a caller-supplied unbounded `agents` list. `agent_get_by_id` also takes the **writer** session despite the call routing `read=True`.

`routes/lifecycle.py:118-148` documents that exact mistake and the incident it caused. It needs the process-wide, loop-keyed budget that file uses — a load-profile change to the highest-frequency endpoint in the API, which does not belong in the same revert unit as request validation. Follow-up PR.

## Follow-ups noted, not done

- The two delete routes are the only raw-dict write bodies in `routes/memories.py`; every other write route takes a Pydantic model, and `schemas.py` documents why. Converting them is the right end state but moves shape errors 400 → 422 and `extra="forbid"` would begin rejecting bodies that pass today — an API change that deserves its own PR.
- `ix_memories_visibility` exists in migrations but is absent from `Memory.__table_args__`, so suites building the schema via `create_all` run without it.
- `contradiction_detector._bounded_gather` is a generic helper sitting private in a service module, next to ~12 hand-rolled bounded fan-outs.


Eldad-Caura:
CI green (0 failing) and the review verdict came back clean with no issues found. Nothing to apply or decline this round.

Over to @erni-a for a human pass. Two things worth your attention specifically:

**1. Tool-surface token budget — needs a decision, not just a review.** The `caura_evolve` cap adds one line to `tools_list_baseline_v1.json` (`"maxItems": 50`), taking `tools/list` from 5312 to 5317 tokens against `CEILING_TOKENS = 5320`. That leaves **3 tokens of headroom**, so the next tool-surface change breaches the ceiling. I flagged it rather than spending the margin quietly. The alternative is enforcing the cap inside `report_outcome` — zero tool-surface cost, covers both boundaries, but agents learn the limit by hitting it rather than reading it.

**2. `oss-0814-l-36` was deliberately pulled from this PR.** The finding is real (sequential heartbeat agent upserts), but the obvious fix — a semaphore constructed inside the handler — gives every in-flight heartbeat its own budget and bounds nothing in aggregate. The storage pool is 5+5 with a 60s timeout, `/fleet/heartbeat` has no rate limit and no per-tenant slot, `agents` is caller-supplied and unbounded, and `agent_get_by_id` takes the **writer** session despite the call routing `read=True`. `routes/lifecycle.py:118-148` documents this exact mistake and the incident it caused. It needs that file's process-wide loop-keyed budget, which is a load-profile change to the highest-frequency endpoint in the API and shouldn't share a revert unit with request validation. Follow-up PR.

Not merging — leaving that to Eldad.

erni-a (approved):

PR #1580, opened 2026-09-17
fix(core-api): stop losing and double-counting in-flight work at shutdown

Eldad-Caura:
Four audit findings that turn out to be one shape: **cancellation and ambiguity were mis-accounted, each in a different direction.**

| Finding | Was |
|---|---|
| `oss-0814-l-11` | A cancellation-aborted audit chunk was credited as **flushed** |
| `oss-0814-m-19` / `oss-0902-m-56` | Shutdown cancelled tracked tasks with no grace **and no record** |
| `oss-0902-l-44` | Billing counters **replayed** after an ambiguous flush |

## L-11 — "unknown" was being booked as success

On an external `CancelledError` the handler set `failed = 0` and then ran `_flushed_count += len(chunk) - failed`, so an interrupted drain of the **audit** path reported itself as a clean one. The comment above it is right that these events aren't failures — avoiding a spurious failure-spike was the intent — but they aren't successes either, and `flushed` is the one bucket that overstates what happened. `interrupted_count` now exists for exactly this.

## M-19 / M-56 — the repair path's storage existed and was never wired

`CancelledError` is a `BaseException`, so it walked past `tracked_task`'s `except Exception` and the function returned having written nothing. A write already ACKed to the caller lost its embed or its enrichment with no row, no retry and nothing to find it by. The embed path is eventually repaired by the backfill; `enrichment_pending` rows have **no sweep at all**.

`background_task_log` already carries a `status` column and an `ix_bg_task_log_tenant_status` index — and its only writer hardcoded `"failed"`, so a column built to be queried by held exactly one value. Cancelled work now writes `status="cancelled"`.

**Stated honestly in the code:** nothing reads that table yet — one writer, zero queries. The row makes the work recoverable *by hand* and makes the damage *countable*, which is the number that decides whether an automatic sweep is worth building. The sweep is not in this PR, and `_TASK_STATUSES` carries a note that a repairing sweep will need a third value, since the row is terminal today.

## L-44 — a deliberate contract change, and the part to review hardest

`/tenant-usage/increment` is an additive upsert with no storage-side dedupe, which is why `increment_tenant_usage` goes out at the `idempotent=False` default. `common/http_retry` already draws the line this needs — connection-phase failures are provably never transmitted, while *"ReadTimeout and 5xx are NOT retried there — the request reached storage and may have committed"*. The meter then replayed on **any** exception one layer up, doing what the client layer refuses to do.

An ambiguous flush now **drops** its counters — logged per row (capped, with a running `dropped_rows` counter) — rather than risk billing a tenant twice. Under-counting is the chosen direction: charging for work that wasn't done is the worse error and the one nobody outside can detect. The durable fix is an idempotency key on that endpoint; `http_retry`'s own docstring names that as the prerequisite.

**This includes the shutdown path, which looks safe to exempt and isn't.** `CoreStorageClient._cancel_safe` shields every request *on purpose* — cancelling mid-request used to strand the pooled connection (incident 2026-06-16) — so a cancelled flush's POST runs to completion and may commit after the frame unwinds, including when storage looks wedged, since the shielded request outlives our giving up on it. My first cut exempted cancellation and would have left that double-bill open.

Five existing usage-meter tests move with the contract. Three simulated "storage down" with a bare `RuntimeError`; since that really *is* a connection-phase failure they now raise `httpx.ConnectError`, preserving each test's intent exactly. Two asserted the cancellation replay directly and now assert the drop, with their original regression still pinned via `dropped_rows`.

## Shutdown ordering — a defect the grace would otherwise have introduced

`cancel_all_tasks()` moves **ahead of** the queue flushes. These tasks are producers for those queues — `process_entity_extraction` calls `log_action` onto the audit queue, metered work calls `usage_meter.record` — so draining them afterwards meant a task the new grace saved handed its audit event to a flusher `stop()` had already torn down: work saved, trail dropped.

It also puts the drain where budget remains. `app.py` awaits its steps sequentially with three 5s timeouts against Cloud Run's 10s window, so as step 6 this was reached with nothing left on exactly the shutdowns that motivated giving it a grace.

## Bounded, because shutdown is the worst moment for a fan-out

Every cancelled `tracked_task` writes a row, and one bulk write in flight schedules ~4 tracked tasks per item against `BULK_MAX_ITEMS=100` — so ~400 tasks can be cancelled at once, each opening its own transaction against core-storage-api's **ten** sessions (`db_pool_size=5` + `db_max_overflow=5`) while the instances still serving traffic need them. A loop-keyed semaphore of 10 bounds it, sized to that pool. The per-task log line is DEBUG for the same reason `usage_service` throttles its own.

`cancel_all_tasks` also re-reads the registry after the grace: tracked coroutines spawn further tracked tasks, and anything created *during* the grace was in neither the snapshot nor the cancel loop — never cancelled, never awaited, with the storage client then closing under it.

## Testing

Tests live with the code they cover — L-11 in `test_audit_queue.py`, L-44 in `test_usage_meter.py`, the shutdown half in `test_shutdown_durability_batch.py`. The shutdown tests patch the module constants `cancel_all_tasks` actually reads rather than passing overrides, so they exercise production values and fail if those are removed.

**All eight properties were confirmed to fail with their fix reverted** — including two where my first check was invalid, and two where the test passed either way until fixed (one of those a pre-existing assertion that read `await_args`, which is the *failed first call* when counts are dropped).

Full suite: 31 failures, set-identical in both directions to a baseline taken on a pristine worktree at the same base commit.

## Follow-ups noted, not done

- **Per-batch idempotency** on `/tenant-usage/increment` (stable `batch_id` + `ON CONFLICT DO NOTHING` in the same transaction), following `audit_log.client_event_id` rather than `idempotency_responses` — which is a per-`(tenant, key)` inbox and the wrong shape for a cross-tenant batch. That makes both the ambiguous drop and the shutdown drop unnecessary.
- **A consumer for `background_task_log`**, modelled on the existing `lifecycle-reconcile` template, plus the terminal-state transition it would need.
- **A single shutdown deadline** with per-step remainders — the repo already uses that idiom in `common/llm/retry.py`. Today the 10s budget is an aspiration with no representation in code, and nothing pins it.


Eldad-Caura:
**Applied** — the finding was correct, and the stale prose was mine.

`UsageMeter.stop()`'s docstring still promised the pre-fix behaviour: that a cancellation landing inside the loop's in-flight flush returns its batch to the buffer, so the cancel "never drops counts and the flush below re-sends them." The `except BaseException:` branch in `flush()` deliberately does the opposite. The two paragraphs of that single docstring contradicted each other — the first promising a re-send, the second explaining why a re-send would double-bill.

Rewritten to describe the code as it stands: a cancellation-path batch is **dropped**, logged per row and counted in `dropped_rows`, because `_cancel_safe` shields the request and the POST very likely commits after the frame unwinds. Only a connection-phase failure — provably never transmitted — returns to the buffer for the final flush to re-send.

I grepped the rest of the batch for sibling claims (`re-send`, `re-buffer`, `replay`, `never drops`) across all ten changed files. This was the only stale one: `flush()`'s own docstring, the inline comments on each of the three handlers, and the tests already describe the current behaviour.

Docstring-only, no behaviour change. Gates on the amended commit: `ruff check` and `ruff format --check` clean, `mypy src/` at the two known pre-existing httpx-vs-httpx2 errors in `common/llm/providers/openai.py`, and 39 tests passing across `test_usage_meter.py`, `test_shutdown_durability_batch.py` and `test_audit_queue.py`.

Amended onto the rebase, one commit, at `8e014c4e`.

Eldad-Caura:
@claude

Eldad-Caura:
**Applied** — correct finding, and a real defect rather than a wording one.

On the timeout path `flush()` has already moved the buffer into its own batch and dropped it, so by the time `except TimeoutError:` runs, `_counts` holds only whatever raced in *during* the flush. `len(self._counts)` therefore reported ~0 rows lost on precisely the path that loses the most. The captured log from the test says it plainly — three consecutive lines from the same run, before the fix:

```
usage meter flush cancelled; 1 counter rows dropped — the shielded request may still commit…
usage meter counter row dropped (flush cancelled at shutdown): tenant=t-1 … count=6
usage meter final flush did not complete within 0.05s; 0 counter rows are lost to shutdown
```

It announced zero lost immediately after naming the row it had just lost.

Fixed by snapshotting `dropped_rows` before the final flush and reporting the delta alongside what is still buffered — both are gone once `stop()` returns, and they are lost for different reasons, so the line now names them separately. I kept the warning rather than deleting it: the blown deadline is its own signal, distinct from the per-row drops `_log_dropped` already emits.

The success-path loop above it is untouched. It is still correct for its own case — a rebuffered connection-phase batch, or counts that raced in during the flush.

**Pinned by a test.** `test_a_wedged_storage_does_not_hold_shutdown_open` already drove this exact path but never asserted the log line, which is how this got through. It now asserts the deadline warning reports the dropped count. Confirmed it fails without the fix: `AssertionError` on `0 counter rows are lost to shutdown`.

**Swept for siblings rather than waiting for a third round.** Both findings so far are leftover consumers of `flush()`'s old post-condition, so I checked every one: `_counts` has exactly two readers after a flush — the success loop (correct) and this timeout log (fixed). Nothing outside the module touches `_counts` or `flush()`; `app.py` only calls `stop(timeout=5.0)`. No prose elsewhere claims the meter replays.

Gates: ruff clean, `mypy src/` at the two known pre-existing httpx-vs-httpx2 errors, 39 tests passing across the three affected files. One commit, `5cba1089`.

One thing I am *not* fixing here, for the record: `dropped_rows` and `interrupted_count` have no production reader — only tests. Same gap as `background_task_log`, and noted in the PR body as follow-up.

Eldad-Caura:
@claude please re-review at `5cba1089`.

Both prior rounds found the same class of defect — a consumer of `flush()`'s **old** post-condition that I failed to update when the batch stopped being returned to the buffer. Round 1 was the `stop()` docstring; round 2 was `stop()`'s `except TimeoutError:` log reading `len(self._counts)`.

Changed since your last pass, and the place to look hardest:

- `usage_meter.stop()` now snapshots `dropped_rows` before the final flush and reports the delta plus what is still buffered.
- `test_a_wedged_storage_does_not_hold_shutdown_open` asserts that warning's contents; revert-confirmed to fail without the fix.

I swept for further instances of that class myself and believe it is exhausted — `_counts` has exactly two readers after a flush (the success loop, which is correct for a rebuffered connection-phase batch, and the timeout log, now fixed), nothing outside the module touches `_counts` or `flush()`, and `app.py` only calls `stop(timeout=5.0)`. Worth an independent check that I did not miss one, since I have now missed two.

Known and deliberate, not defects: an ambiguous *or cancelled* flush drops its counters rather than replaying an additive upsert with no storage-side dedupe, and `dropped_rows` / `interrupted_count` / `background_task_log` have no production reader yet — all three are follow-ups in the PR body.

Eldad-Caura:
@erni-a over to you — CI green on `5cba1089`, automated review clean after three rounds.

**What it changes.** Core-api's shutdown used to cancel background tasks instantly and re-buffer every failed usage flush. Both were wrong in opposite directions: in-flight work was lost without a trace, and ambiguous billing writes were replayed. This adds a bounded drain grace before cancellation, records a cancelled task's disposition as `cancelled` rather than `failed`, distinguishes "unknown disposition" from "failed" on the audit queue, and replays a usage batch only when it provably never reached storage.

**The two things worth your judgement, both deliberate:**

1. **A billing-direction contract change.** An ambiguous *or cancelled* flush now **drops** its counters instead of replaying them. `/tenant-usage/increment` is an additive upsert with no storage-side dedupe, and `CoreStorageClient._cancel_safe` shields every request on purpose (incident 2026-06-16, stranded pooled connections) — so a cancelled POST runs to completion and likely commits after the frame unwinds. Replaying it double-bills. The trade is that a shutdown racing a storage blip can still *under*-count; it can no longer over-charge. Five existing tests moved with this. The durable fix is an idempotency key on that endpoint, which is a follow-up in the PR body.

2. **`background_task_log` still has no reader.** This PR starts writing a `cancelled` status into a table nothing consumes. The column and index already existed with one value in them; this makes the data honest, but a consumer modelled on `lifecycle-reconcile` is still owed. Same gap applies to the new `dropped_rows` and `interrupted_count` counters — tests read them, production does not.

**Review history, since it bears on where to look.** Two rounds each found a real leftover, both in `usage_meter`, both consumers of `flush()`'s old post-condition that I missed when the batch stopped returning to the buffer: first the `stop()` docstring, then its `except TimeoutError:` log reporting `len(self._counts)` — which read ~0 on precisely the path that loses the most. Round 3 came back clean, and independently confirmed `status` is a plain text column with no CHECK constraint, so no migration is required. `usage_meter.stop()` is the part of this PR that has needed the most correction; the rest has been reviewed three times without a finding.

Eight properties are revert-confirmed. Full suite matches baseline in both directions.

Not merging — leaving that to Eldad.

erni-a (approved):

PR #1581, opened 2026-09-17
fix(ci): say when a review runs without recalled guidance

Eldad-Caura:
Mirrors [caura-ai/.github#60](https://github.com/caura-ai/.github/pull/60) into this repo's own copy, which that repo's CLAUDE.md requires for any change to review behaviour — these two scripts are a deliberate fork, not a symlink, and nothing catches drift between them.

Of the two scripts here, `claude_pr_capture.sh` already printed that it was dark. This one — which runs on **every** review rather than after a merge — returned in silence on an empty key:

```sh
[ -n "$agents_key" ] || return 0
```

So a key that stopped resolving would degrade every review here with nothing in any run to read. Not hypothetical: the shared pipeline had the same gap, and it's how recall stayed dead across the org until somebody read the code rather than saw a signal.

A **notice**, not a warning, matching the capture script's existing choice for the identical condition. An absent key is a supported state; a yellow annotation on every review in a repo that hasn't provisioned one is noise that teaches people to ignore annotations.

Verified both directions by sourcing the function: with no key it prints the notice, returns 0, and leaves `GUIDANCE_SECTION` empty — its output contract; with a key set it prints no such notice.

## What this deliberately does not do

**The dual-reads stay.** Everywhere else in the org they were vestigial and came out this week. Here the old spelling is the *live* path:

- this repository is **public**
- both org secrets (`CAURA_AGENTS_KEY`, `MEMCLAW_AGENTS_KEY`) are scoped `visibility: private`, so neither is visible here
- the only key this repo can see is its own **repo-level** secret, under the pre-rename name

Measured rather than assumed — a review run today logged `Injected 10 learned-guidance item(s) from Caura`, which only happens when the fallback resolves.

**Renaming here needs a repo-level `CAURA_AGENTS_KEY` provisioned first**, carrying the same value. This change can't do that, and doing it in the wrong order takes recall dark silently — which, after this commit, is at least something the run would say out loud.

Reviewed my own diff rather than running `/simplify`.


Eldad-Caura:
@erni-a paslr arprove

erni-a (approved):

PR #1582, opened 2026-09-17
ci: read only CAURA_AGENTS_KEY, now that the repo secret exists

Eldad-Caura:
This repository was the one place in the org where the pre-rename spelling was **load-bearing** rather than vestigial. It's public, both org secrets are scoped `visibility: private`, and the only key it could see was its own repo-level secret under the old name — so the `||` wasn't a courtesy, it was the whole supply.

A repo-level `CAURA_AGENTS_KEY` was provisioned on 2026-09-17, so the fallback has nothing left to fall back from: three sites in the workflow and the read in each of the two scripts.

## This PR is also the test

GitHub never returns a secret's value, so the only way to know the new one carries the right key is to make something authenticate with it. A review runs from the merge commit, so **the review on this PR reads `secrets.CAURA_AGENTS_KEY` alone**, with nothing behind it, and annotates the run:

- `Injected N learned-guidance item(s) from Caura` → the secret is right
- `caura recall failed — reviewing without learned guidance` → it isn't

Nothing merges on a failure. The diff is exactly the thing under test.

**Worth stating because it's the risk this closes rather than opens:** `main` is already in this state. `${{ A || B }}` prefers A the moment A is set, so provisioning the new secret already replaced the old one everywhere. If the key is wrong, recall is degraded on `main` right now — this diff neither causes that nor hides it.

## Two stale claims, corrected rather than left to mislead

- The recall docstring said *"DARK AND SILENT without CAURA_AGENTS_KEY"*. It stopped being silent in #1581 — that was the point of that change — and the same paragraph said a repo-level secret didn't exist yet. It does.
- The capture script described reading *"either spelling"*, and its optional-env block said both the key and the URL accept a pre-rename name. Only the URL does now; that's a separate contract and is untouched here.

## Verification

- Sourced the recall function both ways: no key prints the notice, returns 0, leaves `GUIDANCE_SECTION` empty (its output contract); key set prints no such notice
- `shellcheck -x` clean over both scripts
- Old-brand references under `.github/scripts/` and the review workflow: 5 → 0 for the key (the `MEMCLAW_API_URL` dual-read is a separate contract and stays)

Reviewed my own diff rather than running `/simplify`.


Eldad-Caura:
@erni-a plear aspprove

erni-a (approved):

PR #1583, opened 2026-09-17
fix(core-api): bound the fleet heartbeat's per-agent storage fan-out

Eldad-Caura:
Follow-up to #1572, which filed `oss-0814-l-36` and deliberately shipped without it: the fix there was a semaphore built inside the handler, which gives every in-flight heartbeat its own budget and bounds nothing in aggregate. That is a load-profile change to the highest-frequency endpoint and did not belong in a request-validation PR.

## What was wrong

The heartbeat refreshes one agent row per reported agent, awaited one at a time. A node reporting 40 agents paid 40 sequential storage round-trips, every tick.

The upserts now run concurrently against a budget shared by every heartbeat the **worker process** is serving. `routes/lifecycle.py` documents the per-request shape as the cause of its 2026-09-15 incident — it reasons about a single caller while the pool sees the aggregate. core-storage-api runs ten connections (`db_pool_size=5` + `db_max_overflow=5`) with no reader/writer split, so these refreshes compete with live writes for the same slots.

## Three things the finding did not name

**`HeartbeatIn.agents` was unbounded.** `list | None` — untyped, caller-supplied, on a route with no rate limit and no per-tenant slot. A shared budget with no bound on what one caller can enqueue is a queue, not a cap: one node could hold the process's slots and stall its neighbours, which is *worse* than the sequential version, where an oversized payload only ever slowed itself.

The bound lives on the model, beside the two existing free-form caps and for the same 2026-06-28 reason — `body.agents` is also written wholesale to `nodes.agents_json` and read back to the fleet view, so capping only the work list would have left the stored row and the API response unbounded. It truncates rather than dropping to a marker, which would lose every agent instead of the overflow.

**A repeated agent id would have become two racing upserts.** Self-correcting sequentially; under the fan-out it is two in-flight upserts of one row. De-duplicated — preserving a `display_name` that a later bare repeat omits, because `get_or_create_agent` refreshes the name only when the argument is not `None`, so plain last-wins would have erased a name the old loop kept.

**A comment asserted a safety property that was no longer true.** It claimed the read-first shape was still needed to dodge a 500 from `agent_add`'s `IntegrityError` → `rollback()` → re-SELECT path. `agent_add` was rewritten to `INSERT ... ON CONFLICT DO NOTHING RETURNING` with a same-session re-SELECT precisely to remove that pattern. The read-first GET is itself racy — two callers can both miss and both POST — so what makes the fan-out safe is the storage side, not the read.

## Honest limits, documented in the code

- **Per worker, not deployment-wide.** Both services run `--workers 2` and core-api scales horizontally, so the real ceiling is `instances x workers x` the knob. Strictly better than a per-request budget, which bounded nothing even within a worker — but the headline is "predictable per worker".
- **Tenant-blind.** One tenant reporting many agents can delay another's heartbeat on that worker. `per_tenant_storage_slot` is *not* a substitute: it queues unboundedly per `(scope, tenant)`, so N tenants give N x cap and the aggregate is unbounded again. A per-tenant share of this budget is the real fix.
- **The round-trips are made concurrent, not removed.** Steady state is still one `GET /agents/{id}` per agent per tick to learn nothing changed.

## Verification

Seven properties revert-confirmed, each against the shape it exists to catch:

| property | fails against |
|---|---|
| upserts overlap | the sequential loop (peak 1) |
| budget is process-wide | a per-request budget (peak 12 vs 2) |
| repeated id upserted once | no de-dup |
| a bare repeat keeps the name | naive last-wins |
| list capped at the model | no cap (10 vs 4) |
| node row bounded too | a handler-only cap |
| junk env value falls back | a bare `int(...)` |

Per-agent failure isolation and the normal-payload passthrough are preservation tests and pass both ways, deliberately.

Gates: `ruff check`, `ruff format --check`, `mypy src/` at the two known pre-existing httpx-vs-httpx2 errors in `common/llm/providers/openai.py`. Full suite compared against a pristine `origin/main` worktree.

## Follow-ups

1. **Extract the loop-keyed semaphore.** This is its third verbatim copy (`lifecycle.py`, `services/task_tracker.py`, now here), plus a fourth variant in `common/embedding/_service.py`. It belongs in `common/concurrency.py`, with the limit read through a callable so the existing `patch.object(mod, "_X_CONCURRENCY", n)` tests keep working. Not here: it touches two other modules and three test files, which should not share a revert unit with a load-profile change.
2. **A per-tenant share of the budget**, per the tenant-blindness above.
3. **Skip the unchanged reads.** A TTL keyed on the reported tuple — the pattern already in this file for the outdated-plugin warning — would remove nearly all of the steady-state GETs. Note `agent_service` carries an explicit "do NOT cache" for `get_agent`, but that is scoped to the trust/fleet-gate read path where staleness affects authorization; this would need to be a separate structure, not a cache inside `get_agent`.
4. **`GET /agents?tenant_id=&fleet_id=` in one call** is tempting and I did not take it: it is per-*fleet* while the heartbeat is per-*node*, so a 500-agent fleet would return 500 rows to a node reporting 5, every tick. It only wins above a threshold, which would need measuring.

Worth a reviewer's eye: the **truncation is silent to the client** — an oversized payload gets a 200 and a server-side warning, with the overflow agents simply not refreshed that tick. That follows the endpoint's stated contract (the node and command channel is what it promises; the row refresh is best-effort) and the precedent its sibling fields set, but it is a deliberate choice to degrade rather than tell the caller.


Eldad-Caura:
@erni-a plerasr ppaorve

erni-a (approved):

PR #1584, opened 2026-09-17
fix(core-worker): stop the deferred enricher overwriting lifecycle state and caller tags

Eldad-Caura:
Batch 8 of the OSS audit remediation — core-worker. Five findings fixed, one deferred with reasoning.

## The two MEDIUMs share a root cause, and it is worse than filed

CAURA-719 stopped asking the LLM for `status` and `tags`. The deferred enrichment worker's routing tables were never updated to match, so it kept PATCHing both to storage.

**M-65 — `status`.** Filed as "unconditionally re-writes `status='active'`". That understates it. `status` sat in `_ENRICHMENT_ORM_FIELDS`, which — unlike the always-write metadata branch — has **no `llm_ms > 0` guard**. So the heuristic fallback wrote it too, and `fake_enrich` doesn't emit `"active"`: it picks from a keyword guess, `"pending"` for task/plan/commitment and `"confirmed"` for outcome. The case where the **LLM was unavailable** wrote a guessed lifecycle state over a real one.

`status` is owned by explicit setters — `caura_manage`'s transition op, the contradiction detector, the crystallizer, the delete path. The `agent_provided_fields` gate cannot protect it: it is `model_fields_set` on the write request body, a snapshot of what the caller set *at write time*, silent about a transition that happened afterwards. Enrichment is deferred, so "afterwards" is the whole delivery window plus every redelivery. Eleven read paths filter the literal `'active'` (counted, not estimated), so a reverted archive is invisible rather than loud. core-api's inline path guards this by comparing the row's **current** status; the worker never reads the row, so it cannot.

**M-66 — `tags`.** `common/enrichment/_prompts.py` has stated the contract since CAURA-719: *a caller who supplies `metadata["tags"]` keeps them untouched*. The inline path honours it with `if enrichment.tags:`. This path did not — `tags` sat in `_ENRICHMENT_ALWAYS_WRITE_METADATA`, whose guard is `value is not None`, and `[]` is not None. Every deferred enrichment that reached the LLM wrote `tags: []` over the caller's list, in **both** the legacy key and `_system` (the one that wins on read).

The comment sitting on that entry claimed a caller list was "protected by `CALLER_OWNABLE_KEYS` at the `set_system_value` boundary, not here". That boundary lives in `core_api.services.system_metadata`, and this worker PATCHes core-storage-api directly without importing any of it. The protection it named does not cover this path.

Both keep their schema entry, default and validator so existing payloads still deserialise; they simply stop being written.

## The three LOWs

- **L-21** — `reference_datetime` is now typed `datetime` instead of `str`. It was a string the publisher produced with `.isoformat()` and the worker turned back with `fromisoformat`, so the schema accepted anything and the real rejector sat *below* the poison-payload guard: a malformed value raised past the handler, nacked, and redelivered to the DLQ — for a payload that can never parse. Typing the field correctly makes pydantic reject it inside the guard that already ack-drops, and **deletes** the parse rather than moving it. Not a wire change: `model_dump(mode="json")` still emits ISO 8601 and an ISO string still parses inbound, so in-flight messages survive a rolling deploy in both directions.
- **L-54 (config)** — three dead event-bus settings fields removed (zero readers repo-wide, verified). `env_file=".env"` loads onto the settings *object* without exporting to `os.environ`, which the factory reads — so declaring them let `EVENT_BUS_BACKEND=pubsub` in .env produce a worker that looked configured and silently ran the in-process bus. `lifespan` now logs the backend actually resolved.
- **L-54 (cli)** — the embedding backfill refuses rather than publishing into a subscriber-less bus and reporting `published=N`. `--dry-run` is exempt, and `get_event_bus()` sits inside the short-circuit so a dry run never resolves a bus that can raise.

## Deferred

**M-67** (backfill sweep redelivery). Mechanism confirmed — `ack_deadline_seconds: 30` with no `modify_ack_deadline` anywhere, so any sweep over 30s is redelivered into a concurrent duplicate. But the fix is lease extension in the shared Pub/Sub pull loop, already the subject of ent-056 / oss-0814-m-02 / oss-0902-l-12. A per-process guard here would bound nothing across instances.

## Note: this PR touches `common/`

`common/events/memory_enrich_request.py` and `memory_enrich_publisher.py` are the L-21 fix — the field's type *is* the defect. Neither file is on the byte-parity vendored list (`common/__init__.py`, `structlog_config.py`, `serve.py`). Flagging it because it crosses the batch boundary; trivially separable if you'd rather it went with the `common` batch.

## Follow-ups found while verifying, not fixed here

1. **`summary` has the identical caller-clobber bug as `tags`.** `CALLER_OWNABLE_KEYS` is `{"summary", "tags"}`, and the worker has no caller-ownership boundary at all — `MemoryEnrichRequest` carries `agent_provided_fields` for ORM columns and nothing for caller-set *metadata* keys, so the publisher never tells the worker the caller owns `metadata["summary"]`. Unrouting `tags` closes one of two holes in a missing mechanism. The real fix is to carry the caller's metadata key set on the event and gate `_build_patch` on it — then `tags` needs no drop-list entry either.
2. **`InProcessEventBus.publish` should warn when a topic has zero handlers.** The CLI check here is the second inline `isinstance(bus, InProcessEventBus)` in the repo (core-api's `app.py` has the other), and the hazard is live elsewhere: `core-api/src/core_api/routes/lifecycle.py` publishes `embed-backfill` whose only subscriber is core-worker, and OSS `docker-compose.yml` ships no core-worker — so that route returns 200 with the audit row stuck `pending` forever. A warning in the bus is unconditionally correct and covers every publish-only entrypoint.
3. **`get_event_bus()` should log the resolved backend once, in the factory.** core-api resolves the same bus and logs nothing; doing it in `common/events/factory.py` would make core-worker's line redundant and cover every service.

## Verification

- ruff `check` + `format --check` at every path CI gates, including `--isolated` on the vendored trio
- `mypy src/` for core-worker and core-api — 2 pre-existing httpx/httpx2 errors only
- **Four pytest roots run, not one.** `core-worker/tests/` is a separate root with its own `pytest.ini`; it held 4 tests asserting the behaviour M-65/M-66 removes. They are **rewritten, not deleted** — and two of them used `tags` as the "other fields still flow" control, which this change would have made silently vacuous.
- Root `tests/`: 31 failures on the branch, 31 at `origin/main`, **identical sets compared in both directions**; all pre-existing, all in search/recall files untouched here. core-worker 94 passed (80 at baseline). core-operations 81 passed.
- Every fix revert-confirmed: each change reverted in turn and the corresponding tests checked to fail, with an anchor-occurrence assert so a missed revert cannot read as a pass.


Eldad-Caura:
@erni-a pler araprpove

erni-a (approved):

PR #1585, opened 2026-09-17
fix(enrichment): honour the caller's metadata on every write path, not one

Eldad-Caura:
> **⚠️ This PR contains TWO commits — `b77588a2` is the one to review here.**
>
> It sits on top of #1584 and is **based against `main` deliberately**: every meaningful workflow (`ci.yml`, `codeql.yml`, `security-scan.yml`, `claude_code_review.yml`) filters on `branches: [main, dev]`, so while this targeted #1584's branch only `dco` and the legacy-name ratchet ran — and that would NOT have self-corrected, because retargeting on merge fires `pull_request: edited`, which is outside the default trigger set.
>
> Consequence: `9209e71d` (#1584) shows in this diff and **merging this PR would land both**. Merge #1584 first and this reduces to its own commit. Review #1584 separately — it has its own green CI and a clean review.

This is follow-up #1 from #1584's description. Investigating it turned up more than was filed.

## C25 was applied to one of four enrichment write paths

The boundary: `set_system_value` always writes `metadata["_system"][key]`, and mirrors to the legacy top-level `metadata[key]` **only when the caller does not own that key**. `CALLER_OWNABLE_KEYS = {"summary", "tags"}`. Enrichment writes those keys from four places:

| path | before |
|---|---|
| `MergeEnrichmentFields` — single synchronous write | correct since C25 |
| `create_memories_bulk` | wrote `metadata["summary"]` directly |
| `_enrich_memory_background` — every inline deployment | wrote `meta["summary"]` directly |
| core-worker `_build_patch` — deferred | wrote both homes, **and had nothing to decide with** |

So the same payload kept its summary as a single write and lost it as a batch of one.

The shape of the bug is why it survived: a caller's `summary` and a platform-written one are **the same key**, so no path can tell them apart by inspection. The key set has to be captured at write time and carried. `MemoryEnrichRequest` now carries `caller_owned_metadata_keys`, narrowed to `CALLER_OWNABLE_KEYS` before it goes on the wire.

## The worker does not trust that list

It intersects against its own copy of the ownable set on arrival. My first version didn't, and that was a real hole — worth stating plainly because the comment I originally wrote claimed the field was simply "the metadata-side twin" of `agent_provided_fields`, and that symmetry does not hold.

`agent_provided_fields` pins ORM columns the caller could have set directly at write time, so honouring a forged entry grants no capability the caller lacked. `caller_owned_metadata_keys` can name a **platform** key — and some are read from the legacy position only: `governance_remediation` does `md.get("contains_pii")` and `md.get("business_relevance")`, never `_system`. A payload naming those would write the PII verdict to `_system` alone, silently disabling DROP / KEEP_PRIVATE while an auditor reading `system_metadata` still sees PII flagged. Exactly the forgery `PLATFORM_ONLY_KEYS` exists to prevent, through a new channel.

A root-level test asserts the worker's copy equals core-api's — that root is the only one that can import both.

## Response-shape note

The two core-api paths now also write `_system` entries they never populated. That is **required, not incidental**: without it `extract_system_metadata` falls back to the legacy key, which for a caller-owned summary is now the *caller's*, so the platform view would start returning caller text. Rows from those paths therefore echo a `_system` sub-object they previously lacked — converging with single-write rows, which have had it since C25. No reader changes (for a non-ownable key nested == legacy), and `sanitize_caller_metadata` strips `_system` from caller input, so a read-modify-write round trip can't re-inject it.

## Deliberately not fixed here

- `contains_pii` / `pii_types` / `business_relevance` / `weight_source` still write bare top-level keys in those two paths, so `_system` there is present but incomplete. No reader breaks (legacy fallback covers it) and **no clobber is possible** — none are caller-ownable — so routing them is cleanup, not a fix.
- `system_metadata.py` could move to `common/` so all three services share one implementation and one `SYSTEM_NAMESPACE` literal instead of a mirrored frozenset. That would make the trust hole above *impossible to reintroduce* rather than merely tested against. It's a 10-import-site move with a re-export shim; worth doing, not worth burying in this PR.
- `_caller_owned_enrichment_metadata_keys` reads `data.metadata`, which later pipeline steps mutate in place. Correct today only because four independent gates imply `enrichment is None` at every call site. The invariant is now recorded in the docstring; the durable fix is to snapshot the key set at the `sanitize_caller_metadata` chokepoint.

## Verification

- ruff `check` + `format --check` at every CI-gated path incl. `--isolated` on the vendored trio; `mypy src/` for core-api and core-worker at the 2 known pre-existing httpx errors
- **Every write path and both ownable keys revert-confirmed independently** — six call sites, each reverted in turn. The `tags` half initially had zero coverage on any path (a fix that threaded `caller_keys` into the summary call and missed the tags call two lines below would have shipped green); so did both pipeline call sites, where the feature originates.
- **Two would-be vacuous passes caught and guarded.** The bulk merge and the deferred publish both silently no-op when a fresh test tenant resolves `enrichment_enabled` to False — "the caller's summary survived" was true because nothing touched it. Both tests now assert the path actually executed before asserting on its output.
- Three pytest roots: root 7158 passed / 31 failed — **identical failure set to `origin/main` in both directions**, all pre-existing search/recall; core-worker 100 passed; core-operations 81 passed.


Eldad-Caura:
CI caught two real test defects — both of my own integration tests, and both for the same reason.

`test_bulk_write_leaves_a_caller_owned_summary_alone` and `test_bulk_deferred_branch_publishes_the_callers_keys` force the tenant config so enrichment actually runs, but the proxy overrode only `enrichment_enabled`. `create_memories_bulk` gates on **two** things at both the inline merge and the deferred publish:

```
not settings.inline_enrichment and tenant_config.enrichment_enabled and tenant_config.enrichment_provider != "none"
```

CI sets `ENTITY_EXTRACTION_PROVIDER: none`, so `enrichment_provider` resolved to `"none"` and the merge never executed — the codebase's own note in `organization_settings.py` names this case ("survives `enrichment_provider=none`, e.g. CI"). Locally it resolves to `openai`, which is why it passed here and not there.

Both failures were the **execution guards** firing (`stub.await_count == 1`, `publish_spy.call_count == 1`), not the assertions under test. That is the guards doing their job: without them these two would have passed in CI while exercising nothing at all.

Fixed by extracting one `_enrichment_actually_runs()` helper that overrides both properties on a proxy over the real resolved config, used by both tests.

Verified rather than assumed: with `ENTITY_EXTRACTION_PROVIDER=none USE_LLM_FOR_MEMORY_CREATION=false` set locally, the file goes 15/15; removing just the `enrichment_provider` override under that same env reproduces exactly the two CI failures.

No source changes — this is test-harness only.

Eldad-Caura:
@erni-a pleaseasr prpove

erni-a (approved):

PR #1586, opened 2026-09-18
fix(core-storage-api): make audit and entity read paths answer the question asked

Eldad-Caura:
Four findings from the OSS 08/14 + 09/02 audit, one shape: **the response looks well-formed and is wrong in a way the caller cannot see from it.** Each premise was re-verified against the source before any code was written, and two turned out worse than filed.

| Finding | Severity | What it actually was |
|---|---|---|
| `oss-0902-m-59` | MEDIUM | `GET /audit-logs` filtered **after** SQL's `LIMIT` |
| `oss-0902-m-62` | MEDIUM | `entity_list` paged with `OFFSET` and no `ORDER BY` |
| `oss-0814-l-31` | LOW | entity `memory_count` counted soft-deleted memories |
| `oss-0902-l-53` | LOW | `truncated` told you to paginate, with no cursor |

## M-59 — worse than filed

Filed as "silently drops matching rows". It's more than that: `action` / `resource_type` / `offset` were applied in Python to rows the query had **already truncated**. So a filter searched only the newest `limit` rows and returned `[]` when the matches were older — and because `offset` sliced a list itself capped at `limit`, `offset=limit` was empty **for every tenant, always**. The endpoint could not paginate past page 1 at all.

All three now filter in SQL, ordered `(created_at DESC, id)` — `created_at` is not unique, and a stable tiebreak is what makes `OFFSET` paging coherent.

## M-62

Postgres guarantees no row order without `ORDER BY`, so a caller walking pages could be served an entity twice and never see another, with nothing in the response to show it. Ordered by `id`: the only unique column here, and uniqueness is the property a stable sort needs. It's also already indexed, so no sort node is added — and it matches the convention a sibling query in this same file already documents.

## L-31

The count is rendered beside `/with-memories`, which excludes soft-deleted rows. The two endpoints disagreed about the same entity, and the gap was exactly the memories the caller had deleted.

Fixed at the count, deliberately **not** in the shared `_link_within_tenant` helper — that predicate answers a different question (tenancy), liveness is a separate axis, and folding them would have silently changed its two other callers.

## L-53 — made resumable rather than just re-worded

The prose promised pagination the API couldn't do; a chain past the route's 500k cap could never have its tail verified. Adds `start_seq` in / `next_seq` out, plus a `missing_predecessor` refusal — resuming from a seq whose predecessor is gone must not verify against genesis, since that would accept a chain whose head had been cut off exactly at `start_seq`.

### Two security corrections found in review

**The terminal window's tail check is anchored, not skipped.** Skipping it is the obvious reading of "past the end" and it is a hole: `truncated` *also* skips that check, so on a chain whose length is an exact multiple of `limit`, the last non-empty window skips it (truncated) and the next window skips it again (empty) — **no window runs it**, and a walk over a chain with rows deleted off the tail reports `valid: true`. `DELETE ... WHERE seq > 100000` reaches that, 100000 being the documented default `limit`, and a contiguous tail deletion raises no `seq_gap`. On an empty window the seed *is* the tail row's hash, so the check anchors there instead.

**The docstring no longer claims a windowed walk equals a full pass.** It's weaker. One pass reads every row in a single `REPEATABLE READ` snapshot; a walk is N transactions across N snapshots, and under `read_database_url` possibly N replicas at differing lag — so a tamper *behind* the cursor goes undetected until a later run covers that region. `start_seq` is also a plain integer, not a cursor bound to the previous window, so contiguity is caller-enforced and unauthenticated. All of that is now stated rather than implied.

`core-api`'s storage client gains `start_seq` and documents that `valid: true` on a truncated window does **not** mean the chain is intact — it was the one in-repo consumer for which the original motivation was still unfixed.

## Verification

| Gate | Result |
|---|---|
| `core-storage-api/tests/` | **433 passed** (baseline 419, +14) |
| root `tests/` | **31 failed / 7142 passed — failure set identical to baseline** |
| root `tests/test_audit_chain.py` | 22 passed (signature change) |
| `ruff check` + `format --check` | clean |
| `mypy` | exactly baseline (2 pre-existing / 168 pre-existing), no new errors |
| mutation checks | 7/7 discriminate |

Every new test was confirmed to **fail without its fix**. Seven mutations were run against the suite — hash-recompute limited to the first window, dropped `id` tiebreak, non-unique sort column, since-genesis `verified_count`, dropped SQL `OFFSET`, and both directions of the tail-check fix — and each kills at least one test.


Eldad-Caura:
@erni-a pelra arppvoe

erni-a (approved):

PR #1589, opened 2026-09-18
fix(core-storage-api): recover correctly from the conflicts these write paths already detect

Eldad-Caura:
Three findings, one shape: **the INSERT detects the conflict correctly and the recovery from it is wrong** — so a race the database handled cleanly becomes a 500, a dropped item, or silent data loss the caller is told did not happen.

Every one was **reproduced against Postgres before being fixed**, and two turned out worse than filed.

| Finding | Severity | What it actually was |
|---|---|---|
| `oss-0902-m-63` | MEDIUM | `entity_add`'s race recovery was **unreachable code** |
| `oss-0814-l-16` | LOW | concurrent first-time org writes lost data **and reported success** |
| `oss-0902-l-51` | LOW | re-query disagreed with its own arbiter about `fleet_id` |

## M-63 — the recovery path could never run

```
InvalidRequestError: Can't operate on closed transaction inside context manager.
```

`get_session()` yields inside `session.begin()`, so the mid-block `rollback()` closed the transaction the context manager still owned. The re-SELECT died before it could execute — **every entity dedup race 500'd, and the `winner is None` guard beneath it had never executed once.**

The fix isn't new: `agent_add`'s docstring already says it uses `ON CONFLICT DO NOTHING RETURNING` specifically to avoid "the `flush() → IntegrityError → rollback() → re-SELECT` pattern's mid-session rollback, which is brittle." That shape was simply never applied here.

## L-16 — the response lied

Reproduced with two concurrent first-time writers under one namespace:

```
writer-a -> {'settings': {'enrichment': {'provider': 'openai'}}, 'changed': True}
writer-b -> {'settings': {'enrichment': {'enabled': True}},     'changed': True}
FINAL    -> {'enrichment': {'enabled': True}}      # provider silently gone
```

Both were told `changed: True` with their own settings echoed back, and both wrote audit rows — a success response quoting a value that was never stored.

The cause was **two disagreeing merge semantics in one method**: a deep `deep_merge()` in Python and a shallow JSONB `||` in SQL, so the stored result depended on which path the write took. The `||` was justified by a comment claiming top-level keys are independent — true for writers touching *different* namespaces, exactly wrong for two first-time writers under the *same* one.

Rather than making the SQL merge deeper, the two semantics are collapsed into one: claim the row with `ON CONFLICT DO NOTHING`, and any writer that loses the claim redoes the read-merge-write under a lock that now has a row to hold, through the same `deep_merge` as every other path.

## L-51 — contradicting its own documented invariant

`memory_add_all`'s comment says the re-query matches the index scope `(tenant_id, COALESCE(fleet_id, ''), client_request_id)` **exactly**. Sixty lines below, it branched on `fleet_id is not None`. A retry spelling fleetless as `""` against a row stored `NULL` conflicts in the index and misses in the lookup, so the item fell through to `id: None` — a per-item failure reported for a write that had already committed.

## Also: a model/migration schema drift this exposed

`uq_entities_tenant_type_name_fleet` existed **only in migration 001**, never on the `Entity` model. Any schema built from the metadata rather than the migration chain — `tests/conftest.py` uses `Base.metadata.create_all` — therefore had **no unique index on `entities` at all**:

- the new `ON CONFLICT` target has nothing to infer against there (`there is no unique or exclusion constraint matching the ON CONFLICT specification`)
- and the **old** code had nothing to violate, so `entity_add` silently inserted duplicates in those environments instead of deduping

Now declared on the model; a fresh `create_all` emits a definition identical to the migration's. The `Memory` model already declares its equivalent for exactly this reason.

The three fleet predicates that must agree with a COALESCE-grouped index now share one helper instead of being written out per call site.

## Verification

| Gate | Result |
|---|---|
| `core-storage-api/tests/` | **440 passed** (baseline 433, +7) |
| root `tests/` | **31 failed / 7142 passed — failure set identical to baseline** |
| `ruff check` + `format --check` | clean |
| `mypy` | exactly baseline (2 pre-existing), no new errors |
| mutation checks | 5/5 discriminate |

Each fix was reproduced before being written, and every test confirmed to **fail without its fix**. Five mutations were run — entity fleet predicate back to falsiness-branching, `memory_add_all` re-query back to NULL-branching, `entity_add` fully reverted, org settings fully reverted to the shallow `||`, and the org row seeded before the diff check — and each kills only the tests it should. The two COALESCE mutations isolate cleanly: reverting just `entity_add`'s predicate produces exactly `ValueError: conflict but re-select returned nothing`.


Eldad-Caura:
@erni-a pleaer approve

erni-a (approved):

PR #1591, opened 2026-09-18
test(conftest): stop the root suite writing to the database developers work in

Eldad-Caura:
## What

`pytest tests/` with no environment set pointed at **`caura`** — the database the
local stack itself runs against. The suite builds its schema with `create_all` and
writes real rows through the service layer, so a plain run did DDL on a developer's
working data and left rows behind under dozens of tenant prefixes.

`core-storage-api/tests/conftest.py` already defaults to its own `caura_storage`
for exactly these reasons, and [says so in a comment](core-storage-api/tests/conftest.py#L20-L31).
This is the same isolation for the other suite: **`caura_test`**.

## Why it stayed invisible

`create_all` creates missing **tables** only — it never adds a column to a table
that already exists. Once the models move on, the schema silently stops matching
them and failures surface far from the cause. Measured on a developer machine:
**538 failures from one absent column**.

The sweep meant to contain the damage covered **7 of 20** tenant-scoped tables, and
its own comment conceded the rest *"leak the same way ... a separate cleanup"*.
Measured on a developer database beforehand: **802 leaked rows, 794 of them in the
13 tables the literal omitted** (`agents` 267, `background_task_log` 61,
`fleet_nodes` and `fleet_commands` 38 each).

Two mechanisms kept that from ever being noticed, and both are fixed here:

1. **Every statement swallowed its exception into a bare `pass`** — which, as the
   docstring itself conceded, made a DELETE naming a missing table indistinguishable
   from one that worked. That is the only way a sweep quietly stops sweeping.
   Failures are now collected into a warning. Deliberately **not** raised: this runs
   in session teardown, where raising would hide whatever the suite was reporting.

2. **The whole sweep runs inside one `engine.begin()`**, and in PostgreSQL a
   statement that errors aborts that transaction — so the first failure took every
   later DELETE with it, and the new warning would have named twenty broken tables
   instead of the one that broke. Each statement now runs in its own savepoint.

## Changes

- `tests/conftest.py` — default DB `caura` → `caura_test`, hoisted to
  `DEFAULT_TEST_DB_URL`; swept tables derived from `Base.metadata` (reversed, since
  `sorted_tables` is parent-first and a DELETE sweep must run child-first); failures
  warned rather than swallowed; one savepoint per statement, which also collapses
  four duplicated `try/except` blocks into one helper.
- `CONTRIBUTING.md` — names both databases with a `createdb` block, and drops a
  pasteable `createdb memclaw_storage` that named a database nothing provisions any
  more (ci.yml:572 and the storage conftest both say `caura_storage`).

## Verification

- Full root suite against a **fresh** `caura_test`: **7161 passed, 31 failed** — the
  same 31 as the local baseline, which are pre-existing environment gaps. No new
  failures; two baseline failures are gone, both AST scanners over source that this
  branch does not touch (they were failing on the baseline's own branch).
- Four new tests, **each confirmed to fail without its fix**: the swept set equals
  the tenant-scoped set; children are deleted before parents; the default is not a
  database anyone works in; one broken table does not abort the rest of the sweep.
- `ruff check` and `ruff format --check` clean. CI runs mypy on `src/` and
  `scripts/`, not `tests/`.


Eldad-Caura:
## Declined as filed — both failure modes are prevented by a cascade the review assumed absent. Reordered anyway.

The finding rests on *"a dedicated manual query is needed for `memory_entity_links` at all suggests there is no `ON DELETE CASCADE`"*. That inference is the part that doesn't hold — the dedicated query exists because the table has no `tenant_id`, so the metadata-derived loop cannot reach it. Cascade and tenant-scoping are independent.

Both foreign keys are `ON DELETE CASCADE`, on the model and in the database:

```
memory_entity_links_memory_id_fkey | c | FOREIGN KEY (memory_id) REFERENCES memories(id) ON DELETE CASCADE
memory_entity_links_entity_id_fkey | c | FOREIGN KEY (entity_id) REFERENCES entities(id) ON DELETE CASCADE
```

So neither branch occurs:

- **No FK violation** — deleting `memories` cascades, so the DELETE cannot fail. Corroborated independently: the full root suite emits **zero** purge warnings, and a warning on every session is exactly what the "likely outcome" branch predicts.
- **No orphaned rows** — the cascade removes the link rows, so the later statement is a no-op against an empty set rather than a leak.

### What I took from it

The *ordering* instinct is sound even though its stated consequences aren't, so I moved the statement ahead of the loop. To check whether that is worth anything rather than assume it, I recreated the constraints without the cascade — the schema the review assumed — and ran both arrangements against it:

| arrangement | with cascade (actual) | without cascade |
|---|---|---|
| DELETE **after** the loop (before this push) | passes | **fails** — `ForeignKeyViolationError` on `DELETE FROM memories`, `memories` leaks behind a warning |
| DELETE **before** the loop (now) | passes | passes |

So the mechanism described is real; it is simply not reachable under the current schema. Running the statement first costs nothing and removes the dependency.

Added `test_a_memory_and_its_entity_link_are_both_swept`, which seeds a memory, an entity and the row joining them, sweeps, and requires all three gone with no warning — so a future schema change that drops the cascade fails here instead of leaking silently.

One note for the record: the deadlock ordering in `tests/test_entity_link_lock_order.py` depends on the *creation order* of those two constraints, because Postgres names RI triggers after the constraint OID and fires them in that order. Recreating one FK on its own flips it and that test fails. I hit it while testing the above and restored the original order; it is a property of the database, not of this change.

Eldad-Caura:
@erni-a plears aprpove

erni-a (approved):

Eldad-Caura:
@claude

The previous round is answered above, and the code in the flagged area changed after your pass, so this asks for a look at the delta rather than a re-review of the whole PR.

The finding was declined on evidence: both `memory_entity_links` foreign keys are `ON DELETE CASCADE` (model and live database), so neither the FK violation nor the orphaned-rows branch can occur.

Changed since: the `memory_entity_links` DELETE now runs ahead of the `_tenant_scoped_tables()` loop rather than after it, plus one new test. Worth checking specifically:

- whether running it first has any consequence I have not considered, given the cascade already removes those rows
- whether `_delete_or_record`'s savepoint is correct in the `engine.begin()` the sweep is handed

This push was a force-push (single-commit convention), which is why the automatic review skipped.

PR #1592, opened 2026-09-18
fix(events): hold the Pub/Sub lease while a batch drains, instead of letting finished work redeliver

Eldad-Caura:
Closes four findings that are one mechanism in one file: `oss-0814-m-02`, `oss-0902-l-12`, `oss-0902-m-67`, and `ent-056` (the enterprise duplicate of this vendored file — fixed here so the re-vendor carries it, rather than patched there).

## The defect

`_pull_loop` pulls up to `max_messages` (25), dispatches them **sequentially**, and calls `acknowledge` **once at the end**. Nothing extended the ack deadline — the only `modify_ack_deadline` in the file is the nack, with `ack_deadline_seconds: 0`.

So the message dispatched first stays unacked for the whole drain. A batch whose aggregate time passes the 30s deadline has its **already-completed** work redelivered, and the redelivered copy takes just as long — a subscription that falls behind this way stays behind.

Handlers must be idempotent, so this is duplicated work and duplicated provider spend rather than corruption. But two instances running the same org sweep concurrently is not a cost anyone chose, and that is `m-67` exactly.

## Why not just raise `ack_deadline_seconds`

It is set at subscription creation, and `_ensure_broadcast_subscription` already notes that an existing subscription *keeps whatever it was created with*. Raising the constant reaches **no deployment already running**. Extending the lease per batch does.

## Why a timer, not a refresh between messages

A single handler that pages through a whole org can outlast the deadline **on its own**, and a between-messages refresh would never run during it. That is the embed-backfill sweep — which is why one change covers all four findings.

Dispatch stays sequential: the constructor comment is explicit that the serialisation is the only thing holding per-instance handler concurrency near one, and that lifting it needs a bound of its own. The fix is on the lease, not the concurrency.

## Values, and the tradeoffs behind them

| choice | value | why not otherwise |
|---|---|---|
| extension | **60s** | not the 600s the API allows — an extension is also how long a message stays invisible if this process dies holding it; 600 turns a pod kill into a ten-minute stall for everything in flight |
| refresh | **15s** | 15s of slack against the initial 30s deadline on the first refresh, 45s after |
| total | **unbounded** | capping would hand the message to another instance while this one is still inside the handler — the duplicate-sweep outcome this exists to prevent. A hung handler also stalls the loop outright, so it surfaces as a stopped subscription, not a lost message |

The keeper is cancelled **before** the acks go out, so a refresh can never name ids an `acknowledge` is retiring in the same instant, and cancelled again in a `finally` so no exit path leaves one extending messages nothing is processing. A failed extension is logged and the loop continues — losing the lease means redelivery, which is what this guards against and what handlers already tolerate; halting consumption would be strictly worse.

## Verification

- Full root suite: **7167 passed / 31 failed** — the same 31 pre-existing environment gaps as `main`, no new failures.
- `tests/test_events/` 152 passed; `core-worker/tests/` 100 passed.
- mypy clean on this file across core-api, core-storage-api and core-worker (the 2 errors those report are pre-existing, in `common/llm/providers/openai.py`).
- `ruff check common/` clean — the check CI actually runs on this path.
- **3 of the 4 new tests fail without the fix.** The 4th (an empty pull must not spawn a keeper) is a non-regression guard and passes either way; called out rather than counted as proof.

The extension test is deterministic, not timed: the first dispatch blocks until an extension is actually observed, so without the keeper it fails by timeout rather than by a race that would pass on a fast machine.


Eldad-Caura:
## Confirmed — and it caught an over-claim, which is the more important half

The finding is right, and the comment it quotes was mine asserting a property the code does not have. I had considered this race while designing and judged it benign, then wrote *"a refresh can never race an `acknowledge`"* anyway. Benign is not never.

Measured rather than reasoned about, since the whole point is that I had already reasoned about it wrongly once:

```python
t = asyncio.create_task(keeper())     # keeper awaits run_in_executor(slow_rpc)
await asyncio.sleep(0.05)             # let it get INSIDE the executor call
t.cancel()
await asyncio.sleep(0.5)
# task cancelled: True
# executor work  : ['rpc ran to completion']
```

`cancel()` interrupts the *await*, not the thread. So a refresh already inside `run_in_executor` completes regardless, exactly as described.

## Applied

**The claim is corrected.** The comment now says the cancel *narrows* the window and does not close it, and says why — with the measurement noted so the next reader does not have to re-derive it.

**Shutdown is now a signal plus a cancel.** A `stop_extending` event lets the keeper leave cleanly *between* refreshes, which is the common case; the cancel stays as the backstop for one parked inside a refresh. This shrinks the window rather than just documenting it.

**The log is classified**, which is the part you flagged. On failure the keeper reads `stop_extending`:

- set → the refresh lost the race to the ack. `logger.debug`, no `exc_info`, message says there is nothing to do.
- not set → the lease really is lapsing and this batch really will redeliver. Stays `WARNING` with `exc_info`.

I took this over option 1 (inspecting gRPC status codes) deliberately: matching on `NOT_FOUND`/`INVALID_ARGUMENT` infers intent from a code that a genuine IAM or request-shape fault can also produce, so it would downgrade real failures. Whether we asked the keeper to stop is a fact we own, not an inference.

## Tests

Two added, one per side of the branch, and each confirmed to fail under the mutation that breaks it:

| mutation | which test fails |
|---|---|
| always `warning` (the behaviour you reviewed) | `test_a_refresh_that_loses_the_race_to_the_ack_is_not_warned_about` |
| always `debug` | `test_a_genuine_extension_failure_is_still_warned_about` |

The second exists because downgrading the whole branch would have silenced the signal this PR adds — a fix for alert fatigue that quietly removes the alert is not a fix.

Full root suite: 7169 passed / 31 failed, the same pre-existing baseline. `tests/test_events/` 154 passed.

Eldad-Caura:
@erni-a plearas eaprpove

erni-a (approved):

PR #1594, opened 2026-09-18
fix(core-storage-api): make status writes respect soft-delete, and the org purge report what it deletes

Eldad-Caura:
Closes `oss-0814-l-18` and `oss-0902-l-50`. Both are the code disagreeing with a contract stated a few lines away from it.

`oss-0902-m-60` is from the same file and same batch but is **not** here — see the bottom.

## L-18 — status writes reached soft-deleted memories

`memory_update_status` scoped its UPDATE to id + tenant, with no `deleted_at` guard. Its sibling `memory_update` has always guarded, and so does nearly every read path.

The route stated the contract explicitly, directly above the call:

> `memory_update_status` returns False when the target row doesn't exist (**or was already deleted**); surface as 404 so the caller doesn't silently treat a no-op as success.

It did not. A deleted row matched, updated, and returned True — so the route answered **200 for a write its own comment says is impossible**.

What gets rewritten matters beyond the deleted row: `supersedes_id` is an edge into a memory that is *still live*, so a flip landing on a deleted row changes what the contradiction detector reads about a row nobody deleted.

The route's two inline SQL paths had the same gap and are guarded with it. The retraction path also answers honestly now — zero rows is reached three ways (missing, deleted, genuinely taken by another writer) and only the last is a conflict the caller can fix by re-reading, so it classifies with one SELECT on the failure path rather than reporting `stale_retraction` for all three.

## L-50 — the purge under-reported, but nothing survived

**The finding's framing is wrong and I want to be explicit about it.** It says `memory_conflicts` and `memory_derivations` survive an org hard-delete. They do not: all four of their foreign keys are `ON DELETE CASCADE` and `NOT NULL`, verified on the model and in the live database, so they were always deleted along with `memories`. The existing `test_every_tenant_scoped_table_is_purged_or_deliberately_retained` accepts a cascade as a valid fate and passes — correctly.

The real defect is narrower. The module's rule is:

- a table with its own `tenant_id` → **listed explicitly**, *"so their per-table counts are reported rather than hidden inside a cascade"*
- a table without one (`memory_entity_links`, `recall_candidate`) → left to ride the cascade

These two (migration 036) have a `tenant_id` and were unlisted, so they fell between the rules: never at risk, but **invisible in the purge result and in the deletion preview an operator reads before destroying an org**.

Listing `memory_conflicts` for the tenant purge forced listing it for the *fleet* purge too — it carries its own `fleet_id`, so a fleet teardown would otherwise leave its rows behind in a shared tenant. An existing test caught that, not review.

A new test pins the narrower rule the existing one cannot, since its cascade exemption is exactly what it is designed to allow.

## Verification

- `core-storage-api/tests/` **447 passed**; full root suite **7163 passed / 31 failed**, the same pre-existing environment gaps as `main`.
- mypy clean (92 files); `ruff check` and `ruff format --check` clean on `core-storage-api/src/` and `tests/`.
- **6 of the 10 tests fail without their fix.** The other four are non-regression guards and are labelled as such rather than counted as proof: a live memory still updates, a stale retraction on a live row is still a 409, and the two pre-existing purge invariants still hold.

## Not in this PR: `oss-0902-m-60`

Same file, same batch, deliberately left open — it needs a decision, not a patch.

`expires_at` is accepted, stored, returned, and offered as a sort key, and **no read path anywhere filters it**. There *is* an expiry sweep, but `memory_archive_expired` filters `ts_valid_end`, a **different column** — so the machinery that looks like it enforces expiry enforces something else.

I did not enforce it unilaterally because there is nothing to enforce *against*: `expires_at` appears in no doc, no MCP tool schema, no plugin manifest, no published fixture. Turning it on would stop serving rows customers can see today with no written contract saying it should. Options are recorded on the tracker row.


Eldad-Caura:
@erni-a plase rapprove

erni-a (approved):

Eldad-Caura:
## Confirmed real, accepted deliberately, now documented in the code

Verified rather than assumed: `get_session()` yields inside `session.begin()`, so the UPDATE and the classification SELECT do share a transaction — but Postgres runs READ COMMITTED, where each **statement** takes a new snapshot. A commit landing between them is visible to the SELECT. The window is real.

What I did not take on faith is the consequence. Enumerating the interleavings, the answer stays correct *about the row as it now stands*, and each is self-correcting on the caller's retry:

| interleaving | answered | correct? |
|---|---|---|
| stale pointer, then row deleted | 404 | yes — the row **is** deleted |
| deleted, then pointer moved | 404 | yes — still deleted |
| stale pointer, then pointer restored | 409 | yes — a re-read now shows the expected pointer, so the retry succeeds |

The one answer that would be plainly wrong — live row, pointer never moved, yet the UPDATE matched nothing — **cannot arise**: that is precisely the case the UPDATE succeeds in.

So this is not a wrong status code that misleads a caller into the wrong action; it is a status code that describes a state the row genuinely reached, one moment later than the statement that failed.

## Why not the CTE / lock

Both suggestions work. Both cost more than the defect:

- A `SELECT ... FOR UPDATE` before the UPDATE takes a row lock on **every** retraction to sharpen a status code on a rare failing one.
- A single-statement CTE does close it, at the price of hand-written multi-CTE SQL in a router whose whole point in this PR was that its inline SQL had drifted from the service method beside it. Trading a drift-prone inline statement for a more complex drift-prone inline statement is the wrong direction here.
- `SERIALIZABLE`/`REPEATABLE READ` for this endpoint introduces serialization failures the caller must now handle — a new retry contract to fix a case that already resolves on retry.

Taking your third option, explicitly: the window is documented in place, with the interleaving table and the reason a lock was not worth it, so the next reader finds the analysis instead of redoing it — or has what they need to disagree and add the lock.

No behaviour change in this push; the storage suite is still 447 passing.

PR #1595, opened 2026-09-18
fix(core-storage-api): make POSTGRES_REQUIRE_SSL real, and stop .env.example overriding the pool baseline

Eldad-Caura:
Closes `oss-0902-m-01` and `oss-0902-m-02` — two database settings that documented themselves into being ignored.

## M-01 — a security control that could not be switched on

`POSTGRES_REQUIRE_SSL` is published in **`.env.example`, `AGENT-INSTALL.md` and both compose files** as *"Set true in production"*, and was read by nothing. No `connect_args`, no `sslmode`, no `ssl=` anywhere in the tree — including the AlloyDB URL builder.

`Settings.model_config` sets `extra="ignore"`, which is exactly why it was silent: pydantic-settings accepted the variable and dropped it. **The operator got neither TLS nor an error.**

> A security control that cannot be switched on is worse than an absent one — the absent one doesn't tell you it's protecting you.

It is now bound to a real setting and applied through one `db_connect_args()` helper at **all three** places this service opens a connection: the app's engines, the migration runner, and the preflight script. One helper rather than a flag read per site, because a migration running in cleartext against a database the app reaches over TLS would defeat the setting while looking configured.

**`"require"` was measured, not assumed** — against the local non-TLS Postgres:

```
default (no ssl arg)   CONNECTED
ssl='prefer'           CONNECTED
ssl='require'          REFUSED -> ConnectionError: ... rejected SSL upgrade
```

So it genuinely enforces, and `prefer` would have left the same false promise in place while looking fixed.

It deliberately does **not** verify the certificate — that is `"verify-full"`, which needs a CA bundle shipped and configured, so offering it here without one would be the same empty promise in a new place. Verification goes on the DSN, which asyncpg honours.

Off by default, and the connect args are empty when off, so **the default deployment's engine call is byte-identical to what it was**.

## M-02 — the example file overrode the baseline it was meant to illustrate

`.env.example` set `DB_POOL_SIZE=50` / `DB_MAX_OVERFLOW=50` — ten times the `5 + 5` in the source.

That baseline is deliberate, and the config comment says so: defaults were lowered from `20 + 20` after over-allocating against the managed database's connection ceiling produced `asyncpg.TooManyConnectionsError` under load, *"so the source default is now the safe baseline rather than a value that requires environment-side correction."*

Recommending `50 + 50` in the file everyone copies reintroduced exactly that correction — at **2.5x the value that caused the incident**.

Now commented out, so the code default applies and raising it is a decision rather than an inheritance, with the connection-budget arithmetic stated for anyone who does need to.

## Verification

- `core-storage-api/tests/` **446 passed**; full root suite **7169 passed / 31 failed** — the same pre-existing environment gaps as `main`.
- mypy clean (92 files); `ruff check` and `ruff format --check` clean.
- **All six tests confirmed to fail without their fix**, across three separate mutations: reverting `.env.example`; removing the connect args from a *single* one of the three sites; and renaming the settings field out from under the published variable name.

That last mutation is the one worth keeping: the test asserts the **published env-var name**, not the field. Under `extra="ignore"`, renaming the field without updating the docs would silently make the variable inert again — which is the whole bug — and every other test here would still pass.


Eldad-Caura:
## Confirmed, and it was my comment that was wrong

Measured rather than reasoned about, on the versions this repo pins (SQLAlchemy 2.0.50, asyncpg 0.31.0), by reading the kwargs actually handed to the driver:

| DSN | `connect_args` | what asyncpg receives |
|---|---|---|
| `?ssl=verify-full` | `{}` | `ssl='verify-full'` |
| `?ssl=verify-full` | `{"ssl": "require"}` | **`ssl='require'`** |

The mechanism is one line in `sqlalchemy/engine/create.py`:

```python
cparams = util.immutabledict(_cparams).union(pop_kwarg("connect_args", {}))
```

`connect_args` is unioned **on top of** the URL-derived params, so the kwarg wins. My `preflight_012.py` comment asserted the exact opposite ("A DSN that already specifies its own `ssl` wins, since asyncpg reads the URL after these kwargs") and the `config.py` docstring pointed operators at `?ssl=verify-full` as the supported way to get verification — which this PR would then have turned off for them, silently, with no error. That is the same defect class the PR exists to fix, reintroduced one file over.

### The fix: the setting is a floor, not an override

`db_connect_args()` now takes the DSN:

- DSN already at `require` or stronger → returns `{}`, the DSN stands.
- DSN weaker (`disable`/`allow`/`prefer`), unrecognised, or absent → `require`.

Deferring to *any* DSN-supplied `ssl` was the tempting one-liner and it is wrong: `?ssl=disable` would then defeat `POSTGRES_REQUIRE_SSL=true` in silence, which is precisely the failure the setting exists to stop. `allow` and `prefer` fall back to cleartext, so neither satisfies a control named "require". Unrecognised values fail closed — an unknown mode is not evidence of encryption.

All three call sites pass their own URL (`_build_engine(url)`, the migration runner, `preflight_012`'s arbitrary `dsn`).

### On the suggested scope

I did not add `sslmode` handling. `asyncpg.connect()` has no `sslmode` parameter and no `**kwargs` (checked the signature), so `?sslmode=` on a `postgresql+asyncpg://` URL is a `TypeError` at connect time today, with or without this setting — recognising it here would paper over a broken config instead of surfacing it. `ssl` is the only knob this driver has.

### Tests

10 new cases, and the one that matters asserts on **what SQLAlchemy actually hands asyncpg**, captured through the public `do_connect` event, not on what the helper returns in isolation:

```python
engine = create_async_engine(dsn, connect_args=db_connect_args(dsn))
@event.listens_for(engine.sync_engine, "do_connect")
def _capture(dialect, conn_rec, cargs, cparams):
    captured.update(cparams)
    raise _Captured  # fires before the DBAPI connect: no server needed
```

A unit test on the return value could not have caught the original bug — the return value was correct in isolation; the merge was the problem. This one fails if the helper stops reading the DSN **and** if a future SQLAlchemy flips the merge direction, which would make the care unnecessary and the comment wrong again.

`test_every_engine_this_service_opens_carries_the_policy` now also asserts each site passes a non-empty argument, so dropping the DSN at one site is caught rather than just dropping the whole call.

Revert-checked, three mutations, each failing the intended tests and nothing else:

| mutation | fails |
|---|---|
| helper ignores the DSN (the bug you found) | 4 — the three `already_guarantees_tls` params + the `verify-full` end-to-end case |
| helper defers to any DSN `ssl` | 3 — `disable`, `allow`, `prefer` |
| one call site drops its DSN | 1 — the source-site test |

### Gates

`core-storage-api/tests/` 456 passed. Full root suite 7169 passed / 31 failed, identical to the pre-existing baseline on `main`. `mypy src/` clean (92 files). `ruff check` and `ruff format --check` clean.

@claude

Eldad-Caura:
## Both confirmed. One of them is a better finding than its own writeup.

### Medium — `--dsn` no longer survives broken service config: correct, applied

Verified the premise rather than the conclusion: `_resolve_dsn` wraps its `from core_storage_api.config import settings` in `try/except Exception` *precisely* because constructing `Settings()` can fail in the shell this script is run from, and `--dsn` skips `_resolve_dsn` altogether. My unconditional import put that dependency back on every path, including the one documented as the escape hatch.

Fixed at the mechanism rather than with a `try/except` around the import. The policy moved into a new `core_storage_api/db_tls.py` that constructs no `Settings` and imports nothing that does; `config.db_connect_args` is now a two-line delegation to it, and the preflight uses the configured helper when it loads and the same policy over `POSTGRES_REQUIRE_SSL` from the environment when it does not.

Deliberately **not** your parenthetical first option ("falls back to no `connect_args`"). Answering "your config did not load" by dropping TLS is the same silent downgrade this PR exists to delete — it would just fail in a different room.

### Low — case sensitivity: real, but the consequence is the opposite of the one stated

You wrote that `?ssl=Verify-Full` is *"silently dropping certificate verification"*. Measured against asyncpg 0.31.0:

```
ssl='verify-full' -> ClientConfigurationError: root certificate file ... does not exist   (accepted, then wants a CA)
ssl='Verify-Full' -> ClientConfigurationError: `sslmode` parameter must be one of: disable, allow, prefer, require, verify-ca, verify-full
ssl='REQUIRE'     -> ClientConfigurationError: `sslmode` parameter must be one of: ...
ssl='require'     -> ConnectionError: server rejected SSL upgrade                          (accepted, then enforced)
```

asyncpg is case-sensitive and **rejects** a mis-cased mode outright. So deferring to it is not silent at all — the operator gets an error naming the six valid spellings. The genuinely bad branch is the one my code took: overriding it with `require`, which **connects**, converting a precise, loud config error into a working connection at a weaker mode than was asked for.

Which changes the shape of the fix. Recognition is now case-insensitive, but the DSN's own text is **never rewritten** — `?ssl=Verify-Full` is understood as verify-full, left exactly as written, and asyncpg still raises. Lowercasing the value on its way through would have masked the operator's typo in a new place.

### On my own test that nearly passed for the wrong reason

The new subprocess test asserts `db_tls` imports under an env where `Settings()` raises. My first version set `ALLOYDB_HOST` alone and **the assertion passed while proving nothing** — `resolve_alloydb_database_url` returns early when `database_url` was set explicitly, and the suite sets `DATABASE_URL`. The vacuity check in the same test (assert the env really does break `Settings()`) is what caught it. `DATABASE_URL` now comes out of the subprocess env, and both halves are asserted.

### Tests

28 in this file now (was 16). New: three mis-cased DSN forms; seven `POSTGRES_REQUIRE_SSL` spellings through the env fallback; the preflight fallback with `core_storage_api.config` forced to `ImportError`; and the import-independence subprocess above.

Revert-checked, three mutations:

| mutation | fails |
|---|---|
| recognition back to case-sensitive | 3 — every mis-cased form |
| preflight imports config unconditionally | 1 — `ModuleNotFoundError` on the fallback test |
| `db_tls` imports `config` | the file, on a circular import — which is the coupling itself |

### Gates

`core-storage-api/tests/` 468 passed. Full root suite 7169 passed / 31 failed, identical to the baseline on `main`. `mypy src/` clean (93 files). `ruff check` and `ruff format --check` clean.

@claude

Eldad-Caura:
Clean verdict on the amended head (`e89e2e16`), all 11 CI checks green.

Two rounds of findings, both real and both applied: `connect_args` merge precedence (my comment had it backwards, which would have downgraded `?ssl=verify-full` to `require` in silence), and the preflight's `--dsn` escape hatch losing its independence from service config. The policy now lives in `core_storage_api/db_tls.py`, which constructs no `Settings`, so both the configured path and the environment fallback run the same code.

Ready for you, @erni-a.

erni-a (approved):

PR #1599, opened 2026-09-18
fix(reports): delete the dead report_type filter, make fleet_id on /reports/latest real

Eldad-Caura:
Closes OSS audit **08/14 L-30** and **09/02 L-48** — `report_type` and `fleet_id` on the report lookups.

## What was wrong

`GET /reports/running` and `GET /reports/latest` both declared `report_type`; `/latest` also declared `fleet_id`. core-api's storage client forwarded all of them, and `_reserve_report` wrote `"report_type": "crystallization"` into the create body. **Nothing filtered on either one.**

| layer | declared | applied |
|---|---|---|
| `GET /reports/running` | `report_type` | — |
| `GET /reports/latest` | `report_type`, `fleet_id` | — |
| `storage_client` | both, on both calls | forwards them |
| `_reserve_report` create body | `report_type` | dropped by `_filter_fields` |
| `analysis_reports` | — | **no `report_type` column exists** |

## The two are fixed in opposite directions

**`report_type` is removed.** There is one report type, the model is literally `CrystallizationReport`, and no column exists to hold it. A parameter accepted and plumbed through three layers reads as a scoping guarantee to everyone downstream; the honest form is to stop claiming a filter exists. Safe to remove because FastAPI ignores unknown query params — an older core-api still sending `?report_type=` is unaffected, which is asserted rather than assumed.

**`fleet_id` on `/reports/latest` is implemented**, because it has a caller that depends on it:

```python
latest = await sc.get_latest_report(tenant_id, fleet_id)   # crystallizer_service
```

`_type_ii_watermark` uses that report's `completed_at` to decide which subjects the nightly Type-II sweep may skip. Handed a *different* fleet's more recent run, a fleet whose own sweep is older skips subjects that have changed since — `select_candidates` compares the watermark to `created_at` with `>`. That helper's own docstring names this the unrecoverable direction:

> paying twice is a cost bug, skipping a changed subject is a correctness bug, and only one of those is recoverable on the next run.

So this is a live correctness bug in any multi-fleet tenant, not only dead code.

## The test that was green the whole time

`oss-0902-l-43` added fleet scoping and a test for it:

```python
async def test_watermark_is_scoped_to_this_fleet_and_report_type():
    """A different fleet's (or a non-crystallization) report would be the wrong
    clock — it says nothing about when THESE subjects were last swept."""
    await cs._type_ii_watermark(sc, "t1", "fleet-a", "cur")
    assert sc.calls == [("t1", "fleet-a", "crystallization")]
```

It asserts the arguments were **passed**, against a fake. They were — into a route that ignored them. The property the docstring names was false for as long as the test was green. It now asserts the **effect**, against a real database for the storage half.

## The asymmetry, on purpose

Absent `fleet_id` still means **any fleet** on `/reports/latest`, unlike the sibling `/reports/running` where absent means `IS NULL`. The two answer different questions:

- `/running` asks "is a run in flight for exactly this scope" — the tenant-wide run is its own scope and must not be blocked by a fleet's.
- `/latest` also backs `GET /crystallize/latest`, a user-facing "show me my most recent report" with **no fleet concept in its API**. Filtering to `IS NULL` there would 404 every tenant whose runs are all fleet-scoped.

That leaves the tenant-wide run able to read across fleets, which the server cannot close without breaking that endpoint. It is closed in `_type_ii_watermark`, where the run knows its own scope, in the same safe direction as its other guards. Both halves are documented where the next reader will be standing, and both are tested.

## Tests

10 new (7 storage + 3 root), revert-checked against three mutations:

| mutation | fails |
|---|---|
| route stops forwarding `fleet_id` | 2 — the sibling-fleet and empty-fleet cases |
| service ignores `fleet_id` again | the same 2 |
| watermark drops its scope guard | 2 — the cross-fleet and tenant-wide cases |

The `report_type` sweep is **tokenised, not grepped**: all three layers now *document* the removal, so a line-wise search matches the explanations and reports the fix as the defect. Only `NAME` tokens count.

## Gates

`core-storage-api/tests/` 447 passed · full root suite 7171 passed / 31 failed, identical to the pre-existing baseline on `main` · `mypy src/` clean for core-storage-api (92 files) and core-api at its known 2-error `common/llm/providers/openai.py` httpx baseline · `ruff check` and `ruff format --check` clean.

## Not in scope

`oss-0902-l-31` — five crystallizer *hygiene checks* ignore their `fleet_id` — is the same disease in a different place, and belongs to @erni-a's queue rather than this PR.


Eldad-Caura:
CI red on `bc11770a` was mine, and the cause is worth recording because it was a bad note rather than a bad change.

\`\`\`
Run Ruff formatter check (root tests)
  uv run ruff format --check tests/
  unformatted: tests/test_low_oss_0902_l_43_type_ii_watermark.py
\`\`\`

I had been working from a recorded belief that CI does not format-check root `tests/`. It does — `.github/workflows/ci.yml:258`. So my local gate sweep, which was built from that note, skipped the one directory this PR touches outside `core-storage-api/`. Reading the workflow instead of trusting the note would have caught it before the push; I have replaced the note with the file.

Fixed in `349211a1`. `ruff format` on that file changes **only lines this PR added** — three call sites wrapped — and touches nothing pre-existing, which also confirms the file is format-clean on `main` under the same ruff.

Re-ran the complete lint scope as the workflow actually defines it this time, not as I remembered it: `ruff format --check` over `core-api/src`, `core-storage-api/{src,tests}`, `core-operations/{src,tests}`, `core-worker/{src,tests}`, `scripts/`, `tests/` (892 files, all formatted), `ruff check` over the same plus `common/` (all passed), and the isolated `common/` three-file check. Watermark tests still 11 passed.

No source change, so the review verdict above still stands on this head.

@erni-a

erni-a (approved):

PR #1600, opened 2026-09-18
fix(dead-code): drop four unreachable helpers, and make IS_STANDALONE reach the flag it names

Eldad-Caura:
Closes six OSS audit findings — **08/14 L-46, L-47, L-51** and **09/02 L-15, L-17, L-45**.

Five are removals. One is a bug the removals were sitting next to, and it is the reason this PR is worth reading.

## The one that isn't cleanup: the standalone guard could not be switched on

`core-operations` has a defensive short-circuit whose own comment explains the stakes:

> OSS standalone deployments should not deploy core-operations at all; this flag is a defensive short-circuit so that an accidentally-started instance exits cleanly rather than firing cron jobs against a single-tenant standalone DB.

The field was named `standalone`, which binds the environment variable **`STANDALONE`**. Nothing in the tree sets that name. What everything sets is **`IS_STANDALONE`**:

| sets `IS_STANDALONE` | `.env.example`, `env.dev`, `.env.test`, `README.md`, `AGENT-INSTALL.md`, `docs/self-hosting.md`, `docs/public-api-stability.md`, `docs/local-embedder.md`, `.github/workflows/ci.yml`, `tests/conftest.py`, and core-api's own `is_standalone` |
|---|---|
| sets `STANDALONE` | *nothing* |

`model_config` sets `extra="ignore"`, so pydantic accepted `IS_STANDALONE` and discarded it. The guard therefore never fired: an accidentally-started core-operations in a standalone deployment ran its cron jobs against a single-tenant DB **and logged `standalone: False` while doing it** — the misconfiguration reporting itself as correct configuration.

Same mechanism as `POSTGRES_REQUIRE_SSL` in #1595, one service over. A knob that is published, documented, and bound to nothing.

Renamed rather than aliased, and the PR asserts why that is safe rather than claiming it: nothing sets the old spelling, so nothing loses an override. That assertion is a test, with a vacuity check — a scan whose filters exclude every relevant file finds no offenders either, and would pass exactly like a clean tree, so it first asserts it *did* see the files that set `IS_STANDALONE`.

## The removals

**`cache_delete`, `cache_get_json`, `cache_set_json`** (`core-api/cache.py`) — no caller in this repo or any sibling checkout. The one surviving mention of `cache_delete` is a test docstring that names it while the test patches `cache_delete_if`; the detector has always used the compare-and-delete variant, so that reference is corrected rather than left pointing at a function this PR deletes.

**`paginated_order_by`** (`core-api/pagination.py`) — unreferenced, and its docstring asserts a contract nothing honours:

> Both pagination call sites — the repository's `list_by_filters` and the admin route's inline query — must call this helper so the tiebreaker stays in sync with the `tuple_(created_at, id)` cursor predicate.

Neither calls it, and neither **can**: both sites are `memory_admin_list` and `memory_list_by_filters` in **core-storage-api**, while the helper is in **core-api**. It was left behind when the SQL moved services — core-api has no repository layer at all any more, only two prose mentions of one.

**`core_storage_api_url`** (`core-operations/config.py`) — dead, and already known to be: the comment two fields below says "Nothing in this service talks to core-storage-api — all three call sites POST to core-api". That comment was added by an earlier fix that renamed the neighbouring timeout and left this one standing.

## Deleting a docstring is deleting a requirement

`paginated_order_by`'s stated concern is real even though the helper was dead: both sites inline the `(sort, id)` pairing, which is correct, but that is two copies 180 lines apart with nothing asserting they agree. A page ordered by `created_at` alone puts rows sharing a timestamp — a bulk-write tranche collides to ms precision, `status` collides trivially — in an order Postgres may change between executions, so consecutive pages duplicate some rows and skip others.

So the requirement moves from a docstring nobody checked into a test that fails if either site drops it, using this repo's existing `_captured_sql` pattern and its existing argument for why ordering is a statement-shape property rather than a behavioural one. `_captured_sql` gained one optional argument — these are read paths and open `get_read_session`, so patching only `get_session` captures nothing.

## Tests

5 new in `core-operations/tests/`, 4 new parameter sets in `core-storage-api/tests/`, revert-checked:

| mutation | fails |
|---|---|
| rename `is_standalone` back to `standalone` | 4 of the 5 — the 5th is the deployment scan, correctly independent of the rename |
| drop the id tiebreaker at **one** of the two pagination sites | 2 — that site's desc and asc cases, and only that site's |

One assumption of mine was wrong and the test said so: I expected SQLAlchemy to omit an explicit `ASC`. Both sites call `.asc()` rather than leaving the default, so the keyword is emitted — the expectation is now what the code actually produces, with a note saying why.

## Gates

`core-operations/tests/` 86 passed · `core-storage-api/tests/` 444 passed · full root suite 7169 passed / 31 failed, identical to the pre-existing baseline on `main` · `mypy src/` clean for core-storage-api (92 files) and core-operations (6 files), core-api at its known 2-error `common/llm/providers/openai.py` httpx baseline · `ruff check` and `ruff format --check` clean across the full scope the workflow defines, root `tests/` and `scripts/` included.


Eldad-Caura:
## Declining the High — the deletion is the finding, not an accident

`core_storage_api_url` was removed on purpose. It is **08/14 L-51**, one of the six findings this PR closes, and the PR description says so. The rename sitting four lines away in the same hunk is an unlucky diff adjacency, and I can see how it reads.

The premise to check is the consequence, not the intent: *"any code reading `settings.core_storage_api_url` will now raise `AttributeError`."* There is no such code.

Every settings attribute this service reads, taken from the source rather than from memory:

```
settings.agent_digest_http_timeout_s   settings.environment
settings.core_api_admin_api_key        settings.is_standalone
settings.core_api_http_timeout_s       settings.lifecycle_crystallize_every_hours
settings.core_api_url                  settings.log_file
settings.embed_backfill_enabled        settings.log_format_json
                                       settings.log_level
```

The only dynamic access is `getattr(settings, hour_attr)` behind `_daily_at`/`_weekly_at`, whose literals are all `*_run_at_hour` / `*_run_at_weekday` passed at registration.

Corroborating, from the tree itself:

- `config.py` already says it, in the comment four lines below the deleted field: *"Nothing in this service talks to core-storage-api — all three call sites POST to core-api."* That comment was written by an earlier fix which renamed the neighbouring timeout (`storage_http_timeout_s` → `core_api_http_timeout_s`) for exactly this reason and left this field standing.
- `tasks.py:243` calls core-operations *"a separate deployable with no import path into core-storage-api"*.
- `mypy src/` is clean and all 86 tests passed — an `AttributeError` on a pydantic field is not a silent failure mode here.

## What I took from it anyway

The finding is wrong, but the question behind it is a good one, and my answer to it was a grep. A grep is an argument from absence: it says nothing about the reader added next month, and it is exactly the kind of evidence that is convincing right up until it isn't.

So the property is now asserted rather than argued. New test, added this round:

```python
def test_every_setting_this_service_reads_is_one_it_declares() -> None:
```

It walks the AST of every module in `core_operations` and fails if any `settings.<name>` does not resolve to a declared field — which catches a deleted setting that still has a reader, and a typo'd one that never had a value.

Revert-checked by writing your exact scenario back in:

| mutation | result |
|---|---|
| add `settings.core_storage_api_url` back into `tasks.py` | **fails**, naming the file and line |
| restore | 6 passed |

The dynamic `getattr` path is out of AST's reach here (the literal is an argument to the helper, not to `getattr`), and the test says so rather than implying coverage it does not have — that path is covered behaviourally by `test_all_lifecycle_jobs_registered_and_wall_clock_aligned`, which invokes every registered task's `delay_provider` and would raise on a bad name.

## Gates on `736f4979`

`core-operations/tests/` 87 passed · `ruff check` / `ruff format --check` clean · `mypy src/` clean (6 files). Source is unchanged from the reviewed head — this round adds a test only.

@erni-a — flagging explicitly that a **High was declined** rather than applied, so it is a deliberate call and not something buried in a green tick. The evidence is above; if you read it differently, the field is one line to restore.

erni-a (approved):

Eldad-Caura:
CI green on the amended head `736f4979` — 11 checks pass, 0 fail. Verdict handled (the High declined with evidence above, plus the AST test that answers it mechanically). Ready for review.

PR #1601, opened 2026-09-18
fix(events): let the nack be the last word, and stop assuming a 30s deadline

### LINKED ISSUE #1597: Leftover lease extension can override a nack's immediate-redelivery deadline
## Summary

A lease extension already in flight when `_hold_leases` is cancelled can land **after** the batch's nack and re-extend a deliberately-nacked message's deadline from 0 back to `LEASE_EXTENSION_SECONDS`, delaying its redelivery by up to 60s instead of making it immediate.

Introduced by #1592. The comment above the cancel reasons carefully about this race for the **ack** path and concludes it is harmless there — which is right — but the same leftover refresh is *not* harmless on the **nack** path, and that case is not considered.

## Why the ack reasoning does not carry over

On `main` @ `6fe9fdb`, `common/events/pubsub.py`:

```python
stop_extending.set()
if lease_keeper is not None:
    lease_keeper.cancel()
```

The comment states, correctly, that `cancel()` only interrupts the keeper while it is *waiting*, so a refresh already inside `run_in_executor` runs to completion on its thread regardless. For `acknowledge` that is fine: the extension names ids the ack is retiring, so it is a no-op.

The nack is a different operation. It works by **setting the deadline to zero** to force immediate redelivery (L1308-1318):

```python
nack_request = {
    "subscription": sub_path,
    "ack_ids": nack_ids,
    "ack_deadline_seconds": 0,
}
```

A leftover refresh carries `ack_deadline_seconds: 60` for the *whole* leased batch, `nack_ids` included. If it reaches the server after the zero-deadline call, the failed message sits invisible for another 60s. The intent of the nack — redeliver now, ideally to a healthy instance — is silently inverted.

## Why the ordering is genuinely unguaranteed

The two RPCs are submitted to the **same** executor, but it is multi-threaded (L954-957):

```python
self._pull_executor = concurrent.futures.ThreadPoolExecutor(
    max_workers=max(4, 2 * topic_count),
    thread_name_prefix="pubsub-pull",
)
```

With `max_workers` at least 4, the leftover refresh and the nack run on **different threads** concurrently, so neither the submission order nor the arrival order at the Pub/Sub server is ordered. (Had the executor been single-threaded, the queue would have serialised them and this could not happen.)

## Severity

Low-to-medium and bounded: delayed redelivery, never loss, and only when a refresh is mid-flight at the instant the batch ends **and** a handler failed in that batch. The window is roughly one RPC duration in every `LEASE_REFRESH_INTERVAL_SECONDS`. But it is silent, and it defeats an explicit "redeliver now" — the kind of thing that is very hard to diagnose from the outside later.

## Possible fixes

1. **Re-issue the zero-deadline call for `nack_ids` after the keeper is known to be finished.** Simple and local; costs one extra RPC only on batches that actually nacked.
2. **Await the cancelled keeper before ack/nack.** Closes the window completely, but #1592's comment argues deliberately against awaiting — it would fold an outer cancellation (`stop()` cancelling the pull task) into the keeper's own `CancelledError`. Any fix here needs to handle that, so it is not the one-liner it looks like.
3. **Nack first, then ack**, and accept the ack-side race only. Does not actually close it — the refresh can still land after the nack.

Option 1 looks like the best cost/benefit, and keeps the existing cancellation semantics untouched.

## How this surfaced

Found in review of the enterprise re-vendor of #1592 (caura-ai/caura-enterprise#1825). Both the keeper and the cancel block are byte-identical between the two trees, and the executor is configured identically, so this applies equally to both — reporting it here rather than patching downstream so the fix re-vendors normally instead of creating divergence.

### LINKED ISSUE #1598: Lease refresh interval assumes a 30s ack deadline that durable subscriptions need not have
## Summary

`LEASE_REFRESH_INTERVAL_SECONDS = 15.0` is justified by a comment reasoning from a 30s initial ack deadline, but that 30s only holds for the ephemeral broadcast subscriptions this module creates itself. Durable subscriptions are provisioned externally, and Pub/Sub permits deadlines as low as 10s. On such a subscription the keeper's first refresh at t=15s arrives **after** the lease has already lapsed — so the exact defect #1592 fixes would still occur there.

## The assumption

From #1592, `common/events/pubsub.py`:

> Subscriptions are created with `ack_deadline_seconds: 30` and an existing subscription keeps whatever it was created with (see `_ensure_broadcast_subscription`), so raising that constant would not reach a single deployment already running.

That is accurate for `_ensure_broadcast_subscription`, which this module owns. Ordinary durable subscriptions are not created here, so nothing in the code constrains their deadline. The rest of the comment's arithmetic — "15s of slack against the initial 30s deadline on the first refresh" — silently inherits the 30s for subscriptions that never promised it.

The failure is quiet: below roughly `2 × LEASE_REFRESH_INTERVAL_SECONDS`, the keeper is running but always a step behind, so the batch redelivers exactly as it did before the fix.

## Not currently live in our own deployment

Checked before filing, so this is a robustness gap rather than an outage: across both Terraform environments every configured `ack_deadline_seconds` is **60 or 600**, and the module default is `optional(number, 60)`. Nothing is below 30 anywhere. But nothing *enforces* that either, and as an open-source component this file runs against subscriptions we do not provision.

## Suggested fix

**Issue the first extension immediately when the keeper is spawned**, before the first wait. Two lines, no extra RPC on idle loops (the keeper only spawns when a batch exists), and it removes the dependence on the initial deadline entirely rather than documenting an assumption that callers cannot see.

Alternatives, for completeness:

- Read the subscription's real `ack_deadline_seconds` once at pull-loop startup via `get_subscription` and derive the interval from it. Most correct, but adds an RPC and a failure mode to startup.
- Document the invariant and fail loudly at startup if a subscription's deadline is below the threshold. Turns a silent degradation into a visible one, but needs the same `get_subscription` call.

The first option gets most of the benefit for the least machinery: even on a 10s deadline, an immediate extension to `LEASE_EXTENSION_SECONDS` puts the lease comfortably ahead of the 15s refresh cadence.

## How this surfaced

Review of the enterprise re-vendor of #1592 (caura-ai/caura-enterprise#1825). The constants are byte-identical between the two trees, so it applies to both.

Eldad-Caura:
Closes #1597. Closes #1598.

Both live in the lease keeper #1592 added, both are small in effect, and both are in the same function — so they land together rather than as two PRs that would conflict on the same lines.

## #1597 — a leftover extension could invert a nack

`_pull_loop` cancelled the keeper before acking, and the comment there reasoned that a refresh already inside `run_in_executor` runs to completion on its thread regardless, which is harmless because it names ids the ack is retiring. That reasoning is correct — and it only covers the **ack**.

A nack is a `modify_ack_deadline` of **0**: redeliver now. The leftover refresh carries `LEASE_EXTENSION_SECONDS` for the *whole* leased batch, `nack_ids` included. Landing behind the nack, it returns the **failed** message to 60s of invisibility and silently inverts the one operation whose entire point is immediacy.

The ordering is genuinely open, which is what makes this real rather than theoretical: `_pull_executor` is a `ThreadPoolExecutor(max_workers=max(4, 2 * topic_count))`, so the leftover refresh and the nack run on different threads with nothing ordering them.

**Fix:** await the keeper instead of cancelling it. `stop` is already set, so it finishes at most the one refresh it is inside and returns — and because the keeper awaits its own `run_in_executor`, its task completing means that RPC has *returned*. #1592's objection to awaiting was about folding an outer cancellation into the keeper's own `CancelledError`; that applies to cancel-then-await and does not arise when the keeper is left to finish normally. `suppress(Exception)`, not `BaseException`, so `CancelledError` still propagates.

A lock was the obvious alternative and would **not** have worked: mutual exclusion orders nothing, so a nack and a refresh queued together could still take it in either order.

## #1598 — the first refresh could arrive after the lease had lapsed

The 15s interval was justified against *"subscriptions are created with `ack_deadline_seconds: 30`"*. True for the ephemeral subscriptions this module creates; durable ones are provisioned outside this codebase, and Pub/Sub permits as little as 10s. Against a 10s deadline the first refresh at 15s arrives after the lease is already gone, and the keeper then stays a step behind for the whole batch — running, issuing RPCs, fixing nothing.

**Fix:** `LEASE_FIRST_REFRESH_SECONDS = 5.0`, used for the first wait only. It clears the 10s floor with room for a slow RPC and still costs nothing in the common case — the keeper stops when the batch ends, so a batch draining in under 5s issues no extension at all. Refreshing immediately on spawn would also be correct and is simpler, but it puts an RPC on **every** non-empty pull, including the fast majority that need none.

## Verification

- **Each new test fails against its own fix reverted, and only its own.** Restoring the cancel fails the nack test; restoring the steady interval as the first delay fails the first-refresh test.
- **The nack test uses a real `ThreadPoolExecutor`, deliberately.** The entire defect is that cancelling does not stop work already running on a pool thread — which the inline `run_in_executor` stand-in the other lease tests use cannot express. Under that stand-in the test would pass either way.
- **The five existing lease tests needed updating, and that is worth flagging.** They patched only `LEASE_REFRESH_INTERVAL_SECONDS`, so the new first wait left them waiting the real 5s against their own 5s timeout: two failed outright and a third was a coin flip. They now patch both constants. This was a test-visible consequence of the timing change, not a behaviour change slipped past them — and it is the reason the constant is separate rather than folded into the existing one.
- `tests/test_events/` **156 passed**; `test_pubsub_bus.py` **81 passed**.
- `ruff check common/` and `ruff check tests/` clean; `ruff format --check tests/` 477 files already formatted. mypy: **zero errors in `common/events/pubsub.py`**; the 3 core-api reports are pre-existing in `src/core_api/mcp_server.py` and identical on `main`.

## Provenance

Both issues were found reviewing the enterprise re-vendor of #1592 (caura-ai/caura-enterprise#1825). The keeper is byte-identical between the two trees, so fixing here lets the improvement re-vendor normally instead of creating divergence in a file whose reconciliation depends on staying comparable.


Eldad-Caura:
@erni-a plear arppove

erni-a (approved):

PR #1602, opened 2026-09-18
fix(schema): stop autogenerate proposing to drop tables the models cannot see

Eldad-Caura:
Closes **08/14 L-25** and **09/02 L-13, L-14, L-46** — the models and the migration chain disagreed, and `alembic revision --autogenerate` turns disagreement into `DROP`.

## Measured first

I ran autogenerate against a correctly-migrated database before changing anything. It emitted **27 drop statements**:

```python
op.drop_table('recall_candidate')
op.drop_table('recall_event')
op.drop_table('tenant_suppression')
op.drop_index(op.f('ix_memories_embedding_hnsw'), ...)      # HNSW vector index
op.drop_index(op.f('ix_documents_embedding_hnsw'), ...)     # HNSW vector index
op.drop_index(op.f('ix_entities_name_embedding_hnsw'), ...) # HNSW vector index
op.drop_index(op.f('ix_memories_search_vector'), ...)       # GIN full-text
op.drop_index(op.f('uq_audit_log_tenant_seq'), ...)         # audit-chain uniqueness
... 19 more
```

Not one of those losses would fail a test. Search would quietly stop using an index and start scanning.

After this PR the same run emits **20**, and every remaining one is in a set this PR asserts exactly.

## Three separate gaps fed it

**`env.py` imported thirteen model modules by hand and omitted `recall_log`.** `recall_event` and `recall_candidate` are created by migration 027 and exist in every deployed database; autogenerate could not see them, so it proposed dropping them. It now imports the package — which makes the failure impossible to reintroduce by forgetting a line, since a new model registers by being exported from `common/models/__init__.py`, which is where a new model gets added anyway. `organization_settings` and `recall_log` were the two modules the package itself never exported.

**Seven indexes existed only in the models.** Every one came from `index=True` on a column, which generates SQLAlchemy's own `ix_<table>_<column>` name *beside* a differently-named `Index(...)` in `__table_args__` that the migration actually created — so the model described two indexes where the database had one.

| index | reality | fix |
|---|---|---|
| `ix_memory_conflicts_tenant_id` | duplicate of `ix_memory_conflicts_tenant` | drop the `index=True` |
| `ix_recall_event_tenant_id` | migration built `ix_recall_event_tenant_ts` (composite) | declare what exists |
| `ix_recall_candidate_recall_event_id` | migration built `ix_recall_candidate_event` | declare what exists |
| `ix_recall_candidate_memory_id` | migration built `ix_recall_candidate_memory` | declare what exists |
| `ix_documents_data` | GIN `jsonb_ops`, and **no query can use it** | delete |
| `ix_analysis_reports_tenant_started` | genuinely missing | **migration 049** |
| `ix_memory_derivations_tenant_id` | genuinely missing | **migration 049** |

`ix_documents_data` is deleted rather than built because every access to that column is `data['k'].astext` equality — which a `jsonb_ops` GIN index does not serve. There is no containment query anywhere in the tree. Building it would have cost a GIN index to answer nothing.

The two real gaps are worth having. `ix_analysis_reports_tenant_started` serves `report_get_latest_completed`'s tenant filter plus `ORDER BY started_at DESC LIMIT 1`. `ix_memory_derivations_tenant_id` serves the org hard-purge, which deletes from that table by `tenant_id` and had no index — while its sibling `memory_conflicts` has had `ix_memory_conflicts_tenant` all along.

**Migration 027 claimed RLS the schema does not have.** *"Both tables are RLS-scoped by `tenant_id` like every other tenant table."* There is no `ENABLE ROW LEVEL SECURITY` and no `CREATE POLICY` anywhere. A reader auditing tenant isolation would have taken that as a database-enforced boundary and stopped looking for the query-level one that actually does the work. `023_capability_usage` carried the same implication from the other side — *"RLS is intentionally NOT enabled on **this table**"* — and is corrected too.

## Why the tests use `compare_metadata`

A name match is not agreement, and this PR proved it on itself: migration 049 first built `ix_analysis_reports_tenant_started` **ASC**, where the model declares `started_at.desc()`. My name-based check called them identical. Alembic's `compare_metadata` reported a drop-and-rebuild, which is how the direction got caught. The tests use Alembic's comparison for that reason.

Two other things the tests caught about themselves:

- My first RLS test swept for any mention of "RLS" and failed on `023`, which mentions it precisely to say the schema does not use it — an accurate statement worth keeping. An assertion and a denial read almost alike to a text search, so the check now asks **the database** (`pg_class.relrowsecurity`, `pg_policies`) and matches only the affirmative phrase that was wrong.
- One intended mutation turned out not to be one: reverting `memory_derivations` to `index=True` produces the *same* index name migration 049 creates, so there was no drift to detect. That model edit is consistency, not a fix — the fix there is the migration. Said plainly rather than counted as a caught regression.

## Residual, pinned rather than left implicit

`tenant_suppression` has no ORM model at all (migration 019, read through raw SQL), and **17 partial/GIN/HNSW indexes are expressed only in migrations**. Both are asserted as exact sets, so an eighteenth fails the test instead of joining a crowd nobody is counting. Transcribing seventeen load-bearing index predicates into the models is a change with its own risk profile and belongs in its own PR.

## Tests

5 new, revert-checked against 5 mutations:

| mutation | fails |
|---|---|
| `recall_log` un-registered again | the table-visibility test (and the residual set shifts) |
| database rolled back to 048 | the model-index test — both missing indexes |
| duplicate tenant index back on `memory_conflicts` | the model-index test |
| unusable GIN index back on `documents.data` | the model-index test |
| 027 says "RLS-scoped" again | the phrase test |

## Gates

`core-storage-api/tests/` 491 passed · `core-operations/tests/` 81 passed · full root suite 7168 passed / **31 failed, identical to the pre-existing baseline** (the name-level diff against the pre-fix run shows exactly one change: `test_single_head` resolved after bumping its pinned head to `049`, which that test's own docstring asks you to do by hand). One intermediate run reported 32 without captured names and did not reproduce; treating it as a flake rather than claiming three clean runs.

`mypy src/` clean for core-storage-api (94 files) and core-operations; core-api and core-worker at their known 2-error `common/llm/providers/openai.py` httpx baseline. `ruff check` and `ruff format --check` clean across the full workflow scope.


Eldad-Caura:
CI is green across all 12 checks and the review came back with no issues, so this is ready for a human pass.

@erni-a the one thing worth your eyes is the residual allowlist in `core-storage-api/tests/test_models_match_the_migrated_schema.py`. After this PR, autogenerate still emits 20 statements against a correctly-migrated database — down from 27, and every one of them is now pinned by name in `_TABLES_WITHOUT_MODELS` (`tenant_suppression`) or `_INDEXES_ONLY_IN_MIGRATIONS` (17 entries). The point of pinning them as exact sets rather than a count is that an 18th drift fails the test instead of joining a crowd nobody is counting. Transcribing those 17 index predicates into the models is real work and belongs in its own PR — I did not attempt it here.

Two notes on what this PR does *not* claim:

- The `memory_conflicts` / `memory_derivations` `index=True` removals are consistency edits, not bug fixes. Reverting `memory_derivations` to `index=True` generates the same index name that migration 049 creates, so there was no drift to catch there — I checked, and it is worth saying plainly.
- Migration 049 is the part with production effect: `ix_memory_derivations_tenant_id` did not exist, and the org purge deletes from that table by `tenant_id`. Its sibling `memory_conflicts` has had that index all along. Both new indexes are `CREATE INDEX CONCURRENTLY IF NOT EXISTS` inside an `autocommit_block`, so they are re-runnable and do not take a write lock.

Not merging — that is yours.

erni-a (approved):

PR #1603, opened 2026-09-18
fix(dead-code): delete three subsystems that ship and cannot run

Eldad-Caura:
Four findings, one shape: a complete, plausible implementation with no path from a running process to it. None of them failed a test, because tests were the only callers.

## What was unreachable, and how I checked

**The pluggable-backend subsystem** (`oss-0814-m-48`, `oss-0902-m-20`). `SqliteBackend` (461 lines), `InProcessQueue`, `ConfigIdentity`, `ManualResolver`, and the four factories that build them. Every reference outside the package itself is in `tests/test_provider_protocols.py` — and unlike `get_stm_backend`, which `settings.stm_backend` really does select between two live backends, there is no `storage_backend` / `job_queue` / `identity_resolver` / `conflict_resolver` name anywhere in `core_api.config`. The only way to reach one was to call the factory by hand.

It had also already drifted: `SqliteBackend` carried its own `CREATE TABLE memories` DDL that no migration maintains. A second schema definition for the central table, kept current by nobody.

The four protocols and their four supporting dataclasses go with it. `STMBackend` stays.

**The payment package** (`oss-0902-m-54`). 102 lines plus eight settings — seven `PADDLE_*` and `payment_provider` — that nothing read. The live implementation is in the enterprise repo as `platform_admin_api.services.payment`, a fork of this same code under a different package name; that is what `platform-admin-api`'s compose block feeds `PADDLE_*` into, and it never imports the OSS copy.

Worth stating plainly, because the finding pairs dead code with a replay gap: **the enterprise copy already verifies timestamp freshness** (signature first, then age). The unlimited-replay window lives only in this dead copy, so deleting it retires the last carrier of a bug that was already fixed where it mattered. Nothing here closes a live hole, and I would rather say so than let the deletion imply otherwise.

**`auth.hash_key` and `auth.enforce_org_admin`** (`oss-0814-l-45`). Both dead — but the finding's title says the org-admin gate "is enforced nowhere," and **that half is wrong**. It is enforced, at five skills-inbox routes. What happened is stranger than absence: the predicate `is_admin or org_role == "admin"` is written out inline three times (`skills_inbox` twice, `documents` once), each as `bool(getattr(auth, "is_admin", False)) or (getattr(auth, "org_role", None) == "admin")`, while the shared raiser that says exactly that sat unused.

So the fix is not only deletion. The three copies now share `AuthContext.is_org_admin`, and the raiser nobody reached for is gone. `_require_inbox_admin` keeps raising `SKILLS_INBOX_FORBIDDEN` — no error code a caller can still reach changed. The `getattr` spelling went too: `auth` is a real `AuthContext` at all three sites, and the defensive form hid that.

## What the new tests ask

`tests/test_shipped_code_is_reachable.py` asks the question these subsystems' own tests could not ask about themselves: not "does this work?" but "can anything running reach it?"

The settings scan is the one worth reading. After this change, **exactly three of core-api's 92 declared settings are read by nothing**, and they are the `crystallizer_*` knobs — `CRYSTALLIZER_ENABLED` among them, a flag that can neither enable nor disable anything. That is open finding `oss-0814-m-47`, and it is pinned by name in the test rather than fixed here: whether the crystallizer should honour those knobs or has outgrown them is a decision about the crystallizer, not about dead code. The fourth knob in the same block, `crystallizer_dedup_threshold`, *is* read — which is why the three are named individually instead of waving at the prefix.

Two things I got wrong while writing these, both caught by the assertions themselves:

- The settings scan passed while proving nothing, because the test file names the settings it expects to find unread, and `tests/` is in the scanned roots — so the allowlist made its own entries look read. The scan now excludes itself.
- The org-admin scan was a regex and matched the two comments that *quote* the predicate to explain it. Prose describing a defect read as the defect. It compares ASTs now.

Each assertion was mutation-tested: reintroducing the inline predicate, re-adding an unread setting, re-declaring `aiosqlite`, importing a deleted module, and re-exporting an unselected factory each fail the intended test and only that test.

## Gates

`ruff check` and `ruff format --check` clean across the CI scope. `mypy src/` on core-api: 2 errors, the known pre-existing httpx conflict in `common/llm/providers/openai.py`, over 197 files (was 205 — the 8 deleted). Root suite **31 failed, 7140 passed** — the 31 are the standing baseline, unchanged.

Rebased onto `main` after #1602 merged; re-ran lint and the touched suites after the rebase.

Closes `oss-0814-m-48`, `oss-0902-m-20`, `oss-0902-m-54`, `oss-0814-l-45`.


Eldad-Caura:
@erni-a pleras sapprovd

erni-a (approved):

PR #1604, opened 2026-09-18
fix(lifecycle): unpair entity-link from crystallize, and give cross-link a budget it can name

Eldad-Caura:
Fixes the chain behind the 2026-09-18 staging outage, where one entity-link message blew past 120s ten times, dead-lettered, and then failed **every staging deploy for ~17 hours** — a DLQ message is durable and the smoke gate requires a depth of zero.

Three changes, one per link in that chain.

## 1. Entity-link gets its own hour

`lifecycle-crystallize` and `lifecycle-entity-link` both read `lifecycle_pipeline_run_at_hour`, so the two heaviest sweeps fired on the same minute and **no operator setting could separate them**. caura-enterprise's stagger (2026-09-16) removed the six-way 02:00 pile-up but deliberately left this pair together, calling them "the heaviest pair" — there was no second knob to split them with.

New `lifecycle_entity_link_run_at_hour`. It defaults to `2` like its siblings rather than tracking the pipeline hour: a default that *followed* the other setting would silently re-pair them for anyone who moved the pipeline hour, which is precisely the failure this exists to make impossible. The enterprise deploy moves it to 06.

Also added to the `0..23` validator — that list is explicit, so a new field that isn't named in it silently accepts hour 25.

## 2. Cross-link discovery gets a budget that wins

`POST /entities/discover-cross-links` had **no application-level cap**. Its only limit was the writer's 120s Cloud Run request timeout — *equal* to the storage client's 120s httpx read. The client's own docstring already names this shape:

> Equal values would 50/50 race.

Whichever fired first, the caller learned nothing. The dead-lettered failure recorded `failures=["ReadTimeout('')"]` — an empty string where the reason belongs, ten times over.

`cross_link_request_timeout_seconds` (100s) sits below **both** ceilings, so the step's own cancellation wins and the failure names the tenant and the budget it exceeded. Same shape as the bulk route's `asyncio.wait_for` (CAURA-602), which exists to settle exactly this race.

**This does not make slow runs succeed — it makes them legible.** Changes 1 and 3 are what buy headroom.

### The budget is enforced, not just documented — review finding

Verified and correct: `cross_link_request_timeout_seconds` was env-overridable with nothing checking it against the ceilings it has to beat, unlike `interview_request_timeout_seconds`, which `_validate_timeout_ordering` has guarded since it was added. An operator debugging this very timeout reaches for "raise the timeout", and at or above 120 that change is *worse than a no-op*: the step stops cancelling first and the failure reverts to the opaque `ReadTimeout('')` this PR exists to eliminate.

Two things the review prompt left open, both load-bearing:

**`>=`, not the `>` the interview check uses.** That budget only has to *fit* under the ceiling; this one has to *win* against it. At equality the two timers race — and that race **is** the defect: 120 against 120 is exactly what failed staging. Weakening the comparison to `>` lets the 120.0 case straight through, which is why the test parametrises both.

**The httpx read ceiling is now named** (`STORAGE_READ_TIMEOUT_SECONDS`), so the check is `min(platform, storage_read)` rather than the platform ceiling alone. Both are 120.0 today and that is a *coincidence*: one is this process's client, the other is what Cloud Run/nginx sever inbound. Checking only the platform value would let someone raise it to 180, set this budget to 170, and hand httpx the race back — the same misconfiguration in new clothes.

## 3. Half the per-run cross-link batch

`LifecycleAudit.entity_link` runs the pipeline **once** per org per tick with no batch loop, so `CROSS_LINK_MEMORY_BATCH_SIZE` is the entire per-run budget: 200 candidates × a per-candidate ANN search over the tenant's entities. Now 100. The other consumer (`_discover_cross_links_for_memory`) passes a single `target_memory_ids`, so the cap never binds there. The cost is convergence, not coverage — a tenant with more under-linked memories needs more nightly ticks.

## Verification

- **Both new tests fail against their own fix reverted, and only their own.**
- **The budget test carries an outer `wait_for`, and that matters**: without the fix there is nothing to cancel the stalled call, so the test would *hang the suite* rather than fail. I found that by running the mutation, not by reasoning about it — the first attempt hung.
- **`test_run_at_hour_override_is_threaded_through` asserted entity-link follows the pipeline hour**, so change 1 necessarily breaks it. Rewritten to cover crystallize, with the new test pinning both directions of the split (moving one knob must not move the other). A test-visible consequence of an intended change, not something slipped past it.
- **Removing the new guard fails both parametrised cases and nothing else. Weakening `>=` to `>` fails only the 120.0 case** — the operator choice is proven, not asserted.
- core-operations **89 passed**; `tests/pipeline/test_entity_linking.py` **9 passed**; `test_storage_client_pool_recovery.py` passes with the renamed constant. `test_health_gate.py`'s 1 failure + 8 errors are the same MCP SDK mismatch, byte-identical on a clean `origin/main` checkout.
- `ruff check` + `ruff format --check` clean on `core-api/src`, `core-operations/src`, `core-operations/tests`, `tests`. mypy clean on core-operations and on every file this touches; core-api's 3 errors are pre-existing in `mcp_server.py` and identical on `main` — a local MCP SDK mismatch that also explains 43 unrelated collection errors there.

## Rollout

Land this first. `Settings` uses `extra="ignore"`, so the enterprise PR setting `LIFECYCLE_ENTITY_LINK_RUN_AT_HOUR=6` before this ships is harmless — but it would be silently ignored, so this wants to go first for the var to mean anything.


Eldad-Caura:
@claude re-review please — pushed `693f8bee`.

Your Medium was right and is fixed: `cross_link_request_timeout_seconds` now has a startup guard in `_validate_timeout_ordering`, alongside the interview one.

Two deliberate departures from the prompt, both load-bearing:

- **`>=`, not `>`.** The interview budget only has to *fit* under its ceiling; this one has to *win* against it, so equality is a 50/50 race — and 120-against-120 is precisely what failed staging. The test parametrises `120.0` and `150.0`; weakening to `>` fails only the `120.0` case.
- **The httpx read timeout is now a named constant** (`STORAGE_READ_TIMEOUT_SECONDS`), so the check is `min(platform, storage_read)`. You flagged it as optional-if-named; it was hardcoded `read=120.0`, and naming it is what stops someone raising the platform ceiling to 180, setting this to 170, and handing httpx the race back. Only `read` is named — `write` shares the value but carries no ordering contract.

Eldad-Caura:
@claude re-review please — head is `90d0d641`.

The rebase onto current `main` was a `synchronize` push, so the review job skipped silently and there is no verdict on this SHA. The change itself is byte-identical to `5413db14`, which you last reviewed — that round's Low is fixed: the timeout failure no longer claims "no links landed", since cancelling `wait_for` cannot roll back a transaction that may already have committed.

Worth checking against the storage side rather than the wording alone: `entity_discover_cross_links` folds candidate-find, the pgvector pass and the insert into **one** `get_session()` transaction, and the insert is `on_conflict_do_nothing(index_elements=["memory_id", "entity_id"])`. So the message now says the batch is all-or-nothing and that re-running is safe, rather than hedging to "may or may not" and stopping there. A test asserts the string never contains "no links landed" again.

Eldad-Caura:
@erni-a plrasr pproev

erni-a (approved):

PR #1605, opened 2026-09-18
fix(dead-code): remove declarations that describe a contract the code lacks

Eldad-Caura:
Two findings, one shape: a declaration that describes a contract the code does not have. Neither could fail a test — a declaration nothing reads has no behaviour to get wrong.

## `Settings.host` / `Settings.port` (`oss-0814-l-52`)

Neither core-storage-api nor core-worker ever reads them. Both take their address from the Dockerfile CMD: `python -m common.serve … --host 0.0.0.0 --port 8002` for storage, `uvicorn … --port 8080` for the worker. `common.serve` imports the settings singleton only to configure logging; `--host`/`--port` are argparse arguments.

The worker's field carried the comment `# Cloud Run port.` — and Cloud Run *does* set `PORT`, so pydantic-settings bound it to a field nothing consulted while uvicorn bound the literal 8080 beside it. Today that costs nothing, because no reader exists to disagree. It is the kind of value that looks authoritative and answers no question.

The finding names only core-storage-api. core-worker has the identical field and I found it while verifying, so it goes too — calling that out explicitly since it is beyond what was filed.

## The seven abandoned request schemas (`oss-0902-l-49`)

All seven are unreferenced — each appears exactly once in the tree, at its own `class` statement. But the interesting part is what happened after they were abandoned. I diffed each against the handler for the endpoint it names:

| schema | endpoint | keys the handler reads that the schema omits | keys the schema declares that nothing reads |
|---|---|---|---|
| `ScoredSearchRequest` | `/scored-search` | `readable_tenant_ids`, `date_range_start`, `date_range_end`, `history_query`, `temporal_window_days` | — |
| `SemanticDuplicateRequest` | `/semantic-duplicate` | `agent_id`, `exclude_id`, `min_similarity`, `visibility` | `content_hash`, `threshold` |
| `NearDuplicatesRequest` | `/near-duplicates` | `batch_size`, `neighbor_limit`, `offset`, `threshold` | — |
| `MemoryEntityLinksRequest` | `/entity-links` | `memory_ids` | `entity_ids` |
| `EntityFTSRequest` | `/fts-search` | `strict_fleet_scoping` | — |
| `GraphExpandRequest` | `/expand-graph` | `seed_entity_ids` | `seed_ids` |
| `ContradictionCandidatesRequest` | — | — | names an endpoint that does not exist |

`MemoryEntityLinksRequest` is the clearest: it has one field, and that field is the wrong name. `ScoredSearchRequest` is the dangerous one: `readable_tenant_ids` is how a caller is granted reads beyond its home tenant.

**This is why they are deleted rather than corrected.** A reader who found these and did the obvious, tidy thing — annotate the handler with the schema it looks written for — would have dropped every key in the middle column. Silently: pydantic ignores unknown fields by default, so the request would still succeed and cross-tenant read scoping would just stop arriving. A stale schema that looks like the contract is worse than no schema at all, because the no-schema case makes you read the handler.

## What the tests pin

`tests/test_declarations_are_not_stale.py`:

- **Both halves of the address claim**, so neither can rot alone: no `Settings` declares `host`/`port`, *and* the Dockerfiles still carry `--port`. Asserting only the first would leave "the CMD is what binds it" as an unchecked story in a commit message.
- **No unreferenced request model in `schemas.py`** — with a vacuity check that the BaseModel detector finds models in a control module that has them. That check earned its keep immediately: my first control module (`skill_factory.py`) turned out to have no `BaseModel` subclasses at all, so the test was passing on an empty set against an empty set.
- **The three scoping keys are still read** by `/scored-search`, `/fts-search` and `/entity-links` — the facts that made those schemas wrong, recorded where a future reviewer will find them.

Each assertion mutation-tested: restoring the worker's `port`, re-adding an unreferenced model, and making `/scored-search` stop reading `readable_tenant_ids` each fail the intended test and only that one.

## Gates

`ruff check` and `ruff format --check` clean across the CI scope. mypy: core-storage-api **clean, 94 files**; core-worker 2 errors, the standing httpx conflict in `common/llm/providers/openai.py`. Suites: storage **491 passed**, core-worker **100 passed**, root **31 failed / 7180 passed** — the 31 are the unchanged baseline.

Rebased onto `main` after #1603 merged; lint and both new test files re-run after the rebase.

Closes `oss-0814-l-52`, `oss-0902-l-49`.


Eldad-Caura:
All 12 checks pass and the review came back with no issues, so this is ready for a human pass.

@erni-a the part worth your eyes is the table in the description rather than the diff, because the diff is only deletions and the reason for them is not visible in it.

Six of the seven removed schemas are not merely unreferenced — they disagree with the endpoint they name. `MemoryEntityLinksRequest` has exactly one field and it is the wrong one (`entity_ids`, where `/entity-links` reads `memory_ids`). `ScoredSearchRequest` is missing five keys the handler reads, `readable_tenant_ids` among them, which is how a caller is granted reads beyond its home tenant.

That is the argument for deleting rather than fixing them. Someone tidying this file would reasonably annotate the handler with the schema that looks written for it — and pydantic drops unknown fields by default, so the request would keep succeeding while those keys stopped arriving. No error, no 422, just narrower reads. `tests/test_declarations_are_not_stale.py` records the three keys that matter so the next person has the receipts.

Two notes on scope:

- The `Settings.host`/`port` removal extends past what `oss-0814-l-52` filed: it names core-storage-api, and core-worker has the identical dead field. I took both. The worker's carried the comment `# Cloud Run port.`, and Cloud Run does set `PORT` — so pydantic bound it to a field nothing reads while uvicorn bound the literal 8080 next to it. Harmless today precisely because no reader exists to disagree.
- The test asserts both halves of that claim — no `Settings` declares host/port, **and** the Dockerfiles still carry `--port`. Asserting only the first would leave "the CMD is what binds the address" as a story in a commit message with nothing checking it.

Not merging — that is yours.

erni-a (approved):

PR #1607, opened 2026-09-18
fix(events): log the lease keeper's escapes instead of suppressing them

Eldad-Caura:
## What this fixes

#1601 changed the drain before ack/nack from `lease_keeper.cancel()` to
`with contextlib.suppress(Exception): await lease_keeper`. **Awaiting was the fix and
stands.** The bare suppress was not, and it cost something.

`_hold_leases` already catches and logs its own RPC failures, so nothing ordinary
reaches this line. What reaches it is a **bug in the keeper** — and that is precisely
what the suppress discarded, with no log, no metric, no trace of any kind.

Awaiting is what made the log necessary. While the task was only cancelled and never
awaited, an exception it had *already* raised stayed unretrieved, and asyncio surfaced
it at GC as `Task exception was never retrieved`. Late, untargeted, and only in that one
ordering — but not nothing. Retrieving the exception here ends that route, so this has
to take the job over and do it properly, naming the subscription.

`except Exception`, not `BaseException`: `CancelledError` is a `BaseException`, so an
outer `stop()` cancelling the pull task still propagates. The `finally` backstop keeps
its `cancel()` — it covers the paths that never reach the drain, and an un-awaited
cancel there is still the right shape.

## Provenance

Found by the enterprise reviewer on the re-vendor of #1601
(caura-ai/caura-enterprise#1827). Fixing it **here** rather than there keeps the two
copies of this function comparable, which is what makes the next reconciliation a digest
check instead of a re-read.

## Verification

- **The new test fails against the fix reverted, and nothing else does.** Restoring the
  bare suppress fails it alone; the other 81 tests in `test_pubsub_bus.py` are unaffected.
- **The test also asserts the batch is still acked.** Logging the escape is half the
  requirement — a keeper bug must not additionally cost the batch its acknowledgement
  and send finished work round again.
- **It patches `_dispatch_all`, and that is load-bearing rather than copied from its
  neighbours.** `_lease_harness` sets `_stopping` during the pull, and the loop's own
  `except Exception` *returns silently* while stopping. A dispatch that raised would
  therefore look exactly like a keeper escape that went unlogged — the failure under
  test. I found that by instrumenting a test that was failing for the wrong reason, not
  by reading it.
- `tests/test_events/` **157 passed**.
- `ruff check common/ tests/` clean · `ruff format --check tests/` 479 files already
  formatted · mypy via `cd core-api && mypy src/`: the same 3 pre-existing
  `mcp_server.py` errors as on `main`, nothing new.


Eldad-Caura:
@erni-a palr approve

erni-a (approved):

PR #1608, opened 2026-09-18
fix(migrations): stop reporting work that was never done

Eldad-Caura:
Two places where green means "recorded as done" rather than "done". Neither can fail a test, because the record is what later runs consult.

## `CREATE INDEX CONCURRENTLY IF NOT EXISTS` is not idempotent on its own

An interrupted build leaves the index in `pg_index` with `indisvalid = false`. `IF NOT EXISTS` then *skips* it — "already exists, skipping" — the planner refuses to use it, and every re-run takes the same skip. The migration reports success over an index that is permanently useless and still charges write overhead on every insert.

**I verified this against a real database rather than asserting it.** On a throwaway copy of the schema, marking `ix_memories_run_id` invalid and re-running migration 017:

| | result |
|---|---|
| pre-fix 017 | `indisvalid = f` — still broken, migration reported success |
| post-fix 017 | `indisvalid = t` — repaired |

Eight migrations already handled this, as copies of one query spread across 005 / 007 / 026 / 035 / 040 / 041 / 046 / 048. Having no shared spelling is exactly why the other seven never got one, so the fix is to extract it: `drop_invalid_indexes()` in `core_storage_api.database.migration_helpers`, called by **009, 011, 016, 017, 024, 033 and 049**. The full chain 001→049 runs clean on a fresh database.

**049 is mine, from #1602.** I described its two statements as re-runnable in that PR's handoff. They do not error on re-run, which is not the same thing, and I had not read this finding before writing them.

The finding named six migrations; I measured seven, the seventh being 049. It also excluded 012 as self-healing, and my independent scan agrees — 012, 037 and 045 drop before they create.

## `init_database` stamped the whole chain on the strength of one table

The old decision, in full:

```python
if has_tables and not has_alembic:   # has_tables == "a memories table exists"
    command.stamp(alembic_cfg, "head")
```

Stamping asserts that every migration through head has already been applied. A schema built by `Base.metadata.create_all` — which is what `tests/conftest.py` uses — satisfies that test while missing everything migrations add that models do not declare: the 17 partial/GIN/HNSW indexes and `tenant_suppression` itself, both measured in #1602. And the error is permanent in the quiet direction: once stamped, every future `upgrade head` is a no-op, so nothing ever repairs it and nothing ever says so. Search stops using an index and starts scanning.

The decision is now a pure function over three probes, so it can be read and tested without a database:

```python
schema_bootstrap_action(has_tables=..., has_alembic_version=..., has_chain_evidence=...)
    -> "upgrade" | "stamp" | "refuse"
```

`has_chain_evidence` is the presence of `tenant_suppression` — created in raw SQL by migration 019, declared by no ORM model, so present in a chain-built database and absent from a `create_all` one. With it, stamping records what the schema already shows. Without it, the service refuses and names the two deliberate ways forward. `CONTRIBUTING.md:70-80` has documented this hazard for developers all along; documentation was the only thing between an operator and a permanently under-provisioned database.

A test asserts the sentinel is still migration-only, because if a model ever declares that table the discriminator silently stops discriminating.

## On the scan in the test

It resolves calls through the AST rather than grepping, and that is not fastidiousness — grepping got this wrong three times while I was writing it:

- counted `CREATE INDEX CONCURRENTLY` inside a **comment** as a build (005, 007, 026);
- read `{_INDEX_NAME}` f-strings as an index literally named `ON` (045, 048);
- missed the cleanup in 035 and 040 entirely, because their helper is defined outside `upgrade()`.

Each wrong answer looked plausible, and two of them had me about to "fix" migrations that were already correct. The repo's other guard here, `test_no_plain_create_index_on_large_tables`, is a regex — and 035/040/048 carry comments explaining the contortions needed to stay inside its coverage, while 007 and 026 do the right thing and are invisible to it.

Two things I checked and am **not** claiming:

- Replaying migrations over an already-complete schema fails at 018 (`CREATE TABLE dedup_reviews`, no `IF NOT EXISTS`). That is pre-existing and unrelated — migrations are not designed to replay — and I only hit it constructing the test scenario above.
- The `memory_conflicts` style consistency edits are not part of this; nothing here changes what any query does.

## Gates

`ruff check` / `ruff format --check` clean across the CI scope. `mypy src/` on core-storage-api: **clean, 95 files**. Storage suite **520 passed** (491 + 29 new). Root suite **31 failed / 7142 passed** — the standing baseline, unchanged.

Every new assertion mutation-tested: reverting 049's cleanup, pointing the sentinel at a table that has a model, and restoring the stamp-on-sight decision each fail the intended tests. Throwaway probe databases were created on the test container's port and dropped; nothing else was touched.

Rebased onto `main` after #1606; lint and the storage suite re-run after the rebase.

Closes `oss-0902-m-57`, `oss-0902-m-58`.


Eldad-Caura:
Applied — the finding is correct, and it is the same defect this PR is about.

`init_database`'s docstring still said "If tables already exist (e.g., created by the legacy backend), stamps the current revision so Alembic skips the initial migration." That describes the branch I replaced. A reader would have taken it as the deploy behaviour against a `create_all`-built database, which is now exactly the case that refuses. I changed the code and left the prose asserting the property the code no longer has — the thing this PR removes from migrations, in the docstring of the function doing the removing.

The docstring now describes all three outcomes, and says what the old behaviour was and why it went.

I also coupled it mechanically rather than leaving it to the next reviewer: `test_the_bootstrap_docstring_names_every_action_it_can_take` enumerates every value `schema_bootstrap_action` can return over all eight probe combinations and asserts each name appears in the docstring. Adding or renaming an action now forces the description to be revisited. Mutation-tested — restoring the old one-line docstring fails it.

Gates after the amend: `ruff check` / `format --check` clean, `mypy src/` clean over 95 files, storage suite **521 passed**.

@erni-a one thing worth a human eye, since it is a deploy-path behaviour change rather than a refactor: this PR makes `init_database` **refuse to start** when a database has a `memories` table, no `alembic_version`, and no `tenant_suppression`. That is deliberate — stamping such a database certifies migrations that never ran, and the mistake is permanent because every later `upgrade head` is then a no-op. But it does mean a deployment that was previously booting on a silently under-provisioned schema will now fail loudly instead, with a message naming the two deliberate ways forward. I believe loud is right here; you may want to confirm no real deployment is in that state before merging.

erni-a (approved):

PR #1610, opened 2026-09-18
fix(scripts): repair developer tooling that cannot run as shipped

Eldad-Caura:
Six operator-facing files that no suite exercises. That is why none of these was noticed: they are tools, not tests, so "broken" and "working" look identical from CI.

## The two benchmarks could not reach the API at all

`latency_test.py` and `hyperagent_test.py` both built their base URL as `{url}/api`. Every router is mounted under `/api/v1` (`app.py`), so each 404'd on its first call — the health probe — and exited. Against its own documented default target, `http://localhost:8000`.

Two more dead calls came out of checking the rest of what they address against the OpenAPI spec:

- **`hyperagent_test.py` posted to `/admin/keys`.** That route exists nowhere — not in this repo, not in the enterprise one. The route that mints tenant-scoped keys is `/admin/agent-keys/provision`, the one core-api's own MCP error messages name, and it is served by the **enterprise admin API**. So this step 404s against stock OSS by design: MCP refuses an admin key, and OSS has no way to mint a tenant-scoped one. That is a real limitation of the benchmark rather than a bug to paper over, so the script now says it in those words instead of printing a bare status code.
- **`latency_test.py` finished by deleting `/admin/tenants/{tenant}`.** There is no tenant-delete route in core-api and never has been — `DELETE /api/v1/memories?tenant_id=…` is the cleanup, and it was already there on the line above. The 404 went unread because the response was discarded, so every run left its benchmark tenant behind. A failed cleanup is now reported.

## An assertion that could not pass

`gateway_integration_test.py` posts a heartbeat carrying a list of nine tools and `tools_registered: 9`, then asserts the persisted value equals **13**. No server behaviour makes that check pass. Both now derive from one list, so the body and the assertion cannot disagree again.

## A backup whose filename misdescribes its contents

`backup-db.sh` piped a `pg_dump -Fc` through `gzip` and named the result `.sql.gz`. Custom format is already compressed, so the gzip is pure waste — but the name is the real problem: it promises something `gunzip | psql` can restore, and what is in there is a binary archive only `pg_restore` reads. The name misdirects the one operation a backup exists for.

It also wrote straight to the final path. `set -e` aborts on a failed dump, but the redirect creates the file the moment it opens, so a dump killed halfway left a truncated file in the backup directory under a name indistinguishable from a good one — which the retention sweep then kept for seven days. Now: `.dump`, written to a `.partial` name and renamed only on success, with a `trap` cleaning up the partial, and the restore command printed. The retention `find` matches both the new and old names, so backups written by the previous version are still reaped.

## Two pieces of config addressed to nobody

- `docker-compose.dev.yml` bind-mounted `./migrations`, which does not exist in the repo. Compose creates an empty directory in the developer's working tree for a missing bind source, so this silently mounted nothing over nothing — on **core-api**, which has no migrations at all (they belong to core-storage-api).
- `env.dev`'s six `ALLOYDB_*` lines are fed only to core-api, in both compose files. core-api's settings no longer contain any database fields — `config.py` says so outright: the `postgres_*` / `db_pool_*` / `database_url` settings "were removed with the engine", because it talks to core-storage-api over HTTP. `extra="ignore"` swallowed all six, password included (a `changeme` placeholder, not a live credential).

## What the test checks

`tests/test_scripts_call_routes_that_exist.py` extracts every path these scripts build and checks it against **core-api's OpenAPI spec**, not a list maintained beside it — so it follows the app rather than needing to be told when the app changes. It would have caught all three defects above in one run.

The single path deliberately not served here is named with its reason rather than silently skipped, which makes the enterprise dependency reviewable instead of invisible.

Vacuity-checked both ways: the test asserts the spec has >50 versioned paths and that the scan found calls in every script, because an empty spec or a regex that stopped matching would otherwise pass everything.

Mutation-tested: reverting the `/api/v1` prefix and pointing a script at a nonexistent route each fail the intended assertion.

## Gates

`ruff check` and `ruff format --check` clean across the CI scope. `docker compose -f docker-compose.dev.yml config` parses. `bash -n` and `shellcheck` on `backup-db.sh` clean (one pre-existing SC1091 info about sourcing `.env`). Root suite **31 failed / 7157 passed** — the standing baseline, unchanged.

Closes `oss-0902-l-05`, `oss-0902-l-06`, `oss-0902-l-60`, `oss-0902-l-61`, `oss-0814-l-32`, `oss-0902-l-57`.


Eldad-Caura:
Legacy-name ratchet was right to fail this, and the gate's message told me exactly which of its four markers applied.

The line it caught is the retention sweep:

```
find "$BACKUP_DIR" \( -name "caura_*.dump" -o -name "memclaw_*.sql.gz" \) -mtime +7 -delete
```

The file's total count actually *fell* (4 → 3), but the text on that line is new, which is the thing the gate checks — correctly, since a falling total would otherwise let a new mention in behind an unrelated removal.

Marked `legacy-name-floor`, which the gate reserves for a literal the rename will never reach. That claim holds here and I checked it the way the message asks: the pattern matches files that already exist on operators' disks, written by the previous version of this script. Rewording it would not leave the line correct — it would simply stop matching them, and those backups would accumulate forever. So the literal is load-bearing, nothing is declared by it, and the marker states that in the diff.

I also had to fix my own first attempt. The explanatory comment I wrote above the `find` spelled the old name out too, which would have tripped the same gate on lines where the literal is *not* load-bearing — prose can always be reworded. The comment now says `*.sql.gz` and the branded literal appears once, on the line that needs it.

Verified locally with the gate itself rather than by eye:

```
Gate passes: no new lines currently fail it.
Range history: 1 added, 1 annotated, 2 removed, 0 excused moves (-2 net).
```

Net −2: this PR removes more of the old name than it adds.

Re-ran after the amend: `ruff check` / `format --check` clean, `bash -n` and `shellcheck` clean on the script (one pre-existing SC1091 about sourcing `.env`), and the new route-existence test still passes.

Eldad-Caura:
@erni-a plerar aoorive

erni-a (approved):

PR #1611, opened 2026-09-18
fix(events): close the gaps between what the bus promises and what it does

Eldad-Caura:
Three findings, one shape: the bus is described in one place and behaves in another, and the two disagreed.

## The only schema in `common/events` that forbids extra fields

`MemoryEmbedRequest` carried `extra="forbid"`. The other five wire schemas all carry `extra="ignore"`, and `MemoryEnrichRequest` — the direct sibling, same publisher→consumer direction — spells out why:

> `forbid` would silently ack-drop additive fields during a worker-first deploy.

The mechanism is real and I traced it: `core_worker.consumer.handle_embed_request` catches `ValidationError` and returns, **acking** the message. That is deliberate — the comment says "ack-drop so the subscription doesn't loop on a poison message" — and it is the right call for a genuinely malformed payload. But it means that during a rolling deploy where core-api ships first with any new optional key, every embed request for the length of the deploy window is acknowledged and thrown away. Those memories keep a NULL embedding, and nothing retries them: the request event was the retry.

5-against-1 is what settles it. This was the last schema not following a policy the repo states three times.

## A subscription that fell between two mechanisms

`events_manifest.json` — read by caura-enterprise's `check_pubsub_provisioning.py` to require a Terraform subscription — omitted `caura.org.suppression-changed`, which core-worker really does consume.

The generator has **two** ways of knowing about a subscription, and this one fit neither:

- Topics registered by shared helpers are captured *dynamically*, by invoking the helper against a recording bus. But `_capture_helper_topics` patched `get_event_bus` on `lifecycle_handlers` specifically, so for any other helper module the patch landed on the wrong module and the real bus would have been called. The mechanism existed and was unavailable.
- `test_direct_subscribes_match_consumer_files` covers `bus.subscribe` written directly in a service's consumer file. core-worker calls `register_suppression_consumer(...)`, so there is no direct subscribe to find.

A shared registrar in a module the generator didn't name was therefore invisible to both. The patch target is now resolved from the registrar's own `__module__`, `suppression_handlers` is registered, and regenerating adds exactly one line. A consumed topic with no provisioned subscription is the failure that took staging down when `insights-requested` shipped — the generator's own docstring says so.

## Two claims about delivery, in opposite directions, neither true

- `base.py` said raising "propagates to the bus (in-process buses **re-raise in tests**…)". No such mechanism has ever existed; `_safe_invoke` swallows unconditionally.
- `inprocess.py` called that swallow a match for "**Pub/Sub-style fire-and-forget semantics**". Pub/Sub is not fire-and-forget — `base.py` three lines away says a handler that raises gets redelivered. It redelivers precisely what this bus discards.

The swallow itself is right and stays: one bad subscriber must not break its siblings, and `test_handler_exception_does_not_break_other_subscribers` already pins that. What changes is the description — the in-process bus is **at-most-once and cannot be otherwise**, since it holds events in memory — and the log line, which now says the event was *dropped* rather than only that a handler raised. On this backend those are the same fact, and only one spelling tells a reader of the logs that nothing will retry. It carries `dropped=True`, matching the alert hook the worker's own ack-drop paths use.

This matters most where it is least visible: the in-process bus is the standalone and OSS default.

## One test I wrote and then deleted

I added a check that the two retracted claims stay retracted. It failed — because the corrected prose **quotes each claim in order to retract it**, so any search that finds the old wording finds the new wording too.

That is the third time this session a text search could not tell an assertion from its denial (the org-admin predicate in #1603, the RLS sweep in #1602). Here there was no honest way around it: I could contort the documentation to dodge a grep, or keep a check that cannot distinguish the two. I removed the test and recorded why in the module docstring. What is testable is tested.

## What the remaining tests pin

- **The structural gap, not just its instance**: every module in `common/events` that calls `bus.subscribe` must be registered with the generator. Mutation-tested by restoring the hardcoded patch module — the original bug — which the manifest assertion catches.
- **The schema policy** across every wire schema, plus a behavioural test that an unknown field really is tolerated rather than only that the flag reads `ignore`.
- Vacuity-checked: the scan asserts it found subscribing modules at all.

## Gates

`ruff check` clean across the CI scope; `format --check` clean on the checked set and `--isolated` on the three `common/` files CI checks. `common/` was **not** blanket-formatted — the diff is 52 lines across four files. `mypy`: `common/` 2 errors (the standing httpx conflict), core-worker 2 (same file), scripts clean. `gen_events_manifest.py --check` reports up to date. Suites: events **165 passed**, core-worker **100 passed**, root **31 failed / 7162 passed** — the standing baseline.

`common/events/{base,inprocess}.py` are vendored into caura-enterprise as `"manual"` entries, so that copy needs a deliberate re-vendor. No signature changes.

Closes `oss-0814-m-27`, `oss-0902-m-09`, `oss-0814-l-24`.


Eldad-Caura:
The red on `Broker OpenAPI has no breaking change vs main` is infrastructure, not this change. Exit code 125 is docker failing to start the container, so no diff was ever computed:

```
Unable to find image 'tufin/oasdiff:v1.26.0@sha256:95a3f551...' locally
docker: Error response from daemon: Get "https://registry-1.docker.io/v2/tufin/oasdiff/manifests/...":
  net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)
```

Docker Hub was unreachable from the runner while pulling the pinned image.

Worth confirming rather than asserting, since "no breaking change" is exactly the claim a skipped gate cannot make: this PR touches six files — `common/events/{base,inprocess,memory_embed_request}.py`, `events_manifest.json`, `scripts/gen_events_manifest.py`, and one test. `git diff origin/main..HEAD -- core-api/openapi.broker.json core-api/src/core_api/routes/` is empty, so there is no route or schema change for oasdiff to have an opinion about.

Re-ran the failed job.

Eldad-Caura:
@erni-a pleasr aprpvoe

erni-a (approved):

PR #1612, opened 2026-09-18
fix(api): three operations that reported success for something else

Eldad-Caura:
Three operations that returned success for something other than what was asked.

## A delete that deleted the wrong rows

`DELETE /fleet/nodes/{node_name}` deleted the node's **commands** and returned `{"ok": true}` with the node row still in place. The node kept reporting in, kept appearing in `GET /fleet/nodes`, and a second DELETE succeeded exactly as loudly as the first — nothing in the response distinguished them.

The command cleanup it *did* do was also conditional on the node having a `fleet_id`, which a delete does not depend on, so an unfleeted node's commands were left behind too.

**The lookup underneath reaches further than the finding says.** `fleet_get_node_id` used `scalar_one()`, which raises `NoResultFound` for a name the tenant doesn't have — out of **all four** by-name fleet endpoints, as a 500. And it makes the `if node is None: 404` sitting two lines below two of those calls unreachable: nothing ever returned to compare against None.

`GET /commands` had already been written against the contract the service didn't provide:

```python
if node_id is None and node_name:
    node_id = await _svc.fleet_get_node_id(...)   # could only raise
```

So the fix is at the lookup: `UUID | None`, with each caller handling None on its own terms.

**One of those callers is a trap worth naming.** `fleet_list_commands(node_id=None)` drops the node filter. Fixing the 500 by simply letting None through would make `GET /commands?node_name=typo` return *every command in the tenant* — a worse answer than the 500 it replaced. It returns `[]`, and there's a test that fails if that regresses.

## A commit metered as one write

`POST /ingest/commit` charged a flat `check_and_increment(tenant, "write")` for a commit that writes `len(body.facts)` memories. A 50-fact ingest cost the same as a 1-fact one.

The billing gap is the smaller half. The writes counter feeds `_is_over_plan_limits`, so **the cheapest route past a write cap was to ingest in bulk** rather than write one at a time.

Now `bulk_check_and_increment(tenant_id, len(body.facts))` — the same call `caura_write`'s bulk branch makes — before the write, matching the ordering `test_billing_happens_before_the_write` already pins for the MCP surface. That test's own comment says why: *"Two orderings for one operation across two surfaces is the drift this area keeps producing."* The admin skip (`if auth.tenant_id`) is preserved and now has a test of its own.

## A cron tick that reported failure for work that succeeded

`_fire_fanout` called `resp.json()` unguarded on a 2xx. A success body that isn't JSON — an intermediary's HTML page, a truncated response — raised `JSONDecodeError` out of `_fire_fanout`, out of the tick, and was logged as a **failed cron tick**. But the fanout had already fired: core-api dispatched the work before writing that body. So the scheduler fanned the same action out again on the next tick.

The function's docstring says it doesn't raise ("the scheduler retries on the next tick, so re-raising would just produce duplicate stack traces"). This was the one path that did.

## Verification

Every one of the four behaviours mutation-tested — reverting to command-only delete, restoring `scalar_one()`, restoring flat metering, and unguarding `resp.json()` each fail the intended tests and only those.

Two things worth flagging about my own process here:

- I first wrote the fleet mutation runs with `env $SENV pytest …`, and **zsh doesn't word-split unquoted variables**, so those two runs errored on a missing module instead of testing anything. They looked like failures rather than passes, which is the harmless direction, but I re-ran them properly rather than reading an error as a result.
- I introduced a mypy error (`Result` has no `rowcount`) and then found this same file already answers it five times, with a comment explaining why. I matched the house form instead of inventing a sixth.

## Gates

`ruff check` / `format --check` clean across the CI scope. mypy: core-storage-api **clean, 95 files** (it was clean before and is clean again), core-api 2 errors — the standing httpx conflict — and core-operations clean. Suites: storage **530 passed**, core-operations **90 passed**, root **31 failed / 7166 passed**, the unchanged baseline.

Closes `oss-0814-m-20`, `oss-0902-m-29`, `oss-0814-l-14`.


Eldad-Caura:
@erni-a plerse approve

erni-a (approved):

PR #1613, opened 2026-09-18
fix(quickstart): stop publishing default credentials, and delete a control surface that controls nothing

Eldad-Caura:
Two findings about the OSS deployment surface: one publishes more than it should, the other advertises control it doesn't have.

## Default credentials on every interface

`docker compose up -d` is the first command in both README.md and AGENT-INSTALL.md. It started Postgres as `caura`/`changeme` with no TLS, and Redis with **no password at all**, published as `"5432:5432"` and `"6379:6379"` — the two-part form, which Docker binds to 0.0.0.0. Anything that could route to the host could reach them: other machines on the LAN, and on a cloud VM with a permissive security group, the internet.

Redis is the sharper of the two: weak credentials are a different problem from *no* credentials, and its `CONFIG` command can write files.

Both now bind to `127.0.0.1`, overridable via `DB_BIND` / `REDIS_BIND` for anyone who wants to publish deliberately.

**Nothing in the stack loses access.** Service-to-service traffic goes over the Compose network, not the published port. And loopback is what the documentation already promised — `docs/self-hosting.md` lists these as `localhost:5432` and `localhost:6379` — so this closes a gap between the compose file and the page describing it rather than narrowing the supported surface.

`docker-compose.dev.yml` had the same two exposures. I only found them because the guard test ran against both files after I'd fixed the quickstart; that file already bound core-storage-api to `127.0.0.1`, so the datastores were the stragglers.

## A crystallizer control surface that controls nothing

`.env.example` advertised `CRYSTALLIZER_ENABLED`, `CRYSTALLIZER_STALE_DAYS` and `CRYSTALLIZER_DEDUP_THRESHOLD`. All three were bound to `Settings` fields nothing read.

This was the finding I'd said needed your decision — wire them up or delete them. It doesn't, and the reason is that **the capability already exists elsewhere and works**. The crystallizer is configured per organisation: `crystallizer.auto_crystallize` is the live on/off switch (`lifecycle_audit` skips the org when it's false) and `crystallizer.dedup_threshold` the live cosine floor, each with its own rationale in `organization_settings.py` — including why the threshold is deliberately per-tenant and default-unset. The env block was a second surface advertising controls that already existed. All four `Settings.crystallizer_*` fields are deleted and `.env.example` now points at the real one.

## A correction to my own work in #1603

That PR's test pinned three crystallizer settings as known-unread, with a comment asserting `crystallizer_dedup_threshold` "IS read, which is what makes the other three worth naming individually."

**That comment was wrong.** The scan had found `cfg.crystallizer_dedup_threshold` in the crystallizer service — where `cfg` is the per-org `ResolvedConfig`, which exposes a property of the same name. The `Settings` field was as dead as the other three. Matching a bare `.name` cannot tell two objects apart, so the scan reported the one *live* control surface as evidence that the dead one was live.

Fixing it took two attempts, and the first was worse than the bug:

1. I narrowed the match to `settings.` / `app_settings.` — and got **10** newly-"unread" settings, nine of them false. `organization_settings.py` imports the singleton as `global_settings` to use as the per-org fallback, a third spelling I hadn't accounted for.
2. So the scan now **derives** the aliases from each file's imports rather than holding a list. That cleared all nine.

A hand-kept alias set would have gone stale the first time a fourth spelling appeared, and it over-reports dead settings — the noisy direction, but still a scan reporting something it cannot see.

`local_embedding_model` remains as the single allowlist entry, and it's a genuinely different category from the crystallizer knobs: it *declares* an env var that `common/embedding/_registry.py` reads from `os.environ` directly, because that registry is shared with core-worker and must not import a service's config. The control works; only the attribute is unread. Its default now comes from the shared constant — the same remedy the `embedding_provider` field directly above it already uses, whose comment records that the two copies disagreed once before.

## The guard, and a hole I left in it

`tests/test_quickstart_exposure.py` asserts db and redis publish only to loopback, in both compose files, *and* that the bind stays overridable — both halves, so neither can be lost.

The first version had a real hole. It accepted any mapping starting with `${`, meaning to allow `${DB_BIND:-127.0.0.1}` — which also accepted `${DB_PORT:-5432}:5432`, the bug itself wearing a variable. A mutation that reintroduced exactly that passed the loopback assertion and was caught only by the sibling test.

Tightening it surfaced a second bug in my own code: `str.split(":")` on a Compose mapping breaks on the colon inside `${VAR:-default}`, so every mapping in both files read as non-loopback — including the plain `127.0.0.1:` one that had been correct all along. There's now a `${...}`-aware splitter with its own direct assertions, because the splitter is the part that can be wrong.

Both exposure mutations — republishing on all interfaces, and flipping the default to `0.0.0.0` — now fail both tests.

## Gates

`ruff check` / `format --check` clean across the CI scope. `mypy src/` on core-api: 2 errors, the standing httpx conflict. Both compose files parse under `docker compose config`. Root suite **31 failed / 7175 passed** — the unchanged baseline.

Closes `oss-0902-m-68`, `oss-0814-m-47`.


Eldad-Caura:
Applied — and the finding lands on the same defect this PR is about, one level up.

I added `DB_BIND` / `REDIS_BIND` as the override for the new loopback default, then documented them only in the compose files. `.env.example` is the file an operator copies; compose comments are the thing the override exists to save them from reading. A control that works but isn't where you'd look for it is close kin to the crystallizer block this PR deletes, which was in `.env.example` and didn't work.

Checking the premise turned up a bit more than the finding claimed: `.env.example` documented **none** of the compose publishing variables — `DB_PORT`, `REDIS_PORT` and `API_PORT` were already absent before this change. I added all five rather than only my two.

The new block also names a distinction that is easy to get wrong in silence:

```
# These are read by docker-compose.yml, NOT by the application. Keep them
# distinct from POSTGRES_PORT above: that is where core-storage-api CONNECTS,
# these are where the host PUBLISHES.
```

`POSTGRES_PORT` and `DB_PORT` sitting ten lines apart, one meaning "connect here" and the other "publish there", is a trap worth labelling.

Coupled mechanically rather than left to the next reviewer: `test_the_override_is_documented_where_operators_look` asserts each bind variable appears in `.env.example`, so removing the block fails the build. Mutation-tested — deleting it fails both parametrisations.

Gates after the amend: `ruff check` / `format --check` clean, 17 tests across the two guard files pass.

Eldad-Caura:
@claude

Eldad-Caura:
All 11 checks pass and the re-review after the amend came back clean, so this is ready for a human pass.

@erni-a two things worth your eyes, and one of them is a gap in what was reviewed.

**The behaviour change is a default, not a capability.** Postgres and Redis now publish on `127.0.0.1` instead of every interface. Anyone who was reaching them from another machine — CI against a shared runner, a colleague's laptop, a container outside this Compose network — loses that until they set `DB_BIND=0.0.0.0` / `REDIS_BIND=0.0.0.0`. I believe that is the right default given the credentials involved (`caura`/`changeme` with no TLS, and Redis with no password at all), and `docs/self-hosting.md` already described these as `localhost:5432` / `localhost:6379` — so the compose file was the thing out of step. But it is a default flip on the documented quickstart, so it is your call rather than mine.

**The automated review explicitly did not look at the tests.** Its own note: *"`tests/test_quickstart_exposure.py` and `tests/test_shipped_code_is_reachable.py` are `test_*.py` files ... and were excluded from source-level review as instructed."* That matters more than usual here, because the reasoning in this PR lives in those two files — the compose parser, the loopback check, and the corrected settings scan. I mutation-tested all of it (republishing on all interfaces, flipping the default to `0.0.0.0`, reintroducing an inert setting, deleting the `.env.example` block), and I found and fixed two bugs in my own test while doing it: a `${`-prefix check that accepted `${DB_PORT:-5432}:5432`, and a `str.split(":")` that broke on the colon inside `${VAR:-default}`. Still — nothing but me has read that code.

Not merging — that is yours.

erni-a (approved):

PR #1614, opened 2026-09-18
test(storage): pin which session each service method opens, and why draining it is not mechanical

Eldad-Caura:
`oss-0902-l-52` is right: dozens of pure read paths open the transactional writer session. **This PR does not convert them**, and that decision is the deliverable — so here is the evidence for it.

## What building the classifier turned up

Three hazards, each of which a find-and-replace walks straight into:

**1. A method can be all-SELECT and still be a write.**

```python
async def agent_update_fleet(self, agent_id, tenant_id, fleet_id) -> None:
    async with get_session() as session:
        result = await session.execute(select(Agent).where(...))
        agent = result.scalar_one_or_none()
        if agent is not None:
            agent.fleet_id = fleet_id          # ← the write
```

No INSERT, no UPDATE, no `session.add`. The flush at transaction exit is the whole write. Routed to the reader — which opens no transaction, and on a replica holds a read-only connection — that update silently does not happen.

**2. A method can read its own writes.** `entity_resolve_duplicates` says so in its own docstring: *"NOT `get_read_session` — the merge loop re-reads rows it mutates"*, inside SAVEPOINTs an HTTP boundary cannot express. The codebase had already answered that one.

**3. Read-after-write is a property of the caller, not the method.** A method that only selects is still wrong on a replica if a request writes and then calls it. Replica lag isn't visible from inside `postgres_service`, so no amount of analysis *of that file* settles it.

And the reason none of this would be caught late: **in OSS the conversion is a no-op.** `get_read_engine()` returns the writer engine unless `read_database_url` is set. A wrong conversion passes every test in this repo and fails only in a replica deployment.

## What this ships instead

The population, pinned: **138 methods open a writer session, 65 show no write marker.** Those numbers move only when someone changes them deliberately. And the classifier itself, committed, for whoever does the per-method caller analysis a real conversion needs.

## The classifier was wrong four times

Each version looked right and reported a plausible number:

| miss | effect |
|---|---|
| `sql_update(...)` — the aliased import this file uses | writers read as reads |
| `text(f"UPDATE ...")` — a `JoinedStr`, not a `Constant` | 8 archive/apply/restore methods read as reads |
| `agent.fleet_id = ...` — ORM attribute assignment, no call at all | the hazard above read as a read |
| a trailing `\b` applied to the whole alternation, so the `UPDATE\s+\w` branch ended mid-word and could never match | 6 raw-SQL writers read as reads |

That last one is worth dwelling on: the pattern was syntactically fine, matched other branches, and silently disabled exactly one. It reported **73 pure reads**; the real number is 65.

What caught each round was a cross-check from an independent signal — the method's own name. A function called `memory_archive_expired` classifying as a pure read is cheap to notice and impossible to argue with, and that cross-check is committed as a test rather than left as something I did once.

## Verification

Both mutations fail the guard: converting `agent_update_fleet` to the reader (the exact data-loss case this finding invites), and reintroducing the broken regex.

`ruff check` / `format --check` clean. Storage suite **536 passed**.

The finding stays open — this narrows it to "convert the 65, with caller analysis" rather than closing it. I'd rather leave it open and honest than mark it done on a change that didn't do it.

Refs `oss-0902-l-52`.


Eldad-Caura:
CI is green across all 12 checks. The review verdict, though, should not be read as a review of this PR.

Its own note: *"This PR only adds `core-storage-api/tests/test_session_choice.py`, a `test_*.py` file under a `tests/` directory. Per review scope, test files ... are excluded from review, and no source files were changed in this diff."*

So the ✅ means nothing was examined. That is correct behaviour for the scope it was given, and it happens to be the whole of this PR — the entire change is one test file. I would rather say that than let a green check stand in for a read.

@erni-a which makes this one genuinely dependent on a human, and the part worth your time is the **judgement**, not the code:

This PR deliberately does NOT do what `oss-0902-l-52` asks. The finding is right that 65 read-only methods open the transactional writer session; I measured it. I did not convert them, because three things make it unsafe as a mechanical change:

- `agent_update_fleet` runs one SELECT and then assigns `agent.fleet_id = fleet_id`. The flush at transaction exit is the write — no INSERT, no UPDATE, no `session.add`. On the reader it would silently not happen.
- `entity_resolve_duplicates` already documents why it must stay: its merge loop re-reads rows it mutates.
- Read-after-write is a property of the caller. A method that only selects is still wrong on a replica if a request writes and then calls it, and nothing in `postgres_service` can tell you that.

And the trap underneath all three: **in OSS this conversion is a no-op**, because `get_read_engine()` returns the writer unless `read_database_url` is set. A wrong conversion passes every test in this repo and fails only in a replica deployment — which is to say, only in yours.

So I left the finding open and shipped the measurement plus the classifier. If you would rather I convert a scoped subset — say the report/audit/fleet list paths, after checking their callers — that is a reasonable next step and I will do the caller analysis properly rather than by name.

Not merging — that is yours.

erni-a (approved):

PR #1617, opened 2026-09-19
fix(events): make a failed publish visible to the platform, not just the SDK

Eldad-Caura:
Closes enterprise finding `ent-060`. It has to be fixed **here** rather than in caura-enterprise: `common/events/pubsub.py` is vendored there, so patching the copy would be drift.

## What was actually wrong

`publish()` fires into the client's batch queue and returns without blocking on the returned Future. The comment there explains why at length, and it is right: `.result(timeout=30)` would tie up a `publish_concurrency` thread on a slow Pub/Sub, and admin-API request handlers await this call, so a sustained outage would wedge them.

**That stays. What was wrong is that the Future was then discarded.** A publisher-side failure — a 403 on the topic, an exhausted quota — produced log lines from a `google.cloud` logger and nothing this platform watches, while `is_healthy` stayed true because every condition it checks is about the pull side.

## The fix

A done-callback, which closes the gap without reintroducing the blocking the comment exists to prevent: it runs on the SDK's own commit thread when the batch settles, so `publish()` still returns as soon as the message is queued. There is a test for exactly that.

It logs at ERROR with **`dropped=True`** — the field five other places in `common/events` already use to mark an event lost, so existing log-based alerting covers this path with **no new rule**. A publish that failed *is* a dropped event: there is no queue to return it to, because it never reached one.

**Deliberately not flipped into `is_healthy`.** That drives readiness, and draining a pod from the load balancer on one transient publish failure trades a lost event for an outage, and would flap.

## A docstring that was wrong

`is_healthy` opened:

> True when the bus is in a state where it can still deliver events **end-to-end**.

It is not end-to-end. A publisher-only bus whose every publish is 403ing reports healthy, because nothing in that function looks at publishing. Now says so, and says where publish failures surface instead.

## On the finding's wording

`ent-060` called publisher-side failures "permanently invisible". That overstated it — the SDK's own logger did record them, and the audit's own residual note already said as much. The real gap was an **app-level** signal, which is what this adds. Worth stating plainly rather than quietly fixing something narrower than the ticket claimed.

## Evidence

5 tests. Only `test_a_failed_publish_is_logged_as_a_dropped_event` distinguishes before from after — verified failing with the Future discarded again:

```
E  AssertionError: a publish that failed produced no app-level signal;
   the SDK's own logger is the only thing that saw it
1 failed, 4 passed
```

The other four pass either way **by design**, and the module docstring says so: they guard the properties the fix had to preserve (publish stays non-blocking, readiness does not flip) and the ways it could have been wrong without failing a test (logging unconditionally, raising on the SDK's thread where an exception is swallowed and would make the failure *more* invisible).

They use a real `concurrent.futures.Future`, not the shared fixture's `MagicMock` — whose `add_done_callback` records the call and never invokes it, so the callback would never have run.

## Gates

`ruff check` clean across the full CI scope; `ruff format --check` clean on the same minus `common/`, plus `--isolated` on the three `common/` files CI checks. mypy: core-storage-api and core-operations clean; core-api and core-worker each report the 2 standing errors in `common/llm/providers/openai.py` (the httpx/openai conflict) — verified identical on a clean `origin/main` checkout. `gen_events_manifest.py --check`: up to date. Legacy-name ratchet: no new lines. Suites: events **165 → 170**, core-worker **100**, root **31 failed / 7187 passed** — 31 is the standing baseline.

## ⚠️ Enterprise follow-up this creates

`common/events/pubsub.py` is a **`manual`** entry in caura-enterprise's `vendored_files_manifest.json`. Merging this moves the OSS baseline, so that repo's drift gate will fail on every PR until someone reconciles it:

```
FAIL  common/events/pubsub.py: OSS SOURCE MOVED since this copy was last reconciled
```

The 6-hourly re-vendor bot will **not** do it — it skips manual entries by design. That is the same loop that went stale this morning and turned every enterprise PR red. It needs a deliberate re-vendor, and enterprise's copy carries its own divergence (the `DeliveryMetadata` / `DiscardHandler` discard API) that has to be preserved through it.


Eldad-Caura:
@erni-a ready for your review — CI green, automated review clean (no issues found).

**What this is:** a done-callback on the Pub/Sub publish Future so publisher-side failures become an app-level signal. `publish()` still returns as soon as the message is queued — it deliberately does not block on the SDK Future (blocking would occupy a `publish_concurrency` thread and admin-API handlers await this call), and `test_publish_does_not_block_on_the_future` pins that.

**Two notes for the reviewer:**

1. The failure log uses `dropped=True`, the field five other sites in `common/events` already emit, so existing alerting covers it without a new rule.
2. `is_healthy` is deliberately **not** flipped on publish failure. It drives readiness; draining a pod on one transient publish error trades a lost event for an outage. Its docstring did claim "deliver events end-to-end" while checking only the pull side — that wording is corrected here.

Of the 5 new tests, only `test_a_failed_publish_is_logged_as_a_dropped_event` distinguishes before from after (verified failing with the Future discarded); the other four are pins on properties this change must not break, and are labelled as such in the module docstring.

**Downstream: this will turn caura-enterprise's vendored-drift gate red on every open PR once merged.** `common/events/pubsub.py` is a `manual` entry in that repo's `vendored_files_manifest.json`, and the 6-hourly re-vendor bot skips manual entries by design — a human has to run `scripts/check_vendored_drift.py --approve-manual --only common/events/pubsub.py` there. I'll do that reconcile as the immediate follow-up; enterprise's copy carries its own `DeliveryMetadata`/`DiscardHandler` divergence that has to survive it.

Not merging — leaving that to @Eldad-Caura as always.

Eldad-Caura:
@erni-a pleraas rpproar

erni-a (approved):

PR #1618, opened 2026-09-19
refactor(storage): drop three zero-caller service methods, and gate the topic registry

Eldad-Caura:
Closes **OSS-0814-L-53** and **OSS-0814-L-44** — both filed as "unused code", both sitting partial since the 2026-08-14 audit.

## L-53 — three methods nobody calls

`memory_compute_usage_stats`, `relation_list`, `fleet_get_node_ids_for_fleet`. Deleted, 82 lines.

Verified on three axes rather than one grep, because a single `grep *.py` is how a "dead" method turns out to be reached from a fixture or a script:

| axis | result |
|---|---|
| every file type in the tree, not just `.py` | each name appears in exactly one file — its own definition |
| dynamic dispatch | the two `getattr` sites resolve an interview sweep name and a service *constant*; neither reaches these |
| re-exports / `__all__` | none |

**They were also in a backlog, which is the part worth knowing.** Two of the three open a writer session while only reading — the population `core-storage-api/tests/test_session_choice.py` pins, and that [#1614](https://github.com/caura-ai/caura/pull/1614) documented as needing per-caller analysis before any conversion is safe. Deleting them retires two entries **without** that analysis. The pinned counts move 138 → 136 and 65 → 63; the test's docstring anticipated the numbers falling by conversion and now names deletion too, since not needing the method is the cheaper way off the list.

## L-44 — needs no deletion, and establishing that is most of the work

The orphans are already gone. `Memory.CREATED` went in [#1039](https://github.com/caura-ai/caura/pull/1039); the `Pipeline` family and the two comments that referenced it have **zero hits anywhere in the tree**.

`Audit.EVENT_RECORDED` does remain, and removing it would be wrong. `topics.py` states the case in its own words: nothing in this repo publishes or subscribes it, it is consumed downstream, and declaring it keeps the two repos' flip state aligned so a publisher added here later inherits the current name rather than the outgoing one.

**What was actually left is the docs half.** `component-registry.yaml` carries an `event_topic_families` section — a family name and a hand-counted number — and the audit found it claiming a `Pipeline` family whose topics were deleted, next to a `Memory` count that no longer matched. Nothing failed, because nothing was checking: `test_component_registry.py` gated `cron_ticks` in both directions and left this section as prose.

A count a human maintains by hand is a comment wearing a number's clothes. It is now a comparison, walking the `Topics` facade the same way `topics.all_topics` does, so a family added there is covered without anyone remembering to.

Verified by mutation against all three drift shapes — the first two are the audit's own findings, reproduced:

| mutation | caught by |
|---|---|
| `Memory: 4` → `5` (a stale count) | `test_every_declared_topic_family_matches_the_registry` |
| re-add a `Pipeline` family (deleted, still documented) | `test_every_registry_topic_family_is_declared` |
| drop `Org` (a real family missing from docs) | `test_every_declared_topic_family_matches_the_registry` |

## Gates

`ruff check` and `ruff format --check` across the CI scope; `mypy src/` clean on core-storage-api; events manifest up to date; legacy-name ratchet reports no new lines; tenant-scope gate passes with category counts unchanged (194 service methods, down from 197).

Both suites were measured **against a clean `origin/main` worktree** rather than read as absolutes, because both carry standing failures that would otherwise drown the signal:

- core-storage-api: **11 failed / 525 passed / 7 errors** on each side, identical.
- root: the same **31 failures** on each side, **7182 → 7185** passing — the three added by the new gate.

That comparison earned its keep. The first run came back 12 failed against the baseline's 11, and the one extra was the session pin above; read as an absolute against a suite with 11 standing failures, it would have looked like noise.

**`common/` is untouched** — it carries pre-existing `ruff format` drift, and no CI step format-checks it.


Eldad-Caura:
@erni-a ready for your review — CI fully green, automated review clean on the first round, one commit.

**What it is:** two "unused code" findings from the 2026-08-14 audit, `OSS-0814-L-53` and `OSS-0814-L-44`.

**L-53** deletes three methods with no caller anywhere — `memory_compute_usage_stats`, `relation_list`, `fleet_get_node_ids_for_fleet`. Checked on three axes rather than one grep (every file type, dynamic `getattr` dispatch, re-exports); each name appeared exactly once in the repo, its own definition.

Two of the three were in the writer-session backlog `test_session_choice.py` pins — readers holding a writer session, the population #1614 said needs per-caller analysis before conversion. **Deleting them retires two entries without that analysis**, which is why the pin moves by 2 and not 3: `memory_compute_usage_stats` already used `get_read_session` and was never in that population.

**L-44 needed no deletion**, and that is the part worth a second pair of eyes. The orphan topics are already gone — the tracker was stale, not the code. `Audit.EVENT_RECORDED` does remain, and I left it deliberately: `topics.py` records that nothing here publishes or subscribes it, that it is consumed downstream, and that declaring it keeps the two repos' flip state aligned so a publisher added here later inherits the current name. **If you disagree with that reading, it is the one judgement call in this PR.**

What was actually left of L-44 is the docs half: `component-registry.yaml`'s `event_topic_families` is a hand-counted number, and that is how it came to claim a deleted `Pipeline` family next to a wrong `Memory` count. `test_component_registry.py` already gated `cron_ticks` in both directions and left this section as prose. It is now a comparison, verified by mutation against all three drift shapes — two of which reproduce the audit's own findings.

**On the test results in the description:** both suites were measured against a clean `origin/main` worktree rather than read as absolutes, because both carry standing failures. That mattered — the first run was 12 failed against the baseline's 11, and the one extra was a real regression (the session pin). Against a suite with 11 standing failures it would have read as noise.

Not merging — that stays with @Eldad-Caura.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

Issue #1616, opened 2026-09-19
entity-link cannot complete for one tenant: cross-link discovery exceeds 120s every night


Eldad-Caura:
## Symptom

One staging tenant, `knowledge-update-benchmark-0d1a47`, cannot complete
`entity-link` and dead-letters every night. It has now produced two DLQ messages
on `caura.lifecycle.entity-link-requested` (2026-09-18 and 2026-09-19), each
after 10 delivery attempts, and each one fails every staging deploy until drained
because a DLQ message is durable and the smoke gate requires depth zero.

Nothing else in the tenant is unhealthy: `crystallize` (02:00) and `insights`
(05:00) both process it without error on the same nights.

## What is measured

The storage side returns a Cloud Run request timeout:

```
POST /api/v1/storage/entities/discover-cross-links
504  latency 119.990881178s
"The request has been terminated because it has reached the maximum request timeout"
```

The caller gives up first, at its own budget, and now says so legibly (#1604):

```
RuntimeError: entity-linking pipeline failed for org knowledge-update-benchmark-0d1a47:
failures=[{'error': 'cross-link discovery exceeded its 100.0s budget; ...
```

Every other org in the same 06:00 run finished in under 20 seconds
(`Created 0 cross-links` for six of them between 06:00:02 and 06:00:20).

So this is not a race and not a flake: the call genuinely exceeds 120s, by a
margin, for this tenant only.

**#1604 did not fix this and was not expected to.** Its budget made the failure
legible — that is why this issue can be written at all — and its
`CROSS_LINK_MEMORY_BATCH_SIZE` 200 → 100 was headroom, not a cure. Moving
entity-link to its own 06:00 hour (caura-ai/caura-enterprise#1826) ruled out
contention with crystallize: it now runs alone and still fails.

## What is inferred, and not yet proven

The likely bottleneck is the per-candidate ANN in `entity_discover_cross_links`:

```sql
JOIN LATERAL (
    SELECT ... FROM entities e
    WHERE e.tenant_id = :tenant_id
      AND (1 - (e.name_embedding <=> m.embedding)) >= :threshold
    ORDER BY e.name_embedding <=> m.embedding
    LIMIT 10
) e ON true
```

`ix_entities_name_embedding_hnsw` (migration 033) is on `name_embedding` ALONE —
there is no tenant-scoped or composite vector index on `entities`, and 033's own
docstring frames the fix as "each lookup is O(log n)" without addressing the
[...]

PR #1620, opened 2026-09-19
fix(storage): name the phase when cross-link discovery times out

Eldad-Caura:
Instrumentation for #1616, so tomorrow's 06:00 run answers the question instead of the next guess.

## The problem this solves

`knowledge-update-benchmark-0d1a47` fails entity-link every night. Storage returns `504` at `119.990881178s` on `POST /api/v1/storage/entities/discover-cross-links`; six other orgs in the same run finish between 06:00:02 and 06:00:20. A 504 names the endpoint and nothing else — and this endpoint does four things:

| phase | what it is | why it could be the 120s |
|---|---|---|
| `candidates` | `LEFT JOIN … GROUP BY m.id HAVING COUNT(mel.entity_id) < 3 ORDER BY m.created_at DESC LIMIT n` | aggregates **every** memory in the tenant before it can sort or limit; carries `content` + `embedding` through that aggregate; there is no index on `(tenant_id, created_at)` for the sort to ride |
| `lateral` | pgvector LATERAL, `LIMIT 10` per candidate | `ix_entities_name_embedding_hnsw` (migration 033) is on `name_embedding` **alone**, so `tenant_id` and the threshold are post-filters on a global ANN scan — once per candidate |
| `text_verify` | Python substring filter | |
| `insert` | one multi-VALUES `ON CONFLICT DO NOTHING` | |

The first two want **opposite** fixes, and halving the batch size earlier helped neither — it halved the number of deep scans, not their cost. I could not separate them from outside the database, and a wrong guess here ships a patch to the wrong query.

## What this adds

`PhaseTimer` in `core_storage_api/observability.py` — next to `Timer`, which answers "how much of this request was the database". This answers the question a request that never returns otherwise cannot answer at all: **which step was still running when we were killed**.

Three things in the design are load-bearing and each has a test:

- **Reporting happens in `__aexit__`, not from a `try/except`.** A severed Cloud Run request arrives as `asyncio.CancelledError` — a `BaseException`, so it slips straight past the `except Exception` normally written for this job.
- **A phase that does not finish is never filed as completed.** `phase()` records on the line *after* its `yield`, so an interrupted phase stays in flight. The `finally` that reflex suggests would file it as complete, clear the in-flight name, and leave the warning saying `(between phases)` — the signal destroyed, silently, with every test still green. `test_phase_timer_does_not_file_an_unfinished_phase_as_completed` asserts it directly.
- **Nothing is logged on the happy path.** Every ingest takes targeted mode through this call; a line per phase per call would be paid for forever. The existing summary line carries `breakdown()` instead, so log volume is unchanged until something fails.

`other=` in the breakdown is not padding — the timer wraps `get_session()`, so connection acquisition and COMMIT land there. If *that* is the 120s, no named phase would ever have shown it.

`__aexit__` is typed `Literal[False]`, not `bool`: the annotation is the "never swallows" promise, and mypy enforces it from the other side (as `bool` it makes every `async with` a possible swallow point).

## Not in this PR

The fix. `memory_scored_search` in this same file already runs a bounded two-stage ANN candidate pool (`_ann_pool_size`, `hnsw.iterative_scan`, gated on pgvector ≥ 0.8) whose own warning names this exact failure mode; cross-link discovery never got it. That makes the eventual patch "apply the idiom already in this file" rather than new architecture — but two things need settling first, both noted on #1616: the parity analysis its comment cites (`docs/plans/hnsw-two-stage-retrieval.md`) **does not exist in either repo**, and cross-link has a different recall contract from search (a missed link never forms; it is not merely ranked lower). Also unknown: pgvector's version on these deployments — the probe only runs when `ann_pool_size > 0`, and there is no `ann_pool:` line in 30 days of staging storage logs.

## Verification

- `ruff check core-storage-api/src/` · `ruff check tests/` · `ruff format --check core-storage-api/src/` · `ruff format --check tests/` — all clean, run separately at CI's scopes.
- `mypy src/` in `core-storage-api` — clean (95 files). It caught the `bool` vs `Literal[False]` issue above.
- **7 mutations, each caught by its intended test**: `finally` in `phase()`; log on success; `__aexit__` returns True; in-flight falls back to the last completed phase; `breakdown()` drops `other`; the service logs without the breakdown; report only for `Exception` (the guard that misses cancellation).
- Suites: `test_observability.py`, `test_ph6_entity_linking_storage.py`, `tests/pipeline/test_entity_linking.py`, `test_entity_linking_wiring.py`, `test_entity_link_lock_order.py`, `test_request_observation.py` → **98 passed, 4 failed**. All four failures reproduce **identically on pristine `origin/main`** (verified in a detached worktree at `f5076bac`): one stale-local-DB `uq_relations_natural_key` constraint, three health-router probe tests.
- `git diff -w` shows the real change in `postgres_service.py` — much of the diff there is indentation from the four `with` blocks. The one non-mechanical edit is hoisting both candidate queries out of the `if/else` into one `execute`, so the `candidates` phase times the query and not the branch that chose it.


Eldad-Caura:
@claude Both halves of your finding were real — I reproduced each before changing anything — and both are fixed. Rebased onto `f26e45d5` and force-pushed; head is now `21eb94e8`. Please re-review.

**(a) `other` absorbed the unfinished phase.** Reproduced exactly as described:

```
before: demo interrupted after 0.4s during phase=lateral [candidates=0.1s other=0.3s]
after:  demo interrupted after 0.4s during phase=lateral [candidates=0.1s lateral=0.3s(unfinished) other=0.0s]
```

`_current` now carries `(name, started)`, so an in-flight phase is billed to itself. One deviation from your suggested shape: the marker goes *after* the `=` (`lateral=119.5s(unfinished)`) rather than in the name (`lateral(in-flight)=`), so that grepping `lateral=` still finds both forms — which is how anyone would actually search these lines.

**(b) `other` excluded COMMIT on the success path.** Confirmed: `get_session` yields from inside `session.begin()`, so COMMIT runs in its `__aexit__`, after the body.

You offered docstring-or-move. I moved the log, because the docstring fix would have documented the instrument's own gap away — a large `other` is supposed to *mean* "acquisition or commit", and with the summary emitted pre-commit that claim was simply false. The summary now sits outside the `async with`. Since no runtime assertion can see the difference on a test database that commits in microseconds, `test_discover_logs_its_summary_after_the_commit` walks the AST and fails if the call is ever nested back inside.

I also froze the clock at `__aexit__` — a latent version of the same bug, where a breakdown taken after the block kept inflating `other` with the caller's own logging time.

**Ten mutations, each caught by its intended test** (three are new and encode your finding): `finally` in `phase()` · log on success · `__aexit__` returns True · in-flight falls back to the last *completed* phase · `breakdown()` drops `other` · service logs without the breakdown · report only for `Exception` · **fold the unfinished phase back into `other`** · **clock never frozen** · **summary logged back inside the session block**.

ruff (4 scopes, separately) and mypy clean; 50 passed with one pre-existing `uq_relations_natural_key` failure that reproduces identically on pristine `origin/main`.

Eldad-Caura:
@claude Verified and fixed. Rebased onto `1f73f863`, head is now `39d41a81` — please re-review.

The collision is worse than "silently overwritten": `logging` **raises** rather than overwriting a `LogRecord` attribute. Measured:

```
module     -> KeyError: "Attempt to overwrite 'module' in LogRecord"
message    -> KeyError: "Attempt to overwrite 'message' in LogRecord"
process    -> KeyError: "Attempt to overwrite 'process' in LogRecord"
tenant_id  -> logged fine
```

Which lands precisely where it hurts most: the only `extra` this class ever passes is the one in `__aexit__`, so the KeyError would fire mid-incident, inside the unwind, and replace the `CancelledError` the caller needs to see with an exception about logging. Instrumentation destroying the signal it exists to emit.

So the check runs at **construction**, not at logging time — a bad call site then fails on an ordinary healthy request, loudly, where it is a one-line fix, instead of first appearing during the outage. It raises `ValueError` to match `log_request`'s reserved-`slow` idiom already in this file. The reserved set is derived from a sample `LogRecord` rather than hand-listed, so it cannot drift from what `Logger.makeRecord` actually enforces.

Two tests: one parametrised over `module`/`message`/`process`/`phase`/`total_s`, and a drift guard asserting `_OWN_LOG_FIELDS` equals the set `__aexit__` actually puts on the record — because if `__aexit__` grows a fourth field and the tuple isn't updated, that field becomes silently overwritable again and the guard quietly stops covering it.

**12 mutations now, each caught by its intended test** — the two new ones being *guard removed* and *`__aexit__` grows a fourth field without updating the tuple*.

I deliberately did **not** extend the same guard to `log_request`, which has the identical latent issue (arbitrary `**fields` into `extra`, only `slow` checked). It is a real bug but a pre-existing one in a function this PR doesn't touch, and adding validation there could turn a currently-crashing call site into a differently-failing one outside this change's blast radius. Worth its own issue.

ruff (4 scopes) and mypy clean; 56 passed with the one pre-existing `uq_relations_natural_key` failure that reproduces on pristine `origin/main`.

Eldad-Caura:
@erni-a please approve

erni-a (dismissed):

Eldad-Caura:
@claude Applied. Head is now `8abfef29` on base `42e537b2` — please re-review.

`phase()` now raises `RuntimeError` when a phase is already in flight. I took the suggestion rather than arguing the single-call-site line, because declining it would have been inconsistent with the reserved-field check I added one round ago on exactly this reasoning: a mistake in instrumentation must surface on a healthy request, where it is a one-line fix, and never first appear during the outage it was installed to explain. That rule now covers three things — reserved field names, nesting, and phase completion — and the commit message states it as the rule rather than three coincidences.

Two tests: the nesting case, and one asserting the guard does **not** misfire on ordinary sequential phases (the failure mode a naive `if True` guard would have, which is mutation 14).

**14 mutations, each caught by its intended test.**

One thing worth flagging, unrelated to the code. This branch was force-rebased twice by automation while the review was running — `bed6b638`, then `68883256` — so the clean verdict you posted at 15:07 was against `39d41a81`, a SHA that no longer exists on the branch. I verified both rebases preserved my commit's patch byte-for-byte (only blob hashes and hunk offsets moved; `#1618` had removed ~82 lines earlier in the same file) before building on them. Flagging it because with `main` merging roughly every ten minutes right now, a verdict can go stale between being posted and being read.

ruff (4 scopes, separately) and mypy clean; 58 passed with the one pre-existing `uq_relations_natural_key` failure that reproduces on pristine `origin/main`.

Eldad-Caura:
@erni-a pleaee appoei

erni-a (approved):

PR #1622, opened 2026-09-19
chore(client-typescript): add a Prettier lint gate to the TypeScript client CI

### LINKED ISSUE #1403: chore(ci): TypeScript client has no lint gate, while the Python client CI runs ruff
## What

`.github/workflows/client-python-ci.yml` gates the Python client on lint:

```yaml
      - name: Lint (ruff)
        run: ruff check src tests
      - name: Test (pytest)
        run: pytest -q
```

`.github/workflows/client-typescript-ci.yml` has no equivalent:

```yaml
      - name: Install
        run: npm install
      - name: Build + test
        run: npm test
```

`npm test` is `tsc && node --test dist/index.test.js`. `tsconfig.json` sets `"strict": true`, so type errors are caught — but nothing checks formatting or catches the lint-class problems ruff catches on the Python side (unused imports, unreachable branches, shadowed bindings).

## Why it matters

Two clients, published side by side from one repo, held to different standards. A contributor who lints their Python change and not their TypeScript one isn't doing anything wrong — the repo never asked.

## What to change

Two viable routes; **pick one and say why in the PR** rather than doing both:

1. **Prettier check only** — `npx prettier --check src` as a step, plus a `.prettierrc` in `clients/typescript/`. Zero runtime deps, catches formatting drift, no rule bikeshedding. This is the smaller change and probably the right first move.
2. **ESLint with `typescript-eslint`** — real lint rules, closer to ruff in what it catches, but adds a dev-dependency tree and a config to maintain for a ~230-line source file.

Whichever you choose:
- Add it as a named step **before** `Build + test`, matching the Python workflow's ordering.
- Add the matching `"lint"` script to `clients/typescript/package.json` so it runs locally the same way CI runs it.
- Fix whatever the first run reports, in the same PR.

Note `npm install` (not `npm ci`) and the Node-20-only matrix in this workflow are already tracked in #973 — leave those alone here so the two PRs stay reviewable.

## How to verify

The workflow is path-scoped to `clients/typescript/**` and `.github/workflows/client-typescript-ci.yml`, so a PR touching either triggers it. Green run with the new step visible in the job log.

## Scope

One workflow, one package.json, one config file, plus any fixes the first run surfaces.

---

_Filed unlabeled — the account that opened this lacks label-write permission. Intended labels: `good-first-issue, help-wanted, kind/chore, area/infra, status/needs-triage`._

erni-a:
## Summary

Closes #1403.

The Python client CI runs ruff before pytest; the TypeScript client CI ran only `tsc` and `node --test`, so nothing caught formatting drift. This takes the smaller of the two routes the issue offers, **Prettier `--check`, not ESLint**: one dev dependency, one config key, and the source is about 300 lines, which does not justify a rule set to maintain. ESLint stays available as a later step if the client grows.

- `clients/typescript/.prettierrc`: `printWidth: 110`, matching the `line-length = 110` the repo's ruff config already uses for Python, so both clients wrap at the same column. Everything else is Prettier's default.
- `package.json`: `prettier ^3.6` as a devDependency (pinned through the lockfile, so CI and local runs agree), plus `lint` and `format` scripts so contributors run exactly what CI runs.
- `client-typescript-ci.yml`: a `Lint (prettier)` step before `Build + test`, mirroring the Python workflow's ordering.
- `src/`: the reformat the first run reports. No logic change.

`npm install` and the Node-20-only matrix are left alone; #973 tracks them. Stacked on #1621 (same workflow file); merge that first and this rebases clean.

## Related Issue

Closes #1403.

## Type of Change

- [x] Other: CI gate for the TypeScript client

## How Has This Been Tested?

- `npm run lint` fails on `main`'s source and passes after the reformat.
- `npm test`: tsc clean, 31/31 pass on Node 20.
- `.prettierrc` is not in the package `files` list, so the published tarball is unchanged.
- The workflow is path-scoped to `clients/typescript/**`, so this PR triggers it and the new step is visible in the job log.

## Checklist

- [x] Tests: existing suite covers the reformatted code; the lint step is the new gate
- [x] Every commit is signed off (DCO)
- [x] Not user-facing, no CHANGELOG entry


arkash20 (dismissed):

arkash20 (approved):

erni-a:
Rebased onto main after #1561 (User-Agent header) touched the same test file and package.json. Conflicts resolved by taking main's versions and re-applying the lint-gate edits; tsc clean, 33/33 pass, lint clean. The force-push dismissed the earlier approval, so re-requesting review, @arkash20. Diff vs main is unchanged in intent: workflow step, .prettierrc, package.json scripts + prettier devDependency, and the reformat.

arkash20 (approved):

PR #1626, opened 2026-09-19
fix(documents): let GET accept the `id` that POST returns (ax-0917-h-07)

arkash20:
## The bug

`POST /documents` returns **both** keys:

- `id` — the row's primary key (a UUID)
- `doc_id` — the caller's own key

`GET /documents/{doc_id}` only ever looked up by `(tenant, collection, doc_id)`. So an agent that stored the returned `id` — the conventional thing to keep from a create response — got a **404 reading back the document it had just written**.

Found by an agent probe against prod. It breaks the most basic loop there is: write something, read it back.

## The fix

`document_get_by_pk` is the new lookup. The route falls back to it when the natural key misses **and** the path segment parses as a UUID.

**Order is load-bearing:** the natural key is tried *first*, so a caller whose own `doc_id` happens to be UUID-shaped still resolves to **their** document rather than to whatever row shares that primary key.

### Two things the fallback does not relax

**Tenant scoping.** A primary key is globally unique, so without the tenant predicate this would be a cross-tenant read for anyone who learned an id — being a pk lookup is exactly *why* that can't be skipped. `readable_tenant_ids` widens it the same way `document_get_by_doc_id` does, so a cross-tenant credential that can read a sibling by `doc_id` isn't refused by `id`.

**The collection.** The pk identifies the row on its own, so `collection` isn't part of that query — but silently returning a document from a *different* collection than the caller named would make the parameter a lie. A mismatch is a miss.

Non-UUID doc_ids never reach the fallback: parsing fails closed rather than raising, which matters because most doc_ids are slugs and the forge writes `forge/<slug>`.

## Tests

12 added; **6 fail against `main`**. The lookup and its tenant scoping, cross-tenant widening, the try-natural-key-first ordering, the collection-mismatch miss, fail-closed parsing, and the slug shapes that must keep working.

275 passed / 0 failed on the `document or doc_` selection.

## Note for anyone running this locally

`main` has moved the suite onto its own database — `conftest` now defaults to `caura_test`, not `caura`. If you get `InvalidCatalogNameError`, the conftest names the fix: `createdb caura_test && psql -d caura_test -c 'CREATE EXTENSION IF NOT EXISTS vector'`.


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #1627, opened 2026-09-19
fix(documents): say when a document is unsearchable instead of returning a bare 0 (ax-0917-h-08)

arkash20:
## What happened

An agent probe created a document, queried it seconds later using words from its own body, and got `{count: 0, results: [], items: []}` with HTTP 200. Retried, same. **It concluded search was broken.**

Search was working. The document was never searchable.

`document_search` filters `embedding IS NOT NULL`, and a document is embedded only when its write resolves an embed source — `data["summary"]`, or `data["description"]` for skills. A document written without one is stored, readable by id, and **permanently invisible to search**.

That part is by design and isn't changed here. What was wrong is that **nobody could tell**: the write said 200, the read by id worked, the search returned a plain zero. The write path even recorded `indexed` in its audit row — the caller was the one party who couldn't see it.

## Fixed at both ends

**Write** — the response now carries `indexed`, set from the embedding the route just resolved, so an agent knows at write time whether what it stored will ever come back from search.

Optional rather than defaulting to `False`: read paths (GET, query, list) don't know a row's embedding state, and asserting `False` there would be untrue.

**Search** — a zero-result search now reports `unindexed_count` plus a note naming the remedy, so *"your query matched nothing"* and *"nothing here was searchable"* stop being the same response. A bare count wouldn't tell an agent what to **do**, so the note names `data.summary`.

The count is taken **only when there are no hits** — the normal path pays nothing, and the number is only interesting when it explains a zero. It's wrapped: diagnostics must never turn a successful empty search into an error.

`document_count_unindexed` mirrors the search's own predicates (tenant, collection, fleet, readable set) and selects the exact complement of what search takes — `embedding IS (None)` against search's `IS NOT (None)`. A count over a different scope would answer a question the caller didn't ask; a test pins both halves.

## Tests

11 added; **all 11 fail against `main`**. Wire-up at both ends, the `None` default, audit/response agreement, the no-hits-only cost, the non-fatal wrapper, the remedy in the note, and the counter's scope and complement.

274 passed on the `document or doc_` selection (263 on `main` + these 11); `mypy` clean on the touched files.


arkash20:
@Eldad-Caura please approve

Eldad-Caura (dismissed):

erni-a (approved):

PR #1628, opened 2026-09-19
fix(bulk): flag a deferred embedding so bulk callers can see it (ax-0917-h-06)

arkash20:
## The bug

`embedding_pending` is public API — `MemoryOut` documents it, `caura_write`'s tool description tells agents to read it, and core-worker clears it when the vector lands. The single-write path sets it in `write_memory_row`.

**The bulk path never did**, and `MemoryOut`'s own comment admitted it:

> the bulk path sets neither flag — so a bulk caller cannot read pendingness off its own write response.

That's the wrong way round. **Bulk is where the deferred window is longest** — thousands of rows queued behind one backfill — so it's the caller most in need of the signal, and the only one without it.

An agent probe wrote a memory, searched ~2 minutes later, found other memories but not its own, and had nothing in the write response to explain why. Until the vector lands a row is reachable by FTS and the non-semantic list, but not by semantic similarity — so a paraphrase search comes back empty while an exact-words search doesn't.

## The fix

Set the flag on bulk items written without a vector.

**Conditional on the vector actually being absent** — a strong-mode bulk item embeds inline and is immediately searchable, so flagging it pending would send callers to wait for nothing.

**Through `set_system_value`** like every other platform key (C25), so it lands where a caller can't forge it and matches how the single-write path sets the same flag. A test pins that the two paths agree — a caller shouldn't have to know which one ran to know what the flag means.

**`enrichment_pending` deliberately still isn't set for bulk**: enrichment defers there unconditionally, so there's no inline case for that flag to distinguish, and setting it would be noise rather than signal.

The `MemoryOut` comment that told callers bulk was exempt is corrected in the same change — left in place it would keep teaching the old rule.

## Tests

6 added; **4 fail against `main`** (the other 2 are guards that correctly hold both ways — the single-write path's behaviour and the deliberate absence of `enrichment_pending`).

Pre-existing failures on the `bulk or memory_service or write_mode or embedding` selection are unchanged: `main` fails 4, this branch fails 3 — a strict subset.

## Relevance beyond the audit row

This is the signal the PersonaMem benchmark harness would have needed: it writes via `/memories/bulk`, and a 3.5k-chunk corpus has the longest possible deferred window. See `pm-0918-c-01`.


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #1629, opened 2026-09-19
fix(idempotency): record the key even when the caller disconnects (ax-0917-h-09)

arkash20:
## What happened

An agent probe POSTed a document. It ran **~52s**, the client lost the response, and the server **had already committed**. The retry got `409 "still in progress"`, and a later replay with the same key performed a **second upsert**.

So the idempotency key failed in precisely the scenario it exists for: a lost response.

## Why

The write being durable was never the problem.

`IdempotencyGuard.record` is awaited **on the request task**. A client disconnect cancels that task, and `CancelledError` is a `BaseException` — so the `except Exception` around the storage call never saw it.

That explains the probe's whole sequence: the claim stayed unfinished (→ the 409 on retry), then aged out, and the replay was treated as a fresh request.

## The fix

Shield the record, matching `storage_client._cancel_safe`. It lands even though the caller has gone, and the caller **still observes `CancelledError` immediately** — the request really was cancelled, and pretending otherwise would hang it.

A shielded task that outlives its request needs somewhere for its outcome to go, or a failure surfaces as *"exception was never retrieved"* attributed to nothing — hence the done-callback. It checks `task.cancelled()` **before** `task.exception()`, because `exception()` raises on a cancelled task and the logger must not become the thing that throws.

Ordinary storage failures are still swallowed, unchanged: the client already has the live response, so losing the cache must not turn a successful write into an error.

## Not addressed here

**A document POST taking 52s at all.** The doc write embeds on the request path (`background=False`), so a slow provider blocks the write — which is what opened this window in the first place. This fix makes the key survive the window rather than closing it, and the latency deserves its own row.

## Tests

8 added; **7 fail against `main`**. The record surviving cancellation, the cancellation still propagating, ordinary failures still swallowed, the happy path, and the callback's cancelled-before-exception ordering.

57 passed on the `idempot` selection; `mypy` clean on the touched file.


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #1631, opened 2026-09-19
perf(mcp): stop pretty-printing and ASCII-escaping MCP responses

arkash20:
**ax-0917-m-11.** Every MCP response is read by a model, not a person, so `json.dumps`'s two defaults are both pure cost on this surface.

`indent=2` spends tokens on whitespace that carries no meaning to the reader.

`ensure_ascii=True` is the larger one, and it **only bites non-English tenants**: it renders every non-ASCII character as `\uXXXX` — six ASCII characters where one character was meant — and those escapes tokenise badly. A Hebrew memory costs roughly three times what the same memory costs in English, for no reason.

### Measured

cl100k tokens, as the payload reaches the client — JSON-RPC envelope included, not just our own string:

| payload | before | after | |
|---|---:|---:|---:|
| 10-item English recall | 671 | 454 | **−32%** |
| 5-item Hebrew recall | 1,846 | 564 | **−69%** |

### Why the envelope is safe to widen

Our string is carried as `TextContent.text` inside an envelope the MCP SDK serialises itself. If that outer pass re-escaped non-ASCII, the larger saving would be zero — and invisibly so, because our own output would still look correct. It doesn't: pydantic's `model_dump_json` emits UTF-8. A test pins that behaviour, since an SDK change there is the only way we'd find out.

### Scope

All 26 call sites now go through one `_dumps` helper instead of each deciding for itself — that drift is exactly why the old behaviour was inconsistent, with some sites indented and some not. `default=str` is carried into the helper unchanged, so a `datetime` or `UUID` reaching a response still degrades rather than failing to serialise mid-tool.

`doc_indexing._render_value` already reached this conclusion for text fed to an embedder and an LLM; this applies it to the transport.

### Verification

- 9 new tests, all 9 fail against the parent commit and pass on this one.
- `-k "mcp or recall or serial"`: 615 passed. The 7 failures are identical to `main` (local search-pipeline schema drift), byte-for-byte.
- `ruff check`, `ruff format --check`, `mypy core-api/src/core_api/mcp_server.py` (0 errors in this file) all clean.

No behaviour change beyond the encoding: output is byte-identical after `json.loads`.


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #1632, opened 2026-09-19
fix(observability): reject log field names that logging refuses to overwrite

Eldad-Caura:
`log_request` passed arbitrary caller-supplied fields straight into `logger.log(..., extra=fields)`, guarding only the reserved `slow` kwarg.

## The bug

Python's logging does not let `extra` set a name that already exists on a `LogRecord`. `Logger.makeRecord` raises rather than overwriting — reproduced, not inferred:

```
>                   raise KeyError("Attempt to overwrite %r in LogRecord" % key)
E                   KeyError: "Attempt to overwrite 'module' in LogRecord"
/Library/Frameworks/Python.framework/Versions/3.14/lib/python3.14/logging/__init__.py:1637
```

So a field named `module`, `message`, `process`, `args`, `name`, `levelname`, `filename`, `lineno` or `asctime` did not produce a malformed log line — **it raised out of `log_request`**.

That matters because of *where* this runs: the router-level timing path of the hot reads, inside a `finally`, after the request has already done its work. **The request succeeded and the log line is the only thing that failed** — and the caller gets a 500.

## The fix

Reuses `_RESERVED_LOG_FIELDS`, added by #1620 for `PhaseTimer`, rather than writing a second list. It is derived from a sample record — `frozenset(vars(LogRecord(...))) | {"message", "asctime"}` — so it cannot drift from what `makeRecord` enforces. That derivation is doing real work: `message` and `asctime` are **not attributes of a fresh record at all**, they are filled in during formatting, which is exactly what a hand-written list gets wrong.

The `slow` check stays separate rather than folding into the same set. Its message says *why* that name is reserved — it is computed here from `db_ms`, and a caller passing it would silently drop the INFO→WARNING upgrade for slow queries. Collapsing the two would trade that for a generic message.

## No existing call site collides

Checked **before** changing anything, and against the derived set programmatically rather than by eye:

| call site | fields |
|---|---|
| `scored-search` | `tenant_id, top_k, total_ms, db_ms, row_count, has_date_range, has_temporal_window, error` |
| `load-by-ids` | `tenant_id, total_ms, db_ms, row_count, id_count, error` |
| `memory-get` | `tenant_id, total_ms, db_ms, hit, error` |

None is reserved. Every call site is a literal keyword call — no splat, no alias, no re-export — so the static check is complete rather than indicative. **Nothing was renamed, and there is no pre-existing runtime crash to report.**

Also confirmed `path` is safe to assign after validation: `pathname` is a `LogRecord` attribute, `path` is not, and `path` is a positional parameter so Python's own binding already rejects a duplicate.

## Tests

**5 parametrised cases over real reserved names, each verified failing without the guard** — and failing with the `KeyError` above, not a soft assertion.

Two further tests pass either way, **by design**, and say so in their docstrings rather than padding the count:

- `test_the_reserved_set_matches_what_logging_enforces` drives every name in the derived set through a real logger and asserts `KeyError`. It fails on a Python release that adds a `LogRecord` attribute — the moment the guard would otherwise start silently missing one.
- `test_log_request_still_accepts_ordinary_fields` passes every name the three production call sites use, pinning that the guard did not widen.

## Gates

Run separately, at CI's exact scopes — all clean:

`ruff check core-storage-api/src/` · `ruff check tests/` · `ruff format --check core-storage-api/src/` · `ruff format --check tests/` · `cd core-storage-api && mypy src/` (95 files)

Both suites measured **against a pristine `origin/main` worktree**, because both carry standing failures:

| suite | baseline | this branch |
|---|---|---|
| root `tests/` | 31 failed, 7424 passed | 31 failed, **7431** passed |
| `core-storage-api/tests/` | 11 failed, 533 passed, 7 errors | **identical** |

Same 31 root failures on both sides — including the `test_ph6_entity_linking_storage` and `test_request_observation` ones already failing on pristine main here — with +7 passing, which is exactly this PR's new tests.

**One environment note:** `tests/test_observability.py` runs under system `python3` 3.14 as expected (34 passed), but the *whole* root suite cannot collect there — 45 collection errors from the `mcp` in 3.14 site-packages lacking `CacheHint`. That is identical on pristine main, so it is environmental, but it is why the whole-suite comparison above ran under the core-api venv.


Eldad-Caura:
@erni-a plerasr approve

erni-a (approved):

PR #1633, opened 2026-09-19
fix(api): make the request-budget 504 readable instead of `{"detail":...}`

arkash20:
**ax-0917-h-01 / h-02**, partial — the diagnosis half.

The only 504 this service emits itself is `RequestTimeoutMiddleware` giving up on its own 45s budget. It answered:

```json
{"detail": "request timeout"}
```

No code, no budget, no elapsed time — and a shape nothing else on the surface uses. Every other 4xx and 5xx here, REST and MCP alike, returns `{"error": {"code", "message", "details"}}`, so a client branching on `error.code` got **nothing** from this one and could only report a wall-clock number.

That is exactly what h-01 and h-02 came back as: three probe attempts against `/recall` and `/search` on 2026-09-17, ~45s each, CRUD healthy throughout, and no diagnosis to show for them. The response could not say whether a backend had failed or whether *we* had cancelled a handler that was still working — opposite problems, in opposite systems.

### After

```json
{
  "error": {
    "code": "REQUEST_BUDGET_EXCEEDED",
    "message": "Request exceeded the 45.0s server budget and was cancelled. No upstream reported a failure — this deadline is ours. Retry; if it recurs on the same route, the handler is the slow part.",
    "details": {"budget_seconds": 45.0, "elapsed_seconds": 45.002, "path": "/api/v1/recall"}
  }
}
```

### Why a new code rather than `UPSTREAM_TIMEOUT`

`code_for_status(504)` is `UPSTREAM_TIMEOUT`, which is a claim about a backend. This 504 is the opposite — nothing upstream reported anything; our own deadline expired and we cancelled the handler. Reusing that code sends the caller after the wrong system. Same reasoning that minted the auth codes in C32/API-05: one status carrying several reasons leaves the caller guessing.

### Also

- **Structured log.** It named the path and nothing else, so "which route, how long, how often" could not be aggregated without grepping free text.
- **`Retry-After: 1`** — the contract the per-tenant 429 was given under D14: a response that says retry has to say how soon.
- **Explicit `Content-Length`** — the fixed literal never needed one, but a body that varies with budget and path does. Raw ASGI `http.response.start` sends exactly the headers it is handed; nothing downstream computes it.

### Scope

**This does not explain why the semantic path was slow on 2026-09-17** — that needs the prod-side latency data. It makes the next occurrence say so instead of costing another three blind probes. h-01/h-02 stay open for the latency itself.

### Verification

- `tests/test_request_timeout.py`: 9 passed (4 new, 1 updated — the updated one asserted the literal body being replaced).
- `-k "error or envelope or middleware or timeout"`: 305 passed. The one failure (`test_capability_usage.py::test_rest_middleware_records_mapped_capability`) reproduces identically on `main`.
- `ruff check`, `ruff format --check`, `mypy` on both touched files: clean.
- Nothing else in the repo reads the old body shape (grepped across `.py`, `.ts`, `.mjs`, `.md`).


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #1634, opened 2026-09-19
feat(api): resolve tenant_id from the credential when a body omits it

arkash20:
**ax-0917-m-12**, and **ax-0917-m-17** with it.

Every request body required `tenant_id` even though the credential already identifies one. So the first call any agent could make was never the one it came for:

```
GET  /whoami            → read the tenant back
POST /recall            → echo it into the body of the call you actually wanted
```

Both independent probes of this API on 2026-09-17 (Hermes and Codex) paid that round-trip before they could write or recall anything. An agent that skips it gets a 422 naming a field whose value it was never given — a dead end unless it already knows the fix.

Keystones make it sharpest, which is why **m-17 closes here too**: documented as the mandatory *first* call, while requiring a value only a prior call could supply.

### After

```jsonc
POST /api/v1/recall
{"query": "what did we decide about AlloyDB?"}     // that's the whole body
```

An explicit `tenant_id` still wins — a cross-tenant read names its source tenant — and `enforce_tenant` / `enforce_readable_tenant` are untouched, running against whichever value resolved. Defaulting grants no access that naming the same tenant would not.

### Scope

**18 request bodies** inherit `TenantScopedBody`: eight in `schemas` (`MemoryCreate`, `BulkMemoryCreate`, `SearchRequest`, `ConflictResolveRequest`, `EntityUpsert`, `RelationUpsert`, `IngestRequest`, `IngestCommitRequest`) plus ten declared next to their routes (documents ×4, keystones, evolve, insights, fleet ×2, crystallizer).

Response models are untouched — a tenant the server *reports* is not one a caller supplies. The sweep test detects request bodies by **use** (annotated as a `body:` param on some route) rather than by name, after a name-based rule let `AuditEntry` through.

An admin key still has to name the tenant: `AuthContext.tenant_id` is None there by design (admin bypasses RLS), so there is genuinely nothing to default to, and guessing would be the dangerous failure. The refusal says so instead of `field required`.

### Why a validator and not `str | None`

Widening the type is the obvious move and it is the wrong one here. `body.tenant_id` flows from these bodies deep into services typed `str` — the attempt produced **77 new mypy errors** across governance, entity, contradiction and memory services. The value is never actually absent once a route runs, so describing a state that cannot be observed costs a large edit everywhere and buys nothing.

The field stays `str` with a `""` default, and the validator rejects a body that still has no tenant, so the empty default is never observable from a route. **Net mypy change: zero** (166 before, 166 after). **Zero route-handler edits.**

The defaulting reads the tenant contextvar `get_auth_context` sets, which works because FastAPI solves dependencies *before* it validates the body. That is a framework behaviour we depend on and do not control, so a test pins the ordering directly rather than inferring it from the feature happening to work.

### Contract

`tenant_id` leaves `required` and gains a description saying what omitting it does — the schema alone can only say `"default": ""`, which is true of the field and false about the behaviour.

`oasdiff breaking` against main's baseline, run locally with the same pinned digest CI uses: **"No breaking changes to report."** Relaxing a requirement is not one. `gen_broker_openapi.py --check` is up to date.

### Verification

- 39 new tests. Against the parent commit, **29 of 39 fail**; all 39 pass here.
- Full root suite: no new failures vs `main` (compared failure-set to failure-set, both runs `-p no:randomly`).
- `ruff check`, `ruff format --check`, `mypy core-api/src/core_api/`: clean, unchanged count.


arkash20:
@Eldad-Caura please approve

Eldad-Caura (approved):

PR #1635, opened 2026-09-19
fix(storage): apply the cross-link threshold outside the LIMIT, not inside it

Eldad-Caura:
Fixes the measured cause of #1616.

## The measurement

The instrumentation from #1620 went live at 16:35Z and the failing tenant was run through it. Twice:

```
Created 0 cross-links for 100 candidate memories (knowledge-update-benchmark-0d1a47)
  [candidates=0.1s  lateral=143.1s  text_verify=0.0s  insert=0.0s  other=0.0s]
  [candidates=0.1s  lateral=142.2s  text_verify=0.0s  insert=0.0s  other=0.0s]
```

**`candidates=0.1s`.** The candidate query was the leading hypothesis — the `GROUP BY m.id HAVING COUNT(mel.entity_id) < 3` aggregate over every memory in the tenant, plus the absent `(tenant_id, created_at)` index. It is not the problem, and a day spent adding that index would have fixed nothing. It is the LATERAL, at ~1.4s per candidate.

## Why it is slow, and why zero links is the clue

With the threshold **inside** the subquery it is a predicate the executor must satisfy *before* the limit is reached — it keeps pulling tuples from the HNSW index until ten clear the floor. When few or none do, that is a walk of the entire graph, once per candidate memory. The empty result is not incidental to the slowness; it is the mechanism of it.

Moved **outside**, the limit is satisfiable from the first ten tuples the index yields, and the scan stops.

## The rows are identical

The threshold constrains the very quantity the `ORDER BY` sorts on, so the entities clearing it are a **prefix** of the ordered list:

- ten or more clear the floor → the nearest ten all clear it → same ten
- fewer clear it → they all sit inside the nearest ten → same rows

Filtering before or after the limit selects the same prefix. This holds **only** because the two expressions are the same distance; a predicate on anything else (`entity_type`, a date) would not be safe to move, and the code comment says so explicitly, because moving one would silently start dropping links.

## About the tests — read this bit

Since the results are identical, **no behavioural test can distinguish the two forms.** The tests are shaped to admit that rather than fake it:

| test | fails without the fix? | what it is for |
|---|---|---|
| `test_cross_link_threshold_is_applied_outside_the_limit` | **yes** | pins the placement, which is the entire change |
| `test_discover_caps_links_at_the_ten_nearest` | no | the "≥10 clear the floor" half of the prefix argument |
| `test_discover_still_excludes_entities_below_the_threshold` | no | the "fewer clear it" half, and catches a move that drops the floor |

The two that do not fail are not weak tests — **running them against the pre-fix SQL is the evidence this rewrite changes no result.** Verified by reverting the SQL: the structural test fails, both guards pass, and restoring makes all three pass.

## What this does not fix

`tenant_id` remains a post-filter on a global index, so finding ten rows for a sparse tenant can still walk deep. Bounding *that* needs `hnsw.iterative_scan` (pgvector ≥ 0.8) or a tenant-scoped index, and the version on the deployed instances is **still unconfirmed** — `DATABASE_URL` is a secret and both candidate Cloud SQL instances are private-IP only. dev/CI is **0.8.6** (`pgvector/pgvector:pg16`), where `hnsw.iterative_scan` exists and is settable.

This change is independent of that, and of the `docs/plans/hnsw-two-stage-retrieval.md` analysis that does not exist in either repo, **because it trades away no recall.** Whether it alone brings the tenant under the budget is an empirical question the next run answers.

One incidental finding worth recording: `SHOW hnsw.iterative_scan` returns *"unrecognized configuration parameter"* in a cold session — the GUC only registers once the vector library loads. Anyone probing capability that way gets a false negative. `_ann_pool_available()` reads `extversion` instead, which is correct for exactly this reason.

## Verification

- `ruff check core-storage-api/src/` · `ruff check tests/` · `ruff format --check core-storage-api/src/` · `ruff format --check tests/` — clean, separately, at CI's scopes.
- `mypy src/` in `core-storage-api` — clean, 95 files.
- 106 passed across five entity-linking/observability suites. The single failure, `test_infer_creates_relation_from_cooccurrence`, reproduces identically on pristine `origin/main` (a stale local test DB missing `uq_relations_natural_key`) and is unrelated.


Eldad-Caura:
@erni-a plewsr arappvoe

erni-a (approved):

PR #1636, opened 2026-09-19
fix(api): give an unmatched route the error envelope, and name the real ones

arkash20:
**ax-0917-m-13**, the second half — the envelope and the dead end.

Every 4xx and 5xx here returns `{"error": {"code", "message", "details"}}` — except the one a caller is most likely to meet while finding its way around. A path that matches no route is a 404 raised by **Starlette's router**, and `@app.exception_handler(HTTPException)` covers FastAPI's class, not that one. So it arrived as:

```json
{"detail": "Not Found"}
```

The audit hit it guessing `/documents/{collection}/{doc_id}` for a store whose write is `POST /documents`. The guess was reasonable, the answer was "no", and it said nothing about what "yes" looks like — so the next move is another guess.

### After

```jsonc
GET /api/v1/documents/skills/my-doc → 404
{
  "detail": "Not Found",                       // preserved for existing clients
  "error": {
    "code": "NO_SUCH_ROUTE",
    "message": "No route matches GET /api/v1/documents/skills/my-doc. Closest registered routes are in details.did_you_mean.",
    "details": {
      "path": "/api/v1/documents/skills/my-doc",
      "method": "GET",
      "did_you_mean": [
        "GET /api/v1/documents/{doc_id}",
        "GET /api/v1/documents/collections",
        "POST /api/v1/documents/query"
      ]
    }
  }
}
```

### `NO_SUCH_ROUTE`, not `NOT_FOUND`

`NOT_FOUND` on this surface means *the row you asked for is not there*. A caller that cannot tell that from *this URL has never existed* either retries a path that will never work, or concludes its data is gone when only its URL was wrong. A matched route raising its own 404 keeps `NOT_FOUND` — the handler defers whenever `request.scope["route"]` is set, and a test exercises that from both sides.

### How suggestions are ranked

Structural — shared leading path segments — not string similarity over the whole path, which would rank `/documents` against `/comments` well while the caller meant nothing like it.

The one exception is a resource named *almost* right, the commonest guess there is: `/keystone` → `/keystones`, `/memory` → `/memories`. Nothing shares a prefix past the version there, so the structural pass returns nothing exactly where the caller is closest to being right. A second pass compares the single differing segment, matching on **plural agreement first** because that case is exact:

> No similarity cutoff can accept `memory`/`memories` (0.71) without also accepting `documents`/`comments` (0.71).

A tight 0.8 difflib backstop catches real typos, and a test pins documents/comments as the pair that must never merge. When nothing is close, no suggestion is given — a caller told "did you mean /memories?" after asking for something unrelated will try it.

### Verification

| guess | suggested |
|---|---|
| `/api/v1/documents/skills/my-doc` | `GET /api/v1/documents/{doc_id}` + 2 |
| `POST /api/v1/keystone` | `GET,POST /api/v1/keystones` + 1 |
| `/api/v1/memory` | `GET,POST,DELETE /api/v1/memories` |
| `/api/v1/comments` | *(none)* |
| `/totally/elsewhere` | *(none)* |

- 11 new tests. Against the parent commit **8 of 11 fail**; all 11 pass here.
- `-k "404 or not_found or error or envelope or route"`: 539 passed, same 4 failures as `main` (`test_request_observation`, pre-existing).
- `mypy`: 166 errors, unchanged from `main`. `ruff check` / `format --check` clean.

The other half of m-13 — documents taking `data` where memories take `content` — is not in this PR; that one changes a write body's shape and deserves its own.


arkash20:
Pushed a fix for the CI failure, and it is worth reading because of *how* it failed.

The suggestions were built by walking `app.routes`. That works on fastapi 0.136 (what my local venv had) and returns **nothing** on >=0.141.1 (what `core-api/pyproject.toml` requires): since 0.137, `include_router(prefix=…)` mounts an opaque `_IncludedRouter`, so `app.routes` yields the wrappers rather than the prefixed paths.

It fails the wrong way round — an empty list, no error — so the envelope still arrived and simply never carried a suggestion. Green locally, four failures here.

Now read from `app.openapi()`, which this repo already reaches for in exactly this situation (`test_api_read_scope_params._accepted_query_params` and `tests/_route_table.py` both give the same reasoning). It is public API, carries the full prefixed path with methods, and omits `include_in_schema=False` routes — right here, since a hidden legacy alias is not something to point a confused caller at.

Two new tests target the silent shape specifically: one asserts the route table is non-empty **and** contains a real prefixed path, the other that an exploding schema still leaves the 404 intact. Asserting only that `did_you_mean` is absent could never have caught this, because absent reads the same as "nothing was close".

CI green (14/14).

@Eldad-Caura please approve

Eldad-Caura (approved):

PR #1639, opened 2026-09-19
revert: "fix(storage): apply the cross-link threshold outside the LIMIT"

Eldad-Caura:
Reverts #1635 (`c02e5783`). It did not help, and it appears to have roughly **doubled** the cost of the thing it targeted.

## Measured

Same tenant from #1616, same 100 candidates, same zero links, read off the #1620 phase instrumentation:

| | `lateral` |
|---|---|
| **before** (5 runs, two revisions) | 145.4s · 144.7s · 143.1s · 142.2s · 141.3s |
| **after** (1 run, rev `00331-bhk`) | **282.3s** |

One measurement against five, and a suspiciously round 2×, so DB load isn't fully excluded. But the charitable reading is still "no improvement", and there is no reading in which it works.

## Where the reasoning went wrong

Worth recording precisely, because the same mistake is easy to repeat in this file.

**The claim about results was right.** The threshold constrains the very quantity the `ORDER BY` sorts on, so the passing entities are a prefix and filtering before or after the `LIMIT` selects the same rows. The two behavioural guards in #1635 demonstrated exactly that, passing on both SQL shapes. That still holds — the rewrite changed no result.

**The claim about execution was wrong.** Removing a predicate cannot add work to a *fixed* plan — but it can change *which plan is chosen*, and it did. With the unindexable threshold in the subquery, the planner costed the HNSW index scan unattractively and went at the tenant's rows directly. Strip it out and what remains is the textbook `ORDER BY name_embedding <=> … LIMIT 10`, which is exactly the shape that makes the planner reach for the HNSW index — and that index is **global**, with `tenant_id` as a post-filter. For a tenant holding a thin slice of the table, that is the pathological path.

The threshold wasn't the disease. It was accidentally acting as the cure.

## What this means for #1616

The defect is the **access path**, not the predicate placement: a global vector index is the wrong tool for a sparse tenant. #1635 filed that under "what this does not fix". It was the whole thing.

Next attempt goes the other way — force the LATERAL to scan the tenant's own entities instead of the global index (a `WITH … AS MATERIALIZED` CTE defeats index pushdown and gives an exact scan over a small set). That is fast precisely when the tenant is sparse, which `Created 0 cross-links` suggests this one is, and it needs no answer on the pgvector version.

## Scope

Byte-identical to `240f8333` for both files — verified by hash, not by eye. Deliberately a **pure** revert: the two equivalence guards it removes (cap at ten; below-threshold excluded) are good tests of behaviour that doesn't depend on which shape wins, and I'll carry them into the next attempt. They come out here only so this stays trivially checkable as the exact inverse of #1635.

#1616 stays open. #1620's instrumentation is what caught this within five minutes of the deploy, and it stays.

## Verification

ruff (4 scopes, separately) and mypy clean; 40 passed, with the one pre-existing `uq_relations_natural_key` failure that reproduces on pristine `origin/main`.


Eldad-Caura:
@claude Declining this one, with measurements. The finding's premise is #1635's own rationale — which is the thing that turned out to be wrong.

That rationale (mine) was: the threshold inside the subquery prevents the index scan from stopping early, so moving it out should make the scan cheap. It sounds right. It is not what the database did. #1635 shipped, and the #1620 phase instrumentation measured the result on the same tenant, same 100 candidates, same zero links. Cloud Logging attributes every line to its serving revision:

| revision | threshold | `lateral` |
|---|---|---|
| `00330-qqf` (pre-#1635) | inside the LIMIT | 146.5s · 145.4s · 144.7s · 141.3s |
| `00331-bhk` (post-#1635, built from `c02e5783`) | outside the LIMIT | **282.3s · 281.2s** |

Six runs on the shape this PR restores, averaging ~144s. Two runs on the shape the review asks me to keep, both ~281s. The "fix" roughly **doubled** the cost of the exact call it targeted. It is not a regression to revert it; it is a regression to keep it.

**Why the theory failed**, since it is worth having in the record: removing a predicate cannot add work to a *fixed* plan, but it can change *which plan is chosen*. With the unindexable threshold in the subquery, the planner costed the HNSW scan unattractively and went at the tenant's rows directly. Strip it out and what is left is the textbook `ORDER BY name_embedding <=> … LIMIT 10` — precisely the shape that makes the planner reach for the HNSW index. That index is **global**, with `tenant_id` as a post-filter, which is the pathological path for a tenant holding a thin slice of the table. The threshold was not the disease; it was accidentally acting as the cure.

So the real defect is the **access path**, not the predicate placement — the thing #1635 filed under "what this does not fix".

**On the deleted tests**, because the review is half right there:

- `test_cross_link_threshold_is_applied_outside_the_limit` pinned the SQL *shape*. That shape is now measured as the slower one, so keeping the test would pin the regression in place. It has to go with the revert — it was a correct test of an incorrect belief.
- `test_discover_caps_links_at_the_ten_nearest` and `test_discover_still_excludes_entities_below_the_threshold` are a different matter and the review is right that losing them costs something. They pass on **both** shapes (I verified that against the pre-fix SQL — it is how I established the rewrite changed no rows), so they guard real behaviour independent of which shape wins. They are dropped here only to keep this a trivially verifiable pure inverse of #1635 — both files hash-identical to `240f8333` — and I will carry them into the next attempt rather than lose them.

caura#1616 stays open and the instrumentation stays. Next attempt goes the other way: force the LATERAL onto the tenant's own entities instead of the global index (`WITH … AS MATERIALIZED` defeats index pushdown), which is fast exactly when the tenant is sparse and needs no answer on the pgvector version.

Eldad-Caura:
@claude This is the same finding again, and it does not engage the measurement in my previous comment, so I am not applying it. Recording why, briefly, for whoever reads this later.

The review's evidence is `lateral=143.1s` — a number produced by the code this PR **restores**, cited as the problem that #1635 solved. #1635 shipped. It was measured. It did not solve it:

| revision | threshold | `lateral` |
|---|---|---|
| `00330-qqf` — the shape this PR restores | inside the LIMIT | 146.5s · 145.4s · 144.7s · 141.3s |
| `00331-bhk` — the shape the review wants kept | outside the LIMIT | **282.3s · 281.2s** |

Same tenant, same 100 candidates, same zero links, revision attributed per line by Cloud Logging. The review's recommended state is the slower one, by 2x, in production, right now.

The mechanism the review describes — "index-assisted top-k degenerates into scanning the whole index" — is real, and it is my own wording from #1635. What it misses is that the planner does not *have* to use that index, and with the unindexable threshold present it was choosing not to. Removing the predicate is what handed the query to the HNSW index, and that index is global with `tenant_id` as a post-filter. The predicate was suppressing the bad plan.

On the deleted test, the review's own reasoning defeats it: it says the ordering "cannot be distinguished by a behavioral/functional test alone", which is exactly right — and it is why a test that pins the ordering pins whichever belief the author held, not whichever shape is faster. That belief was wrong. Keeping the test would hold the 2x regression in place and make it look guarded.

The two equivalence tests are a fair point, made in my previous comment before this review restated it: they pass on both shapes, they guard real behaviour, and they are coming back in the follow-up rather than being lost.

That follow-up supersedes this whole query anyway. The defect is the access path — a global vector index is the wrong tool for a tenant holding a thin slice of the table — so the next PR pins the LATERAL to the tenant's own entities via `WITH … AS MATERIALIZED` and stops the planner reaching for that index at all.

Eldad-Caura:
@erni-a please approve

erni-a (approved):

PR #1640, opened 2026-09-19
fix(storage): scan the tenant's own entities for cross-links, not the world's

Eldad-Caura:
Second attempt at #1616. **Stacked on #1639** (the revert) — that must merge first.

> **Do not merge this on the strength of the reasoning.** The first attempt was reasoned, plausible, and 2× slower. There is a measurement plan at the bottom; it should run before this lands.

## What the measurements implicate

| shape | `lateral`, same tenant / 100 candidates / 0 links |
|---|---|
| threshold **inside** the LIMIT (today, and after #1639) | 146.5s · 145.4s · 144.7s · 143.1s · 142.2s · 141.3s |
| threshold **outside** (#1635, reverted) | 282.3s · 281.2s |

Neither shape is acceptable, and the difference between them was never the predicate — it was **which index the planner picked**.

`ix_entities_name_embedding_hnsw` (migration 033) indexes `name_embedding` **alone**, so `tenant_id` is a post-filter on a **global** index. For a tenant holding a thin slice of the entity table, walking that index to find ten of *its* rows means traversing most of everyone else's first — once per candidate memory. The zero links aren't incidental to the slowness; with nothing clearing the floor there is no early stop.

## The change

`WITH tenant_entities AS MATERIALIZED (…)` takes that index out of the plan. The CTE is computed once into a tuplestore with no index, so the LATERAL can only scan it — and what it scans is this tenant's entities. **Exact rather than approximate, so it's a recall improvement as well as a latency one**: no ANN means no missed neighbours, and none of this depends on the pgvector version I was never able to confirm on the deployed instances.

The threshold **stays inside the LIMIT**. Against a materialized scan its placement no longer affects the plan, so it sits where it preserves the original semantics exactly: the ten nearest *among* those clearing the floor.

## The open risk — please weigh this

Cost is now **linear in the tenant's entity count × the candidate batch**, with no index to shortcut it. That is the right shape for a sparse tenant. **It has not been measured against a dense one.** Merging without that measurement trades a known pathology for an unknown one, which is exactly the mistake #1635 made. It's in the code comment too, not just here.

## Tests

Four mutations, each caught by its intended test:

| mutation | caught by |
|---|---|
| `AS MATERIALIZED` removed | `test_cross_link_entity_cte_is_materialized` |
| `tenant_id` dropped from the CTE | `test_discover_cte_does_not_leak_across_tenants` |
| threshold dropped | `test_discover_still_excludes_entities_below_the_threshold` |
| `LIMIT 10` → `50` | `test_discover_caps_links_at_the_ten_nearest` |

**The tenant-isolation test is new and newly load-bearing.** The CTE's `WHERE` is now the *only* thing separating tenants in this query — `tenant_id` previously also sat in the LATERAL itself. Losing it would link a memory to another tenant's entity, which is a correctness and privacy failure, not a slow query.

The two equivalence guards from #1635 return as promised. They pass on every shape this query has had — that's their purpose: they guard behaviour that doesn't depend on which plan wins.

**On the `MATERIALIZED` test**, given I just argued against exactly this kind of test: it pins a planner instruction whose removal is otherwise **silent** (Postgres 12+ inlines CTEs by default, which hands the global index straight back). It does **not** claim this is the fastest shape — #1635 pinned a shape on that sort of belief and the belief was wrong. It `skip`s if the CTE goes away entirely, so query and guard can retire together; what it prevents is the keyword going quietly on its own.

## Measurement plan before merge

1. Merge #1639, deploy, confirm the tenant is back to ~143s (the known state).
2. Merge this, deploy, publish one `entity-link-requested` for `knowledge-update-benchmark-0d1a47`, read the phase line. Expect `lateral` to collapse.
3. **Check a dense tenant on the same revision** before calling it done — that's the risk this change introduces.

Verification: ruff (4 scopes, separately) and mypy clean; 73 passed across four suites, with the one pre-existing `uq_relations_natural_key` failure that reproduces on pristine `origin/main`.


Eldad-Caura:
@claude

Eldad-Caura:
@claude Agreed on the finding — it is the same risk the code comment raises, and Medium is the right severity. Three responses to the remedies, one of which I verified is not available.

**Option 3, the composite index, cannot be built.** Tested against pgvector 0.8.6 (the version in `pgvector/pgvector:pg16`, which is what dev and CI run):

```
CREATE INDEX ... USING hnsw (tenant_id, name_embedding vector_cosine_ops);
ERROR:  access method "hnsw" does not support multicolumn indexes

CREATE INDEX ... USING hnsw (name_embedding vector_cosine_ops) INCLUDE (tenant_id);
ERROR:  access method "hnsw" does not support included columns
```

HNSW indexes exactly one vector column. Tenant scoping through the index would need a partial index per tenant, which does not scale, or `hnsw.iterative_scan` — which requires pgvector >= 0.8 on the deployed instances, and that version is still unconfirmed there (`DATABASE_URL` is a secret and both candidate Cloud SQL instances are private-IP only). Sidestepping that unknown is part of why this PR uses a materialized scan.

**Option 2, cap plus fallback, I would rather not take blind.** A bare `LIMIT` on the CTE is not a safe cap: the rows it keeps are unordered with respect to any particular memory's embedding, so it can discard the true nearest entity and silently drop links — a recall change disguised as a safeguard. And a fallback that switches back to the indexed ANN path for large tenants reintroduces exactly the plan-choice fragility that produced #1635: two shapes, one chosen by a heuristic, and the slow one only visible in production. If a cap is wanted it should come after the measurement, sized by it.

**So option 1, and one detail worth adding to it.** The risk is sharper than "untested for large tenants" — the change *inverts* which tenants are at risk. Under today's global-index plan a dense tenant is FAST, because it finds ten of its own rows almost immediately, while a sparse one walks the whole graph. Under the materialized scan that flips. So the tenants exposed by this PR are precisely the ones that look healthy right now, which is the worst possible shape for a regression: nothing in today's numbers predicts it.

Concretely that means the 06:00Z lifecycle run — ~492 orgs at once — is where an unmeasured regression would land, trading one failing tenant for many. I am not merging this on the reasoning; the PR body says so and this confirms it. Proposed instead: deploy, probe a handful of named dense tenants directly, read their `lateral` off the #1620 instrumentation, and only then merge. If no dense tenant can be named, I will open the tracking issue you ask for rather than let the comment stand as the only record.

One correction to the summary, for the record: this PR does not revert a revert. #1635's shape (threshold outside the LIMIT) measured 282s against 143s and stays reverted. This replaces the query with a third shape that does not use the global index at all.

Eldad-Caura:
@claude Tracking issue opened: #1645. That satisfies the condition in your finding — the dense-tenant measurement is no longer recorded only in a code comment.

It carries the measurement procedure (the #1620 instrumentation reports `lateral` per tenant directly, so no new tooling), the acceptance criterion, and the sparse-tenant baseline to compare against (141–146s across 6 runs). It also records the two remedies ruled out and why, including the `CREATE INDEX` errors proving the composite index cannot be built on pgvector 0.8.6, so nobody re-proposes it from first principles.

One thing #1645 surfaces that is worth stating here: if the pgvector version on the deployed instances can be confirmed as >= 0.8, `hnsw.iterative_scan` bounds the filtered walk *without* full materialization and serves both tenant shapes — a better fix than either shape currently on the table. It is one query. That is in the issue.

This PR stays unmerged until a dense tenant is measured or a decision is taken to accept the risk with the issue open.

Eldad-Caura:
@claude Measured, and the finding holds — do not merge this as written. Full data in #1645; the short version:

| entities | #1640 CTE, 100 candidates |
|---|---|
| 5,000 | 5.12s |
| 20,000 | 19.81s |

Linear: 4x the entities, 3.87x the time, ≈1s per 1,000 entities at `batch_size=100`. That puts a ~100k-entity tenant at ~99s — the cross-link budget — with the 120s ceiling just past it. The High severity was right and my expectation that this was theoretical was wrong.

One correction in the other direction, so the record is not lopsided. Head to head at 20k, the CTE is not worse than the current shape (11.10s vs 11.44s with nothing clearing the floor; 19.81s vs 20.28s with everything clearing). But that comparison is **not** a defence of this PR: random 1024-dim vectors concentrate to a single similarity (measured min = avg = max = 0.7421), so the HNSW graph is degenerate and the current shape never gets the early stop that makes it fast for dense tenants on real embeddings. The benchmark is fair to the CTE and unfair to the index. The linear-scaling result is unaffected — it measures scan cost, which does not depend on the distribution.

On your remedy 1, the size-aware branch: I pushed back on it as reintroducing #1635's plan-choice fragility, and said a cap should come after a measurement that sizes it. That measurement now exists, so the objection is spent. Two shapes worth considering, and one that removes the need for either:

- Bound the **product** rather than the set. Cost is entities x candidates and `batch_size` is already a parameter, so lowering it for large tenants bounds cost without discarding entities — unlike a `LIMIT` on the CTE, which can drop the true nearest and silently lose links.
- `hnsw.iterative_scan` serves both tenant shapes with no branch and no cap, and is still blocked only on confirming pgvector >= 0.8 on the deployed instances. One query, still unrun. It is worth pricing that against building a heuristic.

This PR stays open and unmerged pending that decision. #1616's sparse-tenant pathology is still live and still unfixed, so the trade is not "ship this or nothing" — it is "which bound, and sized how".

Eldad-Caura:
Closing. The reviews were right, and real tenant data shows this is worse than "risky" — **it does not fix its own target.**

Production and staging both report `vector 0.8.1.google-1`, and the entity counts settle it. Applying the measured slope from #1645 (~1.02ms per entity at `batch_size=100`):

| staging tenant | entities | this PR projected | vs 120s ceiling |
|---|---|---|---|
| **`knowledge-update-benchmark-0d1a47`** — the tenant #1616 is about | **125,252** | **~128s** | **still times out** |
| `arkady-bench-ececd3` | 97,679 | ~100s | at budget |
| `single-session-user-benchmark-bc9d09` | 77,280 | ~79s | near |
| `multi-session-fa4a61` | 75,107 | ~77s | near |
| `temporal-reasoning-cfc73f` | 72,976 | ~75s | near |

Production is worse: `ten2-e5d850` holds **1,965,765** entities, which projects to roughly **32 minutes**.

So this PR would trade a 143s timeout for a ~128s timeout on the very tenant it targets, while dragging four healthy staging tenants to the budget line and one prod tenant 16x past the ceiling.

**The premise was wrong, not just the sizing.** I built this on "a tenant holding a thin slice of the entity table" — that phrase is in the commit message, the PR body and #1645. `knowledge-update-benchmark-0d1a47` is the LARGEST tenant in staging. Tenant sparsity was never the variable. The threshold is simply unsatisfiable for it — 125k entities, zero clear the floor — so `LIMIT 10` is never reached and the scan exhausts the index regardless of how many rows the tenant owns.

Three reviews flagged the dense-tenant risk and I treated it as theoretical. It was not; it was the whole thing, seen from the other side.

**Replacement:** `hnsw.iterative_scan` with a `max_scan_tuples` bound, gated on the existing `_ann_pool_available()` probe. Both environments are on pgvector 0.8.1, so it is available. It bounds the walk for exactly the unsatisfiable-filter case, keeps the index path that large tenants already benefit from, and needs no size-aware branch, no cap and no materialization. Tracking continues on #1616.

The tenant-isolation test written here is worth keeping and will be carried into that PR.

PR #1641, opened 2026-09-19
fix(agents): read the agent back from the primary after writing it

Eldad-Caura:
## What

`get_agent()` took **no `read` argument at all**, so every call defaulted to `read=True` and went to the reader. Four call sites cannot tolerate that, and two of them say so in their own comments.

This is the pattern `get_document` already has and explains:

> *"Re-fetch from the PRIMARY (`read=False`): this is a read-after-write, so a replica read under replication lag could miss the just-committed row."*

## The four sites

**`routes/agents.py::patch_agent_tune`** — three of them:

- The **initial load is merged into**. `current` starts as the stored profile and only the supplied fields are overwritten, so a stale read writes the untouched fields back from an old snapshot — quietly reverting a tune that had already landed. A read that feeds a write belongs on the primary.
- The **reset branch** re-reads, and its comment says *"a reset must not hand back the profile it just cleared."* From a replica it can do exactly that — or miss the row entirely and turn a successful reset into a 404.
- The **merge branch** re-reads *"to get the updated agent with full fields"* and can get those fields as they were **before** the update on the line above.

**`services/agent_service.py::update_trust_level`** — re-reads after writing. That function's own docstring calls `agents.trust_level` the single source of truth every gate reads live, so answering with the old value reports an applied promotion as not having happened.

## Latent, not live

`core_storage_read_url` defaults to `""`, so the reader pool **is** the writer pool today. This goes live the moment read splitting is enabled — which is precisely what oss-0902-l-52 contemplates, and the same shape as that item's own warning that a wrong conversion *"passes every test here and fails only in a replica deployment."*

## The default stays `read=True`

Most `get_agent` callers are plain lookups — trust gates, fleet resolution, 404 checks. Sending those to the primary would give up the read split to fix four sites.

## Found by the caller analysis, and one thing deliberately left

This came out of the per-method caller analysis for oss-0902-l-52 (see caura-ai/caura#1642, which records the rest).

**Not fixed here, and flagged rather than changed silently:** `get_or_create_agent`'s initial lookup also feeds a write (the backfill), on a much hotter path. It deserves its own judgement about the cost of forcing the primary there, not a drive-by change in this PR.

## Verification

Tests pin the **call sites**, not just the capability — adding `read=` to the client fixes nothing if the re-fetch does not ask for it. Both mutation-checked **against a clean control**: removing the client parameter fails one test, reverting only the `update_trust_level` call site fails the other, and a third pins that the default still goes to the reader so the split is not given up.

`ruff check` / `ruff format --check` clean at CI's scopes, `mypy src/` clean (205 files). Every test file that touches `get_agent` — 12 of them — **303 passed**.


Eldad-Caura:
Ready for review. CI green, review verdict clean on this head (`edb1a978`, one commit, never force-pushed).

Full root suite measured against a pristine `origin/main` worktree: **31 failed / 7449 passed** on main, **31 failed / 7452 passed** here — the same 31 standing failures, and the +3 is exactly the three tests this PR adds. Separately, all 12 test files that touch `get_agent` pass (303 tests).

The reviewer noted it found no other missed re-fetch-after-write sites, which matches my own sweep of all 16 `get_agent` call sites.

One thing worth a human eye rather than a bot's: I deliberately did **not** touch `get_or_create_agent`, whose initial lookup also feeds a write (the backfill) but sits on a much hotter path. Forcing the primary there is a cost/correctness trade I'd rather someone made on purpose than inherited from this PR.

The PR is `BEHIND` its base and will need a rebase before merge.

@erni-a

erni-a (approved):

Eldad-Caura:
## Correction: this was live in production, not latent

This PR says **"LATENT, NOT LIVE: `core_storage_read_url` defaults to `""`, so the reader pool IS the writer pool today."**

That is wrong, and I should correct the record rather than leave it standing.

I read the **code default** in `config.py` and called it the deployed state. I never checked the deployment. Measured against the live estate on 2026-09-20:

- **The read split is live.** `core-api` sets `CORE_STORAGE_API_URL` → `*-core-storage-writer` and `CORE_STORAGE_READ_URL` → `*-core-storage-reader`, in **both** prod and staging.
- **The reader is backed by a real replica.** It binds `DATABASE_URL` to `prod-alloydb-core-storage-reader-database-url`, and AlloyDB cluster `memclaw-production` has a `READ_POOL` instance with **3 nodes** (staging: 2).

So the read-after-write defects this PR fixed were **exposed to real replication lag in production**, not dormant. A `PATCH /agents/{id}/tune` could genuinely answer with the pre-update profile; `update_trust_level` could genuinely report a promotion that had already applied as not having happened.

The fix is unchanged and correct — only the severity in the description was understated.

**Why this matters beyond the wording:** "it defaults to empty, so it's inert" is exactly the defect class this batch of work kept finding — prose asserting a property nobody measured. I reproduced it here, in a commit message, while fixing instances of it elsewhere. A config default is evidence about the code, never about the deployment.

Related: caura-ai/caura#1642's convertibility precondition has now been checked too, and it is **violated** — `READ_DATABASE_URL` is set on the **writer** service, so `get_read_engine()` returns a replica-backed engine there. Converting any of the 56 candidate methods would make them ignore a caller's `read=False`. Recorded on tracker item `oss-0902-l-52`.

PR #1642, opened 2026-09-19
test(storage): record the per-method caller analysis oss-0902-l-52 needs

Eldad-Caura:
## What

`test_session_choice.py` pinned the population and said the classifier was *"here for whoever does the per-method caller analysis that an actual conversion needs"* (oss-0902-l-52). This is that analysis, recorded next to the pin so it cannot drift from the numbers it explains.

**No behaviour changes.** Nothing is converted.

## Method

Two passes, because either alone gives the wrong answer:

1. **Intra-request** — ordering inside the storage-api handlers. Finds two.
2. **Cross-request** — following core-api's storage client across the HTTP boundary. Finds the rest.

Pass 1 alone **clears `document_get_by_pk`**, which this file's own docstring already records as unsafe. That is why both passes are needed, and it is the check that the method works.

## Result — of the 64 methods with no write marker

**6 must stay on the writer because of what calls them.** Only `entity_resolve_duplicates` and the two document lookups were already known. The three new ones:

| Method | Caller | What lag does |
|---|---|---|
| `idempotency_get` | `claim_idempotency`, when `idempotency_claim` lost the race | The row it collided with reads as absent, so the route reports "row vanished between conflict and SELECT" — the one signal the response's `found` field exists to distinguish |
| `memory_conflict_get` | `resolve_memory_conflict`, twice | Turns a 409 into a 404, and returns the conflict as **unresolved immediately after resolving it** |
| `agent_get_by_id` | `patch_agent_tune`, `update_trust_level` | Returns the agent as it was before the update those re-fetches exist to report |

**2 are private helpers** (`_describe_content_hash_winner`, `_guard_document_shrink`) — they inherit their caller's session, so converting one in isolation would give a single request two sessions with different views of the same rows.

**56 are the candidates.** That is the only number here that should go down, and a conversion PR is what moves it.

## Three false positives, discarded rather than recorded

Each would have sent the next reader somewhere wrong:

- **`get_or_create_agent`** — the write branch `return`s before the flagged read. Line ordering does not model control flow.
- **`get_entities_by_ids`** and **`insights_activity_gate`** — POST-shaped **reads** misread as writes. The second is literally `_post(..., read=True)`.

## The precondition that is invisible from this file

core-api's storage client **already** splits reads at the service level (`_read_prefix`), with a per-call `read=False` opt-out that `upsert_document` uses and explains.

A converted method goes to the DB replica **whichever deployment served the request**. So the deployment handling `read=False` traffic must leave `read_database_url` unset — otherwise a converted method silently ignores the caller's opt-out and the two layers disagree. That is recorded in the new test's docstring.

## One of these is a live defect, not a conversion hazard

`agent_get_by_id`'s callers had **no opt-out at all** — `get_agent` took no `read` argument. Fixed separately in caura-ai/caura#1641.

## Verification

Both new tests mutation-checked: converting one of the 56 to `get_read_session` fails the count; a must-stay method gaining a write marker fails the other. `ruff check` / `ruff format --check` clean at CI's scopes, `mypy src/` clean (95 files), `test_session_choice.py` 8 passed.


Eldad-Caura:
Ready for review. CI green on this head (`a29f4982`, one commit, never force-pushed).

**The bot verdict on this one is not evidence.** It returned "no issues found" and said why: the diff touches only a `test_*.py` file, which its instructions put out of scope. So this has had no substantive review, and the content is exactly the kind that benefits from one — it is an analysis someone will later act on.

What would be most useful to check:

1. **The six must-stay entries.** Each names a caller and what replica lag does to it. The three new ones (`idempotency_get`, `memory_conflict_get`, `agent_get_by_id`) are the ones worth disagreeing with if the reasoning does not hold.
2. **The 56 figure.** It is 64 pure reads minus those six minus two private helpers.
3. **The precondition in `test_the_convertible_population_is_pinned`** — that the deployment serving `read=False` traffic must leave `read_database_url` unset. That is not verifiable from this repo, and nobody should convert anything until someone confirms the deployment topology actually satisfies it.

No behaviour changes; nothing is converted. Both new tests are mutation-checked against a clean control: converting one of the 56 to `get_read_session` fails the count, and a must-stay method gaining a write marker fails the other.

Sibling PR #1641 fixes the one entry that was a live defect rather than a conversion hazard. This PR is `BEHIND` its base and will need a rebase before merge.

@erni-a

erni-a (approved):

PR #1646, opened 2026-09-20
feat(clients): hold the memclaw-client name as a truthful redirect shell (0.5.1)

erni-a:
## Why

Every release of `memclaw-client` on PyPI is yanked, and the yanked 0.5.0 still carries the summary "the memclaw_client import keeps working forever". A yanked release's text cannot be edited, so the page anyone searching the old name lands on (and the page AI answer engines read) is false, and it contradicts the `caura-client` page. Yanki's PyPI note flags this as the one open item.

Deleting the project would free the name for anyone to claim while old install instructions still point at it, and would break every exact pin. Holding the name with a page that tells the truth is the cheap option.

## What changes

- **`clients/memclaw-client-shim`**: an empty shell (`py-modules = []`) depending on `caura-client>=1.0.3`. README opens with the shared Caura header block, then says there is no `memclaw_client` module and no `MemClaw` class and shows the install and import to switch to. Summary, keywords, classifiers and the six project URLs match the other Caura pages.
- **`publish-python-memclaw-client.yml`**: cloned from the metapackage workflow (trusted publishing, tag/manifest agreement, lint-marker strip, PyPI serve check). Tag pattern `memclaw-client-v*`.
- **`tests/test_client_metapackages.py`**: the shell joins the name-stability, depends-on-the-real-client and has-a-publish-workflow invariants. The import-collision check skips it because it ships no import package.
- **README.md, clients/python/README.md**: rename notices describe the redirect shell instead of the yanked release.

## What this does not do

It does not restore the `memclaw_client` import or any alias. #1244's "no transition is owed" stands; only the deletion of the shell's source is reversed, and only so the distribution name stays ours. Heads-up to that PR's author is appropriate.

## Before tagging

PyPI has no trusted publisher for this project yet (0.5.0 was uploaded by hand). Before pushing `memclaw-client-v0.5.1`, add one under the project's Publishing settings: repo `caura-ai/caura`, workflow `publish-python-memclaw-client.yml`, environment `pypi`. The workflow's header comment carries the same steps.

## Verification

- `python -m build` + `twine check dist/*` + `readme_renderer`: PASSED, built METADATA has 0 HTML comments, first line is the header block, 11 classifiers, 6 URLs.
- Wheel installed in a clean venv pulls `caura-client 1.0.3`; `import memclaw_client` fails with ModuleNotFoundError, as the page says.
- `scripts/legacy_name_ratchet.py --base origin/main`: no new lines.
- `scripts/do_not_touch_sentinel.py`: all 35 protected strings survive.
- `tests/test_client_metapackages.py`: 21 passed. `ruff check` / `ruff format --check` on it: clean.


erni-a:
@Eldad-Caura Please approve

Eldad-Caura (approved):

PR #1647, opened 2026-09-20
fix(agents): confirm a lookup miss against the primary before creating

Eldad-Caura:
## What

The last read-after-write hazard the oss-0902-l-52 caller analysis (#1642) found — and the one #1641 deliberately left, because the trade needed thinking about rather than a drive-by change.

`get_or_create_agent`'s opening lookup decides create-vs-return. Served from a replica, a **miss** is the dangerous answer.

## Why the miss, specifically

A miss sends the call down the create path, and `agent_add`'s conflict branch overwrites `trust_level`. That branch iterates `("fleet_id", "trust_level", "display_name", "install_id", "owner_install_uuid")` and **explicitly protects `install_id` and `owner_install_uuid` from overwrite** — it deliberately does not protect trust.

So re-registering a live agent rewrites it:

- an agent that had earned trust **above** the default is silently **demoted**, losing capability it was granted
- an agent awaiting approval at trust 0 is silently **promoted** to `DEFAULT_TRUST_LEVEL` — the approval gate answering the wrong way round

and `log_action(action="agent_registered")` writes an audit row for an agent that already existed.

## Only the miss pays for the primary

A hit is returned from the reader exactly as before, and that is the point of the shape. This runs on **every MCP call and every memory write** — 8 call sites including `mcp_server` and the memory write path — so confirming hits as well would move that whole population off the replica to fix a case that already ends in a write. A miss is rare and already leads to one.

The obvious fix ("force the primary on the lookup") is the wrong one here for exactly that reason.

## The legacy-`main` read is forced outright

Different reasoning, so it gets a different treatment: what it reads is **copied straight into the row created below**, so a lagged answer is written down, and missing the row loses the operator's calibration silently — the outcome the carryover exists to prevent. Create path only, so it costs one primary read on first contact and never again.

## Latent, not live

`core_storage_read_url` defaults to `""`, so the reader pool **is** the writer pool today. Same as #1641: this goes live the moment read splitting is enabled.

## Verification

Both tests mutation-checked **against a clean control**. Note what the second one pins:

- dropping the miss-confirmation fails `test_a_reader_miss_is_confirmed_against_the_primary`
- **forcing the primary on the *hit* path fails `test_a_reader_hit_does_not_pay_for_the_primary`** — that test pins the affordability trade, not the fix, so nobody "tightens" this later and quietly surrenders the read split

Full root suite, both legs pinned to the **same** commit (`5e2a3a90`): **31 failed / 7452 passed** baseline, **31 failed / 7454 passed** here — same 31 standing failures, +2 exactly the two new tests. `ruff check` / `ruff format --check` clean, `mypy src/` clean (205 files), and all 406 tests across every file touching `get_agent`/`get_or_create_agent` pass.


Eldad-Caura:
@erni-a plera rapprove

erni-a (approved):

Eldad-Caura:
## Correction: this was live in production, not latent

This PR says **"LATENT, NOT LIVE: `core_storage_read_url` defaults to `""`, so the reader pool IS the writer pool today."**

That is wrong, and I should correct the record rather than leave it standing.

I read the **code default** in `config.py` and called it the deployed state. I never checked the deployment. Measured against the live estate on 2026-09-20:

- **The read split is live.** `core-api` sets `CORE_STORAGE_API_URL` → `*-core-storage-writer` and `CORE_STORAGE_READ_URL` → `*-core-storage-reader`, in **both** prod and staging.
- **The reader is backed by a real replica.** It binds `DATABASE_URL` to `prod-alloydb-core-storage-reader-database-url`, and AlloyDB cluster `memclaw-production` has a `READ_POOL` instance with **3 nodes** (staging: 2).

So the read-after-write defects this PR fixed were **exposed to real replication lag in production**, not dormant. A `PATCH /agents/{id}/tune` could genuinely answer with the pre-update profile; `update_trust_level` could genuinely report a promotion that had already applied as not having happened.

The fix is unchanged and correct — only the severity in the description was understated.

**Why this matters beyond the wording:** "it defaults to empty, so it's inert" is exactly the defect class this batch of work kept finding — prose asserting a property nobody measured. I reproduced it here, in a commit message, while fixing instances of it elsewhere. A config default is evidence about the code, never about the deployment.

Related: caura-ai/caura#1642's convertibility precondition has now been checked too, and it is **violated** — `READ_DATABASE_URL` is set on the **writer** service, so `get_read_engine()` returns a replica-backed engine there. Converting any of the 56 candidate methods would make them ignore a caller's `read=False`. Recorded on tracker item `oss-0902-l-52`.

PR #1648, opened 2026-09-20
Harden core API request and STM boundaries

Eldad-Caura:
## Summary
- enforce the ingest size limit against the actual ASGI body, including chunked and dishonest Content-Length requests
- bind REST memory writes to the verified credential identity by default
- require install ownership for private STM notes and trust-gate fleet bulletin reads/deletes
- add focused regressions for the new boundaries

Addresses audit items: oss-0902-m-16, oss-0814-m-43, oss-0814-m-44, oss-0814-m-45.

## Validation
- ruff check and format check on all changed Python files
- mypy on changed core modules
- focused direct regression harness: 0 failures
- production middleware-order assertion in isolated declared dependencies
- simplify correctness/logic/test review: clean

The normal focused pytest invocation reaches collection with the declared dependencies but its session fixture is blocked locally because PostgreSQL database caura_test is not provisioned. CI should run the repository suite with its database service.


Eldad-Caura:
Applied the verified Critical finding in the amended single commit. REST writes now bind real verified identities, while a legacy verified `main` fails closed with dedicated re-provisioning guidance instead of collapsing or accepting an unauthenticated body identity. The rejection is preflighted before single/bulk idempotency claims and repeated at the inner boundary. Added symmetric single/bulk coverage for default-on rejection, pre-idempotency ordering, and the explicit rollback, plus mutation checks proving the regressions detect the unsafe and late-rejection variants.

Eldad-Caura:
@Codex

Eldad-Caura:
@claude

Eldad-Caura:
@claude

Eldad-Caura:
Addressed the rollout concern in caaa4eab. Normal verified identities remain bound by default. A reserved verified main now follows the existing allow/warn/reject migration policy: allow/warn preserve the unique body-ID escape hatch, warn emits reserved-agent telemetry exactly once per REST request, and reject fails closed before idempotency with REST-specific credential re-provisioning guidance. The explicit bind flag remains the emergency rollback. Added single and bulk coverage for warn, reject ordering, and rollback, plus mutation checks for duplicate observation.

Eldad-Caura:
@claude

Eldad-Caura:
Handled the remaining review findings in f1bce053. Applied: tenant-admin STM bulletin read/clear authority is now explicit in code and symmetric tests; reserved-main rollout telemetry now emits exactly once at the effective service boundary, only for real warn-stage writes (not cached replays, rejects, rollback traffic, or pre-gate failures); and the config, identity docs, tests, and authz ratchet now state plainly that warn mode still permits any non-placeholder body identity for verified main, so those two self-plane gaps remain tracked until policy=reject. Declined the ingest Content-Length fast path: trusting a declared small length would restore the dishonest-header bypass this audit item closes; the authoritative buffer is capped at 3 MB and runs inside the request timeout.

Eldad-Caura:
@claude

Eldad-Caura:
Addressed the final Claude Low finding in amended head `5cbdbc87`:

- install credentials rejected from STM bulletin read/clear now return the existing `AGENT_CREDENTIAL_FORBIDDEN` machine-readable code instead of the misleading trust-level code
- both read and clear regression tests assert the wire error code
- boundary harness: `failures=0`; ruff check/format and focused mypy pass
- mutation check: restoring `AGENT_TRUST_TOO_LOW` makes both new assertions fail

The bounded ingest buffering observation remains an intentional trade-off to enforce the actual 3 MB body limit against chunked or dishonest `Content-Length` requests.

Eldad-Caura:
@erni-a

erni-a (approved):

Issue #1649, opened 2026-09-20
knowledge-update-benchmark: 125k entities with embeddings, zero cross-link matches — likely embedding-space mismatch


Eldad-Caura:
Found while diagnosing #1616. The performance work there is real, but it may be treating a symptom: this tenant's cross-link discovery might be *correct* and finding nothing, because there is genuinely nothing to find.

## The observation

`knowledge-update-benchmark-0d1a47` in staging holds **125,252 entities with a non-null `name_embedding`** — the largest tenant in that environment. Every nightly cross-link run over its 100 candidate memories creates **zero** links, and the phase instrumentation from #1620 confirms the work is real rather than skipped:

```
Created 0 cross-links for 100 candidate memories (tenant knowledge-update-benchmark-0d1a47)
  [candidates=0.1s lateral=143.1s text_verify=0.0s insert=0.0s other=0.0s]
```

`text_verify=0.0s` means the LATERAL handed it almost nothing — the similarity floor rejected effectively everything, not the name-matching filter.

## Why that is implausible

125,252 entities, 100 memories, and not one pair clears the threshold. For a benchmark corpus about knowledge updates, where entity names are extracted *from* those memories, a near-zero match rate is not a plausible distribution. Six other orgs in the same nightly run complete in under 20s and do create links.

## Hypothesis

`entities.name_embedding` and `memories.embedding` are not in the same vector space for this tenant. Candidates:

- the two columns were populated by **different embedding models** (or different model versions) — a backfill run against one and not the other;
- `name_embedding` was computed from something other than the canonical name (raw payload, an id, an empty string), so the vectors are meaningless but non-null;
- a dimensionality or normalisation mismatch that still satisfies the `vector(1024)` type constraint.

The cosine floor would reject essentially all pairs under any of these, which is exactly what is observed.

## Why it matters beyond this tenant

The #1616 work makes the query *fast*. It does not make it *find links*. If the embeddings are mismatched, a fixed query returns zero links in 2s instead of zero links in 143s — an operational win and a product no-op. Worse, the timeout is currently the only signal that anything is wrong; remove it and the tenant fails silently.

It is also the mechanism behind the performance bug: an unsatisfiable threshold is what prevents `LIMIT 10` from ever being reached, which is what makes the index scan exhaust the graph. Same root, two symptoms.

## How to check

Against `memclaw-staging-alloydb`:

```sql
-- 1. Distribution of best-match similarity for a sample of this tenant's memories.
WITH m AS (
[...]

PR #1650, opened 2026-09-20
fix(storage): bound the cross-link ANN scan with hnsw.iterative_scan

Eldad-Caura:
Third attempt at #1616, and the first aimed at the actual mechanism. Supersedes #1635 (reverted in #1639) and #1640 (closed).

## The mechanism — corrected

Every previous attempt, including my own framing on #1616, said this was about a tenant holding *a thin slice* of the entity table. **That is wrong.** Real counts from AlloyDB:

| staging tenant | entities with `name_embedding` |
|---|---|
| **`knowledge-update-benchmark-0d1a47`** | **125,252** |
| `arkady-bench-ececd3` | 97,679 |
| `single-session-user-benchmark-bc9d09` | 77,280 |

The failing tenant is the **largest** in staging.

The real mechanism: the threshold is **unsatisfiable** for it — 125,252 entities with embeddings, and *zero* clear the similarity floor against any of the 100 candidate memories. So `LIMIT 10` is never reached, the scan cannot stop, and it exhausts the index once per candidate. Tenant size was never the variable; **threshold satisfiability** is. That also explains why halving the batch size changed nothing except the number of exhaustive scans.

## The change

One GUC, set for the transaction. The query is untouched.

```python
if await _ann_pool_available():
    await session.execute(select(func.set_config("hnsw.iterative_scan", "relaxed_order", True)))
```

`relaxed_order` keeps the scan walking past `ef_search` until the LIMIT is satisfied, and pgvector's own `hnsw.max_scan_tuples` (default 20k) is what makes that **bounded** — a filter nothing satisfies degrades to an under-filled result instead of a full crawl. That is precisely this tenant's case, and the existing comment in `memory_scored_search` already describes it.

Gated on `_ann_pool_available()`, the probe `memory_scored_search` already uses, so a database below pgvector 0.8 keeps today's behaviour instead of failing on a GUC that does not exist there.

**Version confirmed, not assumed.** Storage runs on **AlloyDB PG17** (clusters `memclaw-staging-alloydb`, `memclaw-production`) — not the Cloud SQL instances earlier notes pointed at. Both report `vector 0.8.1.google-1`, and that string parses to `(0, 8, 1)` through the existing parser, which stops at the non-numeric `google-1`. `alloydb_scann` is **not** installed.

## Why a GUC rather than another rewrite

Both previous attempts reshaped the SQL. Both were measured worse than doing nothing:

| attempt | result |
|---|---|
| #1635 — threshold outside the LIMIT | **282s** vs 143s. Removing the unindexable predicate is what let the planner reach for the global HNSW index. Reverted. |
| #1640 — tenant-scoped `MATERIALIZED` CTE | ~1.02ms/entity ⇒ **~128s for this tenant** (still a timeout), and **~32 min** for prod's `ten2-e5d850` at 1,965,765 entities. Closed. |

The shape was never the problem.

## Tests

Three mutations, each caught by its intended test:

| mutation | caught by |
|---|---|
| GUC never set | `test_discover_pins_iterative_scan_when_pgvector_supports_it` |
| probe gate dropped (set unconditionally) | `test_discover_omits_iterative_scan_below_pgvector_08` |
| `tenant_id` dropped from the LATERAL | `test_discover_does_not_link_across_tenants` |

The first two assert on **the SQL actually sent to the database**, not on source text — both previous attempts pinned query *shape*, and both pinned the wrong shape. Two details that would each have made the assertion vacuous, found by the tests failing first:

- `set_config` passes its name and value as **bind parameters**, so the GUC name never appears in the statement text; the capture records params too.
- `conftest` points the service's session factory at its own `_engine` fixture, not `get_engine()` — listening on the latter records nothing.

## What this does not fix

**It makes the tenant fail fast, not succeed.** 125,252 entities and zero matches is not a plausible distribution; it points at `name_embedding` and memory `embedding` occupying different vector spaces for this tenant. Filed as #1649. A query fix returns zero links in seconds instead of zero links in 143s — an operational win and a product no-op, and worth knowing before this is called done.

## Verification

ruff (4 scopes, separately) and mypy clean. 77 passed across five suites; the 6 failures all reproduce on pristine `origin/main` (5 in `test_ann_pool_behavior.py`, 1 `uq_relations_natural_key`) — baselined in a detached worktree, not assumed.


Eldad-Caura:
Closing this. I measured it, and **the fix has the sign backwards** — it makes the
lateral scan **36.7× slower**, not faster.

### Measurement

Local rig, pgvector 0.8.6, 20,000 entities in one tenant, 1024-dim, HNSW
`(m=16, ef_construction=64)` — the same index migration 033 builds. Memory
embeddings near-orthogonal to the entity cloud, so max similarity is **0.1333**
against a **0.85** threshold: the incident's shape, a threshold nothing
satisfies. `EXPLAIN (ANALYZE, BUFFERS)`, 10 candidate memories:

| `hnsw.iterative_scan` | tuples visited / memory | ms / memory | total |
|---|---|---|---|
| `off` (today) | **40** | 5.0 | 50 ms |
| `relaxed_order` (this PR) | **19,998** | 183.5 | 1,835 ms |

`Rows Removed by Filter: 40` vs `19998`. Both plans are
`Index Scan using ix_entities_name_embedding_hnsw`.

### Why I had it backwards

I wrote that `max_scan_tuples` is "what makes the walk BOUNDED rather than
unbounded". It isn't. With `iterative_scan = off`, the scan already stops at
`ef_search` (40) and returns an under-filled result — that is the *fast* path.
`relaxed_order` is what makes it keep walking to `max_scan_tuples`. For an
unsatisfiable threshold the LIMIT is never reached, so `relaxed_order` pays the
full 20k walk **every time** and still returns zero rows.

So this PR is either a no-op (if production is already seq-scanning) or a 36.7×
regression (if it is index-scanning). Neither is a fix.

### On the review finding

@claude the Medium on `hnsw.max_scan_tuples` is pointing at something real from
the wrong side: `max_scan_tuples` is what *limits the damage this PR causes*, not
what makes the query safe. Its suggested remedy — "pick an explicit bound based
on the 125,252-entity case" — would have raised the bound above 20k and made the
regression proportionally worse.

Two things I verified rather than assumed, both of which hold:
- No `hnsw.*` flag is set on either AlloyDB instance (`databaseFlags` on
  `memclaw-staging-alloydb-primary` and `memclaw-production-primary` contains
  only `password.enforce_complexity`), so `max_scan_tuples` is at pgvector's
  compiled-in default. A cluster-level change there is a Terraform diff, not a
  silent one. A per-database or per-role `SET` would not show up this way and I
  cannot rule it out from outside — `pg_db_role_setting` would show it.
- `max_scan_tuples` is set nowhere in this repo; `memory_scored_search` carries
  the identical assumption in its own comment and has run on it in production.

### What the measurement points at instead

With the HNSW index, 100 candidates cost ~0.5s. The incident is ~143s. So the
143s is very likely **not being served by the index at all**. Forcing the
seq-scan plan in the same rig gives 68.4 ms/memory at 20k entities, which scales
to ~430 ms at 125,252 and ~43s for 100 candidates — the right order of magnitude
for what #1616 measured, on hardware that isn't AlloyDB.

That makes the open question "why isn't `ix_entities_name_embedding_hnsw` being
used in production", not "how do we bound the walk". Tracking in #1616.

Superseded by whatever that turns up. Branch `cross-link-iterative-scan` left in
place for now.

PR #1652, opened 2026-09-20
feat(documents): record which agent wrote a document

arkash20:
**ax-0917-m-14.** `documents` had no author column, and `DocWriteRequest` is `extra="forbid"`, so there was no field to send one in either. A fleet whose agents all write to the same collection could not answer *who wrote this* — the question every other row in this schema can answer, since `memories` has carried `agent_id` from the beginning.

The probe that found this put `owner` inside `data` instead. That is worse than it looks: `data` is replaced wholesale on every upsert, so the attribution survives only as long as each writer remembers to re-send it, and no query can find it without knowing the convention.

### There is history here

SAFE-01 (#1005) used **`agent_id` itself** as its example of junk on this endpoint, with the reasoning *"the documents routes take identity from auth"*.

Half of that was true. Identity did come from auth — and then went nowhere, because the table had no column to put it in. So the plugin was refused a field it was sending for a real reason, and documents carried no author at all. That test now uses `owner`, which is still not a field, and is what the probe resorted to when the body would not take an author.

### Behaviour

```jsonc
POST /api/v1/documents
{"collection": "notes", "doc_id": "d1", "data": {...}, "agent_id": "writer-1"}
→ {"agent_id": "writer-1", ...}        // and readable back through GET
```

- **The credential wins** (`auth.agent_id or body.agent_id`) — the same rule `caller_agent_id` follows on a search: a caller must not write a document under a name that is not its own. A tenant-scoped key has no identity to impose, so there the body is taken at its word.
- **Omitting it stores no author.** Not the tenant's first agent, not `"unknown"` — a fabricated author reads exactly like a real one.
- **An upsert records the agent that wrote *this* version.** The upsert replaces the document; keeping the first writer would attribute someone else's edit to them.
- The field is **present-and-null** on a read, not absent, so a client can tell "no author" from "old server".

### Both storage paths

An **indexed** write (one carrying `data.summary`) goes through `document_upsert_returning_xmax`, not `document_upsert`. Wiring only the first would have dropped the author for exactly the documents that are searchable — so a test asserts both.

### Migration 050

Nullable, no backfill; `ix_documents_tenant_agent` mirrors `ix_memories_tenant_agent`, built `CONCURRENTLY` in an autocommit block matching 049. `ADD COLUMN` of a nullable column with no default is metadata-only on PG 11+, so the column itself needs no such care. `test_single_head` bumped 049 → 050 per its own docstring.

### Verification

- 9 new tests. Against the parent commit **all 9 fail**; all 9 pass here.
- `test_unknown_field_rejection.py`: 21 passed (the documents case rewritten as above).
- `-k "migration or concurrently or large_table or model_index or document"`: 168 passed, 0 failures.
- `ruff check` (incl. `common/` under its own config), `mypy`: unchanged from `main` — 166 core-api, 11 core-storage-api, both pre-existing.
- Broker OpenAPI baseline: unaffected (`--check` clean).

One local-environment note for reviewers: the test DB is built by `Base.metadata.create_all`, which creates missing **tables** but never adds columns — so an existing local `caura_test` needs the column applied by hand or it fails 31 tests with `UndefinedColumnError`. CI builds fresh, so it is not affected.


arkash20:
Pushed a fix for the CI failure — `test_authz_gate_inventory` refused this route, correctly.

Adding a caller-supplied `agent_id` puts the route on the self plane, and the inventory requires such a route to call `enforce_self_agent` or justify itself in `SELF_GATE_ALLOWLIST`. My `auth.agent_id or body.agent_id` did already prevent spoofing — the verified identity won — but it won **silently**, and by the allowlist's own taxonomy that is the weaker of the two behaviours: a caller that names a peer has made a claim, and quietly rewriting it means the caller never learns its attribution was wrong.

So it refuses now. `enforce_self_agent` fires only for a credential that *has* an identity, so a tenant-scoped key may still name any of its agents — the same latitude `POST /memories` gives — and omitting the field always passes. The gate runs on the body value, before resolution; gating the resolved `author` would compare the verified identity against itself and never fire, so a test pins the order.

That also broke six tests in `test_documents_rest_doc_memory`, whose minimal `_Auth` double had no such method. I gave the double the real gate's shape rather than a no-op — a no-op would keep passing if the route ever handed it the resolved author, which is the mistake that makes the gate unfireable in production.

CI green (11/11).

@Eldad-Caura please approve

Eldad-Caura (approved):

arkash20:
Fixed — the failing check was `test_green_means_applied`, and it caught a real defect in migration 050.

A killed `CREATE INDEX CONCURRENTLY` leaves the index in `pg_index` with `indisvalid = false`. `IF NOT EXISTS` then **skips** it — "already exists, skipping" — and the planner refuses to use it. The migration stamps green over an index that is permanently useless and still charges write overhead on every insert, and no re-run repairs it because each one takes the same skip. (035 measured the invalid index at 547x slower on the very query it existed for.)

`drop_invalid_indexes` exists for this and 049 calls it. I copied 049's CONCURRENTLY-inside-`autocommit_block` shape and missed the line above it — which is the half of that pattern easy to copy without.

**Why it was green locally:** the gate lives in `core-storage-api/tests/`, which a plain `pytest tests/` from the repo root never reaches — that suite's conftest fails to import `common` without the repo root on PYTHONPATH, and I let that stop me earlier instead of fixing the invocation.

Verified both directions: removing the line again gives 1 failed / 30 passed; with it, 31 passed. Full `core-storage-api` suite: **569 passed**. Root-suite interacting tests (m-14, m-12, m-13, unknown-field rejection, doc-memory minting, authz gate inventory): **107 passed**.

arkash20:
CI fully green now — 0 failing across 14 checks.

Two failures, only the first of which was mine:

1. **Migration 050 was genuinely wrong** — missing `drop_invalid_indexes` before the concurrent build. Fixed in `98857dab`; details in the comment above.
2. **Docker Hub flake** on the `Broker OpenAPI has no breaking change vs main` step: `connection reset by peer` pulling the pinned `tufin/oasdiff` image, exit 125. I ran the identical gate locally with the same digest — *"No changes detected"*, exit 0 — and re-triggered the job, which passed.

The force-push dismissed the prior approval, so this needs one more look.

@Eldad-Caura please approve

Eldad-Caura (approved):

PR #1654, opened 2026-09-20
Install namespaced skills with safe frontmatter

Eldad-Caura:
## Summary
- recognize the server-issued forge/ and agent/ skill catalog namespaces while keeping filesystem paths on validated bare slugs
- fail closed when multiple catalog IDs normalize to the same disk slug
- quote synthesized name and description values as valid YAML scalars

Tracks Erni audit items oss-0902-m-74 and oss-0902-m-75.

## Validation
- plugin test suite: 473 passed
- TypeScript compilation via npm test
- mutation proofs: namespace install, duplicate-slug fail-closed behavior, and special-character frontmatter tests each fail without their fix
- git diff --check
- /simplify correctness, logic, and test reviews: clean after collision handling


Eldad-Caura:
Addressed the verified Low finding in the existing commit. Exact duplicate catalog rows are now idempotent only when their fully materialized SKILL.md bytes match; same-ID rows with conflicting content still fail closed. Added regressions for normalized-identical metadata and conflicting content, with mutation proof. Full plugin suite: 475/475 passing; TypeScript compilation and diff check pass; final simplify review is clean.

Eldad-Caura:
@claude

Eldad-Caura:
@erni-a

erni-a (approved):

Eldad-Caura:
Ready to merge: fresh Claude review is clean, Erni approved, all required checks are green, and the branch remains one signed commit.

PR #1655, opened 2026-09-20
fix(storage): report schema drift at startup

Eldad-Caura:
## Summary
- report every invalid non-system PostgreSQL index at ERROR on writer startup, with the concurrent rebuild repair path
- warn when cosine_distance still has procost 1 because migration 044 could not apply
- keep both catalog checks non-fatal and skip them entirely on reader replicas

## Verification
- demonstrated all six new behavior tests fail before the production fix
- 552 core-storage-api tests passed
- Ruff check passed at all 8 current CI scopes
- Ruff format --check passed at all 8 current CI scopes
- mypy passed at all 6 current CI scopes

Refs #1616.


Eldad-Caura:
Addressed the review finding: the procost probe now resolves the exact cosine_distance(vector, vector) signature targeted by migration 044, so the halfvec and sparsevec overloads cannot affect the result. The integration test now verifies those other overloads may remain at the default cost without causing a false warning.

Revalidated: focused schema-drift tests 6/6, full core-storage-api suite 552/552, storage Ruff check and format check, and storage mypy.

Eldad-Caura:
@claude

Eldad-Caura:
@erni-a

erni-a (approved):

